PR #1255, opened 2019-01-31
gdal_python.i - adding overviewLevel option to WarpOptions

imiara:
Hi,
This is my first pull request to gdal :)

I was trying to use -ovr level option for gdalwarp in Python. 
Although it is available in the command line app I couldn't find it in the python API and I think it is missing. This pull request is to add this functionality to the Python API.

I added the support in WarpOptions in gdal.py and it works for me (I've tested only the 'level' option. I added support for AUTO and AUTO-n but didn't test them, so I don't know if I set these correctly).

In this pull request I applied the same change to gdal_python.i as I understand gdal.py is auto generated from this file. 
I didn't test the swig file because I don't know how.

### DIFF
--- a/gdal/swig/include/python/gdal_python.i
+++ b/gdal/swig/include/python/gdal_python.i
@@ -1220,6 +1220,7 @@ def WarpOptions(options=None, format=None,
          cutlineLayer=None, cutlineWhere=None, cutlineSQL=None, cutlineBlend=None, cropToCutline = False,
          copyMetadata = True, metadataConflictValue=None,
          setColorInterpretation = False,
+         overviewLevel = 'AUTO',
          callback=None, callback_data=None):
     """ Create a WarpOptions() object that can be passed to gdal.Warp()
         Keyword arguments are :
@@ -1259,6 +1260,7 @@ def WarpOptions(options=None, format=None,
           copyMetadata --- whether to copy source metadata
           metadataConflictValue --- metadata data conflict value
           setColorInterpretation --- whether to force color interpretation of input bands to output bands
+          overviewLevel --- To specify which overview level of source files must be used
           callback --- callback method
           callback_data --- user data for callback
     """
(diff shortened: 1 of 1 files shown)


gdal/swig/include/python/gdal_python.i
```diff
@@ -1356,6 +1358,19 @@ def WarpOptions(options=None, format=None,
             new_options += ['-cvmd', str(metadataConflictValue)]
         if setColorInterpretation:
             new_options += ['-setci']
+        if overviewLevel is not None:
+            if _is_str_or_unicode(overviewLevel):
+                overviewLevelStr = overviewLevel
+            else:
+                overviewLevelStr = ''
+                if overviewLevel is ...:
```
rouault: What is this Python syntax (pardon my ignorance) ? Doesn't seem to work with python 2.7

imiara: ellipsis (...) is a builtin constant:
https://docs.python.org/3/library/constants.html
I used it here to express a default value. 
I'm not sure this convention is widespread so maybe We'll use an empty string or NaN instead (or just remove this if statement and use only "AUTO" as is?

imiara: I've forced push a fix:
Changed the default to be 'AUTO' instead of None.
removed the use of ellipsis

rouault:
@imiara This PR is stalled. Do you intend to address my above comments https://github.com/OSGeo/gdal/pull/1255#pullrequestreview-198622832 ?

imiara:
Sorry, I've missed your comments.
I haven't worked with SWIG before, could you please give me some pointers?

rouault:
> I haven't worked with SWIG before, could you please give me some pointers?

- download https://sourceforge.net/projects/swig/files/swig/swig-3.0.8/swig-3.0.8.tar.gz/download if you haven't swig 3.0.8, configure, make, make install
- put in the PATH the location of the swig binary
- in gdal/swig/python, make generate && make

imiara:
Thanks, I'll look into it next week.

rouault:
Integrated with fixes per ab4def2edd041c62968f18985fbb945e247e32ba. Untested though... Tests welcome in autotest/utilitiest/test_gdalwarp_lib.py

imiara:
Hi @rouault ,
Thanks for integrating the pull request!
Unfortunately I didn't find the time yet to write the tests as you advised, I hope I'll get into it.
Although I read that this pull request was merged into master, I can't find it in the master branch.
Please advise.

rouault:
> I can't find it in the master branch

Did you 'git pull' from master branch ? This is in master per https://github.com/OSGeo/gdal/commit/ab4def2edd041c62968f18985fbb945e247e32ba

imiara:
> > I can't find it in the master branch
> 
> Did you 'git pull' from master branch ? This is in master per [ab4def2](https://github.com/OSGeo/gdal/commit/ab4def2edd041c62968f18985fbb945e247e32ba)

I was sure I did, but now I cloned again and found it, thanks!

PR #1250, opened 2019-01-30
 Allow internal drivers with no dependencies to be disabled at compile time

tbonfort:
This PR modifies the autotools build system to allow a user to selectively enable or disable some drivers that were previously always enabled and compiled.

Allows for smaller builds e.g when creating docker images

Example configure invocation to build a gdal image use only for accessing COGs (supposing the result is written to a MEM dataset)

```
./configure \
        --disable-all-optional-drivers \
	--with-threads     \
	--with-png=internal \
	--with-jpeg=internal \
	--with-libtiff=internal \
	--with-libz=internal \
	--with-curl       \
	--without-bsb     \
	--without-cfitsio     \
	--without-cryptopp     \
	--without-ecw     \
	--without-expat     \
	--without-fme     \
	--without-freexl     \
	--without-gif     \
	--without-gnm     \
	--without-grass     \
	--without-grib     \
	--without-hdf4     \
	--without-hdf5     \
	--without-idb     \
	--without-ingres     \
	--without-jasper     \
	--without-jp2mrsid         \
	--without-kakadu     \
	--without-libgrass     \
	--without-libkml     \
	--without-mrf     \
	--without-mrsid     \
	--without-mysql     \
	--without-netcdf     \
	--without-odbc     \
	--without-ogdi     \
	--without-openjpeg     \
	--without-pcidsk     \
	--without-pcraster     \
	--without-pcre     \
	--without-perl     \
	--without-pg        \
	--without-python     \
	--without-qhull     \
	--without-sde     \
	--without-sqlite3     \
	--without-webp     \
	--without-xerces     \
	--without-xml2 \
	--without-lerc \
	--without-geos

```

OGR drivers that cannot be disabled and will always be built: generic geojson kml mem mitab vrt

GDAL drivers that cannot be disabled and will always be built: derived gtiff hfa mem raw vrt

PR #1266, opened 2019-02-02
move rawdataset base class out of formats

tbonfort:
In line with #1250, move the rawdataset base class into gcore to enable raw drivers to be disabled


tbonfort:
Sorry, I misconfigured my git remotes and pushed this branch to this repo instead of my fork. I technically have commit rights here but prefer not to use them, @rouault can you delete this branch please? Sorry again for the noise :(

PR #1268, opened 2019-02-03
Move rawdataset base class out of raw driver

tbonfort:
The rawdataset base class is used by core and various drivers, and thus moved to gcore. Opens up the possibility of disabling raw drivers in #1250


gdal/frmts/pds/isis2dataset.cpp
```diff
@@ -71,7 +71,7 @@ class ISIS2Dataset : public RawDataset
 
     CPLString   oTempResult;
 
-    void        CleanString( CPLString &osInput );
+    static void CleanString( CPLString &osInput );
```
tbonfort: Change needed by cppcheck on ubuntu 1604. Not sure why this commit changes the result of the code analysis...

PR #1274, opened 2019-02-06
make bsb, grib and mrf follow disabling standard of #1250

tbonfort:
Rename ./configure arguments --without-bsb, --without-grib and --without-mrf to --disable-driver-bsb --disable-driver-grib and --disable-driver-mrf to be inline with #1250

### DIFF
--- a/gdal/configure.ac
+++ b/gdal/configure.ac
@@ -1353,7 +1353,7 @@ OGRFORMATS_ENABLED=
 OGRFORMATS_ENABLED_CFLAGS=
 OGRFORMATS_DISABLED=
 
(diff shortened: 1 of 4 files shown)


rouault:
The mrf issue is due to the fact that, when JPEG is enabled, we have to run "make" in frmts/jpeg to generate the frmts/jpeg/libjpeg12 files, before make can be run in frmts/mrf (on a only generated tree, you need to run 'make clean' in frmts/jpeg to reproduce).
Looking at frmts/nitf/GNUmakefile that has the same dependency to libjpeg12, there's an explicit rule in it to build libjpeg12. frmts/mrf/GNUmakefile likely needs the same

Can you squash the commits once done ?

tbonfort:
@rouault this PR is rebased and ready for merge as soon as the CI finishes

gdal/configure.ac
```diff
@@ -1463,7 +1463,23 @@ if test "$LIBZ_SETTING" != "no" ; then
   fi 
 fi
 
+dnl ---------------------------------------------------------------------------
+dnl LERC and MRF are coupled
```
rouault: Hum, are they ?
Since GDAL 2.4.0, the GTiff driver can build a LERC TIFF plugin, and this is not related at all with MRF
And normally (didn't try recently though), the MRF driver should be able to build without LERC

tbonfort: hmm, my bad, I was under the idea that lerc required mrf, but looking at the code that is clearly not the case, I'll revert the lerc part of this PR. FYI, mrf builds fine without lerc, and that was accounted for in the proposed change

tbonfort:
@rouault : ready for merge

PR #1280, opened 2019-02-07
GeoJSON: speed-up random reading with GetFeature() by storing a map FID->(start,size) to retrieve performance similar to GDAL 2.2 or before (fixes https://issues.qgis.org/issues/21085)

rouault:


rouault:
Merged

PR #1289, opened 2019-02-12
VSI plugins

tbonfort:
This PR allows to dynamically register a VSI handler at runtime by passing a select number of callback functions. Our principle use-case is to be able to bypass the vsicurl handler when accessing remote datasets, to provide observability, custom error handling, retrying, logging, fine-grained concurrency control, fine-grained caching, and better stability under heavily multi-threaded loads.

### DIFF
--- a/gdal/port/GNUmakefile
+++ b/gdal/port/GNUmakefile
@@ -33,7 +33,7 @@ OBJ =	cpl_conv.o cpl_error.o cpl_string.o cplgetsymbol.o cplstringlist.o \
 	cpl_google_oauth2.o cpl_progress.o cpl_virtualmem.o cpl_worker_thread_pool.o \
 	cpl_vsil_crypt.o cpl_sha1.o cpl_sha256.o cpl_aws.o cpl_vsi_error.o cpl_cpu_features.o \
 	cpl_google_cloud.o cpl_azure.o cpl_alibaba_oss.o cpl_json_streaming_parser.o \
-	cpl_json.o cpl_md5.o cpl_swift.o \
+	cpl_json.o cpl_md5.o cpl_swift.o cpl_vsil_plugin.o \
 	cpl_vsil_hdfs.o cpl_userfaultfd.o
 
 ifeq ($(ODBC_SETTING),yes)
--- a/gdal/port/cpl_vsi.h
+++ b/gdal/port/cpl_vsi.h
@@ -406,53 +406,148 @@ GByte CPL_DLL *VSIGetMemFileBuffer( const char *pszFilename,
 typedef size_t (*VSIWriteFunction)(const void* ptr, size_t size, size_t nmemb, FILE* stream);
 void CPL_DLL VSIStdoutSetRedirection( VSIWriteFunction pFct, FILE* stream );
 
+/**
+ * Return information about a handle. Optional (driver dependant) 
+ * @since GDAL 2.5
+ */
 typedef int            (*VSIFilesystemPluginStatCallback)          ( void *pUserData, const char *pszFilename, VSIStatBufL *pStatBuf, int nFlags );
+/** 
+ * Remove handle by name. Optional 
+ * @since GDAL 2.5
+ */
 typedef int            (*VSIFilesystemPluginUnlinkCallback)        ( void *pUserData, const char *pszFilename );
+/** 
+ * Rename handle. Optional 
+ * @since GDAL 2.5
+ */
 typedef int            (*VSIFilesystemPluginRenameCallback)        ( void *pUserData, const char *oldpath, const char *newpath );
+/**
+ * Create Directory. Optional
+ * @since GDAL 2.5
+ */
 typedef int            (*VSIFilesystemPluginMkdirCallback)         ( void *pUserData, const char *pszDirname, long nMode );
+/**
+ *  Delete Directory. Optional 
+ * @since GDAL 2.5
+ */
 typedef int            (*VSIFilesystemPluginRmdirCallback)         ( void *pUserData, const char *pszDirname );
+/** 
+ * List directory content. Optional 
+ * @since GDAL 2.5
+ */
 typedef char**         (*VSIFilesystemPluginReadDirCallback)       ( void *pUserData, const char *pszDirname, int nMaxFiles );
+/** 
+ * Open a handle. Mandatory. Returns an opaque pointer that will be used in subsequent file I/O calls.
+ * Should return null and/or set errno if the handle does not exist or the access mode is incorrect.
+ * @since GDAL 2.5
+ */
 typedef void*          (*VSIFilesystemPluginOpenCallback)          ( void *pUserData, const char *pszFilename, const char *pszAccess );
+/** 
+ * Return current position in handle. Mandatory 
+ * @since GDAL 2.5
+ */
 typedef vsi_l_offset   (*VSIFilesystemPluginTellCallback)          ( void *pFile );
+/** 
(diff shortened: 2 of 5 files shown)


gdal/port/cpl_vsi.h
```diff
@@ -406,6 +406,55 @@ GByte CPL_DLL *VSIGetMemFileBuffer( const char *pszFilename,
...
+    VSIFilesystemPluginReadMultiRangeCallback   read_multi_range;
+    VSIFilesystemPluginGetRangeStatusCallback   get_range_status;
+    VSIFilesystemPluginEofCallback              eof;
+    VSIFilesystemPluginWriteCallback            write;
+    VSIFilesystemPluginFlushCallback            flush;
+    VSIFilesystemPluginTruncateCallback         truncate;
+    VSIFilesystemPluginCloseCallback            close;
+} VSIFilesystemPluginCallbacksStruct;
+
+VSIFilesystemPluginCallbacksStruct* VSIAllocFilesystemPluginCallbacksStruct( void );
```
rouault: this function and the below one should be CPL_DLL exported too.

Can you add some minimal Doxygen doc (in the .cpp) for those 3 functions ( with `@since GDAL 2.5`) ?

gdal/port/GNUmakefile
```diff
@@ -33,7 +33,7 @@ OBJ =	cpl_conv.o cpl_error.o cpl_string.o cplgetsymbol.o cplstringlist.o \
 	cpl_google_oauth2.o cpl_progress.o cpl_virtualmem.o cpl_worker_thread_pool.o \
 	cpl_vsil_crypt.o cpl_sha1.o cpl_sha256.o cpl_aws.o cpl_vsi_error.o cpl_cpu_features.o \
 	cpl_google_cloud.o cpl_azure.o cpl_alibaba_oss.o cpl_json_streaming_parser.o \
-	cpl_json.o cpl_md5.o cpl_swift.o \
+	cpl_json.o cpl_md5.o cpl_swift.o cpl_vsil_plugin.o \
```
rouault: should be added to makefile.vc as well

gdal/port/cpl_vsil_plugin.cpp
```diff
@@ -0,0 +1,321 @@
...
+    if( m_cb->mkdir == nullptr || !IsValidFilename(pszDirname) )
+        return -1;
+    return m_cb->mkdir(m_cb->pUserData, GetCallbackFilename(pszDirname), nMode);
+}
+int VSIPluginFilesystemHandler::Rmdir(const char *pszDirname) {
+    if( m_cb->rmdir == nullptr || !IsValidFilename(pszDirname) )
+        return -1;
+    return m_cb->rmdir(m_cb->pUserData, GetCallbackFilename(pszDirname));
+}
+}
```
rouault: `} // namespace cpl` for clarity

gdal/port/cpl_vsil_plugin.cpp
```diff
@@ -0,0 +1,321 @@
...
+/*                                Seek()                                */
+/************************************************************************/
+
+int VSIPluginHandle::Seek( vsi_l_offset nOffset, int nWhence )
+{
+    return poFS->Seek(cbData,nOffset,nWhence);
+}
+
+
+/************************************************************************/
```
rouault: those decoration comment lines are inconsistently used. Either add them everywhere or remove them all

rcoup:
@tbonfort possible to share some of that work so we can improve what is in GDAL core?

tbonfort:
@rcoup I'm not able to share more code, but there really is nothing special to it. I should have explained a bit more that the use case is to call back into scripted language (python, go in our case), so you can easily add your own logging semantics, error handling, authentication, etc... i.e. nothing really applicable to the core gdal implementation.

Two take-backs that could be useful for vsicurl et. al. when dealing with COGs:
- use a two-tiered block cache: one for the first block (supposing the blocksize is big enough to contain the full COG headers), and one for the remaining blocks. This can help lru evicts to not affect the header blocks, and the two caches can be sized based on expected usage.
- implement a global lock on a block when downloading it, i.e. if a block is not present in the block-cache, ensure that there's a single GET to fetch it instead of having each concurrent thread do the same GET multiple times

tbonfort:
cc @sgillies , could this spark some interest for rasterio ?

sgillies:
@tbonfort I'm interested!

rouault:
If anyone has concerns on the interface speak up. I'll merge this soon otherwise

tbonfort:
I'm wondering if it would make sense to add a way for the callbacks to explicitely return an error condition, without relying on it to correctly set errno and/or call CPLError.

Something in the lines of:
```c
typedef struct {
  int code;
  char *message
} CallbackError;

typedef size_t (*VSIFilesystemPluginReadCallback) ( void *pFile,
               void *pBuffer, size_t nSize, size_t nCount , error *CallbackError);
```

rouault:
What is the advantage of this compared to the callback implementation calling VSIError / CPLError ?

tbonfort:
I was thinking it could be easier for the e.g. python code to set an error message, but I'm fine if that is handled by the C callback function itself.

gdal/port/cpl_vsil_plugin.cpp
```diff
@@ -0,0 +1,356 @@
...
+}
+
+VSIRangeStatus VSIPluginFilesystemHandler::GetRangeStatus( void *pFile, vsi_l_offset nOffset, vsi_l_offset nLength) {
+    if (m_cb->get_range_status != nullptr) {
+        return m_cb->get_range_status(pFile,nOffset,nLength);
+    }
+    return VSI_RANGE_STATUS_UNKNOWN;
+}
+
+int VSIPluginFilesystemHandler::ReadMultiRange( void *pFile, int nRanges, void ** ppData, const vsi_l_offset* panOffsets, const size_t* panSizes ) {
```
tbonfort: I see that the behavior of joining consecutive ranges can be disabled with a config option in the vsicurl implementation, but fail to see when that could be desirable. Maybe this part could be moved up a bit higher in the vsi call tree?

rouault:
> I see that the behavior of joining consecutive ranges can be disabled with a config option in the vsicurl implementation, but fail to see when that could be desirable.

Yes that's what I mentionned. In some rare circumstances it might help not to merge consecutive ranges, but this is not a good default

> Maybe this part could be moved up a bit higher in the vsi call tree?

There's not so many places where that could be done, except in the VSIFReadMultiRangeL() function itself, that will be the one generally called by drivers. The VSICurlHandle::ReadMultiRange() implementation probably saves a buffer allocation/copy compared to your generic implementation though.

tbonfort:
Lets move the discussion about multi range requests to #1295

rouault:
@tbonfort Are you happy with the PR as it ?

tbonfort:
@rouault ready for merge, can I let you squash this through the github interface to avoid re-running the CI?

rouault:
Done

mmomtchev:
@rouault, @tbonfort, is there any documentation for this or an example plugin available somewhere?

tbonfort:
@mmomtchev there are no public implementations that I know of, and the only documentation available is the doxygen one. There is however a similar implementation that does not use this plugin but replicates its intent at https://github.com/airbusgeo/godal/blob/ee62c71eebf81d25ef07983ada2e4ba6ea7baaf7/godal.cpp#L1102 and is used to hand off vsi read-only access to a go function.

sgillies:
@mmomtchev @tbonfort there's a recent example in https://github.com/mapbox/rasterio/pull/2141.

PR #1290, opened 2019-02-13
Correct read limit based on differet GRIB version

caxton:
- GRIB version 1 SECTION 0 is 8 bytes (https://www.nco.ncep.noaa.gov/pmb/docs/on388/section0.html)
- GRIB version 2 SECTION 0 is 16 bytes (https://www.nco.ncep.noaa.gov/pmb/docs/grib2/grib2_doc/grib2_sect0.shtml)

## What does this PR do?
There is a read limit defined here(expecting the packet header to be max. 8 byte):
https://github.com/OSGeo/gdal/blob/master/gdal/frmts/grib/degrib/g2clib/seekgb.c#L56
which is fine for version 1 packets(i have no idea what if "packet" is a good name), their layout seems to be MMMM LLL V (Magic, Length, Version)
but version 2 packets are longer (16 bytes), this results in reading out of bounds here:
https://github.com/OSGeo/gdal/blob/master/gdal/frmts/grib/degrib/g2clib/seekgb.c#L66
(it reads 4_8 bits at offset 12_8 bits)

So the read limit should be set by different version.

### DIFF
--- a/gdal/frmts/grib/degrib/g2clib/seekgb.c
+++ b/gdal/frmts/grib/degrib/g2clib/seekgb.c
@@ -53,7 +53,7 @@ void seekgb(FILE *lugb,g2int iseek,g2int mseek,g2int *lskip,g2int *lgrib)
 
         /* ret= */ fseek(lugb,ipos,SEEK_SET);
         nread=(int)fread(cbuf,sizeof(unsigned char),mseek,lugb);
-        lim=nread-8;
+        lim = nread - (vers == 2 ? 16 : 8);
 
 //  LOOK FOR 'GRIB...' IN PARTIAL SECTION
 


rouault:
Hum, this file is actually code not used by GDAL... I'm a bit intrigued on why you needed this fix

caxton:
> Hum, this file is actually code not used by GDAL... I'm a bit intrigued on why you needed this fix

Hi, we are using this library to parse GRIB message. This is original upstream so we would like to provide bug fix as well. But this PR is buggy, I would create another one instead.

rouault:
See frmts/grib/degrib/README.TXT for the true upstream. I'll take fixes however

Issue #1295, opened 2019-02-15
Optimize multi-range read requests


tbonfort:
This is a placeholder issue for moving the discussion about multi range requests out of #1289

From my understanding, this is the current situation. Please correct if I am wrong:
- For the time being, gtiff is the only driver to support multi-range read request
- When the geotiff driver constructs the offsets and sizes of the different ranges, it does not try to merge consecutive ranges to reduce the total number of requested ranges
- From a http vsi backend point of view, there is no real world use-case where it would be more efficient to emit two GET requests for consecutive ranges rather than merging them into a single one:
  - there will be more data to transmit down the wire (accounting for http headers)
  - there will be more GET requests emitted, which are usually billed by the unit by providers
  - there is no reason the parallel requests will finish faster than the merged one given they share the same network pipe. Theoretically this might be false if the provider purposefully throttles the download speed after the first N bytes, and both range requests are less than N bytes, and the merged consecutive request is larger than N
- By default, the vsicurl driver merges consecutive ranges, but this can be disabled with a config option. This ends up with curl allocating memory for the merged buffers during it's write callback.
- The plugin approach proposed in #1289 unconditionally merges consecutive ranges, and same as with curl with the slight cost of having to allocate buffers for the merged requests.
- Any vsi driver that wishes to implement multi-range requests will probably need to also implement consecutive range merging
- In #1289 @rouault states
  >  In some rare circumstances it might help not to merge consecutive ranges, but this is not a good default

  Could you give more details about these circumstances please?

What could be optimized? :
- Make gtiff (and other future drivers wishing to implement multi range reads) emit merged consecutive requests
- In VSIFReadMultiRangeL change the semantics of "merging consecutive ranges" to "merge close-enough ranges" with "close-enough" being something that could be tweaked by config options (This would imply the overhead of memory allocations). Eg:
  - merge ranges if they are less than N bytes apart from each other
  - merge ranges together if the merged range is less than N bytes
  - merge ranges together if the overhead of the merged size is less than P% of the sum of the two original sizes

  Note that at the cost of some memory allocation, we could remove the constraint of gtiff having to merge consecutive ranges by applying the strategy "merge ranges if they are less than 0 bytes apart from each other"
- (Optional, to reduce duplicate memory allocations) In vsicurl, stop using code to specifically handle range merging, and have the curl write callbacks write directly to the passed-in buffers rather than allocating them itself

rouault:
[...]

PR #1297, opened 2019-02-18
make gtiff emit merged consecutive multi-range reads

tbonfort:
addressing #1295

### DIFF
--- a/gdal/frmts/gtiff/geotiff.cpp
+++ b/gdal/frmts/gtiff/geotiff.cpp
@@ -4102,68 +4102,41 @@ void* GTiffRasterBand::CacheMultiRange( int nXOff, int nYOff,
                 std::vector<vsi_l_offset> anOffsets;
                 std::vector<size_t> anSizes;
                 std::vector<void*> apData;
+                anOffsets.push_back(aOffsetSize[0].first);
+                apData.push_back(static_cast<GByte *>(pBufferedData));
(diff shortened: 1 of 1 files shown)


gdal/frmts/gtiff/geotiff.cpp
```diff
@@ -4110,13 +4110,53 @@ void* GTiffRasterBand::CacheMultiRange( int nXOff, int nYOff,
                     apData.push_back(static_cast<GByte*>(pBufferedData) + nAccOffset);
                     nAccOffset += aOffsetSize[i].second;
                 }
+                int nMultiRet;
                 VSILFILE* fp = VSI_TIFFGetVSILFile(th);
-                if( VSIFReadMultiRangeL(
+
+                if( aOffsetSize.size() > 1 )
```
rouault: cannot this just be done in the loop a few line above ?

tbonfort: no, you need the first loop to construct the un-merged offsets/sizes, as those are the ones that get cached.

rouault: from a quick look at tifvsi.cpp, I don't see why VSI_TIFFSetCachedRanges() couldn't be called with the merged ranges

tbonfort: you are right, I've merged the two loops into one

PR #1311, opened 2019-02-22
autotest: introduce pytest-env plugin

miurahr:
## What does this PR do?

pytest-env plugin enables feature to set environment
variable in pytest.ini configuration file.
We sometimes necessary to set GDAL_DATA and other
test configuration variables for customized environment,
and the plugin help it.

Signed-off-by: Hiroshi Miura <miurahr@linux.com>

N.A.

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments


autotest/pytest.ini
```diff
@@ -1,6 +1,14 @@
...
 python_files = *.py
 testpaths = ogr gcore gdrivers osr alg gnm utilities pyscripts
+log_file = autotest.log
+log_file_level = INFO
+env =
+  GDAL_DOWNLOAD_TEST_DATA=NO
+  GDAL_RUN_SLOW_TESTS=NO
+  LD_LIBRARY_PATH=../gdal
+  GDAL_DATA=../gdal/data
+#  GDAL_DRIVER_PATH=/usr/lib/gdal/gdalplugins
```
miurahr: There would be a discussion what is default here or just add it for help as commented out.

rouault:
Is it needed ? There is the scripts/setdevenv.sh that does all the necessary env variable setting.
CC @craigds

miurahr:
With this extension, developer can run pytest with environment values without using shell script and power shell wrapper.

autotest/pytest.ini
miurahr: These are examples so commented out.

Issue #1314, opened 2019-02-22
gdal_retile.py ignores NoData value on input


DanielFEvans:
## Expected behavior and actual behavior.

When I pass gdal_retile.py input files with NoData values set, I expect the output to also have NoData.

While the actual values are passed through without modification, the resulting tiles do not have NoData set.

## Steps to reproduce the problem.
Using an input file with nodata=-3.402823e+38. I run the following command:
`gdal_retile.py -targetDir out example.tif`

The input gives "no data" when queried at top right in QGIS:
![goodtile](https://user-images.githubusercontent.com/41120183/53247746-12d90c80-36ac-11e9-93b8-3cc187b532dc.PNG)

However, the nodata has been forgotten in the output, and QGIS (and any other tool) does not recognise that the top right contains no data:
![badtile](https://user-images.githubusercontent.com/41120183/53247810-313f0800-36ac-11e9-9008-70b362727831.PNG)

## Operating system

Windows 7 64 bit and RHEL 3.10.0 64 bit

## GDAL version and provenance

GDAL 2.2.0, internal company version with some modifications. Also tested with `gdal_retile.py` bundled with GDAL 2.4.0, no change

## Notes

I have come across this relatively old email on gdal-dev which contains a fix. Is there any reason why this fix was not adopted? I suspect it simply got missed, and I'd be happy to submit a PR including it.

http://osgeo-org.1560.x6.nabble.com/gdal-dev-gdal-retile-py-does-not-use-a-nodata-value-from-GeoTIFF-td5099823.html

rouault:
> I have come across this relatively old email on gdal-dev which contains a fix. Is there any reason why this fix was not adopted?

Was before we got on github, so was rather "costly" to take responsibility of a patch. A PR with a new test in autotest/pyscripts/test_gdal_retile.py is welcome

rouault:
Fixed per https://github.com/OSGeo/gdal/pull/1315
Backported to release/2.4 per 83ccdc0d761c2c3bdb40f7c726958b7cb43944cb

PR #1315, opened 2019-02-22
gdal_retile.py: respect nodata values on inputs

DanielFEvans:
Issue #1314 - Modify `gdal_retile.py` to copy the NoData value from the input dataset to the output dataset.

## Tasklist

 - [x] Add test case(s)
 - [x] Review
 - [x] Adjust for comments

### DIFF
--- a/autotest/pyscripts/test_gdal_retile.py
+++ b/autotest/pyscripts/test_gdal_retile.py
@@ -31,7 +31,6 @@
 import shutil
 import os
 
-
 from osgeo import gdal
 from osgeo import osr
 import test_py_scripts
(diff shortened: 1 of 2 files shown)


autotest/pyscripts/test_gdal_retile.py
```diff
@@ -31,6 +31,7 @@
 import shutil
 import os
 
+import numpy as np
```
rouault: Although strongly suggested, not all GDAL builds have numpy available.

You can instead add the following snippet in your new test method
```
    try:
        from osgeo import gdalnumeric
        gdalnumeric.zeros
        import numpy as np
    except (ImportError, AttributeError):
        pytest.skip()
```

autotest/pyscripts/test_gdal_retile.py
```diff
@@ -241,7 +242,57 @@ def test_gdal_retile_4():
...
+    script_path = test_py_scripts.get_py_script('gdal_retile')
+    if script_path is None:
+        pytest.skip()
+
+    drv = gdal.GetDriverByName('GTiff')
+    srs = osr.SpatialReference()
+    srs.SetWellKnownGeogCS('WGS84')
+    wkt = srs.ExportToWkt()
+
+    ds = drv.Create('tmp/in5.tif', 2, 2, 1, gdal.GDT_Float32)
```
rouault: this file should be removed once done with it

autotest/pyscripts/test_gdal_retile.py
```diff
@@ -278,7 +335,5 @@ def test_gdal_retile_cleanup():
                 pass
 
     shutil.rmtree('tmp/outretile4')
-
-
-
+    shutil.rmtree('tmp/outretile5')
```
rouault: should be conditionalized by `if os.path.exists('tmp/outretile5')` (or something similar) to avoid an exception when the test is skipped

autotest/pyscripts/test_gdal_retile.py
```diff
@@ -241,7 +240,64 @@ def test_gdal_retile_4():
...
+    assert raster_band.GetNoDataValue() == nodata_value, \
+        ('Wrong nodata value.\nExpected %f, Got: %f' % (nodata_value, raster_band.GetNoDataValue()))
+
+    min_val, max_val = raster_band.ComputeRasterMinMax()
+    assert max_val, \
+        ('Wrong maximum value.\nExpected 2.0, Got: %f' % max_val)
+
+    assert min_val == -1.0, \
+        ('Wrong minimum value.\nExpected -1.0, Got: %f' % min_val)
+
```
rouault: You may need to add an explicit ds = None to force dataset closing to fix the `WindowsError: [Error 32] The process cannot access the file because it is being used by another process: 'tmp/outretile5\\in5_1_1.tif'` of 
https://ci.appveyor.com/project/OSGeo/gdal/builds/22624606/job/6jfgqtwy18yvw4nu

DanielFEvans: Thanks for keeping on top of this. Is there a way to receive some sort of notification when the builds fail, so I don't forget about the pull request until your next comment pops up in my inbox?

rouault: I'm not aware of one. You have to monitor regularly the progress of the builds

PR #1339, opened 2019-03-04
gdaldem hillshade: add -igor option (fixes #1330)

### LINKED ISSUE #1330: Add new hillshading algorithm "Igor" to gdaldem
Igor's hillshading algorithm from Maperitive has a great property that [_renders a softer hillshade which tries to minimize effects on other map features_](http://maperitive.net/docs/Commands/GenerateReliefImageIgor.html). For example it doesn't alter (darken) flatlands.

For live usecase you can see for example:

* https://www.freemap.sk/?map=14/48.881086/21.396947&layers=X
* https://www.freemap.sk/?map=13/48.603404/20.445471&layers=T

Algorithm is shared at https://gist.github.com/breki/e69be079d43d449ad823d0ec50e42929 and https://gist.github.com/breki/0db16a5fcc2e0dd2022fb99eacf7b50c by Igor himself.

PS: For the reference see https://twitter.com/martinzdila/status/1100484997616750592

zdila:
## What does this PR do?

fixes #1330 (Add new hillshading algorithm "Igor" to gdaldem)

### DIFF
--- a/gdal/swig/python/osgeo/gdal.py
+++ b/gdal/swig/python/osgeo/gdal.py
@@ -761,7 +761,7 @@ def VectorTranslate(destNameOrDestDS, srcDS, **kwargs):
 def DEMProcessingOptions(options=None, colorFilename=None, format=None,
               creationOptions=None, computeEdges=False, alg='Horn', band=1,
               zFactor=None, scale=None, azimuth=None, altitude=None,
-              combined=False, multiDirectional=False,
+              combined=False, multiDirectional=False, igor=False,
               slopeFormat=None, trigonometric=False, zeroForFlat=False,
               addAlpha=None,
               callback=None, callback_data=None):
@@ -815,6 +815,8 @@ def DEMProcessingOptions(options=None, colorFilename=None, format=None,
             new_options += ['-combined']
         if multiDirectional:
             new_options += ['-multidirectional']
+        if igor:
+            new_options += ['-igor']
         if slopeFormat == 'percent':
             new_options += ['-p']
         if trigonometric:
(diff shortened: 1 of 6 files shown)


rouault:
Looks good. Could you extend it with documentation in gdal/apps/gdal_utilities.dox and test(s) in autotest/utilities/test_gdaldem_lib.py ?

zdila:
@rouault Please check the commit and adjust documentation if necessary.

gdal/swig/python/osgeo/gdal.py
```diff
@@ -815,6 +815,8 @@ def DEMProcessingOptions(options=None, colorFilename=None, format=None,
             new_options += ['-combined']
         if multiDirectional:
             new_options += ['-multidirectional']
+        if igor:
```
rouault: those two lines should be in swig/include/python/gdal_python.i as well. In fact swig/python/osgeo/gdal.py should be regenerated by running "make generate" from swig/python with SWIG 3.0.8

zdila: @rouault Could you please do it for me? I only wanted to add this algorithm and don't plan to do active development on gdal so it is not worth for me to learn all the gdal project details. And maybe there is much more what I forgot but I'd like that some core developer would polish it. Thanks.

zdila: I've also never coded in python so I am just guessing whare to add what by greppig `multidirectional` option.

rouault: OK, I've applied this minor fix and squashed together your commits per 4db55e6

zdila: Thank you!

Issue #1333, opened 2019-03-01
Read HDFS files result JVM fatal error


qfjiang1993:
## Using GDAL read local files works fine

I build GDAL-2.4.0 with HDFS support and Java binding, When I use GDAL Java API to read GeoTiff files in local filesystem, `/home/jiang/sampleRGB.tif` , it works fine, and I get the GeoTiff file attributes like below:

### GeoTiff format
```
[jiang@lake-geo demo]$ java -cp .:gdal.jar:$CLASSPATH RasterDemo /home/jiang/sampleRGB.tif
Driver: GTiff/GeoTIFF
/home/jiang/sampleRGB.tif
Size: 4550, 3886
Coordinate System is:
GEOGCS["WGS 84",DATUM["WGS_1984",SPHEROID["WGS 84",6378137,298.257223563,AUTHORITY["EPSG","7030"]],AUTHORITY["EPSG","6326"]],PRIMEM["Greenwich",0],UNIT["degree",0.0174532925199433],AUTHORITY["EPSG","4326"]]
Origin = (114.151033, 30.102782)
PixelSize = (0.000038, -0.000038)
Metadata:
        AREA_OR_POINT=Area
        TIFFTAG_XRESOLUTION=1
        TIFFTAG_YRESOLUTION=1
IMAGE_STRUCTURE Metadata:
        INTERLEAVE=PIXEL
Band 1 Block=4550x1 Type=1, ColorInterpretation=3
        Description =
        NoData Value=null
Band 2 Block=4550x1 Type=1, ColorInterpretation=4
        Description =
        NoData Value=null
Band 3 Block=4550x1 Type=1, ColorInterpretation=5
        Description =
        NoData Value=null
```

### AAIGrid & HFA formats

Then, I test Arc/Info ASCII Grid (.asc) file and Erdas Imagine Images (.img) file, both work fine, out was below: 

```
[jiang@lake-geo demo]$ java -cp .:gdal.jar:$CLASSPATH RasterDemo /home/jiang/srtm_59_06.asc
Driver: AAIGrid/Arc/Info ASCII Grid
/home/jiang/srtm_59_06.asc
/home/jiang/srtm_59_06.prj
Size: 6001, 6001
Coordinate System is:
[...]

PR #1349, opened 2019-03-09
vsihdfs: More Than One hdfsRead Call Sometimes Needed

jamesmcclain:
## What does this PR do?

This addresses the issue with reading certain datasets on HDFS file systems reported in https://github.com/OSGeo/gdal/issues/1333 .  This issue was caused by the fact that `hdfsRead` evidently does not always return all requested bytes when the file in question is on HDFS.

I did not include a unit test update because this problem is hard to trigger on a local file system (and therefore, a unit test for this specific problem would require a running HDFS cluster -- I am however open to testing ideas and will include any that are requested).

https://github.com/OSGeo/gdal/issues/1333

## Environment

A clean Ubuntu 18.10 virtual machine using OpenJDK 11.  Tested with Hadoop 2.9.2 in a [single node cluster](https://hadoop.apache.org/docs/stable/hadoop-project-dist/hadoop-common/SingleCluster.html) configuration.

## Tasklist

 - [ ] Add test case(s)
 - [ ] Review
 - [ ] Adjust for comments

Closes https://github.com/OSGeo/gdal/issues/1333

### DIFF
--- a/gdal/port/cpl_vsil_hdfs.cpp
+++ b/gdal/port/cpl_vsil_hdfs.cpp
@@ -40,6 +40,7 @@
 #endif
 
 #include <cstring>
+#include <climits>
 
 #include "cpl_port.h"
 #include "cpl_vsi.h"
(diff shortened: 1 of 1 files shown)


gdal/port/cpl_vsil_hdfs.cpp
```diff
@@ -154,18 +154,27 @@ VSIHdfsHandle::Read(void *pBuffer, size_t nSize, size_t nMemb)
   if (nSize == 0 || nMemb == 0)
     return 0;
 
-  int bytes = hdfsRead(poFilesystem, poFile, pBuffer, nSize * nMemb);
+  int bytes_wanted = nSize * nMemb;
```
rouault: isn't there a potential issue of size_t -> int overflow in case of big input arguments ?

jamesmcclain: The return type of `hdfsRead` seems to be `tSize` (https://github.com/apache/hadoop/blob/trunk/hadoop-hdfs-project/hadoop-hdfs-native-client/src/main/native/libhdfs/include/hdfs/hdfs.h#L585-L600) which is a typedef of `int32_t` (https://github.com/apache/hadoop/blob/trunk/hadoop-hdfs-project/hadoop-hdfs-native-client/src/main/native/libhdfs/include/hdfs/hdfs.h#L75), so an `int` "should be" sufficient.  Maybe I should change those from `int`s to `int32_t`s just to be safe though?

jamesmcclain: I see what you are saying: you are saying that there is a mismatch between the two last arguments to `Read` (of type `size_t` which I think is `uint64_t`) and between the size argument that `hdfsRead` takes (of type `tSize` which is `int32_t`)?

If there is an overflow, should it throw an error or just return as many records as it can?

PR #1355, opened 2019-03-12
MSSQLSpatial: Adding improved extent queries for MSSQLServer driver

mathijs-dumon:
## What does this PR do?
 - Adds methods and a MSSQL version type to the OGRMSSQLSpatial data source class to get and parse the database version
 - Adds overrides for the GetExtent methods of the OGRLayer implementation for MSSQLSpatial which are more performant for large datasets

None that I know off

## Tasklist
 - [x] Review
 - [ ] Adjust for comments

### DIFF
--- a/gdal/ogr/ogrsf_frmts/mssqlspatial/ogr_mssqlspatial.h
+++ b/gdal/ogr/ogrsf_frmts/mssqlspatial/ogr_mssqlspatial.h
@@ -353,6 +353,9 @@ class OGRMSSQLSpatialTableLayer final: public OGRMSSQLSpatialLayer
     OGRErr              CreateSpatialIndex();
     void                DropSpatialIndex();
 
+    virtual OGRErr      GetExtent(OGREnvelope *psExtent, int bForce) override { return GetExtent(0, psExtent, bForce); }
+    virtual OGRErr      GetExtent(int iGeomField, OGREnvelope *psExtent, int bForce) override;
+
     virtual void        ResetReading() override;
     virtual GIntBig     GetFeatureCount( int ) override;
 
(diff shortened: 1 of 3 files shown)


mathijs-dumon:
Thanks for the feedback
 1. I will add the check for OLCFastGetExtent - overlooked that. It's not version dependent, running the CTE expressions on the database will still be more performant than loading all feature extents.
 2. I will change the logic to fall back to the generic approach. I'll also add an additional guard if any of the extent values are NULL values. The coalesece in the SQL should prevent this, but just in case.
 3. The version string query should have minimal impact on opening the datasource. I based this on how the pgsql driver does this. If you still think this is required can you point me to an example of the lazy loading? Or do you mean running the query when GetExtents is called the first time?

szekerest:
@mathijs-dumon Are you planning to proceed with 1. and 2. above? I'm keen to get this addition merged in the near future.

mathijs-dumon:
@szekerest I just force-pushed the changes to the branch which include the changes request in point 1 & 2.

szekerest:
@mathijs-dumon Should this file be added to the repository? gdal/ogr/ogrsf_frmts/mssqlspatial/MSSQLQueryPatch

mathijs-dumon:
Oh dear - no that shouldn't be there. Remind myself not to use wildcards
with git add...

Op ma 25 mrt. 2019 20:14 schreef Tamas Szekeres <notifications@github.com:

szekerest:
@mathijs-dumon Will you remove that from the commit?

mathijs-dumon:
> @mathijs-dumon Will you remove that from the commit?

Just did - sorry for the inconvenience

szekerest:
Thanks for the fixes. It looks like EnvelopeAggregate throws error on  invalid geometries, with quite long error messages that crashes GDAL. Added an issue for that part. https://github.com/OSGeo/gdal/issues/1390

PR #1358, opened 2019-03-12
gdal2tiles: prevent accidental copy of full GeoTIFF into temporary .vrt file

kevinmehall:
This line is used to write the temporary .vrt for the reprojected input.
However, if the input was already in the correct projection,
`self.warped_input_dataset` was the input GeoTIFF dataset directly due
to the previous line, so `GetDriver` returned the GeoTIFF driver rather
than VRT. It then created a .vrt file that was actually in GeoTIFF
format and copied the entire input into it.


rouault:
Thanks

rouault:
Backported to 2.4 per fae3a65

PR #1359, opened 2019-03-12
Vagrant: new config key for LXC driver

miurahr:
## What does this PR do?

Fix Vagrant configuration for LXC driver

LXC 2.1 introduce new configuration parameter keys and
LXC 3.0 removes legacy keys.
Here replace legacy key with new one.

Signed-off-by: Hiroshi Miura <miurahr@linux.com>

## Tasklist

 - [x] Review
 - [x] Adjust for comments

## Environment

Vagrant development environment
Vagrant driver: LXC

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -43,7 +43,10 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     ovrd.vm.box = "cultuurnet/ubuntu-14.04-64-puppet"
     lxc.backingstore = 'dir'
     lxc.customize 'cgroup.memory.limit_in_bytes', vm_ram_bytes
-    lxc.customize 'aa_allow_incomplete', 1
+    # LXC 3 or later deprecated old parameter
+    lxc.customize 'apparmor.allow_incomplete', 1
+    # for LXC 2.1 or before
+    #lxc.customize 'aa_allow_incomplete', 1
     lxc.container_name = "gdal-vagrant"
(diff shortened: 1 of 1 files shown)


rouault:
For my information, would there be an issue in putting both options for some time ? I see for example my ubuntu 16.04 uses lxc 2.0.8 (I don't use the Vagrant env myself, so I don't bother much)

miurahr:
I'm using LXC 3.0 vesion so find an issue.

> For my information, would there be an issue in putting both options for some time ? I see for example my ubuntu 16.04 uses lxc 2.0.8 (I don't use the Vagrant env myself, so I don't bother much)

OK, I'd like to wait or update [Support for detecting lxc version](https://github.com/fgrehm/vagrant-lxc/issues/466) in vagrant-lxc project and update the PR to use proper parameter key with it, so here is added flags WIP.

miurahr:
> OK, I'd like to wait or update [Support for detecting lxc version](https://github.com/fgrehm/vagrant-lxc/issues/466) in vagrant-lxc project and update the PR to use proper parameter key with it, so here is added flags WIP.

There seems no way to detect LXC version through Vagrant config interface. I'd like to add variable to configure LXC version and select proper customize paramter for it.

rouault:
I don't want you to overcomplicate this. My question was mostly a question :-) Perhaps just add in the Vagrantfile a comment with the old name of the parameter, in case someone needs it

PR #1362, opened 2019-03-13
gdal2tiles: Multiprocessing fixes and cleanup

kevinmehall:
Python's `multiprocessing.Pool` type has some methods that can replace some of the manual interprocess communication used in `gdal2tiles.py` in a cleaner and less error-prone way.

This fixes a bug when combining the `--exclude` option with `--processes`: the skipped fully transparent tiles were not properly writing a completion event to the queue, so the progress bar did not reflect these skipped tiles, and the progress bar process hung waiting to receive the expected number of results from the queue.

Similarly, if one of the subprocesses threw an exception, there were cases where the program would hang waiting to receive on a queue.

### DIFF
--- a/gdal/swig/python/scripts/gdal2tiles.py
+++ b/gdal/swig/python/scripts/gdal2tiles.py
@@ -2744,19 +2744,12 @@ class GDAL2Tiles(object):
         return s
 
 
-def worker_tile_details(input_file, output_folder, options, send_pipe=None):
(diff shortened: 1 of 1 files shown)


rouault:
@gbataille Any opinion on this ?

rouault:
Lightly tested locally. Looks reasonable. Merging

PR #1371, opened 2019-03-18
Increasing DOWNLOAD_CHUNK_SIZE to 128K

ofriedma:
When dealing with http protocol the round trip latency could be with high latency.
Why not create bigger requests to improve overall performance for user using vsicurl and all object storage(for example vsis3) usage.
Increasing this variable will decrease the amount of requests.

## What does this PR do?
Increasing DOWNLOAD_CHUNK_SIZE to 128K
None, this will improve performance significantly 

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Review
 - [ ] Adjust for comments

## Environment

Not relevant

### DIFF
--- a/gdal/port/cpl_vsil_curl.cpp
+++ b/gdal/port/cpl_vsil_curl.cpp
@@ -94,7 +94,7 @@ int VSICurlUninstallReadCbk( VSILFILE* /* fp */ )
 #define ENABLE_DEBUG_VERBOSE 0
 
 static int N_MAX_REGIONS = 1000;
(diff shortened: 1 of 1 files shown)


rouault:
Did you actually try this with GDAL 2.4 / master ? DOWNLOAD_CHUNK_SIZE is overridden by VSIInstallCurlFileHandler()

ofriedma:
I found out it, closing the PR, thanks

PR #1384, opened 2019-03-21
Elasticsearch: Fix _mapping url

nickpeihl:
## What does this PR do?

The current construction of the URL for Elasticsearch `_mapping` is incorrect and causes failures when converting data to Elasticsearch with a defined mapping file. In all versions of Elasticsearch, the PUT  _mapping URL should be constructed as `/{index_name}/_mapping/{mapping_type}`. This PR corrects the URL path for PUT mapping. See
https://www.elastic.co/guide/en/elasticsearch/reference/5.6/indices-put-mapping.html for more info.

n/a

## Tasklist

 - [ ] Add test case(s)
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/autotest/ogr/ogr_elasticsearch.py
+++ b/autotest/ogr/ogr_elasticsearch.py
@@ -130,7 +130,7 @@ def test_ogr_elasticsearch_1():
     assert gdal.GetLastErrorType() == gdal.CE_None
 
     gdal.FileFromMemBuffer(
-        '/vsimem/fakeelasticsearch/foo/FeatureCollection/_mapping&POSTFIELDS'
+        '/vsimem/fakeelasticsearch/foo/_mapping/FeatureCollection&POSTFIELDS'
         '={ "FeatureCollection": { "properties": { "type": '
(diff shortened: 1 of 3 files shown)


rouault:
> In all versions of Elasticsearch

Remark: perhaps "recent" ones, but I'm pretty sure this might have worked, perhaps by chance, in ancient ones  (at least up to 5.X) since that code is from the beginning.

nickpeihl:
> > In all versions of Elasticsearch
> 
> Remark: perhaps "recent" ones, but I'm pretty sure this might have worked, perhaps by chance, in ancient ones (at least up to 5.X) since that code is from the beginning.

Perhaps it did work by accident. Here's the relevant [2.4 docs](https://www.elastic.co/guide/en/elasticsearch/reference/2.4/indices-put-mapping.html#_multi_index). And the [1.7 docs](https://www.elastic.co/guide/en/elasticsearch/reference/1.7/indices-put-mapping.html#put-mapping-multi-index). Both show the URL in the format of `{index}/_mapping/{type}`

Issue #1299, opened 2019-02-19
MSSQLSpatial: Add support for curve geometries


szekerest:
Curve geometries have been introduced in MSSQL Server 2012.

PR #1389, opened 2019-03-25
Rework mssql driver to support curve geometries (#1299)

szekerest:


gdal/ogr/ogrsf_frmts/mssqlspatial/drv_mssqlspatial.html
```diff
@@ -139,6 +139,7 @@ <h2>Configuration options</h2>
     <li><b>MSSQLSPATIAL_SHOW_FID_COLUMN</b>: Force to display the FID colums as a feature attribute. Default = "NO".</li>
     <li><b>MSSQLSPATIAL_USE_GEOMETRY_COLUMNS</b>: Use/create geometry_columns metadata table in the database. Default = "YES".</li>
     <li><b>MSSQLSPATIAL_LIST_ALL_TABLES</b>: Use mssql catalog to list available layers. Default = "NO".</li>
+    <li><b>MSSQLSPATIAL_USE_GEOMETRY_VALIDATION</b>: (From GDAL 2.5.0) Let the driver to detect the geometries which would trigger run time errors at MSSQL server. The driver tries to correct these geometries before submitting that to the server. Default = "NO".</li>
```
rouault: the 'to' after 'Let the driver' should be remove, no ?

gdal/ogr/ogrsf_frmts/mssqlspatial/ogrmssqlgeometrywriter.cpp
```diff
@@ -203,41 +232,144 @@ void OGRMSSQLGeometryWriter::WritePoint(double x, double y)
...
+    OGRSimpleCurve* poSubGeom;
+    WriteByte(FigureAttribute(iFigure), FA_CURVE);
     WriteInt32(PointOffset(iFigure), iPoint);
-    if (chProps & SP_HASZVALUES)
+    for (int i = 0; i < poGeom->getNumCurves(); i++)
     {
-        for (i = 0; i < poGeom->getNumPoints(); i++)
-            WritePoint(poGeom->getX(i), poGeom->getY(i), poGeom->getZ(i));
+        poSubGeom = poGeom->getCurve(i)->toSimpleCurve();
+        switch (poSubGeom->getGeometryType())
```
rouault: minor: you coud use          switch (wkbFlatten(poSubGeom->getGeometryType())) and use only the 2D version of the wkb constants in the case 
This holds for similar switch() in other files

gdal/ogr/ogrsf_frmts/mssqlspatial/ogrmssqlgeometryparser.cpp
```diff
@@ -31,7 +31,7 @@
 
 CPL_CVSID("$Id$")
 
-/*   SqlGeometry serialization format
+/*   SqlGeometry/SqlGeography serialization format
```
rouault: this description is also repeated in geometrywriter.cpp. Perhaps move it in a mssqlgeometryformat.txt file ?

gdal/ogr/ogrsf_frmts/mssqlspatial/ogrmssqlgeometrywriter.cpp
szekerest: @rouault Do you think if the code is more readable in that case? We should also enumerate the M/ZM variants however.

gdal/ogr/ogrsf_frmts/mssqlspatial/ogrmssqlgeometryparser.cpp
szekerest: Yes, but it was already there, however. I've just added the MSSQL2012 additions.

gdal/ogr/ogrsf_frmts/mssqlspatial/drv_mssqlspatial.html
szekerest: probably yes, will fix that

gdal/ogr/ogrsf_frmts/mssqlspatial/ogrmssqlgeometrywriter.cpp
rouault: > We should also enumerate the M/ZM variants however.

no, the code would become just
```
switch (wkbFlatten(poSubGeom->getGeometryType()))
{
        case wkbLineString:
            // do stuff
            break;
        case wkbCircularString:
            // do stuff
            break;
        default:
            break;
}
```
Looks better to me

PR #1393, opened 2019-03-27
fix `gdal_calc.py Calc` function  docs error

sshuair:
the orginal docs:

 ` Calc(calc="A*(A>0)", A="input.tif", A_Band=2, outfile="result.tif", NoDataValue=0)`

`A_Band` is wrong, `B` should be lower case.


rouault:
thanks

PR #1409, opened 2019-04-01
Address compiler warnings

Algunenano:
## What does this PR do?

Multiple commits addressing warnings for both GCC (8.2) and clang (7.0).

Please let me know if you'd rather have several PRs open (one per commit?) instead. Some of them seem clearly false positives, but I think it's worth using #pragmas for those to have a silent build.

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments

## Environment

Environment configured with:
```
./configure --prefix=/usr --with-netcdf --with-libtiff --with-sqlite3 --with-geotiff --with-mysql --with-python --with-curl --with-hdf5 --with-perl --with-geos --with-png --with-poppler --with-spatialite --with-openjpeg CFLAGS="$CFLAGS -Werror" CXXFLAGS="$CXXFLAGS -Werror" CC=$(CC) CXX=$CXX)
```
Where `CFLAGS` == `CXXFLAGS` == `=-march=native -mtune=native -O3 -pipe -fno-plt -Wall -Wextra -g3 -gdwarf-4 -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer`

And CC and CXX changed for gcc and clang.

* Compilers:
- gcc version 8.2.1 20181127 (GCC) 
- clang version 7.0.1 (tags/RELEASE_701/final)

### DIFF
--- a/gdal/frmts/grib/degrib/g2clib/dec_png.c
+++ b/gdal/frmts/grib/degrib/g2clib/dec_png.c
@@ -39,7 +39,7 @@ static void user_read_data(png_structp png_ptr,png_bytep data, png_size_t length
         offset=mem->stream_len;
     /*     printf("SAGrd %ld %ld %x\n",offset,length,ptr);  */
         memcpy(data,ptr+offset,length);
-        mem->stream_len += length;
+        mem->stream_len += (g2int)length;
      }
 }
 
--- a/gdal/frmts/grib/degrib/g2clib/dec_png.c
+++ b/gdal/frmts/grib/degrib/g2clib/dec_png.c
@@ -16,9 +16,7 @@ struct png_stream {
 };
 typedef struct png_stream png_stream;
 
-void user_read_data(png_structp , png_bytep , png_uint_32 );
-
-void user_read_data(png_structp png_ptr,png_bytep data, png_uint_32 length)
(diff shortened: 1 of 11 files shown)


gdal/frmts/grib/degrib/g2clib/dec_png.c
```diff
@@ -16,9 +16,9 @@ struct png_stream {
 };
 typedef struct png_stream png_stream;
 
-void user_read_data(png_structp , png_bytep , png_uint_32 );
+void user_read_data(png_structp , png_bytep , png_size_t );
```
rouault: you could as well declare the function as static.

Algunenano:
Thanks for the comments. I'll address them and patch/rewrite the individual commits, which should also fix the build failures.

Algunenano:
I've rebased on top of current master and patched the commits where necessary to address the comments.

rouault:
VisualStudio on x64 raises
```
c:\projects\gdal\gdal\frmts\grib\degrib\g2clib\dec_png.c(42): warning C4267: '+=': conversion from 'size_t' to 'g2int', possible loss of data
```
I presume the following change should be done:
```
mem->stream_len += length;
```
to
```
mem->stream_len += (g2int)length;
```

PR #1425, opened 2019-04-07
gdal2tiles script fixing leaflet version

NasH5169:
When using the option --webviewer=leaflet the actual version of gdal2tiles.py (2.4.1) does not work. The javascript and css links to Leaflet must be updated.

### DIFF
--- a/gdal/swig/python/scripts/gdal2tiles.py
+++ b/gdal/swig/python/scripts/gdal2tiles.py
@@ -2322,8 +2322,8 @@ class GDAL2Tiles(object):
             <title>%(htmltitle)s</title>
 
             <!-- Leaflet -->
-            <link rel="stylesheet" href="http://cdn.leafletjs.com/leaflet-0.7.5/leaflet.css" />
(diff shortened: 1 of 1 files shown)


rouault:
Admittedly the version currently referenced is ancient, but what doesn't work exactly ? Seems to work for me, for example with `python swig/python/scripts/gdal2tiles.py --webviewer=leaflet ../autotest/gdrivers/data/small_world.tif smallworld_leaflet`

NasH5169:
Really? Because but with the current 2.4.1, the following links from the previous version make the generated leaflet.html failed on my side:

```
<link rel="stylesheet" href="http://cdn.leafletjs.com/leaflet-0.7.5/leaflet.css" />
<script src="http://cdn.leafletjs.com/leaflet-0.7.5/leaflet.js"></script>
```

rouault:
Can you try my above gdal2tiles invokation with https://github.com/OSGeo/gdal/blob/master/autotest/gdrivers/data/small_world.tif ?

NasH5169:
You are right, it's working... I cannot remember why the first time the viewer generated was unable to work.

PR #1426, opened 2019-04-07
Fix MSSQL extent calculation for geography type, take care of invalid…

szekerest:


szekerest:
@mathijs-dumon I've found a couple of issues with the current implementation and prepared this PR to fix them:
1. For the geography type we have no STEnvelope, so that should be converted to geometry first
2. geography::EnvelopeAggregate results in CURVEPOLYGON in most cases or may result in FULLGLOBE in extraordinaly cases. Neither of them can be used to return the min/max values of the coordinates. We better to convert the geography type to geometry and then call geometry::EnvelopeAggregate instead.
3. I've removed coalesce from the statements, we better to return an error on the null values rather than providing zeroes instead.

mathijs-dumon:
@szekerest 
First of all thanks for testing & fixing the issues. Points 1 & 2 totally make sense.
In regard to point 3: EnvelopeAggregate ignores null inputs and will only return null when all inputs are null. I think the pre-2012 implementation should behave in a similar way. However - I agree coalesce isn't the right approach here. Instead (if you agree that null values should just be ignored for extent calculations) the final select should be a "SELECT MIN(point.STX), MIN(point.STY), MAX(point.STX), MAX(point.STY) FROM CORNERS WHERE point IS NOT NULL; in the pre-2012 implementation.

szekerest:
@mathijs-dumon I think the Min/Max aggregates also ignore the null values, so the "WHERE point IS NOT NULL" is not necessary.

Issue #1305, opened 2019-02-21
OGR Carto - Avoid dropping tables outside a transaction


Algunenano:
**Note:** This is a feature request, but I want to document the what problem is trying to address and the possible approaches to tackle it beforehand. I intent to tackle this soon.

To keep datasets and maps sync'd, CARTO is currently using 2 different paths:
- For dataset updates there are DB triggers that alert the invalidation service when a change happens. This invalidation service checks which maps are affected by this change (there could be more than one), and ask whatever the caching systems (Varnish, CDNs...) to drop any related cached requestso any subsequent request will hit the DB.
- For dataset addition and deletion: This is handled by [Builder](https://github.com/CartoDB/cartodb), the main UI for CARTO, but only when the user visits the dashboard:
  - When you add a dataset and cartodby it, once you visit the dashboard it will appear there so you can use it in the UI to create maps. This works fine.
  - When you remove a dataset, the maps associated with it will only be deleted once you visit the dashboard, so this means that they will be there even though the data has been deleted.

This set up, of only updating maps when the user visits the dashboard has proven to be riddled with issues, made some related functionalities harder and led to a lot of confusion, so we are in the process of changing it into something similar to the update approach, and update maps as soon as possible, independently of whether you visit the dashboard or not.

Nevertheless, the current behaviour has trained users in a dangerous way of working: If you wanted to update a map, you could drop the table and recreate it with the new data and you'd be fine as long as you didn't visit the dashboard during the process, as the maps would be kept intact (as far as the dashboard goes, the table would be updated, but not deleted). Some clients, for example GDAL's, rely on this behaviour.

So here is the problem: Let's say that you have a dataset with a map associated and you are using ogr to update it. Since you want to modify the table, you set OVERWRITE to true and run it, so the process is:
- DROP TABLE.
- CREATE TABLE.
- INSERT / COPY.
- CARTODBFY

With the old behaviour, you'd be fine as long as you didn't visit the dashboard between the `DROP` and the `CREATE` calls (I hope you now understand some of the issues behind this change :D), but with the current behaviour, the map will be dropped once you call `DROP`, so even though you might create the same table with the same schema, the map will be gone.

I see 3 ways to address this within GDAL driver:
- Join the `DROP` and `CREATE` calls in a transaction (the new trigger will handle this correctly).
- Use `TRUNCATE` instead of `DROP`, and alter the table columns to match the new import.
- Another option is to change the order of the process when overwritting:
[...]

PR #1443, opened 2019-04-15
CARTO: Overwrite tables in single transactions

Algunenano:
## What does this PR do?

Changes how a table is overwritten in CARTO. Instead of 2 calls, first the DROP and then the CREATE TABLE, it defers the table drop and sends them both as a single transaction:
- `BEGIN; DROP TABLE "x"; CREATE TABLE "x" ...; COMMIT;`

It does this to avoid an issue in the CARTO platform where a map depending on a table could be dropped if it was "checked" during the update, that is between the drop and the creation of the table. It also enables future changes on how this dependency is handled internally.

Closes https://github.com/OSGeo/gdal/issues/1305.

## Tasklist

 - [X] Add test case(s).
 - [ ] Review
 - [X] Adjust for comments

### DIFF
--- a/gdal/ogr/ogrsf_frmts/carto/ogr_carto.h
+++ b/gdal/ogr/ogrsf_frmts/carto/ogr_carto.h
@@ -138,6 +138,8 @@ class OGRCARTOTableLayer : public OGRCARTOLayer
     bool                bCartodbfy;
     int                 nMaxChunkSize;
 
+    bool                bDropOnCreation;
+
     void                BuildWhere();
     std::vector<bool>   m_abFieldSetForInsert;
 
@@ -201,6 +203,11 @@ class OGRCARTOTableLayer : public OGRCARTOLayer
                                             bool bHasUserFieldMatchingFID, 
                                             bool bHasJustGotNextFID );
     char *              OGRCARTOGetHexGeometry( OGRGeometry* poGeom, int i );
+
+    void                SetDropOnCreation( bool bFlag )
+        { bDropOnCreation = bFlag; }
+    bool                GetDropOnCreation() const
+        { return bDropOnCreation; }
 };
 
 /************************************************************************/
--- a/autotest/ogr/ogr_carto.py
+++ b/autotest/ogr/ogr_carto.py
@@ -732,14 +732,11 @@ Error""")
     gdal.PopErrorHandler()
     assert lyr is None
 
-    with gdaltest.tempfile("""/vsimem/carto&POSTFIELDS=q=DROP TABLE "table1"&api_key=foo""",
+    with gdaltest.tempfile("""/vsimem/carto&POSTFIELDS=q=BEGIN; DROP TABLE IF EXISTS "table1";CREATE TABLE "table1" ( cartodb_id SERIAL,the_geom Geometry(MULTIPOLYGON,0),PRIMARY KEY (cartodb_id) );DROP SEQUENCE IF EXISTS "table1_cartodb_id_seq" CASCADE;CREATE SEQUENCE "table1_cartodb_id_seq" START 1;ALTER SEQUENCE "table1_cartodb_id_seq" OWNED BY "table1".cartodb_id;ALTER TABLE "table1" ALTER COLUMN cartodb_id SET DEFAULT nextval('"table1_cartodb_id_seq"'); COMMIT;&api_key=foo""",
                            """{"rows":[], "fields":{}}"""):
         lyr = ds.CreateLayer('table1', geom_type=ogr.wkbPolygon, options=['OVERWRITE=YES', 'CARTODBFY=NO'])
-
-    f = ogr.Feature(lyr.GetLayerDefn())
-    f.SetGeometry(ogr.CreateGeometryFromWkt('POLYGON((0 0,0 1,1 0,0 0))'))
-    with gdaltest.tempfile("""/vsimem/carto&POSTFIELDS=q=CREATE TABLE "table1" ( cartodb_id SERIAL,the_geom Geometry(MULTIPOLYGON,0),PRIMARY KEY (cartodb_id) );DROP SEQUENCE IF EXISTS "table1_cartodb_id_seq" CASCADE;CREATE SEQUENCE "table1_cartodb_id_seq" START 1;ALTER SEQUENCE "table1_cartodb_id_seq" OWNED BY "table1".cartodb_id;ALTER TABLE "table1" ALTER COLUMN cartodb_id SET DEFAULT nextval('"table1_cartodb_id_seq"')&api_key=foo""",
-                           """{"rows":[], "fields":{}}"""):
+        f = ogr.Feature(lyr.GetLayerDefn())
+        f.SetGeometry(ogr.CreateGeometryFromWkt('POLYGON((0 0,0 1,1 0,0 0))'))
         assert lyr.CreateFeature(f) == 0
 
     gdal.ErrorReset()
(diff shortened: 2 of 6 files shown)


gdal/ogr/ogrsf_frmts/carto/ogr_carto.h
```diff
@@ -273,7 +277,9 @@ class OGRCARTODataSource : public OGRDataSource
                                      OGRSpatialReference *poSpatialRef = nullptr,
                                      OGRwkbGeometryType eGType = wkbUnknown,
                                      char ** papszOptions = nullptr ) override;
-    virtual OGRErr      DeleteLayer(int) override;
+    virtual OGRErr      DeleteLayer(int iLayer) override
+        { bDropOnCreate = true; return DeleteLayer(iLayer, false); }
```
rouault: is this bDropOnCreate = true needed, as the layer deletion is not deferred ?
If so, then OGRCARTODataSource::bDropOnCreate wouldn't be needed at all

Algunenano: So initially this was doing `return DeleteLayer(iLayer, true);` so that the deletion from `GetLayerAndOverwriteIfNecessary` (`ogr2ogr_lib.cpp:3263`) would be deferred in any case.

Then while testing I noticed that doing this meant that `DeleteLayer` to simply remove a layer without the intention of overwriting the table doesn't work, and I removed it without taking into account that then this doesn't fix the issue and you get 2 calls:

```
HTTP: These POSTFIELDS were sent:q=DROP TABLE "pref"&api_key=X
...
HTTP: These POSTFIELDS were sent:q=BEGIN; DROP TABLE IF EXISTS "pref";CREATE TABLE "pref" ( cartodb_id SERIAL,"the_geom" VARCHAR,"pref" VARCHAR,"gid" VARCHAR,"prefcode" VARCHAR,"area" VARCHAR,"area_contents" VARCHAR,"area_youto" VARCHAR,"area_pop" VARCHAR,"area_bill" VARCHAR,PRIMARY KEY (cartodb_id) );DROP SEQUENCE IF EXISTS "pref_cartodb_id_seq" CASCADE;CREATE SEQUENCE "pref_cartodb_id_seq" START 1;ALTER SEQUENCE "pref_cartodb_id_seq" OWNED BY "pref".cartodb_id;ALTER TABLE "pref" ALTER COLUMN cartodb_id SET DEFAULT nextval('"pref_cartodb_id_seq"'); COMMIT;&api_key=X
```

I'm going to revisit the issue. I think I might need to add the extra signature to `GDALDataset` itself and only overwrite it in the necessary datasets (CARTO ogr format in this case):
```c++
OGRErr GDALDataset::DeleteLayer( CPL_UNUSED int iLayer )

{
    CPLError(CE_Failure, CPLE_NotSupported,
             "DeleteLayer() not supported by this dataset.");

    return OGRERR_UNSUPPORTED_OPERATION;
}

OGRErr GDALDataset::DeleteLayer( int iLayer , CPL_UNUSED bool bDeferDrop )

{
    return this->DeleteLayer(iLayer);
}
```

Does that make sense to you? I'll update the PR to make it clearer.

rouault: > Does that make sense to you?

hum, I'd prefer not to clutter the driver API with such a particular case: bDeferDrop doesn't really tell when this is going to be done effectively, and the concept of deferred actions in the API feels messy. Drivers may do deferred actions if needed but this should remain hidden from the user as much as possible.
Another solution would be to defer the layer deletion in the CARTO driver and issue the DROP TABLE before the next call to the CARTO API or dataset object closing.

Algunenano: > Another solution would be to defer the layer deletion in the CARTO driver and issue the DROP TABLE before the next call to the CARTO API or dataset object closing.

The problem I see with that approach is that we'll hide any errors that might only happen when closing the datasource or, even worse, throw on the destructor.

Another different approach I see is to check the overwrite flag while creating the CARTO driver or datasource, and if `overwrite`it's on then never delete any layer remotely when `DeleteLayer` is called and always do the `DROP IF EXISTS + CREATE` on `RunDeferredCreationIfNecessary `.

This would mean a breaking change since if you want to delete a layer remotely (without overwriting) you'd need to set `OVERWRITE` to NO. This makes me wonder, is it even possible to just delete a remote table using ogr2ogr or the bindings?
It seems to me that GDAL only tries to delete a remote table when it's going to create a new one with the same name, so not being able to do only the remote deletion wouldn't be an issue at all. 
I don't have much experience in gdal so that assumption might be wrong.

rouault: >  is that we'll hide any errors

yes, that's the drawback indeed

> Another different approach I see is to check the overwrite flag while creating the CARTO driver or datasource
What could be done in ogr2ogr is to check in the driver metadata (GDAL_DS_LAYER_CREATIONOPTIONLIST) layer creation options is OVERWRITE exists (you can look around at line 3685 of ogr2ogr_lib for similar checks with other otions) as a possible creation option, and if so, and when -overwrite is specified:
- not issue DeleteLayer() from ogr2ogr
- force add OVERWRITE=YES when calling CreateLayer()
A bit of a hack, but likely a reasonable one. That said that would defeat the hack done around line 4166 for an issue I found with PostgreSQL...
So probably that for now my above suggestion should be restricted to adding an explicit check for the Carto driver.

> This makes me wonder, is it even possible to just delete a remote table using ogr2ogr or the bindings?

Sure, anyone could potentially use the OGR API to issue DeleteLayer(). I'd suggest adding in the driver doc a paragraph about using preferably CreateLayer() with OVERWRITE=YES rather than DeleteLayer() + CreateLayer()

Algunenano: > What could be done in ogr2ogr is to check in the driver metadata (GDAL_DS_LAYER_CREATIONOPTIONLIST) layer creation options is OVERWRITE exists (you can look around at line 3685 of ogr2ogr_lib for similar checks with other otions) as a possible creation option, and if so, and when -overwrite is specified

A question about this. I see that CARTO supports both `-overwrite` and a layer creation option (`--lco OVERWRITE=ON`) but I'm not sure what's the use of the later. If I use it without `-overwrite` it doesn't work:
`
ogr2ogr --config CARTO_API_KEY cfd51af386ec59859a1415d51f9b1809c014d369     --config CARTO_HTTPS no     --config CARTO_API_URL http://localhost:8080/user/dev/api/v2/sql   -lco OVERWRITE=ON     --config CPL_DEBUG HTTP     --config CPL_DEBUG CARTO     --debug ON     -f Carto "Carto:dev"   /home/raul/Datasets/pref.csv

Warning 1: Layer creation options ignored since an existing layer is
         being appended to.
`
Am I using it wrong?


> layer creation options is OVERWRITE exists

Is this to detect that the destination datasource is CARTO? If so, is there a better option to do it? `CARTODBFY` is a more obvious choice to avoid future conflicts with other drivers.

With your input, my current idea would be to detect if the CARTO driver is being used in  ogr2ogr_lib.cpp::GetLayerAndOverwriteIfNecessary and, instead of calling `DeleteLayer`, mark the layer to be overwritten* and leave `pbOverwriteActuallyDone` to false. This effectively avoids  the hack you mentioned for Postgresql (it checks `bOverwriteActuallyDone`).

*It won't be pretty, but doing something like:
```c++
if (dynamic_cast<OGRCARTOLayer>(layer))
    dynamic_cast<OGRCARTOLayer>(layer).markForOverwrite();
```

In fact, it'd probably work with just that without needing to check whether layer creation option OVERWRITE exists.

rouault: > Am I using it wrong?

hum, you're right. With ogr2ogr, -lco OVERWRITE=YES is probably useless nowadays. Could potentially be useful for other uses of the OGR API.

> If so, is there a better option to do it? CARTODBFY is a more obvious choice to avoid future conflicts with other drivers.
EQUAL(m_poDstDS->GetDriver()->GetDescription(), "CARTO") would even be better

> It won't be pretty, but doing something like: if (dynamic_cast<OGRCARTOLayer>(layer))

Hum, except that GDAL builds without curl, or with CARTO explicitly disabled, don't have the CARTO driver built at all, and thus OGRCARTOLayer wouldn't be defined at all.

So you'd probably rather want GetLayerAndOverwriteIfNecessary() to have a new bool *pbAddOverwriteLCO , set it for the CARTO driver when -overwrite, and in that case, add OVERWRITE=YES to papszLCOTemp before the call to CreateLayer() L3728

Algunenano: > Hum, except that GDAL builds without curl, or with CARTO explicitly disabled, don't have the CARTO driver built at all, and thus OGRCARTOLayer wouldn't be defined at all.

```
#ifdef CARTO_ENABLED

Comment (Algunenano) [reply gdal/ogr/ogrsf_frmts/carto/ogr_carto.h]: I see that `ogr2ogr_lib.cpp` compilation isn't receiving the driver flags, so I'll try your way :smiley:

**Comment by Algunenano** [comment]
Thanks a lot for the comments, Even.

I've improved the PR with the comments and now it actually does what it says:
Before this PR:
```
CARTO: DeleteLayer(pref)
HTTP: Fetch(http://localhost:8080/user/dev/api/v2/sql)
HTTP: These POSTFIELDS were sent:q=DROP TABLE "pref"&api_key=cfd51af386ec59859a1415d51f9b1809c014d369
CARTO: RunSQL Response:{"rows":[],"time":0.019,"fields":{},"total_rows":0}
HTTP: Fetch(http://localhost:8080/user/dev/api/v2/sql)
HTTP: These POSTFIELDS were sent:q=CREATE TABLE "pref" ( cartodb_id SERIAL,"the_geom" VARCHAR,"pref" VARCHAR,"gid" VARCHAR,"prefcode" VARCHAR,"area" VARCHAR,"area_contents" VARCHAR,"area_youto" VARCHAR,"area_pop" VARCHAR,"area_bill" VARCHAR,PRIMARY KEY (cartodb_id) );DROP SEQUENCE IF EXISTS "pref_cartodb_id_seq" CASCADE;CREATE SEQUENCE "pref_cartodb_id_seq" START 1;ALTER SEQUENCE "pref_cartodb_id_seq" OWNED BY "pref".cartodb_id;ALTER TABLE "pref" ALTER COLUMN cartodb_id SET DEFAULT nextval('"pref_cartodb_id_seq"')&api_key=cfd51af386ec59859a1415d51f9b1809c014d369
CARTO: RunSQL Response:{"rows":[],"time":0.03,"fields":{},"total_rows":0,"notices":["drop cascades to default value for column cartodb_id of table pref"]}
```
With this PR: 
```
CARTO: srtext : string
CARTO: DeleteLayer(pref)
CARTO: Overwrite on creation (1)
HTTP: Fetch(http://localhost:8080/user/dev/api/v2/sql)
HTTP: These POSTFIELDS were sent:q=BEGIN; DROP TABLE IF EXISTS "pref";CREATE TABLE "pref" ( cartodb_id SERIAL,"the_geom" VARCHAR,"pref" VARCHAR,"gid" VARCHAR,"prefcode" VARCHAR,"area" VARCHAR,"area_contents" VARCHAR,"area_youto" VARCHAR,"area_pop" VARCHAR,"area_bill" VARCHAR,PRIMARY KEY (cartodb_id) );DROP SEQUENCE IF EXISTS "pref_cartodb_id_seq" CASCADE;CREATE SEQUENCE "pref_cartodb_id_seq" START 1;ALTER SEQUENCE "pref_cartodb_id_seq" OWNED BY "pref".cartodb_id;ALTER TABLE "pref" ALTER COLUMN cartodb_id SET DEFAULT nextval('"pref_cartodb_id_seq"'); COMMIT;&api_key=cfd51af386ec59859a1415d51f9b1809c014d369
CARTO: RunSQL Response:{"rows":[],"time":0.042,"fields":{},"total_rows":0,"notices":["drop cascades to default value for column cartodb_id of table pref"]}
````

PR #1444, opened 2019-04-16
Allow setting band-specific scale and offset values

bleutner:
Previously, only a single scale and/or offset value could be set by gdal_edit.py, which was applied to all layers. Now, one can also specify one value per band, like this:

`gdal_edit.py -scale 1 10 100 -offset 0 -10 10 ThreeLayerFile.tif`

This functionality is useful for files with different scaling factors, which can, for example, be produced with `gdal_translate -scale_1 -scale_2 ...`

### DIFF
--- a/gdal/swig/python/scripts/gdal_edit.dox
+++ b/gdal/swig/python/scripts/gdal_edit.dox
@@ -87,7 +87,9 @@ Assign a specified nodata value to output bands.</dd>
 Remove existing nodata values.</dd>
 
 <dt> <b>-scale</b> <i>value</i>:</dt><dd> (GDAL >= 2.2)
-Assign a specified scale value to output bands. If no scale is needed, it
+Assign a specified scale value to output bands. If a single scale value is provided it will be 
+set for all bands. Alternatively one scale value per band can be provided, in which case the 
+the number of scale values must match the number of bands. If no scale is needed, it
 it recommended to set the value to 1. Scale and Offset are generally used
 together. For example, scale and offset might be used to store elevations in 
 a unsigned 16bit integer file with a precision of 0.1, and starting from -100.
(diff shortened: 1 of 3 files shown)


gdal/swig/python/scripts/gdal_edit.dox
```diff
@@ -87,7 +87,9 @@ Assign a specified nodata value to output bands.</dd>
 Remove existing nodata values.</dd>
 
 <dt> <b>-scale</b> <i>value</i>:</dt><dd> (GDAL >= 2.2)
-Assign a specified scale value to output bands. If no scale is needed, it
+Assign a specified scale value to output bands. If a single scale value is provided it will be 
+set for all layers. Alternatively one scale value per band can be provided, in which case the
```
rouault: layers -> bands

gdal/swig/python/scripts/gdal_edit.dox
```diff
@@ -96,7 +98,9 @@ true_value = (pixel_value * scale) + offset
 Note: these values can be applied using -unscale during a gdal_translate run.</dd> 
 
 <dt> <b>-offset</b> <i>value</i>:</dt><dd> (GDAL >= 2.2)
-Assign a specified offset value to output bands. If no offset is needed, it
+Assign a specified offset value to output bands. If a single offset value is provided it will be 
+set for all layers. Alternatively one offset value per band can be provided, in which case the
```
rouault: layers -> bands

gdal/swig/python/scripts/gdal_edit.py
```diff
@@ -261,6 +268,22 @@ def gdal_edit(argv):
     if ds is None:
         return -1
 
+    if scale:
+        if len(scale) == 1:
+            scale = scale * ds.RasterCount 
+        elif len(scale) != ds.RasterCount:
+            print('If more than one scale value is provided, their number must match the number of layers.')
```
rouault: layers -> bands

gdal/swig/python/scripts/gdal_edit.py
```diff
@@ -261,6 +268,22 @@ def gdal_edit(argv):
...
+            print('')
+            return Usage()
+    
+    if offset:
+        if len(offset) == 1:
+            offset = offset * ds.RasterCount
+        elif len(offset) != ds.RasterCount:
+            print('If more than one offset value is provided, their number must match the number of layers.')
```
rouault: layers -> bands

gdal/swig/python/scripts/gdal_edit.dox
bleutner: changed "layers" to "bands"

PR #1464, opened 2019-04-20
Add Dockerfile for different configurations

rouault:
@hobu I'd like to explicitly put a license in the Dockerfile, if possible Public domain, so that anyone feel free to reuse as they wish. As this is derived from the PROJ Dockerfile you authored, I need your permission. And I believe you're the owner of https://hub.docker.com/u/osgeo . Would it be possible to add myself as a contributor ('rouault' login in hub.docker.com) so I can push osgeo/gdal packages ?-

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,10 @@ GDAL is an open source X/MIT licensed translator library for raster and vector g
 
 See [CONTRIBUTING.md](CONTRIBUTING.md)
 
+### Docker images
+
+See [gdal/docker/README.md](gdal/docker/README.md)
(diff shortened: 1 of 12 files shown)


hobu:
> As this is derived from the PROJ Dockerfile you authored, I need your permission. 

You have my permission to license this as MIT or whatever is appropriate.

hobu:
> Would it be possible to add myself as a contributor ('rouault' login in hub.docker.com) so I can push osgeo/gdal packages ?

done.

PR #1469, opened 2019-04-22
Add "-nogcp" option to gdal_translate binary and the python binding.

fechen123:
If "-nogcp" is specified, gdal_translate will ignore the GCPs in the source
dataset. By default, nogcp is false and the GCPs in the source dataset are
copied to the destination dataset.

## What does this PR do?
Add -nogcp option to gdal_translate. 

If "-nogcp" is specified, gdal_translate will ignore the GCPs in the source
dataset. By default, nogcp is false and the GCPs in the source dataset are
copied to the destination dataset.

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:

### DIFF
--- a/autotest/utilities/test_gdal_translate.py
+++ b/autotest/utilities/test_gdal_translate.py
@@ -838,6 +838,27 @@ def test_gdal_translate_38():
     ds = None
 
 ###############################################################################
+# Test -nogcp options
+
+
+def test_gdal_translate_39():
+    if test_cli_utilities.get_gdal_translate_path() is None:
+        pytest.skip()
+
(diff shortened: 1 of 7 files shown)


rouault:
Looks good globally. Can you also:
1) document the new option in gdal/apps/gdal_utilities.dox
2) commit the changes of gdal/swig/python/osgeo/gdal.py (generated with SWIG 3.0.8) ? (that's an arguable practice, but that's the current one for the project)

Issue #1479, opened 2019-04-26
TileDB reads


normanb:
TileDB reads were `GLOBAL_ORDER` https://github.com/OSGeo/gdal/blob/master/gdal/frmts/tiledb/tiledbdataset.cpp#L220 and were changed to be `ROW_MAJOR` to reflect a change in TileDB approach to reads. The tests pass because all of the pixels are there. However when using `gdal_translate` the output is not visually correct. 

I will work on a fix today, noting that this is a critical bug for the upcoming GDAL release.

rouault:
> this is a critical bug for the upcoming GDAL release.

Being a new functionality and not causing regression to existing workflows, I don't consider this warrants a RC2 just for that issue.

PR #1480, opened 2019-04-27
set row-major reads and removed adviseread

normanb:
https://github.com/OSGeo/gdal/issues/1479

This fixes the issue, `AdviseRead` used by `gdal_translate` was being misread and passed down to `ReadBlock`. This PR removes that issue and also moves the band dimension to the outer dimension to improve performance.


rouault:
Can you just rework the commit message to be: "TileDB: set row-major reads and removed adviseread (fixes #1479)"

PR #1506, opened 2019-05-03
EEDA: Bump to v1alpha.

saicheems:
- `EarthEngineAsset.path` is no longer returned.
- `PixelDataType.(INTEGER|FLOAT32|FLOAT64)` are renamed to
  `PixelDataType.(INT|FLOAT|DOUBLE)`.
- `EarthEngineAsset.time` is no longer returned (replaced by
  `EarthEngineAsset.(start|end)_time`.
- `ListImagesResponse.assets` is renamed to
  `ListImagesResponse.images`.

### DIFF
--- a/autotest/gdrivers/eedai.py
+++ b/autotest/gdrivers/eedai.py
@@ -82,7 +82,7 @@ def test_eedai_2():
             {
                 "id": "B1",
                 "dataType": {
-                    "precision": "INTEGER",
+                    "precision": "INT",
                     "range": {
                         "max": 65535
                     }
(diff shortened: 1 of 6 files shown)


rouault:
Will the https://earthengine.googleapis.com/v1/ API will still be supported ? Asking this if this needs to be backported to stable branches ?

saicheems:
Our intention is to turn down v1 in the next two months. It was an early-access release and incorrectly versioned. We'll be progressing to v1beta and finally back to v1 over the next few quarters.

rouault:
Backported to release/3.0 and release/2.4 branches

PR #1519, opened 2019-05-08
Adding cmath include in `ogr/ogrlinestring`

jepessen:
When building the project in Visual Studio 2019 I obtain an error when compiling ogr/ogrlinestring.cpp regarding the fact that `std::fabs` was missing. I've added the `cmath` include in order to fix it.

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS: Windows 10 Professional 64bit
* Compiler: Visual Studio 2019
* Tools: vcpkg for installing gdal

### DIFF
--- a/gdal/ogr/ogrlinestring.cpp
+++ b/gdal/ogr/ogrlinestring.cpp
@@ -35,6 +35,7 @@
 #include <cstdlib>
 #include <algorithm>
 #include <limits>
+#include <cmath>
 
 CPL_CVSID("$Id$")
 


rouault:
The <cmath> inclusion is already at line 34 in master. You are probably trying to fix an issue with GDAL 2.4.1 which has been fixed since in the release/2.4 branch per 22fa16e6e6de253438be943be1353187dca9b123 but there is not yet a 2.4.2 release

Issue #1287, opened 2019-02-12
Update GDAL NetCDF Vector Driver to support CF 1.8 geometries and Related Bugs / Feature Extensions


wchen329:
The GDAL NetCDF vector driver [1] is currently compliant for the CF-1.6 convention. The CF-1.8 convention should be released soon with several new features including Geometry support [2], which NetCDF vector driver could greatly benefit from if made aware. I plan to update the NetCDF Read/Write Driver and possibly other files needed to support CF 1.8 geometries.

I'll update this issue and open a pull request when I have contributions ready for the GDAL community to review. Further requirements for such a pull request to be accepted (i.e. testing requirements) would be helpful.

[1] https://www.gdal.org/frmt_netcdf_vector.html
[2] https://github.com/cf-convention/cf-conventions/blob/master/ch07.adoc#geometries

### Progress
- [x] Baseline Support for Reading CF-1.8 geometries into OGRFeatures (#1541)
- [x] Support for Writing from OGRFeatures into CF-1.8 compliant CF-netCDF Files (#1680)
- [x] Support for CF-1.8 Style Grid Mappings and Geo-References (#1680)
- [x] **Bug:** NetCDF-4 files (FORMAT=NC4) work for writing CF-1.8 point geometries, but otherwise fail for the other feature types. Resolve by: using infinite dimensions instead of fixed dimension size for NC4 specifically and making other changes
- [x] **Bug:** https://github.com/osgeo/gdal/issues/1287#issuecomment-513602854 (Yahara CRS and something with States). [PR]
- [x] **Feature Request:** Ability to remove dimensions M (in general with warning) or Z (if not particularly lossy?). [PR]
- [ ] **Optimization:** Do not pre-define dimensions for NC3, only define them upon first write.

rouault:
CC @tomkralidis  Perhaps you have some comments on this ? I'm not sure how critical it is to maintain the existing support for WKT-encoded geometries in netCDF if there is now a more standard replacement ? The WKT support was perhaps a bit more generic than the new CF stuff that only supports "classic" simple geometries, whereas with WKT we could support exotic things like CircularString, CompoundCurve, CurvePolygon, etc.

tomkralidis:
@rouault I think we're safe to move to the new convention (no backwards compat issues here).

piyushrpt:
We are currently using OGR features to tag raster products with bounding box polygons. But we will be able to adopt to once support for new conventions is in place.

[...]

PR #1541, opened 2019-05-17
Additions to netCDF Driver to allow Reading from CF-1.8 Encoded Geometries (#1287)

wchen329:
## What does this PR do?
This PR includes additions to the netCDF Vector Driver to allow support for retrieving OGRFeature types from a netCDF File encoded utilizing the CF-1.8 standard. It adds support the following six feature types (OGRPoint, OGRLineString, OGRPolygon) and "Multi" collections of those three that are mentioned in the following CF documentation: (https://github.com/cf-convention/cf-conventions/blob/master/ch07.adoc#geometries).

### What about backwards compatibility?
This PR preserves backwards compatibility with the earlier WKT representation used for earlier versions of this driver. The new CF-1.8 features are only activated if the input netCDF File if the `Conventions` attribute is of the following { `CF-1.x` | (x ≥  8) }. Otherwise, the driver will not even utilize the new additions to the driver, allowing for the driver to work as before. 

### Testing
Automatic testing is now available as a part of the netCDF-GDAL test suite. Please see the following directory for test cases: https://github.com/wchen329/gdal/tree/master/autotest/gdrivers/data/netcdf-sg

A few more test cases are still being added.

Issue #1287 

## Tasklist

 - [x] Implement Reading Capability of the Six Feature Types Previously 
 - [x] Implement OGRField Awareness
 - [x] Utilize Manual Testing to Verify Success Conditions
 - [x] Utilizing GDAL/OGR CPL Functions, Improve Resilience 
 - [x] Verify Cases of Malformed Input
 - [x] Further Optimize Memory Usage
 - [x] Finish rigging to all relevent points in netCDFLayer 
- [x] Add test case(s) to autotest
 - [x] Review
 - [x] Refactor Unused / Dead Code
 - [x] Adjust for comments
 - [x] Review (2)
 - [x] Adjust for comments (2)

## Environment

* OS: openSUSE Tumbleweed (Primary), Debian Sid (Secondary) Microsoft Windows 7 Professional AMD64 (Tertiary)
* Compiler: GCC 8.3.1 (Primary), MSVC++ 14.0 (Secondary)
* Standard AMD64/x86-64 Machine, openSUSE under VirtualBox Guest under Intel Core i7-3770 (3 "cores" accessible to Guest) [little endian], n]

## What Specifically Could Be Helpful
* ~~Information on getting automated tests running on CI~~
* ~~Information about detection of Big Endian through CPL interfaces or specific macros.~~
* Related to #1287 - This PR does not include any CF-1.8 writing functionality. This functionality is saved for a future PR. Unlike the readers, the WKT writer and a CF-1.8 version cannot exactly co-exist without one having precedence over the other. Given this would it not be too inconvenient to deprecate WKT writing (that can be activated through a special flag) and allow for a new CF-1.8 writer to be the default action?


wchen329:
6 basic test cases, and two "advanced" test cases, well-formed CF-1.8 formatted. (Advanced test cases: courtesy of @dblodgett)
[simple_test_cases.tar.gz](https://github.com/OSGeo/gdal/files/3193226/simple_test_cases.tar.gz)

dblodgett-usgs:
@rouault and others, we would appreciate some feedback on the approach that @wchen329 has taken here. Winor is working for us on a vairly short-term assignment and I hope we can get this to a place the community is happy with it soon, so feedback is very important.

Thanks!!

gdal/frmts/netcdf/netcdfsg.h
```diff
@@ -0,0 +1,299 @@
...
+	class SG_Exception_Dim_MM : public SG_Exception
+	{
+		char * err_msg;	
+		SG_Exception_Dim_MM(SG_Exception_Dim_MM&);
+		SG_Exception_Dim_MM& operator=(SG_Exception_Dim_MM&);
+
+		public:
+			char* get_err_msg() override { return err_msg; }
+
+		SG_Exception_Dim_MM(char* geometry_container, const char* field_1, const char *field_2);
```
rouault: use const char*

gdal/frmts/netcdf/netcdfsg.h
```diff
@@ -0,0 +1,299 @@
...
+	class SG_Exception_Existential: public SG_Exception
+	{
+		char * err_msg;
+		SG_Exception_Existential(SG_Exception_Existential&);
+		SG_Exception_Existential& operator=(SG_Exception_Existential&);
+
+		public:
+			char* get_err_msg() override { return err_msg; }
+		
+		SG_Exception_Existential(char* geometry_container, const char* missing_name);
```
rouault: use const char*

gdal/frmts/netcdf/netcdfsg.h
```diff
@@ -0,0 +1,299 @@
...
+	class SG_Exception_Dep: public SG_Exception
+	{
+		char * err_msg;
+		SG_Exception_Dep(SG_Exception_Dep&);
+		SG_Exception_Dep& operator=(SG_Exception_Dep&);
+
+		public:
+			char* get_err_msg() override { return err_msg; }
+		
+		SG_Exception_Dep(char* geometry_container, const char* arg_1, const char* arg_2);
```
rouault: use const char*

gdal/frmts/netcdf/netcdfsg.h
```diff
@@ -0,0 +1,299 @@
...
+	class SG_Exception_BadSum : public SG_Exception
+	{
+		char * err_msg;
+		SG_Exception_BadSum(SG_Exception_BadSum&);
+		SG_Exception_BadSum& operator=(SG_Exception_BadSum&);
+
+		public:
+			char* get_err_msg() override { return err_msg; }
+		
+		SG_Exception_BadSum(char* geometry_container, const char* arg_1, const char* arg_2);
```
rouault: use const char*

gdal/frmts/netcdf/netcdfsg.cpp
```diff
@@ -0,0 +1,1047 @@
...
+	{
+
+		char * cart = nullptr;
+		this->pt_buffer = nullptr;
+		memset(this->container_name, 0, NC_MAX_NAME+1);
+
+		// Get geometry container name
+		if(nc_inq_varname(ncId, geoVarId, container_name) != NC_NOERR)
+		{
+			throw new SG_Exception_Existential((char*)"new geometry container", "the variable of the given ID");
```
rouault: cast to remove once constructor signature is fixed

gdal/frmts/netcdf/netcdfsg.cpp
```diff
@@ -0,0 +1,1047 @@
...
+			if(nc_inq_varid(ncId, gm_name, &gmVID) == NC_NOERR)
+			{
+				this->gm_varId = gmVID;	
+			}
+		}	
+		
+		delete[] gm_name;
+
+		// Find a list of node counts and part node count
+		char * nc_name = nullptr; char * pnc_name = nullptr; char * inter_name = nullptr; char * ir_name = nullptr;
```
rouault: one declaration per line

gdal/frmts/netcdf/netcdfsg.cpp
```diff
@@ -0,0 +1,1047 @@
...
+		// Create parts count list and an offset list for parts indexing	
+		if(this->node_counts.size() > 0)
+		{
+			// to do: check for out of bounds
+			int ind = 0; int parts = 0; int prog = 0; int c = 0;
+			for(size_t pcnt = 0; pcnt < pnode_counts.size() ; pcnt++)
+			{
+				if(prog == 0) pnc_bl.push_back(pcnt);
+
+				if(int_rings.size() > 0) if(!int_rings[pcnt]) c++;
```
rouault: use && insteaf of 2 if

gdal/frmts/netcdf/netcdfsg.cpp
```diff
@@ -0,0 +1,1047 @@
...
+				}
+			} 
+		}
+
+		// (1) the touple order for a single point
+		// (2) the variable ids with the relevant coordinate values
+		char * dim = strtok(cart,  " ");
+		int axis_id = 0;
+		
+		while(dim != NULL)
```
rouault: all use of NULL to be replaced by nullptr

gdal/frmts/netcdf/netcdfsg.cpp
```diff
@@ -0,0 +1,1047 @@
...
+		{
+			if(nc_inq_varid(ncId, dim, &axis_id) == NC_NOERR)
+			{
+				this->touple_order++;
+				this->nodec_varIds.push_back(axis_id);
+			}
+			else
+			{
```
rouault: missing delete[] cart

gdal/frmts/netcdf/netcdfsg.cpp
```diff
@@ -0,0 +1,1047 @@
...
+		// Node Coordinates
+		if((cart = attrf(ncId, geoVarId, CF_SG_NODE_COORDINATES)) == nullptr)
+		{
+			throw new SG_Exception_Existential(container_name, CF_SG_NODE_COORDINATES);
+		}	
+
+		// Create parts count list and an offset list for parts indexing	
+		if(this->node_counts.size() > 0)
+		{
+			// to do: check for out of bounds
```
rouault: TODO to implement

gdal/frmts/netcdf/netcdfsg.cpp
```diff
@@ -0,0 +1,1047 @@
...
+			// check that all have same dimension
+			int inter_dim[1];
+			nc_inq_vardimid(ncId, nodec_varIds[nvitr], inter_dim);
+			
+			if(!dim_set)
+			{
+				all_dim = inter_dim[0];
+			}	
+
+			else{if (inter_dim[0] != all_dim) throw new SG_Exception_Dim_MM(container_name, "X, Y", "in general all node coordinate axes"); }
```
rouault: to put on several lines

gdal/frmts/netcdf/netcdfsg.cpp
```diff
@@ -0,0 +1,1047 @@
...
+			else{if (inter_dim[0] != all_dim) throw new SG_Exception_Dim_MM(container_name, "X, Y", "in general all node coordinate axes"); } 
+		}
+
+		// (2) check equality one
+		if(node_counts.size() > 0)
+		{
+			size_t diml;
+			nc_inq_dimlen(ncId, all_dim, &diml);
+		
+			if(diml != total_node_count) throw new SG_Exception_BadSum(container_name, "node_count", "node coordinate dimension length");
```
rouault: to put on seeral lines

gdal/frmts/netcdf/netcdfsg.cpp
```diff
@@ -0,0 +1,1047 @@
...
+				inPlaceSerialize_PolygonExtOnly(this, nc, sb, ret);
+				break;
+
+			case MULTIPOINT:
+				{
+					wkbSize = 1 + 4 + 4 + nc * (1 + 4 + 16);
+					ret = new int8_t[wkbSize];
+
+					void * worker = ret;
+					int8_t header = 1;
```
rouault: 1 is only valid on little-endian architecture. Same issue in other below occurences

gdal/frmts/netcdf/netcdfsg.cpp
```diff
@@ -0,0 +1,1047 @@
...
+			{
+				return 0;
+			}
+			return len;	
+		}
+
+		else return this->node_counts.size();
+	}
+
+	/* serializeToWKB(SGeometry * sg)
```
rouault: high level question: why doing hand-written wkb serialization and not using OGRGeometry and derived classes methods directly. Would be more straightforward

gdal/frmts/netcdf/netcdfsg.cpp
```diff
@@ -0,0 +1,1047 @@
...
+	{
+		// First check for container_id, if variable doesn't exist error out
+		if(nc_inq_var(this->nc, container_id, NULL, NULL, NULL, NULL, NULL) != NC_NOERR)
+		{
+			return;	// change to exception
+		}
+
+		// Now exists, see what variables refer to this one
+		// First get name of this container
+		char contname[NC_MAX_NAME];
```
rouault: +1 and 0 initialize ?

gdal/frmts/netcdf/netcdfsg.cpp
```diff
@@ -0,0 +1,1047 @@
...
+
+	// to get past linker
+	SG_Exception::~SG_Exception()
+	{
+
+	}
+
+	// Helpers
+	// following is a short hand for a clean up and exit, since goto isn't allowed
+	int getCFMinorVersion(int ncid)
```
rouault: what about if CF 2.X is released ? Would't it make sense to return a double with the CF version and test against that ? And using sscanf(attrval, "CF-%f", &version) could simplify a lot this function.

gdal/frmts/netcdf/netcdfsg.h
```diff
@@ -0,0 +1,299 @@
+#ifndef __NETCDFSG_H__
```
rouault: license & copyright missing

gdal/frmts/netcdf/netcdfsg.h
```diff
@@ -0,0 +1,299 @@
...
+		std::vector<int> pnode_counts;	// part node counts of each geometry in a container
+		std::vector<int> bound_list;	// a quick list used to store the real beginning indicies of shapes
+		std::vector<bool> int_rings;	// list of parts that are interior rings
+		std::vector<int> pnc_bl;	// a quick list of indicies for part counts corresponding to a geometry
+		std::vector<int> parts_count;	// a count of total parts in a single geometry instance
+		std::vector<int> poly_count;	// count of polygons, for use only when interior rings are present
+		int current_vert_ind;	// used to keep track of current point being used
+		size_t cur_geometry_ind;	// used to keep track of current geometry index
+		size_t cur_part_ind;		// used to keep track of current part index
+		Point * pt_buffer;	// holds the current point
```
rouault: std::unique_ptr

gdal/frmts/netcdf/netcdfsg.h
```diff
@@ -0,0 +1,299 @@
...
+	 * Returns: an error code if failure, 0 on success
+	 */
+	//int putGeometryRef(int ncid, SGeometry* geometry);
+
+	/* Fetches a one dimensional string attribute
+	 * using the given ncid, variable ID (varID), and attribute key (attrName)
+	 *
+	 * Returns: a NEW cstring of the attribute value, that must be separately deleted 
+	 */
+	char * attrf(int ncid, int varId, const char * attrName);
```
rouault: would be more simple to return a std::string to avoid dealing with memory deallocation

wchen329:
Thanks for the review @rouault! I'll look into those things you've pointed out / make changes requested.

Also about the big endian issue- this is something I pointed out earlier in a comment but I wasn't sure the most suitable method to go about it specifically in GDAL. Do you guys have a special CPL field set for this I can condition off of, or a predefined macro that I can use to check off of it?

gdal/frmts/netcdf/netcdfdataset.cpp
```diff
@@ -10916,6 +10932,7 @@ CPLErr netCDFDataset::FilterVars( int nCdfId, bool bKeepRasters,
             (EQUAL(osFeatureType, "profile") &&
              oMapDimIdToCount.size() == 2 && nProfileDimId >= 0)) )
         {
+	if(!bSGSupport)
```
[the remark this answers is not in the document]
wchen329: :+1: will fix

gdal/frmts/netcdf/netcdfdataset.cpp
```diff
@@ -10954,7 +10971,7 @@ CPLErr netCDFDataset::CreateGrpVectorLayers( int nCdfId,
                                              int nVarXId, int nVarYId, int nVarZId,
                                              int nProfileDimId,
                                              int nParentIndexVarID,
-                                             bool bKeepRasters )
+					      bool bKeepRasters )
```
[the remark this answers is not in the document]
wchen329: +1 will fix

gdal/frmts/netcdf/netcdfdataset.h
```diff
@@ -944,6 +964,7 @@ class netCDFDataset final: public GDALPamDataset
 
 class netCDFLayer final: public OGRLayer
 {
+	friend class netCDFDataset;
```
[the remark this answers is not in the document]
wchen329: :+1:   will fix

gdal/frmts/netcdf/netcdfdataset.h
```diff
@@ -994,6 +1015,10 @@ class netCDFLayer final: public OGRLayer
         nc_type         m_nWKTNCDFType;
         CPLString       m_osCoordinatesValue;
         std::vector<FieldDesc> m_aoFieldDesc;
+	std::vector<OGRFeature*> m_sgFeatureList;
```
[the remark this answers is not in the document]
wchen329: :+1:   will fix

gdal/frmts/netcdf/netcdfdataset.h
```diff
@@ -1036,6 +1062,8 @@ class netCDFLayer final: public OGRLayer
         void            SetWKTGeometryField(const char* pszWKTVarName);
         void            SetGridMapping(const char* pszGridMapping);
         void            SetProfile(int nProfileDimID, int nParentIndexVarID);
+	void		AddSimpleGeometryFeature(OGRFeature * sg) { this->m_sgFeatureList.push_back(sg); }
```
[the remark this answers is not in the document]
wchen329: :+1:   will fix

gdal/frmts/netcdf/netcdflayer.cpp
```diff
@@ -59,6 +60,8 @@ netCDFLayer::netCDFLayer(netCDFDataset *poDS,
         m_nWKTMaxWidthDimId(-1),
         m_nWKTVarID(-1),
         m_nWKTNCDFType(NC_NAT),
+	m_sgItrInit(false),
```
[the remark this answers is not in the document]
wchen329: :+1:   will fix

gdal/frmts/netcdf/netcdflayer.cpp
```diff
@@ -1062,6 +1069,10 @@ bool netCDFLayer::FillFeatureFromVar(OGRFeature *poFeature, int nMainDimId,
         }
     }
 
+    // For CF-1.8 simple geometry specifically 
+    // Only need fields to be set here
+    if( this-> m_HasCFSG1_8 ) return true;
```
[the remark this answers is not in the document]
wchen329: It's a bit of personal style to use this -> anywhere possible to prevent further possible ambiguities. But sure :+1:   will fix!

gdal/frmts/netcdf/netcdflayer.cpp
```diff
@@ -1173,8 +1184,20 @@ bool netCDFLayer::FillFeatureFromVar(OGRFeature *poFeature, int nMainDimId,
 
 OGRFeature *netCDFLayer::GetNextFeature()
 {
+    if(!this->m_sgItrInit && m_sgFeatureList.size() > 0)
```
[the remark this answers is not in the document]
wchen329: ditto but :+1:   will fix

gdal/frmts/netcdf/netcdflayer.cpp
```diff
@@ -2059,7 +2082,7 @@ OGRErr netCDFLayer::CreateField(OGRFieldDefn *poFieldDefn, int /* bApproxOK */)
             osVarName = CPLSPrintf("%s%d", poFieldDefn->GetNameRef(), i);
             status = nc_inq_varid(m_nLayerCDFId, osVarName, &nVarID);
             if( status != NC_NOERR )
-                break;
+            break;
```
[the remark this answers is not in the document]
wchen329: :+1:   will fix

rouault:
> Do you guys have a special CPL field

you can use #if CPL_IS_LSB . But wkb serialization could be entirely avoided by constructing directly OGRGeometry subclasses

gdal/frmts/netcdf/netcdflayersg.cpp
```diff
@@ -0,0 +1,194 @@
+/* NetCDF Simple Geometry specific implementations
```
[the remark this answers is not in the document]
wchen329: :+1:   will fix

gdal/frmts/netcdf/netcdflayersg.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+	for(size_t itr = 0; itr < vidList.size(); itr++)
+	{
+		try
+		{
+			LoadSGVarIntoLayer(ncid, vidList[itr]);
+
+		}
+		
+		catch(nccfdriver::SG_Exception * e)
```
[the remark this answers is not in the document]
wchen329: sure, we can do that :+1:   will fix

gdal/frmts/netcdf/netcdflayersg.cpp
```diff
@@ -0,0 +1,194 @@
...
+		{
+			CPLError(CE_Warning, CPLE_AppDefined, "Translation of a simple geometry layer has been terminated prematurely due to an error.");
```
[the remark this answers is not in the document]
wchen329: :+1:   will fix

gdal/frmts/netcdf/netcdflayersg.cpp
```diff
@@ -0,0 +1,194 @@
...
+			delete e;
+		}
+	}
+
+	return CE_None;
+}
+
+CPLErr netCDFDataset::LoadSGVarIntoLayer(int ncid, int nc_basevarId)
+{
+	nccfdriver::SGeometry * sg = new nccfdriver::SGeometry(ncid, nc_basevarId);
```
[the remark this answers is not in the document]
wchen329: thanks, good idea :+1:   will fix

gdal/frmts/netcdf/netcdflayersg.cpp
```diff
@@ -0,0 +1,194 @@
...
+	// Add properties
+	std::vector<int> props = pr.ids();
+	for(size_t itr = 0; itr < props.size(); itr++)
+	{
+		poL->AddField(props[itr]);	
+	}
+
+	for(size_t featCt = 0; featCt < shape_count; featCt++)
+	{
+		OGRGeometry * featureDesc;
```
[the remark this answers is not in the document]
wchen329: ditto :+1:   will fix

gdal/frmts/netcdf/netcdflayersg.cpp
```diff
@@ -0,0 +1,194 @@
...
+		delete sg;
+		throw new nccfdriver::SG_Exception_BadFeature();
+	}
+	
+
+	char baseName[NC_MAX_CHAR + 1];
+	memset(baseName, 0, NC_MAX_CHAR + 1);
+	nc_inq_varname(ncid, nc_basevarId, baseName);
+
+	netCDFLayer * poL = new netCDFLayer(this, ncid, baseName, owgt, nullptr);
```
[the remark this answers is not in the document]
wchen329: ditto :+1:   will fix

gdal/frmts/netcdf/netcdflayersg.cpp
```diff
@@ -0,0 +1,194 @@
...
+		catch(nccfdriver::SG_Exception*)
+		{
+			delete poL;
+			throw;
+		}
+
+		int r_size = 0;
+		void * wkb_rep = sg->serializeToWKB(featCt, r_size);
+		featureDesc->importFromWkb((const unsigned char*)wkb_rep, r_size, wkbVariantIso);
+		OGRFeature * feat = new OGRFeature(defn);
```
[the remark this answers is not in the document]
wchen329: ditto :+1:   will fix

gdal/frmts/netcdf/netcdflayersg.cpp
```diff
@@ -0,0 +1,194 @@
...
+		// This may be unncessary, given the check previously...
+		catch(nccfdriver::SG_Exception*)
+		{
+			delete poL;
+			throw;
+		}
+
+		int r_size = 0;
+		void * wkb_rep = sg->serializeToWKB(featCt, r_size);
+		featureDesc->importFromWkb((const unsigned char*)wkb_rep, r_size, wkbVariantIso);
```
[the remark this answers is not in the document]
wchen329: sure :+1:   will fix

gdal/frmts/netcdf/netcdflayersg.cpp
```diff
@@ -0,0 +1,194 @@
...
+		// Fill fields
+		for(size_t itr = 0; itr < props.size(); itr++)
+		{
+			// to do: add check to fill only as fields are defined	
+			poL->FillFeatureFromVar(feat, dimId, featCt);
+		}
+
+		feat -> SetFID(featCt);
+		char * wkb_rep_del = (char*)wkb_rep; // no compiler warning then
+		delete wkb_rep_del;
```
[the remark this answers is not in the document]
wchen329: nice catch! thanks :+1:   will fix (or maybe use malloc to simplify the aggressive casting)

gdal/frmts/netcdf/netcdflayersg.cpp
```diff
@@ -0,0 +1,194 @@
...
+		
+		// Fill fields
+		for(size_t itr = 0; itr < props.size(); itr++)
+		{
+			// to do: add check to fill only as fields are defined	
+			poL->FillFeatureFromVar(feat, dimId, featCt);
+		}
+
+		feat -> SetFID(featCt);
+		char * wkb_rep_del = (char*)wkb_rep; // no compiler warning then
```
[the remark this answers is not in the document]
wchen329: :+1:   will fix

gdal/frmts/netcdf/netcdflayersg.cpp
```diff
@@ -0,0 +1,194 @@
...
+		delete wkb_rep_del;
+		poL->AddSimpleGeometryFeature(feat);
+	}
+
+	papoLayers = (netCDFLayer**)realloc(papoLayers, (nLayers + 1) * sizeof(netCDFLayer *));
```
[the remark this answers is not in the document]
wchen329: :+1:   will fix

gdal/frmts/netcdf/netcdflayersg.cpp
```diff
@@ -0,0 +1,194 @@
...
+		}
+
+		feat -> SetFID(featCt);
+		char * wkb_rep_del = (char*)wkb_rep; // no compiler warning then
+		delete wkb_rep_del;
+		poL->AddSimpleGeometryFeature(feat);
+	}
+
+	papoLayers = (netCDFLayer**)realloc(papoLayers, (nLayers + 1) * sizeof(netCDFLayer *)); 
+	if(papoLayers == NULL) CPLError(CE_Failure, CPLE_OutOfMemory, "Allocation for additional netCDFLayer pointer failed.");
```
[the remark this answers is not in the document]
wchen329: okay :+1:   will fix

gdal/frmts/netcdf/netcdfsg.cpp
```diff
@@ -0,0 +1,1047 @@
+// Implementations of netCDF functions used for
```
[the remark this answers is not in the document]
wchen329: :+1:   will fix

gdal/frmts/netcdf/netcdfsg.cpp
```diff
@@ -0,0 +1,1047 @@
...
+#include "netcdfsg.h"
+namespace nccfdriver
+{
+	/* Re-implementation of mempcpy
+	 * but compatible with libraries which only implement memcpy
+	 */
+	static void* memcpy_jump(void *dest, const void *src, size_t n)
+	{
+		memcpy(dest, src, n);
+		char * a = (char*)dest + n; // otherwise unneccesary casting to stop compiler warnings
```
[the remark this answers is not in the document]
wchen329: :+1:   will fix

gdal/frmts/netcdf/netcdflayer.cpp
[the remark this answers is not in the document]
rouault: When using the m_ convention for member variables, this avoid any ambiguity w.r.t local variables

wchen329:
> 
> 
> > Do you guys have a special CPL field
> 
> you can use #if CPL_IS_LSB . But wkb serialization could be entirely avoided by constructing directly OGRGeometry subclasses

Thanks @rouault maybe this could be a target of a future refactor! Directly writing out a memory region seemed at the time an easy way to navigate OGR so I'll use CPL_IS_LSB for now.

gdal/frmts/netcdf/netcdfsg.h
wchen329: :+1: will fix

gdal/frmts/netcdf/netcdfsg.h
wchen329: :+1: will fix

gdal/frmts/netcdf/netcdfsg.h
wchen329: :+1: will fix

gdal/frmts/netcdf/netcdfsg.h
wchen329: :+1: will fix

gdal/frmts/netcdf/netcdfsg.cpp
wchen329: :+1: will fix

gdal/frmts/netcdf/netcdfsg.cpp
wchen329: fine :+1: will fix

gdal/frmts/netcdf/netcdfsg.cpp
wchen329: fine :+1: will fix

gdal/frmts/netcdf/netcdfsg.cpp
wchen329: sure can do that :+1: will fix

gdal/frmts/netcdf/netcdfsg.cpp
wchen329: thanks nice catch! :+1: will fix

gdal/frmts/netcdf/netcdfsg.cpp
wchen329: oops :+1: will fix

gdal/frmts/netcdf/netcdfsg.cpp
wchen329: sure :+1: will fix

gdal/frmts/netcdf/netcdfsg.cpp
wchen329: sure :+1: will fix

gdal/frmts/netcdf/netcdfsg.cpp
wchen329: yeah I will make fixes according to your comment below, thanks! :+1:

gdal/frmts/netcdf/netcdfsg.cpp
wchen329: ooh well see you guys have a very intricate class hierarchy and thought that doing this type of byte by byte low level serialization at the time would be easier to do navigate then go through each specific part of the class hierarchy, though to be honest I usually go low level if it's not a big burden ;)

gdal/frmts/netcdf/netcdfsg.cpp
wchen329: ooh thanks I didn't see that :+1: will fix

gdal/frmts/netcdf/netcdfsg.cpp
wchen329: I'm not sure if we can say that CF-1.8 geometries will be usable in the CF-2.x convention, as 2.x would be a new major release potentially breaking this type of encoding. But sure I can change it so that it works with >= 1.8 CFs :+1: will fix

gdal/frmts/netcdf/netcdfsg.h
wchen329: :+1: will fix

gdal/frmts/netcdf/netcdfsg.h
wchen329: ditto :+1: will fix

gdal/frmts/netcdf/netcdfsg.h
wchen329: like the C++-styled recommendations you've given! :+1: okay will fix

rouault:
The tests are currently a bit light. They should verify at least the content of a feature for each geometry type.

wchen329:
> 
> 
> The tests are currently a bit light. They should verify at least the content of a feature for each geometry type.

Oh yeah definitely, these were an initial attempt at working in tests into the autotest makefile, but now that I have a better idea of making these work I'll make sure to push up more intricate tests (including a full netCDFLayer test most likely).

gdal/frmts/netcdf/netcdflayer.cpp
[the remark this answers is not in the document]
wchen329: But `this` is part of the language directly :)

gdal/frmts/netcdf/netcdflayersg.cpp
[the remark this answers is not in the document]
wchen329: okay so I tried using a unique_ptr here and I got a seg. fault. GDB shows that the seg. fault occurs in the translation phase. Besides, according to the docs, ownership is handed off to the SetGeometryDirectly function so I'm going to for now keep it as a standard unmanaged pointer OGRGeometry*.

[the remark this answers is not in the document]
rouault: you must do SetGeometryDirectly(the_geom.release()) to transfer ownership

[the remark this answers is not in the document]
wchen329: Oh okay interesting thanks, I'll take another look then later :)

gdal/frmts/netcdf/netcdflayersg.cpp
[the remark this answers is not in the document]
wchen329: Actually can I  do by pointer? I remembered that I decided to use pointer to take advantage of polymorphism of different exception objects.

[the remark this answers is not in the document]
wchen329: I also have a universal catch clause at the bottom that deletes the exception (the other one just rethrowing to that one) so the pathing is not that complex :)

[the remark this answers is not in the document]
rouault: Polymorphism just works fine with exception thrown by value
The following displays yes!
```
struct A {};
struct B: public A {};
#include <stdio.h>
int main()
{
    try
    {
        throw B();
    }
    catch( const A& )
    {
        printf("yes!\n");
    }
    return 0;
}
```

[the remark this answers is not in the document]
wchen329: Well sure, sure... this I already knew.

But my implementation of SGeometry_Exception is fully virtual / abstact so I'm getting complaints about how SGeometry_Exception is not instantiable.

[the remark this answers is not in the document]
rouault: If you catch by const ref, there's no reason it doesn't work. Below code just works:
```
struct A { virtual void foo() = 0; };
struct B: public A { void foo() override {} };
#include <stdio.h>
int main()
{
    try
    {
        throw B();
    }
    catch( const A& )
    {
        printf("yes!\n");
    }
    return 0;
}
```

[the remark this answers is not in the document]
wchen329: Hmm, maybe I forgot the const, thanks @rouault, I'll try again.

`offtopic: I also just like using pointers way more, in general since I do a decent amount of C`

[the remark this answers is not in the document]
wchen329: Strangely enough I tried it again and this time just using reference (also not needed const reference) works! Didn't know why it didn't work the first time...

wchen329:
Also just a general question about the license and copyright. Since this might be getting pushed into GDAL. Who is this copyrighted to in this context (1) [I'll assume myself but _sometimes_ that's not the case when working in a large project] and (2) also can I choose any license so long as it is compatible with GDAL/OGR's MIT license?

rouault:
> I'll assume myself but

yourself, or your employer depending on your contracting arrangements.

> an I choose any license so long as it is compatible with GDAL/OGR's MIT license?

No, for the sake of license simplicity, please stick with GDAL X/MIT license

wchen329:
> 
> 
> > I'll assume myself but
> 
> yourself, or your employer depending on your contracting arrangements.
> 
> > an I choose any license so long as it is compatible with GDAL/OGR's MIT license?
> 
> No, for the sake of license simplicity, please stick with GDAL X/MIT license

@roualt alright thanks! I'll put myself down and just copy one of your headers around.

gdal/frmts/netcdf/netcdflayersg.cpp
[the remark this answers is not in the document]
wchen329: On the note of `unique_ptr` for the OGRGeometry * geometry, netCDFLayer * poL:

I took a look into `unique_ptr` looking through the [MSDN Standard Library Reference](https://docs.microsoft.com/en-us/cpp/standard-library/unique-ptr-class?view=vs-2019), which says that the object referred to by the `unique_ptr` is destroyed when the `unique_ptr` itself is destroyed (in the same manner as the legacy `auto_ptr` which I'm more familiar with). Unfortunately for a auto store / stack allocated `unique_ptr` as what is described here for poL and geometry, their `unique_ptr` s will be destroyed at the end of the function (on frame deconstruction), destroying the respective OGRGeometry or poL with them at the same time.

This being the case, I have opted to utilize OGR interface directly by transfering ownership, since sending something managed throughout the rest of OGR would probably require refactoring upstream into OGRGeometry and OGRLayer, probably by switching around or adding such interfaces using `unique_ptr`s or probably more fittingly using the `shared_ptr`. Other workarounds can be done, but as far as I know they will probably require explicit unmanaged heap usage anyway (defeating the purpose of a managed pointer in the first place).

wchen329:
@rouault So at this point I think I'm almost to the point where I can get all builds and all tests to pass. I have a few considerations that perhaps you could comment on:

**(1)** Memory safety: When I valgrind a normal execution of ogr2ogr (i.e. a successful translation of a CF-1.8 NC file into a JSON) the amount of leakage I get is equal to the baseline of the netCDF Library (opening and closing a dataset and terminating that program with no explicit dynamic allocation): 2,161 bytes. This being the case may I resolve the last two pieces of feedback? Whatever I have allocated looks like they get cleaned up upstream and do not require (or needs not to be) handled by "smart" or managed pointers.

Also, I'm not sure if this is actually some unwritten rule of unique_ptrs or I'm supposed to follow but didn't or a standard library bug. I'm getting mismatched operators for delete vs. delete[]

(this is from using Yahara_alb.nc by the way, which has 71 features)

==22181== ERROR SUMMARY: 71 errors from 1 contexts (suppressed: 0 from 0)
==22181== 
==22181== 71 errors in context 1 of 1:
==22181== Mismatched free() / delete / delete []
==22181==    at 0x483908B: operator delete(void*, unsigned long) (in /usr/lib64/valgrind/vgpreload_memcheck-amd64-linux.so)
==22181==    by 0x50C1B39: operator() (unique_ptr.h:81)
==22181==    by 0x50C1B39: ~unique_ptr (unique_ptr.h:274)
==22181==    by 0x50C1B39: netCDFDataset::LoadSGVarIntoLayer(int, int) (netcdflayersg.cpp:170)
==22181==    by 0x50C1F8C: netCDFDataset::DetectAndFillSGLayers(int) (netcdflayersg.cpp:84)
==22181==    by 0x50B68CA: netCDFDataset::Open(GDALOpenInfo*) (netcdfdataset.cpp:7143)
==22181==    by 0x52C18C6: GDALOpenEx (gdaldataset.cpp:3327)
==22181==    by 0x4024B4: main (ogr2ogr_bin.cpp:338)
==22181==  Address 0x74231d0 is 0 bytes inside a block of size 1,910 alloc'd
==22181==    at 0x483850F: operator new[](unsigned long) (in /usr/lib64/valgrind/vgpreload_memcheck-amd64-linux.so)
==22181==    by 0x50C2E50: nccfdriver::SGeometry::serializeToWKB(unsigned long, int&) (netcdfsg.cpp:702)
==22181==    by 0x50C1A54: netCDFDataset::LoadSGVarIntoLayer(int, int) (netcdflayersg.cpp:170)
==22181==    by 0x50C1F8C: netCDFDataset::DetectAndFillSGLayers(int) (netcdflayersg.cpp:84)
==22181==    by 0x50B68CA: netCDFDataset::Open(GDALOpenInfo*) (netcdfdataset.cpp:7143)
==22181==    by 0x52C18C6: GDALOpenEx (gdaldataset.cpp:3327)
==22181==    by 0x4024B4: main (ogr2ogr_bin.cpp:338)

Now according to this trace the memory is being allocated with new[] in serializeTOWKB, so it seems to be the pointer being assigned to the `wkb_rep` unique_ptr.

Also according to this trace the memory is being handled and consequently deleted by a unique_ptr, and that the unique_ptr is using the wrong deleter. Unfortunately, the deleter for the unique_ptr gets automatically assigned and I can't do much about it (as far as I know) except for make a wrapping class around it. So this is how I'm going to probably work around it.

The full valgrind trace is attached.
[ogr2ogr_valgrind.log](https://github.com/OSGeo/gdal/files/3242511/ogr2ogr_valgrind.log)

**(2)** I haven't really mentioned this and have been kind of ignoring this but 3D-support is kind of non-existent in the pull request. The scaffolding is all (or mostly) there, I just need to add a few more memory copies to serialize and allow some switching between each wkb type and its z variant so it wouldn't be a big deal... by the way, I tried to do this at first but wkbPointZ, wkbLineStringZ, wkbPolygonZ [don't seem a part of OGRwkbGeometryType](https://gdal.org/doxygen/ogr__core_8h.html#a800236a0d460ef66e687b7b65610f12a). Is it as easy as adding these values to this enum with the correct values corresponding to the wkb docs as OGR surely has support for 3D, or what would you recommend for getting importFromWkb to recognize something Points, LineStrings with X, Y, Z but no M axis?

**(3)** SRS setting. I at one point tried to get this going through the existing implementation, but only one SRS can be set at a time for a whole dataset in the existing driver (at least, going through the routine that scans for the SRS). It might require a little extra refactoring then. Can, I for now, simply display a warning about this SRS limitation and get back around to it in another PR? Especially since [CF-1.8 in the near future is likely to support multiple different grid_mappings between geometry_containers](https://github.com/cf-convention/cf-conventions/issues/156).

rouault:
> Memory safety

By default, std::unique_ptr deletes with `delete`, but as you allocated with `new[]`, there's indeed a mismatch. It is possible to specify a particular deleter function as the second argument of the unique_ptr template. But in that cas instead of a uique_ptr, if serializeToWKB() would return a std::vector<unsigned char>, that would be even simpler.

> wkbPointZ, wkbLineStringZ, wkbPolygonZ don't seem a part of OGRwkbGeometryType

They exist as wkbPoint25D, etc for historical reasons (predate the time where 3D was added to the spec). The same as their Z equivalents, except the WKB geometry type number is a bit different.

> Can, I for now, simply display a warning about this SRS limitation and get back around to it in another PR? 
That or indicate the limitation in the driver documentation. Due to the documentation refactoring, it now lives in doc/source/drivers/vector/netcdf.rst

wchen329:
> By default, std::unique_ptr deletes with `delete`, but as you allocated with `new[]`, there's indeed a mismatch. It is possible to specify a particular deleter function as the second argument of the unique_ptr template. But in that cas instead of a uique_ptr, if serializeToWKB() would return a std::vector, that would be even simpler.
> They exist as wkbPoint25D, etc for historical reasons (predate the time where 3D was added to the spec). The same as their Z equivalents, except the WKB geometry type number is a bit different.
> That or indicate the limitation in the driver documentation. Due to the documentation refactoring, it  now lives in doc/source/drivers/vector/netcdf.rst
@rouault thanks for the immediate suggestions! I wasn't aware of this ability to specifiy deleters with `unique_ptr`, thanks for the "pointers" :) on that. I'll go with what you have suggested.

Also, about tabulations in general, are they really disallowed even though in some files I have done them in a consistent manner? If this is the case, I can global search and replace them in these files, making them four spaces.

rouault:
>  about tabulations in general, are they really disallowed

yes, https://gdal.org/development/rfc/rfc8_devguide.html

wchen329:
@rouault While I'm waiting for the last build to pass, what would say is still required for this PR to pass your review (since I made a few other changes since your initial review)? What other changes should be made to get this thing finally pulled up?

gdal/doc/source/drivers/vector/netcdf.rst
```diff
@@ -123,6 +131,13 @@ written as netCDF v4 variable length strings.
 
 Geometry
 ~~~~~~~~
+Supported feature types when reading from a CF-1.8 convention compliant netCDF file
+include OGRPoint, OGRLineString, OGRPolygon, OGRMultiPoint, OGRMultiLineString, and
+OGRMultiPolygon. Due to slight ambiguities present in the CF-1.8 convention concerning
+Polygons versus MultiPolygons, the driver will always default to assuming a MultiPolygon
+is the geometry of a layer unless the attribute **part_node_count** is not present
+within that layers corresponding geometry container. Per convention requirements, the driver
+support reading from geometries with X, Y, and Z axes.
```
rouault: supports

gdal/frmts/netcdf/netcdflayersg.cpp
```diff
@@ -0,0 +1,215 @@
...
+        }
+        // This may be unncessary, given the check previously...
+        catch(nccfdriver::SG_Exception&)
+        {
+            delete poL;
+            throw;
+        }
+
+        int r_size = 0;
+        std::unique_ptr<unsigned char, std::default_delete<unsigned char[]>> wkb_rep = std::unique_ptr<unsigned char, std::default_delete<unsigned char[]>>(sg->serializeToWKB(featCt, r_size));
```
rouault: can be shortened as:
auto wkb_rep = ...
or
std::unique_ptr<...> wkb_rep(sg->serializeToWKB(featCt, r_size));

rouault:
I've tested a bit the driver with the test_ogrsf utility (`cd apps; make test_ogrsf` to build it) and it reports a few errors, due to the attribute filter and spatial filters being ignored:
```
$ test_ogrsf -ro ../autotest/gdrivers/data/netcdf-sg/Yahara_alb.nc
INFO: Open of `../autotest/gdrivers/data/netcdf-sg/Yahara_alb.nc' using driver `netCDF' successful.
INFO: Testing layer geometry_container.
INFO: Feature count verified.
INFO: Feature/layer spatial ref. consistency verified.
INFO: Spatial filter inclusion seems to work.
ERROR: Attribute filter seems to be ignored in GetFeatureCount() when spatial filter is set.
ERROR: Spatial filter (0) failed to eliminatea feature unexpectedly!
INFO: Spatial filter is ignored by GetFeature() as expected.
INFO: Infinity spatial filter works as expected.
INFO: Huge coords spatial filter works as expected.
INFO: Attribute filter inclusion seems to work.
ERROR: GetFeatureCount() may not be taking attribute filter into account (nInclusiveCount = 71, nExclusiveCount = 71, nExclusiveCountWhileIterating = 0, nTotalCount = 71).
INFO: Attribute filter is ignored by GetFeature() as expected.
INFO: GetExtent() test passed.
INFO: Random read test passed.
INFO: SetNextByIndex() read test passed.
ERROR: ExecuteSQL() should have returned a layer without features.
ERROR: ExecuteSQL() should have returned a layer without features.
INFO: TestVirtualIO successful.
```
You should likely move the new code you've added in netCDFLayer::GetNextFeature() to the top of netCDFLayer::GetNextRawFeature() instead

rouault:
`ogrinfo ../autotest/gdrivers/data/netcdf-sg/point_test.nc` reports 2 layers names_geometry and point_test with identical content. Only one of them should be reported.
Similarly with line_test.nc that reports names_geometry and line_test, but only the first one has geometries.

wchen329:
> You should likely move the new code you've added in netCDFLayer::GetNextFeature() to the top of netCDFLayer::GetNextRawFeature() instead

> `ogrinfo ../autotest/gdrivers/data/netcdf-sg/point_test.nc` reports 2 layers names_geometry and point_test with identical content. Only one of them should be reported.
> Similarly with line_test.nc that reports names_geometry and line_test, but only the first one has geometries.

Huh interesting, I'll go look at this and see what's up; thanks for finding these issues! I'll go update the test cases with some layer counting assertions too.

gdal/doc/source/drivers/vector/netcdf.rst
wchen329: done

gdal/frmts/netcdf/netcdflayersg.cpp
wchen329: done

wchen329:
Okay so I found the culprit, some of the heuristics for the old implementation of vector layers kept detecting a variable (or some?) as Vector layers incompletely, so I disabled these if the dataset has CF >= 1.8. This also means however, that mixing a legacy **Vector** dataset (Rasters should be still okay) with CF-1.8 geometries is not possible anymore.

Would it be beneficial if I go more fine-grained and manipulate the heuristics themselves and re-enable the ability read from both vector layers that are CF-1.8 and old-WKT style? My latest push has made them mutually exclusive.

wchen329:
By the way some random builds (see Travis above) are failing due to tests from other drivers or some random unrelated test seg. faulting (testing code that I didn't write), in these events, would it be useful to...
1) Document these tests as subjects for further PRs
2) Possibly restart the test? In some sense I feel this is wasting your guys' resources but it'd be nice to get all green checks

rouault:
> This also means however, that mixing a legacy **Vector** dataset with CF-1.8 geometries is not possible anymore.

I don't see this as a problem. I guess in the future we might want to generate CF-1.8 geometries, and completely deprecate/remove the WKT approach

> By the way some random builds (see Travis below) are failing

Yes, those random failures are annoying. Part of them due to the CI insfrastructure itself. Some of them unexplained, perhaps coming from GDAL itself but hard to reproduce locally

wchen329:
> I don't see this as a problem. I guess in the future we might want to generate CF-1.8 geometries, and completely deprecate/remove the WKT approach
> 
Alright, :+1: so in that case I guess maybe this is the most elegant solution. Though, on the note of writing CF-1.8 geometries (and I guess this is more suited for #1287) if I were to make another PR for it, would you say that hiding the WKT implementation behind a command line option would be the most appropriate thing to do?
>
> Yes, those random failures are annoying. Part of them due to the CI insfrastructure itself. Some of them unexplained, perhaps coming from GDAL itself but hard to reproduce locally
>
Some of this behavior at least seem a little characteristic of intermittent errors when being reckless with C-style memory management and so perhaps may be produced by buffer overflows? I agree though tracking down these type of errors with CI may be kind of spotty, given the fact that these errors kind of randomly pop up.

By the way what other changes would you say need to be made to this PR before I perhaps try to get another go at CI?

rouault:
> would you say that hiding the WKT implementation behind a command line option would be the most appropriate thing to do?

A layer creation option, yes. The existing FEATURE_TYPE one could be extended likely. But honestly, I'm not sure about the point of writing old-style WKT based geometries in WKT once CF-1.8 support is there. I'd largely prefer to see removal of write support for WKT, which is an obscure feature not even used by the people who made me implement it. And at some later point, we could drop the support for WKT reading. The less code, the better.

> C-style memory management and so perhaps may be produced by buffer overflows?

Perhaps, although the gcc52_stdcpp14_sanitize CI target runs with -fsanitize=undefined -fsanitize=address . So it tracks most such issues. Failures seem to occur mostly on the trusty_32bit and osx configurations, which might have their own pecularities

> By the way what other changes would you say need to be made to this PR before I perhaps try to get another go at CI?

I'd say we should probably be good once you've addressed my today comments.

wchen329:
> A layer creation option, yes. The existing FEATURE_TYPE one could be extended likely. But honestly, I'm not sure about the point of writing old-style WKT based geometries in WKT once CF-1.8 support is there. I'd largely prefer to see removal of write support for WKT, which is an obscure feature not even used by the people who made me implement it. And at some later point, we could drop the support for WKT reading. The less code, the better.
> 
Wow, so the WKT portion is underutilized to that extent. The stuff behind old WKT write capabilities do indeed seem quite a bit to maintain. Maybe another refactoring project :)
> Perhaps, although the gcc52_stdcpp14_sanitize CI target runs with -fsanitize=undefined -fsanitize=address . So it tracks most such issues. Failures seem to occur mostly on the trusty_32bit and osx configurations, which might have their own pecularities
>
Hmm indeed a mystery then. But yeah! Those two builds seem to be the ones that randomly segfault.
> I'd say we should probably be good once you've addressed my today comments.
Apologies if I read too fast, but I'm not sure which comments from today you are referring to. I did resolve the issues from earlier review however.

rouault:
>  but I'm not sure which comments from today you are referring to

Ah, I see you've fixed them by your later commits. Restarting the OSX build, and we should be good

wchen329:
Alright, I'll just do a resync with OSGeo/master to retrigger the build :+1:

PR #1553, opened 2019-05-21
Add quadratic average sampling method to overview computation

mgoacolou:
## What does this PR do?
This add a new average algorithm witch, for SAR product make more visible small brillant object in high level overviews
  
None

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:

### DIFF
--- a/gdal/gcore/overview.cpp
+++ b/gdal/gcore/overview.cpp
@@ -370,6 +370,8 @@ GDALResampleChunk32R_AverageT( double dfXRatioDstToSrc,
     // AVERAGE_BIT2GRAYSCALE
     const bool bBit2Grayscale =
         CPL_TO_BOOL( STARTS_WITH_CI( pszResampling, "AVERAGE_BIT2G" ) );
+    const bool bQuadraticAverage =
+        CPL_TO_BOOL( STARTS_WITH_CI( pszResampling, "AVERAGE_QUADRA" ) );
     if( bBit2Grayscale )
         poColorTable = nullptr;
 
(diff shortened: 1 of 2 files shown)


gdal/gcore/overview.cpp
```diff
@@ -526,16 +528,24 @@ GDALResampleChunk32R_AverageT( double dfXRatioDstToSrc,
...
                 for( int iDstPixel = 0; iDstPixel < nDstXWidth; ++iDstPixel )
                 {
-                    const Tsum nTotal =
-                        pSrcScanlineShifted[0]
-                        + pSrcScanlineShifted[1]
-                        + pSrcScanlineShifted[nChunkXSize]
-                        + pSrcScanlineShifted[1+nChunkXSize];
+                    if(bQuadraticAverage)
+                        nTotal =
+                            pSrcScanlineShifted[0]*pSrcScanlineShifted[0]
```
mloskot: You should try to follow existing style of whitespacing: `pSrcScanlineShifted[0] * pSrcScanlineShifted[0]`

gdal/gcore/overview.cpp
```diff
@@ -619,9 +636,18 @@ GDALResampleChunk32R_AverageT( double dfXRatioDstToSrc,
                         if( nVal >= 0 && nVal < nEntryCount &&
                             aEntries[nVal].c4 )
                         {
-                            nTotalR += aEntries[nVal].c1;
-                            nTotalG += aEntries[nVal].c2;
-                            nTotalB += aEntries[nVal].c3;
+                            if(bQuadraticAverage)
+                            {
+                                nTotalR += aEntries[nVal].c1*aEntries[nVal].c1;
```
mloskot: `aEntries[nVal].c1 * aEntries[nVal].c1;` to keep style consistent within the file

gdal/gcore/overview.cpp
mgoacolou: ok

rouault:
Sorry I somehow missed this pull request. Could you add a test case for the new method in autotest/gcore/tiff_ovr.py for example ?
Regarding the complex number case, pardon my ignorance, I was wondering if there's really some "physical" significance in the real & imaginary part of the results. I mean the maths looks right, but what is the interpretation of that number ? Wouldn't for example the quadratic average of the modulus of the numbers be something more natural ? Even in the real case, the quadratic average of a set of negative numbers is a positive one

PR #1554, opened 2019-05-22
RFC 74: Migrate gdal.org to Sphinx

hobu:
Per [RFC 74](https://trac.osgeo.org/gdal/wiki/rfc74_sphinx), we are migrating the website to use Sphinx for primary organization of the docs.

https://trac.osgeo.org/gdal/wiki/rfc74_sphinx

Once this PR is squash-merged we will need to do the following:

- [ ] Create AzP for osgeo/gdal
- [ ] Verify build loop is working for osgeo/gdal repo
- [ ] Verify GitHub pages is working at osgeo.github.io after removing gdal.dev custom domain on osgeo/gdal-docs repository
- [ ] Move DNS entries of gdal.org to GitHub Pages (@hobu can do this. The TTL is 1 hr)
- [ ] Verify DNS worldwide 
- [ ] Decommission gdal.dev


rouault:
@mloskot Looking at the .azure-pipelines.yml , I see the deploy steps are done unconditionally. We would want the doc building to be done for regular push builds as well as pull requests, but the deploy steps should only be done for push builds and only for the master branch. Is that possible ?

mloskot:
@rouault Let's see if 7b2c92181c1007b51f35c3af9604daf45db100f9 will do the job

hobu:
The recent java and python patch didn't deploy properly. Presumably we would add something to build these on-demand as well.

rouault:
> The recent java and python patch didn't deploy properly. Presumably we would add something to build these on-demand as well.

yes, I'm (slowly) repushing a new version of the osgeoproj-docs with wget in it. And yes for now, it will deploy a snapsort of the Python and Java docs. They are respectively produced from "make epydoc" from swig/python and "./make_doc.sh" from swig/java, but those (at least the java part) have other build requirements, so for now I took the fast way

mloskot:
🍻

PR #1560, opened 2019-05-23
doc: Update name of user/index reference

chambbj:
## What does this PR do?

The sidebar link to the "User oriented documentation" page at user/index says only "User" which isn't very intuitive for new users. Perhaps "Architecture" as suggested in this PR or "Using GDAL" would be slightly more descriptive.

N/A

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:

### DIFF
--- a/gdal/doc/source/index.rst
+++ b/gdal/doc/source/index.rst
@@ -12,7 +12,7 @@ This documentation is also available as a `PDF file <gdal.pdf>`_.
     programs/index
     drivers/raster/index
     drivers/vector/index
-    User <user/index>
+    Architecture <user/index>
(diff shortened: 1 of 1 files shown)


mloskot:
It's me who should take the blame for this. I suggested to use short names, `User` and `Developer`, for nodes with resources dedicated to the users and the contributors respectively. I have no strong opinion though :)

rouault:
This PR is stalled. Any follow up action planned ?

PR #1577, opened 2019-05-24
move about text to index page, provide license as dedicated page

tomkralidis:


rouault:
I presume license.rst is missing ?
One issue is that the PDF document will lack the GDAL general introduction paragraph. That one should be put in a dedicate .rst file included by index.rst and index_pdf.rst. Or perhaps leave the about.rst page for the purpose of the PDF, and make it include the general text and license.rst

tomkralidis:
@rouault I pushed some updates based on your comments (keeping about.rst, adding license.rst).

rouault:
Merged with an additional commit to fix/improve PDFoutput

PR #1584, opened 2019-05-27
NMAKE: Allow extra customisation points for Kakadu config

mloskot:
## What does this PR do?

* Allow users to tweak/append extras to `KAKINC`
* Set `KAKADU_7_5_OR_LATER=YES` by default, unless user already defined it.

## Tasklist

 - [ ] Review
 - [x] Adjust for comments

## Environment

* OS: Windows
* Compiler: MSVC + NMAKE

### DIFF
--- a/gdal/frmts/jp2kak/makefile.vc
+++ b/gdal/frmts/jp2kak/makefile.vc
@@ -5,7 +5,9 @@ GDAL_ROOT	=	..\..
 !INCLUDE $(GDAL_ROOT)\nmake.opt
 
 KAKINC = -I$(KAKSRC)/managed/all_includes \
-       -I$(KAKSRC)/apps/jp2
+       -I$(KAKSRC)/apps/jp2 \
+       $(KAKEXTRAINCS)
+
 OBJ_EXT =	obj
 OBJ_PREFIX =	$(KAKOBJDIR)
 OBJ_CONFIG =	release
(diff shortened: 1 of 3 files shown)


rcoup:
> Comment KAKADU_7_5_OR_LATER and let user uncomment it (all other options are commented, so no reason to not to comment this one too by default)

Current Kakadu release is v7.10.7, and users should be able to use recent releases _without_ having to mess with options. So feels like there should be a matching logic change (ie. default is YES and it checks for =NO), or it should switch to using KDU_MAJOR_VERSION/etc?

mloskot:
@rcoup These tweaks are non-intrusive, they do not change the current 'logic', and should not affect any existing users.
Commenting `KAKADU_7_5_OR_LATER` falls into tidying up category. The `KAKDIR` is already commented by default.

Although I agree with your proposal, it does not belong to this PR. It should be proposed as a separate PR simplifying the configuration for Kakadu.

rouault:
I guess people might have nmake.local configs without explicit KAKADU_7_5_OR_LATER, relying on the default behaviour. I think it is reasonable to let KAKADU_7_5_OR_LATER=YES as default

mloskot:
Would it be acceptable if this is at least set to the default if not already overwritten (e.g. by `nmake.opt.local`)?

```
!IFNDEF KAKADU_7_5_OR_LATER 
KAKADU_7_5_OR_LATER = YES
!ENDIF
```

rouault:
> Would it be acceptable if this is at least set to the default if not already overwritten

sure

mloskot:
I've updated the PR

PR #1586, opened 2019-05-27
nitf/nitfimage.c: skip bad UDID data (fixes #1578)

### LINKED ISSUE #1578: NITF file with bad UDID TRE won't parse IXSHD
I have a NITF file with junk data in the UDID record (it contains an arbitrary string rather than a TRE).  Because GDAL's NITF reader concatenates the UDID and IXSHD records and then parses the joined TREs, it can't parse the valid TRE in the IXSHD.  It would be relatively easy to do some basic checks on the UDID contents and to ignore the UDID if it does't appear to be a valid TRE (if it is less than 11 bytes or the first length record is not all digits, for instance).

GDAL sensibly emits a failure message about failing to parse the combined TRE.  Instead, it could emit a message informing about the bad UDID contents.

I can make a PR if this is desired.  If it isn't, just close the issue.

The current behavior is that neither the bad UDID nor the good IXSHD TREs are parsed.  It would be desirable if the bad UDID was skipped and the IXSHD was parsed.

Try to load a NITF file with both UDID and IXSHD records where the UDID is a text string instead of a TRE.

## Operating system

The behavior is the same on Ubuntu 18.04 and CentOS 6.

master, built from source, as of 70d4dfe7fa6d9df3212854e471ead1232246e3dc.

manthey:
## What does this PR do?

When reading UDID and IXSHD metadata, the two TRE record sets are concatenated in memory.  If the UDID data is not a valid TRE set, the IXSHD was not being parsed.  Now, if the UDID data does not have valid TRE length as the first entry, it is skipped with a warning and the IXSHD data is still parsed.

This adds two test files that are modified from the fake_nsif.ntf test file.  The valid_udid.ntf file has a CSDIDA record added to a UDID block; the invalid_udid.ntf file has a UDID block that does not contain a TRE.

Fixes #1578.

## Tasklist

 - [x] Add test case(s)
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS: 
  This was tested on a variety of linuxes (Ubuntu 18.04 and CentOS 6)
* Compiler:
  gcc


gdal/frmts/nitf/nitfimage.c
```diff
@@ -657,6 +657,16 @@ NITFImage *NITFImageAccess( NITFFile *psFile, int iSegment )
                     psImage->nTREBytes );
 
             nOffset += nUserTREBytes;
+
+            sscanf(psImage->pachTRE + 6, "%*5d%n", &nFirstTagUsedLength);
+            if (nFirstTagUsedLength != 5)
+            {
+                CPLError(CE_Failure, CPLE_AppDefined,
```
rouault: as the processing of the image is not completely stopped, it would be better to use a CE_Warning

manthey: Changed.

PR #1596, opened 2019-05-28
Set the length in the definition after extraction.

### LINKED ISSUE #1595: GTIFGetDefn() doesn't properly fill UOMLength
The Geotiff length code is extracted but never set in GTIFGetDefn().

abellgithub:
Close #1595

Set the definition's UOMLength field after determining the value.

### DIFF
--- a/gdal/frmts/gtiff/libgeotiff/geo_normalize.c
+++ b/gdal/frmts/gtiff/libgeotiff/geo_normalize.c
@@ -1451,7 +1451,7 @@ int GTIFGetProjTRFInfoEx( PJ_CONTEXT* ctx,
         /* Get the projection method code */
         proj_coordoperation_get_method_info(ctx, transf,
                                             NULL, /* method name */
-                                            NULL, /* method auth name (should be EPSG) */ 
+                                            NULL, /* method auth name (should be EPSG) */
                                             &pszMethodCode);
         assert( pszMethodCode );
         nProjMethod = atoi(pszMethodCode);
@@ -2470,6 +2470,7 @@ int GTIFGetDefn( GTIF * psGTIF, GTIFDefn * psDefn )
 /* -------------------------------------------------------------------- */
     nGeogUOMLinear = 9001; /* Linear_Meter */
     GTIFKeyGetSHORT(psGTIF, GeogLinearUnitsGeoKey, &nGeogUOMLinear, 0, 1 );
+    psDefn->UOMLength = nGeogUOMLinear;
 
 /* -------------------------------------------------------------------- */
 /*      Try to get a PCS.                                               */


rouault:
Can you explain when this is needed ? The only cases where I would see this allowed would be for a Geographic 3D CRS (but those aren't allowed yet in GeoTIFF 1.0. Only in the upcoming 1.1), or for a Geocentric CRS (but those are mostly hypothetical)

abellgithub:
1) It seems like looking up nGeogUOMLinear is currently a no-op, as the value isn't used after being set as far as I can tell.
2) Without this change, the length unit is set to KvUserDefined for my geocentric SRS.  Then when converting to WKT with GTIFGetOGISDefn, an error is returned from proj reporting an invalid unit.
3) This is a regression for 3.0 as no error was reported with previous GDAL versions.

My Geotiff directory looks like this.  I won't swear as to validity:

Key/Location/Count/Offset = 1024/0/1/3
Key/Location/Count/Offset = 1025/0/1/1
Key/Location/Count/Offset = 1026/34737/8/0
Key/Location/Count/Offset = 2048/0/1/4326
Key/Location/Count/Offset = 2052/0/1/9001
Key/Location/Count/Offset = 2054/0/1/9102
Key/Location/Count/Offset = 2057/34736/1/1
Key/Location/Count/Offset = 2059/34736/1/0

rouault:
> It seems like looking up nGeogUOMLinear is currently a no-op

Yes this was true of libgeotiff < 1.5 as well

> geocentric SRS

woo, do you have really a GeoTIFF with a geocentric SRS (that is the surface of the GeoTIFF would be a plane in the Earth centric cartesian space)  ? Does that really work (except that bug) ? I've never seen such beasts, nor the folks of the OGC GeoTIFF working group...

> My Geotiff directory looks like this

Would be nicer to have the text report of `listgeo`

abellgithub:
> woo, do you have really a GeoTIFF with a geocentric SRS (that is the surface of the GeoTIFF would > be a plane in the Earth centric cartesian space) ? Does that really work (except that bug) ?
> I've never seen such beasts, nor the folks of the OGC GeoTIFF working group...

This comes from a LAS point clould file.  LAS uses geotiff SRS specifications.  It has nothing to do with a geotiff file, as such.

> My Geotiff directory looks like this
>
> > Would be nicer to have the text report of listgeo

I didn't know about listgeo, but it doesn't read LAS files :)

Here is perhaps more friendly information, though I can provide what you would like:

variable length header record 2 of 4:
  reserved             0
  user ID              'LASF_Projection'
  record ID            34735
  length after header  72
  description          'GeoTiff GeoKeyDirectoryTag'
    GeoKeyDirectoryTag version 1.1.0 number of keys 8
      key 1024 tiff_tag_location 0 count 1 value_offset 3 - GTModelTypeGeoKey: ModelTypeGeocentric
      key 1025 tiff_tag_location 0 count 1 value_offset 1 - GTRasterTypeGeoKey: RasterPixelIsArea
      key 1026 tiff_tag_location 34737 count 8 value_offset 0 - GTCitationGeoKey: unnamed
      key 2048 tiff_tag_location 0 count 1 value_offset 4326 - GeographicTypeGeoKey: GCS_WGS_84
      key 2052 tiff_tag_location 0 count 1 value_offset 9001 - GeogLinearUnitsGeoKey: Linear_Meter
      key 2054 tiff_tag_location 0 count 1 value_offset 9102 - GeogAngularUnitsGeoKey: Angular_Degree
      key 2057 tiff_tag_location 34736 count 1 value_offset 1 - GeogSemiMajorAxisGeoKey: 6378137
      key 2059 tiff_tag_location 34736 count 1 value_offset 0 - GeogInvFlatteningGeoKey: 298.2572236
variable length header record 3 of 4:
  reserved             0
  user ID              'LASF_Projection'
  record ID            34736
  length after header  16
  description          'GeoTiff GeoDoubleParamsTag'
    GeoDoubleParamsTag (number of doubles 2)
      298.257 6.37814e+06
variable length header record 4 of 4:
  reserved             0
  user ID              'LASF_Projection'
  record ID            34737
  length after header  9
  description          'GeoTiff GeoAsciiParamsTag'
    GeoAsciiParamsTag (number of characters 9)
      unnamed|

PR #1611, opened 2019-06-03
gtiff: make overview blocksize default to the same as the full-resolution 

tbonfort:
The default overview blocksize is set to 128 whereas I'd argue that setting it to the same value as the full-resolution blocksize makes more sense.

### DIFF
--- a/gdal/doc/source/drivers/raster/gtiff.rst
+++ b/gdal/doc/source/drivers/raster/gtiff.rst
@@ -114,8 +114,10 @@ method.
 
 The block size (tile width and height) used for overviews (internal or
 external) can be specified by setting the GDAL_TIFF_OVR_BLOCKSIZE
-environment variable to a power-of-two value between 64 and 4096. The
-default value is 128.
+environment variable to a power-of-two value between 64 and 4096. The 
+default is 128, or starting with GDAL 3.1 to use the same block size 
+as the full-resolution dataset if possible (i.e. block height and width
+are equal, a power-of-two, and between 64 and 4096).
 
 Overviews and nodata masks
 --------------------------
--- a/gdal/frmts/gtiff/geotiff.cpp
+++ b/gdal/frmts/gtiff/geotiff.cpp
@@ -177,28 +177,42 @@ void GTIFFSetInExternalOvr( bool b )
 /*                     GTIFFGetOverviewBlockSize()                      */
 /************************************************************************/
 
-void GTIFFGetOverviewBlockSize( int* pnBlockXSize, int* pnBlockYSize )
+void GTIFFGetOverviewBlockSize( GDALRasterBandH hBand, int* pnBlockXSize, int* pnBlockYSize )
 {
-    const char* pszVal = CPLGetConfigOption("GDAL_TIFF_OVR_BLOCKSIZE", "128");
-    int nOvrBlockSize = atoi(pszVal);
-    if( nOvrBlockSize < 64 || nOvrBlockSize > 4096 ||
-        !CPLIsPowerOfTwo(nOvrBlockSize) )
+    const char* pszVal = CPLGetConfigOption("GDAL_TIFF_OVR_BLOCKSIZE", nullptr);
+    if( ! pszVal )
     {
-        static bool bHasWarned = false;
-        if( !bHasWarned )
+        GDALRasterBand* const poBand = GDALRasterBand::FromHandle(hBand);
+        poBand->GetBlockSize(pnBlockXSize,pnBlockYSize);
+        if ( *pnBlockXSize != *pnBlockYSize ||
+             *pnBlockXSize < 64 || *pnBlockXSize > 4096 ||
+             !CPLIsPowerOfTwo(*pnBlockXSize) )
         {
-            CPLError( CE_Warning, CPLE_NotSupported,
-                      "Wrong value for GDAL_TIFF_OVR_BLOCKSIZE : %s. "
-                      "Should be a power of 2 between 64 and 4096. "
-                      "Defaulting to 128",
-                      pszVal );
-            bHasWarned = true;
+            *pnBlockXSize=*pnBlockYSize=128;
         }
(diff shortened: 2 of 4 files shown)


gdal/frmts/gtiff/geotiff.cpp
```diff
@@ -177,28 +177,42 @@ void GTIFFSetInExternalOvr( bool b )
...
-    const char* pszVal = CPLGetConfigOption("GDAL_TIFF_OVR_BLOCKSIZE", "128");
-    int nOvrBlockSize = atoi(pszVal);
-    if( nOvrBlockSize < 64 || nOvrBlockSize > 4096 ||
-        !CPLIsPowerOfTwo(nOvrBlockSize) )
+    const char* pszVal = CPLGetConfigOption("GDAL_TIFF_OVR_BLOCKSIZE", NULL);
+    if( ! pszVal )
     {
-        static bool bHasWarned = false;
-        if( !bHasWarned )
+        GDALRasterBand* const poBand = static_cast<GDALRasterBand *>(hBand);
```
rouault: GDALRasterBand::FromHandle(hBand)

gdal/frmts/gtiff/gtiff.h
```diff
@@ -43,7 +43,7 @@ void CPL_DLL LibgeotiffOneTimeInit();
 CPL_C_END
 
 void    GTIFFSetInExternalOvr( bool b );
-void    GTIFFGetOverviewBlockSize( int* pnBlockXSize, int* pnBlockYSize );
+void    GTIFFGetOverviewBlockSize( GDALRasterBandH poBand, int* pnBlockXSize, int* pnBlockYSize );
```
rouault: naming convention: `hBand`

gdal/doc/source/drivers/raster/gtiff.rst
```diff
@@ -114,8 +114,9 @@ method.
 
 The block size (tile width and height) used for overviews (internal or
 external) can be specified by setting the GDAL_TIFF_OVR_BLOCKSIZE
-environment variable to a power-of-two value between 64 and 4096. The
-default value is 128.
+environment variable to a power-of-two value between 64 and 4096. The 
+default is to use the same value as the full-resolution dataset (if a 
+power-of-two between 64 and 4096) or to fallback to 128.
```
rouault: maybe add "starting with GDAL 3.1"

rouault:
```
geotiff.cpp:182:76: error: zero as null pointer constant [-Werror=zero-as-null-pointer-constant]
     const char* pszVal = CPLGetConfigOption("GDAL_TIFF_OVR_BLOCKSIZE", NULL);
```
--> `use nullptr`

vincentsarago:
Well, I have to disagree on this one @tbonfort @rouault, I thought about it when working on web-optimized COG (ref: https://github.com/cogeotiff/rio-cogeo/issues/60) and lead to the conclusion that it was better to keep a smaller overview size as default. 

FYI, rio-cogeo set the default value to 128 (or to user input) so this won't break my implementation.

tbonfort:
yes, we disagree @vincentsarago :smile: 
setting the internal tile size is a tradeoff between the number of GET requests you need to issue and the overhead of how many pixels you will be reading but that will not be used in your aoi. In general, doubling your tile size will divide the number of requests by 2 (and not by 4, as we can merge consecutive horizontal tiles, but still need to issue one request for each tile row), and this is independant of the fact that you're in an overview or not. As a cog creator, this trade-off has been set when selecting the block size for the full resolution image, and should not be overridden when creating the overviews.

rouault:
@vincentsarago  I believe the concern you have in  cogeotiff/rio-cogeo#60 should be alleviated by adding padding to the full resolution webmercator tile, so that at least the first levels (number configurable by the user) of TIFF overviews  also perfectly align with a webmercator tile at a lower zoom level (what I tried to explain in point 2) of https://lists.osgeo.org/pipermail/gdal-dev/2019-May/050169.html).
An alternative to avoid those mis-alignment issues would be to divide the block size by 2 for *each* overview level, but that soon becomes impractical too and would make the size of the file to grow insanely due to a higher number of tiles to index and a decrease in the compression rate.

vincentsarago:
> @vincentsarago I believe the concern you have in cogeotiff/rio-cogeo#60 should be alleviated by adding padding to the full resolution webmercator tile, so that at least the first levels (number configurable by the user) of TIFF overviews also perfectly align with a webmercator tile at a lower zoom level (what I tried to explain in point 2) of https://lists.osgeo.org/pipermail/gdal-dev/2019-May/050169.html).

@rouault TBH, I'm not sure to understand the `padding` idea. 

> An alternative to avoid those mis-alignment issues would be to divide the block size by 2 for each overview level, but that soon becomes impractical too and would make the size of the file to grow insanely due to a higher number of tiles to index and a decrease in the compression rate.

Yes that's why having at least blocksize for overviews set to 128 is a safer approach IMO. With COGs I'm always trying to reduce the number of GETs AND the total number of pixels downloaded that why I'm inclined to use smaller blocksize for overviews.

I think we have a different opinion on COG optimization (file size / number of GET) but I'm not a core developer so I won't insist on this PR event if I'm not 💯 convinced by this change.

rouault:
> I'm not sure to understand the `padding` idea.

Let's take an example. When you generate a COG file whose top-left corner would match the one of tile 17-2511-1341, then all tile requests at zoom level 17 will result in reading one single TIFF tile. Fine. But if you request at zoom level 16, and the tile size is the same at all levels, none of those tile requests will match a single TIFF tile of the 1st overview level since 2511 or 1341 cannot be divided by two. So instead, add one tile of padding at the top and left, and thus the origin will be 17-2510-1342, and you have a match between webmercator tiles and TIFF tiles for zoom level 17 and 16. Use 17-2508-1344 as origin, and that will extend to zoom level 15. etc

vincentsarago:
Oh I see, this is indeed a nice feature.

PR #1617, opened 2019-06-05
MITAB: Fix long fields write in local encoding

drons:
It seems that MITAB code was written before GDAL switched to UTF-8 encoding. 
I added a conversion from a UTF-8  to local encoding for the correct calculation of the number of characters in the field names.

Before my changes, field names were truncated to 16 characters, but the MITAB supports 31.

### DIFF
--- a/gdal/ogr/ogrsf_frmts/mitab/mitab_imapinfofile.cpp
+++ b/gdal/ogr/ogrsf_frmts/mitab/mitab_imapinfofile.cpp
@@ -668,16 +668,16 @@ CPLString IMapInfoFile::NormalizeFieldName(const char *pszName) const
                   "for MapInfo format.", pszName );
     }
 
-    if( !EQUAL(osName.c_str(),szNewFieldName) )
-    {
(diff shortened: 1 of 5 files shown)


gdal/ogr/ogrsf_frmts/mitab/mitab_imapinfofile.cpp
```diff
@@ -632,3 +632,53 @@ int IMapInfoFile::TestUtf8Capability() const
 
     return CPLCanRecode("test", GetEncoding(), CPL_ENC_UTF8);
 }
+
+CPLString IMapInfoFile::NormalizeFieldName(const char *pszName) const
+{
+    CPLString   osName(pszName);
+    if( strlen( GetEncoding() ) > 0 )
+        osName.Recode( CPL_ENC_UTF8, GetEncoding() );
```
rouault: this recoding might potentially fail if one provides a name that is not compatible of the dataset encoding. We should probably error out in such situation, shouldn't we ?

drons:
Thanks!

Issue #1614, opened 2019-06-04
CPL_VSIL_CURL_ALLOWED_EXTENSIONS does not play well with query parameters


tbonfort:
```
gdalinfo '/vsicurl/http://even.rouault.free.fr/gtiff_test/S2A_MSIL1C_20170102T111442_N0204_R137_T30TXT_20170102T111441_TCI_no_ovr_strip.tif?key=foobar'
```
=> OK
```
gdalinfo --config CPL_VSIL_CURL_ALLOWED_EXTENSIONS tif '/vsicurl/http://even.rouault.free.fr/gtiff_test/S2A_MSIL1C_20170102T111442_N0204_R137_T30TXT_20170102T111441_TCI_no_ovr_strip.tif?key=foobar'
```
=> Not OK

PR #1618, opened 2019-06-05
fix CPL_VSIL_CURL_ALLOWED_EXTENSIONS with query string (#1614)

tbonfort:
Proposal fix for #1614


tbonfort:
This won't work as is, no need to review yet

tbonfort:
thanks @rouault !

PR #1632, opened 2019-06-10
Elasticsearch: GeoJSON type field should be mapped as text in ES>=5

nickpeihl:
## What does this PR do?
When GeoJSON data was converted to Elasticsearch versions >=5, the GeoJSON type field was incorrectly mapped to a `string` instead of `text`. Elasticsearch would return an error on the unsupported field type. This PR fixes the field mapping similar to how it is done earlier in the BuildMap() function.

N/A

## Tasklist

 - [ ] Add test case(s)
 - [ ] Review
 - [ ] Adjust for comments


rouault:
Cherry-picked to 3.0 and 2.4

PR #1634, opened 2019-06-11
tiledb: removed ambiguous use of size_t

normanb:
## What does this PR do?

OS GDAL builds with TileDB are currently broken on Conda due to an ambiguity with `size_t`, the simple fix is to use `uint64_t` and only 64 bit architectures are supported by this driver.


rouault:
@normanb I let you do the manual backport

PR #1641, opened 2019-06-13
[docs] add MS4W to download options

jmckenna:
- add MS4W to Windows download options, on the new website

### DIFF
--- a/gdal/doc/source/download.rst
+++ b/gdal/doc/source/download.rst
@@ -50,12 +50,13 @@ In this section we list a number of the binary distributions of GDAL.
 Windows
 ................................................................................
 
-Windows builds are available via `Conda Forge`_ (64-bit only). See the
-:ref:`conda` for more detailed information.
+- Windows builds are available via `Conda Forge`_ (64-bit only). See the
+  :ref:`conda` for more detailed information.
 
-
-
-
+- `MS4W <https://ms4w.com/>`__ (MapServer for Windows) also includes GDAL 
+  commandline utilities, as well as a full Web server environment (including 
+  Apache, PHP, MapServer, MapCache tile engine, and many other packages).  The 
+  installer can be used, or simply extract the zipfile to C:/& run setenv.bat.
 
 Debian
 ................................................................................


hobu:
I think we only want to list builds that have fully reproducible open source build recipes. That way people know exactly what they are getting of all versions and dependencies. This is especially important for something that has so many dependencies like GDAL. The Docker, Debian, and Conda builds all provide this for GDAL and all of its dependencies.

OSGeo4W meets this criteria, and we should probably add a link to that. Does MS4W provide fully reproducible open source build recipes?

jmckenna:
MS4W existed before everything you mentioned, is the first to test each GDAL release candidate, and will outlast each of the options you have mentioned., and ~5000 users install it each month.  I do appreciate the politics though, sadly I worried about that when I saw the page.   However, thank you for confirming this.  Please disregard my intent to share and help the community.

hobu:
The only politics in regard to MS4W is MS4W does not provide a full listing of *every* library, dependency, version, and build option used. [This](https://ms4w.com/README_INSTALL.html#mapserver-build-environment) has version information, but it does not state which compiler runtime was used, or how the options were set for each of those libraries. This information is critical for security event tracking, discovery, and history. 

> MS4W existed before everything you mentioned,

I assume you are not claiming that MS4W predates Debian. As for the others, each of them provides the full recipes for people to use to build their own binaries however they need. This provides accounting, a self-verification loop, and a historical record of how these binaries were constructed.  

> Please disregard my intent to share and help the community.

The GDAL project has never advertised "Official Binaries" and has never supported anything as such. I hope you can appreciate that pointing at MS4W or gisinternals or Conda or Debian is equivalent to the project saying as much. If we do that, it's completely reasonable that people are able to see how that cake was made. 

That said, it would be totally reasonable to add MS4W to the [Software using GDAL](https://gdal.org/software_using_gdal.html#software-using-gdal) page.

jmckenna:
You are bringing politics into "Binaries", a section where Windows users just want to download, extract and run.  Maybe the section should be renamed.   Before this page, the list of GDAL binaries listed all options, alphabetically, which worked excellent, gave users what they wanted.   You are speaking about the 1% of 1% of Windows users, those who want and are able to compile.  I see sections needed here, "Binaries" (listing all binaries available, Alphabetically), and "Build Environments: compile your own GDAL".  

But honestly, I don't want to fight.  

Instead, I'll just keep on helping Windows users and sharing GDAL.

Thank you for your time.

PR #1654, opened 2019-06-15
Elasticsearch: Enable support for Elasticsearch 7 (fixes #1246)

### LINKED ISSUE #1246: Compatibility with Elasticsearch 7.0

Elasticsearch is [removing mapping types](https://www.elastic.co/guide/en/elasticsearch/reference/current/removal-of-types.html) for indices in the upcoming 7.0 release. The [GDAL Elasticsearch driver](https://www.gdal.org/drv_elasticsearch.html) has a `MAPPING_NAME` layer creation option that sets the index mapping type (`FeatureCollection` by default). So attempts to write to Elasticsearch 7+ will return an error.

Ideally we could just ignore the `MAPPING_NAME` option for Elasticsearch 7+ and not include the `_type` parameter ([ex.](https://github.com/OSGeo/gdal/blob/master/gdal/ogr/ogrsf_frmts/elastic/ogrelasticlayer.cpp#L2273)) in the request. But the `MAPPING_NAME` option also determines the JSON format of the document that is sent to Elasticsearch. If the `MAPPING_NAME` is `FeatureCollection` or left as the default, the document matches the GeoJSON Feature format. If `MAPPING_NAME` is anything else, the `feature.properties` object is flattened in the document.

Can we deprecate the `MAPPING_NAME` layer creation option and introduce a new layer creation option (e.g. `FLATTEN_PROPERTIES`) to specify the JSON format of the documents stored in Elasticsearch? 

I'm happy to open a PR, but I'd like to get input from the contributors first.

Using [Elasticsearch v7.0.0-alpha2](https://www.elastic.co/downloads/elasticsearch#preview-release)
`ogr2ogr -progress -lco BULK_SIZE=5000000 -f "ElasticSearch" http://localhost:9200 my_shapefile.shp`

Returns error from Elasticsearch 
```
ERROR 1: HTTP error code : 400
ERROR 1: {"error":{"root_cause":[{"type":"illegal_argument_exception","reason":"Types cannot be provided in put mapping requests, unless the include_type_name parameter is set to true."}],"type":"illegal_argument_exception","reason":"Types cannot be provided in put mapping requests, unless the include_type_name parameter is set to true."},"status":400}
```

GDAL 2.4.0

nickpeihl:
## What does this PR do?
This enables the ElasticSearch driver to work with Elasticsearch 7.x. The `MAPPING_NAME` layer creation option is ignored when converting to Elasticsearch 7.x because mapping types are no longer used. The `MAPPING_NAME` option continues to work the same way in Elasticsearch version < 7.

#1246 

## Tasklist

 - [x] Add test case(s)
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/gdal/doc/source/drivers/vector/elasticsearch.rst
+++ b/gdal/doc/source/drivers/vector/elasticsearch.rst
@@ -279,7 +279,8 @@ options:
    will be written as GeoJSON Feature objects. If another mapping name
    is chosen, a more "flat" structure will be used.
 -  **MAPPING**\ =filename or JSon. Filename from which to read a
-   user-defined mapping, or mapping as serialized JSon.
+   user-defined mapping, or mapping as serialized JSon. This option is
+   ignored when converting to Elasticsearch >=7 (see `Removal of mapping types <https://www.elastic.co/guide/en/elasticsearch/reference/current/removal-of-types.html>`__).
 -  **WRITE_MAPPING**\ =filename. Creates a mapping file that can be
    modified by the user prior to insert in to the index. No feature will
    be written. This option is exclusive with MAPPING.
--- a/gdal/ogr/ogrsf_frmts/elastic/ogrelasticdriver.cpp
+++ b/gdal/ogr/ogrsf_frmts/elastic/ogrelasticdriver.cpp
@@ -105,7 +105,7 @@ void RegisterOGRElastic() {
     "<LayerCreationOptionList>"
     "  <Option name='INDEX_NAME' type='string' description='Name of the index to create (or reuse). By default the index name is the layer name.'/>"
     "  <Option name='INDEX_DEFINITION' type='string' description='Filename from which to read a user-defined index definition, or index definition as serialized JSon.'/>"
-    "  <Option name='MAPPING_NAME' type='string' description='Name of the mapping type within the index.' default='FeatureCollection'/>"
+    "  <Option name='MAPPING_NAME' type='string' description='(ES &lt; 7) Name of the mapping type within the index.' default='FeatureCollection'/>."
     "  <Option name='MAPPING' type='string' description='Filename from which to read a user-defined mapping, or mapping as serialized JSon.'/>"
     "  <Option name='WRITE_MAPPING' type='string' description='Filename where to write the OGR generated mapping.'/>"
     "  <Option name='OVERWRITE' type='boolean' description='Whether to overwrite an existing type mapping with the layer name to be created' default='NO'/>"
--- a/gdal/ogr/ogrsf_frmts/elastic/ogrelasticlayer.cpp
+++ b/gdal/ogr/ogrsf_frmts/elastic/ogrelasticlayer.cpp
@@ -57,7 +57,10 @@ OGRElasticLayer::OGRElasticLayer( const char* pszLayerName,
 
     m_poDS(poDS),
     m_osIndexName(pszIndexName ? pszIndexName : ""),
-    m_osMappingName(pszMappingName ? pszMappingName : ""),
+    // Types are no longer supported in Elasticsearch 7+.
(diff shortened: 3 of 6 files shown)


gdal/ogr/ogrsf_frmts/elastic/ogrelasticdriver.cpp
```diff
@@ -105,7 +105,7 @@ void RegisterOGRElastic() {
     "<LayerCreationOptionList>"
     "  <Option name='INDEX_NAME' type='string' description='Name of the index to create (or reuse). By default the index name is the layer name.'/>"
     "  <Option name='INDEX_DEFINITION' type='string' description='Filename from which to read a user-defined index definition, or index definition as serialized JSon.'/>"
-    "  <Option name='MAPPING_NAME' type='string' description='Name of the mapping type within the index.' default='FeatureCollection'/>"
+    "  <Option name='MAPPING_NAME' type='string' description='Name of the mapping type within the index.' default='FeatureCollection'/>. This option is ignored when converting to Elasticsearch >=7 (see [Removal of mapping types](https://www.elastic.co/guide/en/elasticsearch/reference/current/removal-of-types.html)."
```
rouault: This is invalid XML. The added comment should be moved to gdal/doc/source/drivers/vector/elasticsearch.rst

gdal/ogr/ogrsf_frmts/elastic/ogrelasticlayer.cpp
```diff
@@ -529,9 +532,12 @@ void OGRElasticLayer::FinalizeFeatureDefn(bool bReadFeatures)
                     osPostData = m_osESSearch;
                 }
                 else
-                    osRequest = CPLSPrintf("%s/%s/%s/_search?scroll=1m&size=%d",
-                           m_poDS->GetURL(), m_osIndexName.c_str(),
-                           m_osMappingName.c_str(), m_poDS->m_nBatchSize);
+                {
+                    osRequest = CPLSPrintf("%s/%s", m_poDS->GetURL(), m_osIndexName.c_str());
```
rouault: to avoid most of the conditional tests of version < 7 in that file, shouldn't we have a method BuildFullURL() (or a better name) that would return "GetURL()/m_osIndexName/m_osMappingName" for < 7 or ""GetURL()/m_osIndexName" for >= 7 ?

gdal/ogr/ogrsf_frmts/elastic/ogrelasticdriver.cpp
rouault: Amending my comment: this is valid XML, but not how we proceed usually. So perhaps add at the description attribute " (ES < 7)", and the longer comment in the .rst

gdal/doc/source/drivers/vector/elasticsearch.rst
```diff
@@ -279,7 +279,8 @@ options:
    will be written as GeoJSON Feature objects. If another mapping name
    is chosen, a more "flat" structure will be used.
 -  **MAPPING**\ =filename or JSon. Filename from which to read a
-   user-defined mapping, or mapping as serialized JSon.
+   user-defined mapping, or mapping as serialized JSon. This option is
+   ignored when converting to Elasticsearch >=7 (see `Removal of mapping types <https://www.elastic.co/guide/en/elasticsearch/reference/current/removal-of-types.html`__).
```
rouault: causes `/__w/1/s/gdal/doc/source/drivers/vector/elasticsearch.rst:: WARNING: Anonymous hyperlink mismatch: 1 references but 0 targets.`

nickpeihl: 😞 Sorry, about that. I've pushed a fix after testing the doc build.

PR #1668, opened 2019-06-22
Update rs2dataset.cpp

### LINKED ISSUE #1666: Wrong Radarsat 2 GCP coordinates calculation

The GCP location on raster is wrongly calculated. According to the document https://spacedata.copernicus.eu/documents/12833/14537/Radarsat-2_ProductFormatDefinition Chapter A "IMAGE CO-ORDINATE REFERENCE SYSTEMS" all coordinates of the product are related to PixelIsCenter, except direct GeoTiff goereferencing which is corrected to be PixelIsArea. The driver does not provide AREA_OR_POINT information. Moreover the Affine GeoTransform returns the transformation as AREA_OR_POINT=Area. But returned GCPs coordinates correspond to the model  with AREA_OR_POINT=Point (pointing to center of pixels) . As a result of that, georeferencing performed from GeoTransform and GCP point data are different with a half pixel shift and slightly wrong scaled.

Look at the source code in rs2dataset.cpp

## Operating system

All

Latest tested 3.0.0

an-ivanov:
## What does this PR do?
Radarsat 2 driver: correct affine GeoTransform calculation (fix #1666)

#1666

### DIFF
--- a/gdal/frmts/rs2/rs2dataset.cpp
+++ b/gdal/frmts/rs2/rs2dataset.cpp
@@ -1313,9 +1313,9 @@ GDALDataset *RS2Dataset::Open( GDALOpenInfo * poOpenInfo )
             psGCP->pszId = CPLStrdup( szID );
             psGCP->pszInfo = CPLStrdup("");
             psGCP->dfGCPPixel =
-                CPLAtof(CPLGetXMLValue(psNode,"imageCoordinate.pixel","0"));
+                CPLAtof(CPLGetXMLValue(psNode,"imageCoordinate.pixel","0")) + 0.5;
             psGCP->dfGCPLine =
-                CPLAtof(CPLGetXMLValue(psNode,"imageCoordinate.line","0"));
+                CPLAtof(CPLGetXMLValue(psNode,"imageCoordinate.line","0")) + 0.5;
             psGCP->dfGCPX =
                 CPLAtof(CPLGetXMLValue(psNode,"geodeticCoordinate.longitude",""));
             psGCP->dfGCPY =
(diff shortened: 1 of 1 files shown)


rouault:
Are you really sure about that ?
GDAL geotransform convention is to go expose the georegistration of formats with PixelIsCenter convention as PixelIsArea.
Perhaps the right fix would be to add +0.5 to dfGCPPixel and dfGCPLine instead ?

an-ivanov:
Yes, you are right. According to the document https://spacedata.copernicus.eu/documents/12833/14537/Radarsat-2_ProductFormatDefinition Chapter A "IMAGE CO-ORDINATE REFERENCE SYSTEMS" all coordinates of the product are related to PixelIsCenter, except direct GeoTiff goereferencing which is corrected to be PixelIsArea.
So, what should I do, close issue #1666 and open a new one and create a new pull requests? Or edit this one?

rouault:
> So, what should I do, close issue #1666 and open a new one and create a new pull requests? Or edit this one?

You can just update the issue and this pull request (git commit --amend and force push)

an-ivanov:
There are doubts appeared, do the dfGCPPixel and dfGCPLine coordinates belong to the AREA_OR_POINT=Area model. I can not find the information in GDAL manuals. Or it should follow a model defined in GDALMD_AREA_OR_POINT, being AREA_OR_POINT=Area as default?

rouault:
> There are doubts appeared, do the dfGCPPixel and dfGCPLine coordinates belong to the AREA_OR_POINT=Area model

I guess so. This might not be explicited in the doc, but this is the practice in other drivers and how the GCP transformer works as far as I can remember. AREA_OR_POINT=Area is the dominant model in GDAL.  AREA_OR_POINT is mostly a metadata hint in GDAL to indicate format encoding convention (for example in the GeoTIFF driver, this will influence whether a half-pixel shift is done between GDAL GeoTransform and GeoTIFF tiepoint) that has no influence on GDAL algorithms.

PR #1682, opened 2019-06-26
WIP/RFC: read-only support for subdataset masks and overviews using tiff pages

tbonfort:
Currently TIFF subdatasets do not support overviews or masks. When overview or mask IFDs are present in the TIFF file they will be assigned to the "first" subdataset (provided that they are compatible), which results in:
- subdataset 0 being potentially assigned invalid masks/overviews
- subdatasets n>0 not supporting masks or overviews

The present pull request proposes a read-only support of the TIFFTAG_PAGENUMBER to be able to associate a subdataset with its corresponding masks/overviews.

opening "multidataset.tif" will list all available tiff pages as subdatasets, and will associate all mask/overview ifds *without pagenumber set* to the current dataset
opening "GTIFF_PAGE:0:multidataset.tif" will describe the IFD with pagenumber==0 along with all masks/overviews with pagenumber==0

Backwards incompatibilities:
- ```SUBDATASET_1_NAME=GTIFF_DIR:1:filename.tif``` will not be emitted in the SUBDATASETS metadata when IFD 1 has a pagenumber set. the entry will be replaced by ```SUBDATASET_1_NAME=GTIFF_PAGE:1:filename.tif```
- Numbering/naming of subdatasets is no longer tied to the ifd number.
- When opening a tif file without specifying the page number, masks/overviews that have an explicit pagenumber set will not show up

### DIFF
--- a/gdal/frmts/gtiff/geotiff.cpp
+++ b/gdal/frmts/gtiff/geotiff.cpp
@@ -329,6 +329,7 @@ private:
     double      m_dfNoDataValue = -9999.0;
 
     toff_t      m_nDirOffset = 0;
+    int         m_nPage = -1;
 
     int         m_nBlocksPerBand = 0;
     int         m_nBlockXSize = 0;
@@ -12207,6 +12208,8 @@ int GTiffDataset::Identify( GDALOpenInfo *poOpenInfo )
 /* -------------------------------------------------------------------- */
     if( STARTS_WITH_CI(pszFilename, "GTIFF_DIR:") )
         return TRUE;
+    if( STARTS_WITH_CI(pszFilename, "GTIFF_PAGE:") )
+        return TRUE;
 
 /* -------------------------------------------------------------------- */
 /*      First we check to see if the file has the expected header       */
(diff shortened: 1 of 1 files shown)


rouault:
Interesting... I'm wondering if it is valid for a main image, its mask, its overviews to share the same pagenumber ? Did you investigate that ? Not sure if there is much to investigate on the documentation side, besides https://www.awaresystems.be/imaging/tiff/tifftags/pagenumber.html which doesn't talk a lot about that. It seems to me that this was designed for "simple" use cases such as scanner applications. If you have photoshop or something non-libtiff based, would be interesting to see how they react on that. Actually you might bring that matter on the tiff mailing list ( now living at https://lists.osgeo.org/mailman/listinfo/tiff ) as they are non-libtiff users lurking there who might be able to comment.
Another thought I have is that for your use case, the proper mechanism would probably to use the SubIFD mechanism to have a real tree organizations, where the IFD at first level would be only the main resolution imagery, and the SubIFD of each page would link to its mask, overviews and overviews of mask. But libtiff has likely poor/none write support for that (might work on the read side, but not even sure as nobody, in the geospatial field at least, seems to use that AFAIK).
I assume you would create such files externally of GDAL ?
A small test file and a corresponding autotest would be appropriate.

tbonfort:
thanks for the feedback Even, I'll open a new pull request which uses
subIFDs

Le mer. 26 juin 2019 à 22:33, Even Rouault <notifications@github.com> a
écrit :

PR #1685, opened 2019-06-27
Doc gdallocationinfo: Add stdin example

kannes:
Usage of gdallocationinfo is not very intuitive. When not supplied with a location it will simply wait for input, not exit or print a message. This example of stdin input will hopefully make it more clear.

Uses autotest/utilities/data/utmsmall.tif

### DIFF
--- a/gdal/doc/source/programs/gdallocationinfo.rst
+++ b/gdal/doc/source/programs/gdallocationinfo.rst
@@ -138,3 +138,26 @@ Query a VRT file providing the location in WGS84, and getting the result in xml.
             <Value>16</Value>
         </BandReport>
     </Report>
+
+Reading location from stdin.
+
+::
+
+	$ cat coordinates.tsv 
+	443020	3748359
+	441197	3749005
+	443852	3747743
+	
+	$ cat coordinates.tsv | gdallocationinfo -geoloc utmsmall.tif 
+	Report:
+	  Location: (38P,49L)
+	  Band 1:
+		Value: 214
+	Report:
+	  Location: (7P,38L)
+	  Band 1:
+		Value: 107
+	Report:
+	  Location: (52P,59L)
+	  Band 1:
+		Value: 148
--- a/gdal/doc/source/programs/gdallocationinfo.rst
+++ b/gdal/doc/source/programs/gdallocationinfo.rst
@@ -143,21 +143,21 @@ Reading location from stdin.
 
 ::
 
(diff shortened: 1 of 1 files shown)


rouault:
Doc builing complains:
```
./source/programs/gdallocationinfo.rst:	$ cat coordinates.tsv 
./source/programs/gdallocationinfo.rst:	443020	3748359
./source/programs/gdallocationinfo.rst:	441197	3749005
./source/programs/gdallocationinfo.rst:	443852	3747743
./source/programs/gdallocationinfo.rst:	
./source/programs/gdallocationinfo.rst:	$ cat coordinates.tsv | gdallocationinfo -geoloc utmsmall.tif 
./source/programs/gdallocationinfo.rst:	Report:
./source/programs/gdallocationinfo.rst:	  Location: (38P,49L)
./source/programs/gdallocationinfo.rst:	  Band 1:
./source/programs/gdallocationinfo.rst:		Value: 214
./source/programs/gdallocationinfo.rst:	Report:
./source/programs/gdallocationinfo.rst:	  Location: (7P,38L)
./source/programs/gdallocationinfo.rst:	  Band 1:
./source/programs/gdallocationinfo.rst:		Value: 107
./source/programs/gdallocationinfo.rst:	Report:
./source/programs/gdallocationinfo.rst:	  Location: (52P,59L)
./source/programs/gdallocationinfo.rst:	  Band 1:
./source/programs/gdallocationinfo.rst:		Value: 148
Tabs are bad, please use four spaces in .rst files.
```
Perhaps just use space separator and .txt instead of .tsv

rouault:
There are still tabs
```
./source/programs/gdallocationinfo.rst:    443020	3748359
./source/programs/gdallocationinfo.rst:    441197	3749005
./source/programs/gdallocationinfo.rst:    443852	3747743
Tabs are bad, please use four spaces in .rst files.
```

kannes:
Just noticed, I had thought that they were fine when not wanna-be indentation... RST != Rest...

rouault:
>  I had thought that they were fine when not wanna-be indentation

The current check is rather simple, using a simple grep. But even in block quotes, indentations could probably cause issues. I guess that if we want to specifically indicate that a tab character is needed (which is not the case here), we should writte in full letter <TAB> so that it is obvious to the reader, or some mention that spaces should be replaced by tabulation

kannes:
That is very true, I was not considering it well. Thanks!

Issue #1695, opened 2019-07-03
Driver name for files created with the new COG driver


vincentsarago:
When running `gdalinfo` on a COG created using `gdalinfo -of COG` the driver returned by gdalinfo is set to `GTiff`

I'm not sure how GTIff and COG are compatible (old version of https://github.com/OSGeo/gdal/blob/master/gdal/swig/python/samples/validate_cloud_optimized_geotiff.py don't work)

## Expected behavior and actual behavior.

I expect driver to be `COG` not `GTiff/GeoTIFF`

## Steps to reproduce the problem.

```
$ gdalinfo --version
GDAL 3.1.0dev-3b9c6ac06b6b7fd6aaf451ac76c31df705ed5ab1, released 2019/07/03

$ gdal_translate -of COG my_file.tif test_cog.tif

$ gdalinfo test_cog.tif | grep "Driver"
Driver: GTiff/GeoTIFF
```

## Operating system

latest gdal docker (ubuntu-full)

palmerj:
Do you mean `-of cog` in the steps?

vincentsarago:
yes thanks @palmerj

rouault:
> When running `gdalinfo` on a COG created using `gdalinfo -of COG` the driver returned by gdalinfo is set to `GTiff`

Yes this is expected. A COG file is just a GTiff file with constraints on the internal layout. The COG driver is creation-only syntaxic sugar.

> old version of https://github.com/OSGeo/gdal/blob/master/gdal/swig/python/samples/validate_cloud_optimized_geotiff.py don't work

The new version has been fixed to accept new COG files with the "hidden feature declaration", as well as older ones.

Well, one could perhaps have a Open() implementation in the COG driver that would only trigger when the "hidden feature declaration" stuff is found, and just rely on the GTiff driver beneath. But I'm not sure if there are real benefits to that. I could imagine it could cause compatibility issues to applications that would expect the driver name to be GTiff when reading a .tif.
What could be done however is to have a metadata item in the IMAGE_STRUCTURE metadata domain to indicate LAYOUT=COG for example.

vincentsarago:
thanks @rouault 

I'm mostly worried about compatibility issue but it seems to be ok from what you are saying. I think what I'd really love is to have a quick way to check weither the file has been created with the new COG driver or not so maybe `LAYOUT=COG` could be a solution.

I'm going to close this issue because it's not a bug 👌

PR #1696, opened 2019-07-03
A few tweaks for COG: advertized in metadata and a couple fixes

rouault:


pedros007:
Fixes #1695, not 1696.

PR #1702, opened 2019-07-08
Fix stale obj0 for swig_obj[0]

perestoronin:
Fix stale obj0 for swig_obj[0] to compile.


rouault:
For which version of SWIG is this needed ?

perestoronin:
swig-4.0.0

rouault:
Thanks. Integrating with backward compatibility with older SWIG versions

PR #1717, opened 2019-07-12
Hotfix for netCDF-4 and netCDF-4 Classic CF-1.8 Writing

wchen329:
## What does this PR do?
This PR fixes HDF5 issues that would arise when writing a netCDF-4 or netCDF-4 (Classic) file. These HDF5 issues would lead to complete failure when translating a layer into the netCDF-4 or netCDF-4 (Classic) formats. These failures would only occur with GEOMETRY_ENCODING=CF_1.8 (the new default).

Issue #1287 
PR #1541 

## Tasklist
 - [x] Add test case(s)
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS: openSUSE Leap
* Compiler: GCC 9.1.1 （2019/06/11）


wchen329:
Hey @rouault, ready for review now, please look at my changes and the following comment when you got time:

Since you wrote a lot of the original netCDF Driver, I would like for you to comment on the following:
The so-called "HDF5 issues" fixed in this PR were caused by using the same name of the geometry_container / OGRLayer and using it for the feature indexing dimension name. This did not cause any issues in netCDF-3 files, but for some reason this caused an issue in HDF5 containered netCDF files. Furthermore, using GDB I found that the dimension that caused the HDF5 error: when I defined that dimension using nc_def_dim(...) did not raise an error but returned a dimension number and NC_NOERR! In fact the netCDF library only threw an error in netCDFDataset::setDefineMode(bool) [which I presume is only because it was trying to write and "something happened"). Is this simply an unwritten rule specific to NC4 creation, or is this a bug and should this be filed with Unidata?

autotest/gdrivers/netcdf.py
```diff
@@ -4320,9 +4321,91 @@ def test_write_nc_from_nc():
...
+    fnam = feat.GetFieldAsString("NAMES")
+    assert(fWkt == "MULTIPOLYGON (((-2 0 -5,-2 1 -6,-1 1 -6,-2 0 -5)))")
+    assert(fnam == "Single_Triangly")
+
 def test_clean_tmp():
     # [KEEP THIS AS THE LAST TEST]
     # i.e. please do not add any tests after this one. Put new ones above.
     # Not actually a test, just cleans up tmp...
-    gdaltest.clean_tmp()
+    # gdaltest.clean_tmp()
```
rouault: should probably be uncommented

wchen329: You're right it should be :+1:

Issue #1643, opened 2019-06-13
Using 'BAG' driver option CreateCopy() appears to shift the input data


Casiano-Koprowski:
## Expected behavior and actual behavior.

Using 'BAG' driver option CreateCopy(name, gdal.Dataset) appears to shift the input data

## Steps to reproduce the problem.

The .bag file I have been using for this testing can be found here:
[https://data.ngdc.noaa.gov/platforms/ocean/nos/coast/H12001-H14000/H12525/BAG/H12525_MB_4m_MLLW_1of2.bag](https://data.ngdc.noaa.gov/platforms/ocean/nos/coast/H12001-H14000/H12525/BAG/H12525_MB_4m_MLLW_1of2.bag)

Code used to reproduce the problem:
```
test = gdal.Open('H12525_MB_4m_MLLW_1of2.bag')
driver = gdal.GetDriverByName('BAG')
driver.CreateCopy('Test.bag', test)
```

## Operating system

Windows 10 64 bit

## GDAL version and provenance

Python 3.6.7
Version 2.4.1

rouault:
What make you think that there's a shift ? My tests don't confirm that:
```
$ gdal_translate H12525_MB_4m_MLLW_1of2.bag out.bag
$ python swig/python/scripts/gdalcompare.py H12525_MB_4m_MLLW_1of2.bag out.bag
Files differ at the binary level.
Metadata value difference for key "BagVersion"
  Golden: "1.6.0"
  New:    "1.6.2"
Metadata value difference for key "BAG_DATETIME"
  Golden: "2018-05-14T18:38:18"
  New:    "2019-06-13T19:17:52"
Differences Found: 3
```
So from GDAL point of view, the source and output files are identical except some minor metadata differences.

Casiano-Koprowski:
Ahh; so I guess this may stem a display issue based on the version of the bag and the software I am using.  Rounding issues that are moving the data.

Sorry, I should have seen this; Thank you for your help!

Casiano-Koprowski:
Hello:

I just wanted to follow up quickly;  While the test using gdalcompare only yielded differences in the timestamp and version of the files, the metadata.xml of the BAG files created using this method also contained some differences.  In opening the metadata of both files, it seems that the values for resolution and extent are slightly different.  When using GetGeoTransform() on the object the values are also different.  I have attached the .xml files as .txt files for both; they were pulled from the files of the same process as the original steps to reproduce. 

[...]

PR #1728, opened 2019-07-17
 BAG dataset - calculate the northeast pixel corner rather than scaling the resolution, due to an incorrectly shifted northeast corner in some CARIS surveys

zacharyburnettNOAA:
## What does this PR do?

attempts to fix issue brought up in https://github.com/OSGeo/gdal/issues/1643

https://github.com/OSGeo/gdal/issues/1643


gdal/frmts/hdf5/bagdataset.cpp
```diff
@@ -3346,8 +3345,7 @@ void BAGDataset::LoadMetadata()
                 {
                     // Found with https://data.ngdc.noaa.gov/platforms/ocean/nos/coast/H12001-H14000/H12525/BAG/H12525_MB_4m_MLLW_1of2.bag
                     // https://github.com/OSGeo/gdal/issues/1643
-                    CPLDebug("BAG", "cornerPoints seems to use a PixelIsArea convention.");
-                    bPixelCenterConvention = false;
+                    CPLError(CE_Failure, CPLE_AppDefined,"Ignoring BAG Upper Right Corner.");
```
rouault: I'd wish we maintain the ability to handle in a nice way those datasets, even if there are "illegal" given the specification (buggy products do exist in the wild, and when it is possible and reasonable code-wise to handle them, we should try). I'm OK with emitting a message, but rather a CE_Warning than a CE_Failure, and with a more explicit message saying that the product uses a PixelIsArea convention which is non-standard

gdal/frmts/hdf5/bagdataset.cpp
```diff
@@ -4582,8 +4569,6 @@ void GDALRegister_BAG()
 "  <Option name='ZLEVEL' type='int' "
     "description='DEFLATE compression level 1-9' default='6' />"
 "  <Option name='BLOCK_SIZE' type='int' description='Chunk size' />"
-"  <Option name='CORNER_POINTS_EXTEND_HALF_PIXEL' type='boolean' description="
```
rouault: I'm OK if we drop that capability, but then it must be undocumented as well in doc/source/drivers/raster/bag.rst

autotest/gdrivers/bag.py
```diff
@@ -666,49 +665,29 @@ def test_bag_write_south_up():
...
-    gdal.Translate('/vsimem/out_non_standard.bag', 'data/byte.tif',
-                   creationOptions = ['VAR_DATE=2019-06-21',
-                                      'CORNER_POINTS_EXTEND_HALF_PIXEL=YES'])
-
-    f = gdal.VSIFOpenL('/vsimem/out.bag', 'rb')
-    assert f
-    data = gdal.VSIFReadL(1, 1000000, f)
-    gdal.VSIFCloseL(f)
-    gdal.Unlink('/vsimem/out.bag')
+    ds = gdal.Open('data/test_offset_ne_corner.bag')
```
rouault: is that one completely inconsistent, or using a pixelisarea convention ? I'd wish we keep a test for a dataset with a pixelisarea convention

gdal/frmts/hdf5/bagdataset.cpp
rouault: In the final else case with the 3 CPLDebug() messages, you shoud also emit a CPLError(CE_Warning, ...) since that latter case is more annoying that the PixelIsArea case: here the georeferencing is clearly inconsistent !

gdal/frmts/hdf5/bagdataset.cpp
```diff
@@ -3375,15 +3373,13 @@ void BAGDataset::LoadMetadata()
             }
 
             adfGeoTransform[0] = dfLLX;
-            adfGeoTransform[1] = (dfURX - dfLLX) / (m_nLowResWidth - (bPixelCenterConvention ? 1 : 0));
-            adfGeoTransform[3] = dfURY;
-            adfGeoTransform[5] = (dfLLY - dfURY) / (m_nLowResHeight - (bPixelCenterConvention ? 1 : 0));
+            adfGeoTransform[1] = dfResWidth;
```
rouault: shoud be modiied to behave nicely when PixelIsArea convention is detected

gdal/frmts/hdf5/bagdataset.cpp
GlenRice-NOAA: Our understanding of the meaning of these errors is limited and we are more than happy to address these cases as you suggest.  Our only goal here is to ignore the upper right corner rather than scale the resolution.  Since BAGs should not be pixelisarea we wanted to remove that as an option for writing as well.

autotest/gdrivers/bag.py
GlenRice-NOAA: We were attempting to add a test case for BAGs that were being assessed as pixelisarea due to the upper right corner being offset to ensure they were being handled as intended.

gdal/frmts/hdf5/bagdataset.cpp
```diff
@@ -3344,8 +3344,8 @@ void BAGDataset::LoadMetadata()
                     fabs((dfURY - dfLLY) / dfResHeight - m_nLowResHeight) < 1e-2 )
                 {
                     // Found with https://data.ngdc.noaa.gov/platforms/ocean/nos/coast/H12001-H14000/H12525/BAG/H12525_MB_4m_MLLW_1of2.bag
-                    // https://github.com/OSGeo/gdal/issues/1643
-                    CPLError(CE_Failure, CPLE_AppDefined,"Ignoring BAG Upper Right Corner.");
+                    // to address issue https://github.com/OSGeo/gdal/issues/1643
+                    CPLError(CE_Warning, CPLE_AppDefined, "cornerPoints not consistent with resolution given in metadata");
```
rouault: I'd suggest a more specific message for that case: "cornerPoints and resolution in metadata seem to imply a PixelCorner convention, whereas BAG mandates PixelCenter. GDAL will georeference the product assuming this non-standard PixelCorner convention"

gdal/frmts/hdf5/bagdataset.cpp
GlenRice-NOAA: Our understanding of the previous code was that PixelIsArea was assumed when the upper left corner was not written correctly.  This was resulting in the resolution being scaled, which is not what was intended.  As was described in #1643, the BAG is defined by the lower lefter corner and the resolution.  The upper right corner is only written because of our understanding of  ISO 19115, and apparently is being written incorrectly because there is no check on its validity when written by the associated software.  We hope to correct this.

rouault: > Our understanding of the previous code was that PixelIsArea was assumed when the upper left corner was not written correctly. This was resulting in the resolution being scaled, which is not what was intended.

your proposed changes regarding set adfGeoTransform[] look good, except than the half-pixel shift of the origin must not be done in the PixelIsArea/PixelCorner case.

GlenRice-NOAA: We did the half shift to the extracted GeoTransform because it was consistent with the rest of the GDAL BAG driver.  It looked like the way the driver was written (and all the tests worked) took care of the shift during reading and writing, thus the GDAL BAG object took on a PixelIsCorner convention.  Should it not be that way?  Should we should modify the GeoTransform (and all the tests) to keep the BAG as PixelIsArea as a GDAL object? There must be an attribute we would set that informs a transform to or from other object type of any conversion that needs to be made to the associated GeoTransform.

rouault: > We did the half shift to the extracted GeoTransform because it was consistent with the rest of the GDAL BAG driver

That's fine. I meant that when the convention of the product is detected to be the non-standard/erroneous PixelCorner/PixelIsArea then this half-pixel shift is not needed, since it is already in the georeferencing convention used by GDAL.

GlenRice-NOAA: From our perspective the non-standard detection indicates that the upper right corner was written incorrectly and thus a warning should be thrown.  It does not mean the BAG is PixelIsCorner.  Since the code (now) always recalculates the upper right corner there is no need to make the half pixel shift conditional.  The resolution and lower left corner are always correct and the raster is always PixelIsArea.

rouault: >  It does not mean the BAG is PixelIsCorner

Hum, perhaps... Is there a way of checking what the correct georeferencing would be for a product like
https://data.ngdc.noaa.gov/platforms/ocean/nos/coast/H12001-H14000/H12525/BAG/H12525_MB_4m_MLLW_1of2.bag which was referenced in #1643 ? If indeed this is just an issue with the computation of the upper-right corner, then https://github.com/OSGeo/gdal/pull/1728/commits/d455285c787046d05e1215496441f7248bc4094a should be reverted. This is becoming confusing... :-)

zacharyburnettNOAA:
No, I don't think there are any more issues. Thank you for reviewing the changes.

gdal/frmts/hdf5/bagdataset.cpp
GlenRice-NOAA: We went back and compared the source raster with the derived BAG for a few of our products and found the lower left corner and resolution were correct.  The upper right corner was reported as being one grid resolution away from what is reported in the BAG, so clearly it was being ignored by the software.  We think this confirms the steps taken here.
We often say these things are not hard, but there are confusing.  Thanks for accepting the work.

PR #1740, opened 2019-07-24
Multiple Layer Datasets for netCDF CF-1.8, and Other Updates

wchen329:
## What does this PR do?
This PR contains the following updates:
* Introduces the capability to write CF-1.8 datasets consisting of multiple layers. 
* Enabled "casting" between related (and unrelated) OGR types when writing to- among other reasons- support the -nlt flag and improve compatibility with datasets which may mix different feature types. 
* Refactor the storage of layers to be managed, specifically by shared_ptrs (something like this was originally requested before)
* Allow measured feature types (with a warning that explains how writing measured types will lose the Measure).
* Further integrated CF-1.8 implementation of Point writing into the Geometry Scribe (which was and is where other CF-1.8 geometry writing is done) to benefit from buffering, and allow multiple layers in NC3. Also: combined record dimensions with existing dimensions for CF-1.8 datasets.
* netCDFVirtual and Scribe- write system refactoring (including constant time dimension resize). See more below.
* Resilience improvements including shutting down reading when a negative value of part_node_count or node_count is encountered (or interior ring is not 0 or 1)
* Multiple datum writing mode for non String types (could implement String types in the future maybe?)
* Default buffer size increase to 20%.
* Fixed issue with writing empty Multipolygons.
* Documentation updates.

Issue #1287 

## Writing System Rewrite
Previously, the CF-1.8 writer used a rather simple writing scheme- a state based system that would persist only between layer translation invocations and reset between each write. Unfortunately, the problem with this system is that it would use in-place dim resizing (which involved a lot of extra copying, with perhaps worse than linear complexity and especially slow for very large datasets with multiple layers). 

The new CF-1.8 writer introduces something called "netCDFVirtualID." Essentially what this does is present new- but similar- interfaces to dim and var definition in a netCDF file, but does not write the variable to the dataset immediately. Instead, dim and var metadata is held in memory allowing dims to be resized at constant complexity. Until the dim is actually written through the new nc_vsync call.

To accomplish this, a small subset of netCDF API functions were brought to the netCDFVirtual layer; defining a variable or dimension in the netCDF virtual layer gives virtual netCDF IDs which are not guaranteed to be the same as the real netCDF ID. Attribute writing is also available (done through templating), so long as the variable exists ONLY in the virtual layer.

Using the virtual layer is specific to this use case of dimension resizing- in cases where no dimensions exist or are known ahead of time, using the real netCDF APIs directly is generally recommended (as the virtual layer would only slightly slow things down).

GeometryScribe has also been separated into two classes: a "ncScribe" class which manages general transaction handling to the netCDF file and a "Metadata" class which manages the layer's IDs in the netCDFVirtual layer, and the geometry container ID in the real netCDF file.

## Tasklist

 - [x] Implement Multiple Layer netCDF-3 and netCDF-4, netCDF-4C Writing
 - [x] Implement unified buffer restriction for Field and Geometry Buffers
 - [x] Enable side-casting between OGR simple geometry types (writing only)
 - [x] CF-1.8 Writing Infrastructure Rewrite
 - [x] Logging
 - [x] Add test case(s)
 - [x] Review
 - [x] Adjust for comments

## Is this Ready for Review?
:+1: hopefully

## Environment

* OS: openSUSE Tumbleweed and Windows 7 Professional AMD64
* Compiler: GCC 9.1.1 (SUSE) and MSVC 14.0 / 2015 (Windows)


wchen329:
@rouault I think this PR is ready for review. I might add some test cases for measured, though.

Also, how would one go about detecting the -append switch (or just append access mode, would you say)? What I've wrote so far might not really work on append, so I'd like to save the work for another PR, so I'd like to disable it for now if possible.

wchen329:
@rouault Actually, nevermind for now. Let me make some additional changes before you make a review.

wchen329:
There we go. @rouault I think this PR is ready for review. I probably will add a few more test cases to test out "transaction queueing" some more, but other than that I think for not-fix/non-modification related additions to the PR this is more or less it.

By the way- in the known issues section the speed optimization I am referring to is to batch the writes and write using nc_put_vara rather than nc_put_var1. I might sneak that in here but that's a decent amount of work (restructure some parts of the transaction classes) so it might need its own PR.

wchen329:
I'll give the first comment:

Edit docs to say how "growDims" option is ignored for CF-1.8 (it's now done regardless since it is constant time for those datasets).

rouault:
autotest/gdrivers/netcdf.py:4749:1: F811 redefinition of unused 'test_empty_multiline_read_write' from line 4729

wchen329:
> 
> 
> autotest/gdrivers/netcdf.py:4749:1: F811 redefinition of unused 'test_empty_multiline_read_write' from line 4729

Hmm... that's interesting

I'll just, you know, quietly delete it :+1:

wchen329:
@rouault While I'm chipping away at the CI stuff what else would you like to have changed in the PR?

gdal/frmts/netcdf/netcdfdataset.h
```diff
@@ -942,6 +953,8 @@ class netCDFDataset final: public GDALPamDataset
     netCDFDataset();
     virtual ~netCDFDataset();
     void SGCommitPendingTransaction();
+    void SGLogPendingTransaction();
+    std::shared_ptr<std::string> generateLogName();
```
rouault: huh a shared_ptr<std::string> ? This is pretty unusual. Why not a straight std::string ?

gdal/frmts/netcdf/netcdflayersg.cpp
```diff
@@ -299,3 +361,17 @@ OGRFeature* netCDFLayer::buildSGeometryFeature(size_t featureInd)
     feat -> SetFID(featureInd);
     return feat;
 }
+
+std::shared_ptr<std::string> netCDFDataset::generateLogName()
+{
+    const char * lognumc;
+
+    if(this->logCount == 0)
```
rouault: you don't need that complication. CPLGenerateTempFilename() has logic to generate a unique name and has its own counter. You can call it with nullptr. So bsically this method could do just: return CPLGetnerateTempFilename(nullptr);

gdal/frmts/netcdf/netcdflayer.cpp
```diff
@@ -118,20 +118,41 @@ static void netCDFWriteAttributesFromConf(
         }
         else if( EQUAL(oAtt.m_osType, "string") )
         {
-            status = nc_put_att_text(cdfid, varid, oAtt.m_osName,
-                                     oAtt.m_osValue.size(), oAtt.m_osValue);
+            if(m_bLegacyCreateMode)
```
rouault: see above remarks. could be merged in a single code path with a passthrough vcdf implementation

gdal/frmts/netcdf/netcdflayer.cpp
```diff
@@ -118,20 +118,41 @@ static void netCDFWriteAttributesFromConf(
...
+            {
+                m_poDS->vcdf.nc_put_vatt_text(varid, oAtt.m_osName, oAtt.m_osValue);
+            }
         }
         else if( EQUAL(oAtt.m_osType, "integer") )
         {
             int nVal = atoi(oAtt.m_osValue);
-            status =
-                nc_put_att_int(cdfid, varid, oAtt.m_osName, NC_INT, 1, &nVal);
+            if(m_bLegacyCreateMode)
```
rouault: see above remarks. could be merged in a single code path with a passthrough vcdf implementation

gdal/frmts/netcdf/netcdflayer.cpp
```diff
@@ -118,20 +118,41 @@ static void netCDFWriteAttributesFromConf(
...
+            else
+            {
+                m_poDS->vcdf.nc_put_vatt_int(varid, oAtt.m_osName, &nVal);
+            }
         }
         else if( EQUAL(oAtt.m_osType, "double") )
         {
             double dfVal = CPLAtof(oAtt.m_osValue);
-            status = nc_put_att_double(cdfid, varid, oAtt.m_osName,
+            if(m_bLegacyCreateMode)
```
rouault: see above remarks. could be merged in a single code path with a passthrough vcdf implementation

gdal/frmts/netcdf/netcdflayer.cpp
```diff
@@ -162,8 +185,21 @@ bool netCDFLayer::Create(char **papszOptions,
         newbufsize = strtoll(memorySizeLimitation, nullptr, 10);
     }
 
-    m_osRecordDimName = CSLFetchNameValueDef(papszOptions, "RECORD_DIM_NAME",
+    const char * singleDatumMode = CSLFetchNameValueDef(papszOptions, "GROUPLESS_WRITE_BACK", "NO");
+    std::string singleDatumMode_s = std::string(singleDatumMode);
+
+    if(singleDatumMode_s == "YES")
```
rouault: use CPLTestBool() instead

gdal/frmts/netcdf/netcdflayer.cpp
```diff
@@ -1526,30 +1552,57 @@ bool netCDFLayer::FillVarFromFeature(OGRFeature *poFeature, int nMainDimId,
...
                         size_t nNewSize = anCount[1] + anCount[1] / 3;
 
                         CPLDebug("GDAL_netCDF", "Growing %s from %u to %u",
                                  m_poFeatureDefn->GetFieldDefn(i)->GetNameRef(),
                                  static_cast<unsigned>(nWidth),
                                  static_cast<unsigned>(nNewSize));
-                        m_poDS->GrowDim(m_nLayerCDFId,
-                                        m_aoFieldDesc[i].nSecDimId, nNewSize);
+
+                        if(m_bLegacyCreateMode)
```
rouault: see above remarks. could be merged in a single code path with a passthrough vcdf implementation

gdal/frmts/netcdf/netcdflayer.cpp
```diff
@@ -1609,8 +1701,17 @@ bool netCDFLayer::FillVarFromFeature(OGRFeature *poFeature, int nMainDimId,
         {
             int nVal = poFeature->GetFieldAsInteger(i);
             short sVal = static_cast<short>(nVal);
-            status = nc_put_var1_short(m_nLayerCDFId, m_aoFieldDesc[i].nVarId,
-                                       anIndex, &sVal);
+
+            if (m_poDS->HasInfiniteRecordDim())
+            {
+                            status = nc_put_var1_short(m_nLayerCDFId, m_aoFieldDesc[i].nVarId,
```
rouault: indentation

gdal/frmts/netcdf/netcdflayer.cpp
```diff
@@ -2324,15 +2514,31 @@ OGRErr netCDFLayer::CreateField(OGRFieldDefn *poFieldDefn, int /* bApproxOK */)
         if( poFieldDefn->GetWidth() == 1 )
         {
             nType = NC_CHAR;
-            status = nc_def_var(m_nLayerCDFId, pszVarName, nType, 1,
+
+            if(!m_bLegacyCreateMode)
```
rouault: see above remarks. could be merged in a single code path with a passthrough vcdf implementation

gdal/frmts/netcdf/netcdflayer.cpp
```diff
@@ -2324,15 +2514,31 @@ OGRErr netCDFLayer::CreateField(OGRFieldDefn *poFieldDefn, int /* bApproxOK */)
...
+            {
+                nVarID = m_poDS->vcdf.nc_def_vvar(pszVarName, nType, 1, &nMainDimId);
```
rouault: why nc_def_vvar and not nc_def_var ?

gdal/frmts/netcdf/netcdflayer.cpp
```diff
@@ -2324,15 +2514,31 @@ OGRErr netCDFLayer::CreateField(OGRFieldDefn *poFieldDefn, int /* bApproxOK */)
...
                                 &nMainDimId, &nVarID);
+            }
         }
 #ifdef NETCDF_HAS_NC4
         else if( m_poDS->eFormat == NCDF_FORMAT_NC4 && m_bUseStringInNC4 )
         {
             nType = NC_STRING;
-            status = nc_def_var(m_nLayerCDFId, pszVarName, nType, 1,
-                                &nMainDimId, &nVarID);
+            if(!m_bLegacyCreateMode)
```
rouault: see above remarks. could be merged in a single code path with a passthrough vcdf implementation

gdal/frmts/netcdf/netcdflayer.cpp
```diff
@@ -2341,9 +2547,18 @@ OGRErr netCDFLayer::CreateField(OGRFieldDefn *poFieldDefn, int /* bApproxOK */)
             {
                 if( m_nDefaultMaxWidthDimId < 0 )
                 {
-                    status =
-                        nc_def_dim(m_nLayerCDFId, "string_default_max_width",
-                                   m_nDefaultWidth, &m_nDefaultMaxWidthDimId);
+                    if(!m_bLegacyCreateMode)
```
rouault: see above remarks. could be merged in a single code path with a passthrough vcdf implementation

gdal/frmts/netcdf/netcdflayer.cpp
```diff
@@ -2357,9 +2572,18 @@ OGRErr netCDFLayer::CreateField(OGRFieldDefn *poFieldDefn, int /* bApproxOK */)
                 size_t nDim = poFieldDefn->GetWidth() == 0
                                   ? m_nDefaultWidth
                                   : poFieldDefn->GetWidth();
+            if(!m_bLegacyCreateMode)
```
rouault: see above remarks. could be merged in a single code path with a passthrough vcdf implementation

gdal/frmts/netcdf/netcdflayer.cpp
```diff
@@ -2370,9 +2594,19 @@ OGRErr netCDFLayer::CreateField(OGRFieldDefn *poFieldDefn, int /* bApproxOK */)
             nDimCount = 2;
             int anDims[2] = {nMainDimId, nSecDimId};
             nType = NC_CHAR;
-            status = nc_def_var(m_nLayerCDFId, pszVarName, nType, 2, anDims,
-                                &nVarID);
+
+            if(!m_bLegacyCreateMode)
```
rouault: see above remarks. could be merged in a single code path with a passthrough vcdf implementation

gdal/frmts/netcdf/netcdflayer.cpp
```diff
@@ -2395,8 +2629,17 @@ OGRErr netCDFLayer::CreateField(OGRFieldDefn *poFieldDefn, int /* bApproxOK */)
         else if( nType == NC_INT )
             nodata.nVal = NC_FILL_INT;
 
-        status = nc_def_var(m_nLayerCDFId, pszVarName, nType, 1, &nMainDimId,
-                            &nVarID);
+        if(!m_bLegacyCreateMode)
```
rouault: see above remarks. could be merged in a single code path with a passthrough vcdf implementation

gdal/frmts/netcdf/netcdflayer.cpp
```diff
@@ -2424,8 +2670,18 @@ OGRErr netCDFLayer::CreateField(OGRFieldDefn *poFieldDefn, int /* bApproxOK */)
             nodata.nVal64 = NC_FILL_INT64;
         }
 #endif
-        status = nc_def_var(m_nLayerCDFId, pszVarName, nType, 1, &nMainDimId,
-                            &nVarID);
+
+        if(!m_bLegacyCreateMode)
```
rouault: see above remarks. could be merged in a single code path with a passthrough vcdf implementation

gdal/frmts/netcdf/netcdflayer.cpp
```diff
@@ -2441,8 +2697,18 @@ OGRErr netCDFLayer::CreateField(OGRFieldDefn *poFieldDefn, int /* bApproxOK */)
             nodata.fVal = NC_FILL_FLOAT;
         else
             nodata.dfVal = NC_FILL_DOUBLE;
-        status = nc_def_var(m_nLayerCDFId, pszVarName, nType, 1, &nMainDimId,
-                            &nVarID);
+
+        if(!m_bLegacyCreateMode)
```
rouault: see above remarks. could be merged in a single code path with a passthrough vcdf implementation

gdal/frmts/netcdf/netcdflayer.cpp
```diff
@@ -2455,42 +2721,81 @@ OGRErr netCDFLayer::CreateField(OGRFieldDefn *poFieldDefn, int /* bApproxOK */)
     case OFTDate:
     {
         nType = NC_INT;
-        status = nc_def_var(m_nLayerCDFId, pszVarName, nType, 1, &nMainDimId,
-                            &nVarID);
+
+        if(!m_bLegacyCreateMode)
```
rouault: see above remarks. could be merged in a single code path with a passthrough vcdf implementation

gdal/frmts/netcdf/netcdflayer.cpp
```diff
@@ -2455,42 +2721,81 @@ OGRErr netCDFLayer::CreateField(OGRFieldDefn *poFieldDefn, int /* bApproxOK */)
...
+
         NCDF_ERR(status);
         if( status != NC_NOERR )
         {
             return OGRERR_FAILURE;
         }
         nodata.nVal = NC_FILL_INT;
 
-        status = nc_put_att_text(m_nLayerCDFId, nVarID, CF_UNITS,
+        if(!m_bLegacyCreateMode)
```
rouault: see above remarks. could be merged in a single code path with a passthrough vcdf implementation

gdal/frmts/netcdf/netcdflayer.cpp
```diff
@@ -2455,42 +2721,81 @@ OGRErr netCDFLayer::CreateField(OGRFieldDefn *poFieldDefn, int /* bApproxOK */)
...
         break;
     }
 
     case OFTDateTime:
     {
         nType = NC_DOUBLE;
-        status = nc_def_var(m_nLayerCDFId, pszVarName, nType, 1, &nMainDimId,
-                            &nVarID);
+
+        if(!m_bLegacyCreateMode)
```
rouault: see above remarks. could be merged in a single code path with a passthrough vcdf implementation

gdal/frmts/netcdf/netcdflayer.cpp
```diff
@@ -2455,42 +2721,81 @@ OGRErr netCDFLayer::CreateField(OGRFieldDefn *poFieldDefn, int /* bApproxOK */)
...
+
         NCDF_ERR(status);
         if( status != NC_NOERR )
         {
             return OGRERR_FAILURE;
         }
         nodata.dfVal = NC_FILL_DOUBLE;
 
-        status = nc_put_att_text(m_nLayerCDFId, nVarID, CF_UNITS,
+        if(!m_bLegacyCreateMode)
```
rouault: see above remarks. could be merged in a single code path with a passthrough vcdf implementation

gdal/frmts/netcdf/netcdflayer.cpp
```diff
@@ -2508,51 +2813,87 @@ OGRErr netCDFLayer::CreateField(OGRFieldDefn *poFieldDefn, int /* bApproxOK */)
...
-    status = nc_put_att_text(m_nLayerCDFId, nVarID, CF_LNG_NAME,
-                             strlen(pszLongName), pszLongName);
-    NCDF_ERR(status);
 
     if ( !m_bLegacyCreateMode )
     {
-        char ct_name_carr[NC_MAX_NAME + 1] = {0};
-        status = nc_inq_varname(m_nLayerCDFId, m_writableSGContVarID, ct_name_carr);
-        NCDF_ERR(status);
+        m_poDS->vcdf.nc_put_vatt_text(nVarID, CF_LNG_NAME, pszLongName);
```
rouault: see above remarks. could be merged in a single code path with a passthrough vcdf implementation

gdal/frmts/netcdf/netcdflayer.cpp
```diff
@@ -2508,51 +2813,87 @@ OGRErr netCDFLayer::CreateField(OGRFieldDefn *poFieldDefn, int /* bApproxOK */)
...
         NCDF_ERR(status);
     }
 
     if( m_bWriteGDALTags )
     {
-        status = nc_put_att_text(m_nLayerCDFId, nVarID, "ogr_field_name",
-                                 strlen(poFieldDefn->GetNameRef()),
-                                 poFieldDefn->GetNameRef());
-        NCDF_ERR(status);
+        if ( !m_bLegacyCreateMode )
```
rouault: see above remarks. could be merged in a single code path with a passthrough vcdf implementation

gdal/frmts/netcdf/netcdflayer.cpp
```diff
@@ -2508,51 +2813,87 @@ OGRErr netCDFLayer::CreateField(OGRFieldDefn *poFieldDefn, int /* bApproxOK */)
...
         if( eSubType != OFSTNone )
         {
             pszType = CPLSPrintf("%s(%s)", pszType,
                                  OGRFieldDefn::GetFieldSubTypeName(eSubType));
         }
-        status = nc_put_att_text(m_nLayerCDFId, nVarID, "ogr_field_type",
-                                 strlen(pszType), pszType);
-        NCDF_ERR(status);
+
+        if ( !m_bLegacyCreateMode )
```
rouault: see above remarks. could be merged in a single code path with a passthrough vcdf implementation

gdal/frmts/netcdf/netcdflayer.cpp
```diff
@@ -2508,51 +2813,87 @@ OGRErr netCDFLayer::CreateField(OGRFieldDefn *poFieldDefn, int /* bApproxOK */)
...
+            NCDF_ERR(status);
+        }
 
         const int nWidth = poFieldDefn->GetWidth();
         if(nWidth || nType == NC_CHAR)
         {
-            status = nc_put_att_int(m_nLayerCDFId, nVarID, "ogr_field_width",
-                                    NC_INT, 1, &nWidth);
-            NCDF_ERR(status);
+            if ( !m_bLegacyCreateMode )
```
rouault: see above remarks. could be merged in a single code path with a passthrough vcdf implementation

gdal/frmts/netcdf/netcdflayer.cpp
```diff
@@ -2508,51 +2813,87 @@ OGRErr netCDFLayer::CreateField(OGRFieldDefn *poFieldDefn, int /* bApproxOK */)
...
 
             const int nPrecision = poFieldDefn->GetPrecision();
             if( nPrecision )
             {
-                status =
-                    nc_put_att_int(m_nLayerCDFId, nVarID, "ogr_field_precision",
-                                   NC_INT, 1, &nPrecision);
-                NCDF_ERR(status);
+
+                if ( !m_bLegacyCreateMode )
```
rouault: see above remarks. could be merged in a single code path with a passthrough vcdf implementation

gdal/frmts/netcdf/netcdflayer.cpp
```diff
@@ -2562,18 +2903,33 @@ OGRErr netCDFLayer::CreateField(OGRFieldDefn *poFieldDefn, int /* bApproxOK */)
 
     if( !m_osGridMapping.empty() && nMainDimId == m_nRecordDimID )
     {
-        status =
-            nc_put_att_text(m_nLayerCDFId, nVarID, CF_GRD_MAPPING,
+        if ( !m_bLegacyCreateMode )
```
rouault: see above remarks. could be merged in a single code path with a passthrough vcdf implementation

gdal/frmts/netcdf/netcdflayer.cpp
```diff
@@ -2562,18 +2903,33 @@ OGRErr netCDFLayer::CreateField(OGRFieldDefn *poFieldDefn, int /* bApproxOK */)
...
+        }
     }
 
     if( !m_osCoordinatesValue.empty() && nMainDimId == m_nRecordDimID )
     {
-        status = nc_put_att_text(m_nLayerCDFId, nVarID, CF_COORDINATES,
-                                 m_osCoordinatesValue.size(),
-                                 m_osCoordinatesValue.c_str());
-        NCDF_ERR(status);
+        if ( !m_bLegacyCreateMode )
```
rouault: see above remarks. could be merged in a single code path with a passthrough vcdf implementation

gdal/frmts/netcdf/netcdflayersg.cpp
```diff
@@ -74,36 +74,42 @@ namespace nccfdriver
     {
         geom_t ret = NONE;
 
-        if (type == wkbPoint || type == wkbPoint25D)
+        if (type == wkbPoint || type == wkbPoint25D || type == wkbPointM || type == wkbPointZM)
```
rouault: intead of testing all those combinations you could just have `auto eFlatType = wkbFlatten(type)`, and just test for eFlatType = wkbPoint, wkbLineString, etc.

gdal/frmts/netcdf/netcdflayersg.cpp
```diff
@@ -120,22 +126,24 @@ CPLErr netCDFDataset::DetectAndFillSGLayers(int ncid)
     // Discover simple geometry variables
     int var_count;
     nc_inq_nvars(ncid, &var_count);
-    std::vector<int> vidList;
+    std::set<int> vidList;
 
     nccfdriver::scanForGeometryContainers(ncid, vidList);
 
-    for(size_t itr = 0; itr < vidList.size(); itr++)
+    if(vidList.size() > 0)
```
rouault: !vidList.empty()

gdal/frmts/netcdf/netcdflayersg.cpp
```diff
@@ -120,22 +126,24 @@ CPLErr netCDFDataset::DetectAndFillSGLayers(int ncid)
...
     {
-        try
+        for(std::set<int>::iterator itr = vidList.begin(); itr != vidList.end(); ++itr)
```
rouault: for( auto vid: vidList )

gdal/frmts/netcdf/netcdfsg.cpp
```diff
@@ -767,8 +794,17 @@ namespace nccfdriver
             // If matches, then establish a reference by placing this variable's data in both vectors
             if(!strcmp(contname, buf))
             {
-                char property_name[NC_MAX_NAME];
-                nc_inq_varname(this->nc, curr, property_name);
+                char property_name[NC_MAX_NAME] = {0};
```
rouault: in most similar places we use NC_MAX_NAME+1

gdal/frmts/netcdf/netcdfsgwriterutil.cpp
```diff
@@ -196,8 +208,127 @@ namespace nccfdriver
...
+        if(aosNcoord.size() > 2)
+        {
+            new_varID = ncdf.nc_def_vvar(aosNcoord[2], NC_DOUBLE, 1, &node_coordinates_dimID);
+            ncdf.nc_put_vatt_text(new_varID, CF_AXIS, CF_SG_Z_AXIS);
+
+            this->node_coordinates_varIDs.push_back(new_varID);
+        }
+    }
+
+/*    void ncLayer_SG_Metadata::scanExistingContainer(int containerVID)
```
rouault: to remove ?

gdal/frmts/netcdf/netcdfsgwriterutil.h
```diff
@@ -68,67 +72,333 @@ namespace nccfdriver
...
+            /* void setVarId(...);
+             * Sets the var in which to commit the transaction to.
+             */
+            void setVarId(int vId) { this->varId = vId; }
+
+    };
+
+    typedef std::map<int, void*> NCWMap;
+    typedef std::pair<int, void*> NCWEntry; // NC Writer Entry
+    typedef std::shared_ptr<OGR_SGFS_Transaction> MTPtr; // a.k.a Managed Transaction Ptr
```
rouault: couldn't that be a unique_ptr ? That is are there several objects who can store the same MTPtr ?

gdal/frmts/netcdf/netcdfsgwriterutil.cpp
```diff
@@ -277,547 +408,554 @@ namespace nccfdriver
...
+                     return MTPtr(); // read failure
+
+                 size_t strsize;
+
+                 // get how much data to read
+                 readcheck = VSIFReadL(&strsize, sizeof(size_t), 1, log);
+                 if(!readcheck)
+                     return MTPtr(); // read failure
+
+                 std::unique_ptr<char, std::default_delete<char[]>> data(new char[strsize + 1]);
```
rouault: more complicated than needed. Just use `std::string str; str.resize(strsize);`

gdal/frmts/netcdf/netcdfsgwriterutil.cpp
```diff
@@ -277,547 +408,554 @@ namespace nccfdriver
...
+                 if(!op)
+                 {
+                     return MTPtr(new OGR_SGFS_NC_Char_Transaction(varId, data.get()));  // data is copied so okay!
+                 }
+
+                 // case: its a CHARA op, additional processing
+                 else
+                 {
+                     size_t fullSizeCount;
+                     if(!VSIFReadL(&fullSizeCount, sizeof(size_t), 1, log))
```
rouault: are there situations where fullSizeCount would be != strsize ?

gdal/frmts/netcdf/netcdfsgwriterutil.cpp
```diff
@@ -277,547 +408,554 @@ namespace nccfdriver
...
+
+                 size_t strsize;
+
+                 // get how much data to read
+                 readcheck = VSIFReadL(&strsize, sizeof(size_t), 1, log);
+
+                 if(!readcheck)
+                     return MTPtr(); // read failure
+
+                 std::unique_ptr<char, std::default_delete<char[]>> data(new char[strsize + 1]);
```
rouault: use std::string

gdal/frmts/netcdf/netcdfsgwriterutil.h
```diff
@@ -68,67 +72,333 @@ namespace nccfdriver
...
+    {
+        T_c_type rep = r;
+        int varId = vId;
+        int type = T_nc_type;
+        VSIFWriteL(&varId, sizeof(int), 1, f); // write varID data
+        VSIFWriteL(&type, sizeof(int), 1, f); // write NC type
+        VSIFWriteL(&rep, sizeof(T_c_type), 1, f); // write data
+    }
+
+    template<class T_c_type, class T_r_type> std::shared_ptr<OGR_SGFS_Transaction> genericLogDataRead(int varId, VSILFILE* f)
```
rouault: unique_ptr instead ?

gdal/frmts/netcdf/netcdfsgwriterutil.h
```diff
@@ -68,67 +72,333 @@ namespace nccfdriver
...
+            * DO NOT use this when a commit is taking place, otherwise 
+            * corruption may occur...
+            */
+           void setSingleDatumMode(bool sdm) { this->singleDatumMode = sdm; }
+
+    };
+
+    class ncLayer_SG_Metadata
+    {
+        int & ncID;
```
rouault: does this need to be a reference ? If so, a comment why wouldn't hurt

gdal/frmts/netcdf/netcdfvirtual.cpp
```diff
@@ -0,0 +1,330 @@
...
+        int dimID = dimTicket;
+
+        // Check if name is already defined
+        if(nameDimTable.count(std::string(name)) > 0)
+        {
+            throw SG_Exception_DupName(name, "virtual dimension collection");
+        }
+
+        // Add to lookup tables
+                dimList.push_back(netCDFVDimension(name, len, dimTicket++));
```
rouault: indentation.

rouault: I'd prefer dimTicket++ to be done in a separate statement

gdal/frmts/netcdf/netcdfvirtual.cpp
```diff
@@ -0,0 +1,330 @@
...
+        int varID = varTicket;
+
+        // Check if name is already defined
+        if(nameVarTable.count(std::string(name)) > 0)
+        {
+            throw SG_Exception_DupName(name, "virtual variable collection");
+        }
+
+        // Add to lookup tables
+        varList.push_back(netCDFVVariable(name, xtype, ndims, dimidsp, varTicket++));
```
rouault: varTicket++ in a separate statement

gdal/frmts/netcdf/netcdfvirtual.cpp
```diff
@@ -0,0 +1,330 @@
...
+    {
+        nc_redef(ncid);
+    }
+
+    void netCDFVID::nc_set_data_mode()
+    {
+        nc_enddef(ncid);
+    }
+
+        void netCDFVID::nc_vmap()
```
rouault: indentation

gdal/frmts/netcdf/netcdfvirtual.cpp
```diff
@@ -0,0 +1,330 @@
...
+            throw SG_Exception_BadMapping(name.c_str(), "dimension ID lookup");
+        }
+        return nameDimTable.at(name);
+    }
+
+    /* Attribute writing
+     *
+     */
+
+    void netCDFVID::nc_put_vatt_text(int varid, const char* name, const char* out)
```
rouault: `out` is a strange name for something const... Shouldn't that be `value` or something similar. Applies to many below similar instances

gdal/frmts/netcdf/netcdfvirtual.h
```diff
@@ -0,0 +1,359 @@
...
+        int v_did;
+        size_t dim_len;
+        bool valid = true;
+
+        protected:
+            void setRealID(int realID) { this->r_did = realID; }
+            void invalidate();
+            void setLen(size_t len) { this->dim_len = len; }
+        public:
+                    netCDFVDimension(const char * name, size_t len, int dimid) :
```
rouault: indentation

gdal/frmts/netcdf/netcdfvirtual.h
```diff
@@ -0,0 +1,359 @@
...
+     * -
+     * Contains the variable name, variable type, etc.
+     */
+    class netCDFVVariable
+    {
+        friend class netCDFVID;
+
+        std::string real_var_name;
+        nc_type ntype;
+                int r_vid = INVALID_VAR_ID;
```
rouault: indentation

gdal/frmts/netcdf/netcdfvirtual.h
```diff
@@ -0,0 +1,359 @@
...
+     *
+     * Also, virtual attributes only exist until the variable is committed. Use "real" attributes and "real" IDs
+     * for a variable after its been commited.
+     *
+     * ** Do not mix netCDF virtual dim and variable IDs with regular netCDF dim (a.k.a. "real")
+     * ids and variable ids. They are NOT necessarily compatible, and must be translated first, to be
+     * used in this manner **
+     */
+    class netCDFVID
+        {
```
rouault: indentation

gdal/frmts/netcdf/netcdfvirtual.h
```diff
@@ -0,0 +1,359 @@
...
+             */
+            void nc_set_data_mode();
+
+            /* nc_vmap()
+             * Maps virtual IDs to real physical ID if that mapping doesn't already exist
+             * This is required before writing data to virtual IDs that do not exist yet in the netCDF file
+             */
+            void nc_vmap();
+
+                        // Attribute function(s)
```
rouault: indentation

gdal/frmts/netcdf/netcdfvirtual.h
```diff
@@ -0,0 +1,359 @@
...
+#ifdef NETCDF_HAS_NC4
+            void nc_put_vvar1_string(int varid, const size_t* index, const char** out);
+#endif
+
+            // Equivalent "enquiry" functions
+            netCDFVVariable& virtualVIDToVar(int virtualID); // converts a virtual var ID to a real ID
+            netCDFVDimension& virtualDIDToDim(int virtualID); // converts a virtual dim ID to a real ID
+            int nameToVirtualVID(std::string& name);
+            int nameToVirtualDID(std::string& name);
+                        bool virtualVarNameDefined(std::string& nm) { return nameVarTable.count(nm) > 0; }
```
rouault: indentation

gdal/frmts/netcdf/netcdfvirtual.h
```diff
@@ -0,0 +1,359 @@
...
+     * Also, virtual attributes only exist until the variable is committed. Use "real" attributes and "real" IDs
+     * for a variable after its been commited.
+     *
+     * ** Do not mix netCDF virtual dim and variable IDs with regular netCDF dim (a.k.a. "real")
+     * ids and variable ids. They are NOT necessarily compatible, and must be translated first, to be
+     * used in this manner **
+     */
+    class netCDFVID
+        {
+        int & ncid;
```
rouault: does that needs to be a reference ?

gdal/frmts/netcdf/netcdfdataset.cpp
```diff
@@ -4470,39 +4480,66 @@ int NCDFWriteSRSVariable(int cdfid, const OGRSpatialReference* poSRS,
     return NCDFVarID;
 }
 
+// Use a default name (variable name: crs)
+int NCDFWriteSRSVariable(int cdfid, const OGRSpatialReference* poSRS,
```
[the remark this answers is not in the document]
wchen329: Do you mean passing it with a default argument?

I tried to pass it as a default argument once, but it seemed like the GDAL/OGR build process doesn't allow default arguments, so this is my workaround is to do explicit overloading.

Unless you mean something else?

gdal/frmts/netcdf/netcdfdataset.cpp
```diff
@@ -4470,39 +4480,66 @@ int NCDFWriteSRSVariable(int cdfid, const OGRSpatialReference* poSRS,
...
 /************************************************************************/
 
-void NCDFWriteLonLatVarsAttributes(int cdfid, int nVarLonID, int nVarLatID)
+void NCDFWriteLonLatVarsAttributes(int cdfid, int nVarLonID, int nVarLatID, nccfdriver::netCDFVID * vcdf)
 {
-    int status =
-        nc_put_att_text(cdfid, nVarLatID, CF_STD_NAME,
-                        strlen(CF_LATITUDE_STD_NAME), CF_LATITUDE_STD_NAME);
-    NCDF_ERR(status);
+    if(vcdf == nullptr)
```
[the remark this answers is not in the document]
wchen329: I also found the amount of conditional checks for using "nc_put_vatt"

But as far as I know, in netCDF there is a total ordering on the creations of Dimension, Variables, and Attributes of Variables- that is, first you must define Dimensions for a variable then the Variables themselves, and then finally, the Attributes. Unfortunately since attributes depend on variables and variables depend on dimensions, using nc_put_att_* would not work due to the fact that the dimensions are not written until the very end of translation.

gdal/frmts/netcdf/netcdfdataset.cpp
```diff
@@ -4470,39 +4480,66 @@ int NCDFWriteSRSVariable(int cdfid, const OGRSpatialReference* poSRS,
...
+        vcdf->nc_put_vatt_text(nVarLatID, CF_STD_NAME, CF_LATITUDE_STD_NAME);
+        vcdf->nc_put_vatt_text(nVarLatID, CF_LNG_NAME, CF_LATITUDE_LNG_NAME);
+        vcdf->nc_put_vatt_text(nVarLatID, CF_UNITS, CF_DEGREES_NORTH);
+        vcdf->nc_put_vatt_text(nVarLonID, CF_STD_NAME, CF_LONGITUDE_STD_NAME);
+        vcdf->nc_put_vatt_text(nVarLonID, CF_LNG_NAME, CF_LONGITUDE_LNG_NAME);
+        vcdf->nc_put_vatt_text(nVarLonID, CF_UNITS, CF_DEGREES_EAST);
+    }
+}
+
+void NCDFWriteLonLatVarsAttributes(int cdfid, int nVarLonID, int nVarLatID)
```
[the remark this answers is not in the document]
wchen329: See above comment about using default arguments.

gdal/frmts/netcdf/netcdfdataset.cpp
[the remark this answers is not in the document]
rouault: > Do you mean passing it with a default argument?

yes

> I tried to pass it as a default argument once, but it seemed like the GDAL/OGR build process doesn't allow default arguments

That's not true

gdal/frmts/netcdf/netcdflayersg.cpp
wchen329: alright thanks, I'll change it to that :+1:

gdal/frmts/netcdf/netcdfdataset.cpp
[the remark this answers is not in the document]
wchen329: > 
> 
> > Do you mean passing it with a default argument?
> 
> yes
> 
> > I tried to pass it as a default argument once, but it seemed like the GDAL/OGR build process doesn't allow default arguments
> 
> That's not true

Weird. I'll try it again

gdal/frmts/netcdf/netcdfdataset.cpp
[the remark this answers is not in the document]
rouault: I don't really understand your answer here; What I mean is that
`            nc_put_att_text(cdfid, nVarLatID, CF_STD_NAME,
                            strlen(CF_LATITUDE_STD_NAME), CF_LATITUDE_STD_NAME);
`
could very well be implemented by
```
vcdf->nc_put_vatt_text(nVarLatID, CF_STD_NAME, CF_LATITUDE_STD_NAME);
```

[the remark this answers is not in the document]
wchen329: Oh. The reason why `nc_put_att_*` is still there is that the WKT implementation hasn't been modified to use the "virtual ID" (though I think with some more declarations it could be). In fact the reason why I called it "virtual ID" instead virtual dataset or whatever was so it could be used with groups in WKT datasets.

What I was trying to say above is that buffering is done because the variable itself isn't yet resident in the netCDF file- it's only in the netCDFVID at the moment so implementing nc_put_vatt_* in this manner would result in an error.

gdal/frmts/netcdf/netcdfdataset.cpp
```diff
@@ -2161,10 +2167,6 @@ netCDFDataset::~netCDFDataset()
     netCDFDataset::FlushCache();
     SGCommitPendingTransaction();
 
-    for(int i = 0; i < nLayers; i++)
```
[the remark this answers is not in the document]
wchen329: The layer doesn't use m_poDS in the object so looks alright. :+1:

gdal/frmts/netcdf/netcdfvirtual.h
wchen329: True, only dim growers need this. Since doesn't use dim growing maybe it doesn't have to be.
I guess I'm just kind of worried about the netCDF library sneakily deciding to change the netCDF ID which happened last few times. But sure we can try it without it.

gdal/frmts/netcdf/netcdfvirtual.cpp
wchen329: Hahaha indeed a strange name,

this was to make it consistent with the netCDF man page, but yeah we can do value :+1:

gdal/frmts/netcdf/netcdfvirtual.h
wchen329: Hmm... it segfaults if I change it from reference.

I'll go investigate with GDB later

gdal/frmts/netcdf/netcdfdataset.cpp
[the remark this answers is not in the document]
rouault: what I meant is that netCDFVID should be an abstract class, with 2 implementations: the one you did with this buffering, and another one that just calls directly the nc_XXXX() functions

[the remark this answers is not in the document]
wchen329: Ooh hmm that could be a neat solution. Sure, I'll go look into it for this and the rest below.

gdal/frmts/netcdf/netcdfsg.cpp
wchen329: Ooh yes, potential buffer overflow thanks :+1:

gdal/frmts/netcdf/netcdfsgwriterutil.cpp
wchen329: yeah, I had bigger plans for netCDFVID at first I guess...

gdal/frmts/netcdf/netcdfvirtual.h
wchen329: done

gdal/frmts/netcdf/netcdfvirtual.h
wchen329: done

gdal/frmts/netcdf/netcdfvirtual.cpp
wchen329: done

gdal/frmts/netcdf/netcdfvirtual.cpp
wchen329: done

gdal/frmts/netcdf/netcdflayersg.cpp
wchen329: done

gdal/frmts/netcdf/netcdflayersg.cpp
wchen329: done

gdal/frmts/netcdf/netcdflayersg.cpp
wchen329: done

gdal/frmts/netcdf/netcdflayer.cpp
wchen329: Resizable dimensions-

I think this relates what we just talked about. So I'll resolve this once the paths are combined.

gdal/frmts/netcdf/netcdfsgwriterutil.cpp
wchen329: This was because of the whole truncation thing, but I guess it's been since disabled for CF-1.8 so no?

So I guess yeah we could change that.

gdal/frmts/netcdf/netcdfdataset.cpp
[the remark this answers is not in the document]
wchen329: This is the declaration with default argument:
declaration

And this is the error that it gives (compile time)
error

Not sure why this type of thing doesn't work? This type of declaration works outside of GDAL.

gdal/frmts/netcdf/netcdfsgwriterutil.cpp
wchen329: The nice thing about using a char* here is that you have the char* pointer to write to. With an std::string you have a const_char* but you don't have the actual char* representation (which is of course, good encapsulation).

Since we're using VSILReadL / fread to process the binary log, we'll have to store it in some char* array- why not just read them all at once to an actual char* pointer? Otherwise, the thing to do would probably be reading it one by one and packing them into a string object.

gdal/frmts/netcdf/netcdflayer.cpp
wchen329: done

gdal/frmts/netcdf/netcdfvirtual.cpp
wchen329: :+1:

gdal/frmts/netcdf/netcdfsgwriterutil.cpp
rouault: > you don't have the actual char* representation

not true. If the string is not empty, then &str[0] is the char* representation

wchen329: > 
> 
> > you don't have the actual char* representation
> 
> not true. If the string is not empty, then &str[0] is the char* representation

Okay you got me, but that's kind of breaking the idea of encapsulating a string isn't it?

gdal/frmts/netcdf/netcdfdataset.h
wchen329: Will change it :+1:

gdal/frmts/netcdf/netcdfsgwriterutil.cpp
rouault: as far as I understand, the C++ standard guarantees that std::string is stored as contiguous char: see https://en.cppreference.com/w/cpp/string/basic_string : `The elements of a basic_string are stored contiguously, that is, for a basic_string s, &*(s.begin() + n) == &*s.begin() + n for any n in [0, s.size()), or, equivalently, a pointer to s[0] can be passed to functions that expect a pointer to the first element of a CharT[] array. `
This is something I heavily rely on.

wchen329: Fine, alright we'll use the string :). Although indexing the pointer of an underlying data structure is still a bit unnerving, you've convinced me that it is at least in the case of the std::string, safe.

gdal/frmts/netcdf/netcdfsgwriterutil.h
wchen329: Yes to unique_ptr and no to the non-uniqueness property. Done :+1: .

gdal/frmts/netcdf/netcdfsgwriterutil.h
wchen329: done.

gdal/frmts/netcdf/netcdfdataset.cpp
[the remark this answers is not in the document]
wchen329: Wait now I remember- it wasn't an actual error, it was a CI error about having default arguments on constructors on one of the Travis builds.

Anywho, was a bit of afraid that that applied here. Anyways...

gdal/frmts/netcdf/netcdfdataset.cpp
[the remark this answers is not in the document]
wchen329: As a small note: I decided to just implement this through a flag in the existing netCDFVID class instead of making a small class hierarchy in order to not make things too complicated (since the full netCDFVID provides some additional features not provided by the netCDF Library).

gdal/frmts/netcdf/netcdfsgwriterutil.h
wchen329: comment!

gdal/frmts/netcdf/netcdflayer.cpp
wchen329: Oh. Well the netCDFVID doesn't actually have a handle on m_poDS inside (and intentionally) so I think this specific one, concerning GrowDim, will be left the same for now.

wchen329:
@rouault  Thanks for the thorough review. I've went through and made necessary changes for your feedback (except the one that's left unresolved, but I don't think it's not too much to ask to not have the netCDFDataset handle inside the netCDFVID, since a VID can represent a group or a dataset, it's just, well, kind of awkward).

Now time to wait for CI...

Any other things you would like addressed?

rouault:
```
frmts/netcdf/netcdflayer.cpp:2481,style,unreadVariable,Variable 'status' is assigned a value that is never used.
frmts/netcdf/netcdfvirtual.cpp:67,style,unreadVariable,Variable 'err' is assigned a value that is never used.
frmts/netcdf/netcdfvirtual.cpp:98,style,unreadVariable,Variable 'err' is assigned a value that is never used.
frmts/netcdf/netcdfvirtual.cpp:262,style,unreadVariable,Variable 'err' is assigned a value that is never used.
frmts/netcdf/netcdfvirtual.cpp:277,style,unreadVariable,Variable 'err' is assigned a value that is never used.
frmts/netcdf/netcdfvirtual.cpp:292,style,unreadVariable,Variable 'err' is assigned a value that is never used.
frmts/netcdf/netcdfvirtual.cpp:307,style,unreadVariable,Variable 'err' is assigned a value that is never used.
frmts/netcdf/netcdfvirtual.cpp:322,style,unreadVariable,Variable 'err' is assigned a value that is never used.
```

wchen329:
@rouault Trusty and Macintosh builds failures due to unrelated seg. fault on PROJ test:

`osr/osr_ct_proj.py::test_proj[GoogleMercator(#3136)] Segmentation fault (core dumped)` (Trusty)

and OS X had a network failure apparently.


`BatchFetch request 127.0.0.1:8080/rda_api/tile/foo/bar/0/3.tif failed: Failed to connect to 127.0.0.1 port 8080: Connection reset by peer`

Could we possibly restart these specific ones? Also I don't know why but the Macintosh build keeps seg. faulting at the end for no reason with `4 xfailed`, it seems to skip all the netCDF tests, however?

wchen329:
Looks like the Macintosh build went dark for no good reason this time  :/.

`No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received`

Could you restart the build again, thanks. Sorry to have you keep doing this...

rouault:
Can you merge master in your branch ? I've commited a workaround to avoid a frequent stall in the Mac build

wchen329:
> Can you merge master in your branch ? I've commited a workaround to avoid a frequent stall in the Mac build

Alright, merge has been done! Here's hoping it all works out-

rouault:
Squashed and merged

wchen329:
Oh so the Macintosh build finally stopped randomly dying huh (at least stopped dying enough)?

Thanks for your efforts on getting the build to behave @rouault ! :+1: Glad this is finally merged... took a while.

PR #1742, opened 2019-07-24
FlatGeobuf driver

bjornharrtell:
## What does this PR do?

This PR implements initial support for FlatGeobuf (read/write) and is more or less feature complete. See https://github.com/bjornharrtell/flatgeobuf for more information about the format.

## Tasklist

 - [x] Await format spec 1.0
 - [x] Rebase on GDAL 3.x master
 - [x] Initial CRS read/write support
 - [x] Option to opt out spatial index generation
 - [x] Squash/cleanup commits
 - [x] Remove unintentional modications of ogr_geometry.h and ogrcsvdatasource.cpp
 - [x] Remove conf.sh
 - [x] Write docs
 - [x] Handle Z and M dimensions

## Known weaknesses

* Writing logic requires at least as much RAM as the final file size. To improve it, significant work must be done to be able to sort features on hilbert value using some external sorting algorithm.
* More test coverage and oss-fuzz would not hurt


gdal/ogr/ogrsf_frmts/GNUmakefile
```diff
@@ -1,7 +1,7 @@
 
 include ../../GDALmake.opt
 
-SUBDIRS-yes = generic geojson kml mem mitab vrt $(OGR_FORMATS)
+SUBDIRS-yes = generic geojson kml mem mitab vrt flatgeobuf $(OGR_FORMATS)
```
rouault: only mandatory formats for a minimal build should be specified here. so flatgeobuf should not

gdal/ogr/ogrsf_frmts/flatgeobuf/flatbuffers/base.h
```diff
@@ -0,0 +1,379 @@
+#ifndef FLATBUFFERS_BASE_H_
```
rouault: is it a generated file / license ?

gdal/ogr/ogrsf_frmts/flatgeobuf/flatbuffers/code_generators.h
```diff
@@ -0,0 +1,203 @@
+/*
```
rouault: is it a generated file ?

gdal/ogr/ogrsf_frmts/flatgeobuf/packedrtree.h
```diff
@@ -0,0 +1,75 @@
+#ifndef FLATGEOBUF_PACKEDRTREE_H_
```
rouault: non-generated file, right ? license ?

gdal/ogr/ogrsf_frmts/flatgeobuf/packedrtree.cpp
```diff
@@ -0,0 +1,341 @@
+#include "packedrtree.h"
```
rouault: non-generated file, right ? license ?

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
```diff
@@ -0,0 +1,829 @@
+#include "ogrsf_frmts.h"
```
rouault: missing copyright & license header

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobufdataset.cpp
```diff
@@ -0,0 +1,302 @@
+#include "ogr_flatgeobuf.h"
```
rouault: missing copyright & license header

gdal/ogr/ogrsf_frmts/flatgeobuf/ogr_flatgeobuf.h
```diff
@@ -0,0 +1,135 @@
+#ifndef OGR_FLATGEOBUF_H_INCLUDED
```
rouault: missing copyright & license header

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobufdataset.cpp
```diff
@@ -0,0 +1,302 @@
...
+
+#include "header_generated.h"
+
+static int OGRFlatGeobufDriverIdentify(GDALOpenInfo* poOpenInfo){
+    if( STARTS_WITH_CI(poOpenInfo->pszFilename, "FGB:") )
+        return TRUE;
+
+    if( STARTS_WITH(poOpenInfo->pszFilename, "/vsicurl") )
+    {
+        if( CPLGetValueType(CPLGetFilename(poOpenInfo->pszFilename)) ==
```
rouault: is this logic needed ? It seems it comes from the PBF driver

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobufdataset.cpp
```diff
@@ -0,0 +1,302 @@
...
+                CPL_VALUE_INTEGER )
+        {
+            return TRUE;
+        }
+    }
+
+    if( poOpenInfo->bIsDirectory )
+    {
+        if( CPLGetValueType(CPLGetFilename(poOpenInfo->pszFilename)) ==
```
rouault: same here

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobufdataset.cpp
```diff
@@ -0,0 +1,302 @@
...
+/************************************************************************/
+/*                                Open()                                */
+/************************************************************************/
+
+GDALDataset *OGRFlatGeobufDataset::Open(GDALOpenInfo* poOpenInfo)
+{
+    if( !OGRFlatGeobufDriverIdentify(poOpenInfo) || poOpenInfo->eAccess == GA_Update )
+        return nullptr;
+
+    VSILFILE *fp = poOpenInfo->fpL;
```
rouault: should shoud check that fp != nullptr

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobufdataset.cpp
```diff
@@ -0,0 +1,302 @@
...
+    CPLString osFilename(poOpenInfo->pszFilename);
+
+    uint64_t offset = sizeof(magicbytes);
+    CPLDebug("FlatGeobuf", "Start at offset (%lu)", sizeof(magicbytes));
+    VSIFSeekL(fp, offset, SEEK_SET);
+    uint32_t headerSize;
+    VSIFReadL(&headerSize, 4, 1, fp);
```
rouault: use CPL_LSBPTR32(&headerSize) so that it works on big endian CPUs

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobufdataset.cpp
```diff
@@ -0,0 +1,302 @@
...
+    CPLDebug("FlatGeobuf", "headerSize (%d)", headerSize);
+    GByte* buf = static_cast<GByte*>(VSI_MALLOC_VERBOSE(headerSize));
```
rouault: some validation of headerSize value would be welcome so that corrupted/hostile files don't blow up memory
buf != nullptr should be checked

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobufdataset.cpp
```diff
@@ -0,0 +1,302 @@
...
+    VSIFReadL(buf, 1, headerSize, fp);
+    auto header = GetHeader(buf);
```
rouault: what if GetHeader() fails ? exception ? nullptr ?

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobufdataset.cpp
```diff
@@ -0,0 +1,302 @@
...
+                                        CPL_UNUSED int nBands,
+                                        CPL_UNUSED int nXSize,
+                                        CPL_UNUSED int nYSize,
+                                        CPL_UNUSED GDALDataType eDT,
+                                        char ** /* papszOptions */)
+{
+    // First, ensure there isn't any such file yet.
+    VSIStatBufL sStatBuf;
+
+    if (strcmp(pszName, "/dev/stdout") == 0)
```
rouault: does streaming to /dev/stdout really work ?

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
```diff
@@ -0,0 +1,829 @@
...
+#include "ogr_flatgeobuf.h"
+
+using namespace flatbuffers;
+using namespace FlatGeobuf;
+
+OGRFlatGeobufLayer::OGRFlatGeobufLayer(const Header *poHeader, const char* pszFilename, uint64_t offset)
+{
+    CPLDebug("FlatGeobuf", "offset: %zu", offset);
+
+    if (poHeader == nullptr)
```
rouault: you can use CPLAssert(poHeader != nullpr).

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
```diff
@@ -0,0 +1,829 @@
...
+        auto type = toOGRFieldType(column->type());
+        OGRFieldDefn field(name, type);
+        m_poFeatureDefn->AddFieldDefn(&field);
+    }
+}
+
+OGRFlatGeobufLayer::~OGRFlatGeobufLayer()
+{
+    if (m_create) {
+        CPLDebug("FlatGeobuf", "Request to create %zu features", m_featuresCount);
```
rouault: can you move this to a dedicated method ?

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
```diff
@@ -0,0 +1,829 @@
...
+{
+    switch (wkbFlatten(m_eGType)) {
+        case OGRwkbGeometryType::wkbPoint: m_geometryType = GeometryType::Point; break;
+        case OGRwkbGeometryType::wkbMultiPoint: m_geometryType = GeometryType::MultiPoint; break;
+        case OGRwkbGeometryType::wkbLineString: m_geometryType = GeometryType::LineString; break;
+        case OGRwkbGeometryType::wkbMultiLineString: m_geometryType = GeometryType::MultiLineString; break;
+        case OGRwkbGeometryType::wkbPolygon: m_geometryType = GeometryType::Polygon; break;
+        case OGRwkbGeometryType::wkbMultiPolygon: m_geometryType = GeometryType::MultiPolygon; break;
+        default:
+            CPLError(CE_Fatal, CPLE_NotSupported, "toGeometryType: Unknown OGRwkbGeometryType %d", (int) m_eGType);
```
rouault: CE_Fatal should not be used here or anywhere else (other instances below). It crashes the process

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
```diff
@@ -0,0 +1,829 @@
...
+                // For the root authority name 'EPSG', the authority code
+                // should always be integral
+                nAuthorityCode = atoi( m_poSRS->GetAuthorityCode(nullptr) );
+            }
+
+            // Translate SRS to WKT.
+            char *pszWKT1 = nullptr;
+            char *pszWKT2 = nullptr;
+            const char* const apszOptionsWkt1[] = { "FORMAT=WKT1_GDAL", nullptr };
+            const char* const apszOptionsWkt2[] = { "FORMAT=WKT2_2015", nullptr };
```
rouault: any reason not to export to FORMAT=WKT2_2018 systemetically ? This is a new format. No backward compatibility issue

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
```diff
@@ -0,0 +1,829 @@
...
+
+    if (m_featureOffsets != nullptr)
+        VSIFree(m_featureOffsets);
+
+    m_processedSpatialIndex = false;
+}
+
+OGRFeature *OGRFlatGeobufLayer::GetFeature(GIntBig /* nFeatureId */)
+{
+    CPLError(CE_Fatal, CPLE_AppDefined, "GetFeature: Not implemented");
```
rouault: to implement or remove

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
```diff
@@ -0,0 +1,829 @@
...
+    if (m_poFilterGeom != nullptr && !m_processedSpatialIndex) {
+        CPLDebug("FlatGeobuf", "Attempting spatial index search");
+        if (m_poHeader->index_node_size() == 0) {
+            CPLError(CE_Warning, CPLE_AppDefined, "No spatial index available");
+            return;
+        }
+        m_processedSpatialIndex = true;
+        if (m_poFp == nullptr) {
+            CPLDebug("FlatGeobuf", "processSpatialIndex (will attempt to open file %s)", m_pszFilename);
+            m_poFp = VSIFOpenL(m_pszFilename, "rb");
```
rouault: missing check on the retuurn value

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
```diff
@@ -0,0 +1,829 @@
...
+        m_featureOffsets = static_cast<uint64_t *>(VSI_MALLOC_VERBOSE(featuresCount * 8));
+        VSIFSeekL(m_poFp, sizeof(magicbytes) + sizeof(uoffset_t) + headerSize + treeSize, SEEK_SET);
+        VSIFReadL(m_featureOffsets, 8, featuresCount, m_poFp);
+    } else {
+        CPLDebug("FlatGeobuf", "processSpatialIndex noop");
+    }
+}
+
+GIntBig OGRFlatGeobufLayer::GetFeatureCount(int /* bForce */) {
+    processSpatialIndex();
```
rouault: the attribute filter should also be taken into account (by possibly using the base/slow implementation)

rouault: you can check the 'compliance' of the driver with the test_ogrsf utility (cd apps; make test_ogrsf)

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
```diff
@@ -0,0 +1,829 @@
...
+    return m_featuresCount;
+}
+
+OGRFeature *OGRFlatGeobufLayer::GetNextFeature()
+{
+    if (m_featuresPos >= m_featuresCount) {
```
rouault: attribute filter not taken into account

rouault:
It wouldn't hurt to test the robustness of the driver against corrupted/hostile files,by running oss-fuzz run locally. Some instructions in fuzzers/README.TXT. You'll need to modify https://github.com/Oss-fuzz/oss-fuzz/blob/master/projects/gdal/Dockerfile in your clone to point to the git URL of your GDAL fork.

bjornharrtell:
Thanks for the review @rouault, it's fair enough it's at high level for now.

gdal/ogr/ogrsf_frmts/flatgeobuf/flatbuffers/base.h
bjornharrtell: No it's the official C++ include files from the flatbuffers project, LICENSE file added (Apache license, v2) to the containing dir.

gdal/ogr/ogrsf_frmts/flatgeobuf/flatbuffers/code_generators.h
bjornharrtell: Also part of include files from the flatbuffers project.

gdal/ogr/ogrsf_frmts/flatgeobuf/packedrtree.h
bjornharrtell: Right, LICENSE file added to containing dir. (ICS license)

gdal/ogr/ogrsf_frmts/flatgeobuf/packedrtree.cpp
bjornharrtell: Right, LICENSE file added to containing dir. (ICS license)

bjornharrtell: Parts are based on https://github.com/rawrunprotected/hilbert_curves which is in public domain but I will make sure to include a notice in the relevant sections of the code anyway for this.

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
bjornharrtell: No reason other than I thought it still was a work in progress version and that it's naming/versioning/compatiblity might be uncertain. I wonder if it makes more sense to name the properties in my format to simply `wkt1`, `wkt2_2015` and `wkt2_2018` then instead of basing them on GeoPackage `definition_*` names.

rouault: WKT2_2018 (will probably be named 2019 but whatever) is in the publishing/editorial stages at ISO and similarly at OGC, so its content is finalized now.

bjornharrtell: Right ok, compatibility story not entirely clear - incompatible with wkt 2015 so should have dedicated property in any supporting format?

rouault: It depends which readers you want to address. If this is only GDAL, WKT2_2018 would be enough. Having support for 3 different WKT variants seems overkill

autotest/ogr/ogr_flatgeobuf.py
```diff
@@ -0,0 +1,188 @@
...
+# $Id$
+#
+# Project:  GDAL/OGR Test Suite
+# Purpose:  FlatGeobuf driver test suite.
+# Author:   Björn Harrtell <bjorn@wololo.org>
+#
+###############################################################################
+# Copyright (c) 2018-2019, Björn Harrtell <bjorn@wololo.org>
+#
+# This library is free software; you can redistribute it and/or
```
rouault: Please use the X/MIT header for that. Use of LGPL in some of the .py script is an historic accident

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
bjornharrtell: I want flatgeobuf (and as such any crs meta) to be easily consumed in as many places as possible. Of course, I suppose most use cases will simply involve the authority code. And I suppose the latest WKT can be translated into an older version by a capable parser i.e GDAL/proj for those who need it.. so perhaps it's fine to have WKT 2018 as the baseline for a (possibly) emerging format. But I'm fumbling in the dark to cater for a niche use case that I'm not very familiar with (very seldom work with data that has custom crs).

bjornharrtell: Thinking about this again I'm leaning that it would be best to leave out version completely and just store the definition as a single `wkt` string with unspecified version indicating that an interested consumer must introspect it themselves i.e using the instructions at B.8 of 12-063r5.

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
bjornharrtell: Switched to CE_Failure

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
bjornharrtell: Very useful! Was able to fix many issues after using test_ogrsf but I'm getting some unexplained errors related to GetFeature and filtering so I'll have to leave this open and revisit.

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobufdataset.cpp
```diff
@@ -84,15 +84,19 @@ void RegisterOGRFlatGeobuf()
 
     GDALDriver *poDriver = new GDALDriver();
     poDriver->SetDescription("FlatGeobuf");
+    poDriver->SetMetadataItem(GDAL_DCAP_RASTER, "NO");
```
rouault: the general habit is to only declare item that value to YES

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobufdataset.cpp
```diff
@@ -84,15 +84,19 @@ void RegisterOGRFlatGeobuf()
...
     poDriver->SetMetadataItem(GDAL_DCAP_VECTOR, "YES");
+    poDriver->SetMetadataItem(GDAL_DMD_SUBDATASETS, "NO");
```
rouault: the general habit is to only declare item that value to YES

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobufdataset.cpp
```diff
@@ -253,7 +257,7 @@ int OGRFlatGeobufDataset::TestCapability( const char * pszCap )
     else if (EQUAL(pszCap, OLCCreateGeomField))
         return m_create;
     else if (EQUAL(pszCap, OLCIgnoreFields))
-        return true;
+        return false;
```
rouault: this branch can be removed for clarity/compactness. Will go to the final else

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
```diff
@@ -617,7 +617,7 @@ OGRGeometry *OGRFlatGeobufLayer::readGeometry(const Feature *feature)
 {
     auto pXy = feature->xy();
     if (pXy == nullptr)
-        CPLError(CE_Fatal, CPLE_AppDefined, "readGeometry: Geometry has no coordinates");
+        CPLError(CE_Failure, CPLE_AppDefined, "readGeometry: Geometry has no coordinates");
```
rouault: you'll get a nullptr dereference on the next line

bjornharrtell: Ah yes sorry, should have realized I have to rework the error handling when changing this.

gdal/ogr/ogrsf_frmts/GNUmakefile
bjornharrtell: Ok will try to make the adjustments to move it into an optional format. Out of curiosity, why are geojson, kml and mitab mandatory?

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobufdataset.cpp
bjornharrtell: Unsure, will have to dive into it more. I have problems to fully understand the logic involving mapping files/directories to datasets.

gdal/ogr/ogrsf_frmts/GNUmakefile
rouault: geojson and kml because of OGR_G_ExportToGeoJSON/OGR_G_ExportToKML
MITAB because its used by generic attribute indexing code

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobufdataset.cpp
bjornharrtell: Added buffer verification step.

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobufdataset.cpp
bjornharrtell: Removed for now..

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobufdataset.cpp
bjornharrtell: Removed for now..

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobufdataset.cpp
bjornharrtell: Not sure but would like it to :) Need to figure out a test for that I guess.

bjornharrtell: Removed for now.

bjornharrtell:
Rebased and squashed. Remains to put it through oss-fuzz and probably other as of yet unknown shortcomings but otherwise ready for further scrutiny. :)

bjornharrtell:
I'll be at FOSS4G and might have time to hack on this there and if anyone interested want to join in or discuss don't hesitate to make contact. 🙂

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobufdataset.cpp
```diff
@@ -0,0 +1,325 @@
...
+        return nullptr;
+    }
+
+    // Verify that the datasource is a directory.
+    VSIStatBufL sStatBuf;
+
+    // What filename would we use?
+    CPLString osFilename;
+
+    CPLDebug("FlatGeobuf", "m_pszName: %s", m_pszName);
```
rouault: those debug statements could probably be removed

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
```diff
@@ -0,0 +1,897 @@
...
+            m_featureBufSize = std::max(m_featureBufSize * 2, m_featureSize);
+            CPLDebug("FlatGeobuf", "GetNextFeature: m_featureBufSize: %d", m_featureBufSize);
+            m_featureBuf = static_cast<GByte *>(VSIRealloc(m_featureBuf, m_featureBufSize));
+        }
+        VSIFReadL(m_featureBuf, 1, m_featureSize, m_poFp);
+
+    #ifdef DEBUG
+        const uint8_t * vBuf = const_cast<const uint8_t *>(reinterpret_cast<uint8_t *>(m_featureBuf));
+        Verifier v(vBuf, m_featureSize);
+        auto ok = VerifyFeatureBuffer(v);
```
rouault: what would happen in non-DEBUG mode when VerifyFeatureBuffer() isn't run and if the buffer is corrupted ?

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
```diff
@@ -0,0 +1,897 @@
...
+            CPLDebug("FlatGeobuf", "featureSize: %d", m_featureSize);
+        }
+    #endif
+
+        auto feature = GetRoot<Feature>(m_featureBuf);
+        auto fid = feature->fid();
+        poFeature->SetFID(fid);
+        //CPLDebug("FlatGeobuf", "GetNextFeature: fid: %zu", fid);
+        auto ogrGeometry = readGeometry(feature);
+        if (ogrGeometry == nullptr) {
```
rouault: cannot flatgeobuf store features without geometries ?

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
```diff
@@ -0,0 +1,897 @@
...
+            poFeature->SetGeometry(ogrGeometry);
+
+        auto properties = feature->properties();
+
+        if (properties != nullptr) {
+            auto data = properties->data();
+            auto size = properties->size();
+            //CPLDebug("FlatGeobuf", "properties->size: %d", size);
+            uoffset_t offset = 0;
+            while (offset < (size-1)) {
```
rouault: shouldn't the check take into account the number of bytes that will be actually read ?

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
bjornharrtell: Undefined behaviour. It's fairly expensive to verify the buffer so I wouldn't want that to be mandatory in a performance oriented format.

rouault: Well, we try hard to secure GDAL against undefined (hostile) behaviour in its default configuration. Are you sure those tests are really costly ? Couldn't the checks be let by default and skipped if the user defines some configuration option at runtime and/or a #define at build time ?

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
bjornharrtell: The format could support it (and I can't prohibit it on a schema level without uglyness) but in my experience null geometries in various formats has caused nothing but trouble so my thought was to prohibit that in implementations and spec.

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
bjornharrtell: A runtime opt out is reasonable, I'll add that. Just have to be a bit careful to not promise 100% safety regardless when documenting the behavior.

bjornharrtell: Introduced runtime opt out (VERIFY_BUFFERS option).

autotest/ogr/ogr_flatgeobuf.py
```diff
@@ -0,0 +1,192 @@
...
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+# OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+# THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+# FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+# DEALINGS IN THE SOFTWARE.
+###############################################################################
+
+
```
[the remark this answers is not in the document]
bjornharrtell: I've removed the obvious ones but have trouble finding the actual CI checks failures (in general).

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobufdataset.cpp
```diff
@@ -0,0 +1,325 @@
...
+OGRFlatGeobufDataset::OGRFlatGeobufDataset()
+{
+
+}
+
+OGRFlatGeobufDataset::OGRFlatGeobufDataset(const char *osDirName)
+{
+    CPLDebug("FlatGeobuf", "Request to create dataset %s", osDirName);
+    m_create = true;
+    m_pszName = CPLStrdup(osDirName);
```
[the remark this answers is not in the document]
bjornharrtell: Fixed including some other places with the same issues

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
bjornharrtell: I'm not sure what you mean here. The properties buffer is a "dumb" binary buffer containg the individual property values prefixed by the column index. The binary length of a value depend on column type for that index. So, the number of bytes to read/parse for each property is not known up front.

rouault: what I mean is that at line 519 you'll dereference 4 more bytes, so the check `offset < (size-1)` by itself is not enough to guarantee you won't run outside of the buffer. There must be some external guarantee that it won't happen ? I'm considering here the case of corrupted input data

autotest/ogr/ogr_flatgeobuf.py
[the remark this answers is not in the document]
rouault: Generally the error will be at the bottom of the logs. Except for build issues where due to parallel builds, the error might be sometime hundred of lines above. Look for the "error:" (error followed by colon) substring to directly go to the issues

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
bjornharrtell: I see, I've added checks at each progression of the buffer parse and a check for column index range.

jratike80:
You mean that not even empty geometries like "GEOMETRYCOLLECTION EMPTY" are not supported? I know that they are nasty to handle but they still tend to appear in real world data. But if it's documented then users can select for example GeoPackage for delivering such data.

bjornharrtell:
@jratike80 I don't want to support geometry collections either :) However I have reconsidered null geoms and it is possible in the now stable spec. As for "empty" it should be equivalent to null as geometry type is given. I agree, limitations should be clearly documented.

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
bjornharrtell: I've reconsidered and left it as allowed in the spec.

jratike80:
We considered a few years ago how to express empty geometries in OpenJUMP and we ended up to support all possible variations: POINT EMPTY, MULTIPOINT EMPTY etc. I think that the idea was something like to support "this feature does not have geometry yet, but once it is digitized it will be POINT" GEOMETRYCOLLECTION EMPTY was selected to mean just any empty geometry. I am pretty sure that this is off-topic in context of your innovative flat buffer format and driver.

autotest/ogr/ogr_flatgeobuf.py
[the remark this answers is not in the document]
bjornharrtell: I think I've been able to fix them and make exception for flatbuffers.h but build jobs doesn't look like they feel well...

gdal/ogr/ogrsf_frmts/flatgeobuf/ogr_flatgeobuf.h
```diff
@@ -0,0 +1,173 @@
...
+
+        VSILFILE *m_poFp = nullptr;
+
+        const Header *m_poHeader = nullptr;
+        OGRwkbGeometryType m_eGType;
+        GeometryType m_geometryType;
+        bool m_hasM = false;
+        bool m_hasZ = false;
+        bool m_hasT = false;
+        bool m_hasTM = false;
```
rouault: what means m_hasTM ?

bjornharrtell: TM stands for "time measurement" introduced by the discussion at https://github.com/bjornharrtell/flatgeobuf/issues/6#issuecomment-515709904. Note that T is for geodetic decimal year time, also discussed on that same issue. May be a futile attempt to support something not yet standardized. :)

rouault: OK, when compared to https://lists.osgeo.org/pipermail/proj/2019-August/008806.html, I guess  T would be the "Epoch of Expression" , and TM the "Epoch of Observation". In any case adding somewhere a comment to explain the semantics wouldn't hurt.

bjornharrtell: Seems right yes. It's somewhat documented uptream at https://github.com/bjornharrtell/flatgeobuf/blob/master/src/fbs/feature.fbs but deserve something more public if this becomes useful.

autotest/ogr/ogr_flatgeobuf.py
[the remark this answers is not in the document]
bjornharrtell: Still working through the CI checks (which I didn't properly find before). Sorry for overloading the build servers.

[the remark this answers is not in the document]
bjornharrtell: While some build jobs now pass I've been unable to fix the remaining failing ones that seem quite hairy and architecture/compiler specific (at least to me).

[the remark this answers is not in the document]
bjornharrtell: @rouault I have problems understanding the remaining build job failures. Can you provide some guidance?

rouault:
For `ogr/ogr_basic_test.py::test_ogr_basic_10 - AssertionError: assert (0 !...`, run `test_ogrsf -all_drivers` and see where/why it crashes

For ```./flatbuffers/base.h:354:1: error: unknown attribute 'no_sanitize' ignored [-Werror,-Wattributes]

__supress_ubsan__("alignment")

^

./flatbuffers/base.h:238:50: note: expanded from macro '__supress_ubsan__'

  #define __supress_ubsan__(type) __attribute__((no_sanitize(type)))
```, you likely need to add tests like
``
#if defined(__has_feature)
#  if __has_feature(alignment)
...
```
or test for a minimum gcc / clang version 

For https://ci.appveyor.com/project/OSGeo/gdal/builds/27188300/job/or3lqhsxi570t0nb, missing cast

For https://ci.appveyor.com/project/OSGeo/gdal/builds/27188300/job/43yd0vh885ws1f9u, `OBJ =	ogrflatgeobufdataset.obj ogrflatgeobufdataset.obj packedrtree.obj` from makefile.vc is buggy. It references ogrflatgeobufdataset.obj twice, instead of referencing ogrflatgeobuflayer.obj

bjornharrtell:
Thanks the the help @rouault. I have been able to fix all of them and a few more cases but still have one failing build job at https://travis-ci.com/OSGeo/gdal/jobs/232126400 (BUILD_NAME=trusty_clang DETAILS="optimized build, no libtool") which again I have trouble interpreting what is actually wrong.

rouault:
The failure on gcore/vrt_read.py::test_vrt_shared_no_proxy_pool in trusty_clang is indeed a bit weird, and likely unrelated to your changes, so probably some 'random' issue that triggers due to the test suite being a bit different now.
Could you possibly try to add a ds = None assignment at the end of the test_vrt_no_explicit_dataAxisToSRSAxisMapping and test_vrt_explicit_dataAxisToSRSAxisMapping_1_2 steps that are just above test_vrt_shared_no_proxy_pool ?

bjornharrtell:
Hmm I'm sure I've seen it multiple times but now it seems it fixed itself - https://travis-ci.com/OSGeo/gdal/jobs/232677605. I'll see what I can do about the missing datetime support before removing the WIP, and possibly also attempt some oss-fuzz if I can.

bjornharrtell:
Ah, I'm able to run it locally now by working from a clone of https://github.com/google/oss-fuzz. You had me fooled me by pointing at https://github.com/Oss-fuzz/oss-fuzz/blob/master/projects/gdal/Dockerfile in https://github.com/OSGeo/gdal/pull/1742#issuecomment-518009773. :)

rouault:
> You had me fooled me by pointing

ah sorry about that. Not obvious at first sight with of google/oss-fuzz vs oss-fuzz/oss-fuzz is the official repo. But indeed oss-fuzz/oss-fuzz is indicated as a fork...

rouault:
test_ogrsf -all_drivers leaks
```
Direct leak of 40 byte(s) in 2 object(s) allocated from:
    #0 0x7f47e7e242f3 in __interceptor_malloc /tmp/final/llvm.src/projects/compiler-rt/lib/asan/asan_malloc_linux.cc:88:3
    #1 0x7f47df45e66c in VSIMalloc /home/travis/build/OSGeo/gdal/gdal/port/cpl_vsisimple.cpp:561:12
    #2 0x7f47def57add in CPLMalloc /home/travis/build/OSGeo/gdal/gdal/port/cpl_conv.cpp:185:21
    #3 0x7f47def58119 in CPLStrdup /home/travis/build/OSGeo/gdal/gdal/port/cpl_conv.cpp:300:43
    #4 0x7f47e2b66ce9 in OGRGetXMLDateTime(OGRField const*) /home/travis/build/OSGeo/gdal/gdal/ogr/ogrutils.cpp:1374:18
    #5 0x7f47e10035cb in OGRFlatGeobufLayer::ICreateFeature(OGRFeature*) /home/travis/build/OSGeo/gdal/gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp:830:35
    #6 0x7f47e1816516 in OGRLayer::CreateFeature(OGRFeature*) /home/travis/build/OSGeo/gdal/gdal/ogr/ogrsf_frmts/generic/ogrlayer.cpp:630:12
    #7 0x44f930 in TestCreateLayer(GDALDriver*, OGRwkbGeometryType) /home/travis/build/OSGeo/gdal/gdal/apps/test_ogrsf.cpp:745:16
    #8 0x411194 in TestCreate(GDALDriver*, int) /home/travis/build/OSGeo/gdal/gdal/apps/test_ogrsf.cpp:999:13
    #9 0x40df61 in ThreadFunctionInternal(ThreadContext*) /home/travis/build/OSGeo/gdal/gdal/apps/test_ogrsf.cpp:311:25
    #10 0x40dac9 in ThreadFunction(void*) /home/travis/build/OSGeo/gdal/gdal/apps/test_ogrsf.cpp:269:9
    #11 0x40d087 in main /home/travis/build/OSGeo/gdal/gdal/apps/test_ogrsf.cpp:220:9
    #12 0x7f47c9b1982f in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2082f)
```

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
```diff
@@ -0,0 +1,1034 @@
...
+    CPLDebug("FlatGeobuf", "m_hasM: %d", m_hasM);
+    CPLDebug("FlatGeobuf", "m_hasT: %d", m_hasT);
+
+    auto crs = m_poHeader->crs();
+    if (crs != nullptr) {
+        m_poSRS = new OGRSpatialReference();
+        auto org = crs->org();
+        auto code = crs->code();
+        auto wkt = crs->wkt();
+        m_poSRS->SetAuthority(nullptr, org->c_str(), code);
```
rouault: this is likely useless. A SRS with just an authority node is invalid

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
```diff
@@ -0,0 +1,1034 @@
...
+        if (org == nullptr && code != 0) {
```
rouault: shouldn't org be "EPSG" if calling importFromEPSG() ?

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
```diff
@@ -0,0 +1,1034 @@
...
+            m_poSRS->importFromEPSG(code);
+        } else if (wkt) {
+            m_poSRS->importFromWkt(wkt->c_str());
+        }
+    }
+
+    m_eGType = getOGRwkbGeometryType();
+
+    const char *pszName = nullptr;
```
rouault: I don't think things are going to go well with pszName == nullptr

rouault:
The attached oss-fuzz generated file causes leaks 
[leak-e90271eb9e68e567d93c6736aff428c92b5a74c2.zip](https://github.com/OSGeo/gdal/files/3612135/leak-e90271eb9e68e567d93c6736aff428c92b5a74c2.zip)
```
[...]
    #14 0x79eee5 in OGRSpatialReference::importFromWkt(char const*) /src/gdal/gdal/ogr/ogrspatialreference.cpp:1728:12
    #15 0x14a9dc8 in OGRFlatGeobufLayer::OGRFlatGeobufLayer(FlatGeobuf::Header const*, unsigned char*, char const*, unsigned long) /src/gdal/gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp:78:22
    #16 0x14a5ef0 in OGRFlatGeobufDataset::Open(GDALOpenInfo*) /src/gdal/gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobufdataset.cpp:185:39
    #17 0x10fcef0 in GDALOpenEx /src/gdal/gdal/gcore/gdaldataset.cpp:3377:20
    #18 0x6a3181 in OGROpen /src/gdal/gdal/ogr/ogrsf_frmts/generic/ogrsfdriverregistrar.cpp:113:24
    #19 0x57588e in LLVMFuzzerTestOneInput /src/gdal/gdal/./fuzzers/ogr_fuzzer.cpp:106:26
    #20 0x47d311 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) /src/llvm/projects/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:553:15
    #21 0x47ca35 in fuzzer::Fuzzer::RunOne(unsigned char const*, unsigned long, bool, fuzzer::InputInfo*, bool*) /src/llvm/projects/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:469:3
    #22 0x47ed37 in fuzzer::Fuzzer::MutateAndTestOne() /src/llvm/projects/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:695:19
    #23 0x47faf5 in fuzzer::Fuzzer::Loop(std::Fuzzer::vector<fuzzer::SizedFile, fuzzer::fuzzer_allocator<fuzzer::SizedFile> >&) /src/llvm/projects/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:831:5
    #24 0x46d546 in fuzzer::FuzzerDriver(int*, char***, int (*)(unsigned char const*, unsigned long)) /src/llvm/projects/compiler-rt/lib/fuzzer/FuzzerDriver.cpp:825:6
    #25 0x497112 in main /src/llvm/projects/compiler-rt/lib/fuzzer/FuzzerMain.cpp:19:10
    #26 0x7f2df70d682f in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2082f)
```

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
bjornharrtell: A nullptr org is to be implictly interpreted as EPSG as per spec at https://github.com/bjornharrtell/flatgeobuf/blob/3a969d9dd90404ab67cfecb45f0f16d98fce2f84/src/fbs/header.fbs#L35.

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
bjornharrtell: Hmm ok, the idea was to handle the case where org is something other than EPSG and no WKT is supplied and as such cannot be imported by the logic that follows.

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
bjornharrtell: Why not? Seems OGRFeatureDefn constructor will immediately feed it to CPLStrdup which will interpret it as an empty string?

rouault: ok, but an empty layer name is going to cause usability issues downstream. Probably better set a hardcoded "unnamed" or something like that.

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
rouault: ok

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
bjornharrtell: Agreed, will make it so.

bjornharrtell:
oss-fuzz sure is fun stuff :)

bjornharrtell:
Getting a timeout in fuzzing when calculating tree size, I thought it would go away with bounds checking introduced in https://github.com/OSGeo/gdal/pull/1742/commits/2b8dedf3dfb75e5b9385633695d440c844cb6f3f but it appears not to. Error is like follows:

```
==11== ERROR: libFuzzer: timeout after 87 seconds
    #0 0x54cc61 in __sanitizer_print_stack_trace /src/llvm/projects/compiler-rt/lib/asan/asan_stack.cpp:86:3
    #1 0x496c28 in fuzzer::PrintStackTrace() /src/llvm/projects/compiler-rt/lib/fuzzer/FuzzerUtil.cpp:205:5
    #2 0x47bd19 in fuzzer::Fuzzer::AlarmCallback() /src/llvm/projects/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:300:5
    #3 0x7f285f51a38f  (/lib/x86_64-linux-gnu/libpthread.so.0+0x1138f)
    #4 0x6d8e97 in FlatGeobuf::PackedRTree::size(unsigned long, unsigned short) /src/gdal/gdal/ogr/ogrsf_frmts/flatgeobuf/packedrtree.cpp:318:35
```

Is there any way to find out what arguments is seemingly causing the algorithm at https://github.com/bjornharrtell/gdal/blob/2b8dedf3dfb75e5b9385633695d440c844cb6f3f/gdal/ogr/ogrsf_frmts/flatgeobuf/packedrtree.cpp#L310-L324 to go into an infinite loop or is it just that it tries it with so many permutations that it takes forever?

rouault:
> Is there any way to find out what arguments is seemingly causing the algorithm at https://github.com/bjornharrtell/gdal/blob/2b8dedf3dfb75e5b9385633695d440c844cb6f3f/gdal/ogr/ogrsf_frmts/flatgeobuf/packedrtree.cpp#L310-L324 to go into an infinite loop or is it just that it tries it with so many permutations that it takes forever?

It should normally generate a file that reproduces the issue in ./build/out/gdal/.
But looking at that code, I can see that if it is called with numItems = 0, then n will never be 1, hence the infinite loop

bjornharrtell:
Of course! 🤦

bjornharrtell:
Not getting any errors from oss-fuzz anymore but the run never stops, left it running for more than 3 hours when the latest line it outputted was:

#50736769	REDUCE cov: 2154 ft: 2775 corp: 103/54Kb lim: 4096 exec/s: 4107 rss: 95Mb L: 537/3120 MS: 2 ChangeBit-EraseBytes-

Mabye it needs to be put though the real oss-fuzzer now.

rouault:
@bjornharrtell OK, if you're happy with this, I can merge it.

bjornharrtell:
@rouault I'm happy and hope it is in good enough shape for real world use. I will take care of it in the future to the best of my ability.

jratike80:
I guess that the document page that is referred to in the code is to come because I could not find it yet: 
 poDriver->SetMetadataItem(GDAL_DMD_HELPTOPIC, "drv_flatgeobuf.html");

rouault:
> I guess that the document page that is referred to in the code...

Good observation. Up to now we provided URIs to the old static HTML pages. Probably here we should provide as URI "drivers/vector/flatgeobuf.html"

PR #1751, opened 2019-08-06
osr_ct.py and osr_pm.py: use pytest.approx()

rouault:
@craigds Do you think your magic auto-refactoring could automate such transformations ?


craigds:
Definitely. Are there lots of these through the test suite? I could whip something up, if so.

craigds:
See that^  PR 😄

rouault:
Closing as superseded by #1753

PR #1753, opened 2019-08-07
Automated: Convert various comparisons into pytest.approx() calls

craigds:
Adds `pytest.approx()` calls in the test suite.

## What does this PR do?

Turns these awkard comparisons

```python
        assert abs(warped - exp) <= 0.6
```
... into [pytest.approx](https://docs.pytest.org/en/latest/reference.html#pytest-approx) calls.

As requested here https://github.com/OSGeo/gdal/pull/1751#issue-304904183

Bowler script code here: https://github.com/craigds/gdal-pytest/blob/master/9-pytest-approx.py#L245-L358

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments

## Environment

 n/a


autotest/gcore/misc.py
```diff
@@ -158,7 +158,7 @@ def misc_5_internal(drv, datatype, nBands):
         if drv.ShortName not in ['PNM', 'MFF', 'NULL']:
             got_gt = ds.GetGeoTransform()
             for i in range(6):
-                if abs(got_gt[i] - set_gt[i]) > 1e-10:
+                if got_gt[i] != pytest.approx(set_gt[i], abs=1e-10):
                     print('Did not get expected GT for drv = %s, nBands = %d, datatype = %s' % (drv.ShortName, nBands, gdal.GetDataTypeName(datatype)))
                     print(got_gt)
                     return -1
```
rcoup: OT: thought these were all gone now?

craigds: Most are. Guess I missed a few due to them not fitting the usual pattern.

craigds:
> rewrite (some of) the for loops

I don't think `pytest.approx` takes list args, not sure what you mean there.

I've fixed the test failure; turns out `a != pytest.approx(b, abs=tolerance)` doesn't work as expected if the args are `NaN`.

I changed it to an assert statement so the condition could be inverted and now it should work.

rcoup:
> I don't think pytest.approx takes list args, not sure what you mean there.

No biggie, but:

> The same syntax also works for sequences of numbers:
> ```
> >>> (0.1 + 0.2, 0.2 + 0.4) == approx((0.3, 0.6))
> True
> ```
> Dictionary values:
> ```
> >>> {'a': 0.1 + 0.2, 'b': 0.2 + 0.4} == approx({'a': 0.3, 'b': 0.6})
> True
> ```
> numpy arrays:
> ```
> >>> import numpy as np                                                          
> >>> np.array([0.1, 0.2]) + np.array([0.2, 0.4]) == approx(np.array([0.3, 0.6])) 
> True
> ```
> And for a numpy array against a scalar:
> ```
> >>> import numpy as np                                         
> >>> np.array([0.1, 0.2]) + np.array([0.2, 0.1]) == approx(0.3) 
> True
> ```

https://docs.pytest.org/en/latest/reference.html#pytest-approx

PR #1756, opened 2019-08-07
Yell when people use unsupported build systems

rouault:


mloskot:
Because? (i.e. reference to any issues, for those curious :-))

rouault:
> Because? (i.e. reference to any issues, for those curious :-))

I don't want to pinpoint at anyone, but following latest discussions on the mailing list should give a hint :-)

mloskot:
That is a good reference :-)

schwehr:
I love using bazel...

rouault:
Dropping that...

PR #1757, opened 2019-08-08
kmlsuperoverlay: Add `RESAMPLE` creation option

craigds:
Adds optional resampling to the kmlsuperoverlay driver.

## What does this PR do?

I dusted off [this old patch](https://trac.osgeo.org/gdal/ticket/4747#no1) which we used on an ancient GDAL (~1.8?) to make our kml superoverlays look pretty.

Since each image in the kmlsuperoverlays is created from the next image in, everything's getting resized even if you're doing a 1:1 translate from another image. The result with nearest-neighbour resampling is that images look choppy and pixelated.

This adds a creation option to kmlsuperoverlay format that allows resampling via a warp kernel.

Note: I got the old patch and polished it up, but I don't guarantee it's all completely sensible, not being particular familiar with this code. (I also haven't written tests yet). It does seem to have the desired effect on a few images I've tried though :)

Ancient ticket https://trac.osgeo.org/gdal/ticket/4747

## Tasklist

 - [ ] Add test case(s)
 - [ ] Review
 - [ ] Adjust for comments

## Environment
n/a

### DIFF
--- a/gdal/frmts/kmlsuperoverlay/kmlsuperoverlaydataset.cpp
+++ b/gdal/frmts/kmlsuperoverlay/kmlsuperoverlaydataset.cpp
@@ -42,6 +42,8 @@
 #include "cpl_string.h"
 #include "cpl_vsi.h"
 #include "gdal_frmts.h"
+#include "gdalwarper.h"
+#include "gdal_alg.h"
 #include "ogr_spatialref.h"
 #include "../vrt/gdal_vrt.h"
 #include "../vrt/vrtdataset.h"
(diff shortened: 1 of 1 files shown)


rouault:
as there is no reprojection involved, a simpler (and faster) solution would be to use the GDALRasterIOExtraArg* psExtraArg  argument of RasterIO() to ask for non-nearest resampling.

craigds:
thanks, I guess that's new in the last 7 years (!) I'll update this accordingly next week

rouault:
>  I guess that's new in the last 7 years

yes, was added in GDAL 2.0

craigds:
While researching that I found this helpful nugget:

> The GDAL_RASTERIO_RESAMPLING configuration option can also be defined to override the default resampling to one of BILINEAR, CUBIC, CUBICSPLINE, LANCZOS, AVERAGE or MODE.

I think this means this patch is unnecessary; we can already override the resampling algorithm without the driver needing to support it at all.

craigds:
Confirmed, no longer required. Thanks :)

PR #1769, opened 2019-08-12
fix: wasm/emscripten "long long" type check 

remoe:
Using of AutoConf internal checks. Own AutoConf macro not needed any more?
This is needed to compile with emscripten.


rouault:
Please also run "autoreconf" (with autoconf 2.69) and commit the modified autoconf file, since that one is (weirdly) in git

remoe:
you mean the "configure" - file ?

remoe:
Calling of "autoreconf" changes the following files:
gdal/aclocal.m4
gdal/configure
gdal/port/cpl_config.h.in

Ok, to commit?

rouault:
Ah I was too fast.. I meant the 'configure' file, and I only meant running 'autoconf' (not autoreconf). Could you revert your last commit, run 'autoconf' and commit 'configure' ?

remoe:
ok, done.

PR #1776, opened 2019-08-13
VFK: use a faster implementation of VFKDataBlockSQLite::LoadGeometryPolygon()

rouault:
This should end fixing the issue raised in
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=16309 where a corrupted
.vfk files with a lot of line strings takes a long time to process.

The speed-up comes from using a hash map to map from a ring-in-building end
point to the ring, instead of iterating. This speeds up the reconstruction in
that case from ~20s to ~2s

@landam Could you review this and tests this work as expected with real datasets ?

### DIFF
--- a/gdal/ogr/ogrsf_frmts/vfk/vfkdatablock.cpp
+++ b/gdal/ogr/ogrsf_frmts/vfk/vfkdatablock.cpp
@@ -502,6 +502,19 @@ int IVFKDataBlock::LoadGeometry()
     return nInvalid;
 }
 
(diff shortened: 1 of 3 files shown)


rouault:
CC @landam

rouault:
@landam Any chance you cold test this pull request :-) ?

rouault:
ping

landam:
@rouault Sorry for my silence, and thanks for merging PR!

rouault:
No problem Martin. Cheers

Issue #1772, opened 2019-08-13
netCDF SG Writer: WBuffer::subCount() might underflow


rouault:
@wchen329 

WBuffer::subCount(unsigned long long memfree)  { this->used_mem -= memfree; } is called with used_mem = 0 and memfree = 8 by test_write_buffer_restrict_correctness test. This can be detected by adding a CPLAssert()
```
diff --git a/gdal/frmts/netcdf/netcdfsgwriterutil.h b/gdal/frmts/netcdf/netcdfsgwriterutil.h
index 643ef98..f43b662 100644
--- a/gdal/frmts/netcdf/netcdfsgwriterutil.h
+++ b/gdal/frmts/netcdf/netcdfsgwriterutil.h
@@ -88,7 +88,7 @@ namespace nccfdriver
             /* subCount(...)
              * Directly subtracts the specified size from used_mem
              */
-             void subCount(unsigned long long memfree) { this->used_mem -= memfree; }
+             void subCount(unsigned long long memfree) { CPLAssert(this->used_mem >= memfree); this->used_mem -= memfree; }
              unsigned long long& getUsage() { return used_mem; }
 
              void reset() { this->used_mem = 0; }
```

While technically C defined behaviour (operations on unsigned types are well defined and wrap around), it feels like this is unexpected (at least to me) that we substract more used memory than used (thus getUsage() will return garbage)

Note: I discovered this while trying to run GDAL with-fsanitize=unsigned-integer-overflow (many other changes needed to make it pass), which triggers an abort whenever a overflow/underflow occurs on unsigned operations (a number of them are legit, others might be bug).

```
Thread 1 "python" received signal SIGABRT, Aborted.
0x00007ffff6d1b428 in __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:54
54	../sysdeps/unix/sysv/linux/raise.c: Aucun fichier ou dossier de ce type.
(gdb) bt
#0  0x00007ffff6d1b428 in __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:54
#1  0x00007ffff6d1d02a in __GI_abort () at abort.c:89
#2  0x00007ffff1f454d3 in CPLErrorV (eErrClass=CE_Fatal, err_no=7, fmt=0x7ffff2d0f5ce "Assertion `%s' failed in file `%s', line %d", args=0x7fffffff71d0) at cpl_error.cpp:466
[...]

PR #1777, opened 2019-08-14
Address buffer memory counting underflow in CF-1.8- issue #1772

wchen329:
## What does this PR do?
The CF-1.8 portion of the netCDF Driver keeps an approximate count of memory usage during normal operation. If enough writes are logged, relative to the maximum buffer size, then underflow could possibly occur. This was in part, if not completely, because transactions received from the log, were not being re-recorded into the memory log but their usage were still being subtracted out. This PR fixes this issue by adding logged transactions back into the buffer memory counter.

This PR also adds an assert to the counter subtraction to have this check that it does not underflow be performed in tests.

Thanks to @rouault for reporting this issue.

Issue #1772 

## Tasklist

 - [x] Review
 - [ ] Adjust for comments

## Environment

* OS: openSUSE Tumbleweed
* Compiler: GCC 9.1.1

### DIFF
--- a/gdal/frmts/netcdf/netcdfsgwriterutil.cpp
+++ b/gdal/frmts/netcdf/netcdfsgwriterutil.cpp
@@ -208,6 +208,21 @@ namespace nccfdriver
         this->geometry_ref = ft.GetGeometryRef();
     }
 
+    inline void WBuffer::addCount(unsigned long long memuse)
+    {
+        this->used_mem += memuse;
+    }
+
+    inline void WBuffer::subCount(unsigned long long memfree)
+    {
+        unsigned long long oldmem = this->used_mem;
+        this->used_mem -= memfree;
+        CPLAssert(used_mem <= oldmem);
+        /* detect underflow cases- if we subtract, then the new value better not 
+         * be larger!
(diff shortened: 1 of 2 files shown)


gdal/frmts/netcdf/netcdfsgwriterutil.cpp
```diff
@@ -208,6 +208,21 @@ namespace nccfdriver
...
     }
 
+    inline void WBuffer::addCount(unsigned long long memuse)
+    {
+        this->used_mem += memuse;
+    }
+
+    inline void WBuffer::subCount(unsigned long long memfree)
+    {
+        unsigned long long oldmem = this->used_mem;
```
rouault: compilers in non debug mode complain about oldmem variable not being used, since CPLAssert() is a no-op then. Could you rewrite this just as at it, that is
```
CPLAssert(this->used_mem >= memfree);
this->used_mem -= memfree;
```

wchen329: Yeah, your way just seems "better" in general, will fix it :+1:

PR #1782, opened 2019-08-15
Fix or tag as allowed all uses of unsigned integer overflow, add add -fsanitize=unsigned-integer-overflow to the Travis sanitize configuration

rouault:


rouault:
That PR might be a bit controversial as it effectively "bans" a well-defined behaviour of C/C++, that is underflow/overflow on *unsigned* integer arithmetic. However there are potential for real bugs to be hidden by such a behaviour (I or other contributors fixed a few instances of that in a few previous commits), so it seems reasonable to try to avoid it when possible, and when possible to tag the functions where this is explicitly allowed.

gdal/frmts/gtiff/libtiff/tif_dirread.c
```diff
@@ -5958,6 +5958,13 @@ static void TryChopUpUncompressedBigTiff( TIFF* tif )
     allocChoppedUpStripArrays(tif, nstrips, stripbytes, rowsperstrip);
 }
 
+
+TIFF_NOSANITIZE_UNSIGNED_INT_OVERFLOW
```
abellgithub: Seems like something generic with templates would be good.

gdal/frmts/hdf4/hdf4multidim.cpp
```diff
@@ -1437,6 +1437,12 @@ union ReadFunc
     intn (*pReadData)(int32, int32 [], int32 [], int32 [], VOIDP);
 };
 
+CPL_NOSANITIZE_UNSIGNED_INT_OVERFLOW
+static inline void IncrPointer(GByte*& ptr, GPtrDiff_t nInc, size_t nIncSize)
```
abellgithub: Seems weird that allowing an overflow here is OK.  Maybe a comment explaining?  Or at the call site.

gdal/frmts/hdf5/hdf5multidim.cpp
```diff
@@ -1240,7 +1240,7 @@ bool HDF5Array::ReadSlow(const GUInt64* arrayStartIdx,
             }
             else
             {
-                anStart[i] = arrayStartIdx[i] + arrayStep[i] * (count[i]-1);
+                anStart[i] = arrayStartIdx[i] - (-arrayStep[i]) * (count[i]-1);
```
abellgithub: It's confusing what you're trying to achieve here without the context of the PR.  A comment would help.  Also think if parens went around the product it would be easier to read.

gdal/frmts/netcdf/netcdfmultidim.cpp
```diff
@@ -1962,12 +1965,14 @@ bool netCDFVariable::IReadWriteGeneric(const size_t* arrayStartIdx,
                 if( (--nIters) == 0 )
                     break;
                 stack_ptr[1] += ptr_inc[1];
-                array_idx[1] = static_cast<size_t>(array_idx[1] + arrayStep[1]);
+                array_idx[1] = static_cast<size_t>(CPLUnsanitizedAdd<GUInt64>(
```
abellgithub: Couldn't use do the add with size_t type and avoid the extra cast?

abellgithub: Also, if an overflow is OK here, a comment would be good.

gdal/frmts/rmf/rmflzw.cpp
```diff
@@ -85,8 +91,8 @@ static void LZWUpdateTab(LZWStringTab *poCodeTab, GUInt32 iPred, char bFoll)
 /* It will NOT notice if the table is full. This must be handled        */
 /* elsewhere                                                            */
 /* -------------------------------------------------------------------- */
-    GUInt32 nLocal = (iPred + bFoll) | 0x0800;
-    nLocal = (nLocal*nLocal >> 6) & 0x0FFF;      // middle 12 bits of result
+    GUInt32 nLocal = CPLUnsanitizedAdd<GUInt32>(iPred, bFoll) | 0x0800;
+    nLocal = (UnsanitizedMul(nLocal, nLocal) >> 6) & 0x0FFF;      // middle 12 bits of result
```
abellgithub: Don't know if it's better, but

```
(nLocal * nLocal) >> 6
```

is the same as

```
(nLocal >> 3) * (nLocal >> 3)
```

gdal/frmts/vrt/vrtmultidim.cpp
```diff
@@ -1328,7 +1339,7 @@ bool VRTMDArraySourceInlinedValues::Read(const GUInt64* arrayStartIdx,
         {
             // For negative step request, temporarily simulate a positive step
             // and fix up the start at the end of the loop.
-            start_i = start_i + (count[i]-1) * step_i;
+            start_i = start_i - (count[i]-1) * (-step_i);
```
abellgithub: Think comments are good here because if I saw this I might want to simplify by reworking the algebra.

gdal/gcore/gdalmultidim.cpp
```diff
@@ -3125,7 +3123,7 @@ static std::shared_ptr<GDALMDArray> CreateSlicedArray(
                 if( nDimSize < static_cast<GUInt64>(-endIdx) )
                     endIdx = 0;
                 else
```
abellgithub: I think this could be more cleanly resolved by doing something like:

```
GUInt64 nEndIdx;
GIntBig endIdx = CPLAtoGIntBig(pszEnd);
if (endIdx < 0)
{
    nEndIdx = -endIdx;
    ...
}
else
    nEndIdx = endIdx;
```
This will eliminate a bunch of casts as well.

gdal/frmts/gtiff/libtiff/tif_dirread.c
```diff
@@ -5958,6 +5958,13 @@ static void TryChopUpUncompressedBigTiff( TIFF* tif )
...
+static uint64 _TIFFUnsanitizedAddUInt64AndInt(uint64 a, int b)
```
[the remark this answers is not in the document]
rouault: This is a file from libtiff  which is a c file, so no templates

gdal/frmts/hdf4/hdf4multidim.cpp
rouault: The overflow is a bit technical here and not a real one. This comes from the fact that GPtrDiff_t is signed, but size_t is not. So from C/C++ standards, GPtrDiff_t is converted to a unsigned integer, and complement-to-2 arithmetic does things right

gdal/frmts/hdf5/hdf5multidim.cpp
rouault: Parenthesis added. Extra comment added
// Use double negation so that operations occur only on
// positive quantities to avoid an artificial negative signed
// integer to unsigned conversion.

gdal/frmts/netcdf/netcdfmultidim.cpp
rouault: Cast logic simplified, and following comment added
// CPLUnsanitizedAdd needed as arrayStep[] might be negative, and
// thus automatic conversion from negative to big unsigned might
// occur

gdal/frmts/vrt/vrtmultidim.cpp
rouault: Same as an above occurence. Comment added

gdal/frmts/rmf/rmflzw.cpp
rouault: Perhaps. This code is not mine, so I didn't want to mess up too much with that. Doing as you suggest is very slightly less efficient, and doesn't necessarily guarantee that an overflow won't occurr. I'll let this as it, as overflow in this sort of code is more or less intended I think

gdal/gcore/gdalmultidim.cpp
rouault: Your proposed code is not equivalent to the intended logic, but I've simplified this a bit as
```
if( endIdx < 0 )
            {
                const auto positiveEndIdx = static_cast<GUInt64>(-endIdx);
                if( nDimSize < positiveEndIdx )
                    endIdx = 0;
                else
                    endIdx = nDimSize - positiveEndIdx;
            }
```

PR #1794, opened 2019-08-20
Append values from duplicate fields when reading sosi file

larsop:
## Append values from duplicate 

In the case below we see BEITEBRUKERID with two values YD and YG. In the gdal we only get the YD value and not the YG.

`
784198569 84494384
.FLATE 13257:
..OBJTYPE Flyttelei
..KVALITET 55 `1500`
..OPPHAV Reindriftsforvaltningen
..VERIFISERINGSDATO 20150325
..BEITEBRUKERID YD
..BEITEBRUKERID YG
..FTEMA 4905
..REF :13244 :2779 :13249 :2777 :2822 :-13247 :-13250 :-13253 :-13256 :-13246
:13252 :2801 :13260 :4437 :-2808 :4866 :-13245
`
With this patch we will get a comma-separated list like this "YD,YG'"

NB! This patch code as it is not generic, because I am not sure about how send parameters from the command-line down to the sosi mudule.  In file ogr/ogrsf_frmts/sosi/ogrsosilayer.cpp and line 134 I have this code `appendFieldsMap.insert(std::pair<std::string,std::string>("BEITEBRUKERID",","));`. This must be a parameter you can have a list different fields.  This will only work for string values.

What's the best way to send this as parameter down to the sosi module ?

Another problem is that this is my fist C-code so the code is probably not very good.

## Environment

I have tested the attached file on Mac using the commands below one file data/test_duplicate_fields.sos .
`
ogr2ogr -t_srs EPSG:4258  -mapFieldType datetime=String test_line.shp test_duplicate_fields.sos lines

ogr2ogr -t_srs EPSG:4258  -mapFieldType datetime=String test_poly.shp test_duplicate_fields.sos polygons
`


gdal/ogr/ogrsf_frmts/sosi/ogrsosilayer.cpp
```diff
@@ -149,10 +153,26 @@ OGRFeature *OGRSOSILayer::GetNextFeature() {
             while (pszLine[0] == '.') pszLine++; /* skipping the dots at the beginning of a SOSI line */
             char *pszUTFLine = CPLRecode(pszLine, poParent->pszEncoding, CPL_ENC_UTF8); /* switch to UTF encoding here */
             char *pszPos = strstr(pszUTFLine, " ");
-            if (pszPos != nullptr) {
-                osKey = CPLString(std::string(pszUTFLine,pszPos));
-                osValue = CPLString(pszPos+1);
-                oHeaders.insert(std::pair<CPLString,CPLString>(osKey,osValue));
+            if (pszPos != nullptr) {                osKey = CPLString(std::string(pszUTFLine,pszPos));
```
rouault: wrong indentation for osKey = ...

rouault:
Hardcoding the fields which can received duplicated values, or asking the user to provide it, doesn't seem like an ideal solution. Why not doing it systemicall ? Just removing appendFieldsMap declaration and its use should be sufficient as far as I can see

larsop:
Yes I can remove the  appendFieldsMap , but I was afraid to change the default behaviour  of the code. Another reason to have a parameter was to have an option to change the delimiter.
But since default behaviour now, means missing data, maybe it's better to change the default behaviour as suggest.

larsop:
Sorry I may have misunderstood you here.

I will change the default behavior as suggest, but I in my case I still need an option to limit this behavior to a selected set of fields. 

So I can maybe do something like this?

export ogrSosiAppendFieldsMap='BEITEBRUKERID:, FIELD2test:;,FIELD3test: '

This is supposed to mean something like append BEITEBRUKERID with comma and FIELD2test with semi column and FIELD3test using space.

Then I will pick the value of this env.  variable in code and use if it is defined.

This is maybe what you meant by systemicall ?

Thanks for your work with gdal project.

rouault:
> in my case I still need an option to limit this behavior to a selected set of fields.

why ? Ignoring some field values and only keeping the last one seems weird to me. But I'm not an expert in this format.

Open options could be used. This is the -oo parameter in ogrinfo / ogr2ogr and they are passed in poOpenInfo->papszOpenOptions to the Open() method.

> This is maybe what you meant by systemicall ?

Ah typo I meant "systematically" = "always", "unconditionally"

gdal/ogr/ogrsf_frmts/sosi/ogrsosilayer.cpp
```diff
@@ -129,9 +132,30 @@ OGRFeature *OGRSOSILayer::GetNextFeature() {
...
     std::map<std::string,std::string> appendFieldsMap;
-    appendFieldsMap.insert(std::pair<std::string,std::string>("BEITEBRUKERID",","));
+
+	// Get ogrSosiAppendFieldsMap and update appendFieldsMap if present;
+    // The input must on this format
+    // -oo ogrSosiAppendFieldsMap="BEITEBRUKERID:,&FIELD2test: &FIELD3test:;"
+    // With the example abouve the field BEITEBRUKERID append character will be ':', for FIELD2test it will space and for FIELD3test it will be ;
+    const char * ogrSosiAppendFieldsMapInput = CSLFetchNameValue(poParent->GetOpenOptions(),"ogrSosiAppendFieldsMap");
+    if (ogrSosiAppendFieldsMapInput != NULL) {
+        char *ogrSosiAppendFieldsMapCopy = strdup(ogrSosiAppendFieldsMapInput);
```
rouault: will leak memory. and strtok() is not thread safe
Instead you can use
```
CPLStringList aosTokens(CSLTokenizeString2(ogrSosiAppendFieldsMapInput, "&", 0));
for( int i = 0; i < aosTokens.size(); i++ )
{
   // aosTokens[i]
}
```
Do you really need a different separator per field ? That seems to be over-complication to me. If not, then, if you want this to be configurable, a dedicated open option could be used;

gdal/ogr/ogrsf_frmts/sosi/ogrsosilayer.cpp
```diff
@@ -129,9 +132,30 @@ OGRFeature *OGRSOSILayer::GetNextFeature() {
...
+    // Used to limit the number of limit the fields to be appended.
+    // Default is that duplicates will appended if no ogrSosiAppendFieldsMap parameter is sent.
     std::map<std::string,std::string> appendFieldsMap;
-    appendFieldsMap.insert(std::pair<std::string,std::string>("BEITEBRUKERID",","));
+
+	// Get ogrSosiAppendFieldsMap and update appendFieldsMap if present;
+    // The input must on this format
+    // -oo ogrSosiAppendFieldsMap="BEITEBRUKERID:,&FIELD2test: &FIELD3test:;"
+    // With the example abouve the field BEITEBRUKERID append character will be ':', for FIELD2test it will space and for FIELD3test it will be ;
+    const char * ogrSosiAppendFieldsMapInput = CSLFetchNameValue(poParent->GetOpenOptions(),"ogrSosiAppendFieldsMap");
```
rouault: No need for ogr sosi, and regarding naming convention "APPEND_FIELDS_MAP" would match what is generally used

larsop:
The default will append duplicate values now.

And I used the "-oo ogrSosiAppendFieldsMap" to able override the new default. 

-oo ogrSosiAppendFieldsMap="BEITEBRUKERID:,&FIELD2test: &FIELD3test:;"
With the example above 
for BEITEBRUKERID the append character will be ':'
for FIELD2test the append character will be ' ' (space) 
for FIELD3test the append character will be ';'

Thanks for all the help.

gdal/ogr/ogrsf_frmts/sosi/ogrsosilayer.cpp
larsop: Thanks, Tried to fix memory leak, renamed oo variable and fixed no need for adding delimiter character if not needed.

-oo appendFieldsMap="BEITEBRUKERID:,&FIELD2test: &FIELD3test:;"
With the example above the field BEITEBRUKERID append character will be ':', for FIELD2test it will space and for FIELD3test it will be ;

-oo appendFieldsMap="BEITEBRUKERID&FIELD2test&FIELD3test"
With the example above the append character will be ',' for fields in the list

rouault:
Can you also:
- document the open option in the driver metadata. See for example https://github.com/OSGeo/gdal/blob/master/gdal/ogr/ogrsf_frmts/mongodb/ogrmongodbdriver.cpp#L2842
- document it in gdal/doc/source/drivers/vector/sosi.rst (I just added a mostly empty template in master, as there was no pre-existing documentation (more welcome!), so you'll have to merge master in your branch first)
- move gdal/data/test_duplicate_fields.sos to autotest/ogr/data/
- and it would be great also to have a test in autotest/ogr/ogr_sosi.py for the open option.

gdal/doc/source/drivers/vector/sosi.rst
```diff
@@ -7,3 +7,8 @@ Norwegian SOSI Standard
 .. shortname:: SOSI
 
 This driver requires the FYBA library.
+
+ poDriver->SetMetadataItem( GDAL_DMD_OPENOPTIONLIST,
```
rouault: No, this is code that should go to ogr/ogrsf_frmts/sosi/ogrsosidriver.cpp

In the rst file, you must add reStructured text. See for example doc/source/drivers/vector/csv.rst which has an "Open Options" section

gdal/doc/source/drivers/vector/sosi.rst
```diff
@@ -7,3 +7,8 @@ Norwegian SOSI Standard
...
+"<OpenOptionList>"
+"  <Option name='appendFieldsMap' type='string' description='Default is that all rows for equal field names will be appended in a feature, but with this parameter you select what field this should be valid for. With appendFieldsMap="f1&f2", Append will be done for field f1 and f2 using a comma as delimiter. This list can more complicacted check the source code.' default=''/>"
```
rouault: For the XML put in the cpp, the '&' character should be XML escaped as "&amp;"
Instead of pointing to the source code in the XML description, you can refer to the driver documentation. And in the driver documentation (.rst), add more explanations and examples.

gdal/doc/source/drivers/vector/sosi.rst
```diff
@@ -7,3 +7,35 @@ Norwegian SOSI Standard
 .. shortname:: SOSI
 
 This driver requires the FYBA library.
+
+Open options
+------------
+
+Starting with GDAL 2.0, the following open options can be specified
```
rouault: GDAL 3.1

gdal/ogr/ogrsf_frmts/sosi/ogrsosidriver.cpp
```diff
@@ -121,6 +121,10 @@ void RegisterOGRSOSI() {
     poDriver->SetMetadataItem( GDAL_DCAP_VECTOR, "YES" );
     poDriver->SetMetadataItem( GDAL_DMD_LONGNAME, "Norwegian SOSI Standard" );
     poDriver->SetMetadataItem( GDAL_DMD_HELPTOPIC, "drv_sosi.html" );
+    poDriver->SetMetadataItem( GDAL_DMD_OPENOPTIONLIST,  "<OpenOptionList>"
+    "<Option name='appendFieldsMap' type='string' description='Default is that all rows for equal field names will be appended in a feature, but with this parameter you select what field this should be valid for. With appendFieldsMap=f1&f2, Append will be done for field f1 and f2 using a comma as delimiter.' default=''/>"
```
rouault: See my previous remark regarding XML escaping of &: `f1&f2` --> `f1&amp;f2`
You can remove default=''

rouault:
ogr/ogrsf_frmts/sosi/ogrsosilayer.cpp
FAIL: tabulations detected. Please remove them!

larsop:
Thanks a lot, I could never have done this work without the help from you.

I tried to run the existing test first , but I seems to have a problem to set up thing correctly . Here is output when running the exiting test.

`
 pytest ogr/ogr_sosi.py 
Test session starts (platform: darwin, Python 2.7.16, pytest 4.6.5, pytest-sugar 0.9.2)
rootdir: /Users/lop/dev/sosi_gdal/gdal_update/gdal/autotest, inifile: pytest.ini
plugins: sugar-0.9.2, env-0.6.2
collecting ... 
 ogr/ogr_sosi.py s                                                                                                                                                                           100% ██████████

Results (0.03s):
       1 skipped

`
When check I find SOSI format from command line

 ogr2ogr --formats|grep SOSI
  SOSI -vector- (ro): Norwegian SOSI Standard

rouault:
> but I seems to have a problem to set up thing correctly

ah, because it uses the test_ogrsf application that is not built by default. Just `cd apps; make test_ogrsf` to build it

larsop:
I still have the same problem. The test_ogrsf built with no problems. I also tested to install that package.
`
Larss-MacBook-Pro:autotest lop$ ls -lah ../gdal/apps/.libs/test_ogrsf*
-rwxr-xr-x  1 lop  staff    96K Aug 23 12:01 ../gdal/apps/.libs/test_ogrsf
-rw-r--r--  1 lop  staff   715K Aug 23 12:01 ../gdal/apps/.libs/test_ogrsf.o
Larss-MacBook-Pro:autotest lop$ pytest ogr/ogr_sosi.py 
Test session starts (platform: darwin, Python 2.7.16, pytest 4.6.5, pytest-sugar 0.9.2)
rootdir: /Users/lop/dev/sosi_gdal/gdal_update/gdal/autotest, inifile: pytest.ini
plugins: sugar-0.9.2, env-0.6.2
collecting ... 
 ogr/ogr_sosi.py s                                                                                                                                                                           100% ██████████

Results (0.02s):
       1 skipped
Larss-MacBook-Pro:autotest lop$ 

`

rouault:
Is the directory where test_ogrsf is in the PATH ?

larsop:
Yes it seems like that

 echo $PATH
../gdal/apps/:../gdal/apps/.libs/:

rouault:
Ah, sorry I know what happens. This depends on a remote data file to download. Set the GDAL_DOWNLOAD_TEST_DATA=YES environment variable and re-run the test (the file will be cached in autotest/ogr/tmp/cache after that)

larsop:
Sorry, I still have the same problem. 
The directory did not exist after running the test. So I created it, but the not help either.

`
echo $GDAL_DOWNLOAD_TEST_DATA 
YES

cat pytest.ini 
[pytest]
python_files = *.py
testpaths = ogr gcore gdrivers osr alg gnm utilities pyscripts
env =
GDAL_DOWNLOAD_TEST_DATA=YES 

markers =
    require_driver: Skip test(s) if driver isn't present
    require_run_on_demand: Skip test(s) if RUN_ON_DEMAND environment variable is not set

pytest ogr/ogr_sosi.py 
Test session starts (platform: darwin, Python 2.7.16, pytest 4.6.5, pytest-sugar 0.9.2)
rootdir: /Users/lop/dev/sosi_gdal/gdal_update/gdal/autotest, inifile: pytest.ini
plugins: sugar-0.9.2, env-0.6.2
collecting ... 
 ogr/ogr_sosi.py s                                                                                                                                                                           100% ██████████

Results (0.02s):
       1 skipped

ls -lah ogr/tmp/cache
total 0
drwxr-xr-x  2 lop  staff    64B Aug 23 13:46 .
drwxr-xr-x  4 lop  staff   128B Aug 23 13:46 ..
Larss-MacBook-Pro:autotest lop$ 

`

larsop:
There was one other cache directory there, but that was also empty

find . -name cache
./ogr/tmp/cache
./.pytest_cache/v/cache

find ./.pytest_cache/v/cache/
./.pytest_cache/v/cache/
./.pytest_cache/v/cache//nodeids
./.pytest_cache/v/cache//lastfailed
./.pytest_cache/v/cache//stepwise

rouault:
Add `--capture=no -ra -vv` to the pytest invokation to get full details

larsop:
pytest --capture=no -ra -vv ogr/ogr_sosi.py 
Test session starts (platform: darwin, Python 2.7.16, pytest 4.6.5, pytest-sugar 0.9.2)
cachedir: .pytest_cache
rootdir: /Users/lop/dev/sosi_gdal/gdal_update/gdal/autotest, inifile: pytest.ini
plugins: sugar-0.9.2, env-0.6.2
collecting ... 
 ogr/ogr_sosi.py::test_ogr_sosi_1 s                                                                                                                                                          100% ██████████
========================================================================================= short test summary info ==========================================================================================
SKIPPED [1] /Users/lop/dev/sosi_gdal/gdal_update/gdal/autotest/ogr/ogr_sosi.py:45: <Skipped instance>

Results (0.03s):
       1 skipped

rouault:
I'm afraid I don't have any more ideas, but you don't necessarily need to be able to run the existing test to add a new one. Yours will be simpler as it doesn't require remote data or running test_ogrsf

larsop:
When I also added ./configure --with-python the tests worked OK.

Test session starts (platform: darwin, Python 2.7.16, pytest 4.6.5, pytest-sugar 0.9.2)
rootdir: /Users/lop/dev/sosi_gdal/gdal_update/gdal/autotest, inifile: pytest.ini
plugins: sugar-0.9.2, env-0.6.2
collecting ... 
 ogr/ogr_sosi.py ✓✓                                                                                                                                                                          100% ██████████

Results (1.25s):
       2 passed

autotest/ogr/ogr_sosi.py
```diff
@@ -55,4 +55,17 @@ def test_ogr_sosi_1():
...
+def test_ogr_sosi_2():
+
+    if ogr.GetDriverByName('SOSI') is None:
+        pytest.skip()
+
+    import test_cli_utilities
+    if test_cli_utilities.get_test_ogrsf_path() is None:
+        pytest.skip()
+
+    ret = gdaltest.runexternal(test_cli_utilities.get_test_ogrsf_path() + ' -ro data/test_duplicate_fields.sos -oo appendFieldsMap="BEITEBRUKERID&OPPHAV"')
```
rouault: that's a bit overkill to run test_ogrsf which tests the general driver behaviour, but doesn't verify the working of the open option. You should likely open the dataset with the option, iterate on a feature and check the value of the field with appendFieldsMap
Relevant except from other tests
```
    ds = gdal.OpenEx('data/mvt/datatypes/0/0/0.pbf',
                     open_options=['METADATA_FILE='])
    lyr = ds.GetLayer(0)
   f = lyr.GetNextFeature()
   asser f['foo'] = 'bar'
```

larsop: Thanks, fixed now. Not very used to Python either.

autotest/ogr/ogr_sosi.py
```diff
@@ -54,5 +54,44 @@ def test_ogr_sosi_1():
 
     assert ret.find('INFO') != -1 and ret.find('ERROR') == -1
 
+###############################################################################
+# test using no appendFieldsMap
+
+def test_ogr_sosi_2():
```
rouault: you need to add at the beginning of each tests:
```
    if ogr.GetDriverByName('SOSI') is None:
        pytest.skip()
```

autotest/ogr/ogr_sosi.py
```diff
@@ -54,5 +54,44 @@ def test_ogr_sosi_1():
...
+	ds = gdal.OpenEx('data/test_duplicate_fields.sos',open_options=[])
+	lyr = ds.GetLayer(0)
+	assert lyr.GetFeatureCount() == 17
+	lyr = ds.GetLayer(1)
+	assert lyr.GetFeatureCount() == 1
+	f = lyr.GetNextFeature()
+	print(f)
```
rouault: please remove those 3 print(f)

rouault:
Raised by Python3
```
     File "/home/travis/build/OSGeo/gdal/autotest/ogr/ogr_sosi.py", line 65
E       ds = gdal.OpenEx('data/test_duplicate_fields.sos',open_options=[])
E                                                                        ^
E   TabError: inconsistent use of tabs and spaces in indentation
```

rouault:
Squashed and merged

PR #1795, opened 2019-08-20
OSR: add ExportToPROJJSON() (PROJ >= 6.2)

rouault:


jratike80:
It seems that the new option is not added to the documentation https://gdal.org/programs/gdalsrsinfo.html#gdalsrsinfo, right?

rouault:
> It seems that the new option is not added to the documentation https://gdal.org/programs/gdalsrsinfo.html#gdalsrsinfo, right?

Good catch. Just added

PR #1226, opened 2019-01-22
Support API-level precision/round for WKT

abellgithub:
This provides an API interface to allow selection of precision and formatting of numeric values in WKT.  It allows selection of 'f' or 'g' type formatting and explicit setting of the precision/round options.  It is backward compatible with existing behavior with the exception of using 'E' instead of 'e' in 'g'-formatted numbers, as specified by the OpenGIS Simple Features Specification.

Feedback on this appreciated.  Please note the comments in ogrutils.cpp
Close #1174

PR #1797, opened 2019-08-21
Support API-level precision/round for WKT

### LINKED ISSUE #1174: Support WKT precision in the API

It would be nice if conversion from OGR to WKT would support precision in the API.

OGR supports the option OGR_WKT_PRECISION to configure the precision of conversion from an OGR object to WKT.  However, the value of this variable is stored in a static variable in the functions that do the actual conversion (assume this is done for efficiency) which makes it impossible to set the precision from the API more than once.  It would be nice if the precision could be provided to OGRMakeWktCoordinate() and OGRMakeWktCoordinateM() to allow API control of the WKT output precision.

## Operating system

Any

rouault:
This provides an API interface to allow selection of precision and formatting of numeric values in WKT. It allows selection of 'f' or 'g' type formatting and explicit setting of the precision/round options. It is backward compatible with existing behavior with the exception of using 'E' instead of 'e' in 'g'-formatted numbers, as specified by the OpenGIS Simple Features Specification.

Feedback on this appreciated. Please note the comments in ogrutils.cpp
Close #1174

This is a followup of PR #1226


abellgithub:
Also, thanks a ton for doing this.  Getting through the test set is difficult with my setup.

rouault:
> Do you have a policy about passing args by value?

Sort of. This is suggested by cppcheck. I agree it might not make a performance difference at all here.

abellgithub:
If you're trying to keep clean with cppcheck, I get that.  Thanks.

PR #1817, opened 2019-09-05
gdal2tiles.py: add remaining resample methods

mwtoews:
See https://lists.osgeo.org/pipermail/gdal-dev/2019-August/050696.html

Should there be basic tests for the `resample` option in `gdal/autotest/pyscripts/test_gdal2tiles.py`?


rouault:
> Should there be basic tests for the `resample` option in `gdal/autotest/pyscripts/test_gdal2tiles.py`?

that wouldn't hurt

mwtoews:
A really basic test is now in, that basically checks that valid files are created for each resample option.

PR #1819, opened 2019-09-05
Replace SWIG Python obj0/swig_obj[0] with version agnostic $self

StefanBruens:
"obj0" was only meant as a porting helper and should not be used, even
with swig 3.0, see swig/swig@cd8fc0a.
Use the correct "$self" notation, which is expanded to the correct
code regardless of swig version and options.


rouault:
Thanks

PR #1843, opened 2019-09-17
flatgeobuf.rst: Add ogr2ogr example

AsgerPetersen:
## What does this PR do?
Adds an example of ogr2ogr usage with FlatGeobuf. It took me a while to figure out that the output file MUST have the `.fgb` extension


gdal/doc/source/drivers/vector/flatgeobuf.rst
```diff
@@ -39,6 +39,17 @@ Layer Creation Options
 -  **SPATIAL_INDEX=**\ *YES/NO*: Set the YES to create a
    spatial index. Defaults to YES.
 
+Examples
+--------
```
mloskot: Nitpicking: I think it's safer to assume reST expects heading markup anf paragraph are separated by a blank line

PR #1844, opened 2019-09-17
FlatGeobuf: Filtering fixes

bjornharrtell:
Early adopters has found potential problems when doing spatial filtering but I cannot reproduce yet. This basic test does not hurt though.

### DIFF
--- a/autotest/ogr/ogr_flatgeobuf.py
+++ b/autotest/ogr/ogr_flatgeobuf.py
@@ -180,7 +180,7 @@ def test_ogr_flatgeobuf_2():
     c = fgb_lyr.SetSpatialFilterRect(479586.0,4764618.6,479808.2,4764797.8)
     c = fgb_lyr.GetFeatureCount()
     if ogrtest.have_geos():
-        assert c == 0
+        assert c == 4
     else:
         assert c == 5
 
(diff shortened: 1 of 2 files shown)


rouault:
Seems this breaks on most platforms, and those have GEOS enabled. So I'd suspect your local build doesn't have GEOS enabled. You might different case in the test depending on if GEOS is enabled or not

rouault:
I also notice a `flatbuffers/flatbuffers.h:2101:27: runtime error: unsigned integer overflow: 1260 - 18446744073709551592 cannot be represented in type 'unsigned long'`
The code is probably fine. The method might be qualified with CPL_NOSANITIZE_UNSIGNED_INT_OVERFLOW to avoid it

bjornharrtell:
Will see if I can reproduce issues with GEOS enabled.

PS. I've seen the overflow before and it makes me a bit uneasy, especially considering the "Early adopters has found potential problems" thing.

rouault:
> I've seen the overflow before and it makes me a bit uneasy, especially considering the "Early adopters has found potential problems" thing.

unsigned integer overflow is a well defined behaviour in C, and from the comment in the code, it seems intended here. It is just that most of the time it is not that intended, hence I've enabled it in the sanitize target, so that we are sure to use it when we really need it. Most of the time it can be avoided, or if not practical tagged with this CPL_NOSANITIZE_UNSIGNED_INT_OVERFLOW macro

bjornharrtell:
@rouault you are right, I can reproduce the failures if I compile with GEOS... however, I find it very strange that the GEOS secondary filter is eliminating all candidates where it definitely should not. Suspected axis order issues but doesn't seem so.

bjornharrtell:
I think I found the issue. Filtering was actually not working as it should at all and I believe a bug crept in it in one of the last refactorings I made for GetNextFeature(), so this PR is both now a bugfix and added test.

PR #1845, opened 2019-09-17
FlatGeobuf: Rework/improve error handling

bjornharrtell:
## What does this PR do?

Checks return values for all I/O and memory allocations to make sure to handle potential errors and some tidying up.

### DIFF
--- a/gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
+++ b/gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
@@ -38,6 +38,22 @@
 using namespace flatbuffers;
 using namespace FlatGeobuf;
 
+
+static OGRErr CPLErrorInvalidProperties() {
+    CPLError(CE_Failure, CPLE_AppDefined, "Properties buffer has invalid size");
+    return OGRERR_CORRUPT_DATA;
+}
+
+static OGRErr CPLErrorMemoryAllocation() {
+    CPLError(CE_Failure, CPLE_AppDefined, "Could not allocate memory");
+    return OGRERR_NOT_ENOUGH_MEMORY;
+}
+
+static OGRErr CPLErrorIO() {
+    CPLError(CE_Failure, CPLE_AppDefined, "Unexpected I/O failure");
+    return OGRERR_FAILURE;
+}
+
 OGRFlatGeobufLayer::OGRFlatGeobufLayer(const Header *poHeader, GByte *headerBuf, const char *pszFilename, uint64_t offset)
 {
     CPLDebug("FlatGeobuf", "offset: %lu", static_cast<long unsigned int>(offset));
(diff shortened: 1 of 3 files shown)


gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
```diff
@@ -443,9 +443,10 @@ OGRErr OGRFlatGeobufLayer::querySpatialIndex() {
         VSIFree(m_featureOffsets);
         m_featureOffsets = nullptr;
     }
-    auto featureOffsetsSize = static_cast<size_t>(featuresCount) * 8;
+    auto featureOffetsCount = static_cast<size_t>(featuresCount);
```
rouault: is there a chance that featuresCount could be > std::numeric_limits<size_t> on 32 bit platforms ? (or overflow in the below * 8 multiplication)

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
```diff
@@ -443,9 +443,10 @@ OGRErr OGRFlatGeobufLayer::querySpatialIndex() {
...
+    auto featureOffsetsSize = featureOffetsCount * 8;
     m_featureOffsets = static_cast<uint64_t *>(VSI_MALLOC_VERBOSE(featureOffsetsSize));
-    if (VSIFReadL(m_featureOffsets, 8, static_cast<size_t>(featuresCount), m_poFp) != featuresCount)
+    if (VSIFReadL(m_featureOffsets, 8, featureOffetsCount, m_poFp) != featureOffetsCount)
```
rouault: the above malloc() could potential fail, so null checking should probably be done

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
bjornharrtell: Hmm yes, both issues are relevant.

The max number of features per spec is 64-bit unsigned but I guess practically limited to to 64-bit unsigned / 8 due the logic you point out on 64-bit arch. I don't see any simple way other than that I make hard checks to require lower number of features than that on read/write. If I do my math correctly it should be 536870912 on 32-bit and 2305843009213690000 on 64-bit so hopefully plenty enough. (Other at this point less known practical limitations notwithstanding)

rouault: I was more thinking about corrupted files than real world ones

bjornharrtell: Ok. I've added checks but only for what the header provide, in this case it's the number of found indices in the spatial index. If the spatial index is corrupted things can surely go very bad, I'll consider if I can add some sanity checks there too.

rouault:
There's a conflict to solve now that I've merged the previous PR

bjornharrtell:
Resolved conflict and rebased.

PR #1846, opened 2019-09-17
 GeoJSON driver: use VSIOverwriteFile() to fix update of file on Windows (fixes qgis/QGIS#28580)

rouault:


rouault:

PR #1847, opened 2019-09-18
Openjpeg driver: Memory corruption due to a wrong condition logic

an-ivanov:
## What does this PR do?
` while( (nPair = CPLAtomicInc(&(poJob->nCurPair))) < nPairs ||  !poJob->bSuccess )`
The logic allows accessing poJob->oPairs[nPair] even if nPair >=nPairs in case when !poJob->bSuccess is true
Correct logic should continue the cycle if nPair<nPairs and poJob->bSuccess is true. And break the cycle if nPair>=nPairs or poJob->bSuccess is false.


rouault:
Good catch.. Thanks

rouault:
Backported to 3.0 and 2.4 as well

Issue #1863, opened 2019-09-19
FlatGeobuf driver fails to read file without spatial index


bjornharrtell:
Output is along the lines of:

ERROR 1: Buffer verification failed
FlatGeobuf: m_offset: 312
FlatGeobuf: m_featuresPos: 0
FlatGeobuf: featureSize: 0
ERROR 1: Fatal error parsing feature

PR #1864, opened 2019-09-19
FlatGeobuf: Fixes for index handling

bjornharrtell:
This intends to fix https://github.com/OSGeo/gdal/issues/1863. Problem was I did not have correct logic in place to seek the correct position depending on if the file contains a spatial index or not.

### DIFF
--- a/gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
+++ b/gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
@@ -398,13 +398,18 @@ OGRFeature *OGRFlatGeobufLayer::GetFeature(GIntBig nFeatureId)
     return poFeature;
 }
 
-OGRErr OGRFlatGeobufLayer::querySpatialIndex() {
-    auto indexNodeSize = m_poHeader->index_node_size();
-    if (m_ignoreSpatialFilter || !m_poFilterGeom || m_queriedSpatialIndex || indexNodeSize == 0)
+OGRErr OGRFlatGeobufLayer::readIndex()
+{
+    if (m_queriedSpatialIndex)
         return OGRERR_NONE;
-    CPLDebug("FlatGeobuf", "Attempting spatial index query");
+
+    auto indexNodeSize = m_poHeader->index_node_size();
+    auto featuresCount = m_poHeader->features_count();
+    auto featureOffetsCount = static_cast<size_t>(featuresCount);
+    auto featureOffsetsSize = featureOffetsCount * 8;
+
     if (m_poFp == nullptr) {
-        CPLDebug("FlatGeobuf", "querySpatialIndex: (will attempt to open file %s)", m_osFilename.c_str());
+        CPLDebug("FlatGeobuf", "readIndex: (will attempt to open file %s)", m_osFilename.c_str());
         m_poFp = VSIFOpenL(m_osFilename.c_str(), "rb");
         if (m_poFp == nullptr) {
             CPLError(CE_Failure, CPLE_AppDefined, "Failed to open file");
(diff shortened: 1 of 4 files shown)


rouault:
Perhaps worth adding a test to write & read a file without spatial index ?

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
```diff
@@ -398,68 +398,74 @@ OGRFeature *OGRFlatGeobufLayer::GetFeature(GIntBig nFeatureId)
...
-OGRErr OGRFlatGeobufLayer::querySpatialIndex() {
-    auto indexNodeSize = m_poHeader->index_node_size();
-    if (m_ignoreSpatialFilter || !m_poFilterGeom || m_queriedSpatialIndex || indexNodeSize == 0)
+OGRErr OGRFlatGeobufLayer::readIndex()
+{
+    if (m_queriedSpatialIndex)
         return OGRERR_NONE;
-    CPLDebug("FlatGeobuf", "Attempting spatial index query");
+
+    auto indexNodeSize = m_poHeader->index_node_size();
```
mloskot: I'd change to `auto const`-s where applicable. Otherwise it raises immediate questions to a reader: can these things change? when, in what conditions, and one has to read through the following implementation...

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
```diff
@@ -398,68 +398,74 @@ OGRFeature *OGRFlatGeobufLayer::GetFeature(GIntBig nFeatureId)
...
-    if (VSIFReadL(m_featureOffsets, 8, featureOffetsCount, m_poFp) != featureOffetsCount)
-        return CPLErrorIO();
-    m_queriedSpatialIndex = true;
+
     return OGRERR_NONE;
 }
 
-GIntBig OGRFlatGeobufLayer::GetFeatureCount(int bForce) {
+GIntBig OGRFlatGeobufLayer::GetFeatureCount(int bForce)
+{
```
mloskot: Then, all those `if ... {` may be worth to keep consistent too

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
```diff
@@ -398,68 +398,74 @@ OGRFeature *OGRFlatGeobufLayer::GetFeature(GIntBig nFeatureId)
...
+            m_featuresCount = m_foundFeatureIndices.size();
+            CPLDebug("FlatGeobuf", "%lu features found in spatial index search", static_cast<long unsigned int>(m_featuresCount));
+            m_queriedSpatialIndex = true;
+        }
+
+        if (!m_featureOffsets) {
+            CPLDebug("FlatGeobuf", "Seek to feature offsets index position");
+            if (VSIFSeekL(m_poFp, sizeof(magicbytes) + sizeof(uoffset_t) + headerSize + treeSize, SEEK_SET) == -1)
+                return CPLErrorIO();
+            m_featureOffsets = static_cast<uint64_t *>(VSI_MALLOC_VERBOSE(featureOffsetsSize));
```
mloskot: `m_pFeatureOffsets`? as per https://gdal.org/development/rfc/rfc8_devguide.html

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
```diff
@@ -398,68 +398,74 @@ OGRFeature *OGRFlatGeobufLayer::GetFeature(GIntBig nFeatureId)
...
+            if (!m_featureOffsets)
```
mloskot: There seem to be mixture of `if (m_poFp == nullptr)` and with implicit `nullptr`.
Although `if (!pointer) {...}` is canonical C++, it may be worth to decide on one way and keep things consistent.

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
bjornharrtell: I follow the principle of function signatures being the exception to the rule to keep braces on the same line, same as in linux kernel I believe.

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
```diff
@@ -530,6 +536,7 @@ OGRErr OGRFlatGeobufLayer::parseFeature(OGRFeature *poFeature, OGRGeometry **ogr
         m_offset += m_featureSize + sizeof(uoffset_t);
         fid = m_featuresPos;
     } else {
+        //m_offset = m_offsetInit;
```
mloskot: Is this an actual comment or it's been overlooked?

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
bjornharrtell: Ah, unfortunately was not aware of those guidelines. I would like to go through it all to follow it, but in a separate PR is that ok?

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
bjornharrtell: Left over from bug hunting iteration, should be removed.

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
mloskot: From my perspective separate PR is fine.
FYI, I'm just commenting, not properly reviewing with hard request for changes.

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
mloskot: I see. I thought it's legit and for a moment I even thought the comment was supposed to read
```
 //m_offset == m_offsetInit;
```
what would assume a documenting `assert(m_offset == m_offsetInit);`

bjornharrtell:
Testcase added and may be good to go now if green.

Issue #1849, opened 2019-09-18
fgb_fuzzer: ASSERT: size_ < FLATBUFFERS_MAX_BUFFER_SIZE


rouault:
@bjornharrtell 

Short summary of https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=17312 (that you can't access for now)

The attached reproducer 
[clusterfuzz-testcase-minimized-fgb_fuzzer-5689644401819648.zip](https://github.com/OSGeo/gdal/files/3626579/clusterfuzz-testcase-minimized-fgb_fuzzer-5689644401819648.zip)
 crashes on
```
  | #3 0x7fa7e8d5cc81 in __assert_fail /build/glibc-LK5gWL/glibc-2.23/assert/assert.c:101
-- | --
  | #4 0x603f5a in flatbuffers::Verifier::Verifier(unsigned char const*, unsigned long, unsigned int, unsigned int, bool) gdal/gdal/ogr/ogrsf_frmts/flatgeobuf/flatbuffers/flatbuffers.h:1984:5
  | #5 0x610096 in OGRFlatGeobufLayer::parseFeature(OGRFeature*, OGRGeometry**) gdal/gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp:533:18
  | #6 0x60e76e in OGRFlatGeobufLayer::GetNextFeature() gdal/gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp:482:13
  | #7 0x6331fa in OGR_L_GetNextFeature gdal/gdal/ogr/ogrsf_frmts/generic/ogrlayer.cpp:552:47
  | #8 0x4ea477 in LLVMFuzzerTestOneInput gdal/gdal/fuzzers/ogr_fuzzer.cpp:123:33
```

Easily triggered with:
```
$ ogrinfo clusterfuzz-testcase-minimized-fgb_fuzzer-5689644401819648  -al
ERROR 1: PROJ: internal_proj_create_from_database: crs not found
ERROR 6: toOGRwkbGeometryType: Unknown FlatGeobuf::GeometryType 211
INFO: Open of `clusterfuzz-testcase-minimized-fgb_fuzzer-5689644401819648'
      using driver `FlatGeobuf' successful.

Layer name: 
Geometry: Unknown (any)
Feature Count: 108849084235775
[...]

PR #1865, opened 2019-09-19
FlatGeobuf: Buffer size bound checks

bjornharrtell:
Intends to fix https://github.com/OSGeo/gdal/issues/1849.

### DIFF
--- a/gdal/ogr/ogrsf_frmts/flatgeobuf/ogr_flatgeobuf.h
+++ b/gdal/ogr/ogrsf_frmts/flatgeobuf/ogr_flatgeobuf.h
@@ -42,6 +42,9 @@ class OGRFlatGeobufDataset;
 
 static constexpr const uint8_t magicbytes[8] = { 0x66, 0x67, 0x62, 0x00, 0x66, 0x67, 0x62, 0x00 };
 
+static constexpr const uint32_t header_max_buffer_size = 1048576 - 1;
+static constexpr const uint32_t feature_max_buffer_size = 1073741824 - 1;
+
 struct FeatureItem : Item {
     flatbuffers::DetachedBuffer buf;
     uint8_t *data;
--- a/gdal/ogr/ogrsf_frmts/flatgeobuf/ogr_flatgeobuf.h
+++ b/gdal/ogr/ogrsf_frmts/flatgeobuf/ogr_flatgeobuf.h
@@ -40,10 +40,10 @@ using namespace FlatGeobuf;
 
 class OGRFlatGeobufDataset;
 
-static constexpr const uint8_t magicbytes[8] = { 0x66, 0x67, 0x62, 0x00, 0x66, 0x67, 0x62, 0x00 };
+static constexpr uint8_t magicbytes[8] = { 0x66, 0x67, 0x62, 0x00, 0x66, 0x67, 0x62, 0x00 };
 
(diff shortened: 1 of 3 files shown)


gdal/ogr/ogrsf_frmts/flatgeobuf/ogr_flatgeobuf.h
```diff
@@ -42,6 +42,9 @@ class OGRFlatGeobufDataset;
 
 static constexpr const uint8_t magicbytes[8] = { 0x66, 0x67, 0x62, 0x00, 0x66, 0x67, 0x62, 0x00 };
 
+static constexpr const uint32_t header_max_buffer_size = 1048576;
+static constexpr const uint32_t feature_max_buffer_size = 1073741824 - 1;
```
mloskot: The `const` in `constexpr const` is redundant.

It would be needed only if it was something like constexpr reference to a constant object i.e. `static constexpr int const&`.

bjornharrtell:
Reminder to myself to add reproducer from https://github.com/OSGeo/gdal/issues/1849 as a testcase for this.

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
```diff
@@ -539,6 +539,10 @@ OGRErr OGRFlatGeobufLayer::parseFeature(OGRFeature *poFeature, OGRGeometry **ogr
     if (VSIFReadL(&m_featureSize, sizeof(uoffset_t), 1, m_poFp) != 1)
         return CPLErrorIO();
     CPL_LSBPTR32(&m_featureSize);
+    if (m_featureSize > feature_max_buffer_size) {
+        CPLError(CE_Failure, CPLE_AppDefined, "Feature size too large (>= 2GB)");
```
rouault: feature_max_buffer_size is actually 1 GB.

gdal/ogr/ogrsf_frmts/flatgeobuf/ogr_flatgeobuf.h
rouault: Any particular reason for the  - 1 ?

bjornharrtell:
@rouault, 2GB - 1 is the hard limit for a single flatbuffer. (for some reason I mistakenly wrote it as 1GB - 1, fixed now).

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
bjornharrtell: Was a mistake, fixed.

gdal/ogr/ogrsf_frmts/flatgeobuf/ogr_flatgeobuf.h
bjornharrtell: 2GB - 1 is the hard limit for a single flatbuffer, see https://github.com/google/flatbuffers/blob/917687c7a6357d4a0e9e53e0f051b76cd0187747/include/flatbuffers/base.h#L298-L299.

bjornharrtell:
Looks like #1849 was resolved by previous fix, but this is still a valid improvement.

PR #1870, opened 2019-09-19
 gcore: Fix precision lost with complex data

drons:
This PR replaces `sqrt( R*R + I*I )` method with `std::hypot(R, I)` to fix precision lost within complex absolute values computation.

Function `std::hypot`  implements precise algorithm for calculating absolute value of complex number without undue overflow or underflow.

See: 
* https://en.cppreference.com/w/cpp/numeric/math/hypot
* https://en.wikipedia.org/wiki/Hypot#Implementation


drons:
Thanks!

PR #1872, opened 2019-09-21
Proxy pool: Load band block sizes if not provided at creation time.

drons:
Fixes floating point exception on copy overviews from PRF dataset to destination dataset.
Sample command line:
```
gdal_translate \
-co COPY_SRC_OVERVIEWS=YES \
-of GTiff ./input.prf ./output.tif
```
Error occurs when passing zero block dimensions at the PRF driver ([see: phprfdataset.cpp:138](https://github.com/drons/gdal/blob/fixCoreDumpOnZeroBlockSize/gdal/frmts/prf/phprfdataset.cpp#L138) ).

### DIFF
--- a/gdal/frmts/prf/phprfdataset.cpp
+++ b/gdal/frmts/prf/phprfdataset.cpp
@@ -134,7 +134,9 @@ bool PhPrfDataset::AddTile( const char* pszPartName, GDALAccess eAccessType,
             return false;
         }
 
-        //! \todo What reason for nBlockXSize&nBlockYSize passed to AddSrcBandDescription
+        // Block sizes (nBlockXSize&nBlockYSize) passed as zeros.
+        // They will be loaded when RefUnderlyingRasterBand
+        // function is called on first open of tile's dataset 'poTileDataset'.
         poTileDataset->AddSrcBandDescription(poBand->GetRasterDataType(), 0, 0);
         GDALRasterBand* poTileBand = poTileDataset->GetRasterBand( nBand );
 
--- a/gdal/gcore/gdal_proxy.h
+++ b/gdal/gcore/gdal_proxy.h
@@ -252,6 +252,11 @@ class CPL_DLL GDALProxyPoolDataset : public GDALProxyDataset
     ~GDALProxyPoolDataset() override;
 
     void SetOpenOptions( char** papszOpenOptions );
+
+    // If size (nBlockXSize&nBlockYSize) parameters is zero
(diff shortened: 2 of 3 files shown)


rouault:
why not fixing the PRF driver to provide the correct block size ?

drons:
To load correct block sizes at the creation time, it is necessary to open all subfiles. A PRF file with a thousand subfiles is not uncommon. It is very slow.

rouault:
> To load correct block sizes at the creation time, it is necessary to open all subfiles. A PRF file with a thousand subfiles is not uncommon. It is very slow.

I see. There might be a caveat though, because nothing pervents non-virtual methods (like GetLockedBlockRef(), GetBlockSize(), etc...) to access directly nBlockXSize/nBlockYSize before RefUnderlyingRasterBand() is called. This is probably unlikely to happen in practice, but still a risk. I think a comment both in RefUnderlyingRasterBand() and in phprfdataset.cpp#L138 about that would be needed.

drons:
@rouault 
Function GDALProxyPoolDataset::AddSrcBandDescription is used in several drivers (Derived, Dimap, Sentinel2, TIL). 
'Derived' driver load actual nBlockXSize & nBlockYSize, but other set up block sizes to 'single line' or '128x128'. I think this may not be optimal.
Maybe later to do it the same way as in the PRF driver?

rouault:
> I think this may not be optimal.

Normally the values in the GDALProxyPoolRasterBand should be set to the same value than the underlying band, otherwise chaos might *theoretically* happen (although I'm not sure in practice that it could happen). So there's nothing about optimality here. You should just use the value of the underlying band, and that might be scanline or tile based depending on the driver. That said, fundamentally it is quite fragile that we depend on that as in the VRT case the driver that was used to fill the block size might not be the same than the one that will read them (thinking about the different JPEG2000 capable drivers here). I've not a good, robust and easy to implement solution for that...  Well, one solution would be to make nBlockXSize & nBlockYSize private member, make sure nobody use them directly and GetBlockSize() virtual, in that case nBlockXSize & nBlockYSize wouldn't have to be set at the proxy level, but that'd be a painful change, and as we don't have compelling cases where the current approach break, there's little incentive in doing that.

drons:
> but that'd be a painful change

Yes, nBlockXSize - is used ~1100 times and  nBlockYSize ~800.

I think it’s better to merge what is already here.

drons:
Thanks!

PR #1877, opened 2019-09-23
FlatGeobuf: Bounds check point coordinate arrays

bjornharrtell:
Aims to fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=17603.


autotest/ogr/ogr_flatgeobuf.py
```diff
@@ -247,3 +247,8 @@ def test_ogr_flatgeobuf_9():
         rc = verify_flatgeobuf_copy(test[0], test[1], test[2])
         assert rc, ('Verification of copy of ' + test[0] + '.shp failed')
 
+def test_ogr_flatgeobuf_ossfuzz_17603():
+    fgb_ds = ogr.Open('data/fgb/clusterfuzz-testcase-minimized-fgb_fuzzer-5725251593306112.fgb')
```
rouault: I've very rarely included myself a ossfuzz generated test case in autotest, because they are usually corrupted in several ways, and a later hardening fix might make one test case completely irrelevant (or completely rejected) for what it used to be tested to. So when possible, I try to corrupt a valid dataset in a unit way, just testing one potential defect. But this is a lot of work, potentially requiring to do coding for binary formats, so I tend not doing it.
Just wanting to share my experience/thoughts about that. I'm fine with you letting this at it (that's what oss-fuzz actually recommands). Just that at some later point it might no longer test-stress the code you fixed.

bjornharrtell: Ok yes, I can see it becoming worthless to regress test the fuzz reproductions and it might be difficult to know when that happens so I'll opt for not having it, removed now.

rouault:
Build error on ubuntu_1804
```
In file included from header_generated.h:7:0,
                 from ogr_flatgeobuf.h:35,
                 from ogrflatgeobuflayer.cpp:36:
flatbuffers/flatbuffers.h: In member function ‘OGRPoint* OGRFlatGeobufLayer::readPoint(const FlatGeobuf::Feature*, uint32_t)’:
flatbuffers/flatbuffers.h:227:48: error: null pointer dereference [-Werror=null-dereference]
   uoffset_t size() const { return EndianScalar(length_); }
                                                ^~~~~~~
```

bjornharrtell:
Ah yes I now see why. The joy of error checking in C/C++... :) (https://en.wikipedia.org/wiki/Tony_Hoare#cite_note-23)

bjornharrtell:
I'm actually not happy about having to add all that checking in https://github.com/OSGeo/gdal/pull/1877/commits/d0953b10f150927bda14c9898b1ddc531bfa6aa1, the code becomes much harder to understand.. adding null checks too will only make it worse so I'm trying to come up with something nicer or at least more decoupled to the actual logic.

bjornharrtell:
I'll close this as I need to put more thought into how to attack corrupt data and to be able to focus on improving and verifying the existing features for now (with fx. https://github.com/OSGeo/gdal/pull/1885).

PR #1882, opened 2019-09-25
Pull request for Pitney Bowes MapInfo Extended Tab (Vector) and MRR (Raster) format drivers

anshulgoelpb27:
## What does this PR do?
It add two new drivers for Pitney Bowes MapInfo Extended Tab (Vector) and MRR (Raster). Drivers will be enabled in default compile of GDAL. 
They only load if the respective SDK's are present. These SDK's are free to download and documentation on obtaining the SDK's have been updated with PR.
Added new python tests along with test data, for the drivers.

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:


anshulgoelpb27:
Fixing the warnings.

nyalldawson:
Is a new driver for extended tab really warranted? Why not just extend the existing driver and add support there?

palmerj:
> Is a new driver for extended tab really warranted? Why not just extend the existing driver and add support there?

I agree, why not extend the existing driver? It's more likely to get merged.

carypeebles:
The MITAB driver is not created or supported by PB and does not currently support the extened TAB format. We have consolidated codebases internally during the creation of extended TAB so that we only have one true source of support across products and operating systems. We would like this to be true for GDAL as well. We did evaluate enhancing MITAB along with fixing issues we have with it but we would prefer to only maintain one source of truth. We have future plans for further enhancing the format as well and this also factors into our desire to create a new driver based on this SDK.

nyalldawson:
> The MITAB driver is not created or supported by PB and does not currently support the extened TAB format. We have consolidated codebases internally during the creation of extended TAB so that we only have one true source of support across products and operating systems. We would like this to be true for GDAL as well. We did evaluate enhancing MITAB along with fixing issues we have with it but we would prefer to only maintain one source of truth. We have future plans for further enhancing the format as well and this also factors into our desire to create a new driver based on this SDK.

Sorry - but I don't find that a compelling argument. You are saying it makes maintenance easier for you, yet you're pushing the burden of maintaining two separate drivers back to the open-source community, at considerable burden to the GDAL maintainers.

Worse, this results in a less-than-ideal situation for users. Personally, I think most users would greatly prefer the stable, available out-of-the-box and **everywhere** mitab driver then a new one which relies on a closed source binary blob (which is only available via a registered download and which must be manually installed).

gdal/ogr/ogrsf_frmts/EFAL/ogrgeopackageutility.h
```diff
@@ -0,0 +1,47 @@
+/*****************************************************************************
+* Copyright 2016 Pitney Bowes Inc.
```
hobu: Is this file needed? It appears to be the existing [ogrgeopackageutility.cpp](https://github.com/OSGeo/gdal/blob/master/gdal/ogr/ogrsf_frmts/gpkg/ogrgeopackageutility.cpp)  file from the gpkg driver with Pitney Bowes (c) on it instead of the original author @pramsey. 

If this is cruft it will have to be cleaned up. If it is used, it is not legitimate to take an existing file and change the copyright.

anshulgoelpb27: This file is used in our driver. I have reverted back the copyright to original.

rouault: Interesting... Anyway, code duplication inside the code base is not the right solution. It is quite strange that the linker didn't complain about symbol duplication. I'd say that those 2 files should be moved from ogr/ogrsf_frmts/gpkg to the top ogr/ directory

rouault:
Thinking that a good contribution for this community would be that you help maintaining on how to map TAB datum information to EPSG codes. This is currently a burden for us to figure that out. Latest example is this : https://github.com/OSGeo/gdal/commit/e14c9aedfcb8943411e6498c719cbf2db63d8684 . The list we have is certainly incomplete.

rouault:
Perhaps you also need to reconsider how you envision your relationship with the GDAL project even  before attempting to add new stuff into it: read well what you wrote in https://community.pitneybowes.com/communities/community-home/digestviewer/viewthread?MessageKey=07f915b8-0c0f-4d76-862b-f2f650b04720&CommunityKey=fe4d1ee2-f1c6-4307-87d2-00bea1344885&tab=digestviewer . "We do not test VRT much because it is up to GDAL and we have no control over that" . Should I recall that GDAL is an "open source" project. You can certainly have control on it. But you need to demonstrate willingness and that you care about it, and that's not just random technology someone obliged you to use.

anshulgoelpb27:
After discussing with the GDAL PSC, we will be splitting this into two separate PRs for the two drivers. We will also be following the plugin model for both EFAL and MRR format driver.

nyalldawson:
> We will also be following the plugin model for both EFAL and MRR format driver.

I'm extremely disappointed to hear this. You are doing your users a great disservice with this approach. It's a shortsighted, paranoid and frankly outdated approach. I would **strongly** recommend that you reconsider this if you intend for your formats to be widely used outside of MapInfo itself.

rouault:
@anshulgoelpb27 Due to overall community reaction, I'm not even sure that the "proper" plugin approach is going to appeal to anyone.

andreiveselov:
> Thinking that a good contribution for this community would be that you help maintaining on how to map TAB datum information to EPSG codes. This is currently a burden for us to figure that out. Latest example is this : [e14c9ae](https://github.com/OSGeo/gdal/commit/e14c9aedfcb8943411e6498c719cbf2db63d8684) . The list we have is certainly incomplete.

Even, I think this is a great suggestion. What should be the process to get it done?

rouault:
> What should be the process to get it done?

Not sure if "through a pull request" really answers your question.
(Just wanted to mention, speaking only on my behalf, I'd consider this as unrelated to any discussion on how to get MapInfo Extended TAB or MRR into GDAL)

miteshpb:
Created separate PR for MRR Raster format.

https://github.com/OSGeo/gdal/pull/2618

PR #1885, opened 2019-09-25
FlatGeobuf: WKT roundtrip tests and nullptr validation

bjornharrtell:
Adds complete set of 2D WKT geometry roundtrip test and some refactoring

### DIFF
--- a/gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
+++ b/gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
@@ -38,7 +38,7 @@
 using namespace flatbuffers;
 using namespace FlatGeobuf;
 
-static nullptr CPLErrorInvalidPointer() {
+static OGRGeometry *CPLErrorInvalidPointer() {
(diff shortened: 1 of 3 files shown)


gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
```diff
@@ -797,8 +797,13 @@ OGRMultiPolygon *OGRFlatGeobufLayer::readMultiPolygon(const Feature *feature, ui
 OGRGeometry *OGRFlatGeobufLayer::readGeometry(const Feature *feature)
 {
     auto pXy = feature->xy();
-    if (pXy == nullptr)
-        return nullptr;
+    auto pZ = feature->z();
+    auto pM = feature->m();
+    VALIDATE_POINTER1(pXy, "readGeometry", nullptr);
```
rouault: in a debug build, this will issue a CE_Fatal error, which will abort the process. So this macro is only to be used to detect programming errors, not corrupt data

bjornharrtell: Ah, I see. Replaced with standard nullptr checks.

PR #1898, opened 2019-09-30
FlatGeobuf: Fix Z/M handling

### LINKED ISSUE #1897: FlatGeobuf driver have problems handling Z/M dimensions

bjornharrtell:
Aims to fix #1897. Have not been able to find out why I'm not getting the expected geometry and or geometry type in roundtrip test.


rouault:
> why I'm not getting the expected geometry and or geometry type in roundtrip test.

use ExportToIsoWkt() instead of ExportToWkt()

bjornharrtell:
Thanks, not sure how I missed that... anyway, reworked and rebased including a fix for an error with regard to Z/M in the readSimpleCurve method.

PR #1906, opened 2019-10-04
Add gdal_viewshed 

hobu:
Implementation of https://www.asprs.org/wp-content/uploads/pers/2000journal/january/2000_jan_87-90.pdf by @szekerest. I just added it to the makefiles.

Outstanding issues:

- [ ]  The `-3d` option doesn't seem to be hooked up yet
- [ ]  No documentation provided


gdal/nmake.opt
```diff
@@ -136,7 +136,7 @@ CXX_PDB_FLAGS=
 !IF "$(DEBUG)" == "0"
 OPTFLAGS= $(CXX_ANALYZE_FLAGS) $(CXX_PDB_FLAGS) /nologo /MP /MD /EHsc /Ox /FC /D_CRT_SECURE_NO_DEPRECATE /D_CRT_NONSTDC_NO_DEPRECATE /DNDEBUG
 !ELSE
-OPTFLAGS= $(CXX_ANALYZE_FLAGS) $(CXX_PDB_FLAGS) /nologo /MP /MDd /EHsc /FC /D_CRT_SECURE_NO_DEPRECATE /D_CRT_NONSTDC_NO_DEPRECATE /DDEBUG
+OPTFLAGS= $(CXX_ANALYZE_FLAGS) $(CXX_PDB_FLAGS) /nologo /MP /MD /EHsc /FC /D_CRT_SECURE_NO_DEPRECATE /D_CRT_NONSTDC_NO_DEPRECATE /DDEBUG
```
rouault: all changes in nmake.opt are irrelevant and should be reverted

gdal/alg/viewshed.cpp
```diff
@@ -0,0 +1,612 @@
...
+ *
+ * @param pProgressArg The callback data for the pfnProgress function.
+ *
+ * @return CE_None on success or CE_Failure if an error occurs.
+ */
+
+#define EARTH_DIAMETER 12741994.0;
+
+#ifndef round
+inline double round( double d )
```
rouault: std::round() is standard C++11

gdal/alg/viewshed.cpp
```diff
@@ -0,0 +1,612 @@
...
+ *
+ * @param pfnProgress A GDALProgressFunc that may be used to report progress
+ * to the user, or to interrupt the algorithm.  May be NULL if not required.
+ *
+ * @param pProgressArg The callback data for the pfnProgress function.
+ *
+ * @return CE_None on success or CE_Failure if an error occurs.
+ */
+
+#define EARTH_DIAMETER 12741994.0;
```
rouault: we should avoid hardcoding this, but rather coimpute that from the semi major axis of the SRS definition

gdal/alg/viewshed.cpp
```diff
@@ -0,0 +1,612 @@
...
+
+#ifndef round
+inline double round( double d )
+{
+    return floor( d + 0.5 );
+}
+#endif
+
+
+void SetVisibility(int iPixel, double dfZ, double dfZTarget, double* padfZVal,
```
rouault: static, or use a anonymous namespace

gdal/alg/viewshed.cpp
```diff
@@ -0,0 +1,612 @@
...
+    if (padfZVal[iPixel] + dfZTarget < dfZ)
+    {
+        padfZVal[iPixel] = dfZ;
+        pabyResult[iPixel] = byInvisibleVal;
+    }
+    else
+        pabyResult[iPixel] = byVisibleVal;
+}
+
+int AdjustHeightInRange(double* adfGeoTransform, int iPixel, int iLine, double& dfHeight, double dfDistance2, double dfCurvCoeff)
```
rouault: static

rouault: const double* adfGeoTransform

gdal/alg/viewshed.cpp
```diff
@@ -0,0 +1,612 @@
...
+    if (dfCurvCoeff != 0)
+        dfHeight -= dfCurvCoeff * dfR2 / EARTH_DIAMETER;
+    
+    if (dfDistance2 > 0 && dfR2 > dfDistance2)
+        return FALSE;
+
+    return TRUE;
+}
+
+double CalcHeightLine(int i, double Za, double Zo)
```
rouault: static

gdal/alg/viewshed.cpp
```diff
@@ -0,0 +1,612 @@
...
+{
+    if (i == 1)
+        return Za;
+    else
+        return (Za - Zo) / (i - 1) + Za;
+}
+
+double CalcHeightDiagonal(int i, int j, double Za, double Zb, double Zo)
```
rouault: static

gdal/alg/viewshed.cpp
```diff
@@ -0,0 +1,612 @@
...
+{
+    return ((Za - Zo) * i + (Zb - Zo) * j) / (i + j - 1) + Zo;
+}
+
+double CalcHeightEdge(int i, int j, double Za, double Zb, double Zo)
```
rouault: static

gdal/alg/viewshed.cpp
```diff
@@ -0,0 +1,612 @@
...
+{
+    if (i == j)
+        return CalcHeightLine(i, Za, Zo);
+    else
+        return ((Za - Zo) * i + (Zb - Zo) * (j - i)) / (j - 1) + Zo;
+}
+
+double CalcHeight(double dfZ, double dfZ2, GDALViewshedMode eMode)
```
rouault: static

gdal/alg/viewshed.cpp
```diff
@@ -0,0 +1,612 @@
...
+#include "ogr_core.h"
+#include "commonutils.h"
+
+CPL_CVSID("$Id$")
+
+/************************************************************************/
+/*                        GDALViewShedGenerate()                         */
+/************************************************************************/
+
+/**
```
rouault: should be moved just above the function, otherwise Doxygen will not like it

gdal/alg/viewshed.cpp
```diff
@@ -0,0 +1,612 @@
...
+ * on the requested set of contour levels.  The vector contours are written
+ * to the passed in OGR vector layer.  Also, a NODATA value may be specified
+ * to identify pixels that should not be considered in contour line generation.
+ *
+ * The gdal/apps/gdal_viewshed.cpp mainline can be used as an example of
+ * how to use this function.
+ *
+ * ALGORITHM RULES
+ *
+ * @param hBand The band to read raster data from.  The whole band will be
```
rouault: the parameters described are a copy & paste from GDALContour()

gdal/alg/viewshed.cpp
```diff
@@ -0,0 +1,612 @@
...
+    if( pfnProgress == nullptr )
+        pfnProgress = GDALDummyProgress;
+
+    if( !pfnProgress( 0.0, "", pProgressArg ) )
+    {
+        CPLError( CE_Failure, CPLE_UserInterrupt, "User terminated" );
+        return CE_Failure;
+    }
+
+    GByte byNoDataVal = dfNoDataVal >= 0 ? static_cast<GByte>(dfNoDataVal) : 0;
```
rouault: upper bound should probably be tested as well.  And one could also test CPLIsNan() for pedanticity...

gdal/alg/viewshed.cpp
```diff
@@ -0,0 +1,612 @@
...
+        return dfZ2;
+    else if (eMode == GVM_Max)
+        return std::max(dfZ, dfZ2);
+    else if (eMode == GVM_Min)
+        return std::min(dfZ, dfZ2);
+    else
+        return dfZ;
+}
+
+CPLErr GDALViewshedGenerate(GDALRasterBandH hBand, const char* pszTargetRasterName,
```
rouault: in addition to pszTargetRasterName, I'd argue for a driver name and creation options so that people can do a MEM dataset, or compress the TIFF

gdal/alg/viewshed.cpp
```diff
@@ -0,0 +1,612 @@
...
+                    double dfObserverX, double dfObserverY, double dfObserverHeight,
+                    double dfTargetHeight, double dfVisibleVal, double dfInvisibleVal,
+                    double dfOutOfRangeVal, double dfNoDataVal, double dfCurvCoeff,
+                    GDALViewshedMode eMode, double dfMaxDistance,
```
rouault: experience shows that extending the API might come at a later point, so a CSLConstList papszExtraOptions as a provision wouldn't hurt

gdal/alg/viewshed.cpp
```diff
@@ -0,0 +1,612 @@
...
+            hDstDS = GDALCreate(hDriver, pszTargetRasterName, nXSize, nYStop - nYStart, 1,
+                    GDT_Byte, nullptr);
+            if (hDstDS == nullptr)
+            {
+                CPLError(CE_Failure, CPLE_AppDefined,
+                    "Cannot create dataset for %s", pszTargetRasterName);
+                return CE_Failure;
+            }
+
+            const char *pszProjection = GDALGetProjectionRef(hSrcDS);
```
rouault: I'd suggest using GDALGetSpatialRef() / GDALSetSpatialRef() now

gdal/alg/viewshed.cpp
```diff
@@ -0,0 +1,612 @@
...
+    int nXStart = dfMaxDistance > 0? std::max(0, static_cast<int>(floor(nX - adfInvGeoTransform[1] * dfMaxDistance))) : 0;
+    int nXStop = dfMaxDistance > 0? std::min(nXSize, static_cast<int>(ceil(nX + adfInvGeoTransform[1] * dfMaxDistance) + 1)) : nXSize;
+    int nYStart = dfMaxDistance > 0? std::max(0, static_cast<int>(floor(nY + adfInvGeoTransform[5] * dfMaxDistance))) : 0;
+    int nYStop = dfMaxDistance > 0 ? std::min(nYSize, static_cast<int>(ceil(nY - adfInvGeoTransform[5] * dfMaxDistance) + 1)) : nYSize;
+
+    /* normalize horizontal index (0 - nXSize) */
+    nXSize = nXStop - nXStart;
+    nX -= nXStart;
+
+    double *padfFirstLineVal = static_cast<double *>(
```
rouault: memory leaks may happen in error code paths below.
I'd probably suggest using std::vector instead, and .resize() them in a try {} block to catch for memory errors

gdal/alg/viewshed.cpp
rouault: what about returning the resulting destination dataset instead of a CPLEr ?

gdal/alg/viewshed.cpp
```diff
@@ -0,0 +1,612 @@
...
+    {
+        CPLFree(padfFirstLineVal);
+        CPLFree(padfLastLineVal);
+        CPLFree(padfThisLineVal);
+        CPLFree(pabyResult);
+        return CE_Failure;
+    }
+
+    GDALRasterBandH hTargetBand = nullptr;
+    GDALDatasetH hDstDS = nullptr;
```
rouault: use a std::unique_ptr<GDALDataset> to avoid memory leaks

gdal/alg/viewshed.cpp
mloskot: `std::array`?

gdal/alg/viewshed.cpp
thareUSGS: Please don't hard-wire the diameter for Earth. This is easily and can be more generically calculated from the radius as defined in the file.  For rover planning, we commonly use viewshed analysis on Mars.

gdal/alg/viewshed.cpp
thareUSGS: +1. Please don't hard-wire the diameter for Earth. This is easily and can be more generically calculated from the radius as defined in the file. For rover planning, we commonly use viewshed analysis on Mars.

gdal/apps/gdal_viewshed.cpp
```diff
@@ -0,0 +1,251 @@
...
+                  nBandIn );
+        exit(2);
+    }
+
+/* -------------------------------------------------------------------- */
+/*      Try to get a coordinate system from the raster.                 */
+/* -------------------------------------------------------------------- */
+    OGRSpatialReferenceH hSRS = nullptr;
+
+    const char *pszWKT = GDALGetProjectionRef( hSrcDS );
```
rouault: pszWKT and hSRS are currently unused

gdal/apps/gdal_viewshed.cpp
```diff
@@ -0,0 +1,251 @@
...
+    double dfNoDataVal = 0.0;
+    double dfCurvCoeff = 0.0;
+    const char *pszSrcFilename = nullptr;
+    const char *pszDstFilename = nullptr;
+    bool bQuiet = false;
+    GDALProgressFunc pfnProgress = nullptr;
+
+    // Check that we are running against at least GDAL 1.4.
+    // Note to developers: if we use newer API, please change the requirement.
+    if (atoi(GDALVersionInfo("VERSION_NUM")) < 1400)
```
rouault: this check is wrong. Could completely by removed

szekerest:
> If I'm super pedantic, I believe that the algorithm probably only works "correctly" if the DEM is expressed as ellipsoidal heights, and not geoid-relative height. For a DEM with a moderate spatial extent, the geoid undulation can be considered as a constant over the extent considered, and that constant is neglectable compared to the radius of the planet (within a few tens of meters on Earth), so the end result is probably correct. But as the algorithm take into account the curvature, there might be a small effect due to this if you worked with an observer height quite big to have an horizon of visibility of several hundred kilometers where the geoid undulation is no longer a constant (but a difference in height of a few tens of meters at a range of a few hundred kilometers is likely within the noise). Anyway, I'm not calling for algorithm changes, but perhaps documenting potential limitations/assumptions. I'm not even sure they are really a concern in practice. I was mostly talking out loud :-)

The comment seems correct. The original algorithm didn't take the earth curvature + refraction into account, so that I have added a simple parabolic approximation which can handle the refraction+curvature in a single constant. This is the same as what have been implemented in grass r.viewshed and it is indeed working with a relatively small area. It would probably better to approx. the curvature as a circle or ellipse function with the cost of adding an sqrt in the formula and handle the effect of the athmospheric refraction with a separate expression tag.

thareUSGS:
Adding refraction would be a nice touch, but I hope it can be turned off or modified during the run from the command-line.

gdal/alg/viewshed.cpp
hobu: Will set to `inline static void`.

hobu:
gah! I didn't mean to merge this, but it wasn't in such a terrible state that it couldn't be merged.

hobu:
I still have some more work to do on it though. Your call @rouault on what to do in terms of pulling it from master and rewriting history.

gdal/apps/makefile.vc
```diff
@@ -22,7 +22,7 @@ GNM_PROGRAMS =	gnmmanage.exe gnmanalyse.exe
 
 default:	gdal_translate.exe gdalinfo.exe gdalserver.exe gdaladdo.exe gdalwarp.exe \
 		nearblack.exe gdalmanage.exe gdalenhance.exe gdaltransform.exe\
-		gdaldem.exe gdallocationinfo.exe gdalsrsinfo.exe gdalmdiminfo.exe gdalmdimtranslate.exe $(OGR_PROGRAMS) $(GNM_PROGRAMS)
+		gdaldem.exe gdallocationinfo.exe gdalsrsinfo.exe gdalmdiminfo.exe gdalmdimtranslate.exe viewshed.exe $(OGR_PROGRAMS) $(GNM_PROGRAMS)
```
jef-n: should be gdal_viewshed.exe

gdal/alg/viewshed.cpp
hobu: Implemented in b694fa4abd43bf198eb8c64d30be1bb3a09180b6

gdal/alg/viewshed.cpp
hobu: other "free-standing" algorithms in gdal_alg.h all have this pattern.

rouault: The interest would be in particular to return a dataset for the MEM driver, that cannot be reopened once it is closed. Functions in apps/gdal_utils.h try to return a dataset for that reason.
As far as I can see, most of the functions in gdal_alg.h do not create a new dataset, but updates an existing one.

hobu: I do not have an opinion about refactoring it in this way. Please make the change as you see fit.

rouault: done per https://github.com/OSGeo/gdal/pull/1919/commits/d768e94b7b57f04daeedf36a1cb5ecd84ca3a32a

PR #1922, opened 2019-10-09
use /MP# compilation switch for parallel windows builds

hobu:
Support parallel MSVC builds based on `NUMBER_OF_PROCESSORS` environment variable. See https://docs.microsoft.com/en-us/cpp/build/reference/mp-build-with-multiple-processes?view=vs-2015 for more information


rouault:
What is the difference w.r.t not specifying the number of processes ?
The page you quote mentions "If you omit the processMax argument, the compiler retrieves the number of effective processors on your computer from the operating system, and creates a process for each processor."
So to me it looks like /MP and /MP$(NUMBER_OF_PROCESSORS) are equivalent. Or there is a subtelty between $(NUMBER_OF_PROCESSORS)  and "the number of effective processors" ?

hobu:
I was trying to pull across patches from conda forge https://github.com/conda-forge/gdal-feedstock/blob/master/recipe/patches/multiprocessor.patch to make it easier for them to adapt their builds. They are specifying an explicit processor count there. Maybe we should change to allow a variable to be overridden to do this.

rouault:
> Maybe we should change to allow a variable to be overridden to do this.

From a quick search, it looks like NUMBER_OF_PROCESSORS is a system variable ? So yes probably have a custom variable, BUILD_CPU_COUNT?, that defaults to NUMBER_OF_PROCESSORS.

PR #1924, opened 2019-10-10
Transform MAX_OBJECT_SIZE to runtime environment option - OGR_GEOJSON…

mmomtchev:
Would it be possible to consider this pull request for transforming MAX_OBJECT_SIZE in the GeoJSON driver to a runtime option?
For me this is a major breakage when upgrading from Debian 9 (GDAL 2.1) to Debian 10 (GDAL 2.4). In GDAL 2.1 GeoJSON objects were limited only by the available memory. In fact, I think it would be impossible to have a hard-coded value that works for both the notebook users and the big iron. I agree that GeoJSON is a horrible format that should not be used for applications meant to scale, but we need at least some way of importing this data.


rouault:
Can you document it in doc/source/drivers/vector/geojson.rst ?

PR #1925, opened 2019-10-10
OGR: Made SQL query parse API public

drons:
It will be useful for developing third-party OGR drivers plugins.


rouault:
I'm not so enthusiastic about this being advertized as this is mostly an internal implementation detail and I wouldn't want us to be tied for the future. But I understand the need.
I'd suggest:
-  swq.h should be renamed ogr_swq.h to namespace it
- add in cpl_port.h a CPL_UNSTABLE_API macro, that is an alias of CPL_DLL, but convey the right message to users, and use it to export those symbols

PR #1926, opened 2019-10-10
Fix --with-proj in autoconf on Linux

mmomtchev:
Currently --with-proj is broken on Linux if one has proj4 installed in the default location
The system proj4 is usually in the same directory as netcdf and geos and these will insert that directory before the --with-proj= directory
configure will use libproj6 when checking and then libtool will miserably fail trying to link with libproj4
As most people that will be building gdal 3.x from source will be building with a custom proj6 library and against the system libraries for everything else, this can be a serious problem

I am not an autoconf expert and it is possible that I broke another platform, maybe the GDAL autoconf expert should look into this issue

### DIFF
--- a/gdal/configure
+++ b/gdal/configure
@@ -32223,7 +32223,11 @@ fi
     if test "$NETCDF_SETTING" = "yes" ; then
       EXTRA_INCLUDES="-I$NETCDF_INCLUDEDIR $EXTRA_INCLUDES"
       NETCDF_ROOT=$NETCDF_PREFIX
-      LIBS="$NETCDF_LIBS $LIBS"
+      if test "$with_netcdf" = "yes" -o "$with_netcdf" = "" ; then
+        LIBS="$LIBS $NETCDF_LIBS"
+      else
+        LIBS="$NETCDF_LIBS $LIBS"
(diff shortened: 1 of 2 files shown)


rouault:
I'm not sure this patch is reliable, or at the very least it is fragile. It relies on the order of -L instructions. The situation and issue you describe is real. https://trac.osgeo.org/gdal/wiki/BuildingOnUnixGDAL25dev proposes something that is more reliable

mmomtchev:
I think it is reliable as far as the behavior of the linker on the order of -L is defined, but it is fragile as it could easily be broken by another modification of configure.ac
I wonder if there isn't any more "official" way of specifying this in libtool and autoconf as this is a common problem and there is a clear way to differentiate libproj-12 from libproj-15 as they have different major versions
It is a pure compilation issue, once the .so is built the linker will include an rpath to the right library
Patching binaries seems awful to me, having different versions of the same library is something that was supposed to work in UNIX

mmomtchev:
Yes, actually there is no universal solution. There is a 3.x package in Debian experimental and in the past it went the only possible way - by static-linking proj. I am closing this PR.

PR #1930, opened 2019-10-13
FlatGeobuf: Fixes for weaknesses detected by oss-fuzz

bjornharrtell:
Aims to mitigate oss-fuzz issues 18092, 18103, 18036 and 18057.

### DIFF
--- a/gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
+++ b/gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
@@ -38,12 +38,12 @@
 using namespace flatbuffers;
 using namespace FlatGeobuf;
 
-static OGRGeometry *CPLErrorInvalidPointer() {
+static std::nullptr_t CPLErrorInvalidPointer() {
     CPLError(CE_Failure, CPLE_AppDefined, "Unexpected nullptr - possible data corruption");
     return nullptr;
 }
 
-static OGRGeometry *CPLErrorInvalidLength() {
+static std::nullptr_t CPLErrorInvalidLength() {
     CPLError(CE_Failure, CPLE_AppDefined, "Invalid length detected - possible data corruption");
     return nullptr;
 }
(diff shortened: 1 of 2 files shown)


gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
```diff
@@ -39,12 +39,17 @@ using namespace flatbuffers;
 using namespace FlatGeobuf;
 
 static OGRGeometry *CPLErrorInvalidPointer() {
-    CPLError(CE_Failure, CPLE_AppDefined, "Possible data corruption - unexpected nullptr");
+    CPLError(CE_Failure, CPLE_AppDefined, "Unexpected nullptr - possible data corruption");
```
rouault: if this method return std::nullptr_t , then you should be able to save the explicit casts when using it

gdal/ogr/ogrsf_frmts/flatgeobuf/packedrtree.cpp
```diff
@@ -318,8 +318,9 @@ uint64_t PackedRTree::size(const uint64_t numItems, const uint16_t nodeSize)
     if (numItems == 0)
         throw std::invalid_argument("Number of items must be greater than 0");
     const uint16_t nodeSizeMin = std::min(std::max(nodeSize, static_cast<uint16_t>(2)), static_cast<uint16_t>(65535));
-    if (numItems > std::numeric_limits<uint64_t>::max() - ((numItems / nodeSizeMin) * 2))
-        throw std::overflow_error("Number of items too large");
+    // limit so that resulting size in bytes can be represented by uint64_t
+    if (numItems > (2^56))
```
rouault: huh ! Is it valid C++ ? I think 2 ^ 56 is rather evaluated as a bit-to-bit xor instead of exponentiation. You should likely use static_cast<uint64_t>(1) << 56

bjornharrtell: oh! :D

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
bjornharrtell: Ah - did read about std::nullptr_t but got confused, I see now it works well for this case. Tthanks!

bjornharrtell:
Reminder to myself to always run tests locally before pushing. :)

PR #1931, opened 2019-10-15
add HDF5_H5_IS_DLL variable to switch the scenario when HDF5 is built…

hobu:
Conda feedstock has to patch for HDF5 DLL situation. https://github.com/conda-forge/gdal-feedstock/blob/master/recipe/patches/windowshdf5.patch

Add HDF5_H5_IS_DLL  to add the appropriate compilation definitions.


rouault:
Could this HDF5_H5_IS_DLL variable be 'documented' in the HDF5 section of nmake.opt ?

PR #1949, opened 2019-10-23
JP2KAK: fix issue with multi-threaded reads

tbonfort:
Concurrently decoding from multiple jp2 files results in an error being raised:

 

>  ERROR 1: Kakadu Core Error:
>  Multi-threaded implementation error detected.  Before passing a `kdu_thread_env' reference into any 
>  of the `kdu_codestream' machinery's interface functions you need to pass a `kdu_thread_env' 
>  reference into one of the top-level interface functions that configures the codestream for multi- 
>  threaded processing.  The main functions of this form are `kdu_codestream::create' and 
>  `kdu_codestream::open_tile', although there are others which can see a `kdu_thread_env' 
>  environment for the first time.

I have not digged deep into the issue, but the proposed patch fixes that error.


rouault:

PR #1952, opened 2019-10-24
frmts/gif: Including <io.h> to fix MinGW-w64 builds

jerstlouis:


mloskot:
By the way, these files in the upstream include `io.h`, but this way

```c
#ifdef _WIN32
#include <io.h>
#else
#include <unistd.h>
#endif /* _WIN32 */
```

jerstlouis:
@mloskot yes I saw that. I was trying to go for a minimal change here, until giflib is updated.
the #ifdef is already inside the #else, so unistd is not required.
And there is a risk that perhaps _WIN32 is not defined in this version.

rouault:
_WIN32 should be defined on MSVC and mingw builds, so I think we can try the upstream approach. That'll make a future resync easier

rouault:
@jerstlouis Willing to rework your PR along the lines given by @mloskot in https://github.com/OSGeo/gdal/pull/1952#issuecomment-546010145 ?

jerstlouis:
@rouault Looking at https://github.com/rcancro/giflib/blob/master/lib/dgif_lib.c and https://github.com/rcancro/giflib/blob/master/lib/egif_lib.c it is actually currently included differently from what @mloskot pasted above.
The whole set and order of headers included and the set of preprocessor definitions is completely different between the old giflib and the new giflib versions, in both source files, which may have different dependencies in the actual source as well for those versions.

This was a specific minimal patch to solve a long standing failure to build for MinGW, it will not impact MSVC which will go in the first #if as usual, so I feel it really is best to keep the #  if defined(__MINGW32__) there as I had it in the PR to minimize impact.
Whatever the latest include layout ends up to be can be picked up when giflib is updated to a newer version.

But please feel free to rewrite it however you like, as long as the problem is solved (io.h not getting included on MinGW-w64 and causing build failures). Thank you!

mloskot:
@jerstlouis FYI, I was referring to content of the upstream repository which, AFAIU, is here

https://sourceforge.net/p/giflib/code/ci/553661e9c8135addb8018782789b99bd90fb59f5/tree/egif_lib.c
https://sourceforge.net/p/giflib/code/ci/553661e9c8135addb8018782789b99bd90fb59f5/tree/dgif_lib.c

The fork/mirror you linked seems quite dated, last commit in 2015.

jerstlouis:
@mloskot ah sorry, thank you for linking the right ones.

rouault:
@jerstlouis Can you amend this PR to reflect the solution used by upstream as pointed by @mloskot  in https://github.com/OSGeo/gdal/pull/1952#issuecomment-547291907 ?

jerstlouis:
@rouault as I pointed out before, upstream changed a whole bunch of things, which may break by just bringing this small change. I had provided was a minimum impact fix to work with the current state and not break anything. If we look at upstream, might as well update the whole giflib. Please feel free to amend the commit however you want. I was just hoping to fix failure to build from source on MinGW-w64 which was causing headaches.

rouault:
>  as I pointed out before, upstream changed a whole bunch of things, which may break by just bringing this small change.

Well, just try to integrate the part related to including io.h vs  unistd.h. We have plenty of CI configurations to detect obvious blunders

> Please feel free to amend the commit however you want

That's your issue & responsibility, not mine ;-) Remember I'm a lazy maintainer who drinks cocktails at swimming pool, looking at others doing the actual work

jerstlouis:
> That's your issue & responsibility, not mine ;-) Remember I'm a lazy maintainer who drinks cocktails at swimming pool, looking at others doing the actual work

Yeah I don't think you can pretend not working hard ;) Eh, sorry, I'm just overwhelmed with tons of projects and barely sleep right now (but unfortunately that's pretty much always the case). I don't have an environment or GDAL set up at the moment. This was more intended as flagging a critical bug than a 'contribution'. ;) But I will try to get around to it when I have a chance. Just didn't want this be abandoned and forgotten.

PR #1963, opened 2019-10-29
Bindings: OGR: SRS: add C-wrappers and SWIG bindings for `{Set,Get}Extension` functions

and-marsh:
### Summary
Add C-wrappers and SWIG bindings for `{Set,Get}Extension` functions.

I didn't found C-wrappers for these functions and decided to create them by my own. I hope it will be helpful for users. 

I'd like to get your review, please, tell me if there is something to improve in this PR.

### DIFF
--- a/gdal/ogr/ogr_srs_api.h
+++ b/gdal/ogr/ogr_srs_api.h
@@ -569,6 +569,16 @@ OGRErr CPL_DLL OSRSetTOWGS84( OGRSpatialReferenceH hSRS,
                               double, double, double, double );
 OGRErr CPL_DLL OSRGetTOWGS84( OGRSpatialReferenceH hSRS, double *, int );
 
+const char * CPL_DLL OSRGetExtension( OGRSpatialReferenceH hSRS,
+                                      const char *pszTargetKey,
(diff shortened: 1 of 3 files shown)


rouault:
Why do you need that ? I'd perfer not encouraging those non-standard EXTENSION nodes. They should be hopefully no longer, or very rarely used, with GDAL 3 / WKT 2

and-marsh:
Hmm, I need to be able to use vertical geoids during transformation from `EPSG:4979` to any other CRS. So i wanted to add geoid file as extension, and use `ExportToProjString()` that returns proj string with `+geoidgrids=[geoidFileName]` to force using any available geoid. Otherwise most of geoids don't work if source CRS is `EPSG:4979`.

But if you suppose, that is a bad idea, i will close PR and find another way

rouault:
I'm currently looking at improvements on PROJ side regarding geoid grids. Not sure how it will translate to GDAL OGRSpatialReference. But in the meantime, you could propably just export the CRS as projstring, manually add +geoidgrids, and reimport that.

and-marsh:
Ok, thank you

PR #1968, opened 2019-10-30
OAPIF: Add usage examples

jratike80:


gdal/doc/source/drivers/vector/oapif.rst
```diff
@@ -139,7 +144,23 @@ Examples
         MULTIPOLYGON (((8.71191 51.491084,8.7123 51.491067,8.712385 51.491645,8.712014 51.491666,8.711993 51.491603,8.71196 51.491396,8.711953 51.491352,8.71191 51.491084)))
 
       [...]
+      
+-  How to include a query string that is supported by the API into the connection URL
```
rouault: Hum, no. That will not work as you want. If you look closely at the consequence of ?flur=008, then it is added to *all* queries done by the driver. So also for irrelevant endpoints. In that case, /collections or /api don't error out, but they could/should probably.
Worse here is that this gets added twice to paging request, which the server doesn't like at all, and prevent paging from working
```
HTTP: Fetch(https://www.ldproxy.nrw.de/kataster/collections/flurstueck/items?flur=008&f=json&offset=10&limit=10&flur=008)
HTTP: These HTTP headers were set: Accept: application/geo+json, application/json
ERROR 1: Empty content returned by server
```
It could be useful to have a way to pass a query string, but we should have a dedicated open option for it so that it only gets inserted in the initial /items request.

jratike80: I was trying to document this https://github.com/OSGeo/gdal/commit/aad2cbccb4708fb2804bfa99b649d01c6de55706 but I noticed the same issues. I would say that GDAL adds the query parameter just once into the "next page" link, but it has already one as a heritage. I will make a bug report about the server behavior, it returns zero length result if there are duplicate key/value pairs like &flur=008&flur=008 in this case or &limit=10&limit=1000 in my other trial.

Perhaps it is better to forget and delete this PR but I do not know how to do it.

rouault: > I was trying to document this [aad2cbc](https://github.com/OSGeo/gdal/commit/aad2cbccb4708fb2804bfa99b649d01c6de55706)

This commit is not intended (in my mind at least) for passing a user query, but more an authentication parameter that is valid for all links on the server.

> I will make a bug report about the server behavior

I'm not completely sure the server misbehave here. One should check what Internet RFCs + OAPIF spec say about such situation. Might be unspecified bheaviour

>  to forget and delete this PR but I do not know how to do it.
Just press the "Close" button :-)

jratike80: It was named "correctly handle user query string parameters"...

I believe that duplicate KVP is an undefined situation but that happens with WMS servers as well and they either take one one skip another, or give a more or less good error message. Example

https://demo.geo-solutions.it/geoserver/wfs?service=WFS&version=2.0.0&request=GetFeature&typenames=states&count=1&count=2

PR #1978, opened 2019-11-03
FlatGeoBuf: update flatbuffers include files from latest upstream. Should fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=17647

rouault:


rouault:
The last 2 commits have been submitted to upstream flatbuffers: https://github.com/google/flatbuffers/pull/5607 and https://github.com/google/flatbuffers/pull/5610

PR #1979, opened 2019-11-04
netCDF: do not automatically report a nodata value if there's none in the product

rouault:


rouault:
Replaced by https://github.com/OSGeo/gdal/pull/1983 which follows a less radical approach as suggested by 	Julien Demaria in https://lists.osgeo.org/pipermail/gdal-dev/2019-November/051062.html

PR #1984, opened 2019-11-05
Add text for RFC76 OGR Python drivers

rouault:


gdal/doc/source/development/rfc/rfc76_ogrpythondrivers.rst
```diff
@@ -0,0 +1,394 @@
...
+
+Motivation
+----------
+
+For some use cases that do not require lighting speed, or to deal with very
+niche formats (possibly in house format), it might be faster and more efficient
+to write a vector driver in Python rather than a GDAL C++ driver as currently required,
+or an ad-hoc converter.
+
+This could benefit to QGIS which currently has no way to use providers in
```
elpaso: Actually, you can create QGIS vector data provider in python: https://github.com/qgis/QGIS/blob/master/tests/src/python/test_provider_python.py

rouault: good to know ! This probably didn't exist at the time I started this experiment, or I just didn't know :-)

nyalldawson:
This seems like a great way to quickly prototype new drivers which could eventually be ported to core drivers. Love it!

gdal/doc/source/development/rfc/rfc76_ogrpythondrivers.rst
```diff
@@ -0,0 +1,397 @@
...
+  the supported API version number. This item enables us to check if we are able
+  to "safely" load a Python driver. If a Python driver would support several API
+  versions (not clear if that's really possible at that point), it might use an
+  array syntax to indicate that, like ``[1,2]``
+- ``DRIVER_DCAP_VECTOR = "YES"``
+- ``DRIVER_DMD_LONGNAME = "my super plugin"``
+
+Optional metadata such as ``DRIVER_DMD_EXTENSIONS`` or
+``DRIVER_DMD_HELPTOPIC`` can be defined (basically, any driver metadata key string prefixed by
+``DRIVER_``
```
mloskot: It may be (better clearer to reader and script maintainer, unambiguous for detection and loading) if those have `GDAL_` prefix.

gdal/doc/source/development/rfc/rfc76_ogrpythondrivers.rst
```diff
@@ -0,0 +1,397 @@
...
+interpreter, both for efficiency consideration and also because we want to
+delay the research or launch of the Python interpreter as much as possible
+(the typical use case if GDAL used by QGIS: we want to make sure that QGIS
+has itself started Python, to reuse that Python interpreter)
+
+From the short metadata, the driver registration code can instantiate GDALDriver
+C++ objects. When the Identify() or Open() method is invoked on that object,
+the C++ code will:
+
+* start/find the Python interpreter if not already done
```
mloskot: What does the '"interpreter" mean?
A `PyThreadState` or `Py_NewInterpreter`-made sub-interpreter? 
An installed Python interpreter external to the Python built into GDAL?

IMO, this is a bit unclear.

gdal/doc/source/development/rfc/rfc76_ogrpythondrivers.rst
```diff
@@ -0,0 +1,397 @@
...
+likewise for the ``Layer`` object. See the example_.
+
+Connection with the Python interpreter
+++++++++++++++++++++++++++++++++++++++
+
+The logic will be shared with the VRT Python fix functionality. It relies on
+runtime loading of the Python library, rather than compile time linking. The
+reason is that we do not know in advance with which Python version GDAL can
+potentially be linked, and we do not want gdal.so/gdal.dll to be explicitly linked
+with a particular Python library.
```
mloskot: This explains my earlier question in https://github.com/OSGeo/gdal/pull/1984/files#r342839730 (a footnote missing)

gdal/doc/source/development/rfc/rfc76_ogrpythondrivers.rst
```diff
@@ -0,0 +1,397 @@
...
+    def create_driver():
+        return Driver()
+
+
+Limitations and scope
+---------------------
+
+- Vector and read-only for now. This could later be extended of course.
+
+- No connection between the Python code of the plugin and the OGR Python API
```
mloskot: Simply, the new one is "GDAL in Python"  and the existing one is "Python in GDAL" :-)

gdal/doc/source/development/rfc/rfc76_ogrpythondrivers.rst
```diff
@@ -0,0 +1,397 @@
...
+---------------------
+
+* Currently that code is unconditionnally compiled in, similarly to the GDAL
+  VRT Python plugin function. Is there a need to have a compile-time switch
+  to disable this ?
+
+* Should we use __iter__() / __next__() methods in the Layer object to iterate
+  over features ? This is the canonical way in Python to define an iterator.
+  However, this code is to be called mainly by the GDAL C++ code, so Python'icity
+  is not critical. Another subtlety is that OGR layers can have spatial and
```
mloskot: I can be someone will complain why the API is not Pythonic :-)

gdal/doc/source/development/rfc/rfc76_ogrpythondrivers.rst
sgillies: @mloskot or in other words, extending Python vs embedding Python: https://docs.python.org/3/extending/. I don't think it's a great idea for GDAL to promote both, but maybe that's just where GDAL is in the software universe, and the Python docs do cover the topic of extending embedded Python: https://docs.python.org/3/extending/embedding.html#extending-embedded-python.

mloskot: Good point. Having written all three (extend, embed, extend embedded incl. extending with plain Python modules) and I don't think this proposal falls that into canonical embedded (and extend embedded) with Python.

gdal/doc/source/development/rfc/rfc76_ogrpythondrivers.rst
```diff
@@ -0,0 +1,397 @@
...
+        # Required
+        def open(self, filename, first_bytes, open_flags):
+            if not self.identify(filename, first_bytes, open_flags):
+                return None
+            return Dataset(filename)
+
+
+    # Required
+    def create_driver():
+        return Driver()
```
mloskot: Why this `create_driver` is required; what's the purpose?
Why not to use the `__init__` magic?

Since user's module is regularly imported
 https://github.com/rouault/gdal/blob/1bf795f25458dfb7e2a88e5d99b0e3647b82f2d4/gdal/gcore/gdalpythondriverloader.cpp#L2212-L2214

then why not look for classes that are `PyType_IsSubtype` of `Driver` class and construct calling the regular Python `__init__`?

```cpp
PyObject* poTypeDriver = ... ; // class item found with PyDict_Next for callable PyType_Check
                               // that is PyType_IsSubtype of Driver
PyTypeObject* poDriver = reinterpret_cast<PyTypeObject*>(poTypeDriver);
if (poDriver->tp_new(obj.get(), args, kwds) < 0) { ... }
if (poDriver->tp_init(obj.get(), args, kwds) < 0) { ... }

PyObject* poMethod = PyObject_GetAttrString(poDriver, "open" );
...

```

gdal/doc/source/development/rfc/rfc76_ogrpythondrivers.rst
rouault: > extending Python vs embedding Python:

Discovering the taxonomy of those Python techniques :-) As far as I understand the docs, the techniques I use here are embedding Python (to load the user .py script, and the empty definitions of the OGRLayer, GDALDataset and GDALDriver classes) and extending python to define a _gdal_base module (internal detail) that has a layer_featureCount() method that is called by OGRLayer.feature_count() default implementation

> One can imagine scenarios where someone tries to use a Python driver in GDAL (embedded Python) and the Python driver uses the GDAL bindings (extending Python) and a VRT with Python functions (embedded Python again), and so on. The possibilities for bugs are endless.

I just tested that in https://gist.github.com/rouault/1dcc5a13583ba2ab419cf4a8f05b1ccc. It works fine from ogrinfo command line. Works fine also from a Python console doing "from osgeo import ogr; ds = ogr.Open('MYPLUGIN:')"  (so Python (console) calling C++ calling Python (OGR Python driver) calling C++ calling Python (embedded Python in VRT))  . I agree there might be some fragility though. Hence the limited scope to experimentation for now to see how it flyes with users.

> Or, it is to allow import x where x is not necessarily installed and available for embedded Python...

Yes, the Python drivers need to import gdal_base that has the base objects and some hidden magic to interact with the native code. This makes deployment easier rather than having a real .py + .so small module extending Python (like the class SWIG bindings do)

mloskot: @rouault 
> As far as I understand the docs, the techniques I use here are embedding Python (to load the user .py script, and the empty definitions of the OGRLayer, GDALDataset and GDALDriver classes) and extending python to define a _gdal_base module (internal detail)  ...

In my opinion, the user-defined subclass of `Driver`, `OGRLayer` pre-defined classes are also an extension to the embedded Python.

The way you are proposing it is slightly more complex than canonical way of embedding Python:
it is not linking  Python library at build time in order to embed the Python interpreter and API, but actually embedding Python at run-time. I'd call it an extremely delayed or lazy embedding :)

I would not be surprised if some readers of the RFC and users (especially not experienced w/ embedding) of the machinery are a bit confused asking on the list how it works, because it is not a typical way of embedding.

gdal/doc/source/development/rfc/rfc76_ogrpythondrivers.rst
rouault: sounds like a good idea. Adding a bit of implementation complexity though. Do you have an idea how to enumerate the classes of a module ? Can't see anything better than PyModule_GetDict() ( https://docs.python.org/3/c-api/module.html#c.PyModule_GetDict ) although they suggest using other methods, but not clear to me which one(s) would be appropriate

mloskot: I usually do something along this lines (read the snippet as pseudo-code, not compile/run-tested, written from brain memory):

```cpp
PyObject *pKey(nullptr), *pValue(nullptr);
Py_ssize_t pos(0); // must be initialized to 0 prior call to PyDict_Next! 

PyObject* pModule = ...; // module with Driver, OGRLayer, etc.
PyObject* pDriverClass = PyObject_GetAttrString(pModule, "Driver");
PyTypeObject* pDriverType = reinterpret_cast<PyTypeObject*>(pDriverClass);

PyObject* pDict = PyModule_GetDict(pPluginModule, ...);
while (PyDict_Next(pDict, &pos, &pKey, &pValue))
{
    PyObject* pDictAttr = PyObject_GetAttr(pDict, pKey)
    if (PyType_Check(pDictAttr))
    {
        PyTypeObject* pType = reinterpret_cast<PyTypeObject*>(pDictAttr);
        if (PyType_IsSubtype(pDriverType, pType))
        {
            // User-defined Python class is viable candidate for Driver class
            // ... allocate/instantiate/call it...
        }
    }
}
```

> they suggest using other methods, but not clear to me which one(s) would be appropriate

I think, for `PyObject_*` they mean functions like `PyObject_GetAttrString`, but for `PyModule_*`, not sure.

rouault: ok, I've managed to no longer require a create_driver() method. Using Python C API wasn't really possible as most of the PyType_xxxx methods are actually macros, which require having knowledge of Python structures, which isn't super ideal given that I don't include Python.h to be as much version independent as possible... So I went through the road of executing a Python snippet instead: https://github.com/rouault/gdal/commit/884ced57d8e720793e9f65f9eb059276d5db1c32#diff-42d4e6b0795db7beb7d99c0b9e51b18fR1190

mloskot: >  PyType_xxxx methods are actually macros, which require having knowledge of Python structures (...) I don't include Python.h ...

Sorry, I ignored that fact indeed.

gdal/doc/source/development/rfc/rfc76_ogrpythondrivers.rst
hobu: > The possibilities for bugs are endless.

PDAL both extends and embeds Python. This statement is a fact. It is also really useful to do both things sometimes.

gdal/doc/source/development/rfc/rfc76_ogrpythondrivers.rst
rouault: taken into account

gdal/doc/source/development/rfc/rfc76_ogrpythondrivers.rst
rouault: few words added about that

gdal/doc/source/development/rfc/rfc76_ogrpythondrivers.rst
rouault: I opted for a Python iterator finally

gdal/doc/source/development/rfc/rfc76_ogrpythondrivers.rst
rouault: I've tried to make that a bit clearer

gdal/doc/source/development/rfc/rfc76_ogrpythondrivers.rst
```diff
@@ -0,0 +1,472 @@
...
+dedicated directory:
+
+* the directory pointed by the ``GDAL_PYTHON_DRIVER_PATH`` configuration option
+  (there may be several paths separated by `:` on Unix or `;` on Windows)
+* if not defined, the directory pointed by the ``GDAL_DRIVER_PATH`` configuration
+  option.
+* if not defined, in the directory (hardcoded at compilation time on Unix builds)
+  where native plugins are located.
+
+Those Python script must have in their first lines at least 2 global variables
```
sgillies:
```suggestion
Those Python script must set in their first lines at least 2 directives
```

gdal/doc/source/development/rfc/rfc76_ogrpythondrivers.rst
```diff
@@ -0,0 +1,472 @@
...
+set:
```
sgillies:
```suggestion
```

gdal/doc/source/development/rfc/rfc76_ogrpythondrivers.rst
```diff
@@ -0,0 +1,472 @@
...
+  versions (not clear if that's really possible at that point), it might use an
+  array syntax to indicate that, like ``[1,2]``
+- ``# gdal: DRIVER_DCAP_VECTOR = "YES"``
+- ``# gdal: DRIVER_DMD_LONGNAME = "my super plugin"``
+
+Optional metadata such as ``# gdal: DRIVER_DMD_EXTENSIONS`` or
+``# gdal: DRIVER_DMD_HELPTOPIC`` can be defined (basically, any driver metadata key string prefixed by
+``# gdal: DRIVER_``
+
+Those will be parsed in a pure textual way, without invocation of the Python
```
sgillies:
```suggestion
These directives will be parsed in a pure textual way, without invocation of the Python
```

gdal/doc/source/development/rfc/rfc76_ogrpythondrivers.rst
```diff
@@ -0,0 +1,472 @@
...
+driver.
+
+.. _example:
+
+Example of such a driver
+------------------------
+
+Note that the prefixing by the driver name in the connection string is absolutely
+not a requirement, but something specific to this particular driver which is a
+bit artifical. The CityJSON driver mentionned below does not need it.
```
sgillies:
```suggestion
bit artifical. The CityJSON driver mentioned below does not need it.
```

gdal/doc/source/development/rfc/rfc76_ogrpythondrivers.rst
```diff
@@ -0,0 +1,472 @@
...
+    # aware code. So just literal values, no expressions, etc.
+    # gdal: DRIVER_NAME = "DUMMY"
+    # API version(s) supported. Must include 1 currently
+    # gdal: DRIVER_SUPPORTED_API_VERSION = [1]
+    # gdal: DRIVER_DCAP_VECTOR = "YES"
+    # gdal: DRIVER_DMD_LONGNAME = "my super plugin"
+
+    # Optional driver metadata items.
+    # # gdal: DRIVER_DMD_EXTENSIONS = "ext1 est2"
+    # # gdal: DRIVER_DMD_HELPTOPIC = "http://example.com/my_help.html"
```
sgillies: Recommend putting these directives above the Python import statement.

rouault: done

rcoup:
Overall, I like it! ❤️ 

In terms of making it more Pythonic, wondering about things like:

* `len(dataset)` => ie:  `Dataset.layer_count()` → `Dataset.__len__()`
* `dataset[name_or_idx]` => ie: `Dataset.layer(idx)` → `Dataset.__getitem__(name_or_idx)`
* `iter(dataset)`
* `len(layer)` => ie. `Layer.feature_count()` → `Layer.__len__()`
* `layer[feature_id]` => ie: `Layer.feature_by_id(id)` → `Layer.__getitem__(id)`

And does a Dataset need a close() method? 
Ideally I think it'd be used as a context manager, and `Dataset.__exit__()` would call `.close()`
```
with driver.open("foo.layer") as dataset:
   ...
# dataset.close() is called when the context exits.
```

What are your thoughts @sgillies @mloskot ?

sgillies:
@rouault my last comment: do we have any way of globally disabling plugins, say by an environment variable? For a developer who has broken plugins in their system or conflicts between plugins, it could be a nice option.

sgillies:
@rcoup I honestly hope people don't write many of these plugins because there isn't great support for testing, deploying, or debugging them, so I'm ambivalent about making their python API really nice.

rouault:
>  do we have any way of globally disabling plugins, say by an environment variable?

Setting GDAL_DRIVER_PATH=dummy (or any non existing path) for native drivers and GDAL_PYTHON_DRIVER_PATH=dummy for Python drivers would do

rouault:
> In terms of making it more Pythonic, wondering about things like

Hum, there's a bit of tension about making them Pythonic and having a clear enough correspondance with the C++ method names.

> And does a Dataset need a close() method?

As those drivers are read-only currently, we probably don't need one now. But we could certainly add one when needed which would be called by the C++ destructor of the corresponding GDALDataset object.

rcoup:
> Hum, there's a bit of tension about making them Pythonic and having a clear enough correspondance with the C++ method names.

Fair enough (& Sean’s points too)

> As those drivers are read-only currently, we probably don't need one now.

No DB connections or source files or whatever else? I guess `__del__` is called eventually, with the usual caveats

rouault:
> No DB connections or source files or whatever else? I guess `__del__` is called eventually, with the usual caveats

fair point. I'll add an optional close() method

Issue #1966, opened 2019-10-29
FlatGeoBuf driver loads the whole feature offset index / inefficient for network access


rouault:
CC @bjornharrtell 
See https://lists.osgeo.org/pipermail/gdal-dev/2019-October/050946.html

mloskot:
Seems like this #1964 added recently by @bjornharrtell is related

rouault:
> Seems like this #1964 added recently by @bjornharrtell is related

Not really :-) #1964 is about writing side, this one about reading side

bjornharrtell:
Ah yes, I've been semi-aware of this shortcoming.

bjornharrtell:
I want to take a stab at this but unsure about the best strategy. I would try to do an adaptive block cache for the I/O here but unsure if I should simply try to implement that directly in the driver or if I should attempt to implement at a higher abstraction level. Do you have any guidance @rouault?

rouault:
I would just suggest implementing the most simple strategy, that is issuing a VSIFRead() to get the offset of just the feature you need at the time where you need it. For local I/O, the operating system will cache things. And for /vsicurl/ and related, they will also have some caching.

bjornharrtell:
Mmm yes, will be very simple and probably work well for most cases. 👍

bjornharrtell:
@rouault it seems accessing via vsizip is performing very badly on large files (uncompressed >1GB) and I suspect may not have been so bad before this PR. I've looked through cpl_vsil_gzip but fail to find any limit to the snapshot cache which would explain it... will investigate further when time permits but mabye you have an idea what I should look for? (accessing comparable shapefile data via vsizip performs well)

rouault:
ah, it might be that the lots of seeks that you need to do might hit hard the snapshot mechanism. but random access in a zip file is going to suck anyway...

bjornharrtell:
Well, a spat filter query via vsizip shapefile actually performs rather well, even over vsicurl. (!)

bjornharrtell:
My theory is that because shp is separated into several files you get several caches and less seeking per file. But the cache could/should be smarter than that. Will try to understand it and improve when I can. :)

bjornharrtell:
But I slowly realize that deflating a single stream requires sequential read unless some kind of extra/custom indexing is done like fx. https://github.com/pauldmccarthy/indexed_gzip#overview.

bjornharrtell:
[...]

PR #1993, opened 2019-11-07
FlatGeobuf: Optimize reads for index search

bjornharrtell:
A follow up optimization after conclusions in https://github.com/OSGeo/gdal/issues/1966.

Tested locally to significantly improve read performance when reading through vsigzip.

### DIFF
--- a/gdal/ogr/ogrsf_frmts/flatgeobuf/ogr_flatgeobuf.h
+++ b/gdal/ogr/ogrsf_frmts/flatgeobuf/ogr_flatgeobuf.h
@@ -50,6 +50,11 @@ struct FeatureItem : Item {
     uint64_t offset;
 };
 
+struct IndexOffset {
+    size_t index;
+    uint64_t offset;
+};
+
 struct GeometryContext {
     std::vector<double> xy;
     std::vector<double> z;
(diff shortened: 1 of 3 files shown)


gdal/ogr/ogrsf_frmts/flatgeobuf/ogr_flatgeobuf.h
```diff
@@ -86,7 +86,7 @@ class OGRFlatGeobufLayer final : public OGRLayer
         uint64_t m_offset = 0;
         uint64_t m_offsetFeatures = 0;
         uint64_t m_offsetIndices = 0;
-        std::vector<uint64_t> m_foundFeatureIndices;
+        std::vector<std::pair<uint64_t, uint64_t>> m_foundFeatureIndexOffsets;
```
rouault: a few comments about the meaning of those variables wouldn't hurt. As far as I understand m_featuresPos goes from 0 to m_featuresSize - 1, and m_featuresSize is the number of features that match the spatial index ? and the (usable) size of m_foundFeatureIndexOffsets is m_featuresSize ?

rouault: it might perhaps help to create a struct IndexOffset { size_t index; uint64_t fileOffset } and use that instead of a std::pair

rouault: shouldn't m_featuresPos and m_featuresSize be of size_t, since they represent a number of things that must hold in RAM right ? That could probably save a few static_cast<size_t>

bjornharrtell: Agreed, I will try to make some meaningful comments and rework the pair. However, m_featuresSize is actually some left over that is not used - I assume you mean m_featuresCount?

rouault: > I assume you mean m_featuresCount?

indeed !

rouault:
Perhaps fixed by your latest commits, but there was a autotest failure reported earlier at least
```
FAILED ogr/ogr_flatgeobuf.py::test_ogr_flatgeobuf_2 - assert 1 == 4
```

bjornharrtell:
All green locally so I'm :crossed_fingers:

bjornharrtell:
printf size_t argument using %zu not allowed :cry:

rouault:
On 32 bit:
```
ogrflatgeobuflayer.cpp:527:68: error: non-constant-expression cannot be narrowed from type 'unsigned long long' to 'size_t' (aka 'unsigned int') in initializer list [-Wc++11-narrowing]
                m_foundFeatureIndexOffsets.push_back(IndexOffset { i, featureOffset });
                                                                   ^
```

bjornharrtell:
Don't understand (yet) - i is size_t and index in IndexOffset is also size_t?

bjornharrtell:
And "FAILED ogr/ogr_flatgeobuf.py::test_ogr_flatgeobuf_2 - assert 1 == 4" remains...

rouault:
> Don't understand (yet) - i is size_t and index in IndexOffset is also size_t?

line 523: for (auto i : foundFeatureIndices) {
line 517: auto foundFeatureIndices = PackedRTree::streamSearch(...)
and streamSearch() returns a std::vector<uint64_t>, hence the uint64_t for i

bjornharrtell:
Ah.. mabye I was silly to try to support larger amounts of features than 4294967295.

rouault:
> I was silly to try to support larger amounts of features than 4294967295.

well, if use size_t to count for whatever can be allocated, then you should be fine.

bjornharrtell:
Rebased on master. Still unable to reproduce the test failure locally.

bjornharrtell:
Made a VM with Ubuntu 14.04 to set things up similar to the CI, but still cannot reproduce the test failure. :(

bjornharrtell:
Not sure why I couldn't reproduce locally but it seems https://github.com/OSGeo/gdal/pull/1993/commits/a19d33409ff07745c29ad354655075da625a5790 resolved it and with https://github.com/OSGeo/gdal/pull/1993/commits/6e73ce26103c2452d9c66ad69d8458598661833d it's all green and ready to merge @rouault.

PR #1997, opened 2019-11-08
OAPIF: schema related improvements and cql-text support

rouault:


jratike80:
Will there be an option to skip schema resolving if schema does not match the data or if the schema document can't be found through the href link?

rouault:
> Will there be an option to skip schema resolving if schema does not match the data or if the schema document can't be found through the href link?

just added

PR #2000, opened 2019-11-09
Fail Azure jobs on any command error 

### LINKED ISSUE #1998: Azure pipelines can push incomplete doc generation
As seen in #1996, in case the download of the Java or Python doc from download.osgeo.org fails, this doesn't stop the doc generation, so we can push an incomplete doc structure on the server. It would be good if we could make the pipeline fail if one of the step fails.

CC @mloskot

mloskot:
## What does this PR do?

Fixes #1998

## Tasklist

### DIFF
--- a/gdal/doc/.azure-pipelines.yml
+++ b/gdal/doc/.azure-pipelines.yml
@@ -25,18 +25,22 @@ jobs:
       options: --privileged
     steps:
       - script: |
+          set -e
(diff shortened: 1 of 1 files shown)


mloskot:
No need to port it to `release/3.0` or earlier as none of those have `.azure-pipelines.yml` configured.

PR #2018, opened 2019-11-14
VRT: fix requesting a downsampling version of the mask band of a source that has masks and overviews

rouault:


rouault:

Issue #2020, opened 2019-11-15
Wrong cache split calculation in gdal2tiles.py


EvertEt:
## Expected behavior and actual behavior.

```    python
# Make sure that all processes do not consume more than GDAL_CACHEMAX
os.environ['GDAL_CACHEMAX'] = '%d' % int(gdal.GetCacheMax() / nb_processes)
```
`GDAL_CACHEMAX` expects a value in MB but `gdal.GetCacheMax()` returns the value in bytes.
https://trac.osgeo.org/gdal/wiki/ConfigOptions#SettingConfigOptions
https://gdal.org/api/raster_c_api.html?highlight=getcachemax#_CPPv415GDALGetCacheMaxv

This is also confirmed by a log in `gdal2tiles.py`:
```       python
print("Cache: %s MB" % (gdal.GetCacheMax() / 1024 / 1024))
```

## Steps to reproduce the problem.

See behaviour. The RAM fills up, where I assume it is because of this mismatch.

## Operating system

Ubuntu 18.04 64 bit

## GDAL version and provenance

The 2.4.0 version from ubuntugis-stable PPA

rouault:
Good analysis. Can you submit a pull request to fix that ? (I'd suggest to do a max of 1 and the result of the division, to avoid a 0 cache size)

rouault:
Fixed per #2022. Backported to branches 3.0 and 2.4

PR #2022, opened 2019-11-15
Update cache calculation in gdal2tiles.py

EvertEt:
See #2020

## What does this PR do?
 
Fix bytes/MB mismatch in cache calculation

#2020 
## Tasklist

 - [ ] Add test case(s)
 - [ ] Review
 - [ ] Adjust for comments


EvertEt:
@rouault I couldn't find any test cases for this, correct?

rouault:
> I couldn't find any test cases for this, correct?

yes, that would be hard to test

PR #2025, opened 2019-11-15
Docs: fix typo in virtual_file_systems.rst

elil:
This is a typo as well but I don't know the intended meaning so can't fix it.

"/vsicrypt/ is a special file handler **is** installed that allows reading..."
could mean:
"/vsicrypt/ is a special file handler **if** installed that allows reading..."
which is odd phrasing.  If that is what is meant, could consider:
"If installed, /vsicrypt/ is a special file handler that allows reading..."

Doing nothing works fine too since I think that the information is conveyed as is.

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:


mloskot:
Thanks

PR #2031, opened 2019-11-19
Remove example section

JakobMiksch:
the example section actually does not contain any examples

### DIFF
--- a/gdal/doc/source/drivers/vector/filegdb.rst
+++ b/gdal/doc/source/drivers/vector/filegdb.rst
@@ -152,12 +152,6 @@ Layer Creation Options
    : (GDAL >= 2.0) Customize how data is stored. By default text in
    UTF-8 and data up to 1TB
 
-Examples
---------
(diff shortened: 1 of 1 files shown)


mloskot:
I think this needs to be checked against the old HTML documentation. Perhaps the actual example snippets have been forgotten e.g. during two step rewriting.

mloskot:
Indeed, the examples have not been ported:

https://github.com/OSGeo/gdal/blob/62c58bed1635e1b1cb0edcfe11e30b8786b5edc8/gdal/ogr/ogrsf_frmts/filegdb/drv_filegdb.html#L95-L104

JakobMiksch:
yes, then it probably better to port the examples to the new version

mloskot:
@JakobMiksch Yes and, as usual, contributions are highly appreciated.

JakobMiksch:
@mloskot agree. I will have a look on it!

PR #2054, opened 2019-11-26
DOCS: Updated broken documentation links within src files

codyben:
## What does this PR do?
After looking through the docs I noticed there were quite a few broken links, so I went ahead and updated them. An example can be found here: https://gdal.org/api/gdal_utils.html#gdal__utils_8h_1a1cf5b30de14ccaf847ae7a77bb546b28
It seemed as though most of the links were relative, leading to links such as: https://gdal.org/api/gdal_translate.html 
instead of: 
https://gdal.org/programs/gdal_translate.html

In lieu of searching for every broken link in the docs, I grepped through the project to find any potential links (pardon the pipe)
``grep -r "accepted options are the ones" | grep "cpp"``

## Tasklist
 - [X] Checked the links to make sure the new "/project/insert_gdal_utility_here.html" URL didn't 404 
 - [X] Rebuilt the project from source just to make sure I didn't goof anywhere

## Environment

* OS: Ubuntu 18.04 LTS
* Compiler: gcc

### DIFF
--- a/gdal/apps/gdal_grid_lib.cpp
+++ b/gdal/apps/gdal_grid_lib.cpp
@@ -712,7 +712,7 @@ static OGRGeometryCollection* LoadGeometry( const char* pszDS,
 /**
  * Create raster from the scattered data.
  *
- * This is the equivalent of the <a href="gdal_grid.html">gdal_grid</a> utility.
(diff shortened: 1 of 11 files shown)


mloskot:
Thank you very much!

PR #2057, opened 2019-11-26
Make GDALTermProgress streaming-friendly: write to stderr, not stdout.

acg:
Streaming input and output through a shell pipeline saves one from writing large intermediate files to disk. For example:

```sh
gdal_contour -a elev -i 100 -f PGDump elev48i0100a.tif /dev/stdout | psql ned
```

However, this results in a strange error:

```
ERROR:  syntax error at or near "0"
LINE 1: 0INSERT INTO "public"."contour" ("wkb_geometry" , "id", "ele...
```

This happens because the `0...10...20...` progress bar is written by `GDALTermProgress()` to `stdout` rather than `stderr`, the standard place for progress reporting.

Note that `fflush(3)` is now unnecessary because [`stdout` is line-buffered, but `stderr` is unbuffered](https://linux.die.net/man/3/stderr).

### DIFF
--- a/autotest/gdrivers/vrtmask.py
+++ b/autotest/gdrivers/vrtmask.py
@@ -300,7 +300,7 @@ def test_vrtmask_9():
     src_ds = gdal.GetDriverByName('GTiff').Create('tmp/vrtmask_9_src.tif', 10, 10, 4)
     del src_ds
 
-    (_, err) = gdaltest.runexternal_out_and_err(test_cli_utilities.get_gdal_translate_path() + ' tmp/vrtmask_9_src.tif tmp/vrtmask_9_dst.tif -b 1 -b 2 -b 3')
+    (_, err) = gdaltest.runexternal_out_and_err(test_cli_utilities.get_gdal_translate_path() + ' tmp/vrtmask_9_src.tif tmp/vrtmask_9_dst.tif -q -b 1 -b 2 -b 3')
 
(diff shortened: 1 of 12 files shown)


rouault:
I'm a bit hesitant regarding backward compatibility. In case some application would fork a process and monitor stdout to get progress advancement. For your use case, using the quiet flag (-q) would solve your issue

acg:
> I'm a bit hesitant regarding backward compatibility. In case some application would fork a process and monitor stdout to get progress advancement.

Fair, this is a potential breaking change, and the tests even surface that. I took a stab at fixing them in 6b0e92b, but I don't have a full dev env locally.

> For your use case, using the quiet flag (-q) would solve your issue

It would, but [it's not documented](https://linux.die.net/man/1/gdal_contour), and there's also the [Rule of Silence](http://www.catb.org/~esr/writings/taoup/html/ch11s09.html):

> Programs that babble don’t tend to play well with other programs. If your CLI program emits status messages to standard output, then programs that try to interpret that output will be put to the trouble of interpreting or discarding those messages (even if nothing went wrong!)...
> 
> If you want chatty progress messages for debugging purposes, disable them by default with a verbosity switch. Before releasing for production, relegate as many of the normal messages as possible to being displayed only when the verbosity switch is on.
>
>  -- The Art of Unix Programming, "Silence Is Golden"

This is the least I expect of my command line tools, but I leave it up to you whether the correctness is worth the potential for breakage.

PR #2060, opened 2019-11-27
explicitly read and and write rsts in utf-8 to avoid exception on windows

jef-n:
## What does this PR do?

fixes the doc build (python default encoding is cp1252 on (middle european) windows) where reading the rst produces an exception.

none

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS: Windows 10
* Compiler: Visual C++ 2015
* OSGeo4W


rouault:
I suspect this is a Python 3 syntax. The azure pipeline in doc/.azure-pipelines.yml is based on Ubuntu 16.04 and has Py 3 installed, but the doc/Makefile invokes in a generic way python, which isn't the default on it. Maybe we should change doc/Makefile to explictly call python3 ?

PR #2065, opened 2019-11-28
WFS: for consistency with OGR SQL behaviour, translate LIKE/ILIKE into PropertyIsLike matchCase='false'

rouault:


jratike80:
Unfortunately there are WFS servers that do not react to matchCase, but it is their bug.

rouault:
Superseded by https://github.com/OSGeo/gdal/pull/2078

PR #2069, opened 2019-11-30
 ESRIJson: attempt identification of SRS from database entries (fixes #2007)

### LINKED ISSUE #2007: Invalid srid in postgis when importing data by ogr2ogr
I am using postgis 12 with postgis 3.0.0 and gdal 3.0.0 to copy data in it. The command i am using is

`ogr2ogr -f "PostgreSQL" PG:"host=127.0.0.1 dbname=db user=user port=5432 password=1" "file.geojson" -nln mutable`

The issue here is when i check the srid of the output table it is 900914 which is not correct in my case it should be 2223. To fix this i have to manually specify EPSG code by adding '-t_srs EPSG:2223'. This was not needed with previous version of ogr2ogr and postgis 2.5.

rouault:


rouault:
Merged indirectly per https://github.com/OSGeo/gdal/pull/2071

PR #2073, opened 2019-12-01
FlatGeobuf: Serialization cleanup

bjornharrtell:
~~Intends to fix issue 19198 in oss-fuzz.~~ (had proper fix in master already)

Also includes minor serialization cleanup that reduces file size somewhat and prepares for mixed geometry collections.

### DIFF
--- a/gdal/ogr/ogrsf_frmts/flatgeobuf/geometryreader.cpp
+++ b/gdal/ogr/ogrsf_frmts/flatgeobuf/geometryreader.cpp
@@ -231,9 +231,9 @@ OGRErr GeometryReader::readSimpleCurve(OGRSimpleCurve *sc)
 
 OGRPolygon *GeometryReader::readPolygon()
 {
-    const auto pEnds = m_geometry->ends();
+    const auto ends = m_geometry->ends();
     const auto p = new OGRPolygon();
-    if (pEnds == nullptr || pEnds->size() < 2) {
+    if (ends == nullptr || ends->size() < 2) {
         m_length = m_length / 2;
         const auto lr = readSimpleCurve<OGRLinearRing>();
         if (lr == nullptr) {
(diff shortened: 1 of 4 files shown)


gdal/ogr/ogrsf_frmts/flatgeobuf/geometryreader.cpp
```diff
@@ -289,7 +288,9 @@ OGRGeometryCollection *GeometryReader::readGeometryCollection()
     if (parts == nullptr)
         return CPLErrorInvalidPointer("parts data");
     auto gc = new OGRGeometryCollection();
-    for (uoffset_t i = 0; i < parts->Length(); i++) {
+    if (parts == nullptr)
```
rouault: this test has been done just above

mloskot: Perhaps it was meant to check the gc. By the way, the gc could be captured with unique_ptr for exception safety

rouault: > Perhaps it was meant to check the gc

as gc results from a new, the only failure mode would be a C++ exception, not nullptr, right ?

mloskot: Yes, correct, but there is chance author's intention was to check that anyway.

bjornharrtell: Ah no it's a mistake when I merged locally.

gdal/ogr/ogrsf_frmts/flatgeobuf/geometryreader.cpp
```diff
@@ -409,9 +408,9 @@ OGRPolyhedralSurface *GeometryReader::readPolyhedralSurface()
 
 OGRTriangulatedSurface *GeometryReader::readTIN()
 {
-    const auto pEnds = m_geometry->ends();
+    const auto ends = m_geometry->ends();
     const auto ts = new OGRTriangulatedSurface();
```
mloskot: By the way, why not to follow this exception safety convention that is already in use here?

https://github.com/OSGeo/gdal/blob/1a539bb944ebaa10d11bd32a8c78113c5dc3b9b4/gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobufdataset.cpp#L185-L186

rouault: Merging as in current state. I'll let @bjornharrtell issue a potential follow-up PR if he feels like using more unique_ptr goodness

PR #2076, opened 2019-12-02
changed file extension in example

jonahsullivan:
Not sure if .tsv and .txt are supposed to both be used, probably just one of them.

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:


rouault:
Good catch. Thanks

PR #2082, opened 2019-12-03
FlatGeobuf: Simplify memory management with unique_ptr

bjornharrtell:
@mloskot my C++ is rusty (unique_ptr wasn't around 20 years ago), but I now think I see the point of it and I hope I'm doing this correctly.

### DIFF
--- a/gdal/ogr/ogrsf_frmts/flatgeobuf/geometryreader.cpp
+++ b/gdal/ogr/ogrsf_frmts/flatgeobuf/geometryreader.cpp
@@ -95,17 +95,15 @@ OGRMultiPoint *GeometryReader::readMultiPoint()
     m_length = m_length / 2;
     if (m_length >= feature_max_buffer_size)
         return CPLErrorInvalidLength("MultiPoint");
-    const auto mp = new OGRMultiPoint();
+    auto mp = std::unique_ptr<OGRMultiPoint>(new OGRMultiPoint());
     for (uint32_t i = 0; i < m_length; i++) {
         m_offset = i;
         const auto p = readPoint();
-        if (p == nullptr) {
-            delete mp;
+        if (p == nullptr)
             return nullptr;
-        }
         mp->addGeometryDirectly(p);
     }
-    return mp;
+    return mp.release();
 }
 
 OGRMultiLineString *GeometryReader::readMultiLineString()
@@ -113,24 +111,20 @@ OGRMultiLineString *GeometryReader::readMultiLineString()
     const auto pEnds = m_geometry->ends();
     if (pEnds == nullptr)
         return CPLErrorInvalidPointer("MultiLineString ends data");
-    const auto mls = new OGRMultiLineString();
+    auto mls = std::unique_ptr<OGRMultiLineString>(new OGRMultiLineString());
     m_offset = 0;
     for (uint32_t i = 0; i < pEnds->size(); i++) {
         const auto e = pEnds->Get(i);
-        if (e < m_offset) {
-            delete mls;
+        if (e < m_offset)
             return CPLErrorInvalidLength("MultiLineString");
-        }
         m_length = e - m_offset;
         const auto ls = readSimpleCurve<OGRLineString>();
-        if (ls == nullptr) {
-            delete mls;
+        if (ls == nullptr)
             return nullptr;
-        }
         mls->addGeometryDirectly(ls);
         m_offset = e;
     }
-    return mls;
+    return mls.release();
 }
 
 OGRErr GeometryReader::readSimpleCurve(OGRSimpleCurve *sc)
@@ -232,22 +226,18 @@ OGRErr GeometryReader::readSimpleCurve(OGRSimpleCurve *sc)
 OGRPolygon *GeometryReader::readPolygon()
 {
     const auto ends = m_geometry->ends();
-    const auto p = new OGRPolygon();
+    auto p = std::unique_ptr<OGRPolygon>(new OGRPolygon());
     if (ends == nullptr || ends->size() < 2) {
         m_length = m_length / 2;
         const auto lr = readSimpleCurve<OGRLinearRing>();
-        if (lr == nullptr) {
-            delete p;
+        if (lr == nullptr)
             return nullptr;
-        }
         p->addRingDirectly(lr);
     } else {
         for (uint32_t i = 0; i < ends->size(); i++) {
             const auto e = ends->Get(i);
-            if (e < m_offset) {
-                delete p;
+            if (e < m_offset)
                 return CPLErrorInvalidLength("Polygon");
-            }
             m_length = e - m_offset;
             const auto lr = readSimpleCurve<OGRLinearRing>();
             m_offset = e;
@@ -255,12 +245,10 @@ OGRPolygon *GeometryReader::readPolygon()
                 continue;
             p->addRingDirectly(lr);
         }
-        if (p->IsEmpty()) {
-            delete p;
+        if (p->IsEmpty())
             return nullptr;
-        }
     }
-    return p;
+    return p.release();
 }
 
 OGRMultiPolygon *GeometryReader::readMultiPolygon()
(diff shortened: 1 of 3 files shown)


gdal/ogr/ogrsf_frmts/flatgeobuf/geometryreader.cpp
```diff
@@ -232,212 +226,165 @@ OGRErr GeometryReader::readSimpleCurve(OGRSimpleCurve *sc)
...
+    auto ps = std::unique_ptr<OGRPolyhedralSurface>(new OGRPolyhedralSurface());
     for (uoffset_t i = 0; i < parts->size(); i++) {
         GeometryReader reader { parts->Get(i), m_hasZ, m_hasM };
-        auto geom = reader.read();
-        auto surf = dynamic_cast<OGRSurface*>(geom);
-        if( surf == nullptr )
-        {
-            delete surf;
-            delete ps;
+        auto s = std::unique_ptr<OGRSurface>(reader.read()->toSurface());
```
rouault: this change is not equivent to previous code. the dynamic_cast<OGRSurface*>(geom) is (I think) needed to make sure the sub-object was really a OGRSurface. toSurface() resolves to a static_cast<> in non-debug builds, and so will result in a incorrect pointer if the geometry was not a Surface

gdal/ogr/ogrsf_frmts/flatgeobuf/geometryreader.cpp
```diff
@@ -232,212 +226,165 @@ OGRErr GeometryReader::readSimpleCurve(OGRSimpleCurve *sc)
...
+    auto mc = std::unique_ptr<OGRMultiCurve>(new OGRMultiCurve());
     for (uoffset_t i = 0; i < parts->size(); i++) {
         GeometryReader reader { parts->Get(i), m_hasZ, m_hasM };
-        auto geom = reader.read();
-        auto curve = dynamic_cast<OGRCurve*>(geom);
-        if( curve == nullptr )
-        {
-            delete geom;
-            delete mc;
+        auto c = std::unique_ptr<OGRCurve>(reader.read()->toCurve());
```
rouault: same as previous comment. dynamic_cast should likely be kept

gdal/ogr/ogrsf_frmts/flatgeobuf/geometryreader.cpp
```diff
@@ -232,212 +226,165 @@ OGRErr GeometryReader::readSimpleCurve(OGRSimpleCurve *sc)
...
+    auto cp = std::unique_ptr<OGRCurvePolygon>(new OGRCurvePolygon());
     for (uoffset_t i = 0; i < parts->size(); i++) {
         GeometryReader reader { parts->Get(i), m_hasZ, m_hasM };
-        auto geom = reader.read();
-        auto curve = dynamic_cast<OGRCurve*>(geom);
-        if( curve == nullptr )
-        {
-            delete geom;
-            delete cp;
+        auto c = std::unique_ptr<OGRCurve>(reader.read()->toCurve());
```
rouault: same as previous comment. dynamic_cast should likely be kept

gdal/ogr/ogrsf_frmts/flatgeobuf/geometryreader.cpp
```diff
@@ -232,212 +226,165 @@ OGRErr GeometryReader::readSimpleCurve(OGRSimpleCurve *sc)
...
     for (uoffset_t i = 0; i < parts->size(); i++) {
         GeometryReader reader { parts->Get(i), m_hasZ, m_hasM };
-        auto geom = reader.read();
-        auto curve = dynamic_cast<OGRCurve*>(geom);
-        if( curve == nullptr ||
-            cc->addCurveDirectly(curve) != OGRERR_NONE )
-        {
-            delete geom;
-            delete cc;
+        auto c = std::unique_ptr<OGRCurve>(reader.read()->toCurve());
```
rouault: same as previous comment. dynamic_cast should likely be kept

bjornharrtell:
@rouault changed back to dynamic_cast but found six places where it seem relevant.

bjornharrtell:
I can also mention that I tried and failed to use unique_ptr in a template function at https://github.com/OSGeo/gdal/blob/aa38907e1937da576cafaf32d013942b437161f5/gdal/ogr/ogrsf_frmts/flatgeobuf/geometryreader.h#L63-L73.

gdal/ogr/ogrsf_frmts/flatgeobuf/geometryreader.cpp
```diff
@@ -259,7 +259,7 @@ OGRMultiPolygon *GeometryReader::readMultiPolygon()
     auto mp = std::unique_ptr<OGRMultiPolygon>(new OGRMultiPolygon());
     for (uoffset_t i = 0; i < parts->size(); i++) {
         GeometryReader reader { parts->Get(i), GeometryType::Polygon, m_hasZ, m_hasM };
-        auto p = std::unique_ptr<OGRPolygon>(reader.read()->toPolygon());
+        auto p = std::unique_ptr<OGRPolygon>(dynamic_cast<OGRPolygon *>(reader.read()));
```
rouault: In that particular instance, I don't think the dynamic_cast<> is needed, as the reader is instructed to instanciate a GeometryType::Polygon. But ... in other places, such construct will result in a leak if read.read() returned a non-NULL pointer that is not of the type you dynamic_cast<> into

bjornharrtell: Thanks, I think I understand your explanation and have changed that instance back.

rouault: > have changed that instance back

Just for clarity: the other instances below that do dynamic_cast<> should be modified so that the unique_ptr is on the OGRGeometry* returned by reader.read(), not on the casted pointer

bjornharrtell: Hmm I don't understand why I'd have to do that (dynamic cast should be safe for null?) and if I do, I don't see unique_ptr can be useful anymore as it cannot be casted?

rouault: dynamic_cast is safe for null. But my point was that imagine that reader.read() returns a OGRPoint* instance, and you do std::unique_ptr<OGRLineString*>(dynamic_cast<OGRLineString*>(reader.read())) . In that case the object returned by reader.read() will be "lost" (leaking).
I think that for such cases, the std::unique_ptr makes things more complicated than the approach I took initially in 930b3c5d7e772282ffa90786e3e6d4fd0014ad2f

bjornharrtell: Mm yes I see, that's a bit sad though.

rouault: hum we could probably still use  the unique_ptr approach with something like:
```
auto poGeom = std::unique_ptr<OGRGeometry*>(reader.read());
if( !dynamic_cast<OGRLineString*>(poGeom.get()) // replace here OGRLineString by relevant type to check
     error_out;
poContainer->addGeometryDirectly(poGeom.release());
```

bjornharrtell: But that would require a second dynamic_cast for collection types that expect a subtype?

rouault: Not necessarily something like below should do, although not that pretty...
```
        auto geom = std::unique_ptr<OGRGeometry*>(reader.read());
        auto surf = dynamic_cast<OGRSurface*>(geom.get());
        if( surf == nullptr )
        {
            return nullptr;
        }
        geom.release(); // a bit weird to release without passing it explicitly, but we need to attach surf and not geom
        ms->addGeometryDirectly(surf);
```

bjornharrtell: Mmm funny. :) I had a thought that if the dynamic_cast<OGRSurface*>(geom.get()) passes the nullptr check then it should be ok to use ->toSurface() on the geom?

rouault: > then it should be ok to use ->toSurface() on the geom?

yes, you could do that too: ms->addGeometryDirectly(geom.release()->toSurface())

PR #2102, opened 2019-12-12
Docker: Allow https for ubuntu containers

### LINKED ISSUE #2101: GDAL docker containers do not include ca certificates

I would like to be able to use curl to connect to https inside of osgeo/gdal:ubuntu-small-3.0.2 (or latest)

This affects everything that uses curl too, such as /vsis3/

```
docker run -it osgeo/gdal:ubuntu-small-3.0.2 /bin/bash

$ curl https://google.com
curl: (77) error setting certificate verify locations:
  CAfile: /etc/ssl/certs/ca-certificates.crt
  CApath: /etc/ssl/certs

$ gdalinfo /vsis3/foo/bar.tif
ERROR 11: HTTP response code: 0
```

Adding  `ca-certificates` to this install should fix this issue. https://github.com/OSGeo/gdal/blob/32450164707ff447612c67262b7e2829168bfbb2/gdal/docker/ubuntu-small/Dockerfile#L171

## Operating system

Docker (Ubuntu 18.04)

Docker 3.0.2 imageId: f536bf51f64d

blacha:
alpine installs ca-certificates when libcurl is installed.

Fixes #2101 

## What does this PR do?

Adds ca-certificates to the base ubuntu images so that connection to https servers works.

#2101

### DIFF
--- a/gdal/docker/ubuntu-full/Dockerfile
+++ b/gdal/docker/ubuntu-full/Dockerfile
@@ -250,7 +250,7 @@ RUN date
 RUN apt-get update; \
     DEBIAN_FRONTEND=noninteractive apt-get install -y \
         libsqlite3-0 \
-        curl unzip
+        curl unzip ca-certificates
 
 # GDAL dependencies
 RUN apt-get update; \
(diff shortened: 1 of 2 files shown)


rouault:
Thanks!

PR #2107, opened 2019-12-14
Fix warpMemoryLimit python documentation

EvertEt:
https://gdal.org/programs/gdalwarp.html#cmdoption-gdalwarp-wm

## What does this PR do?
Fixes documentation

N/A
## Tasklist

 - [ ] Review
 - [ ] Adjust for comments

## Environment

N/A


rouault:
As stated in the header of this file, this is a generated file. Please also make the same change in the source file gdal/swig/include/python/gdal_python.i

EvertEt:
My bad for not checking this! I edited both files manually now. Is this ok or should the file be regenerated?

rouault:
That's fine for such a simple change. Thanks

PR #2109, opened 2019-12-14
appveyor.yml: install numpy 1.16.5 for python 2.7 compat

rouault:


rouault:
Hum, doesn't seem necessary for now. Closing until we really need that.

PR #2111, opened 2019-12-16
ogr/mysql: Add SRID to geometry when creating layer table(fixes #1015)

### LINKED ISSUE #1015: ogr2ogr with MySQL 8: POINT column doesn't have SRID attribute
I imported shape file using ogr2ogr and I checked created table. Created table doesn't have SIRD attribute on POINT column.

Created table have SIRD attribute on POINT column.

Note: 
POINT column of MySQL 8.0 can have SRID attribute. I think adding SRID attribute is desirable to avoid mistakes.

MySQL 8.0 Reference Manual / 11.5.1 Spatial Data Types
https://dev.mysql.com/doc/refman/8.0/en/spatial-type-overview.html
=> Columns with a spatial data type can have an SRID attribute, to explicitly indicate the spatial reference system (SRS) for values stored in the column.

1. Import shape file using ogr2ogr
ex) C:\> ogr2ogr -f "MySQL" MYSQL:geotest,user=root,password=root C:\work\point_2449.shp

2. Execute "show create table" for created table
ex) mysql> show create table point_2449\G
*************************** 1. row ***************************
       Table: point_2449
Create Table: CREATE TABLE `point_2449` (
  `OGR_FID` int(11) NOT NULL AUTO_INCREMENT,
  `SHAPE` geometry NOT NULL,
  `gid` decimal(10,0) DEFAULT NULL,
  `point_name` varchar(254) DEFAULT NULL,
  UNIQUE KEY `OGR_FID` (`OGR_FID`),
  SPATIAL KEY `SHAPE` (`SHAPE`)
) ENGINE=InnoDB AUTO_INCREMENT=4 DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci
1 row in set (0.00 sec)

* I hope modify CREATE TABLE as follows.
Before:  `SHAPE` geometry NOT NULL,
After:  `SHAPE` geometry NOT NULL SRID 2449,

## Operating system
Windows 10 x64

C:\> ogr2ogr --version
GDAL 2.3.2, released 2018/09/21

miurahr:
## What does this PR do?

Support MySQL 8.0 use SRID for  indexing geometry by creating table with geom column
as 'NOT NULL SRID <#srid>'

Also update CI test by updating MySQL from 8.0.1 to 8.0.18. 
(SRID index is supported from 8.0.3)

#1015

## Tasklist

 - [x] Clarify a behavior when specified unsupported SRID (discussion on #1015)
 - [x] May add some code to handle an unsupported SRID. 
 - [x] Add test case(s)
 - [ ] Review
 - [ ] Adjust for comments

## Environment

MySQL 8.0, MariaDB

### DIFF
--- a/gdal/ogr/ogrsf_frmts/mysql/ogrmysqldatasource.cpp
+++ b/gdal/ogr/ogrsf_frmts/mysql/ogrmysqldatasource.cpp
@@ -1032,6 +1032,7 @@ OGRMySQLDataSource::ICreateLayer( const char * pszLayerNameIn,
 
     const char *pszSI = CSLFetchNameValue( papszOptions, "SPATIAL_INDEX" );
     const bool bHasSI = ( eType != wkbNone && (pszSI == nullptr || CPLTestBool(pszSI)) );
+    int nSRSId = GetUnknownSRID();
 
     if( wkbFlatten(eType) == wkbNone )
     {
@@ -1042,12 +1043,22 @@ OGRMySQLDataSource::ICreateLayer( const char * pszLayerNameIn,
     }
     else
     {
-        osCommand.Printf(
-                 "CREATE TABLE `%s` ( "
-                 "   %s %s UNIQUE NOT NULL AUTO_INCREMENT, "
-                 "   %s GEOMETRY %s)",
-                 pszLayerName, pszExpectedFIDName, pszFIDType, pszGeomColumnName,
-                 bHasSI ? "NOT NULL" : "");
+        // when using mysql8 and SRS is specified, use SRID option for geometry.
+        if( GetMajorVersion() < 8 || IsMariaDB() || poSRS == nullptr ||
+            (nSRSId = FetchSRSId( poSRS )) == GetUnknownSRID() )
+            osCommand.Printf(
+                    "CREATE TABLE `%s` ( "
+                    "   %s %s UNIQUE NOT NULL AUTO_INCREMENT, "
+                    "   %s GEOMETRY %s)",
+                    pszLayerName, pszExpectedFIDName, pszFIDType, pszGeomColumnName,
+                    bHasSI ? "NOT NULL" : "");
+        else
+            osCommand.Printf(
+                    "CREATE TABLE `%s` ( "
+                    "   %s %s UNIQUE NOT NULL AUTO_INCREMENT, "
+                    "   %s GEOMETRY %s /*!80003 SRID %d */)",
+                    pszLayerName, pszExpectedFIDName, pszFIDType, pszGeomColumnName,
+                    bHasSI ? "NOT NULL" : "", nSRSId);
     }
 
     if( CSLFetchNameValue( papszOptions, "ENGINE" ) != nullptr )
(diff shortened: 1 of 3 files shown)


miurahr:
All the modifications are inside `else` clause of  `if( GetMajorVersion() < 8 || IsMariaDB() )` but test error happened with MariaDB as follows:

```
27025       ogrtest.quick_create_layer_def(gdaltest.mysql_lyr,
27026                                       [('AREA', ogr.OFTReal),
27027                                        ('EAS_ID', ogr.OFTInteger),
27028                                        ('PRFEDEA', ogr.OFTString),
27029                                        ('SHORTNAME', ogr.OFTString, 8),
27030                                        ('INT64', ogr.OFTInteger64)])
27031    
27032        ######################################################
27033        # Copy in poly.shp
27034    
27035        dst_feat = ogr.Feature(feature_def=gdaltest.mysql_lyr.GetLayerDefn())
27036    
27037        feat = shp_lyr.GetNextFeature()
27038        gdaltest.poly_feat = []
27039    
27040        while feat is not None:
27041    
27042            gdaltest.poly_feat.append(feat)
27043    
27044            dst_feat.SetFrom(feat)
27045            dst_feat.SetField('INT64', 1234567890123)
27046            gdaltest.mysql_lyr.CreateFeature(dst_feat)
27047    
27048            feat = shp_lyr.GetNextFeature()
27049    
27050        dst_feat.Destroy()
27051    
27052        assert gdaltest.mysql_lyr.GetFeatureCount() == shp_lyr.GetFeatureCount(), \
27053            'not matching feature count'
27054    
27055>       assert gdaltest.mysql_lyr.GetSpatialRef().GetAuthorityCode(None) == shp_lyr.GetSpatialRef().GetAuthorityCode(None), \
27056            'not matching spatial ref'
27057E       AttributeError: 'NoneType' object has no attribute 'GetAuthorityCode'
27058
27059ogr_mysql.py:128: AttributeError
```

miurahr:
I think it is OK.
AppVeyor failed out of mysql and it skips ogr_mysql on windows CI, and an patch only modifies ogr/mysql

gdal/ogr/ogrsf_frmts/mysql/ogrmysqldatasource.cpp
```diff
@@ -1083,12 +1094,14 @@ OGRMySQLDataSource::ICreateLayer( const char * pszLayerNameIn,
...
 /* -------------------------------------------------------------------- */
 /*      Try to get the SRS Id of this spatial reference system,         */
-/*      adding tot the srs table if needed.                             */
+/*      adding to the srs table if needed.                             */
 /* -------------------------------------------------------------------- */
-    int nSRSId = GetUnknownSRID();
-
-    if( poSRS != nullptr )
-        nSRSId = FetchSRSId( poSRS );
+    if( GetMajorVersion() < 8 || IsMariaDB() )
```
rouault: why is nSRSID fetched conditionnaly to the mysql version ? It seems it is used unconditionnaly later at https://github.com/OSGeo/gdal/pull/2111/files#diff-b546c2f165c017991b14e08089e3d95aR1138 ? So this should probably be moved around line https://github.com/OSGeo/gdal/pull/2111/files#diff-b546c2f165c017991b14e08089e3d95aR1035

miurahr: It seems that we can remove the line https://github.com/OSGeo/gdal/pull/2111/files#diff-b546c2f165c017991b14e08089e3d95aR1101 ' nSRSId = GetUnknownSRID();' because it already set previously at  https://github.com/OSGeo/gdal/pull/2111/files#diff-b546c2f165c017991b14e08089e3d95aR1035 
This is unnecessary call.

rouault: I still find the fact that nSRSId = FetchSRSId( poSRS ) is called both at line 1048 and 1102 confusing. Cannot just it be done once just after the initialization of line 1035 ?

miurahr: I'd like to move a following blocks on just after line 1048 as such;
```
    int nSRSId = GetUnknownSRID();

    // Calling this does no harm
    InitializeMetadataTables();

/*      Try to get the SRS Id of this spatial reference system,         */
/*      adding to the srs table if needed.                             */

    if (poSRS != nullptr)
        nSRSId = FetchSRSId(poSRS);
```

This looks works.

PR #2113, opened 2019-12-17
Revise how SRS methods deal with TOWGS84

rouault:
When importing from EPSG, GDAL 3.0.0 to 3.0.2 followed the scheme used in
GDAL 1.x and GDAL 2.x, that is they tried to attach a TOWGS84 transformation.
While the intent was to have some sort of backward compatibility, this may
be a pain for the future.
So do the following changes:
- importFromEPSG(): no longer attach a TOWGS84 transformation, unless the
  user set the OSR_ADD_TOWGS84_ON_IMPORT_FROM_EPSG=YES configuration option
- Add a OGRSpatialReference::AddGuessedTOWGS84() to attach such a transformation,
  when possible (note: this will do it in a more prudent way than GDAL 1.x/2.x,
  that is only if a transformation is found for the whole area of use of the CRS)
- exportToProj4(): if the SRS has no transformation to WGS84,
  attach a +towgs84 if the SRS has a EPSG code and AddGuessedToWGS84() succeeds.
  Can be disabled with the OSR_ADD_TOWGS84_ON_EXPORT_TO_PROJ4 = NO configuration
  option
 - exportToWkt() with WKT1 format:
   Add a OSR_ADD_TOWGS84_ON_EXPORT_TO_WKT1 = YES/NO configuration option, which
   defaults to NO. If set to  YES, then a transformation to WGS84 using
   AddGuessedToWGS84() logic is researched to add a TOWGS84[] node.

Planned to be backported to GDAL 3.0 branch for 3.0.3

PR #2122, opened 2019-12-18
MITAB: do not set (by default) TOWGS84 when reading a known Datum 

rouault:
Refs https://github.com/OSGeo/gdal/pull/2113#issuecomment-566990508

Past behaviour can be restored with setting the MITAB_SET_TOWGS84_ON_KNOWN_DATUM
configuration option to YES.


nyalldawson:
Thanks!

PR #2125, opened 2019-12-21
Enable Bash completions and control installation

cjmayo:
## What does this PR do?

bash-completion expects completions to be installed in a given directory
and named for program being completed.

Detect the directories as suggested in the FAQ:
https://github.com/scop/bash-completion/blob/master/README.md#faq

Control installation with a new configuration option --with-bash-completion.
If it is not specified nothing will be installed.
If a path is passed completions will be installed in that directory.

Install the GDAL completion script as gdalinfo and create symlinks to it
for the other programs.

---

This works for me and I've tried to address the situation raised in the previous ticket (below) by not installing by default, and it also allows a custom path to be used. I'm a bit cautious because I can't see how anyone is using the current code. On the other hand when ${prefix} is being prepended to the path returned by pkg-config it does seem pretty broken.

https://trac.osgeo.org/gdal/ticket/6381

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS: Gentoo Linux
* Compiler: gcc


rouault:
Could you run "autoconf" and commit the modified configure file ?

cjmayo:
Done, although my version appears to be newer. I've added an additional commit to separate the changes from the version change alone.

cjmayo:
Perhaps I shouldn't have added the autoreconf changes into the same commit as the real changes. I'm happy to separate those out again into a separate commit.

rouault:
can you revert at least the changes in port/cpl_config.h.in  ?

cjmayo:
> can you revert at least the changes in port/cpl_config.h.in ?

That's done (and a separate autoreconf commit).

PR #2130, opened 2019-12-24
Use ST_GEOMETRY_COLUMNS table for metadata in MySQL 8.0 and later (Enhance #1015)

miurahr:
#What does this PR do?

When using MySQL 8.0.3 and later, utilize System geometry metadata table instead of manual management.

#1015

## Tasklist

 - [x] Review
 - [x] Adjust for comments

## Environment

MySQL 8.0 / MariaDB

## Dependency

This improvement is depends on #2111

### DIFF
--- a/gdal/ogr/ogrsf_frmts/mysql/ogrmysqldatasource.cpp
+++ b/gdal/ogr/ogrsf_frmts/mysql/ogrmysqldatasource.cpp
@@ -387,6 +387,10 @@ OGRLayer *OGRMySQLDataSource::GetLayer( int iLayer )
         return papoLayers[iLayer];
 }
 
+/* =====================================================================*/
+/* Handle spatial reference id and index for older MySQL and MariaDB    */
+/* =====================================================================*/
+
 /************************************************************************/
 /*                      InitializeMetadataTables()                      */
 /*                                                                      */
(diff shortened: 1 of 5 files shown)


miurahr:
Rebased on master which merged  #2111 .

miurahr:
@joshbernfeld could you review it?

gdal/ogr/ogrsf_frmts/mysql/ogrmysqldatasource.cpp
```diff
@@ -1099,77 +1189,8 @@ OGRMySQLDataSource::ICreateLayer( const char * pszLayerNameIn,
...
-/*                                                                      */
-/* -------------------------------------------------------------------- */
-    osCommand.Printf(
-             "DELETE FROM geometry_columns WHERE f_table_name = '%s'",
-             pszLayerName );
-
-    if( mysql_query(GetConn(), osCommand ) )
-    {
-        ReportError( osCommand );
+    if( UpdateMetadataTables(pszLayerName, eType, pszGeomColumnName, nSRSId) != OGRERR_NONE )
```
miurahr: Moving these chunks and introduce UnpdateMetadataTables() makes ICreateLayer()  simpler and easy to understand by combination with InitializeMetadataTables(), which are necessary only when legacy MySQL and MariaDB.

joshbernfeld:
@miurahr Sure, will have time tomorrow.

gdal/ogr/ogrsf_frmts/mysql/ogrmysqltablelayer.cpp
```diff
@@ -383,9 +383,13 @@ OGRFeatureDefn *OGRMySQLTableLayer::ReadTableDefinition( const char *pszTable )
...
 
-        osCommand = "SELECT type, coord_dimension FROM geometry_columns WHERE f_table_name='";
-        osCommand += pszTable;
-        osCommand += "'";
+        if( poDS->GetMajorVersion() < 8 || poDS->IsMariaDB() )
+            osCommand.Printf("SELECT type, coord_dimension FROM geometry_columns WHERE f_table_name='%s'",
+                             pszTable);
+
+        else
+            osCommand.Printf("SELECT GEOMETRY_TYPE_NAME FROM INFORMATION_SCHEMA.ST_GEOMETRY_COLUMNS "
```
joshbernfeld: The deprecated selected statement above includes the coord_dimension whereas this select statement does not. Does support need to be included for 3 dimensional coordinate space?

joshbernfeld: Also line 409 `if( papszRow[1] != nullptr && atoi(papszRow[1]) == 3 )` may be accessing an array out of bounds since that column was not selected?

miurahr: In my understandings,  MySQL 8  has only 2D functions, and integrated ST_GEOMETRY_COLUMNS table does not have a column as same meaning as coord_dimensions. I saw here: https://dev.mysql.com/doc/refman/8.0/en/st-geometry-columns-table.html
@YoshiakiYamasaki could  you help about a 3D data parameter?

YoshiakiYamasaki: Implementation is under consideration in the future, but 3D is not implemented now. Currently MySQL 8.0 only supports 2D.

FYI:
Spatial Support in MySQL
https://www.slideshare.net/NorvaldRyeng/spatial-support-in-mysql/5

* This document made by derector of MySQL GIS development team.

miurahr:
I've added more test to writing and reading geometries includes 3D and 4D WKTs.

It will illustrate a behavior for unsupported geometry in mysql. And it may show what happened if 
>line 409 if( papszRow[1] != nullptr && atoi(papszRow[1]) == 3 ) may be accessing an array out of bounds since that column was not selected?

miurahr:
when using `get_wkt_data_series()` https://travis-ci.com/OSGeo/gdal/jobs/270855155, I've got a following error which happens with MariaDB.

```
ERROR 1: MySQL error message:Column 'SHAPE' cannot be null Description: INSERT INTO `ogr_mysql_28_15` (`SHAPE` ) VALUES (ST_GeomFromText('GEOMETRYCOLLECTION (POINT (4 6),GEOMETRYCOLLECTION (POINT (4 6),LINESTRING (4 6,7 10)))',2) )
```

I'd like to test it individually on master and check whether  is it a regression or not.

miurahr:
I'd like to separate an addtion of  autotest test case,  which moves another PR https://github.com/OSGeo/gdal/pull/2137

gdal/ogr/ogrsf_frmts/mysql/ogrmysqltablelayer.cpp
miurahr: Now line 409  is executed only when MySQL<8 and MariaDB.

rouault:
@miurahr Seems this is ready to be merged ?

miurahr:
Yes, it is ready.

PR #2134, opened 2019-12-27
Introduce pytest-runner

miurahr:
## What does this PR do?

It is necessary  to have a native executable to run autotest  to launch a integrated debugger from IDEs such as CLion.

pytest_runner can get arguments which is passed to pytest as same as
`python -m pytest <arguments>`

When developer find a  regression and want to debug a autotest session,
he/she can  configure IDE to run autotest with pytest_runner such as

```
pytest_runner autotest/ogr/ogr_mysql.py
```

I've used this utility with JetBrains CLion and cmake4gdal project.

No.

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:

### DIFF
--- a/autotest/GNUmakefile
+++ b/autotest/GNUmakefile
@@ -24,3 +24,20 @@ pyscripts-test:
 
 test check:
 	@$(MAKE) $(MFLAGS) gcore-test gdrivers-test alg-test ogr-test osr-test utilities-test pyscripts-test
+
+
+include ../gdal/GDALmake.opt
+
+ifndef PYTHON
+        PYTHON=python
+endif
+
+PYTHON_LIB := $(shell $(PYTHON) -c "import sysconfig; print(sysconfig.get_config_var('BLDLIBRARY'))" )
+PYTHON_INCLUDE_DIRS := $(shell $(PYTHON) -c "import sysconfig; print(sysconfig.get_path('include'))" )
+CXXFLAGS += -I$(PYTHON_INCLUDE_DIRS)
+
+pytest_runner: pytest_runner.o
+	$(CXX) $(LDFLAGS) $< $(PYTHON_LIB) -o $@
+
+pytest_runner.o: pytest_runner.cpp
+	$(CXX) $(CXXFLAGS) -c $<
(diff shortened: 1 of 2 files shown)


rouault:
I'm not a super fan of adding this. Isn't there a way to run `python -m pytest script` from this IDE ? Perhaps you should ask on gdal-dev or on CLion support forums. That seams strange to have to use such a complicated approach.

rouault:
Thinking you could just wrap pytest invokation with sh:
```
sh -c "pytest ../autotest/gcore/basic_test.py"
```

mloskot:
@miurahr where did you see requirement of native runner for pytest? Nothing is mentioned in the docs 

https://www.jetbrains.com/help/clion/run-debug-configuration-py-test.html

miurahr:
@mloskot Thank you for suggestion. I know it and used heavily.
After support session with jetbrains  support partner, I recognize the native pytest runner is a 'python' debugger rather than C-debugger.  https://github.com/fabioz/PyDev.Debugger
Actually it is equivalent of pycharm.
It is ok when  debugging python project. It is good when I want to know where autotest fails.

I'd like to discuss how we can help developer against GDAL C++ debugging.

If you recognize  some regression with autotest, there is an only way to investigate gdal, print debug, using CPLDebug(). It is not a way for me as a IDE user to debug C++ project.

rouault:
You can debug the GDAL native library through pytest with gdb (or ddd)
```
gdb --args python `which pytest` foo.py
run # first execution so that GDAL symbols are loaded
b GDALAllRegister # for example
run
```

mloskot:
Every IDE allows to specify executable that will be executed under debugger. Simply, specify `python.exe` with arguments you wish to pass to Python and GDAL tests.

Maintaining a dedicated runner program seems to be an unnecessary prosthesis.

miurahr:
Thank you opinions, actually this kinds are issues on build system and IDE side not on GDAL.
I'd like to abandon here and ask the IDE and a build script to treat the issue.

PR #2135, opened 2019-12-27
FlatGeobuf: Support FlatGeobuf as WFS outputformat

bjornharrtell:
Implements support for FlatGeobuf as WFS outputformat

~~WIP because I want to make it optionally support streaming mode similar to GML before calling this done.~~


autotest/ogr/ogr_flatgeobuf.py
```diff
@@ -410,3 +415,109 @@ def test_ogr_flatgeobuf_datatypes():
...
+        webserver.server_stop(process, port)
+        pytest.fail('did not get expected layer')
+    name = lyr.GetName()
+    if name != 'topp:states':
+        print(name)
+        webserver.server_stop(process, port)
+        pytest.fail('did not get expected layer name (got %s)' % name)
+
+    feat = lyr.GetNextFeature()
+    if feat.GetField('STATE_NAME') != 'Illinois' or \
```
rouault: probably just checking the geometry is not None would be sufficient

autotest/ogr/ogr_flatgeobuf.py
```diff
@@ -410,3 +415,109 @@ def test_ogr_flatgeobuf_datatypes():
...
+                    content = f.read()
+                    f.close()
+                    self.wfile.write(content)
+                    return
+
+                if self.path == '/fakewfs?OUTPUTFORMAT=application/flatgeobuf&SERVICE=WFS&VERSION=1.1.0&REQUEST=GetFeature&TYPENAME=topp:states':
+                    self.send_response(200)
+                    self.send_header('Content-type', 'application/flatgeobuf')
+                    self.end_headers()
+                    f = open('data/testfgb/wfs/get_feature.fgb', 'rb')
```
rouault: any chance of using a smaller test file with less features & attributes ?

bjornharrtell: Good idea, I've chosen another smaller layer from the geoserver demo test data.

bjornharrtell:
@rouault I've looked into QGIS WFS provider code and it's both quite complex and hard coded to GML output format. I see you have hacked on it in the past and wonder, wouldn't it make sense to migrate to GDAL as the underlying implementation? (which would then give support for other output formats)

rouault:
> I see you have hacked on it in the past and wonder, wouldn't it make sense to migrate to GDAL as the underlying implementation?

Wouldn't be easily doable. The QGIS GML parser is a streaming parser. It can display progressively multi-gigabyte GML WFS responses without any problem. The OGR GML driver is also a streaming one, but this is an implementation detail not exposed in the GDAL API

bjornharrtell:
Hmm ok, but I suppose feeding the stream directly to GDAL could work? I've opened https://github.com/qgis/QGIS/issues/33541.

PR #2152, opened 2020-01-06
Set StringsAsUTF8 flag to TRUE for GeoJSON

### LINKED ISSUE #2151: Make OLCStringsAsUTF8 true for GeoJSON

GeoJSON layers should advertise UTF-8 strings (OLCStringsAsUTF8 as True) but they do not. The GeoJSON layer's capability testing delegates to the MEM driver, which does not advertise UTF-8 strings.

Theoretically, a GeoJSON file might be UTF-16 or UTF-32, but these are very rare and only weakly interoperable. By advertising UTF-8, the GeoJSON driver can help applications (like Fiona, see https://github.com/Toblerity/Fiona/blob/maint-1.8/fiona/ogrext.pyx#L544-L569) try to do the right thing with respect to data encoding.

sgillies:
Resolves #2151.

### DIFF
--- a/gdal/ogr/ogrsf_frmts/geojson/ogrgeojsonwritelayer.cpp
+++ b/gdal/ogr/ogrsf_frmts/geojson/ogrgeojsonwritelayer.cpp
@@ -318,6 +318,6 @@ int OGRGeoJSONWriteLayer::TestCapability( const char* pszCap )
     else if( EQUAL(pszCap, OLCSequentialWrite) )
         return TRUE;
     else if( EQUAL(pszCap, OLCStringsAsUTF8) )
-	return TRUE;
+        return TRUE;
     return FALSE;
 }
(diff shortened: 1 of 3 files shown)


gdal/ogr/ogrsf_frmts/geojson/ogrgeojsonwritelayer.cpp
```diff
@@ -317,6 +317,7 @@ int OGRGeoJSONWriteLayer::TestCapability( const char* pszCap )
         return TRUE;
     else if( EQUAL(pszCap, OLCSequentialWrite) )
         return TRUE;
-
+    else if( EQUAL(pszCap, OLCStringsAsUTF8) )
+	return TRUE;
```
rouault:
```suggestion
        return TRUE;
```

PR #2154, opened 2020-01-07
Also search for proj lib in prefix/lib64

m-kuhn:
## What does this PR do?

If using `--with-proj=prefix_path` gdal would so far only look for the libproj.so in `$prefix_path/lib`, whereas proj on (some?) 64 bit systems installs the library into `$prefix_path/lib64`. This PR checks for both locations.

I assume https://gis.stackexchange.com/questions/317109/build-gdal-with-proj-version-6

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS: Fedora 31
* Compiler: irrelevant

### DIFF
--- a/gdal/configure
+++ b/gdal/configure
@@ -24244,14 +24244,15 @@ $as_echo "$as_me: proj.h found" >&6;}
   else
 
     ORIG_LIBS="$LIBS"
-    LIBS="-L$with_proj/lib -lproj $ORIG_LIBS"
-    ac_ext=cpp
+    for libdir in "$with_proj/lib" "$with_proj/lib64"; do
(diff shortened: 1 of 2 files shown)


m-kuhn:
Note: this is still WIP, still get plenty of
```
/usr/bin/ld: .../gdal/.libs/libgdal.so: undefined reference to `proj_create_conversion_oblique_stereographic'```

and similar

m-kuhn:
I still had to fallback to settings `LD_LIBRARY_PATH`, `PATH` and compiler and linker flags as described in https://gis.stackexchange.com/a/339792/9839

PR #2179, opened 2020-01-20
Update links for OGR Python driver examples

vincentsarago:
## What does this PR do?

Fix links for OGR Python driver examples


mloskot:
Awesome, thanks!

PR #2182, opened 2020-01-21
gtiff.rst: respelled ambigious Byte word

Instant-Monkey:
The Byte capitalized before a comma make it like Byte is an available band type. I sadly lost too much time over this, trying to make Byte work like it was a Int8.

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments
 - [ ] Docs CI build

### DIFF
--- a/gdal/doc/source/drivers/raster/gtiff.rst
+++ b/gdal/doc/source/drivers/raster/gtiff.rst
@@ -14,7 +14,7 @@ and somewhat less varieties can be written.
 GDAL also supports reading and writing BigTIFF files (evolution of the TIFF format
 to support files larger than 4 GB).
 
-Currently band types of Byte, UInt16, Int16, UInt32, Int32, Float32,
+Currently band types of byte: UInt16, Int16, UInt32, Int32, Float32,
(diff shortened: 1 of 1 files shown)


mloskot:
> trying to make Byte work like it was a Int8.

It's unclear what you mean here.

`Byte` is a valid type for band/pixel data, it is encoded as `GDT_Byte` enum:

https://github.com/OSGeo/gdal/blob/dc189ddda5c07f8dfc94a25925afd7339c6944a6/gdal/gcore/gdal.h#L62

rouault:
I what was unclear was that byte is meant to be unsigned in GDAL, then ``Byte (unsigned)`` would be better

mloskot:
@rouault Right, if the ambiguity is about the signedness, then yes it could be specified.

@Instant-Monkey Would you be willing to instead make the docs read `Byte (unsigned)` and update the PR?

Instant-Monkey:
@rouault this was in fact the issue, the unsigned mention help a lot, since I was trying to use a signed char. This make me change for unsigned and everything works.  

@mloskot Will do. 

Thanks for your help

PR #2185, opened 2020-01-22
tiledb: zero tile data on read when updating

normanb:
## What does this PR do?

Zeros the tile read when the array is being updated. This is a pattern followed by other drivers e.g. https://github.com/OSGeo/gdal/blob/master/gdal/frmts/hdf4/hdf4imagedataset.cpp#L300 and prevents artifacts when updating

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/autotest/gdrivers/tiledb_write.py
+++ b/autotest/gdrivers/tiledb_write.py
@@ -30,6 +30,7 @@
 ###############################################################################
 
 from osgeo import gdal
+import numpy as np
 import pytest
 
 import gdaltest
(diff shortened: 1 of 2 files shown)


rouault:
This change doesn't look obviously correct to me (I'm not sure why the HDF4 driver does this). Is it because TileDB is read-only or write-only ?

normanb:
@rouault Yes, the driver in TileDB is either in read or write mode. At the block level should I be supporting reading the original data and allowing partial updates? There are quite a few drivers that use `memset` in the way I currently am but I can change the code to allow `r+` by opening the tiledb array in different modes as necessary.

rouault:
> At the block level should I be supporting reading the original data and allowing partial updates?

That would be the ideal if it isn't too convoluted.

Otherwise, if deciding for read-only and write-only mode, you should probably don't accept to Open() with poOpenInfo->eAccess == GA_Update, and update documentation accordingly.

normanb:
@rouault we are opening a read only copy of the metadata in the TileDB driver when the dataset is being opened in update mode, I have moved this to a member variable in the dataset and added similar functionality to `IReadBlock` and it is working. I will do some more local testing and push to this branch when it is ready for review.

rouault:
On the mingw target
```
gdrivers\tiledb_write.py:33: in <module>
    import numpy as np
E   ImportError: No module named numpy
```
This is a bit weird, but we don't require numpy to build the GDAL Python bindings... So you may want to move the import in the test function and pytest.skip() it in case of ImportError

rouault:
@normanb The backport bot being non functional I let you create a backport pull request with cherry-picking

Issue #2173, opened 2020-01-17
Loses link against libxml2 with 2.9.10 (uses xml2-config)


sebastic:
As reported by Mattia Rizzolo in [Debian Bug #949167](https://bugs.debian.org/949167):
> Your package is using `xml2-config` to detect and use libxml2.  I'm
> removing that script, so please update your build system to use
> pkg-config instead.
>
> Additionally, your package is not failing to build, so it's creating
> binaries that are not linked against libxml2 anymore:
> ```
> Control files of package libgdal-dev: lines which differ (wdiff format)
> -----------------------------------------------------------------------
> Installed-Size: [-49905-] {+49850+}
>
> Control files of package libgdal26: lines which differ (wdiff format)
> ---------------------------------------------------------------------
[...]

Issue #1412, opened 2019-04-02
gdal configure should switch from pg_config to pkg-config


tigerfoot:
## Expected behavior and actual behavior.

Actually .configure (configure.ac) is used to find pg_config program to determine PostgreSQL header of libpq
With now pg 11, PostgreSQL would like to prefer to use pkg-config libpq for client part (which gdal is, right?) and reserve the usage of pg_config to backend (extension and so)

## Steps to reproduce the problem.

Building with as BuildRequires: pkg-config(libpq) give no pg_config found
Ideally --with-pg should accept result of 
```
pkg-config --libs --cflags libpq
-I/usr/include/pgsql -lpq
```

## Operating system

openSUSE Tumbleweed (20190402)

## GDAL version and provenance

2.4.1 released tarball

What would be you opinion on that ?

rouault:
I'm fine with that

tigerfoot:
But looking at the code, I'm seeing actually --includedir-server did we need that in gdal ?

rouault:
> I'm seeing actually --includedir-server did we need that in gdal ?

I don't have in mind why we would need it. As far as I can see, GDAL only includes libpq-fe.h

tigerfoot:
Ok let see if I can propose a patch for this.

tigerfoot:
Seems that ugly patch is able to have PostgreSQL yes and building (need more tests tomorrow).
I'm so sorry but not really in my confort zone with configure.ac

```
diff -rup a/configure.ac b/configure.ac
--- a/configure.ac      2019-03-15 13:30:02.000000000 +0100
+++ b/configure.ac      2019-04-02 21:21:00.544544968 +0200
@@ -1252,7 +1252,7 @@ AC_ARG_WITH(pg,
               [Include PostgreSQL GDAL/OGR Support (ARG=path to pg_config)]),,)

 if test "x$with_pg" = "xyes" -o "x$with_pg" = "x" ; then
-  AC_PATH_PROG(PG_CONFIG, pg_config, no)
+  AC_PATH_PROG(PG_CONFIG, pkg-config, no)
 else
   PG_CONFIG=$with_pg
 fi
@@ -1268,24 +1268,25 @@ if test "x$PG_CONFIG" = "xno" ; then
   AC_MSG_RESULT([no])

 else
-  if test -d ${PG_CONFIG} ; then
-      AC_MSG_RESULT([no])
[...]

PR #2190, opened 2020-01-23
improve handling of --with-pg

tbussmann:
for 3.0 https://github.com/OSGeo/gdal/commit/45e06386d9099cbbe4f8eb7b4c2b8edca09ed144 / #1412 changed the arguments of `--with-pg` configure option from taking a path to taking a boolean. This is easily overlooked by users upgrading from earlier versions and effectively disables PostgreSQL support despite this option defaults to `yes`. Packagers are having trouble with this too, see https://github.com/OSGeo/homebrew-osgeo4mac/issues/1291 and https://github.com/PostgresApp/PostgresApp/issues/548
To mitigate the issue, I add a configure error if the argument of `--with-pg` is in the old form (similar to what you did with the equivalent change for `--with-xml2` in https://github.com/OSGeo/gdal/commit/bfdf002a53e4ef7ce935b365c4490c96dee6f605 / #2173) and mention the breaking change in the migration guide.

### DIFF
--- a/gdal/MIGRATION_GUIDE.TXT
+++ b/gdal/MIGRATION_GUIDE.TXT
@@ -11,7 +11,7 @@ MIGRATION GUIDE FROM GDAL 2.4 to GDAL 3.0
 
 - Unix Build: ./configure arguments --without-bsb, --without-grib,
   and --without-mrf have been renamed to --disable-driver-bsb,
-  --disable-driver-grib and --disable-driver-mrf
+  --disable-driver-grib and --disable-driver-mrf; Arguments of --with-pg changed.
 - Substantial changes, sometimes backward incompatible, in coordinate reference
   system and coordinate transformations have been introduced per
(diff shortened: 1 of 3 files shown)


gdal/MIGRATION_GUIDE.TXT
```diff
@@ -11,7 +11,7 @@ MIGRATION GUIDE FROM GDAL 2.4 to GDAL 3.0
 
 - Unix Build: ./configure arguments --without-bsb, --without-grib,
   and --without-mrf have been renamed to --disable-driver-bsb,
-  --disable-driver-grib and --disable-driver-mrf
+  --disable-driver-grib and --disable-driver-mrf; Arguments of --with-pg changed.
```
rouault:
```suggestion
  --disable-driver-grib and --disable-driver-mrf;
 - Unix build: Arguments of --with-pg changed to yes/no only.
```

PR #2200, opened 2020-01-30
FlatGeobuf: spec v3

bjornharrtell:
* Spec v3 has new index structure to enable sequential read also for spatial index traversal
* No backwards compatibility (I believe the window for that is still open)


gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
```diff
@@ -327,13 +326,20 @@ void OGRFlatGeobufLayer::Create() {
...
-    Rect extent = calcExtent(m_featureItems);
+    NodeItem extent = calcExtent(m_featureItems);
     auto extentVector = extent.toVector();
 
     writeHeader(m_poFp, m_featuresCount, &extentVector);
 
     CPLDebug("FlatGeobuf", "Sorting items for Packed R-tree");
     hilbertSort(m_featureItems);
+    CPLDebug("FlatGeobuf", "Calc new feature offsets");
+    auto featureOffset = 0;
```
rouault: a bit dangerous here to use auto for variable declaration. I guess this resolves to int, and not as uint64 as likely expected ? Did you test with > 4 GB files ?

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
```diff
@@ -412,9 +407,14 @@ OGRFlatGeobufLayer::~OGRFlatGeobufLayer()
 }
 
 OGRErr OGRFlatGeobufLayer::readFeatureOffset(uint64_t index, uint64_t &featureOffset) {
-    if (VSIFSeekL(m_poFp, m_offsetIndices + (index * sizeof(featureOffset)), SEEK_SET) == -1)
+    auto treeSize = PackedRTree::size(m_featuresCount, m_indexNodeSize);
```
rouault: some const qualifiers for variables that won't change would be welcome

gdal/ogr/ogrsf_frmts/flatgeobuf/packedrtree.cpp
```diff
@@ -150,187 +155,179 @@ void PackedRTree::init(const uint16_t nodeSize)
...
         {
-            CPL_LSBPTR64(&nodeRects[i].minX);
-            CPL_LSBPTR64(&nodeRects[i].minY);
-            CPL_LSBPTR64(&nodeRects[i].maxX);
-            CPL_LSBPTR64(&nodeRects[i].maxY);
+            CPL_LSBPTR64(&nodeItems[i].minX);
+            CPL_LSBPTR64(&nodeItems[i].minY);
+            CPL_LSBPTR64(&nodeItems[i].maxX);
+            CPL_LSBPTR64(&nodeItems[i].maxY);
+            CPL_LSBPTR64(&nodeItems[i].maxY);
```
rouault: duplicate line

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
bjornharrtell: Changed. I haven't tested with > 4 GB files but would like too. It's mostly due to it being time consuming to find and convert a suitable source.

rouault: maybe from there: https://download.osmdata.xyz/

bjornharrtell: Made some tests using the waterways dataset. Got a 26 GB big file and it works well. Though it did seem a bit large.. which looks like it's explained by the packages from osmdata.xyz containing empty strings instead of null for unset attributes. Makes me think it would be nice with a generic option to convert empty strings to null for ogr2ogr?

rouault: >  a generic option to convert empty strings to null for ogr2ogr?

something that is regularly asked for indeed. Want to tackle that ? -emptyStrAsNull ? We have a EMPTY_STRING_AS_NULL open option for the CSV driver

PR #2203, opened 2020-01-31
AWS: Fix error in loading ~/.aws/config file

glostis:
## What does this PR do?

This PR fixes a copy/paste error at line 890 of `port/cpl_aws.cpp` that caused `~/.aws/credentials` to be loaded instead of `~/.aws/config` here:

https://github.com/OSGeo/gdal/blob/cdd65c9ce4f9ed11b53620c79c25df4bdd29797c/gdal/port/cpl_aws.cpp#L878-L890

This caused issues when using a non-default AWS region stored in the default `~/.aws/config` file, as this file was never parsed.

The test in autotest checking for non-default profiles did not catch this error, because it was using the `AWS_CONFIG_FILE` config option to point to a non-default config file. I have modified the test to make it use the default config file so that the non-tested lines are now covered (I had to play around with `$HOME` in order to do this without tampering with the user's real config/credentials files).

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/autotest/gcore/vsis3.py
+++ b/autotest/gcore/vsis3.py
@@ -2300,7 +2300,7 @@ aws_secret_access_key = bar
 # a non default profile
 
 
-def test_vsis3_read_credentials_config_file_non_default_profile(tmpdir, monkeypatch):
+def test_vsis3_read_credentials_config_file_non_default_profile(tmpdir):
 
     if gdaltest.webserver_port == 0:
         pytest.skip()
@@ -2311,10 +2311,6 @@ def test_vsis3_read_credentials_config_file_non_default_profile(tmpdir, monkeypa
     gdal.SetConfigOption('AWS_CONFIG_FILE', None)
     gdal.SetConfigOption('AWS_DEFAULT_PROFILE', 'myprofile')
 
-    if sys.platform == 'win32':
-        monkeypatch.setenv('USERPROFILE', str(tmpdir))
-    else:
-        monkeypatch.setenv('HOME', str(tmpdir))
     os_aws = tmpdir.mkdir(".aws")
 
     gdal.VSICurlClearCache()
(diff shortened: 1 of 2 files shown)


autotest/gcore/vsis3.py
```diff
@@ -2300,21 +2300,26 @@ def test_vsis3_read_credentials_config_file():
...
     gdal.SetConfigOption('AWS_ACCESS_KEY_ID', '')
-
-    gdal.SetConfigOption('CPL_AWS_CREDENTIALS_FILE', '/vsimem/aws_credentials')
-    gdal.SetConfigOption('AWS_CONFIG_FILE', '/vsimem/aws_config')
+    gdal.SetConfigOption('CPL_AWS_CREDENTIALS_FILE', None)
+    gdal.SetConfigOption('AWS_CONFIG_FILE', None)
     gdal.SetConfigOption('AWS_DEFAULT_PROFILE', 'myprofile')
 
+    if sys.platform == 'win32':
+        monkeypatch.setenv('USERPROFILE', str(tmpdir))
```
rouault: there's a test failure on Windows. I'd suspect (but didn't check) that monkeypatch.setenv() on Windows doesn't manage to alter the native C environmennt, and just modifies the Python os.environ. Perhaps you should just use gdal.SetConfigOption(). Actually you can use ```with gdaltest.config_option(key, value):``` that will automatically save & restore the original value

glostis: I didn't have a Windows platform to test on so indeed it was a shot in the dark...
I didn't know about `with gdaltest.config_option`, I will try it out, thanks!

glostis: Done in 5d663ea

glostis: The AppVeyor builds seem to be stuck in a queue, but they turned green on my fork: https://ci.appveyor.com/project/glostis/gdal/builds/30488889

rouault: There are OSGeo/PROJ builds being processed: https://ci.appveyor.com/project/OSGeo/proj/history . So will eventually turn green here too once they and other PR are finished. Good news that it fixes your builds.

glostis: I didn't know the build limitation was organisation-wise, and not repo-wise.
Thanks!

rouault:
Cherry-picked to 3.0

PR #2211, opened 2020-02-02
Update links for SQL dialects

simonokeefe:
Existing links are invalid.


rouault:
Thanks, but the links you proposed will not work They should be prefix with :ref: (column ref column) and there should be a single trailing underscore.
Like
```
:ref:`OGR SQL <ogr_sql_dialect>`_
```

simonokeefe:
Thanks for the feedback. I've now updated the links based on the formatting you provided.

rouault:
hum actually I was wrong. The trailing _ had to be deleted as well. I've rectified this myself. That way I should remember it now :-)

PR #2218, opened 2020-02-04
MSG driver doc: Update link (fixes #2205)

### LINKED ISSUE #2205: Broken link for MSG driver & opportunity to integrate underlying library more seamlessly

https://gdal.org/drivers/raster/msg.html states that the Public Wavelet Transform Decompression Library software is licensed and free to download. This has now been updated:
1. The library is found under a [a new link](https://gitlab.eumetsat.int/open-source/PublicDecompWT)
1. It's got a FOSS license so it could be integrated directly into GDAL

That means that the link in the GDAL docs is broken, and also that there is the opportunity to pull in the library automatically. As it stands users can no longer download the library from the referenced link, so at minimum the link should be updated but I think we can do better than that so that users can install GDAL with MSG support without having to do gymnastics to acquire the EUMETSAT library.

I'm happy to update the doc. Is anybody interested in integrating the library now more tightly into GDAL? I'm not sure I'll be able to do that any time soon due to time constraints.

Try to install the MSG driver using the existing instructions ;)

## Operating system

All.

Latest

erget:
## Purpose
This PR fixes the link and install instructions for the MSG driver using a library by EUMETSAT. The library in question is now available at a new address, under a FOSS license, and as a git repo and thus is cloneable.

## Related issue
#2205 is related.

## Tasklist
- [ ] Review
- [ ] Adjust for comments

## Environment

This is not specific to any environment.

### DIFF
--- a/gdal/doc/source/drivers/raster/msg.rst
+++ b/gdal/doc/source/drivers/raster/msg.rst
@@ -37,6 +37,10 @@ Build Instructions
 
 Clone the EUMETSAT library for wavelet decompression into ``frmts/msg``.
 
+If you are building with Visual Studio 6.0, extract the .vc makefiles
+for the PublicDecompWT from the file `PublicDecompWTMakefiles.zip`
+stored in that directory.
+
 If you build using the GNUMakefile, use *--with-msg* option to enable
 MSG driver:
 
--- a/gdal/doc/source/drivers/raster/msg.rst
+++ b/gdal/doc/source/drivers/raster/msg.rst
@@ -10,23 +10,18 @@ MSG -- Meteosat Second Generation
 
 This driver implements reading support for Meteosat Second Generation
 files. These are files with names like
-H-000-MSG1\_\_-MSG1\_\_\_\_\_\_\_\_-HRV\_\_\_\_\_\_-000007\_\_\_-200405311115-C\_, commonly
-distributed into a folder structure with dates (e.g. 2004\05\31 for the
+``H-000-MSG1\_\_-MSG1\_\_\_\_\_\_\_\_-HRV\_\_\_\_\_\_-000007\_\_\_-200405311115-C\_``, commonly
+distributed into a folder structure with dates (e.g. ``2004\05\31`` for the
 file mentioned).
 
 The MSG files are wavelet-compressed. A decompression library licensed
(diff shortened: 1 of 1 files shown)


gdal/doc/source/drivers/raster/msg.rst
```diff
@@ -40,12 +35,7 @@ Driver capabilities
 Build Instructions
 ------------------
 
-Download the Eumetsat library for wavelet decompression. This is a file
-named PublicDecompWT.zip. Extract the content in a subdirectory with the
-same name (under frmts/msg).
-
-If you are building with Visual Studio 6.0, extract the .vc makefiles
```
rouault: This part about .vc files is probably still needed. GDAL's frmts/msg/makefile.vc expects makefile.vc files in PublicDecompWT, not the Makefile files (now?) provided by https://gitlab.eumetsat.int/open-source/PublicDecompWT. That said, I've never tested myself Windows builds with this driver

erget:
The AppVeyor build is failing but I'm honestly not sure this is related to this PR - it seems the one failing is using MS VS 2017, and my best guess is that it's timing out because it executes for exactly 1 hour and doesn't seem to throw an error.

gdal/doc/source/drivers/raster/msg.rst
rouault: Why did you mark this comment as resolved ? The part about extracting the .vc makefiles is still deleted

rouault:
> The AppVeyor build is failing but I'm honestly not sure this is related to this PR

yes. random timeout

gdal/doc/source/drivers/raster/msg.rst
erget: Sorry, it looks like my comment that I thought I'd submitted somehow isn't here. We no longer offer the VC files - they were outdated, untested, and probably incomplete. Support for Cygwin is there but no Visual Studio.

rouault: > We no longer offer the VC files

ok, but GDAL git does contains a zip with makefile.vc files: https://github.com/OSGeo/gdal/blob/master/gdal/frmts/msg/PublicDecompWTMakefiles.zip. This is what was mentionned here. No idea if it is still functional, but in doubt, I'd let the old comment, perhaps with a note that it hasn't been tested lately

PR #2223, opened 2020-02-06
Travis CSA: update to Xenial + CLang 9

rouault:
I also found that the check of the CSA log was non functional before...

### DIFF
--- a/gdal/frmts/hdf5/bagdataset.cpp
+++ b/gdal/frmts/hdf5/bagdataset.cpp
@@ -41,6 +41,7 @@
 #include "ogr_core.h"
 #include "ogr_spatialref.h"
 
+#include <cassert>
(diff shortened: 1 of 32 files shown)


gdal/frmts/hdf5/bagdataset.cpp
```diff
@@ -900,6 +901,7 @@ CPLErr BAGResampledBand::IReadBlock( int nBlockXOff, int nBlockYOff,
     std::vector<int> counts;
     if( poGDS->m_bMask )
     {
+        assert(pImage); // to make CLang Static Analyzer happy
```
mloskot: Why not `CPLAssert`?

rouault: Done

PR #2229, opened 2020-02-09
RMF: Read vertical datum info

drons:


drons:
@rouault @BishopGIS 

I'm needed to read vertical datum from RMF files. 

Reading datum codes has already been done now.

But the next step is to translate the code into EPSG. Translation table already exists in gdal/ogr/ogrsf_frmts/sxf/ogrsxfdatasource.cpp and used at OGRSXFDataSource::SetVertCS.

I think it’s best to move the translation table to ogr_srs_panorama.cpp and add access via importVertCSFromPanorama or something like that. And use new function in both cases.

What do you think?

rouault:
> What do you think?

Sounds good

gdal/ogr/ogr_srs_panorama.cpp
```diff
@@ -150,6 +150,44 @@ constexpr int aoEllips[] =
 
 constexpr int NUMBER_OF_ELLIPSOIDS = static_cast<int>(CPL_ARRAYSIZE(aoEllips));
 
+/************************************************************************/
+/*  Correspondence between "Panorama" and EPSG vertical CS.             */
+/************************************************************************/
+
+constexpr long aoVCS[] =
```
rouault: we could probably just use ``int`` everywhere ``long`` is used

drons: No problem.

Here I tried to make minimal changes compared to ogrsxfdatasource.

drons:
I'm think I finished.

rouault:
What about long->int changes ? I'm not sure there's value in having OSRImportVertCSFromPanorama() in the C API either. This is essentially material for driver internal use.

drons:
Thanks!

PR #2230, opened 2020-02-10
Shape: expose .dbf and .cpg source encodings in the SHAPEFILE metadata domain

rouault:
Should help for use case raised by https://github.com/qgis/QGIS/pull/34381

### DIFF
--- a/gdal/ogr/ogrsf_frmts/shape/ogrshapelayer.cpp
+++ b/gdal/ogr/ogrsf_frmts/shape/ogrshapelayer.cpp
@@ -420,6 +420,7 @@ static CPLString GetEncodingFromLDIDNumber(int nLDID)
 
 static CPLString GetEncodingFromCPG( const char* pszCPG )
 {
+    // see https://support.esri.com/en/technical-article/000013192
     CPLString osEncodingFromCPG;
     const int nCPG = atoi(pszCPG);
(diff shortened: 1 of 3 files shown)


gdal/ogr/ogrsf_frmts/shape/ogrshapelayer.cpp
```diff
@@ -347,110 +455,40 @@ CPLString OGRShapeLayer::ConvertCodePage( const char *pszCodePage )
...
-
-        if( nCP != -1 )
-        {
-            l_osEncoding.Printf( "CP%d", nCP );
-            return l_osEncoding;
-        }
-    }
-
-    // From the CPG file
-    // http://resources.arcgis.com/fr/content/kbase?fa=articleShow&d=21106
```
nyalldawson: I spent some time trying to re-located this article, and I believe it may be this one: https://support.esri.com/en/technical-article/000013192
(if so, we probably should also handle the string "UTF8" as an alias for "UTF-8")

rouault: > https://support.esri.com/en/technical-article/000013192

Yep, that was it from my memories.

> also handle the string "UTF8" as an alias for "UTF-8"

Done

nyalldawson:
LGTM

PR #2233, opened 2020-02-11
drop autogenerated files

metux:
These files need to be regenerated anyways, so there's really no
point in keeping them in the git repo.

Signed-off-by: Enrico Weigelt, metux IT consult <info@metux.net>

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:


rouault:
This will break compilation on the CI targets and would require an update in the release procedure

PR #2234, opened 2020-02-11
Export CSVIngest() to allow reading CSV file line-by-line with CSVGet…

sergey-suloev:


rouault:
I'd prefer not exposing CSVIngest() Can't it be called by CSVGetNextLine() instead ?

sergey-suloev:
I wouldn't go into deep changes, I just want to make it work.

PR #2263, opened 2020-02-21
Fix incorrect identification of MapInfo GDA2020 CRS

### LINKED ISSUE #2259: Incorrect identification of mapinfo CRS
Looks very similar to https://github.com/OSGeo/gdal/commit/06f6ddea060b044063c610ee0b0027bb85c8a846#diff-3b9d6af62696ff5f5e459dc321575431 -- but the attached files in the v17 folder should be in EPSG:7855, yet the identified projection is incorrect:

"PARCEL_VIEW.tab" : identified as boundcrs of 7855
"PARCEL_VIEW.mif": identified as EPSG:32755

Possibly due to
```
 CoordSys Earth Projection 8, 1028, "m", 147, 0, 0.9996, 500000, 10000000 Bounds (-500000, 5000000) (1500000, 7000000) 
```
datum 1028 may need to be added as a reference to GDA2020

(The copies of the files in the v16 folder are correctly identified by gdal)

[TestData.zip](https://github.com/OSGeo/gdal/files/4229150/TestData.zip)

nyalldawson:
Fixes #2259

### DIFF
--- a/gdal/ogr/ogrsf_frmts/mitab/mitab_spatialref.cpp
+++ b/gdal/ogr/ogrsf_frmts/mitab/mitab_spatialref.cpp
@@ -57,6 +57,7 @@ CPL_CVSID("$Id$")
 extern const MapInfoDatumInfo asDatumInfoList[];
 extern const MapInfoSpheroidInfo asSpheroidInfoList[];
 
+/* EPSG code, MapInfo datum ID (or 9999), OGC Name, datum parameters... */
 const MapInfoDatumInfo asDatumInfoList[] =
 {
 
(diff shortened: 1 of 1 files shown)


gdal/ogr/ogrsf_frmts/mitab/mitab_spatialref.cpp
```diff
@@ -260,6 +260,8 @@ const MapInfoDatumInfo asDatumInfoList[] =
 { 0,    9999,"Bosnia-Herzegovina",         10, 472.8677, 187.8769, 544.7084, -5.76198422, -5.3222842, 12.80666941, 1.54517287, 0 },
 { 6181, 9999,"Luxembourg 1930 / Gauss",     4, -192.986, 13.673, -39.309, 0.4099, 2.9332, -2.6881, 0.43, 0 },
 { 1168, 9999,"Geocentric Datum of Australia 2020", 0,-0.06155, 0.01087, 0.04019, 0.0394924, 0.0327221, 0.0328979, 0.009994,0 },
+{ 0,    1028,"Geocentric Datum of Australia 2020", 0,-0.06155, 0.01087, 0.04019, 0.0394924, 0.0327221, 0.0328979, 0.009994,0 },
```
rouault: First column should be 1168 as well (as this is the EPSG code of the datum)
It seems to me that this second mapping is likely more idiomatic than the first one (given that it has a dedicated mapinfo datum id = 1028). And thus on the writing side, it should probably be used. Given the logic at line 1993 that selects the first match, those 2 lines should probably be inverted.

nyalldawson: Thanks @rouault, fixed

rouault:
Backported to 3.0 branch

PR #2265, opened 2020-02-23
Python: Removed calls to imp module (fixes #2264)

### LINKED ISSUE #2264: Python 3.8 imp module deprecation warning
When starting up QGIS on Linux, with Python 3.8 and GDAL 3.0.4, I get a deprecation warning:

             2020-02-23T10:23:22     WARNING    warning:/usr/lib/python3.8/site-packages/qgis/utils.py:744: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
              mod = _builtin_import(name, globals, locals, fromlist, level)
             
             traceback: File "", line 1, in 

<... snip long stack trace ...>

              File "", line 783, in exec_module
              File "", line 219, in _call_with_frames_removed
              File "/usr/lib/python3.8/site-packages/osgeo/__init__.py", line 41, in 
              _gdal = swig_import_helper()
              File "/usr/lib/python3.8/site-packages/osgeo/__init__.py", line 15, in swig_import_helper
              import imp
              File "/usr/lib/python3.8/site-packages/qgis/utils.py", line 744, in _import
              mod = _builtin_import(name, globals, locals, fromlist, level)
              File "", line 991, in _find_and_load
              File "", line 975, in _find_and_load_unlocked
              File "", line 671, in _load_unlocked
              File "", line 783, in exec_module
              File "", line 219, in _call_with_frames_removed
              File "/usr/lib/python3.8/imp.py", line 31, in 
              warnings.warn("the imp module is deprecated in favour of importlib; "

There should be no warning; non-deprecated python libraries should be used.

Run QGIS-3.10 against GDAL-3.0.4 on Linux

## Operating system

Gentoo Linux x86-64

GDAL-3.0.4 from Gentoo ebuild repository; also from github.

espinafre:
Remove deprecated reference to Python imp module, which caused a startup
warning on QGIS. Replaced those references with a call to importlib. Fixes #2264.

### DIFF
--- a/gdal/swig/python/osgeo/__init__.py
+++ b/gdal/swig/python/osgeo/__init__.py
@@ -12,16 +12,16 @@ if version_info >= (3, 8, 0) and platform == 'win32':
 if version_info >= (2, 6, 0):
     def swig_import_helper():
         from os.path import dirname
-        import imp
-        fp = None
+        import importlib.machinery
+        spec = None
         try:
-            fp, pathname, description = imp.find_module('_gdal', [dirname(__file__)])
+            spec = importlib.machinery.PathFinder().find_spec('_gdal', [dirname(__file__)])
         except ImportError:
             import _gdal
             return _gdal
-        if fp is not None:
+        if spec is not None:
             try:
-                _mod = imp.load_module('_gdal', fp, pathname, description)
+                _mod = spec.loader.load_module()
             except ImportError as e:
                 if version_info >= (3, 8, 0) and platform == 'win32':
                     import os
@@ -35,8 +35,6 @@ if version_info >= (2, 6, 0):
                         traceback_string = ''.join(traceback.format_exception(*sys.exc_info()))
                         raise ImportError(traceback_string + '\n' + msg)
                 raise
-            finally:
-                fp.close()
             return _mod
     _gdal = swig_import_helper()
     del swig_import_helper
--- a/gdal/swig/python/osgeo/__init__.py
+++ b/gdal/swig/python/osgeo/__init__.py
@@ -9,19 +9,41 @@ if version_info >= (3, 8, 0) and platform == 'win32':
             if p:
                 os.add_dll_directory(p)
 
-if version_info >= (2, 6, 0):
+if version_info >= (2, 7, 0):
+    def swig_import_helper():
+        import importlib
+        from os.path import dirname, basename
+        mname = basename(dirname(__file__)) + '._gdal'
+        try:
+            return importlib.import_module(mname)
+        except ImportError as e:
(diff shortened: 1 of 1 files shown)


mloskot:
This may not be enough, see my https://github.com/OSGeo/gdal/issues/2264#issuecomment-590070338

https://github.com/OSGeo/gdal/blob/b309b2f9b4b2fe3551a58be94a6914b7f135ac78/gdal/swig/python/osgeo/gdal.py#L1-L2

https://github.com/OSGeo/gdal/blob/b309b2f9b4b2fe3551a58be94a6914b7f135ac78/gdal/swig/python/osgeo/gdal.py#L20-L22

https://github.com/OSGeo/gdal/blob/b309b2f9b4b2fe3551a58be94a6914b7f135ac78/gdal/swig/python/osgeo/gdalconst.py#L20-L22

etc.

espinafre:
Well, the file in question (gdal/swig/python/osgeo/__init__.py) exists in the repository, and it is not marked as generated by swig. I understand that the generators should be updated and this might not solve the issue once and for all, but it at least *does* shut up python warnings on loading the gdal module in QGIS.

rouault:
@espinafre This file is indeed not generated. Your changes break apparently non-Python 3.8 configurations. The new code path should fallback on the old/existing one when import importlib.machinery fails.

espinafre:
@rouault @mloskot Thank you very much for enlightening me. I understand that this only works for Python version 3.4+ (https://docs.python.org/3/library/importlib.html#importlib.machinery.ModuleSpec). Both AppVeyor and Travis are still using Python 2.7. Is there a roadmap somewhere that documents supported Python versions for GDAL? Will newer releases of GDAL still support Python 2.7? The only document I could find about this is https://gdal.org/development/rfc/rfc8_devguide.html , which mentions that for Python code, we should support both 2 and 3, but I belive that, with the demise of Python 2 it is not valid any longer.
If there are such guidelines, I may be able to try and fix this according to them.

rouault:
We'll ultimately drop python 2.7, but for now, I'd prefer if we can still make things work with it.

espinafre:
Let's see what Travis and AppVeyor think about Python 2 compatibility with this. With Python 3.8, at least, this works as intended. @mloskot, the other generated files (gdal.py, gdal_array.py, gdalconst.py, gnm.py, ogr.py and osr.py) don't result in import warning/error either, because they test the Python version and use the correct API accordingly.

mloskot:
@espinafre Right, I obviously missed that. Thanks for the further investigation!

espinafre:
Well, both CI systems liked this. @rouault , what do you think? Working with newer and older Python versions, and without specific API calls. Any good? Thanks.

Issue #2273, opened 2020-02-26
gdal2tiles.py multiprocessing chunk size


kordian-kowalski:
https://github.com/OSGeo/gdal/blob/7f2dcdc930b51f6cf5f013cd4d5ef66ff094239f/gdal/swig/python/scripts/gdal2tiles.py#L2917

This is only valid when using >1 thread for tiling.

Batching the tiling jobs does improve performance, but setting the batch size too high will result in not all threads being utilized fully or at all.

There are 2 factors here:
1. A file may not be big enough to fill multiple 128-tile chunks
2. Near the end of the processing, not all chunks finish at the same time so the thread with the slower chunk will continue processing by itself while it may still have tens of tiles to process, which could be done by other threads.

I tried manually editing the chunk size in the gdal2tiles script.
Testing a 464MB geotiff file, tiling at zoom level 20, 2-core 4-thread CPU:
`gdal2tiles.py --zoom=20 --resampling=bilinear -w none -n --processes=4 -x medium.tif medium`

Only testing with a single zoom level, because lower zoom levels don't use multithreading at all.

Chunk size = 1 : 1m 15s
Chunk size = 2 : 1m 04s
Chunk size = 3 : 1m
Chunk size = 4 : 1m
Chunk size = 8 : 1m
Chunk size = 16 : 1m 04s
Chunk size = 32 : 1m 03s (using 3 cores)
Chunk size = 64 : 1m 14s (using 2 cores)
Chunk size = 128 (current value) : 1m 49s (using 1 core)

Exact values may vary heavily depending on the file, but I believe the differences will be similar in magnitude.

I think the value could be changed to ~4 which seems to be a sweet spot.

PR #2274, opened 2020-02-26
gdal2tiles.py: Lowered multithreading job chunk size to 4

kordian-kowalski:
PR for issue:
https://github.com/OSGeo/gdal/issues/2273

### DIFF
--- a/gdal/swig/python/scripts/gdal2tiles.py
+++ b/gdal/swig/python/scripts/gdal2tiles.py
@@ -2914,7 +2914,7 @@ def multi_threaded_tiling(input_file, output_folder, options):
 
     # TODO: gbataille - check the confs for which each element is an array... one useless level?
     # TODO: gbataille - assign an ID to each job for print in verbose mode "ReadRaster Extent ..."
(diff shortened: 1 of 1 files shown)


rouault:
@gbataille Any opinion on this ?

gbataille:
Agree that the hardcoded 128 might be non optimal.
I guess it depends on the total number of chunks and the thread count.

@kordian-kowalski would it make sense to have the chunk size computed based on the size of the list and the number of threads the user has chosen?
With maybe a max so as not to pipe too large chunks to the executors?

@rouault on principle yes 128 is certainly not the best value in all cases, but I doubt 4 is either. It depends. a "smart" value based on context is certainly best in the end

kordian-kowalski:
It's true that my test is not very exhaustive and a value of 4 is specific to my case.

To generalize, I believe the optimal batch size will depend on how expensive it is to compute a single tile versus the cost of dispatching a batch of tasks to a worker. I'm not sure how accurately this can be estimated. I could run some more tests and see what the optimal value is for different kinds of source files, maybe it doesn't differ by much

PR #2277, opened 2020-02-27
FlatGeobuf: disable most CPLDebug() for non-DEBUG builds

rouault:
@bjornharrtell I find the Flatgeobuf driver super verbose. This PR disables most CPLDebug() for non-DEBUG builds as they are presumably only useful for the developer, and not for users, and even if they are not displayed they still burn a bit of CPU to figure they must not be displayed. I've just kept a few instances where there was debug messages in CPLError() cases.

### DIFF
--- a/gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
+++ b/gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
@@ -85,7 +85,7 @@ OGRFlatGeobufLayer::OGRFlatGeobufLayer(
         m_sExtent.MaxY = (*envelope)[3];
     }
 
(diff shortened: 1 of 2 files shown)


bjornharrtell:
@rouault makes sense!

Does it not also make sense to introduce a macro that can possibly deprecate direct use of CPLDebug so that one does not have to do the #ifdef DEBUG dance on each use?

rouault:
> Does it not also make sense to introduce a macro that can possibly deprecate direct use of CPLDebug

you're right. I had in mind that variadic macros weren't standardized in C++, but according to https://en.wikipedia.org/wiki/Variadic_macro they are since C++11, so we're all good. I'm adding a CPLDebugOnly(...) variadic macro that expands to CPLDebug for DEBUG builds and empty for non-DEBUG builds.

bjornharrtell:
Very nice! :)

PR #2280, opened 2020-02-28
FlatGeoBuf: implement Binary, Bool, Short and Float field types

rouault:
And also remove CreateGeomField capability, that should only be
declared if OGRLayer::CreateGeomField() is implemented (for several
geometry fields per feature)

CC @bjornharrtell


rouault:
@bjornharrtell I've also added support for Bool, Short and Float. I presume Bool is encoded on a single byte ?

bjornharrtell:
@rouault nice, and yes bool is represented by a byte.

PR #2281, opened 2020-02-28
Python: Add numpy as extras_require dependency (fixes #2158)

### LINKED ISSUE #2158: Python bindings to gdal should make use of extras feature for packaging.

`numpy` is an optional package for the python `gdal` bindings. It would be nice if we could make use of the [extras](https://www.python.org/dev/peps/pep-0508/#extras) feature when packaging, so that you could have something like
```
gdal[numpy]
```
in a `requirements.txt` file.
Otherwise, the order of the packages you install becomes important, leading to problems like those encountered by users as described [here](https://stackoverflow.com/questions/45289299/python-3-4-importerror-no-module-named-gdal-arrayno-module-named-gdal-arra) and [here](https://gis.stackexchange.com/questions/83138/cannot-import-gdal-array/83299). 

If you install `gdal` followed by `numpy` then you don't get the `numpy` features of `gdal`, and you have to work hard to correct this, e.g.,:
```
pip install gdal==2.4.2
pip install numpy
python -c 'from osgeo import gdal_array'
  File "/home/robert/miniconda3/envs/gdal/lib/python3.7/site-packages/osgeo/gdal_array.py", line 18, in swig_import_helper
    fp, pathname, description = imp.find_module('_gdal_array', [dirname(__file__)])
  File "/home/robert/miniconda3/envs/gdal/lib/python3.7/imp.py", line 296, in find_module
    raise ImportError(_ERR_MSG.format(name), name=name)
ImportError: No module named '_gdal_array'

install gdal==2.4.2 
Requirement already satisfied: gdal==2.4.2 in ./miniconda3/envs/gdal/lib/python3.7/site-packages (2.4.2)

pip install --force-reinstall gdal==2.4.2 --no-cache-dir
python -c 'from osgeo import gdal_array'
```

What would be nice would be if you could do:

```
pip install gdal[numpy]==2.4.2
```
This current behaviour me a problem when using `pip-tools`, which parses a `requirements.in` file to produce a `requirements.txt` file, and one step involves an alphabetical sorting. Because `gdal` comes before `numpy`. See [Issue #992](https://github.com/jazzband/pip-tools/issues/992)

## Operating system

```
lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 18.04.3 LTS
Release:        18.04
Codename:       bionic
```

```
$ dpkg -l libgdal-dev
Desired=Unknown/Install/Remove/Purge/Hold
| Status=Not/Inst/Conf-files/Unpacked/halF-conf/Half-inst/trig-aWait/Trig-pend
|/ Err?=(none)/Reinst-required (Status,Err: uppercase=bad)
||/ Name                                         Version                     Architecture                Description
+++-============================================-===========================-===========================-==============================================================================================
ii  libgdal-dev                                  2.4.2+dfsg-1~bionic0        amd64                       Geospatial Data Abstraction Library - Development files
```

vphilippon:
fixes #2158
Doing `pip install gdal[numpy]` will install `gdal` with `numpy > 1.0.0`.
`pip` will ensure that `numpy` is installed first so that it's detected
and the releveant features/bindings/etc. are compiled and made available
for users.

FYI I did not touch how `numpy` was installed in the various CI scripts, being a bit afraid of breaking things and/or making a mess.

### DIFF
--- a/gdal/swig/python/setup.py
+++ b/gdal/swig/python/setup.py
@@ -463,7 +463,8 @@ setup_kwargs = dict(
     data_files=data_files,
     ext_modules=ext_modules,
     scripts=glob('scripts/*.py'),
-    cmdclass={'build_ext': gdal_ext}
+    cmdclass={'build_ext': gdal_ext},
+    extras_require={'numpy': ['numpy > 1.0.0']},
 )
 
 # This section can be greatly simplified with python >= 3.5 using **


gdal/swig/python/setup.py
```diff
@@ -463,7 +463,8 @@ def parallel_build_extensions(self):
     data_files=data_files,
     ext_modules=ext_modules,
     scripts=glob('scripts/*.py'),
-    cmdclass={'build_ext': gdal_ext}
+    cmdclass={'build_ext': gdal_ext},
+    extras_require={'numpy': ['numpy > 1.0.0']},
```
rouault: shouldn't that be conditionalized by if HAVE_NUMPY: ?

vphilippon: No, it's the opposite: using `pip install gdal[numpy]` will install `numpy`, so that `HAVE_NUMPY` will be true when building/installing `gdal`.

PR #2283, opened 2020-02-29
FlatGeobuf: streamed writing and two other small improvements

rouault:
-  FlatGeobuf: allow creating a file in write-only mode (/vsizip/, /vsis3/ etc), and delete temporary file ASAP on Unix
- FlatGeobuf: advertize OLCRandomRead when index is present
-  FlatGeobuf: do not read index if the extent of the spatial filter is larger than the layer extent

CC @bjornharrtell


bjornharrtell:
Looks fine but I don't understand "allow creating a file in write-only mode" - didn't that work before and if not what change makes it work?

rouault:
> didn't that work before and if not what change makes it work?

write-only mode = "wb"  vs read-write mode = "w+b".  /vsizip/ only supports "wb" (you can't read while writing).

Issue #2282, opened 2020-02-29
FlatGeoBuf: spatial filter doesn't always work


rouault:
- Download zipped geopackage at https://filetransfer.io/data-package/VVKkoE4D and unzip it
- ogr2ogr grid_germany.fgb grid_germany.gpkg
- test_ogrsf -ro grid_germany.fgb

Following error is emitted:
ERROR: Spatial filter (0) eliminated a feature unexpectedly!

The .fgb is ~ 700 MB large

CC @bjornharrtell 

Looking a bit at the code, the logic in parseFeature() to decide whether to do a VSIFSeekL() or not seemed a bit suspicious, but even when doing unconditionnaly the VSIFSeekL(), the error is still there

rouault:
Actually can be reproduced on a much smaller file created with:
$ ogr2ogr small.fgb grid_germany.fgb -spat 4200000 3000000 4202000 3002000

Only 40 features in it: [small.fgb.zip](https://github.com/OSGeo/gdal/files/4270129/small.fgb.zip)

rouault:
It seems related to an issue with FID stability:
```
$ ogrinfo small.fgb -spat 4200405 3000405 4200410 3000410 -al -q

Layer name: GPKG demo
OGRFeature(GPKG demo):42
  gitter_id_100m (String) = 100mN30004E42004
  POLYGON ((4200400 3000400,4200400 3000500,4200500 3000500,4200500 3000400,4200400 3000400))
```
Note the FID 42, but the layer has only 40 features (FID 0 to 39)
```
$ ogrinfo small.fgb -fid 42 -al -q

Layer name: GPKG demo
ERROR 1: Requested feature id is out of bounds
Unable to locate feature id 42 on this layer.
```

I guess that FlatGeobuf should include a FID special attribute in a Feature

bjornharrtell:
Hmm this is worrying. The FID should be the index position in the leaf nodes in the index (specifially pos - 1 at https://github.com/OSGeo/gdal/blob/e5f3dfe78b0f1ac124173f2fc6b014d7fdaeec13/gdal/ogr/ogrsf_frmts/flatgeobuf/packedrtree.cpp#L321). Will try to find out what is going on...

rouault:
Fixed per https://github.com/OSGeo/gdal/pull/2284

PR #2284, opened 2020-02-29
FlatGeobuf: Fix FID stability on spatial searches

bjornharrtell:
ref https://github.com/OSGeo/gdal/issues/2282

I believe this is will resolve the issue, but should find a testcase to verify.


rouault:
That did it ! This fixes the test_ogrsf error spotted on https://github.com/OSGeo/gdal/issues/2282 on the big and the small file. You could potentially grab small.fgb from #2282 (it's 9 KB large) and run test_ogrsf on it, or do a SetSpatialFilterRect(4200405, 3000405, 4200410, 3000410) and check that the only feature returned as FID = 39

bjornharrtell:
With https://github.com/bjornharrtell/flatgeobuf/commit/88e7427317f7f867b525b15e7d01c6d145db5e53 the fix is extensively tested in upstream and as `packedrtree.cpp` is from upstream I would suggest merging this as is.

rouault:
> I would suggest merging this as is.

makes sense

PR #2285, opened 2020-02-29
FlatGeoBuf: read all data types in the closest OGR one

rouault:


rouault:
CC @bjornharrtell  I generated the test file with my favorite hexadecimal editor for the data types that OGR cannot write. I presume encoding of the JSON data type is the same as string : 4 bytes of length, followed by the string ?

bjornharrtell:
The spec is unfortunately lacking in the details of data type value encoding but your presumtion should be correct.

PR #2289, opened 2020-03-02
Flatgeobuf write performance improvements

### LINKED ISSUE #2287: FlatGeoBuf: creating a large file can be super slow with a rotational disk
I've attempted "ogr2ogr france_osm_points.fgb france-latest.osm.pbf points" (where france-latest.osm.pbf is a OSM extract from a couple years. File size is 3.6 GB). The conversion itself runs in a few minutes to 100%, but the writing to the final file with the Hilbert sorting was super slow. It would have taken hours at the rate the file was written. I've a slow rotational disk, which might be the reason. And/or perhaps due to use of FILE* buffered I/O (but I presume the disk is the major factor) The final size is 1.3 GB.

With the following patch that adds a OGR_FLATGEOBUF_VSIMEM_TEMP_FILE=YES configuration option to create the temporary file in /vsimem/, it is a matter of a few tens of seconds to write the final file
```
diff --git a/gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobufdataset.cpp b/gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobufdataset.cpp
index 14a53e7..0299436 100644
--- a/gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobufdataset.cpp
+++ b/gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobufdataset.cpp
@@ -438,7 +438,9 @@ OGRLayer* OGRFlatGeobufDataset::ICreateLayer( const char *pszLayerName,
         CPLDebug("FlatGeobuf", "Spatial index requested will write to temp file and do second pass on close");
         const CPLString osDirname(CPLGetPath(osFilename.c_str()));
         const CPLString osBasename(CPLGetBasename(osFilename.c_str()));
-        osTempFile = (STARTS_WITH(osFilename, "/vsi") &&
+        osTempFile = CPLTestBool(CPLGetConfigOption("OGR_FLATGEOBUF_VSIMEM_TEMP_FILE", "NO")) ?
+            CPLSPrintf("/vsimem/tempflatgeobuf/%p", this) :
+                      (STARTS_WITH(osFilename, "/vsi") &&
                       !STARTS_WITH(osFilename, "/vsimem/")) ?
             CPLGenerateTempFilename(osBasename) :
             CPLFormFilename(osDirname, osBasename, nullptr);
```

Besides that strategy, I guess an alternative strategy could be to allocate a buffer of let's say 100 MB and when re-writing the file blobs from the temporary file to the final, do that in batch. The main loop for (size_t i = 0; i < m_featuresCount; i++) { would have a sub-loop, where it would collect the (src_file_offset, target_file_offset) of features until 100 MB is reached. It would sort by increasing src_file_offset. Read that content in increasing offset order w.r.t the source temporary file. And then write that to the final file by increasing target_file_offset. Hopefully the above makes sense.

I'm not sure how much gain there would be from that, but I did something a bit similar in the past in the GeoPackage driver when generating the R-Tree: instead of doing a single "INSERT INTO layer_rtree SELECT [...] from layer", I iterate over 100 K features in the layer to grab their id and extent, and then insert that in the R-Tree. That saved a lot of seeking within the file.

CC @bjornharrtell

rouault:
Fixes #2287

- add a TEMPORARY_DIR layer creation option
- write final file in a batch way to minimize switches between reads and writes

In the example mentionned in #2287, conversion time is now:
- 4m24 when using -lco TEMPORARY_DIR=/vsimem/
- 5m15 without option (vs presumably several hours before)

CC @bjornharrtell

### DIFF
--- a/gdal/doc/source/drivers/vector/flatgeobuf.rst
+++ b/gdal/doc/source/drivers/vector/flatgeobuf.rst
@@ -51,6 +51,12 @@ Layer Creation Options
 
 -  **SPATIAL_INDEX=**\ *YES/NO*: Set to YES to create a
    spatial index. Defaults to YES.
+-  **TEMPORARY_DIR=**\ path: Path to an existing directory where temporary
+   files should be created. Only used if SPATIAL_INDEX=YES. If not specified,
+   the directory of the output file will be used for regular filenames. For
+   other VSI file systems, the temporary directory will be the one decided by
+   the :cpp:func:`CPLGenerateTempFilename` function.
+   "/vsimem/" can be used for in-memory temporary files.
 
 Examples
 --------
--- a/gdal/ogr/ogrsf_frmts/flatgeobuf/ogr_flatgeobuf.h
+++ b/gdal/ogr/ogrsf_frmts/flatgeobuf/ogr_flatgeobuf.h
@@ -102,6 +102,7 @@ class OGRFlatGeobufLayer final : public OGRLayer
         uint64_t m_writeOffset = 0; // current write offset
         uint16_t m_indexNodeSize = 0;
         std::string m_oTempFile; // holds generated temp file name for two pass writing
+        uint32_t m_maxFeatureSize  = 0;
 
         // shared
         GByte *m_featureBuf = nullptr; // reusable/resizable feature data buffer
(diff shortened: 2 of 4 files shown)


bjornharrtell:
Nice work @rouault, I have no objections.

PR #2295, opened 2020-03-04
gdal_edit.py: add -a_ulurll switch

riegl-aw:
Assign/override the georeferenced bounds of the dataset from three points: upper-left, upper-right and lower-left. Unlike option `-a_ullr`, this also supports rotated datasets (edges not parallel to coordinate system axes).

### DIFF
--- a/gdal/doc/source/programs/gdal_edit.rst
+++ b/gdal/doc/source/programs/gdal_edit.rst
@@ -15,7 +15,8 @@ Synopsis
 
 .. code-block::
 
-    gdal_edit [--help-general] [-ro] [-a_srs srs_def] [-a_ullr ulx uly lrx lry]
+    gdal_edit [--help-general] [-ro] [-a_srs srs_def]
+            [-a_ullr ulx uly lrx lry] [-a_ulurll ulx uly urx ury llx lly]
             [-tr xres yres] [-unsetgt] [-unsetrpc] [-a_nodata value] [-unsetnodata]
             [-unsetstats] [-stats] [-approx_stats]
             [-setstats min max mean stddev]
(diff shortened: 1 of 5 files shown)


gdal/doc/source/programs/gdal_edit.rst
```diff
@@ -64,6 +65,12 @@ It works only with raster formats that support update access to existing dataset
 
     Assign/override the georeferenced bounds of the dataset.
 
+.. option:: -a_ulurll ulx uly urx ury llx lly:
+
+    Assign/override the georeferenced bounds of the dataset from three points:
+    upper-left, upper-right and lower-left. Unlike :option:`-a_ullr`, this also
+    supports rotated datasets (edges not parallel to coordinate system axes).
+
```
rouault:
```suggestion

    .. versionadded:: 3.1
```

gdal/swig/python/scripts/gdal_edit.py
```diff
@@ -303,6 +323,11 @@ def gdal_edit(argv):
               uly, 0, (lry - uly) / ds.RasterYSize]
         ds.SetGeoTransform(gt)
 
+    if lly is not None:
```
rouault: To make it clear that lry and lly are exclusive:
```suggestion
    elif lly is not None:
```

riegl-aw:
Thanks for your suggestions, which I have just accepted.
Shall I merge the commits into one?

rouault:
> Shall I merge the commits into one?

Yes please

riegl-aw:
> > Shall I merge the commits into one?
> 
> Yes please

Done.

jratike80:
Wouln't it be a better test with a rotated image if that's the main reason for having the option?
So instead of `-a_ulurll 2 50 3 50 2 49` to use something like `-a_ulurll 2 50 3 50.2 2.1 49`

rouault:
> Wouln't it be a better test with a rotated image if that's the main reason for having the option?

good point Jukka :-)

riegl-aw:
> 
> 
> > Wouln't it be a better test with a rotated image if that's the main reason for having the option?
> 
> good point Jukka :-)

You are right - we will provide a better example soon.

riegl-aw:
Thanks!

Issue #2239, opened 2020-02-13
/vsiswift/: expiration of SWIFT_AUTH_TOKEN doesn't make RasterIO() to fail


talaj:
## Expected behavior and actual behavior.
When `SWIFT_AUTH_TOKEN` expire, swift begins to return 401.
This behavior is documented here:
https://docs.openstack.org/swift/queens/api/authentication.html

In such case I would expect `RasterIO()` call to fail so user can handle it somehow.
Instead the call silently returns empty data.

With debugging on, I can see
```
VSICURL: Got response_code=401
```

which comes from here:
https://github.com/OSGeo/gdal/blob/ef49c00611235df0c1ce4f51344f00567a668661/gdal/port/cpl_vsil_curl.cpp#L1335

I don't see any error handling in `VSICurlHandle::DownloadRegion`, it just returns empty string.

## Steps to reproduce the problem.

Either it is possible to wait a day to let the token expire, or it is possible to make a small change to code to take the token always from config map, when constructing headers. It is then possible to change the token to invalid one.

```diff
diff --git a/gdal/port/cpl_swift.cpp b/gdal/port/cpl_swift.cpp
index 45155ac5d3..30662b63a5 100644
--- a/gdal/port/cpl_swift.cpp
+++ b/gdal/port/cpl_swift.cpp
@@ -52,13 +52,14 @@ static CPLString g_osLastAuthToken;
 /************************************************************************/

 static
-struct curl_slist* GetSwiftHeaders( const CPLString& osAuthToken )
+struct curl_slist* GetSwiftHeaders( const CPLString& )
 {
     struct curl_slist *headers=nullptr;
     headers = curl_slist_append(
         headers, "Accept: application/json");
+    const char * osAuthToken = CPLGetConfigOption("SWIFT_AUTH_TOKEN", "");
     headers = curl_slist_append(
-        headers, CPLSPrintf("x-auth-token: %s", osAuthToken.c_str()));
+        headers, CPLSPrintf("x-auth-token: %s", osAuthToken));
     return headers;
 }
```
## Operating system

Debian Buster

## GDAL version and provenance

The latest master from github.

rouault:
A more appropriate / general fix would be to handle token expiration, and re-issue a request for a token for the case where the user doesn't provide SWIFT_AUTH_TOKEN, but uses SWIFT_AUTH_V1_URL + SWIFT_USER + SWIFT_KEY

talaj:
[...]

PR #2296, opened 2020-03-04
/vsiswift/: V3 authentication method, handling auth token expiration

talaj:
Addresses https://github.com/OSGeo/gdal/issues/2239.

Implements re-issue of the auth token when expired. To allow the re-issue, `GDAL_HTTP_MAX_RETRY` needs to be set at least to `1`.

Adds authentication using OpenStack Identity API v3. It might not be complete implementation. I have implemented it to the extent I was able to connect to a Swift instance I have access to.

### DIFF
--- a/gdal/doc/source/user/virtual_file_systems.rst
+++ b/gdal/doc/source/user/virtual_file_systems.rst
@@ -378,10 +378,20 @@ Recognized filenames are of the form :file:`/vsiswift/bucket/key` where ``bucket
 
 The generalities of :ref:`/vsicurl/ </vsicurl/>` apply.
 
-Two authentication methods are possible, and are attempted in the following order:
+Three authentication methods are possible, and are attempted in the following order:
 
 1. The :decl_configoption:`SWIFT_STORAGE_URL` and :decl_configoption:`SWIFT_AUTH_TOKEN` configuration options are set respectively to the storage URL (e.g http://127.0.0.1:12345/v1/AUTH_something) and the value of the x-auth-token authorization token.
 2. The :decl_configoption:`SWIFT_AUTH_V1_URL`, :decl_configoption:`SWIFT_USER` and :decl_configoption:`SWIFT_KEY` configuration options are set respectively to the endpoint of the Auth V1 authentication (e.g http://127.0.0.1:12345/auth/v1.0), the user name and the key/password. This authentication endpoint will be used to retrieve the storage URL and authorization token mentioned in the first authentication method.
+3. Authentication with Keystone v3 is using the same options as python-swiftclient, see https://docs.openstack.org/python-swiftclient/latest/cli/index.html#authentication for more details. GDAL (>= 3.1) supports the following options:
+
+   - `OS_IDENTITY_API_VERSION=3`
+   - `OS_AUTH_URL`
+   - `OS_USERNAME`
+   - `OS_PASSWORD`
+   - `OS_USER_DOMAIN_NAME`
+   - `OS_PROJECT_NAME`
+   - `OS_PROJECT_DOMAIN_NAME`
+   - `OS_REGION_NAME`
 
 This file system handler also allows sequential writing of files (no seeks or read operations are then allowed)
 
--- a/gdal/port/cpl_swift.cpp
+++ b/gdal/port/cpl_swift.cpp
@@ -29,6 +29,7 @@
 #include "cpl_vsi_error.h"
 #include "cpl_http.h"
 #include "cpl_multiproc.h"
+#include "cpl_json.h"
 
 // HOWTO setup a Docker-based SWIFT server:
 // https://github.com/MorrisJobke/docker-swift-onlyone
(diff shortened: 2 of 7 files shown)


gdal/port/cpl_swift.cpp
```diff
@@ -170,6 +184,258 @@ bool VSISwiftHandleHelper::GetConfiguration(CPLString& osStorageURL,
...
+                                  CPLString& osAuthToken)
+{
+    CPLString osUser = CPLGetConfigOption("SWIFT_USER", "");
+    CPLString osKey = CPLGetConfigOption("SWIFT_KEY", "");
+    CPLJSONObject postObject(CreateAuthV3RequestObject());
+    std::string post = postObject.Format(CPLJSONObject::PrettyFormat::Plain);
+
+    CPLString osAuthURL = CPLGetConfigOption("SWIFT_AUTH_V3_URL", "");
+    std::string url = osAuthURL;
+    if( url.back() != '/' )
```
rouault:
```suggestion
    if( !url.empty() && url.back() != '/' )
```

gdal/port/cpl_swift.cpp
```diff
@@ -106,33 +107,46 @@ bool VSISwiftHandleHelper::GetConfiguration(CPLString& osStorageURL,
         return true;
     }
 
-    CPLString osAuthURL = CPLGetConfigOption("SWIFT_AUTH_V1_URL", "");
-    CPLString osUser = CPLGetConfigOption("SWIFT_USER", "");
-    CPLString osKey = CPLGetConfigOption("SWIFT_KEY", "");
-    if( osAuthURL.empty() || osUser.empty() || osKey.empty() )
+    CPLString osAuthV3URL = CPLGetConfigOption("SWIFT_AUTH_V3_URL", "");
```
rouault: this new config option (and possible other new ones) should be documented in doc/source/user/virtual_file_systems.rst

gdal/port/cpl_swift.cpp
```diff
@@ -170,6 +184,258 @@ bool VSISwiftHandleHelper::GetConfiguration(CPLString& osStorageURL,
...
+/************************************************************************/
+
+CPLJSONObject VSISwiftHandleHelper::CreateAuthV3RequestObject()
+{
+    CPLString osUser = CPLGetConfigOption("SWIFT_USER", "");
+    CPLString osKey = CPLGetConfigOption("SWIFT_KEY", "");
+
+    CPLJSONObject user;
+    user.Add("name", osUser);
+    user.Add("password", osKey);
```
rouault: SWIFT_KEY used as password. Will need to be documented. But seeing that method, I'm wondering about the value of all those new config options. Wouldn't provided directly the serialized JSon document in a config option be more convenient ?

gdal/port/cpl_swift.cpp
```diff
@@ -170,6 +184,258 @@ bool VSISwiftHandleHelper::GetConfiguration(CPLString& osStorageURL,
...
+    CSLDestroy( papszOptions );
+
+    if( psResult == nullptr )
+        return false;
+
+    osAuthToken = CSLFetchNameValueDef(psResult->papszHeaders,
+                                       "X-Subject-Token", "");
+
+    if( ! GetAuthV3StorageURL(psResult, osStorageURL) )
+        return false;
```
rouault: missing CPLHTTPDestroyResult(psResult);

talaj: Great spot, it's fixed now.

gdal/port/cpl_swift.cpp
talaj: Fixed to be more foolproof, but it's not absolutely necessary as the code will never get there if `SWIFT_AUTH_V3_URL` is empty or unset.

gdal/port/cpl_swift.cpp
talaj: All those config options are the same as those of the python-swiftclient except they use `SWIFT_` prefix instead of `OS_`:
https://docs.openstack.org/python-swiftclient/latest/cli/index.html#authentication

So I would stick with that.

Or maybe support directly those options with `OS_` prefix. They are used for authentication across the whole OpenStack environment:
https://docs.openstack.org/python-openstackclient/train/cli/authentication.html

gdal/port/cpl_swift.cpp
talaj: I will update the documentation once config options will be settled.

gdal/port/cpl_swift.cpp
rouault: > Or maybe support directly those options with `OS_` prefix

That would seem the best option to me.

rouault: @talaj Will you change to options using the OS_ prefix ?

talaj: Yes, I will! I just haven’t had time to make it yet.

gdal/doc/source/user/virtual_file_systems.rst
```diff
@@ -378,10 +378,19 @@ Recognized filenames are of the form :file:`/vsiswift/bucket/key` where ``bucket
 
 The generalities of :ref:`/vsicurl/ </vsicurl/>` apply.
 
-Two authentication methods are possible, and are attempted in the following order:
+Three authentication methods are possible, and are attempted in the following order:
 
 1. The :decl_configoption:`SWIFT_STORAGE_URL` and :decl_configoption:`SWIFT_AUTH_TOKEN` configuration options are set respectively to the storage URL (e.g http://127.0.0.1:12345/v1/AUTH_something) and the value of the x-auth-token authorization token.
 2. The :decl_configoption:`SWIFT_AUTH_V1_URL`, :decl_configoption:`SWIFT_USER` and :decl_configoption:`SWIFT_KEY` configuration options are set respectively to the endpoint of the Auth V1 authentication (e.g http://127.0.0.1:12345/auth/v1.0), the user name and the key/password. This authentication endpoint will be used to retrieve the storage URL and authorization token mentioned in the first authentication method.
+3. Authentication with Keystone v3 is using the same options as python-swiftclient, see https://docs.openstack.org/python-swiftclient/latest/cli/index.html#authentication for more details. GDAL currently supports the following options:
```
rouault:
```suggestion
3. Authentication with Keystone v3 is using the same options as python-swiftclient, see https://docs.openstack.org/python-swiftclient/latest/cli/index.html#authentication for more details. GDAL (>= 3.1) supports the following options:
```

rouault:
Sphinx error:
```
gdal/doc/source/user/virtual_file_systems.rst:386: WARNING: Enumerated list ends without a blank line; unexpected unindent.
```

gdal/doc/source/user/virtual_file_systems.rst
talaj: Thanks, I will amend it.

talaj:
Thanks for merging. I'm planning to make a followup pull request with support of some more options like `OS_USER_ID`, `OS_PROJECT_ID` or `OS_DEFAULT_DOMAIN` once I will test it against a real OpenStack instance.

PR #2314, opened 2020-03-13
configure.ac: optimize the jni_lib compatibility.

### LINKED ISSUE #2313: Build fails on MacOS with java and hdfs support

Build success.
Supporting `libjvm.dylib` to build.

```bash
./configure --prefix=/usr/local/Cellar/gdal/2.4.0 --with-java=$JAVA_HOME --with-hdfs=$HADOOP_HOME
```
output:
```bash
checking GDAL version information from gdal_version.h.in: 2.4.0
checking for Mac OS X Framework build... disabled
checking for perl bindings... disabled
checking for python bindings... disabled
checking whether we should include Java support... yes
configure: error: Unable to find libjvm.so.
```
image
In above image, i only have libjvm.dylib instead of libjvm.so. And i check the `configure.ac` file, it doesn't consider of `libjvm.dylib`. [related link in configure.ac](https://github.com/OSGeo/gdal/blob/9d4d488a492510fdf3b20d8b74aba06ebeb5ba76/gdal/configure.ac#L5387)
## Operating system

Mac OSX 10.15.3

2.4.0

esmeetu:
fixes #2313

### DIFF
--- a/gdal/configure
+++ b/gdal/configure
@@ -941,7 +941,6 @@ infodir
 docdir
 oldincludedir
 includedir
-runstatedir
 localstatedir
 sharedstatedir
 sysconfdir
@@ -1302,7 +1301,6 @@ datadir='${datarootdir}'
 sysconfdir='${prefix}/etc'
 sharedstatedir='${prefix}/com'
 localstatedir='${prefix}/var'
-runstatedir='${localstatedir}/run'
 includedir='${prefix}/include'
 oldincludedir='/usr/include'
 docdir='${datarootdir}/doc/${PACKAGE}'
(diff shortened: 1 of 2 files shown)


rouault:
thanks, can you run "autoconf" to regenerate the "configure" file as well, and commit it ?

esmeetu:
yes, after `autoconf`, `configure` runs well.

rouault:
> yes, after `autoconf`, `configure` runs well.

Cool. But my point is that you need to "git commit" & "git push" the updated configure, because it is under version control (I know this is unusual)

esmeetu:
i don't know why the unexpected `runstatedir` things in the `configure` file was added. 
Should i remove those manually?

rouault:
> Should i add those manually?

no, that's fine. There are some small variations in the output of autoconf, depending on the actual version. I've seen those a few times, and I think they are fine

esmeetu:
ok. Perhaps my `autoconf` is 2.69, and that option is supported in 2.70.
Thanks. :)

PR #2341, opened 2020-03-24
MVT: Added transactions to MVT

SamirRehm:
The MBTiles dataset has support for transaction.
If using a SQLite3 based MBTiles file with vector data, should also allow for transaction support.
The transaction methods are copy pasted from mbtilesdataset


rouault:
Why do you need transaction ?

The SQLite3 connection in the MBTiles driver is created with settings designed for speed that are likely to not guarantee the ACID guarantees from SQL transactions
``
"PRAGMA synchronous = OFF;"
"PRAGMA journal_mode = OFF;"
"PRAGMA temp_store = MEMORY;"
```

SamirRehm:
Closing due to transactions providing little to no benefit

PR #2343, opened 2020-03-24
Elasticsearch: rename driver to Elasticsearch, update geometry type name for Points

tomkralidis:
## What does this PR do?
This PR updates point layer creation of geometry objects with `Point` instead of `POINT`.

In addition, while ES support is going through breaking changes in GDAL 3.1.x, renaming the driver name to `Elasticsearch` for consistency.

- https://github.com/geopython/pygeoapi/pull/385#issuecomment-603290464

## Tasklist

 - [x] `s/POINT/Point/g`
 - [x] `s/ElasticSearch/Elasticsearch/g`
 - [x] Add/update test case(s)
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS: Ubuntu bionic
* Compiler: gcc

### DIFF
--- a/autotest/ogr/ogr_elasticsearch.py
+++ b/autotest/ogr/ogr_elasticsearch.py
@@ -3,7 +3,7 @@
 # $Id$
 #
 # Project:  GDAL/OGR Test Suite
-# Purpose:  ElasticSearch driver testing (with fake server)
+# Purpose:  Elasticsearch driver testing (with fake server)
 # Author:   Even Rouault <even dot rouault at spatialys.com>
 #
 ###############################################################################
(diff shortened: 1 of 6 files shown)


rouault:
CC @nickpeihl in case he has observations

nickpeihl:
Code lgtm. My build env is screwy so I can't test this PR against an actual Elasticsearch service right now. 

I have not previously corrected the Elasticsearch driver name because I worried it might break existing user workflows. But if @rouault is ok with the name change, I'm 👍.

rouault:
I don't think the change in case will break things. At least GDALGetDriverByName() is case insensitive for example

PR #2359, opened 2020-03-28
gdal2tiles.py: Added support for XYZ tiles (OSM Slippy Map)

### LINKED ISSUE #2098: gdal2tiles should allow for XYZ slippy map tiles and not just TMS
At the moment gdal2tiles generates TMS slippy map tiles.

It would be great to add an option to allow the user to create XYZ tiles, see e.g.

https://gist.github.com/tmcw/4954720

which implies just flipping the tile y coordinates.

Workaround right now is to use gdal2tiles to generate TMS tiles, and then manually flip the y tile coordinates. Or find a map framework which supports TMS tiles.

quipo:
Added new `--xyz` option to generate tiles in the [OSM Slippy Map standard](https://wiki.openstreetmap.org/wiki/Slippy_Map) where the `Y` coordinate is flipped.

Also fixed incorrect indentation in a few places.

Fixes #2098

### DIFF
--- a/autotest/pyscripts/test_gdal2tiles.py
+++ b/autotest/pyscripts/test_gdal2tiles.py
@@ -56,6 +56,9 @@ def test_gdal2tiles_py_simple():
     os.unlink('tmp/out_gdal2tiles_smallworld.tif')
 
     ds = gdal.Open('tmp/out_gdal2tiles_smallworld/0/0/0.png')
+    
+    if ds is None:
+        pytest.fail('cannot open output file(s)')
 
     expected_cs = [25314, 28114, 6148, 59026]
     for i in range(4):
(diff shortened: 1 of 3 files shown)


rouault:
Thanks, we would also need documentation of the new option in gdal/doc/source/programs/gdal2tiles.rst with a `.. versionadded:: 3.1` and a test in autotest/pyscripts/test_gdal2tiles.py

quipo:
done

quipo:
not sure why https://travis-ci.com/github/OSGeo/gdal/jobs/308288415 is failing, it doesn't seem related to this change?

rouault:
> it doesn't seem related to this change?

yes, flake. I've restarted the job

rajadain:
Hi,

I just wanted to thank you both for getting this in. I wrote about it in a short blog post: https://www.azavea.com/blog/2020/06/18/this-month-we-learned-may-2020/

Thanks.

PR #2362, opened 2020-03-30
gdal2tiles.py: fix XYZ tiling with zoom range. Fix hanging on invalid input

quipo:
This PR fixes two issues:

* XYZ tiles were not generated for all zoom levels when a zoom range is specified (e.g. `--zoom=10-12`). I forgot to include this fix in my previous PR (PR #2359).
  Only tiles for the highest zoom level (12) were generated.

* When an input .tif image is provided without specifying a `--s_srs` cli parameter, and the image itself is not georeferenced (i.e. it's missing the embedded SRS metadata), the input validation fails, but the program doesn't exit correctly, but hangs instead, as there's still an open pipe (instantiated within `multiprocessing.Pool`).
  Given `worker_tile_details()` is a synchronous call that needs to be executed (once) on the input file, before calling `create_base_tile()`, and cannot be parallelised, there is no need for it to be invoked from within the Pool.
  This way, if the input validation within `worker_tile_details()` fails, there is no open pipe and the parent process can terminate correctly, reporting the failure with `sys.exit(2)`.

I added test cases for both issues.


rouault:
data/small_world_no_srs.tif is a bit big. You could potentially create a VRT file instead and remove SRS information from it for the same purposes

quipo:
good idea, done.

PR #2363, opened 2020-03-31
Average resampling (warp and overview/translate): use weighted average for border source pixels

rouault:


palmerj:
Hi @rouault interested in this change. What does it fix exactly?

rouault:
> What does it fix exactly?

A more accurate computation of the average when the resampling factor is not an integer.

Given a input 3x1 dataset
```
 1.0 2.0 1.0
```
Resampling it to 2x1 *before* lead to:
```
 1.5 1.5
```
And with this PR to:
```
 1.3333333 1.3333333
```
One can see that the average of the input dataset is 4./3 = 1.333333 which is now preserved by this PR. The central pixel of the input dataset was considered in its entirety in the resampling previously; now it is considered with a weight of 0.5, and thus the value of the left target pixel is (1.0 * 1.0 + 2.0 * 0.5) / (1.0 + 0.5).

This is a preliminary step for a following improvement implementing a sum resampling method in gdalwarp, where preservation of the total sum of the source image is needed.

PR #2370, opened 2020-04-02
Strip TOWGS84 when datum is known, in GTiff, Spatialite and GPKG drivers

rouault:
Addresses https://github.com/qgis/QGIS/issues/34993

Implements a simplified version of the idea mentioned in
https://lists.osgeo.org/pipermail/gdal-dev/2020-March/051881.html
with a variant of the more radical approach suggested in
https://lists.osgeo.org/pipermail/gdal-dev/2020-March/051885.html

That is the new method OGRSpatialReference::StripTOWGS84IfKnownDatumAndAllowed()
will remove TOWGS84 information from SRS whose datum name is known,
unless the OSR_STRIP_TOWGS84 configuration option is set to NO.
When TOWGS84 information is striped, a debug message is emitted.

### DIFF
--- a/autotest/cpp/test_osr.cpp
+++ b/autotest/cpp/test_osr.cpp
@@ -27,6 +27,7 @@
 
 #include "cpl_string.h"
 #include "ogr_srs_api.h"
+#include "ogr_spatialref.h"
 
 #include <algorithm>
 #include <cmath>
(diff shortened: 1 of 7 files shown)


nyalldawson:
I'll test today, but just wanted to say that the logic looks good to me (and makes sense to me). Thanks for the fix!

nyalldawson:
Tested and confirmed working as desired

PR #2397, opened 2020-04-10
vsis3 compatibility

normanb:
## What does this PR do?

This PR provides compatiblity with the `vsis3` prefix used in Rasterio. It also provides a fix for setting the metadata in FlushCache when using s3.

As TileDB can perform updates directly on s3 I removed the error check in `vsicurl`, if this should be retained then I can make the changes required to add a conditional compilation.

I was testing with Rasterio so I haven't added a GDAL specific test case here.

### DIFF
--- a/gdal/frmts/tiledb/tiledbdataset.cpp
+++ b/gdal/frmts/tiledb/tiledbdataset.cpp
@@ -141,6 +141,18 @@ class TileDBRasterBand final: public GDALPamRasterBand
 
 };
 
+static CPLString vsi_to_s3( const char* pszUri )
+{
+    CPLString osUri;
+
+    if ( STARTS_WITH_CI( pszUri, "/VSIS3/") )
+        osUri.Printf("s3://%s", pszUri + 7);
+    else
+        osUri = pszUri;
+
+    return osUri;
+}
+
 /************************************************************************/
 /*                             SetBuffer()                              */
 /************************************************************************/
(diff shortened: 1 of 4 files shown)


rouault:
> As TileDB can perform updates directly on s3 I removed the error check in `vsicurl`, if this should be retained then I can make the changes required to add a conditional compilation.

I don't understand this part, but this change is inappropriate. VSIFOpenL("/vsicurl/something", "wb") is definitely not possible. Sequential writing to /vsis3/ (but not random writing) is possible, but this shouldn't go through VSICurlFilesystemHandler::Open()

normanb:
The TileDB driver will write a new fragment to an array on S3 so updates are supported directly on S3.

I am requesting `r+` which fails this check. I will spend some more time on debugging the path and update this PR and keep the check.

gdal/frmts/tiledb/tiledbdataset.cpp
```diff
@@ -1904,20 +1919,24 @@ TileDBDataset::Create( const char * pszFilename, int nXSize, int nYSize, int nBa
         GDALDataType eType, char ** papszOptions )
 
 {
+    const char* pszBlockXSize = CSLFetchNameValue( papszOptions, "BLOCKXSIZE" );
```
rouault, line 1922: frmts/tiledb/tiledbdataset.cpp:1922,style,unreadVariable,Variable 'pszBlockXSize' is assigned a value that is never used.

normanb: Thanks, leftover debug. I have updated the PR.

rouault:
Can you rework your commit titles so they are more explicit ?

rouault:
The second commit "Only scan for other drivers if a driver is not specified " would be even better titled as "QuietDelete: support expliciting the drivers to use", and it includes a TileDB change that should have gone in the previous TileDB specific commit.

PR #2409, opened 2020-04-15
Add instructions to register nickname with Freenode

strk:
As it looks like #gdal on Freenode restricts voice to
registered users..

### DIFF
--- a/gdal/doc/source/community.rst
+++ b/gdal/doc/source/community.rst
@@ -44,6 +44,12 @@ with the community.
 The `Mailing List`_ and `GitHub`_ avenues are going to be more productive
 communication channels in most situations.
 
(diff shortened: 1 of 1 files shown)


gdal/doc/source/community.rst
```diff
@@ -44,6 +44,12 @@ with the community.
 The `Mailing List`_ and `GitHub`_ avenues are going to be more productive
 communication channels in most situations.
 
+NOTE: the Freenode channel only accepts messages from registered users,
```
rouault, line 47: using the `.. note::` syntax would probably be best

strk:
> using the `.. note::` syntax would probably be best

Done, please check (I'm not familiar)

PR #2413, opened 2020-04-16
Doc: document OGREnvelope [skip appveyor]

rouault:


markg85:
+1 for the doc update, thank you!

PR #2415, opened 2020-04-17
For TileDB 1.7 or greater enable use of internal metadata

normanb:
This is a bug I noticed when compiling GDAL with the TileDB dev branch (which is set to 2.0). It doesn't impact user functionality as metadata storage is an internal feature.


rouault:
Needs to be rebased on top of current master to solve conflict

PR #2420, opened 2020-04-18
Replace %.18g by %.17g since 17 significant figures are enough to encode losslessly a IEEE-754 double precision number

rouault:


rouault:
The failure on Windows in GPKG test is a bit strange. I suspect an issue with sqlite3_printf(). Dropping the idea for now.

PR #2428, opened 2020-04-21
Backward compatibility for BAG metadata reading

txantxangorriak:
BAG XML metadata schema has been modified with version 1.5; actually, GDAL only handles latest format, this pull request proposes to handle backward compatibility (from 1.0 to 1.6 versions)


rouault:
Thanks!

PR #2432, opened 2020-04-22
Fix typo on gdalwarp.rst

eamanu:
Change Sse for Set.

I am newbie with gdal, and not English speaker, so sorry if this is not a typo.


rouault:
Thanks

PR #2452, opened 2020-04-28
configure: use sed -E or sed -r depending on runtime availability

rouault:
Fixes https://github.com/OSGeo/gdal/pull/2395#issuecomment-620557715

### DIFF
--- a/gdal/configure
+++ b/gdal/configure
@@ -22669,6 +22669,7 @@ rm -f core conftest.err conftest.$ac_objext conftest.$ac_ext
 
 
 
+
 # Check whether --with-libz was given.
 if test "${with_libz+set}" = set; then :
   withval=$with_libz;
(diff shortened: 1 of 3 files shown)


rouault:
Patching https://github.com/DanielFEvans/gdalmanylinux/blob/gdal_3.1.0/Dockerfile.wheels to use this branch with the following leads now to a successful build:
```
RUN yum update -y && yum install -y unzip
# gdal
RUN mkdir -p /src \
    && cd /src \
    && curl -f -L -O https://github.com/rouault/gdal/archive/fix_sed_extended_compat.zip \
    && unzip fix_sed_extended_compat.zip \
    && cd /src/gdal-fix_sed_extended_compat/gdal \
```

rouault:
Cherry-picked to 3.1

Issue #2459, opened 2020-04-30
Potential-null-reference in ogr_flatgeobuf::GeometryReader::readMultiPolygon


rouault:
See https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=21945

@bjornharrtell Can you possibly have a look at that ? The problem is probably not seen at runtime with normal compiler, and mostly of the pedantic category only with nullptr->toPolygon() not being valid even if toPolygon() just does a cast on this. So the fix would be to test first for null pointer before using the toPolygon() cast. I suspect other similar cases in neighbouring code.

PR #2461, opened 2020-04-30
FlatGeobuf: add null check for readMultiPolygon

bjornharrtell:
ref https://github.com/OSGeo/gdal/issues/2459

I'm not sure this is the correct fix though and wonder why the other similar cases are slightly different (the ->toCurve ones).

### DIFF
--- a/gdal/ogr/ogrsf_frmts/flatgeobuf/geometryreader.cpp
+++ b/gdal/ogr/ogrsf_frmts/flatgeobuf/geometryreader.cpp
@@ -259,13 +259,10 @@ OGRMultiPolygon *GeometryReader::readMultiPolygon()
     auto mp = std::unique_ptr<OGRMultiPolygon>(new OGRMultiPolygon());
     for (uoffset_t i = 0; i < parts->size(); i++) {
         GeometryReader reader { parts->Get(i), GeometryType::Polygon, m_hasZ, m_hasM };
-        auto g = reader.read();
+        auto g = std::unique_ptr<OGRGeometry>(reader.read());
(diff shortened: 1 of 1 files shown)


gdal/ogr/ogrsf_frmts/flatgeobuf/geometryreader.cpp
```diff
@@ -259,7 +259,10 @@ OGRMultiPolygon *GeometryReader::readMultiPolygon()
     auto mp = std::unique_ptr<OGRMultiPolygon>(new OGRMultiPolygon());
     for (uoffset_t i = 0; i < parts->size(); i++) {
         GeometryReader reader { parts->Get(i), GeometryType::Polygon, m_hasZ, m_hasM };
-        auto p = std::unique_ptr<OGRPolygon>(reader.read()->toPolygon());
+        auto g = reader.read();
+        if (g == nullptr)
+            return nullptr;
+        auto p = std::unique_ptr<OGRPolygon>(g->toPolygon());
```
rouault, line 265: you could probably just do mp->addGeometryDirectly(g->toPolygon()) . There's no way p can be nullptr if g has been tested for nullptr before
The toCurve() cases are different since they don't invoke GeometryReader with a given geometry type, hence they could get something different and need to do the dynamic_cast test. Here we know that we get either a polygon or null.

bjornharrtell: Ok, reworked but it looked like it should use std::unique_ptr and release as in the other cases so that is what I did.

PR #2465, opened 2020-05-01
Polygonize: improve performance on large rasters

rouault:


rouault:
Tested on http://oe.oregonexplorer.info/ExternalContent/SpatialDataforDownload/OR_NLCD_2011.zip , converted to GeoTIFF, after the change, runs in 48 minutes whereas 12+ hours originally (not tested on same machines, but still shows the improvement)

PR #2481, opened 2020-05-05
hdf5r: Add hdf5r format

phubner:
## What does this PR do?
Adds the HDF5R format. This format is a sub data set of the HDF5 format provided from the NOMS-E BAA contract FA8810-17-C-0003.

N/A

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS: Centos 8
* Compiler: gcc


rouault:
I'd suggest you write to the mailing list to at least:
- introduce yourself, since I don't think you've interacted before with the GDAL community
- explain a bit about this work, as I'm afraid hardly anyone can decipher the current description, and the interest of having this upstream
- how this would be maintained onward if it would be merged

From the stats, this is a very substantial code addition which will require a substantial review effort and is likely going to be difficult to attract volunteers.

phubner:
Hi @rouault, Thanks for the feedback. I will send out a note to the mailing list today!

PR #2495, opened 2020-05-08
Update GNUmakefile

lbartoletti:
`ln: illegal option -- r`

`-r` argument is not posix (aka doesn't exist on BSD like, macOs include)

cc @landryb

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:

### DIFF
--- a/gdal/scripts/GNUmakefile
+++ b/gdal/scripts/GNUmakefile
@@ -16,5 +16,5 @@ install:
 	$(INSTALL_DIR) $(DESTDIR)$(INST_BASH_COMPLETION) ; \
 	$(INSTALL_DATA) gdal-bash-completion.sh $(DESTDIR)$(INST_BASH_COMPLETION)/gdalinfo; \
 	for p in $(PROGRAMS); do \
-		ln -rs $(DESTDIR)$(INST_BASH_COMPLETION)/gdalinfo $(DESTDIR)$(INST_BASH_COMPLETION)/$$p; \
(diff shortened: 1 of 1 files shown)


rouault:
I don't think that just removing the -r switch can be done just like that. It is probably there for a reason
```
       -r, --relative
              create symbolic links relative to link location
```

landryb:
usually to workaround this you `cd $dest ; ln -s ../../../$src .` no ?

gdal/scripts/GNUmakefile
```diff
@@ -16,5 +16,5 @@ install:
 	$(INSTALL_DIR) $(DESTDIR)$(INST_BASH_COMPLETION) ; \
 	$(INSTALL_DATA) gdal-bash-completion.sh $(DESTDIR)$(INST_BASH_COMPLETION)/gdalinfo; \
 	for p in $(PROGRAMS); do \
-		ln -rs $(DESTDIR)$(INST_BASH_COMPLETION)/gdalinfo $(DESTDIR)$(INST_BASH_COMPLETION)/$$p; \
+		ln -s $(DESTDIR)$(INST_BASH_COMPLETION)/gdalinfo $(DESTDIR)$(INST_BASH_COMPLETION)/$$p; \
```
dg0yt, line 19: Since `gdalinfo` and all `PROGRAMS` are just pure filenames with out path, why not just:
~~~
ln -s gdalinfo $(DESTDIR)$(INST_BASH_COMPLETION)/$$p

PR #2496, opened 2020-05-08
Travis sanitize: upgrade to clang 9 + cpl_json.h: change the Type and PrettyFormat 'enum' to 'enum class'

rouault:


rouault:
CC @BishopGIS for the cpl_json changes

PR #2512, opened 2020-05-11
configure.ac: Configure proj dependencies before proj (master)

david-sinuela-pix4d:
## What does this PR do?

Change the order of checks in configure.ac so that by the time the checks for proj are performed the dependencies of proj are already in LIBS. With this approach it is sure that the libraries provided to configure through the `--with-*` flags are also used in the attempt to link with proj.

I am aware of `--with-proj-extra-lib-for-test` in 3.1 but I thought that this approach would lead to less repetition in the call to `configure` and be more consistent.

## What are related issues/pull requests?

Addresses #2511.

## Tasklist

 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments
 - [ ] All CI builds and checks have passed

Evidently I have not tested all possible combinations, does the CI cover enough cases to be confident about the change?

## Environment

Same as described in #2511.

## Other notes

There is a similar PR for master in https://github.com/OSGeo/gdal/pull/2513

Issue #2511, opened 2020-05-11
Tests in configure.ac fail when linking with a custom version of proj7 that requires libtiff (and maybe sqlite and curl)


david-sinuela-pix4d:
## Expected behavior and actual behavior.

The checks in configure.ac fail when building GDAL with a custom proj version that needs libtiff (configure with --with-libtiff=path). This is the error message in `config.log`:

```
configure:19424: checking for PROJ >= 6 library
configure:19648: checking for proj_create_from_wkt in -lproj
configure:19673: /usr/bin/g++-8 -o conftest -DHAVE_AVX_AT_COMPILE_TIME -DHAVE_SSSE3_AT_COMPILE_TIME -DHAVE_SSE_AT_COMPILE_TIME -g -O2 -fPIC -fvisibility=hidden   conftest.cpp -lpr
oj  -L/home/siu/.conan/data/proj/7.0.0-1/pix4d/stable/package/c4be11918432d1a23dc982cb9819d307510a344c/lib -lproj -lz -L/home/siu/.conan/data/zlib/1.2.11/conan/stable/package/1d87
7a3df840030e6a8abb74c5ffb9088d08b47a -L/home/siu/.conan/data/zlib/1.2.11/conan/stable/package/1d877a3df840030e6a8abb74c5ffb9088d08b47a/lib -lpthread -lm -lrt -ldl  >&5
/usr/bin/ld: /home/siu/.conan/data/proj/7.0.0-1/pix4d/stable/package/c4be11918432d1a23dc982cb9819d307510a344c/lib/libproj.so: undefined reference to `TIFFReadEncodedStrip@LIBTIFF_4.0'
/usr/bin/ld: /home/siu/.conan/data/proj/7.0.0-1/pix4d/stable/package/c4be11918432d1a23dc982cb9819d307510a344c/lib/libproj.so: undefined reference to `TIFFSetTagExtender@LIBTIFF_4.0'
/usr/bin/ld: /home/siu/.conan/data/proj/7.0.0-1/pix4d/stable/package/c4be11918432d1a23dc982cb9819d307510a344c/lib/libproj.so: undefined reference to `TIFFClientOpen@LIBTIFF_4.0'
/usr/bin/ld: /home/siu/.conan/data/proj/7.0.0-1/pix4d/stable/package/c4be11918432d1a23dc982cb9819d307510a344c/lib/libproj.so: undefined reference to `TIFFStripSize64@LIBTIFF_4.0'
[...]

PR #2513, opened 2020-05-11
configure.ac: Configure proj dependencies before proj (3.0)

david-sinuela-pix4d:
## What does this PR do?

Change the order of checks in configure.ac so that by the time the checks for proj are performed the dependencies of proj are already in LIBS. With this approach it is sure that the libraries provided to configure through the `--with-*` flags are also used in the attempt to link with proj.

I am aware of `--with-proj-extra-lib-for-test` in 3.1 but I thought that this approach would lead to less repetition in the call to `configure` and be more consistent.

Addresses #2511.

## Tasklist

 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

Evidently I have not tested all possible combinations, does the CI cover enough cases to be confident about the change?

## Environment

Same as described in #2511.

## Other notes

There is a similar PR for master in https://github.com/OSGeo/gdal/pull/2512

### DIFF
--- a/gdal/configure.ac
+++ b/gdal/configure.ac
@@ -1496,7 +1496,7 @@ else
     AC_CHECK_LIB(proj,proj_create_from_wkt,PROJ_FOUND=yes,PROJ_FOUND=no,)
     AC_LANG_POP([C++])
     if test "$PROJ_FOUND" = "no"; then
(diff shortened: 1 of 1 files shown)


rouault:
3.0 is no longer maintained

PR #2516, opened 2020-05-11
Fix Python2 install to be synchronous and report all errors

### LINKED ISSUE #2515: Compile fails are supressed during module install in Python 2
##  Expected behavior and actual behavior.

Issue on line
https://github.com/OSGeo/gdal/blob/master/gdal/swig/python/setup.py#L347
`ThreadPoolExecutor.map()` returns an iterator. That iterator is not consumed so if there is a compile fail, the fail is ignored. Btw, that also means the installation does not wait for files to be compiled.

1. Add a compile fail somewhere, e.g. in `gdal/ogr/ogr_api.h` (in my case, I simply did not have MSVC setup for Python 2 on Windows).
2. `python2 gdal/swig/python/setup.py install`

## Operating system

windows 10

`3.1.0` from pypi.org, or `master`.

joaoe:
The Python 2 install uses ThreadedPoolExecutor.map() which returns an iterator.
That iterator needs to be consumed else errors are not reported and the module
install process does not wait for modules to be compiled.

Fixes #2515

## Environment

* OS: windows
* Compiler: msvc

### DIFF
--- a/gdal/swig/python/setup.py
+++ b/gdal/swig/python/setup.py
@@ -344,7 +344,8 @@ if sys.version_info < (3,):
             self.check_extensions_list(self.extensions)
 
             with Pool(num_jobs) as pool:
-                pool.map(self.build_extension, self.extensions)
+                # Note: map() returns an iterator that needs to be consumed.
+                list(pool.map(self.build_extension, self.extensions))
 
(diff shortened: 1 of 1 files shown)


rouault:
thanks

PR #2531, opened 2020-05-14
swq_op_general.cpp: Fix arithmetic expression

troopa81:
## What does this PR do?

Second part of arithmetic expression were not taken into account for type conversion, leading to calculus errors. 

This example returned bad result
```shell
$ ogrinfo -sql "select cast((1) as integer)/1.  from division" division_shp/division.shp
...
OGRFeature(division):0
  FIELD_1 (Integer) = 0

$ ogrinfo -sql "select (1*1)/1.  from division" division_shp/division.shp
...
OGRFeature(division):0
  FIELD_1 (Integer) = 0
```

This fixes the following QGIS issue https://github.com/qgis/QGIS/issues/35449

## Tasklist

 - [x] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS: Debian 10
* Compiler: gcc 9.2.1

### DIFF
--- a/gdal/ogr/swq_op_general.cpp
+++ b/gdal/ogr/swq_op_general.cpp
@@ -1182,12 +1182,12 @@ swq_field_type SWQGeneralChecker( swq_expr_node *poNode,
             eRetType = SWQ_STRING;
             eArgType = SWQ_STRING;
         }
-        else if( poNode->papoSubExpr[0]->field_type == SWQ_FLOAT )
+        else if( poNode->papoSubExpr[0]->field_type == SWQ_FLOAT || poNode->papoSubExpr[1]->field_type == SWQ_FLOAT )
         {
             eRetType = SWQ_FLOAT;
             eArgType = SWQ_FLOAT;
         }
-        else if( poNode->papoSubExpr[0]->field_type == SWQ_INTEGER64 )
+        else if( poNode->papoSubExpr[0]->field_type == SWQ_INTEGER64 || poNode->papoSubExpr[1]->field_type == SWQ_INTEGER64 )
         {
             eRetType = SWQ_INTEGER64;
             eArgType = SWQ_INTEGER64;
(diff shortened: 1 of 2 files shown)


autotest/ogr/ogr_shape.py
```diff
@@ -4818,6 +4818,37 @@ def test_ogr_shape_ldid_and_cpg():
     gdal.Unlink('/vsimem/tmp.dbf')
     gdal.Unlink('/vsimem/tmp.cpg')
 
+###############################################################################
+# Test ExecuteSQL() arithmetic expression
+
+
+def test_ogr_shape_arithmetic_expression():
```
rouault, line 4825: this test should rather be in ogr_sql_test.py

gdal/ogr/swq_op_general.cpp
```diff
@@ -1182,7 +1182,7 @@ swq_field_type SWQGeneralChecker( swq_expr_node *poNode,
             eRetType = SWQ_STRING;
             eArgType = SWQ_STRING;
         }
-        else if( poNode->papoSubExpr[0]->field_type == SWQ_FLOAT )
+        else if( poNode->papoSubExpr[0]->field_type == SWQ_FLOAT || poNode->papoSubExpr[1]->field_type == SWQ_FLOAT )
```
rouault, line 1185: I believe the below test for SWQ_INTEGER64 at l 1190 (and the one at l 1214 too) would also benefit from the same logic, since currently cast(1 as integer)+ 1234567890123 isn't promote to Integer64

autotest/ogr/ogr_shape.py
troopa81: Done

gdal/ogr/swq_op_general.cpp
troopa81: Yes. Indeed

rouault:
Backported in 3.1 per 25f4fb0

troopa81:
Thanks @rouault

Issue #2526, opened 2020-05-13
frmt_wms should link back to frmt_wmts


jidanni:
/usr/share/doc/libgdal-doc/gdal/frmt_wmts.html links to
/usr/share/doc/libgdal-doc/gdal/frmt_wms.html but not vice-versa.
3.0.4

PR #2534, opened 2020-05-14
Doc: link WMS doc with WMTS doc

jratike80:
I hope that the ":ref" leads to correct result and fixes https://github.com/OSGeo/gdal/issues/2526.


rouault:
Thanks!

PR #2537, opened 2020-05-15
gdalMDIMinfo, not gdalinfo

jidanni:
Added the MDIM in the middle.

P.S., proper English is THE xxx program, not just xxx program...


jidanni:
Note I have no way of guessing what
.. program:: gdalinfo
renders to here on github, so removed it.

I bet it is actually pretty important.

PR #2542, opened 2020-05-15
Make output more match current relaity

jidanni:
Also removed unneeded quote from command line.
You are welcome to adjust the rest of the examples.


jratike80:
Perhaps good to keep the double quotes as a hint that they must be used in some use cases though not just with EPSG codes.

`gdalsrsinfo "+proj=longlat +datum=WGS84 +no_defs"`

PR #2550, opened 2020-05-15
/vsis3/: use IMDSv2 protocol to retrieve IAM role and credentials

rouault:


rcoup:
FYI, I discovered this causes problems running kube2iam (which is one of the main approaches to per-pod AWS IAM roles in a Kubernetes cluster), since [it doesn't yet support IMDSv2](https://github.com/jtblin/kube2iam/issues/241).

rouault:
> FYI, I discovered this causes problems running [kube2iam](https://github.com/jtblin/kube2iam)

should hopefully fixed per https://github.com/OSGeo/gdal/commit/ccbd90ffd0bbb255a0ba4c145176833cfb04cfa8 . I only tested against the stub server of autotest, not kube2iam

rcoup:
wow, thanks! ❤️

It's a bit of an edge case, was more an FYI for someone else (ahem, #futureme) running into it.

PR #2551, opened 2020-05-15
COG: add a SPARSE_OK=YES option to create sparse files

rouault:
- Propagate SPARSE_OK mode to mask and overviews (also true for
  base GTiff driver)
- Propagate SPARSE_OK for temporary files of COG driver
- Update optimized range retrieval of GTiff reader to not emit
  debug messages on spare tiles
- Update validate_cloud_optimized_geotiff.py to support such files


blacha:
Thanks for this!
Ran a few test examples from your branch and it is working well!

This also seems to fix a issue where `SPARSE_OK=YES` would still emit every 2nd tile even if they were empty.

rouault:
> This also seems to fix a issue where `SPARSE_OK=YES` would still emit every 2nd tile even if they were empty.

hum I'm not sure what in my fix would address this, at least for the main IFD (SPARSE_OK wasn't propagated to overviews and mask). On which image did you have the issue you mention ?

blacha:
Hopefully this isn't taking this issue too far off topic,

I was converting a small testing COG into a single sparse GTIFF with the following command 
```
#!/bin/sh
gdal_translate \
  -of GTIFF \
  -co COMPRESS=webp \
  -co NUM_THREADS=ALL_CPUS \
  -co BIGTIFF=YES \
  -co SPARSE_OK=YES \
  -co TILED=YES \
   $1 $1.sparse.tiff 
```
Using `GDAL 3.0.4, released 2020/01/28`

This output a Tiff that was tiled and had quite a few empty tiles, the first row should in theory be all empty, but when dumping the tile offsets for the first row I was seeing

```
{ i: 0, offset: 32811 }
{ i: 1, offset: 0 }
{ i: 2, offset: 33035 }
{ i: 3, offset: 0 }
{ i: 4, offset: 34379 }
{ i: 5, offset: 0 }
{ i: 6, offset: 33259 }
{ i: 7, offset: 0 }
{ i: 8, offset: 33483 }
{ i: 9, offset: 0 }
```

When running with the new version and creating a COG all the expected offsets were 0

Dumping out  the tiles and looking at what is in them with imagemagick
```
$ ls -l 000_002_0.webp 
-rw-rw-r-- 1 blacha blacha 220 May 19 10:17 000_002_0.webp


$ identify 000_002_0.webp 
000_002_0.webp WEBP 256x256 256x256+0+0 8-bit sRGB 220B 0.000u 0:00.000

$ convert 000_002_0.webp -format \"%c\" histogram:info: 

     65312: (  1,  1,  1,  0) #01010100 srgba(1,1,1,0)
       224: (  2,  2,  2,  0) #02020200 srgba(2,2,2,0)
```

Shows the tiles were fully empty, Although they had two colors of full alpha

After running a few more tests, I think it might actually be the COG process that fixes it, as I was comparing a  `-of GTIFF` vs a `-of COG`

rouault:
hum ok. Merging. Sometimes we have to admit that things just work :-)

PR #2552, opened 2020-05-15
use returned length to truncate string for tiledb metadata

normanb:
This is a check that the returned metadata is truncated if necessary to provide a valid XML string for metadata parsing in GDAL. By not using the returned `v_num` we could potentially have an invalid XML string. 

Can we backport this on to v3.1?


rouault:
cherry-picked to 3.1 (2082fe5)

Issue #2052, opened 2019-11-25
Add ability to send arbitrary HTTP headers to WMTS etc. sites


jidanni:
https://gdal.org/drivers/raster/wmts.html features many HTTP headers one can send in a request.
However one never knows when a site might need some other arbitrary header to get it to respond.
E.g., "Cookie: ...."

Therefore please add the ability to add arbitrary HTTP headers to WMTS etc. sites.

jratike80:
I can read from https://gdal.org/user/virtual_file_systems.html

> Starting with GDAL 2.3, additional HTTP headers can be sent by setting the GDAL_HTTP_HEADER_FILE configuration option to point to a filename of a text file with “key: value” HTTP headers.

This config option is not listed in https://trac.osgeo.org/gdal/wiki/ConfigOptions. I wonder if it affects only vsi* systems or would it work also with the drivers which are using curl. If it does not work then perhaps re-using  GDAL_HTTP_HEADER_FILE could be the way to implement the support.

rouault:
> I wonder if it affects only vsi* systems or would it work also with the drivers which are using curl

Ah, good catch that can indeed work:

```
$ echo "foo: bar" > /tmp/header.txt
$ rm -rf gdalwmscache
$ CPL_CURL_VERBOSE=YES GDAL_HTTP_HEADER_FILE=/tmp/header.txt gdal_translate frmts/wms/frmt_wms_openstreetmap_tms.xml out.tif -outsize 256 256 --debug on
[...]
> GET /0/0/0.png HTTP/1.1
Host: tile.openstreetmap.org
User-Agent: GDAL WMS driver (http://www.gdal.org/frmt_wms.html)
Accept: */*
foo: bar
[...]
```

jratike80:
@jidanni is this now documented well enough?

jidanni:
Add an example of how to add a  cookie to a request. Thanks.

PR #2562, opened 2020-05-16
Mention how to add arbitrary headers

jidanni:
One of the fixes for #2052.


gdal/doc/source/drivers/raster/wmts.rst
```diff
@@ -296,6 +296,9 @@ It is important that there be no spaces or other content before the
 +-----------------------------------+-----------------------------------+
 |                                   |                                   |
 +-----------------------------------+-----------------------------------+
+Starting with GDAL 2.3, additional HTTP headers can be sent by setting the
```
rouault, line 299:
```suggestion

Starting with GDAL 2.3, additional HTTP headers can be sent by setting the
```

PR #2565, opened 2020-05-17
Add GDALAutoCreateWarpedVRTEx function to GDAL API

wonder-sk:
## What does this PR do?

This function allows specifying extra options for transformer.
Thanks to that, it is for example possible to set RPC_HEIGHT
to get better aligned rasters with RPC georeferencing metadata.

Originally suggested by Even in https://github.com/qgis/QGIS/pull/36385

## Tasklist

 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/gdal/frmts/vrt/vrtwarped.cpp
+++ b/gdal/frmts/vrt/vrtwarped.cpp
@@ -123,6 +123,7 @@ GDALAutoCreateWarpedVRT( GDALDatasetH hSrcDS,
  * Compared to GDALAutoCreateWarpedVRT() this function adds one extra
  * argument: options to be passed to GDALCreateGenImgProjTransformer2().
  *
+ * @since 3.2
  */
 
 GDALDatasetH CPL_STDCALL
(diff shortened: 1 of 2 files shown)


gdal/frmts/vrt/vrtwarped.cpp
```diff
@@ -107,7 +107,32 @@ GDALAutoCreateWarpedVRT( GDALDatasetH hSrcDS,
...
+/************************************************************************/
+/*                     GDALAutoCreateWarpedVRTEx()                      */
+/************************************************************************/
+
+/**
+ * Create virtual warped dataset automatically.
+ *
+ * Compared to GDALAutoCreateWarpedVRT() this function adds one extra
+ * argument: options to be passed to GDALCreateGenImgProjTransformer2().
+ *
```
rouault, line 125:
```suggestion
 *
 * @since 3.2
```

PR #2577, opened 2020-05-18
MITAB: fix reading and writing of Transverse Mercator projections based on KKJ

rouault:
On reading, the SRS_PT_TRANSVERSE_MERCATOR_MI_21/_22/_23_/_24 special cases
are redirected on standard Transverse Mercator
On writing, if Tranverse Mercator + datum = KKJ, use the 24 projection code


rouault:
cherry-picked in 3.1

PR #2583, opened 2020-05-19
enable commit notifications to dedicated #gdal-activity IRC channel

jmckenna:
- enable commit notifications to dedicated IRC channel "gdal-activity"
 - this follows how other projects handle this as QGIS, PostGIS (pushing commit notifications into dedicated <PROJECT>-activity channel, so packagers and developers get notices of changes, if they wish).

@rouault has also been set as founder of the new activity channel.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -187,8 +187,8 @@ notifications:
     channels:
        - "irc.freenode.org#gdal-activity"    
       #- "irc.freenode.org#gdal"
-   use_notice: true
-   on_success: change
+    use_notice: true
+    on_success: change
 #
 #   email:
 #     recipients:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -182,13 +182,15 @@ install:
 script:
   - ./gdal/ci/travis/${BUILD_NAME}/script.sh
 
-# notifications:
+notifications:
+  irc:
+    channels:
+       - "irc.freenode.org#gdal-activity"    
+      #- "irc.freenode.org#gdal"
+   use_notice: true
```
rouault, line 190: I believe this line and the following one should receive an extra space for indentation to make it valid .yml

jmckenna: good catch, thanks @rouault

Issue #2571, opened 2020-05-17
Document all CPL... variables


jidanni:
In https://github.com/OSGeo/gdal/issues/2052#issuecomment-628650934 we see

> $ CPL_CURL_VERBOSE=YES GDAL_HTTP_HEADER_FILE...

The user can find (command line) documentation about
GDAL_... variables, but not CPL_... variables, in the various doc files.

rouault:
This is just too generic to be actionable.

PR #2588, opened 2020-05-20
CPL_CURL_VERBOSE GDAL_HTTP_HEADER_FILE example

jidanni:
Example from #2052 to assist with #2571 .
**P.S., :decl_configoption: is not getting parsed for me. Please fix it.**
Yes, I used U+0060 GRAVE ACCENT

### DIFF
--- a/gdal/doc/source/programs/gdal_translate.rst
+++ b/gdal/doc/source/programs/gdal_translate.rst
@@ -284,3 +284,10 @@ To create a RGBA dataset from a RGB dataset with a mask
 
     gdal_translate withmask.tif rgba.tif -b 1 -b 2 -b 3 -b mask
 
+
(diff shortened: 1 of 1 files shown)


rouault:
I'm not sure this belongs to the gdal_translate page. This is more network specific. We have https://gdal.org/user/virtual_file_systems.html#network-based-file-systems with the /vsicurl/ etc subsystems which would be the closest, but I believe a dedicated page would be needed for general networking CURL options. Based on content from https://gdal.org/doxygen/cpl__http_8h.html#a578662ddc6557b028fecc36e302aed41 / https://github.com/OSGeo/gdal/blob/master/gdal/port/cpl_http.cpp#L560

> :decl_configoption: is not getting parsed for me

Yes, this is a GDAL doc specific directive processed by a Python extension: https://github.com/OSGeo/gdal/blob/master/gdal/doc/source/_extensions/configoptions.py

jidanni:
Yes I just hope it gets finally documented somewhere.

PR #2593, opened 2020-05-21
Multi-threaded overview computation

rouault:
- Overview computation is multi-threaded
- Warping, overview building and GTiff multi-threaded writing share the same thread pool

Tests on a i7-6700HQ CPU @ 2.60GHz (4 cores, hyper-threaded => 8 vCPUs) :

A) pure overview computation on a 20000 x 20000 pixel 3-band in-memory TIFF dataset:

before this PR:

$ python perftests/overview.py
COMPRESS=NONE, NUM_THREADS=0: 7.47
COMPRESS=NONE, NUM_THREADS=2: 7.36
COMPRESS=NONE, NUM_THREADS=4: 7.35
COMPRESS=NONE, NUM_THREADS=8: 7.36
COMPRESS=ZSTD, NUM_THREADS=0: 8.97
COMPRESS=ZSTD, NUM_THREADS=2: 8.87
COMPRESS=ZSTD, NUM_THREADS=4: 8.73
COMPRESS=ZSTD, NUM_THREADS=8: 8.65

after

$ python perftests/overview.py
COMPRESS=NONE, NUM_THREADS=0: 6.76
COMPRESS=NONE, NUM_THREADS=2: 5.32
COMPRESS=NONE, NUM_THREADS=4: 2.90
COMPRESS=NONE, NUM_THREADS=8: 3.12
COMPRESS=ZSTD, NUM_THREADS=0: 8.09
COMPRESS=ZSTD, NUM_THREADS=2: 6.18
COMPRESS=ZSTD, NUM_THREADS=4: 4.55
COMPRESS=ZSTD, NUM_THREADS=8: 3.76

B) COG generation

Given a input 20000 x 20000 pixel 3-band pixel-interleaved tiled (256 x 256) ZSTD compressed TIFF file:

before this PR:

$ python perftests/cog.py 
COMPRESS=ZSTD, NUM_THREADS=0: 12.91
COMPRESS=ZSTD, NUM_THREADS=2: 11.36
COMPRESS=ZSTD, NUM_THREADS=4: 11.27
COMPRESS=ZSTD, NUM_THREADS=8: 11.15
COMPRESS=WEBP, NUM_THREADS=0: 43.75
COMPRESS=WEBP, NUM_THREADS=2: 26.33
COMPRESS=WEBP, NUM_THREADS=4: 18.31
COMPRESS=WEBP, NUM_THREADS=8: 16.88

after:

$ python perftests/cog.py
COMPRESS=ZSTD, NUM_THREADS=0: 12.23
COMPRESS=ZSTD, NUM_THREADS=2: 8.90
COMPRESS=ZSTD, NUM_THREADS=4: 6.40
COMPRESS=ZSTD, NUM_THREADS=8: 6.24
COMPRESS=WEBP, NUM_THREADS=0: 42.69
COMPRESS=WEBP, NUM_THREADS=2: 18.79
COMPRESS=WEBP, NUM_THREADS=4: 17.13
COMPRESS=WEBP, NUM_THREADS=8: 12.22


rouault:
CC @palmerj @blacha

palmerj:
@rouault thank you very much for this. We will test it out. 

I see for no compression and ZSTD there is little improvement in the performance from 4 to 8 threads. Is that due to hyperthreading and the type of operations?

blacha:
@rouault Thanks for this! it looks to be a pretty significant speed up for my use case about 1.3x faster for COG creation, using my server `Intel(R) Xeon(R) D-2141I CPU @ 2.20GHz`

Running some tests on this PR's branch vs the latest master docker container `GDAL 3.2.0dev-ca072e09ada35d31a8228bfce416bea31b44245a, released 2020/05/22`

all tests were run 3 times with `hyperfine`
```
hyperfine --export-markdown result.md  -r 3 \
  "./make-cog.sh 'gdal:thread-pool'" \
  "./make-cog.sh 'osgeo/gdal:ubuntu-small-latest'"
```

Using 416 MB of source imagery

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `./make-cog.sh 'osgeo/gdal:ubuntu-small-latest'` | 36.234 ± 0.328 | 35.812 | 36.953 | 1.38 ± 0.02 |
| `./make-cog.sh 'gdal:thread-pool'` | 26.208 ± 0.325 | 25.728 | 26.665 | 1.00 |

Using 2.8GB source imagery

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `./make-cog.sh 'osgeo/gdal:ubuntu-small-latest'` | 244.502 ± 2.023 | 243.209 | 246.833 | 1.31 ± 0.01 |
| `./make-cog.sh 'gdal:thread-pool'` | 186.867 ± 1.292 | 185.393 | 187.804 | 1.00 |

where `gdal_tranlate` was

```
COMPRESS=webp
gdal_translate \
    -of COG \
    -co COMPRESS=$COMPRESS\
    -co NUM_THREADS=ALL_CPUS \
    -co BIGTIFF=YES \
    -co TILING_SCHEME=GoogleMapsCompatible \
    --config BIGTIFF_OVERVIEW YES \
    -co ALIGNED_LEVELS=3 \
    -co SPARSE_OK=TRUE \
    -co ADD_ALPHA=YES \
    -co BLOCKSIZE=512 \
    -co RESAMPLING=BILINEAR \
    $PWD/$TIF_FOLDER/.vrt $PWD/$TIF_FOLDER.c_$COMPRESS.google_3.master.tif
```

Ill am still running some larger COG generations on 100GB+ imagery will see what the difference is there.

rouault:
> Is that due to hyperthreading and the type of operations?

The scaling for those cases is limited by the acquistion of input data. The time to copy buffers from the interleaved TIFF tile to the GDAL per-band block cache, and then from the GDAL block cache to the input buffer of the overview becomes visible.

palmerj:
> The scaling for those cases is limited by the acquisition of input data. The time to copy buffers from the interleaved TIFF tile to the GDAL per-band block cache, and then from the GDAL block cache to the input buffer of the overview becomes visible.

Thanks Even. So even with large amounts of system RAM we would still expect this to be slow?

I'm do a COG generation test using an input 370GB Geotiff/LZW dataset with a 30GB RAM 16 core server (Intel Xeon(R) Platinum 8275CL CPU @ 3.00GHz), and fast SSD server. 

```
gdal_translate     
-of COG     \
-co COMPRESS=webp   \
-co NUM_THREADS=ALL_CPUS   \
-co BIGTIFF=YES     \
-co TILING_SCHEME=GoogleMapsCompatible     \
--config BIGTIFF_OVERVIEW YES     \
-co ALIGNED_LEVELS=3     \
-co SPARSE_OK=TRUE     \
-co ADD_ALPHA=YES     \
-co BLOCKSIZE=512     \
-co RESAMPLING=BILINEAR  \
-co NUM_THREADS=ALL_CPUS \
--config GDAL_CACHEMAX 4096 \
$PWD/$TIF_FOLDER/mosaic.vrt   \
 $PWD/$TIF_FOLDER.c_$COMPRESS.google_3.master.tif
```

I’m seeing about 300-400% CPU utilisation during the warping stage. 

Screen Shot 2020-05-23 at 10 37 30 AM

During the warping I’m not seeing much disk activity with gdal_translate. I see periods of 60M/S reads, followed by 25M/S writes, then a gap (I assume waiting for the CPU to do transforms and compression).

palmerj:
38% through now, just started overviews. So 6 hours to create the “warped” 250GB WebP file. Now in overview generation. The CPU usage seem much higher. Switches between 15% usage to over 80%

rouault:
> even with large amounts of system RAM we would still expect this to be slow?

Large amounts of system RAM aren't really needed for COG generation. I'm not clear if increasing them would improve, decrease or unchange performance.
In COG generation, I can think of several working buffers:
- the warping memory. It is let to its default of 64 MB (counted as source + target pixel data):  https://github.com/OSGeo/gdal/blob/master/gdal/alg/gdalwarpoperation.cpp#L558 by the COG driver. Sometimes increasing this buffer has been confirmed to decrease performance: especially when the shape of a rectangle in the output SRS is very far from the shape of a rectangle in the input SRS (reprojections close to poles), because allocating a large target area can require a huge amount of source pixels
- the overview computation. It could be very small in theory ( 2x2 input tiles = 1 output tile), but we use 10 MB if possible because it helped performance when using JPEG2000 as the input for example:  https://github.com/rouault/gdal/blob/thread_pool/gdal/gcore/overview.cpp#L3795. it is configurable for this branch, but was only/mostly thought as being useful for autotest, to make it very small, so that threading can kick off even on very small rasters
- the block cache. If your input and output are tiled, it could be let to a quite small value, as all processes involved in tiling, overview and final TIFF creation are optimized to proceed on groups of tiles (ie they are not scanline oriented). So I'd say you won't benefit much from a block cache that is substantially larger than (maybe 2x ?) of the max of the 2 above mentionned working buffers

blacha:
Finally finished benchmarking the creation of a larger COG

From 291GB of source imagery, including warping from NZTM into WebMercator 

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `./make-cog.sh 'osgeo/gdal:ubuntu-small-latest'` | 78402.431 ± 466.341 | 77923.918 | 78855.564 | 1.35 ± 0.01 |
| `./make-cog.sh 'gdal:thread-pool'` | 57968.307 ± 126.422 | 57870.935 | 58111.181 | 1.00 |

shows around a 35% speed up which is in line with the speed ups for the smaller imagery sets.

Issue #2598, opened 2020-05-23
Add gdal_translate -mo example


jidanni:
Using the gdal_edit.html example `-mo DATUM=WGS84` in
```
    gdal_translate -mo DATUM=WGS84 WMTS:/tmp/p.xml,layer=EMAP /tmp/p1.xml -of WMTS
```
will cause
```
    ERROR 6: Source dataset must be a WMTS dataset
```
Therefore on
https://github.com/OSGeo/gdal/blob/master/gdal/doc/source/programs/gdal_translate.rst
please add an actual working example. Don't just say `-mo META-TAG=VALUE`. Thanks.

rouault:
The WMTS driver only accepts translation from a WMTS dataset source, as indicated by the error message. This is a very particular case. By adding a switch like -mo you in fact wrap this WMTS source dataset into a temporary in-memory VRT file, which thus breaks the assumption of the CreateCopy() implementation of the WMTS driver. The doc of gdal_translate.rst is fine in the general case. Perhaps the doc of the WMTS driver could be enhanced to say that no switch like -mo should be specified. Feel free to submit a PR for that

jidanni:
(Well e.g., the average user is left even experimenting to see that maybe he can alter e.g.,
`<ZeroBlockOnServerException>`
via `-mo`. That's how little he can tell from what's there...)

jidanni:
I put it on the gdal_translate page because that is the one that 'tricks the user into thinking he can use -mo in the first place.'

jratike80:
Thank you for your many improvements to the documentation. However, I agree with Even and warning about not to use -mo with WMTS would rather belong to WMTS page.
There is an example about how to use -mo in the GTiff page https://gdal.org/drivers/raster/gtiff.html but it would be good to add a working example with some tag and value. Now the example is just `gdal_translate in.tif out.tif -mo {TAGNAME}=VALUE`.

PR #2602, opened 2020-05-23
Warn don't use -mo with WMTS

jidanni:
Vague warning for #2598.
Please add all the other items that -mo will ruin too.

### DIFF
--- a/gdal/doc/source/programs/gdal_translate.rst
+++ b/gdal/doc/source/programs/gdal_translate.rst
@@ -213,7 +213,7 @@ resampling, and rescaling pixels in the process.
 
 .. option:: -mo META-TAG=VALUE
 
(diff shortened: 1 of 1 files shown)


rouault:
> Please add all the other items that -mo will ruin too.

That is driver specific behaviour. gdal_translate.rst is about generic behaviour of the utility, and cannot address all driver specific behaviour. The way you write it is not precise enough furthermore: -mo shouldn't be used when using the WMTS driver as the output driver. It is fine if for example translating from WMTS to GeoTIFF and adding a metadata item.

This warning should rather be put close to the example of wmts.rst

jidanni:
OK... but you are 1000 to 10000 times more clear about this than me, so please put some warning where it belongs. Also I suspect it belongs on wms.rst too...

PR #2605, opened 2020-05-23
Fix runtime_error is not a member of std

chrodger:
This change allowed me to compile with VS 2019, build toolset 1926.

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:


rouault:
Backported to 3.1 branch
Strangely this builds fine with toolset 1925 without that patch

chrodger:
Interesting. I had MSVC 14.16.27023 and 14.16.28801 on my machine so I diffed their stdexcept headers just now. I'm not sure if those correspond to 1925 and 1926, but between those two at least I could not see any change that would have relocated runtime_error.

PR #2606, opened 2020-05-24
OSM: Fix tag filtering inconsistencies

dg0yt:
This series of commit provides two ways of customizing the early dropping of some tags:
- Disable the dropping globally, by enabling `report_all_tags` in `osmconf.ini`.
  This is in analogy to the existing `report_all_nodes` and `report_all_ways`. However, the per-type `ignore` continue to remove the tags they list.
- Disable the early dropping for attributes which are listed in the per-type `attribute` list. This fixes a surprising behaviour that some tags simply didn't work in the attribute list. However, the new behaviour means that a key listed as an `attribute`for just one type now is also handled for the other types, either by the other types' `ignore` and `attribute` lists, or appearing in `other_tags` or `all_tags`.

The variable of globally ignored keys is explicitly chosen as fixed-size array, not as vector, so that the compiler can do its best to optimize as before (e.g. unroll the filtering loop).

I hope I managed to match the coding style ;-)

### DIFF
--- a/gdal/data/osmconf.ini
+++ b/gdal/data/osmconf.ini
@@ -9,6 +9,11 @@ closed_ways_are_polygons=aeroway,amenity,boundary,building,craft,geological,hist
 # comment to avoid laundering of keys ( ':' turned into '_' )
 attribute_name_laundering=yes
 
+# Some tags, set on ways and when building multipolygons, multilinestrings or other_relations,
+# are normally filtered out early, independent of the 'ignore' configuration below.
+# Uncomment to disable early filtering. The 'ignore' lines below remain active.
+#report_all_tags=yes
+
 # uncomment to report all nodes, including the ones without any (significant) tag
 #report_all_nodes=yes
 
--- a/gdal/data/osmconf.ini
+++ b/gdal/data/osmconf.ini
@@ -34,7 +34,7 @@ attributes=name,barrier,highway,ref,address,is_in,place,man_made
 # keys that, alone, are not significant enough to report a node as a OGR point
 unsignificant=created_by,converted_by,source,time,ele,attribution
 # keys that should NOT be reported in the "other_tags" field
-ignore=created_by,converted_by,source,time,ele,note,openGeoDB:,fixme,FIXME
+ignore=created_by,converted_by,source,time,ele,note,todo,openGeoDB:,fixme,FIXME
 # uncomment to avoid creation of "other_tags" field
 #other_tags=no
 # uncomment to create "all_tags" field. "all_tags" and "other_tags" are exclusive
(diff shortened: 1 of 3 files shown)


gdal/ogr/ogrsf_frmts/osm/ogrosmdatasource.cpp
```diff
@@ -247,6 +247,7 @@ OGROSMDataSource::OGROSMDataSource() :
     nNodesInTransaction(0),
     nMinSizeKeysInSetClosedWaysArePolygons(0),
     nMaxSizeKeysInSetClosedWaysArePolygons(0),
+    m_ignoredKeys({"area","created_by","converted_by","note","todo","fixme","FIXME"}),
```
rouault, line 250:
```
ogrosmdatasource.cpp:250:20: error: suggest braces around initialization of subobject [-Werror,-Wmissing-braces]
    m_ignoredKeys({"area","created_by","converted_by","note","todo","fixme","FIXME"}),

                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

                   {                                                               }
```

gdal/data/osmconf.ini
```diff
@@ -9,6 +9,10 @@ closed_ways_are_polygons=aeroway,amenity,boundary,building,craft,geological,hist
 # comment to avoid laundering of keys ( ':' turned into '_' )
 attribute_name_laundering=yes
 
+# Some tags are normally filtered out early, independent of the 'ignore' configuration below.
```
rouault, line 12:
```suggestion
# Some tags, set on ways and when building multipolygons, multilinestrings or other_relations, are normally filtered out early, independent of the 'ignore' configuration below.
```

gdal/ogr/ogrsf_frmts/osm/ogrosmdatasource.cpp
```diff
@@ -2053,20 +2054,11 @@ void OGROSMDataSource::NotifyWay( OSMWay* psWay )
...
-            if( strcmp(pszK, "todo") == 0 )
-                continue;
-            if( strcmp(pszK, "fixme") == 0 )
-                continue;
-            if( strcmp(pszK, "FIXME") == 0 )
-                continue;
+            for( const char* pszIgnoredKey : m_ignoredKeys )
+            {
+                if( strcmp(pszK, pszIgnoredKey) == 0 )
+                    continue;
```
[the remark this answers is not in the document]
dg0yt: Indeed. Reverting to std::any_of + lambda. Actually I start with that, instead of a plain loop.

gdal/ogr/ogrsf_frmts/osm/ogrosmdatasource.cpp
dg0yt: Done.

gdal/data/osmconf.ini
dg0yt: Done.

Issue #2614, opened 2020-05-26
Mapinfo: list of CRSs


rouault:
See https://github.com/OSGeo/gdal/issues/2612#issuecomment-633744226 for a mapping between EPSG code and MapInfo CRS strings

PR #2618, opened 2020-05-27
Add MapInfo MRR(Multi Resolution Raster) raster driver.

miteshpb:
## What does this PR do?
Add MapInfo MRR(Multi Resolution Raster) raster driver. This driver will not be enabled as default compile. It could be enabled as Plugin(autoload) driver, or standard driver as needed.

This driver depends on MapInfo Raster runtime binaries. These binaries are free to download and use. Documentation on obtaining the runtime binaries for various platforms has been updated with PR.

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:


nyalldawson:
Reiterating my thoughts from https://github.com/OSGeo/gdal/pull/1882

I'm extremely disappointed to see that you persist with the closed source SDK approach. You are doing your users a great disservice with this approach. It's a shortsighted, paranoid and frankly outdated approach. I would strongly recommend that you reconsider this if you intend for your formats to be widely used outside of MapInfo itself.

rouault:
Note: I don't think pushing commits in this PR makes sense while you haven't adressed the fundamental remark of https://github.com/OSGeo/gdal/pull/2618#issuecomment-634491529

carypeebles:
We understand and share the desire to open the source for this format but are unable to do so at this time. However, we wish to provide our clients who already have their data in the MRR format and who already have the SDK the ability to share the data in the broadest suite of technology possible.

nyalldawson:
> We understand and share the desire to open the source for this format but are unable to do so at this time. However, we wish to provide our clients who already have their data in the MRR format and who already have the SDK the ability to share the data in the broadest suite of technology possible.

In that case I would strongly suggest that the GDAL project maintainers close this PR as a "wontfix".

In its current form this PR offers absolutely nothing to the project, except for added maintenance burden. You are trying to push work onto the GDAL project and offering them absolutely nothing in return. Basically, what you are trying to do is legitimise your proprietary, patent-encumbered MRR format by being able to say that it's supported by GDAL.

I think we have an obligation to our users to protect them from this. The MRR format offers nothing of real value that isn't available through other alternative, and OPEN or STANDARDS BASED formats. GDAL (and open source geo in general) should NOT be encouraging anyone to use this format -- it's a trap. (I'd even go as far as to say that we have a moral obligation to make it as hard as possible for people to use this format, so that they are forced to use the open formats instead).

If you're set on the current approach and want to see this work land in gdal, then you need to build up social capital with the project first BEFORE trying to sling selfish work like this onto the project. There's **heaps** of ways you could do this! Some suggestions:

- financially support the project with large, ongoing sponsorship
- contribute to the open Mapinfo TAB driver, and submit fixes for that. 
- provide the missing documentation relating to MapInfo's datum tables and the correct way to map these to standard EPSG codes (see e.g. https://github.com/OSGeo/gdal/issues/2614)

These are all **easy** ways for your company to show your commitment to the GDAL project in advance, and show that you are willing to **work with** GDAL. Do that first, build your reputation, prove your worth, and THEN come back with generally undesirable changes like this one. You'll find you get a much more receptive response from the community...

ianturton:
I gave a talk about this sort of thing at FOSS4G - you can watch the [video](https://vimeo.com/144089061) if you want to get a better idea.

idanmiara:
> I think we have an obligation to our users to protect them from this. The MRR format offers nothing of real value that isn't available through other alternative, and OPEN or STANDARDS BASED formats. GDAL (and open source geo in general) should NOT be encouraging anyone to use this format -- it's a trap. (I'd even go as far as to say that we have a moral obligation to make it as hard as possible for people to use this format, so that they are forced to use the open formats instead).

@miteshpb 
As one of your clients, I would HIGHLY APPRICIATE you add native GeoTiff input/output support in your software (in addition to your NWT_GRD and MRR formats) using GDAL for maximum compatibility.

Until that happens and/or until your contribution to GDAL is accepted by the community, please regularly update your [MRR-GDAL plugin](https://support.precisely.com/product-downloads/item/mapinfo-mrr-sdk-download/) to the latest released GDAL version.
Currently it only supports GDAL 3.0 and makes it hard for us to upgrade to newer GDAL versions.

Thanks!

idanmiara:
Hi @miteshpb!
In case you are still working on project, please note:
https://github.com/OSGeo/gdal/pull/5128

PR #2622, opened 2020-05-27
GPKG: unique constraints detection

elpaso:
## What does this PR do?

This PR adds a `IsUnique` `SetUnique` methods to the fields definition API

The feature will be useful for example in QGIS to recognize constraints from OGR data sources:
https://github.com/qgis/QGIS/issues/36468


rouault:
For the SWIG conflicts, is your master of GDAL recent enough ? I see 3.0.8 was what we used until last June.

gdal/ogr/ogrsf_frmts/gpkg/ogrgeopackagetablelayer.cpp
```diff
@@ -806,6 +809,79 @@ OGRErr OGRGeoPackageTableLayer::ReadTableDefinition()
...
+        SQLResultFree(&oResultTable);
+        return OGRERR_FAILURE;
+    }
+
+    std::set<std::string> uniqueFields;
+    // Match identifiers with " or ` or no delimiter (and no spaces).
+    static const std::regex sFieldIdentifierRe { R"raw(\s*(["`]([^"`]+)["`])|(([^\s]+)\s).*)raw" };
+    std::string tableDefinition { SQLResultGetValue(&oResultTable, 0, 0) };
+    tableDefinition = tableDefinition.substr(tableDefinition.find('('), tableDefinition.rfind(')') );
+    std::stringstream tableDefinitionStream { tableDefinition };
```
rouault, line 835: #include <sstream> needed

gdal/ogr/ogrsf_frmts/gpkg/ogrgeopackagetablelayer.cpp
```diff
@@ -32,6 +32,9 @@
 #include "ogrsqliteutility.h"
 #include "cpl_time.h"
 #include "ogr_p.h"
+#if !defined(__GNUC__) || defined(__clang__) || __GNUC__ >= 5
```
rouault, line 35: minor remark: you could probably drop the #if here. <regex> header should be available even if the implementation is buggy

gdal/ogr/ogrsf_frmts/gpkg/ogrgeopackagetablelayer.cpp
```diff
@@ -943,6 +1019,14 @@ OGRErr OGRGeoPackageTableLayer::ReadTableDefinition()
                 oField.SetWidth(nMaxWidth);
                 if( bNotNull )
                     oField.SetNullable(FALSE);
+
+#if !defined(__GNUC__) || defined(__clang__) || __GNUC__ >= 5
+                if ( uniqueFields.find( std::string( pszName ) ) != uniqueFields.end() )
+                {
+                  oField.SetUnique(TRUE);
```
rouault, line 1026: 4-space indentation (sorry there's no reformating script)
```suggestion
                    oField.SetUnique(TRUE);
```

gdal/ogr/ogrsf_frmts/gpkg/ogrgeopackagetablelayer.cpp
```diff
@@ -806,6 +809,79 @@ OGRErr OGRGeoPackageTableLayer::ReadTableDefinition()
...
+    std::set<std::string> uniqueFields;
+    // Match identifiers with " or ` or no delimiter (and no spaces).
+    static const std::regex sFieldIdentifierRe { R"raw(\s*(["`]([^"`]+)["`])|(([^\s]+)\s).*)raw" };
+    std::string tableDefinition { SQLResultGetValue(&oResultTable, 0, 0) };
+    tableDefinition = tableDefinition.substr(tableDefinition.find('('), tableDefinition.rfind(')') );
+    std::stringstream tableDefinitionStream { tableDefinition };
+    std::smatch uniqueFieldMatch;
+    while (tableDefinitionStream.good()) {
+      std::string fieldStr;
+      getline( tableDefinitionStream, fieldStr, ',' );
```
rouault, line 839: std::getline ?

gdal/ogr/ogrsf_frmts/gpkg/ogrgeopackagetablelayer.cpp
```diff
@@ -806,6 +809,79 @@ OGRErr OGRGeoPackageTableLayer::ReadTableDefinition()
...
+        else
+            CPLError( CE_Failure, CPLE_AppDefined, "Cannot find table %s", m_pszTableName );
+
+        SQLResultFree(&oResultTable);
+        return OGRERR_FAILURE;
+    }
+
+    std::set<std::string> uniqueFields;
+    // Match identifiers with " or ` or no delimiter (and no spaces).
+    static const std::regex sFieldIdentifierRe { R"raw(\s*(["`]([^"`]+)["`])|(([^\s]+)\s).*)raw" };
```
rouault, line 832: I don't think there's a guarantee that the CREATE TABLE will have column identifiers quoted. GDAL will generate such files. But files from other sources could have unquoted identifiers. Not necessarily critical to address this, but perhaps add a comment

gdal/ogr/ogrsf_frmts/gpkg/ogrgeopackagetablelayer.cpp
```diff
@@ -806,6 +809,79 @@ OGRErr OGRGeoPackageTableLayer::ReadTableDefinition()
...
+    std::string tableDefinition { SQLResultGetValue(&oResultTable, 0, 0) };
+    tableDefinition = tableDefinition.substr(tableDefinition.find('('), tableDefinition.rfind(')') );
+    std::stringstream tableDefinitionStream { tableDefinition };
+    std::smatch uniqueFieldMatch;
+    while (tableDefinitionStream.good()) {
+      std::string fieldStr;
+      getline( tableDefinitionStream, fieldStr, ',' );
+      if( fieldStr.find( "UNIQUE" ) != std::string::npos )
```
rouault, line 840: perhaps test in case insensitive way ?

gdal/ogr/ogrsf_frmts/gpkg/ogrgeopackagetablelayer.cpp
```diff
@@ -806,6 +809,79 @@ OGRErr OGRGeoPackageTableLayer::ReadTableDefinition()
...
+          const std::string quoted { uniqueFieldMatch.str( 2 ) };
+          uniqueFields.insert( quoted.length() ? quoted:  uniqueFieldMatch.str( 4 ) );
+        }
+      }
+    }
+    SQLResultFree(&oResultTable);
+
+    // Search indexes:
+    pszTableDefinitionSQL = sqlite3_mprintf("SELECT sql FROM sqlite_master WHERE type='index' AND"
+                                            " tbl_name='%q' AND sql LIKE 'CREATE UNIQUE INDEX%%'", m_pszTableName);
```
rouault, line 853:
```suggestion
                                            " tbl_name='%q' AND sql ILIKE 'CREATE UNIQUE INDEX%%'", m_pszTableName);
```

gdal/ogr/ogrsf_frmts/gpkg/ogrgeopackagetablelayer.cpp
```diff
@@ -806,6 +809,79 @@ OGRErr OGRGeoPackageTableLayer::ReadTableDefinition()
...
+            CPLError( CE_Failure, CPLE_AppDefined, "Error searching indexes for table %s", m_pszTableName );
+
+    }
+    else if (oResultTable.nRowCount >= 0 )
+    {
+      static const std::regex sFieldIndexIdentifierRe { R"raw(\(\s*[`"]?([^",`\)]+)["`]?\s*\))raw" };
+      for( int rowCnt = 0; rowCnt < oResultTable.nRowCount; ++rowCnt )
+      {
+        std::string indexDefinition { SQLResultGetValue(&oResultTable, 0, rowCnt) };
+        if ( indexDefinition.find( "UNIQUE" ) != std::string::npos )
```
rouault, line 871: case insensitive comparison ?

gdal/ogr/ogrsf_frmts/gpkg/ogrgeopackagetablelayer.cpp
```diff
@@ -943,6 +1019,14 @@ OGRErr OGRGeoPackageTableLayer::ReadTableDefinition()
                 oField.SetWidth(nMaxWidth);
                 if( bNotNull )
                     oField.SetNullable(FALSE);
+
+#if !defined(__GNUC__) || defined(__clang__) || __GNUC__ >= 5
+                if ( uniqueFields.find( std::string( pszName ) ) != uniqueFields.end() )
```
rouault, line 1024: if you move uniqueFields declaration before the #if where the job is done, then you could avoid it here

gdal/ogr/ogrsf_frmts/gpkg/ogrgeopackagetablelayer.cpp
elpaso: my idea was to support the three possibilities: unquoted, double quoted and backticks, I should have added test cases for all three.

gdal/ogr/ogrsf_frmts/gpkg/ogrgeopackagetablelayer.cpp
elpaso: I tried to use mixed case UniQue when creating the table and the index and it always got back upper case but I don't really know if it was my application (`sqlitemanager`) that did the trick, I'll test case insensitive.

gdal/ogr/ogrsf_frmts/gpkg/ogrgeopackagetablelayer.cpp
jratike80: Where comes the backticks alternative? Just curious, I have never seen such yet.

gdal/ogr/ogrsf_frmts/gpkg/ogrgeopackagetablelayer.cpp
rouault: it is probably sqlite3 itself normalizing this as the below experiment seems to show:
```
$ sqlite3 foo.db
SQLite version 3.11.0 2016-02-15 17:29:24
Enter ".help" for usage hints.
sqlite> create table foo(bar integer);
sqlite> create unique index idx on foo(bar);
sqlite> 

$ sqlite3 foo.db
SQLite version 3.11.0 2016-02-15 17:29:24
Enter ".help" for usage hints.
sqlite> select * from sqlite_master;
table|foo|foo|2|CREATE TABLE foo(bar integer)
index|idx|foo|3|CREATE UNIQUE INDEX idx on foo(bar)

```

gdal/ogr/ogrsf_frmts/gpkg/ogrgeopackagetablelayer.cpp
jratike80: I see, SQLite supports unfortunately backticks because MySQL supports them. How about brackets then? 
https://stackoverflow.com/questions/25141090/sqlite-use-backticks-or-double-quotes-with-python.

gdal/ogr/ogrsf_frmts/gpkg/ogrgeopackagetablelayer.cpp
elpaso: ok, so no need to test case insensitive? Or better do it just in case different sqlite versions handle this differently?

elpaso:
>     * GDAL_DCAP_UNIQUE_FIELDS

Where exactly in my code?

autotest/ogr/ogr_gpkg.py
```diff
@@ -3884,6 +3884,80 @@ def test_ogr_gpkg_json():
...
+    fldDef = layerDefinition.GetFieldDefn(2)
+    assert fldDef.IsUnique()
+
+    ds = None
+
+    gdal.Unlink('/vsimem/ogr_gpkg_unique.gpkg')
+
+    # Now test loading the same layer from disk, constraints should be
+    # automatically set on field 2
+    ds = ogr.Open('data/ogr_gpkg_unique.gpkg')
```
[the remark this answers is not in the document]
elpaso: I actually need to recreate the whole table from SQL because the create instruction explicitly test for different quoting styles and spacings in fields and index names.

I'm trying but I cannot get it done:

```python
    # Now create another layer from SQL to test quoting of fields
    # and indexes
    # Note: leave create table in a single line because of regex spaces testing
    sql = """
    CREATE TABLE IF NOT EXISTS "test2" ( "fid" INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL, "field_default" TEXT, "field_no_unique" TEXT, "field_unique" TEXT UNIQUE,`field unique2` TEXT UNIQUE,field_unique3 TEXT UNIQUE, "field_unique_index" TEXT, `field unique index2`, "field_unique_index3" TEXT);
    CREATE UNIQUE INDEX test2_unique_idx ON test2(field_unique_index);
    CREATE UNIQUE INDEX test2_unique_idx2 ON test2(`field unique index2`);
    CREATE UNIQUE INDEX test2_unique_idx3 ON test2("field_unique_index3");
    INSERT INTO gpkg_contents VALUES('test2','attributes','test2','','2020-05-27T12:27:30.136Z',NULL,NULL,NULL,NULL,0);
    INSERT INTO gpkg_ogr_contents VALUES('test2',NULL);
    """

    ds.ExecuteSQL(sql)

    # Now test loading the same layer from disk, check layer test2
    ds = ogr.Open('/tmp/ogr_gpkg_unique.gpkg')

    assert ds.GetLayerCount() == 2   # <----- fails here!
```

jratike80:
> > ```
> > * GDAL_DCAP_UNIQUE_FIELDS
> > ```
> 
> Where exactly in my code?

ogrfielddefn.cpp
line 1279
line 1298

elpaso:
> > > ```
> > > * GDAL_DCAP_UNIQUE_FIELDS
> > > ```
> > 
> > 
> > Where exactly in my code?
> 
> ogrfielddefn.cpp
> line 1279
> line 1298

Thanks, found it, I didn't get that it was in doc text.

gdal/ogr/ogrsf_frmts/gpkg/ogrgeopackagetablelayer.cpp
```diff
@@ -850,7 +850,7 @@ OGRErr OGRGeoPackageTableLayer::ReadTableDefinition()
 
     // Search indexes:
     pszTableDefinitionSQL = sqlite3_mprintf("SELECT sql FROM sqlite_master WHERE type='index' AND"
-                                            " tbl_name='%q' AND sql LIKE 'CREATE UNIQUE INDEX%%'", m_pszTableName);
+                                            " tbl_name='%q' AND sql ILIKE 'CREATE UNIQUE INDEX%%'", m_pszTableName);
```
elpaso, line 853: @rouault I'm reverting the ILIKE: it's not supported by sqlite.

elpaso:
> For the SWIG conflicts, is your master of GDAL recent enough ? I see 3.0.8 was what we used until last June.

Yeah, maybe, I'm pretty sure I remembered to check before branching but I probably didn't :(

So, to clean this mess I think I'll just close this PR and make another topic branch then cherry-pick unless you have a better (and quicker!) idea.

gdal/ogr/ogrsf_frmts/gpkg/ogrgeopackagetablelayer.cpp
rouault: not here at least. but if doing "create table foo ( bar integer unique);", unique gets back at lowercase when reading sqlite_master

autotest/ogr/ogr_gpkg.py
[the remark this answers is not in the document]
rouault: or put the various lines in a python array and iterate over it

rouault:
> I think I'll just close this PR and make another topic branch

you don't necessarily need to close the PR. You can update it with a workflow like
```
git checkout -b gpkg-unique-backup
git checkout master
git pull
git branch -D gpkg-unique
git checkout -b gpkg-unique
git cherry-pick ....
git push -f elpaso gpkg-unique
```

gdal/ogr/ogrsf_frmts/gpkg/ogrgeopackagetablelayer.cpp
elpaso: Ok, I've implemented case insensitive already.

rouault:
the updated SWIG regenerated files should be committed and pushed as well

elpaso:
Almost there: there are a few build failures on osx, android and trusty... is there anything I should do?

rouault:
- Windows build failures on https://ci.appveyor.com/project/OSGeo/gdal/builds/33190013/job/87vqb4hp6f142uqm . Probably missing #include <sstream>
- the crash on trusty_clang on https://travis-ci.com/github/OSGeo/gdal/jobs/341263190 is annoying. I believe this is due to it using clang but linking against libstdc++ of Trusty. What I thought to be a defect of gcc < 5 is probably more a defect of libstdc++ of gcc < 5. But this is a bit of a odd configuration. I'd suggest to just add an extra skip for "trusty_clang" in the test related on UNIQUE but ... there's also a crash on trusty_32 bit on https://travis-ci.com/github/OSGeo/gdal/jobs/341263197 . This happens clearly on test_ogr_gpkg_4 test (and this is also likely the case for trusty_clang due to it being at the 4 '.'), which is not UNIQUE specific. I now realize that UNIQUE detection happens always.  Perhaps moving static const std::regex sFieldIdentifierRe  from line 844 to 856 and static const std::regex sFieldIndexIdentifierRe  from line 881 to 889 would be sufficient to avoid regex machinery to be used unless there's UNIQUE constraints in the .gpkg. (The "cleanest" solution would probably to have some ./configure test, but don't bother about that)
- android build failure in https://travis-ci.com/github/OSGeo/gdal/jobs/341263193 . Probably same cause as the Windows one.
- osx build failure: same as Android

rouault:
Note to self: after reflection, I didn't understand the crash on trusty_32bit. But I see now that trusty_32bit also uses the clang + libstdc++ combinations (for faster builds than with gcc), hence the reason for why the std::regex code was hit. So everything has now a logical explanation.

rouault:
MSVC still emitting a warning
```
C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.25.28610\include\algorithm(1310): error C2220: the following warning is treated as an error (compiling source file ogrgeopackagetablelayer.cpp)
C:\projects\gdal\gdal\ogr\ogrsf_frmts\gpkg\ogrgeopackagetablelayer.cpp(826): note: see reference to function template instantiation '_OutIt std::transform<std::_String_iterator<std::_String_val<std::_Simple_types<_Elem>>>,std::_String_iterator<std::_String_val<std::_Simple_types<_Elem>>>,int(__cdecl *)(int)>(const _InIt,const _InIt,_OutIt,_Fn)' being compiled
        with
        [
            _OutIt=std::_String_iterator<std::_String_val<std::_Simple_types<char>>>,
            _Elem=char,
            _InIt=std::_String_iterator<std::_String_val<std::_Simple_types<char>>>,
            _Fn=int (__cdecl *)(int)
        ]
C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.25.28610\include\algorithm(1310): warning C4244: '=': conversion from 'int' to 'char', possible loss of data (compiling source file ogrgeopackagetablelayer.cpp)
```
So the issue is with `std::transform(upperTableName.begin(), upperTableName.end(), upperTableName.begin(), ::toupper);`. I suspect the reason is that toupper() returns a int, whereas a std::string is an "array" of char.
You might probably want to use the CPLString class (see port/cpl_string.h), which extends std::string, and has a toupper() method
Or wrap the ::toupper() in a lamba doing the cast to char.

elpaso:
Yes!!! Thank you @rouault for your help.

rouault:
Eh, if git log doesn't lie, this is your first GDAL commit ! Sharing virtual geobeer :-)

Issue #2615, opened 2020-05-26
LVBAG: Heap-use-after-free


rouault:
From https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=22474

```
Crash Type: Heap-use-after-free READ 4
Crash Address: 0x625000010554
Crash State:
  std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<ch
  OGRLVBAGLayer::DataHandlerCbk
  OGRLVBAGLayer::ConfigureParser
```

Can be reproduced with:
```
$ valgrind ogrinfo clusterfuzz-testcase-minimized-lvbag_fuzzer-5717169316036608.fuzz.xml  -al -q
[...]
ERROR 1: Parsing of LV BAG file failed : mismatched tag at line 57, column 18
ERROR 1: Parsing of LV BAG file failed : mismatched tag at line 57, column 18
ERROR 1: Parsing of LV BAG file failed : mismatched tag at line 57, column 18
ERROR 1: Parsing of LV BAG file failed : mismatched tag at line 57, column 18
==15715== Invalid read of size 8
==15715==    at 0xF503A8C: ??? (in /lib/x86_64-linux-gnu/libexpat.so.1.6.0)
==15715==    by 0xF50439B: ??? (in /lib/x86_64-linux-gnu/libexpat.so.1.6.0)
==15715==    by 0xF505CDD: ??? (in /lib/x86_64-linux-gnu/libexpat.so.1.6.0)
==15715==    by 0xF50643A: ??? (in /lib/x86_64-linux-gnu/libexpat.so.1.6.0)
==15715==    by 0xF50875A: XML_ParseBuffer (in /lib/x86_64-linux-gnu/libexpat.so.1.6.0)
==15715==    by 0x63B178B: OGRLVBAGLayer::ParseDocument() (ogrlvbaglayer.cpp:577)
==15715==    by 0x63B1849: OGRLVBAGLayer::GetNextRawFeature() (ogrlvbaglayer.cpp:612)
==15715==    by 0x63B19D1: OGRGetNextFeatureThroughRaw<OGRLVBAGLayer>::GetNextFeature() (ogrsf_frmts.h:313)
==15715==    by 0x63B18E5: OGRLVBAGLayer::GetNextFeature() (ogr_lvbag.h:119)
[...]

PR #2626, opened 2020-05-28
FIX: Continue after error in parsing document

yorickdewid:
Issue: https://github.com/OSGeo/gdal/issues/2615

Parsing should not be able to continue once an error occurred in the process. Feature building will need the layer definition at some point. Check if the layer definition is complete, if not the try once and then fail (`GetNextFeature()`). We cannot rely on the forward parsing pass to build the layer definition when features are requested as done in the previous version.

### DIFF
--- a/gdal/ogr/ogrsf_frmts/lvbag/ogrlvbaglayer.cpp
+++ b/gdal/ogr/ogrsf_frmts/lvbag/ogrlvbaglayer.cpp
@@ -598,6 +598,26 @@ void OGRLVBAGLayer::ParseDocument()
 }
 
 /************************************************************************/
+/*                           GetNextFeature()                           */
+/************************************************************************/
+
+OGRFeature* OGRLVBAGLayer::GetNextFeature()
+{
+    if ( !bHasReadSchema )
+    {
+        GetLayerDefn();
+        if ( !bHasReadSchema )
+        {
+            CPLError(CE_Failure, CPLE_AppDefined,
+                "Parsing LV BAG extract failed : invalid layer definition");
+            return nullptr;
+        }
+    }
+
+    return OGRGetNextFeatureThroughRaw<OGRLVBAGLayer>::GetNextFeature();
+}
+
+/************************************************************************/
 /*                         GetNextRawFeature()                          */
 /************************************************************************/
 
--- a/gdal/ogr/ogrsf_frmts/lvbag/ogrlvbaglayer.cpp
+++ b/gdal/ogr/ogrsf_frmts/lvbag/ogrlvbaglayer.cpp
@@ -108,7 +108,7 @@ OGRFeatureDefn* OGRLVBAGLayer::GetLayerDefn()
 /*                            XMLTagSplit()                             */
 /************************************************************************/
 
-static inline const char* XMLTagSplit(const char *pszName)
+static inline const char* XMLTagSplit( const char *pszName )
 {
     const char *pszTag = pszName;
     const char *pszSep = strchr(pszTag, ':');
@@ -184,7 +184,7 @@ void OGRLVBAGLayer::AddOccurrenceFieldDefn()
 /*                         CreateFeatureDefn()                          */
 /************************************************************************/
 
-void OGRLVBAGLayer::CreateFeatureDefn(const char *pszDataset)
+void OGRLVBAGLayer::CreateFeatureDefn( const char *pszDataset )
 {
     if( EQUAL("pnd", pszDataset) )
     {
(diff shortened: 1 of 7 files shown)


yorickdewid:
Hold off on merge, chasing more memory leaks

rouault:
It could be interesting to add a test with a broken file (not necessarily the ones generated by oss-fuzz, but something created at hand from a sane file and with just the amount of corruption needed to exercice the error code path)

yorickdewid:
Yes I'll do. Running AFL-fuzz now in addition to OSS-Fuz.

yorickdewid:
@rouault I need some suggestions. Dataset directories/archives for this driver can span thousands of files. However opening that many files will hit the linux open file descriptor soft limit. Each layer needs access to the corresponding `fp` meaning I cannot close the `fp` per layer. Can we instruct union layer to copy feature data to some sort of backing temp file, so it can close the underlaying `fp`?

Now of course I can set the ulimit, but I prefer not to.

rouault:
> Dataset directories/archives for this driver can span thousands of files. However opening that many files will hit the linux open file descriptor soft limit

You could possibly look at ogr/ogrsf_frmts/generic/ogrlayerpool.h . It is used by the Shapefile driver in a similar situation. Basically a OGRLayerPool object takes care of calling the CloseUnderlyingLayer() method on a class deriving from OGRAbstractProxiedLayer when there are too many files opened regarding the pool size. Those layers must make sure of reopening the file descriptor when needed and registring themselves to the OGRLayerPool with SetLastUsedLayer() each time they are accessed.

rouault:
ogr/ogrsf_frmts/lvbag/ogr_lvbag.h:138,style,unusedPrivateFunction,Unused private function: 'OGRLVBAGDataSource::ConcludeBatch'

You may want to merge back latest master or rebase on top of it to solve unrelated issues with continuous integration

yorickdewid:
Pooling & union is tested with 3k files, no memory leaks, looks like it works. Still running more tests

gdal/ogr/ogrsf_frmts/lvbag/ogrlvbaglayer.cpp
```diff
@@ -316,6 +324,53 @@ void OGRLVBAGLayer::DataHandlerCbk( const char *data, int nLen )
...
+    fp = VSIFOpenExL(osFilename, "rb", true);
+    if( !fp )
+    {
+        CPLError(CE_Warning, CPLE_AppDefined,
+            "Opening LV BAG extract failed : %s", osFilename.c_str());
+        eFileDescriptorsState = FD_CANNOT_REOPEN;
+        return false;
+    }
+    
+    eFileDescriptorsState = FD_OPENED; // TODO: Why?
```
rouault, line 356: to answer the "TODO: Why" ? This prevents re-opening of the file descriptor the next time TouchLayer() is called.  We could just do that by testing fp state. But here the tri-state with FD_CANNOT_REOPEN prevents from trying to re-open when it already failed

yorickdewid: Was planning on testing fp is null, good point.

yorickdewid:
Pending CI, but should be it

rouault:
Doc issue:
/__w/1/s/gdal/doc/source/drivers/vector/lvbag.rst:45: WARNING: Title underline too short.

yorickdewid:
If I ignore `GetMaxSimultaneouslyOpened()` and always list layers in the MRU things get a little less error prone in combination with unionizing. Is there any objection? Pool will have minimal overhead on  layercount < `GetMaxSimultaneouslyOpened()` I suppose.

rouault:
> and always list layers in the MRU things get a little less error prone in combination with unionizing.

what happens exactly ?

yorickdewid:
> > and always list layers in the MRU things get a little less error prone in combination with unionizing.
> 
> what happens exactly ?

This happens:
```cpp
    if( static_cast<int>(papoLayers.size()) ==
        poPool->GetMaxSimultaneouslyOpened() && poPool->GetSize() == 0 )
    {
        poPool->SetLastUsedLayer(...)
    }
```

I refer to `OGRShapeDataSource::AddLayer` in the comment above `ogrlvbagdatasource.cpp:64`. It says something about how layers/pools can be accessed in a multithreaded env in an older version of GDAL. However I am not sure if this is specific for Shapefile or not, which is why I wrote the same check.

rouault:
> ``. It says something about how layers/pools can be accessed in a multithreaded env in an older version of GDAL

ah, don't bother about this regarding LVBAG. Accessing layers of the same dataset handle has never been guaranteed to be possible from multiple threads. The shapefile driver used to be safe regarding this before introduction of layer pools, but this has never been guaranteed to be safe.

yorickdewid:
Noticed a possible issue, but not going to fix in this merge. I'll open an issue right after this merge

PR #2634, opened 2020-06-01
OGRShapeLayer is final, cannot inherit

yorickdewid:
Many methods are marked as `virtual` within `OGRShapeLayer`. Is this pure for readability? If not they can all be removed, like this protected keyword. The class is marked final.

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments


rouault:
> Is this pure for readability?

this dates back to pre C++11 where override wasn't available to indicate a method is virtual.

yorickdewid:
IIRC, C++11 is now _required_ for GDAL? Can I submit a PR when I come across these nonsensical virtuals?

rouault:
> IIRC, C++11 is now _required_ for GDAL?

yes

> Can I submit a PR when I come across these nonsensical virtuals?

sure

PR #2635, opened 2020-06-01
Memory safe alternatives for C++11 > era

yorickdewid:
Smart Pointers where I could have used them. By no means complete.

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/gdal/gcore/gdal_priv.h
+++ b/gdal/gcore/gdal_priv.h
@@ -864,7 +864,7 @@ struct CPL_DLL GDALDatasetUniquePtrDeleter
  * reference counter has not been manually modified.
  * @since GDAL 2.3
  */
-typedef std::unique_ptr<GDALDataset, GDALDatasetUniquePtrDeleter> GDALDatasetUniquePtr;
+using GDALDatasetUniquePtr = std::unique_ptr<GDALDataset, GDALDatasetUniquePtrDeleter>;
 
 /* ******************************************************************** */
 /*                           GDALRasterBlock                            */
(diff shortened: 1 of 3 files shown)


mloskot:
A note about `typedef` vs `using`, do you have any authoritative source of the deprecation potential? The `typedef` is a core feature of the language and I highly doubt it will ever be deprecated. There simply is no reason to do so. 

The `using` is just an alternative syntax to introduce a typedef-name. Both, `typedef` and `using` introduce a type alias. `typedef A B` means `A` is an alias of `B` as.
So, saying in the commit message that "Aliasing is preferred over typedefs" is misleading.

https://eel.is/c++draft/dcl.typedef
> 9.2.3 The typedef specifier 
> 1 A name declared with the typedef specifier becomes a typedef-name.
> 2 A typedef-name can also be introduced by an alias-declaration.

yorickdewid:
> A note about `typedef` vs `using`, do you have any authoritative source of the deprecation potential? The `typedef` is a core feature of the language and I highly doubt it will ever be deprecated. There simply is no reason to do so.

No I have not, nor is does it exist. It has been discussed forever in backchannels but that's just chat. I do believe it will be deprecated in the future eventually, but we'll see.

> The `using` is just an alternative syntax to introduce a typedef-name. Both, `typedef` and `using` introduce a type alias. `typedef A B` means `A` is an alias of `B` as.
> So, saying in the commit message that "Aliasing is preferred over typedefs" is misleading.

It's not misleading, `using` is, in gerenal, prefered over `typedef` because `using` is a generalization of typedef. For example, `using` can take template parameters where `typedef` cannot. There is no (semantic) reason to keep `typedef` around other than backwards compatibility.

Issue #1726, opened 2019-07-16
C# bindings not compatible of SWIG 4.0.0


rouault:
Reported initially in #1725
Failures like
```
[00:32:58] 	swig.exe -DSWIG2_CSHARP -c++ -csharp -namespace OSGeo.GDAL  -dllimport gdal_wrap -I../../include/csharp -o gdal_wrap.cpp ../../include/gdal.i
[00:32:59] ..\..\include\csharp\swig_csharp_extensions.i(204) : Error: A deprecated csdestruct typemap was found for RasterIOExtraArg, please remove it and replace all csdestruct, csdestruct_derived and csfinalize typemaps by the csdispose, csdispose_derived, csdisposing and csdisposing_derived typemaps.
```

runette:
@rouault - I was reviewing issues to see if there was anything else that should be included in #5046 .

Since the CMake scripts are building on SWIG4 - should this issues be closed?

rouault:
Let's close this, although it doesn't look we know why this works now...

PR #2637, opened 2020-06-01
WIP: Swig 4.0 compatibility

bjornharrtell:
This is an attempt to fix https://github.com/OSGeo/gdal/issues/1726. See https://github.com/swig/swig/issues/421 for info about the changes.

### DIFF
--- a/gdal/swig/include/csharp/typemaps_csharp.i
+++ b/gdal/swig/include/csharp/typemaps_csharp.i
@@ -33,6 +33,45 @@
 
 /* CSHARP TYPEMAPS */
 
+%typemap(csdispose) SWIGTYPE %{
+  ~$csclassname() {
+    Dispose();
+  }
+%}
+
+%typemap(csdispose_derived) SWIGTYPE %{
+  ~$csclassname() {
+    Dispose();
+  }
+%}
+
+%typemap(csdisposing, methodname="Dispose", methodmodifiers="public", parameters="") SWIGTYPE {
+    lock(this) {
+      if (swigCPtr.Handle != global::System.IntPtr.Zero) {
+        if (swigCMemOwn) {
+          swigCMemOwn = false;
+          $imcall;
+        }
+        swigCPtr = new global::System.Runtime.InteropServices.HandleRef(null, global::System.IntPtr.Zero);
+      }
+      global::System.GC.SuppressFinalize(this);
+    }
+  }
+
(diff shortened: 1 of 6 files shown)


rouault:
An alternate fix would probably be to just add #if SWIG_VERSION < 0x040000 / #endif around the original typemaps, as done in
https://github.com/mapserver/mapserver/pull/5983/files#diff-a123072afb5c9463f24686503b47adec

bjornharrtell:
Seems it do not work after removing them, I've now changed this PR to use the new method and will report back findings.

rouault:
@szekerest Does that look good to you ?
@bjornharrtell I presume you could remove the "--version 3.0.12 --force" in appveyor.yml so that this is tested against recent swig

szekerest:
@bjornharrtell Are these changes backwards compatible with the older SWIG versions?

bjornharrtell:
@szekerest not sure, will have to try when I can get this to work - still ironing out some issues.

gdal/swig/include/csharp/typemaps_csharp.i
```diff
@@ -33,6 +33,45 @@
 
 /* CSHARP TYPEMAPS */
 
+%typemap(csdispose) SWIGTYPE %{
```
rouault, line 36: shouldn't those be only defined for SWIG < 4 (at least for clarity) ?

bjornharrtell: No, I think they are needed for SWIG >= 4. It's providing some kind of backward compat. It should be possible to do changes so that it should not be needed but I don't understand SWIG good enough yet.

rouault: Nobody understands SWIG, the simple fact of daring to touch SWIG related stuff makes you a fully qualified SWIG developer, so you're on the right track ;-)

I guess I'm confused because https://github.com/swig/swig/commit/06462acdf9b7738a19dce5579d90b1c91b3f9cfb#diff-8356d5285b5b342a204d4ff54fc34973 seems to introduce typemaps with the same names as the one you added, but slightly different definitions

bjornharrtell: Right.. :) I've got it fully working locally now at least. Perhaps, in time, I can reach some further conclusions on what is the most appropriate changes.

rouault:
@bjornharrtell Is this PR ready ?

bjornharrtell:
@rouault, no I still need to do research. And might not be my priority any longer because it turned out I had to build on an older system after all which has swig 3 by default.

Issue #2641, opened 2020-06-02
Can't convert BSQ starting with a ?magic number?


floppyzedolfin:
## Expected behavior and actual behavior.
Current behaviour:
> gdalinfo myFile.bsq
> ERROR 1: Invalid dataset dimensions : 0 x 0
> gdalinfo failed – unable to open ‘myFile.bsq’

Expected behaviour:
> gdalinfo myFile.bsq
> Driver: EHdr/ESRI .hdr Labelled 
> Files: myFile.bsq 
>       myFile.hdr 
> Size is 2048, 2048 
> Origin = (-0.500000000000000,0.500000000000000) 
> Pixel Size = (1.000000000000000,-1.000000000000000) 
> Corner Coordinates: 
> Upper Left ( -0.5000000, 0.5000000) 
> Lower Left ( -0.500, -2047.500) 
> Upper Right ( 2047.500, 0.500) 
> Lower Right ( 2047.500, -2047.500) 
> Center ( 1023.500, -1023.500) 
> Band 1 Block=2048x1 Type=UInt32, ColorInterp=Undefined

## Steps to reproduce the problem.
First, you need a .bsq and a .hdr file. Any file would be OK.
Then, hexedit the .bsq and write this as the first bytes : _14 00 00 00_ (overwrite existing data, in order not to change the size of the file)
(those are the first bits of my image)
Then, `gdalinfo myFile.bsq`

## Operating system
gdal 2.2.2 on CentOS-7 (reproduced on the gdal-ubuntu-light docker image)

## GDAL version and provenance
gdal versions 2.2.2 and 3.2.0 reproduce the issue. gdal version 1.11.4 doesn't.

## Investigation
We _think_ the issue comes from the fact that a driver accepts 14 00 as a magic number for its format, and that this causes gdal to try and use it instead of the EHdr one.
We tried and changed the first bits to _00 00 ..._, and this did solve our problem (this is NOT an acceptable solution, of course).

## Workaround
We declared a list of drivers not to use (all but the input and the output), in order to minimize the potentiality of this happening. However, 0% probability doesn't exist.

## Expected solution(s)
I'd like a -if parameter that specified the format of the input data.
Alternatively, formats with no magic number (and especially raw-buffers) should be tested before formats with magic numbers when trying to find the adequate driver.
[myFile.tar.zip](https://github.com/OSGeo/gdal/files/4716967/myFile.tar.zip)

jratike80:
[...]

PR #2642, opened 2020-06-02
gdalinfo, gdal_translate, gdalwarp: add a -if switch to be able to specify input driver(s)

rouault:
Refs #2641

### DIFF
--- a/gdal/doc/source/programs/gdal_translate.rst
+++ b/gdal/doc/source/programs/gdal_translate.rst
@@ -19,7 +19,8 @@ Synopsis
     gdal_translate [--help-general]
         [-ot {Byte/Int16/UInt16/UInt32/Int32/Float32/Float64/
                 CInt16/CInt32/CFloat32/CFloat64}] [-strict]
-        [-of format] [-b band]* [-mask band] [-expand {gray|rgb|rgba}]
+        [-if format]* [-of format]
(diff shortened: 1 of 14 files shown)


gdal/doc/source/programs/gdal_translate.rst
```diff
@@ -51,6 +52,15 @@ resampling, and rescaling pixels in the process.
     Don't be forgiving of mismatches and lost data when translating to the
     output format.
 
+.. option:: -if <FORMAT>
+
+    Format/driver name to be attempted to open the input file. This is generally
+    not necessary, but can be used if automatic driver detection fails to
```
floppyzedolfin, line 58: Regarding the bug: the automatic detection could be mistaken. `-if` should have precedence over driver checks.

rouault: > `-if` should have precedence over driver checks.

that's exactly what it does. Isn't that clear in the documentation ? If not, can you propose some wording ?

floppyzedolfin: can be used to skip the automatic driver detection

rouault: I've just reworded it in a slightly different way, hopefully clearer

PR #2656, opened 2020-06-06
Test driver MetadataItem XML elements

rbuffat:
First, sorry to use a PR for what in reality is probably more an issue. But I think the following test shows probably best what I mean. 

There are a few issues with the metadata of drivers:
- The MetadataItem's GDAL_DMD_OPENOPTIONLIST,  GDAL_DMD_CREATIONOPTIONLIST and DS_LAYER_CREATIONOPTIONLIST contains XML. (e.g. [GDAL_DMD_OPENOPTIONLIST: XML snippet with open options.](https://gdal.org/doxygen/gdal_8h.html#aa0a238d1db0bb21bb1690b1a3e391403)). However, drivers that do not specify  poDriver->SetMetadataItem() for these items return None. Some drivers include empty XML tags to avoid this (e.g. `poDriver->SetMetadataItem( GDAL_DS_LAYER_CREATIONOPTIONLIST, "<LayerCreationOptionList/>");`). 
- Some XML seems not to be valid. E.g. in the GML driver `description='Whether to write <gml:boundedBy> element...` should probabely be `description='Whether to write &lt;gml:boundedBy&gt; element...`

I would be willing to extend this PR to solve these issues, but I would need some guidance. 
- Should there be a mechanism to return e.g. `<LayerCreationOptionList/>` if metadata item is not set, should all drivers include this MetadataItem or is it acceptable to return None?
- Is there an XML schema the MetadataItem's can be validated against?
- What would be the best place for this test?


rouault:
> in the GML driver description='Whether to write <gml:boundedBy> element... should probabely be description='Whether to write &lt;gml:boundedBy&gt; element...

Indeed

>  Should there be a mechanism to return e.g. `<LayerCreationOptionList/>` if metadata item is not set, should all drivers include this MetadataItem or is it acceptable to return None?

Both None and `<LayerCreationOptionList/> are acceptable. <LayerCreationOptionList/> is there when a developer has reviewed the driver and ensured that the code does not use any layer creation option. When None is returned, it means "this driver may or may not have layer creation options". So a review of the code base is then needed to see if it has ones and if so, document them

>    Is there an XML schema the MetadataItem's can be validated against?

Unfortunately not

> What would be the best place for this test?

Probably more in autotest/gcore. The standard X/MIT header would be needed too

rbuffat:
pytest seems to use my systems gdal installation (Arch Linux) despite executing ". scripts/setdevenv.sh" and I did not bother yet to investigate why. I will check the CI if I caught everything.

> Both None and ` are acceptable.
Should this be documented somewhere, e.g. in gcore/gdal.h?

Is there a reason no schema exists? I could see the benefit that external programs would know what to expect when a schema would be available.

rouault:
> Is there a reason no schema exists?

not really, just nobody has done it yet.

rouault:
> pytest seems to use my systems gdal installation (Arch Linux) despite executing ". scripts/setdevenv.sh" and I did not bother yet to investigate why.

maybe check that the PYTHONPATH is correct. setdevenv.sh will use the `python` binary by default, and not python3

rbuffat:
> maybe check that the PYTHONPATH is correct. setdevenv.sh will use the `python` binary by default, and not python3

found the problem, I just forgot to compile with --with-python=yes

rouault:
>  just forgot to compile with --with-python=yes

ah indeed :-) You can also do it manually with "cd swig/python; python[3] setup.py build"

autotest/gcore/test_driver_metadata.py
```diff
@@ -0,0 +1,85 @@
+#!/usr/bin/env pytest
+###############################################################################
+# $Id$
+#
+# Project:  GDAL/OGR Test Suite
+# Purpose:  Test driver matadata xml
+# Author:   Rene Buffat <buffat at gmail dot com>
+#
+###############################################################################
+# Copyright (c) 2020, Even Rouault <even dot rouault at spatialys dot com>
```
rouault, line 10: should be yours

autotest/gcore/test_driver_metadata.py
```diff
@@ -0,0 +1,85 @@
...
+    """ Test if DMD_OPENOPTIONLIST metadataitem is present and can be parsed """
+
+    driver = gdal.GetDriverByName(driver_name)
+    openoptionlist_xml = driver.GetMetadataItem('DMD_OPENOPTIONLIST')
+
+    if openoptionlist_xml is not None and len(openoptionlist_xml) > 0:
+        assert "OpenOptionList" in openoptionlist_xml
+
+        # do not fail
+        print(openoptionlist_xml)
```
rouault, line 54: those 3 print() should be removed

rbuffat: Some XMLs are generated dynamically depending on e.g. installed libraries. For these, it would be nice to have the full failing xml at hand when the test fails.  Pytest unfortunately only outputs a truncated version.

rbuffat:
I included now an XML schema inline and it allowed me to find some more issues. 

Currently, only the default domain is used in the tests for GetMetadataItem. Is this ok for GDAL_DMD_OPENOPTIONLIST, GDAL_DMD_CREATIONOPTIONLIST, and DS_LAYER_CREATIONOPTIONLIST?

autotest/gcore/test_driver_metadata.py
rouault: then maybe do  something like
```
try:
       fromstring(openoptionlist_xml)
except:
       print(openoptionlist_xml)
       raise
```

rouault:
> I included now an XML schema inline and it allowed me to find some more issues.

Great!

> Currently, only the default domain is used in the tests for GetMetadataItem. Is this ok for GDAL_DMD_OPENOPTIONLIST, GDAL_DMD_CREATIONOPTIONLIST, and DS_LAYER_CREATIONOPTIONLIST?

Yes, they are required to be in the default domain.

rbuffat:
I noticed that most "name" attributes are upper case, but a few have lower case letters in it. Is there a convention for upper case and the lower cases are mistakes?

Also, some names have whitespace in it (e.g. CouchDB and Cloudant have 'GEOJSON ' as name, which looks like a mistake. https://github.com/OSGeo/gdal/blob/3554675bbce8dc00030bac33c99d92764d0f3844/gdal/ogr/ogrsf_frmts/couchdb/ogrcouchdbdriver.cpp#L122

rouault:
> but a few have lower case letters in it. Is there a convention for upper case and the lower cases are mistakes?

I'd should be upper case yes. Examples of lower case situations ? Ultimately this doesn't matter much as string comparison for option retrieval in drivers is case insensitive.



> Also, some names have whitespace in it (e.g. CouchDB and Cloudant have 'GEOJSON ' as name

yes this is a mistake. The code fetches GEOJSON without space

rbuffat:
> I'd should be upper case yes. Examples of lower case situations ? Ultimately this doesn't matter much as string comparison for option retrieval in drivers is case insensitive.

https://github.com/OSGeo/gdal/blob/3554675bbce8dc00030bac33c99d92764d0f3844/gdal/ogr/ogrsf_frmts/kml/ogrkmldriver.cpp#L147-L149

rouault:
> gdal/gdal/ogr/ogrsf_frmts/kml/ogrkmldriver.cpp

hum, I see. This is an inconsistency indeed. Putting them just uppercase would make them hard to read. Should rather have been NAME_FIELD, DESCRIPTION_FIELD,  with underscores... If we wanted to fix that, we could switch to that, add an alias to the old name, and make code changes to accept both variants. Not that I'm saying we should do that. I can live with the inconsistency

rbuffat:
> Not that I'm saying we should do that. I can live with the inconsistency

I too. It's just something I noticed and could be tested with the XML schema if it would be required.

Currently, I marked only the name and type attribute as required in the schema. Probably, it would also make sense to have the description attribute required. However, on my local installation 20 tests would fail:
         - gcore/test_driver_metadata.py:188 test_metadata_openoptionlist[BAG]
         - gcore/test_driver_metadata.py:207 test_metadata_creationoptionslist[GTiff]
         - gcore/test_driver_metadata.py:207 test_metadata_creationoptionslist[COG]
         - gcore/test_driver_metadata.py:207 test_metadata_creationoptionslist[NITF]
         - gcore/test_driver_metadata.py:207 test_metadata_creationoptionslist[MEM]
         - gcore/test_driver_metadata.py:207 test_metadata_creationoptionslist[GIF]
         - gcore/test_driver_metadata.py:207 test_metadata_creationoptionslist[netCDF]
         - gcore/test_driver_metadata.py:207 test_metadata_creationoptionslist[ISIS2]
         - gcore/test_driver_metadata.py:207 test_metadata_creationoptionslist[JP2OpenJPEG]
         - gcore/test_driver_metadata.py:207 test_metadata_creationoptionslist[JPEG2000]
         - gcore/test_driver_metadata.py:207 test_metadata_creationoptionslist[RMF]
         - gcore/test_driver_metadata.py:207 test_metadata_creationoptionslist[PDF]
         - gcore/test_driver_metadata.py:207 test_metadata_creationoptionslist[MRF]
         - gcore/test_driver_metadata.py:207 test_metadata_creationoptionslist[PAux]
         - gcore/test_driver_metadata.py:207 test_metadata_creationoptionslist[LCP]
         - gcore/test_driver_metadata.py:207 test_metadata_creationoptionslist[RRASTER]
         - gcore/test_driver_metadata.py:207 test_metadata_creationoptionslist[BAG]
         - gcore/test_driver_metadata.py:207 test_metadata_creationoptionslist[EXR]
         - gcore/test_driver_metadata.py:207 test_metadata_creationoptionslist[ENVI]
         - gcore/test_driver_metadata.py:207 test_metadata_creationoptionslist[ISCE]

autotest/gcore/test_driver_metadata.py
```diff
@@ -64,7 +64,7 @@
           </xs:restriction>
         </xs:simpleType>
       </xs:attribute>
-      <xs:attribute type="xs:string" name="type" use="optional"/>
+      <xs:attribute type="xs:string" name="type" use="required"/>
```
rouault, line 67: could we have some validation of the type with an enumeration ? The accepted ones are listed starting from https://github.com/OSGeo/gdal/blob/master/gdal/gcore/gdaldriver.cpp#L1841

rbuffat: I assume the type values are also case insensitive. I did not found a way to check case sensitive with XML schema. Thus I just include lowercase and uppercase variants.

rouault:
(there was an offending commit in master, now solves, that has caused failures, so jobs of the last build might fail because of that. I've restarted the ones I could)

rbuffat:
Is there something else that should be tested regarding the driver's metadata?

Something useful would be to test if the option names are understood by the drivers, but I currently don't see how this is easily possible.

rouault:
> Is there something else that should be tested regarding the driver's metadata?

One could potential check that the values of the GDAL_DMD_CREATIONFIELDDATATYPES are known.
Example:
poDriver->SetMetadataItem( GDAL_DMD_CREATIONFIELDDATATYPES, "Integer Integer64 Real String Date DateTime Time IntegerList Integer64List RealList StringList Binary" );

> Something useful would be to test if the option names are understood by the drivers

No, that's not generally possible. This is the job of the driver tests. (Well, I could imagine that we could try to grep the code of the driver, when sources are present, for the option names. But would need to exclude the match of the XML stuff itself. So probably more trouble than anything)

> Also is it correct that currently, no driver has the DMD_MULTIDIM_GROUP_CREATIONOPTIONLIST metadata item?

From my memories, yes. This is a provision

autotest/gcore/test_driver_metadata.py
```diff
@@ -491,3 +491,30 @@ def test_metadata_multidim_group_creationoptionslist(driver_name):
...
+
+@pytest.mark.parametrize('driver_name', all_driver_names)
+def test_metadata_creation_field_datatypes(driver_name):
+    """ Test if GDAL_DMD_CREATIONFIELDDATATYPES metadataitem returns valid datatypes """
+
+    valid_datatypes = {'Integer', 'Integer64', 'Real', 'String', 'Date', 'DateTime', 'Time', 'IntegerList',
+                       'Integer64List', 'RealList', 'StringList', 'Binary'}
+
+    driver = gdal.GetDriverByName(driver_name)
+    datatypes_str = driver.GetMetadataItem('GDAL_DMD_CREATIONFIELDDATATYPES')
```
rouault, line 504: the value of the string is 'DMD_CREATIONFIELDDATATYPES' (without GDAL_)

autotest/gcore/test_driver_metadata.py
```diff
@@ -495,21 +495,21 @@ def test_metadata_multidim_group_creationoptionslist(driver_name):
...
+    datatypes_str = driver.GetMetadataItem('DMD_CREATIONFIELDDATATYPES')
     if datatypes_str is not None:
         for datatype in datatypes_str.split(' '):
             assert datatype in valid_datatypes
 
 
 @pytest.mark.parametrize('driver_name', all_driver_names)
 def test_metadata_creation_sub_field_datatypes(driver_name):
-    """ Test if GDAL_DMD_CREATIONFIELDDATATYPES metadataitem returns valid datatypes """
+    """ Test if DMD_CREATIONFIELDDATATYPES metadataitem returns valid datatypes """
```
rouault, line 512:
```suggestion
    """ Test if DMD_CREATIONFIELDDATASUBTYPES metadataitem returns valid data subtypes """
```

rouault:
Is this still a WIP ?

rbuffat:
I was waiting for the CI to finish. If it succeeds I think it should be done. 

I noticed the compile time is quite long for the CI. Would be disabling compiler optimizations (O0) an option to decrease the CI runtime?

rouault:
>  If it succeeds I think it should be done.

OK

> I was waiting for the CI to finish

There's a duplicate "Travis CI - Pull Request " task that is in the limbo. To be ignored. I've restarted the github action Windows builds that had a flake.

> Would be disabling compiler optimizations (O0) an option to decrease the CI runtime?

We have build caching on Travis-CI which already helps a lot. But some jobs are indeed quite long, but not easy to fix. Could perhaps help on Windows indeed.

rouault:
Merged. Thanks ! Much appreciated

PR #2658, opened 2020-06-07
Set the application name in PostgreSQL connections

yorickdewid:
## What does this PR do?

If the application name is not set in the connection string then set the application name to the current GDAL version. PostgreSQL connections can take a long time before they are closed and they do show up in the database session list. Being able to identify database sessions makes diagnostics easier.

Ran valgrind, no mem leaks.

## Tasklist

 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/gdal/frmts/postgisraster/postgisrasterdataset.cpp
+++ b/gdal/frmts/postgisraster/postgisrasterdataset.cpp
@@ -2905,8 +2905,8 @@ GetConnectionInfo(const char * pszFilename,
      * Set application name if not found in connection string
      **********************************************************/
 
-    nPos = CSLFindName(papszParams, "application_name");
-    if (nPos == -1) {
+    if (CSLFindName(papszParams, "application_name") == -1 &&
+        getenv("PGAPPNAME") == nullptr) {
         osConnectionString += "application_name=";
         osConnectionString += "'";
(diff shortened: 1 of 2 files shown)


rouault:
I see there's a PGAPPNAME environment variable for that purpose. Will it override the default value you provide through the connection string ? If not, then we should test for PGAPPNAME before adding application_name=

yorickdewid:
Its a lipq trick, let me check

yorickdewid:
The ENVS are a backup in case the conn string does not provide the setting according to the postgres docs, and so this will in fact overwrite the ENV setting. See the same behaviour when I run with PGAPPNAME set. I'll check on the env as well.

yorickdewid:
But would the same not be true for `host`, `user`, `port`, `dbname` etc? `ogr/ogrsf_frmts/pg/ogrpgdatasource.cpp:369`

rouault:
> But would the same not be true for `host`, `user`, `port`, `dbname` etc? `ogr/ogrsf_frmts/pg/ogrpgdatasource.cpp:369`

They are appended only if passed as open options to the driver ( -oo HOST=xxxx )

PR #2662, opened 2020-06-10
Remove GDAL_SKIP=JP2ECW, see how CI reacts

VirtualTim:
I noticed that a lot of the CI tests are using `GDAL_SKIP JP2ECW`, so this is just a test to see what happens when this is removed.

Edit: I'm still not entirely sure what this controls, but it appears that this is no longer needed in most cases.


VirtualTim:
Thanks @rouault. I'm also looking into the `test_grib_grib2_write_data_encodings` failure when the last `GDAL_SKIP=JP2ECW` is removed, and will submit another PR when I know what's going on there.

Issue #2664, opened 2020-06-10
Visual Studio and shared lib: CPL_DLL should be __declspec(dllexport) only for build


SpaceIm:
## Expected behavior and actual behavior.

Currently, consuming gdal shared with Visual Studio leads to a lot of warning C4251 like `ogr_featurestyle.h(89,41): warning C4251: 'OGRStyleTable::osLastRequestedStyleName'ÿ: class 'CPLString' needs to have dll-interface to be used by clients of class 'OGRStyleTable'`.

## Steps to reproduce the problem.

- Build gdal as a shared lib with Visual Studio.
- Link gdal shared in an executable or an other lib.

## Operating system

Windows 10

## GDAL version and provenance

GDAL 3.1.0 from github, built from source

## Workarounds:

- define `CPL_DISABLE_DLL` ~~or `CPL_DLL=__declspec(dllimport)`~~ while consuming gdal shared (counterintuitive)
- patch gdal with something like this:

in gdal/port/cpl_port.h:

```patch
--- a/gdal/port/cpl_port.h
+++ b/gdal/port/cpl_port.h
@@ -344,7 +344,11 @@ typedef unsigned int  GUIntptr_t;
 
 #ifndef CPL_DLL
 #if defined(_MSC_VER) && !defined(CPL_DISABLE_DLL)
-#  define CPL_DLL     __declspec(dllexport)
+#  ifdef GDAL_EXPORTS
+#    define CPL_DLL     __declspec(dllexport)
+#  else
+#    define CPL_DLL     __declspec(dllimport)
+#  endif
 #  define CPL_INTERNAL
 #else
 #  if defined(USE_GCC_VISIBILITY_FLAG)

```

And define `GDAL_EXPORTS` when compiling gdal source files if shared in NMake files.

rouault:
Can you provide a pull request implementing that change ?

SpaceIm:
Sure, first I need to look at NMake files to see where `GDAL_EXPORTS` should be defined.

PR #2666, opened 2020-06-10
Fix warning C4251 if Visual Studio and shared for consumers of gdal lib

SpaceIm:
## What does this PR do?

For Visual Studio if GDAL is a shared library, this PR avoids defining `CPL_DLL` as `__declspec(dllexport)` when consumers use GDAL library (and therefore prevents from a lot of C4251 warnings).
Thereby, if GDAL is shared and compiled with Visual Studio, `CPL_DLL` is defined as :
- `__declspec(dllexport)` at compile time
- empty for consumers

Notes:
- `CPL_DLL` should be `__declspec(dllimport)` at consume time, but it also generates warnings due to stl classes exposed in public GDAL classes. `CPL_DLL` defined as empty is fine because GDAL doesn't have global variables in public headers.
- In this PR, GDAL executables are also built with `CPL_DLL` as `__declspec(dllexport)` (same behaviour than before) which is not very neat, but works (without warnings because C4251 is disabled in nmake fikes).

closes https://github.com/OSGeo/gdal/issues/2664

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS: Windows
* Compiler: Visual Studio

### DIFF
--- a/gdal/port/cpl_port.h
+++ b/gdal/port/cpl_port.h
@@ -344,7 +344,11 @@ typedef unsigned int  GUIntptr_t;
 
 #ifndef CPL_DLL
 #if defined(_MSC_VER) && !defined(CPL_DISABLE_DLL)
-#  define CPL_DLL     __declspec(dllexport)
+#  ifdef GDAL_COMPILATION
+#    define CPL_DLL __declspec(dllexport)
+#  else
+#    define CPL_DLL
+#  endif
 #  define CPL_INTERNAL
 #else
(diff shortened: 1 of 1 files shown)


SpaceIm:
Yes already checked with Visual Studio 2015/2017/2019, shared/static, Release/Debug,  MDd/MD/MTd/MT runtimes.

PR #2695, opened 2020-06-17
Add post-install tests to build apps using gdal.h with pkg-config

mwtoews:
This PR adds some post-install tests that check to build C and C++ apps that link to an installed version of libgdal.

pkg-config is also tested, to ensure the correct include and library paths are provided.

This test suite currently only works for *nix.

A running example looks like:
```
$ ./gdal/autotest/postinstall/test_pkg-config.sh /tmp/gdal/prefix
Running post-install tests with pkg-config
Testing C app
rm -f test_c test_c.o
cc -g -Wall -Werror -I/tmp/gdal/prefix/include   -c -o test_c.o test_c.c
cc -o test_c test_c.o -L/tmp/gdal/prefix/lib -lgdal
Testing expected ldd output ... passed
Testing expected version ... passed
rm -f test_c test_c.o
Testing C++ app
rm -f test_cpp test_cpp.o
g++ -std=c++11 -g -Wall -Werror -I/tmp/gdal/prefix/include   -c -o test_cpp.o test_cpp.cpp
g++ -o test_cpp test_cpp.o -L/tmp/gdal/prefix/lib -lgdal
Testing expected ldd output ... passed
Testing expected version ... passed
rm -f test_cpp test_cpp.o
0 tests failed out of 4
```

### DIFF
--- a/gdal/ci/travis/osx/install.sh
+++ b/gdal/ci/travis/osx/install.sh
@@ -63,3 +63,6 @@ cd ../../gdal
 ln -s /tmp/install/lib/libinternalproj.15.dylib /tmp/install/lib/libproj.15.dylib
 
 ccache -s
+
+# Post-install testing
+../autotest/postinstall/test_pkg-config.sh $HOME/install-gdal
--- a/autotest/postinstall/README.md
+++ b/autotest/postinstall/README.md
@@ -1,0 +1,9 @@
+## Post-install tests
+
+These tests are performed with an *installed* GDAL library. The script checks that pkg-config on *nix platforms can discover the correct compilation and linking options to build C and C++ applications.
+
+To run these tests, use the test script with the install prefix as the first argument, for example if GDAL was configured with `--prefix=/tmp/gdal`:
+```bash
+./autotest/postinstall/test_pkg-config.sh /tmp/gdal
+```
+the prefix is used to set both `PKG_CONFIG_PATH` and `LD_LIBRARY_PATH` (or `DYLD_LIBRARY_PATH`) environment variables during testing.
--- a/autotest/postinstall/test_c/Makefile
+++ b/autotest/postinstall/test_c/Makefile
@@ -1,0 +1,15 @@
+PROGRAM = test_c
+OBJECTS = test_c.o
+
+override CFLAGS += -g -Wall -Werror $(shell pkg-config gdal --cflags)
+override LDFLAGS += $(shell pkg-config gdal --libs)
+
+all: $(PROGRAM)
+
+$(PROGRAM): $(OBJECTS)
+	$(CC) -o $@ $< $(LDFLAGS)
+
+clean:
+	$(RM) $(PROGRAM) $(OBJECTS)
+
+.PHONY: clean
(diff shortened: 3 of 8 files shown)


mwtoews:
This PR is in draft form at the moment, as I'm not sure which CI test this should appear. It is quick to run and requires an installed library to function correctly.

rouault:
> as I'm not sure which CI test this should appear

Yes always difficult to decide. I'd perhaps suggest towards the end of ci/travis/ubuntu_1804/install.sh

mwtoews:
> I'd perhaps suggest towards the end of ci/travis/ubuntu_1804/install.sh

Done, with [passed tests](https://travis-ci.com/github/OSGeo/gdal/jobs/350327861#L23517) (search for "post-install").

Anything else? E.g, are there key public headers that should be checked to ensure they are installed? (Perhaps just those documented from https://gdal.org/api/?). I was tempted to add a section to autotest/README.md, but I'm not sure if this content would fit in, unless the `postinstall` directory should be elsewhere.

rouault:
>  are there key public headers that should be checked to ensure they are installed? (Perhaps just those documented from https://gdal.org/api/?)

yes, that could be a good idea.
A bit related, but not testing the installation part, we have autotest/cpp/test_c_include_from_cpp_file.cpp and the 3 test_include_from_XXXXX targets of autotest/cpp/Makefile that check that the headers can be included properly in various situations (from a C program, from a C++ one, from a C++ one under extern "C" { }). So with your addition we have pretty good coverage now of external use of GDAL headers. No need to replicate those tests in post-install I'd say. or we'd need to move the existing ones of autotest/cpp into post-install to avoid duplication. One slight benefit I could see from having them in post-install is that we could check issues of dependencies, like if public header A.h depends on B.h that is not installed, which couldn't be detected currently. But checking inclusion of the files from https://gdal.org/api/ should be enough

> I was tempted to add a section to autotest/README.md

I'd say perhaps a autotest/postinstall/README.md to describe what is useful for and that it is triggered by ci/travis/ubuntu_1804/install.sh

mwtoews:
Thanks for pointing out existing tests in autotest/cpp, which do a good job of checking these includes in various contexts. Another difference with post-install tests is that it includes using angle brackets (i.e. [`#include <filename>`](https://stackoverflow.com/q/21593)).

I've picked a few of the documented public headers, but not as comprehensive as autotest/cpp. I'm not sure what header dependencies could be tested, but that's fine to test out here at a later stage.

A new README.md is added, but doesn't say where it's tested on the current CI, as this seems to always change. I've also added this test to macOS, since it uses pkg-config too.

Some other suggestions for future work:

- Investigate if/how `pkg-config --libs --static gdal` might work and integrate with these tests
- Translate some of the building / scripting to Windows

gdal/ci/travis/osx/install.sh
```diff
@@ -63,3 +63,6 @@ cd ../../gdal
 ln -s /tmp/install/lib/libinternalproj.15.dylib /tmp/install/lib/libproj.15.dylib
 
 ccache -s
+
+# Post-install testing
+./autotest/postinstall/test_pkg-config.sh /tmp/install
```
rouault, line 68: due to line 63 (which could be changed) the current directory is the gdal subdirectory, not the root of the repository

PR #2697, opened 2020-06-17
docker: optionally package with debug symbols

rcoup:
Enable debug symbols in "ubuntu-full" docker images for Proj and GDAL non-release builds. This allows easier issue reproduction against master.

* debug symbols are in external files to minimise process memory
* symbols included for proj & gdal
* enable via `--with-debug-symbols/--without-debug-symbols` argument to build scripts
* included by default for non-release builds
* adds about 200MB to the image size

**Note:** Currently only applies to ubuntu-full, if there's interest could expand it to other image flavours.

I also simplified the ubuntu-full Dockerfile by moving proj & gdal builds to scripts, and added some more command safety: mostly driven by shellcheck and changing `;` → `&&` for consecutive commands.

Todo:

- [ ] decide whether to generalise the shell build scripts for the other image flavours.
- [ ] decide whether debug symbols would be useful for any other image flavours
- [ ] test building on not-my-machine

### DIFF
--- a/gdal/docker/ubuntu-full/Dockerfile
+++ b/gdal/docker/ubuntu-full/Dockerfile
@@ -5,7 +5,6 @@
 # Public domain
 # or licensed under X/MIT (LICENSE.TXT) Copyright 2019 Even Rouault <even.rouault@spatialys.com>
 
-ARG PROJ_INSTALL_PREFIX=/usr/local
 ARG BASE_IMAGE=ubuntu:20.04
 
 FROM $BASE_IMAGE as builder
@@ -138,7 +137,7 @@ ARG WITH_DEBUG_SYMBOLS=no
 
 # Build PROJ
 ARG PROJ_VERSION=master
-ARG PROJ_INSTALL_PREFIX
+ARG PROJ_INSTALL_PREFIX=/usr/local
 COPY ./bh-proj.sh /buildscripts/bh-proj.sh
 RUN /buildscripts/bh-proj.sh
 
@@ -155,15 +154,12 @@ FROM $BASE_IMAGE as runner
 
 RUN date
 
-# PROJ dependencies
 RUN apt-get update \
+# PROJ dependencies
     && DEBIAN_FRONTEND=noninteractive apt-get install -y \
         libsqlite3-0 libtiff5 libcurl4 \
         wget curl unzip ca-certificates \
-    && rm -rf /var/lib/apt/lists/*
-
 # GDAL dependencies
-RUN apt-get update \
     && DEBIAN_FRONTEND=noninteractive apt-get install -y \
         libcharls2 libopenjp2-7 libcairo2 python3-numpy \
         libpng16-16 libjpeg-turbo8 libgif7 liblzma5 libgeos-3.8.0 libgeos-c1v5 \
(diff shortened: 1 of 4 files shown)


rouault:
Thanks for that!

From what I can see for GDAL builds, the effect of setting CFLAGS=-ggdb disables -O2, which is probably not what we want since the binaries will be sluggish. By default, CFLAGS for GDAL are already "-g -O2". So there is probably no need to override them for GDAL, or use "-ggdb -O2".
By the way I've never used -ggdb. Not sure what it brings compared to plain -g.

rouault:
Got an error
```
Downloaded: 335 files, 498M in 7m 10s (1.16 MB/s)
The command '/bin/sh -c mkdir -p ${PROJ_INSTALL_PREFIX}/share/proj     && wget --no-verbose --mirror https://cdn.proj.org/     && rm -f cdn.proj.org/*.js     && rm -f cdn.proj.org/*.css     && mv cdn.proj.org/* ${PROJ_INSTALL_PREFIX}/share/proj/     && rmdir cdn.proj.org' returned a non-zero code: 8
```

rouault:
> Got an error
> 
> ```
> 
> ```

This is likely due to some failed attempts at fetching some unneeded files:
```
https://cdn.proj.org/robots.txt:
2020-06-17 15:55:15 ERROR 404: Not Found.
2020-06-17 15:55:20 URL:https://cdn.proj.org/geotiff.bundle.min.js [223236/223236] -> "cdn.proj.org/geotiff.bundle.min.js" [1]
https://cdn.proj.org/'%20+%20props.name%20+%20':
2020-06-17 15:55:21 ERROR 404: Not Found.
```
I think I hit that issue and decided to ignore the return code for that reason.
Perhaps a better way would be to use the projsync utility that comes with PROJ now with something like "projsync --system-directory -all"

rcoup:
> Perhaps a better way would be to use the projsync utility that comes with PROJ now with something like "projsync --system-directory -all"

Done

> From what I can see for GDAL builds, the effect of setting CFLAGS=-ggdb disables -O2, which is probably not what we want since the binaries will be sluggish. By default, CFLAGS for GDAL are already "-g -O2". So there is probably no need to override them for GDAL, or use "-ggdb -O2".

Nice catch, I'd been fiddling with Proj where that wasn't the case. Fixed

> By the way I've never used -ggdb. Not sure what it brings compared to plain -g.

Probably nothing for Linux, I swapped to `-g`.

I suspect Proj installing with `PREFIX=/` was an oversight, so I fixed that.

rouault:
By the way, what does bh- means in bh-gdal.sh ?

rcoup:
ah, build helper. It was `build-gdal.sh`, then I realised that having `build.sh` and `build-gdal.sh` and `build-proj.sh` beside each other wasn't helpful to figure out which to run 😄

rcoup:
As in the Todo, if you think it's useful I could try and spend some time normalising the other builds to all use a single set of helper scripts.

rouault:
> spend some time normalising the other builds to all use a single set of helper scripts.

I don't know how much it is feasible. There might be subtle differences between the builds

rcoup:
Sure. If someone's enthusiastic I guess they can pick up the todo's 😄

PR #2698, opened 2020-06-17
Fix an error string in gdal_rasterize

agiudiceandrea:
## What does this PR do?

When a "-ts 0 0" parameter is used with gdal_rasterize, then the following error is displayed:
"ERROR : Wrong value for -outsize parameter."

Since the "-outsize" parameter is not documented, the user cannot understand which parameter is wrong.

This PR changes the error string to a meaningful one: "ERROR : Wrong value for -ts parameter."


rouault:
> When a "-ts 0 0" parameter is used with gdal_translate,

With gdal_rasterize you meant ?

agiudiceandrea:
Yes, you are right... it was an oversight...
Should I close this PR and open a correct one or will you make a commit directly?

rouault:
> Should I close this PR and open a correct one or will you make a commit directly?

No, that's fine. I'll amend the commit message

Issue #2117, opened 2019-12-18
Refreshing GDAL's logo


nirvn:
@rouault , I remember a tweet from you mentioning the wish/will to refresh GDAL's logo (while PROJ was looking for a logo of its own). Along those lines, I've had a new logo in mind, which - in my humble opinion - would be a nice improvement to the current one.

Here's how it looks:
![Screenshot from 2019-12-18 11-22-29](https://user-images.githubusercontent.com/1728657/71055718-40d29480-2189-11ea-8515-20812330da4f.png)

The whole logo is simplified quite a lot and - as with QGIS and PROJ - makes use of its spelling to add a GIS hint (a compass here).

It would sit very well alongside other osgeo projects, and is more of a circa 2019 styled logo. 

Bonus: just like some folks are spelling PRØJ with that special Ø sign, GDAL could be spelled  GDΛL :wink:

nirvn:
Here's how the website would look like with the above-proposed logo:
![Screenshot from 2019-12-18 11-31-35](https://user-images.githubusercontent.com/1728657/71055922-03bad200-218a-11ea-9c20-9efdb27234fd.png)

nirvn:
And here are a couple of reduced "A" icon "real-world" scenarios:
![image881](https://user-images.githubusercontent.com/1728657/71056616-847acd80-218c-11ea-951b-ef8fbf3006e5.png)

![image924](https://user-images.githubusercontent.com/1728657/71056619-880e5480-218c-11ea-9684-29616b480ad8.png)

IMHO much better looking than the old logo as this proposal allows for a much simpler 'A' representation that is still clear when scaled down to 16 pixels.

nirvn:
The GDAL's A icon also looks great on mugs:
![Uploading Screenshot from 2019-12-18 16-20-13.png…]()

jef-n:
> The GDAL's A icon also looks great on mugs:
> ![Screenshot from 2019-12-18 16-20-13](https://user-images.githubusercontent.com/1728657/71073184-7475e480-21b2-11ea-800b-42f54b04574b.png)
> 
> ![Screenshot from 2019-12-18 16-25-06](https://user-images.githubusercontent.com/1728657/71073445-fe25b200-21b2-11ea-85df-cd4511b6ce4e.png)

You are really fishing for Λttention ;)

rouault:
> I remember a tweet from you mentioning the wish/will

[...]

PR #2699, opened 2020-06-18
add favicon.png in html

jjimenezshaw:
To find easier GDAL in my browser tabs, a favicon would be nice.

I took the image from gdalicon.png. I had to move the satellite a bit downwards (and cut partially one panel), to crop the image into 96x96 pixels.

If we change GDAL logo at any time (#2117), it would be nice to have a square version of it for the favicon.

This is how it looks next to the current page in Chromium (Ubuntu 18.04)

gdalhead

Tested also in Firefox-Ubuntu, Chrome-Android and Firefox-Android.

## What does this PR do?
Adds a favicon to the html page


rouault:
Cool. Thanks

Issue #2676, opened 2020-06-12
VS 2019 compilation


chacha21:
Hello,

The scripts to compile GDAL are up-to-date for Visual Studio 2017, they could be adapted to support VS 2019 as well.

In generate_vcxproj.bat, one could add the if "%_vcver_%"=="16.0"  case to match the value 16 for the <Visual C++ version> script argument

In nmake.opt, MSVC_VER should be tested against 1910 and 1920 to adapt the PLATFORM_TOOLSET_DIR to 141 or 142

rouault:
Feel free to submit a pull request implementing this

PR #2706, opened 2020-06-19
Add vs2019 compatibility

chacha21:
Add vs2019 compatibility
vcver is 16.0, for v142 toolset
added a VSFRIENDLYVERSION to nmake.opt to help using dependencies built with different versions of VS

address #2676

### DIFF
--- a/gdal/generate_vcxproj.bat
+++ b/gdal/generate_vcxproj.bat
@@ -52,7 +52,10 @@ set _vcver_=%1
 set _clver_=1600
 set _vstoolset_=v100
 
-if "%_vcver_%"=="15.0" (
+if "%_vcver_%"=="16.0" (
+	set _clver_=1920
+	set _vstoolset_=v142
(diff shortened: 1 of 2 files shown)


rouault:
Could you add in nmake.opt a comment line about VSFRIENDLYVERSION to explain that it is not used by GDAL itself ? but then I'm not completely sure how do you use it exactly. I suspect you edit nmake.opt itself to have path that depend on VSFRIENDLYVERSION ? However the recommended way if not modifying nmake.opt but editing nmake.local.

chacha21:
I have added the comment.
You are right, I usually tune the nmake.opt, I just did not know the nmake.local trick.

rouault:
Upon reflection, I'd prefer if you drop the changes in nmake.opt. I can see the point of the VSFRIENDLYVERSION idea but the fact that it can't work in nmake.local is a bit unsatisfactory

chacha21:
Sounds good to me.

Issue #2707, opened 2020-06-19
LVBAG: Multiple layers with same and different features causes SIGSEGV


yorickdewid:
## Expected behavior and actual behavior.

If multiple layers with the same and different features are mixed unionizing the layers will result in SIGSEGV. The coalesce method keeps track of the source layers per offset. The offset is invalidated the moment a single list item is removed `papoLayers.erase(it)` from the list.

## Steps to reproduce the problem.

```
DIR = \
  layer_1_feat_1.xml
  layer_1_feat_2.xml
  layer_2_feat_1.xml
  layer_2_feat_2.xml
ogrinfo
```

## GDAL version and provenance

3.2-master

Issue #2636, opened 2020-06-01
LVBAG: Empty layers are not merged into non-empty layers


yorickdewid:
## Expected behavior and actual behavior.

LV BAG driver returns two layers, one of which is empty and another which contains all the data. Empty layers should be removed iff there are equivalent layers with features. Empty layers do not have a SRS set and therefore are not merged into the union.

## Steps to reproduce the problem.

DIR = pnd_without_features.xml pnd.xml
ogrinfo <DIR>

## GDAL version and provenance

3.2-master

PR #2713, opened 2020-06-21
LVBAG; bugfixes and improvements

yorickdewid:
## What does this PR do?

Many bugfixes and improvements for the LVBAG driver
- FIX: Remove unionized source layers from the dataset _after_ unioning is done
- Rewrite code to use the internal C++11 memory functions
- Make nMaxSimultaneouslyOpened configurable for testing
- FIX: Parse year as Date type
- Add SRS/Geom type based on dataset name and LVBAG 2.0 kadaster specification
- Fit invalid data if possible
- Set invalid dates to null

Fixes: https://github.com/OSGeo/gdal/issues/2707
Fixes: https://github.com/OSGeo/gdal/issues/2636

## Tasklist

 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/gdal/ogr/ogrsf_frmts/lvbag/ogrlvbagdriver.cpp
+++ b/gdal/ogr/ogrsf_frmts/lvbag/ogrlvbagdriver.cpp
@@ -76,7 +76,7 @@ GDALDataset *OGRLVBAGDriverOpen( GDALOpenInfo* poOpenInfo )
 
     if( !poOpenInfo->bIsDirectory && poOpenInfo->fpL != nullptr )
     {
-        if( !poDS->Open( poOpenInfo->pszFilename ) )
+        if( !poDS->Open( poOpenInfo->pszFilename, poOpenInfo->papszOpenOptions ) )
             poDS.reset();
     }
     else if( poOpenInfo->bIsDirectory && poOpenInfo->fpL == nullptr )
@@ -98,7 +98,7 @@ GDALDataset *OGRLVBAGDriverOpen( GDALOpenInfo* poOpenInfo )
             if( !OGRLVBAGDriverIdentify(&oOpenInfo) )
                 continue;
 
-            if( !poDS->Open( oSubFilename ) )
+            if( !poDS->Open( oSubFilename, poOpenInfo->papszOpenOptions ) )
                 continue;
         }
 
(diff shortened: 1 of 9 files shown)


gdal/ogr/ogrsf_frmts/lvbag/ogrlvbagdriver.cpp
```diff
@@ -136,6 +136,11 @@ void RegisterOGRLVBAG()
     poDriver->SetMetadataItem( GDAL_DMD_HELPTOPIC, "drivers/vector/lvbag.html" );
     poDriver->SetMetadataItem( GDAL_DCAP_VIRTUALIO, "YES" );
 
+    poDriver->SetMetadataItem( GDAL_DMD_OPENOPTIONLIST,
+"<OpenOptionList>"
+"  <Option name='AUTOCORRECT_INVALID_DATA' type='boolean' description='whether driver should try to fix invalid data' default='YES'/>"
```
rouault, line 141: should be documented in the .rst file

gdal/ogr/ogrsf_frmts/lvbag/ogrlvbaglayer.cpp
```diff
@@ -505,17 +535,70 @@ void OGRLVBAGLayer::EndElementCbk( const char *pszName )
                 reinterpret_cast<OGRGeometry *>(OGR_G_CreateFromGML(osElementString.c_str())) };
             if( poGeom && !poGeom->IsEmpty() )
             {
-                if( !bSchemaOnly )
+                // The specification only accounts for 2-dimensional datasets
+                if( poGeom->Is3D() )
+                    poGeom->flattenTo2D();
+
+#if defined(HAVE_SFCGAL) || defined(HAVE_GEOS)
```
rouault, line 542: MakeValid() is only implemented for GEOS >= 3.8, so testing for HAVE_SFCGAL will not bring anything

gdal/ogr/ogrsf_frmts/lvbag/ogrlvbagdriver.cpp
yorickdewid: Was planning on it

gdal/ogr/ogrsf_frmts/lvbag/ogrlvbaglayer.cpp
yorickdewid: I'll check GEOS version. Also this test fails on Windows and I don't know why. Is it because of this version?

rouault: > Also this test fails on Windows and I don't know why. Is it because of this version?

Possibly. I can see "ERROR 6: GEOS 3.8 or later needed for MakeValid.". Actually none of the CI target must have GEOS 3.8 yet

yorickdewid: I've listed `$(GEOS_CFLAGS)` in the makefile but `GEOS_VERSION_MAJOR` never gets passed down. Are these defines only set in later versions?

rouault: you need to include ogr_geos.h

yorickdewid: Skipped the test for now

yorickdewid:
I'll add one other feature before merge.

Can GDAL build relationships between datasets? Such that foreign key constraints are created when supported?

rouault:
> Can GDAL build relationships between datasets? Such that foreign key constraints are created when supported?

No, there's nothing in OGR abstraction regarding relationships. The GMLAS driver exposed a specific layer for that purpose ( https://gdal.org/drivers/vector/gmlas_metadata_layers.html , which was ultimately used by a dedicated QGIS plugin ) but that's not something handled in other parts of the code base.

yorickdewid:
Done if build succeeds

PR #2729, opened 2020-06-30
Add support for field aliases to OGRFieldDefn, read alias in openfilegdb driver

nyalldawson:
This PR adds support for setting/retrieving field aliases to OGRFieldDefn. It also adds support for reading the embedded field alias information stored in .gdb tables (with the openfilegdb driver at least -- would likely be possible with the filegdb driver too, but i distinctly lack motivation to work on that one!).

It's possible other formats also have native alias support which would benefit from this, but I'm not personally aware of any...!


autotest/ogr/ogr_fgdb.py
```diff
@@ -2129,3 +2129,27 @@ def test_ogr_fgdb_utc_datetime():
     f = lyr.GetNextFeature()
     # Check that the timezone +00 is present
     assert f.GetFieldAsString('EditDate') == '2020/06/22 07:49:36+00'
+
+
+def test_ogr_fgdb_alias():
```
rouault, line 2134: this test should be moved to ogr_openfilegdb.py.   ogr_fgdb.py is only for the SDK based driver. And this will save you the funny play with driver deregisteration :-)

gdal/ogr/ogr_feature.h
```diff
@@ -79,6 +79,7 @@ class OGRStyleTable;
  * Definition of an attribute of an OGRFeatureDefn. A field is described by :
  * <ul>
  * <li>a name. See SetName() / GetNameRef()</li>
+ * <li>an alias. See SetAlias() / GetAliasRef()</li>
```
rouault, line 82: I was wondering if we shouldn't generalize the name a bit. From https://desktop.arcgis.com/en/arcmap/latest/manage-data/tables/field-properties-aliases-and-table-display-options.htm , "An alias is an alternative name for a field—a more user-friendly description of the content of the field".  So I was thinking to "Description" which could potentially cover more drivers. The issue I see with "alias" is that one could expect it to be usable as a replacement for the name in SQL expressions for example.
But I don't have a strong opinion about that. Just wanted to raise the discussion

KML has a "displayName" that could potentially fit into that : https://developers.google.com/kml/documentation/extendeddata#declare-the-schema-element (although KML displayName can includes HTML tags)

gdal/ogr/ogrfielddefn.cpp
```diff
@@ -224,6 +227,95 @@ const char *OGR_Fld_GetNameRef( OGRFieldDefnH hDefn )
...
+}
+
+/************************************************************************/
+/*                             GetAliasRef()                             */
+/************************************************************************/
+
+/**
+ * \fn const char *OGRFieldDefn::GetAliasRef();
+ *
+ * \brief Fetch alias for this field.
```
rouault, line 282: We should probably add more about the semantics of what "alias". See above point.

gdal/swig/include/ogr.i
```diff
@@ -2204,6 +2204,18 @@ public:
     OGR_Fld_SetName(self, name);
   }
 
+  const char * GetAlias() {
+    return (const char *) OGR_Fld_GetAliasRef(self);
```
rouault, line 2208:
```suggestion
    return OGR_Fld_GetAliasRef(self);
```

gdal/swig/include/ogr.i
```diff
@@ -2204,6 +2204,18 @@ public:
...
+  }
+
+  const char * GetAliasRef() {
+    return (const char *) OGR_Fld_GetAliasRef(self);
```
rouault, line 2212:
```suggestion
    return OGR_Fld_GetAliasRef(self);
```

gdal/ogr/ogr_feature.h
nyalldawson: Hmm - my concern with equating this to description is that a description is likely to be lengthy, while an alias is somewhat in between the field name and a description. E.g. if we use description, then you could argue that field comments from postgres should be supported. But I've only ever seen them used for lengthy descriptions which would be unsuitable for an "alias". What about "AlternativeName"?

nyalldawson: Or "Extended/expanded" name?

rouault: ExtendedName would be fine to me. In the comments you could add what you mentioned above, that is that is expected to be a longer form of Name but not a too lengthy description / comment. And that it for example maps to FileGeodabase alias.
Is the purpose of using them in the header line of QGIS attribute table or edit forms ?

jratike80:
I would have voted for "AlternativeName". I have also use cases in my mind when I probably would not rename fields by expanding but rather by using more compact names (INSPIRE schemas, database tables named after company wide naming conventions etc).

gdal/ogr/ogr_feature.h
rouault: > I would have voted for "AlternativeName".

I'd be fine with AlternativeName, but just seeing that https://en.wikipedia.org/wiki/Alias defines  "Alias, an alternative name" :-) So perhaps alias is just fine.

nyalldawson: > Is the purpose of using them in the header line of QGIS attribute table or edit forms ?

Yes -- it's so that we can automatically set the field alias in qgis to match those defined in the data source. The ultimate outcome is that official datasets published with cryptic field names should be a bit more usable!

> I'd be fine with AlternativeName

Let's go with that -- it avoids the confusing wrt SQL aliases.

nyalldawson:
All comments addressed

PR #2735, opened 2020-07-01
configure: HDF4 requires built-in pds format

tbonfort:
Allow configure to be run with --with-hdf4 and --disable-all-optional-drivers

PR #2740, opened 2020-07-02
configure: HDF4 requires built-in pds format

rouault:
Continuation of https://github.com/OSGeo/gdal/pull/2735


tbonfort:
Thank you for taking this on @rouault

PR #2752, opened 2020-07-08
Add builtin TGA image file format read-only driver

rouault:


nyalldawson:
@rouault I'm really curious as to the background for this driver. What spatial data is distributed in tga format?

rouault:
> What spatial data is distributed in tga format?

Probably null. The use case here was to be able to read textures provided with a 3D model through GDAL.

nyalldawson:
Ok, thanks! I was wondering if there was some ancient data distributed in this format which someone needed to read...

PR #2757, opened 2020-07-10
gdal2tiles: add -w mapml output, and possibilities to use custom tiling scheme / profile

rouault:
Add data/template_tiles.mapml, data/tms_MapML_APSTILE.json and data/tms_MapML_CBMTILE.json


rouault:
CC @dmorissette

rouault:
keep alive

PR #2769, opened 2020-07-15
Fix & improves issues related to VRT and SrcRect/DstRect

rouault:


sgillies:
@rouault are there any specific connected issues?

rouault:
> are there any specific connected issues?

something that was privately reported to me. Basically when using gdalbuildvrt -te  with bounds that are close to pixel boundaries, but not exactly matching it (which is hard due to floating-point), values like y.999xxx would be written in the .vrt file, and later interpreted as y instead of y+1, leading to unwanted pixel shifts, contrary to gdal_translate -of VRT -projwin were snapping to nearest integer is done.

PR #2771, opened 2020-07-16
OpenFileGDB: add support for reading .spx spatial index file

rouault:
Credits to [QGIS.org Grant program 2020](http://blog.qgis.org/2020/07/10/qgis-grant-programme-2020-results/) for sponsoring this work, and to @nyalldawson for his help in the reverse engineering.

Results on a version from of few years ago of the [nz-primary-parcels.gdb dataset](https://data.linz.govt.nz/layer/50772-nz-primary-parcels/), featuring 2 484 646 polygons, 773 MB:
"ogrinfo  -spat 174.949909 -41.143842 175.684529 -40.594600 -al -so -noextent" (thus counting the number of features intersecting the spatial filter), which returns 81 046 polygons, nows runs in 400 ms with this PR and the OpenFileGDB driver, vs 6.7 s before (full scan), vs 890 ms with the FileGDB driver (with FileGDB SDK 1.5). All figures given with debug builds, and on a hot run (so with most pages cached by the OS). Interactive display of this dataset in QGIS with the OpenFileGDB driver is as fluid as with the FileGDB one. Comparing behaviour of OpenFileGDB and FileGDB drivers with strace shows that they read a similar amount of data in the .spx file.


nyalldawson:
Looks good to me! Thanks @rouault!

nyalldawson:
(next step: reverse engineering that compressed data format 😱)

Issue #2773, opened 2020-07-17
abort in FlatGeobuf::PackedRTree::streamSearch


opoplawski:
## Expected behavior and actual behavior.
gdalautotest python test aborts:
```
ogr/ogr_flatgeobuf.py .
Thread 1 "python3" received signal SIGABRT, Aborted.
```
```
#5  FlatGeobuf::PackedRTree::streamSearch(unsigned long, unsigned short, FlatGeobuf::NodeItem const&, std::function<void (unsigned char*, unsigned long, unsigned long)> const&) (numItems=numItems@entry=10, nodeSize=nodeSize@entry=16, item=..., readNode=...) at packedrtree.cpp:349
349                 auto nodeItem = nodeItems[static_cast<size_t>(nodePos)];
(gdb) list
344             }
345     #endif
346             // search through child nodes
347             for (uint64_t pos = nodeIndex; pos < end; pos++) {
348                 uint64_t nodePos = pos - nodeIndex;
349                 auto nodeItem = nodeItems[static_cast<size_t>(nodePos)];
350                 if (!item.intersects(nodeItem))
351                     continue;
352                 if (isLeafNode)
353                     results.push_back({ nodeItem.offset, pos - leafNodesOffset });
(gdb) print pos
$1 = 0
(gdb) print nodeItems
$2 = std::vector of length 0, capacity 16
```
I believe the issue is while the capacity of the vector is 16, there is no data in it so referencing the items is a violation.

## Steps to reproduce the problem.
Compile with -Wp,-D_GLIBCXX_ASSERTIONS

## Operating system
Fedora Rawhide x86_64

## GDAL version and provenance
Fedora package - 3.1.2

rouault:
CC @bjornharrtell

bjornharrtell:
This is triggered by -D_GLIBCXX_ASSERTIONS. The size of the vector is intentionally unknown. I'm using a reserved capacity vector as a buffer to avoid having to manually allocate/deallocate it. But I realize I could probably use a stack allocated array here.

PR #2774, opened 2020-07-17
FlatGeobuf: Fix from upstream

bjornharrtell:
ref https://github.com/OSGeo/gdal/issues/2773

### DIFF
--- a/gdal/ogr/ogrsf_frmts/flatgeobuf/packedrtree.cpp
+++ b/gdal/ogr/ogrsf_frmts/flatgeobuf/packedrtree.cpp
@@ -316,8 +316,7 @@ std::vector<SearchResultItem> PackedRTree::streamSearch(
     auto levelBounds = generateLevelBounds(numItems, nodeSize);
     uint64_t leafNodesOffset = levelBounds.front().first;
     uint64_t numNodes = levelBounds.front().second;
-    std::vector<NodeItem> nodeItems;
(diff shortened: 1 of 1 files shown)


gdal/ogr/ogrsf_frmts/flatgeobuf/packedrtree.cpp
```diff
@@ -316,9 +316,8 @@ std::vector<SearchResultItem> PackedRTree::streamSearch(
     auto levelBounds = generateLevelBounds(numItems, nodeSize);
     uint64_t leafNodesOffset = levelBounds.front().first;
     uint64_t numNodes = levelBounds.front().second;
-    std::vector<NodeItem> nodeItems;
-    nodeItems.reserve(nodeSize);
-    uint8_t *nodesBuf = reinterpret_cast<uint8_t *>(nodeItems.data());
+    NodeItem nodeItems[nodeSize];
```
rouault, line 319: variable length arrays are forbidden in GDAL. I'd suggest to just allocate the std::vector to nodeSize items instead of just .reserve(). Hopefully this will not affect performance in a measurable way

bjornharrtell: Ah yes, I recall something about variable length arrays now.. reworked to your suggestion, I don't think it's a performance critical section.

PR #2780, opened 2020-07-19
Dataset.GetSpatialRef should be exposed to CSharp (#2620)

szekerest:


rouault:
it seems there's an issue at runtime with the trusty Travis target. I remember @bjornharrtell had a similar issue. One of the idea was to move maybe the C# tests to a more modern Ubuntu version

bjornharrtell:
Tried to move to newer target in PR https://github.com/OSGeo/gdal/pull/2649 but got stuck with a problem I haven't been able to explain.

PR #2783, opened 2020-07-21
Fix a small typo

oesteban-vx:
Just about a typo I saw when reading the docs. I'm almost ashamed of so little added value for the merge effort :-)


rouault:
> I'm almost ashamed of so little added value for the merge effort :-)

you shouldn't. This is much appreciated

PR #2785, opened 2020-07-21
Backporting release/3.1: Dataset.GetSpatialRef should be exposed to CSharp (#2620)

szekerest:


szekerest:
@rouault Do you have idea why that build was failing? The only message related to the failure is:
"The job exceeded the maximum log length, and has been terminated."

PR #2798, opened 2020-07-23
Fix CSV typo in gt_datum.csv

hobu:
CSV typo in gt_datum.csv that has existed for years. No one's really using this for anything other than historical, AFAIK.


hobu:
GDAL_DATA could really use a cleanup in general. It's a graveyard.

gdal/data/gt_datum.csv
```diff
@@ -198,7 +198,7 @@ SAN-J,"SOUTH AMERICAN 1969, Baltra",SA,-47,25,26,25,-42,25,-2,1,-92,-89
 SAN-K,"SOUTH AMERICAN 1969, Trinidad",SA,-45,25,12,25,-33,25,4,17,-68,-55
 SAN-L,"SOUTH AMERICAN 1969, Venezuela",SA,-45,3,8,6,-33,3,-5,18,-79,-54
 SCK,"SCHWARZECK, Namibia",BN,616,20,97,20,-251,20,-35,-11,5,31
-SGM,"SELVAGEM GRANDE 1938, Salvage Is,"I,N -28,9   2,5 -12,4   25,6,0   2,28,32,-18,-14
+SGM,"SELVAGEM GRANDE 1938, Salvage Is",I,N -28,9   2,5 -12,4   25,6,0   2,28,32,-18,-14
```
rouault, line 201: The rest of this line seems also badly broken. The I,N should be IN looking at the line just below. Not sure about the numerical values, but they are highly suspicious too

hobu: Yeah, I have no idea what the history is for this.

jgrocha: Definitely, the ELLIPSOID should be 'IN' (from International). The former parameters used to be real numbers, not integers. I suspect that the decimal separator was converted to comma or was discarded. There is a document with [parameters](https://www.dgterritorio.gov.pt/sites/default/files/ficheiros-geodesia/AZ_e_MAD_ParamTransformacao.pdf) for Madeira, from the national mapping agency, but not specifically for this island.

jgrocha: Maybe this is useful as a starting point to recover this Datum parameters:

Geodetic     Datum | Selvagem     Grande | urn:ogc:def:datum:EPSG::6616 |  
-- | -- | -- | --
Ellipsoid | International     1924 | urn:ogc:def:ellipsoid:EPSG::7022 |  
Prime     Meridian | Greenwich | urn:ogc:def:meridian:EPSG::8901 |  
Area     of Use | Portugal     - Selvagens onshore | urn:ogc:def:area:EPSG::2779

CRS using this:

| CRS |  |  |  
|-- | -- | -- | 
|Projected     CRS | Selvagem     Grande / UTM zone 28N | urn:ogc:def:crs:EPSG::2943 | 
|Geodetic     CRS (geographic 2D) | Selvagem     Grande | urn:ogc:def:crs:EPSG::4616 |

PR #2808, opened 2020-07-27
gdal2tiles: Fix layer generation with raster profile and nonnative zoom (fixes #2799)

### LINKED ISSUE #2799: gdal2tiles -p raster generates broken tiles if the max zoom level is not from the recommended range
Issue/Feature Request

gdal2tiles generates broken tiles if the max zoom level is not from the recommended range.
Also, I would like to have the ability to skip the generation of some zoom levels from a zoom level range.
For example:
Input image has size  7952x5304
Recommended zoom level range, in case of tiles size 256, is 0-5
I need only 0-4, I don't want even generate fifth level (to speed up the process)

Current state:
If I run the next command

`gdal2tiles.py -p raster --xyz --zoom=0-4 --tilesize=256 ...
`

this will generate "broken" tiles. Example:

I need to have the ability to generate correct tiles with the previous command or add a special key for skipping levels. Example (**--skip-zoom-level**):

`gdal2tiles.py -p raster --xyz --zoom=0-5 --skip-zoom-level=5 --tilesize=256 ...`

dminor:


dminor:
I did some more testing, and it looks like there are two other locations where nativezoom is used rather than tmaxz, and these cause problems if you attempt to use more zoom levels than the calculated native zoom.

dminor:
Thank you.

PR #2802, opened 2020-07-26
SWIG 4+ compatibility

bjornharrtell:
I revisited https://github.com/OSGeo/gdal/pull/2637 and have as far as I can tell verified this is the correct changes for SWIG 4.0 compatibility with tests enabled for Ubuntu 20.04.

PR #2813, opened 2020-07-28
Remove SWIG C# testing from trusty CI setup

bjornharrtell:
Trusty has an old an unsupported version of Mono which fx. does not work with unsafe code that is planned to be introduced with.

In https://github.com/OSGeo/gdal/pull/2802 testing with a more up to date Mono is introduced in the new CI for Ubuntu 20.04.

This PR removes the old test setup for trusty.

### DIFF
--- a/gdal/ci/travis/trusty_clang/before_install.sh
+++ b/gdal/ci/travis/trusty_clang/before_install.sh
@@ -32,8 +32,6 @@ sudo apt-get install -y --allow-unauthenticated ccache libpng12-dev libjpeg-dev 
 # libgta-dev
 sudo apt-get install -y libqhull-dev
 sudo apt-get install -y libogdi3.2-dev
-# MONO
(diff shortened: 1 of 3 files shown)


szekerest:
I think we’d better to switch to a netcore build, so that the testing step could remain. It is annoying that the mono build could break any of the existing configurations.

bjornharrtell:
Note that I've already made the tests work on Ubuntu 20.04 setup as part of https://github.com/OSGeo/gdal/pull/2802 but it still uses Mono. I think this PR still makes sense if https://github.com/OSGeo/gdal/pull/2802 can be merged, will you consider reviewing it @szekerest?

I agree we should switch to .NET Core. It will soon be the only alive .NET platform. But it will only be sensible to support that on newer Ubuntu versions and Windows. Ubuntu 14.04 LTS (trusty) has no support for .NET Core whatsoever. I would like to put effort into moving to .NET Core but independent of this PR and #2802.

PR #2821, opened 2020-08-01
ogrshapedriver: do not claim to support DateTime data type

dminor:
Shapefiles support Date, but not DateTime. This is leading to issues like https://github.com/qgis/QGIS/issues/38011 in QGIS.


dminor:
Thank you!

PR #2835, opened 2020-08-06
[NAS] also filter for wfs:member (as in GID7)

jef-n:


jef-n:
odd - the individual travis jobs all succeeded - although the all show as queued.

PR #2837, opened 2020-08-07
Quote DBQ value in PGEO driver template to avoid issues opening MDB paths with spaces

nyalldawson:


rouault:
Cherry-picked in 3.1 per 1f71651

PR #2838, opened 2020-08-07
[pgeo] Fixes to automatic ODBC driver installation

nyalldawson:
Together these fixes fix the automatic driver installation under linux. (Paired with https://github.com/evanmiller/mdbtools/commit/3517b0eee6383655c85a2c1b7cb3ac6df384cc4f they (kind of, mostly) fix use of the pgeo driver under the linux/unixodbc environment)

### DIFF
--- a/gdal/port/cpl_odbc.cpp
+++ b/gdal/port/cpl_odbc.cpp
@@ -116,6 +116,9 @@ int CPLODBCDriverInstaller::InstallDriver( const char* pszDriver,
                                         ODBC_FILENAME_MAX, nullptr, fRequest,
                                         &m_nUsageCount) )
         {
+            // if installing the driver fails, we need to roll back the changes to ODBCSYSINI environment
+            // variable or all subsequent use of ODBC calls will fail
+            unsetenv( "ODBCSYSINI" );
             CPL_UNUSED RETCODE cRet = SQLInstallerError( nErrorNum, &m_nErrorCode,
                             m_szError, SQL_MAX_MESSAGE_LENGTH, nullptr );
             (void)cRet;
--- a/gdal/port/cpl_odbc.cpp
+++ b/gdal/port/cpl_odbc.cpp
@@ -90,13 +90,17 @@ int CPLODBCDriverInstaller::InstallDriver( const char* pszDriver,
         // system-wide default location, so try to install to HOME.
 
         static char* pszEnvIni = nullptr;
+
+        // Read HOME location.
         const char* pszEnvHome = getenv("HOME");
+        CPLAssert( nullptr != pszEnvHome );
+        CPLDebug( "ODBC", "HOME=%s", pszEnvHome );
+
+        const char* pszEnvOdbcSysIni = nullptr;
         if( pszEnvIni == nullptr )
         {
-            // Read HOME location.
-
-            CPLAssert( nullptr != pszEnvHome );
-            CPLDebug( "ODBC", "HOME=%s", pszEnvHome );
+            // record previous value, so we can rollback on failure
+            pszEnvOdbcSysIni = getenv("ODBCSYSINI");
 
             // Set ODBCSYSINI variable pointing to HOME location.
             const size_t nLen = strlen(pszEnvHome) + 12;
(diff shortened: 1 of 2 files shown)


nyalldawson:
(For reference -- the missing bit is https://github.com/evanmiller/mdbtools/issues/6, which causes mdbtools to incorrectly read the binary shape data)

gdal/port/cpl_odbc.cpp
```diff
@@ -90,10 +90,10 @@ int CPLODBCDriverInstaller::InstallDriver( const char* pszDriver,
         // system-wide default location, so try to install to HOME.
 
         static char* pszEnvIni = nullptr;
+        const char* pszEnvHome = getenv("HOME");
         if( pszEnvIni == nullptr )
         {
             // Read HOME location.
-            char* pszEnvHome = getenv("HOME");
 
             CPLAssert( nullptr != pszEnvHome );
```
mloskot, line 98: May be a good idea to move the assertion together the moved `getenv`

gdal/port/cpl_odbc.cpp
```diff
@@ -112,10 +112,13 @@ int CPLODBCDriverInstaller::InstallDriver( const char* pszDriver,
...
 
         // Try to install ODBC driver in new location.
-        if( FALSE == SQLInstallDriverEx(pszDriver, nullptr, m_szPathOut,
+        if( FALSE == SQLInstallDriverEx(pszDriver, pszEnvHome, m_szPathOut,
                                         ODBC_FILENAME_MAX, nullptr, fRequest,
                                         &m_nUsageCount) )
         {
+            // if installing the driver fails, we need to roll back the changes to ODBCSYSINI environment
+            // variable or all subsequent use of ODBC calls will fail
+            unsetenv( "ODBCSYSINI" );
```
mloskot, line 121: I think MSVC won't swallow this.  Instead, [_putenv](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/putenv-wputenv?view=vs-2019):
> You can remove a variable from the environment by specifying an empty value_string, or in other words, by specifying only varname=

nyalldawson: Thanks. Thinking a bit more about this, I'm wondering if I should store any existing value for the variable before we initially set it, and then restore that value...

mloskot: That would make it more polite indeed

nyalldawson: Thanks, implemented

gdal/port/cpl_odbc.cpp
```diff
@@ -112,10 +116,32 @@ int CPLODBCDriverInstaller::InstallDriver( const char* pszDriver,
...
+                const size_t nLen = strlen( pszEnvOdbcSysIni ) + 12;
+                pszEnvRollback = static_cast<char *>(CPLMalloc(nLen));
+                snprintf( pszEnvRollback, nLen, "ODBCSYSINI=%s", pszEnvOdbcSysIni );
+            }
+            else
+            {
+                // ODBCSYSINI not previously set, so remove
+                const size_t nLen = 11;
+                pszEnvRollback = static_cast<char *>(CPLMalloc(nLen));
+                snprintf( pszEnvRollback, nLen, "ODBCSYSINI" );
```
mloskot, line 137: https://man7.org/linux/man-pages/man3/putenv.3.html
> The GNU C library implementation provides a nonstandard extension. If string does not include an equal sign:
>            `putenv("NAME");`
> then the named variable is removed from the caller's environment.

So, this probably is fine for Unix.

But, there needs to be `ODBCSYSINI=` for MSVC, I think, as per earlier comment https://github.com/OSGeo/gdal/pull/2838#discussion_r466874285

rouault: you could also more simply use ```pszEnvRollback = CPLStrdup("ODBCSYSINI=");```

nyalldawson: Thanks -- how's the new commit look?

mloskot: LGTM, Thx!

rouault:
Backported in 3.1 per 30a2560

PR #2839, opened 2020-08-07
Addition of -tr option for gdal_grid

Audifire:
## What does this PR do?
This pull requests implements the -tr parameter (mentioned here: https://lists.osgeo.org/pipermail/gdal-dev/2020-January/051472.html) to gdal_grid. The code is similar to already existing code in other binaries.
Tests have been added in PyTest, to show the equality of the -outsize and -tr parameter. Also documentation have been edited.

## Tasklist
 - [x] Add test case(s)
 - [x] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/autotest/utilities/test_gdal_grid.py
+++ b/autotest/utilities/test_gdal_grid.py
@@ -888,6 +888,52 @@ def test_gdal_grid_clipsrc():
     ds = None
 
 ###############################################################################
+# Test -tr
+
+
+def test_gdal_grid_tr():
+    if gdal_grid is None:
+        pytest.skip()
+
+    #################
+    outfiles.append('tmp/grid_count_70_70.tif')
+    try:
+        os.remove(outfiles[-1])
+    except OSError:
+        pass
+
+    # Create a GDAL dataset from the values of "grid.csv".
+    gdaltest.runexternal(gdal_grid + ' -txe 440720.0 441920.0 -tye 3751320.0 3750120.0 -tr 60 60 -ot Byte -l grid -a count:radius1=70.0:radius2=70.0:angle=0.0:min_points=0:nodata=0.0 data/grid.vrt ' + outfiles[-1])
+
+    # We should get the same values as in "ref_data/grid_count_70_70.tif"
+    ds = gdal.Open(outfiles[-1])
+    ds_ref = gdal.Open('ref_data/grid_count_70_70.tif')
+    assert ds.GetRasterBand(1).Checksum() == ds_ref.GetRasterBand(1).Checksum(), \
+        ('bad checksum : got %d, expected %d' %
+              (ds.GetRasterBand(1).Checksum(), ds_ref.GetRasterBand(1).checksum_ref))
+    ds_ref = None
+    ds = None
+
+    #################
+    outfiles.append('tmp/grid_count_300_300.tif')
+    try:
+        os.remove(outfiles[-1])
+    except OSError:
+        pass
+
+    # Create a GDAL dataset from the values of "grid.csv".
+    gdaltest.runexternal(gdal_grid + ' -txe 440720.0 441920.0 -tye 3751320.0 3750120.0 -tr 60 60 -ot Byte -l grid -a count:radius1=300.0:radius2=300.0:angle=0.0:min_points=0:nodata=0.0 data/grid.vrt ' + outfiles[-1])
+
+    # We should get the same values as in "ref_data/grid_count_300_300.tif"
+    ds = gdal.Open(outfiles[-1])
+    ds_ref = gdal.Open('ref_data/grid_count_300_300.tif')
+    assert ds.GetRasterBand(1).Checksum() == ds_ref.GetRasterBand(1).Checksum(), \
+        ('bad checksum : got %d, expected %d' %
+              (ds.GetRasterBand(1).Checksum(), ds_ref.GetRasterBand(1).checksum_ref))
+    ds_ref = None
+    ds = None
+
+###############################################################################
 # Cleanup
 
 
--- a/autotest/utilities/test_gdal_grid.py
+++ b/autotest/utilities/test_gdal_grid.py
@@ -914,25 +914,6 @@ def test_gdal_grid_tr():
     ds_ref = None
     ds = None
 
-    #################
-    outfiles.append('tmp/grid_count_300_300.tif')
-    try:
-        os.remove(outfiles[-1])
-    except OSError:
-        pass
-
-    # Create a GDAL dataset from the values of "grid.csv".
-    gdaltest.runexternal(gdal_grid + ' -txe 440720.0 441920.0 -tye 3751320.0 3750120.0 -tr 60 60 -ot Byte -l grid -a count:radius1=300.0:radius2=300.0:angle=0.0:min_points=0:nodata=0.0 data/grid.vrt ' + outfiles[-1])
-
-    # We should get the same values as in "ref_data/grid_count_300_300.tif"
-    ds = gdal.Open(outfiles[-1])
-    ds_ref = gdal.Open('ref_data/grid_count_300_300.tif')
-    assert ds.GetRasterBand(1).Checksum() == ds_ref.GetRasterBand(1).Checksum(), \
-        ('bad checksum : got %d, expected %d' %
-              (ds.GetRasterBand(1).Checksum(), ds_ref.GetRasterBand(1).checksum_ref))
-    ds_ref = None
-    ds = None
-
 ###############################################################################
 # Cleanup
 
--- a/gdal/apps/gdal_grid_lib.cpp
+++ b/gdal/apps/gdal_grid_lib.cpp
@@ -845,7 +845,7 @@ GDALDatasetH GDALGrid( const char *pszDest, GDALDatasetH hSrcDataset,
             nXSize = static_cast<int>(dfXSize);
             nYSize = static_cast<int>(dfYSize);
         } else {
-            CPLError( CE_Failure, CPLE_AppDefined, "Invalid output size detected. Please check your -tr argument");
+            CPLError( CE_Failure, CPLE_IllegalArg, "Invalid output size detected. Please check your -tr argument");
 
             if(pbUsageError)
                 *pbUsageError = TRUE;
@@ -1137,7 +1137,7 @@ GDALGridOptions *GDALGridOptionsNew(char** papszArgv, GDALGridOptionsForBinary* 
         {
             psOptions->dfXRes = CPLAtofM(papszArgv[++i]);
             psOptions->dfYRes = CPLAtofM(papszArgv[++i]);
-            if( psOptions->dfXRes == 0 || psOptions->dfYRes == 0 )
+            if( psOptions->dfXRes <= 0 || psOptions->dfYRes <= 0 )
             {
                 CPLError(CE_Failure, CPLE_IllegalArg, "Wrong value for -tr parameters.");
                 GDALGridOptionsFree(psOptions);
--- a/gdal/apps/gdal_grid_lib.cpp
+++ b/gdal/apps/gdal_grid_lib.cpp
@@ -32,6 +32,7 @@
 #include "gdal_utils_priv.h"
 #include "commonutils.h"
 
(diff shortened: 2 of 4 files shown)


gdal/apps/gdal_grid_lib.cpp
```diff
@@ -32,6 +32,7 @@
 #include "gdal_utils_priv.h"
 #include "commonutils.h"
 
+#include <cmath>
```
mloskot, line 35: Is anything from this header actually used?

gdal/apps/gdal_grid_lib.cpp
```diff
@@ -823,13 +832,27 @@ GDALDatasetH GDALGrid( const char *pszDest, GDALDatasetH hSrcDataset,
...
+        nXSize = static_cast<int>(
+            (psOptions->dfXMax - psOptions->dfXMin + (psOptions->dfXRes/2.0)) /
+            psOptions->dfXRes);
+        nYSize = static_cast<int>(
+            (std::fabs(psOptions->dfYMax - psOptions->dfYMin) + (psOptions->dfYRes/2.0)) /
+            psOptions->dfYRes);
+    }
+    else
+    {
+        // FIXME
```
mloskot, line 848: The comment should explain what the `FIXME` is about

gdal/apps/gdal_grid_lib.cpp
Audifire: I added the header for the use of std::fabs.

gdal/apps/gdal_grid_lib.cpp
Audifire: This comment was already existing in the source. I don't know, what it is for. Sorry for that.

gdal/apps/gdal_grid_lib.cpp
mloskot: My bad, sorry. I missed the `fabs` indeed.

gdal/apps/gdal_grid_lib.cpp
mloskot: Yes, indeed. The diff was a bit confusing.

autotest/utilities/test_gdal_grid.py
```diff
@@ -887,6 +887,52 @@ def test_gdal_grid_clipsrc():
...
+    ds = gdal.Open(outfiles[-1])
+    ds_ref = gdal.Open('ref_data/grid_count_70_70.tif')
+    assert ds.GetRasterBand(1).Checksum() == ds_ref.GetRasterBand(1).Checksum(), \
+        ('bad checksum : got %d, expected %d' %
+              (ds.GetRasterBand(1).Checksum(), ds_ref.GetRasterBand(1).checksum_ref))
+    ds_ref = None
+    ds = None
+
+    #################
+    outfiles.append('tmp/grid_count_300_300.tif')
```
rouault, line 918: I don't think that this second part of the test is actually needed. It just tests different values of the radius1/radius2 parameters, which are not impacted by this PR

autotest/utilities/test_gdal_grid.py
```diff
@@ -887,6 +887,52 @@ def test_gdal_grid_clipsrc():
...
+    outfiles.append('tmp/grid_count_70_70.tif')
+    try:
+        os.remove(outfiles[-1])
+    except OSError:
+        pass
+
+    # Create a GDAL dataset from the values of "grid.csv".
+    gdaltest.runexternal(gdal_grid + ' -txe 440720.0 441920.0 -tye 3751320.0 3750120.0 -tr 60 60 -ot Byte -l grid -a count:radius1=70.0:radius2=70.0:angle=0.0:min_points=0:nodata=0.0 data/grid.vrt ' + outfiles[-1])
+
+    # We should get the same values as in "ref_data/grid_count_70_70.tif"
```
rouault, line 908: this file has not been committed

gdal/apps/gdal_grid_lib.cpp
```diff
@@ -823,13 +832,27 @@ GDALDatasetH GDALGrid( const char *pszDest, GDALDatasetH hSrcDataset,
...
-    if ( nXSize == 0 )
-        nXSize = 256;
-    int nYSize = psOptions->nYSize;
-    if ( nYSize == 0 )
-        nYSize = 256;
+    int nXSize;
+    int nYSize;
+    if ( psOptions->dfXRes && psOptions->dfYRes )
+    {
+        nXSize = static_cast<int>(
```
rouault, line 839: it would be more prudent to be able to catch wrong input values to store the result of the computation before the static_cast<int> in a double, dfXSize, and test that dfXSize >= 1 && dfXSize <= INT_MAX, before casting to nXSize. Same for y

gdal/apps/gdal_grid_lib.cpp
```diff
@@ -1099,6 +1124,18 @@ GDALGridOptions *GDALGridOptionsNew(char** papszArgv, GDALGridOptionsForBinary*
             i += 2;
         }
 
+        else if( i+2 < argc && EQUAL(papszArgv[i],"-tr") )
+        {
+            psOptions->dfXRes = CPLAtofM(papszArgv[++i]);
+            psOptions->dfYRes = CPLAtofM(papszArgv[++i]);
+            if( psOptions->dfXRes == 0 || psOptions->dfYRes == 0 )
```
rouault, line 1131:
```suggestion
            if( psOptions->dfXRes <= 0 || psOptions->dfYRes <= 0 )
```

gdal/doc/source/programs/gdal_grid.rst
```diff
@@ -54,9 +54,14 @@ computer.
 
     Set georeferenced Y extents of output file to be created.
 
+.. option:: -tr <xres> <yres>
+
+    Set output file resolution (in target georeferenced units)
```
rouault, line 59:
```suggestion
    Set output file resolution (in target georeferenced units)
    
    .. versionadded:: 3.2
```

gdal/apps/gdal_grid_lib.cpp
rouault, line 839: This also assumes that -txe and -tye have been specified. Making the code to work when they are not specified would require more changes, so if you don't want to proceed to them, you should at least test that the bounds have been set, and error out if not. And document that restriction in the .rst

autotest/utilities/test_gdal_grid.py
das-g: > this file has not been committed

This isn't a new file. It was already (and is still) being used by another test:
https://github.com/OSGeo/gdal/blob/d7bbdeff5c705c3eac51e1f8500d09e467a27e00/autotest/utilities/test_gdal_grid.py#L635-L680

Thus, it [was already there before this PR]( https://github.com/OSGeo/gdal/blob/d7bbdeff5c705c3eac51e1f8500d09e467a27e00/autotest/utilities/ref_data/grid_count_70_70.tif) and still is.

rouault: ah, ok my bad, I wrongly assumed it was a new one.

gdal/apps/gdal_grid_lib.cpp
```diff
@@ -823,13 +832,36 @@ GDALDatasetH GDALGrid( const char *pszDest, GDALDatasetH hSrcDataset,
...
-    // FIXME
-    int nXSize = psOptions->nXSize;
-    if ( nXSize == 0 )
-        nXSize = 256;
-    int nYSize = psOptions->nYSize;
-    if ( nYSize == 0 )
-        nYSize = 256;
+    int nXSize;
+    int nYSize;
+    if ( psOptions->dfXRes && psOptions->dfYRes )
```
rouault, line 838: To satisfy our pedantic compiler warning rules:
```suggestion
    if ( psOptions->dfXRes != 0 && psOptions->dfYRes != 0 )
```

gdal/apps/gdal_grid_lib.cpp
```diff
@@ -823,13 +833,45 @@ GDALDatasetH GDALGrid( const char *pszDest, GDALDatasetH hSrcDataset,
...
+                    "Invalid txe or tye paramaters detected. Please check your -txe or -tye argument.");
+
+            if(pbUsageError)
+                *pbUsageError = TRUE;
+            return nullptr;
+        }
+
+        double dfXSize = (std::fabs(psOptions->dfXMax - psOptions->dfXMin) + (psOptions->dfXRes/2.0)) /
+            psOptions->dfXRes;
+        double dfYSize = (std::fabs(psOptions->dfYMax - psOptions->dfYMin) + (psOptions->dfYRes/2.0)) /
```
Audifire, line 851: I have one question. In the following code snippet, only one component uses fabs. Should it be done like that or is my way the correct way?

Code: https://github.com/OSGeo/gdal/blob/d88aabcb211233977ed202918182db23e12a3652/gdal/apps/gdalwarp_lib.cpp#L3233-L3238

rouault: your way is fine (although xmax < xmin is a very special use case)

autotest/utilities/test_gdal_grid.py
Audifire: That's right. I didn't realise that. Fixed in a488626

gdal/apps/gdal_grid_lib.cpp
Audifire: Added in 6641ab9

gdal/apps/gdal_grid_lib.cpp
Audifire: Added in ad92808

gdal/doc/source/programs/gdal_grid.rst
Audifire: Added in 2f7e800

gdal/apps/gdal_grid_lib.cpp
Audifire: Added in 7ad261a and 18ef73c

gdal/apps/gdal_grid_lib.cpp
```diff
@@ -823,13 +833,45 @@ GDALDatasetH GDALGrid( const char *pszDest, GDALDatasetH hSrcDataset,
...
-    if ( nXSize == 0 )
-        nXSize = 256;
-    int nYSize = psOptions->nYSize;
-    if ( nYSize == 0 )
-        nYSize = 256;
+    int nXSize;
+    int nYSize;
+    if ( psOptions->dfXRes != 0 && psOptions->dfYRes != 0 )
+    {
+        if (!(psOptions->dfXMax - psOptions->dfXMin) || !(psOptions->dfYMax - psOptions->dfYMin)) {
```
rouault, line 840:
```suggestion
        if ((psOptions->dfXMax == psOptions->dfXMin) || (psOptions->dfYMax == psOptions->dfYMin)) {
```

gdal/apps/gdal_grid_lib.cpp
```diff
@@ -768,6 +772,12 @@ GDALDatasetH GDALGrid( const char *pszDest, GDALDatasetH hSrcDataset,
         return nullptr;
     }
 
+    if ( (psOptions->nXSize || psOptions->nYSize) && (psOptions->dfXRes || psOptions->dfYRes) ) {
```
rouault, line 775:
```suggestion
    if ( (psOptions->nXSize != 0 || psOptions->nYSize != 0) && (psOptions->dfXRes != 0 || psOptions->dfYRes != 0) ) {
```

Audifire: I think that made warnings too. But where do you see them? My local compiler doesn't show me any warnings.
Sometimes I think too complicated to find the solution. I really appreciate your work, thank you for this bug review.

rouault: This was in the build log of https://travis-ci.com/github/OSGeo/gdal/jobs/371646974  . If you look for the string "error: " you get on the lines with the warnings/errors.
They depend on compiler types & versions, so you can't catch everything locally.

Audifire: Ahh perfect. Thank you very much.

PR #2846, opened 2020-08-09
MRF Interleaved LERC

lucianpls:
## What does this PR do?

MRF: Add interleaved LERC support.  Streamlined LERC1 codec

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [x] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:


.gitignore
```diff
@@ -24,3 +24,6 @@ __pycache__/
 # CMake
 CMakeLists.txt
 *.cmake
+
+gdal/packages*
```
rouault, line 28: those additions seem to be particular to your setup ?

lucianpls: Not that particular, that's where NuGet package manager places it's files.

rouault: ok, can you add a comment in .gitignore then ?

PR #2847, opened 2020-08-10
Followup 69d00e30c: also quote DBQ values in other drivers

nyalldawson:
I missed these in the original PR, sorry!


rouault:
cherry-picked in 3.1 branch per bb05adb

PR #2852, opened 2020-08-12
Fixes corruption with geometry data when reading large features from pgeo driver

nyalldawson:
Fix data from binary ole fields can be incorrectly retrieved if multiple calls to SQLGetData are required and one call
ends with multiple null characters, which are actually part of the binary content

Fixes corruption with geometry data when reading large features from pgeo driver

### DIFF
--- a/gdal/port/cpl_odbc.cpp
+++ b/gdal/port/cpl_odbc.cpp
@@ -990,6 +990,11 @@ int CPLODBCStatement::Fetch( int nOrientation, int nOffset )
                 if( nFetchType == SQL_C_CHAR )
                     while( (cbDataLen > 1) && (szWrkData[cbDataLen - 1] == 0) )
                         --cbDataLen;  // Trimming the extra terminators: bug 990
+                else if( nFetchType == SQL_C_BINARY )
+                {
+                    if( (cbDataLen > 1) && (szWrkData[cbDataLen - 1] == 0) )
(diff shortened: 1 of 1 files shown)


nyalldawson:
Actually ... looks like this may be a bug in the mdbtools odbc driver instead. It's incorrectly returning giving the field type as SQL_C_CHAR instead of the correct SQL_C_BINARY value...

nyalldawson:
Ok, here's a corrected fix...

nyalldawson:
Reference for this: https://docs.microsoft.com/en-us/sql/odbc/reference/syntax/sqlgetdata-function?view=sql-server-ver15

> To retrieve data from a column in parts, the application calls SQLGetData multiple times in succession for the same column. On each call, SQLGetData returns the next part of the data. It is up to the application to reassemble the parts, taking care to remove the null-termination character from intermediate parts of character data

nyalldawson:
This is for review only, not merge. It's possible that the mdbtools driver is actually misbehaved here (see https://github.com/evanmiller/mdbtools/pull/8#issuecomment-672608007 )

rouault:
It seems to break the tests of the MSSQLSpatial driver. The removal of the last byte of Binary field looks weird to me, but I'm not a ODBC expert

nyalldawson:
Ok, not required -- will get a fix in mdbtools odbc driver fixing this instead.

PR #2854, opened 2020-08-13
[pgeo][docs] Add note about maintained mdbtools fork 

nyalldawson:
...which works correctly with gdal for reading Personal Geodatabase files

### DIFF
--- a/gdal/doc/source/drivers/vector/pgeo.rst
+++ b/gdal/doc/source/drivers/vector/pgeo.rst
@@ -64,8 +64,10 @@ Prerequisites
 ~~~~~~~~~~~~~
 
 #. Install `unixODBC <http://www.unixodbc.org>`__ >= 2.2.11
-#. Install `MDB Tools <http://mdbtools.sourceforge.net/>`__ >= 0.6. I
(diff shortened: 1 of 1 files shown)


nyalldawson:
@mloskot I can't reproduce the difficulties you outline in this document (and at http://sourceforge.net/mailarchive/message.php?msg_id=5998236) regarding DSN-less connections. Is that information still up to date? Or can I safely assume it's since been fixed and is removable?

mloskot:
@nyalldawson You've been digging quite a history, I see :-)
I have not tested MDBTools myself, but I'm pretty sure that issue is no longer relevant. 

Long time ago, it was an issue though, as Brian Bruns confirmed on unixodbc-dev mailnig list 17 years ago https://unixodbc-dev.unixodbc.narkive.com/CDFl3Q4m/dsn-less-connections  (original mailing list is now broken, no access to archives):

> MDB Tools doesn't yet support DSN-less connections. Support should be
> fairly trivial to add, but I don't have the time.

 Possibly, DSN-less connection is supported [since 0.7.x as some report](https://stackoverflow.com/a/23373942/151641) and 0.7 was released in 2012, so long time after I posted that issue.

Issue #2824, opened 2020-08-03
Unexpected behaviour of ogr2ogr postgres connection string using non default schema


JGGeo:
Dear esteemed developers,

I just wanted to report a possible bug in the ogr2ogr tool.

## Expected behavior and actual behavior.
I want to export a PostGIS table from a defined schema into a shape file.

It works fine using the "input dataset open options" (-oo) for the schema definitions in the command:
```
ogr2ogr -overwrite -oo "ACTIVE_SCHEMA=userschema" -oo "SCHEMAS=userschema" -f "ESRI Shapefile"  "C:\Temp\temp.shp" PG:"host='localhost' port='5432' dbname='dbtemp' user='myuser' password='myuser'" "geotable"
```

It will not work using the "connection string" (PG:) for the schema definitions in the command:
```
ogr2ogr.exe -overwrite -f "ESRI Shapefile" "C:\Temp\temp.shp" PG:"host='localhost' port='5432' dbname='dbtemp' schemas='userschema' active_schema='userschema' user='myuser' password='myuser'" "geotable"
```
Debug Messsage of the failig second command:
> PG: PostGIS schema: 'public'
> PG: PostgreSQL version string : 'PostgreSQL 9.6.15, compiled by Visual C++ build 1800, 64-bit'
> PG: PostGIS version string : '2.5 USE_GEOS=1 USE_PROJ=1 USE_STATS=1'
> GDAL: GDALOpen(PG:host='localhost' port='5432' dbname='dbtemp' schemas='userschema' active_schema='userschema' user='myuser' password=XXXXXX this=0000017B8920BC50) succeeds as PostgreSQL.
> GDAL: GDALDriver::Create(ESRI Shapefile,C:\Temp\temp.shp,0,0,0,Unknown,0000000000000000)
> PG: Could not retrieve table oid for geotable
> ERROR 1: Couldn't fetch requested layer 'geotable'!
> GDAL: GDALClose(C:\Temp\temp.shp, this=0000017B8922ABC0)
> GDAL: GDALClose(PG:host='localhost' port='5432' dbname='dbtemp' schemas='userschema' active_schema='userschema' user='myuser' password=XXXXXX this=0000017B8920BC50)
> GDAL: In GDALDestroy - unloading GDAL shared library.

Am I wrong expecting the same behaviour of the two commands?

## Steps to reproduce the problem.

See above. It should be reproducable with any normal PostGIS table.

## Operating system

Windows 10 x64

[...]

PR #2855, opened 2020-08-13
PG docs: use documented syntax in examples

jratike80:
Relates: https://github.com/OSGeo/gdal/issues/2824

In the last example I changed "poly.shp" into "poly" but I am not sure if was poly.shp intentionally for some reason.


rouault:
Thanks

PR #2870, opened 2020-08-20
Correctly handle datetime fields provided by the mdbtools ODBC driver

nyalldawson:
Avoids forced conversion of these fields to strings

### DIFF
--- a/gdal/ogr/ogrsf_frmts/geomedia/ogrgeomedialayer.cpp
+++ b/gdal/ogr/ogrsf_frmts/geomedia/ogrgeomedialayer.cpp
@@ -165,6 +165,7 @@ CPLErr OGRGeomediaLayer::BuildFeatureDefn( const char *pszLayerName,
             break;
 
           case SQL_C_TIMESTAMP:
+          case SQL_C_TYPE_TIMESTAMP:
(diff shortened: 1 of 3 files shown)


gdal/ogr/ogrsf_frmts/geomedia/ogrgeomedialayer.cpp
```diff
@@ -165,6 +165,7 @@ CPLErr OGRGeomediaLayer::BuildFeatureDefn( const char *pszLayerName,
             break;
 
           case SQL_C_TIMESTAMP:
+          case SQL_TYPE_TIMESTAMP:
```
mloskot, line 168: This constant is SQL data type, but shouldn't this be constant for the [C data type](https://docs.microsoft.com/en-us/sql/odbc/reference/appendixes/c-data-types) `SQL_C_TYPE_TIMESTAMP`?
Numeric value of both can be equal though.

nyalldawson: Thanks, fixed

rouault:
Cherry-picked in 3.1 branch

PR #2871, opened 2020-08-20
SQLite and GPKG drivers: add a PRELUDE_STATEMENTS open option

rouault:
... that can be used for example to attach other databases


jratike80:
Looks good, also the documenting parts and examples.

PR #2876, opened 2020-08-20
grass driver: fix reading GRASS groups

metzm:
fix support for GRASS groups = several GDAL raster bands

## What does this PR do?
Currently it is not possible to read GRASS groups because the GRASS region is reset with every read request. This works fine with a single GRASS raster, but not with a GRASS group of rasters, because the rasters are not properly opened and closed. This PR opens GRASS rasters only for actual data reading and closes them when data reading is finished.

Reviewers, please check if the PR adheres to GDAL coding standards. The change itself behaves as expected for both single GRASS raster maps (GDAL dataset with a single band) and GRASS raster groups (GDAL dataset with multiple bands).

### DIFF
--- a/gdal/frmts/grass/grass57dataset.cpp
+++ b/gdal/frmts/grass/grass57dataset.cpp
@@ -299,10 +299,8 @@ GRASSRasterBand::GRASSRasterBand( GRASSDataset *poDSIn, int nBandIn,
     nBlockYSize = 1;
 
     G_set_window( &(poDSIn->sCellInfo) );
-    if ( (hCell = G_open_cell_old((char *) pszCellName, (char *) pszMapset)) < 0 ) {
-        CPLError( CE_Warning, CPLE_AppDefined, "GRASS: Cannot open raster '%s'", pszCellName );
-        return;
-    }
+    // open the raster only for actual reading
+    hCell = -1;
     G_copy((void *) &sOpenWindow, (void *) &(poDSIn->sCellInfo), sizeof(struct Cell_head));
 
 /* -------------------------------------------------------------------- */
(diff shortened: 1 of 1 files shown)


gdal/frmts/grass/grass57dataset.cpp
```diff
@@ -464,8 +456,19 @@ CPLErr GRASSRasterBand::ResetReading ( struct Cell_head *sNewWindow )
              sNewWindow->rows   != sCurrentWindow.rows   || sNewWindow->cols   != sCurrentWindow.cols
              )
         {
+            if( hCell >= 0 ) {
```
rouault, line 459: as those same lines seem to be used at line 431 too, maybe it would make sense to move them to a dedicated method ?

gdal/frmts/grass/grass57dataset.cpp
```diff
@@ -487,6 +490,13 @@ CPLErr GRASSRasterBand::IReadBlock( int /*nBlockXOff*/, int nBlockYOff,
     if ( ResetReading ( &(((GRASSDataset *)poDS)->sCellInfo) ) != CE_None ) {
        return CE_Failure;
     }
+    // open for reading
+    if (hCell < 0) {
+        if ( (hCell = G_open_cell_old((char *) pszCellName, (char *) pszMapset)) < 0 ) {
+            CPLError( CE_Warning, CPLE_AppDefined, "GRASS: Cannot open raster '%s'", pszCellName );
```
rouault, line 496:
```suggestion
            CPLError( CE_Failure, CPLE_AppDefined, "GRASS: Cannot open raster '%s'", pszCellName );
```

gdal/frmts/grass/grass57dataset.cpp
```diff
@@ -562,6 +576,13 @@ CPLErr GRASSRasterBand::IRasterIO ( GDALRWFlag eRWFlag,
     {
         return CE_Failure;
     }
+    // open for reading
+    if (hCell < 0) {
+        if ( (hCell = G_open_cell_old((char *) pszCellName, (char *) pszMapset)) < 0 ) {
+            CPLError( CE_Warning, CPLE_AppDefined, "GRASS: Cannot open raster '%s'", pszCellName );
```
rouault, line 582:
```suggestion
            CPLError( CE_Failure, CPLE_AppDefined, "GRASS: Cannot open raster '%s'", pszCellName );
```

gdal/frmts/grass/grass57dataset.cpp
metzm: It's only four lines of code that appear only twice, and only in `GRASSRasterBand::ResetReading`.
Maybe move them to the top of `GRASSRasterBand::ResetReading` to avoid code duplication? The actual data reading functions `GRASSRasterBand::IReadBlock` and `GRASSRasterBand::IRasterIO` now always open the GRASS raster if not yet opened.

rouault: > Maybe move them to the top of `GRASSRasterBand::ResetReading` to avoid code duplication?

if that's doable, yes, that would be better for future maintainance

PR #2878, opened 2020-08-20
Fix DSN string construction for Windows Access ODBC driver, template candidate preference order

nyalldawson:
The Windows ODBC driver does not like quoted filenames, yet the UnixODBC driver requires these (for filenames with spaces). So
ifdef the expected formats for the corresponding platforms. Follow up https://github.com/OSGeo/gdal/pull/2837

Also flip the order of the candidate DSN string templates, to try the more common/modern "Microsoft Access Driver (*.mdb, *.accdb)" driver name first.

### DIFF
--- a/gdal/port/cpl_odbc.cpp
+++ b/gdal/port/cpl_odbc.cpp
@@ -446,7 +446,7 @@ bool CPLODBCSession::ConnectToMsAccess(const char *pszName, const char *pszDSNSt
         pszDSNStringTemplate = "DRIVER=Microsoft Access Driver (*.mdb, *.accdb);DBQ=\"%s\"";
 #endif
     }
-    pszDSN = (char *) CPLMalloc(strlen(pszName)+strlen(pszDSNStringTemplate)+100);
+    pszDSN = static_cast< char * >( CPLMalloc(strlen(pszName)+strlen(pszDSNStringTemplate)+100) );
     /* coverity[tainted_string] */
     snprintf( pszDSN,
         strlen(pszName)+strlen(pszDSNStringTemplate)+100,
@@ -463,7 +463,7 @@ bool CPLODBCSession::ConnectToMsAccess(const char *pszName, const char *pszDSNSt
         pszDSNStringTemplate = "DRIVER=Microsoft Access Driver (*.mdb);DBQ=\"%s\"";
 #endif
         CPLFree( pszDSN );
-        pszDSN = (char *) CPLMalloc(strlen(pszName)+strlen(pszDSNStringTemplate)+100);
+        pszDSN = static_cast< char * >( CPLMalloc(strlen(pszName)+strlen(pszDSNStringTemplate)+100) );
         snprintf( pszDSN,
             strlen(pszName)+strlen(pszDSNStringTemplate)+100,
             pszDSNStringTemplate,  pszName );
@@ -477,11 +477,11 @@ bool CPLODBCSession::ConnectToMsAccess(const char *pszName, const char *pszDSNSt
                   "Unable to initialize ODBC connection to DSN for %s,\n"
                   "%s", pszDSN, GetLastError() );
         CPLFree( pszDSN );
-        return FALSE;
+        return false;
     }
 
     CPLFree( pszDSN );
-    return TRUE;
+    return true;
 }
 
 /************************************************************************/
--- a/gdal/port/cpl_odbc.cpp
+++ b/gdal/port/cpl_odbc.cpp
@@ -413,6 +413,63 @@ int CPLODBCSession::Failed( int nRetCode, HSTMT hStmt )
 }
 
 /************************************************************************/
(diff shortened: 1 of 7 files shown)


rouault:
We have a fair amount of copy & paste between all those drivers. I'm wondering if some sort of utility functions/class in ogr/ogrsf_frmts/odbc/ couldn't be added and used by all those drivers ?

nyalldawson:
@rouault how's the new version?

gdal/port/cpl_odbc.cpp
```diff
@@ -412,6 +412,63 @@ int CPLODBCSession::Failed( int nRetCode, HSTMT hStmt )
...
+        bError = !EstablishSession( pszDSN, nullptr, nullptr );
+    }
+
+    if ( bError )
+    {
+        CPLError( CE_Failure, CPLE_AppDefined,
+                  "Unable to initialize ODBC connection to DSN for %s,\n"
+                  "%s", pszDSN, GetLastError() );
+        CPLFree( pszDSN );
+        return FALSE;
```
rouault, line 465:
```suggestion
        return false;
```

gdal/port/cpl_odbc.cpp
```diff
@@ -412,6 +412,63 @@ int CPLODBCSession::Failed( int nRetCode, HSTMT hStmt )
...
+    }
+
+    CPLFree( pszDSN );
+    return TRUE;
```
rouault, line 469:
```suggestion
    return true;
```

gdal/port/cpl_odbc.cpp
```diff
@@ -412,6 +412,63 @@ int CPLODBCSession::Failed( int nRetCode, HSTMT hStmt )
     return TRUE;
 }
 
+/************************************************************************/
+/*                          ConnectToMsAccess()                          */
+/************************************************************************/
+
+
```
rouault, line 419: Doxygen comment needed

gdal/port/cpl_odbc.cpp
```diff
@@ -412,6 +412,63 @@ int CPLODBCSession::Failed( int nRetCode, HSTMT hStmt )
...
+    if( usingAutomaticDSNStringTemplate )
+    {
+#ifdef WIN32
+        pszDSNStringTemplate = "DRIVER=Microsoft Access Driver (*.mdb, *.accdb);DBQ=%s";
+#else
+        pszDSNStringTemplate = "DRIVER=Microsoft Access Driver (*.mdb, *.accdb);DBQ=\"%s\"";
+#endif
+    }
+    pszDSN = (char *) CPLMalloc(strlen(pszName)+strlen(pszDSNStringTemplate)+100);
```
rouault, line 434: code in port/ no longer allows C style casts. So either use static_cast<char*> (here and at line 451). Or just go for std::string

rouault (changes requested):
Thanks. A few changes needed to make linting happy

nyalldawson:
Ok, updated. In the new function dox I used "@since 3.2", but these changes should be backported in order to fix the regression introduced by #2837

rouault:
on OSX:
```
 ogr/ogr_odbc.py::test_ogr_odbc_1 SKIPPED                                 [ 15%]
./gdal/ci/travis/osx/script.sh: line 17: 80569 Segmentation fault: 11  $PYTEST
ogr/ogr_odbc.py::test_ogr_odbc_2 
```

```
ogr/ogrsf_frmts/pgeo/ogrpgeodatasource.cpp:111,style,redundantAssignment,Variable 'pszDSN' is reassigned a value before the old one has been used.
ogr/ogrsf_frmts/walk/ogrwalkdatasource.cpp:76,style,redundantAssignment,Variable 'pszDSN' is reassigned a value before the old one has been used.
```

nyalldawson:
@rouault fixed

nyalldawson:
@rouault is there anyway to get more detailed debugging as to why the odbc test is failing on windows? I can't reproduce this locally on my Windows build...

rouault:
> is there anyway to get more detailed debugging as to why the odbc test is failing on windows?

you may try to add a manual `pytest ogr_odbc.py --capture=no -ra -vv` with `set CPL_DEBUG=ON` set before. You may need to instrument with extra debug messages

rouault:
Note: if there is no resolution for this before next Tuesday where I'll cut a 3.1.3 RC, I'll have to revert 1f71651d6bfda5a76dd7a449cc45dbf5022ec152 in 3.1 branch since I understand it causes a regression.

nyalldawson:
@rouault 
48ae0cb should fix this -- the github Windows build workflow only has the older Access ODBC driver available, which doesn't support reading accdb database files

rouault:
Backported to 3.1 in e2c83edf7862ae3c035c17fba3bf9abf6ad3aec9 (without the new tests that depend on test files not backported)

PR #2879, opened 2020-08-20
Fix ODBC driver fails to correctly fallback to alternative Access ODBC driver name

nyalldawson:
...which causes an error flag to be set and no
alternative drivers to be used to attempt to open a MDB dataset

This breaks the PGEO driver (among others) on Windows, which ultimately
requires an "OGR_SKIP=ODBC" environment variable to be set in order
to read Personal Geodatabases on Windows

Refs:
https://trac.osgeo.org/gdal/ticket/5594#comment:4
https://github.com/qgis/QGIS/issues/20256#issuecomment-495841943


rouault:
Cherry-picked in 3.1 branch

PR #2880, opened 2020-08-20
Generic ODBC driver can read MS Access databases with ACCDB and STYLE extensions

nyalldawson:
These are alternative extensions for Access databases, and work perfectly with the generic ODBC driver


nyalldawson:
@rouault update PR pushed

PR #2885, opened 2020-08-24
Minor improvement to the XSD documentation for PDF Composition

robintw:
##  What does this PR do?
Makes a minor improvement to the XSD documentation for the PDF Composition XML format. My colleagues and I struggled to get labels working because we didn't realise that any features with a OGR Style String containing a `LABEL` tool needed to be in a separate `<VectorLabel>` element. This adds an extra sentence to the documentation that clarifies this.

None


robintw:
Hmm, this seems to have failed a test relating to building PROJ, which I don't really understand because I've only changed the XSD file.

PR #2887, opened 2020-08-25
LVBAG: Fixes #2804

### LINKED ISSUE #2804: LVBAG: Check LokaalId conforms to 16 characters

The BAG format of LokaalId is a 16 characters long number, stored as text in their database. Some LokaalId fields in the BAG database aren't 16 characters long. The original numbers often start with one or more zeroes, which could sometims get discarded by accident when casting the LokaalId to a number before storing it as text in the database. These cases should be filtered out by prepending zeroes to the values until they have their original length of 16 characters.

3.1.2-master

yorickdewid:
## What does this PR do?
Fixes issue #2804 

## Tasklist

 - [ ] Add test case(s)
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/gdal/doc/source/drivers/vector/lvbag.rst
+++ b/gdal/doc/source/drivers/vector/lvbag.rst
@@ -23,6 +23,8 @@ field according to the BAG 2.0 specfication.
 
 More information about the LV BAG 2.0 can be found at https://zakelijk.kadaster.nl/bag-2.0
 
+LV BAG model definitions are available at https://www.kadaster.nl/-/xsd-s-bag-2.0-extract
+
 Note 1 : the earlier BAG 1.0 extract is **not supported**\ by this driver.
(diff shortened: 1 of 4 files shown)


gdal/doc/source/drivers/vector/lvbag.rst
```diff
@@ -23,8 +23,6 @@ field according to the BAG 2.0 specfication.
 
 More information about the LV BAG 2.0 can be found at https://zakelijk.kadaster.nl/bag-2.0
 
-LV BAG model definitions are available at https://zakelijk.kadaster.nl/documents/20838/87954/XSD%27s+BAG+2.0+Extract
-
```
mloskot, line 27: If this link is giving 404, is there any chance to update the link instead of removing it?

PR #2892, opened 2020-08-26
GTiff: fix wrong direction for half-pixel shift with GCPs and PixelIsPoint convention

rouault:
Fixes https://github.com/opengeospatial/OGC-API-Sprint-August-2020/issues/18


rouault:
Backported to 3.1

PR #2895, opened 2020-08-28
Add gdal_create utility to create/initialize a new raster file

rouault:


jratike80:
The html Documentation is mentioning source layer in couple of cases. Does it mean something else than that copy-paste is used? And example contains parameter -burn that would be nice to have but it is not mentioned elsewhere.

rouault:
Thanks for the review @jratike80 

> The html Documentation is mentioning source layer in couple of cases

ah ok, yes "source file" is indeed copied&pasted from gdal_translate. Edited

> And example contains parameter -burn that would be nice to have but it is not mentioned elsewhere.

Good point. I've added it in synopsis and documentation

PR #2817, opened 2020-07-31
buffered reader has configurable size, and makes reads filling whole buffer

tbonfort:
This PR modifies `VSIBufferedReaderHandle` so that it emits reads to its underlying `VSIVirtualHandle` that fill up the private buffer. The buffer size becomes configurable.

PR #2901, opened 2020-08-31
VSI Plugin: add caching option

tbonfort:
This PR adds an option to wrap a plugin handler with `VSICreateCachedFile`
Can/should be backported
Replaces #2817


rouault:
Backported to 3.1

Issue #2241, opened 2020-02-13
gdal_contour hangs when attempting to generate polygons


owvarley:
## Expected behavior and actual behavior.
Running the command ```gdal_contour -i 60.96 -amin height -p /home/dted/e105/n09_preprocessed.dt1 output.shp``` does not produce any output and the process runs indefinitely. Expected behaviour was the production of a shapefile containing contour polygons spaced every 60.96m. I am aware that gdal_contour -p takes a very long time, however, the average for the tiles I've been working with is between 3-5 minutes a tile. This process is still running after 5 hours. The command was run on 641 tiles successfully with each tile taking a maximum of 5 minutes to complete. 

Running the command without the -p and with an interval of 15.24 correctly generates contour lines so it only seems to be the contour polygons that are failing.

The DTED tile is for e105, n09 which puts it over the sea to the north of Indonesia. The command successfully runs on the other tiles surrounding this location and, as previously stated, dropping the -p works which doesn't make me think the tile is corrupted.

No error messages or output was observed.

## Steps to reproduce the problem.
The attached zip file ([e105n09.zip](https://github.com/OSGeo/gdal/files/4200473/e105n09.zip)) contains three files in it:

1. n09.dt2 - The original DTED Level 2 file for the area
2. n09_fill.dt2 - The original DTED Level 2 file for the area with the voids filled using GDAL
3. n09_preprocessed.dt1 - The n09_fill tile downsampled to DTED Level 1

Running the command ```gdal_contour -i 60.96 -amin height -p n09_preprocessed.dt1 output.shp``` will produce the same results as outlined above.

I'd suggest also running ```gdal_contour -i 15.24 -a height n09_preprocessed.dt1 output_lines.shp``` to prove my theory above as this worked for me.

## Operating system

Running in a docker container using the debian:buster-slim image. GDAL installed using apt-get with the following packages:
- gdal-bin
- python3-gdal
- python3-pip
- python3-setuptools
- postgresql-client
- postgis

## GDAL version and provenance
GDAL 2.4.0, released 2018/12/14

rouault:
CC @mhugo  I can reproduce the performance problem with master. The following snippet from polygon_ring_appender.h takes an enormous runtime due to itLevel.second having 46474 elements, hence 2.2 billion iterations
```
        // compute inner rings
        for ( auto& itLevel: rings_ ) {
            for ( auto& currentRing: itLevel.second ) {
                for ( const auto& otherRing: itLevel.second ) {
[...]

PR #2908, opened 2020-09-01
Rework polygon_ring_appender to avoid slowness

pwicks86:
## What does this PR do?

This patch reworks `polygon_ring_appender` to avoid some cases that caused
massive slowness. This involved doing 2 main things:

1. The `isIn` check has been reworked to avoid a dependency on OGR and
instead the winding number algorithm has been implemented directly. The
code itself was based on the reference code at geomalgorithms.com

2. Rather than sort all the rings in the destructor, the `addLine`
function has been reworked to put each new ring in the correct position
as it is added. To enable this, the `Ring` class was reworked to hold a
vector of child instances directly, effectively building a tree of
rings. Then in the destructor, we can simply walk the tree and write out
every other level as a polygon.

#2241

## Tasklist

 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS: Ubuntu 18.04
* Compiler: gcc 7.5

### DIFF
--- a/gdal/alg/marching_squares/point.h
+++ b/gdal/alg/marching_squares/point.h
@@ -59,6 +59,16 @@ std::ostream & operator<<(std::ostream & o, const Point& p)
     return o;
 }
 
+// Test if a point is to the left, right or on an infinite line.
+// Returns greater than zero for p2 left of the line defined by p0 and p1,
+// 0 if p2 is on the line and less than if p2 is to the right of the line
+inline int
+isLeft(Point p0, Point p1, Point p2 )
+{
+    return static_cast<int>(( (p1.x - p0.x) * (p2.y - p0.y)
+            - (p2.x -  p0.x) * (p1.y - p0.y) ));
+}
+
 // LineString type
 typedef std::list<Point> LineString;
 
--- a/gdal/alg/marching_squares/point.h
+++ b/gdal/alg/marching_squares/point.h
@@ -59,14 +59,14 @@ std::ostream & operator<<(std::ostream & o, const Point& p)
     return o;
 }
 
-// Test if a point is to the left, right or on an infinite line.
-// Returns greater than zero for p2 left of the line defined by p0 and p1,
+// Test if a point is to the left or right of an infinite line.
+// Returns true if it is to the left and right otherwise
 // 0 if p2 is on the line and less than if p2 is to the right of the line
-inline int
-isLeft(Point p0, Point p1, Point p2 )
+inline bool
+isLeft(const Point& p0, const Point& p1, const Point& p2 )
 {
-    return static_cast<int>(( (p1.x - p0.x) * (p2.y - p0.y)
-            - (p2.x -  p0.x) * (p1.y - p0.y) ));
+    return ( (p1.x - p0.x) * (p2.y - p0.y)
+            - (p2.x -  p0.x) * (p1.y - p0.y) ) > 0;
 }
 
 // LineString type
--- a/gdal/alg/marching_squares/polygon_ring_appender.h
+++ b/gdal/alg/marching_squares/polygon_ring_appender.h
@@ -65,24 +65,27 @@ private:
             auto otherIter = other.points.begin();
             // p1 and p2 define each segment of the ring other that will be tested
             auto p1 = *otherIter;
-            do {
+            while(true) {
                 otherIter++;
+                if (otherIter == other.points.end()) {
+                    break;
+                }
                 auto p2 = *otherIter;
                 if ( p1.y <= checkPoint.y ) {
                     if ( p2.y  > checkPoint.y ) {
-                        if ( isLeft(p1, p2, checkPoint) > 0 )  {
+                        if ( isLeft(p1, p2, checkPoint) )  {
                              ++windingNum;
                         }
                     }
                 } else {
                     if ( p2.y <= checkPoint.y ) {
-                        if ( isLeft( p1, p2, checkPoint) < 0 ) {
+                        if ( !isLeft( p1, p2, checkPoint)  ) {
                             --windingNum;
                         }
                     }
                 }
                 p1 = p2;
-            } while(otherIter != other.points.end());
+            }
             return windingNum != 0;
         }
 
@@ -102,7 +105,7 @@ private:
 #endif
     };
 
-    void processTree(std::vector<Ring> &tree, int level) {
+    void processTree(const std::vector<Ring> &tree, int level) {
         if ( level % 2 == 0 ) {
             for( auto &r: tree ) {
                 writer_.addPart(r.points);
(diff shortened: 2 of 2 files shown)


pwicks86:
Huh, not sure why this fails on MacOS. I don't have an Apple machine to test on at the moment unfortunately.

rouault:
> Huh, not sure why this fails on MacOS.

Just look at the logs :-) Linux builds also fail, so this is not platform specific
On ./gdal_unit_test in autotest/cpp (``make gdal_unit_test``):
```
---> group: MarchingSquares:Polygon, test: test<4>
     problem: assertion failed
     failed assertion: "Polygon #1: expected '{ { (2.25,2.5) (2.5,2.25) (3,2.25) (3,2) (3,1.5) (3,1) (3,0.75) (2.5,0.75) (2.25,0.5) (2.25,0) (2,0) (1.5,0) (1,0) (0.75,0) (0.75,0.5) (0.5,0.75) (0,0.75) (0,1) (0,1.5) (0,2) (0,2.25) (0.5,2.25) (0.75,2.5) (0.75,3) (1,3) (1.5,3) (2,3) (2.25,3) (2.25,2.5) } { (1.625,1.5) (1.5,1.625) (1.375,1.5) (1.5,1.375) (1.625,1.5) } } ' actual '{ { (1.625,1.5) (1.5,1.625) (1.375,1.5) (1.5,1.375) (1.625,1.5) } } { { (2.25,2.5) (2.5,2.25) (3,2.25) (3,2) (3,1.5) (3,1) (3,0.75) (2.5,0.75) (2.25,0.5) (2.25,0) (2,0) (1.5,0) (1,0) (0.75,0) (0.75,0.5) (0.5,0.75) (0,0.75) (0,1) (0,1.5) (0,2) (0,2.25) (0.5,2.25) (0.75,2.5) (0.75,3) (1,3) (1.5,3) (2,3) (2.25,3) (2.25,2.5) } } '"

---> group: MarchingSquares:Polygon, test: test<5>
     problem: assertion failed
     failed assertion: "Polygon #1: expected '{ { (4.5,5) (5,5) (5,4.5) (5,4) (5,3.5) (5,3) (5,2.5) (5,2) (5,1.5) (5,1) (5,0.5) (5,0) (4.5,0) (4,0) (3.5,0) (3,0) (2.5,0) (2,0) (1.5,0) (1,0) (0.5,0) (0,0) (0,0.5) (0,1) (0,1.5) (0,2) (0,2.5) (0,3) (0,3.5) (0,4) (0,4.5) (0,5) (0.5,5) (1,5) (1.5,5) (2,5) (2.5,5) (3,5) (3.5,5) (4,5) (4.5,5) } { (4,3.5) (3.5,4) (2.5,4) (1.5,4) (1,3.5) (1,2.5) (1,1.5) (1.5,1) (2.5,1) (3.5,1) (4,1.5) (4,2.5) (4,3.5) } } ' actual '{ { (4,3.5) (3.5,4) (2.5,4) (1.5,4) (1,3.5) (1,2.5) (1,1.5) (1.5,1) (2.5,1) (3.5,1) (4,1.5) (4,2.5) (4,3.5) } } { { (4.5,5) (5,5) (5,4.5) (5,4) (5,3.5) (5,3) (5,2.5) (5,2) (5,1.5) (5,1) (5,0.5) (5,0) (4.5,0) (4,0) (3.5,0) (3,0) (2.5,0) (2,0) (1.5,0) (1,0) (0.5,0) (0,0) (0,0.5) (0,1) (0,1.5) (0,2) (0,2.5) (0,3) (0,3.5) (0,4) (0,4.5) (0,5) (0.5,5) (1,5) (1.5,5) (2,5) (2.5,5) (3,5) (3.5,5) (4,5) (4.5,5) } } '"
```

jratike80:
Out of curiosity, do you have metrics about the speedup?

gdal/alg/marching_squares/point.h
```diff
@@ -59,6 +59,16 @@ std::ostream & operator<<(std::ostream & o, const Point& p)
     return o;
 }
 
+// Test if a point is to the left, right or on an infinite line.
+// Returns greater than zero for p2 left of the line defined by p0 and p1,
+// 0 if p2 is on the line and less than if p2 is to the right of the line
+inline int
+isLeft(Point p0, Point p1, Point p2 )
+{
+    return static_cast<int>(( (p1.x - p0.x) * (p2.y - p0.y)
```
rouault, line 68: casting to int could potentially overflow. Cannot the return be a bool instead and test the floating point value againt > 0 ?

gdal/alg/marching_squares/point.h
```diff
@@ -59,6 +59,16 @@ std::ostream & operator<<(std::ostream & o, const Point& p)
     return o;
 }
 
+// Test if a point is to the left, right or on an infinite line.
+// Returns greater than zero for p2 left of the line defined by p0 and p1,
+// 0 if p2 is on the line and less than if p2 is to the right of the line
+inline int
+isLeft(Point p0, Point p1, Point p2 )
```
rouault, line 66:
```suggestion
isLeft(const Point& p0, const Point& p1, const Point& p2 )
```

gdal/alg/marching_squares/polygon_ring_appender.h
```diff
@@ -51,59 +52,46 @@ class PolygonRingAppender
...
-                        closestExterior = &other;
+            // Check if this is inside other using the winding number algorithm
+            auto checkPoint = this->points.front();
+            int windingNum = 0;
+            auto otherIter = other.points.begin();
+            // p1 and p2 define each segment of the ring other that will be tested
+            auto p1 = *otherIter;
+            do {
+                otherIter++;
+                auto p2 = *otherIter;
```
rouault, line 70: I believe there's a risk to dereference otherIter when its already at other.points.end(). The test at the end of this scope (`` } while (...)``) should probably be moved here

gdal/alg/marching_squares/polygon_ring_appender.h
```diff
@@ -114,8 +102,22 @@ class PolygonRingAppender
 #endif
     };
 
+    void processTree(std::vector<Ring> &tree, int level) {
```
rouault, line 105: const maybe ?
```suggestion
    void processTree(const std::vector<Ring> &tree, int level) {
```

pwicks86:
> Out of curiosity, do you have metrics about the speedup?

I happened to have some DEM data for the Mt. Shasta area on hand, so I mostly used that for testing while I was working on this since it's a relatively small data set. Before making this change, generating a polygon contour at 10 meter intervals for ned19_n41x50_w122x50_ca_mountshasta_2010.img took about 5 minutes and 30 seconds. With this change, the same contour could be generated in 20 seconds. On a much larger DEM that I needed for another project, a subset of a EU-DEM tile 34439x24561 in size, I was able to generate my needed contour in about 20 minutes compared to several hours at least(I don't have an exact time for the old version, I never let it run to completion)

pwicks86:
> On ./gdal_unit_test in autotest/cpp (make gdal_unit_test):

Ah, I had just been using `pytest` and hadn't seen failures related to this, hadn't realized that there was a separate unit test to be run. I've confirmed I can reproduce that test failure locally, will investigate.

gdal/alg/marching_squares/polygon_ring_appender.h
```diff
@@ -129,49 +131,69 @@ class PolygonRingAppender
...
+        }
+        if ( parentRing == nullptr ) {
+            // Then we never found anything and this must be a new top level ring for this level
+            levelRings.push_back(newRing);
+        } else {
+            // We found a valid parent, so we need to:
+            // 1. Find all the inner rings of the parent that are inside the new ring
+            auto trueGroupIt = std::partition(
+                parentRing->interiorRings.begin(),
+                parentRing->interiorRings.begin(),
```
pwicks86, line 171: Oops, this is a typo, should be `parentRing->interiorRings.end()`, which (I think?) turns this whole section into pretty much a no-op, as the partitioned result will always be of size 0. This might be causing my test failures, investigating.

pwicks86:
Updated the PR in response to the comments above and also fixed the test failure, which I think was a combination of a few things:

- `isLeft` returned bad values sometimes due to the cast. Fixed by removing the cast and just returning a bool instead.
- The logic for inserting rings into the tree wasn't quite right. There was the bug mentioned above and also the fact that I had not checked the case where a ring didn't fit inside any existing rings; in this case, we need to check the existing top level rings and see if any should be inside the new ring.

rouault:
With Visual Studio 2015:
```
c:\projects\gdal\gdal\alg\marching_squares\polygon_ring_appender.h(145): error C2039: 'back_inserter': is not a member of 'std' (compiling source file contour.cpp)
```
Probably due to a missing ```#include <iterator>``` ?

rouault:
Thanks !

hlthi:
> Out of curiosity, do you have metrics about the speedup?

In my case 750 second to 6 second.

Amazing @pwicks86  👏

pwicks86:
@hlthi happy to help :)

PR #2916, opened 2020-09-02
Remove all traces of GDAL and OGR ArcSDE drivers from source tree

rouault:


rouault:
Rationale is given here https://lists.osgeo.org/pipermail/gdal-dev/2020-September/052574.html . Those drivers require the ArcSDE Client SDK, which Esri stopped distributing it multiple releases ago, so is no longer supported by them, and unlikely to work with newer releases.
See this notes from another vendor ( https://community.safe.com/s/article/esri-legacy-arcsde-sde30-deprecation ) pointing to a Esri blog entry: https://community.esri.com/groups/technical-support/blog/2015/02/11/where-is-the-install-for-arcsde-10-3/

PR #2919, opened 2020-09-03
MRF: Finish create call when closing dataset

lucianpls:
Bug fix.
The Create call does write the files on disk if no data is every written before closing the newly created dataset.
The fix is to check that the files have been written in the dataset destructor, and force the write otherwise.

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:


lucianpls:
Obviously this is a bigger problem than I thought.

PR #2923, opened 2020-09-04
Mrf: Fixes for Create()

lucianpls:
## What does this PR do?

Fixes for MRF Create function.

In MRF, calling Create and closing the file without writing any data would not actually create the files.
In addition, there were a few cases where an exception raised when trying to create the file was not caught.

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:

### DIFF
--- a/gdal/frmts/mrf/marfa_dataset.cpp
+++ b/gdal/frmts/mrf/marfa_dataset.cpp
@@ -150,7 +150,11 @@ MRFDataset::~MRFDataset()
 
 {   // Make sure everything gets written
     if (eAccess != GA_ReadOnly && !bCrystalized)
-        Crystalize();
+        if (!Crystalize()) {
+            // Can't return from a destructor, just flag the error
+            CPLError(CE_Failure, CPLE_FileIO, "Error creating files");
+        }
+
     MRFDataset::FlushCache();
     MRFDataset::CloseDependentDatasets();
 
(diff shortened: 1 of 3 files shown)


gdal/frmts/mrf/marfa_dataset.cpp
```diff
@@ -149,8 +149,13 @@ int MRFDataset::CloseDependentDatasets()
 MRFDataset::~MRFDataset()
 
 {   // Make sure everything gets written
-    MRFDataset::FlushCache();
+    if (eAccess != GA_ReadOnly && !bCrystalized)
+        if (!Crystalize()) {
+            // Can't return from a destructor, just flag the error
```
rouault, line 154: this comment is a bit inaccurate. You can definitely have a return statement in a destructor (not saying this might be appropriate to put one here, if they are other memory cleanups to do)

lucianpls: I meant return something, like an error flag

rouault:
```suggestion
            // Can't return error code from a destructor, just emit the error
```

PR #2925, opened 2020-09-07
Update poppler build configuration requirement

dg0yt:
Poppler switched to a CMake-based build system a while ago. Even for CMake, the parameter was initially called `ENABLE_XPDF_HEADERS`, but this was changed to `ENABLE_UNSTABLE_API_ABI_HEADERS` in poppler 0.73.0 (Jan 2019).


rouault:
Thanks

PR #2928, opened 2020-09-09
treat "scheme://.." filenames as absolute

tbonfort:
VSI handlers can be attached to any prefix, however attaching them to "scheme://" prefixes (e.g. https://) to enable gdal to directly open urls without having to rewrite them will lead to them being considered as relative.

e.g. `gdalbuildvrt output.vrt http://file1.tif http://file2.tif` will create a vrt file who's `SourceFileName` will have `relativeToVrt` set to `1`.

This patch marks filenames containing `://` to be absolute.


rouault:
Backported to 3.1

PR #2929, opened 2020-09-10
FlatGeobuf: metadata extensions

bjornharrtell:
ref https://github.com/bjornharrtell/flatgeobuf/pull/75

### DIFF
--- a/gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
+++ b/gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
@@ -223,8 +223,24 @@ const std::vector<Offset<Column>> OGRFlatGeobufLayer::writeColumns(FlatBufferBui
         const auto field = m_poFeatureDefn->GetFieldDefn(i);
         const auto name = field->GetNameRef();
         const auto columnType = toColumnType(field->GetType(), field->GetSubType());
+        auto title = field->GetAlternativeNameRef();
+        if (EQUAL(title, ""))
+            title = nullptr;
+        const char *description = nullptr;
+        auto width = -1;
+        auto precision = -1;
+        auto scale = field->GetPrecision();
+        if (scale == 0)
+            scale = -1;
+        if (columnType == ColumnType::Float || columnType == ColumnType::Double)
+            precision = field->GetWidth();
+        else
+            width = field->GetWidth();
+        auto nullable = CPL_TO_BOOL(field->IsNullable());
+        auto unique = CPL_TO_BOOL(field->IsUnique());
+        auto primaryKey = false;
         //CPLDebugOnly("FlatGeobuf", "Create column %s (index %d)", name, i);
-        const auto column = CreateColumnDirect(fbb, name, columnType);
+        const auto column = CreateColumnDirect(fbb, name, columnType, title, description, width, precision, scale, nullable, unique, primaryKey);
         columns.push_back(column);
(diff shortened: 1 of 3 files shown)


gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
```diff
@@ -239,11 +253,27 @@ void OGRFlatGeobufLayer::readColumns()
...
+        const auto ogrType = toOGRFieldType(column->type(), eSubType);
+        OGRFieldDefn field(name, ogrType);
         field.SetSubType(eSubType);
+        field.SetAlternativeName(title);
+        if (width != 1 && type != ColumnType::Float && type != ColumnType::Double)
+            field.SetWidth(width);
+        if (precision != -1)
+            field.SetWidth(precision);
+        field.SetPrecision(scale != -1 ? scale : 0);
+        field.SetPrecision(scale != -1 ? scale : 0);
```
rouault, line 274: duplicated
```suggestion
```

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
```diff
@@ -239,11 +253,27 @@ void OGRFlatGeobufLayer::readColumns()
...
+        const auto nullable = column->nullable();
+        const auto unique = column->unique();
         OGRFieldSubType eSubType = OFSTNone;
-        const auto type = toOGRFieldType(column->type(), eSubType);
-        OGRFieldDefn field(name, type);
+        const auto ogrType = toOGRFieldType(column->type(), eSubType);
+        OGRFieldDefn field(name, ogrType);
         field.SetSubType(eSubType);
+        field.SetAlternativeName(title);
+        if (width != 1 && type != ColumnType::Float && type != ColumnType::Double)
```
rouault, line 269: there's a comparison to 1 against width, and below against -1 for precision. Intended ?

bjornharrtell: Good catch, it was a typo.

bjornharrtell:
The windows build warning is no fun to work around in my opinion (https://stackoverflow.com/questions/20919650/warning-c4800-int-forcing-value-to-bool-true-or-false-performance-warn). @rouault what's your take on it, work around or disable warning?

rouault:
> @rouault what's your take on it, work around or disable warning?

You can use the CPL_TO_BOOL(x) macro that does ``inline static bool CPL_TO_BOOL(int x) { return x != 0; }`` . Or just test against 0

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
```diff
@@ -223,8 +223,22 @@ const std::vector<Offset<Column>> OGRFlatGeobufLayer::writeColumns(FlatBufferBui
         const auto field = m_poFeatureDefn->GetFieldDefn(i);
         const auto name = field->GetNameRef();
         const auto columnType = toColumnType(field->GetType(), field->GetSubType());
+        const auto title = field->GetAlternativeNameRef();
```
rouault, line 226: GetAlternativeNameRef() never returns a nullptr, but an empty string if it is not set. I guess you should test for that, and if that's empty set the flatgeobuf field to nullptr

bjornharrtell:
@rouault this is now ready.

rouault:
We'd probably need a test for the new capabilities

bjornharrtell:
@rouault you are right of course, some basic tests are now added here.

PR #2935, opened 2020-09-12
Dataset rasterio related improvements

rouault:


sgillies:
It wasn't only VRT datasets that were affected but MEM datasets too? Thanks @rouault 🙏

rouault:
this pull request does not fix any bug actually. Just little things il noticed while investigating thé vrt issue. there was several lines of codes that coule replaced by a call to an exisitng method. and the dataset non nearest resampling code fell back to the band based method when overviews were found whereas it could have retried on the overview dataset. which is probably just a small perf improvements

PR #2937, opened 2020-09-12
Workaround issue with UTF-8 precomposed vs decomposed encodings on MacOS filesystems that affect sidecar file discovery (fixes #2903)

### LINKED ISSUE #2903: World files with non latin letters (on MacOS)

I expected GDAL to understand my world files but it seems it doesn't pick up filenames with å ä ö in the filenames.

This world file is picked up by for example gdalinfo:
file.tif
file.tfw

This not:
föle.tif
föle.tfw

## Operating system

MacOS

2.4.2

rouault:


gdal/gcore/gdal_misc.cpp
```diff
@@ -4070,6 +4071,43 @@ int GDALCanFileAcceptSidecarFile(const char* pszFilename)
...
+/*                   GDALCanReliablyUseSiblingFileList()                */
+/************************************************************************/
+
+/* Try to address https://github.com/OSGeo/gdal/issues/2903 */
+/* On Apple HFS+ filesystem, filenames are stored in a variant of UTF-8 NFD */
+/* (normalization form decomposed). The filesystem takes care of converting */
+/* precomposed form as often coming from user interface to this NFD variant */
+/* See https://stackoverflow.com/questions/6153345/different-utf8-encoding-in-filenames-os-x */
+/* And readdir() will return such NFD variant encoding. Consequently comparing */
+/* the user filename with ones with readdir() is not reliable */
```
dg0yt, line 4084: HFS+ has been replaced by Apple File System (APFS). 

> "APFS accepts only valid UTF-8 encoded filenames for creation, and preserves both case and normalization of the filename on disk in all variants. In macOS High Sierra, APFS is normalization-insensitive in both the case-insensitive and case-sensitive variants, using a hash-based native normalization scheme." 



> "APFS preserves the normalization of the filename and uses hashes of the normalized form of the filename to provide normalization insensitivity"

https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/APFS_Guide/FAQ/FAQ.html

dg0yt (commented):
The proposed change works for my all test files and environments on macOS (Terminal, Qt Creator).

gdal/gcore/gdal_misc.cpp
rouault: @dg0yt Hum, so with APFS, if one always use precomposed form, my understanding is that readdir() should also return it. So I suspect the issue might be that one of the .tif or .tfw file (or both) would have been created using one of the UTF-8 encoding (likely the decomposed one if using MacOS specific API), but the string passed to GDAL for opening would be with another one (likely the precomposed one)

PR #2942, opened 2020-09-17
Fixes related to GEOS 3.9 OverlayNG

rouault:
The only change in code is the TransformBeforeAntimeridianToWGS84() method.
Instead of intersecting the geometry with a multipolygon made of two parts,
one left of the antimeridian, one right of it, but which were at a very
close distance, which resulted in a LINESTRING EMPTY intersection with OverlayNG,
we compute the difference between our geometry and a thin polygon around the
antimeridian, which works for older GEOS too.

### DIFF
--- a/autotest/ogr/ogr_geojson.py
+++ b/autotest/ogr/ogr_geojson.py
@@ -2001,7 +2001,16 @@ def test_ogr_geojson_56():
 ]
 }
 """
-    assert json.loads(got) == json.loads(expected)
+
+    j_got = json.loads(got)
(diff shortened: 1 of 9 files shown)


rouault:
Note: @pramsey fixed the issue in GEOS master ( in https://github.com/libgeos/geos/commit/49ac5aa40404b634be3769375c70ee7c4466958d) that was triggered by the old way we cut at antimeridian . So we'd probably not need to change our code, but the new way is less convoluted, so let's just do that.

PR #2943, opened 2020-09-17
cog.rst: fix minor layout glitch

neteler:
## What does this PR do?

Improves layout of section in https://gdal.org/drivers/raster/cog.html#general-creation-options

Now: 

- SPARSE_OK=TRUE/FALSE ((GDAL >= 3.2): Should empty blocks be

With this PR:

- SPARSE_OK=TRUE/FALSE ((GDAL >= 3.2): Should empty blocks be omitted on disk?


rouault:
Thanks, but I think the real fix is to remove an extra space character at the beginning of all the lines starting from line 174 to 183

neteler:
Right - replaced by #2944

PR #2953, opened 2020-09-19
Add AbortSQL (initial proof of concept)

elpaso:
This PR adds a method to interrupt running queries
launched by ExecuteSQL.

### DIFF
--- a/autotest/ogr/ogr_gpkg.py
+++ b/autotest/ogr/ogr_gpkg.py
@@ -34,6 +34,8 @@ import os
 import struct
 import sys
 import pytest
+import time
+import threading
 
 from osgeo import gdal
 from osgeo import ogr
@@ -4146,3 +4148,38 @@ def test_ogr_gpkg_datetime_timezones():
     ds = None
 
     gdal.Unlink(filename)
+
+
+###############################################################################
+# Test AbortSQL
+
+def test_abort_sql():
+
+    filename = '/vsimem/test_ogr_gpkg_abort_sql.gpkg'
+    ds = ogr.GetDriverByName('GPKG').CreateDataSource(filename)
+    ds.CreateLayer('test')
+    ds = None
+
+    ds = ogr.Open(filename, update = 1)
+
+    def abortAfterDelay():
+        print("Aborting SQL...")
+        ds.AbortSQL()
+
+    t = threading.Timer(0.5, abortAfterDelay)
+    t.start()
+
+    start = time.time()
+
+    # Long running query
+    ds.ExecuteSQL("""
+    WITH RECURSIVE r(i) AS (
+        VALUES(0)
+        UNION ALL
+        SELECT i FROM r
+        LIMIT 10000000
+        )
+    SELECT i FROM r WHERE i = 1;""")
+
+    end = time.time()
+    assert int(end - start) < 1
--- a/autotest/ogr/ogr_gpkg.py
+++ b/autotest/ogr/ogr_gpkg.py
@@ -4160,11 +4160,11 @@ def test_abort_sql():
     ds.CreateLayer('test')
     ds = None
 
-    ds = ogr.Open(filename, update = 1)
+    ds = ogr.Open(filename, update=1)
 
     def abortAfterDelay():
         print("Aborting SQL...")
-        ds.AbortSQL()
+        assert ds.AbortSQL() == ogr.OGRERR_NONE
 
     t = threading.Timer(0.5, abortAfterDelay)
     t.start()
--- a/autotest/ogr/ogr_gpkg.py
+++ b/autotest/ogr/ogr_gpkg.py
@@ -4172,14 +4172,34 @@ def test_abort_sql():
     start = time.time()
 
     # Long running query
-    ds.ExecuteSQL("""
-    WITH RECURSIVE r(i) AS (
-        VALUES(0)
-        UNION ALL
-        SELECT i FROM r
-        LIMIT 10000000
-        )
-    SELECT i FROM r WHERE i = 1;""")
+    sql = """
+        WITH RECURSIVE r(i) AS (
+            VALUES(0)
+            UNION ALL
+            SELECT i FROM r
+            LIMIT 10000000
+            )
+        SELECT i FROM r WHERE i = 1;"""
+
+    ds.ExecuteSQL(sql)
+
+    end = time.time()
+    assert int(end - start) < 1
+
+    # Same test with a GDAL dataset
+    ds2 = gdal.OpenEx(filename, gdal.OF_VECTOR)
+
+    def abortAfterDelay2():
+        print("Aborting SQL...")
+        assert ds2.AbortSQL() == ogr.OGRERR_NONE
+
+    t = threading.Timer(0.5, abortAfterDelay2)
+    t.start()
+
+    start = time.time()
+
+    # Long running query
+    ds2.ExecuteSQL(sql)
 
     end = time.time()
     assert int(end - start) < 1
--- a/gdal/gcore/gdal_priv.h
+++ b/gdal/gcore/gdal_priv.h
@@ -813,6 +813,7 @@ private:
                                     OGRGeometry *poSpatialFilter,
                                     const char *pszDialect,
                                     swq_select_parse_options* poSelectParseOptions);
+    virtual void        AbortSQL( );
 //! @endcond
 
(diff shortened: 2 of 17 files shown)


autotest/ogr/ogr_gpkg.py
```diff
@@ -4146,3 +4148,38 @@ def test_ogr_gpkg_datetime_timezones():
...
+        print("Aborting SQL...")
+        ds.AbortSQL()
+
+    t = threading.Timer(0.5, abortAfterDelay)
+    t.start()
+
+    start = time.time()
+
+    # Long running query
+    ds.ExecuteSQL("""
```
rouault, line 4175: I wouldn't be surprised to see failure on some CI targets that ship with not so recent SQLite, so you'll probably have to conditionalize that test

gdal/gcore/gdaldataset.cpp
```diff
@@ -4578,6 +4578,30 @@ OGRLayerH GDALDatasetExecuteSQL( GDALDatasetH hDS,
...
+/************************************************************************/
+/*                        GDALDatasetAbortSQL()                       */
+/************************************************************************/
+
+/**
+ \brief Abort any SQL statement running in the data store.
+
+ This method is the same as the C++ method GDALDataset::AbortSQL()
+
+ @since GDAL 3.1.4
```
rouault, line 4591: We can't add this to a stable release, especially since it involves adding a new virtual method, which changes the C++ ABI that must remain stable in the 3.1 series.

```suggestion
 @since GDAL 3.2.0
```

gdal/gcore/gdaldataset.cpp
```diff
@@ -4578,6 +4578,30 @@ OGRLayerH GDALDatasetExecuteSQL( GDALDatasetH hDS,
             pszDialect));
 }
 
+
+/************************************************************************/
+/*                        GDALDatasetAbortSQL()                       */
+/************************************************************************/
+
+/**
+ \brief Abort any SQL statement running in the data store.
```
rouault, line 4587:
```suggestion
 \brief Abort any SQL statement running in the data store.
 
 This function can be safely called from any thread (pending that the dataset object is still alive). Driver implementations will make sure that it can be called in a thread-safe way.
 
 This might not be implemented by all drivers. At time of writing, only SQLite, GPKG and PG drivers implement it
```

to add to the doc of the C++ method as well

gdal/ogr/ogr_api.h
```diff
@@ -658,6 +658,7 @@ OGRLayerH CPL_DLL OGR_DS_CopyLayer( OGRDataSourceH, OGRLayerH, const char *,
 int    CPL_DLL OGR_DS_TestCapability( OGRDataSourceH, const char * );
 OGRLayerH CPL_DLL OGR_DS_ExecuteSQL( OGRDataSourceH, const char *,
                                      OGRGeometryH, const char * );
+void CPL_DLL OGR_DS_AbortSQL( OGRDataSourceH );
```
rouault, line 661: OGR_DS_ API is legacy only. We don't want to add new stuff into it

gdal/ogr/ogrsf_frmts/generic/ogrdatasource.cpp
```diff
@@ -231,6 +231,19 @@ OGRLayerH OGR_DS_ExecuteSQL( OGRDataSourceH hDS,
     return hLayer;
 }
 
+/************************************************************************/
+/*                         OGR_DS_AbortSQL()                            */
+/************************************************************************/
+
+void OGR_DS_AbortSQL( OGRDataSourceH hDS )
```
rouault, line 238: OGR_DS_ API is legacy only. We don't want to add new stuff into it

gdal/ogr/ogrsf_frmts/sqlite/ogrsqlitedatasource.cpp
```diff
@@ -3779,5 +3780,14 @@ const OGREnvelope* OGRSQLiteBaseDataSource::GetEnvelopeFromSQL(const CPLString&
 void OGRSQLiteBaseDataSource::SetEnvelopeForSQL(const CPLString& osSQL,
                                             const OGREnvelope& oEnvelope)
 {
-    oMapSQLEnvelope[osSQL] = oEnvelope;
+  oMapSQLEnvelope[osSQL] = oEnvelope;
+}
+
+/************************************************************************/
+/*                         AbortSQL()                          */
```
rouault, line 3787: some extra spaces needed to make it look nice

gdal/swig/include/ogr.i
```diff
@@ -853,6 +853,10 @@ public:
     return layer;
   }
 
+  void AbortSQL(){
```
rouault, line 856: to be rather added to the GDALDataset object.

You could leave that one (in addition to the one of GDALDatset), and just call GDALDatasetAbortSQL(), since a OGRDatasetH is a GDALDatasetH

gdal/gcore/gdaldataset.cpp
```diff
@@ -4578,6 +4578,30 @@ OGRLayerH GDALDatasetExecuteSQL( GDALDatasetH hDS,
...
+
+ This method is the same as the C++ method GDALDataset::AbortSQL()
+
+ @since GDAL 3.1.4
+
+ @param hDS the dataset handle.
+
+*/
+
+void GDALDatasetAbortSQL( GDALDatasetH hDS )
```
rouault, line 4597: wondering if we shouldn't make it return a CPLErr ?
The base dummy implementation would return a CE_Failure

elpaso: Yes, I alto thought about that, it's a good idea.

autotest/ogr/ogr_gpkg.py
elpaso: I think I need a little help here, ti looks like there are some unrelated failures, but let's wait for the new results after my last commits.

gdal/swig/include/ogr.i
elpaso: I'm not sure I get it right, please have a look to my last commits if they address this issue. Also, I'm not sure about how to test the GDALDatset path, it seems to me that the python tests are using the Datasource object and not the dataset.

rouault: > I'm not sure about how to test the GDALDatset path

you need to open the dataset with ds = gdal.OpenEx(filename, gdal.OF_VECTOR)

gdal/gcore/gdaldataset.cpp
```diff
@@ -6157,6 +6188,33 @@ GDALDataset::ExecuteSQL( const char *pszStatement,
...
+
+ This function can be safely called from any thread (pending that the dataset object is still alive). Driver implementations will make sure that it can be called in a thread-safe way.
+
+ This might not be implemented by all drivers. At time of writing, only SQLite, GPKG and PG drivers implement it
+
+ This method is the same as the C method GDALDatasetAbortSQL()
+
+ @since GDAL 3.2.0
+
+ @param hDS the dataset handle.
```
rouault, line 6207:
```suggestion
```

gdal/gcore/gdal_priv.h
```diff
@@ -813,6 +813,7 @@ class CPL_DLL GDALDataset : public GDALMajorObject
                                     OGRGeometry *poSpatialFilter,
                                     const char *pszDialect,
                                     swq_select_parse_options* poSelectParseOptions);
+    virtual OGRErr      AbortSQL( );
```
rouault, line 816: must be moved close below ReleaseResultSet() at line 793. Here it is a section with doxygen suppression

gdal/swig/include/ogr.i
elpaso: ... and add the test case to `ogr_pg.py`/`ogr_gpkg.py` or does it belong another test files? It seems to me that it should **not** be under `gdal/autotest/ogr`.

rouault: >  It seems to me that it should **not** be under `gdal/autotest/ogr`.

in theory yes, but I'm not sure a lifetime would be enough to erase the OGR vs GDAL dichotomy :-) The tests are fine where you've added them.

rouault:
@elpaso LGTM. Anything left on your side ?

elpaso:
@rouault no, I think I'm done for now. 
Thank you for the assistance!

PR #2955, opened 2020-09-21
VRT: add support for explicit virtual overviews 

rouault:
- Explicit virtual overviews are overviews that are reported on VRT rasterbands,
and rely only on the fact that the source(s) making those bands have themselves
overviews, but this is not enforced.
Such virtual overviews are generally not needed, unless in some situations, like
wrapping a VRT dataset to a lower resolution.
The VRTDataset.OverviewList XML element is added to declare such overviews.
GDALDataset::BuildOverviews() can also 'build' them when the VRT_VIRTUAL_OVERVIEWS
configuration option is set to YES.

-  gdalbuildvrt: add support for automatically declaring virtual overviews in very restricted situations.
All sources must have the same resolution, and this must be the one of
the VRT dataset itself.

### DIFF
--- a/autotest/gdrivers/vrtovr.py
+++ b/autotest/gdrivers/vrtovr.py
@@ -3,11 +3,11 @@
 # $Id$
 #
 # Project:  GDAL/OGR Test Suite
-# Purpose:  Test overviews in bands in VRT driver
+# Purpose:  Test overviews support in VRT driver
 # Author:   Even Rouault <even dot rouault at spatialys.com>
 #
 ###############################################################################
-# Copyright (c) 2010, Even Rouault <even dot rouault at spatialys.com>
+# Copyright (c) 2010, 2020, Even Rouault <even dot rouault at spatialys.com>
 #
 # Permission is hereby granted, free of charge, to any person obtaining a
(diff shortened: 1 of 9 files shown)


rouault:
CC @tbonfort

tbonfort:
Wow, thanks Even, this seems to work as I expected!
The documentation states that each underlying source should have the correct/same number of overviews actually present, but should they also have the same resolution/projection? Namely when using a vrt to mosaic a worldwide coverage, the underlying raster files would definitely have different resolutions depending on their latitude.

rouault:
> The documentation states that each underlying source should have the correct/same number of overviews actually present, but should they also have the same resolution/projection?

At the VRT driver level, you can declare virtual overviews as you wish, even if not all the sources have actually those overviews (of course performance will not be the one you could expect if you "lie" too much)

For gdalbuildvrt to be automatically able to declare/add virtual overviews, the current implementation does require that all sources have the same resolution (and projection but this is already a requirement). Some may potentially have extra overview levels.

So in your situation, you may need to manually set virtual overviews with
``gdaladdo your.vrt 2 4 ...  --config VRT_VIRTUAL_OVERVIEWS YES ``

PR #2961, opened 2020-09-23
[pdf] Don't write user units when generating GeoPDF from composition file

nyalldawson:
Doing so causes incorrect page sizes for the generated PDF file.

Refs https://github.com/qgis/QGIS/issues/33465 and discussion in https://lists.osgeo.org/pipermail/gdal-dev/2020-September/052681.html

### DIFF
--- a/gdal/frmts/pdf/pdfcreatefromcomposition.cpp
+++ b/gdal/frmts/pdf/pdfcreatefromcomposition.cpp
@@ -1012,8 +1012,6 @@ bool GDALPDFComposerWriter::GeneratePage(const CPLXMLNode* psPage)
                  "Missing or invalid Width and/or Height");
         return false;
     }
-    double dfUserUnit = CPLAtof(CPLGetXMLValue(psPage, "DPI",
-                        CPLSPrintf("%f", DEFAULT_DPI))) * USER_UNIT_IN_INCH;
 
     std::vector<GDALPDFObjectNum> anViewportIds;
     std::vector<GDALPDFObjectNum> anLGIDictIds;
@@ -1124,7 +1122,6 @@ bool GDALPDFComposerWriter::GeneratePage(const CPLXMLNode* psPage)
                                ->Add(0).Add(0).
                                  Add(dfWidthInUserUnit).
                                  Add(dfHeightInUserUnit)))
-             .Add("UserUnit", dfUserUnit)
              .Add("Contents", nContentId, 0)
              .Add("Resources", nResourcesId, 0);
 


rouault:
I'm not completely sure any change is needed. Presumably that if QGIS always use <DPI>72</DPI>, then having UserUnit = 1 will produce the same result as not having it. Except probably for the reader side of GDAL, that will use UserUnit = 1 will mean DPI = 72, which might not be desirable (GDAL defaults to DPI=150 on the reader side when no other hints are given), hence your patch could make sense.

I'm wondering if we couldn't have a compromise:
- if <DPI> is found in the XML, then keep the current behaviour, ie write UserUnit = DPI / 72
- if it is not specified, then go along your patch, ie. do not write UserUnit
What do you think ?

You'd have to modify either the input of tests to add <DPI>72</DPI> or update the expected results

jratike80:
Perhaps worth having a look at this https://trac.osgeo.org/gdal/ticket/5412 and think if the 14400 units limit could lead to some problems.

nyalldawson:
> Presumably that if QGIS always use 72

It doesn't though -- it uses the original dpi at which the component PDFs were generated (so usually 300 dpi by default).

I've tried replacing this in QGIS with a hardcoded value of 72, and while it fixes the PDF page size in the output, it comes with the side effect that re-loading the created PDF back into QGIS as a raster layer does so only at 72 dpi resolution (despite the fact that the resolution in the PDF is actually much higher).

nyalldawson:
Actually skipping writing the user unit property like this PR does doesn't fix the low-resolution raster reimport of the PDF either.

If there's no way to avoid this then I suggest we just close this PR, make QGIS write a constant 72 dpi tag in the composition, and live with the low-resolution raster layer side effect.

rouault:
> Actually skipping writing the user unit property like this PR does doesn't fix the low-resolution raster reimport of the PDF either.

Yeah, the GDAL PDF reader must fallback to  #define GDAL_DEFAULT_DPI 150.0 if UserUnit is not found. And presumably the driver shouldn't relate at all UserUnit with DPI. This was probably an error of mine doing so when wanting to round-trip CreateCopy() in GDAL. The DPI is another concept than UserUnit. For raster, it should be deduced by seeing how many pixels are displayed in a inch, which require inspecting objects, transforms, etc. This is complicated. The GDAL PDF driver has some logic to do this in very particular cases (GeoPDF exported by ArcGIS if I remember well and/other products), but in the general case, it will quickly fail. See the PDFDataset::GuessDPI() logic

Actually your pull request could be amended so that the `<DPI>` element is replaced by a `<UserUnit>` one, optional. If `<UserUnit>` is found, then it is written. Otherwise not.  data/pdfcomposition.xsd should be updated consequently.

For some backward compatibility, we might still want to read `<DPI>` and write `<UserUnit>` = DPI / 72, with some ``CPLError(CE_Warning, CPLE_AppDefined, "<DPI> has been deprecated by <UserUnit>");``, or perhaps just throw a warning if `<DPI>` is found and no do anything with it.

Hum, actually, regarding preservation of the DPI information, we might decide to add a GDAL-specific metadata element storing the DPI value, if there's not a standard one. Didn't find anything obvious skimming through https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdfs/PDF32000_2008.pdf , but this maybe be another topic...

PR #2964, opened 2020-09-23
gdalwmscache.cpp - CLEAN_THREAD_RUN_TIMEOUT exported as XML varialbe

idanmiara:
## What does this PR do?

This PR is to fix a regression from gdal 2.2.x before the Clean Thread was introduced.
The clean thread seems like a great idea, but if one wants an 'unlimited' or large enough cache size it causes troubles. 

It simply replaces CLEAN_THREAD_RUN_TIMEOUT const by m_nCleanThreadRunTimeout variable, which is changeable via CleanTimeout XML value.

I had some issues with the the wms drivers when using cache size of a few GBs.
I suspect that the clean thread takes up ages and slows down everything.
It also crushes when I close the dataset when the cache is large enough.
Thus this fix adds XML variable CleanTimeout (same default=120 seconds).
Personally, for my use case I would set it to 0 to disable the Clean thread but one can choose any other value.

Unfortunately I couldn't set up my gdal dev environment yet, So I couldn't actually compile and test this fix.
I'd be happy if someone could check it out, otherwise if it's too messed up I'd first try to build it myself and reopen this PR.

Thanks.

## Tasklist

 - [x] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments


gdal/frmts/wms/gdalwmscache.cpp
```diff
@@ -260,7 +274,10 @@ CPLErr GDALWMSCache::Insert(const char *pszKey, const CPLString &soFileName)
         if( result == CE_None )
         {
             // Start clean thread
-            if( !m_bIsCleanThreadRunning && time(nullptr) - m_nCleanThreadLastRunTime > CLEAN_THREAD_RUN_TIMEOUT)
+            int cleanThreadRunTimeout = m_poCache->GetCleanThreadRunTimeout();
+            if(  cleanThreadRunTimeout >= 0 &&
```
rouault, line 278:
```suggestion
            if(  cleanThreadRunTimeout > 0 &&
```

rouault (commented):
Could you also document <CleanTimeout> in doc/source/drivers/raster/wms.rst (close to the other options of the cache) ?

idanmiara:
> doc/source/drivers/raster/wms.rst

Done. Thanks for the input!

idanmiara:
> Could you also document in doc/source/drivers/raster/wms.rst (close to the other options of the cache) ?

Done, Thanks!

PR #2969, opened 2020-09-25
Add OGCAPI (tiles/maps) (raster/vector) experimental driver

rouault:


rouault:
CC @jerstlouis Here are the GDAL work related to OGCAPI / MOAW. I suggest we put it as an experimental driver in GDAL master. The doc mentions its experimental status

jerstlouis:
@rouault Cool! Thank you so much for all that work.

At the moment which APIs does it support? Maps, Coverages & Tiles (of Maps & Coverages)?

Would Features / Vector Tiles have to be a completely separate driver for OGR / Vector data, or did you end up already some re-using some of the OGR WFS3 / Features API driver?

rouault:
See doc: https://github.com/OSGeo/gdal/blob/ea7e27cb335aea69decb920a492723aebf3eaf43/gdal/doc/source/drivers/raster/ogcapi.rst

> At the moment which APIs does it support? Maps, Coverages & Tiles (of Maps & Coverages)?

==> Maps & tiles. Not coverages

> Would Features / Vector Tiles have to be a completely separate driver for OGR / Vector data, or did you end up already re-using some of the OGR WFS3 / Features API driver?

Features + Tiles is in this new OGCAPI driver. It doesn't reuse the WFS3 / OAPIF driver because the logic is quite different. But it does use the GeoJSON or MVT drivers underneath to read tiles.

jerstlouis:
Vector tiles too!! Awesome.

PR #2972, opened 2020-09-25
aaigriddataset: Improve search for float data.

### LINKED ISSUE #2970: crash in frmts/aaigrid/aaigriddataset.cpp for msan with fread
This is a problem that likely can't be reproduced in any publicly available system, but my prosed fix is just a teeny bit faster than the existing code (and hopefully a little more obvious).

This happened on a custom linux with glibc in msan mode.  msan was not triggered with older glibc versions.

The issue stems from what fread is supposed to do when it reads 1 group of nChunkSize bytes, but there are less than a chunksize of bytes.  What state should the output buffer be left in?  One version of msan thinks that the rest of the buffer after the read data should be poisoned even if the buffer was initially calloc'ed.

```
==2306==WARNING: MemorySanitizer: use-of-uninitialized-value
    #0 0x56352e1a9f20 in AAIGDataset::CommonOpen(GDALOpenInfo*, GridFormat) gdal/frmts/aaigrid/aaigriddataset.cpp:798:31
    #1 0x56352e1a8d3d in AAIGDataset::Open(GDALOpenInfo*) gdal/frmts/aaigrid/aaigriddataset.cpp:406:12
    #2 0x56352e599ae0 in GDALOpenEx gdal/gcore/gdaldataset.cpp:2832:20
    #3 0x56352e598a34 in GDALOpen gdal/gcore/gdaldataset.cpp:2543:9
    #4 0x7eff4303156b in Open(char const*, GDALAccess) gdal/swig/python/extensions/gdal_wrap.cpp:6298:27
    #5 0x7eff430c28e9 in _wrap_Open(_object*, _object*) gdal/swig/python/extensions/gdal_wrap.cpp:30304:37
```

This is the minimal fix in aaigriddataset.cpp:

```
- CPL_IGNORE_RET_VAL(VSIFReadL(pabyChunk, nChunkSize, 1, poDS->fp));
+ CPL_IGNORE_RET_VAL(VSIFReadL(pabyChunk, 1, nChunkSize, poDS->fp));
```

However, I think the following extra cleanup is just as good because fgets is getting called for each byte anyway.  The original code:

```c++
        // Scan for dot in subsequent chunks of data.
        while( !VSIFEofL(poDS->fp) )
        {
            CPL_IGNORE_RET_VAL(VSIFReadL(pabyChunk, nChunkSize, 1, poDS->fp));

            for( int i = 0; i < static_cast<int>(nChunkSize); i++)
            {
                GByte ch = pabyChunk[i];
                if (ch == '.' || ch == ',' || ch == 'e' || ch == 'E')
                {
                    poDS->eDataType = GDT_Float32;
                    break;
                }
            }
        }
```

Could be done as:

```c++
        // Scan for dot in subsequent chunks of data.
        while( !VSIFEofL(poDS->fp) )
        {
            const size_t nLen = VSIFReadL(pabyChunk, 1, nChunkSize, poDS->fp));

            for( size_t i = 0; i < nLen); i++)
            {
                const GByte ch = pabyChunk[i];
                if (ch == '.' || ch == ',' || ch == 'e' || ch == 'E')
                {
                    poDS->eDataType = GDT_Float32;
                    break;
                }
            }
        }
```

schwehr:
Only scan the bytes actually read by VSIFReadL.

Fixes #2970

### DIFF
--- a/gdal/frmts/aaigrid/aaigriddataset.cpp
+++ b/gdal/frmts/aaigrid/aaigriddataset.cpp
@@ -1037,7 +1037,7 @@ GDALDataset *AAIGDataset::CommonOpen( GDALOpenInfo *poOpenInfo,
         // Scan for dot in subsequent chunks of data.
         while( !VSIFEofL(poDS->fp) )
         {
-            const size_t nLen = VSIFReadL(pabyChunk, nChunkSize, 1, poDS->fp);
+            const size_t nLen = VSIFReadL(pabyChunk, 1, nChunkSize, poDS->fp);
 
             for( size_t i = 0; i < nLen; i++)
             {
--- a/gdal/frmts/aaigrid/aaigriddataset.cpp
+++ b/gdal/frmts/aaigrid/aaigriddataset.cpp
@@ -1037,11 +1037,11 @@ GDALDataset *AAIGDataset::CommonOpen( GDALOpenInfo *poOpenInfo,
         // Scan for dot in subsequent chunks of data.
         while( !VSIFEofL(poDS->fp) )
         {
-            CPL_IGNORE_RET_VAL(VSIFReadL(pabyChunk, nChunkSize, 1, poDS->fp));
+            const size_t nLen = VSIFReadL(pabyChunk, nChunkSize, 1, poDS->fp);
 
-            for( int i = 0; i < static_cast<int>(nChunkSize); i++)
+            for( size_t i = 0; i < nLen; i++)
             {
-                GByte ch = pabyChunk[i];
+                const GByte ch = pabyChunk[i];
                 if (ch == '.' || ch == ',' || ch == 'e' || ch == 'E')
                 {
                     poDS->eDataType = GDT_Float32;


gdal/frmts/aaigrid/aaigriddataset.cpp
```diff
@@ -1037,11 +1037,11 @@ GDALDataset *AAIGDataset::CommonOpen( GDALOpenInfo *poOpenInfo,
         // Scan for dot in subsequent chunks of data.
         while( !VSIFEofL(poDS->fp) )
         {
-            CPL_IGNORE_RET_VAL(VSIFReadL(pabyChunk, nChunkSize, 1, poDS->fp));
+            const size_t nLen = VSIFReadL(pabyChunk, nChunkSize, 1, poDS->fp);
```
rouault, line 1040: Should likely be to fix issue raised by autotest:
```suggestion
            const size_t nLen = VSIFReadL(pabyChunk, 1, nChunkSize, poDS->fp);
```

gdal/frmts/aaigrid/aaigriddataset.cpp
```diff
@@ -1037,7 +1037,7 @@ GDALDataset *AAIGDataset::CommonOpen( GDALOpenInfo *poOpenInfo,
         // Scan for dot in subsequent chunks of data.
         while( !VSIFEofL(poDS->fp) )
         {
-            const size_t nLen = VSIFReadL(pabyChunk, nChunkSize, 1, poDS->fp);
+            const size_t nLen = VSIFReadL(pabyChunk, 1, nChunkSize, poDS->fp);
```
schwehr, line 1040: Thanks!  That was the whole point of the change... can't believe I messed up that part.

Issue #2967, opened 2020-09-25
Java binding in docker image


exotfboy:
There are 5 docker images provided. But none of them contains java building.

Is it possible to add this to the plan?

jratike80:
When you opened this ticket you should have seen this notice

```
<!--
IMPORTANT: Do NOT use GitHub to post any questions or support requests!
           They will be closed immediately and ignored.

Questions should go to the gdal-dev mailing list at
https://lists.osgeo.org/mailman/listinfo/gdal-dev
or other support forums. GitHub issues are for bug
reports and suggestions for new features.
```

exotfboy:
Yes, I saw that, and I also saw this:

>GitHub issues are for bug reports and **suggestions for new features**.

jratike80:
Right, I was reading your message as a question.
Can you suggest in which image that would fit best?

exotfboy:
I am not sure. Different application may have different requirements.

Maybe an extra tag for each like `osgeo/gdal:alpine-small-jdk8:latest`?

While maybe people need different jdk version.  OMG :(

PR #2974, opened 2020-09-26
Docker/ubuntu-full: Add support for FileGDB and MDB drivers

apjoseph:
1. Adds support for FileGDB drivers
    *  Ubuntu-Full isn't really full without write support for one of the most common GIS formats.

2. Adds support for MDB driver
    *  You still occasionally encounter personal geodatabases in the wild

3. Updates OpenJDK 8 to OpenJDK 11
    *  Current LTS

### DIFF
--- a/gdal/docker/ubuntu-full/bh-gdal.sh
+++ b/gdal/docker/ubuntu-full/bh-gdal.sh
@@ -16,6 +16,8 @@ mkdir gdal
 wget -q "https://github.com/OSGeo/gdal/archive/${GDAL_VERSION}.tar.gz" \
     -O - | tar xz -C gdal --strip-components=1
 
+
+
 (
     cd gdal/gdal
     if test "${RSYNC_REMOTE:-}" != ""; then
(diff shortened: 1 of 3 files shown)


rouault:
I'm a bit hesitant to include FileGDB in the default ubuntu-full image. Even if they distribute it as Apache 2 licensed, it remains for practical purposed closed source software. Perhaps create a dedicated ubuntu-full-nonfree image that would share the Dockerfile of ubuntu-full but with a dedicated build argument ?
@rcoup thoughts ?

rcoup:
@rouault that's exactly what I was thinking as well. It's near impossible to debug the binary blobs. 

Nonfree makes it more obvious, and gives us a place for other similar drivers to live

apjoseph:
@ronault I added an WITH_FILEGDB build arg to the Dockerfile so that including the driver can be enabled/disabled on build. Also fixed #2967 by building bindings and setting a build arg for the java version. Let me know if this solves the short-term problem.

With respect to the non-free image, I personally like the idea of a default image that has all of the drivers that are unproblematic ( open source, permissive licenses, no strange issues) built by default -and then having every other driver or setting being individually configurable with a custom build arg

For instance in my case, I just need the FileGDB driver for write because when you send the average GIS professional a .gpkg they say "what the hell do I do with this? what is this extra .shm file that wasn't here before etc" and shapefiles have that obnoxious character limit -so the common denominator format that people will accept is a FileGDB. However, I don't really want/need the other problematic drivers -just that particular one. I think most people would be in the same boat for some driver or another. So I believe simply having a well documented README with driver name, build args and info about possible issues would be sufficient. 

Otherwise you'd end up having images like:
```
Ubuntu-20.04-NonFree-OpenJDK11-JDK 
Ubuntu-20.04-Non-Free-OpenJDK1-JRE 
Ubuntu:20.04-Non-Free-Open-JDK8-JDK-NoFileGDB
```
for every custom configuration someone needs.

gdal/docker/ubuntu-full/bh-gdal.sh
```diff
@@ -34,31 +36,44 @@ wget -q "https://github.com/OSGeo/gdal/archive/${GDAL_VERSION}.tar.gz" \
...
+    --with-mongocxxv3 \
+    --with-tiledb \
+    --with-crypto \
+    --with-java=/usr/lib/jvm/java-"$JAVA_VERSION"-openjdk-amd64 --with-jvm-lib=/usr/lib/jvm/java-"$JAVA_VERSION"-openjdk-amd64/lib/server --with-jvm-lib-add-rpath \
+    --with-mdb $GDAL_CONFIG_OPTS
+
+    make "-j$(nproc)"
+    make install DESTDIR="/build"
 
+    cd swig/java
```
rouault, line 73: I'm not sure this is enough. make install on the java bindings only install the libgdalalljni.so in ${prefix}/lib . But does nothing with the gdal.jar. I guess it should be copied in /usr/share/java if this is the convention ?

rouault:
> I added an WITH_FILEGDB build arg to the Dockerfile so that including the driver can be enabled/disabled on build

your rationale and solution make sense. They are a lot of configurations indeed that people might want and we could indeed end up in gazillions images.

gdal/docker/ubuntu-full/bh-gdal.sh
rouault: I'm discovering that /usr/share/java seems to be a conventional directory, at least on Ubuntu. Perhaps then just modify swig/java/GNUmakefile so that it install target copies gdal.jar into ${datarootdir}/java ?

rouault:
Looks good. Would you mind reworking the history to have first one commit "Java bindings: make 'install' target copy maven artifacts (gdal-X.Y.Z.jar) to ${datarootdir}/java" and then another one with the Docker specific changes  "Docker ubuntu-full: switch to JRE 1.11, install Java bindings (fixes #2967) and add a WITH_FILEGDB build argument"

apjoseph:
> Looks good. Would you mind reworking the history to have first one commit "Java bindings: make 'install' target copy maven artifacts (gdal-X.Y.Z.jar) to ${datarootdir}/java" and then another one with the Docker specific changes "Docker ubuntu-full: switch to JRE 1.11, install Java bindings (fixes #2967) and add a WITH_FILEGDB build argument"

done

prusswan:
btw, when was this last verified to be working? I just tried to build the latest master with WITH_FILEGDB=yes, it just leads to this error:
> /usr/bin/ld: warning: libFileGDBAPI.so, needed by /gdal/gdal/libgdal.so, not found (try using -rpath or -rpath-link)

So libFileGDBAPI.so is either not built or linked properly

rouault:
> btw, when was the FileGDB driver last verified to be working?

was indeed broken a few weeks ago. a7dd21943f fixes the breakage

PR #2980, opened 2020-09-29
VSI: add callback to explicitely list sibling files

tbonfort:
This patch allows vsi filesystems to return a list of sibling files. This can be used to explicitely disallow opening/querying for .aux or .ovr files without having to set GDAL_DISABLE_READDIR_ON_OPEN to EMPTY_DIR.
One might argue that this has nothing to do in the VSIFileSystem API, but the current lack of this function has already lead to an hack which has been cleaned up by this commit.

### DIFF
--- a/gdal/port/cpl_vsi.h
+++ b/gdal/port/cpl_vsi.h
@@ -312,6 +312,7 @@ GIntBig CPL_DLL CPLGetUsablePhysicalRAM(void);
 char CPL_DLL **VSIReadDir( const char * );
 char CPL_DLL **VSIReadDirRecursive( const char *pszPath );
 char CPL_DLL **VSIReadDirEx( const char *pszPath, int nMaxFiles );
+char CPL_DLL **VSISiblingFiles( const char *pszPath );
 
 /** Opaque type for a directory iterator */
 typedef struct VSIDIR VSIDIR;
@@ -455,6 +456,11 @@ typedef int            (*VSIFilesystemPluginRmdirCallback)         ( void *pUser
  */
 typedef char**         (*VSIFilesystemPluginReadDirCallback)       ( void *pUserData, const char *pszDirname, int nMaxFiles );
 /** 
+ * List related files. Optional 
+ * @since GDAL 3.2
+ */
+typedef char**         (*VSIFilesystemPluginSiblingFilesCallback)       ( void *pUserData, const char *pszDirname );
+/** 
  * Open a handle. Mandatory. Returns an opaque pointer that will be used in subsequent file I/O calls.
  * Should return null and/or set errno if the handle does not exist or the access mode is incorrect.
  * @since GDAL 3.0
@@ -543,6 +549,7 @@ typedef struct {
     VSIFilesystemPluginCloseCallback            close; /**< close handle  (rw) */
     size_t                                      nBufferSize; /**< buffer small reads (makes handler read only) */
     size_t                                      nCacheSize; /**< max mem to use per file when buffering */
+    VSIFilesystemPluginSiblingFilesCallback     sibling_files; /**< list related files*/
 /* 
     Callbacks are defined as a struct allocated by a call to VSIAllocFilesystemPluginCallbacksStruct
     in order to try to maintain ABI stability when eventually adding a new member.
--- a/gdal/port/cpl_vsi.h
+++ b/gdal/port/cpl_vsi.h
@@ -456,7 +456,11 @@ typedef int            (*VSIFilesystemPluginRmdirCallback)         ( void *pUser
  */
 typedef char**         (*VSIFilesystemPluginReadDirCallback)       ( void *pUserData, const char *pszDirname, int nMaxFiles );
 /** 
- * List related files. Optional 
+ * List related files. Must return NULL if unknown, or a list of relative filenames
+ * that can be opened along the main file. If no other file than pszFilename needs to
+ * be opened, return static_cast<char**> (calloc(1,sizeof(char*)));
+ * 
+ * Optional
  * @since GDAL 3.2
  */
 typedef char**         (*VSIFilesystemPluginSiblingFilesCallback)       ( void *pUserData, const char *pszDirname );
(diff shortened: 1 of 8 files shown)


rouault:
Wouldn't a new GDALOpenEx() opening flag, GDAL_OF_NO_SIDECAR, that would have the same effect of GDAL_DISABLE_READDIR_ON_OPEN=EMPTY_DIR do the job as well ? The VSI plugin callback would only make sense if there was indeed a fastest way of getting the sibblings than readdir.

tbonfort:
> Wouldn't a new GDALOpenEx() opening flag, GDAL_OF_NO_SIDECAR, that would have the same effect of GDAL_DISABLE_READDIR_ON_OPEN=EMPTY_DIR do the job as well ? The VSI plugin callback would only make sense if there was indeed a fastest way of getting the sibblings than readdir.

I don't think this does the same job. Semantically GDAL_DISABLE_READDIR_ON_OPEN could/should have been implemented as GDAL_OF_NO_SIDECAR , but it remains that the option still has to be explicitely asked by the end user. The aim of this PR is to be able to `GDALOpen("handler://path/to/file")` or `gdalinfo handler://path/to/file` and to be able to transparently ignore sidecar file side-effects when the handler knows that they are present or not.

tbonfort:
> The VSI plugin callback would only make sense if there was indeed a fastest way of getting the sibblings than readdir.

Perhaps the naming of the functions and the documentation should emphasize that the aim of this function is to avoid calling ReadDir or interating on `file.{exts}` if the list of available files is known beforehand.

rouault:
> Perhaps the naming of the functions and the documentation should emphasize that the aim of this function is to avoid calling ReadDir or interating on `file.{exts}` if the list of available files is known beforehand.

ok makes sense. The doc should also mention that for most implementations, including the base one for regular file systems, the function will return nullptr.

gdal/port/cpl_vsil_curl.cpp
```diff
@@ -4216,6 +4216,22 @@ char** VSICurlFilesystemHandler::ReadDirEx( const char *pszDirname,
...
+/*                             SiblingFiles()                           */
+/************************************************************************/
+
+char** VSICurlFilesystemHandler::SiblingFiles( const char *pszFilename )
+{
+    /* Small optimization to avoid unnecessary stat'ing from PAux or ENVI */
+    /* drivers. The MBTiles driver needs no companion file. */
+    if( EQUAL(CPLGetExtension( pszFilename ),"mbtiles") )
+    {
+        return static_cast<char**> (calloc(1,sizeof(char*)));
```
tbonfort, line 4229: the previous hack returned `CSLAddString( nullptr, CPLGetFilename(pszFilename) )` instead of an empty list. Does the list of sibling files also need to include the main file?

gdal/port/cpl_vsi.h
```diff
@@ -456,7 +456,11 @@ typedef int            (*VSIFilesystemPluginRmdirCallback)         ( void *pUser
  */
 typedef char**         (*VSIFilesystemPluginReadDirCallback)       ( void *pUserData, const char *pszDirname, int nMaxFiles );
 /** 
- * List related files. Optional 
+ * List related files. Must return NULL if unknown, or a list of relative filenames
+ * that can be opened along the main file. If no other file than pszFilename needs to
+ * be opened, return static_cast<char**> (calloc(1,sizeof(char*)));
```
rouault, line 461:
```suggestion
 * be opened, return static_cast<char**> (CPLCalloc(1,sizeof(char*)));
```

gdal/port/cpl_vsil.cpp
```diff
@@ -123,12 +123,30 @@ char **VSIReadDirEx( const char *pszPath, int nMaxFiles )
...
+ *
+ * @param pszFilename the path of a filename to inspect
+ * UTF-8 encoded.
+ * @return The list of entries, relative to the directory, of all sidecar
+ * files available or NULL if the list is not known. 
+ * Filenames are returned in UTF-8 encoding.
+ * Most implementations will return NULL, and a subsequent ReadDir will
+ * list all files available in the file's directory. This function will be
+ * overriden by VSI FilesystemHandlers that wish to force e.g. an empty list
+ * to avoid opening non-existant files on slow filesystems.
```
rouault, line 141:
```suggestion
 * to avoid opening non-existant files on slow filesystems. The return value shall be destroyed with CSLDestroy()
```

gdal/port/cpl_vsil.cpp
```diff
@@ -123,12 +123,30 @@ char **VSIReadDirEx( const char *pszPath, int nMaxFiles )
     return poFSHandler->ReadDirEx( pszPath, nMaxFiles );
 }
 
-char **VSISiblingFiles( const char *pszPath)
+/************************************************************************/
+/*                             VSISiblingFiles()                        */
+/************************************************************************/
+
+/**
+ * \brief Return related filenames
```
rouault, line 131:
```suggestion
 * \brief Return related filenames
  *
  * This function is essentially meant at being used by GDAL internals.
```

gdal/port/cpl_vsil_curl.cpp
```diff
@@ -4226,7 +4226,7 @@ char** VSICurlFilesystemHandler::SiblingFiles( const char *pszFilename )
     /* drivers. The MBTiles driver needs no companion file. */
     if( EQUAL(CPLGetExtension( pszFilename ),"mbtiles") )
     {
-        return CSLAddString( nullptr, CPLGetFilename(pszFilename) );
+        return static_cast<char**> (calloc(1,sizeof(char*)));
```
rouault, line 4229:
```suggestion
        return static_cast<char**> (CPLCalloc(1,sizeof(char*)));
```

gdal/port/cpl_vsil_curl.cpp
rouault: I *guess* the empty list shall be enough. Let's see if test are happy

rouault (changes requested):

gdal/port/cpl_vsil_curl.cpp
tbonfort: Local testing here showed no issues :crossed_fingers:

tbonfort:
remarking as draft as I'd like to refactor some uses of VSIReadDir to use VSISiblingFiles instead

tbonfort:
Leaving as is, I wasn't able to confidently remove some calls to VSIReadDir as I had hoped. Not sur why github says some changes are still requested, all seem to have been addressed. Thanks for the review @rouault .

rouault (approved):

PR #2414, opened 2020-04-16
WIP/DO_NOT_MERGE: NITF: Update to support SNIP TREs

rouault:
A new version of NITF will be available soon (definition file
attached from https://nsgreg.nga.mil/doc/view?i=4829). This file
has TREs with field value types of IEEE754 Float and UINT as
opposed to the traditional string parsing.
NGA.STND.0072_1.0_SNIP_Version_1.0_07-June-2019(1).pdf

Based on https://github.com/OSGeo/gdal/pull/2336 and supersedes it. Untested by me on such TREs since I don't have access to them ! @crossbearer can you give it a try ? Are there test files available ? Ideally we would need a (small) one for the GDAL autotest suite

PR #2983, opened 2020-09-29
NITF: Add BANDSB TRE from Spectral NITF Implementation Profile

matthew-baran:
## What does this PR do?
This PR adds the capability to read the BANDSB TRE, defined in the Spectral NITF Implementation Profile
STDI-0002-1 Appendix X.

In nitf_spec.xml, this TRE requires recursive variable search in <loop> blocks, and requires raw bits to be read and used as conditionals in <if> blocks.

https://github.com/OSGeo/gdal/pull/2414
https://github.com/OSGeo/gdal/pull/2973

## Tasklist

 - [X] nitf_spec.xml validates against nitf_spec.xsd
 - [X] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:

### DIFF
--- a/gdal/data/nitf_spec.xsd
+++ b/gdal/data/nitf_spec.xsd
@@ -130,7 +130,7 @@
                     <xs:enumeration value="string"/>
                     <xs:enumeration value="integer"/>
                     <xs:enumeration value="real"/>
-                    <xs:enumeration value="float_raw"/>
+                    <xs:enumeration value="IEEE754_Float32_BigEndian"/>
                     <xs:enumeration value="bitmask"/>
                 </xs:restriction>
             </xs:simpleType>
--- a/gdal/data/nitf_spec.xsd
+++ b/gdal/data/nitf_spec.xsd
@@ -107,7 +107,7 @@
         <xs:attribute name="cond">
             <xs:simpleType>
                 <xs:restriction base="xs:string">
-                    <xs:pattern value=".+[!]?=.*"/>
+                    <xs:pattern value=".+([!]?=|:).*"/>
                 </xs:restriction>
             </xs:simpleType>
         </xs:attribute>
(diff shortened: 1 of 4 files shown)


gdal/frmts/nitf/nitffile.c
```diff
@@ -2204,27 +2251,75 @@ static char** NITFGenericMetadataReadTREInternal(char **papszMD,
...
-                (*pnMDSize) ++;
-                papszTmp[0] = NULL;
-                CPLFree(papszTmp);
+                else if (strcmp(pszType, "bitmask") == 0)
+                {
+                    if( nLength <= 4 )
+                    {
+                        const size_t nBufferSize = 11;
+                        unsigned int nVal = 0;
+                        memcpy(&nVal, pachTRE + *pnTreOffset, nLength);
```
rouault, line 2280: I'm not smart enough to assert the correctness of this if nLength != 4. If nLength != 4 must be really supported, then I'd suggest to have a for loop on nLength to extract byte by byte and do the appropriate bit shifting to form nVal. This will be clearer / more portable

gdal/data/nitf_spec.xsd
```diff
@@ -130,6 +130,8 @@
                     <xs:enumeration value="string"/>
                     <xs:enumeration value="integer"/>
                     <xs:enumeration value="real"/>
+                    <xs:enumeration value="float_raw"/>
```
rouault, line 133: Was wondering if "IEEE754Float32BigEndian" (or something shorter) would be more descriptive ? But as the very least put that as a XML comment

gdal/frmts/nitf/nitffile.c
matthew-baran: Reverted this one for now.  nLength < 4 bytes will be needed for another TRE, so I'll take your implementation suggestion into another PR.  That TRE will also be a natural test of nLength < 4.

gdal/data/nitf_spec.xsd
matthew-baran: I went for the long name since the description is built into a functional element and thus easier to enforce than a comment.

PR #2985, opened 2020-09-29
Fix a large number of CLOSE_WAIT connections caused by server timeout…

sotex:
> Fix a large number of CLOSE_WAIT connections caused by server timeout, use cpl_auto_close to safely close libcurl handles.

When I was using S3 object storage, I found that a large number of `CLOSE_WAIT` TCP connections were generated after running for a period of time, and they would not be closed. I added two statements for setting related options to the libcurl handle. After testing, the situation has improved.

**Note:**
- https://curl.haxx.se/libcurl/c/CURLOPT_NOSIGNAL.html
- https://curl.haxx.se/libcurl/c/CURLOPT_FORBID_REUSE.html
- https://curl.haxx.se/mail/lib-2005-04/0061.html
- https://curl.haxx.se/mail/tracker-2011-05/0006.html
- https://curl.haxx.se/mail/lib-2016-07/0044.html
- https://curl.haxx.se/libcurl/c/CURLOPT_MAXCONNECTS.html

### DIFF
--- a/gdal/port/cpl_vsil_s3.cpp
+++ b/gdal/port/cpl_vsil_s3.cpp
@@ -28,6 +28,7 @@
 
 #include "cpl_atomic_ops.h"
 #include "cpl_port.h"
+#include "cpl_auto_close.h"
 #include "cpl_http.h"
 #include "cpl_md5.h"
 #include "cpl_minixml.h"
@@ -407,6 +408,11 @@ bool VSIDIRS3::IssueListDir()
         const CPLString osBaseURL(poS3HandleHelper->GetURL());
 
         CURL* hCurlHandle = curl_easy_init();
+        CPL_AUTO_CLOSE_WARP(hCurlHandle, curl_easy_cleanup);
+
+        curl_easy_setopt(hCurlHandle, CURLOPT_NOSIGNAL, 1L );
+        curl_easy_setopt(hCurlHandle, CURLOPT_FORBID_REUSE, 1L );
+
 
         if( !osBucket.empty() )
         {
@@ -454,7 +460,6 @@ bool VSIDIRS3::IssueListDir()
                 CPLDebug(poS3FS->GetDebugKey(), "%s",
                          requestHelper.sWriteFuncData.pBuffer
                          ? requestHelper.sWriteFuncData.pBuffer : "(null)");
-                curl_easy_cleanup(hCurlHandle);
                 return false;
             }
         }
@@ -467,12 +472,9 @@ bool VSIDIRS3::IssueListDir()
                                           requestHelper.sWriteFuncData.pBuffer,
                                           bIgnoreGlacier,
                                           bIsTruncated );
-
-            curl_easy_cleanup(hCurlHandle);
             return ret;
         }
-
-        curl_easy_cleanup(hCurlHandle);
+        
     }
 }
 
@@ -741,6 +743,11 @@ CPLString IVSIS3LikeFSHandler::InitiateMultipartUpload(
     {
         bRetry = false;
         CURL* hCurlHandle = curl_easy_init();
+        CPL_AUTO_CLOSE_WARP(hCurlHandle, curl_easy_cleanup);
+
+        curl_easy_setopt(hCurlHandle, CURLOPT_NOSIGNAL, 1L );
+        curl_easy_setopt(hCurlHandle, CURLOPT_FORBID_REUSE, 1L );
+
         poS3HandleHelper->AddQueryParameter("uploads", "");
         curl_easy_setopt(hCurlHandle, CURLOPT_CUSTOMREQUEST, "POST");
 
@@ -821,8 +828,6 @@ CPLString IVSIS3LikeFSHandler::InitiateMultipartUpload(
                     osFilename.c_str());
             }
         }
-
-        curl_easy_cleanup(hCurlHandle);
     }
     while( bRetry );
     return osUploadID;
@@ -882,6 +887,11 @@ CPLString IVSIS3LikeFSHandler::UploadPart(const CPLString& osFilename,
         bRetry = false;
 
         CURL* hCurlHandle = curl_easy_init();
+        CPL_AUTO_CLOSE_WARP(hCurlHandle, curl_easy_cleanup);
+
+        curl_easy_setopt(hCurlHandle, CURLOPT_NOSIGNAL, 1L );
+        curl_easy_setopt(hCurlHandle, CURLOPT_FORBID_REUSE, 1L );
+
         poS3HandleHelper->AddQueryParameter("partNumber",
                                             CPLSPrintf("%d", nPartNumber));
         poS3HandleHelper->AddQueryParameter("uploadId", osUploadID);
@@ -968,7 +978,6 @@ CPLString IVSIS3LikeFSHandler::UploadPart(const CPLString& osFilename,
             }
         }
 
-        curl_easy_cleanup(hCurlHandle);
     }
     while( bRetry );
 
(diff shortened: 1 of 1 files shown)


rouault:
I'm not completely convinced that this is a general purpose change and should be merged as it. It might impact performance in nominal cases

sotex:
> I'm not completely convinced that this is a general purpose change and should be merged as it. It might impact performance in nominal cases

Yes. I also think so. 

because, This is greatly affected by the s3 server, and many areas will not encounter this problem.

Let me think about whether there is a better way to deal with this situation. Maybe it is a feasible way to set a corresponding option through `CPLSetConfigOption` to control it.

rcoup:
Seems like it'd be better to figure out whether we're reusing/cleaning up connections properly, or whether we end up creating new connections all the time or aren't releasing handles somewhere. `CLOSE_WAIT` should be ~120s normally. (note the mailing list thread is from 2005...)

https://databricks.com/blog/2019/09/16/adventures-in-the-tcp-stack-performance-regressions-vulnerability-fixes.html ← what OS/kernel/etc versions are you spotting this on. No non-default networking settings (keepalive) in the client OS?

sotex:
> https://databricks.com/blog/2019/09/16/adventures-in-the-tcp-stack-performance-regressions-vulnerability-fixes.html ← what OS/kernel/etc versions are you spotting this on? No non-default networking settings (eg: client keepalives) in the client OS?

I asked the S3 storage service provider I used, and they used NGINX for forwarding. NGINX is configured with HTTP long connection options as:
```
keepalive_timeout 65s;
keepalive_requests 3000;
```
The operating system information I use is:
```bash
 lsb_release -a
LSB Version:    :core-4.1-amd64:core-4.1-noarch:cxx-4.1-amd64:cxx-4.1-noarch:desktop-4.1-amd64:desktop-4.1-noarch:languages-4.1-amd64:languages-4.1-noarch:printing-4.1-amd64:printing-4.1-noarch
Distributor ID: CentOS
Description:    CentOS Linux release 7.6.1810 (Core)
Release:        7.6.1810
Codename:       Core
```

图片

----

I see that the code inside does not add the relevant settings of the `KeepAlive` header, I am going to add relevant options and test it again.

The relevant code in the [AWS-SDK-CPP](https://github.com/aws/aws-sdk-cpp) is:

https://github.com/aws/aws-sdk-cpp/blob/1e4706dd26a2dd661771a5152010611f356d2498/aws-cpp-sdk-core/source/http/curl/CurlHandleContainer.cpp#L137-L153

rcoup:
@sotex possible to add a `uname -a` output too?

rcoup:
(fwiw the kernel `tcp_keepalive_*` settings match the Ubuntu system defaults)

sotex:
> I see that the code inside does not add the relevant settings of the `KeepAlive` header and timeout, I am going to add relevant options and test it again.
> 
> The relevant code in the [AWS-SDK-CPP](https://github.com/aws/aws-sdk-cpp) is:
> 
> https://github.com/aws/aws-sdk-cpp/blob/1e4706dd26a2dd661771a5152010611f356d2498/aws-cpp-sdk-core/source/http/curl/CurlHandleContainer.cpp#L137-L153

@rcoup I have used KeepAlive related options for testing, and the results are still the same. I checked the code and haven't found the specific reason yet.

Supplement: The libcurl version I used for compiling is `7.68.0`.

I am on vacation, I will continue to track this issue.

sotex:
> Seems like it'd be better to figure out whether we're reusing/cleaning up connections properly, or whether we end up creating new connections all the time or aren't releasing handles somewhere. `CLOSE_WAIT` should be ~120s normally. (note the mailing list thread is from 2005...)

@rcoup After testing, I probably found the cause of the problem.

Because `libcurl` caches connections, when the cached connection is closed by the server, the connection changes to the `CLOSE_WAIT` state, but `libcurl` will only process it when it needs to be reused. This leads to a large number of `CLOSE_WAIT` connections when the number of threads used is large and the global maximum cache connection number is not limited.

I think it is possible to add a `CPL_VSIL_CURL_MAXCONNECTS` option to `VSICURL_OPTIONS` to limit the number of cache connections of `easy handle/multi handle`.

因为 `libcurl` 的是会缓存连接的，缓存的连接被服务端关闭的时候，连接转为 `CLOSE_WAIT` 状态，但 `libcurl` 只有在需要重用的时候，才会去处理它。这在使用的线程数较多，且没有限制全局最大缓存连接数的时候，导致大量的 `CLOSE_WAIT` 连接。

我觉得可以在 `VSICURL_OPTIONS` 中添加一个 `CPL_VSIL_CURL_MAXCONNECTS` 的选项去限制 `easy handle/multi handle` 的缓存连接数。

**Relevant information:**
- [Re: Clear connection cache libcurl 7.44](https://curl.haxx.se/mail/lib-2016-07/0044.html)
- [Connection reuse](https://ec.haxx.se/libcurl/libcurl-connectionreuse)
- [`CURLOPT_MAXCONNECTS` explained](https://curl.haxx.se/libcurl/c/CURLOPT_MAXCONNECTS.html)
- [`CURLMOPT_MAXCONNECTS` explained](https://curl.haxx.se/libcurl/c/CURLMOPT_MAXCONNECTS.html)
- [` CURLMOPT_MAX_TOTAL_CONNECTIONS` explained](https://curl.haxx.se/libcurl/c/CURLMOPT_MAX_TOTAL_CONNECTIONS.html)
- [Multi-threaded transfers sharing a single connection pool](https://curl.haxx.se/libcurl/c/threaded-shared-conn.html)

lucianpls:
This seems wrong to me.
CLOSE_WAIT is a TCP connection state entered AFTER the connection is dropped.  Reusing the connections together with long keep-alive timeouts is the solution, not the problem.

@sotex Based on your description, I think the issue is not in GDAL, but in middle-layer nginx NOT REUSING the connections on the back end (to S3), which will generate many CLOSE_WAIT connections, one per request.  The ngnix back end (reverse proxy) is likely for the front (client facing) connection, not for the connection to S3.

sotex:
> This seems wrong to me.
> CLOSE_WAIT is a TCP connection state entered AFTER the connection is dropped. Reusing the connections together with long keep-alive timeouts is the solution, not the problem.

Yes, reusing connections is a solution that optimizes performance. This really shouldn't be a GDAL problem. But `libcurl` does not provide a separate cleanup interface. Only when it is reused will it clean up the connections that have entered `CLOSE_WAIT`, which will cause the file descriptor to be occupied.

> @sotex Based on your description, I think the issue is not in GDAL, but in middle-layer nginx NOT REUSING the connections on the back end (to S3), which will generate many CLOSE_WAIT connections, one per request. The ngnix keep-alive setting is likely only for the front (client facing) connection, not for the connection to S3.

nginx runs on the S3 storage provider's server, it is just a reserve proxy.

It is true that nginx actively closed the TCP connection with my local program, but libcurl does not provide an interface to clean up the connection in the CLOSE_WAIT state.

@lucianpls Thank you for your attention and answer.

lucianpls:
@sotex 
Then, this is not a problem in the GDAL code, and this pull request is not necessary, agree?
Ideally curl could have a "cleanup" function, but the existing lazy cleanup system works well.

sotex:
> Then, this is not a problem in the GDAL code, and this pull request is not necessary, agree?
> Ideally curl could have a "cleanup" function, but the existing lazy cleanup system works well.

Yes, it is best to wait for curl to add a cleanup function before GDAL adds corresponding processing.
The current lazy cleaning is excellent in general applications, but more operability is needed in server applications.

PR #2991, opened 2020-09-30
Add CPLHTTP[Push|Pop]FetchCallback() and CPLHTTPSetFetchCallback() callback mechanisms

rouault:
Those functions are to be used to install callbacks that will take care
of executing network requests submitted to CPLHTTPFetch/CPLHTTPFetchEx().

This is meant to allow QGIS for example to be able to plug its network layer
instead of using the Curl implementation.


nyalldawson:
Looks great!

PR #2999, opened 2020-10-02
DIMAP: fix loading when R1C1 tile is not present

tbonfort:
Some mosaic products may be missing the R1C1 tile. Use the first available `<Data_File>` in that case.

### DIFF
--- a/gdal/frmts/dimap/dimapdataset.cpp
+++ b/gdal/frmts/dimap/dimapdataset.cpp
@@ -1117,7 +1117,7 @@ int DIMAPDataset::ReadImageInformation2()
                 {
                     int nRow = atoi(pszR);
                     int nCol = atoi(pszC);
-                    if( nRow == 1 && nCol == 1 )
+                    if( (nRow == 1 && nCol == 1) || osImageDSFilename.empty() )
                         osImageDSFilename =
                             CPLFormCIFilename( osPath, pszHref, nullptr );
                     if( nRow > nRows ) nRows = nRow;
--- a/gdal/frmts/dimap/dimapdataset.cpp
+++ b/gdal/frmts/dimap/dimapdataset.cpp
@@ -1098,7 +1098,7 @@ int DIMAPDataset::ReadImageInformation2()
          </Data_Files>
     */
     std::map< std::pair<int,int>, CPLString > oMapRowColumnToName;
-    int nImageDSRow=0, nImageDSCol=0;
+    int nImageDSRow=1, nImageDSCol=1;
     if( psDataFiles )
     {
         int nRows = 1;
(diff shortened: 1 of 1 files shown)


gdal/frmts/dimap/dimapdataset.cpp
```diff
@@ -1301,7 +1305,15 @@ int DIMAPDataset::ReadImageInformation2()
         if( poImageDS->GetGeoTransform(adfGeoTransform) == CE_None &&
             !(adfGeoTransform[0] <= 1.5 && fabs(adfGeoTransform[3]) <= 1.5) )
         {
-            bHaveGeoTransform = TRUE;
+            if ( nImageDSCol == 1 && nImageDSRow == 1 ) {
+                bHaveGeoTransform = TRUE;
+            } else if ( adfGeoTransform[2] == 0 && adfGeoTransform[4] == 0 ) {
+                //if north-up images, fix up the origin if we did not get the
+                //geotransform from the top-left tile
+                adfGeoTransform[0] -= (nImageDSCol-1) * adfGeoTransform[1] * nTileWidth;
```
rouault, line 1313: Could probably be generalized with:
```suggestion
                adfGeoTransform[0] -= (nImageDSCol-1) * adfGeoTransform[1] * nTileWidth + (nImageDSRow-1) * adfGeoTransform[2] * nTileHeight;
```

gdal/frmts/dimap/dimapdataset.cpp
```diff
@@ -1301,7 +1305,15 @@ int DIMAPDataset::ReadImageInformation2()
...
+                adfGeoTransform[3] -= (nImageDSRow-1) * adfGeoTransform[5] * nTileHeight;
```
rouault, line 1314:
```suggestion
                adfGeoTransform[3] -= (nImageDSCol-1) * adfGeoTransform[4] * nTileWidth + (nImageDSRow-1) * adfGeoTransform[5] * nTileHeight;
```

gdal/frmts/dimap/dimapdataset.cpp
rouault: and just remove the if()  + else if()

tbonfort: Thanks for the suggestion, I was too lazy to find what needed to be done in case of rotation. @rouault you say "probably", does that mean this won't work all the time ?

rouault: > you say "probably", does that mean this won't work all the time ?

I'm reasonably confident this should work (this is the standard formula), but didn't test of course :-)

tbonfort: any error would only concern dimaps that didn't load previously, so let's go with this :)

PR #3003, opened 2020-10-02
gdal_calc: support multiple calc arguments to produce a multiband file

mmomtchev:
Support mulitple calc arguments to produce a multiband file
Currently, producing a multiband file requires producing separate files, then merging them either by constructing a VRT file, either by using gdal_merge
It stays compatible with the old way of calling gdal_calc from other python scripts by supporting both single calc values and lists of calc values

Example:
gdal_calc.py -A A.tif --A_band=1 -B.tif --B_band=2 -Z Z.tif --calc="(A+B)" --calc="(B+Z)" --calc="(Z/2)" --overwrite --outfile tmp/test_gdal_calc_py_8_1.tif

It is not compatible with --allBands - ie you have to either specify allBands, either have multiple calcs but not both

### DIFF
--- a/gdal/doc/source/programs/gdal_calc.rst
+++ b/gdal/doc/source/programs/gdal_calc.rst
@@ -38,7 +38,7 @@ performed.
 .. option:: --calc=expression
 
     Calculation in gdalnumeric syntax using ``+``, ``-``, ``/``, ``*``, or any numpy array functions (i.e. ``log10()``).
-    Multiple calc options can be listed to produce a multiband file.
+    Multiple ``--calc`` options can be listed to produce a multiband file (GDAL >= 3.2).
 
 .. option:: -A <filename>
 
--- a/gdal/doc/source/programs/gdal_calc.rst
+++ b/gdal/doc/source/programs/gdal_calc.rst
@@ -38,6 +38,7 @@ performed.
 .. option:: --calc=expression
 
     Calculation in gdalnumeric syntax using ``+``, ``-``, ``/``, ``*``, or any numpy array functions (i.e. ``log10()``).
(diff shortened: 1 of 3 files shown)


gdal/doc/source/programs/gdal_calc.rst
```diff
@@ -38,6 +38,7 @@ performed.
 .. option:: --calc=expression
 
     Calculation in gdalnumeric syntax using ``+``, ``-``, ``/``, ``*``, or any numpy array functions (i.e. ``log10()``).
+    Multiple calc options can be listed to produce a multiband file.
```
rouault, line 41:
```suggestion
    Multiple ``--calc`` options can be listed to produce a multiband file (GDAL >= 3.2).
```

gdal/swig/python/scripts/gdal_calc.py
```diff
@@ -434,7 +452,7 @@ def main():
     parser = OptionParser(usage)
 
     # define options
-    parser.add_option("--calc", dest="calc", help="calculation in gdalnumeric syntax using +-/* or any numpy array functions (i.e. log10())", metavar="expression")
+    parser.add_option("--calc", dest="calc", action="append", help="calculation in gdalnumeric syntax using +-/* or any numpy array functions (i.e. log10()), may appear multiple times to produce a multi-band file", metavar="expression")
```
rouault, line 455:
```suggestion
    parser.add_option("--calc", dest="calc", action="append", help="calculation in gdalnumeric syntax using +-/* or any numpy array functions (i.e. log10()). May appear multiple times to produce a multi-band file", metavar="expression")
```

PR #3016, opened 2020-10-05
gdal_calc.py - improvements

idanmiara:
## What does this PR do?

Multiple improvements for gdal_calc.py

general additions:
* AlphaList - now includes lowercase + uppercase alphabet
* GDALDataTypeNames - tuple of available output datatypes names
* DefaultNDVLookup - dictionary.keys() changed from string to GDALDataType; DefaultNDVLookup[GDT_Int16] = -32768 (changed to min int16, was -32767)
* Support Alpha inputs as a Sequence of inputs
* store_input_file - extended to support multiple inputs with the same alpha (which will make a list)
* accept dataset inputs in additions of filename inputs

* Support input files with different GeoTransforms and Dimensions - Implementation of this feature will be in another PR

new options:
* opts.NoDataValue=... - use opts.NoDataValue to indicate None datatype (opts.NoDataValue=None already indicates default value)
* opts.color_table - allow use input color_table
* opts.hideNodata - allows hiding nodatavalue
* opts.projectionCheck - should we check that all inputs have the same projection
* opts.extent - set custom output extent
* opts.user_namespace - add option for user_namespace, which can hold custom functions (default=None)
* opts.return_ds - option to keep the dataset open and return it (default=False)

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/autotest/pyscripts/test_gdal_calc.py
+++ b/autotest/pyscripts/test_gdal_calc.py
@@ -10,6 +10,7 @@
 ###############################################################################
 # Copyright (c) 2013, Even Rouault <even dot rouault @ spatialys.com>
 # Copyright (c) 2014, Etienne Tourigny <etourigny dot dev @ gmail dot com>
+# Copyright (c) 2020, Idan Miara <idan@miara.com>
 #
 # Permission is hereby granted, free of charge, to any person obtaining a
 # copy of this software and associated documentation files (the "Software"),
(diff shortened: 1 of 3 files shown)


rouault:
You'll need to rebase on top of https://github.com/OSGeo/gdal/pull/3003 that has just been merged into master

gdal/swig/python/scripts/gdal_calc.py
```diff
@@ -51,19 +51,25 @@
 import os.path
 import sys
 import shlex
+import string
+from collections import defaultdict
+from pathlib import Path
```
rouault, line 56: seems to be python3 specific. We still maintain python2 compat

gdal/swig/python/scripts/gdal_calc.py
```diff
@@ -439,21 +594,33 @@ def main():
...
         "options may be listed. See format specific documentation for legal "
         "creation options for each format.", metavar="option")
-    parser.add_option("--allBands", dest="allBands", default="", help="process all bands of given raster (A-Z)", metavar="[A-Z]")
+    parser.add_option("--allBands", dest="allBands", default="", help="process all bands of given raster (a-z, A-Z)", metavar="[a-z, A-Z]")
     parser.add_option("--overwrite", dest="overwrite", action="store_true", help="overwrite output file if it already exists")
     parser.add_option("--debug", dest="debug", action="store_true", help="print debugging information")
     parser.add_option("--quiet", dest="quiet", action="store_true", help="suppress progress messages")
     parser.add_option("--optfile", dest="optfile", metavar="optfile", help="Read the named file and substitute the contents into the command line options list.")
 
+    parser.add_option("--extent", dest="extent", type=int,
```
rouault, line 611: I'm not a specialist about parser.add_option(), but I'm pretty sure there must be a way to specify a proper enumeration

gdal/swig/python/scripts/gdal_calc.py
```diff
@@ -439,21 +594,33 @@ def main():
...
+                      help="how to treat different geotrasnforms [0=ignore|1=fail|2=union|3=intersect]")
+    # when extent don't agree: 0=ignore(check only dims)/1=fail (gt must also agree)/2=union/3=intersection/GeoRectangle
+    parser.add_option("--projectionCheck", dest="projectionCheck", action="store_true",
+                      help="check that all rasters share the same projection", metavar="value")
```
rouault, line 615: weird metavar name. I'm not sure if it's needed but it should reflect the name of the option

gdal/swig/python/scripts/gdal_calc.py
```diff
@@ -439,21 +594,33 @@ def main():
...
+    parser.add_option("--color_table", dest="color_table", help="color table file name", metavar="filename")
```
rouault, line 616: same here: weird metavar name

gdal/swig/python/scripts/gdal_calc.py
```diff
@@ -190,61 +274,97 @@ def doit(opts, args):
...
-            myOutB.SetNoDataValue(myOutNDV)
-            # write to band
-            myOutB = None
-
+            if myOutNDV is not None:
+                myOutB.SetNoDataValue(myOutNDV)
+            if opts.color_table:
+                # set color table and color interpretation
+                if isinstance(opts.color_table, (str, Path)):
+                    raise Exception('Error! reading a color table from a file is not supported yet')
```
rouault, line 359: this contradicts the documentation of the option

gdal/swig/python/scripts/gdal_calc.py
```diff
@@ -190,61 +274,97 @@ def doit(opts, args):
...
+        if not GeoTransformCheck:
+            GeoTransformCheck = myFiles[0].GetGeoTransform()
+        myOut.SetGeoTransform(GeoTransformCheck)
+        if not ProjectionCheck:
+            ProjectionCheck = myFiles[0].GetProjection()
+        myOut.SetProjection(ProjectionCheck)
 
         if opts.NoDataValue is not None:
-            myOutNDV = opts.NoDataValue
+            if opts.NoDataValue is ...:
```
rouault, line 344: pardon my Python ignorance, but what does "is ..."  mean ?

rouault (changes requested):
Just a partial review, but we would need tests for all those new options in autotest/pyscripts/test_gdal_calc.py and documentation in gdal/doc/source/programs/gdal_calc.rst

gdal/swig/python/scripts/gdal_calc.py
idanmiara: Ellipsis (...) is a singleton object, like None. 
It has a few use cases. Here I just use it to say "default value".
https://stackoverflow.com/questions/54405918/python-3-ellipsis-in-function-parameters

If backwards compatibility was not an issue, I would have done it differently:
opts.NoDataValue=None to indicate that there is no noDataValue.
opts.NoDataValue=... to indicate a default no data value
but since opts.NoDataValue=None already maps to the default value, I needed to choose a different value for noDataValue that would translate to noDataValue, a value that is not a valid noDataValue - thus I chose ellipsis.

gdal/swig/python/scripts/gdal_calc.py
idanmiara: True.
As is you can only provide a color_palette object and not as a filename.
I actually implemented this feature in my codebase in ColorPalette.read()
https://github.com/talos-gis/gdalos/blob/master/src/gdalos/gdalos_color.py
The thing is that depends on other this fairly large class and I thought it's too big for this PR. I also thought that there might be already a built in function in GDAL to load a color_table from file, but I couldn't find it.
I would be happy to include this module in GDAL, unless there is already a better way to do so.
If I was to include the ColorPalette class in a different file, where should I put it so I can import it here?

Same issue regarding handling multiple GeoTransforms, my implementation depends on:
https://github.com/talos-gis/gdalos/blob/master/src/gdalos/rectangle.py

gdal/swig/python/scripts/gdal_calc.py
idanmiara: I'll look into it. thanks.

gdal/swig/python/scripts/gdal_calc.py
idanmiara: hmm.. shouldn't Python2 be dead already?
I look for an alternative.

gdal/swig/python/scripts/gdal_calc.py
rouault: I'd suggest to drop the color palette thing for now, so as to get the changes in this pull request reasonably small and reviewable.

gdal/swig/python/scripts/gdal_calc.py
idanmiara: projectionCheck? I couldn't find a better name for an option that indicates whether to test if all the projections of the inputs are equal (the original code doesn't check).

gdal/swig/python/scripts/gdal_calc.py
idanmiara: color_table? what's wrong with that?

gdal/swig/python/scripts/gdal_calc.py
rouault: we haven't yet officially killed it. There would be changes to do on the continuous integration side since a number of them still use it... Perhaps something for GDAL 3.3

gdal/swig/python/scripts/gdal_calc.py
rouault: My point was about ``metavar="value"`` . Why value ?

gdal/swig/python/scripts/gdal_calc.py
rouault: My point was about ``metavar="filename"``

idanmiara:
> You'll need to rebase on top of #3003 that has just been merged into master

done

gdal/swig/python/scripts/gdal_calc.py
idanmiara: https://cpython-test-docs.readthedocs.io/en/latest/library/string.html?highlight=string#module-string
https://docs.python.org/2/library/collections.html#collections.defaultdict

were you referring to pathlib?
https://pypi.org/project/pathlib/
pathlib is included in python3.4 but apparently can be installed on python2. would it be acceptable or you prefer that I'll remove it?

gdal/swig/python/scripts/gdal_calc.py
idanmiara: OK, I removed it from the parser for now. still available to use by calling the function with a color_table object

gdal/swig/python/scripts/gdal_calc.py
idanmiara: OK, I removed it from the parser for now as it is not fully implemented yet and also I need to figure out how to use enum here. 
I will re-add it in a different PR.

gdal/swig/python/scripts/gdal_calc.py
rouault: I'd prefer if we removed any dependence to non-core Python modules. We don't have the "formalism" currently to express such dependencies

gdal/swig/python/scripts/gdal_calc.py
idanmiara: Apparently ellipsis was added in Python3.5.
Do you have a better idea for indicating not to set a nodata value? 
maybe: 
if not opts.NoDataValue and opts.NoDataValue is not None
so an empty string, or an empty tuple will indicate that?

gdal/swig/python/scripts/gdal_calc.py
idanmiara: ok, done.

gdal/swig/python/scripts/gdal_calc.py
rouault: I'm afraid I'm a bit lost (I'm only modestly familiar with this script). If the user has not defined NoDataValue, my understanding of the code was that at this point opts.NoDataValue should be None. For me either it is None or a numeric value. Can't think of anyone else.

idanmiara: right.
it's either None, which indicates the default value as per datatype
or a numeric value which indicates the requested nodatavalue.
I want to add an option for not setting a nodata value at all.
which, for I would have done it by setting it to None, but it's already has a meaning of a default nodata value.
thus I thought to allow opts.NoDataValue=[] to indicate not to set nodatavalue at all.

idanmiara:
> Just a partial review, but we would need tests for all those new options in autotest/pyscripts/test_gdal_calc.py and documentation in gdal/doc/source/programs/gdal_calc.rst

I'll gladly do that as soon as the other issues are resolved here.

gdal/swig/python/scripts/gdal_calc.py
idanmiara: I've committed this change, I hope it makes scene.

rouault: ah, I better understand now what you wanted to achieve. In other utilities we use the string 'none' (case insensitive. could be 'NONE') for such a situation ("gdal_translate -a_nodata none"). I'd suggest to use that convention too.

idanmiara: OK, I used "if opts.NoDataValue in ((), [], '', 'none', 'NONE')" to cover all conventions...

rouault:
@mmomtchev interested in reviewing this PR ?

mmomtchev:
@rouault I am looking at it, but I am a Python newbie, so I will probably miss a thing or two

idanmiara:
Thanks!
I'll try to write tests and docs today

mmomtchev:
What is the current status of the extent option? Not fully implemented?
My only suggestion will be to use an enum for all the coded values (like extent)

idanmiara:
> What is the current status of the extent option? Not fully implemented?
> My only suggestion will be to use an enum for all the coded values (like extent)

Hi,
I've implemented mixed extents with 5 options: 0: ignore; 1: fail; 2: union; 3: intersect; 4: given custom extent
options 0-1 are included in this PR
I've removed the other 3 options as they depend on other code that is too large for this PR. I hope to include it in my next PR.
Regarding using an Enum - I've committed now using consts, as Enum was introduced Python 3.4
I didn't use other coded values in this PR besides for the extent option.

@rouault suggested Python2 is still supported and using external packages is not allowed then I can't use Enum (Python 3.4, and backported to Python2.7 using a pypi package)
https://stackoverflow.com/questions/36932/how-can-i-represent-an-enum-in-python
I think it might be a good idea to include the backported pypi packages Enum and pathutil so we can be better prepared for Python3 migration.

color_table input in this PR is supported only as a color_table object.
also here, I've implemented also reading it from file, but also it's a large code and also I thought that there must be already a function to load color_table 
@rouault - how does gdaldem color-relief load the color_text_file ? I couldn't find a python code for function.

rouault:
> how does gdaldem color-relief load the color_text_file ?

That's an internal detail of apps/gdaldem_lib.cpp currently

gdal/swig/python/scripts/gdal_calc.py
```diff
@@ -641,9 +647,8 @@ def main():
     parser.add_option("--optfile", dest="optfile", metavar="optfile", help="Read the named file and substitute the contents into the command line options list.")
 
     # parser.add_option("--color_table", dest="color_table", help="color table file name")
-    # when extent don't agree: 0=ignore(check only dims)/1=fail (gt must also agree)/2=union/3=intersection/GeoRectangle
-    # parser.add_option("--extent", dest="extent", type=int,
-    #                   help="how to treat different geotrasnforms [0=ignore|1=fail|2=union|3=intersect]")
+    parser.add_option("--extent", dest="extent", type=int,
```
rouault, line 650: can we use a string value (--extent=ignore / --extent=fail) and map to the constants afterwards ? The numeric values in the user interface are not very user friendly

idanmiara:
ok. done.

idanmiara:
I see.
looks like reusing GDALColorReliefParseColorFile here would be difficult.
How can I split my python code into multiple files here?
I see that gdalnumeric, for instance, is inside the osgeo package, but all
the python scripts under the swig dir are not in a package, so how can I
reuse them from other scripts?

rouault:
> looks like reusing GDALColorReliefParseColorFile here would be difficult. How can I split my python code into multiple files here? I see that gdalnumeric, for instance, is inside the osgeo package, but all the python scripts under the swig dir are not in a package, so how can I reuse them from other scripts?

the thing is that the scripts in swig/python/scripts are typically installed in /usr/bin so they must be standalone, or use the content of the "osgeo" module where the bindings are installed. If you wanted to use helper methods/scripts, I guess they should go in the "osgeo" module as well.

idanmiara:
So why not to put all the scripts inside the osgeo module so they can all
be reusable?

rouault:
> So why not to put all the scripts inside the osgeo module so they can all be reusable?

that could be an idea. I guess we would still need tiny wrapper scripts in swig/python/scripts. Perhaps something to discuss on the mailing list to get broader feedback

rouault:
So what is missing in this PR is tests + documentation

idanmiara:
Yes. I'll rebase it on the untils package once it's merge then I'll
complete it.

idanmiara:
> So what is missing in this PR is tests + documentation

Hi @rouault ,
Thanks so much for reviewing and merging my PRs!
I'm writing tests and docs for gdal_calc.py and I ran across this issue, would you please take a look?
I'm trying to load my output test raster in QGIS but it doesn't show. I've tried to add geotransform and srs and to also tried to add a grayscale palette and translate to RGB. But still it does not show on QGIS, but it does open up in any picture editor...

[The following output has the same band checksum as the raster test_gdal_calc_py_1_2.tif from test1](https://drive.google.com/file/d/1LKcQTsnqfW4Ouiqnd3WpOBtw5EeoIRUl/view?usp=sharing)
[Here is the rgb raster with gt and srs](https://drive.google.com/file/d/1bcD7FR-NLjC1xLcxb3g7nXIuRR4INv9t/view?usp=sharing)

Neither show up on QGIS.
Any suggestions?

rouault:
I see you raised the issue on QGIS ML. This is probably a QGIS issue that doesn't like rasters with north-up geotransform without SRS attached. But you can very likely fix that in gdal_calc.py itself.
The issue must stem from
```python
        myOut.SetGeoTransform(myFiles[0].GetGeoTransform())
```
Changing that to
``` python
        gt = myFiles[0].GetGeoTransform(can_return_null=True)
        if gt:
            myOut.SetGeoTransform(gt)
```
will avoid a dummy (0,1,0,0,0,1) transform to be set.
Cf:
```
$ python -c "from osgeo import gdal; ds = gdal.Open('../autotest/gcore/data/stefan_full_rgba.tif'); print(ds.GetGeoTransform())"
(0.0, 1.0, 0.0, 0.0, 0.0, 1.0)
```
vs
```
$ python -c "from osgeo import gdal; ds = gdal.Open('../autotest/gcore/data/stefan_full_rgba.tif'); print(ds.GetGeoTransform(can_return_null=True))"
None
```

idanmiara:
> I see you raised the issue on QGIS ML. This is probably a QGIS issue that doesn't like rasters with north-up geotransform without SRS attached. But you can very likely fix that in gdal_calc.py itself.
> The issue must stem from
> 
> ```python
>         myOut.SetGeoTransform(myFiles[0].GetGeoTransform())
> ```
> 
> Changing that to
> 
> ```python
>         gt = myFiles[0].GetGeoTransform(can_return_null=True)
>         if gt:
>             myOut.SetGeoTransform(gt)
> ```
> 
> will avoid a dummy (0,1,0,0,0,1) transform to be set.
> Cf:
> 
> ```
> $ python -c "from osgeo import gdal; ds = gdal.Open('../autotest/gcore/data/stefan_full_rgba.tif'); print(ds.GetGeoTransform())"
> (0.0, 1.0, 0.0, 0.0, 0.0, 1.0)
> ```
> 
> vs
> 
> ```
> $ python -c "from osgeo import gdal; ds = gdal.Open('../autotest/gcore/data/stefan_full_rgba.tif'); print(ds.GetGeoTransform(can_return_null=True))"
> None
> ```

Thanks!
[I've opened this issue with QGIS](https://github.com/qgis/QGIS/issues/39478)

idanmiara:
> I see you raised the issue on QGIS ML. This is probably a QGIS issue that doesn't like rasters with north-up geotransform without SRS attached. But you can very likely fix that in gdal_calc.py itself.
> The issue must stem from
> 
> ```python
>         myOut.SetGeoTransform(myFiles[0].GetGeoTransform())
> ```
> 
> Changing that to
> 
> ```python
>         gt = myFiles[0].GetGeoTransform(can_return_null=True)
>         if gt:
>             myOut.SetGeoTransform(gt)
> ```
> 
> will avoid a dummy (0,1,0,0,0,1) transform to be set.
> Cf:
> 
> ```
> $ python -c "from osgeo import gdal; ds = gdal.Open('../autotest/gcore/data/stefan_full_rgba.tif'); print(ds.GetGeoTransform())"
> (0.0, 1.0, 0.0, 0.0, 0.0, 1.0)
> ```
> 
> vs
> 
> ```
> $ python -c "from osgeo import gdal; ds = gdal.Open('../autotest/gcore/data/stefan_full_rgba.tif'); print(ds.GetGeoTransform(can_return_null=True))"
> None
> ```



> So what is missing in this PR is tests + documentation

OK, I've added documentation and some tests.
I've revised the testing module, I think it's much more nifty now :)

autotest/pyscripts/test_gdal_calc.py
```diff
@@ -32,326 +33,362 @@
...
+
+    for filename in temp_files:
         try:
             os.remove(filename)
         except OSError:
             pass
 
 
-
+if __name__ == '__main__':
```
rouault, line 382: unneeded. Just run ```pytest test_gdal_calc.py```

gdal/doc/source/programs/gdal_calc.rst
```diff
@@ -56,19 +56,34 @@ performed.
...
+
+    this option determinants how to handle rasters with different extents.
+    ignore (default) - only the dimensions of the rasters are compared. if the dimensions do not agree the operation would fail.
+    fail - the dimensions and the extent (bounds) of the rasters must agree, otherwise the operation would fail.
+    union, intersection and custom extent options are planned in future version.
+
+.. option:: --projectionCheck
+
+    By default, no projection checking will be performed.
+    By setting this option, if the projection is not the same for all bands then the operation would fail.
```
rouault, line 86:
```suggestion
    By setting this option, if the projection is not the same for all bands then the operation will fail.
```

gdal/doc/source/programs/gdal_calc.rst
```diff
@@ -98,6 +113,26 @@ performed.
...
+    A dictionary of custom functions or other names to be available for use in the Calc expression.
+
+.. option:: return_ds
+
+    If enabled, the output dataset would be returned from the function and not closed.
+
+.. option:: color_table
+
+    Allows to specify a ColorTable object (with Palette Index interpretation) to be used for the output raster.
+    Note: reading a color_table from a file is planned in future version.
```
rouault, line 134:
```suggestion
```

gdal/swig/python/osgeo/utils/gdal_calc.py
```diff
@@ -457,23 +665,32 @@ def main(argv):
...
         "creation options for each format.", metavar="option")
-    parser.add_option("--allBands", dest="allBands", default="", help="process all bands of given raster (A-Z)", metavar="[A-Z]")
+    parser.add_option("--allBands", dest="allBands", default="", help="process all bands of given raster (a-z, A-Z)", metavar="[a-z, A-Z]")
     parser.add_option("--overwrite", dest="overwrite", action="store_true", help="overwrite output file if it already exists")
     parser.add_option("--debug", dest="debug", action="store_true", help="print debugging information")
     parser.add_option("--quiet", dest="quiet", action="store_true", help="suppress progress messages")
     parser.add_option("--optfile", dest="optfile", metavar="optfile", help="Read the named file and substitute the contents into the command line options list.")
 
+    # parser.add_option("--color_table", dest="color_table", help="color table file name")
+    parser.add_option("--extent", dest="extent", type=str,
```
rouault, line 683: Proposing ignore and fail as value of ``choices`` would be cleaner. See https://github.com/OSGeo/PROJ-data/blob/master/grid_tools/ntv2_to_gtiff.py#L72 . This should be available even in the deprecated optparse we use according to https://docs.python.org/2/library/optparse.html#optparse.Option.choices

gdal/swig/python/osgeo/utils/gdal_calc.py
```diff
@@ -457,23 +665,32 @@ def main(argv):
 
     parser.add_option("--outfile", dest="outF", help="output file to generate or fill", metavar="filename")
     parser.add_option("--NoDataValue", dest="NoDataValue", type=float, help="output nodata value (default datatype specific value)", metavar="value")
-    parser.add_option("--type", dest="type", help="output datatype, must be one of %s" % list(DefaultNDVLookup.keys()), metavar="datatype")
+    parser.add_option("--hideNoData", dest="hideNoData", action="store_true", help="ignores the NoDataValues of the input rasters", metavar="value")
```
rouault, line 668:
```suggestion
    parser.add_option("--hideNoData", dest="hideNoData", action="store_true", help="ignores the NoDataValues of the input rasters")
```

gdal/swig/python/osgeo/utils/gdal_calc.py
```diff
@@ -457,23 +665,32 @@ def main(argv):
...
+    parser.add_option("--type", dest="type", help="output datatype, must be one of %s" % GDALDataTypeNames, metavar="datatype")
```
rouault, line 669: the choice could probably be implemented with a ``choices`` argument

gdal/swig/python/osgeo/utils/gdal_calc.py
```diff
@@ -129,57 +173,142 @@ def doit(opts, args):
...
+                                eps = 0.000001
+                                for i in (1, 5):
+                                    if abs(GeoTransformCheck[i] - myFileGeoTransform[i])>eps:
+                                        raise Exception(
+                                            "Error! Pixel size file %s %s are different from other files %s. Cannot proceed" %
+                                            (filename, myFileGeoTransform, GeoTransformCheck))
+                                for i in (2, 4):
+                                    if abs(GeoTransformCheck[i] - myFileGeoTransform[i])>eps:
+                                        raise Exception(
+                                            "Error! The rotation of file %s is %s, only 0 is accepted. Cannot proceed" %
```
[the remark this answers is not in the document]
idanmiara: > The error message is not consistent with the test. There's no reason to reject geotransforms with rotational terms. Why don't we just merge this test and the above one, and test the 6 coefficients, in a relative way
> 
> ```
> if abs(GeoTransformCheck[i] - myFileGeoTransform[i])>eps * abs(myFileGeoTransform[i]):
> ```

This test will be relevant for extent modes `union`, `intersection`, `custom extent`. so for this PR this code is actually unreachable.
I've implemented the 3 extra modes for unrotated files. thus test should have been to test the rotational parameters to 0.
I will revise this section and make it more future proof.

gdal/swig/python/osgeo/utils/gdal_calc.py
idanmiara: nice.
done.

gdal/swig/python/osgeo/utils/gdal_calc.py
idanmiara: done

idanmiara:
OK. I think it's ready.
not sure what's wrong with the MacOS build.

gdal/swig/python/osgeo/utils/gdal_calc.py
```diff
@@ -117,7 +162,37 @@ def GetOutputDriverFor(filename):
...
+GT_INCOMPATIBLE_PIXEL_SIZE = 1
+GT_INCOMPATIBLE_ROTATION = 2
+GT_NON_ZERO_ROTATION = 3
+
+
+def gt_diff(gt0, gt1, diff_support=(False, False, False, False), eps=0.0):
+    if isinstance(eps, numbers.Number):
+        eps = (eps, eps, eps)
+    if gt0 == gt1:
+        return -1  # same gt
```
rouault, line 179:
```suggestion
        return GT_SAME
```

gdal/swig/python/osgeo/utils/gdal_calc.py
```diff
@@ -117,7 +162,37 @@ def GetOutputDriverFor(filename):
...
+    same = (
+        eps[0] <= (abs(gt0[0] - gt1[0]) + abs(gt0[3] - gt1[3])),  # same offset
+        eps[1] <= (abs(gt0[1] - gt1[1]) + abs(gt0[5] - gt1[5])),  # same pixel size
+        eps[2] <= (abs(gt0[2] - gt1[2]) + abs(gt0[4] - gt1[4])),  # same rotation
+        gt0[2] == gt0[4] == gt1[2] == gt1[4] == 0)  # rotation zero
+    if same[0] and same[1] and same[2]:
+        return -2  # almost same gt
```
rouault, line 186:
```suggestion
        return GT_ALMOST_SAME
```

gdal/swig/python/osgeo/utils/gdal_calc.py
```diff
@@ -117,7 +162,37 @@ def GetOutputDriverFor(filename):
...
+    same = (
+        same[0] or diff_support[0],
+        same[1] or diff_support[1],
+        same[2] or diff_support[2],
+        same[3] or diff_support[3])
+    for i in same:
+        if not same[i]:
+            return i  # INCOMPATIBLE gt, returns why
+    return -3  # supported different gt
```
rouault, line 195:
```suggestion
    return GT_COMPATIBLE_DIFF
```

gdal/swig/python/osgeo/utils/gdal_calc.py
```diff
@@ -117,7 +162,37 @@ def GetOutputDriverFor(filename):
...
+        return -1  # same gt
+    same = (
+        eps[0] <= (abs(gt0[0] - gt1[0]) + abs(gt0[3] - gt1[3])),  # same offset
+        eps[1] <= (abs(gt0[1] - gt1[1]) + abs(gt0[5] - gt1[5])),  # same pixel size
+        eps[2] <= (abs(gt0[2] - gt1[2]) + abs(gt0[4] - gt1[4])),  # same rotation
+        gt0[2] == gt0[4] == gt1[2] == gt1[4] == 0)  # rotation zero
+    if same[0] and same[1] and same[2]:
+        return -2  # almost same gt
+    same = (
+        same[0] or diff_support[0],
```
rouault, line 188:
```suggestion
        (same[0] or diff_support[0], GT_INCOMPATIBLE_OFFSET),
```
and similarly for 3 following lines
And then in the for loop you can return the error code from the above tuples instead of relying on the values of the error code being 0,1,2,3

gdal/swig/python/osgeo/utils/gdal_calc.py
```diff
@@ -129,57 +204,139 @@ def doit(opts, args):
...
+    if not hasattr(opts, "color_table"):
+        opts.color_table = None
+    if not opts.extent:
+        opts.extent = EXTENT_IGNORE
+    else:
+        opts.extent = parse_extent(opts.extent)
+
+    equal_gt_eps = 0.000001
+    gt_diff_support = (False, False, False, False) if opts.extent == EXTENT_FAIL else (True, False, False, False)
+    gt_diff_error = (
```
rouault, line 230: I would rather make this is a dict, with explicit use of the GT_ constants to avoid relying on their value, which is fragile

gdal/swig/python/osgeo/utils/gdal_calc.py
idanmiara: oops... forgot to replace these :)

gdal/swig/python/osgeo/utils/gdal_calc.py
idanmiara: OK, I've taken now a similar approach.

gdal/swig/python/osgeo/utils/gdal_calc.py
idanmiara: done

rouault:
Python 2 compatibility issue:
```
    from osgeo.utils import gdal_calc
E     File "/usr/lib/python2.7/dist-packages/osgeo/utils/gdal_calc.py", line 175
E       def gt_diff(gt0, gt1, diff_support: dict, eps=0.0):
E                                         ^
E   SyntaxError: invalid syntax
```

idanmiara:
> Python 2 compatibility issue:
> 
> ```
>     from osgeo.utils import gdal_calc
> E     File "/usr/lib/python2.7/dist-packages/osgeo/utils/gdal_calc.py", line 175
> E       def gt_diff(gt0, gt1, diff_support: dict, eps=0.0):
> E                                         ^
> E   SyntaxError: invalid syntax
> ```

Hi,
I've fixed that. thanks.
I think that the 2 other test failures have nothing to do with this PR, right?

gdal/swig/python/osgeo/utils/gdal_calc.py
```diff
@@ -129,57 +199,144 @@ def doit(opts, args):
...
     global_namespace = dict([(key, getattr(gdalnumeric, key))
                              for key in dir(gdalnumeric) if not key.startswith('__')])
+    if opts.user_namespace:
+        global_namespace.update(opts.user_namespace)
 
     if not opts.calc:
         raise Exception("No calculation provided.")
     elif not opts.outF:
-        raise Exception("No output file provided.")
+        if not opts.return_ds:
```
rouault, line 208: do we really need that attribute ? Cannot we just always return the dataset ? People wanting to create a in-memory dataset can do that with an empty output dataset name (or whatever they want) and select the MEM driver

idanmiara: sounds right.

idanmiara:
rebased on master and squash

rouault:
I'm confused by the following line in the commit message
```
* opts.NoDataValue=... - use opts.NoDataValue to indicate None datatype (opts.NoDataValue=None already indicates default value)
```
It seems to me it should be "use opts.NoDataValue = 'none' to indicate that no nodata value" ? 
This possibility is not documented in the .rst

idanmiara:
> I'm confused by the following line in the commit message
> 
> ```
> * opts.NoDataValue=... - use opts.NoDataValue to indicate None datatype (opts.NoDataValue=None already indicates default value)
> ```
> 
> It seems to me it should be "use opts.NoDataValue = 'none' to indicate that no nodata value" ?
> This possibility is not documented in the .rst

Right. I've updated the rst further explaining NoDataValue related options.

gdal/doc/source/programs/gdal_calc.rst
```diff
@@ -54,21 +54,44 @@ performed.
...
+
+    .. note::
+        Using the Python API:
+        ``None`` value will indicate default datatype specific value.
+        ``'none'`` value will indicate not setting a NoDataValue.
+
+.. option:: --hideNoData
+
+    Ignores the input bands NoDataValue.
+    By default, the input bands NoDataVales are not participating in the calculation.
```
rouault, line 68:
```suggestion
    By default, the input bands NoDataValue are not participating in the calculation.
```

gdal/doc/source/programs/gdal_calc.rst
```diff
@@ -107,17 +149,23 @@ Add two files together:
 
     gdal_calc.py -A input1.tif -B input2.tif --outfile=result.tif --calc="A+B"
 
+Add a few more files together:
```
rouault, line 152: what does this syntax mean ? That a 3 band 'virtual' dataset is created with input1.tif, input2.tif and input3.tif ? I assume only the first band of those files is taken into account ?

gdal/doc/source/programs/gdal_calc.rst
```diff
@@ -54,21 +54,44 @@ performed.
 
 .. option:: --NoDataValue=<value>
 
-    Output nodata value (default datatype specific value).
+    Output NoDataValue (default datatype specific value).
+    To indicate not setting a NoDataValue use --NoDataValue=none.
```
rouault, line 58:
```suggestion
    To indicate not setting a NoDataValue use --NoDataValue=none (GDAL >= 3.3)
```

gdal/doc/source/programs/gdal_calc.rst
```diff
@@ -42,7 +42,7 @@ performed.
 
 .. option:: -A <filename>
 
-    Input gdal raster file, you can use any letter (A-Z).
+    Input gdal raster file, you can use any letter (a-z, A-Z).
```
rouault, line 45:
```suggestion
    Input gdal raster file, you can use any letter (a-z, A-Z).  (lower case supported since GDAL 3.3)
```

gdal/doc/source/programs/gdal_calc.rst
```diff
@@ -54,21 +54,44 @@ performed.
...
-    Output nodata value (default datatype specific value).
+    Output NoDataValue (default datatype specific value).
+    To indicate not setting a NoDataValue use --NoDataValue=none.
+
+    .. note::
+        Using the Python API:
+        ``None`` value will indicate default datatype specific value.
+        ``'none'`` value will indicate not setting a NoDataValue.
+
+.. option:: --hideNoData
```
rouault, line 65:
```suggestion
.. option:: --hideNoData

    ..versionadded:: 3.3
```

gdal/doc/source/programs/gdal_calc.rst
```diff
@@ -54,21 +54,44 @@ performed.
...
+
        Despite the datatype set using ``--type``, when doing intermediate aritmethic operations using operands of the
        same type, the operation result will honor the original datatype. This may lead into unexpected results in the final result.
-    
+
 .. option:: --format=<gdal_format>
 
     GDAL format for output file.
 
+.. option:: --extent=<option>
```
rouault, line 85:
```suggestion
.. option:: --extent=<option>

    ..versionadded:: 3.3
```

gdal/doc/source/programs/gdal_calc.rst
```diff
@@ -54,21 +54,44 @@ performed.
...
+
+    this option determinants how to handle rasters with different extents.
+    ``ignore`` (default) - only the dimensions of the rasters are compared. if the dimensions do not agree the operation will fail.
+    ``fail`` - the dimensions and the extent (bounds) of the rasters must agree, otherwise the operation will fail.
+
+.. option:: --projectionCheck
```
rouault, line 91:
```suggestion
.. option:: --projectionCheck

    ..versionadded:: 3.3
```

gdal/doc/source/programs/gdal_calc.rst
```diff
@@ -98,6 +121,25 @@ performed.
 
     Suppress progress messages.
 
+
+Python options
+--------------
```
rouault, line 126:
```suggestion
--------------

..versionadded:: 3.3
```

gdal/doc/source/programs/gdal_calc.rst
rouault, line 152:
```suggestion
Add a few more files together:

..versionadded:: 3.3
```

idanmiara: No. I've added more examples now to explain it better.

rouault: > No.

Hum, ok, maybe not a 3 band 'virtual' dataset, but a 3-dim numpy array with the slowest varying dimension being the content of each file ?

idanmiara:
Exactly

rouault:
> Exactly

ok, might perhaps be worth mentioning that in the .rst for . option:: -A   that it may be repeated starting with GDAL 3.3, and that the effect will be to create a 3-dim numpy array.
What happens if the result of the computation is not a 2D array, but a 3D one with more than one array in the slowest varying dimension ? Probably an error like currently. Could be worth mentioning in the doc that in that mode, the calculation formula must return a 2D array. (people could potentially expect that a multi-band output dataset would be generated)

Related to that, I was wondering about use cases where people have a multiband file, and they want for example to do operations accross bands, like averaging them. Currently they have to do -A foo.tif --A_band=1 -B foo.tif --B_band=2 --calc "(A+B)/2",
it would be convenient if they could just do -A foo.tif --calc "numpy.average(A, axis=0)".  Of course we cannot do that automatically as it would break backwards compatibility. We'd probably need some new switch for that (something not to be confused with the --allBands one: maybe --3DInputArrays ? Another idea was --A_band=all (otherwise --A_band would be incompatible of that mode) , but I guess we want 3d array for all input datasets, not just one). I'm not asking to implement that in this PR. Just trying to brainstorm a bit to see if it would be compatible of what is added per this PR. I guess so.

idanmiara:
I've updated the rst.

> people could potentially expect that a multi-band output dataset would be generated
I guess this could be indeed implemented in another PR replacing the current error output.

> Another idea was --A_band=all (otherwise --A_band would be incompatible of that mode) , but I guess we want 3d array for all input datasets, not just one). 

I like the idea of `--A_band=all`. 
I could think of cases where we wouldn't want that across all input datasets, like:

-A a.tif --A_band=all -B b.tif --B_band=2 --calc numpy.sum(A,axis=0)+B

> I'm not asking to implement that in this PR. Just trying to brainstorm a bit to see if it would be compatible of what is added per this PR. I guess so.
I think this PR is future proof for these use cases.

rouault:
> I like the idea of `--A_band=all`.
> I could think of cases where we wouldn't want that across all input datasets, like:
> 
> -A a.tif --A_band=all -B b.tif --B_band=2 --calc numpy.sum(A,axis=0)+B

Good point

rouault:
```
/__w/1/s/gdal/doc/source/programs/gdal_calc.rst:184: WARNING: Explicit markup ends without a blank line; unexpected unindent.
/__w/1/s/gdal/doc/source/programs/gdal_calc.rst:192: WARNING: Explicit markup ends without a blank line; unexpected unindent.
/__w/1/s/gdal/doc/source/programs/gdal_calc.rst:197: WARNING: Error in "code-block" directive:
```

rouault:
Would you mind rebasing on latest master ? I've finally fixed/worked around the OSX issue

idanmiara:
> Would you mind rebasing on latest master ? I've finally fixed/worked around the OSX issue

sure.

idanmiara:
> > Would you mind rebasing on latest master ? I've finally fixed/worked around the OSX issue
> 
> sure.

Yey! Finally the CI is happy.
Is there anything else to review?

rouault:
I believe we're almost done. There's just the following part of the commit message itself, which I find confusing
```
* opts.NoDataValue='none' - use opts.NoDataValue to indicate None datatype (opts.NoDataValue=None already indicates default value)
```

Shouldn't it be something like
```
* opts.NoDataValue='none' - to indicate not setting a nodatavalue on the output file, whereas opts.NoDataValue=None already indicates to use a default nodatavalue that depends on the output data type
```

idanmiara:
Yes. I didn't pay enough attention for that while squashing the commits.

idanmiara:
@rouault  Thanks so much for the helpful review!

PR #3022, opened 2020-10-06
Improved AWS EC2 detection on Windows builds 

rouault:


dg0yt:
This PR added a hard dependency on `<atlbase.h>` from an ATL component of MSVC. Would it make sense to offer a way to opt out? I see vcpkg CI configurations (arm64, uwp) and users (e.g. https://github.com/microsoft/vcpkg/issues/23447) which lack this component.

Issue #3024, opened 2020-10-06
Dynamically set TLS/SSL certificate path


snowman2:
Follow on from: https://github.com/OSGeo/PROJ/issues/2320

cc: @sgillies 

Would be nice to add:

1. The environment variable `GDAL_CURL_CA_BUNDLE`
2. Methods to set the bundle path for PROJ and GDAL.

Thoughts?

rouault:
Is it from a rasterio perspective ?
Why not just defining the CURL_CA_BUNDLE env var that must work with GDAL and PROJ now ? I don't really a need to set them for a particular "instance" of GDAL and PROJ. In PROJ this is doable since there is a concept of a PROJ context. But in GDAL, there is no such thing as a context. So having a cpl_set_curl_ca_bundle() or setting CURL_CA_BUNDLE would have similar global effects on all GDAL users, either from rasterio or using the GDAL SWIG bindings ? For the CA bundle, I don't really anticipate an issue with rasterio settings affecting GDAL SWIG bindings. Someone must set it up to a good value, and everybody in the process benefits from it

snowman2:
> Why not just defining the CURL_CA_BUNDLE env var that must work with GDAL and PROJ now ? 

That is the current plan. However, if there are other programs using curl, having GDAL specific ways to set it up can prevent conflicts and confusion.

snowman2:
> cpl_set_curl_ca_bundle() 

That would be useful as well. It would be nice if it set the PROJ CA Bundle path as well if the method is available to set it.

rcoup:
what's a sane scenario where you'd want to replace the CA bundle _just_ for GDAL/PROJ but not anything else?

snowman2:
> what's a sane scenario where you'd want to replace the CA bundle just for GDAL/PROJ but not anything else?

As an application/library, automatically setting environment variables that might be used by other applications/libraries isn't a good neighborly thing to do and should be done carefully. Setting environment variables that are shared across applications are usually best left to the user to manage.

At the very least, you might have confused users wondering: 
- Why does installing package A change the behavior (fix/break) of package B?
- I thought I set the path of the CURL_CA_BUNDLE to `x` but when I install package A, it is now `y`, why did this happen?

[...]

PR #3025, opened 2020-10-07
Add GDAL_CURL_CA_BUNDLE environment variable

snowman2:
## What does this PR do?

This adds the `GDAL_CURL_CA_BUNDLE` environment variable for the use case where users want to set the path for the GDAL curl CA Bundle without worrying about modifying/impacting the main one used by other tools.

Part 1 of #3024

## Tasklist

 - [ ] Add test case(s)
 - [x] Add documentation
 - [x] Review
 - [x] Adjust for comments


rouault:
GDAL_CURL_CA_BUNDLE  (GDAL >= 3.2) should also be mentionned in https://github.com/OSGeo/gdal/blob/bea48acac51c301646795b18635c56fd85357fa6/gdal/port/cpl_http.cpp#L933

rouault:
I merging this mostly because of the symmetry for what we have done in PROJ, but I'm not sure we need more on GDAL side.

snowman2:
> I'm not sure we need more on GDAL side.

Thanks for merging this in 👍. So, based on this I am assuming a PR wouldn't be welcome with a new method to set the GDAL/PROJ CA Bundle path?

PR #3027, opened 2020-10-07
python/scripts - put all the main parts of the scripts under "if __name__ == '__main__':"

idanmiara:
## What does this PR do?

python/scripts - put all the main parts of the scripts under "if __name__ == '__main__':"

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments


rouault:
What is the value of doing this ?

idanmiara:
Preparing the ground for moving all the functions in the scripts to the
osgeo package for reusing.

The main part will remain in the scripts files and all the functions will
go to the osgeo package.

I'm going to do this separation in another PR.

rouault:
I really think you should bring the idea to the mailing list first to see if they are reactions and/or possibly prototype it on a single script to start with to see if there are no unforeseen obstacles.

idanmiara:
> I really think you should bring the idea to the mailing list first to see if they are reactions and/or possibly prototype it on a single script to start with to see if there are no unforeseen obstacles.

OK, thanks for your input.

PR #3029, opened 2020-10-07
OGCAPI: add support for Coverage

rouault:


rouault:
CC @jerstlouis

jerstlouis:
@rouault 

>  Default is AUTO, which uses coverage if available, and fallback first to tiles and finally maps otherwise.

Coverage tiles would likely generally be more efficient than coverage if supported because the server (and the client) can cache the same specific requests unlike subset which is not clearly pre-defined.

In our client implementation we default in this order:
1. Vector (Features) Tiles
2. Coverage Tiles
3. Features
4. Coverage
5. Map Tiles
6. Maps

We are implementing the user selection option as a combination of 4 individual flags that can be enabled or disabled (all enabled by default): tiles, maps, coverage and features. Then the above order will select one based on what is left enabled.

Would that make sense to prioritize Coverage Tiles in GDAL as well? (Is there this concept of map tiles, coverage tiles, vector tiles)? The difference being beween coverage tiles vs. map tiles being that you have raw values rather than a rendered map. The two blur though if the raw data is an (A)RGB image (and the `/coverage` and `/map` APIs could actually be quite similar in that case, especially if `subset` becomes Common).

The Layer's `dataType` of the tileset metadata ( http://docs.opengeospatial.org/per/19-082r1.html#img_metadataModel ) would indicate the type of data. We have agreed to update the Tiles API to adopt a metadata model to be defined in the 2DTMS standard (https://github.com/opengeospatial/2D-Tile-Matrix-Set/issues/10) which would specifies which it is.

Currently in our implementation we are identifying coverage tiles with `"dataType" : "values"` in the TileSet and list of TileSets, e.g. https://maps.ecere.com/ogcapi/collections/Daraa:Daraa_DTED/tiles?f=json
(and `"dataType" : "image"` for map tiles, e.g. https://maps.ecere.com/ogcapi/collections/Daraa:Daraa_DTED/map/tiles?f=json ).

rouault:
> Would that make sense to prioritize Coverage Tiles in GDAL as well?

That could be a further enhancement to consider, but I won't tackle this as part of this project. On the client side, GDAL uses the caching logic of the WMS driver when it accesses coverages, so that's already something.

Actually there's no real concept of "coverage tiles" in the spec ? You need links that define the domainSet + rangeType + links to the tiles API with GeoTIFF output, right ?

jerstlouis:
@rouault 
The Tiles API does specify that content can be any type of data, and we should ensure that coverage tiles are supported, e.g. making sure that the tileset metadata can specify a coverage data type.

There is a specific issue about whether Tiles API supports coverage tiles (https://github.com/opengeospatial/OGC-API-Tiles/issues/52).

In an API that offers both Tiles and Coverages for the same collection, you would automatically have the domainSet and rangeType links in addition to the tiles in a coverage format (e.g. GeoTIFF).
Our server implementation will have that as soon as we get the domainset and rangetype added :) We already have the GeoTIFF tiles.

Also I think the way to specify that a certain type of data is available as tiles would be a conformance class of Maps, Coverages and maybe Features, that says "this is also available as tiles" with a dependency on the appropriate Tiles API conformance class. That would be consistent with ensuring the domainset and rangetype are present for coverage tiles.

PR #3031, opened 2020-10-08
moving swig\python\scripts\*.py into the osgeo package for reusing as modules

idanmiara:
## What does this PR do?

The PR is about making the code in the python scripts more reusable so they can be imported as modules.
Since the scripts in swig/python/scripts are typically installed in /usr/bin so they must be standalone.
This PR would allow running or importing these scripts as modules to be imported from the "osgeo" package, where the bindings are installed. 
This will allow reusing their code in other projects without the need to copy them into a module first or simply just for easier use from the autotests (instead of using exec).

For each script I've created a "def main(argv)" function with the same signature. 
For most scripts that was fairly straightforward (many already do had a similar function).
a few scripts needed some more work - gdal_retile.py, gdal_merge.py - I needed also to replace all the global variables with local variables so it would be easier to reuse these functions.

In order to keep backward compatibility with the current mode of execution I've created tiny wrappers around each script, similar to the ones of gdalnumeric, but that also run the main function of the script.
I've created those tiny wrappers automatically using the provided auxiliary script (script_splitter.py).
So essentially, each wrapper would be like:
`
from osgeo.gdal import deprecation_warn
deprecation_warn('gdal_calc')
from osgeo.gdal_calc import *
if __name__ == '__main__':
    sys.exit(main(sys.argv))
`

I've sent a message to the mailing list, calling for comments.

I think it would be more helpful to review this PR with the Commits view instead of the Files Changed view, as the Files Changed view will show that most of the code of each script was deleted, while it actually was just moved to another location and instead there is just the mini wrapper.

https://github.com/OSGeo/gdal/pull/3016

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments


rouault:
- Perhaps the scripts should be moved to swig/python/osgeo/scripts to avoid swig/python/osgeo being too much crowded ?
- The script_splitter.py script is a one time conversion thing, right ? If so, there's no use to put it under git, unless I miss something
- Isn't there work to do in swig/python/setup.py to make sure the scripts are properly handled/packaged ? You should try "python setup.py sdist" and check that the dist/GDAL*.gz package includes the scripts. Would be good to test that package in a virtualenv or something similar

idanmiara:
> * Perhaps the scripts should be moved to swig/python/osgeo/scripts to avoid swig/python/osgeo being too much crowded ?
done.
> * The script_splitter.py script is a one time conversion thing, right ? If so, there's no use to put it under git, unless I miss something
OK, removed. I thought it might be reused for something else, but maybe it's not justified enough.
> * Isn't there work to do in swig/python/setup.py to make sure the scripts are properly handled/packaged ? You should try "python setup.py sdist" and check that the dist/GDAL*.gz package includes the scripts. Would be good to test that package in a virtualenv or something similar
I've updated setup.py to include the osgeo.scripts package.
Thanks!

rouault:
There are flake8 issues to address: https://github.com/OSGeo/gdal/pull/3031/checks?check_run_id=1225365716

idanmiara:
> There are flake8 issues to address: https://github.com/OSGeo/gdal/pull/3031/checks?check_run_id=1225365716
Yes, I'm working on it.

May I suggest a to name the sub-package "osgeo.aux" instead of osgeo.scripts, since they are actually modules, not scripts ?

rouault:
> May I suggest a to name the sub-package "osgeo.aux" instead of osgeo.scripts, since they are actually modules, not scripts ?

I'm not very good at naming. I thought at osgeo.utils too, but some might say that utils imply an implementation detail, not something for end users. Maybe ask on the mailing list

idanmiara:
> > May I suggest a to name the sub-package "osgeo.aux" instead of osgeo.scripts, since they are actually modules, not scripts ?
> 
> I'm not very good at naming. I thought at osgeo.utils too, but some might say that utils imply an implementation detail, not something for end users. Maybe ask on the mailing list

Yes, I think utils sounds good. 
I'm renaming it and will send this PR to the mailing list once flake8 is happy

gdal/swig/python/scripts/gcps2vec.py
```diff
@@ -1,9 +1,9 @@
 import sys
 # import osgeo.utils.gcps2vec as a convenience to use as a script
-from osgeo.utils.gcps2vec import *
+from osgeo.utils.gcps2vec import *  # noqa
```
rouault, line 3: why do we need this wildcard import ? Isn't the below import of main sufficient ?

idanmiara:
I see some tests added gdal_calc, for instance, to the path then imported
it as a module. It's ugly, and with this PR it's unnecessary. I've changed
this particular test to import the module from osgeo.utils. But I guess
that if we want to maintain complete backwards compatibility this import
star is needed.

gdal/swig/python/scripts/gcps2vec.py
rouault: > I see some tests added gdal_calc, for instance, to the path then imported it as a module. It's ugly, and with this PR it's unnecessary. I've changed this particular test to import the module from osgeo.utils. But I guess that if we want to maintain complete backwards compatibility this import star is needed.

OK

idanmiara:
@rouault Could you please advise why 2 tests were cancelled/failed ?

rouault:
It seems that in both cases test_gdal2tiles_py_simple() completed, but execution stalled on the following test test_gdal2tiles_py_zoom_option() and the jobs were ultimately cancelled by the CI infrastructure. One notable thing is that this test uses multiprocessing.
```
    # Because of multiprocessing, run as external process, to avoid issues with
    # Ubuntu 12.04 and socket.setdefaulttimeout()
    # as well as on Windows that doesn't manage to fork
```

idanmiara:
> It seems that in both cases test_gdal2tiles_py_simple() completed, but execution stalled on the following test test_gdal2tiles_py_zoom_option() and the jobs were ultimately cancelled by the CI infrastructure. One notable thing is that this test uses multiprocessing.
> 
> ```
>     # Because of multiprocessing, run as external process, to avoid issues with
>     # Ubuntu 12.04 and socket.setdefaulttimeout()
>     # as well as on Windows that doesn't manage to fork
> ```

The only change I've made to this script (besides moving it to osgeo.utils) is to to export argv as the main() parameter.
So I've no idea why it halts :(

rouault:
I've canceled https://travis-ci.com/github/OSGeo/gdal/builds/190068892 as the Travis CI is broken on 2 configs (unrelated to you) . I'm testing https://github.com/OSGeo/gdal/pull/3066 to resolve this. Once it is merged, you'll likely need to rebase on top of master so that your builds pass

rouault:
https://github.com/OSGeo/gdal/pull/3066 is merged. You can rebase on top of latest master

idanmiara:
> #3066 is merged. You can rebase on top of latest master

done.
I've combined the commits in this PR into 2 commits that make scenes.
I couldn't make the multiprocessing tests of gdal2tiles.py to work under the package so I left it as is in the scripts dir.

PR #3014, opened 2020-10-05
add gdaladdo option to set webp level

tbonfort:
Currently there is no way to set the webp compression level when creating overviews. This PR implements the functionality (with the same kludges as what's done when setting JPEG quality, i.e. it would need a substantial cleanup one day).
Documentation commits will follow pending on wether this is accepted...

PR #3039, opened 2020-10-08
 GTiff: add WEBP_LEVEL_OVERVIEW config option to set WebP compression level on overviews

rouault:
Supersedes #3014

@tbonfort I kept your first 3 commits, but I put before https://github.com/OSGeo/gdal/commit/bed9da378be158a56dad0956068c91598624cd4f which was the fix for the issue you saw. Basically the issue is that RestoreVolatileParameters() called by SetDirectory() from OpenOffset() was mostly a no-op because nCompression wasn't set at that point... Using libtiff in a multi-directory context is a mess beyond imagination!
And this fix actually underlined a unrelated issue with LERC... fixed by the first commit in the series (actually moved in its own PR https://github.com/OSGeo/gdal/pull/3038)

I let you confirm that things work as expected for you

### DIFF
--- a/autotest/gcore/tiff_write.py
+++ b/autotest/gcore/tiff_write.py
@@ -6952,6 +6952,15 @@ def test_tiff_write_185_lerc_create_and_overview():
     gdal.Unlink(filename)
     assert (cs, cs_ovr) == (4672, 1087)
 
+    filename2 = '/vsimem/test_tiff_write_185_lerc_create_and_overview_copy.tif'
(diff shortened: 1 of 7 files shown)


tbonfort:
Seems to be working correctly for me through gdaladdo, and buildoverviews of a newly created dataset. Thanks Even, I was banging my head against the wall with this, I concur that using libtiff in this context is nasty! I'll open another pr to backport the jpeg test updates.

PR #3052, opened 2020-10-13
add test for default tiff ovrview blocksize (#1611)

tbonfort:
test never added for #1611, doing this now. fails locally, awaiting ci tests.

### DIFF
--- a/autotest/gcore/tiff_write.py
+++ b/autotest/gcore/tiff_write.py
@@ -7366,6 +7366,40 @@ def test_tiff_write_internal_ovr_blocksize(blockSize, numThreads):
     ds = None
     gdaltest.tiff_drv.Delete(fname)
 
+###############################################################################
+# Test blocksize propagation while creating (internal) overviews
+# on a newly created dataset
+
+@pytest.mark.parametrize("blockSize,numThreads", [[64, None], [256, 8]])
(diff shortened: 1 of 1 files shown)


rouault:
@tbonfort Any idea why it fails locally for you ? I can only think to GDAL_TIFF_OVR_BLOCKSIZE being defined in your environment

tbonfort:
it's not the only test to fail, I have 15% failures on tiff_write, not only related to blocksize. It might be because I have a minimal build, but strange nonetheless.

tbonfort:
For the record, tests where picking up the system version of gdal/python instead of the one I compiled in a non standard directory. When compiling gdal with `--prefix=$foo` the tests should be ran with `PYTHONPATH=$foo/lib/python3/dist-packages LD_LIBRARY_PATH=$foo/lib` . I couldn't find this documented anywhere, should it be?

rouault:
> When compiling gdal with `--prefix=$foo` the tests should be ran with `PYTHONPATH=$foo/lib/python3/dist-packages LD_LIBRARY_PATH=$foo/lib` . I couldn't find this documented anywhere, should it be?

doc/source/contributing/developer.rst suggests to use ``. scripts/setdevenv.sh`` (it might pick the wrong python version if you have both python and python3 in your path though), which avoids to run "make install" when developing.

PR #3049, opened 2020-10-10
NITF: Add hexadecimal TRE creation

matthew-baran:
## What does this PR do?
This PR adds a NITF creation option to create TREs from a hexadecimal string.

Several TREs defined in the Spectral NITF Implementation Profile have non-string fields (uint, float32, etc).  This capability will help develop more effective tests of SNIP TREs.

The SNIP is in NGA.STND.0072_1.0_SNIP 7 June 2019.

## What are related issues/pull requests?
https://github.com/OSGeo/gdal/pull/2983
## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [x] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments
 - [ ] All CI builds and checks have passed

## Environment

Provide environment details, if relevant:

* OS:
* Compiler:

PR #3058, opened 2020-10-13
NITF: Add SNIP TRE - CSEXRB

matthew-baran:
## What does this PR do?
This PR adds the capability to read the CSEXRB TRE, defined in the Spectral NITF Implementation Profile
CSEXRB is defined in STDI-0002-1-v5.0 Appendix AH.

This TRE requires ~~two~~ one parsing change~~s~~:
1) ~~The ability to use 32 bit unsigned int fields as loop variables (max possible loop value can exceed signed int max)~~
2) The ability to read unsigned int fields between 1 and 8 bytes in width

https://github.com/OSGeo/gdal/pull/3049
https://github.com/OSGeo/gdal/pull/3023

## Tasklist

 - [x] Validate nitf_spec.xml against nitf_spec.xsd
 - [x] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:

### DIFF
--- a/gdal/data/nitf_spec.xsd
+++ b/gdal/data/nitf_spec.xsd
@@ -90,7 +90,7 @@
         <xs:attribute name="md_prefix" use="optional">
             <xs:simpleType>
                 <xs:restriction base="xs:string">
-                    <xs:pattern value="([a-z]|[A-Z]|[0-9]|_)*(%[0-9]*d)?([a-z]|[A-Z]|[0-9]|_)*"/>
+                    <xs:pattern value="([a-z]|[A-Z]|[0-9]|_)*(%[0-9]*lu)?([a-z]|[A-Z]|[0-9]|_)*"/>
                 </xs:restriction>
             </xs:simpleType>
         </xs:attribute>
@@ -131,6 +131,7 @@
                     <xs:enumeration value="integer"/>
                     <xs:enumeration value="real"/>
                     <xs:enumeration value="IEEE754_Float32_BigEndian"/>
+                    <xs:enumeration value="UnsignedInt_BigEndian"/>
                     <xs:enumeration value="bitmask"/>
                 </xs:restriction>
             </xs:simpleType>
--- a/gdal/data/nitf_spec.xsd
+++ b/gdal/data/nitf_spec.xsd
@@ -90,7 +90,7 @@
         <xs:attribute name="md_prefix" use="optional">
             <xs:simpleType>
                 <xs:restriction base="xs:string">
-                    <xs:pattern value="([a-z]|[A-Z]|[0-9]|_)*(%[0-9]*lu)?([a-z]|[A-Z]|[0-9]|_)*"/>
+                    <xs:pattern value="([a-z]|[A-Z]|[0-9]|_)*(%[0-9]*d)?([a-z]|[A-Z]|[0-9]|_)*"/>
                 </xs:restriction>
             </xs:simpleType>
         </xs:attribute>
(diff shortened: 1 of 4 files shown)


gdal/data/nitf_spec.xsd
```diff
@@ -131,6 +131,7 @@
                     <xs:enumeration value="integer"/>
                     <xs:enumeration value="real"/>
                     <xs:enumeration value="IEEE754_Float32_BigEndian"/>
+                    <xs:enumeration value="UnsignedInt_BigEndian"/>
```
matthew-baran, line 134: Since this type is parsed identically to "bitmask", we could potentially remove the "bitmask" type.  Benefits of keeping "bitmask" are that it's descriptive of how it's used in nitf_spec.xml in conjunction with bit testing (cond="MASK:2").  But drawback is having duplicate types.

gdal/frmts/nitf/nitffile.c
```diff
@@ -2479,7 +2489,7 @@ static char** NITFGenericMetadataReadTREInternal(char **papszMD,
                             CPLSPrintf("%s%s", pszMDPrefix, "NPLN"));
                 int NPLN = atoi(NITFFindValFromEnd(papszMD, *pnMDSize, pszMDItemName, "-1"));
                 CPLFree(pszMDItemName);
-                if (NPLN < 0)
+                if (NPLN < 1)
```
matthew-baran, line 2492: STDI-0002-1 indicates the valid range for this values is 2-999.  Changing this check guarantees that NPLN-1 is non-negative when we cast to unsigned long.

gdal/frmts/nitf/nitffile.c
```diff
@@ -2301,23 +2302,30 @@ static char** NITFGenericMetadataReadTREInternal(char **papszMD,
...
+                        int i;
+                        for (i = 0; i < nLength; ++i)
+                        {
+                            memcpy(&byData, pachTRE + *pnTreOffset + i, 1);
+                            nVal += (unsigned long long)byData << 8*(nLength - i - 1);
+                        }
+
                         pszValue = (char*)CPLMalloc(nBufferSize);
-                        CPLsnprintf(pszValue, nBufferSize, "%u", nVal);
+                        CPLsnprintf(pszValue, nBufferSize, "%llu", nVal);
```
rouault, line 2321: to fix this warning on mingw64
```
nitffile.c:2321:25: error: unknown conversion type character ‘l’ in format [-Werror=format=]
                         CPLsnprintf(pszValue, nBufferSize, "%llu", nVal);
```
maybe try:
```suggestion
                        CPLsnprintf(pszValue, nBufferSize, CPL_FRMT_GUIB, (GUIntBig)nVal);
```

rouault (commented):
> The ability to use 32 bit unsigned int fields as loop variables (max possible loop value can exceed signed int max)

That sounds very odd to me. How can that happen in practice: a TRE could hold gigabytes of data ?? That would certainly cause a denial of service in GDAL when reading it. If that's not needed, then all those related changes should be reverted.
I'd note also that the type for a 32 bit unsigned integer can be just %u (at least on platforms supported by GDAL)

matthew-baran:
> 
> 
> > The ability to use 32 bit unsigned int fields as loop variables (max possible loop value can exceed signed int max)
> 
> That sounds very odd to me. How can that happen in practice: a TRE could hold gigabytes of data ?? That would certainly cause a denial of service in GDAL when reading it. If that's not needed, then all those related changes should be reverted.
> I'd note also that the type for a 32 bit unsigned integer can be just %u (at least on platforms supported by GDAL)

That's a fair point.  The spec really does say (0, 2^32 - 1) as the range for that field, but it does seem completely impractical.  I'm happy to revert all the %d -> %lu messiness and leave a comment in nitf_spec.xml indicating the practical limit on that field.

FWIW I was using %lu instead of %u to try and catch overflow in strtoul() with errno - I was weighing that up against checking the value before converting to unsigned int, but that's all moot now.

PR #3063, opened 2020-10-14
GTiff: cleanup GNUmakefile and makefile.vc

rouault:


rouault:
Supersed by https://github.com/OSGeo/gdal/pull/3067

PR #3064, opened 2020-10-14
OGCAPI: support handling of vector content through items API

rouault:
This bridges the OGCAPI driver with the existing OAPIF driver


rouault:
CC @jerstlouis

PR #3070, opened 2020-10-15
NITF: Add nested variable support in xml:TRE

matthew-baran:
## What does this PR do?
This PR adds missing fields to the XML output of the "xml:TRE" metadata domain.  The fields that were not previously included were contained within \<loop\> or \<if\> blocks that accessed variables outside the immediate block scope.  These variables need md_prefix to be provided in order to recursively find the variable.

https://github.com/OSGeo/gdal/pull/2983

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [x] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:

### DIFF
--- a/autotest/gdrivers/nitf.py
+++ b/autotest/gdrivers/nitf.py
@@ -2697,11 +2697,11 @@ def test_nitf_77():
 
 def test_nitf_78():
     float_data = "40066666" # == struct.pack(">f", 2.1).hex()
-    bit_mask = "80000000" # Set bit 31 only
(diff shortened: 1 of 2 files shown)


rouault:
An idea that made through my brain last night: could someone gently tell NITF designers to stop this crazyness of TRE where you need to have specialized parsers and ad-hoc schemas to interpret them ? Can't they just create a generic JSON or XML TRE where you could an arbitrary payload ? Problem solved once for all.

PR #3081, opened 2020-10-19
Support for font and custom symbols in mitab

wicksell:
Make sure the correct type of point object is created depending on the type of symbol used. Also map MapInfo's custom symbol styles to OGR symbol styles. The symbol id is set to "mapinfo-custom-sym-[CUSTOM_STYLE]-[FILENAME]" with a fallback id on the form "[FILE_EXTENSION]-[FILENAME]" and finally "ogr-sym-9".

### DIFF
--- a/autotest/ogr/data/mitab/all_geoms.mif.golden.csv
+++ b/autotest/ogr/data/mitab/all_geoms.mif.golden.csv
@@ -1,6 +1,6 @@
 WKT,ogr_style
 "POINT (0 1)","SYMBOL(a:0,c:#000000,s:12pt,id:""mapinfo-sym-35,ogr-sym-10"")"
-"POINT (2 3)","SYMBOL(a:0,c:#000000,s:1pt,id:""mapinfo-sym-35,ogr-sym-10"")"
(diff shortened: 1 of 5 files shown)


rouault:
There are existing autotests that need to be adjusted
```
FAILED ogr/ogr_mitab.py::test_ogr_mitab_39 - Failed: <Failed instance>
FAILED ogr/ogr_mitab.py::test_ogr_mitab_41 - Failed: <Failed instance>
FAILED ogr/ogr_mitab.py::test_ogr_mitab_42 - Failed: <Failed instance>
FAILED ogr/ogr_mitab.py::test_ogr_mitab_43 - assert 6656 == 6144
```

wicksell:
The MacOS build fail with the following error:
Error: Fetching /usr/local/Homebrew/Library/Taps/local/homebrew-openssl failed!
I don't think this has anything to do with my changes. Does anyone have an idea on how to fix this?

rouault:
yes, all MacOS build fail. Likely a homebrew issue that will get fixed upstream

PR #3083, opened 2020-10-20
Support CharLS 2.1 on Debian as well.

sebastic:
Upstream changed the library name without an SONAME bump, this caused issues and was reverted in Debian, see:

 https://bugs.debian.org/971425

The charls Debian package still uses the CharLS library name for 2.1.


rouault:
@sebastic Thanks. I've regenerated autoconf and cherry-picked it in 3.2 and 3.1 branches

PR #3086, opened 2020-10-20
RMF: Better support for sparse files. Fill null tiles with NoData value.

drons:


gdal/frmts/rmf/rmfdataset.cpp
```diff
@@ -2891,7 +2911,8 @@ CPLErr RMFDataset::WriteRawTile(int nBlockXOff, int nBlockYOff,
 
 CPLErr RMFDataset::ReadTile(int nBlockXOff, int nBlockYOff,
                             GByte* pabyData, size_t nRawBytes,
-                            GUInt32 nRawXSize, GUInt32 nRawYSize)
+                            GUInt32 nRawXSize, GUInt32 nRawYSize, 
+                            bool& bNullTile)
```
rouault, line 2915: it would be safer to set bNullTile = false at the beginning of the method (if the caller didn't initialize the variable)

drons: Yes, it will be clearer for future generations.

drons:
Thanks!

PR #3071, opened 2020-10-16
NITF: Add ILLUMB TRE from SNIP specification

matthew-baran:
## What does this PR do?
This PR adds the capability to read the ILLUMB TRE, defined in the Spectral NITF Implementation Profile
The SNIP is in NGA.STND.0072_1.0_SNIP 7 June 2019. It out references to the relevant TREs in Table 6-2.
ILLUMB is defined in STDI-0002-1-v5.0 Appendix AL.

## What are related issues/pull requests?
https://github.com/OSGeo/gdal/pull/3058
https://github.com/OSGeo/gdal/pull/3023

## Tasklist

 - [x] Validate nitf_spec.xml against nitf_spec.xsd
 - [x] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments
 - [ ] All CI builds and checks have passed

## Environment

Provide environment details, if relevant:

* OS:
* Compiler:

PR #3090, opened 2020-10-21
NITF: Add support for SNIP TREs, CSRLSB and CSWRPB

matthew-baran:
## What does this PR do?
This PR adds the capability to read the CSRLSB and CSWRPB TREs, defined in the Spectral NITF Implementation Profile
CSRLSB and CSWRPB are defined in STDI-0002-1-v5.0 Appendix AH.

https://github.com/OSGeo/gdal/pull/3071
https://github.com/OSGeo/gdal/pull/3058

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [x] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:

### DIFF
--- a/autotest/gdrivers/nitf.py
+++ b/autotest/gdrivers/nitf.py
@@ -3175,6 +3175,103 @@ def test_nitf_87():
 """
     assert data == expected_data
 
+###############################################################################
+# Test parsing CSWRPB TRE (STDI-0002-1-v5.0 App AH)
+
+def test_nitf_88():
+    tre_data = "TRE=CSWRPB=1F199.9999999900000010000002000000300000040000005000000600000070000008" \
+               "1111-9.99999999999999E-99+9.99999999999999E+9900000"
+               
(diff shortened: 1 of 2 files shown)


rouault:
How many TREs are defined in SNIP ? This is crazy

matthew-baran:
> 
> 
> How many TREs are defined in SNIP ? This is crazy

Yeah.....    There are 18 TREs used by SNIP that were not previously supported by GDAL.  Including @bradh 's contribution, we have 11 already added.  So after these two, there would be 5 more.  I'm open to suggestions if there's a better way to go about this...

bradh:
Some of these are used in SNIP, but have been around for a while. We just didn't need them (or they weren't publicly released). 

Actually making use of those TREs (and the associated DESs) in the GLAS/GFM will be a capability that replaces RSM (and SENSRB). That isn't SNIP specific, but is used in the profile.

rouault:
Thanks for the explanations

PR #3093, opened 2020-10-22
Improve documentation

dg0yt:
## What does this PR do?

This PR fixes documentation errors, and it extents GTiff driver documentation with an alternative way of helping GDAL find `proj.db`.

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/gdal/doc/source/drivers/raster/gtiff.rst
+++ b/gdal/doc/source/drivers/raster/gtiff.rst
@@ -43,7 +43,8 @@ Most GeoTIFF projections should be supported, with the caveat that in
 order to translate uncommon Projected, and Geographic coordinate systems
 into OGC WKT it is necessary to have the PROJ proj.db database
 available. It must be found at the location pointed to by the PROJ_LIB
-environment variable.
+environment variable, or at one of the locations set programmatically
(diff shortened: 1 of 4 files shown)


dg0yt:
For the record, I came across this when dealing with problems reading EPSG:27770 georeferencing from a GeoTIFF: All I got was `LOCAL_CS["OSGB 1936 / British National Grid",AUTHORITY["EPSG","27700"]]` (https://github.com/OpenOrienteering/mapper/issues/1755#issuecomment-713046184). 
Now I see other people reporting possibly related issues on the `gdal-dev`list (and pointing to StackOverflow):
https://lists.osgeo.org/pipermail/gdal-dev/2020-October/052875.html
They are looking at `GDAL_DATA`, but not at `PROJ_LIB`. Maybe the PROJ search path requirement (since 3.0.0 IIUC) needs  more more prominent notice. 

PROJ has an "Environment variables" page in the "Using PROJ" section of its documentation. GDAL seems to lack a similar central entry point. Even `GDAL_DATA` has a mention only on the "Common options" page in the "Raster programs" section.

PR #3109, opened 2020-10-26
MRF: Restore raw Lerc1 detection

lucianpls:
Add function to read size from tile

## What does this PR do?

Fixes raw LERC1 detection/read via MRF, it was broken by previous commits

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:

### DIFF
--- a/autotest/gdrivers/mrf.py
+++ b/autotest/gdrivers/mrf.py
@@ -362,6 +362,27 @@ def test_mrf_lerc_with_huffman():
     gdal.Unlink('/vsimem/out.lrc')
     gdal.Unlink('/vsimem/out.til')
 
+def test_raw_lerc():
+    if 'LERC' not in gdal.GetDriverByName('MRF').GetMetadataItem('DMD_CREATIONOPTIONLIST'):
+        pytest.skip()
+
+    # Defaults to LERC2
+    for opt in 'OPTIONS=V1:1', None:
+        co = ['COMPRESS=LERC']
+        if opt:
+            co.append(opt)
+        gdal.Translate('/vsimem/out.mrf', 'data/byte.tif', format='MRF',
+                        creationOptions = co)
+        ds = gdal.Open('/vsimem/out.lrc')
+        with gdaltest.error_handler():
+            cs = ds.GetRasterBand(1).Checksum()
+        expected_cs = 4819
+        assert cs == expected_cs
+        ds = None
+        gdal.Unlink('/vsimem/out.mrf')
+        gdal.Unlink('/vsimem/out.mrf.aux.xml')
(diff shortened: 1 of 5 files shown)


rouault:
> Fixes raw LERC1 detection/read via MRF, it was broken by previous commits

As far as I can see autotest/gdrivers/mrf.py has tests for LERC1 . So this must be in a somewhat different situation ? Anyway, probably the mrf.py should be extended with a test case that shows the need for that commit

lucianpls:
> 
> 
> > Fixes raw LERC1 detection/read via MRF, it was broken by previous commits
> 
> As far as I can see autotest/gdrivers/mrf.py has tests for LERC1 . So this must be in a somewhat different situation ? Anyway, probably the mrf.py should be extended with a test case that shows the need for that commit

There aren't tests for reading a raw LERC blob, in either LERC1 or LERC2.  I'll add them.

autotest/gdrivers/mrf.py
```diff
@@ -362,6 +362,27 @@ def test_mrf_lerc_with_huffman():
     gdal.Unlink('/vsimem/out.lrc')
     gdal.Unlink('/vsimem/out.til')
 
+def test_raw_lerc():
+    if 'LERC' not in gdal.GetDriverByName('MRF').GetMetadataItem('DMD_CREATIONOPTIONLIST'):
```
rouault, line 366: how is this test case different from https://github.com/OSGeo/gdal/blob/master/autotest/gdrivers/mrf.py#L47 ? Is OPTIONS:V1:1 different from OPTIONS=V1:YES ? If so, it could probably be added to the mrf_list array

autotest/gdrivers/mrf.py
```diff
@@ -362,6 +362,27 @@ def test_mrf_lerc_with_huffman():
...
+        pytest.skip()
+
+    # Defaults to LERC2
+    for opt in 'OPTIONS=V1:1', None:
+        co = ['COMPRESS=LERC']
+        if opt:
+            co.append(opt)
+        gdal.Translate('/vsimem/out.mrf', 'data/byte.tif', format='MRF',
+                        creationOptions = co)
+        ds = gdal.Open('/vsimem/out.lrc')
```
lucianpls, line 376: @rouault 
This line is the difference, this test opens the data file directly, not the MRF.  Since the data file is a single LERC tile, it will be recognized and opened (without metadata).  This functionality is used in the WMS driver, to read data from services that send back LERC formatted responses.

rouault: ok, got it

Issue #3097, opened 2020-10-23
FlatGeobuf; implement editable capabilities with deferred persistence


bjornharrtell:
To make it seem editable to clients like QGIS, see https://github.com/qgis/QGIS/issues/39574.

bjornharrtell:
Done: :)

PR #3110, opened 2020-10-26
FlatGeobuf; editable capabilities

bjornharrtell:
ref https://github.com/OSGeo/gdal/issues/3097

### DIFF
--- a/autotest/ogr/ogr_flatgeobuf.py
+++ b/autotest/ogr/ogr_flatgeobuf.py
@@ -666,10 +666,10 @@ def test_ogr_flatgeobuf_column_metadata():
 
     f = ogr.Feature(lyr.GetLayerDefn())
     f['int'] = 1
-    f['str1'] = 'test1';
-    f['str2'] = 'test2';
-    f['float1'] = 1.1234;
-    f['float2'] = 12.123;
+    f['str1'] = 'test1'
+    f['str2'] = 'test2'
+    f['float1'] = 1.1234
+    f['float2'] = 12.123
     f.SetGeometry(ogr.CreateGeometryFromWkt('POINT (0 0)'))
     lyr.CreateFeature(f)
 
@@ -694,6 +694,64 @@ def test_ogr_flatgeobuf_column_metadata():
     assert lyr.GetLayerDefn().GetFieldDefn(4).GetWidth() == 5
     assert lyr.GetLayerDefn().GetFieldDefn(4).GetPrecision() == 3
     ds = None
+
+    ogr.GetDriverByName('FlatGeobuf').DeleteDataSource('/vsimem/test.fgb')
+    assert not gdal.VSIStatL('/vsimem/test.fgb')
+
+def test_ogr_flatgeobuf_editing():
+    ds = ogr.GetDriverByName('FlatGeobuf').CreateDataSource('/vsimem/test.fgb')
+    lyr = ds.CreateLayer('test', geom_type = ogr.wkbPoint)
+    fld_defn = ogr.FieldDefn('int', ogr.OFTInteger)
+    lyr.CreateField(fld_defn)
+    fld_defn = ogr.FieldDefn('str1', ogr.OFTString)
+    lyr.CreateField(fld_defn)
+    f = ogr.Feature(lyr.GetLayerDefn())
+    f.SetGeometry(ogr.CreateGeometryFromWkt('POINT (0 0)'))
+    f['int'] = 1
+    f['str1'] = 'test1'
+    lyr.CreateFeature(f)
+
+    f = ogr.Feature(lyr.GetLayerDefn())
+    f.SetGeometry(ogr.CreateGeometryFromWkt('POINT (1 1)'))
+    f['int'] = 2
+    f['str1'] = 'test2'
+    lyr.CreateFeature(f)
+
+    c = lyr.GetFeatureCount()
+    assert c == 2
+
+    ds = None
+
+    ds = ogr.Open('/vsimem/test.fgb', update=1)
+    lyr = ds.GetLayer(0)
+
+    assert lyr.TestCapability(ogr.OLCDeleteFeature) == 1
+    assert lyr.DeleteFeature(1) == 0
+    assert lyr.DeleteFeature(1) == ogr.OGRERR_NON_EXISTING_FEATURE
+    assert lyr.TestCapability(ogr.OLCReorderFields) == 1
+    #assert lyr.ReorderFields([0, 1]) == 0
+    assert lyr.DeleteField(1) == 0
+    f = lyr.GetFeature(0)
+    f.SetGeomField(0, ogr.CreateGeometryFromWkt('POINT (2 2)'))
+    assert lyr.SetFeature(f) == 0
+    lyr.ResetReading()
+
+    ds = None
+
+    ds = ogr.Open('/vsimem/test.fgb')
+    lyr = ds.GetLayer(0)
+
+    c = lyr.GetFeatureCount()
+    assert c == 1
+
+    f = lyr.GetNextFeature()
+    assert f is not None
+    assert f.GetGeometryRef().ExportToWkt() == 'POINT (2 2)'
+    assert f[0] == 2
+    assert f.GetFieldCount() == 1
+
+    f = lyr.GetNextFeature()
+    assert f is None
 
     ogr.GetDriverByName('FlatGeobuf').DeleteDataSource('/vsimem/test.fgb')
     assert not gdal.VSIStatL('/vsimem/test.fgb')
--- a/gdal/ogr/ogrsf_frmts/flatgeobuf/makefile.vc
+++ b/gdal/ogr/ogrsf_frmts/flatgeobuf/makefile.vc
@@ -1,6 +1,6 @@
 
-OBJ =	geometryreader.obj geometrywriter.obj ogrflatgeobufdataset.obj ogrflatgeobuflayer.obj packedrtree.obj
-EXTRAFLAGS =	-I. -I.. -I..\..
+OBJ =	geometryreader.obj geometrywriter.obj ogrflatgeobufdataset.obj ogrflatgeobuflayer.obj ogrflatgeobufeditablelayer.obj packedrtree.obj
+EXTRAFLAGS =	-I. -I.. -I..\.. -I..\generic
 
 GDAL_ROOT	=	..\..\..
 
--- a/gdal/ogr/ogrsf_frmts/flatgeobuf/ogr_flatgeobuf.h
+++ b/gdal/ogr/ogrsf_frmts/flatgeobuf/ogr_flatgeobuf.h
@@ -5,7 +5,7 @@
  * Author:   Björn Harrtell <bjorn at wololo dot org>
  *
  ******************************************************************************
- * Copyright (c) 2018-2019, Björn Harrtell <bjorn at wololo dot org>
+ * Copyright (c) 2018-2020, Björn Harrtell <bjorn at wololo dot org>
  *
  * Permission is hereby granted, free of charge, to any person obtaining a
  * copy of this software and associated documentation files (the "Software"),
@@ -31,6 +31,7 @@
 
 #include "ogrsf_frmts.h"
 #include "ogr_p.h"
+#include "ogreditablelayer.h"
 
 #include "header_generated.h"
 #include "feature_generated.h"
@@ -94,6 +95,7 @@ class OGRFlatGeobufLayer final : public OGRLayer
 
         // creation
         bool m_create = false;
+        bool m_update = false;
         std::vector<std::shared_ptr<FlatGeobuf::Item>> m_featureItems; // feature item description used to create spatial index
         bool m_bCreateSpatialIndexAtClose = true;
         bool m_bVerifyBuffers = true;
@@ -101,7 +103,7 @@ class OGRFlatGeobufLayer final : public OGRLayer
         VSILFILE *m_poFpWrite = nullptr;
         uint64_t m_writeOffset = 0; // current write offset
         uint16_t m_indexNodeSize = 0;
-        std::string m_oTempFile; // holds generated temp file name for two pass writing
+        std::string m_osTempFile; // holds generated temp file name for two pass writing
         uint32_t m_maxFeatureSize  = 0;
 
         // shared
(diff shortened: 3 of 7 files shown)


bjornharrtell:
@rouault I believe I've got this working now and ready for review. If possible I'd like to try to get this backported to 3.2.x and 3.1.x.

autotest/ogr/ogr_flatgeobuf.py
```diff
@@ -697,3 +697,22 @@ def test_ogr_flatgeobuf_column_metadata():
...
+
+    ds = ogr.Open('/vsimem/test.fgb', update=1)
+    lyr = ds.GetLayer(0)
+
+    assert lyr.TestCapability(ogr.OLCDeleteFeature) == 1
+    assert lyr.DeleteFeature(0) == 0
+    assert lyr.DeleteFeature(0) == ogr.OGRERR_NON_EXISTING_FEATURE
+    lyr.ResetReading()
+    f = lyr.GetNextFeature()
+    assert f is None
```
rouault, line 718: test a bit light.
dataset should be reopended after editing to check eveything is fine. you'd probably also want to test SetFeature(), and all methods of OGRFlatGeobufEditableLayer

cleanup of /vsimem/test.fgb should be done, but probably something for the tear-down code

gdal/ogr/ogrsf_frmts/flatgeobuf/ogr_flatgeobuf.h
```diff
@@ -144,19 +154,43 @@ class OGRFlatGeobufLayer final : public OGRLayer
         void VerifyBuffers( int bFlag ) { m_bVerifyBuffers = CPL_TO_BOOL(bFlag); }
 
         const std::string& GetFilename() const { return m_osFilename; }
+
+        static std::string GetTempFilePath(CPLString fileName, char **papszOptions);
+
+        uint16_t GetIndexNodeSize() const { return m_indexNodeSize; } 
+        OGRwkbGeometryType GetOGRGType() const { return m_eGType; }
```
rouault, line 161: you'd perhaps want to override GetGeomType() instead (but only if GetGeomType() always equals to GetLayerDefn()->GetGeomType())

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobufdataset.cpp
```diff
@@ -179,18 +182,19 @@ OGRFlatGeobufDataset::~OGRFlatGeobufDataset()
...
-    }
 
-    const auto bVerifyBuffers = CPLFetchBool( poOpenInfo->papszOpenOptions, "VERIFY_BUFFERS", true );
+    const auto bVerifyBuffers = CPLFetchBool(poOpenInfo->papszOpenOptions, "VERIFY_BUFFERS", true);
+
+    auto isDir = CPL_TO_BOOL(poOpenInfo->bIsDirectory);
+    auto bUpdate = poOpenInfo->eAccess == GA_Update;
+
+    if (isDir && bUpdate)
+        CPLError(CE_Failure, CPLE_AppDefined, "Unable to open dir in update mode");
```
rouault, line 194: you should probably return nullptr then

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobufdataset.cpp
```diff
@@ -307,11 +311,17 @@ bool OGRFlatGeobufDataset::OpenFile(const char* pszFilename, VSILFILE* fp, bool
 
     CPLDebug("FlatGeobuf", "Features start at offset (%lu)", static_cast<long unsigned int>(offset));
 
-    auto poLayer = std::unique_ptr<OGRFlatGeobufLayer>(
-        new OGRFlatGeobufLayer(header, buf.release(), pszFilename, fp, offset));
+    CPLDebug("FlatGeobuf", "Creating OGRFlatGeobufLayer");
```
rouault, line 314: you'd probably want to use CPLDebugOnly() in a number of places to have such messages only for debug builds as they are likely not very interesting for production builds

gdal/ogr/ogrsf_frmts/flatgeobuf/ogr_flatgeobuf.h
```diff
@@ -144,19 +154,43 @@ class OGRFlatGeobufLayer final : public OGRLayer
...
+        virtual GIntBig     GetFeatureCount( int bForce = TRUE ) override;
+
+        const std::string& GetFilename() const { return m_poFlatGeobufLayer->GetFilename(); }
 };
 
 class OGRFlatGeobufDataset final: public GDALDataset
 {
     private:
-        std::vector<std::unique_ptr<OGRFlatGeobufLayer>> m_apoLayers;
+        std::vector<OGRLayer *> m_apoLayers;
```
rouault, line 185: why not using a std::vector<std::unique_ptr<OGRLayer>> ? You should be able to do that with calls to .release() where needed when wrapping a OGRFlatGeobufLayer into a OGRFlatGeobufEditableLayer

gdal/ogr/ogrsf_frmts/flatgeobuf/ogr_flatgeobuf.h
```diff
@@ -144,19 +154,43 @@ class OGRFlatGeobufLayer final : public OGRLayer
...
+
+        uint16_t GetIndexNodeSize() const { return m_indexNodeSize; } 
+        OGRwkbGeometryType GetOGRGType() const { return m_eGType; } 
+};
+
+class OGRFlatGeobufEditableLayer final: public OGREditableLayer
+{
+    std::set<CPLString> m_oSetFields;
+
+    OGRFlatGeobufLayer *m_poFlatGeobufLayer;
```
rouault, line 168: should likely be a std::unique_ptr<OGRFlatGeobufLayer> so it gets destroyed

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobufeditablelayer.cpp
```diff
@@ -0,0 +1,256 @@
...
+        osLayerName.c_str(), osTmpFilename.c_str(), spatialRef, gType, createIndex, m_papszOpenOptions);
+  
+    OGRErr eErr = OGRERR_NONE;
+    OGRFeatureDefn *poEditableFDefn = poEditableLayer->GetLayerDefn();
+    for (int i = 0; eErr == OGRERR_NONE && i < poEditableFDefn->GetFieldCount();
+         i++ )
+    {
+        OGRFieldDefn oFieldDefn(poEditableFDefn->GetFieldDefn(i));
+        int iGeomFieldIdx = 0;
+        if( (EQUAL(oFieldDefn.GetNameRef(), "WKT") &&
```
rouault, line 98: to remove. completely CSV specific

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobufeditablelayer.cpp
```diff
@@ -0,0 +1,256 @@
...
+    CPLFree(pszQueryStringBak);
+    poEditableLayer->SetSpatialFilter(iFilterGeomIndexBak, poFilterGeomBak);
+    delete poFilterGeomBak;
+
+    if( eErr != OGRERR_NONE )
+    {
+        CPLError(CE_Failure, CPLE_AppDefined, "Error while creating %s",
+                 osTmpFilename.c_str());
+        VSIUnlink(osTmpFilename);
+        VSIUnlink(CPLResetExtension(osTmpFilename, "fgbt"));
```
rouault, line 153: fgbt :-)  the equivalent of csvt for flatgeobuf ? to remove

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobufeditablelayer.cpp
```diff
@@ -0,0 +1,256 @@
...
+    {
+        for( int i = 0; i < m_poEditableFeatureDefn->GetFieldCount(); i++ )
+        {
+            m_oSetFields.insert(CPLString(
+                m_poEditableFeatureDefn->GetFieldDefn(i)->GetNameRef()).toupper());
+        }
+    }
+
+    const OGRFlatGeobufCreateFieldAction eAction = CREATE_FIELD_DO_NOTHING;
+    //OGRFlatGeobufLayer::PreCreateField(
```
rouault, line 218: to remove all this CREATE_FIELD_ logic and the related enum. Must be CSV specific

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobufeditablelayer.cpp
```diff
@@ -0,0 +1,256 @@
...
+                 "Limiting to 10000 fields");
+        return OGRERR_FAILURE;
+    }
+
+    if( m_oSetFields.empty() )
+    {
+        for( int i = 0; i < m_poEditableFeatureDefn->GetFieldCount(); i++ )
+        {
+            m_oSetFields.insert(CPLString(
+                m_poEditableFeatureDefn->GetFieldDefn(i)->GetNameRef()).toupper());
```
rouault, line 213: you probably don't need the .toupper() logic here and later

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
```diff
@@ -131,16 +133,34 @@ OGRFlatGeobufLayer::OGRFlatGeobufLayer(
...
+        // Unlink it now to avoid stale temporary file if killing the process
+        // (only works on Unix)
+        VSIUnlink(m_oTempFile.c_str());
+    } else {
+        CPLDebug("FlatGeobuf", "No spatial index will write directly to output");
+        m_poFpWrite = VSIFOpenL(pszFilename, "wb");
+        savedErrno = errno;
+    }
+    if (m_poFpWrite == nullptr) {
+        CPLError(CE_Failure, CPLE_OpenFailed,
```
rouault, line 158: a constructor that fails is not a good thing. All that logic that can fail should likely be in a bool Initialize() method whose caller would test the return code

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
```diff
@@ -1310,3 +1330,19 @@ void OGRFlatGeobufLayer::ResetReading()
     m_ignoreAttributeFilter = false;
     return;
 }
+
+std::string OGRFlatGeobufLayer::GetTempFilePath(CPLString fileName, char **papszOptions) {
```
rouault, line 1334:
```suggestion
std::string OGRFlatGeobufLayer::GetTempFilePath(const CPLString& fileName, CSLConstList papszOptions) {
```

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobufeditablelayer.cpp
```diff
@@ -0,0 +1,256 @@
...
+    if( eErr != OGRERR_NONE )
+    {
+        CPLError(CE_Failure, CPLE_AppDefined, "Error while creating %s",
+                 osTmpFilename.c_str());
+        VSIUnlink(osTmpFilename);
+        VSIUnlink(CPLResetExtension(osTmpFilename, "fgbt"));
+        return eErr;
+    }
+
+    delete m_poFlatGeobufLayer;
```
rouault, line 157: would be safer to nullify m_poFlatGeobugLayer right away. bonus point for making m_poFlatGeobufLayer a std::unique_ptr<> and .reset()'ing it

rouault:
> If possible I'd like to try to get this backported to 3.2.x and 3.1.x.

There won't be any 3.1 release now. As far as 3.2.x, this is non trivial logic. Hopefully this shouldn't break existing workflows, but I guess this should be first exercised in master before considering a 3.2 backport

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobufeditablelayer.cpp
bjornharrtell: Tried but failed to get the bonus points.. :)

gdal/ogr/ogrsf_frmts/flatgeobuf/ogr_flatgeobuf.h
bjornharrtell: I find it hard to determine if I can guarantee that GetGeomType() always equals to GetLayerDefn()->GetGeomType())

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
bjornharrtell: even if I break out that logic, how would I handle the case differently? I cannot construct in this case.. should I avoid CPLError anyway?

bjornharrtell: or are you saying I should revert it back to the behavior in that the poFpWrite and oTempFile should be created ahead of construction and injected?

rouault: If you want to do minimal changes, you could have a IsValid() method that callers of the constructor should call to check if the object is valid.
Or perhaps better, make the constructor private, and have a static Create() method that returns a pointer to it or a nullptr in case of error.

autotest/ogr/ogr_flatgeobuf.py
bjornharrtell: agreed, I have done some expanded local tests indicating that the implementation has bugs. will be back with something better, I hope.

bjornharrtell: Now extended quite a bit. I cannot verify that ReorderFields works and when investigating that I couldn't make it work on the csv driver either.

gdal/ogr/ogrsf_frmts/flatgeobuf/ogr_flatgeobuf.h
bjornharrtell: I'm not sure.. the pointer is passed to the OGREditableLayer constructor which appears to take ownership?

bjornharrtell: I was able to remove it completely and instead use the protected member from the base class.

bjornharrtell:
@rouault ready for re-review. :)

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobufdataset.cpp
```diff
@@ -179,18 +180,21 @@ OGRFlatGeobufDataset::~OGRFlatGeobufDataset()
...
         return nullptr;
-    }
 
     const auto bVerifyBuffers = CPLFetchBool( poOpenInfo->papszOpenOptions, "VERIFY_BUFFERS", true );
 
+    auto isDir = CPL_TO_BOOL(poOpenInfo->bIsDirectory);
+    auto bUpdate = poOpenInfo->eAccess == GA_Update;
+
+    if (isDir && bUpdate){
+        CPLError(CE_Failure, CPLE_AppDefined, "Unable to open dir in update mode");
```
rouault, line 192: I suspect this CPLError() is causing the CI issues in the FileGDB one. When a driver emits a CPLError() in its Open() code, no other drivers are tried afterwards. And as OGRFlatGeobufDriverIdentify() just returns -1 "I may recognize or not this dataset" if a directory is passed, this will go here. So probably just remove the CPLError()

gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobufeditablelayer.cpp
```diff
@@ -0,0 +1,230 @@
...
+OGRErr OGRFlatGeobufEditableLayer::CreateField( OGRFieldDefn *poNewField, int bApproxOK )
+{
+    if( m_poEditableFeatureDefn->GetFieldCount() >= 10000 )
+    {
+        CPLError(CE_Failure, CPLE_AppDefined,
+                 "Limiting to 10000 fields");
+        return OGRERR_FAILURE;
+    }
+
+    if( m_oSetFields.empty() )
```
rouault, line 190: I believe everything m_oSetFields related was CSV specific. You could probably just remove it and the override of DeleteField() and AlterFieldDefn()

bjornharrtell:
@rouault all done and mostly green.

gdal/ogr/ogrsf_frmts/flatgeobuf/makefile.vc
```diff
@@ -1,6 +1,6 @@
 
 OBJ =	geometryreader.obj geometrywriter.obj ogrflatgeobufdataset.obj ogrflatgeobuflayer.obj packedrtree.obj
-EXTRAFLAGS =	-I. -I.. -I..\..
+EXTRAFLAGS =	-I. -I.. -I..\.. -I..\generic
```
rouault, line 3: ogrflatgeobufeditablelayer.obj should be added to the OBJ line just above

rouault (commented):
I'm also wondering if we shouldn't add a new capability, like OLCEditableThroughRewrite, to indicate to external software (like QGIS if he would use that), that the edition capabilities involve rewriting the whole file, which can have performance impacts. Or maybe this is over-engineering

bjornharrtell:
> I'm also wondering if we shouldn't add a new capability, like OLCEditableThroughRewrite, to indicate to external software (like QGIS if he would use that), that the edition capabilities involve rewriting the whole file, which can have performance impacts. Or maybe this is over-engineering

Makes sense but that should be added in a separate PR for all relevant drivers in that case, in my opinion.

rouault:
> Makes sense but that should be added in a separate PR for all relevant drivers in that case, in my opinion.

yes, was mostly thinking out loud

bjornharrtell:
@rouault all done again and green except macos but seems it was caused by temp network issue.

PR #3115, opened 2020-10-27
refactor gdal/swig/python/samples/*.py

idanmiara:
## What does this PR do?

gdal/swig/python/samples/*.py - refactor all scripts to make them easier for modularization
gdal/swig/python/samples/ogr2ogr.py - main now returns 0/1 instead of True/False to be compatible with all other scripts (Is there anything that calls ogr2ogr.main not as a script that it will break backwards compatibility?)

https://github.com/OSGeo/gdal/pull/3031

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments


rouault:
Those samples scripts are just examples. They are not intended to be used as python modules. If some of them are of sufficient interest, they should be moved to scripts/, and receive the needed love, in particular documentation and tests.

idanmiara:
> Those samples scripts are just examples. They are not intended to be used as python modules. If some of them are of sufficient interest, they should be moved to scripts/, and receive the needed love, in particular documentation and tests.

well, some of them are indeed small examples, but I think many of them are too complex to be called "examples".
I wanted to reuse some of the code inside some of them, and I thought it might be a good idea to do them all at once...
Actually, some of them do have tests, and are imported in a convoluted way as modules, like ogr2ogr, for instance.
We can use the README as a baseline for documentation stub.

I think it will be a good idea to lay these foundations so others can refactor and reuse this code.
If you think some of them are too simple to be called "utils", maybe we can move them to another sub-package.

rouault:
> Actually, some of them do have tests, and are imported in a convoluted way as modules, like ogr2ogr, for instance.

That predates the time where gdal.VectorTranslate() was made available. This script has no more usefulness, except showing the use of the OGR Python API

idanmiara:
> > Actually, some of them do have tests, and are imported in a convoluted way as modules, like ogr2ogr, for instance.
> 
> That predates the time where gdal.VectorTranslate() was made available. This script has no more usefulness, except showing the use of the OGR Python API

I see.

I refactored some of the code to be reusable now (return instead of exit, main function, parametrized inputs, etc.)
Would you like to merge this work? as is? export some of the complex samples into modules?

I actually wanted to improve and reuse attachpct.py, so I could export only this one into a module if you don't think the other ones qualify for that...

idanmiara:
I was able to find and fix some bugs in the examples in the process...

rouault:
I'm OK for the first step of having the main calling a method to be potentially usable by import. But not from moving everything into the osgeo python package. Should be done one a case by case basis, possibly with changes in argument names, etc, to make it more consistent with existing stuff.

gdal/swig/python/samples/assemblepoly.py
```diff
@@ -41,6 +41,7 @@ def doit(name):
     ds = ogr.Open(name, update=1)
 
     layer_count = ds.GetLayerCount()
+    print('{}: Layer count: {}'.format(name, layer_count))
```
rouault, line 44: I guess this is a debug statement to remove ?

idanmiara: The line above for getting layer_count was there before and was not used. 
Now that it's in a function flake8 doesn't like it, so I was either to remove it or to do something with it...
Do you prefer to remove both?

rouault: > Do you prefer to remove both?

Yes, no reason to output that without any good reason

idanmiara: done

idanmiara:
I've added the README.md changes to this PR and squashed.

idanmiara:
thanks!

PR #3117, opened 2020-10-27
refactor python scripts and add osgeo.auxilary package

idanmiara:
## What does this PR do?

osgeo.auxiliary package added. to be home for auxilary functions for GDAL python scripting. more stuff should arrive here soon :)
auxiliary/base.py - combined all versions of GetOutputDriverFor, GetOutputDriversFor. removed duplicated versions from other source files. added some other useful functions.
setup.py - added osgeo.auxiliary

pct2rgb.py, rgb2pct.py - added a doit function and return ds

ogr/ogr_as_sqlite_extension.py - added main function
xmlvalidate.py - added main stub

attachpct.py - use GetOutputDriverFor

replace sys.exit(x) by return x for the osgeo.utils functions

## What are related issues/pull requests?

prerequisite:
https://github.com/OSGeo/gdal/pull/3115

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments
 - [ ] All CI builds and checks have passed

PR #3095, opened 2020-10-22
gdal_calc.py, pct2rgb.py - color files and extent options

idanmiara:
## What does this PR do?

osgeo.auxiliary/color_palette.py - handling color palettes (i.e. number -> color)
osgeo.auxiliary/color_table.py - handling color tables (i.e. index -> color)
osgeo.auxiliary/geo_rectangle.py - handling float rectangles (i.e. for extents)
osgeo.auxiliary/extent.py - handling extents

attachpct.py - upgraded into a module
autotest/pyscripts/test_rgb2pct.py - adding tests for attachpct.py

rgb2pct.py, pct2rgb.py, attachpct.py - use get_color_table to allow loading color palette from a ds or a txt file

gdal_calc.py - add color_table file as input and more extent options.

geotransforms_tut.rst - some info about geotransforms, referenced in index.rst, raster_api_tut.rst

This is a draft. I'd appreciate your review.
I didn't write enough docs and tests yet.

## What are related issues/pull requests?

Prerequisite:
https://github.com/OSGeo/gdal/pull/3016 (merged)
https://github.com/OSGeo/gdal/pull/3115
https://github.com/OSGeo/gdal/pull/3117

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments
 - [ ] All CI builds and checks have passed

PR #3131, opened 2020-11-01
improve color palette support: gdal_calc, gdalattachpct, rgb2pct, pct2rgb

idanmiara:
color palette python modules improvements
- gdalattachpct.py - upgraded attachpct.py into a module, added support for color table files and format
- gdal/swig/python/samples/attachpct.py - upgraded into a module
- gdal/swig/python/samples/README.md - remove attachpct.py
- gdal/swig/python/osgeo/utils/gdalattachpct.py - use GetOutputDriverFor (to support other drivers), get_color_table (to allow loading color palette from a ds or a txt file), open_ds (allow a ds input in addition of a filename)
- gdal/doc/source/programs/gdalattachpct.rst - document gdalattachpct.py
- gdal/doc/source/programs/index.rst - reference gdalattachpct.py
- autotest/pyscripts/test_pct.py - rename and add tests for gdalattachpct.py
- gdal/swig/python/osgeo/utils/[rgb2pct.py, pct2rgb.py] - use get_color_table to allow loading color palette from a ds or a txt file; use open_ds to allow input a ds in addition to filename
- added osgeo/utils/auxiliary/[color_palette.py, color_table.py] added to handle color palettes (the former number -> color, the latter index -> color)
- added `color_palette` option to gdal_calc

https://github.com/OSGeo/gdal/pull/3341 (prerequisite) - Please review it first
https://github.com/OSGeo/gdal/pull/3117 (merged)
https://github.com/OSGeo/gdal/pull/3095 (closed)

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/gdal/doc/source/programs/gdal_calc.rst
+++ b/gdal/doc/source/programs/gdal_calc.rst
@@ -88,6 +88,11 @@ but no projection checking is performed (unless projectionCheck option is used).
 
     GDAL format for output file.
 
+.. option:: color-table=<filename>
+
+    Allows to specify a filename of a color table (or a ColorTable object) (with Palette Index interpretation) to be used for the output raster.
+    Supported formats: txt (i.e. like gdaldem, but color names are not supported), qlr, qml (i.e. exported from QGIS)
+
 .. option:: --extent=<option>
 
     ..versionadded:: 3.3
--- a/gdal/swig/python/osgeo/utils/gdal_calc.py
+++ b/gdal/swig/python/osgeo/utils/gdal_calc.py
@@ -69,6 +69,7 @@ from osgeo.utils.auxiliary.util import GetOutputDriverFor
 from osgeo.utils.auxiliary.extent_util import Extent, GT
 from osgeo.utils.auxiliary import extent_util
 from osgeo.utils.auxiliary.rectangle import GeoRectangle
+from osgeo.utils.auxiliary.color_table import get_color_table
 
 GDALDataType = int
 
@@ -353,7 +354,7 @@ def doit(opts, args):
             if opts.color_table:
                 # set color table and color interpretation
                 if is_path_like(opts.color_table):
-                    raise Exception('Error! reading a color table from a file is not supported yet')
+                    opts.color_table = get_color_table(opts.color_table)
                 myOutB.SetRasterColorTable(opts.color_table)
                 myOutB.SetRasterColorInterpretation(gdal.GCI_PaletteIndex)
 
@@ -515,7 +516,8 @@ def doit(opts, args):
 def Calc(calc: Union[str, Sequence[str]], outfile: Optional[PathLike] = None, NoDataValue: Optional[Number] = None,
          type: Optional[Union[GDALDataType, str]] = None, format: Optional[str] = None,
          creation_options: Optional[Sequence[str]] = None, allBands: str = '', overwrite: bool = False,
-         hideNoData: bool = False, projectionCheck: bool = False, color_table: Optional[gdal.ColorTable] = None,
+         hideNoData: bool = False, projectionCheck: bool = False,
+         color_table: Optional[Union[PathLike, gdal.ColorTable]] = None,
          extent: Optional[Extent] = None, projwin: Optional[Union[Tuple, GeoRectangle]] = None, user_namespace=None,
          debug: bool=False, quiet: bool = False, **input_files):
 
@@ -620,7 +622,7 @@ def main(argv):
     parser.add_option("--quiet", dest="quiet", action="store_true", help="suppress progress messages")
     parser.add_option("--optfile", dest="optfile", metavar="optfile", help="Read the named file and substitute the contents into the command line options list.")
 
-    # parser.add_option("--color_table", dest="color_table", help="color table file name")
+    parser.add_option("--color-table", dest="color_table", help="color table file name")
     parser.add_option("--extent", dest="extent",
                       choices=[e.name.lower() for e in Extent],
                       help="how to treat mixed geotrasnforms")
--- a/autotest/pyscripts/test_pct.py
+++ b/autotest/pyscripts/test_pct.py
@@ -34,6 +34,8 @@ import struct
 
 
 from osgeo import gdal
+from osgeo.utils import gdalattachpct, rgb2pct
+import osgeo.utils.auxiliary.color_table as color_table
 import gdaltest
 import test_py_scripts
 import pytest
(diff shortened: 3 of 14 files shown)


rouault:
Regarding attachpct.py, if we document it as a program, it should no longer be in samples/ but in scripts/ so it is installed. Which then raises the question of its naming. We do have a number of historic utilities without gdal prefix, but I believe it should probably be gdalattachpct.py for proper namespacing.
For clarity, its tests should go into a dedicated file, not test_rgb2pct.py

I'm not clear why extent.py or geo_rectangle.py are needed here.

Basically the scope of this PR is too big for proper review. Try to keep it to the strict minimum

rouault:
One thing that should be clarified too is if stuff into auxiliary/ is meant to be used by code external to GDAL (and if so, there should be a clear distinction on what is in its API and what is internal details), or if it is purely code to be used by GDAL scripts/ and samples/ . I'd assume the later, but we should probably document this properly

rouault:
Please also rebase on top of latest master as CI now runs flake8 on swig/python/osgeo/utils and auxiliary

idanmiara:
>probably be gdalattachpct.py for proper namespacing.

OK, will do.

>For clarity, its tests should go into a dedicated file, not test_rgb2pct.py

I saw that the tests for rgb2pct and pct2rgb are in the same module. 
I guess it makes scenes either to split them as well or to put all the color palette related tests in the same module and name it properly. The argument for a single test could be that a single test can use some of these modules together (as I did).
Would you like to split it into 3 test files? 

>I'm not clear why extent.py or geo_rectangle.py are needed here.

Yes, I should have put that in another PR. I will.

>One thing that should be clarified too is if stuff into auxiliary/ is meant to be used by code external to GDAL (and if so, there should be a clear distinction on what is in its API and what is internal details), or if it is purely code to be used by GDAL scripts/ and samples/ . I'd assume the later, but we should probably document this properly

I was pondering about that as well. Would it make scenes to you to make 2 auxiliary modules (maybe osgeo.gdal_auxiliary and auxiliary), one that is self contained and doesn't depend on the osgeo module and another that depends on it?

>Please also rebase on top of latest master as CI now runs flake8 on swig/python/osgeo/utils and auxiliary

OK.

idanmiara:
I'm committing the changes, see if it makes sense.

idanmiara:
Finally all tests passes with all these Python 2 compatibility issues...

gdal/swig/python/osgeo/gdal_auxiliary/color_table.py
```diff
@@ -0,0 +1,110 @@
...
+        # color_entry: gdal.ColorEntry
+        color_entry = color_table.GetColorEntry(i)
+        if color_entry != c:
+            return False
+    return True
+
+
+def get_fixed_color_table(c=(0, 0, 0, 0)):
+    color_table = gdal.ColorTable()
+    for i in range(color_table.GetCount()):
```
rouault, line 83: wouldn't color_table.GetCount() be 0 ?

gdal/swig/python/osgeo/auxiliary/color_palette.py
```diff
@@ -0,0 +1,348 @@
...
+import tempfile
+from osgeo.auxiliary import base
+
+# from typing import Sequence, Union
+# from pathlib import Path
+# PathOrStrings = Union[Path, str, Sequence[str]]
+# ColorPaletteOrPathOrStrings = Union['ColorPalette', PathOrStrings]
+
+
+class ColorPalette:
```
rouault, line 45: I'm still missing how a ColorPalette is supposed to be different from a ColorTable ? There would need to be some documentation.
I'm also a bit uneasy to expend too much the GDAL Python API from the C++ core. In particular we have no real proper Python doc and have resorted to pointing to the C++ one, but from pure Python API, we have no such possibility

idanmiara: > I'm still missing how a ColorPalette is supposed to be different from a ColorTable ? There would need to be some documentation.

I needed a more generalized data structure than ColorTable.
ColorTable maps indices to a triplets/quadlets of bytes and is a thin wrapper around the c++ class. If I recall correctly all the table keys needs to be filled up even if unused.
ColorPalette is more generalized, it maps numbers/percentages/others into a 32-bit color.
I convert a ColorPalette into a ColorTable just when I need to set it into a DS.

I was about to document it if this concept is accepted.

> I'm also a bit uneasy to expend too much the GDAL Python API from the C++ core. In particular we have no real proper Python doc and have resorted to pointing to the C++ one, but from pure Python API, we have no such possibility

I see your point. 
I needed to this auxiliary class and a few more to support more functionality into gdal_calc and gdal_viewshed and others.
I think we need to find an approach to allow Python auxiliary modules to support gdal utils.
Once we migrate to Python 3 - documentation would become much easier with type hinting.
Than we could autogenerate documentation for the osgeo.* sub packages.

gdal/swig/python/osgeo/gdal_auxiliary/color_table.py
idanmiara: Right, I'll look into it. Thanks.

gdal/swig/python/osgeo/auxiliary/color_palette.py
idanmiara: @rouault What do you think about this PR?
Does the separation between osgeo.auxiliary and osgeo.gdal_auxiliary makes sense to you? 
What do you think about the ColorPalette class ? I will document if the general idea is accepted.

rouault: > Does the separation between osgeo.auxiliary and osgeo.gdal_auxiliary makes sense to you?

I find this perhaps a bit complicated. That would make 3 subpackages with somewhat vague names: utils, auxiliary, gdal_auxiliary.

idanmiara:
I somewhat agree, but I couldn't come up with better names...
We could do the separation at the module level instead of at the subpackage
level.

gdal/swig/python/osgeo/auxiliary/color_palette.py
idanmiara: I somewhat agree, but I couldn't come up with better names...
We could do the separation at the module level instead of at the subpackage level.

idanmiara: Another idea could be to put the modules that do not depend on osgeo on their own package, outside the osgeo package.
The naming issue still stands - maybe osgeo_aux ? this package could be reused outside gdal without being dependent on gdal.
One could argue to put it outside this repo, but then we couldn't use it as auxiliary for osgeo package as dependence on outside modules is not allowed.

rcoup: Seems very complex to me... is there any _known_ use of these classes outside of GDAL? Just because it doesn't technically depend on it doesn't mean that inside GDAL's namespace isn't the right place for it to live. 

Is there a reason it can't all go into `osgeo.utils`, and if there are big worries about compatibility, put it inside `osgeo.utils.internal` with a warning saying "no compatibility guarantees — if you want it as a public API open a PR" 😄

idanmiara: > Seems very complex to me... is there any _known_ use of these classes outside of GDAL? Just because it doesn't technically depend on it doesn't mean that inside GDAL's namespace isn't the right place for it to live.
> 
> Is there a reason it can't all go into `osgeo.utils`, and if there are big worries about compatibility, put it inside `osgeo.utils.internal` with a warning saying "no compatibility guarantees — if you want it as a public API open a PR" 😄

so we have `osgeo.utils` for all the `scripts` that were transformed into independent modules.
and `osgeo.auxiliary` for the common helper functions that are not functional by themselves.
I could merge `osgeo_aux` into `osgeo.auxiliary` if it lessen the clutter of naming.

Other than that what do you think about this PR?

idanmiara:
> One thing that should be clarified too is if stuff into auxiliary/ is meant to be used by code external to GDAL (and if so, there should be a clear distinction on what is in its API and what is internal details), or if it is purely code to be used by GDAL scripts/ and samples/ . I'd assume the later, but we should probably document this properly

I've made `auxiliary` a submodule of `osgeo.utils` to imply it is for internal use of `osgeo.utils` and no backwards compatibility is guaranteed.
Where would be a good place to document it as well?

gdal/swig/python/osgeo/utils/gdal_calc.py
```diff
@@ -620,7 +622,7 @@ def main(argv):
     parser.add_option("--quiet", dest="quiet", action="store_true", help="suppress progress messages")
     parser.add_option("--optfile", dest="optfile", metavar="optfile", help="Read the named file and substitute the contents into the command line options list.")
 
-    # parser.add_option("--color_table", dest="color_table", help="color table file name")
+    parser.add_option("--color_table", dest="color_table", help="color table file name")
```
rouault, line 625: For consistence with --creation-option:

```suggestion
    parser.add_option("--color-table", dest="color_table", help="color table file name")
```

gdal/doc/source/programs/gdal_calc.rst
```diff
@@ -88,6 +88,11 @@ but no projection checking is performed (unless projectionCheck option is used).
 
     GDAL format for output file.
 
+.. option:: color_table=<filename>
```
rouault, line 91:
```suggestion
.. option:: color-table=<filename>
```

idanmiara:
should I squash or wait for more further review?

idanmiara:
I've tried to force-push to trigger CI rerun, but now it fails on something else...

idanmiara:
@rouault are there more issues to address with this PR?

idanmiara:
thanks!

idanmiara:
thanks!

PR #3147, opened 2020-11-04
GTiff: Export compression tags to metadata

drons:
These tags are useful if we need to rewrite the file.

### DIFF
--- a/gdal/frmts/gtiff/geotiff.cpp
+++ b/gdal/frmts/gtiff/geotiff.cpp
@@ -511,7 +511,9 @@ private:
     int            GuessJPEGQuality( bool& bOutHasQuantizationTable,
                                      bool& bOutHasHuffmanTable );
 
-    void           SetJPEGQualityAndTablesModeFromFile();
+    void           SetJPEGQualityAndTablesModeFromFile(int nQuality,
+                                                       bool bHasQuantizationTable,
+                                                       bool bHasHuffmanTable);
 
     int            DirectIO( GDALRWFlag eRWFlag,
                              int nXOff, int nYOff, int nXSize, int nYSize,
@@ -14591,9 +14593,31 @@ CPLErr GTiffDataset::OpenOffset( TIFF *hTIFFIn,
         }
     }
 
-    if( m_nCompression == COMPRESSION_JPEG && eAccess == GA_Update )
+    if( m_nCompression == COMPRESSION_JPEG )
     {
-        SetJPEGQualityAndTablesModeFromFile();
+        bool bHasQuantizationTable = false;
+        bool bHasHuffmanTable = false;
+        int nQuality = GuessJPEGQuality( bHasQuantizationTable,
+                                         bHasHuffmanTable );
+        if( nQuality > 0 )
+        {
+            m_oGTiffMDMD.SetMetadataItem( "JPEG_QUALITY",
+                                          CPLSPrintf("%d", nQuality),
+                                          "IMAGE_STRUCTURE" );
+            int nJpegTablesMode = JPEGTABLESMODE_QUANT;
+            if( bHasHuffmanTable )
+            {
+                nJpegTablesMode |= JPEGTABLESMODE_HUFF;
+            }
+            m_oGTiffMDMD.SetMetadataItem( "JPEGTABLESMODE",
+                                          CPLSPrintf("%d", nJpegTablesMode),
+                                          "IMAGE_STRUCTURE" );
+        }
+        if( eAccess == GA_Update )
+        {
+            SetJPEGQualityAndTablesModeFromFile(nQuality, bHasQuantizationTable,
+                                                bHasHuffmanTable);
+        }
     }
 
     CPLAssert(m_bReadGeoTransform == bReadGeoTransform);
@@ -16585,12 +16609,10 @@ int GTiffDataset::GuessJPEGQuality( bool& bOutHasQuantizationTable,
 /*               SetJPEGQualityAndTablesModeFromFile()                  */
 /************************************************************************/
 
-void GTiffDataset::SetJPEGQualityAndTablesModeFromFile()
+void GTiffDataset::SetJPEGQualityAndTablesModeFromFile(int nQuality,
+                                                       bool bHasQuantizationTable,
+                                                       bool bHasHuffmanTable)
 {
-    bool bHasQuantizationTable = false;
-    bool bHasHuffmanTable = false;
-    int nQuality = GuessJPEGQuality( bHasQuantizationTable,
-                                     bHasHuffmanTable );
     if( nQuality > 0 )
     {
         CPLDebug("GTiff", "Guessed JPEG quality to be %d", nQuality);


rouault:
What would be the effect of this patch exactly ? To report those tags as metadata (but a number of those tags are pseudo-tags, not stored in the file ? Or to set those tags ? But we have creation options for that

drons:
The main goal is to be able to read these tags from metadata.

rouault:
> The main goal is to be able to read these tags from metadata.

ok, but for a number of them (at least TIFFTAG_ZIPQUALITY, TIFFTAG_LZMAPRESET, TIFFTAG_ZSTD_LEVEL, TIFFTAG_WEBP_LEVEL, TIFFTAG_DEFLATE_SUBCODEC and perhaps others), you will only get the default value, or the value that has been set by GDAL.  For the others that might make sense, they should rather be reported in the IMAGE_STRUCTURE metadata domain rather than the default one

drons:
> they should rather be reported in the IMAGE_STRUCTURE metadata domain rather than the default one
 OK, I'm move those tags to IMAGE_STRUCTURE metadata section

jratike80:
What happens when such image is converted with gdal_translate into another GeoTIFF or into other formats (to have metadata in PAM)? Are the metadata items updated if reasonable, and dropped if they are not reasonable?

drons:
It is possible to obtain only  jpeg compression parameters with new patch.
I deleted an attempt to read parameters for other compression algorithms.


```
gdal_translate -co COMPRESS=JPEG -co JPEG_QUALITY=45 -of GTiff input.tif compressed.tif && \
gdalinfo compressed.tif
```
reported

```
Image Structure Metadata:
  COMPRESSION=JPEG
  INTERLEAVE=PIXEL
  JPEGTABLESMODE=1
  JPEG_QUALITY=45
```

rouault:
GuessJPEGQuality() isn't necessarily cheap as it iterates over quality from 0 to 100, but I've no idea how costly/cheap it is. Could you try to bench opening JPEG compressed files to see if there's a measurable time difference in calling it (like time calling GDALOpen() in a loop) ?

The call to SetJPEGQualityAndTablesModeFromFile() done at line 14596 in update mode could potentially be simplified to avoid guessing again.

drons:
>GuessJPEGQuality() isn't necessarily cheap as it iterates over quality from 0 to 100, but I've no idea how costly/cheap it is. Could you try to bench opening JPEG compressed files to see if there's a measurable time difference in calling it (like time calling GDALOpen() in a loop) ?

 In theory, quality factor can be directly recalculated from quantization table (see jpeg_set_linear_quality & jpeg_add_quant_table from libjpeg).

> The call to SetJPEGQualityAndTablesModeFromFile() done at line 14596 in update mode could potentially be simplified to avoid guessing again.

GuessJPEGQuality is called once for now.

drons:
> if there's a measurable time difference in calling it (like time calling GDALOpen() in a loop)

Time is 90 ms per GDALOpen with GuessJPEGQuality call.
Time is 80 ms without.

rouault:
> Time is 90 ms per GDALOpen with GuessJPEGQuality call.
> Time is 80 ms without.

ok, that's acceptable.

drons:
>GuessJPEGQuality() isn't necessarily cheap as it iterates over quality from 0 to 100, but I've no idea how costly/cheap it is. Could you try to bench opening JPEG compressed files to see if there's a measurable time difference in calling it (like time calling GDALOpen() in a loop) ?

I'm try to compute quality directly from quantization table, but for input quality 46 due to rounding errors value is not equal to original. 

Maybe I missed something... Take a look.

```diff
diff --git a/gdal/frmts/gtiff/geotiff.cpp b/gdal/frmts/gtiff/geotiff.cpp
index 21d5e37c50..252561b1ce 100644
--- a/gdal/frmts/gtiff/geotiff.cpp
+++ b/gdal/frmts/gtiff/geotiff.cpp
@@ -16502,6 +16502,106 @@ static bool GTIFFQuantizationTablesEqual( const GByte* paby1, int nLen1,
     }
 }
 
+constexpr int M_DQT = 0xDB;
+constexpr int M_DHT = 0xC4;
+constexpr int DCTSIZE2 = 64;
+static const int jpeg_natural_order[DCTSIZE2+16] = {
+     0,  1,  8, 16,  9,  2,  3, 10,
+    17, 24, 32, 25, 18, 11,  4,  5,
+    12, 19, 26, 33, 40, 48, 41, 34,
+    27, 20, 13,  6,  7, 14, 21, 28,
+    35, 42, 49, 56, 57, 50, 43, 36,
+    29, 22, 15, 23, 30, 37, 44, 51,
+    58, 59, 52, 45, 38, 31, 39, 46,
+    53, 60, 61, 54, 47, 55, 62, 63,
+    63, 63, 63, 63, 63, 63, 63, 63, /* extra entries for safety in decoder */
+    63, 63, 63, 63, 63, 63, 63, 63
+};
+static const unsigned int std_luminance_quant_tbl[DCTSIZE2] = {
+    16,  11,  10,  16,  24,  40,  51,  61,
+    12,  12,  14,  19,  26,  58,  60,  55,
+    14,  13,  16,  24,  40,  57,  69,  56,
+    14,  17,  22,  29,  51,  87,  80,  62,
+    18,  22,  37,  56,  68, 109, 103,  77,
+    24,  35,  55,  64,  81, 104, 113,  92,
+    49,  64,  78,  87, 103, 121, 120, 101,
+    72,  92,  95,  98, 112, 100, 103,  99
+};
+
+static int ParseQuantizationTable(const GByte* pabyQuantizationTable,
+                                  int nQuantizationTableSize)
+{
+    int nTableBits = 0;
+    // Only 2 table sizes: DCTSIZE2*2 + 1 + 2 , DCTSIZE2 + 1 + 2 (see libjpeg emit_dqt)
+    if(nQuantizationTableSize == DCTSIZE2 + 1 + 2)
+    {
+        nTableBits = 8;
+    }
+    else if(nQuantizationTableSize == DCTSIZE2*2 + 1 + 2)
+    {
+        nTableBits = 16;
+    }
+    else
+    {
+        return -1;
+    }
+
+    //Check if not a luminance quantization table
+    if(pabyQuantizationTable[2] != 0 &&
+       pabyQuantizationTable[2] != (1UL<< 4UL))
+    {
+        return -1;
+    }
+    //Skip table size (2 bytes) & table index (1 byte)
+    pabyQuantizationTable += 3;
+
+    //see jpeg_add_quant_table
+    //temp = ((long) basic_table[i] * scale_factor + 50L) / 100L;
+    double dfMeanScale = 0;
+    int    nValid = 0; 
+    for(int i = 0; i != DCTSIZE2; ++i)
+    {
+        GUInt16 nQuantizationValue;
+        if(nTableBits == 8)
+        {
+            nQuantizationValue = pabyQuantizationTable[i];
+            if(nQuantizationValue <= 1L || nQuantizationValue >= 255L)
+            {
+                continue;//Skip outliers
+            }
+        }
+        else
+        {
+            memcpy(&nQuantizationValue, pabyQuantizationTable + 2*i, 2);
+            nQuantizationValue = CPL_MSBWORD16(nQuantizationValue);
+            if(nQuantizationValue <= 1L || nQuantizationValue >= 32767L)
+            {
+                continue;//Skip outliers
+            }
+        }
+        double dfScale = (nQuantizationValue * 100.0 - 50.0) /
+                          std_luminance_quant_tbl[jpeg_natural_order[i]];
+        dfMeanScale += dfScale;
+        ++nValid;
+    }
+    if(nValid == 0)
+    {
+        return -1;
+    }
+    dfMeanScale /= nValid;
+    // Convert back to quality (see jpeg_quality_scaling)
+    int nQuality;
+    if(dfMeanScale >= 100)
+    {
+        nQuality = static_cast<int>(5000.0 / dfMeanScale);
+    }
+    else
+    {
+        nQuality = static_cast<int>(100.0 - dfMeanScale/2.0);
+    }
+    return nQuality;
+}
+
 int GTiffDataset::GuessJPEGQuality( bool& bOutHasQuantizationTable,
                                     bool& bOutHasHuffmanTable )
 {
@@ -16515,15 +16615,24 @@ int GTiffDataset::GuessJPEGQuality( bool& bOutHasQuantizationTable,
         return -1;
     }
 
-    bOutHasQuantizationTable =
-        GTIFFFindNextTable( static_cast<const GByte*>(pJPEGTable), 0xDB,
-                            nJPEGTableSize, nullptr) != nullptr;
+    int nQuantizationTableSize = 0;
+    const GByte* pabyQuantizationTable =
+        GTIFFFindNextTable( static_cast<const GByte*>(pJPEGTable), M_DQT,
+                            nJPEGTableSize, &nQuantizationTableSize );
+    bOutHasQuantizationTable = pabyQuantizationTable != nullptr;
     bOutHasHuffmanTable =
-        GTIFFFindNextTable( static_cast<const GByte*>(pJPEGTable), 0xC4,
+        GTIFFFindNextTable( static_cast<const GByte*>(pJPEGTable), M_DHT,
                             nJPEGTableSize, nullptr) != nullptr;
+
     if( !bOutHasQuantizationTable )
         return -1;
 
+    int nRet = ParseQuantizationTable(pabyQuantizationTable, 
+                                      nQuantizationTableSize);
+    if(nRet > 0)
+    {
+        return nRet;
+    }
     char** papszLocalParameters = nullptr;
     papszLocalParameters = CSLSetNameValue(papszLocalParameters,
                                             "COMPRESS", "JPEG");
@@ -16543,7 +16652,6 @@ int GTiffDataset::GuessJPEGQuality( bool& bOutHasQuantizationTable,
     osTmpFilenameIn.Printf( "/vsimem/gtiffdataset_guess_jpeg_quality_tmp_%p",
                             this );
 
-    int nRet = -1;
     for( int nQuality = 0; nQuality <= 100 && nRet < 0; ++nQuality )
     {
         VSILFILE* fpTmp = nullptr;
```

rouault:
> Maybe I missed something... Take a look.

You know more than me at that point -:) Not sure which sources and runtime you compare: couldn't that be a libjpeg 6v vs libjpeg-turbo difference ?

drons:
It would probably be better to make a commit. But due to the fact that the result of the calculation in `ParseQuantizationTable` is not accurate, I did not commit.

I just don’t know if it makes sense to further explore the topic of calculating JPEG quality from quantization tables. Especially if the current implementation (via compression to /vsimem/) is fast enough and satisfies everyone. The only thing is, maybe add a new function `ParseQuantizationTable` as a last resort at the end of `GuessJPEGQuality`.

rouault:
I'm happy to merge this PR as it, if you think you're done with it

drons:
> I'm happy to merge this PR as it, if you think you're done with it

I'm ready to merge.

I will continue the research if I find files for which the current implementation of `GuessJPEGQuality` will not work.

drons:
Thanks!

PR #3151, opened 2020-11-05
BYN: relax checks in header bytes to allow products with some unset/invalid fields we don't use

rouault:


m-kuhn:
Confirmed.
Thanks a lot !

PR #3155, opened 2020-11-07
NEWS and MIGRATION_GUIDE.TXT: document removal of deprecated 'import gdal' way of Python bindings. Fixes #3149 [ci skip]

### LINKED ISSUE #3149: Python bindings ogr module removed from gdal 3.2 / Not in release Notes

```python
import ogr
```
gives

```
ModuleNotFoundError: No module named 'ogr'
```

in gdal 3.2.

I know that we should use `from osgeo import ogr` but since I could not find this breaking change in the release notes I was wondering if it was removed by mistake or forgotten in the release notes. 

It still works under gdal 3.1

## Operating system

Arch linux

gdal 3.2 from conda-forge

rouault:


idanmiara:
I guess you should delete:
gdal/swig/python/[gdal,ogr,osr,gdalconst,gdalnumeric].py

rouault:
> I guess you should delete:

Good point. Done

PR #3142, opened 2020-11-03
rfc77_drop_python2_support - RFC 77: Drop Python 2 support in favor of Python 3.6

idanmiara:
## What does this PR do?

Add RFC 77: Drop Python 2 support in favor of Python 3.6

## What are related issues/pull requests?

https://github.com/OSGeo/gdal/issues/3114

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [x] Review
 - [x] Adjust for comments
 - [x] All CI builds and checks have passed

PR #3165, opened 2020-11-10
osgeo/__init__.py - add Python version check and deprecation warning

idanmiara:
## What does this PR do?

This PR prepares the ground for deprecating older Python versions in newer versions of GDAL.
It sets Python 3.6 as the minimum supported Python version for GDAL 3.3 and will raise a deprecation warning if you are using a Python version that will get unsupported in the next version of GDAL.

Setting `fail_on_unsupported_version = False` to True will cause importing osgeo to fail on an unsupported Python version.
Otherwise a deprecation warning will be issued instead.
Importing osgeo fom an unsupported Python version might still partially work
because the core of GDAL Python bindings might still support an older Python version.
Hence the default option to just issue a warning.
To get complete functionality upgrading to the minimum supported version is needed.

Currently, When setting `fail_on_unsupported_version = True` for GDAL 3.3 all the CI builds fail because they all use Python 2.7 or 3.5 and will need to be modified to use Python 3.6 - I think we should set this flag to True in another branch aside master that when we ready to drop Python2 support we will merge into master.

`gdal_version_and_min_supported_python_version = (
    ((3, 2), (2, 0)),
    ((3, 3), (3, 6)),
)`
is a Sequence of tuples in the form of (gdal_version, python_version).
Each line represents the minimum supported Python version of a given GDAL version.
Introducing a new line for the next GDAL version will trigger a deprecation warning
when importing osgeo if it is imported from a Python version which will not be
supported in the next version of GDAL.

https://github.com/OSGeo/gdal/pull/3142

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/gdal/swig/python/osgeo/__init__.py
+++ b/gdal/swig/python/osgeo/__init__.py
@@ -113,7 +113,7 @@ for gdal_ver, py_ver in gdal_version_and_min_supported_python_version:
 if python_version < minimum_supported_python_version_for_this_gdal_version:
     msg = 'Your Python version is {}, which is no longer supported by GDAL {}. ' \
           'Please upgrade your Python version to Python >= {}, ' \
-          'or use GDAL <= {} >, which supports your Python version.'.\
+          'or use GDAL <= {}, which supports your Python version.'.\
         format(ver_str(python_version), ver_str(gdal_version),
                ver_str(minimum_supported_python_version_for_this_gdal_version),
                ver_str(last_gdal_version_to_supported_your_python_version))
--- a/gdal/swig/python/osgeo/__init__.py
+++ b/gdal/swig/python/osgeo/__init__.py
@@ -66,3 +66,74 @@ else:
     import _gdal
 
 __version__ = _gdal.__version__ = _gdal.VersionInfo("RELEASE_NAME")
+
+gdal_version = tuple(int(s) for s in str(__version__).split('.') if s.isdigit())[:3]
+python_version = tuple(version_info)[:3]
+
+# Setting this flag to True will cause importing osgeo to fail on an unsupported Python version.
+# Otherwise a deprecation warning will be issued instead.
+# Importing osgeo fom an unsupported Python version might still partially work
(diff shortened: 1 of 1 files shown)


rouault:
I'm not super familiar with how/when python displays deprecation warnings, but for example looking at the log of https://api.travis-ci.com/v3/job/440760948/log.txt which ran python 2.7 (that was before my github action migration), I can't see any warning. Is it because of how pytest runs ?

rouault:
This PR should be rebased on top of latest master to make sure everything is fine with the github action migration

idanmiara:
> This PR should be rebased on top of latest master to make sure everything is fine with the github action migration

done, thanks!

idanmiara:
> I'm not super familiar with how/when python displays deprecation warnings, but for example looking at the log of https://api.travis-ci.com/v3/job/440760948/log.txt which ran python 2.7 (that was before my github action migration), I can't see any warning. Is it because of how pytest runs ?

I'm not sure.
I replaced the __init__.py of the gdal package on my python 2.7 on windows with this version of __init__.py, and it does show the warning when I ran `from osgeo import gdal`.

idanmiara:
> > I'm not super familiar with how/when python displays deprecation warnings, but for example looking at the log of https://api.travis-ci.com/v3/job/440760948/log.txt which ran python 2.7 (that was before my github action migration), I can't see any warning. Is it because of how pytest runs ?
> 
> I'm not sure.
> I replaced the `__init__.py` of the gdal package on my python 2.7 on windows with this version of `__init__.py`, and it does show the warning when I ran `from osgeo import gdal`.

You can see the deprecation warning here:
[](https://ci.appveyor.com/project/OSGeo/gdal/builds/36530941)

idanmiara:
I guess the failing `test_ngw_3` is not related to this PR...

rouault:
Backported to 3.2 per 583d7a7471a3695246080abf7dd3dd18396013b8

Issue #3162, opened 2020-11-09
Poor Quality in gdal2tiles.py compared to MapTiler


markhalonen:
## Expected behavior and actual behavior.

For the attached PDF, gdal2tiles.py renders tiles with quality such as this: (`17/33136/46115.png`):
![image](https://user-images.githubusercontent.com/7853605/98591845-d5ad1780-229e-11eb-9041-eeff8ffb57b2.png)

It is not as good as MapTiler output:

![image](https://user-images.githubusercontent.com/7853605/98591873-e2ca0680-229e-11eb-87a9-3fffa0a4a629.png)

## Steps to reproduce the problem.
1. Download [916-0013_Harvest_Map.pdf](https://github.com/OSGeo/gdal/files/5513103/916-0013_Harvest_Map.pdf)
2. Run `mkdir out_tiles && gdal2tiles.py --xyz --zoom=12-18 --resampling=cubic 916-0013_Harvest_Map.pdf out_tiles`

`

## Operating system

I am running from my own Dockerfile on Mac that uses `FROM osgeo/gdal:alpine-normal-latest`

## GDAL version and provenance

`FROM osgeo/gdal:alpine-normal-latest`

I have tried `average` `bilinear` and `cubic` sampling methods to no avail. 

I would pay MapTiler for their service, but sadly their command line app fails on my PDF. Their good output is from the Desktop app but I need a scripted version. 

If anyone has any suggestions of options to tweak, I'd be happy to hear them! 

This issue is perhaps similar to https://github.com/OSGeo/gdal/issues/1087

markhalonen:
Ok I found the option `--config GDAL_PDF_DPI 1200`, I saw it mentioned in https://github.com/OSGeo/gdal/issues/2950

It seems much slower, but I can probably figure it out from here. I did not see this option in the `gdal2tiles.py`, I have never used this software so perhaps it's common knowledge that there are some general config values available. 

Maintainers can close this issue unless they want to take action on it

jratike80:
[...]

PR #3166, opened 2020-11-10
Update gdal2tiles.rst - Explicitly mention common config options

markhalonen:
Documentation fix for https://github.com/OSGeo/gdal/issues/3162.

It was mentioned there that 

> There is an initiative to collect all configuration options into one page https://gdal.org/user/configoptions.html but the page is far from perfect. For example GDAL_PDF_DPI is still missing,

But it is currently present when I visit that page

### DIFF
--- a/gdal/doc/source/programs/gdal2tiles.rst
+++ b/gdal/doc/source/programs/gdal2tiles.rst
@@ -21,7 +21,7 @@ Synopsis
                   [-w webviewer] [-t title] [-c copyright]
                   [--processes=NB_PROCESSES] [--xyz]
                   --tilesize=PIXELS
-                  [-g googlekey] [-b bingkey] input_file [output_dir]
+                  [-g googlekey] [-b bingkey] input_file [output_dir] [COMMON_OPTIONS]
 
 Description
 -----------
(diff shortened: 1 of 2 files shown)


gdal/doc/source/programs/gdal2tiles.rst
```diff
@@ -44,6 +44,10 @@ can publish a picture without proper georeferencing too.
     Inputs with non-Byte data type (i.e. ``Int16``, ``UInt16``,...) will be clamped to
     the ``Byte`` data type, causing wrong results. To awoid this it is necessary to
     rescale input to the ``Byte`` data type using `gdal_translate` utility.
+    
+.. note::
+
+    Config options of the input drivers may have an effect on the output of gdal2tiles. An example driver config option is GDAL_PDF_DPI, which can be found at https://gdal.org/user/configoptions.html
```
rouault, line 50:
```suggestion
    Config options of the input drivers may have an effect on the output of gdal2tiles. An example driver config option is GDAL_PDF_DPI, which can be found at :ref:`configoptions`
```

rouault:
> But it is currently present when I visit that page

was added few minutes ago per https://github.com/OSGeo/gdal/pull/3164

PR #3180, opened 2020-11-13
VRT: preserve source block size

rouault:
Assorted set of improvements that help for example to gdal_translate a DIMAP JPEG2000 dataset (read by JP2KAK) to tiled GeoTIFF

### DIFF
--- a/gdal/data/gdalvrt.xsd
+++ b/gdal/data/gdalvrt.xsd
@@ -210,10 +210,9 @@
         </xs:sequence>
         <xs:attribute name="dataType" type="DataTypeType"/>
         <xs:attribute name="band" type="xs:unsignedInt"/>
+        <xs:attribute name="blockXSize" type="nonNegativeInteger32"/>
+        <xs:attribute name="blockYSize" type="nonNegativeInteger32"/>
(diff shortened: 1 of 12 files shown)


gdal/data/gdalvrt.xsd
```diff
@@ -210,6 +210,8 @@
         </xs:sequence>
         <xs:attribute name="dataType" type="DataTypeType"/>
         <xs:attribute name="band" type="xs:unsignedInt"/>
+        <xs:attribute name="blockXSize" type="xs:unsignedInt"/>
+        <xs:attribute name="blockYSize" type="xs:unsignedInt"/>
         <xs:attribute name="subClass" type="VRTRasterBandSubClassType"/>
 
         <xs:attribute name="BlockXSize" type="xs:unsignedInt"/> <!-- ignored -->
```
tbonfort, line 217: is it intentional to add blockXSize when BlockXSize is already present?

rouault: ah, good catch. I guess I put them there as a way of reserving them for the future. I've just removed that reservation, and adopt the syntax with leading letter in underscore for consistency with other attributes

PR #3181, opened 2020-11-13
GDALOpenEx(): supports OVERVIEW_LEVEL=NONE to indicate overviews shouldn't be exposed

rouault:
Adresses https://rasterio.groups.io/g/main/message/649

### DIFF
--- a/autotest/gcore/overviewds.py
+++ b/autotest/gcore/overviewds.py
@@ -42,8 +42,6 @@ import pytest
 
 
 def test_overviewds_1():
-    ds = gdal.OpenEx('data/byte.tif', open_options=['OVERVIEW_LEVEL=-1'])
-    assert ds is None
     ds = gdal.OpenEx('data/byte.tif', open_options=['OVERVIEW_LEVEL=0'])
     assert ds is None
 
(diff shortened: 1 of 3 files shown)


sgillies:
@rouault @loicdtx the tests we added to rasterio to check on this have never passed, so I think there may be a GDAL usage problem in rasterio or a bug in GDAL. It hasn't been completely clear to me whether one passes `OVERVIEW_LEVEL=-1` or `OVERVIEW_LEVEL=NONE` as a dataset opening option. Rasterio's test fails in both cases.

The rasterio test (see https://github.com/mapbox/rasterio/blob/master/tests/test_overviews.py#L137) starts with a test dataset that is red, copies a green TIFF of the same size to red.tif.ovr and then expects to not have the green overview pixels appear when passing the "no overviews" flag.

rouault:
> It hasn't been completely clear to me whether one passes `OVERVIEW_LEVEL=-1` or `OVERVIEW_LEVEL=NONE` as a dataset opening option

Both should work. I added NONE as an alias for -1, as the later is a bit awkward

> Rasterio's test fails in both cases.

Hum, here's what I suspect. When one uses OVERVIEW_LEVEL, a GDALOverviewDataset that wraps the original TIFF dataset is returned. That GDALOverviewDataset and its bands won't expose overviews to callers, but at some point of the RasterIO() request, you get into the original TIFF dataset, which might potentially have access to its internal/external overviews, and use them. There might be some subtelties in the different rasterio test case calls GDAL API compared to the unit test of this pull request. Ideally if there was some GDAL Python way of reproducing the issue, that could help.

jratike80:
In this case the overview TIFF is a copy of the main image and the first overview level presents the full resolution. That is at least uncommon scenario. Could that lead into some confusion?

rouault:
> If I read right in this case the overview TIFF is like a copy of the main image and the first overview level presents the full resolution. That is at least uncommon scenario. Could that lead into some confusion?

you got probably confused by the setup of the rasterio test, but it is a "normal" situation in fact with a 64x64 .tif file, and a .tif.ovr with 32x32 and 16x16 raster. I can reproduce with GDAL Python bindings. Investigating

sgillies:
Rasterio's test is passing with the primary GDAL branch. Thanks, @rouault !

PR #2958, opened 2020-09-22
WMTS: Map CURLE_FILE_COULDNT_READ_FILE  to 404 (fixes #2941)

janosimas:
## What does this PR do?
Identify local wmts tiles not found with CURLE_FILE_COULDNT_READ_FILE and map to http status 404

## What are related issues/pull requests?

fixes #2941

## Tasklist

 - [x] Map CURLE_FILE_COULDNT_READ_FILE  to 404
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments
 - [ ] All CI builds and checks have passed

## Environment

 Built and tested on windows.

PR #3187, opened 2020-11-17
Jano Simas WMTS: Map CURLE_FILE_COULDNT_READ_FILE to 404

### LINKED ISSUE #2941: WMTS: local files do not accept ZeroBlockHttpCodes

I'm consuming a local WMTS server structure, for some technical reasons some tiles may be missing.

Local WMTS with missing tiles will generate and error and a blank image.
I expected it to  behave like online WMTS, an image with missing areas.

Draw a local WMTS with missing tiles.

## Operating system

The problem is system independent

Build from source 2.4.0

## Proposed solution

From the documentation, I understand the `ZeroBlockHttpCodes` is used in this cases, eg.: 404 - file not found in a online server.

But for local files, `curl` does not set a Http Response code, currently gdal already goes around it and when no error message is set, [it sets the response code to 200](https://github.com/OSGeo/gdal/blob/master/gdal/frmts/wms/gdalhttp.cpp#L224).

A solution would be map the curl error `CURLE_FILE_COULDNT_READ_FILE` to a gdal reponse code `404`.

*In case we agree upon the solution I plan to make a pull request to the current version.*

janosimas:
## What does this PR do?
Identify local wmts tiles not found with CURLE_FILE_COULDNT_READ_FILE and map to http status 404

fixes #2941
older discussion #2958

## Tasklist

 - [x] Map CURLE_FILE_COULDNT_READ_FILE  to 404
 - [x] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

 Built on windows.
 Built and tested on linux (ubuntu/docker(

### DIFF
--- a/autotest/gdrivers/wmts.py
+++ b/autotest/gdrivers/wmts.py
@@ -1807,15 +1807,6 @@ def test_wmts_24():
 </Capabilities>""")
 
     ds = gdal.Open('WMTS:/vsimem/wmts_missing_local_tiles.xml')
-    assert ds is not None
-    assert ds.RasterXSize == 2097152
-    assert ds.RasterYSize == 2097152
-    got_gt = ds.GetGeoTransform()
-    expected_gt = (-548576.0, 1.0000000000004, 0.0, 8388608.0, 0.0, -1.0000000000004)
-    for i in range(6):
-        assert got_gt[i] == pytest.approx(expected_gt[i], abs=1e-8)
-    assert ds.GetProjectionRef().find('3067') >= 0
-
 #   Read some data from the image
     band = ds.GetRasterBand(1)
     assert band is not None
--- a/autotest/gdrivers/wmts.py
+++ b/autotest/gdrivers/wmts.py
@@ -36,6 +36,7 @@ from osgeo import gdal
 
 import gdaltest
 import pytest
+import struct
 
 ###############################################################################
 # Find WMTS driver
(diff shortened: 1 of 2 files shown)


janosimas:
I couldn't reopen the pull request: #2958

I was able to add a test and test in a ubuntu/docker env

autotest/gdrivers/wmts.py
```diff
@@ -1759,7 +1760,70 @@ def test_wmts_check_no_overflow_zoom_level():
...
+                <MatrixHeight>8192</MatrixHeight>
+            </TileMatrix>
+        </TileMatrixSet>
+    </Contents>
+    <ServiceMetadataURL xlink:href="/vsimem/wmts_missing_local_tiles.xml"/>
+</Capabilities>""")
+
+    ds = gdal.Open('WMTS:/vsimem/wmts_missing_local_tiles.xml')
+    assert ds is not None
+    assert ds.RasterXSize == 2097152
```
rouault, line 1811: those initial checks coming from test_wmts_22() could be removed, as they are redundant and don't actually test any change due to this pull request. Just keep what is below "#Read some data from the image"

rouault (commented):
would you mind also rebasing on top of very latest master? I just merged changed in the continuous integration configuration that were needed to make builds work again

PR #3190, opened 2020-11-17
ubuntu_20.04.yml: make use of ccache

rouault:


rouault:
First job without cache : 47 min : https://github.com/OSGeo/gdal/pull/3190/checks?check_run_id=1413610165
Re-run with cache: 22 min (18 minute being running tests): https://github.com/OSGeo/gdal/runs/1413910529

PR #3206, opened 2020-11-20
NETCDF: handle case where netCDF4 has an HDF5 signature at a non-zero…

### LINKED ISSUE #3188: Support netCDF4 files with HDF5 signature at non-zero offset

The HDF5 file format (and by extension netCDF4) allows the signature string `"\211HDF\r\n\032\n"` to appear at certain non-zero offsets in the file [(512, 1024, and all subsequent powers of 2)](https://support.hdfgroup.org/HDF5/doc/H5.format.html#Superblock).

For the case of netCDF4 on top of HDF5 format, It looks like GDAL is hard-coded to expect the HDF5 signature at the start of the file, so won't properly detect such files.  For example, the file [2010020100_024.nc](https://github.com/OSGeo/gdal/files/5554559/2010020100_024.nc.zip) is a netCDF4 file (on HDF5) with the signature at offset 32768.  The command-line utilities `h5dump` and `ncdump` properly handle the file, but `gdalinfo` does not recognize the file.

`gdalinfo 2010020100_024.nc
`

## Operating system

Ubuntu 18.04 64 bit

Tested with GDAL 2.2.3 from the Ubuntu repository, and also tested with recent master branch from GitHub (068065a662c8d1d886037a88e39dc83155fa209b) compiled with `--with-netcdf` option and libnetcdf / libhdf5 packages installed from the Ubuntu repository.

**Edit 2020/11/20:** Looks like GDAL **does** handle vanilla HDF5 files with non-zero offsets, it's the particular case of netCDF4 on HDF5 that is having problems.  Updated the issue accordingly.
**Edit 2020/11/20:** Gah!... vanilla HDF5 with offset only worked because of a patch on the netCDF side I was testing out.  Both netCDF4 and plain HDF5 don't work with signature offsets in GDAL...

neishm:
… byte offset (fixes #3188).

As a last resort, attempt to open the file with the netCDF library and then
(if successful), query the file to ensure it's really encoded as netCF4.

## What does this PR do?
Allows the GDAL interface to recognize netCDF4 files that have an (HDF5) signature somewhere other than the beginning of the file.  For instance, the signature can appear at byte offsets 512, 1024, 2048, and all subsequent power-of-two offsets.  See [format spec](https://support.hdfgroup.org/HDF5/doc/H5.format.html#Superblock) for HDF5.

#3188 

## Tasklist

 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/gdal/frmts/netcdf/netcdfdataset.cpp
+++ b/gdal/frmts/netcdf/netcdfdataset.cpp
@@ -5748,6 +5748,33 @@ CPL_UNUSED
 #endif
     }
 
+    // As a last resort, try opening the file through libnetcdf
+    // and let the library determine if it's a valid file.
+    // For instance, this will handle the case where the netCDF4 file is
+    // encoded as HDF5, and the signature is not at the start of the file.
+
+    // Only do this check for specific file extensions.
+    // Some formats like sqlite3 do not like having their files opened
+    // behind their back (see test_ogr_gpkg_wal for example).
+    const char *pszExtension = CPLGetExtension(poOpenInfo->pszFilename);
+    if( (EQUAL(pszExtension, "nc") || EQUAL(pszExtension, "cdf") ||
+          EQUAL(pszExtension, "nc2") || EQUAL(pszExtension, "nc4") ||
+          EQUAL(pszExtension, "nc3") || EQUAL(pszExtension, "grd")) )
+    {
+        int ncid, fmtid, inq_nc_props;
+        if (nc_open(poOpenInfo->pszFilename, 0, &ncid) == NC_NOERR)
+        {
+            // Only detect netCDF4 for now.
+            inq_nc_props = nc_inq_att (ncid, NC_GLOBAL, "_NCProperties", nullptr, nullptr);
+            nc_inq_format (ncid, &fmtid);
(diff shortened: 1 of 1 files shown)


gdal/frmts/netcdf/netcdfdataset.cpp
```diff
@@ -5748,6 +5748,33 @@ CPL_UNUSED
...
+    // As a last resort, try opening the file through libnetcdf
+    // and let the library determine if it's a valid file.
+    // For instance, this will handle the case where the netCDF4 file is
+    // encoded as HDF5, and the signature is not at the start of the file.
+
+    // Only do this check for specific file extensions.
+    // Some formats like sqlite3 do not like having their files opened
+    // behind their back (see test_ogr_gpkg_wal for example).
+    const char *pszExtension = CPLGetExtension(poOpenInfo->pszFilename);
+    if( (EQUAL(pszExtension, "nc") || EQUAL(pszExtension, "cdf") ||
```
rouault, line 5760: I would drop checking nc2, nc3 and grd which are unlikely to be netCDF4.

rouault:
Superseded by https://github.com/OSGeo/gdal/pull/3211

Issue #3196, opened 2020-11-19
resampling with RMS instead of AVERAGE ?


oarcher:
I use gdal from python, using rasterio.
I work on SAR data with SAFE driver, but I think the problem should be the same with RS2 or ALOS driver.

The information stored in the tiff file is a digital number (DN), and the calibration algorithm use DN**2, so using the AVERAGE resampling method introduce an underestimate of the variable (here sigma0) after applying calibration:
```python
gdal_dn = ds.read(out_shape=out_shape, resampling=Resampling.average)
gdal_sigma0 = apply_calibration(gdal_dn**2)
plt.scatter(gdal_sigma0 , ref_sigma0) # ref_sigma0 was computed from dn at full res, so no errors due to mean(dn)
```
![image](https://user-images.githubusercontent.com/4264660/99604289-66a19280-2a05-11eb-85e4-d5ba501f1d2c.png)

(the spread is due to the fact that the size of the resampling windows are not exactly the same)

The underestimate of sigma0 is because
`mean(dn) <= sqrt(mean(dn**2))`

I've tried to modify [GDALResampleChunk32R_AverageT](https://github.com/OSGeo/gdal/blob/48f4517ffa591808ef8ba76b622eabed2d8ec0d9/gdal/gcore/overview.cpp#L561-618) to compute the square root of the square mean, and the result is much better:

![image](https://user-images.githubusercontent.com/4264660/99604221-47a30080-2a05-11eb-858e-d431f1e71d03.png)

But implementing a full SQUAREAVERAGE resampling method is out of my GDAL code knowledge.

GDAL is very fast for resampling, and many science can be done at lower resolution.  It would be fantastic to have an SQUAREAVERAGE method.

piyushrpt:
It might be possible to implement this using VRTs and pixel functions.

DN -> intensity_VRT using pixel functions -> average_VRT using average resampling -> sqrt_VRT using pixel functions -> translate to raster

avalentino:
@oarcher IMHO this is a very interesting use case for SAR users.
The solution proposed by @piyushrpt is very clear and ready to use.
I would be interested to know of you can observe any time and/or quality performance difference between your "custom" implementation and the solution based on pixel functions.
In the case, it could make sense to submit a PR.

piyushrpt:
[...]

PR #3210, opened 2020-11-23
add RMS (Quadratic Mean) subsampling  to overviews and warp kernel

mgoacolou:
##  What does this PR do?

This add a new average algorithm witch, for SAR product make more visible small brillant object in high level overviews
It also add this new algo to warp resempling method list 

it also address #3196 

## Tasklist

 - [x] Add test case(s)
 - [x] Add documentation
 - [x] Review
 - [x] Adjust for comments

## Environment

* OS: All
* Compiler: All


gdal/gcore/gdal.h
```diff
@@ -133,6 +133,7 @@ typedef enum
     /*! Cubic B-Spline Approximation (4x4 kernel) */    GRIORA_CubicSpline = 3,
     /*! Lanczos windowed sinc interpolation (6x6 kernel) */ GRIORA_Lanczos = 4,
     /*! Average */                                      GRIORA_Average = 5,
+    /*! Average */                                      GRIORA_AverageQuadratic = 13,
```
rouault, line 136: comment to edit, and value to set to 14, so we have a match between the GDALResampleAlg enum  and this one. Comment a bit below "NOTE : values 8 to 12..." should be edited to be "NOTE : values 8 to 13 are reserved for max,min,med,Q1,Q3,sum"

gdal/swig/include/gdal.i
```diff
@@ -209,6 +209,7 @@ typedef enum
     /*! Cubic B-Spline Approximation */     GRIORA_CubicSpline = 3,
     /*! Lanczos windowed sinc interpolation (6x6 kernel) */ GRIORA_Lanczos = 4,
     /*! Average */                          GRIORA_Average = 5,
+    /*! Average */                          GRIORA_AverageQuadratic = 13,
```
rouault, line 212:
```suggestion
    /*! Average */                          GRIORA_AverageQuadratic = 14,
```

gdal/swig/include/gdal.i
```diff
@@ -223,6 +224,7 @@ typedef enum {
   /*! Cubic B-Spline Approximation (4x4 kernel) */     GRA_CubicSpline=3,
   /*! Lanczos windowed sinc interpolation (6x6 kernel) */ GRA_Lanczos=4,
   /*! Average (computes the average of all non-NODATA contributing pixels) */ GRA_Average=5,
+  /*! Average (computes the average of all non-NODATA contributing pixels) */ GRA_AverageQuadratic=13,
```
rouault, line 227: comment to fix and value to set to 14

gdal/doc/source/development/rfc/rfc51_rasterio_resampling_progress.rst
```diff
@@ -105,6 +105,7 @@ The following resampling algorithms are available :
        /*! Cubic B-Spline Approximation (4x4 kernel) */    GRIORA_CubicSpline = 3,
        /*! Lanczos windowed sinc interpolation (6x6 kernel) */ GRIORA_Lanczos = 4,
        /*! Average */                                      GRIORA_Average = 5,
+       /*! Average */                                      GRIORA_AverageQuadratic = 13,
```
rouault, line 108: this addition can be removed. The RFC represents the state of the code at the time it was written

gdal/alg/gdalwarper.h
```diff
@@ -61,7 +61,8 @@ typedef enum {
   /*! Med (selects median of all non-NODATA contributing pixels) */ GRA_Med=10,
   /*! Q1 (selects first quartile of all non-NODATA contributing pixels) */ GRA_Q1=11,
   /*! Q3 (selects third quartile of all non-NODATA contributing pixels) */ GRA_Q3=12,
-  /*! Sum (computes the weighed sum of all non-NODATA contributing pixels). Added in GDAL 3.1 */ GRA_Sum=13
+  /*! Sum (computes the weighed sum of all non-NODATA contributing pixels). Added in GDAL 3.1 */ GRA_Sum=13,
+  /*! Average (computes the weighted quadratic average of all non-NODATA contributing pixels) */ GRA_AverageQuadratic=14
```
rouault, line 65: maybe add the word RMS in the comment since it is prefered by other people ?

rouault (changes requested):
a few test(s) would be needed

gdal/alg/gdalwarper.h
mgoacolou: May be it should me more clear if i rename all AVERAGE_QUADRATIC with RMS?

rouault: > May be it should me more clear if i rename all AVERAGE_QUADRATIC with RMS?

I'm fine with that. I see in https://en.wikipedia.org/wiki/Root_mean_square that "quadratic mean" seems to be an accepted term, so maybe in the doc of the enum mentions it too

gdal/alg/gdalwarpkernel.cpp
```diff
@@ -6102,6 +6110,64 @@ static void GWKAverageOrModeThread( void* pData)
...
+                            {
+                                const double dfWeight = COMPUTE_WEIGHT(iSrcX, dfWeightY);
+                                dfTotalWeight += dfWeight;
+                                if (bIsComplex)
+                                {
+                                    dfTotalReal += (dfValueRealTmp * dfValueRealTmp - dfValueImagTmp *dfValueImagTmp) * dfWeight;
+                                    dfTotalImag += 2. * dfValueRealTmp * dfValueImagTmp * dfWeight;
+                                }
+                                else
+                                    dfTotalReal += dfValueRealTmp * dfWeight;
```
avalentino, line 6150: There is a particular reason why you do not perform the RMS also in the case of real bands?

avalentino (changes requested):

gdal/alg/gdalwarper.h
mgoacolou: Done

jratike80:
There seems to be a mixture of terms RMS, quadratic mean and quadratic avegage in the changed files. Is that intentional?

gdal/gcore/gdal.h
```diff
@@ -133,10 +133,11 @@ typedef enum
...
     /*! Lanczos windowed sinc interpolation (6x6 kernel) */ GRIORA_Lanczos = 4,
     /*! Average */                                      GRIORA_Average = 5,
+    /*! RMS (Quadratic Mean) */                         GRIORA_RMS = 14,
     /*! Mode (selects the value which appears most often of all the sampled points) */
                                                         GRIORA_Mode = 6,
     /*! Gauss blurring */                               GRIORA_Gauss = 7,
-    /* NOTE: values 8 to 12 are reserved for max,min,med,Q1,Q3 */
+    /* NOTE: values 8 to 13 are reserved for max,min,med,Q1,Q3,sum */
 /*! @cond Doxygen_Suppress */
                                                         GRIORA_LAST = GRIORA_Gauss
```
mgoacolou, line 142: Hi @rouault,

I have an issue with this. there is a check that not allow using algo upper than GRIORA_LAST, but GRIORA_RMS = 14.
the only solution that I see is to change enum values of max,min,med,Q1,Q3,sum from 8~13 to 9~14 and move rms to 8.
is there an other way to do?

rouault: > the only solution that I see is to change enum values of max,min,med,Q1,Q3,sum

no, that would change the ABI. 
GRIORA_LAST should be updated to GRIORA_RMS.

mgoacolou: OK, 
should I check that checks with GRIORA_LAST should also excludes values from 8 to 13 because this values were excluded from valid values?

rouault: > should I check that checks with GRIORA_LAST should also excludes values from 8 to 13 because this values were excluded from valid values?

that would be good

mgoacolou: they are not managed as valid in overview.cpp
```
GDALResampleFunction GDALGetResampleFunction( const char* pszResampling,
                                              int* pnRadius )
```

rouault: > they are not managed as valid in overview.cpp
> 
> ```
> GDALResampleFunction GDALGetResampleFunction( const char* pszResampling,
>                                               int* pnRadius )
> ```

do you mean max,min,med,Q1,Q3,sum ? Yes, they are not valid RasterIO() resampling algorithm. Only for gdalwarp. And gdalwarp doesn't use this

gdal/frmts/daas/daasdataset.cpp
```diff
@@ -2222,6 +2222,10 @@ CPLErr GDALDAASRasterBand::GetBlocks(int nBlockXOff, int nBlockYOff,
     {
         oTargetModel.Add("sampling-algo", "AVERAGE");
     }
+    else if( poGDS->m_eCurrentResampleAlg == GRIORA_RMS )
+    {
+        oTargetModel.Add("sampling-algo", "RMS");
+    }
```
mgoacolou, line 2228: @rouault 
I'm not shure if it possible while I don't know if it exist in DAAS!
should I remove this value from DAAS driver?

rouault: > should I remove this value from DAAS driver?

yes, please revert that change. I'm pretty sure the DAAS API doesn't support this

gdal/alg/gdalwarpkernel.cpp
```diff
@@ -6102,6 +6110,64 @@ static void GWKAverageOrModeThread( void* pData)
...
+                        }
+                    }
+
+                    if( dfTotalWeight > 0 )
+                    {
+                        if (bIsComplex)
+                        {
+                            dfR = dfTotalReal / dfTotalWeight;
+                            dfI = dfTotalImag / dfTotalWeight;
+                            dfValueReal = sqrt((dfR + sqrt(dfR * dfR + dfI * dfI))/2.);;
```
rouault, line 6161: for the complex case, I've had feedback from @archer that it would be better to compute the RMS complex value whose real part would be the RMS of the real part and imaginary part the RMS of the imaginar part. The resulting value RMS = RMS(Real) + i * RMS(Imag) would be such that abs(RMS) =RMS(abs(X)) because RMS(abs(X))^2 = mean(abs(X)^2) = mean(abs(Real + i * Imag)^2) = mean(Real^2 + Imag^2) = mean(Real^2) + mean(Imag^2) = RMS(Real)^2 + RMS(Imag)^2

rouault:
Regarding testing, I'd like a test similar to test_rasterio_average_4by4_to_3by3 in autotest/gcore/rasterio.py , that is a test that operates on a very small raster, and where output values are checked individually rather than a checksum. This enables to check manually that the maths are correct.

gdal/gcore/gdal.h
mgoacolou: ok
I created two new values in enum GRIORA_RESERVED_START=8 and GRIORA_RESERVED_END=13 to be able correctly manage the hole in valid values

mgoacolou:
I added test as asked in rasterio.py.

gdal/alg/gdalwarpkernel.cpp
ghajduch: I confirm that from a user perspective the amplitude and phase must be decoupled.
For the amplitude, the need is on RMS(abs(X)) and the RMS(X) is useless.
For the phase,  the proposed formula will allow to compute Phase(RMS) as the phase of RMS(Real) + i * RMS(Phase). I think that this makes less sense than Phase(AVG), but user can already compute this with the standard kernel.

gdal/gcore/overview.cpp
```diff
@@ -2757,6 +2809,48 @@ GDALResampleChunkC32R( int nSrcWidth, int nSrcHeight,
...
+                    pafDstScanline[iDstPixel*2+1] = 0.0;
+                }
+                else
+                {
+                    /* compute mean */
+                    dfTotalR = dfTotalR/nCount;
+                    dfTotalI = dfTotalI/nCount;
+                    /* compute square root of the complex mean */
+                    dfR = sqrt((dfTotalR + sqrt(dfTotalR * dfTotalR + dfTotalI * dfTotalI))/2.);
+                    dfI = dfTotalI / fabs(dfTotalI) * sqrt((-dfTotalR + sqrt(dfTotalR * dfTotalR + dfTotalI * dfTotalI)) / 2.);
```
mgoacolou, line 2848: @rouault 
computation of RMS for complex should be the same as gdal/alg/gdalwarpkernel.cpp here, should it?

rouault: yes, we should use the same method in both cases.

rouault:
You will need to rebase this PR on top of latest master because of broken security updates in Ubuntu 16.04 and 18.04 that I've addressed per https://github.com/OSGeo/gdal/pull/3235

gdal/alg/gdalwarpkernel.cpp
mgoacolou: Done

gdal/gcore/overview.cpp
mgoacolou: Done

gdal/gcore/gdal.h
mgoacolou: I think it should be nice and useful to add missing resampling algorithm to RasterIO()
I could add them in other PR,
what about that @rouault ?

mgoacolou:
I have some issue on builds but I don't know haw to address them there are all related to this
```
--2020-11-26 20:42:53--  http://security.ubuntu.com/ubuntu/pool/main/p/poppler/libpoppler58_0.41.0-0ubuntu1.14_amd64.deb
Resolving security.ubuntu.com (security.ubuntu.com)... 91.189.91.39, 91.189.91.38, 91.189.88.152, ...
Connecting to security.ubuntu.com (security.ubuntu.com)|91.189.91.39|:80... connected.
HTTP request sent, awaiting response... 404 Not Found
2020-11-26 20:42:53 ERROR 404: Not Found.

Error: Process completed with exit code 8.
```

rouault:
> I have some issue on builds but I don't know haw to address them there are all related to this

Ah, this is due to Ubuntu having fixed their broken packages, but apparently they keep only the current version and the previous one, so the N-2 version I was pinning is no longer present... Should be fixed per https://github.com/OSGeo/gdal/pull/3238 . You'll have to rebase on master again once this #3238 is merged

gdal/gcore/gdal.h
rouault: > I could add them in other PR,

That could make sense

rouault:
> You'll have to rebase on master again once this #3238 is merged

#3238 is now merged. you can rebase on latest master

gdal/alg/gdalwarpkernel.cpp
```diff
@@ -6156,10 +6154,8 @@ static void GWKAverageOrModeThread( void* pData)
                     {
                         if (bIsComplex)
                         {
-                            dfR = dfTotalReal / dfTotalWeight;
-                            dfI = dfTotalImag / dfTotalWeight;
-                            dfValueReal = sqrt((dfR + sqrt(dfR * dfR + dfI * dfI))/2.);;
-                            dfValueImag = dfI / fabs(dfI) * sqrt((-dfR + sqrt(dfR * dfR + dfI * dfI)) / 2.);
+                            dfValueReal = sqrt( dfTotalReal / dfTotalWeight );
```
rouault, line 6157: the computation of dfTotalReal and dfTotalImag done a bit above in the bIsComplex case should be fixed to be similar to the one in the !bIsComplex case. Same remark for overview.cpp

mgoacolou: done

rouault (approved):
I've added a follow-up commits with various adjustments/fixes, docs, extra tests

Archer:
1



发自我的iPhone


------------------ Original ------------------

PR #3224, opened 2020-11-25
do not use global code in pytest

### LINKED ISSUE #3223: gdal_calc unit tests are silently skipped and some fail
gdal_calc unit tests seem to be silently skipped and when fixed some of them fail

Actual
```
mmom@mmom-workstation:~/src/gdal/autotest$ pytest  pyscripts/test_gdal_calc.py 
Test session starts (platform: linux, Python 3.7.5, pytest 5.3.5, pytest-sugar 0.9.2)
rootdir: /home/mmom/src/gdal-mmomtchev/autotest, inifile: pytest.ini
plugins: env-0.6.2, sugar-0.9.2
collecting ... 

Results (0.97s):
```

Expected

```
mmom@mmom-workstation:~/src/gdal-mmomtchev/autotest$ pytest  pyscripts/test_gdal_calc.py 
Test session starts (platform: linux, Python 3.7.5, pytest 5.3.5, pytest-sugar 0.9.2)
rootdir: /home/mmom/src/gdal-mmomtchev/autotest, inifile: pytest.ini
plugins: env-0.6.2, sugar-0.9.2
collecting ... 
 pyscripts/test_gdal_calc.py ✓✓✓                                                                                                                                                              30% ███       
```

Do the steps from `CONTRIBUTING.md`

## Operating system

Ubuntu 19.10

git master

mmomtchev:
## What does this PR do?

Remove global code from ` test_gdal_calc.py`
Fix #3223 

#3223

### DIFF
--- a/autotest/pyscripts/test_gdal_calc.py
+++ b/autotest/pyscripts/test_gdal_calc.py
@@ -357,7 +357,11 @@ def test_gdal_calc_py_9():
         kwargs.update(inputs0)
         ds = gdal_calc.Calc(calc='a', outfile=outfile, **kwargs)
         input_file = ds if return_ds else outfile
+        if not return_ds:
+            # the dataset must be closed if we are to read it again
+            del ds
         inputs.append(input_file)
+
         check_file(input_file, checksums[i], i+1)
 
     inputs1 = dict()
@@ -412,6 +416,7 @@ def test_gdal_calc_py_9():
 
 def test_gdal_calc_py_cleanup():
     """ cleanup all temporary files that were created in this pytest """
+    return
     global temp_counter_dict
     global opts_counter_counter
     temp_files = []
--- a/autotest/pyscripts/test_gdal_calc.py
+++ b/autotest/pyscripts/test_gdal_calc.py
@@ -416,7 +416,6 @@ def test_gdal_calc_py_9():
 
 def test_gdal_calc_py_cleanup():
     """ cleanup all temporary files that were created in this pytest """
-    return
     global temp_counter_dict
     global opts_counter_counter
     temp_files = []
--- a/autotest/pyscripts/test_gdal_calc.py
+++ b/autotest/pyscripts/test_gdal_calc.py
@@ -332,6 +332,14 @@ def test_gdal_calc_py_9():
     * single alpha for multiple datasets
     * extent = 'fail'
     """
+
+    if gdalnumeric_not_available:
+        pytest.skip("gdalnumeric is not available, skipping all tests", allow_module_level=True)
+
+    script_path = test_py_scripts.get_py_script('gdal_calc')
+    if script_path is None:
+        pytest.skip("gdal_calc script not found, skipping all tests", allow_module_level=True)
+
     infile = get_input_file()
     test_id, test_count = 9, 9
     out = make_temp_filename_list(test_id, test_count)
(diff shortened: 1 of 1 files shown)


mmomtchev:
This just reenables the unit tests, some of them fail

mmomtchev:
@rouault do not merge, I will look into the failing tests

rouault:
@idanmiara might possibly help w.r.t test_gdal_calc_py_9() failure

mmomtchev:
in `test_gdal_calc_py_4()`, the geotransforms are missing from the output file, the result is otherwise correct

mmomtchev:
In fact, for test #4, I think that the new behavior is the correct one, the input file does not have a valid geotransform

mmomtchev:
No, for #4 the geotransforms do look better than before, but I am getting 3 identical bands when I should not

mmomtchev:
I get different results when I use the deprecated script from the old location - which I think the unit test uses too

mmomtchev:
@roualt It is not as bad as it seems, I think the only problem is that the deprecated script does not import the right `gdal_calc` after building in a dev tree

mmomtchev:
@rouault does the tests work for you? Because now that I am calling the right `gdal_calc` script (I just updated the PR), only test 9 is still failing
And even that one works if I manually call the script
So I guess it is a Python import path issue?

rouault:
The tests now pass with your changes in autotest/pyscripts/test_gdal2tiles.py . I don't need the changes in autotest/pymod/test_py_scripts.py : they don't hurt, but I don't think they're really needed and should be reverted. They must hide something wrong with your dev environment. Do you source the "gdal/scripts/setdevenv.sh" script that will set PYTHONPATH ?

mmomtchev:
Yes, I did source it and `gdalinfo --version` does show the dev tree. But my system GDAL is 3.1.0, so if something is included from it, it explains that test 9 fails

mmomtchev:
@rouault I found the problem - the dataset is not closed after writing and if computing the checksum by reopening the file, the reported checksum is zero
Once the dataset is closed, the checksum is correct
But I wonder, should `gdal_calc.Calc` return a closed dataset? Who shall close the dataset in this case?

rouault:
> But I wonder, should `gdal_calc.Calc` return a closed dataset? Who shall close the dataset in this case?

That's one of the recent change done by @idanmiara . gdal_calc.Calc() returns a dataset that hasn't been closed, so if you want to gdal.Open() it you need to close it before. This is similar to gdal.Translate() or gdal.Warp()

idanmiara:
> > But I wonder, should `gdal_calc.Calc` return a closed dataset? Who shall close the dataset in this case?
> 
> That's one of the recent change done by @idanmiara . gdal_calc.Calc() returns a dataset that hasn't been closed, so if you want to gdal.Open() it you need to close it before. This is similar to gdal.Translate() or gdal.Warp()

If the result of gdal_calc.Calc() is not assigned to a variable then the dataset would be closed by itself, right?
So I guess that shouldn't be a problem.

idanmiara:
> @idanmiara might possibly help w.r.t test_gdal_calc_py_9() failure

sorry, what's w.r.t?

idanmiara:
@mmomtchev would you please explain in which cases some tests are skipped when they shouldn't be skipped ?
Why would the value of gdalnumeric_not_available or script_path would need to be evaluated in each test?
In which case would script_path not be found ?
Wouldn't it be easier/cleaner to run the tests against the gdal_calc module and not as a script anyway?

mmomtchev:
@idanmiara I didn't investigate further - I guess there is something special about the way `pytest` loads modules - but I don't have an idea what

rouault:
I'm merging this as it fixes a regression in testing. @idanmiara feel free to submit follow-up PR

idanmiara:
> I'm merging this as it fixes a regression in testing. @idanmiara feel free to submit follow-up PR

thanks.

Issue #1893, opened 2019-09-29
gdal_calc.py: support option Ignore NoData calculation


bennyistanto:
Hello, I hope gdal_calc.py would support raster calculation with option Ignore Nodata.
If using ArcGIS Desktop, its easy using Cell Statistics within the Spatial Analyst Toolbox with option Ignore NoData calculation enabled/checked.

Example case: 
I would like to get maximum value from two or more raster Temperature using gdal_calc.py
If one of the raster has a lot of missing data, then the result will look necesseraly the same. 
But, if gdal_calc.py has option Ignore NoData then I can enable this option. So the maximum temperature will be calculated from the available data (other raster without missing data)

Thanks

taoyu17:
strongly agreed!

dmichelakis:
Please - this will be hugely appreciated!

rouault:
The description of this ticket is not clear.

Let's me recap the current behaviour of gdal_calc regarding nodata value: for each output pixel, as soon as one of the source pixels correspond to the nodata value of its band, the output pixel is set to the output nodata value.

What is exactly the behavior asked here:
- do not take into consideration source nodata value at all ? That is deal with it as a regular value and let the computation formula potentially take them into account in a special way
- or ... ?

bennyistanto:
How about something like below picture, I hope it's clear enough.
![image](https://user-images.githubusercontent.com/722584/78907962-9bcb2700-7aab-11ea-9ab5-7a8aba943236.png)

it would be good if  gdal_calc have option to ignore NoData in the calculation.

rouault:
> I hope it's clear enough.

how would you implement that concretely if your formula if (A+B+C+D+E+F+G+H+I+J)/10 ?

bennyistanto:
Yeah I know, its difficult.

Example case is, if we have daily rainfall data in 1 year in TIF in a folder and would like to get the AVERAGE, MAX and MIN from it. I imagine by using gdal_calc, can type: `gdal_calc.py -A foldername --outfile=outputmax.tif --calc="MAX(A)" --ignoreNODATA=1`

dmichelakis:
If gdal_calc can use numpy operations could this problem be solved using `numpy.nanmean`?

mmomtchev:
That would be a very useful feature for weather data, I agree, but it would require a substantial modification of `gdal_calc`
Currently NoData is a mask that is computed from all the bands and it is then applied to the output band
[...]

PR #3228, opened 2020-11-25
gdal_calc: map GDAL NoData to numpy.nan

mmomtchev:
## What does this PR do?

map GDAL's NoData to numpy's nan so one could write `--calc="numpy.nanmean([A, B, C], axis=2)"` to get an average without taking into account the NoData's

#1893 

## Tasklist

 - [x] Add test case(s)
 - [x] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS: Ubuntu 19.10
* Compiler: g++ 9 / Python 3.7

### DIFF
--- a/autotest/pyscripts/test_gdal_calc.py
+++ b/autotest/pyscripts/test_gdal_calc.py
@@ -181,10 +181,7 @@ def test_gdal_calc_py_4():
     out = make_temp_filename_list(test_id, test_count)
 
     # some values are clipped to 255, but this doesn't matter... small values were visually checked
-    print('gdal_calc', '-A {} --calc=1 --overwrite --outfile {}'.format(infile, out[0]))
     test_py_scripts.run_py_script(script_path, 'gdal_calc', '-A {} --calc=1 --overwrite --outfile {}'.format(infile, out[0]))
-    print(script_path, 'gdal_calc', '-A {} -B {} --B_band 1 --allBands A --calc=A+B --NoDataValue=999 --overwrite --outfile {}'.format(
-        infile, out[0], out[1]))
     test_py_scripts.run_py_script(script_path, 'gdal_calc', '-A {} -B {} --B_band 1 --allBands A --calc=A+B --NoDataValue=999 --overwrite --outfile {}'.format(infile, out[0], out[1]))
 
     bnd_count = 3
--- a/autotest/pyscripts/test_gdal_calc.py
+++ b/autotest/pyscripts/test_gdal_calc.py
@@ -451,7 +451,6 @@ def test_gdal_calc_py_10():
 
 
 def test_gdal_calc_py_cleanup():
-    return
     """ cleanup all temporary files that were created in this pytest """
     global temp_counter_dict
     global opts_counter_counter
--- a/autotest/pyscripts/test_gdal_calc.py
+++ b/autotest/pyscripts/test_gdal_calc.py
@@ -181,7 +181,10 @@ def test_gdal_calc_py_4():
     out = make_temp_filename_list(test_id, test_count)
 
     # some values are clipped to 255, but this doesn't matter... small values were visually checked
+    print('gdal_calc', '-A {} --calc=1 --overwrite --outfile {}'.format(infile, out[0]))
     test_py_scripts.run_py_script(script_path, 'gdal_calc', '-A {} --calc=1 --overwrite --outfile {}'.format(infile, out[0]))
+    print(script_path, 'gdal_calc', '-A {} -B {} --B_band 1 --allBands A --calc=A+B --NoDataValue=999 --overwrite --outfile {}'.format(
+        infile, out[0], out[1]))
     test_py_scripts.run_py_script(script_path, 'gdal_calc', '-A {} -B {} --B_band 1 --allBands A --calc=A+B --NoDataValue=999 --overwrite --outfile {}'.format(infile, out[0], out[1]))
 
     bnd_count = 3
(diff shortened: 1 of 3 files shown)


idanmiara:
Nice work! 
I'm wondering if it has any impact performance wise on the original use cases.

mmomtchev:
It is probably a little bit faster since `numpy.nan` is provided by the hardware, it is implemented directly in the FPU microcode
But still, unless you are able to squeeze an FFT or linear algebra in the `calc` argument, `gdal_calc` is not limited by CPU performance - it is mostly I/O bound

rouault:
You will need to rebase this PR on top of latest master because of broken security updates in Ubuntu 16.04 and 18.04 that I've addressed per #3235

mmomtchev:
I think there might be a problem with my unit testing, since this unit test passes here, but by looking at it, I think I modified the behavior of `gdal_calc` in this case:
`-A data --calc=1`
* before, if A contained NoData, the corresponding outputs cell would be NoData
* with my PR, if A contains NoData, the corresponding outputs cell will be 1

I will rebuild from scratch and will try to reproduce it, but I suspect that this might be the problem

mmomtchev:
And generally speaking, for constructs like
`--calc="(A > B) * C + (A <= B) * D"`
my PR might have different behavior - it could produce a non-nan value if C or D is not a nan and the right condition is met

mmomtchev:
I found one:
```
>>> numpy.logical_or(numpy.nan > 0, 1 > 0)
True
>>> numpy.logical_or(numpy.nan > 0, 1 > 0) * 5
5
>>> 
```

mmomtchev:
In fact, I ran all my unit tests while importing another copy of `osgeo.utils`... it is remarkable that only one is failing :smile: 

<https://github.com/OSGeo/gdal/blob/ddb85dd65ae0af3adf611af29ba1870312e6423c/gdal/swig/python/osgeo/utils/gdal_calc.py#L516>

It is because of this that `--calc=1` will get transformed to a `numpy.ndarray` - using the NDV's mask to set its shape
I will have to rewrite this part, I had missed this

mmomtchev:
@rouault , @idanmiara The more I look at this

<https://github.com/OSGeo/gdal/blob/eaea495ccbc52271e2ee97740baf24e1bb79516c/autotest/pyscripts/test_gdal_calc.py#L185>

The more I wonder what it is supposed to do? The output data type is `Byte` and the NoDataValue is `999`?
NoDataValue with masks produces a `255` where there was NoData in the input files

rouault:
> The output data type is `Byte` and the NoDataValue is `999`?

That could be a way of avoiding a nodatavalue to be set. Before recent master changes that allow --NoDataValue=none to be specified, there was no other way.

mmomtchev:
@rouault @idanmiara 
What shall be the behavior of 255 + 1 if the output type is Byte? 0 or 255? The current one is 0

rouault:
> What shall be the behavior of 255 + 1 if the output type is Byte? 0 or 255?

No magic here, we use numpy, and thus follow its design constraints, so this will wrap around at 0. Actually what matters is the input data type in this regard, not the output one. This is documented in gdal_calc doc

idanmiara:
Special treatment for overflow/underflow might not be worth the extra
computation time.
If no extra computation time would be required then replacing the value
with with NDV (if provided) or with the DefaultNDV could be a good option.

mmomtchev:
Ok, thanks, in fact using a `numpy.where` with a `numpy.nan` in the parameters will silently convert the numpy array to floating point, thus explaining the different behavior of the truncation
So, using `numpy.nan` would be possible only if the input type is a floating point type

mmomtchev:
Ok, I gave up on unifying the quantum field theory with the general relativity, so I added an option that switches the new behavior - that is all NDV in the input files get converted to `numpy.nan` and all `numpy.nan` in the results get converted to the output NDV, bypassing the old mechanism

This allows for doing explicit operations with the NDV - like for example `numpy.nanmean`

idanmiara:
> Ok, I gave up on unifying the quantum field theory with the general relativity, so I added an option that switches the new behavior - that is all NDV in the input files get converted to `numpy.nan` and all `numpy.nan` in the results get converted to the output NDV, bypassing the old mechanism
> 
> This allows for doing explicit operations with the NDV - like for example `numpy.nanmean`

So now that you are a famous numpy physicist, what's still left to do for this PR to see the light?

mmomtchev:
I will re-submit it, it was not as elegant as initially expected, but it was working well for floating point data
@rouault I guess that this is not the first time someone has the idea to map `numpy.nan` to GDAL NDV - and since it hasn't been done - you probably know something that I do not?
This is of course also true for the C/C++ version, but I guess that there will be portability issues.
I suppose that portability and the ability to have a common abstraction that also applies to integer values, are the reason the FPU `NaN` is not used?

rouault:
> I guess that this is not the first time someone has the idea to map `numpy.nan` to GDAL NDV - and since it hasn't been done

I don't have memories of anyone having tried to do that in gdal_calc

> I suppose that portability and the ability to have a common abstraction that also applies to integer values, are the reason the FPU NaN is not used?

For obvious reasons, the nodata value should be of the same data type as the raster data type, so NaN cannot be indeed used as a universal nodata value. And even for floating point values, it is somewhat annoying that x == nodata fails when x and nodata are both equal to NaN, hence some people prefer to avoid it

mmomtchev:
@rouault Yes, I meant it for the whole library
Well, there should probably be a GDAL-provided test for this case (`GDALIsNoData`) but this is a major breaking change, so probably, who knows, maybe it should be discussed again if there is a 4.0 one day - implementing it in a way that will permit to use a hardware `nan` on platforms/compilers/unit types that support it.

Using `nan` has lots of advantages, as many math libraries handle it very well avoiding the end-user a potentially error-prone in-house implementation of all the edge cases.

@idanmiara I will just stick to the `gdal_calc` case for now.

PR #3254, opened 2020-12-01
VRT: close sources at dataset closing (fixes #3253)

### LINKED ISSUE #3253: VRT File Lock Issue

I've noticed a VRT file locking issue with GDAL 3.0.4 which is as follows:

I've 2 VRT files whose source files names are same but the files (i.e., VRT & source file) are located at different locations but after opening both of them in readonly mode, I closed one of the file. Then, ideally the file I closed must be free to updated but it is not until I close down the other VRT file.

I verified the same in QGIS 3.16.1-Hannover which is using GDAL 3.1.4 and it has the same issue.

Open the both of the attached VRT in QGIS and remove one of them and try to rename the source Tiff file of the removed VRT.

## Operating system

Windows 10

3.0.4
[Dataset.zip](https://github.com/OSGeo/gdal/files/5622350/Dataset.zip)

rouault:


gdal/doc/source/drivers/raster/vrt.rst
```diff
@@ -292,12 +292,6 @@ The relativeToVRT attribute on the SourceFilename indicates whether the
 filename should be interpreted as relative to the .vrt file (value is 1)
 or not relative to the .vrt file (value is 0).  The default is 0.
 
-The shared attribute, on the SourceFilename indicates whether the
-dataset should be shared (value is 1) or not (value is 0). The default is 1.
-If several VRT datasets referring to the same underlying sources are used in a multithreaded context,
-shared should be set to 0. Alternatively, the VRT_SHARED_SOURCE configuration
-option can be set to 0 to force non-shared mode.
-
```
sgillies, line 300: @rouault is this usage guidance no longer relevant? I'm not sure what to make of this change.

rouault: this was true of GDAL < 2.3, and this is also specified in the "multithreading issues" section. No need to confuse users more than needed.

PR #3259, opened 2020-12-03
osgeo.utils -> osgeo_utils individual package

idanmiara:
## What does this PR do?

This is an implementation for RFC78.

## What are related issues/pull requests?

https://github.com/OSGeo/gdal/pull/3260

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments
 - [x] All CI builds and checks have passed

PR #3260, opened 2020-12-03
RFC 78: gdal-utils package

idanmiara:
## What does this PR do?

This RFC suggests to rename osgeo.utils (which were introduced in GDAL 3.2) into osgeo_utils. Essentially moving the utils outside of the osgeo package (which includes the GDAL core bindings), into a new pure-Python package gdal-utils. gdal-utils new distribution wheel will be available in pypi for smooth upgrading via pip install --upgrade gdal-utils.

For maximum backwards compatibility, The gdal distribution wheel would still include both packages: osgeo and osgeo_utils.

This PR includes a suggested implementation.

@rouault @hobu @rcoup 

## Formatted version

https://github.com/talos-gis/gdal/blob/Branch_rfc78_py_modules/gdal/doc/source/development/rfc/rfc78_gdal_utils_package.rst

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments


gdal/doc/source/development/rfc/rfc78_osgeo_utils.rst
```diff
@@ -0,0 +1,184 @@
...
+This RFC sets the new package `osgeo_utils`, as the name implies, it is related to `osgeo.utils`.
+Only one character is different between `osgeo_utils` and `osgeo.utils` (which it mainly replaces)
+should help a smooth transition.
+If you think you have a better name for this package name or other any other name suggested in this RFC - please do say.
+
+Motivation
+----------
+
+As the Python code evolves independently of the GDAL core and is not directly dependent on a specific
+GDAL version, one might want use the latest `osgeo_utils` package with an older `osgeo` GDAL core, or vice versa.
```
rouault, line 38: > As the Python code evolves independently of the GDAL core and is not directly dependent on a specific
GDAL version

Unfortunately, this is not always true. Some GDAL python scripts might rely on new GDAL C API available in updated SWIG Python bindings, or change in behaviour of existing API

gdal/doc/source/development/rfc/rfc78_osgeo_utils.rst
```diff
@@ -0,0 +1,184 @@
...
+
+Conversely, if we drop another Python version, like we just did with dropping Python 2.7,
+One might be still be able to use a newer `osgeo` (which I believe might still support Python 2.7)
+with an older `osgeo_utils`, Which as of now might support Python >= 3.6 and in the future might support
+only a newer version of Python.
+Technically, We could also deliver a first and also final version version of `osgeo_utils`
+that still supports Python 2.7, by taking the code from the GDAL 3.2 branch.
+
+All in all, fixes and improvements to both the `osgeo` and the `osgeo_utils` packages can be delivered faster
+And one can mix and match easily.
```
rouault, line 53: > And one can mix and match easily.

There would be certainly version dependencies between both packages. It would be hard to guarantee that a version X of osgeo_utils would work with a version Y != X of osgeo, because we would have no testing of such combination.

gdal/doc/source/development/rfc/rfc78_osgeo_utils.rst
```diff
@@ -0,0 +1,184 @@
...
+
+Documentation
+-------------
+
+The new layout should be documented, maybe inside each sub-package __init__.py file.
+
+Testing
+-------
+
+No any additional tests should be needed.
```
rouault, line 158: If there are 2 separate Python packages, there's probably an impact on the release procedure (HOWTO_RELEASE) regarding update to pypi

idanmiara:
> Do we need this .t ? We do release bugfixes version of GDAL every 2 months. That's the .z number

Good point. 2 months is a good timeframe. I'll remove that.

idanmiara:
@rouault does the `pypi` section make sense to you? 
Could you please reserve `osgeo` and `osgeo_utils` in `pypi` as they are still available?

rouault:
Shouldn't the pypi package be "gdal" (bindings only) and "gdal_utils" (that would have a dependency on gdal)  ?
"pip install osgeo" or osgeo_utils is confusing. The osgeo part is just a namespacing artifact, and perhaps not a super good one. People want to use GDAL, not OSGeo.

We probably need to mention a backward compatiblity issue if the "gdal" pip project no longer ships the utils

idanmiara:
I'm wondering about the utils namespace should it be `osgeo_utils` or `gdal_utils`?
I think I'd go with `osgeo_utils` because:
* it's a single char difference from osgeo.utils
* the modules inside utils already have a gdal prefix so `from osgeo_utils import gdal_calc` makes sense.
* it leaves room for `from osgeo_utils import ogr_something, osr_something`

> Shouldn't the pypi package be "gdal" (bindings only) and "gdal_utils" (that would have a dependency on gdal) ?
> "pip install osgeo" or osgeo_utils is confusing. The osgeo part is just a namespacing artifact, and perhaps not a super good one. People want to use GDAL, not OSGeo.
> 
> We probably need to mention a backward compatiblity issue if the "gdal" pip project no longer ships the utils

So as you said, one option would be:
`pip install gdal` would install just the `osgeo` package that includes the bindings;
`pip install gdal_utils` would install the new utils package and the bindings as being its dependency.
This would raise another backward compatibility issue as you noted.

Another option would be:
`pip install osgeo` (or gdal_bindings?) would install just the `osgeo` package that includes just the bindings;
`pip install gdal` would install the new utils package and the bindings as being its dependency. 
So the pypi package name `gdal` could also be considered as a meta-package name.
And no further backward compatibility issues with the second option, right?

gdal/doc/source/development/rfc/rfc78_osgeo_utils.rst
```diff
@@ -0,0 +1,280 @@
...
+Folder structure change
+--------------------------
+
+* `gdal/swig/python/osgeo/utils` -> `gdal/python/src/osgeo_utils`
+
+* `gdal/swig/python/osgeo/auxiliary` -> `gdal/python/src/osgeo_utils/auxiliary`
+
+* `gdal/swig/python/scripts` -> `gdal/python/src/osgeo_utils/scripts`
+
+* `autotest/pyscripts` -> `gdal/python/tests`
```
rouault, line 161: Some of the tests in autotest/pyscripts depend on autotest/pymod

gdal/doc/source/development/rfc/rfc78_osgeo_utils.rst
```diff
@@ -0,0 +1,280 @@
...
+============== ============================
+
+Summary
+-------
+
+This RFC is focused on the Python code in the GDAL repo, it would have no impact on non-Python code.
+This RFC suggests that the `osgeo` package will include only the SWIG bindings (As was in GDAL<3.2),
+And mainly suggests to move `osgeo.utils`, `osgeo.auxiliary` and other Python stuff that can be reused
+into a new pure-Python package named `osgeo_utils` and sets a new layout for this package.
+The `osgeo_utils` package will be distributed in pypi.org independently of the `osgeo` package.
```
rouault, line 22: I really don't like those names, at least as pypi packages. For pypi packages, "gdal_bindings" and "gdal_utils" would be better, and "gdal" as the meta.
My feeling is that the ratio complications / advantages of this RFC isn't big enough. And this RFC without the actual pypi packaging splitting (which is where complications would start) doesn't seem to bring much advantages, and I would need to be convinced that there would be many users taking advantages of those split packages.
My vote would be currently somewhere around 0 / -0 on it, besides the naming issue.

gdal/doc/source/development/rfc/rfc78_osgeo_utils.rst
idanmiara: I see.
Do you think `pymod` needs to be a subpacakge of `osgeo_utils` ? or something else?

gdal/doc/source/development/rfc/rfc78_osgeo_utils.rst
idanmiara: I am also for splitting the wheels.

So if I understand you correctly you meant:
* `osgeo` in `gdal_bindings` wheel
* `osgeo_utils` in `gdal_utils` wheel, that depends on the `gdal_bindings` wheel
* `gdal` wheel would depend on both as a meta package ?
Does that sound that right?

If wheel/pypi packaging splitting is implemented would you think you'd vote for it?

gdal/doc/source/development/rfc/rfc78_osgeo_utils.rst
rouault: but if autotest/pyscripts is moved in gdal/python/tests, it won't be tested anymore by running pytest from autotest  ? (I may have missed it from your implementation but the diff is too large to be easily reviewed)

gdal/doc/source/development/rfc/rfc78_osgeo_utils.rst
rouault: > If wheel/pypi packaging splitting is implemented would you think you'd vote for it?

I don't know. I guess I'm not really convinced this package splitting and all the implications it has, foreseen and unforeseen, is really worth it. I can see your use case, but not sure how common it is, and as discussed privately, you could also fallback to vendorizing the utils you're interested in in your library, in case you depend on new features not typically present in the version of the utility of the GDAL package you rely on.

idanmiara: Our maybe we better make `gdal_utils` be also the package name and also the wheel name to avoid further confusion of names.

idanmiara: > > If wheel/pypi packaging splitting is implemented would you think you'd vote for it?
> 
> I don't know. I guess I'm not really convinced this package splitting and all the implications it has, foreseen and unforeseen, is really worth it. I can see your use case, but not sure how common it is, and as discussed privately, you could also fallback to vendorizing the utils you're interested in in your library, in case you depend on new features not typically present in the version of the utility of the GDAL package you rely on.

I understand, thank you.

gdal/doc/source/development/rfc/rfc78_osgeo_utils.rst
idanmiara: See [`autotest/pytest.ini`](https://github.com/OSGeo/gdal/pull/3259/files#diff-b4368b7fec3584d4c01ae8741e1a0df7a942e04713dd4f30b8e9c0d015ea9fb8)
`testpaths = ogr gcore gdrivers osr alg gnm utilities ../gdal/python/tests`

gdal/doc/source/development/rfc/rfc78_osgeo_utils.rst
idanmiara: As you said my implementation already makes a fairly large diff (without wheel split)
So maybe we better of with a more gradual and conservative two step approach:

1. As already implemented, and wouldn't have a big impact on users who don't care about it as everything would be still in the `gdal` wheel and the utils still installed as scripts for backwards compatibility. Users who care about mixing versions could upgrade the utils with the separate utils wheel (which I also implemented in that PR).
2. Evaluate and implement a wheel split in another PR.

I guess the naming should be decided in this PR so not to add more future confusions.

rouault: >  Or maybe we better make gdal_utils be also the package name and also the wheel name to avoid further confusion of names.

Would be better

>     1. Users who care about mixing versions could upgrade the utils with the separate utils wheel (which I also implemented in that PR).

Hum, how that work if someone installs the "gdal" all-in-one package and the "gdal_utils" one. Wouldn't that conflict ?

> I guess the naming should be decided in this PR so not to add more future confusions.

Yes, we don't want to change package names for every RFC :-)

idanmiara: > Hum, how that work if someone installs the "gdal" all-in-one package and the "gdal_utils" one. Wouldn't that conflict ?

I've done some testing. 

It seems that every time you `pip install` a wheel it just overwrites whichever files that may already exist. 
If you `pip install gdal` then `pip install gdal_utils` you'd get the utils from `gdal_utils`.
If later you do again `pip install gdal` with a different version then you'd get the utils from `gdal` again, and so on.
(it doesn't seem that it matters which version is a bigger number, just which one you installed later)

If you `pip install gdal` then `pip install gdal_utils` and then `pip uninstall gdal_utils` then the `utils` would be uninstalled and you'd be left with gdal without utils.
Then you could `pip install gdal --ignore-installed` to get them back again (`--ignore-installed` is not required if you install a different version)

idanmiara: I've been thinking...
`gdal` and `gdal_utils` as wheel/pypi packages sounds good and consistent.
regarding the package inside `gdal` and `gdal_utils`, I'm not sure what sounds more consistent:
option 1:
```
from osgeo import gdal, ogr, osr
from gdal_utils import gdal_calc, ogr_foo, osr_bar
```
option 2:
from osgeo import gdal
from utils_utils import gdal_calc, ogr_foo, osr_bar
```

Regarding the second stage:
I think I'll drop the `gdal` meta package idea. sounds too complicated.
When/if we decide to remove the utils from the `gdal` wheel then they will be already ready inside `gdal_utils` wheel, which depends on the `gdal` wheel anyway.
So `pip install gdal_utils` would install both.

What do you think?

**Comment by idanmiara** [comment]
I've updated the RFC and released a test version of `gdal_utils` to [pypi](https://pypi.org/project/gdal-utils/).

I mistakenly used version "3.3.0" on pypi so we cannot use this version for the release, we could use "3.3.0.0" 
I couldn't figure out why the Project description shows as raw although I specified the format in the setup.py. 
Do you prefer I'll remove `gdal-utils` from pypi and put it only on [test.pypi](https://test.pypi.org/project/gdal-utils/
) till it is approved?

Open issues:
0. @rouault @hobu - Would you like to share more comments? Would you like to discuss this RFC further before voting?
1. Who will be the maintainer of this package? @hobu, as you are the maintainer of the gdal package, would you like to be listed as the maintainer of `gdal-utils`? or myself? or both of us?
2. package naming (`from gdal_utils import gdal_calc, ogr_foo, osr_bar` vs `from osgeo_utils import gdal_calc, ogr_foo, osr_bar`)

**Comment by rouault** [comment]
@idanmiara I'm still lacking enthusiasm for this RFC, but maybe others will be or have suggestions to make it evolve. If you still want to pursue it, the process would be first to advertize it to gdal-dev for discussion before putting it to vote.

**Comment by idanmiara** [comment]
> @idanmiara I'm still lacking enthusiasm for this RFC, but maybe others will be or have suggestions to make it evolve. If you still want to pursue it, the process would be first to advertize it to gdal-dev for discussion before putting it to vote.

Done, thanks!

**Comment by idanmiara** [comment]
I have no idea what causes all these failures on the MacOS build. 
Can someone drop me a hint?

**Comment by idanmiara** [comment]
Hi @rouault,
Would you please advise what's the next step regarding this PR?

**Comment by rouault** [comment]
> Would you please advise what's the next step regarding this PR?

if you want still to have it adopted, you'll need to make a motion for it on the mailing list. I'm still unconvinced though

**Comment by idanmiara** [comment]
> > Would you please advise what's the next step regarding this PR?
> 
> if you want still to have it adopted, you'll need to make a motion for it on the mailing list. I'm still unconvinced though

Any ideas/pointers whatsoever to improve this PR?

**Comment by idanmiara** [comment]
> > Would you please advise what's the next step regarding this PR?
> 
> if you want still to have it adopted, you'll need to make a motion for it on the mailing list. I'm still unconvinced though

@rouault I've revised the text of this RFC today to emphasize the main ideas and motivations. please read it again before you vote. Thanks!

File: gdal/swig/python/gdal-utils/setup.py [hunk 6]
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# Setup script for gdal-utils.
+
+from setuptools import setup, find_packages
+
+__pacakge_name__ = 'gdal-utils'
+__version__ = '3.3.0'
```
rouault, line 8: HOWTO-RELEASE should be updated to ask to update this number in the part where we update version numbers

idanmiara: done.
I've added a build for the gdal-utils wheel as well.

gdal/doc/source/development/rfc/rfc78_gdal_utils_package.rst
```diff
@@ -0,0 +1,242 @@
...
+
+===========================================
+RFC 78: gdal-utils package
+===========================================
+
+============== ============================
+Author:        Idan Miara
+Contact:       idan@miara.com
+Started:       2020-Dec-3
+Last updated:  2021-Match-26
```
rouault, line 11:
```suggestion
Last updated:  2021-March-26
```

gdal/doc/source/development/rfc/rfc78_gdal_utils_package.rst
```diff
@@ -0,0 +1,242 @@
...
+
+- https://gdal.org/development/rfc/rfc77_drop_python2_support.html
+- https://github.com/OSGeo/gdal/pull/3131
+- https://github.com/OSGeo/gdal/pull/3117
+- https://github.com/OSGeo/gdal/pull/3247
+
+Voting history
+--------------
+
+* +1 from EvenR, HowardB, , , DanielM
```
rouault, line 234: DanielM didn't vote AFAICS

gdal/doc/source/development/rfc/rfc78_gdal_utils_package.rst
idanmiara: done

gdal/doc/source/development/rfc/rfc78_gdal_utils_package.rst
idanmiara: right... careless copy paste :)

rouault:
Could you possibly rework a bit the history in a commit with the RFC text and another one with the implementation (for the implementation, potentially split up in bisect-able pieces if that makes sense) ?

idanmiara:
> Could you possibly rework a bit the history in a commit with the RFC text and another one with the implementation (for the implementation, potentially split up in bisect-able pieces if that makes sense) ?

done.

Do you think that `./gdal/swig/python/gdal-utils` is the proper path for the sources? other options I had in mind:
`./gdal/gdal-utils` - as it's not really swig related.
`./gdal-utils` - maybe a better place if we think to split autotest sometime in the future?
I really don't have a solid preference...

rouault:
> done.

except that the commits where you change the osgeo/utils to osgeo_utils, and then the change in the imports should be done in one step, otherwise this breaks bisectability (by bysectability, I mean that ideally we should be able to checkout at a random master commit and have a fully working tree)

> other options I had in mind:
The downside with your alternate proposals is that it is no longer obvious that gdal-utils is python related.

idanmiara:
> except that the commits where you change the osgeo/utils to osgeo_utils, and then the change in the imports should be done in one step, otherwise this breaks bisectability (by bysectability, I mean that ideally we should be able to checkout at a random master commit and have a fully working tree)

The logic behind the individual commits was that it's easier to track the changes I've made in each commit. 
But I see what you mean, so I squashed the 4 following commits to make it bisactable:
* osgeo.utils -> osgeo_utils (search replace)
* move gdal/swig/python/osgeo/utils -> gdal/swig/python/gdal-utils/osgeo_utils
* gdal/swig/python/setup.py - update paths
* test_py_scripts.py - update osgeo_utils path

idanmiara:
Thanks!
@rouault I've sent you an invitation to be an owner of [https://pypi.org/project/gdal-utils/](https://pypi.org/project/gdal-utils/).

Please note that I've released a gdal-utils version 3.3.0 in the past by mistake, so for the first official release could be 3.3.0.X or 3.3.1.
Before the release of gdal 3.3.0 I'll yank all previous versions so only official versions would be available.

I would like to keep a "beta channel" for `gdal-utils` as a playground to test new features before they release officially. 
As I don't want users to mistakenly think these as official versions I think I'll better use a different name, maybe `gdal-utils-beta` (unless you would like to suggest a different name).

rouault:
>  I've released a gdal-utils version 3.3.0 in the past by mistake,

can't the release by removed / the project removed & recreated ?

idanmiara:
> > I've released a gdal-utils version 3.3.0 in the past by mistake,
> 
> can't the release by removed / the project removed & recreated ?

Nope... 
https://github.com/pypa/packaging-problems/issues/74
I'll delete 3.3.0 so it wouldn't be possible to install it even by specifying `gdal-utils == 3.3.0`

PR #3274, opened 2020-12-09
NITF: Add support for ISO8859-1 fields in NITF TREs

matthew-baran:
## What does this PR do?
This PR introduces a new field type to nitf_spec.xml to handle NITF Extended Character Set (ECS).  ECS consists of 1-byte characters that align to the ISO-8859-1 extended ASCII standard.

ECS is defined in MIL-STD-2500C
ECS character \mu (0xB5) is used for several TREs including ILLUMB (STDI-0002 Appendix AL) and PIXMTA (STDI-0002 Appendix AJ).

https://github.com/OSGeo/gdal/pull/3071
https://github.com/OSGeo/gdal/pull/3023

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [x] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:

### DIFF
--- a/autotest/gdrivers/nitf.py
+++ b/autotest/gdrivers/nitf.py
@@ -3107,8 +3107,10 @@ def test_nitf_86():
 # Test parsing ILLUMB TRE (STDI-0002-1-v5.0 App AL)
 
 def test_nitf_87():
+    mu = "B5"   # \mu per ISO-8859-1
     bit_mask = "7A0000"
-    tre_data = "TRE=HEX/ILLUMB=" + hex_string("0001mm                                      8.5192000000E-01") + \
+    tre_data = "TRE=HEX/ILLUMB=" + hex_string("0001") + \
+        mu + hex_string("m                                      8.5192000000E-01") + \
         hex_string("2.5770800000E+00001NUM_BANDS=1 because ILLUMB has no band-dependent content                        ") + \
         hex_string("World Geodetic System 1984                                                      ") + \
         hex_string("WGE World Geodetic System 1984                                                      ") + \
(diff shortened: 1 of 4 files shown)


rouault:
I was wondering if this ECS encoding only for the few fields it is declared or is it potentially true for any string ? In the later case we might save the "ISO8859-1" type and do recoding systematically for string.

matthew-baran:
> I was wondering if this ECS encoding only for the few fields it is declared or is it potentially true for any string ? In the later case we might save the "ISO8859-1" type and do recoding systematically for string.

Yes I think you're right.  Any other string field should be a subset of ASCII which is a subset of 8895-1, so it should be safe to convert all string fields and remove the 8859 type.

matthew-baran:
> 
> 
> > I was wondering if this ECS encoding only for the few fields it is declared or is it potentially true for any string ? In the later case we might save the "ISO8859-1" type and do recoding systematically for string.
> 
> Yes I think you're right. Any other string field should be a subset of ASCII which is a subset of 8895-1, so it should be safe to convert all string fields and remove the 8859 type.

After a deeper look, there is a "UCS" type in the NITF specification which allows for UTF-8 fields.  The only actual usage I could find in all the TREs defined in STDI-0002 is the SECURA TRE which has a UTF-8 field called "SECURITY".

Thoughts?

rouault:
My question was mostly for my information. Your change is fine is the spec distinguishes some fields as having distinct ASCII, ECS or UCS encoding. So the current String type should be understood as ASCII only.

PR #3276, opened 2020-12-10
Attempt to deploy to gdal-master Anaconda channel

hobu:
@JorisVandenBossche [suggested](https://twitter.com/jorisvdbossche/status/1336573419823960065?s=20) we create public channel for deploying builds from master is available at https://anaconda.org/gdal-master/

This PR adds that, but the push only happens if the branch is ``refs/heads/master``

### DIFF
--- a/.github/workflows/conda.yml
+++ b/.github/workflows/conda.yml
@@ -1,8 +1,6 @@
 name: Conda
 
-on:
-  push:
-    branches: '*'
+on: [push, pull_request]
 
 jobs:
   build:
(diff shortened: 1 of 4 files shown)


jorisvandenbossche:
Nice!

If needed you could also run this upload only in a cron-job, eg to do it once a day to potentially limit the number of uploads (I don't know what the typical commit activity is on this repo), since there is a storage limit (I need to check if there are automated ways to remove older artifacts)

rouault:
I presumed we would upload a package with the same version number overriding the previous one ? @hobu will correct me. Otherwise, the current solution is indeed only sustainable if there's some rolling mechanism on the storage side.

jorisvandenbossche:
The upload still failed: https://github.com/OSGeo/gdal/runs/1531517417

Looking at the artifact, it seems that you are indeed using a simple version and fixed build number (so the package is named like `gdal-3.2.0-py36h9cc8abd_2112.tar.bz2`), so then that will probably indeed override the previous one

hobu:
> Looking at the artifact, it seems that you are indeed using a simple version and fixed build number (so the package is named like gdal-3.2.0-py36h9cc8abd_2112.tar.bz2), so then that will probably indeed override the previous one

Yes, the intent was to simply clobber the existing build with a fixed build number. Using "2112" is our indicator where the build came from too (long live Neil Peart). 

I'll make a new PR that fixes the upload bits.

PR #3286, opened 2020-12-14
GeoPackage: Fix handling of invalid SRS ID

mloskot:
## What does this PR do?

Addresses the previous inline comment saying

```
    /* Should we do something special with undefined SRS ? */
```

and distinguishes between SRS ID of `0` and `-1` according to https://www.geopackage.org/spec121/#r11

> GeoPackage 1.2.1
>  1.1.1.1.1. File Format
>   Requirement 11
>    The record with an srs_id of -1 SHALL be used for undefined Cartesian coordinate reference systems.
>    The record with an srs_id of 0 SHALL be used for undefined geographic coordinate reference systems.
> 

## Tasklist

 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS: any
* Compiler: any

### DIFF
--- a/gdal/ogr/ogrsf_frmts/gpkg/ogrgeopackagedatasource.cpp
+++ b/gdal/ogr/ogrsf_frmts/gpkg/ogrgeopackagedatasource.cpp
@@ -349,6 +349,7 @@ OGRSpatialReference* GDALGeoPackageDataset::GetSpatialRef(int iSrsId,
     if( iSrsId == 0 || iSrsId == -1 )
     {
         OGRSpatialReference *poSpatialRef = new OGRSpatialReference();
+        poSpatialRef->SetAxisMappingStrategy( OAMS_TRADITIONAL_GIS_ORDER );
 
         if( iSrsId == 0)
         {
--- a/gdal/ogr/ogrsf_frmts/gpkg/ogrgeopackagedatasource.cpp
+++ b/gdal/ogr/ogrsf_frmts/gpkg/ogrgeopackagedatasource.cpp
@@ -336,12 +336,6 @@ static OGRErr GDALGPKGImportFromEPSG(OGRSpatialReference *poSpatialRef,
 OGRSpatialReference* GDALGeoPackageDataset::GetSpatialRef(int iSrsId,
                                                           bool bFallbackToEPSG)
 {
-    /* Should we do something special with undefined SRS ? */
-    if( iSrsId == 0 || iSrsId == -1 )
-    {
-        return nullptr;
-    }
-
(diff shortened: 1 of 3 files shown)


gdal/ogr/ogrsf_frmts/gpkg/ogrgeopackagedatasource.cpp
```diff
@@ -352,6 +346,29 @@ OGRSpatialReference* GDALGeoPackageDataset::GetSpatialRef(int iSrsId,
         return oIter->second;
     }
 
+    if( iSrsId == 0 || iSrsId == -1 )
+    {
+        OGRSpatialReference *poSpatialRef = new OGRSpatialReference();
```
rouault, line 351: You'll perhaps need to add
```
poSpatialRef->SetAxisMappingStrategy(OAMS_TRADITIONAL_GIS_ORDER);
```

rouault (changes requested):
There are a few unit tests to adapt for this change

gdal/ogr/ogrsf_frmts/gpkg/ogrgeopackagedatasource.cpp
mloskot: Thanks, added.

Of course, I'll look into the unit tests.

mloskot:
I adjusted the `ogr_gpkg.py` test.

I don't quite know how to exercise the `GDALGeoPackageDataset::GetSpatialRef` call with `iSrsId=-1` though.

mloskot:
@mloskot 
> I don't quite know how to exercise the `GDALGeoPackageDataset::GetSpatialRef` call with `iSrsId=-1` though.

Okey, it looks like `gpkg.py` does it in https://github.com/OSGeo/gdal/pull/3286/commits/c2d97be39d8ec6084039743f92964b0b6c70dde9

rouault:
There are some variations of WKT depending on the PROJ version which explain the remaining failures (not all builds use the same one). I'd suggest perhaps using the OGRSpatialReference.IsLocal() method instead of strict WKT string matching

mloskot:
>  I'd suggest perhaps using the OGRSpatialReference.IsLocal() method instead of strict WKT string matching

I'll fiddle with it, thank you.

rouault:
It occurs to me that we would probably want to do something on the writing side of the driver to be able to round-trip nicely those 2 special CRS. Is that something in your scope of interest @mloskot  ?

mloskot:
@rouault I will try to deliver such improvement over the next couple of days

PR #3287, opened 2020-12-15
doc: Recommended default mode for VSIMkdir is 0755

dg0yt:
## What does this PR do?

This PR fixes a documentation error. 

Following the previous instructions (mode 0666), the directory was created without execute permissions, rendering it unusable. E.g. after creating a `../Test/KMZ/123/files` directory:

~~~
$ LANG=C ls -la ../Test/KMZ/123/files
ls: cannot access '../Test/KMZ/123/files/.': Permission denied
ls: cannot access '../Test/KMZ/123/files/..': Permission denied
total 0
d????????? ? ? ? ?             ? .
d????????? ? ? ? ?             ? ..
~~~

Note that looking for `VSIMkdir` in GDAL reveals a number of suboptimal (`0755`) and broken (`0744`) code:
https://github.com/OSGeo/gdal/search?q=vsimkdir
However, I don't know if VSI and OS implementation would properly apply umask in all cases.

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/gdal/port/cpl_vsil.cpp
+++ b/gdal/port/cpl_vsil.cpp
@@ -436,10 +436,10 @@ void VSICloseDir(VSIDIR* dir)
 /**
  * \brief Create a directory.
  *
- * Create a new directory with the indicated mode.  The mode is ignored
- * on some platforms.  A reasonable default mode value would be 0666.
- * This method goes through the VSIFileHandler virtualization and may
- * work on unusual filesystems such as in memory.
+ * Create a new directory with the indicated mode. For POSIX-style systems,
+ * the mode is modified by the file creation mask (umask). However, some
+ * file systems and platforms may not use umask, or they may ignore the mode
+ * completely. So a reasonable cross-platform default mode value is 0755.
  *
  * Analog of the POSIX mkdir() function.
  *


gdal/port/cpl_vsil.cpp
```diff
@@ -437,7 +437,7 @@ void VSICloseDir(VSIDIR* dir)
  * \brief Create a directory.
  *
  * Create a new directory with the indicated mode.  The mode is ignored
- * on some platforms.  A reasonable default mode value would be 0666.
+ * on some platforms.  A reasonable default mode value would be 0777.
```
rouault, line 440: You generally don't want to give write permissions to group and other, so:

```suggestion
 * on some platforms.  A reasonable default mode value would be 0755.
```

dg0yt: No, I often don't want to give write permissions to group and other. And sometimes I even don't want to give any permissions to other. But this is normally well resolved by the mode given to `mkdir` being "modified by the process' file creation mask" aka `umask`: 
https://linux.die.net/man/3/mkdir 
http://manpages.ubuntu.com/manpages/focal/en/man2/mkdir.2.html 

So explicitly: IMO 0777 is right for Linux and POSIX. But if 0777 is unsafe with regard to GDAL's different target OS or VSI implementation, 0755 maybe a better recommendation in GDAL's documentation.

rouault: > No, I often don't want to give write permissions to group and other

I'm a bit confused by your message. 0755 is exactly that: rwx for user, rx for group and other, so it is more restrictive than 0777 and "safer"

dg0yt: Even more explicitly: 

`mkdir("files", 0777)` does *not* mean: Create dir "files" with exact permissions 0777. 
What it really means is: Create dir "files" with the result of `0777 & ~umask`.

`mkdir("files", 0755)` does *not* mean: Create dir "files" with exact permissions 0755. 
What it really means is: Create dir "files" with the result of `0755 & ~umask`.

A typical umask is 0022, i.e. don't give write permissions to group and others. Only if I explicitly change umask to 0000, files and dirs can be created with write permissions for group and other. And this is exactly what I expect when I change umask to 0000.

So in order for a program to behave correctly in the context of `umask`, you have to use `mkdir("file", 0777)`.

dg0yt: (Disclaimer: I'm not an expert on low-level functions.)

rouault, line 440: What do you think of
```suggestion
 * on some platforms.  A reasonable default mode value would be 0777 to grant
 read, write, execute permissions to user, group and other. Or 0755 for read,
 write, execute permissions to user, and only read and execute permissions to group
 and other. On some filesystems, the umask mechanism may further restrict those permissions.
```

dg0yt: I would like to keep it simple, and I don't think having two recommended value serves this purpose. (Why not 0775?) If in doubt, stick with 0755. Everything is better than 0666 (which gave me GDAL crashes in subsequent file access).

But looking at the various `mkdir` man pages, I now think the first sentense needs to be rectified, too: As explained, the function will not create the dir "with the indicated mode". So:

~~~
Create a new directory. The indicated mode is modified by
the file creation mask (umask) on POSIX-style systems,
so a reasonable default mode value is 0777. The mode is
ignored on some platforms.
~~~


We could also wait for others to comment.

dg0yt:
Selected 0755, and updated the wording to reflect why it is not 0777.

Issue #3122, opened 2020-10-28
Warnings in TileDB driver with x86_64-apple-darwin13.4.0-clang++


rouault:
CC @normanb 

Building GDAL on OSX with x86_64-apple-darwin13.4.0-clang++ generates warnings in the TileDB driver. See https://github.com/rouault/gdal/runs/1323732947:

```
tiledbdataset.cpp:399:19: error: equality comparison with extraneous parentheses [-Werror,-Wparentheses-equality]
    if ( ( status == tiledb::Query::Status::FAILED ) )
           ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
tiledbdataset.cpp:399:19: note: remove extraneous parentheses around the comparison to silence this warning
    if ( ( status == tiledb::Query::Status::FAILED ) )
         ~~       ^                                ~~
tiledbdataset.cpp:399:19: note: use '=' to turn this equality comparison into an assignment
    if ( ( status == tiledb::Query::Status::FAILED ) )
                  ^~
                  =
tiledbdataset.cpp:783:56: error: implicit conversion loses integer precision: 'size_t' (aka 'unsigned long') to 'uint32_t' (aka 'unsigned int') [-Werror,-Wshorten-64-to-32]
            oMeta->put_metadata("_gdal", TILEDB_UINT8, strlen( pszTree ), pszTree);
                   ~~~~~~~~~~~~                        ^~~~~~~~~~~~~~~~~
tiledbdataset.cpp:788:58: error: implicit conversion loses integer precision: 'size_t' (aka 'unsigned long') to 'uint32_t' (aka 'unsigned int') [-Werror,-Wshorten-64-to-32]
            m_array->put_metadata("_gdal", TILEDB_UINT8, strlen( pszTree ), pszTree);
                     ~~~~~~~~~~~~                        ^~~~~~~~~~~~~~~~~
tiledbdataset.cpp:1341:63: error: implicit conversion loses integer precision: 'unsigned long long' to 'int' [-Werror,-Wshorten-64-to-32]
                poDS->nBandStart = dims[0].domain<uint64_t>().first;
                                 ~ ~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~
tiledbdataset.cpp:1343:68: error: implicit conversion loses integer precision: 'unsigned long long' to 'int' [-Werror,-Wshorten-64-to-32]
                                - dims[0].domain<uint64_t>().first + 1;
[...]

PR #3293, opened 2020-12-17
TileDB: Add support for array versioning

normanb:
## What does this PR do?

This PR adds support for reading and writing TileDB arrays at user specified timestamps enabling a user to version the array data.

None

## Tasklist

 - [x] Add test case(s)
 - [x] Add documentation
 - [ ] Review
 - [ ] Adjust for comments


rouault:
By the way, could you have a look at https://github.com/OSGeo/gdal/issues/3122 ? (shouldn't be OSX specific. just using clang should be enough to reproduce)

Issue #3289, opened 2020-12-15
System.AccessViolationException reading some NTF files


czielin:
## Expected behavior and actual behavior.

We are using the sample C# code provided in GDALRead.cs and GDALReadDirect.cs here https://github.com/OSGeo/gdal/tree/master/gdal/swig/csharp/apps. We expect to be able to read in a NTF image for further processing and thumbnail generation. It works for some NTF images, but others throw an exception on read.

The exception always occurs when reading the third band of an RGB image. Even if we change the order of the colors.

```
redBand.ReadRaster(0, 0, width, height, r, width, height, 0, 0);
greenBand.ReadRaster(0, 0, width, height, g, width, height, 0, 0);
blueBand.ReadRaster(0, 0, width, height, b, width, height, 0, 0);
```

The exception thrown is:

> Unhandled Exception: System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
>    at OSGeo.GDAL.GdalPINVOKE.Band_ReadRaster__SWIG_0(HandleRef jarg1, Int32 jarg2, Int32 jarg3, Int32 jarg4, Int32 jarg5, IntPtr jarg6, Int32 jarg7, Int32 jarg8, Int32 jarg9, Int32 jarg10, Int32 jarg11)
>    at OSGeo.GDAL.Band.ReadRaster(Int32 xOff, Int32 yOff, Int32 xSize, Int32 ySize, IntPtr buffer, Int32 buf_xSize, Int32 buf_ySize, DataType buf_type, Int32 pixelSpace, Int32 lineSpace)
>    at GDALReadDirect.SaveBitmapDirect(Dataset ds, String filename, Int32 iOverview) in C:\...\GDALReadDirect.cs:line 195
>    at GDALReadDirect.Main(String[] args) in C:\...\GDALReadDirect.cs:line 131
>    at GDAL.Program.Main(String[] args) in C:\...\Program.cs:line 87

## Steps to reproduce the problem.

Use either the GdalRead.cs or GdalReadDirect.cs sample code to read an RGB image. I am not currently able to share the images that cause the exception. If I can get permission to share I will update the issue.

## Operating system

Windows 10

## GDAL version and provenance

2.4.4 (.NET Framework 4.8)

I can see there are newer releases on Github, but 2.4.4 is the latest that is available on Nuget.

rouault:
Check with the gdal_translate utility if the crash can be reproduced independntly of the C# bindings.
[...]

PR #3300, opened 2020-12-18
Add comments on how to fix System.AccessViolationException

czielin:
Add comments on how to fix System.AccessViolationException on some larger images. Related to issue #3289

### DIFF
--- a/gdal/swig/csharp/apps/GDALRead.cs
+++ b/gdal/swig/csharp/apps/GDALRead.cs
@@ -206,6 +206,8 @@ class GDALRead {
         byte[] g = new byte[width * height];
         byte[] b = new byte[width * height];
 
+        // Reading each band individually can sometimes cause a System.AccessViolationException for larger images.
+        // Using the ReadRaster() method on the Dataset instead (as in GDALDatasetRasterIO.cs and GDALDatasetRasterIOEx.cs) may fix this.
         redBand.ReadRaster(0, 0, width, height, r, width, height, 0, 0);
         greenBand.ReadRaster(0, 0, width, height, g, width, height, 0, 0);
         blueBand.ReadRaster(0, 0, width, height, b, width, height, 0, 0);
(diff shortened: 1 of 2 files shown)


rouault:
I'm not super enthusiastic about adding such comment without the root cause having been identified. Would you possibly read more than 2 GB at once ? (wondering if it might be a 32-bit integer overflow somewhere)

czielin:
The NTF source images that are causing a problem are in the 200MB to 400MB range on disk. Not sure how that translates to working memory. I'll run another test when I have a chance and keep on eye on memory utilization.

rouault:
> when running as 32-bit.

eh, still people running 32 bit code those days !?!!

> The Dataset.ReadRaster() version only uses about a third of the memory (and also takes about a third of the time).

That doesn't make sense to me. If you read the same quantity of information with Dataset.ReadRaster() or Band.ReadRaster(), you should have similar memory consumption

Closing this PR as I don't think it is relevant.

czielin:
> eh, still people running 32 bit code those days !?!!

Indeed. In this case it stemmed from our unit tests being configured to run as 32-bit. Though it appears we're not the only ones. The code sample provided in GDALReadDirect.cs uses 32-bit integers and it throws an Arithmetic exception when I run the test application as 64-bit. When I change the calls to ToInt64() it works.

image

> That doesn't make sense to me. If you read the same quantity of information with Dataset.ReadRaster() or Band.ReadRaster(), you should have similar memory consumption

I initially assumed the same, but this is not what we are observing when plugging the same file into the two sets of sample code provided in the repo. I do not know enough about how the ReadRaster method works to know if the problem is in the method itself, or the way the sample code is written.

I still think there is some value for future consumers in mentioning that the approach in these samples will not perform optimally. It certainly would have saved us a good amount of time. 

Happy to create another PR with the ToInt64() fix and/or updated commenting if you think it is valuable.

rouault:
> I initially assumed the same, but this is not what we are observing when plugging the same file into the two sets of sample code provided in the repo

The swig/csharp/apps/GDALReadDirect.cs reads into a RGBA pixel-interleave buffer, whereas swig/csharp/apps/GDALRead.cs reads in separate per-band buffers. Those are 2 valid use cases depending on the needs. You perhaps overallocate things when you use the Band.RasterIO() call

> Happy to create another PR with the ToInt64() fix

would be welcome

PR #3303, opened 2020-12-18
VRT: add a <UseMaskBand>true</UseMaskBand> element of <ComplexSource>…

### LINKED ISSUE #1148: Mask for overlapped layers masks both of them
I'd like to mosaic two overlapped rasters with VRT (see attachments). Both of them has a mask. But result looks like this. Is it expected result?

image

```
gdalbuildvrt -input_file_list list.txt output.vrt
```

[overlapped-rasters.zip](https://github.com/OSGeo/gdal/files/2681162/overlapped-rasters.zip)

rouault:
… to allow compositing of overlapping sources (fixes #1148)

### DIFF
--- a/autotest/gcore/vrt_read.py
+++ b/autotest/gcore/vrt_read.py
@@ -1529,3 +1529,118 @@ def test_vrt_implicit_ovr_with_hidenodatavalue():
     assert got_data[32*64+32] == 10
 
     gdal.Unlink('/vsimem/src.tif')
+
+
+def test_vrt_usemaskband():
+
+    ds = gdal.GetDriverByName('GTiff').Create('/vsimem/src1.tif', 3, 1)
+    ds.GetRasterBand(1).Fill(255)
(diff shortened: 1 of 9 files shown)


rouault:
@hobu We have failures in Conda builds. See https://pipelines.actions.githubusercontent.com/yqaARdsLodWRtkoqajmNaa64Xp4JCsYwhgHYzJtTCVY8q4LFPT/_apis/pipelines/1/runs/11266/signedlogcontent/4?urlExpires=2020-12-18T19%3A17%3A20.9751974Z&urlSigningMethod=HMACV1&urlSignature=lvkDAc8rymyAul2IMyC0qUupgdeQP2xD1aF4%2BoGzywA%3D
==> "gdalinfo: error while loading shared libraries: libicui18n.so.68: cannot open shared object file: No such file or directory" .
Do you think it is just a transient issue (but this also happens after a rebuild attempt) or  a change in Conda dependencies ?

PR #3308, opened 2020-12-20
[NAS] don't skip prescan, if NAS_GFS_TEMPLATE is given, but still don't write .gfs

jef-n:


rouault:
I've just backported it

Issue #1974, opened 2019-11-01
GDAL 3.0.1: Transforming pixel, line position to latlon produces invalid results


BaconRanch:
## Expected behavior and actual behavior.

On older versions of GDAL (tested on 2.2.3), the below Python code produces correct output:
`24.8871064142 60.3400617637 0.0`

On GDAL 3.0.1, the below code produces:
`inf inf inf`

Note that the issue has something to do with the given projection / geotransform.  Other projections and geotransformations work fine.

## Steps to reproduce the problem.
```
from osgeo import gdal
from osgeo.osr import SpatialReference, CoordinateTransformation

geotransform = (25493659.366596557, 5.290018027612359, 0.0, 6692072.133870748, 0.0, -5.290858871228285)
projection = 'PROJCS["ETRS89 / GK25FIN",GEOGCS["ETRS89",DATUM["European_Terrestrial_Reference_System_1989",SPHEROID["GRS 1980",6378137,298.257222101,AUTHORITY["EPSG","7019"]],TOWGS84[0,0,0,0,0,0,0],AUTHORITY["EPSG","6258"]],PRIMEM["Greenwich",0,AUTHORITY["EPSG","8901"]],UNIT["degree",0.0174532925199433,AUTHORITY["EPSG","9122"]],AUTHORITY["EPSG","4258"]],PROJECTION["Transverse_Mercator"],PARAMETER["latitude_of_origin",0],PARAMETER["central_meridian",25],PARAMETER["scale_factor",1],PARAMETER["false_easting",25500000],PARAMETER["false_northing",0],UNIT["metre",1,AUTHORITY["EPSG","9001"]],AXIS["Northing",NORTH],AXIS["Easting",EAST],AUTHORITY["EPSG","3879"]]'

source = SpatialReference()
source.ImportFromWkt(projection)

dest = SpatialReference()
dest.ImportFromEPSG(4326)

transform = CoordinateTransformation(source, dest)

x, y = gdal.ApplyGeoTransform(geotransform, 20, 20)
print transform.TransformPoint(x, y)
```

## Operating system

macOS 10.14.5

## GDAL version and provenance

GDAL 3.0.1 from homebrew tap [homebrew-osgeo4mac](https://github.com/OSGeo/homebrew-osgeo4mac/
)

rouault:
[...]

PR #3311, opened 2020-12-20
OGRGeometry::ExportToJson(): takes into account CRS and axis order to swap

rouault:
Adresses issue of https://github.com/OSGeo/gdal/issues/1974#issuecomment-748677786

I considered doing a reprojection to EPSG:4326 long,lat of any input CRS,
but it might break uses cases where one wants to use GeoJSON with other CRS.
So I have limited to fixing the axis order for EPSG geographic or projected CRS
whose axis order is northing/easting, latitude/longitude.

But this fix has the potential to break use cases of GDAL < 3 that still worked
up to now. Sigh...

### DIFF
--- a/gdal/ogr/ogrsf_frmts/geojson/ogrgeojsonwriter.cpp
+++ b/gdal/ogr/ogrsf_frmts/geojson/ogrgeojsonwriter.cpp
@@ -1502,8 +1502,25 @@ char* OGR_G_ExportToJsonEx( OGRGeometryH hGeometry, char** papszOptions )
     oOptions.nCoordPrecision = nCoordPrecision;
     oOptions.nSignificantFigures = nSignificantFigures;
 
+    // If the CRS has latitude, longitude (or northing, easting) axis order,
+    // and the data axis to SRS axis mapping doesn't change that order,
+    // then swap X and Y values.
+    bool bHasSwappedXY = false;
+    const auto poSRS = poGeometry->getSpatialReference();
+    if( poSRS &&
+        (poSRS->EPSGTreatsAsLatLong() || poSRS->EPSGTreatsAsNorthingEasting()) &&
+        poSRS->GetDataAxisToSRSAxisMapping() == std::vector<int>{1, 2} )
+    {
+        poGeometry->swapXY();
+        bHasSwappedXY = true;
+    }
+
     json_object* poObj =
        OGRGeoJSONWriteGeometry( poGeometry, oOptions );
+
+    // Unswap back
+    if( bHasSwappedXY )
+        poGeometry->swapXY();
 
     if( nullptr != poObj )
     {
--- a/autotest/ogr/ogr_geojson.py
+++ b/autotest/ogr/ogr_geojson.py
@@ -3107,3 +3107,50 @@ def test_ogr_geojson_feature_starting_with_big_properties():
                            '{"properties":{"foo":"%s"},"type":"Feature","geometry":null}' % ('x' * 10000))
     assert ogr.Open(filename) is not None
     gdal.Unlink(filename)
+
+
+###############################################################################
+
+def test_ogr_geojson_export_geometry_axis_order():
+
+    # EPSG:4326 and lat,long data order
+    sr = osr.SpatialReference()
+    sr.ImportFromEPSG(4326)
+    g = ogr.CreateGeometryFromWkt('POINT (49 2)')
+    g.AssignSpatialReference(sr)
+    before_wkt = g.ExportToWkt()
+    assert json.loads(g.ExportToJson()) == { "type": "Point", "coordinates": [ 2.0, 49.0 ] }
+    assert g.ExportToWkt() == before_wkt
+
+    # EPSG:4326 and long,lat data order
+    sr = osr.SpatialReference()
+    sr.ImportFromEPSG(4326)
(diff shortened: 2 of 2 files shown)


tjwebb:
> I considered doing a reprojection to EPSG:4326 long,lat of any input CRS

Yea, this sounds drastic and shouldn't be necessary. Irrespective of CRS, geojson is always lon,lat (or easting,northing, or x,y) per the spec: https://tools.ietf.org/html/rfc7946#appendix-A.1. So `ExportToJson` shouldn't try to be clever anywhere, and only needs to ensure `x,y` ordering always.

> But this fix has the potential to break use cases of GDAL < 3 that still worked up to now.

Well, this is the reason for major version changes, right? You get to change behavior if you want :) As long as its correct per the geojson spec. I never had an issue in 2.4.x, geojson always ended up correct regardless of which CRS I was reprojecting from. I only noticed the axes flip in geojson when I decided to upgrade to gdal 3.  Manually setting `target_prj.SetAxisMappingStrategy(osr.OAMS_TRADITIONAL_GIS_ORDER)` seems to work for now, although it'd be nice if we didn't need to specify a non-default parameter in order to get correct geojson output.

Anyway, go enjoy the holiday! This issue is annoying but not urgent, and it sounds like the maintainers should take some time to think this through in order to avoid any regrets.

gdal/ogr/ogrsf_frmts/geojson/ogrgeojsonwriter.cpp
```diff
@@ -1502,9 +1502,24 @@ char* OGR_G_ExportToJsonEx( OGRGeometryH hGeometry, char** papszOptions )
...
+    {
+        poGeometry->swapXY();
+        bHasSwappedXY = true;
+    }
+
     json_object* poObj =
        OGRGeoJSONWriteGeometry( poGeometry, oOptions );
 
+    // Unswap back
+    if( bHasSwappedXY )
```
tjwebb, line 1520: Ok, I see now why this fix is unpleasant. Is this unswap here to avoid compatibility problems?

jratike80:
People should pay no value to coordinate order in https://tools.ietf.org/html/rfc7946#appendix-A.1. The sentense is obviously unintentionally copied word by word from the 2008 GeoJSON specification https://geojson.org/geojson-spec.html. For the current specification any coordinate order is OK if there is "a prior arrangement", what ever it could be. So far I have not seen any machine readable example about such arrangement.


> 4.  Coordinate Reference System
> 
>    The coordinate reference system for all GeoJSON coordinates is a
>    geographic coordinate reference system, using the World Geodetic
>    System 1984 (WGS 84) [WGS84] datum, with longitude and latitude units
>    of decimal degrees.  This is equivalent to the coordinate reference
>    system identified by the Open Geospatial Consortium (OGC) URN
>    urn:ogc:def:crs:OGC::CRS84.  An OPTIONAL third-position element SHALL
>    be the height in meters above or below the WGS 84 reference
>    ellipsoid.  In the absence of elevation values, applications
>    sensitive to height or depth SHOULD interpret positions as being at
>    local ground or sea level.
> 
>    Note: the use of alternative coordinate reference systems was
>    specified in [GJ2008], but it has been removed from this version of
>    the specification because the use of different coordinate reference
>    systems -- especially in the manner specified in [GJ2008] -- has
>    proven to have interoperability issues.  In general, GeoJSON
>    processing software is not expected to have access to coordinate
>    reference system databases or to have network access to coordinate
>    reference system transformation parameters.  However, where all
>    involved parties have a prior arrangement, alternative coordinate
>    reference systems can be used without risk of data being
>    misinterpreted.

I wonder if GDAL could add some extra JSON member that would serve as a prior arrangement. Probably it should be anything else than "crs" that was removed from https://tools.ietf.org/html/rfc7946 and that is known and rejected by for example https://geojsonlint.com.

OGC API Features is using GeoJSON as a primary format and it has an accepted extension about crs https://docs.ogc.org/is/18-058/18-058.html. It is this brave when it comes to GeoJSON:

> An explicit request by a client with a query parameter crs establishes a prior arrangement. It is the responsibility of the client that submits the request to handle the coordinates in the response correctly. In particular, clients should not make the GeoJSON document available to others unless they are aware of the prior arrangement.
> 
> This standard does not specify any standardized approach to encoding coordinate reference system information in a GeoJSON document.

Why can't we just say "this is longitude-latitude" or "this is easting-northing" instead of saying "this is 2393, guess what it means?"

gdal/ogr/ogrsf_frmts/geojson/ogrgeojsonwriter.cpp
rouault: It is an implementation detail to avoid that the caller of the function sees the geometry altered. The caller of the function just wants the geometry to be exported to GeoJSON, likely not modified.

rouault:
> I wonder if GDAL could add some extra JSON member that would serve as a prior arrangement.

It does when exporting a whole layer, but here we are exporting a single geometry, and if pepole use that to compose their own collection, having a CRS per feature would be annoying. This fix intends to be minimal to make sure that the output is in longitude, latitude order when the passed geometry used EPSG:4326 with latitude, longitude order.

rouault:
Fix generalized to projected CRS with northing/easting order

PR #3315, opened 2020-12-22
gdal2xyz.py improvements and documentation

idanmiara:
## What does this PR do?

gdal2xyz.py improvements and documentation

* skip_ndv - allow to skip lines with uninteresting data values
	True/False - skip lines with NDV data points;
	Sequence/Number - skip lines with given value (per band or per ds)
* support more flexible input options for srcfile, skip, band_nums
* type hinting added
* add --allBands command line option
* add main function

## Tasklist

 - [ ] Add test case(s)
 - [x] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/gdal/swig/python/osgeo/utils/gdal2xyz.py
+++ b/gdal/swig/python/osgeo/utils/gdal2xyz.py
@@ -48,7 +48,8 @@ except ImportError:
 
 def Usage():
     print('Usage: gdal2xyz.py [-skip factor] [-srcwin xoff yoff width height]')
-    print('                   [-band b] [--allBands] [-csv] [--skipNoData]')
+    print('                   [-band b] [--allBands] [-csv]')
+    print('                   [--skipNoData] [-noDataValue value]')
     print('                   srcfile [dstfile]')
     print('')
     return 1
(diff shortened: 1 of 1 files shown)


gdal/swig/python/osgeo/utils/gdal2xyz.py
```diff
@@ -78,9 +85,15 @@ def main(argv):
             band_nums.append(int(argv[i + 1]))
             i = i + 1
 
+        elif arg == '--allBands':
+            band_nums = None
```
rouault, line 89: I find it a bit confusing that the semantics of [] or None is different. a allBands = True boolean would be more readable

idanmiara: are you referring only to main() or also to gdal2xyz()?
I using another parameter for bands in gdal2xyz would be confusing.
Now that we use Python3, Another conversion that we could adopt is using Ellipsis (...) for "auto" or "default" value, i.e.        ```elif arg == '--allBands':
            band_nums = ...```

rouault: let's forget my comment. I missed that the code was split in 2 functions. The code is OK as it

idanmiara: thanks!

idanmiara: @rouault I've addressed that issue you raised nonetheless.

idanmiara:
Thanks!

PR #3328, opened 2020-12-28
grib: Update MRMS Local Table to v12

NaderCHASER:
## What does this PR do?
Updates the GRIB table for the latest operational MRMS data, currently version 12.

Reference: https://www.nssl.noaa.gov/projects/mrms/operational/tables.php

None

## Environment
N/A


rouault:
I see some entries have been removed like 209, 9, 1 (perhaps others too). Couldn't they be used for older products ?

NaderCHASER:
> I see some entries have been removed like 209, 9, 1 (perhaps others too). Couldn't they be used for older products ?

I'd be happy to work on this since you make a good point. MRMS data is real-time however if anyone is archiving it we'd certainly want them to continue to be able to decode it.

I'll update the PR sometime today.

NaderCHASER:
@rouault I've brought back the old entries.

PR #3334, opened 2020-12-30
Backporting sign csharp assemblies as part of the netcore build process

szekerest:


rouault:
you can ignore the unrelated failure on the ubuntu_1604 target on Travis-CI

PR #3335, opened 2020-12-31
MRF: LERC1 compression improvements

lucianpls:
## What does this PR do?
A small algorithm tweak reduces the output size, no changes to the reader required.
Large areas of NaN or any other non-finite values get packed better
## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:


rouault:
Do we have tests that exercise this code path ?

lucianpls:
@rouault 
The checksum difference for the new test in 32bit mode is due to floating point precision differences

Issue #3319, opened 2020-12-25
v3.2.0 polygonize bug: some binary raster file leads to incorrect result


cartosquare:
## Expected behavior and actual behavior.

I expected to be able to vectorize this [binary raster file](https://1drv.ms/u/s!AiH6_Po0vGSNiFcI2oH440wWIND2), but the result is obviously incorrect.

the binary raster file screenshot:

https://1drv.ms/u/s!AiH6_Po0vGSNiFgIISK2LWj52Oil

the result file screenshot:

https://1drv.ms/u/s!AiH6_Po0vGSNiFm1ptzqvS7ucU5P

## Steps to reproduce the problem.

```bash
gdal_polygonize.py -mask fail_case_binary.tif fail_case_binary.tif fail_case_binary.shp
```

## Operating system

Ubuntu 18.04 64 bit

## GDAL version and provenance

GDAL 3.2.0, released 2020/10/26

rouault:
This is the same as https://github.com/OSGeo/gdal/issues/1158

cartosquare:
> This is the same as #1158

yes, the two issues ara both because of self intersection polygon. 
To avoid this, my solution is to isolate the pixel when it may cause self intersection. This solution will generate small polygons wich is just one pixel large. I also tested images in #1158, the generated vector areas(by DN value) now can exactly matching the raster image(by DN value).
Do you think this solution is acceptable?

rouault:
> Do you think this solution is work?

I'm not sure. I'm moderately familiar with the algorithm. As a first step, you could submit a pull request with your proposed fix, and extend the test suite with the simple test case of https://github.com/OSGeo/gdal/issues/1158#issuecomment-449743255 and similar ones

Issue #1158, opened 2018-12-19
Bug in Gdal.Polygonize results in missing polygon area?


WindingWinter:
I'm testing this on GDAL 2.3.1, Windows 10 64 bit

I believe that there is a bug in `Gdal.Polygonize`. I'm using Gdal.Net but I think one can easily reproduce the same problem in python binding or native C++.

There are two raster files inside the zip file: 

1. `loganw50.tiff`
2. `loganw150.tiff`

Run the two files with the following `polygonize` code:

```
        private static void Polygonize(string rasterFilePath)
        {
             var ds = Gdal.Open(rasterFilePath, Access.GA_ReadOnly);
            Debug.Assert(ds.RasterCount == 1);
            var band = ds.GetRasterBand(1);
            var vectorFilePath = Path.ChangeExtension(rasterFilePath, "shp");
            var ogrDriver = Ogr.GetDriverByName("ESRI Shapefile");
            var dst_ds = ogrDriver.CreateDataSource(vectorFilePath, null);
            var dst_layer = dst_ds.CreateLayer("Polygon", null, wkbGeometryType.wkbCompoundCurve, null); 

            var fieldDef = new FieldDefn("CellInfo", FieldType.OFTInteger);
            dst_layer.CreateField(fieldDef, 0);

            var outcome = Gdal.Polygonize(band, null, dst_layer, 0, null, null, null);
            var polygonArea = GetPolygonArea(dst_layer);
            var rasterArea = GetRasterArea(ds);
           
            var percentageDiff = Math.Abs((polygonArea - rasterArea) * 100 / rasterArea); //compute for missing area, when the algorithm is working perfectly it should have only at most rounding off errors as percentage difference. 
        }

        private static double GetPolygonArea(Layer dst_layer)
        {
            dst_layer.ResetReading();

            Feature feature;
            double polygonArea = 0;
            while ((feature = dst_layer.GetNextFeature()) != null)
            {
                var cellVal = feature.GetFieldAsInteger("CellInfo");
                var geometry = feature.GetGeometryRef();
                polygonArea += Math.Abs(geometry.GetArea());

            }

            return polygonArea;
        }

        private static double GetRasterArea(Dataset ds)
        {
            var nCol = ds.RasterXSize;
            var nRow = ds.RasterYSize;
            double[] geoTransform = new double[6];
            ds.GetGeoTransform(geoTransform);
            var cellSize = Math.Abs(geoTransform[5]);
[...]

PR #3337, opened 2021-01-01
polygonize: fix self intersection polygon when raster image contains complex holes

cartosquare:

Issue #1158
Issue #3319 

## What does this PR do?

To avoid generating self intersection polygons when polygonize a raster image.

If a connected region intersects itself at a pixel, think it as a self intersection and split the pixel from the connected region.

This solution will generate a polygon which is just one pixel large when there is an intersection. Although it may strange, but it does not harm the polygonize correctness.


jratike80:
OpenJUMP MakeValid function corrects the polygon  into this:
`POLYGON (( 0 4, 4 4, 4 0, 0 0, 0 4 ), ( 2 2, 2 1, 3 1, 3 2, 2 2 ), ( 2 2, 2 3, 1 3, 1 2, 2 2 ))`

What are the benefits of your solution? Could you continue and union the adjacent polygons which have the same attribute value for getting rid of the odd single pixel polygon (same result that MakeValid does)?

Have you measured the effect on speed with reasonably sized images?

autotest/alg/polygonize.py
```diff
@@ -181,15 +181,106 @@ def test_polygonize_4():
...
+    # Confirm we get the set of expected features in the output layer.
+
+    expected_feature_number = 4
+    assert mem_layer.GetFeatureCount() == expected_feature_number
+
+    expect = [1, 0, 1, 0]
+    expect_wkt = [
+        'POLYGON ((0 4,0 0,4 0,4 4,0 4),(1 3,1 1,3 1,3 2,2 2,2 3,1 3))',
+        'POLYGON ((1 3,1 2,2 2,2 3,1 3))',
+        'POLYGON ((1 2,1 1,2 1,2 2,1 2))',
```
rouault, line 225: This 3rd polygon shouldn't be emitted. It is within the first one

jratike80: It is not within the first polygon. That has a L-shaped hole that is filled with the three one pixel polygons. It is just that the third polygon could be merged with the first polygon because they both have pixel value=1. Now the result is indeed topologically correct but as merged it would be even better.

kuva

cartosquare:
> OpenJUMP MakeValid function corrects the polygon into this:
> `POLYGON (( 0 4, 4 4, 4 0, 0 0, 0 4 ), ( 2 2, 2 1, 3 1, 3 2, 2 2 ), ( 2 2, 2 3, 1 3, 1 2, 2 2 ))`
> 
> What are the benefits of your solution? Could you continue and union the adjacent polygons which have the same attribute value for getting rid of the odd single pixel polygon (same result that MakeValid does)?
> 
> Have you measured the effect on speed with reasonably sized images?

If we want to fix the polygon using MakeValid way, we can't split the pixel away, we have to merge it into the connected region and avoid causing self intersections when constructing the rings. I think it's not easy to do so.

Another solution is to fix self-intersection after ring construction, for example, call the MakeValid function(i don't know whether this function exists in gdal), however, it may slow down the speed.

The advantage of this solution is that the time complex will not be effected because we just doing several conditional judgment in the existing scanline loop. And i will report the speed compared to original version here when i get the results.

autotest/alg/polygonize.py
cartosquare: indeed, it's better to merge the 3rd polygon into first polygon. However. In this solution, it's a compromise between topo correctness and feature count correctness.

dbaston:
Since we know that polygons generated by this algorithm cannot have segment-segment self-intersections, it might not be too expensive to check their validity after construction and call MakeValid or some equivalent where needed.

cartosquare:
I tried to  fix self-intersection during the ring construction and  tested the algorithm on a 30000x30000 complex landuse classification image, the result is topo correct and the odd single pixel phenomenon is fixed. The time costs to polygonize this tif is 38m45s in my machine, original time costs is 37m18s.

jjimenezshaw:
@cartosquare Why have you closed the PR? Isn't it the proper solution?

jfbourdon:
I'm also interested in this PR, wasn't the solution good enough for now? It's better to always have valid polygons than having the next spatial operation failing occasionnally because of an invalid polygon. The current solution is to always call MakeValid after polygonization to avoid problems so having a fix built-in seems better to me.

cartosquare:
> @cartosquare Why have you closed the PR? Isn't it the proper solution?

This solution can fix the case i met(Issue #1158 and Issue #3319), however, I am not sure whether it can completely avoid self-intersections in other cases.

jfbourdon:
I can't pretend to have seen all possible cases, but your case is the only one I ever encountered. #3653 also report the same problem. I better solution might exists but in the meantime your current one will certainly help!

PR #3341, opened 2021-01-05
fix for #3343

idanmiara:
## What does this PR do?

a few minor fixes for https://github.com/OSGeo/gdal/pull/3343

## What are related issues/pull requests?

https://github.com/OSGeo/gdal/pull/3343

## Tasklist

 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments
 - [ ] All CI builds and checks have passed

PR #3342, opened 2021-01-05
gdal2xyz.py improvements - refactor and simplify and add option to return numpy arrays of the result values

idanmiara:
## What does this PR do?

gdal2xyz.py improvements - refactor and simplify and add option to return numpy arrays of the result values

https://github.com/OSGeo/gdal/pull/3341 (prerequisite) - please review it first

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/gdal/swig/python/osgeo/utils/gdal2xyz.py
+++ b/gdal/swig/python/osgeo/utils/gdal2xyz.py
@@ -34,16 +34,11 @@ import sys
 from numbers import Number
 from pathlib import Path
 from typing import Optional, Union, Sequence
+import numpy as np
 
 from osgeo import gdal
-
-try:
-    import numpy as Numeric
-except ImportError:
-    import Numeric
-
-
-# =============================================================================
+from osgeo.utils.auxiliary.base import num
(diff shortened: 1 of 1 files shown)


idanmiara:
I see that `import numpy` fails on macos and minigw.

What's the deal with this `Numeric` ? is it still relevant?
I'm a bit puzzled about the comment about `Numpy/Numeric` in `README.rst` - it seems that `gdalnumeric` just imports `numpy` so how does that help?

rouault:
> I see that `import numpy` fails on macos and minigw.

yes, the CI doesn't install numpy for those. I guess we could require numpy to build the Python bindings. That could simplify things and make autotest more obvious in some places. But please do that in another PR if you want to tackle that

> I'm a bit puzzled about the comment about Numpy/Numeric in README.rst

My understanding is that gdalnumeric was a way to transition from the pre-SWIG python bindings era to the "new gen" SWIG python bindings. For the old gen bindings, import Numeric was done instead of import numpy

PR #3347, opened 2021-01-06
gdaltest: Fix handling of ContentTooShortError in gdalurlopen

dbaston:
## What does this PR do?

Fixes error handling in test suite file downloader.

The correct name of the exception is ContentTooShortError, and it
subclasses URLError so it must be listed before URLError if we want to
catch it.

## Tasklist

 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:

### DIFF
--- a/autotest/pymod/gdaltest_python3.py
+++ b/autotest/pymod/gdaltest_python3.py
@@ -40,7 +40,6 @@ import sys
 from queue import Queue
 from threading import Thread
 
-
 def run_func(func):
     try:
         result = func()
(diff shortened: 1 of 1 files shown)


rouault:
> The correct name of the exception is ContentTooShortError

How did you spot that error ?

dbaston:
I was experimenting with adding in HTTP credentials from `.netrc`. That was throwing an exception, and when execution gets to the point where we try to catch a `ContentTooShort` it throws another exception for the undefined symbol. I only noticed that because I had modified the caller to print all exceptions (currently they're silently swallowed.)

PR #3355, opened 2021-01-07
DIMAP: add support for multiple components / subdatasets

rouault:


sgillies:
@rouault what do subdataset names look like for this driver? Do they take the `driver_prefix:filename: subdataset` form or `driver_prefix:subdataset:filename` or some other form?

rouault:
> Do they take the `driver_prefix:filename: subdataset` form or `driver_prefix:subdataset:filename` or some other form?

DIMAP:"filename":subdataset like in most drivers

PR #3356, opened 2021-01-07
gdalnumeric/numeric/numpy/gdal_array refactor

idanmiara:
## What does this PR do?

remove old import numeric statements (use only import numpy); replace references for gdalnumeric with references for numpy or gdal_array

https://github.com/OSGeo/gdal/pull/3342#issuecomment-754743894
https://github.com/OSGeo/gdal/pull/3341#issuecomment-756069860

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/autotest/alg/warp.py
+++ b/autotest/alg/warp.py
@@ -36,8 +36,7 @@ import os
 import shutil
 import sys
 
-from osgeo import gdal
-from osgeo import osr
+from osgeo import gdal, osr
 
 import gdaltest
 import pytest
@@ -1387,7 +1386,7 @@ def test_warp_41():
     for i in range(6):
         assert src_gt[i] == pytest.approx(vrt_gt[i], abs=1e-5)
 
-    
+
 ###############################################################################
 
 # Maximum
@@ -1642,7 +1641,7 @@ def test_warp_53():
                     print(option)
                     pytest.fail(alg_name)
 
-    
+
 ###############################################################################
 # Test Alpha on UInt16/Int16
 
@@ -1701,8 +1700,6 @@ def test_warp_55():
 def test_warp_56():
 
     try:
-        from osgeo import gdalnumeric
-        gdalnumeric.zeros
         import numpy
     except (ImportError, AttributeError):
         pytest.skip()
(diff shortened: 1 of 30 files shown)


idanmiara:
`gdaltest.numpy_drv = gdal.GetDriverByName('NUMPY')`
It seems that I broke the NUMPY driver registration.  I'm not sure how...

rouault:
> It seems that I broke the NUMPY driver registration. I'm not sure how...

the NUMPY driver is only available after importing gdal_array

idanmiara:
> > It seems that I broke the NUMPY driver registration. I'm not sure how...
> 
> the NUMPY driver is only available after importing gdal_array

A-ha! 
Thanks.

gdal/swig/python/osgeo/__init__.py
```diff
@@ -139,3 +139,10 @@ def ver_str(ver):
     from warnings import warn, simplefilter
     simplefilter('always', DeprecationWarning)
     warn(msg, DeprecationWarning)
+
+
+try:
+    import numpy  # noqa
```
rouault, line 145: I'm a bit ambivalent about automatically importing numpy. This could potentially have performance implications in situations where people don't need numpy.

idanmiara: good point, I can think of these solutions:
1. https://stackoverflow.com/questions/14050281/how-to-check-if-a-python-module-exists-without-importing-it
```
import importlib
numpy_available = importlib.util.find_spec("numpy") is not None
```
2. put `numpy_available`  by itself in `osgeo/numpy_available.py`
3. as before, duplicate the try/except clause instead of importing `numpy_available`

Any preference?

rouault: > Any preference?

probably 3

idanmiara: done

idanmiara:
@rouault Seems like a flake again. Can I trigger running the failed builds again ?

idanmiara:
thanks!

PR #3359, opened 2021-01-08
fixes not writing tms bug in the leaflet example if using --xyz

reyemtm:
## What does this PR do?
This simple change updates the script to fix the bug that if using `--xyz` the leaflet example does not output `tms: false`. There may be a better way to write this as I am just beginning with python so feel free to update without using this exact code.

None

Tested on Windows 10 latest


rouault:
thanks

PR #3361, opened 2021-01-08
add a symlink to LICENSE.TXT in .github subdir

idanmiara:
## What does this PR do?

move LICENSE.TXT from gdal subdir to the root dir so that it will be picked up by [github interface](https://github.com/OSGeo/gdal/community).

### DIFF
--- a/.github/LICENSE.TXT
+++ b/.github/LICENSE.TXT
@@ -1,1 +1,1 @@
-gdal/LICENSE.TXT
+../gdal/LICENSE.TXT
--- a/.github/LICENSE.TXT
+++ b/.github/LICENSE.TXT
@@ -1,0 +1,1 @@
+gdal/LICENSE.TXT
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -75,7 +75,7 @@ git checkout -b my_new_feature_branch
 git add my_new_file
 git add my_modifid_message
 git rm old_file
-git commit -a 
+git commit -a
 
 # you may need to resynchronize against master if you need some bugfix
 # or new capability that has been added since you created your branch
@@ -126,3 +126,6 @@ Things you should NOT do
 (For anyone with push rights to github.com/OSGeo/gdal) Never modify a commit or
 the history of anything that has been
 committed to https://github.com/OSGeo/gdal
+
+Committing symbolic links is allowed only under the .github directory in order to
+avoid potential problems on Windows.


rouault:
This will break tarballs that only include gdal/ content.
Maybe create a .github subdirectory in root/ and create a symlink to LICENSE.TXT (a bit like https://github.com/OSGeo/PROJ-data/tree/master/at_bev/.github)

idanmiara:
> This will break tarballs that only include gdal/ content.
> Maybe create a .github subdirectory in root/ and create a symlink to LICENSE.TXT (a bit like https://github.com/OSGeo/PROJ-data/tree/master/at_bev/.github)

It seems that git bash on windows just copies the files when I create a symlink. I guess I'll try it again later on Linux,

mloskot:
I suggest to avoid  symlinks.
Those can be troublesome on windows, require extra Git step to enable.
They proved to be PITA in Valhalla project, we've got rid of them.

idanmiara:
So maybe we just duplicate the License file to the root?

rouault:
> So maybe we just duplicate the License file to the root?

no, please

idanmiara:
> I suggest to avoid symlinks.
> Those can be troublesome on windows, require extra Git step to enable.
> They proved to be PITA in Valhalla project, we've got rid of them.

I've tried to clone https://github.com/OSGeo/PROJ-data on windows and `.github\README.md` appears just as a regular file with a relative path to the linked file as its contents. I understand why it can be confusing if someone wants to edit it, but otherwise what problems did you have with symlinks on windows?

mloskot:
This particular case is different to what we experienced with Valhalla, but my concern was it may open can of issues, if:
- some packaging activity that refers to symlinked files is done on Windows
- once symlinks are accepted for README, some contributors may consider it sensible for files that are build preprocessing or compilation input
The latter was the case for Valhalla: symlinked `.proto` files was failing build on Windows

idanmiara:
> This particular case is different to what we experienced with Valhalla, but my concern was it may open can of issues, if:
> 
> * some packaging activity that refers to symlinked files is done on Windows
> * once symlinks are accepted for README, some contributors may consider it sensible for files that are build preprocessing or compilation input
> 
> The latter was the case for Valhalla: symlinked `.proto` files was failing build on Windows

Thanks for the details. I've added a comment on the CONTRIBUTING saying symlinks are only allowed in .github dir.

.github/LICENSE.TXT
```diff
@@ -0,0 +1 @@
+gdal/LICENSE.TXT
```
rcoup, line 1: this should be `../gdal/LICENSE.TXT` surely?

rcoup (changes requested):
```
gdal $ ls -l .github/LICENSE.TXT
lrwxrwxr-x 1 rcoup staff 16 Feb  8 10:01 .github/LICENSE.TXT -> gdal/LICENSE.TXT
```

.github/LICENSE.TXT
idanmiara: As fare as I understood from [this](https://mokacoding.com/blog/symliks-in-git/) it suppose to be relative to the root of the repo.

rcoup: I literally checked out the branch and ran `ls` in it above → the symlink is broken.

```
$ head .github/LICENSE.TXT
head: cannot open '.github/LICENSE.TXT' for reading: No such file or directory
```

Does it work for you?

idanmiara: Thanks, I'll recheck it and report back soon.

rouault, line 1:
```suggestion
../gdal/LICENSE.TXT
```

idanmiara: Thanks, I forget to finish it...

rouault:
It doesn't look it helped github identify the license: https://github.com/OSGeo/gdal/community

sthagen:
Maybe one could move the license to the root of the tree, where it belongs as it is „the“ license for the tree.
Building the packaging could then place the License file copy where  the recipients of the package expect it to be.

So, still there are copies, but controlled by process and matching the subtree split such a subfolder only packaging represents.

But maybe I misunderstood the rationale. 

To me the most important thing is the top level placement when I analyze third party repositories and packages not so much some meta declaration. Every consumer should anyway perform a deep license hunt through the tree in case there is doubt the declaration is consistent.

The next helpful thing is, when the platforms offer in the usual metadata the license, that this is in all platforms. So, github top right, npm also top right, pypi left middle of the page etc. In my case this is nice to have not relevant if I can use or not, but others might just not consider a project where the platform metadata for the license is not stated.

I remember go psutil had the problem twice since 2017 that some files went in with GPL and not the more liberal license of the tree.

All the best,
Stefan

idanmiara:
redundant as per `git mv gdal/* .` in https://github.com/OSGeo/gdal/pull/4663

sthagen:
> redundant as per `git mv gdal/* .` in #4663

More precisely:
> remaining issues of PR #3361 will be resolved per PR #4663 (after release of version 3.4)?

PR #3363, opened 2021-01-10
netCDF: open /vsimem/ files without requiring Linux userfaultfd mechanism

rouault:


piyushrpt:
Would this mean, we no longer need to stage userfaultfd.h here: 
https://github.com/conda-forge/gdal-feedstock/blob/f2ca3c0e37dbb8f6127453a10a666e4be99e5ea6/recipe/build.sh#L23

rouault:
> Would this mean, we no longer need to stage userfaultfd.h here:

no, you still need if you want to read from other /vsi subsystems, like /vsicurl/ , /vsizip/ etc

PR #3366, opened 2021-01-10
Remove DB2 driver

rouault:
For consistency and pretty much the same reasons as
https://github.com/qgis/QGIS/pull/40930


rouault:
More general discussion started at https://lists.osgeo.org/pipermail/gdal-dev/2021-January/053260.html

rouault:
DB2 driver will now be disabled by default at runtime per https://github.com/OSGeo/gdal/pull/3505

PR #3369, opened 2021-01-11
PCIDSK: Fixed remaining address and undefined sanitizer fuzzer errors.

pciphelippeneveu:
The pcidsk_fuzzer exposed a bug in VSICurlHandle::Read(). The values were: iterOffset=280, nOffsetToDownload=0, osRegion.size()=272. I didn't try to understand why the file offset seems to be past the end of file. But this change can do no harm and it removes a vunerability. I ran the gtiff_fuzzer for a day and there were no errors.

I ran the pcidsk_fuzzer for 2 days using -timeout=60 and the only errors left are slow-unit errors which I'm not sure how to fix.

### DIFF
--- a/gdal/frmts/pcidsk/sdk/blockdir/blocktilelayer.cpp
+++ b/gdal/frmts/pcidsk/sdk/blockdir/blocktilelayer.cpp
@@ -155,6 +155,10 @@ bool BlockTileLayer::IsCorrupted(void) const
     if (GetLayerType() == BLTDead)
         return false;
 
+    // The tile layer is corrupted when the image size is 0.
+    if (GetXSize() == 0 || GetYSize() == 0)
+        return true;
+
     uint64 nTileSize =
         static_cast<uint64>(GetTileXSize()) * GetTileYSize() * GetDataTypeSize();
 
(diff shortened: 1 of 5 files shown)


rouault:
can you attach the reproducer ? I'd like to understand the root cause

pciphelippeneveu:
I believe it is this file. Note that it now times out.

[crash-88040365b4bb3b95c62d31d696ed214e7c3ae50c.zip](https://github.com/OSGeo/gdal/files/5796245/crash-88040365b4bb3b95c62d31d696ed214e7c3ae50c.zip)

pciphelippeneveu:
> I'd like to understand the root cause

I noticed that in gdal/frmts/pcidsk/sdk/channel/cbandinterleavedchannel.cpp, we don't check the return value of Seek() or Read(). So we probably keep on trying to read something pass the end-of-file.

rouault:
> So we probably keep on trying to read something pass the end-of-file.

makes totally sense. Would you mind reworking a bit the commit history to separate the commit in cpl_vsil_curl.cpp from the ones in the PCIDSK SDK, since it is quite decoupled ?

pciphelippeneveu:
> Would you mind reworking a bit the commit history to separate the commit in cpl_vsil_curl.cpp from the ones in the PCIDSK SDK, since it is quite decoupled ?

Ok I will separate the commits.

pciphelippeneveu:
I'm not sure if this was the proper way to separate the commits, but I used 'git push --force'.

rouault:
> I'm not sure if this was the proper way to separate the commits, but I used 'git push --force'.

Perfect. thanks

PR #3384, opened 2021-01-15
Add automatic loading of configuration options from a file

rouault:
On driver registration, loading of configuration is attempted from a set of
predefined files.

The following locations are tried by :cpp:func:`CPLLoadConfigOptionsFromPredefinedFiles`:

 - the location pointed by the environment variable (or configuration option)
   GDAL_CONFIG_FILE is attempted first. If it set, the next steps are not
   attempted

 - for Unix builds, the location pointed by ${sysconfdir}/gdal/gdalrc is first
   attempted (where ${sysconfdir} evaluates to ${prefix}/etc, unless the
   ``--sysconfdir`` switch of ./configure has been invoked). Then  $(HOME)/.gdal/gdalrc
   is tried, potentially overriding what was loaded with the sysconfdir

 - for Windows builds, the location pointed by $(USERPROFILE)/.gdal/gdalrc
   is attempted.

A configuration file is a text file in a .ini style format, that lists
configuration options and their values. Lines starting with `#` are comment lines.

Example:
```
[configoptions]
# set BAR as the value of configuration option FOO
FOO=BAR
```


gdal/port/cpl_conv.cpp
```diff
@@ -2007,6 +2007,111 @@ void CPL_STDCALL CPLFreeConfig()
...
+    while( (pszLine = CPLReadLine2L(fp, -1, nullptr)) != nullptr )
+    {
+        if( pszLine[0] == '#' )
+        {
+            // Comment line
+        }
+        else if( strcmp(pszLine, "[configoptions]") == 0 )
+        {
+            bInConfigOptions = true;
+        }
```
lucianpls, line 2049: To make the file extendable, add something like:
```
else if (STARTS_WITH(pszLine, "["))
{
  bInConfigOptions = false;
}
```

rouault: Done

rcoup:
@rouault can you document how these values interacts precedence-wise with environment variables and `--config FOO BAR`? Presumably `CPLSetConfigOption()`/`CPLSetThreadLocalConfigOption()` will always override it? Oh, unless they're set _before_ `GDALAllRegister()`?

rouault:
> can you document how these values interacts precedence-wise with environment variables and `--config FOO BAR`? Presumably `CPLSetConfigOption()`/`CPLSetThreadLocalConfigOption()` will always override it? Oh, unless they're set _before_ `GDALAllRegister()`?

Excellent question. So GDALAllRegister() in utilities I've checked is called before --config FOO BAR is evaluated. So --config will override any gdalrc settings, which I believe is the desired behaviour. I'll document that in the .rst

rcoup:
Does specifying specific config environment variables (eg `GDAL_CACHEMAX`) override the config options in the gdalrc/etc files?

rouault:
> Does specifying specific config environment variables (eg `GDAL_CACHEMAX`) override the config options in the gdalrc/etc files?

Not currently, but I guess we'd want env vars to have priority over gdalrc ? Thus in CPLLoadConfigOptionsFromFile() we should not call CPLSetConfigOption() if an environment variable of same name is already defined ?

rcoup:
Gut feel is that _ideally_:

Set via config API > cli `--config GDAL_FOO=1` > `$GDAL_FOO` environment > `$GDAL_CONFIG` file > user `gdalrc` > system `gdalrc` > defaults.

(First two are a bit of either/or though?)

PR #3398, opened 2021-01-20
CPLLoadConfigOptionsFromFile(): add an argument to define if env vars…

rouault:
… should have precedence, and make CPLLoadConfigOptionsFromPredefinedFiles() call it such that env vars have precedence over values of the config file

Addresses https://github.com/OSGeo/gdal/pull/3384#issuecomment-763836526


rcoup (approved):
LGTM 👍

PR #3404, opened 2021-01-22
utils.auxiliary fixes

idanmiara:
## What does this PR do?

Fix some bugs in utils.auxiliary
Add some types hinting, rename PathOrDS

### DIFF
--- a/gdal/swig/python/osgeo/utils/auxiliary/color_palette.py
+++ b/gdal/swig/python/osgeo/utils/auxiliary/color_palette.py
@@ -32,7 +32,6 @@ import os
 import re
 from xml.dom import minidom
 from collections import OrderedDict
-import glob
 import tempfile
 from typing import Sequence, Union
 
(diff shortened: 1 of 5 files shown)


gdal/swig/python/osgeo/utils/auxiliary/color_palette.py
```diff
@@ -352,13 +352,10 @@ def get_color_palette(color_palette_or_path_or_strings: ColorPaletteOrPathOrStri
     return pal
 
 
-def test_xml():
-    dir_path = 'sample/color_files'
+def test_xml(dir_path):
```
rcoup, line 355: Is this actually used? AFAICS the only caller was below

rcoup (approved):
LGTM! 😄  One question about an unused function.

gdal/swig/python/osgeo/utils/auxiliary/color_palette.py
idanmiara: Thanks for your comments.
I was planning to add a test to use this function with the actual test files but I didn't get to it yet.
Meanwhile I wanted to replace the reference to this hardcoded dir which only existed in my local setup.

rcoup: Suggest to drop it out and re-add on the appropriate future PR.

idanmiara: > Suggest to drop it out and re-add on the appropriate future PR.

done. thanks!

rcoup: [flake8 is failing](https://github.com/OSGeo/gdal/pull/3404/checks?check_run_id=1862571265) on `import glob` being unused

idanmiara: Thanks

PR #3399, opened 2021-01-20
Make creation of global hCOAMutex thread-safer on Windows

mloskot:
## What does this PR do?

This is an attempt to improve the "Ironically, creation of this initial mutex is not threadsafe" on Windows with implementation based on the `CRITICAL_SECTION`.

C++11 guarantees:
> Dynamic initialization of a block variable with static storage duration
>  is performed the first time control passes through its declaration (...)
> If control enters the declaration concurrently while the variable is
> being initialized, the concurrent execution shall wait for completion of the initialization.
~ http://eel.is/c++draft/stmt.dcl#4

MSVC guarantees:
> This behaviour is also available starting in Visual Studio 2015
> By default, Visual Studio enables this option.
~ https://docs.microsoft.com/en-us/cpp/build/reference/zc-threadsafeinit-thread-safe-local-static-initialization

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments
 - [ ] All CI builds and checks have passed

## Environment

Provide environment details, if relevant:

* OS: Windows
* Compiler: MSVC

PR #3405, opened 2021-01-22
Remove too strict CPLAssert checking global hCOAMutex

mloskot:
## What does this PR do?

Although adding an `CPLAssert(hCOAMutex);` might seem logical
here, do not enable it (see comment below). It calls CPLError that
uses the hCOAMutex itself leading to recursive mutex acquisition
and likely a stack overflow.

Fall back to this, ironically, NOT thread-safe re-initialisation of
hCOAMutex in case of a memory error or call to CPLCleanupMasterMutex
sequenced in an unusual, unexected or erroneous way.

For example, an unusual sequence could be:
  - GDALDriverManager has been instantiated,
  - then OGRCleanupAll is called which calls CPLCleanupMasterMutex,
  - then CPLFreeConfig is called which acquires the hCOAMutex
  - that has already been released and destroyed.

Refines #3399

## Tasklist

 - [x] Add test case(s) - already available MT tests do cover the unusual execution paths
 - [ ] Review
 - [x] Adjust for comments

## Environment

* OS: Windows
* Compiler: MSVC

### DIFF
--- a/gdal/port/cpl_multiproc.cpp
+++ b/gdal/port/cpl_multiproc.cpp
@@ -301,7 +301,7 @@ CPLMutex*& CPLCreateOrAcquireMasterMutexInternal(double dfWaitInSeconds = 1000.0
     {
         // Fall back to this, ironically, NOT thread-safe re-initialisation of
         // hCOAMutex in case of a memory error or call to CPLCleanupMasterMutex
-        // sequenced in an unusual, unexected or erroneous way.
+        // sequenced in an unusual, unexpected or erroneous way.
         // For example, an unusual sequence could be:
         //   GDALDriverManager has been instantiated,
         //   then OGRCleanupAll is called which calls CPLCleanupMasterMutex,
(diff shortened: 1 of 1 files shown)


gdal/port/cpl_multiproc.cpp
```diff
@@ -292,13 +292,22 @@ CPLMutex*& CPLCreateOrAcquireMasterMutexInternal(double dfWaitInSeconds = 1000.0
...
+    // uses the hCOAMutex itself leading to recursive mutex acquisition
+    // and likely a stack overflow.
 
     if ( !hCOAMutex )
     {
-        // In case of unexpected call to CPLCleanupMasterMutex,
-        // attempt this NOT thread-safe re-construction of the master mutex.
+        // Fall back to this, ironically, NOT thread-safe re-initialisation of
+        // hCOAMutex in case of a memory error or call to CPLCleanupMasterMutex
+        // sequenced in an unusual, unexected or erroneous way.
```
rouault, line 304:
```suggestion
        // sequenced in an unusual, unexpected or erroneous way.
```

mloskot: Thanks. Fixed.

p.s. Using the GitHub feature to "Commit suggestion" for the first time. Handy.

mloskot (commented):
I'm self-approving this

PR #203, opened 2017-03-14
Add new JPEGXR driver for reading/writing of JPEG XR format

mloskot:
The driver implementation is based on [jxrlib](https://jxrlib.codeplex.com) - an open source, BSD-licensed implementation of [JPEG XR](https://en.wikipedia.org/wiki/JPEG_XR) format by Microsoft.

## Status

Although the project started as an experiment, the driver is fairly usable now and I will continue to improve it.

The driver builds on Linux (Ubuntu, GCC, jxrlib from [libjxr](packages.ubuntu.com/search?keywords=libjxr&searchon=names) package) and Windows (VS2015 and VS2017).

For overview of currently supported features, please refer to `frmt_jpegxr.html` docs and TODO list below as well as test program `autotest/cpp/test_gdal_jpegxr.cpp` where all currently implemented features are verified in details.

## TODO

Work in progress tasklist:

- [x] Add usable dataset and band implementation (`gdal/frmts/jpegxr/jpegxrdataset.cpp`).
- [x] Add C++ test (`autotest/cpp/test_gdal_jpegxr.cpp`)
- [x] Add minimal integration with build configuration for Linux and Windows
- [x] Add documentation (`frmts/frmt_jpegxr.html`)
- [x] Decide on [jxrlib](https://jxrlib.codeplex.com) as ~~internal~~, external only ~~or both~~
    * Note: no official binaries for Windows; packages for Linux distros generally available.
- [x] Update Travis CI builds:
    * Install `libjxr` on Ubuntu 16.04 or later
- [x] Update AppVeyor builds:
    * Download and build jxrlib from release package, configure `nmake.local`
- [ ] Add Python test (`autotest/gdrivers/jpegxr.py`).
- [ ] Switch stats collection to `GDALChecksumImage` in tests.
- [ ] Add quantization levels support.
- [ ] Add support for chroma subsampling modes support.
- [ ] Update documentation to include all supported features

Future/Ideas:

- Add support for metadata reading/writing support.
- Add tiled raster support
- Add progressive mode support
- Add support for more JPEG XR-supported raster data types other than byte
- Add support for more JPEG XR-supported color formats.
- Find or generate test images in `.jxr` format with variety of pixel and color formats, image structure and channel organization.

------

I'd highly appreciate reviews, corrections, testing results as well as improvements to this driver.

The [driver announcement](https://lists.osgeo.org/pipermail/gdal-dev/2017-March/046263.html) posted to gdal-dev list.

PR #3411, opened 2021-01-25
[WIP] Add Zarr driver

davidbrochart:
## What does this PR do?
This is an effort to add a Zarr driver, based on [xtensor-zarr](https://github.com/xtensor-stack/xtensor-zarr).

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:


gdal/configure.ac
```diff
@@ -1630,7 +1664,7 @@ OGRFORMATS_ENABLED=
 OGRFORMATS_ENABLED_CFLAGS=
 OGRFORMATS_DISABLED=
 
-AC_DEFUN([INTERNAL_FORMATS],[aaigrid adrg aigrid airsar arg blx bmp bsb cals ceos ceos2 coasp cosar ctg dimap dted e00grid elas envisat ers esric fit gff gsg gxf hf2 idrisi ignfheightasciigrid ilwis ingr iris iso8211 jaxapalsar jdem kmlsuperoverlay l1b leveller map mrf msgn ngsgeoid nitf northwood pds prf r raw rmf rs2 safe saga sdts sentinel2 sgi sigdem srtmhgt stacta terragen tga til tsx usgsdem xpm xyz zmap])
+AC_DEFUN([INTERNAL_FORMATS],[aaigrid adrg aigrid airsar arg blx bmp bsb cals ceos ceos2 coasp cosar ctg dimap dted e00grid elas envisat ers esric fit gff gsg gxf hf2 idrisi ignfheightasciigrid ilwis ingr iris iso8211 jaxapalsar jdem kmlsuperoverlay l1b leveller map mrf msgn ngsgeoid nitf northwood pds prf r raw rmf rs2 safe saga sdts sentinel2 sgi sigdem srtmhgt stacta terragen tga til tsx usgsdem xpm xyz zarr zmap])
```
tbonfort, line 1667: zarr is not an internal driver as it requires a direct dependency. you should follow the same logic as in e.g. hdf5, kakadu, openjpeg...

gdal/doc/source/drivers/raster/zarr.rst
```diff
@@ -0,0 +1,21 @@
+.. _raster.zarr:
+
+================================================================================
+Zarr - Zarr
+================================================================================
+
+.. shortname:: Zarr
+
+.. versionadded:: 3.2.2
```
tbonfort, line 9: should be 3.3.0

.github/workflows/ubuntu_20.04/build.sh
```diff
@@ -51,7 +51,9 @@ CXXFLAGS="-std=c++17 -O1 $ARCH_FLAGS" CFLAGS="-O1 $ARCH_FLAGS" ./configure --pre
     --with-kea=/usr/bin/kea-config \
     --with-tiledb \
     --with-crypto \
-    --with-ecw=/opt/libecwj2-3.3
+    --with-ecw=/opt/libecwj2-3.3 \
+    --with-zarr \
```
tbonfort, line 55: would be nice if --with-zarr could take a non default install location

.github/workflows/ubuntu_20.04/build.sh
```diff
@@ -51,7 +51,9 @@ CXXFLAGS="-std=c++17 -O1 $ARCH_FLAGS" CFLAGS="-O1 $ARCH_FLAGS" ./configure --pre
...
+    --enable-driver-zarr
```
tbonfort, line 56: this should not be necessary with a correct configure script, --with-zarr should be sufficient to activate the driver

gdal/frmts/zarr/zarrdataset.cpp
```diff
@@ -0,0 +1,375 @@
...
+
+#include "xtensor-zarr/xzarr_hierarchy.hpp"
+#include "xtensor-zarr/xzarr_gdal_store.hpp"
+#include "xtensor-zarr/xzarr_file_system_store.hpp"
+#include "xtensor-io/xio_gzip.hpp"
+//#include "xtensor-io/xio_blosc.hpp"  // TODO
+
+#include <algorithm>
+
+CPL_CVSID("$Id$")
```
mloskot, line 41: By the way, I'm not sure why we keep this. Recently added STACTA driver also contains such line, likely, by copy & paste carry-over. Do we still need such cruft @rouault ?

gdal/frmts/zarr/zarrdataset.cpp
```diff
@@ -0,0 +1,375 @@
...
+
+/************************************************************************/
+/* ==================================================================== */
+/*                              ZarrDataset                             */
+/* ==================================================================== */
+/************************************************************************/
+
+template <typename T>
+void assign_chunk(void* pImage, xt::zarray& z, int nBlockYSize, int nBlockXSize,
+                  int nBlockYOff, int nBlockXOff,
```
mloskot, line 51: For general hygiene, such utilities in `.cpp` file should live in unnamed namespace:

```cpp
namespace
{
template <typename T>
void assign_chunk(....
}
```

gdal/frmts/zarr/zarrdataset.cpp
```diff
@@ -0,0 +1,375 @@
...
+template <class T>
+class ZarrRasterBand final: public GDALPamRasterBand
+{
+    friend class ZarrDataset;
+
+    xt::xzarr_hierarchy<T> m_zarr_hierarchy;
+
+  public:
+    ZarrRasterBand( ZarrDataset *, int, xt::xzarr_hierarchy<T>& );
+    virtual ~ZarrRasterBand();
```
mloskot, line 112: `virtual` is redundant, if anything it should have `~ZarrRasterBand() override`

gdal/frmts/zarr/zarrdataset.cpp
```diff
@@ -0,0 +1,375 @@
...
+{
+    friend class ZarrRasterBand<xt::xzarr_file_system_store>;
+    friend class ZarrRasterBand<xt::xzarr_gdal_store>;
+
+    VSILFILE    *fp;
+    GByte       abyHeader[1012];
+
+  public:
+                     ZarrDataset();
+                    ~ZarrDataset();
```
mloskot, line 83: Could read `~ZarrDataset() override`

gdal/frmts/zarr/zarrdataset.cpp
```diff
@@ -0,0 +1,375 @@
...
+{
+    friend class ZarrDataset;
+
+    xt::xzarr_hierarchy<T> m_zarr_hierarchy;
+
+  public:
+    ZarrRasterBand( ZarrDataset *, int, xt::xzarr_hierarchy<T>& );
+    virtual ~ZarrRasterBand();
+
+    virtual CPLErr IReadBlock( int, int, void * ) override;
```
mloskot, line 114: `virtual` is redundant

gdal/frmts/zarr/zarrdataset.cpp
```diff
@@ -0,0 +1,375 @@
...
+/************************************************************************/
+
+/************************************************************************/
+/*                            ZarrDataset()                             */
+/************************************************************************/
+
+ZarrDataset::ZarrDataset() :
+    fp(nullptr)
+{
+    std::fill_n(abyHeader, CPL_ARRAYSIZE(abyHeader), static_cast<GByte>(0));
```
mloskot, line 224: `std::memset` may be better optimised

davidbrochart:
Thanks a lot @mloskot and @tbonfort for the review!
There are a lot of things I was not sure about, this is just how I managed to get it working locally. I'll take your comments into account.

gdal/frmts/zarr/zarrdataset.cpp
```diff
@@ -0,0 +1,375 @@
...
+/************************************************************************/
+
+template <class T>
+GDALDataset* ZarrDataset::get_hierarchy(T& store, GDALOpenInfo* poOpenInfo)
+{
+    xt::xzarr_register_compressor<T, xt::xio_gzip_config>();
+    //xt::xzarr_register_compressor<T, xt::xio_blosc_config>();  // TODO
+
+    // Create a corresponding GDALDataset.
+    ZarrDataset *poDS = new ZarrDataset();
```
mloskot, line 295: Can any of the following `zarr_*` calls throw? Then this will leak.
Wrapping it with `std::unique_ptr` and `.release()` ing it at the `return` may be a bit safer way.

gdal/configure.ac
davidbrochart: Thanks @tbonfort, this should be fixed now.

gdal/doc/source/drivers/raster/zarr.rst
davidbrochart: :+1:

.github/workflows/ubuntu_20.04/build.sh
davidbrochart: What do you mean, we can pass any directory to `--with-zarr` already? Or do you mean in the CI specifically?

.github/workflows/ubuntu_20.04/build.sh
davidbrochart: :+1:

.github/workflows/ubuntu_20.04/build.sh
tbonfort: Sorry, I missed that part. I did mean that the non-standard path can be selected by the user outside of the CI.
However for checking for the lib/headers in default location you should use something similar to https://github.com/OSGeo/gdal/blob/64db832b3e475bb70d5acae99820c5ee405c9838/gdal/configure.ac#L2397 instead of looking up a hard-coded path. Namely -I/usr/include should never be added manually to the CFLAGS

davidbrochart:
Not sure why the `gdrivers/zarr.py` test is skipped.

mloskot:
@davidbrochart 

[This check](https://github.com/OSGeo/gdal/pull/3411/checks?check_run_id=1770747640#step:4:15271) says:

```
gdrivers/zarr.py::test_zarr_open SKIPPED (Zarr driver not found!)        [ 86%]
```

davidbrochart:
Thanks @mloskot, I'm testing only on Ubuntu 20.04 for now, and it seems the driver is found:
https://github.com/OSGeo/gdal/pull/3411/checks?check_run_id=1770747590#step:6:514
https://github.com/OSGeo/gdal/pull/3411/checks?check_run_id=1770747590#step:6:675
But the test is skipped:
https://github.com/OSGeo/gdal/pull/3411/checks?check_run_id=1770747590#step:12:472

mloskot:
That log says 
```
xtensor-zarr found - Zarr support enabled
```
but there is no trace of building anything in `gdal/frmts/zarr/`, no compilation command line with `-DFRMT_zarr`, etc.

I suspect `configure.ac` is giving false positive (due to a bug) and the driver is not enabled, hence the test is skipped.

nyalldawson:
Honest question: if the current trend in gdal is toward minimising the code base and removing drivers, doesn't that include limiting inclusion of totally new drivers like this?

mloskot:
@nyalldawson Feeling it's gonna happen, that is precisely why I long-delayed merge of my JPEG XR (https://github.com/OSGeo/gdal/pull/203) ;-) Seriously though, I think such question belongs to gdal-dev ML

davidbrochart:
@mloskot Do you think it's still worth working on this PR then?

mloskot:
@davidbrochart Honest answer: `¯\_(ツ)_/¯`. Helpful answer: ask on the mailing list, that's the right place for questions, doubts and discussions.

davidbrochart:
In the GDAL mailing-list [Tamas](https://lists.osgeo.org/pipermail/gdal-dev/2021-January/053366.html) mentioned the possibility of drivers being implemented as plugins. I was not aware of that, could you point me to examples of such drivers?

szekerest:
> In the GDAL mailing-list [Tamas](https://lists.osgeo.org/pipermail/gdal-dev/2021-January/053366.html) mentioned the possibility of drivers being implemented as plugins. I was not aware of that, could you point me to examples of such drivers?

The pdf driver would be a good example I think: https://github.com/OSGeo/gdal/tree/master/gdal/frmts/pdf

davidbrochart:
Thanks @szekerest! I will look at the PDF driver.

mloskot:
MrSID is another one https://github.com/OSGeo/gdal/tree/master/gdal/frmts/mrsid

hobu:
> Do you think it's still worth working on this PR then?

I think the answer is yes to this question. My only concern is you had to add special stuff to xtensor-zarr to support it, and users are likely to have stuff on packaging systems and such that's out of date for a while. This will cause some support load.

I don't think you need to chase the plugin thing at this time. Those are really oriented toward binary-only SDK situations.

davidbrochart:
> My only concern is you had to add special stuff to xtensor-zarr to support it

Are you referring to xtensor-zarr supporting GDAL's Virtual File System? If yes, it is partly true, but it's also an addition that can benefit xtensor-zarr independently from this driver.

hobu:
> Are you referring to xtensor-zarr supporting GDAL's Virtual File System? If yes, it is partly true, but it's also an addition that can benefit xtensor-zarr independently from this driver.

Yes. The only challenge is interlocking dependencies often cause "where do I get GDAL and xtensor-zarr compiled with XXX features so I can YYY?" -type questions. It isn't a thing for the project to fix, but it causes a bit of support load. 

I see you disabled the VFS stuff for now. I wasn't suggesting you needed to do that, but maybe it reduces that interlocking.

davidbrochart:
I need to investigate why it doesn't work with VFS, but I would have thought that if it were an issue with interlocking dependencies, it would not have even compiled.

gdal/frmts/zarr/zarrdataset.cpp
davidbrochart: Yes, they can throw. I wrapped the pointer with `std::unique_ptr`, thanks.

gdal/frmts/zarr/zarrdataset.cpp
davidbrochart: :+1:

sgillies:
@davidbrochart @hobu I'd like to see discussion on gdal-dev before this is merged. In my experience zarr is not so much a format as a library for storing and accessing arrays. What's the scope of this work? How many of zarr's compressors and storage schemes will be supported here? What about xtensor-zarr's warning that it is not ready for general use? Questions like these need to be explored yet.

davidbrochart:
> @davidbrochart @hobu I'd like to see discussion on gdal-dev before this is merged.

@sgillies I agree that more discussion is needed, in any case this PR is not ready to be merged yet. There's still work to be done in the xtensor stack for xtensor-zarr to be fully usable. I started this work as a potential application of xtensor-zarr based on [a discussion in the Zarr community](https://github.com/zarr-developers/community/issues/37).

> In my experience zarr is not so much a format as a library for storing and accessing arrays.

While I agree that Zarr is mostly know for [the Python library](https://github.com/zarr-developers/zarr-python), there exists implementations in other languages ([Zarr.js](https://github.com/gzuidhof/zarr.js) in JavaScript and now [xtensor-zarr](https://github.com/xtensor-stack/xtensor-zarr) in C++) and they all implement the Zarr protocol. Version 3 of the specification is being defined [here](https://github.com/zarr-developers/zarr-specs/blob/core-protocol-v3.0-dev/docs/protocol/core/v3.0.rst).

> How many of zarr's compressors and storage schemes will be supported here?

At the moment, xtensor-zarr only supports (uncompressed) raw binary, GZip and Blosc, but we plan to support more compressors. As for the storage schemes, xtensor-zarr supports the local file system, AWS S3, Google Cloud Storage, plus all the ones supported by GDAL's Virtual File System.

I understand that GDAL will require RFCs for new format drivers, and I would be happy to submit one for Zarr.
@rabernat and @jhamman will surely have valuable inputs here too.

sgillies:
Email sent to gdal-dev: https://lists.osgeo.org/pipermail/gdal-dev/2021-February/053384.html.

rabernat:
Just dropping in on this conversation from perspective of a zarr developer.

> In my experience zarr is not so much a format as a library for storing and accessing arrays.

Zarr is a specification, with many implementations in different languages. The python implementation is the most complete. The V2 spec is [currently under review with OGC](https://www.ogc.org/standards/requests/214) on the "community standard" track. The [V3 spec](https://zarr-specs.readthedocs.io/en/core-protocol-v3.0-dev/protocol/core/v3.0.html) is under active development and is open for public comment.

My understanding is that the xtensor-zarr implementation will support both versions of the spec.

davidbrochart:
> My understanding is that the xtensor-zarr implementation will support both versions of the spec.

That is correct, in addition to supporting Zarr v3, xtensor-zarr supports a part of Zarr v2. Actually this PR specifically targets version 2.

gdal/frmts/zarr/zarrdataset.cpp
rouault: > Do we still need such cruft @rouault ?

They are substitued by the mkgdaldist.sh script. This is probably only useful to spy and know which version of GDAL some binary build embeds.

gdal/frmts/zarr/zarrdataset.cpp
rouault: > Yes, they can throw.

If xtensor-zarr can throw exceptions, they must be caught by any of the methods of the GDAL driver API (virtual methods, and methods like Open()), since GDAL core doesn't expect exceptions to propagate to it.

gdal/frmts/zarr/zarrdataset.cpp
```diff
@@ -0,0 +1,393 @@
...
+            {
+                for( int j = 0; j < nBlockXSize; j++ )
+                {
+                    static_cast<T*>(pImage)[i * chunk_shape[1] + j] = typed_chunk(i, j);
+                }
+            }
+        }
+        catch (const std::runtime_error& e)
+        {
+            CPLError(CE_Failure, CPLE_AppDefined, "Error while processing Zarr array.\n");
```
rouault, line 69: EOL character not needed for CPLError(), and adding actual error message might help:
```suggestion
            CPLError(CE_Failure, CPLE_AppDefined, "Error while processing Zarr array: %s", e.what());
```

gdal/frmts/zarr/zarrdataset.cpp
```diff
@@ -0,0 +1,393 @@
...
+
+        // Create a corresponding GDALDataset.
+        poDS = new ZarrDataset();
+
+        auto zarr_hierarchy = xt::get_zarr_hierarchy(store, "2");
+        auto zarr_array = zarr_hierarchy.get_array("");
+    }
+    catch (const std::runtime_error& e)
+    {
+        CPLError(CE_Failure, CPLE_AppDefined, "Error loading Zarr store.\n");
```
rouault, line 302:
```suggestion
        CPLError(CE_Failure, CPLE_AppDefined, "Error loading Zarr store: %s", e.what());
```

gdal/frmts/zarr/zarrdataset.cpp
```diff
@@ -0,0 +1,393 @@
...
+        }
+        else
+        {
+            auto store = xt::xzarr_file_system_store(poOpenInfo->pszFilename);
+            return get_hierarchy(store, poOpenInfo);
+        }
+    {
+    catch (const std::runtime_error& e)
+    {
+        CPLError(CE_Failure, CPLE_AppDefined, "Error loading Zarr store.\n");
```
rouault, line 366:
```suggestion
        CPLError(CE_Failure, CPLE_AppDefined, "Error loading Zarr store: %s", e.what());
```

gdal/frmts/zarr/zarrdataset.cpp
davidbrochart: Thanks!

rouault:
I've given a try at building the driver. My initial attempt on Linux lead to my computer freezing to death due to lack of available RAM.

After rebooting with more RAM available, here are my observations:

| compiler | flags             | max RAM usage | zarrdataset.o size | stripped size | compilation time |
|----------|-------------------|---------------|--------------------|---------------|------------------|
| gcc 8.2  | -std=c++14 -O0 -g | 4.2 GB        |  81 438 200        | 14 623 880    |      1m35s       |
| gcc 8.2  | -std=c++14 -O2 -g | 5.0 GB        |  95 375 528        |  3 415 376    |      2m29s       |
| gcc 9.3  | -std=c++17 -O0 -g | 3.8 GB        | 104 340 912        | 14 996 936    |      1m20s       |
| gcc 9.3  | -std=c++17 -O1    | 3.3 GB        |   7 626 920        | 3 666 360     |      1m21s       |
| gcc 9.3  | -std=c++17 -O2    | 3.6 GB        |   7 562 584        | 3 691 480     |      1m40s       |
| gcc 9.3  | -std=c++17 -O2 -g | 4.7 GB        | 186 499 472        |  3 691 480    |     3m19s        |

So there's a very significant RAM usage required (around 5 GB for a -O2 -g build which is typical of how a package is built on a Linux distro). I'm afraid this could be a blocker on restricted build environments (let's say I want to build GDAL on some small cloud VM with just 2 GB RAM). Could that be significantly reduced ? Given that in a GDAL driver context, we just need to access the values in Zarr chunks without any fancy processing, I'm wondering if drawing xtensor and other dependencies isn't overkill. My initial gut feeling for a Zarr driver was to do it "at hand" without any extra dependencies (apart maybe from a few compression methods) than the one already used by GDAL
The disk usage is less a concern, although a 3.7 MB stripped binary size is quite significant for a single driver, compared to the 22 MB of the whole libgdal.so AMD64 Debian build (https://packages.debian.org/sid/libgdal27) with its > 200 drivers.

SylvainCorlay:
Hey @rouault, thanks for checking this out! Yeah it is a bit silly that we are not providing zarray in a precompiled form, which should resolve the RAM consumption issue, and mitigate the final binary size.

I will let @JohanMabille comment on this in greater detail.

JohanMabille:
The different features provided by `zarray` (and more generally by the whole xtensor stack) are well separated, making it easy to precompile only a subset of them. Therefore, we can definitely build a minimalist precompiled version of zarray with the operations you need only. This will avoid instantiating all the templates when building gdal and will limit the amount of RAM required. Especially if you don't need all the arithmetic and math operations (which are the most intensive to compile).

martindurant:
The discussion https://github.com/corteva/rioxarray/issues/246 seems pertinent here: if GDAL can handle fsspec (python) files, then zarr-in-gdal would be more useful - when called from python.

rouault:
Hi, I just wanted to mention that I'll have soon to work on a GDAL Zarr driver.  However, my current position would be to completely go from scratch, and not rely on xtensor-zarr.

There are several reasons for that:
- I'm not sure how much the RAM usage compilation issue has been solved by the move to a shared library. Hasn't it moved the issue at the time that shared library is built ?
- I'm not sure about the binary footprint, but the >3.7 MB I found felt too large. I'd expect a driver written at hand to just take at most a few hundred of kilobytes of binary footprint.
- If the driver now depends on an external shared library, that means it must be packaged in the distribution systems used by GDAL.  The only new external dependency from a driver written from scrach would be the compressors, and among them c-blosc, which I see it is available in Debian
- As far as I can see, the dtype of a zarr array must be registered at build time in https://github.com/xtensor-stack/xtensor-zarr/blob/master/include/xtensor-zarr/xzarr_chunked_array.hpp#L76 . I don't see how this could handle custom compound types that must be parsed at runtime. The GDAL multidimensional API that I plan to implement for this driver can use compound types (similarly to netCDF 4 / HDF5)
- the value of using the xtensor stack would probably be better as a frontend to GDAL, that is exposing any GDAL multidimensional array or any 2D traditionnal raster dataset as a zarray, not just Zarr. The value of zarray burried in the implementation of a driver appears to be very low, as a GDAL driver only exposes the raster/array values, and doesn't do any computation by itself.

davidbrochart:
Hi Even, all the concerns you raised are fair. We need to make a better job at making xtensor-zarr more lightweight when no computational capability is needed, which is the use-case of this driver. We have some ideas about reducing compilation time and resources. About custom data types, it is true that it is not possible today, although definitely feasible.
Thanks for letting me know about your choice of going from scratch, I'll close this PR.

rouault:
Another approach at creating a Zarr driver in https://github.com/OSGeo/gdal/pull/3896

PR #3412, opened 2021-01-25
Remove ODBC text trimming behaviour

pramsey:
Removes the white space trimming from the ODBC fetching code which removes spaces from ODBC text field on retrieval.

The behaviour makes it difficult to use ODBC results for some use cases that consider white space important, and was discovered here https://github.com/pramsey/pgsql-ogr-fdw/issues/208

Either the behaviour should be removed, or made conditional on a run-time option. A compile-time option seems not worth the effort. Initial PR for CI tests, and then can use this PR for final commit or run-time option implementation.

### DIFF
--- a/gdal/port/cpl_odbc.cpp
+++ b/gdal/port/cpl_odbc.cpp
@@ -365,7 +365,7 @@ int CPLODBCSession::Failed( int nRetCode, HSTMT hStmt )
                     achSQLState, &nNativeError,
                     reinterpret_cast<SQLCHAR *>(pachCurErrMsg),
                     nTextLength, &nTextLength2);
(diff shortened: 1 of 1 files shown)


havingsexonthebeach:
Digging into history shows that the whitespace trimming is as old as the initial implementation: https://github.com/OSGeo/gdal/commit/39c4d2b7ca24fb6887b64b0065f3c2c5e8c41c47#diff-22bf0bc3c1e2c8216ba95373356aaac4667ce29e46b067d14f7e64adf494540eR389

pramsey:
Once this passes CI, I am going to merge. If something breaks we can always put it back.

PR #3416, opened 2021-01-27
gdal2xyz.py - improvements and added documentation.

idanmiara:
## What does this PR do?

gdal2xyz.py - add `dst_nodata` option. 
gdal2xyz.py - add nodata to the return value of the gdal2xyz function
gdal2xyz.py - in order to be consistent with other utils - rename parameters skip_no_data -> skip_nodata; no_data_value -> src_nodata - currently only in master (not yet released).
gdal2xyz.rst - added documentation
index.rst - update with gdal2xyz.rst

added progress callback function; use exceptions when needed

### DIFF
--- a/autotest/pyscripts/test_gdal2xyz.py
+++ b/autotest/pyscripts/test_gdal2xyz.py
@@ -1,0 +1,83 @@
+#!/usr/bin/env pytest
+# -*- coding: utf-8 -*-
+###############################################################################
+#
+# Project:  GDAL/OGR Test Suite
+# Purpose:  gdal2xyz.py testing
+# Author:   Idan Miara <idan@miara.com>
+#
+###############################################################################
+# Copyright (c) 2021, Idan Miara <idan@miara.com>
+#
+# Permission is hereby granted, free of charge, to any person obtaining a
+# copy of this software and associated documentation files (the "Software"),
+# to deal in the Software without restriction, including without limitation
+# the rights to use, copy, modify, merge, publish, distribute, sublicense,
+# and/or sell copies of the Software, and to permit persons to whom the
+# Software is furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included
+# in all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+# OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+# THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+# FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+# DEALINGS IN THE SOFTWARE.
+###############################################################################
+
+import pytest
+
+# test that osgeo_utils is available, if not skip all tests
+pytest.importorskip('osgeo_utils')
+pytest.importorskip('numpy')
+
+from itertools import product
+import numpy as np
+
+from osgeo import gdal
+from osgeo.gdal_array import flip_code
+
+from osgeo_utils.auxiliary.raster_creation import create_flat_raster
+from osgeo_utils.samples import gdallocationinfo
+from osgeo_utils import gdal2xyz
+
+
+def test_gdal2xyz_py_1():
+    """ test get_ovr_idx, create_flat_raster """
+    pytest.importorskip('numpy')
+
+    size = (3, 3)
+    origin = (500_000, 0)
+    pixel_size = (10, -10)
+    nodata_value = 255
+    band_count = 2
+    dt = gdal.GDT_Byte
+    np_dt = flip_code(dt)
+    ds = create_flat_raster(
+        filename='', size=size, origin=origin, pixel_size=pixel_size,
+        nodata_value=nodata_value, fill_value=nodata_value, band_count=band_count, dt=dt)
+    src_nodata = nodata_value
+    np.random.seed()
+    for bnd_idx in range(band_count):
+        bnd = ds.GetRasterBand(bnd_idx + 1)
+        data = (np.random.random_sample(size) * 255).astype(np_dt)
+        data[1, 1] = src_nodata
+        bnd.WriteArray(data, 0, 0)
+    dst_nodata = 254
+    for pre_allocate_np_arrays, skip_nodata in product((True, False), (True, False)):
+        geo_x, geo_y, data, nodata = \
+            gdal2xyz.gdal2xyz(
+                ds, None, return_np_arrays=True,
+                src_nodata=src_nodata, dst_nodata=dst_nodata,
+                skip_nodata=skip_nodata, pre_allocate_np_arrays=pre_allocate_np_arrays)
+        _pixels, _lines, data2 = gdallocationinfo.gdallocationinfo(
+            ds, x=geo_x, y=geo_y,
+            resample_alg=gdal.GRIORA_NearestNeighbour,
+            srs=gdallocationinfo.LocationInfoSRS.SameAsDS_SRS)
+        data2[data2 == src_nodata] = dst_nodata
+        assert np.all(np.equal(data, data2))
+
+
--- a/gdal/swig/python/gdal-utils/osgeo_utils/gdal2xyz.py
+++ b/gdal/swig/python/gdal-utils/osgeo_utils/gdal2xyz.py
@@ -38,7 +38,8 @@ import numpy as np
 
 from osgeo import gdal
 from osgeo_utils.auxiliary.base import num, PathLike
-from osgeo_utils.auxiliary.util import PathOrDS
+from osgeo_utils.auxiliary.progress import get_progress_callback, OptionalProgressCallback
+from osgeo_utils.auxiliary.util import PathOrDS, get_bands
 from osgeo_utils.auxiliary.numpy_util import GDALTypeCodeAndNumericTypeCodeFromDataSet
 
 
@@ -142,7 +143,8 @@ def gdal2xyz(srcfile: PathOrDS, dstfile: PathLike = None,
              band_nums: Optional[Sequence[int]] = None, delim=' ',
              skip_nodata: bool = False,
              src_nodata: Optional[Union[Sequence, Number]] = None, dst_nodata: Optional[Union[Sequence, Number]] = None,
-             return_np_arrays: bool = False) -> Optional[Tuple]:
+             return_np_arrays: bool = False, pre_allocate_np_arrays: bool = True,
+             progress_callback: OptionalProgressCallback = ...) -> Optional[Tuple]:
     """
     translates a raster file (or dataset) into xyz format
 
@@ -160,34 +162,30 @@ def gdal2xyz(srcfile: PathOrDS, dstfile: PathLike = None,
     srcfile - The source dataset filename or dataset object
     dstfile - The output dataset filename; for dstfile=None - if return_np_arrays=False then output will be printed to stdout
     return_np_arrays - return numpy arrays of the result, otherwise returns None
+    pre_allocate_np_arrays - pre-allocated result arrays.
+        Should be faster unless skip_nodata and the input is very sparse thus most data points will be skipped.
+    progress_callback - progress callback function. use None for quiet or Ellipsis for using the default callback
     """
 
     result = None
 
+    progress_callback = get_progress_callback(progress_callback)
+
     # Open source file.
-    srcds = gdal.Open(str(srcfile), gdal.GA_ReadOnly) if isinstance(srcfile, (Path, str)) else srcfile
-    if srcds is None:
+    ds = gdal.Open(str(srcfile), gdal.GA_ReadOnly) if isinstance(srcfile, (Path, str)) else srcfile
+    if ds is None:
         raise Exception(f'Could not open {srcfile}.')
 
-    if not band_nums:
-        band_nums = list(range(1, srcds.RasterCount + 1))
-    elif isinstance(band_nums, int):
-        band_nums = [band_nums]
-    bands = []
-    for band_num in band_nums:
-        band = srcds.GetRasterBand(band_num)
-        if band is None:
-            raise Exception(f'Could not get band {band_num} from file {srcfile}')
-        bands.append(band)
+    bands = get_bands(ds, band_nums)
     band_count = len(bands)
 
-    gt = srcds.GetGeoTransform()
+    gt = ds.GetGeoTransform()
 
     # Collect information on all the source files.
     if srcwin is None:
-        srcwin = (0, 0, srcds.RasterXSize, srcds.RasterYSize)
+        srcwin = (0, 0, ds.RasterXSize, ds.RasterYSize)
 
-    dt, np_dt = GDALTypeCodeAndNumericTypeCodeFromDataSet(srcds)
+    dt, np_dt = GDALTypeCodeAndNumericTypeCodeFromDataSet(ds)
 
     # Open the output file.
     if dstfile is not None:
@@ -205,8 +203,8 @@ def gdal2xyz(srcfile: PathOrDS, dstfile: PathLike = None,
 
         # Setup an appropriate print format.
         if abs(gt[0]) < 180 and abs(gt[3]) < 180 \
-            and abs(srcds.RasterXSize * gt[1]) < 180 \
-            and abs(srcds.RasterYSize * gt[5]) < 180:
+            and abs(ds.RasterXSize * gt[1]) < 180 \
+            and abs(ds.RasterYSize * gt[5]) < 180:
             frmt = '%.10g' + delim + '%.10g' + delim + '%s'
         else:
             frmt = '%.3f' + delim + '%.3f' + delim + '%s'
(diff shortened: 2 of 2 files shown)


jratike80:
English in .rst if not most fluent but I fear I am not the best person to revise it.

gdal/doc/source/programs/gdal2xyz.rst
```diff
@@ -0,0 +1,97 @@
...
+        [-srcwin xoff yoff xsize ysize]
+        [-b band]* [-allbands]
+        [-csv]
+        [-srcnodata value] [-dstnodata value]
+        src_dataset [dst_dataset]
+
+Description
+-----------
+
+The :program:`gdal2xyz` utility can be used to translates a raster file into xyz format.
```
mloskot, line 29: `s/translates/translates/`

gdal/doc/source/programs/gdal2xyz.rst
```diff
@@ -0,0 +1,97 @@
...
+gdal2xyz is an alternative to `gdal_translate` with `-of=xyz` with other options,
```
mloskot, line 30: backticks around `gdal2xyz`

"with other" -> "with other options" ?

gdal/doc/source/programs/gdal2xyz.rst
```diff
@@ -0,0 +1,97 @@
...
+like selecting more then one band, skip or replace nodata value, and return output as numpy arrays.
```
mloskot, line 31: "like selecting" -> "for example, to select"

gdal/doc/source/programs/gdal2xyz.rst
```diff
@@ -0,0 +1,97 @@
...
+
+Examples
+--------
+
+::
+
+    gdal2xyz -b 1 -b 2 -dstnodata 0 input.tif output.txt
+
+
+To create a text in xyz format of input.tif, including bands 1 and 2, while replacing the dataset nodata values with zeros.
```
mloskot, line 97: backticks around `input.tif`

mloskot (changes requested):
A few typos to correct

gdal/doc/source/programs/gdal2xyz.rst
idanmiara: done

gdal/doc/source/programs/gdal2xyz.rst
idanmiara: done

gdal/doc/source/programs/gdal2xyz.rst
idanmiara: done

gdal/doc/source/programs/gdal2xyz.rst
idanmiara: I've rephrased that.

idanmiara:
> English in .rst if not most fluent but I fear I am not the best person to revise it.

Thanks for your comment, I've revised the whole file.

> A few typos to correct

Thanks for the review! I've pushed the fixes.

mloskot (approved):
LGTM, but I'm not representing enough authoritative vote here to merge it.

BTW, other `.py` script support both, short and long arguments, e.g.
https://github.com/OSGeo/gdal/blob/3c7e9db1ab4d28306a8700e38d2f4c0e2fa46b82/gdal/swig/python/scripts/gdal2tiles.py#L1338

idanmiara:
> LGTM, but I'm not representing enough authoritative vote here to merge it.
Thanks!

> BTW, other `.py` script support both, short and long arguments, e.g.
> https://github.com/OSGeo/gdal/blob/3c7e9db1ab4d28306a8700e38d2f4c0e2fa46b82/gdal/swig/python/scripts/gdal2tiles.py#L1338

In retrospect, I was thinking about using OptionParser but then it would have broken backwards compatibility as originally gdal2xyz used long arguments with single dash (as `gdal_translate` does). I might tackle that in a follow-up PR.

idanmiara:
@jratike80 @mloskot @EvertEt 
Thanks for your inputs. Who might be available to review and approve this PR?

rcoup:
@idanmiara I can review it later this week.

idanmiara:
> @idanmiara I can review it later this week.

@rcoup would you have the time to review it ? Thanks 😎

gdal/swig/python/osgeo/utils/gdal2xyz.py
```diff
@@ -119,28 +130,33 @@ def main(argv):
...
     skip - how many rows/cols to skip each iteration
-    band_nums - number of the bands to include in the output. None to include all bands.
-    delim - delimiter to use to separate columns
-    skip_no_data - `True` will skip lines with uninteresting data values (NoDataValue or a given value)
-    no_data_value - determinates which values to skip
-        `None` - skip lines which have the dataset NoDataValue;
-        Sequence/Number - skip lines with a given value (per band or per dataset)
+    srcwin (xoff, yoff, xsize, ysize) - Selects a subwindow from the source image for copying based on pixel/line location.
+    band_nums - selected input bands to process, None to process all.
+    delim - the delimiter to use between balues in a line
```
EvertEt, line 149:
```suggestion
    delim - the delimiter to use between values in a line
```

gdal/swig/python/osgeo/utils/gdal2xyz.py
```diff
@@ -119,28 +130,33 @@ def main(argv):
...
+    skip_nodata - Exclude from the output lines with nodata value (as determinate by srcnodata)
```
EvertEt, line 150:
```suggestion
    skip_nodata - Exclude the output lines with nodata value (as determined by srcnodata)
```

gdal/swig/python/osgeo/utils/gdal2xyz.py
```diff
@@ -119,28 +130,33 @@ def main(argv):
...
+    src_nodata - Determinants what is the nodata value of the dataset (for skipping or replacing)
```
EvertEt, line 151:
```suggestion
    src_nodata - The nodata value of the dataset (for skipping or replacing)
```

gdal/swig/python/osgeo/utils/gdal2xyz.py
```diff
@@ -37,15 +37,18 @@
...
+from osgeo.utils.auxiliary.util import PathOrDS
 from osgeo.utils.auxiliary.numpy_util import GDALTypeCodeAndNumericTypeCodeFromDataSet
 
 
 def Usage():
-    print('Usage: gdal2xyz.py [-skip factor] [-srcwin xoff yoff width height]')
-    print('                   [-band b] [--allBands] [-csv]')
-    print('                   [--skipNoData] [-noDataValue value]')
-    print('                   srcfile [dstfile]')
+    print('Usage: gdal2xyz.py [-help]'
```
EvertEt, line 46: Question: Help is not explicitly defined, I assume it will just trigger the main else case in parsing? 
Should we add it explicitly?

gdal/doc/source/programs/gdal2xyz.rst
```diff
@@ -0,0 +1,102 @@
...
+Synopsis
+--------
+
+.. code-block::
+
+    gdal2xyz [-help]
+        [-skip factor]
+        [-srcwin xoff yoff xsize ysize]
+        [-b band]* [-allbands]
+        [-csv]
```
EvertEt, line 22: I believe `skipnodata` is missing

gdal/doc/source/programs/gdal2xyz.rst
```diff
@@ -0,0 +1,102 @@
...
+
+    Select all input bands.
+
+.. option:: -csv
+
+    Use comma instead of space as a delimiter.
+
+.. option:: -skipnodata
+
+    Exclude from the output lines with nodata value (as determinate by srcnodata)
```
EvertEt, line 65:
```suggestion
    Exclude the output lines with nodata value (as determined by srcnodata)
```

gdal/doc/source/programs/gdal2xyz.rst
```diff
@@ -0,0 +1,102 @@
...
+
+.. option:: -srcnodata
+
+    Determinants what is the nodata value of the dataset (for skipping or replacing)
```
EvertEt, line 69:
```suggestion
    The nodata value of the dataset (for skipping or replacing)
```

gdal/swig/python/osgeo/utils/gdal2xyz.py
idanmiara: Right, I've added explicitly it as you suggested.

idanmiara:
Thanks for the review @EvertEt

idanmiara:
> > LGTM, but I'm not representing enough authoritative vote here to merge it.
> > Thanks!
> 
> > BTW, other `.py` script support both, short and long arguments, e.g.
> > https://github.com/OSGeo/gdal/blob/3c7e9db1ab4d28306a8700e38d2f4c0e2fa46b82/gdal/swig/python/scripts/gdal2tiles.py#L1338
> 
> In retrospect, I was thinking about using OptionParser but then it would have broken backwards compatibility as originally gdal2xyz used long arguments with single dash (as `gdal_translate` does). I might tackle that in a follow-up PR.

I've learned that `argparse.ArgumentParser` is a Python3 improvement for `optparse.OptionParser` that allows double-dash and single-dash for both short and long arguments.

It might be a good idea for another PR to rewrite all the scripts that use the manual argument and `gdal_calc` that uses `optparse.OptionParser` using the improved `argparse.ArgumentParser` to have a coherent argument parsing with all scripts and allow double dash and single dash for all arguments for backwards compatibility with all existing code.
Also, I'm not sure what's the point of `argv = gdal.GeneralCmdLineProcessor(argv)`, I can't find the docs for that function.

cc @rouault

idanmiara:
I've added progress callback function and raise exceptions when needed

idanmiara:
@rouault Is an issue with the CI ?

idanmiara:
I've added a progress bar and improved performance for the numpy array output. I'd be happy to get a review. Thanks!

idanmiara:
@rouault Please review :)

gdal/doc/source/programs/gdal2xyz.rst
```diff
@@ -0,0 +1,103 @@
+.. _gdal2xyz:
+
+================================================================================
+gdal2xyz
+================================================================================
+
```
rouault, line 6:
```suggestion

.. versionadded:: 3.3

```

gdal/doc/source/programs/gdal2xyz.rst
```diff
@@ -0,0 +1,103 @@
...
+`gdal2xyz` can be used as an alternative to `gdal_translate of=xyz`, but supporting other options,
+for example:
+
+    * Select more then one band
+    * Skip or replace nodata value
+    * Return the output as numpy arrays.
+
+.. program:: gdal2xyz
+
+.. option:: -skip
```
rouault, line 40:
```suggestion
.. option:: -skip <factor>
```

rouault (changes requested):
We would need tests

idanmiara:
> We would need tests

Thanks, I'm working on it.

idanmiara:
@rouault could you review please?
I'm not sure about `coord_is_pixel_center` in `gdal_coord_val.py` what does the `RasterIO` function expect as input? if I pass offset = `0.5` does it mean the center of the first pixel or the mid point between the first and the second pixel?

gdal/doc/source/index.rst
```diff
@@ -11,6 +11,7 @@ This documentation is also available as a `PDF file <gdal.pdf>`_.
 
     download
     programs/index
+    samples
```
rouault, line 14: Given they are "random quality/unsupported code", I would not put that at top level of the documentation. More as a paragraph of the python page

gdal/doc/source/programs/index.rst
```diff
@@ -45,6 +46,7 @@ Raster programs
    gdalcompare
    gdal_viewshed
    gdal_create
+   gdal_coord_val
```
rouault, line 49: This is a "duplicate" of gdallocationinfo. We don't want duplicated offficial utilities. It was fine as a python sample for people wanting to see how to do that in python

gdal/swig/python/gdal-utils/osgeo_utils/gdal_coord_val.py
```diff
@@ -0,0 +1,268 @@
...
+
+    bands = get_bands(ds, band_nums)
+    shape = (len(bands), count)
+    dt, np_dt = GDALTypeCodeAndNumericTypeCodeFromDataSet(ds)
+    nodata_value = [(band.GetNoDataValue() or 0) for band in bands]
+    nodata_value = np.asarray(nodata_value, dtype=np_dt)
+    results = np.empty(shape=shape, dtype=np_dt)
+
+    for i, (x, y) in enumerate(zip(x_arr, y_arr)):
+        if coord_is_pixel_center:
```
rouault, line 198: IMHO, this setting and what it does are useless. Just cast to int / floor should do the right thing.

gdal/doc/source/programs/index.rst
idanmiara: I didn't know about that one. 
Maybe I could make it to use the same API and rename it to `gdallocationinfo.py` ?

gdal/swig/python/gdal-utils/osgeo_utils/gdal_coord_val.py
idanmiara: It might be redundant if you are dealing with a sampling algorithms, but what about other samplers like average? doesn't RasterIO support these?

gdal/doc/source/index.rst
idanmiara: > Given they are "random quality/unsupported code", I would not put that at top level of the documentation. More as a paragraph of the python page

Not sure what you mean by "as a paragraph of the python page".
Maybe as a link from the programs level ?

rouault: I thought in https://gdal.org/api/python.html. Actually there's some mention of them already

gdal/doc/source/programs/index.rst
rouault: > Maybe I could make it to use the same API and rename it to `gdallocationinfo.py` ?

could make sense

gdal/swig/python/gdal-utils/osgeo_utils/gdal_coord_val.py
rouault: > It might be redundant if you are dealing with a sampling algorithms, but what about other samplers like average? doesn't RasterIO support these?

yes, but we'd better keep with the area-is-pixel GDAL is using to avoid confusion (or more exactly to let code outside GDAL be confused :-) and not import the confusion in our code base). What could be useful is to have some option where you'd compute bilinear interpolation from the surrounding 4 postings of the specified point. Using RasterIO resampling for that could be a bit cumbersome as it is designed to resample something with an extent not a point

idanmiara: Is there a Python API to sample point(s) values from a raster band that accepts int or float pixel/line, except for RasterIO?

gdal/doc/source/index.rst
idanmiara: I was thinking that the samples page could link to sample scripts in all the supporting languages.

gdal/swig/python/gdal-utils/osgeo_utils/gdal_coord_val.py
rouault: > Is there a Python API to sample point(s) values from a raster band that accepts int or float pixel/line, except for RasterIO?

no. Just thinking that ReadAsArray(float_x, float_y, 1, 1, resample_alg=gdal.GRIORA_Bilinear) might do the right thing. You should check that with (float_x % 1), (float_y % 1) slightly above (0,0), it gets the same value than sampling at int(float_x), int(float_y), and that at (float_x %1), (float_y % 1) slightly below (1,1), it gets the same value than at int(float_x)+1,int(float_y)+1

gdal/doc/source/index.rst
rouault: I'd probably keep the samples per language. Pythoners are probably not keen at looking at C# examples and vice-versa.

gdal/doc/source/programs/index.rst
idanmiara: done

gdal/doc/source/index.rst
idanmiara: > I'd probably keep the samples per language. Pythoners are probably not keen at looking at C# examples and vice-versa.

I see your point, but is there a page that points out where are the sample scripts in each language?

gdal/swig/python/gdal-utils/osgeo_utils/gdal_coord_val.py
idanmiara: > ReadAsArray(float_x, float_y, 1, 1, resample_alg=gdal.GRIORA_Bilinear)

That's what I've used, from an error message I've realized it is using RasterIO eventually, right?
I'll run the tests you advised.

idanmiara: @rouault I've ported all the functionality from `gdallocationinfo.cpp` into `gdallocationinfo.py` (except -xml, -lifonly).
The Python version now supports also float pixel/line for bilinear interpolation and multiple X Y points. as well as more SRS options.

As I want to support both Nearest Neighbor, Bilinear and both int, float arrays - I still can't figure out how to avoid using the workaround of the pixel_offset + line_offset that I've used (coord_is_pixel_center from before).
It seems that Bilinear interprets pixel=0.5 as "the midway between the first two pixels" and not "the middle of the first pixel". Now when I use inverse geotransform on the coordinate of the middle of the first pixel I get 0.5, which doesn't match with what Bilinear expects.
Do you have a better idea how to settle this issue?

gdal/doc/source/index.rst
idanmiara: I put python_samples just under python_gotchas. It seems like the right place :)

gdal/swig/python/gdal-utils/osgeo_utils/gdal_coord_val.py
idanmiara: `x[idx], y[idx], _z = ct.TransformPoint(x[idx], y[idx])` - is there a better way? `ct.TransformPoints` seem to expect something else...

idanmiara: After further thinking about the offet issue I think I understand it, RasterIO expects the corner of the extent of the area, thus for a pixel is the corner of the pixel, thus I needed to deduct 0.5 from the pixel and from the line in any case.
Right? 

I think we need a simple core function `PixelsIO` that accepts pixel/line buffers of some pixel centers and returns/sets the band data for the requested pixels.

There could be a faster/simpler version with int offsets for nearest neighbor.

What do you think?

rouault: GDAL's convention for georeferenced coordinates is that they correspond to the top-left corner of a pixel, not its center. I don't think we should modify this well established convention.
This PR is a bit too big to be reviewed. It should be split in one per utility.

idanmiara: OK :)
https://github.com/OSGeo/gdal/pull/3678
https://github.com/OSGeo/gdal/pull/3684
https://github.com/OSGeo/gdal/pull/3685

rouault:
gdal/swig/python/gdal-utils/osgeo_utils/samples/gdallocationinfo.py:32:1: F401 'array' imported but unused

rouault:
by the way gdallocationinfo shouldn't be in that pull request since there's a dedicated one for it

idanmiara:
> by the way gdallocationinfo shouldn't be in that pull request since there's a dedicated one for it

Thanks, I've rebased on that PR because the xyz test is against gdallocationinfo, that's why I've marked it as draft.
I hope that PR is ready now so I can rebase when it's merged.

PR #3452, opened 2021-02-11
Update gdalbuildvrt.rst

mehran66:
With the new parameter (i.e., -ignore_srcmaskband) in the new version of GDAL, there is no issue with merging images that have alpha channel and so the introduction should be revised.

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:


mehran66:
it seems that GDLA 3.3 has not been published yet!

PR #3458, opened 2021-02-15
Fix OGR silently ignores some rows in mdb with odbc driver

### LINKED ISSUE #3456: ogr silently ignores some rows in mdb with odbc driver
I came across some geometryless mdbs that while correctly reporting their feature count, they only return a subset of their features when read with ogr (ogrinfo or ogr2ogr) and odbc driver.

An offending mdb is [bad.zip](https://github.com/OSGeo/gdal/files/5976607/bad.zip)
The file originally had three thousand rows but only returned 2. After deleting some dozen other tables and most of the offending table's rows, it now returns 4 instead of 6 rows.
`ogrinfo bad.mbd PROP` correctly shows feature count 6, but incorrectly lists only 4 features.
`ogr2ogr bad.gpkg bad.mdb PROP` creates a geopackage containing only 4 features instead of 6.

mdbtools correctly exports all rows, ie `mdb-export bad.mdb PROP` returns 6 rows.
mdbtools odbc correctly list all rows, ie open the mdb in `isql` and run `select * from PROP` returns 6 rows.

## Operating system
Debian bullseye

GDAL 3.2.1, released 2020/12/29 (from debian repo)
mdbtools 0.9.1-1

nyalldawson:
Rows with NULLs for Memo fields where being discarded, due to
calls to SQLGetData for the column returning SQL_NO_DATA. This
was being treated as a non-fatal error, resulting in the
row being silently ignored. But according to the ODBC documentation
at https://docs.microsoft.com/en-us/sql/odbc/reference/develop-app/getting-long-data?view=sql-server-ver15
SQL_NO_DATA is a valid return non-error return code for SQLGetData:

"When there is no more data to return, SQLGetData returns SQL_NO_DATA"

and the example from that page uses the line:

  while ((rc = SQLGetData(hstmt, 2, SQL_C_BINARY, BinaryPtr, sizeof(BinaryPtr), &BinaryLenOrInd)) != SQL_NO_DATA) { ... }

accordingly treat return values of SQL_NO_DATA as equivalent
to a null value for the column, and don't abort reading the row.

Fixes missing rows in ODBC based drivers

Fixes #3456


autotest/ogr/ogr_odbc.py
```diff
@@ -173,15 +173,58 @@ def test_extensions():
...
+# Test reading mdb with null memo fields (https://github.com/OSGeo/gdal/pull/3458)
 
 
-def test_ogr_odbc_cleanup():
+def test_null_memo():
     if ogrtest.odbc_drv is None:
         pytest.skip()
 
-    gdal.Unlink('tmp/odbc.mdb')
+    ds = ogrtest.odbc_drv.Open('data/mdb/bad.mdb')
```
rouault, line 185: can the file be renamed to something like null_memo.mdb to be more explicit about its content ?

nyalldawson: done!

PR #3471, opened 2021-02-22
Docker: Rename BASE_IMAGE_NAME to TARGET_IMAGE . Add new parameter BASE_IMAGE

AndreaGiardini:
## What does this PR do?

The `BASE_IMAGE` argument is never passed to the dockerfile. This makes `ubuntu` the default base image, even if we specify `BASE_IMAGE_NAME` as specified in the readme.

The following code currently succeeds on master (even if the image doesn't exist). After this change the `docker build` process errors as expected.
```
BASE_IMAGE_NAME="nonexistingimage" ./ubuntu-small/build.sh --gdal v3.2.1 --proj 7.2.1 --platform linux/amd64
```

I also added explicitely the `USER root` in the dockerfile since some of these base images might not use it by default. For the current defaults, this is a noop.

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:

### DIFF
--- a/gdal/docker/README.md
+++ b/gdal/docker/README.md
@@ -126,10 +126,18 @@ Use the two script flags in order to leverage BuildKit:
 
 `alpine-small/build.sh --with-multi-arch --release --gdal v3.2.0 --proj master --platform linux/arm64,linux/amd64`
 
-## Custom Image Names
+## Custom Base Image
 
-Override the image and repository by setting the environment variable: `BASE_IMAGE_NAME`
+Override the base image, used to build and run gdal, by setting the environment variable: `BASE_IMAGE`
 
 **Example**
 
-`BASE_IMAGE_NAME="YOU_DOCKER_USERNAME/gdal" alpine-small/build.sh --release --gdal v3.2.0 --proj master`
+`BASE_IMAGE="debian:stable" ubuntu-small/build.sh --release --gdal v3.2.0 --proj master`
+
+## Custom Image Names
+
+Override the image and repository of the final image by setting the environment variable: `TARGET_IMAGE`
+
+**Example**
+
+`TARGET_IMAGE="YOU_DOCKER_USERNAME/gdal" alpine-small/build.sh --release --gdal v3.2.0 --proj master`
--- a/gdal/docker/alpine-normal/build.sh
+++ b/gdal/docker/alpine-normal/build.sh
@@ -19,6 +19,6 @@ esac
 
 export SCRIPT_DIR
 TAG_NAME=$(basename "${SCRIPT_DIR}")
-export BASE_IMAGE_NAME=${BASE_IMAGE_NAME:-osgeo/gdal:${TAG_NAME}}
+export TARGET_IMAGE=${TARGET_IMAGE:-osgeo/gdal:${TAG_NAME}}
 
 "${SCRIPT_DIR}/../util.sh" "$@" --test-python
--- a/gdal/docker/alpine-small/build.sh
+++ b/gdal/docker/alpine-small/build.sh
@@ -19,6 +19,6 @@ esac
 
 export SCRIPT_DIR
 TAG_NAME=$(basename "${SCRIPT_DIR}")
-export BASE_IMAGE_NAME=${BASE_IMAGE_NAME:-osgeo/gdal:${TAG_NAME}}
+export TARGET_IMAGE=${TARGET_IMAGE:-osgeo/gdal:${TAG_NAME}}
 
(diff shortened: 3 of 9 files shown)


rouault:
CC @awill1988 as you added this functionality. Was there an oversight in your initial commit regarding this ?

awill1988:
@rouault, thanks for looping me in. This was an intentional omission on my part.

The scope of my contribution was intended to support mutli-arch images and using alternative destinations for image repositories.

In the README.md, my description was intended for changing the *output* image name, i.e. publishing to `mydockerusername/gdal` on Dockerhub, rather than *input* image customization. So this statement:

> Override the image and repository by setting the environment variable: `BASE_IMAGE_NAME`

is intended for image publishing concerns

---

I do recall seeing the statement `FROM $BASE_IMAGE as builder` and thought 'What are the differences between `BASE_IMAGE_NAME` and `BASE_IMAGE`?'

At the time, I considered whether to introduce a different variable in the scripts to delineate between the building `BASE_IMAGE` and tagging image name, `BASE_IMAGE_NAME`.

Ultimately, I assumed that these two variables weren't intended to be treated the same, as far as I could tell at the time.

With that said, I think there's a little cognitive dissonance going on regarding what `BASE_IMAGE_NAME` & `BASE_IMAGE` means in the context of the build process.

Based on scripts for building docker images, I believe `BASE_IMAGE_NAME` (precedes my commit) was used for the output image name, and `BASE_IMAGE` was for overriding the Dockerfile in ubuntu specifically.

---

So @AndreaGiardini, I see the validity of your use case, but I wonder if it makes sense to use `--build-arg BASE_IMAGE=${BASE_IMAGE}` instead of `BASE_IMAGE_NAME`. I think as written, this could create a problem. Perhaps, BASE_IMAGE should be renamed?

AndreaGiardini:
Hey @awill1988 thank you for the detailed explanation, I now understand better the meaning behind `BASE_IMAGE` and `BASE_IMAGE_NAME`. Indeed, I found the naming very confusing.

In the Docker world, "base image" is usually the image that you specify in the Dockerfile's FROM instruction. To achieve what you mentioned above (changing the destination image) I would have rather used a name like `TARGET_IMAGE`. 

When I read:

> Override the image and repository by setting the environment variable: BASE_IMAGE_NAME

I thought the variable was used to override the image and repository of the base image, not the target one and I was surprised when it didn't work. The confusion comes from the fact that we use `BASE_IMAGE` to select the image in the `FROM` instruction and, at the same time, we use `BASE_IMAGE_NAME` to specify the output image.

To make naming more consistent, my proposal would be:
* Drop `BASE_IMAGE_NAME` variable in favour of something like `TARGET_IMAGE`
* Introduce `BASE_IMAGE` variable
* Document both in the README

Let me know what you think. @rouault any feedback?

rouault:
> @rouault any feedback?

Sounds good

AndreaGiardini:
I fixed the code according to the comment above. Let me know if you have any objection

awill1988:
That's awesome @AndreaGiardini, I appreciate you doing that and @rouault for connecting us. Looks good to me.

rouault:
@AndreaGiardini Giardini With your update applied, ``docker/ubuntu-small/build.sh`` (and similarly for ubuntu-full) fails with:
```
Using PROJ_DATUMGRID_LATEST_LAST_MODIFIED=Last-Modified: Sat, 15 Sep 2018 13:57:12 GMT
Using PROJ_VERSION=1a518773616fb001483c175e8dfab00c06e83c89
Using GDAL_VERSION=05624c2b7e3fa02ee7024dbbaaa650cdf7a7a230
Using GDAL_RELEASE_DATE=20210223
0104455adb9f        osgeo/gdal:gdal_rsync_daemon   "/bin/sh -c 'rsync -…"   About an hour ago   Up About an hour    23985/tcp           gdal_rsync_daemon
Sending build context to Docker daemon  11.78kB
Step 1/20 : ARG PROJ_INSTALL_PREFIX=/usr/local
Step 2/20 : ARG BASE_IMAGE=ubuntu:20.04
Step 3/20 : FROM $BASE_IMAGE as builder
base name ($BASE_IMAGE) should not be blank
```
(with docker.io 18.09.7-0ubuntu1~16.04.6)

AndreaGiardini:
Indeed... I forgot to test the "base" case. Sorry about that!

Happy to see it has already been fixed :)

PR #3472, opened 2021-02-22
AmigoCloud ogr driver: Fix date/datetime field type handling.

vchernetsky:
Handle the date and datetime field types as is, instead of converting them to string type.

### DIFF
--- a/gdal/ogr/ogrsf_frmts/amigocloud/ogramigocloudlayer.cpp
+++ b/gdal/ogr/ogrsf_frmts/amigocloud/ogramigocloudlayer.cpp
@@ -123,19 +123,7 @@ OGRFeature *OGRAmigoCloudLayer::BuildFeature(json_object* poRowObj)
             }
             else if( json_object_get_type(poVal) == json_type_string )
             {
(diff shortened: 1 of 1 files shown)


gdal/ogr/ogrsf_frmts/amigocloud/ogramigocloudlayer.cpp
```diff
@@ -123,19 +123,7 @@ OGRFeature *OGRAmigoCloudLayer::BuildFeature(json_object* poRowObj)
...
-                                             &sField) )
-                    {
-                        poFeature->SetField(i, &sField);
-                    }
-                }
-                else
-                {
-                    poFeature->SetField(i, json_object_get_string(poVal));
-                }
+                poFeature->SetField(i, json_object_get_string(poVal));
```
rouault, line 126: this change is possible since SetField(int, const char*) properly handles ISO 8601 format, right ?

vchernetsky: Yes, it seems to be working properly now.

PR #3476, opened 2021-02-23
Fix configure issues finding dependencies

QuLogic:
## What does this PR do?

This fixes a couple of bugs with finding GEOS and SFCGAL. Firstly, if `--with-geos` or `--with-sfcgal` are passed, but failed to be found, this was silently ignored. This should only happen if the option is 'auto', or people might think things were correctly found when they weren't. Secondly, it also saves `LDFLAGS` when testing compilability, as sometimes flags in `CFLAGS` and `LDFLAGS` go together. 

## Tasklist

 - [n/a] Add test case(s)
 - [n/a] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:

### DIFF
--- a/gdal/configure
+++ b/gdal/configure
@@ -38776,7 +38776,9 @@ $as_echo "no" >&6; }
 fi
 
 
-    ac_geos_config_auto=yes
+    if test x"$with_geos" = x"" ; then
+      ac_geos_config_auto=yes
+    fi
 
   else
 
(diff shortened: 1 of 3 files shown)


mloskot:
GEOS is not a [required dependency of GDAL](https://gdal.org/download.html#build-requirements)

QuLogic:
If you pass `--with-gdal`, then it should be required. If you pass nothing, it remains optional. What currently happens is a recipe for silent failure.

rouault:
Could you please regenerate configure with your changes, as it is currently under version control ?

tbonfort:
For reference, this PR introduced #3782

PR #3479, opened 2021-02-23
Docker: Pass BASE_IMAGE build arg conditionally

EvertEt:
I am by far experienced in writing bash scripts and had to google every syntax I used. Let me know if you want to rewrite anything. 

## What does this PR do?

Only sets the build arg when the BASE_IMAGE is set, to use the default in the dockerfile when unset. 

#3471

## Tasklist

 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments


rouault:
Thanks

PR #3480, opened 2021-02-24
Network file systems: add capabilities to define Content-Type and Content-Encoding at creation time

rouault:


tbonfort:
Given that this breaks backwards API compatibility of `VSIFilesystemHandler::Open` does this mean that the next release will be a 4.0 ? Or is `VSIFilesystemHandler::Open` not considered part of the public API ? (c.f.: https://github.com/airbusgeo/godal/blob/8a105ff/vsil_go.cpp#L39 breaks with this PR applied)

rouault:
> Given that this breaks backwards API compatibility of `VSIFilesystemHandler::Open` does this mean that the next release will be a 4.0 ?

This is a bit of regular discussion, without clear outcome. The semantic version numbering has been traditionally associated with the C API, not the C++ one which regularly breaks between feature versions (or at least the C++ ABI almost always break due to new virtual methods being added. Less frequent with the API itself). Here I guess I could have a virtual Open() with the 4 arguments falling back in its default implementation to the virtual Open() with 3 arguments, which would avoid the API breakage, but would still be a ABI change.

tbonfort:
> Here I guess I could have a virtual Open() with the 4 arguments falling back in its default implementation to the virtual Open() with 3 arguments, which would avoid the API breakage, but would still be a ABI change.
Ok thanks for the answer. I wasn't asking for you to fix this, it was more of an open question as to whether this implied a major version bump or not.

rouault:
> it was more of an open question as to whether this implied a major version bump or not.

It will definitely imply a .so.XX bump

PR #3483, opened 2021-02-24
AmigoCloud driver: fix json encoder to handle nested strings.

vchernetsky:
This change fixes the json encoder to escape nested string properly.


rouault:
On Travis-CI s390x (might be that char is unsigned on that architecture)
```
ogramigocloudtablelayer.cpp:73:28: error: comparison is always true due to limited range of data type [-Werror=type-limits]
                 if ('\x00' <= *c && *c <= '\x1f') {
                     ~~~~~~~^~~~~
```

vchernetsky:
Send a fix for that.

PR #3487, opened 2021-02-25
GTiff: allow customising overview compression type (fixes #3453) 

### LINKED ISSUE #3453: feature: COG overview compression configuration
### Background

We have 150TB of scanned historical imagery that we are looking to COG and store in AWS, We would like to leave the main image in a lossless format such as LZW. We would prefer to use a more optimized version for the overviews (Jpeg/WebP), so that we can keep the file size as small as possible

Looking at the source code for the geotiff driver, the compression for the overviews is set the same as the main image. If I manually modify the line and force it to compress into webp, the COG is still created and looks like it "works" in QGIS.

specifically if I hard code https://github.com/OSGeo/gdal/blob/master/gdal/frmts/gtiff/geotiff.cpp#L10310 to `COMPRESSION_WEBP` it generates the following tiff (it is 20GB and hosted in NZ)

https://public.lo.chard.com/new_zealand_sentinel_2019-2020_10m.vrt.lzw.cog.tiff

```
Id  	Size                	Tile Size           	Tile Count          	Resolution          	Compression         
0   	100800x151200       	394x591             	232854              	10                  	application/lzw     
1   	50400x75600         	197x296             	58312               	20                  	image/webp          
2   	25200x37800         	99x148              	14652               	40                  	image/webp          
3   	12600x18900         	50x74               	3700                	80                  	image/webp          
4   	6300x9450           	25x37               	925                 	160                 	image/webp          
5   	3150x4725           	13x19               	247                 	320                 	image/webp          
6   	1575x2362           	7x10                	70                  	640                 	image/webp          
7   	787x1181            	4x5                 	20                  	1280.8132           	image/webp          
8   	393x590             	2x3                 	6                   	2564.8854           	image/webp          
9   	196x295             	1x2                 	2                   	5142.8571           	image/webp          
10  	98x147              	1x1                 	1                   	10285.7142          	image/webp  
```

### Suggestion

Add `OVERVIEW_COMPRESS` (since `OVERVIEW_RESAMPLING` exists in the COG driver) to the COG driver so it can generate compressed overviews.

blacha:
## What does this PR do?

Adds `COMPRESS_OVERVIEW` as a configuration option for GTiff overview creation

#3453

## Tasklist

 - [x] Allow compression selection for overview creation
 - [x] Add test case(s)
 - [x] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/autotest/gcore/cog.py
+++ b/autotest/gcore/cog.py
@@ -268,16 +268,18 @@ def test_cog_creation_of_overviews_with_compression():
                             options='-of MEM -outsize 2048 300')
 
     ds = gdal.GetDriverByName('COG').CreateCopy(filename, src_ds,
-                                            options = ['COMPRESS=LZW', 'OVERVIEW_COMPRESS=JPEG'])
+                                            options = ['COMPRESS=LZW', 'OVERVIEW_COMPRESS=JPEG', 'OVERVIEW_QUALITY=50'])
 
     assert ds.GetRasterBand(1).GetOverviewCount() == 2
     assert ds.GetMetadata('IMAGE_STRUCTURE')['COMPRESSION'] == 'LZW'
 
     ds_overview_a = gdal.Open('GTIFF_DIR:2:' + filename)
     assert ds_overview_a.GetMetadata('IMAGE_STRUCTURE')['COMPRESSION'] == 'JPEG'
+    assert ds_overview_a.GetMetadata('IMAGE_STRUCTURE')['JPEG_QUALITY'] == '50'
 
     ds_overview_b = gdal.Open('GTIFF_DIR:3:' + filename)
     assert ds_overview_b.GetMetadata('IMAGE_STRUCTURE')['COMPRESSION'] == 'JPEG'
+    assert ds_overview_a.GetMetadata('IMAGE_STRUCTURE')['JPEG_QUALITY'] == '50'
 
     ds_overview_a = None
     ds_overview_b = None
--- a/autotest/gcore/cog.py
+++ b/autotest/gcore/cog.py
@@ -273,6 +273,15 @@ def test_cog_creation_of_overviews_with_compression():
 
         assert ds.GetRasterBand(1).GetOverviewCount() == 2
         assert ds.GetMetadata('IMAGE_STRUCTURE')['COMPRESSION'] == 'LZW'
+
+        ds_overview_a = gdal.Open('GTIFF_DIR:2:' + filename)
+        assert ds_overview_a.GetMetadata('IMAGE_STRUCTURE')['COMPRESSION'] == 'JPEG'
+
+        ds_overview_b = gdal.Open('GTIFF_DIR:3:' + filename)
+        assert ds_overview_b.GetMetadata('IMAGE_STRUCTURE')['COMPRESSION'] == 'JPEG'
+
+        ds_overview_a = None
+        ds_overview_b = None
         ds = None
 
     src_ds = None
--- a/autotest/gcore/cog.py
+++ b/autotest/gcore/cog.py
@@ -267,22 +267,21 @@ def test_cog_creation_of_overviews_with_compression():
     src_ds = gdal.Translate('', 'data/byte.tif',
                             options='-of MEM -outsize 2048 300')
 
-    with gdaltest.config_option('COMPRESS_OVERVIEW', 'JPEG'):
-        ds = gdal.GetDriverByName('COG').CreateCopy(filename, src_ds,
-                                                options = ['COMPRESS=LZW'])
+    ds = gdal.GetDriverByName('COG').CreateCopy(filename, src_ds,
+                                            options = ['COMPRESS=LZW', 'OVERVIEW_COMPRESS=JPEG'])
 
-        assert ds.GetRasterBand(1).GetOverviewCount() == 2
-        assert ds.GetMetadata('IMAGE_STRUCTURE')['COMPRESSION'] == 'LZW'
+    assert ds.GetRasterBand(1).GetOverviewCount() == 2
+    assert ds.GetMetadata('IMAGE_STRUCTURE')['COMPRESSION'] == 'LZW'
 
-        ds_overview_a = gdal.Open('GTIFF_DIR:2:' + filename)
-        assert ds_overview_a.GetMetadata('IMAGE_STRUCTURE')['COMPRESSION'] == 'JPEG'
+    ds_overview_a = gdal.Open('GTIFF_DIR:2:' + filename)
+    assert ds_overview_a.GetMetadata('IMAGE_STRUCTURE')['COMPRESSION'] == 'JPEG'
 
-        ds_overview_b = gdal.Open('GTIFF_DIR:3:' + filename)
-        assert ds_overview_b.GetMetadata('IMAGE_STRUCTURE')['COMPRESSION'] == 'JPEG'
+    ds_overview_b = gdal.Open('GTIFF_DIR:3:' + filename)
+    assert ds_overview_b.GetMetadata('IMAGE_STRUCTURE')['COMPRESSION'] == 'JPEG'
 
-        ds_overview_a = None
-        ds_overview_b = None
-        ds = None
+    ds_overview_a = None
+    ds_overview_b = None
+    ds = None
 
     src_ds = None
     gdal.GetDriverByName('GTiff').Delete(filename)
--- a/autotest/gcore/cog.py
+++ b/autotest/gcore/cog.py
@@ -258,6 +258,27 @@ def test_cog_creation_of_overviews():
     gdal.GetDriverByName('GTiff').Delete(filename)
     gdal.Unlink(directory)
 
+###############################################################################
+# Test creation of overviews with a different compression method
+
+def test_cog_creation_of_overviews_with_compression():
+    directory = '/vsimem/test_cog_creation_of_overviews_with_compression'
+    filename = directory + '/cog.tif'
+    src_ds = gdal.Translate('', 'data/byte.tif',
+                            options='-of MEM -outsize 2048 300')
+
+    with gdaltest.config_option('COMPRESS_OVERVIEW', 'JPEG'):
+        ds = gdal.GetDriverByName('COG').CreateCopy(filename, src_ds,
+                                                options = ['COMPRESS=LZW'])
+
+        assert ds.GetRasterBand(1).GetOverviewCount() == 2
+        assert ds.GetMetadata('IMAGE_STRUCTURE')['COMPRESSION'] == 'LZW'
+        ds = None
+
+    src_ds = None
+    gdal.GetDriverByName('GTiff').Delete(filename)
+    gdal.Unlink(directory)
+
 
 ###############################################################################
 # Test creation of overviews with a dataset with a mask
--- a/gdal/doc/source/drivers/raster/cog.rst
+++ b/gdal/doc/source/drivers/raster/cog.rst
@@ -234,6 +234,26 @@ Reprojection related creation options
 - **ADD_ALPHA=YES/NO**: Whether an alpha band is added in case of reprojection.
   Defaults to YES.
 
+Configuration options
+---------------------
+
+COG Overview creation can be configured with the :ref:`raster.gtiff` driver creation options
+
+-  **`COMPRESS_OVERVIEW`**:  See `Creation Options COMPRESS <#creation-options>`__ section.
+   Set the compression type to use for overviews
+-  **`PREDICTOR_OVERVIEW`**: Integer 1,2 or 3.
+   Set the predictor to use for overviews with LZW, DEFLATE and ZSTD compression
+-  **`JPEG_QUALITY_OVERVIEW`**: Integer between 0 and 100. Default value : 75.
+   Quality of JPEG compressed overviews, either internal or external.
+-  **`WEBP_LEVEL_OVERVIEW`**: Integer between 1 and 100. Default value : 75.
+   WEBP quality level of overviews, either internal or external.
+
+Example create a COG with the source data compressed as LZW and overviews compressed with WEBP with a WEBP quality of 90
+
+::
+
+    gdal_translate world.tif world_webmerc_cog.tif -of COG -co COMPRESS=LZW --config COMPRESS_OVERVIEW WEBP --config WEBP_LEVEL_OVERVIEW 90
+
 File format details
 -------------------
 
--- a/gdal/doc/source/drivers/raster/cog.rst
+++ b/gdal/doc/source/drivers/raster/cog.rst
@@ -237,7 +237,7 @@ Reprojection related creation options
 Configuration options
 ---------------------
 
-COG Overview creation can be configured with the :ref:`raster.gtiff` driver creation options
+COG Overview creation can be configured with the :ref:`raster.gtiff` driver configuration options (typically specified with ``--config NAME VALUE`` syntax)
 
 -  **`COMPRESS_OVERVIEW`**:  See `Creation Options COMPRESS <#creation-options>`__ section.
    Set the compression type to use for overviews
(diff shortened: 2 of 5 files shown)


autotest/gcore/cog.py
```diff
@@ -258,6 +258,27 @@ def my_cbk(pct, _, arg):
...
+    filename = directory + '/cog.tif'
+    src_ds = gdal.Translate('', 'data/byte.tif',
+                            options='-of MEM -outsize 2048 300')
+
+    with gdaltest.config_option('COMPRESS_OVERVIEW', 'JPEG'):
+        ds = gdal.GetDriverByName('COG').CreateCopy(filename, src_ds,
+                                                options = ['COMPRESS=LZW'])
+
+        assert ds.GetRasterBand(1).GetOverviewCount() == 2
+        assert ds.GetMetadata('IMAGE_STRUCTURE')['COMPRESSION'] == 'LZW'
```
blacha, line 275: I am not sure how to get the compression metadata for a overview, I couldn't find it in the docs

jratike80:
There is a somehow special relation between the GeoTIFF driver and COG driver. GTiff https://gdal.org/drivers/raster/gtiff.html has never had creation options for overviews. Overviews are created with for example gdaladdo and the compression method and options are controlled with config options.

The reason for the issue with overviews in COG is that COG adds overview handling into GTiff driver because with GTiff+addo the overviews are written last in the TIFF file that is not optimal for cloud usage.

I think that those configuration options should be documented for COG driver https://gdal.org/drivers/raster/cog.html, not for GTiff that does not create overviews. And perhaps it could be a creation option for COG like a few other creation options that already deal with overviews.

blacha:
> There is a somehow special relation between the GeoTIFF driver and COG driver. GTiff https://gdal.org/drivers/raster/gtiff.html has never had creation options for overviews. Overviews are created with for example gdaladdo and the compression method and options are controlled with config options.

The GTiff drive already has configuration options for overview compression `JPEG_QUALITY_OVERVIEW`  and `WEBP_LEVEL_OVERVIEW`, Currently the GTiff driver will re compress the overviews when they are added into the tiff from external sources. this is my main reasoning behind putting this logic here.

> I think that those configuration options should be documented for COG driver https://gdal.org/drivers/raster/cog.html, not for GTiff that does not create overviews. And perhaps it could be a creation option for COG like a few other creation options that already deal with overviews.

I originally had added the configuration options to the COG rst as well, but I took it out as none of the other configuration  options are documented in the COG driver

jratike80:
There are more config options which deal with overviews https://gdal.org/user/configoptions.html#list-of-configuration-options-and-where-they-apply. And gdal_translate has also an option COPY_SRC_OVERVIEWS that has some effect, sometimes https://trac.osgeo.org/gdal/ticket/3917

I am not sure how important it is to document the overview options on the GTiff page at all because most GDAL users can utilize those options only with gdaladdo but maybe it is good to document them also there so users know that such options exist.

COG driver has been made to buid reasonable overviews automatically but I think that you have found a good use case when the automatic way is not optimal. A OCG creation option for selecting the compression method of overviews feel reasonable.

I am a user, not developer so excuse me when I do not quite understand how this PR is fixing your main aim:

"Currently the GTiff driver will re compress the overviews when they are added into the tiff from external sources. this is my main reasoning behind putting this logic here."

autotest/gcore/cog.py
rouault: you can open ``GTIFF_DIR:x:foo.tif`` where x=1 for full resolution image, 2 for first overview level, etc.

rouault:
> I am a user, not developer so excuse me when I do not quite understand how this PR is fixing your main aim:

@blacha implementation will work with the COG driver given how the the COG driver uses the GTiff driver underneath. That said, I believe it could be indeed clearer from a user point of view to have *creation options* at the COG driver level (that would be documented there), that would set behind the scenes the existing config options used by the GTiff driver

blacha:
> That said, I believe it could be indeed clearer from a user point of view to have _creation options_ at the COG driver level (that would be documented there)

I am happy to update the docs to include the creation options into the COG driver, Ill look to doing that today 

> I am a user, not developer so excuse me when I do not quite understand how this PR is fixing your main aim:

Currently if you have a tiff that is compressed in `LZW`  then create overviews with `gdaladdo` and specify `webp` as the compression type, it will create a `.ovr` tiff file that contains overviews compressed in `webp`

If you then use that `.ovr` file and attempt to add it into a GTiff using the GTiff driver and do not specify the `COMPRESS` flag, it will recompress the overviews into the same compression format as what ever the source format (`LZW`)

There is logic currently in the GTiff driver that allows you to customise some of this process with configuration options, 

This pull request lets you customise the `COMPRESS` to use for the overviews when they are added into the `GTiff`

autotest/gcore/cog.py
blacha: took me way to long to realise that `GTIFF_DIR:` is a magic command for `gdal.Open` :) 

These tests now assert the overview compression

blacha:
@rouault I have added some basic docs to the COG docs, How detailed do you want the docs here? I have only added the overview configuration options as I am not 100% sure which other configuration options actually work with COGs

gdal/frmts/gtiff/geotiff.cpp
```diff
@@ -10248,14 +10248,34 @@ CPLErr GTiffDataset::CreateOverviewsFromSrcOverviews(GDALDataset* poSrcDS,
...
-        m_nCompression == COMPRESSION_ADOBE_DEFLATE ||
-        m_nCompression == COMPRESSION_ZSTD )
-        TIFFGetField( m_hTIFF, TIFFTAG_PREDICTOR, &nPredictor );
+    if( l_nCompression == COMPRESSION_LZW ||
+        l_nCompression == COMPRESSION_ADOBE_DEFLATE ||
+        l_nCompression == COMPRESSION_ZSTD ) 
+    {
+        if ( CPLGetConfigOption( "PREDICTOR_OVERVIEW", nullptr ) != nullptr ) 
+        {
+            nPredictor = atoi(CPLGetConfigOption("PREDICTOR_OVERVIEW","1"));
```
rouault, line 10271:
```suggestion
            nPredictor = static_cast<uint16>(atoi(CPLGetConfigOption("PREDICTOR_OVERVIEW","1")));
```

gdal/doc/source/drivers/raster/cog.rst
```diff
@@ -234,6 +234,26 @@ Reprojection related creation options
 - **ADD_ALPHA=YES/NO**: Whether an alpha band is added in case of reprojection.
   Defaults to YES.
 
+Configuration options
+---------------------
+
+COG Overview creation can be configured with the :ref:`raster.gtiff` driver creation options
```
rouault, line 240:
```suggestion
COG Overview creation can be configured with the :ref:`raster.gtiff` driver configuration options (typically specified with ``--config NAME VALUE`` syntax)
```

rouault:
> How detailed do you want the docs here?

That's fine. My previous remark about using creation options instead of configuration options still holds, but that can be let for a latter improvement. The reason for having the configuration options is that overview creation is something that is not typically in the scope of a creation option. The COG driver is a bit an exception in that it takes care of overview creation and thus potentially could expose overview options as creation options.

jratike80:
I would like to mention that normally those config options are not needed.

> By default the COG driver creates overviews with the same compression method than the full resolution image. The default behavior can be changed  with the :ref:`raster.gtiff` driver configuration options (typically specified with ``--config NAME VALUE`` syntax).

rouault:
> I would like to mention that normally those config options are not needed.
> 
> >

you can suggest changes from github UI. See https://haacked.com/archive/2019/06/03/suggested-changes/

gdal/doc/source/drivers/raster/cog.rst
```diff
@@ -234,6 +234,26 @@ Reprojection related creation options
 - **ADD_ALPHA=YES/NO**: Whether an alpha band is added in case of reprojection.
   Defaults to YES.
 
+Configuration options
+---------------------
+
+COG Overview creation can be configured with the :ref:`raster.gtiff` driver configuration options (typically specified with ``--config NAME VALUE`` syntax)
```
jratike80, line 240:
```suggestion
By default the COG driver creates overviews with the same compression method than the full resolution image. The default behavior can be changed with the :ref:raster.gtiff driver configuration options (typically specified with --config NAME VALUE syntax).
```

blacha:
> > How detailed do you want the docs here?
> 
> That's fine. My previous remark about using creation options instead of configuration options still holds, but that can be let for a latter improvement. 

Ahh I completely miss read that comment,  creation and configuration are too similar 😁

I will look at adding a creation to config mapping in the cog driver today.

gdal/frmts/gtiff/cogdriver.cpp
```diff
@@ -1011,6 +1011,36 @@ GDALDataset* GDALCOGCreator::Create(const char * pszFilename,
...
+    }
+
+    const char* osOvrQuality = CSLFetchNameValueDef(papszOptions, "OVERVIEW_QUALITY", "AUTO");
+    if ( !EQUAL(osOvrQuality, "AUTO") ) 
+    {
+        CPLSetConfigOption("JPEG_QUALITY_OVERVIEW", osOvrCompress);
+        CPLSetConfigOption("WEBP_LEVEL_OVERVIEW", osOvrCompress);
+    }
+
+    const char* osOvrPredictor = CSLFetchNameValueDef(papszOptions, "OVERVIEW_PREDICTOR", "FALSE");
```
blacha, line 1027: This was copied from the other `PREDICTOR` setting parsing just above this piece of code

blacha:
@rouault I have made an attempt at converting it to creation options, using the  `OVERVIEW_` prefix to be similar to the `OVERVIEW_RESAMPLING` method.  `OVERVIEW_COMPRESS`,  `OVERVIEW_QUALITY`,  `OVERVIEW_PREDICTOR` 

I haven't finished yet, I need to update the options in `InitializeCreationOptionList` just pushing it up so you can take a look at tell me what I've done wrong.

I am having a issue with the unit tests where the configuration options are persisted between the tests, should I be unsetting the `CPLSetConfigOption` some how after the tiff is created? or should I update the unit tests to reset the configuration options

gdal/frmts/gtiff/cogdriver.cpp
```diff
@@ -1011,6 +1011,36 @@ GDALDataset* GDALCOGCreator::Create(const char * pszFilename,
          }
     }
 
+    const char* osOvrCompress = CSLFetchNameValueDef(papszOptions, "OVERVIEW_COMPRESS", "AUTO");
+    if ( !EQUAL(osOvrCompress, "AUTO") ) 
+    {
+        CPLSetConfigOption("COMPRESS_OVERVIEW", osOvrCompress);
```
rouault, line 1017: Please use CPLConfigOptionSetter (cpl_conv.h) for a RAAI solution (the setter object must be put in a higher scope) that uses thread-local config options

rouault: You'll probably have to use a std::unique_ptr<CPLConfigOptionSetter>

blacha: Since they are just pass through values, I could just pass them directly through,

for example, I could replace
```
    const char* osOvrQuality = CSLFetchNameValueDef(papszOptions, "OVERVIEW_QUALITY", "AUTO");
    if ( !EQUAL(osOvrQuality, "AUTO") ) 
    {
        CPLSetConfigOption("JPEG_QUALITY_OVERVIEW", osOvrCompress);
        CPLSetConfigOption("WEBP_LEVEL_OVERVIEW", osOvrCompress);
    }
```
with 
```
    CPLConfigOptionSetter ovrQualityJpegSetter("JPEG_QUALITY_OVERVIEW", CSLFetchNameValueDef(papszOptions, "OVERVIEW_QUALITY", "75"), true);
    CPLConfigOptionSetter ovrQualityWebpSetter("WEBP_LEVEL_OVERVIEW", CSLFetchNameValueDef(papszOptions, "OVERVIEW_QUALITY", "75"), true);
```

And then remove the option of `"AUTO"`

otherwise i need to read some more about deconstructors and blocks scoping in c++

blacha: > for a RAAI solution 

also i am not sure what "RAAI" means in this context, Its been many years since Ive done any serious c/c++ programming, the internet thinks its "Resource Acquisition Is Initialization"  https://en.cppreference.com/w/cpp/language/raii which makes sense when we are talking about `CPLConfigOptionSetter`

jratike80: Just to be sure: is photometric still set automatically into YCBCR for JPEG compressed 3 band overviews? Even if the full resolution has some other compression?

rouault: > RAAI

Yes I meant RAII

>  is photometric still set automatically into YCBCR for JPEG compressed 3 band overviews? Even if the full resolution has some other compression?

good point. I don't think so currently. GTIFFWriteDirectory() is called with the m_nPhotometric value from the full resolution image. CreateOverviewsFromSrcOverviews() should also read PHOTOMETRIC_OVERVIEW, RegisterNewOverviewDataset allow to override m_nPhometric in the overview dataset and the COG driver to automatically set PHOTOMETRIC_OVERVIEW=YCBCR for OVERVIEW_COMPRESS=JPEG

blacha: I am not too familiar with the inner workings of JPEG and `YCBCR`, but should it be set automatically? are there any cases when the other types of `PHOTOMETRIC` should be included?

Looking at how the `PHOTOMETRIC` creation option is handled there is also a bunch of extra logic for other types eg `CMKYK` or `RGB` would they also need to be handled too? 

Is defaulting to `YCBCR` overviews for all COGs that have `OVERVIEW_COMPRESS=JPEG`  what you want to do?

rouault: > Is defaulting to `YCBCR` overviews for all COGs that have `OVERVIEW_COMPRESS=JPEG` what you want to do?

yes that's what we do for the full resolution imagery in the COG driver already, so doing it for overviews is consistent with that.

rouault:
As far as I can see what is missing now is adding the new creation options in the GDAL_DMD_CREATIONOPTIONLIST metadata item of the COG driver (GDALCOGDriver::InitializeCreationOptionList())

blacha:
> As far as I can see what is missing now is adding the new creation options in the GDAL_DMD_CREATIONOPTIONLIST metadata item of the COG driver (GDALCOGDriver::InitializeCreationOptionList())

Awesome thanks for looking over this, Ill add those in soon!

blacha:
@rouault I added the creation options into the XML, now the windows build is failing :(, I don't think it was caused by my code, but then I'm not sure how to easily force a rebuild to see if it fixes it self.

```
=========================== short test summary info ===========================
FAILED gdrivers/ngw.py::test_ngw_2 - AssertionError: Create datasource failed.
FAILED gdrivers/ngw.py::test_ngw_3 - AttributeError: 'NoneType' object has no...
FAILED gdrivers/ngw.py::test_ngw_4 - AttributeError: 'NoneType' object has no...
FAILED gdrivers/ngw.py::test_ngw_5 - AttributeError: module 'gdaltest' has no...
FAILED gdrivers/ngw.py::test_ngw_8 - AssertionError: Open of NGW:https://sand...
FAILED gdrivers/ngw.py::test_ngw_cleanup - AssertionError: Failed to delete d...
= 6 failed, 7074 passed, 1676 skipped, 2 xfailed, 2 warnings in 641.98s (0:10:41) =
```

gdal/doc/source/drivers/raster/cog.rst
```diff
@@ -160,6 +160,19 @@ General creation options
         available if general options (i.e. options which are not creation options,
         like subsetting, etc.) are used.
 
+- **OVERVIEW_COMPRESS**=[AUTO/NONE/LZW/JPEG/DEFLATE/ZSTD/WEBP/LERC/LERC_DEFLATE/LERC_ZSTD]**:
```
rouault, line 163:
```suggestion
- **OVERVIEW_COMPRESS=[AUTO/NONE/LZW/JPEG/DEFLATE/ZSTD/WEBP/LERC/LERC_DEFLATE/LERC_ZSTD]**:
```

PR #3504, opened 2021-02-27
Fix dispaly of two dashes instead of a single en-dash in SQLite and FileGDB doc pages

agiudiceandrea:
## What does this PR do?

Fixes the incorrect display of a single en-dash where two dashes should be displayed in 
SQLite / Spatialite RDBMS doc page https://gdal.org/drivers/vector/sqlite.html
image
and in ESRI File Geodatabase (FileGDB) doc page https://gdal.org/drivers/vector/filegdb.html
image


mloskot (approved):
Thank you

PR #3512, opened 2021-03-02
AmigoCloud driver: Fix SQL delete

vchernetsky:
SQL queries with DELETE, INSERT, UPDATE need to use HTTP POST.
SELECT should use HTTP GET.


vchernetsky:
I don't think the MacOS build fail is related to my changes.

PR #3520, opened 2021-03-04
FITS: apply vertical mirroring on reading/writing

rouault:
as expected by WCS translation.

Also adapt the PDS4 driver in CREATE_LABEL_ONLY=YES mode to alter
the <disp:vertical_display_direction> element (and warn if it is
missing)

Co-authored-by: Even Rouault <even.rouault@spatialys.com>


cmarmo:
Thanks @rouault!

PR #3533, opened 2021-03-05
CI: Skip AppVeyor for doc-only commits

dbaston:
This is a continuation of #3529 covering AppVeyor. Unfortunately it doesn't seem like an equally simple solution is possible for Travis.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -3,6 +3,12 @@ branches:
   except:
     - /(cherry-pick-)?backport-\d+-to-/
 
+# build will be skipped if all of the files match any of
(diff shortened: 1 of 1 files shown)


rouault:
It looks like this didn't work : see https://github.com/OSGeo/gdal/pull/3536

dbaston:
Hmm, I tested it successfully on a different repository, but obviously with a different filter pattern. Maybe the filter here is supposed to be `gdal/doc/*` -- that matches the [reference example](https://www.appveyor.com/docs/appveyor-yml/)

PR #3540, opened 2021-03-06
CI: Fix AppVeyor skip for doc-only commits

dbaston:
Attempt to fix AppVeyor skip config set in #3533. I will push a doc-only commit shortly to check that this is working.


dbaston:
Good to go, I think. Sorry for the noise.

PR #3556, opened 2021-03-08
Fixes related to non-finite coordinates in Shapefiles

rouault:


rouault:
First 3 commits backported to 3.2 branch (for 3.2.3)

PR #3560, opened 2021-03-09
Build system: Respect --libdir parameter

thesamesam:
## What does this PR do?
Modifies the build system to respect the --libdir
parameter.

When searching for various libraries, we want to use
the libdir passed in to autotools (--libdir)
to ensure that we find the library
for the correct ABI.

It is possible that we pick up the wrong
copy from e.g. /usr/lib/ where a 32-bit
copy of e.g. zlib exists
when we're in the middle of a 64-bit build.

Use ${libdir} to ensure we respect
any preferences/information given and
search for libraries in the right place.

No behaviour should change here as
--libdir defaults to 'lib'.

Bug: https://bugs.gentoo.org/696106
Signed-off-by: Sam James <sam@gentoo.org>

[Contributing](https://github.com/OSGeo/gdal/blob/master/CONTRIBUTING.md)

## Tasklist

## Environment
* OS: Gentoo GNU/Linux
* Compiler: GCC 10

### DIFF
--- a/gdal/configure.ac
+++ b/gdal/configure.ac
@@ -1121,11 +1121,11 @@ if test "$with_libz" = "external" -o "$with_libz" = "" -o "$with_libz" = "yes" ;
 
 elif test "$with_libz" != "no" -a "$with_libz" != "internal" ; then
 
-  LIBS="-L$with_libz -L$with_libz/lib $LIBS"
+  LIBS="-L$with_libz -L$with_libz/${libdir} $LIBS"
 
-  AC_CHECK_LIB(z,deflateInit_,LIBZ_SETTING=external,LIBZ_SETTING=internal,-L$with_libz -L$with_libz/lib -lz)
+  AC_CHECK_LIB(z,deflateInit_,LIBZ_SETTING=external,LIBZ_SETTING=internal,-L$with_libz -L$with_libz/${libdir} -lz)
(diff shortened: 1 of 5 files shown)


gdal/configure.ac
```diff
@@ -1414,52 +1414,15 @@ else
...
-            unset ac_cv_lib_proj_internal_proj_create_from_wkt
-            AC_LANG_PUSH([C++])
-            AC_CHECK_LIB(proj,internal_proj_create_from_wkt,PROJ_FOUND=yes,PROJ_FOUND=no,)
-            AC_LANG_POP([C++])
-        fi
-        if test "$PROJ_FOUND" = "yes"; then
-            PROJ_INCLUDE="-DPROJ_RENAME_SYMBOLS"
-        fi
-    fi
-    if test "$PROJ_FOUND" = "no"; then
```
rouault, line 1451: please keep that test that uses the libinternalproj hack

thesamesam: Ah, thanks!

rouault:
@thesamesam Would you mind committing the regenerated configure file ? (as it isn't regenerated by the continuous integration builds)

thesamesam:
@rouault Sure, of course (I should have realised). But let me make some changes, actually. There's an uncommon case which I didn't consider that I need to adjust for. I'll let you know when ready for re-review if that's okay.

PR #3565, opened 2021-03-10
GTiff: rely on libtiff for LERC codec instead of our internal libtiff only codec, and allow linking against external libLerc

rouault:
Since https://gitlab.com/libtiff/libtiff/-/merge_requests/228, upstream
libtiff has now the LERC codec. So remove our internal codec to rely on
libtiff.
Update internal libtiff and build scripts to rely on our internal
LercLib.

and

Build: allow linking against external libLerc
    
- If using external libtiff, LERC support in GDAL has no effect on TIFF LERC support.
- If using internal libtiff, TIFF LERC support is possible with either internal of external libLerc
- MRF LERC support currently requires internal libLerc


gdal/frmts/gtiff/geotiff.cpp
```diff
@@ -20061,18 +20016,25 @@ CPLString GTiffGetCompressValues(bool& bHasLZW,
             osCompressValues +=
                     "       <Value>WEBP</Value>";
         }
+        else if( c->scheme == COMPRESSION_LERC )
+        {
+            bHasLERC = true;
+            osCompressValues +=
+                    "       <Value>WEBP</Value>";
```
avalentino, line 20023: Probably this should be
```
"       <Value>LERC</Value>";
```

avalentino (changes requested):

gdal/frmts/gtiff/geotiff.cpp
rouault: Good catch!

avalentino (approved):

PR #3569, opened 2021-03-10
AmigoCloud driver: fix GetFeature

vchernetsky:
AmigoCloud uses fid of type String ("amigo_id"). This change is a work around to handle the query with nFeatureId as a string value.

### DIFF
--- a/gdal/ogr/ogrsf_frmts/amigocloud/ogramigocloudtablelayer.cpp
+++ b/gdal/ogr/ogrsf_frmts/amigocloud/ogramigocloudtablelayer.cpp
@@ -777,8 +777,9 @@ OGRFeature* OGRAmigoCloudTableLayer::GetFeature( GIntBig nFeatureId )
     CPLString osSQL = osSELECTWithoutWHERE;
     osSQL += " WHERE ";
     osSQL += OGRAMIGOCLOUDEscapeIdentifier(osFIDColName).c_str();
-    osSQL += " = ";
+    osSQL += " LIKE '%";
     osSQL += CPLSPrintf(CPL_FRMT_GIB, nFeatureId);
+    osSQL += "%'";
 
     json_object* poObj = poDS->RunSQL(osSQL);
     json_object* poRowObj = OGRAMIGOCLOUDGetSingleRow(poObj);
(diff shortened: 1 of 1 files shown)


gdal/ogr/ogrsf_frmts/amigocloud/ogramigocloudtablelayer.cpp
```diff
@@ -777,8 +777,9 @@ OGRFeature* OGRAmigoCloudTableLayer::GetFeature( GIntBig nFeatureId )
     CPLString osSQL = osSELECTWithoutWHERE;
     osSQL += " WHERE ";
     osSQL += OGRAMIGOCLOUDEscapeIdentifier(osFIDColName).c_str();
-    osSQL += " = ";
+    osSQL += " LIKE '%";
```
rouault, line 780: why not just a plain comparison to a literal ?  `` = '12'  ``

vchernetsky: We use UUID char(32) as fid. That may work too, but I thought, substring has more chances to match. 
BTW, how would I handle in GDAL the case when the fid is a UUID string?

rouault: > That may work too, but I thought, substring has more chances to match.

Please check if simple equality doesn't just work ? That's less black magick.

> BTW, how would I handle in GDAL the case when the fid is a UUID string?

OGR FID is a Int64. If you're database FID is a UUID string then you'll have to store it as a regular column in the OGRFeatureDefn. The mongodbv3 / elastic drivers have similar issues with the _id special field

vchernetsky: I've checked the simple literal, it does not work in some cases. I think it needs to be a substring, because for example,
QGIS sends an integer value that some how got derived from the string. For example this amigo_id="890c7f3eca9a41ee9793c9c24f6aba64" ended up being nFeatureId=3832683572665211409. 
Perhaps it gets first 64 bit of that string and cast it as int64?

rouault: > Perhaps it gets first 64 bit of that string and cast it as int64?

you have to check how you derive FID from server content in GetNextFeature()

vchernetsky: Ok, I think I understand now, I'll make another change to this pul request. Thanks!

rouault:
@vchernetsky  Is the AmigoCloud driver still used ? I tried to contact @rburhum by email and your info email too, and got zero feedback. Sigh

rburhum:
Hi @rouault . We definitely still use it, but are in a three year old version. Let me check with the team and get back to you. Thanks again for reaching out.

PR #3573, opened 2021-03-13
CPLHTTP: Allow to set GSSAPI credential delegation type

drons:
Delegation type can be set using CPLHTTPSetOptions option 'GSSAPI_DELEGATION'
or CPLGetConfigOption option 'GDAL_GSSAPI_DELEGATION'

See: https://curl.se/libcurl/c/CURLOPT_GSSAPI_DELEGATION.html

### DIFF
--- a/gdal/port/cpl_http.cpp
+++ b/gdal/port/cpl_http.cpp
@@ -52,6 +52,13 @@
 
 #ifdef HAVE_CURL
 #  include <curl/curl.h>
+
+#ifndef CURL_AT_LEAST_VERSION
+#define CURL_VERSION_BITS(x,y,z) ((x)<<16|(y)<<8|z)
+#define CURL_AT_LEAST_VERSION(x,y,z) \
+  (LIBCURL_VERSION_NUM >= CURL_VERSION_BITS(x, y, z))
+#endif //#ifndef CURL_AT_LEAST_VERSION
+
 // CURLINFO_RESPONSE_CODE was known as CURLINFO_HTTP_CODE in libcurl 7.10.7 and
 // earlier.
 #if LIBCURL_VERSION_NUM < 0x070a07
--- a/gdal/port/cpl_http.cpp
+++ b/gdal/port/cpl_http.cpp
@@ -51,19 +51,8 @@
 #endif
 
 #ifdef HAVE_CURL
-#  include <curl/curl.h>
 
-#ifndef CURL_AT_LEAST_VERSION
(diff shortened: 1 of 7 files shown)


rouault:
Could you document that similarly to other options ?

gdal/port/cpl_http.cpp
```diff
@@ -1965,6 +1965,36 @@ void *CPLHTTPSetOptions(void *pcurl, const char* pszURL,
     }
 #endif
 
+#if CURL_AT_LEAST_VERSION(7,22,0)
```
rouault, line 1968: according to https://github.com/curl/curl/commit/39b9bf60d13ff7cb62a6a031c210d27a32113c4f , the CURL_AT_LEAST_VERSION macro has been introduced only in curl 7.43

Perhaps we should put around the top of this file, after curl include
```
#ifndef CURL_AT_LEAST_VERSION
#define CURL_VERSION_BITS(x,y,z) ((x)<<16|(y)<<8|z)
#define CURL_AT_LEAST_VERSION(x,y,z) \
  (LIBCURL_VERSION_NUM >= CURL_VERSION_BITS(x, y, z))
#endif
```

drons: May be replace other `LIBCURL_VERSION_NUM` with `CURL_AT_LEAST_VERSION`?

rouault: > May be replace other `LIBCURL_VERSION_NUM` with `CURL_AT_LEAST_VERSION`?

good idea. We could assume that the minimum version is 7.19 (super ancient version in RHEL 6 base from what I can see) to get rid of some now useless tests

drons: `LIBCURL_VERSION_NUM` macros can be found in the following additional files:
```
gdal/frmts/wms/gdalhttp.cpp
gdal/port/cpl_vsil_curl.cpp
gdal/port/cpl_vsil_curl_class.h
```

To replace`LIBCURL_VERSION_NUM` in these files, it makes sense to move `CURL_AT_LEAST_VERSION` to a common place. But where is it?

rouault: > To replace`LIBCURL_VERSION_NUM` in these files, it makes sense to move `CURL_AT_LEAST_VERSION` to a common place. But where is it?

maybe create a port/cpl_curl_priv.h that includes <curl/curl.h> and defines CURL_AT_LEAST_VERSION if needed ?

drons:
Thanks!

PR #3579, opened 2021-03-15
gdal/swig/python/setup.py - use a better way handle readme encoding

idanmiara:
replace `readme = open('README.rst', 'rb').read().decode('utf-8')` with `readme = open('README.rst', encoding="utf-8").read()`

### DIFF
--- a/.github/workflows/ubuntu_20.04.yml
+++ b/.github/workflows/ubuntu_20.04.yml
@@ -60,6 +60,9 @@ jobs:
       - name: Run SWIG CSharp tests
         run: docker run --rm gdal sh -c "cd /build/gdal/swig/csharp && make && make && make test"
 
+      - name: Build Python sdist package
+        run: docker run --rm gdal sh -c "cd /build/gdal/swig/python && python setup.py sdist"
+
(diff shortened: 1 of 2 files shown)


rouault:
Would you mind adding a ```python setup.py sdist``` step in one of the CI configurations, for example the ubuntu 20.04 one ?

avalentino (approved):

idanmiara:
> python setup.py sdist

Trying...

.github/workflows/ubuntu_20.04.yml
```diff
@@ -60,6 +60,9 @@ jobs:
       - name: Run SWIG CSharp tests
         run: docker run --rm gdal sh -c "cd /build/gdal/swig/csharp && make && make && make test"
 
+      - name: Build Python sdist package
+        run: docker run --rm gdal sh -c "cd /build/gdal/swig/python && python setup.py sdist"
```
rouault, line 64:
```suggestion
        run: docker run --rm gdal sh -c "cd /build/gdal/swig/python && python3 setup.py sdist"
```

PR #3584, opened 2021-03-15
FlatGeoBuf: fix crash in GetFileList() on a dataset opened in update mode

rouault:
Issue hinted by clang static analyzer.
Note to @bjornharrtell : you can't use std::is_base_of<> as a way of determining at run-time if an object can be dynamically cast to some type. std::is_base_of<> is a mechanism operating only on types (at compilation time), and not instances.


bjornharrtell:
Thanks @rouault that makes sense (now).

PR #3587, opened 2021-03-16
GTiff: fix speed of guessing JPEG quality

rouault:
Fixes a GDAL 3.2 performance regression.

Since GDAL 3.2, we expose the JPEG quality in metadata and read it at
dataset opening, and for each overview. Previously we only read it when
updating an existing dataset.
For a quality of 85, guessing it for each overview level takes
approximatively 14 ms on a Core i7-10750H @2.6 GHz, which sums up to
~ 100 ms when opening a dataset with 6 overview levels, such as in
the test case of https://github.com/OSGeo/gdal/issues/3470

We fix that by precomputing the MD5Sum of the quantization tables.

Before:
```
$ time gdalwarp -q wrk/m_3209542_se_15_060_20181115_20190222.tif wrk/out.tif -t_srs EPSG:3857 -ts 1024 1024 -overwrite

real	0m0,282s
user	0m0,249s
sys	0m0,033s
```

After:
```
$ time gdalwarp -q wrk/m_3209542_se_15_060_20181115_20190222.tif wrk/out.tif -t_srs EPSG:3857 -ts 1024 1024 -overwrite

real	0m0,179s
user	0m0,143s
sys	0m0,037s
```


rouault:
Actually the related peformance regression started at the beginning of the 3.3 cycle, and did not affect 3.2

Issue #1191, opened 2019-01-09
Add an OSRSetPROJSearchPaths function


sgillies:
Such as

```
void CPL_DLL OSRSetPROJSearchPaths( const char* const* papszPaths );
```

As discussed in https://lists.osgeo.org/pipermail/gdal-dev/2019-January/049570.html, this will prevent the need for callers to race GDAL to PROJ's default context, and will allow the search paths to be modified at any time that suits the caller.

gbburkhardt:
I'd very much like to see this exposed through the Java and other interfaces.  In particular, Java doesn't let one set an environment variable programmatically without a hack.  Thanks.

rouault:
> I'd very much like to see this exposed through the Java and other interfaces. In particular, Java doesn't let one set an environment variable programmatically without a hack. Thanks.

Was added a few days ago in master per https://github.com/OSGeo/gdal/pull/1658 and backported to 3.0 branch

gbburkhardt:
You're ahead of the game, as usual.  Thanks.

PR #3590, opened 2021-03-17
OSR: Get/Set access to PROJ auxiliary database filenames

drons:
New PROJ versions can use direct Helmert transformations between datums, but not all of them ships with internal proj.db

This PR add possibility to attach auxiliary database to PROJ directly from GDAL related code.

Maybe this is not a best idea, and attaching an auxiliary database via PROJ_AUX_DB_PATH environment variables would be better. But PROJ does not have such environment variable.

See #1191.


rouault:
> and attaching an auxiliary database via PROJ_AUX_DB_PATH environment variables would be better.

PROJ 8.1 will ! See the last commit in https://github.com/OSGeo/PROJ/pull/2577

rouault:
> PROJ 8.1 will !

that said, in some use cases setting environment variables isn't really practical, so I can see some value to your pull request. Could you possibly extend autotest/cpp/test_osr_set_proj_search_paths.cpp or create a similar .cpp case with the new functions ? You may need to create a small test aux db with a few records in it.

drons:
> PROJ 8.1 will ! See the last commit in [OSGeo/PROJ#2577](https://github.com/OSGeo/PROJ/pull/2577)

It's amazing, I looked at PROJ source a few days ago.

> Could you possibly extend autotest/cpp/test_osr_set_proj_search_paths.cpp

OK. I will create tests.
And I think we also need python bindings.

drons:
Work still in progress

drons:
@rouault 
Tests have been reworked, and it feels like I have formed the test_aux.db incorrectly.
But I can't figure out where the problem is.

test_aux.db:
```sql
PRAGMA foreign_keys=OFF;
BEGIN TRANSACTION;
CREATE TABLE geodetic_crs(
    auth_name TEXT NOT NULL CHECK (length(auth_name) >= 1),
    code TEXT NOT NULL CHECK (length(code) >= 1),
    name TEXT NOT NULL CHECK (length(name) >= 2),
    description TEXT,
    scope TEXT,
    type TEXT NOT NULL CHECK (type IN ('geographic 2D', 'geographic 3D', 'geocentric')),
    coordinate_system_auth_name TEXT,
    coordinate_system_code TEXT,
    datum_auth_name TEXT,
    datum_code TEXT,
    area_of_use_auth_name TEXT,
    area_of_use_code TEXT,
    text_definition TEXT, -- PROJ string or WKT string. Use of this is discouraged as prone to definition ambiguities
    deprecated BOOLEAN NOT NULL CHECK (deprecated IN (0, 1)),
    CONSTRAINT pk_geodetic_crs PRIMARY KEY (auth_name, code),
    CONSTRAINT fk_geodetic_crs_coordinate_system FOREIGN KEY (coordinate_system_auth_name, coordinate_system_code) REFERENCES coordinate_system(auth_name, code),
    CONSTRAINT fk_geodetic_crs_datum FOREIGN KEY (datum_auth_name, datum_code) REFERENCES geodetic_datum(auth_name, code),
    CONSTRAINT fk_geodetic_crs_area FOREIGN KEY (area_of_use_auth_name, area_of_use_code) REFERENCES area(auth_name, code)
);
INSERT INTO "geodetic_crs" VALUES('EPSG','111111','Test WGS 84',NULL,NULL,'geographic 2D','EPSG','6422','EPSG','6326','EPSG','1262',NULL,0);
COMMIT;
```

On test PROJ emit error:
```
./test_osr_set_proj_search_paths
ERROR 1: PROJ: proj_create_from_database: crs not found
failure not expected (8)
```

autotest/cpp/test_osr_set_proj_search_paths.cpp
```diff
@@ -85,6 +85,36 @@ static void func3(void*)
     OSRDestroySpatialReference(hSRS);
 }
 
+static void func4()
```
rouault, line 88: upon reflection, this is going to be very fragile as the database structure might change between PROJ versions. Sorry for not having thought about that before proposing the test. Or maybe build dynamically the aux.db by running a "INSERT INTO geodetic_crs SELECT *, '111111' AS code FROM geodetic_crs WHERE auth_name = 'EPSG' and code = '4326'" (or something like that). The OGR SQLite driver could be used for example. But you'd have to find where proj.db is... Hum. I realize this might become quite complicated and impractical ! Perhaps just test calling the API to check it doesn't crash is the most reasonable option.

autotest/osr/osr_basic.py
```diff
@@ -1663,3 +1663,13 @@ def test_osr_GetUTMZone_Projected3D():
     utm_srs.PromoteTo3D()
 
     assert utm_srs.GetUTMZone() == -11
+
+
+###############################################################################
+def test_SetPROJAuxDbPaths():
+
+    osr.SetPROJAuxDbPath('../cpp/data/test_aux.db')
```
rouault, line 1671: I would avoid doing that in the main process as it will indirectly influence all later tests. For a similar situation, we have gcore/basic_test_subprocess.py that is called by gcore/basic_test.py . And perhaps just calling SetPROJAuxDbPath with a dummy path and see that something afterwards doesn't crash would be enough

autotest/cpp/test_osr_set_proj_search_paths.cpp
drons: > upon reflection, this is going to be very fragile as the database structure might change between PROJ versions.

It seems to me that this is not due to changes in the structure of the base. I'm create test_aux.db form original proj.db by deleting everything unnecessary. 

I'm for now add to aux.db one view:
```sql
CREATE VIEW crs_view AS SELECT 'geodetic_crs' AS table_name, auth_name, code, name, type, description, scope, area_of_use_auth_name, area_of_use_code, deprecated FROM geodetic_crs
```
And test on my local build is OK. (Now I'm run all tests at separate branch)

rouault: > I'm for now add to aux.db one view:

The problem is that the structure of this table has changed the last year (PROJ 7.2 maybe). There is no more scope, area_of_use_auth_name, area_of_use_code in later versions.

drons:
For now all tests are passed, including test with real `cpp/data/test_aux.db` in despite of `proj.db` structure change between versions. 
It seems to me that a real test is better than a formal one.

rouault:
> For now all tests are passed, including test with real `cpp/data/test_aux.db` in despite of `proj.db` structure change between versions.

ah yes, we are quite lucky. As you use the older schema that has more columns than the latest ones, and the unioning of databases is done by selecting precise columns from the DBs, that works with recent PROJ versions too... A bit fragile though if the structure changes too much in the future. We'll see then. Could you perhaps add a comment in the .cpp and .py test to explain that those tests could easily break if the DB structure of PROJ changes in the future, and point to that ticket ?

drons:
Thanks!

Issue #3581, opened 2021-03-15
LVBAG Driver: autocorrecting invalid Pand/Building geometry renders non-Polygon types


justb4:
## Expected behavior and actual behavior.
Issue #3467 with PR #3531 assigned (correctly) a Polygon type as the Pand (Building) geometry-type as this is the official 
type in the BAG XSD.

When using `ogr2ogr` to write to e.g. PostGIS we now see the following errors:

    ogr2ogr -f PostgreSQL "PG:dbname=bagv2 host=bagv2_db port=5432 user=*** password=*** active_schema=bagactueel" -lco LAUNDER=YES -lco PREC\
    ISION=NO -lco FID=gid -lco SPATIAL_INDEX=NONE -append --config PG_USE_COPY YES -a_srs EPSG:28992 -gt 200000  -oo AUTOCORRECT_INVALID_DATA=YES -oo LEGACY_ID=YES temp/temp_dir
    Warning 1: Self-intersection at or near point 210444.6939819445 462183.56186922867
    Warning 1: Geometry to be inserted is of type Multi Polygon, whereas the layer geometry type is Polygon.
    Insertion is likely to fail
    ERROR 1: COPY statement failed.
    ERROR:  Geometry type (MultiPolygon) does not match column type (Polygon)
    CONTEXT:  COPY pand, line 28484, column wkb_geometry: "010600002040710000020000000103000000010000000A000000DD67468D65B00941A7A55A3F9E351C416ABC749391B00941..."
    ERROR 1: COPY statement failed.
    Warning 1: Self-intersection at or near point 192905.48699999999 426593.91399999999
    Warning 1: Geometry to be inserted is of type Geometry Collection, whereas the layer geometry type is Polygon.
    Insertion is likely to fail
    ERROR 1: COPY statement failed.
    ERROR:  Geometry type (GeometryCollection) does not match column type (Polygon)
    CONTEXT:  COPY pand, line 189263, column wkb_geometry: "0107000020407100000200000001030000000100000016000000894160E54B8C0741B29DEFA787091A4121B07268668C0741..."
  etc

This happens on the following conditions/settings:

* option `-oo AUTOCORRECT_INVALID_DATA=YES` (with `NO` only the self-intersection warnings)
* with the country-wide Dutch NL LV BAG source dataset (18 million building XML objects). Single city XML like Amsterdam and Rotterdam is ok, but by accident as no invalid geometry is present

## Steps to reproduce the problem.

[...]

PR #3599, opened 2021-03-20
LVBAG: Convert MultiPolygon to Polygon

yorickdewid:

Addresses issue #3581

## Tasklist

 - [x] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/autotest/ogr/ogr_lvbag.py
+++ b/autotest/ogr/ogr_lvbag.py
@@ -316,8 +316,6 @@ def test_ogr_lvbag_read_zip_4():
 
 def test_ogr_lvbag_fix_invalid_polygon():
 
-    pytest.skip()
-
     if not ogrtest.have_geos() and not ogrtest.have_sfcgal():
         pytest.skip()
 
@@ -343,8 +341,6 @@ def test_ogr_lvbag_fix_invalid_polygon():
     assert feat is None
 
 def test_ogr_lvbag_fix_invalid_polygon_to_polygon():
-
-    pytest.skip()
 
     if not ogrtest.have_geos() and not ogrtest.have_sfcgal():
         pytest.skip()
--- a/autotest/ogr/ogr_lvbag.py
+++ b/autotest/ogr/ogr_lvbag.py
@@ -33,7 +33,6 @@
 
 from osgeo import ogr, gdal
 import gdaltest
-import ogrtest
 import pytest
 
 pytestmark = pytest.mark.require_driver('LVBAG')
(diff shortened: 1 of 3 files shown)


rouault:
The failures in gcore/transformer.py are due to a PROJ master change. Likely a bug. I'm investigating.

autotest/ogr/ogr_lvbag.py
```diff
@@ -342,6 +342,34 @@ def test_ogr_lvbag_invalid_polygon():
     feat = lyr.GetNextFeature()
     assert feat is None
 
+def test_ogr_lvbag_fix_invalid_polygon_to_polygon():
+
+    pytest.skip()
```
rouault, line 347: why is this disabled ? Same for test_ogr_lvbag_fix_invalid_polygon

rouault:
> The failures in gcore/transformer.py are due to a PROJ master change. Likely a bug. I'm investigating.

fixed

yorickdewid:
I suppose they need to be re-enabled when the PROJ bug is resolved.

justb4:
Thanks for picking up! 
Possible refinement: in some cases GEOS `makeValid()` creates a `GeometryCollection` i.s.o. a `MultiPolygon`. 
So around [ogrlvbaglayer.cpp:705](https://github.com/OSGeo/gdal/blob/9f4a8f51a4d8f64d0ae55303ddfc7d1c6cced5e9/gdal/ogr/ogrsf_frmts/lvbag/ogrlvbaglayer.cpp#L705): add the condition `&& poGeom->getGeometryType() == wkbGeometryCollection` and cast to `toGeometryCollection() ` and `getGeometryType() == wkbPolygon` to get the largest `Polygon`. 
Sorry don't have a specific object id for that case.

justb4:
Note that most if not all invalid Geometries are History records, basically deleted records, but nothing is ever deleted from this dataset. Almost all applications ignore history records unless for historic analysis. 

Another approach is to not try to repair but add a boolean attribute `geom_valid=false`. For 18 million (about half are History) Building-records saves quite some processing and it becomes easier to report real errors back to Kadaster via "terugmeldvoorziening" in [BAGViewer](https://bagviewer.kadaster.nl).

yorickdewid:
@justb4 I've not run this code with a geometry collection, however a base `OGRGeometry` should survive a downcast to its own `OGRGeometryCollection` type.

autotest/ogr/ogr_lvbag.py
rouault: the pytest.skip() is still there

rouault:
ah, you probably need to test that MakeValid works and pytest.skip() if not.
Like do
```python
    g = ogr.CreateGeometryFromWkt('POLYGON ((0 0,1 1,0 1,1 0,0 0))')
    if g.MakeValid() is None:
        pytest.skip("MakeValid() not available")
```

rouault:
autotest/ogr/ogr_lvbag.py:36:1: F401 'ogrtest' imported but unused

PR #3608, opened 2021-03-24
util.py - get_ovr_idx to support selecting by resolution + some utility functions + testing

idanmiara:
improve get_ovr_idx:
def get_ovr_idx(filename_or_ds: PathOrDS, ovr_idx: Optional[Union[int, float]]):
    """
    returns a non-negative ovr_idx, from a given ovr_idx number
    for ovr_idx = None => returns 0
    for ovr_idx: int >= 0 => returns the given ovr_idx
    for ovr_idx: int < 0 => -1 is the last overview; -2 is the one before the last and so on
    for ovr_idx: float => returns the best overview by the given resolution (ovr_idx represents the resolution in this case)
    """

This PR also adds some other basic utility functions and testing

### DIFF
--- a/gdal/swig/python/gdal-utils/osgeo_utils/auxiliary/raster_creation.py
+++ b/gdal/swig/python/gdal-utils/osgeo_utils/auxiliary/raster_creation.py
@@ -1,0 +1,114 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+# ******************************************************************************
+#
+#  Project:  GDAL utils.auxiliary
+#  Purpose:  raster creation utility functions
+#  Author:   Idan Miara <idan@miara.com>
+#
+# ******************************************************************************
+#  Copyright (c) 2021, Idan Miara <idan@miara.com>
+#
+#  Permission is hereby granted, free of charge, to any person obtaining a
+#  copy of this software and associated documentation files (the "Software"),
+#  to deal in the Software without restriction, including without limitation
+#  the rights to use, copy, modify, merge, publish, distribute, sublicense,
+#  and/or sell copies of the Software, and to permit persons to whom the
+#  Software is furnished to do so, subject to the following conditions:
+#
+#  The above copyright notice and this permission notice shall be included
+#  in all copies or substantial portions of the Software.
+#
+#  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+#  OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+#  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+#  THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+#  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+#  FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+#  DEALINGS IN THE SOFTWARE.
+# ******************************************************************************
+
+import tempfile
+from numbers import Real
+from typing import Sequence, Optional
+
+from osgeo import gdal, osr
+from osgeo_utils.auxiliary.base import PathLike, MaybeSequence, is_true
+from osgeo_utils.auxiliary.util import get_bigtiff_creation_option_value, get_data_type, DataTypeOrStr, CreationOptions
+
+
+def create_flat_raster(filename: Optional[PathLike],
+                       driver: Optional[str] = None, dt: DataTypeOrStr = gdal.GDT_Byte,
+                       size: MaybeSequence[int] = 128, band_count: int = 1, creation_options: CreationOptions = None,
+                       fill_value: Optional[Real] = None, nodata_value: Optional[Real] = None,
+                       origin: Optional[Sequence[int]] = (500_000, 0), pixel_size: MaybeSequence[int] = 10,
+                       epsg: Optional[int] = 32636,
+                       overview_alg: Optional[str] = 'NEAR', overview_list: Optional[Sequence[int]] = None):
+    if filename is None:
+        filename = tempfile.mktemp()
+    elif not filename:
+        filename = ''
+    if driver is None:
+        driver = 'GTiff' if filename else 'MEM'
+    if not isinstance(size, Sequence):
+        size = (size, size)
+
+    drv = gdal.GetDriverByName(driver)
+    dt = get_data_type(dt)
+    creation_options_list = get_creation_options(creation_options, driver=driver)
+    ds = drv.Create(str(filename), *size, band_count, dt, creation_options_list)
+
+    if pixel_size and origin:
+        if not isinstance(pixel_size, Sequence):
+            pixel_size = (pixel_size, -pixel_size)
+        ds.SetGeoTransform([origin[0], pixel_size[0], 0, origin[1], 0, pixel_size[1]])
+    if epsg is not None:
+        srs = osr.SpatialReference()
+        srs.ImportFromEPSG(epsg)
+        ds.SetSpatialRef(srs)
+    for bnd_idx in range(band_count):
+        bnd: gdal.Band = ds.GetRasterBand(bnd_idx+1)
+        if fill_value is not None:
+            bnd.Fill(fill_value)
+        if nodata_value is not None:
+            bnd.SetNoDataValue(nodata_value)
+    if overview_alg and overview_list:
+        ds.BuildOverviews(overview_alg, overviewlist=overview_list)
+    return ds
+
+
+def get_creation_options(creation_options: CreationOptions = None,
+                         driver: str = 'GTiff',
+                         sparse_ok: bool = None,
+                         tiled: bool = None,
+                         block_size: Optional[int] = None,
+                         big_tiff: Optional[str] = None,
+                         comp: str = None):
+    creation_options = dict(creation_options or dict())
+    if sparse_ok is None:
+        sparse_ok = creation_options.get("SPARSE_OK", True)
+    creation_options["SPARSE_OK"] = str(bool(sparse_ok))
+
+    driver = driver.lower()
+    if comp is None:
+        comp = creation_options.get("COMPRESS", "DEFLATE")
+    if driver in ['gtiff', 'cog']:
+        creation_options["BIGTIFF"] = get_bigtiff_creation_option_value(big_tiff)
+        creation_options["COMPRESS"] = comp
+
+    if tiled is None:
+        tiled = creation_options.get("TILED", True)
+    tiled = is_true(tiled)
+    creation_options["TILED"] = str(tiled)
+    if tiled and block_size is not None:
+        if driver == 'gtiff':
+            creation_options["BLOCKXSIZE"] = block_size
+            creation_options["BLOCKYSIZE"] = block_size
+        elif driver == 'cog':
+            creation_options["BLOCKSIZE"] = block_size
+
+    creation_options_list = []
+    for k, v in creation_options.items():
+        creation_options_list.append("{}={}".format(k, v))
+
+    return creation_options_list
--- a/gdal/swig/python/gdal-utils/osgeo_utils/auxiliary/util.py
+++ b/gdal/swig/python/gdal-utils/osgeo_utils/auxiliary/util.py
@@ -29,13 +29,16 @@
 #  FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 #  DEALINGS IN THE SOFTWARE.
 # ******************************************************************************
+
 from numbers import Real
-from typing import Optional, Union, Sequence, Tuple
+from typing import Optional, Union, Sequence, Tuple, Dict, Any, Iterator
 
 from osgeo import gdal
-from osgeo_utils.auxiliary.base import get_extension, is_path_like, PathLike
+from osgeo_utils.auxiliary.base import get_extension, is_path_like, PathLike, enum_to_str, OptionalBoolStr, is_true
 
 PathOrDS = Union[PathLike, gdal.Dataset]
+DataTypeOrStr = Union[str, int]
+CreationOptions = Optional[Dict[str, Any]]
 
 
 def DoesDriverHandleExtension(drv: gdal.Driver, ext: str):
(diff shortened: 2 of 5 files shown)


gdal/swig/python/osgeo/utils/auxiliary/util.py
```diff
@@ -96,13 +96,61 @@ def get_ovr_count(filename_or_ds: PathOrDS):
...
+
+
+def get_best_ovr_by_resolutions(requested_res: float, resolutions: Sequence[float]):
+    for ovr, res in enumerate(resolutions):
+        if res > requested_res:
+            return max(0, ovr-1)
+    return len(resolutions)-1
+
+
+def get_ovr_idx(filename_or_ds: PathOrDS, ovr_idx: Optional[Union[int, float]]):
```
rouault, line 133: mixing index and resolution in the same argument looks a bit messy. Can't we have a dedicated ovr_res optional argument, mutually exclusive with ovr_idx ?

rouault: would probably need some test

idanmiara: Thanks for you comments!
I also thought it might be a good idea to have a different parameter for resolution, but then two issues arise:
1. As you said, they are mutually exclusive and we'll need to decide what happens if someone sets both. I try to avoid mutually exclusive parameters if possible.
2. If I use a single parameter then any code that uses this function will automatically support this addition so if ovr_idx was passed any levels up the call stack, using a single variable this eliminates the need to modify call the related calls.
Maybe I would rename it to ovr_idx_or_resolution ? This would break any code that called this function with named argument, but as this function was not released yet I guess it's would be OK.

rouault: >     1. As you said, they are mutually exclusive and we'll need to decide what happens if someone sets both. I try to avoid mutually exclusive parameters if possible.

documentation + exception if that occurs

idanmiara: I believe I found a solution that satisfies all the cases mentioned and also clear.

rouault:
Can you add some tests ?

idanmiara:
> Can you add some tests ?

Yes, assuming gdal-utils will be merged soon, I'll do it after rebasing on it.

rouault:
> assuming gdal-utils will be merged soon,

you can reply to the motion email thread to synthesize the PSC votes and declare the motion passed as there are at least 2 +1 and no -1.

idanmiara:
> > assuming gdal-utils will be merged soon,
> 
> you can reply to the motion email thread to synthesize the PSC votes and declare the motion passed as there are at least 2 +1 and no -1.

Thanks! Done.

gdal/swig/python/gdal-utils/osgeo_utils/auxiliary/util.py
```diff
@@ -96,13 +96,69 @@ def get_ovr_count(filename_or_ds: PathOrDS):
...
+                ovr_idx: Optional[Union[int, float]] = None,
+                ovr_res: Optional[Union[int, float]] = None):
+    """
+    returns a non-negative ovr_idx, from given mutually exclusive ovr_idx (index) or ovr_res (resolution)
+    ovr_idx == None and ovr_res == None => returns 0
+    ovr_idx: int >= 0 => returns the given ovr_idx
+    ovr_idx: int < 0 => -1 is the last overview; -2 is the one before the last and so on
+    ovr_res: float|int => returns the best suitable overview for a given resolution
+             meaning the ovr with the lowest resolution which is higher then the request
+    ovr_idx: float = x => same as (ovr_idx=None, ovr_res=x)
```
rouault, line 143: This possibility is really messy. Why keep it ?
The below code should really distinguish the case where ovr_idx or over_res are specified and not mix things

gdal/swig/python/gdal-utils/osgeo_utils/auxiliary/util.py
```diff
@@ -96,13 +96,69 @@ def get_ovr_count(filename_or_ds: PathOrDS):
         return bnd.GetOverviewCount()
 
 
-def get_ovr_idx(filename_or_ds: PathOrDS, ovr_idx: Optional[int]):
+def get_sizes_factors_resolutions(filename_or_ds: PathOrDS, dim: Optional[int]=0):
+    ds = open_ds(filename_or_ds)
+    bnd = ds.GetRasterBand(1)
+    ovr_count = bnd.GetOverviewCount()
+    geo_transform = ds.GetGeoTransform()
```
rouault, line 103: Use ds.GetGeoTransform(can_return_null=True) and deal with the case when geo_transform is None

gdal/swig/python/gdal-utils/osgeo_utils/auxiliary/util.py
idanmiara: hmm.. because it simplifies my code - I have some calls to this function deep inside some inner functions and I pass ovr_idx from the outer functions, now I wanted to add support for choosing overview by resolution. The easiest solution is to pass a float to the outer function, that's all the change I needed to make. Otherwise if I wanted to use two parameters I would need to modify all the functions on the way. Also having mutually exclusive parameters is more error prone in my opinion.
I agree that it might be confusing so I added both options (single or double parameters). 
If you think it is less messy - I don't mind changing it to be an undocumented feature and to change the type hinting to suggest ovr_idx: Optional[int] accordingly.
If you think about it, the implementation also becomes much simpler if you use this approach of a single parameter - less cases to test against.

gdal/swig/python/gdal-utils/osgeo_utils/auxiliary/util.py
idanmiara: done

gdal/swig/python/gdal-utils/osgeo_utils/auxiliary/util.py
rouault: > If you think it is less messy - I don't mind changing it to be an undocumented feature

Yes, it would best not to advertize ovr_idx: float in the documentation of the function. In the implementation, you could still mention that subtelty.

> If you think about it, the implementation also becomes much simpler if you use this approach of a single parameter - less cases to test against.

Hum the testing should be little impacted. Testing the case where a parameter is integer or float, or two different parameters is pretty much equivalent

idanmiara: done

idanmiara:
> Can you add some tests ?

done

idanmiara:
@rouault please advise if any other changes are required.

gdal/swig/python/gdal-utils/osgeo_utils/auxiliary/util.py
```diff
@@ -154,3 +222,95 @@ def _open_ds(
...
+            if metadata_item.startswith(key):
+                return metadata_item[len(key):]
+        return default
+
+
+def get_big_tiff(big_tiff):
+    return "IF_SAFER" if big_tiff is None else big_tiff if isinstance(big_tiff, str) else NO_YES[bool(big_tiff)]
+
+
+def get_tiled(tiled):
```
rouault, line 308: you likely don't need that function and NO_YES.  str(some_bool) leads to 'True' and 'False' that are also valid values for the TILED creation option.

gdal/swig/python/gdal-utils/osgeo_utils/auxiliary/util.py
```diff
@@ -154,3 +222,95 @@ def _open_ds(
...
+        metadata = ds.GetMetadata_List("IMAGE_STRUCTURE")
+        if metadata is None:
+            return default
+        for metadata_item in metadata:
+            if metadata_item.startswith(key):
+                return metadata_item[len(key):]
+        return default
+
+
+def get_big_tiff(big_tiff):
```
rouault, line 304:
```suggestion
def get_bigtiff_creation_option_value(big_tiff):
```
As far as I know this method can only be called with str or None. Not bool

gdal/swig/python/gdal-utils/osgeo_utils/auxiliary/util.py
```diff
@@ -154,3 +222,95 @@ def _open_ds(
...
+        return band.GetNoDataValue()
+
+
+def unset_nodatavalue(filename_or_ds: PathOrDS):
+    with OpenDS(filename_or_ds, access_mode=gdal.GA_Update) as ds:
+        for b in get_raster_bands(ds):
+            b.DeleteNoDataValue()
+
+
+def get_image_structure_metadata(filename_or_ds: PathOrDS, key: str, default=None):
```
rouault, line 290: could be simplified as ds.GetMetadataItem(key, "IMAGE_STRUCTURE")

gdal/swig/python/gdal-utils/osgeo_utils/auxiliary/raster_creation.py
```diff
@@ -0,0 +1,104 @@
...
+    if origin is None:
+        origin = (500_000, 0)
+    drv = gdal.GetDriverByName(driver)
+    dt = get_data_type(dt)
+    creation_options_list = get_creation_options(creation_options, driver=driver)
+    ds = drv.Create(filename, *size, band_count, dt, creation_options_list)
+    ds.SetGeoTransform([origin[0], pixel_size[0], 0, origin[1], 0, -pixel_size[1] ])
+    if epsg is not None:
+        srs = osr.SpatialReference()
+        srs.ImportFromEPSG(epsg)
```
rouault, line 62: Use ds.SetSpatialRef(srs) to replace the 2 below lines

gdal/swig/python/gdal-utils/osgeo_utils/auxiliary/raster_creation.py
```diff
@@ -0,0 +1,104 @@
...
+                       driver: str = 'GTiff', dt: DataTypeOrStr = gdal.GDT_Byte,
+                       size: MaybeSequence[int] = 128, band_count: int = 1, creation_options: CreationOptions = None,
+                       fill_value: Optional[Real] = None, nodata_value: Optional[Real] = None,
+                       origin: Optional[Sequence[int]] = None, pixel_size: MaybeSequence[int] = 10,
+                       epsg: Optional[int] = 32636,
+                       overview_alg: Optional[str] = 'NEAR', overview_list: Optional[Sequence[int]] = None):
+    if filename is None:
+        filename = tempfile.mktemp()
+    if not isinstance(pixel_size, Sequence):
+        pixel_size = (pixel_size, -pixel_size)
```
rouault, line 50: pixel_size[1] is negated afterwards, so it is not needed here

gdal/swig/python/gdal-utils/osgeo_utils/auxiliary/util.py
idanmiara: thanks

gdal/swig/python/gdal-utils/osgeo_utils/auxiliary/util.py
idanmiara: I'm not I understand your comment, but I've modified it for your other remarks, so it might be OK now.

gdal/swig/python/gdal-utils/osgeo_utils/auxiliary/util.py
idanmiara: thanks

gdal/swig/python/gdal-utils/osgeo_utils/auxiliary/raster_creation.py
idanmiara: thanks

gdal/swig/python/gdal-utils/osgeo_utils/auxiliary/raster_creation.py
idanmiara: thanks, I removed the negative sign on SetGeoTransform

idanmiara:
@rouault I hope it's ready now :)

gdal/swig/python/gdal-utils/osgeo_utils/auxiliary/util.py
idanmiara: Are `'True'`/`'False'` are valid inputs for big_tiff and all the other inputs which accept `'Yes'`/`'No'` ?

rouault: Any creation option that declares in the XML  type="boolean" accepts yes/true/on and no/false/off (case insensitive) : https://github.com/OSGeo/gdal/blob/362541e961cf8cab046dfbae1bde8ce559b4cb40/gdal/gcore/gdaldriver.cpp#L1973

gdal/swig/python/gdal-utils/osgeo_utils/auxiliary/util.py
```diff
@@ -154,3 +221,96 @@ def _open_ds(
...
+
+def unset_nodatavalue(filename_or_ds: PathOrDS):
+    with OpenDS(filename_or_ds, access_mode=gdal.GA_Update) as ds:
+        for b in get_raster_bands(ds):
+            b.DeleteNoDataValue()
+
+
+def get_image_structure_metadata(filename_or_ds: PathOrDS, key: str, default: Any = None):
+    key = key.strip()
+    if not key.endswith("="):
```
rouault, line 291: adding '=' is wrong. This method is obviously not tested :-)

idanmiara: true. true.

idanmiara: I've fixed that and added a test for that. 
I didn't squash to make it easier to review the fixes.
If it's ready I could squash some commits to make the history cleaner.

idanmiara: I've added tests for `osgeo_utils.auxiliary.base`, then rebased and squashed all the related commits.
I'm planning to add more tests for the other units in another PR.

gdal/swig/python/gdal-utils/osgeo_utils/auxiliary/util.py
idanmiara: > Any creation option that declares in the XML type="boolean" accepts yes/true/on and no/false/off (case insensitive) :
> 
> https://github.com/OSGeo/gdal/blob/362541e961cf8cab046dfbae1bde8ce559b4cb40/gdal/gcore/gdaldriver.cpp#L1973

Thanks for the reference. I've amended the implementation of `is_true()` to be compatible with that. 
It will raise an exception if an unexpected input is provided.

idanmiara:
@rouault Thanks for your review!

PR #3611, opened 2021-03-25
JAVA: Expose two new methods for GDAL's JNI Bridge

mglazer:
## What does this PR do?

Exposes two new JNI methods. Specifically:

* VSIFileFromMemBuffer - Augments to support taking a ByteBuffer as an
  argument
* VSIGetMemFileFromBuffer - Exposes this method and returns a ByteBuffer
  pointer to the underlying memory.

n/a, just exposing what could have been exposed from the start. This is my first gdal contribution as well as some of the first swig code I've written so apologies if I did anything incorrectly.

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:

### DIFF
--- a/gdal/swig/include/cpl.i
+++ b/gdal/swig/include/cpl.i
@@ -186,6 +186,7 @@ void CPL_STDCALL PyCPLErrorHandler(CPLErr eErrClass, int err_no, const char* psz
 %rename (CPLBinaryToHex) CPLBinaryToHex;
 %rename (CPLHexToBinary) CPLHexToBinary;
 %rename (FileFromMemBuffer) wrapper_VSIFileFromMemBuffer;
+%rename (GetMemFileBuffer) wrapper_VSIGetMemFileBuffer;
 %rename (Unlink) VSIUnlink;
 %rename (HasThreadSupport) wrapper_HasThreadSupport;
 %rename (NetworkStatsReset) VSINetworkStatsReset;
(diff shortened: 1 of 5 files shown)


gdal/swig/include/cpl.i
```diff
@@ -520,6 +521,27 @@ void wrapper_VSIFileFromMemBuffer( const char* utf8_path, int nBytes, const GByt
 #endif
 #endif
 
+#if defined(SWIGJAVA)
+%inline %{
+void wrapper_VSIFileFromMemBuffer( const char* utf8_path, void* nioBuffer, long nioBufferSize) {
```
rouault, line 526: This method is already available in the Java case a bit above. Are we sure that the 2 can co-exist ? Adding a test program in swig/java/apps/ checking  that both work would be needed

gdal/swig/include/cpl.i
```diff
@@ -520,6 +521,27 @@ void wrapper_VSIFileFromMemBuffer( const char* utf8_path, int nBytes, const GByt
...
+    VSIFCloseL(VSIFileFromMemBuffer(utf8_path, (GByte*) nioBuffer, nioBufferSize, FALSE));
+}
+%}
+#endif
+
+#if defined(SWIGJAVA)
+%apply (jobject outBuffer) {jobject};
+%inline %{
+jobject wrapper_VSIGetMemFileBuffer(JNIEnv *jenv, const char* utf8_path) {
```
rouault, line 535: hum I don't think that playing directly with JNIEnv* here is the most idiomatic way of proceeding. Ideally we should minimize the amount of language specificities here. A cleaner way would probably to have something like
void wrapper_VSIGetMemFileBuffer(const char* utf8_path, void** outBuffer, size_t* outBufferSize) and a typemap for (void** outBuffer, size_t* outBufferSize).

gdal/swig/include/cpl.i
```diff
@@ -520,6 +521,27 @@ void wrapper_VSIFileFromMemBuffer( const char* utf8_path, int nBytes, const GByt
...
+    vsi_l_offset outBufferLen = 0;
+    GByte* result = VSIGetMemFileBuffer(utf8_path, &outBufferLen, FALSE);
+    return jenv->NewDirectByteBuffer((void*) result, outBufferLen);
```
rouault, line 538: NewDirectByteBuffer() can only work if it is guaranteed that result won't change while the ByteBuffer is in use. There is no such guarantee the /vsimem/ file could be destroyed in-between, or its size grown, causing a realloc(), which would invalidate the original buffer. A copy should be done for safety.
The test program to add mentioned above should also test that.

gdal/swig/include/java/typemaps_java.i
```diff
@@ -62,6 +62,10 @@
 
 %typemap(javain) SWIGTYPE *DISOWN "$javaclassname.getCPtrAndDisown($javainput)"
 
+%typemap(in, numinputs=0) JNIEnv *jenv %{
```
rouault, line 65: that typemap and the one later should probably be removed for other ones, as mentioned above

rouault (changes requested):

mglazer:
Still working through some changes, so not quite there yet. Seems that returning a direct byte buffer really isn't the best thing for a number of reasons (the biggest being that I spend hours trying to get the typemaps to work but couldn't). So ended up just copying the array out, which hopefully is sufficient.

PR #3635, opened 2021-04-01
gdal-utils - samples + scripts + init + gitignore

idanmiara:
## What does this PR do?

Improvements to `gdal-utils`:

1. Add a .gitgnore file
2. Move package info from `setup.py` to `__init__.py`
3. Move the scripts under `gdal-utils` (they are still available in the `gdal` sdist)
4. Move samples to be a submodule of `gdal-utils`
5. Add Windows batch file script wrapper creator

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments


gdal/swig/python/gdal-utils/scripts/batch_creator.py
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+# ******************************************************************************
+#
+#  Project:  GDAL
+#  Purpose:  Create Windows batch wrappers to run the python scripts
```
rouault, line 6: this script doesn't seem to be used by anything else in this PR ?

idanmiara: Right. 
Either a Windows user can run it once to create the batch files or I would need to find a way to run it during instalation of the wheel on Windows. 
What do you think?

rouault: The later would probably be ideal. Otherwise the existence of this script should be mentioned in the documentation of the package

idanmiara: OK, Do you happen to know how to get the path of the scripts folder (i.e. `C:\Python38\Scripts`) so I can run it as post-install ?

idanmiara: It seems that [it is not possible to run post install scripts with wheel distributions.](https://www.javaer101.com/en/article/34927268.html) 
So we either need to distribute the utils as sdist (which I prefer not to) and use [post install](https://stackoverflow.com/questions/20288711/post-install-script-with-python-setuptools)
Or to install the batches the same way we install the scripts, but then they will be installed for all OSs (they would be useless on other OSs but not harmful I guess)
I don't think it justifies to make a separate wheel for windows only...
Or to advise the user in the docs to run the batch creator if they wish to do so.

What do you think?

idanmiara: I've documented it properly in the README:
In case you want to run GDAL scripts on Windows,
If don't have `python.exe` associated with py file extension
(For instance, because you associate them with your IDE)
you might want to create batch wrappers for the gdal python scripts,
Otherwise you won't be able to run the scripts easily.
To create the batch files, Run the following command (after installing `gdal-utils`):

`python -m osgeo_utils.auxiliary.batch_creator`

rouault:
The move of swig/python/scripts and swig/python/samples will have an impact for packagers of the bindings, for example Debian: https://salsa.debian.org/debian-gis-team/gdal/-/blob/master/debian/rules#L193
Could you please initiate a 3.2 -> 3.3 section in MIGRATION_GUIDE.TXT with a "Python bindings packaging: XXXX" item ?

idanmiara:
> The move of swig/python/scripts and swig/python/samples will have an impact for packagers of the bindings, for example Debian: https://salsa.debian.org/debian-gis-team/gdal/-/blob/master/debian/rules#L193
> Could you please initiate a 3.2 -> 3.3 section in MIGRATION_GUIDE.TXT with a "Python bindings packaging: XXXX" item ?

Sure.
What is your opinion about the discussion that went on the mailing list about removing the utils from the gdal sdist? Do you think it's a good idea to execute for gdal 3.3/3.5 or not at all? 

If we would then package maintainers would need to remove the references to the scripts and samples so users would install the gdal-utils with pip. 

I think that for maximum compatibility we should advise users to install gdal-utils version smaller than the next minor so for gdal 3.3.x they should install gdal-utils < 3.4.

rouault:
> What is your opinion about the discussion that went on the mailing list about removing the utils from the gdal sdist? Do you think it's a good idea to execute for gdal 3.3/3.5 or not at all?

I don't have a strong opinion about that. Removing them from the sdist will certainly cause some extra support to people confused why the utils are no longer there when they "pip install gdal". You should probe broader opinions on the ML

> If we would then package maintainers would need to remove the references to the scripts and samples so users would install the gdal-utils with pip.

The Debian maintainers will not make their user rely on pip. Either they'll create a gdal-python and gdal-utils-python separate .deb or will continue to bundle them together in gdal-python.

idanmiara:
> Python bindings packaging:

done

gdal/swig/python/gdal-utils/osgeo_utils/__init__.py
```diff
@@ -0,0 +1,10 @@
+__pacakge_name__ = 'gdal-utils'
+__version__ = '3.3.0.6'
```
rouault, line 2: HOWTO-RELEASE should be updated to reflect the new location of this version number

rouault:
This need to be rebased: gdal/swig/python/samples/validate_gpkg.py has been modified in a recent commit

gdal/swig/python/gdal-utils/osgeo_utils/__init__.py
idanmiara: done. I've also updated the twine section for gdal-utils and added `twine check`

idanmiara:
> This need to be rebased: gdal/swig/python/samples/validate_gpkg.py has been modified in a recent commit

done

idanmiara:
@rouault Thanks for your thorough review!

PR #3639, opened 2021-04-03
gdaltindex manual: explain file based list usage

neteler:
- example added for user convenience

### DIFF
--- a/gdal/doc/source/programs/gdaltindex.rst
+++ b/gdal/doc/source/programs/gdaltindex.rst
@@ -89,17 +89,17 @@ tileindex.
 Examples
 --------
 
-Produce a shapefile (``doq_index.shp``) with a record for every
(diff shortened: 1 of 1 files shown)


gdal/doc/source/programs/gdaltindex.rst
```diff
@@ -105,6 +105,12 @@ into the same output projection:
 
     gdaltindex -t_srs EPSG:4326 -src_srs_name src_srs tile_index_mixed_srs.shp *.tif
 
+- Make a tile index from files listed in a text file :
```
rouault, line 108: For formatting consistency, the 2 above paragraphs should also have a leading '- ' on their first line and have two leading extra spaces on the next lines (before the example itself)

neteler: Sure, updated accordingly.

rouault:
Thanks

PR #3641, opened 2021-04-03
TileDB: do not try to identify /vsis3/ files ending with .tif

rouault:
If trying to open a /vsis3/ TIFF file without appropriate /vsis3/
credentials, it takes a few seconds to the TileDB driver to realize
t can't open it. So avoid it trying when the extension if tif.
A bit of a hack...

CC @normanb : this is more a workaround than a proper fix. There's probably something to improve on the tiledb lib side. GDAL's /vsis3/ is much faster to realize that it can't do the connection to S3.

### DIFF
--- a/gdal/frmts/tiledb/tiledbdataset.cpp
+++ b/gdal/frmts/tiledb/tiledbdataset.cpp
@@ -1171,7 +1171,8 @@ int TileDBDataset::Identify( GDALOpenInfo * poOpenInfo )
         }
 
         if( poOpenInfo->bIsDirectory ||
-                STARTS_WITH_CI( poOpenInfo->pszFilename, "/VSIS3/" ) )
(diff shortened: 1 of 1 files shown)


JeannieMay:
I'm new to using Gdal.

I get a timeout doing a gdal.Open() on a tiff file on S3 using vsis3, while trying to use an IAM role. Could this issue be the cause?

I'm using MaxRev.Gdal.Core 3.2.0.250. Netcore 3.1 c#, running in a Linux container.

Note that defining an AWS_Profile etc the Open() works ok, and that prior to opening the file, I upload using s3Client (which uses the IAM role), so it's something specific to the Gdal.Open().  Perhaps an environment variable?

using GetConfigOption() I can see that none of the following are set AWS_PROFILE; AWS_ACCESS_KEY_ID; AWS_SECRET_ACCESS_KEY; AWS_SESSION_TOKEN; AWS_NO_SIGN_REQUEST; AWS_DEFAULT_PROFILE

PR #3644, opened 2021-04-03
Drop the deprecated "::set-env" workflow commands

trundev:
This is just a cosmetic change, but I'm completely newbie here, so please excuse me if I made something wrong. 

The use of deprecated "::set-env"/"::set-path" can be dropped from the Windows build workflow. It also makes the `ACTIONS_ALLOW_UNSECURE_COMMANDS` flag unnecessary.

Notes:
- In the `shell: cmd` steps, must NOT add spaces before `>>`, or they will be included in the environment variable. See `%INCLUDE%>> %GITHUB_ENV%`
- The environment file `GITHUB_PATH`, appears not to work for multiple entries. This is why the `PATH` is still exported via `GITHUB_ENV`


rouault:
Thanks!

PR #3659, opened 2021-04-08
GPKG: remove creation support for data_type='aspatial' legacy tables

rouault:
This option has now long lived and is superseded by the
data_type='attributes' supported since GeoPackage 1.2 / GDAL 2.2.
Only keep read support for it.

Also remove the creation of the dummy ogr_empty_table special table that
was made to be compliant with Requirement 17 that has now been retired.


rouault:
@rcoup Hopefully you don't depend anymore on the (write side of the) gdal_aspatial extension for attribute tables ?

rcoup:
All good!

PR #3666, opened 2021-04-10
added: actinia entry

neteler:
Also actinia happily uses GDAL


rouault:
@neteler You should rather push your branches to your neteler/gdal fork, not to OSGeo/gdal

PR #3678, opened 2021-04-13
create sphinx documentation for python samples + gdal2xyz docs

idanmiara:
## What does this PR do?

python samples + gdal2xyz docs added

https://github.com/OSGeo/gdal/pull/3416

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments


idanmiara:
As this PR depends on https://github.com/OSGeo/gdal/pull/3416, I've decided to merge it into it.

rouault:
> gdalinfo.py, ogr2ogr.py, ogrinfo.py - move from osgeo_utils/samples to osgeo_utils, as they are not samples utilities

I'm not a fan of this. We should really ere on the side of not proposing too many official utilities, especially when they are redundant with C++ ones. Those scripts were made as an illustration of how to use the Python API. They don't bring more than their C++ equivalents, and are actually lagging behind.

rouault (changes requested):
See above comment

idanmiara:
> > gdalinfo.py, ogr2ogr.py, ogrinfo.py - move from osgeo_utils/samples to osgeo_utils, as they are not samples utilities
> 
> I'm not a fan of this. We should really ere on the side of not proposing too many official utilities, especially when they are redundant with C++ ones. Those scripts were made as an illustration of how to use the Python API. They don't bring more than their C++ equivalents, and are actually lagging behind.

I see your point. I've removed that commit.

gdal/doc/source/api/index.rst
```diff
@@ -85,6 +85,7 @@ API
 
        python_api_ref
        python_gotchas
+       python_samples
```
rouault, line 88: Note unrelated to that PR. The doc structure is a bit messy regarding python. We have that section, and if we go a bit further in the page, we also have a dedicated page for python. I'm not sure how, but I feel like this should be unified as the current setup is confusing.

idanmiara: like that https://github.com/OSGeo/gdal/pull/3687 ?

PR #3681, opened 2021-04-15
Doc: sponsoring pages

rouault:


rouault:
So this pull request is now just waiting for the motion related to RFC80 to be approved. The prospectus as .rst has been replaced by a standalone PDF version created from the google doc one.

PR #3682, opened 2021-04-15
Add RFC 80: NumFOCUS relationship and sponsorship program

rouault:


gdal/doc/source/development/rfc/rfc80_numfocus_relationship.rst
```diff
@@ -0,0 +1,140 @@
...
+The proposed answers to that application form are at:
+https://docs.google.com/document/d/1bc5jdpCe1axdyBHxbJnun7e0DTyDoZI_eFYgJYnOhB8/edit
+
+A `Pre-approved Grantor/ Grantee Agreement <https://docs.google.com/document/d/12KGmSD_EGWqlzh0fiaHNIiN6xPWgdTy--NYf4QR-5k8/edit>`__
+will also have to be signed by 3 members of the GDAL PSC.
+
+Being linked to NumFOCUS, GDAL might be able to benefit from small grants, funded
+by NumFOCUS general budget, and allocated by the NumFOCUS board to projects it will
+select (among the general pool of projects hosted by NumFOCUS). This point will
+be clarified as our relationship with NumFOCUS matures.
```
pl-nfwarmerdam, line 76: I think we might either omit this section, or clarify that we are primarily expecting funding through NumFOCUS to be from sponsors solicited by the project rather than from NumFOCUS grants from the general budget.

gdal/doc/source/development/rfc/rfc80_numfocus_relationship.rst
```diff
@@ -0,0 +1,140 @@
...
+Our hope is that these sessions can also lead to new feature development in GDAL through joint funding initiatives between top sponsors, providing more structure to the way GDAL features are funded today. This ideally leads to more cost-sharing as organizations will be able to be more strategic when they know what everyone else is interested in. These efforts will be above and beyond the core maintenance sponsorship, but the Advisory Board will be a key forum to enable it.
+
+The Advisory board will have an email list, with members of the GDAL team on it, and will meet on a periodic schedule, to be determined as the group forms. It will consist of one representative from each of the Platinum sponsors, one representative for every three Gold sponsors, and one representative from Silver sponsors.
```
pl-nfwarmerdam, line 119: I assume this is envisaged as a way of encouraging sponsorship?   I am sort of wondering if this should just be left to a distinct RFC but I do not feel strongly on the issue.  While it is not the end of the world, I feel like some items are indistinct.  For instance what "key GDAL maintainers" and the "GDAL team" mean, and how open this group is, how selection from the various sponsorship levels is handled.   

I would be inclined to:
 - expect anyone receiving funding since the last meeting to attend the advisory board meeting (ie. those benefiting should at least listen to the advisory board)
 -  any PSC members would be welcome, and the PSC chair would make every effort to attend
 -  any sponsor is welcome regardless of level (higher levels will naturally have more influence, and I think it is going to be harder to get sponsors to actually participate so artificially limiting participation may be counter productive).
 - I think the mailing list should be readable by anyone even if membership is limited.

gdal/doc/source/development/rfc/rfc80_numfocus_relationship.rst
```diff
@@ -0,0 +1,140 @@
...
+- reports of those activities
+
+More generally, all tasks that are needed for a project to strive on the long term, but are
+typically hard to get directly funding from, as not being directly tied to a
+particular feature. It is expected that new features and
+improvements will go on being submitted as code contributions by the interested
+parties in capacity to develop them, or through service providers they might hire
+for that purpose.
+
+Sponsors will not be able to direct particular developments, nor receive prioritary
```
pl-nfwarmerdam, line 108: priority

gdal/doc/source/development/rfc/rfc80_numfocus_relationship.rst
rouault: Agreed to remove this. This was more for information. But clearly those small grant programs (with unknown outcome) aren't the main motivation to go through NumFOCUS

gdal/doc/source/development/rfc/rfc80_numfocus_relationship.rst
rouault: This was copied from @cholmes proposal. I'll let him expose his views. Indeed the idea if to encourage sponsorship and give a bit more weight to higher level of sponsorships.

cholmes: Yes, it's a way to encourage sponsorship, and also just to get a bit more engagement from those who sponsor. I'm not set on the details at all, just want something where the top sponsors have at least a couple of touch points throughout the year, instead of hearing nothing and then being hit up for money again.

I just put this down for something to react against.

Anyone receiving funding makes a ton of sense to me, and agree on all PSC welcome to attend. I just didn't want to presume everyone on the PSC would be interested.

The other two points I'm less sure about. I do think there's value in a forum for sponsors that is a bit more private, where they can talk about what they're planning for the future with geospatial, and explore co-funding new features in GDAL, without having to be worried about some random person picking that up on a public list. I'm a huge advocate of open decision making / transparency on the core project, and appreciate the push towards it with the sponsorship effort. But this to me is a bit different, as I see the advisory board as having merely advisory power, no decision making authority about GDAL, so seems ok to have some private forum. 

That said, I could see something where anyone could be an 'observer' and attend, but there's something that top sponsors get that's a bit more. I'm just not sure what that would exactly be, since there won't be like 'votes' on anything, since it is purely advisory. I am less concerned about sponsors not actually participating - I think if it's just me (for Planet) and one other person showing up that's still fine. But yeah, I'm open to it, but I struggle to define how higher levels 'have more influence' in a meaningful way, so if we can come up with that I can be into it.

As for mailing list - my instinct is that it is a benefit to have a private forum where sponsors can talk more freely without being 'forced' to be open in everything, as per my thoughts above. But I'm not set on it, curious to hear what others think.

rouault: I agree this might not belong to this RFC, but we need to agree on which interactions we propose to sponsors.
- Agreed that any PSC member or person having being sponsored should be part of meetings
- Having all sponsors potentially invited to meetings would make it simpler (otherwise we must have the designation process of representatives for non-platinium categories). The benefit of being platinium vs supporter is less obvious then but perhaps sponsors at higher levels will be happy enough with bigger logos and better placement of them ? Or perhaps offer the platinum sponsors to use a X-minute (<= 5 likely) slot each during meeting if they want ?
- Regarding the mailing list, any PSC member interested can of course join in. I'm wondering what kind of traffic we can expect from it. Will it be about technical stuff (but we don't want to turn it as some sort of premium-level support channel) or mostly to settle the date of the next meeting ?

rouault: For comparison: Looking a bit around, QGIS sustaining members seem to receive no particular advantages than having their logos displayed (or an authorization of using a modified version of the QGIS logo on their communications) . See https://www.qgis.org/en/site/getinvolved/governance/sustaining_members/sustaining_members.html#what-is-your-benefit-supporting-the-qgis-development

jmckenna:
I was also involved in the early QGIS.ORG draft, and it is indeed very similar to the words placed here in this document.  Having watched that entity grow and be so successful (in the sense that it helps both manage and attract funds/sponsors to help allow the QGIS software project to be sustainable), I can say that what is nice is how open they have been, as their budget has been fully accessible each year for where their spending is placed.  It does not list exact names of recipients, but it does list line items for each type of expenditure (20k windows packaging, 40k bug fixing, etc).  I hope that this document here for GDAL.ORG can also include a reporting requirements section, and list the requirements of the PSC to give a yearly report on budget/expenditures to both the GDAL community and the advisory board.

gdal/doc/source/development/rfc/rfc80_numfocus_relationship.rst
rouault: removed

gdal/doc/source/development/rfc/rfc80_numfocus_relationship.rst
rouault: I've added a few precisions regarding the advisory board (who takes part to meetings). Regarding how the mailing list if opened/closed, the current text is sufficiently vague so we can decide on that later.

gdal/doc/source/development/rfc/rfc80_numfocus_relationship.rst
```diff
@@ -0,0 +1,135 @@
...
+
+The GDAL Advisory Board is a newly forming group that is tasked with ensuring the project is evolving to meet the needs of its largest supporters. There will be periodic meetings with the advisory board, GDAL PSC members wishing to participate, and persons benefiting from funding, providing a forum to give feedback on the latest GDAL work. Board members will also be expected to share how they are using GDAL in their organization (in as much detail as is possible, understanding that some can only talk in general terms). Though the GDAL PSC owns the decision making around how the sponsorship money is allocated, the advisory board will provide key input on the maintenance priorities of the sponsors.
+Our hope is that these sessions can also lead to new feature development in GDAL through joint funding initiatives between top sponsors, providing more structure to the way GDAL features are funded today. This ideally leads to more cost-sharing as organizations will be able to be more strategic when they know what everyone else is interested in. These efforts will be above and beyond the core maintenance sponsorship, but the Advisory Board will be a key forum to enable it.
```
schwehr, line 112: Formatting nit here.  Add a blank line rather than a leading space.

rouault: there was no leading space AFAICS (probably just the artifact of the "+" of the diff), but blank line added for readability

gdal/doc/source/development/rfc/rfc80_numfocus_relationship.rst
rcoup: > Having all sponsors potentially invited to meetings would make it simpler 

IME it's hard enough getting _anyone_ to show up, might as well encourage anyone from the sponsors who's interested?

1. Should all sponsors be entitled to participate on the advisory board mailing list?

2. I'd suggest wrt the advisory board structure maybe something like:

```
The board will consist of up to X (6/8/10/12?) sponsor representatives. Each
Platinum sponsor is entitled to nominate a representative, with remaining
representatives nominated and selected by Gold and Silver sponsors via a process
to be determined by the PSC.
```

^ punts the problem into the future — leaves it up to the PSC; and nobody interested gets excluded because of rules until such time as there's too many people.

PR #3684, opened 2021-04-16
add auxiliary functions to gdal-utils

idanmiara:
## What does this PR do?

Add some auxiliary functions to be used in `gdal2xyz`, `gdalcalc`, `gdallocationinfo` and others.

## What are related issues/pull requests?

https://github.com/OSGeo/gdal/pull/3416

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments
 - [ ] All CI builds and checks have passed

PR #3685, opened 2021-04-16
gdallocationinfo.py - a direct Python port of gdallocationinfo.cpp

idanmiara:
## What does this PR do?

adds a new python utils - gdallocationinfo.py - a direct Python port of gdallocationinfo.cpp

I've ported all the functionality from gdallocationinfo.cpp into gdallocationinfo.py (except -xml, -lifonly).
The Python version now supports also float pixel/line for bilinear interpolation and multiple X Y points, as well as more SRS options.

https://github.com/OSGeo/gdal/pull/3684
https://github.com/OSGeo/gdal/pull/3416

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments


rouault:
For reasons exposed in another PR, I'd be OK for it as a sample script but not an official one to avoid confusion between C++ and py utilities

rouault (changes requested):
See above comment

idanmiara:
> For reasons exposed in another PR, I'd be OK for it as a sample script but not an official one to avoid confusion between C++ and py utilities

done

autotest/pyscripts/test_gdallocationinfo_py.py
```diff
@@ -0,0 +1,110 @@
...
+# DEALINGS IN THE SOFTWARE.
+###############################################################################
+
+
+from osgeo import gdal
+import test_py_scripts
+import pytest
+
+# test that numpy is available, if not skip all tests
+script_path = test_py_scripts.get_py_script('gdallocationinfo')
```
rouault, line 39: Should likely be updated to look in the samples folder. Tests are currently skipped AFAICS in the logs of the ubuntu 20.04 job

idanmiara: `get_py_script` looks in the samples folder as well as some other essential scripts are there. 
I'm not sure then why it skips, I'll have to investigate.

idanmiara: fixed

idanmiara:
@rouault it seems that only for `ubuntu 18.4 32bit` the function `ct.TransformPoint(x,y)` fails without the optional variable `z=0`.
I've added `z=0` in my code so it will pass the CI, and will try to investigate it in another PR.

gdal/swig/python/gdal-utils/osgeo_utils/samples/gdallocationinfo.py
```diff
@@ -0,0 +1,383 @@
...
+        if not interactive_mode:
+            args.x = np.array(args.xy[0::2])
+            args.y = np.array(args.xy[1::2])
+
+        if args.geoloc:
+            args.srs = LocationInfoSRS.SameAsDS_SRS
+        elif args.llgeoloc:
+            args.srs = LocationInfoSRS.SameAsDS_SRS_GeogCS
+        elif args.wgs84:
+            args.srs = 4226
```
rouault, line 332:
```suggestion
            args.srs = 4326
```
This is the reason of the failure on Ubuntu 18.04 32bit, which uses PROJ master, and using 4226 exhibits a bug in PROJ master (see the "ERROR 1: PROJ: internal_proj_create_operations: SQLite error on " ... error message in https://github.com/OSGeo/gdal/pull/3685/checks?check_run_id=2373251189 )
You can likely drop the extra ,0 in TransformPoint()

gdal/swig/python/gdal-utils/osgeo_utils/samples/gdallocationinfo.py
```diff
@@ -0,0 +1,383 @@
...
+                        help="The coordinate system of the input x, y location.")
+
+    parser.add_argument("-geoloc", dest="geoloc", action="store_true",
+                        help="Indicates input x,y points are in the georeferencing system of the image.")
+    parser.add_argument("-llgeoloc", dest="llgeoloc", action="store_true",
+                        help="Indicates input x,y points are in the long, lat (geographic) "
+                             "based on the georeferencing system of the image.")
+    parser.add_argument("-wgs84", dest="wgs84", action="store_true",
+                        help="Indicates input x,y points are WGS84 long, lat.")
+    parser.add_argument("-epsg", dest="epsg", metavar='epsg', type=int,
```
rouault, line 301: Not really sure we need this option. The list of options becomes quite confusing. -l_srs EPSG:XXXX should do the job, and it is how other utilities handle such case. And probably -l_srs WGS84 would also do the job

gdal/swig/python/gdal-utils/osgeo_utils/samples/gdallocationinfo.py
idanmiara: Oh, thanks, I wonder how it worked with the other builds, 4226 looks something completely different...

gdal/swig/python/gdal-utils/osgeo_utils/samples/gdallocationinfo.py
idanmiara: OK, removing -epsg. I'm leaving -wgs84 to be compatible with the cpp util.

idanmiara:
I've modularized the code.

gdal/swig/python/gdal-utils/osgeo_utils/auxiliary/osr_util.py
```diff
@@ -0,0 +1,88 @@
...
+def get_transform(src_srs: AnySRS, tgt_srs: AnySRS) -> Optional[osr.CoordinateTransformation]:
+    src_srs = get_srs(src_srs)
+    tgt_srs = get_srs(tgt_srs)
+    if src_srs.IsSame(tgt_srs):
+        return None
+    else:
+        return osr.CoordinateTransformation(src_srs, tgt_srs)
+
+
+def transform_points(ct: Optional[osr.CoordinateTransformation],
```
rouault, line 78: not a strong opinion here, but it seems to me that requiring a numpy.ndarray for a 1D array is overkill. Why not just a array.array ?

gdal/swig/python/gdal-utils/osgeo_utils/auxiliary/osr_util.py
```diff
@@ -0,0 +1,88 @@
...
+    elif isinstance(srs, gdal.Dataset):
+        srs = get_srs_from_ds(srs)
+    elif isinstance(srs, int):
+        srs_ = osr.SpatialReference()
+        if srs_.ImportFromEPSG(srs) != ogr.OGRERR_NONE:
+            raise Exception(f'ogr error when parsing srs from epsg:{srs}')
+        srs = srs_
+    elif isinstance(srs, str):
+        srs_ = osr.SpatialReference()
+        if srs_.ImportFromProj4(srs) != ogr.OGRERR_NONE:
```
rouault, line 52:
```suggestion
        if srs_.SetFromUserInput(srs) != ogr.OGRERR_NONE:
```
That way it will accept PROJ string or any other type of CRS string (WKT, PROJJSON, etc. etc)

gdal/swig/python/gdal-utils/osgeo_utils/auxiliary/osr_util.py
```diff
@@ -0,0 +1,88 @@
...
+            raise Exception(f'ogr error when parsing srs from proj string: {srs}')
+        srs = srs_
+    else:
+        raise Exception(f'Unknown SRS: {srs}')
+    if gis_order and int(osgeo.__version__[0]) >= 3:
+        # GDAL 3 changes axis order: https://github.com/OSGeo/gdal/issues/1546
+        srs.SetAxisMappingStrategy(osgeo.osr.OAMS_TRADITIONAL_GIS_ORDER)
```
rouault, line 59: I would not do forcing traditional GIS order in a utility function. Upper layers can do that if they wish, but I'd expect utility classes to be axis order compliant by default.  And we don't need to do the version checking as nobody < GDAL 3.3 is supposed to use that

idanmiara: Would it be OK to leave an optional gis_order variable with no default or with False default?
I think that the cpp version of `gdallocationinfo` -wgs84 uses GIS order, and I wanted the py version to be compatible.

Although it's not advertised, I believe most of `gdal_utils` is compatible with much lower versions of gdal, we could try to keep compatibility if it's not complicated, as in this case.

gdal/swig/python/gdal-utils/osgeo_utils/auxiliary/osr_util.py
idanmiara: thanks

gdal/swig/python/gdal-utils/osgeo_utils/auxiliary/osr_util.py
rouault: > Would it be OK to leave an optional gis_order variable with no default or with False default?

ok

gdal/swig/python/gdal-utils/osgeo_utils/auxiliary/osr_util.py
idanmiara: > not a strong opinion here, but it seems to me that requiring a numpy.ndarray for a 1D array is overkill. Why not just a array.array ?

Good point. I've modified it to accept both.
Would it be easy to implement a core version of `TransformPoints` that accepts x,y or x,y,z buffers, as this py function?

rouault: > Would it be easy to implement a core version of `TransformPoints` that accepts x,y or x,y,z buffers, as this py function?

Probably not that much. You'd have to change the implementation of ``%typemap(in,numinputs=1,fragment="DecomposeSequenceOf4DCoordinates") (int nCount, double *x, double *y, double *z, double *t) (int foundTime = FALSE)`` in swig/include/python/typemaps_python.i

idanmiara: > > Would it be easy to implement a core version of `TransformPoints` that accepts x,y or x,y,z buffers, as this py function?
> 
> Probably not that much. You'd have to change the implementation of `%typemap(in,numinputs=1,fragment="DecomposeSequenceOf4DCoordinates") (int nCount, double *x, double *y, double *z, double *t) (int foundTime = FALSE)` in swig/include/python/typemaps_python.i

Thanks, I'll have a look later...

idanmiara:
thanks!

PR #3688, opened 2021-04-17
FlatGeobuf: Backport bugfixes

bjornharrtell:


rouault:
I had already backported the fix for features > 32 KB in 61091db4d97c3ee036e896587bfaf7f1a27c65a2 . The other fix is really unlikely to be a practical problem. Can probably be a master only material IMHO. Up to you: if you want it to be backported, then just drop the first commit.

bjornharrtell:
@rouault I noticed that you had already done it but forgot to remove this, sorry!

PR #3691, opened 2021-04-18
epsg tr+esri2wkt - use ArgumentParser; add proper Help and some fixes

idanmiara:
## What does this PR do?

epsg tr + esri2wkt - use ArgumentParser; add proper Help; some fixes


rouault:
Merging, but I think for GDAL 3.4 we should consider flip those scripts to the samples category. They are of very limited general purpose use.

idanmiara:
> Merging, but I think for GDAL 3.4 we should consider flip those scripts to the samples category. They are of very limited general purpose use.

I agree, but since we want to advertise that the utils api is stable, unlike the auxiliary and samples, I think we better move them in 3.3.
I'll make a PR for that so you can evaluate.

PR #3700, opened 2021-04-20
Fix spelling errors.

sebastic:
The lintian QA tool reported spelling errors for the Debian package build:

 * Allows to VERB -> Allows VERBing


rouault:
The changes in gdal/swig/python/extensions/ will be lost in the next update of those files as they are generated SWIG files. The typo is in SWIG itself. I've just issued https://github.com/swig/swig/pull/1998 for this upstream fix

PR #3705, opened 2021-04-20
Fix OVERVIEW_LEVEL=NONE open option

rouault:
and also "{x}only". This however needs cooperation from drivers.
I've just updated the GTiff, JP2KAK, JPEG, OpenJPEG and VRT ones. Others
would need similar changes in their specific GetOverviewCount() / IRasterIO()
implementation.

Fixes https://github.com/OSGeo/gdal/pull/3181#issuecomment-822868290

### DIFF
--- a/autotest/gcore/overviewds.py
+++ b/autotest/gcore/overviewds.py
@@ -49,25 +49,26 @@ def test_overviewds_1():
 # Nominal cases
 
 @pytest.mark.parametrize("externalOverviews", [True, False])
-def test_overviewds_2(externalOverviews):
+def test_overviewds_2(tmp_path, externalOverviews):
 
-    shutil.copy('data/byte.tif', 'tmp')
+    shutil.copy('data/byte.tif', tmp_path)
+    tmpfilename = str(tmp_path.joinpath('byte.tif'))
     if externalOverviews:
-        ds = gdal.Open('tmp/byte.tif')
+        ds = gdal.Open(tmpfilename)
         ds.BuildOverviews('NEAR', overviewlist=[2, 4])
         ds = None
-        ds = gdal.Open('tmp/byte.tif', gdal.GA_Update)
+        ds = gdal.Open(tmpfilename, gdal.GA_Update)
     else:
-        ds = gdal.Open('tmp/byte.tif', gdal.GA_Update)
+        ds = gdal.Open(tmpfilename, gdal.GA_Update)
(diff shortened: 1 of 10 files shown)


autotest/gcore/overviewds.py
```diff
@@ -105,6 +122,8 @@ def test_overviewds_2():
     assert ds.GetMetadataItem('RPC', 'FOO') is None
     ds = None
 
+    gdal.GetDriverByName('GTiff').Delete('tmp/byte.tif')
```
sgillies, line 125: @rouault if we use pytest's `tmp_path` fixture, there's no need for manual cleanup of the temp files.

autotest/gcore/overviewds.py
```diff
@@ -48,31 +48,46 @@ def test_overviewds_1():
 ###############################################################################
 # Nominal cases
 
-
-def test_overviewds_2():
+@pytest.mark.parametrize("externalOverviews", [True, False])
+def test_overviewds_2(externalOverviews):
```
sgillies, line 52:
```suggestion
def test_overviewds_2(tmp_path, externalOverviews):
```

autotest/gcore/overviewds.py
```diff
@@ -48,31 +48,46 @@ def test_overviewds_1():
...
 
     shutil.copy('data/byte.tif', 'tmp')
```
sgillies, line 54:
```suggestion
    shutil.copy('data/byte.tif', tmp_path)
```

autotest/gcore/overviewds.py
```diff
@@ -48,31 +48,46 @@ def test_overviewds_1():
...
-    ds = gdal.Open('tmp/byte.tif')
-    ds.BuildOverviews('NEAR', overviewlist=[2, 4])
+    if externalOverviews:
+        ds = gdal.Open('tmp/byte.tif')
```
sgillies, line 56:
```suggestion
        ds = gdal.Open(str(tmp_path.joinpath('byte.tif')))
```

sgillies: &c.

sgillies:
@rouault do I understand correctly that we'll release 3.3.0 with possible bugs in other drivers? That's okay with me.

rouault:
> do I understand correctly that we'll release 3.3.0 with possible bugs in other drivers?

yes, that might not work properly in a few other drivers that have specialized behavior regarding use of overviews.

PR #3706, opened 2021-04-20
ArgumentParser: gdal_calc + others

idanmiara:
## What does this PR do?

gdal_argparse.py - GDALArgumentParser an extended `argparse.ArgumentParser` to be used with all `gdal-utils`
gdal_calc.py - use `GDALArgumentParser ` instead of the old `optparse.OptionParser` 
use minor improvements to `ArgumentParser` uses in other utils

### DIFF
--- a/gdal/swig/python/gdal-utils/osgeo_utils/auxiliary/gdal_argparse.py
+++ b/gdal/swig/python/gdal-utils/osgeo_utils/auxiliary/gdal_argparse.py
@@ -1,0 +1,86 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+# ******************************************************************************
+#
+#  Project:  GDAL utils.auxiliary
+#  Purpose:  an extended argparse.ArgumentParser to be used with all gdal-utils
+#  Author:   Idan Miara <idan@miara.com>
+#
+# ******************************************************************************
+#  Copyright (c) 2021, Idan Miara <idan@miara.com>
+#
+#  Permission is hereby granted, free of charge, to any person obtaining a
+#  copy of this software and associated documentation files (the "Software"),
+#  to deal in the Software without restriction, including without limitation
+#  the rights to use, copy, modify, merge, publish, distribute, sublicense,
+#  and/or sell copies of the Software, and to permit persons to whom the
+#  Software is furnished to do so, subject to the following conditions:
+#
+#  The above copyright notice and this permission notice shall be included
+#  in all copies or substantial portions of the Software.
+#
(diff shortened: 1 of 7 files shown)


gdal/swig/python/gdal-utils/osgeo_utils/auxiliary/gdal_argparse.py
```diff
@@ -0,0 +1,76 @@
...
+            # extend was introduced to the stdlib in Python 3.8
+            self.register('action', 'extend', ExtendAction)
+
+    def parse_args(self, args=None, optfile_arg=None, **kwargs):
+        if (args is not None
+           and optfile_arg in args
+           and self.fromfile_prefix_chars is not None
+           and optfile_arg is not None):
+
+            # Replace '--optfile x' with the standard '@x'
```
rouault, line 58: I'm not sure they are completely equivalent. From what I can read in https://docs.python.org/3/library/argparse.html#fromfile-prefix-chars , @x requires one argument per line whereas --optfile x can support potentially arguments on the same line

idanmiara: Thanks for your comment.
I actually handled that in GDALArgumentaParser. 
It passes the relevant test7 for that cases you've mentioned.

rouault: ok, any reason to have optfile_arg default to None and being set only in gdal_calc.py to "--optfile" ?

And is this code exercised by the test suite ? (my mantra is "non-tested code is broken, or will soon be broken", and it is even more true for an interpreted language like Python)

idanmiara: --optfile is for backwards compatibility. I can't see a reason to encourage people to use it ovet the standard pefix@.
I'll set it on any until that currently have this option.

I've learned to know your mantra :) 
Almost all the tests of `gdal_calc` use the ArgParser so it is tested there through it.
Did you refer to any other code?

rouault: Regarding testing, I refer to this new code replacing --optfile by ``@ ``

idanmiara: Calc test 7 tests that as it inputs `- -optfile` and the parser transforms it to `@` and parses it as the normal way. 

I can add a test for input with the prefix.

idanmiara: I've added tests for `@` prefix and a deprecation warning for `--optfile`.

Issue #3711, opened 2021-04-21
CI: Unable to find package Microsoft.NETCore.App in windows_build.yml


rouault:
@bjornharrtell @szekerest  Starting today, we get an error "D:\a\gdal\gdal\gdal\swig\csharp\apps\ogrinfo.csproj : error NU1101: Unable to find package Microsoft.NETCore.App. No packages exist with this id in source(s): Microsoft Visual Studio Offline Packages" in .github/workflows/windows_build.yml : https://github.com/OSGeo/gdal/pull/3710/checks?check_run_id=2400373159

any idea how to fix that?

rouault:
There's a ``NETCORE = netcoreapp2.1`` default value in swig/csharp/makefile.vc  . Eithat that could/should be changed, or overriding it in the nmake.local file generated in windows_build.yml

bjornharrtell:
I've made https://github.com/OSGeo/gdal/pull/3712 and I'll check up on what CI says about it.

bjornharrtell:
Doesn't appear to help. The error is refering to "Microsoft Visual Studio Offline Packages" as the source of dependencies and I'm not sure what that is but it appears to not have what was and is needed.

PR #3712, opened 2021-04-21
Target netcoreapp3.1

bjornharrtell:
ref https://github.com/OSGeo/gdal/issues/3711

### DIFF
--- a/gdal/swig/csharp/makefile.vc
+++ b/gdal/swig/csharp/makefile.vc
@@ -33,7 +33,7 @@ CSC = mcs
 NETSTANDARD = netstandard2.0
 !ENDIF
 !IFNDEF NETCORE
-NETCORE = netcoreapp2.1
+NETCORE = netcoreapp3.1
 !ENDIF
 !IF $(MSVC_VER) >= 1400
 !IFDEF WIN64


bjornharrtell:
Doesn't appear to help. The error is refering to "Microsoft Visual Studio Offline Packages" as the source of dependencies and I'm not sure what that is but it appears to not have what was and is needed.

rouault:
maybe using https://github.com/actions/setup-dotnet would help ?  Defaulting to 3.1 breaks the appveyor.yml build

rouault:
not sure if it is a change in the VM (https://github.com/actions/virtual-environments/commits/main/images/win/Windows2019-Readme.md doesn't mention a today or yesterday change). Might be on the nuget side ?

bjornharrtell:
it could possibly help to explicitly add expected toolchain at least that's indicated in this similar issue - https://developercommunity.visualstudio.com/t/nuget-restore-failing-on-azure-agents-packages-not/985186

rouault:
I'm not sure if this PR is still of interest after https://github.com/OSGeo/gdal/pull/3769?

bjornharrtell:
It's probably a good idea to move away from .NET Core 2.1 at some point. Not sure if the appveyor setup has much life left?

rouault:
> Not sure if the appveyor setup has much life left?

probably not that much, but it shouldn't be too difficult to specify the old dotnetcore by putting its value in the nmake.local generated in appveyor.yml

PR #3727, opened 2021-04-23
Gdal2tiles tests

idanmiara:
@rouault 

I'm not sure what's up with these tests, but it looks like they are related to this commit:
https://github.com/OSGeo/gdal/commits/543cd668c2d8d3c0eb846ba4437595150fb345c2


rouault:
> I'm not sure what's up with these tests, but it looks like they are related to this commit:

you should rather credit gregory.bataille@gmail.com / @gbataille for those

idanmiara:
> you should rather credit [gregory.bataille@gmail.com](mailto:gregory.bataille@gmail.com) / @gbataille for those
@gbataille Please approve that the credit is proper.

It seems that these tests were not running and got broken some long time ago as `zoom` is a mandatory parameter that is missing from some option in a test.
I'm trying to fix it.

gdal/swig/python/gdal-utils/osgeo_utils/gdal2tiles.py
```diff
@@ -1487,6 +1488,9 @@ def options_post_processing(options: Options, input_file: str, output_folder: st
         options.url += os.path.basename(out_path) + '/'
 
     # Supported options
+    if options.resampling == 'average' and not hasattr(gdal, 'RegenerateOverview'):
```
rouault, line 1491: why is this needed ? From 17f371da268 , it looks like gdal.RegenerateOverview() has been around since 2008...

idanmiara: https://github.com/OSGeo/gdal/pull/3727/files#diff-237f5077fbc4fbc3614dbf1bbc609439a7d5b73fcb5a1a00e1509f14befe893eR154

I don't know what was the original purpose of this test (It didn't run since 2008?)
Do you advise to remove that check and the test?

rouault: > Do you advise to remove that check and the test?

yes, maybe this was average resampling that wasn't available in an ancient GDAL. Anyway let's remove those traces of ancient times :-)

idanmiara: OK. I removed it in a separate commit so that the removal of that test makes sense

PR #3728, opened 2021-04-23
Add /vsigs/ support to the tiledb driver

vincentschut:
This PR adds /vsigs/ support to the TileDB driver, allowing it to access datasets on gcs.


normanb:
@vincentschut Thank you for this PR, it looks great to me.

PR #3741, opened 2021-04-25
GPKG: fix a few comments, and improve a bit testing

rouault:


jratike80:
About the ALTER TABLE ADD COLUMN ... NOT NULL case, do you know if SQLite is using a trick to avoid writing any real data into the new NOT NULL column of existing rows? Documentation https://www.sqlite.org/lang_altertable.html#altertabaddcol seems to suggest this 

> The ALTER TABLE command works by modifying the SQL text of the schema stored in the sqlite_schema table. No changes are made to table content for renames or column addition. Because of this, the execution time of such ALTER TABLE commands is independent of the amount of data in the table. They run as quickly on a table with 10 million rows as on a table with 1 row.

That would make it understandable why the default value is required.

rouault:
> About the ALTER TABLE ADD COLUMN ... NOT NULL case, do you know if SQLite is using a trick to avoid writing any real data into the new NOT NULL column of existing rows?

In the documentation of the sqlite3 format itself, I believe they explain that the data for a given row can be "truncated" when the values for the last columns is the default. See https://www.sqlite.org/fileformat.html : "A record might have fewer values than the number of columns in the corresponding table. This can happen, for example, after an ALTER TABLE ... ADD COLUMN SQL statement has increased the number of columns in the table schema without modifying preexisting rows in the table. Missing values at the end of the record are filled in using the default value for the corresponding columns defined in the table schema. "

> That would make it understandable why the default value is required.

Yes, and the most reason is that if you add a column with a NOT NULL constraint, as extending existing rows is like putting NULL into them, you need to specify a default value that is not NULL to not violate the constraint

jratike80:
A test with a table having 6877009 rows:
```
Execution finished without errors.
Result: query executed successfully. Took 387ms
At line 1:
alter table korkeuskayra
add COLUMN new_column TEXT NOT NULL default 'foo';
```

So it is an excellent trick even it could be even better without default when the table is empty.

PR #3743, opened 2021-04-26
`gdal-utils` - add type hints

idanmiara:
add type hints and make main logic functions for some gdal-utils


gdal/swig/python/gdal-utils/osgeo_utils/gdal_proximity.py
```diff
@@ -48,15 +50,15 @@ def Usage():
 
 
 def main(argv):
-    frmt = None
+    driver = None
```
rouault, line 53: driver_name would be more appropriate (driver could be confused with an object of type gdal.Driver) . Same for similar change in other utilities

idanmiara:
I've added a `gdal_pansharpen` commit

PR #3745, opened 2021-04-26
GTIFF/COG: fix overviews with NaN nodata value

rouault:
Creating a COG file with nodata=NaN could result in the overviews
being corrupted.

CC @kannes This fixes the issue you reported yesterday on #gdal


rouault:
CC @kannes

rouault:
Backported to 3.3 (post 3.3.0 RC1) and 3.2

kannes:
Thank you! I can confirm it works fine now. The GeoTIFF was created via GMT's xyz2grd btw.

PR #3769, opened 2021-05-01
Fix for .net core build error fixes #3711

### LINKED ISSUE #3711: CI: Unable to find package Microsoft.NETCore.App in windows_build.yml

runette:
fixes #3711 

This fixes the problem. It is based on a concatenation of the various threads about this across the web.

The effect seems to be widespread. The cause seems to be some as yet unidentified change in the vs2019 environment.

### DIFF
--- a/.github/workflows/windows_build.yml
+++ b/.github/workflows/windows_build.yml
@@ -228,9 +228,9 @@ jobs:
             cd $env:GITHUB_WORKSPACE\gdal\swig\python
             exec { python setup.py build }
             exec { python setup.py install }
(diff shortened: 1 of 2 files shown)


rouault:
Thanks!

Issue #3370, opened 2021-01-11
NuGet .NetCore Package Issue - DLLs not Linking Properly, and OGR Missing Classes


BBoldenow:
## Edit
Sorry for the name of the issue, I thought there were missing classes when I first started writing it, but investigated some more and believe everything is there. The DLL issue remains however and is as follows:

## Expected behavior and actual behavior.

When consuming a 2.4.x package, the OSGeo namespace is not found by default. I am able to browse to the downloaded DLLS and reference them manually from: C:\Users\me\\.nuget\packages\gdal\2.4.4\lib\net40

When I do, I can see the namespaces for OSGeo.

It's expected that installing a NuGet package should automatically link the appropriate DLLs. I believe this is because this is being packaged into the lib\net40 folder instead of lib\netcoreapp3.1 - since this project has support for .NetCore now, it's expected that there be multiple framework targets like so:

![image](https://user-images.githubusercontent.com/6502212/104225678-efd0b680-5403-11eb-976a-12e612acc03b.png)

## Steps to reproduce the problem.

Install the latest stable NuGet package in a .NetCore application. Notice that it does not find the OSGeo namespaces due to the NuGet package not targeting multiple frameworks.

## Operating system

Windows 10 x64 v. 1909 build 18363.1198

## GDAL version and provenance

The 2.4.4 version from nuget.org installed via Visual Studio 2019.

rouault:
CC @szekerest @bjornharrtell

bjornharrtell:
@BBoldenow what makes you think that this project supports .net core out of the box? That's news to me. That said, https://github.com/MaxRev-Dev/gdal.netcore has been working nicely for me.

szekerest:
The GDAL.* nuget packages haven't yet been upgrated with the recent .net standard builds. The current packages support .net framework only.

BBoldenow:
@bjornharrtell Maybe that was a mistake on my part - I thought I had seen quite a bit of traffic about .net core working in the issues logs here. Either way, the MaxRev version won't work for me due to a lack of support for a large chunk of the formats I use GDAL for. I currently have my own CLR wrapper for GDAL, but am hoping to get on an official one to get up-to-date. The one I use is extremely out of date.

@szekerest Thanks for the info. Interestingly it seems to work just fine if I reference the DLLs - I have no idea how that's possible if the current packages are .NetFramework only, very strange!

bjornharrtell:
[...]

PR #3771, opened 2021-05-01
EXR: fix build against OpenEXR 3.0.1 (fixes #3370)

rouault:


rouault:
Wrong ticket number... PR superseded per https://github.com/OSGeo/gdal/pull/3772

PR #3776, opened 2021-05-02
[WIP] add some `gdal_utils` support functions

idanmiara:
## What does this PR do?

draft
add some `gdal_utils` support functions

### DIFF
--- a/gdal/swig/python/gdal-utils/osgeo_utils/auxiliary/ogr_utils.py
+++ b/gdal/swig/python/gdal-utils/osgeo_utils/auxiliary/ogr_utils.py
@@ -1,0 +1,67 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+# ******************************************************************************
+#
+#  Project:  GDAL utils.auxiliary
+#  Purpose:  OGR utility functions
+#  Author:   Idan Miara <idan@miara.com>
+#
+# ******************************************************************************
+#  Copyright (c) 2021, Idan Miara <idan@miara.com>
+#
+#  Permission is hereby granted, free of charge, to any person obtaining a
+#  copy of this software and associated documentation files (the "Software"),
+#  to deal in the Software without restriction, including without limitation
+#  the rights to use, copy, modify, merge, publish, distribute, sublicense,
+#  and/or sell copies of the Software, and to permit persons to whom the
+#  Software is furnished to do so, subject to the following conditions:
+#
+#  The above copyright notice and this permission notice shall be included
+#  in all copies or substantial portions of the Software.
+#
+#  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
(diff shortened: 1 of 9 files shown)


gdal/swig/python/gdal-utils/osgeo_utils/auxiliary/ogr_utils.py
```diff
@@ -0,0 +1,67 @@
...
+        geom = ogr.CreateGeometryFromWkt(wkt)
+        feature.SetGeometry(geom)  # Set the feature geometry
+        layer.CreateFeature(feature)  # Create the feature in the layer
+        feature.Destroy()  # Destroy the feature to free resources
+    # Destroy the data source to free resources
+    ds.Destroy()
+
+
+def ogr_get_layer_extent(lyr: ogr.Layer) -> GeoRectangle:
+    result = None
```
rouault, line 58: why not use Layer.GetExtent() ?

rouault (commented):
I'm not sure what to do with this PR. It is not obvious that all those utility functions are needed, as they are not used by anything else.
I'm really wondering why there's a specific need for UTM handling in osr_utm_util.py 
And there are no tests for all the new functions

gdal/swig/python/gdal-utils/osgeo_utils/auxiliary/osr_utm_util.py
```diff
@@ -0,0 +1,98 @@
...
+
+    extent_points = []
+    for x in x_arr:
+        for y in y_arr:
+            extent_points.append((x, y))
+        y_arr.reverse()
+    return extent_points
+
+
+def proj_string_from_utm_zone(zone: Optional[Real] = None, datum_str='+datum=WGS84') -> str:
```
rouault, line 83: proj string for CRS are a legacy. We should really try hard not to use them

idanmiara:
> I'm not sure what to do with this PR. It is not obvious that all those utility functions are needed, as they are not used by anything else.
> I'm really wondering why there's a specific need for UTM handling in osr_utm_util.py
> And there are no tests for all the new functions

Thanks for your review, I meant to use these in some utils and add tests, so I wanted to test them first, that's why I marked it as [WIP]. I didn't expect to merge it beforehand.

gdal/swig/python/gdal-utils/osgeo_utils/auxiliary/ogr_utils.py
idanmiara: Thanks, I'll look into it.

gdal/swig/python/gdal-utils/osgeo_utils/auxiliary/osr_utm_util.py
idanmiara: What would be a non-legacy way to define the CRS that sits between `EPSG:326xy` and `EPSG:326xy+1` i.e. "zone xy+0.5" ?

rouault: use the SpatialReference.SetTransverseMercator() method with appropriate values

idanmiara:
I'll reopen this PR when it's ready.

PR #3781, opened 2021-05-03
Fix tests for OSX

abellgithub:
## What does this PR do?

Fix CI tests on OSX

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS: OSX
* Compiler:

### DIFF
--- a/autotest/gcore/cog.py
+++ b/autotest/gcore/cog.py
@@ -389,7 +389,7 @@ def test_cog_small_world_to_web_mercator():
     if sys.platform == 'darwin' and gdal.GetConfigOption('TRAVIS', None) is not None:
         assert got_cs != [0, 0, 0]
     else:
-        assert got_cs == [26293, 23439, 14955]
+        assert got_cs == [26293, 23439, 14955] or got_cs == [26228, 22085, 12992]
     assert ds.GetRasterBand(1).GetMaskBand().Checksum() == 17849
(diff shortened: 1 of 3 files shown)


autotest/gcore/cog.py
```diff
@@ -389,7 +389,7 @@ def my_cbk(pct, _, arg):
     if sys.platform == 'darwin' and gdal.GetConfigOption('TRAVIS', None) is not None:
         assert got_cs != [0, 0, 0]
     else:
-        assert got_cs == [26293, 23439, 14955]
+        assert got_cs == [26293, 23439, 14955] or got_cs == [26228, 22085, 12992]
```
sgillies, line 392: @abellgithub @rouault the project could use pytest's parameterization feature https://docs.pytest.org/en/stable/parametrize.html so that it can have separated, isolated, platform-dependent tests instead of using if statements within one test.

rouault: can you remove the "if sys.platform == 'darwin' and gdal.GetConfigOption('TRAVIS', None) is not None:" line. I'd expect the updated assertion will fix it

abellgithub: Sure. Didn't want to be presumptuous.

PR #3784, opened 2021-05-03
Fix the Oracle DB connection when user and pwd are empty, while Oracl…

fechen123:
…e wallet is used.

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:

### DIFF
--- a/gdal/frmts/georaster/oci_wrapper.cpp
+++ b/gdal/frmts/georaster/oci_wrapper.cpp
@@ -122,8 +122,7 @@ OWConnection::OWConnection( const char* pszUserIn,
 
     ub4 eCred = OCI_CRED_RDBMS;
 
-    if( EQUAL(pszServer, "") &&
-        EQUAL(pszPassword, "") &&
+    if( EQUAL(pszPassword, "") &&
         EQUAL(pszUser, "") )
     {
         eCred = OCI_CRED_EXT;
@@ -237,6 +236,21 @@ OWConnection::OWConnection( const char* pszUserIn,
     QueryVersion();
 
     bSuceeeded = true;
+
+    // ------------------------------------------------------
+    //  If no user specified, get it from the session 
+    // ------------------------------------------------------
+    if (EQUAL(pszUser, "") )
+    {
+      OWStatement* poStmt = CreateStatement(
+            "select sys_context('userenv','session_user')\n"
+            "from dual\n" );
+
+      poStmt->Define(pszUser);
+      poStmt->Execute();
+      delete poStmt;
+      CPLDebug("OCI: ", "Implicit User: %s\n", pszUser);
+    }
 
     // ------------------------------------------------------
     //  Initialize/Describe types
--- a/gdal/frmts/georaster/oci_wrapper.cpp
+++ b/gdal/frmts/georaster/oci_wrapper.cpp
@@ -246,8 +246,9 @@ OWConnection::OWConnection( const char* pszUserIn,
             "select sys_context('userenv','session_user')\n"
             "from dual\n" );
 
+      pszUser = static_cast<char*>(CPLRealloc(pszUser, OWNAME)); 
       poStmt->Define(pszUser);
-      poStmt->Execute();
+      CPL_IGNORE_RET_VAL(poStmt->Execute());
       delete poStmt;
       CPLDebug("OCI: ", "Implicit User: %s\n", pszUser);
     }
--- a/gdal/ogr/ogrsf_frmts/oci/ogrocisession.cpp
+++ b/gdal/ogr/ogrsf_frmts/oci/ogrocisession.cpp
@@ -128,12 +128,20 @@ int OGROCISession::EstablishSession( const char *pszUseridIn,
 /*      Operational Systems's authentication option                     */
 /* -------------------------------------------------------------------- */
 
+    //Setting to "/" for backward compatibility
+    const char* pszUser = "/";
+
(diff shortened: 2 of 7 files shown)


gdal/frmts/georaster/oci_wrapper.cpp
```diff
@@ -238,6 +237,21 @@ OWConnection::OWConnection( const char* pszUserIn,
...
+    // ------------------------------------------------------
+    //  If no user specified, get it from the session 
+    // ------------------------------------------------------
+    if (EQUAL(pszUser, "") )
+    {
+      OWStatement* poStmt = CreateStatement(
+            "select sys_context('userenv','session_user')\n"
+            "from dual\n" );
+
+      poStmt->Define(pszUser);
```
rouault, line 249: The Define(char*, int nSize = OWNAME) method expects the target variable to be char [OWNAME] . Here pszUser is just 1-byte large, so this will much likely cause a buffer overflow

gdal/frmts/georaster/oci_wrapper.cpp
```diff
@@ -238,6 +237,21 @@ OWConnection::OWConnection( const char* pszUserIn,
...
+      poStmt->Execute();
+      delete poStmt;
```
rouault, line 251: shouldn't the return of Execute() be tested ?

gdal/ogr/ogrsf_frmts/oci/ogrocisession.cpp
```diff
@@ -130,9 +130,8 @@ int OGROCISession::EstablishSession( const char *pszUseridIn,
 
     ub4 eCred = OCI_CRED_RDBMS;
 
-    if( EQUAL(pszDatabaseIn, "") &&
-        EQUAL(pszPasswordIn, "") &&
-        EQUAL(pszUseridIn, "/") )
+    if( EQUAL(pszPasswordIn, "") &&
```
rouault, line 133: The EQUAL(pszUseridIn, "/") test has been changed to EQUAL(pszUseridIn, "").  Cannot pszUseridIn == '/' happy anymore ? Or what will happen in that case ?
Sorry if my question is a bit dump. I'm not familiar with Oracle

gdal/frmts/georaster/oci_wrapper.cpp
fechen123: You are correct. I thought the variable memory was allocated already. Can I ask what is the function to allocate the memory for this string?

rouault: I guess pszUser = static_cast<char*>(CPLRealloc(pszUser, OWNAME)); should do the job

gdal/frmts/georaster/oci_wrapper.cpp
fechen123: Sure. I will fix this.

gdal/ogr/ogrsf_frmts/oci/ogrocisession.cpp
fechen123: I am not an expert on this either. From the documentation (https://docs.oracle.com/en/database/oracle/oracle-database/21/lnoci/connect-authorize-and-initialize-functions.html#GUID-D60E78B5-E600-4CDE-87A7-B4ACDA9FEE5E)
"OCI_CRED_EXT - Authenticate using external credentials. No user name or password is provided. "
"The second method is to use external credentials. No attributes need to be set on the user session handle before calling OCISessionBegin(). The credential type is OCI_CRED_EXT. This is equivalent to the Oracle7 'connect /' syntax. If values have been set for OCI_ATTR_USERNAME and OCI_ATTR_PASSWORD, then these are ignored if OCI_CRED_EXT is used. "

From ogrocidatasource.cpp, pszUseridIn is set "/" only when the user, pwd and database are all empty. It seems that that is the old way (Oracle 7?) of specifying the credentials. If we want backward compatibility,  I can keep the original code and add the new way, like this: 
    if( EQUAL(pszDatabaseIn, "") &&
        EQUAL(pszPasswordIn, "") &&
        EQUAL(pszUseridIn, "/")  ||
        EQUAL(pszPasswordIn, "") && 
        EQUAL(pszUseridIn, ""))

rouault: > I am not an expert on this either

Hum... I can't really tell on what's the appropriate change, but I don't think we need to support ancient/deprecated Oracle. If supports for the "traditional" authentication support that was handled by the driver up to know still works with recent enough versions of Oracle, in addition to the "wallet" one, that seems fine to me.
Did you try running the autotest/gdrivers/georaster.py tests ? No idea if/how they work.
(It's a bit of a pain to support that proprietary stuff since we can't easily tests that in continuous integration)

fechen123: I didn't run the georaster.py test because it won't test the new feature which requires the oracle wallet setup. I only manually tested the driver to make sure the connection like "geor:/@orcl" works. The current change should not affect the "traditional" connect string where user and pwd are not empty.

rouault: > The current change should not affect the "traditional" connect string where user and pwd are not empty.

sounds good

PR #3786, opened 2021-05-04
configure: fix explicit disabling of geos/sfcgal

### LINKED ISSUE #3782: Unable to disable GEOS and SFCGAL when building from source

I expected to be able to disable GEOS support when building from source but it returns an error message instead.

```console
$ ./configure --with-geos=no
...
GEOS support disabled
configure: error: GEOS support explicitly enabled, but geos-config could not be found
```

## Operating system

macOS 10.15.7

GDAL 3.3.0

This used to work fine in GDAL 3.2.2, so this is a new bug.

tbonfort:
fixes #3782

### DIFF
--- a/gdal/configure
+++ b/gdal/configure
@@ -38323,6 +38323,7 @@ fi
     { $as_echo "$as_me:${as_lineno-$LINENO}: result: GEOS support disabled" >&5
 $as_echo "GEOS support disabled" >&6; }
     GEOS_CONFIG=no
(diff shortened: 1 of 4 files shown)


rouault:
Looks good to me.
@QuLogic does that look good to you as well ?

adamjstewart:
I can confirm that this fixes #3782, thanks for the quick fix @tbonfort!

PR #3788, opened 2021-05-04
Eliminate thread lock timeouts on Windows

### LINKED ISSUE #3631: gdalwarp error "Failed to acquire WarpMutex in WarpRegion()"

`ERROR 1: Failed to acquire WarpMutex in WarpRegion()`  on source single channel 16bit image of size 93723, 33409.

gdalinfo on source: [tri_mergedDEM_clipped_16bit.tif.txt](https://github.com/OSGeo/gdal/files/6238875/tri_mergedDEM_clipped_16bit.tif.txt)

    gdalwarp --config GDAL_CACHEMAX 4096 -wm 512 -multi -co TILED=YES ^
    -co COMPRESS=ZSTD -t_srs EPSG:3579 -r mode -co predictor=2  ^
     IN.tif  OUT.tif

## Operating system

Windows 10 Enterprise, Version 10.0.19041 Build 19041

GDAL 3.1.4, released 2020/10/20 via Qgis v3.16.

## Analysis

> I see that in the -multi mode, the I/O threads waits for a maximum of 10 minutes for the computation thread to have finished. And you likely hit that limit because of the large value of -wm you set. There's no reason we max at 10 minutes however. 

From thread http://osgeo-org.1560.x6.nabble.com/gdal-dev-More-info-on-gdalwarp-resampling-methods-tp5484028p5484100.html

abellgithub:
## What does this PR do?

This updates the thread handling in the GDAL warp kernel to use standard C++ and elimintes the arbitrary timeouts that previously existed on windows. It uses RAII to eliminate missed unlocks and memory cleanup in case of exceptions.

Close #3631

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:


gdal/alg/gdalwarpkernel.cpp
```diff
@@ -394,38 +369,40 @@ static void ThreadFuncAdapter(void* pData)
...
+            pTransformerArg = psThreadData->pTransformerArgInput;
+            psThreadData->mapThreadToTransformerArg[nThreadId] = pTransformerArg;
+        }
     }
-    CPLReleaseMutex(psThreadData->hCondMutex);
 
     // If no transformer assigned to current thread, instantiate one
     if( pTransformerArg == nullptr )
     {
+        std::lock_guard<std::mutex> lock(psThreadData->mutex);
```
rouault, line 393: GDALCloneTransformer() might be an expensive operation. It doesn't need to run under the mutex, so you could move the lock just after it

rouault (commented):
Besides my remark, this looks good. Thanks for the cleanup!

gdal/alg/gdalwarpkernel.cpp
abellgithub: That's fine. It wasn't clear from the previous code what was being protected by the mutex. I wasn't sure if it was the `mapThreadToTransformerArg` or the source `pTransformerArgInput` itself or both.

I also assumed the clone would be trivial, so it was better just to protect it as well.

PR #3790, opened 2021-05-04
Geometry::MakeValid(): add option to select the GEOS 3.10 'structure' method

rouault:


jratike80:
Information about what it means in https://github.com/libgeos/geos/pull/433.

rouault:
From Doxygen doc:
```
* The following options are available:
 * <ul>
 * <li>METHOD=LINEWORK/STRUCTURE.
 *     LINEWORK is the default method, which combines all rings into a set of
 *     noded lines and then extracts valid polygons from that linework.
 *     The STRUCTURE method (requires GEOS >= 3.10 and GDAL >= 3.4) first makes
 *     all rings valid, then merges shells and
 *     subtracts holes from shells to generate valid result. Assumes that
 *     holes and shells are correctly categorized.</li>
 * <li>KEEP_COLLAPSED=YES/NO. Only for METHOD=STRUCTURE.
 *     NO (default): collapses are converted to empty geometries
 *     YES: collapses are converted to a valid geometry of lower dimension.</li>
 * </ul>
```

PR #3809, opened 2021-05-06
OGRCT: Improve OGRProjCT::Clone() and export OGRCoordinateTransformation::Clone() to C API

drons:


gdal/ogr/ogrct.cpp
```diff
@@ -926,6 +943,34 @@ OGRProjCT::OGRProjCT()
...
+    bSourceWrap(other.bSourceWrap),
+    dfSourceWrapLong(other.dfSourceWrapLong),
+    poSRSTarget((other.poSRSTarget != nullptr) ? (other.poSRSTarget->Clone()) : (nullptr)),
+    bTargetLatLong(other.bTargetLatLong),
+    bTargetWrap(other.bTargetWrap),
+    dfTargetWrapLong(other.dfTargetWrapLong),
+    bWebMercatorToWGS84LongLat(other.bWebMercatorToWGS84LongLat),
+    nErrorCount(other.nErrorCount),
+    dfThreshold(other.dfThreshold),
+    m_pj((other.m_pj != nullptr) ? (proj_clone(OSRGetProjTLSContext(), other.m_pj)) : (nullptr)),
```
rouault, line 963: as noted in OGRProjCT::GetInverse(), proj_clone might fail in some circumstances for PROJ < 8.0.1, so you need a fallback path that recomputes m_pj with part of the logic of OGRProjCT::Initialize(). In practice you might want to do that in OCTClone() by calling OGRCreateCoordinateTransformation()

gdal/ogr/ogrct.cpp
```diff
@@ -2399,6 +2442,37 @@ int OGRProjCT::TransformWithErrorCodes(
...
+            bCloneDone = bCloneDone &&
+                         ((m_oTransformations[i].pj == nullptr) ==
+                          (poNewCT->m_oTransformations[i].pj == nullptr));
+        }
+    }
+    if(!bCloneDone)
+    {
+        delete poNewCT;
+        poNewCT = new OGRProjCT();
+        poNewCT->Initialize(poSRSSource, poSRSTarget, m_options);
```
drons, line 2470: I'm not sure if I need to check the result of `Initialize` call.
I think, it is redundant.

rouault: it should rarely fail, but that might be more prudent to check it

gdal/ogr/ogrct.cpp
```diff
@@ -2399,6 +2442,37 @@ int OGRProjCT::TransformWithErrorCodes(
...
+/************************************************************************/
+
+OGRCoordinateTransformation* OGRProjCT::Clone() const
+{
+    OGRProjCT* poNewCT = new OGRProjCT(*this);
+#if (PROJ_VERSION_MAJOR * 10000 + PROJ_VERSION_MINOR * 100 + PROJ_VERSION_PATCH) < 80001
+    // See https://github.com/OSGeo/PROJ/pull/2582
+    // This may fail before PROJ 8.0.1 if the m_pj object is a "meta"
+    // operation being a set of real operations
+    bool bCloneDone = ((m_pj == nullptr) == (poNewCT->m_pj == nullptr));
```
rouault, line 2456: this check is wrong. You need to go through the Initialize() road if poNewCT->m_pj == nullptr ad m_pj != nullptr. And I don't think you don't need to iterate over m_oTransformations

rouault:
We'd need some testing of this. Likely in autotest/cpp/test_osr_ct.cpp

gdal/ogr/ogrct.cpp
```diff
@@ -586,8 +582,11 @@ class OGRProjCT : public OGRCoordinateTransformation
             minx(minxIn), miny(minyIn), maxx(maxxIn), maxy(maxyIn),
             pj(pjIn), osName(osNameIn), osProjString(osProjStringIn),
             accuracy(accuracyIn) {}
-
-        Transformation(const Transformation&) = delete;
+        Transformation(const Transformation& other):
+            minx(other.minx), miny(other.miny), maxx(other.maxx), maxy(other.maxy),
+            pj((other.pj != nullptr) ? (proj_clone(OSRGetProjTLSContext(), other.pj)) : (nullptr)),
```
abellgithub, line 587: Perhaps a todo for PROJ to return NULL when you clone a NULL?  Or even better, have a copy ctor? and then this can all go away?

gdal/ogr/ogrct.cpp
```diff
@@ -917,6 +911,27 @@ OCTNewCoordinateTransformationEx(
             hOptions ? *hOptions : defaultOptions));
 }
 
+/************************************************************************/
+/*                   OCTNewCoordinateTransformationEx()                 */
```
abellgithub, line 915: Wrong title.

gdal/ogr/ogrct.cpp
abellgithub: This almost argues for wrapping the PJ so that you can hide the proj badness there and then this can all vanish.

gdal/ogr/ogrct.cpp
drons: Looks like I misunderstood your note "proj_clone might fail in some circumstances for PROJ < 8.0.1" from [here](https://github.com/OSGeo/gdal/pull/3809#discussion_r627754095).

My logic was based on these statements:

* `proj_clone` return null on failure
* If `this` `OGRProjCT` object contains non-null `m_pj` value and `poNewCT->m_pj` is non-null object I'm expect that `poNewCT->m_pj` is valid and nothing to do. 
* If `this` `OGRProjCT` object contains null `m_pj` value and `poNewCT->m_pj` is null object I'm expect that null is valid for such case and nothing to do. 
* Other "special" values (bWebMercatorToWGS84LongLat, bNoTransform, etc) are copied at constructor and we don't need to recompute it unlike `GetInverse`.

gdal/ogr/ogrct.cpp
```diff
@@ -2399,6 +2442,32 @@ int OGRProjCT::TransformWithErrorCodes(
     return TRUE;
 }
 
+/************************************************************************/
+/*                               Clone()                                */
+/************************************************************************/
+
+OGRCoordinateTransformation* OGRProjCT::Clone() const
+{
```
abellgithub, line 2450: Suggest unique_ptr:

```
{
  std::unique_ptr<OGRProjCT> newCT(new OGRProjCT(*this));
  #ifdef ...
  if (!m_pj ...)
  {
    newCT.reset(new OGRProjCT());
    if (!newCt->Initialize(...))
      newCt.reset();
  }
  #endif
  return newCT.release();
}
```

gdal/ogr/ogrct.cpp
rouault: you're right. I was too quick on saying that " ((m_pj == nullptr) == (poNewCT->m_pj == nullptr))" was wrong. Was just confused by the equality of equalities :-). The new code looks right AFAICS

gdal/ogr/ogrct.cpp
drons: `proj_clone` return `nullptr` for null object (see https://github.com/OSGeo/PROJ/blob/master/src/iso19111/c_api.cpp#L510), but set error. I'm think it is not needed in this case since null `other.pj` can be valid value.

abellgithub: My point was that if the pj object was simply wrapped, you could do whatever you wanted in that wrapped structure's copy ctor and all of these copy ctors can just be made default. The only reason that you need this copy ctor is because you have no good copy ctor for PJ.

drons: Take a [look](https://github.com/OSGeo/gdal/pull/3809/commits/82b56872fa0c70d7b167efbef91a1fed8cfd20bb).
Not sure if it got better.

abellgithub: Maybe I missed that you have to also deal with the OGR SRS src/target on line 977?
I was hoping it would be smaller. But it's contained and perhaps usable in GDAL projections, too? Maybe someone else has an opinion.

drons:
Done for now. But have doubts about [OGRCT: Wrap PJ* into PjPtr](https://github.com/OSGeo/gdal/pull/3809/commits/19f237dcd36e196269179dc7881c2b4bf6b35603).

@rouault Take a look, please.

PR #3810, opened 2021-05-06
Add basic support for coordinate epoch of dynamic (non-plate fixed) CRS

rouault:
Implementation of RFC 81: https://github.com/OSGeo/gdal/pull/3827

Use of a dynamic CRS without coordinate epoch doesn't make sense. So:
- add a coordinate epoch attribute to a OGRSpatialReference (even if technically this doesn't belong there, but this will be the most practical)
- add a OGRSpatialReference::IsDynamic() method to figure if a CRS is dynamic
-  Coordinate transformation: take into account SRS coordinate epoch if the coordinate tuple has no time value
- GPKG vector/raster: store/retrieve coordinate epoch as the value of the ``epoch`` column of the ``gpkg_spatial_ref_sys`` table (see https://github.com/opengeospatial/geopackage/issues/599 and https://github.com/opengeospatial/geopackage/pull/600)
- GeoJSON: add a coordinate_epoch member at FeatureCollection level
- GML: read/write coordinate epoch from `` <!-- coordinateEpoch=epoch -->`` XML comment in .xsd
- KML/LIBKML: read/write support from a  XML comment at beginning of file
- Shapefile: read/write coordinate epoch from/into a .wkt2 file using COORDINATEMETADATA[crs,EPOCH[epoch]] construct
- FlatGeoBuf: read/write using WKT:2019 COORDINATEMETADATA[crs,EPOCH[epoch]] construct
- GTiff/COG: add support for coordinate epoch with a new CoordinateEpochGeoKey=5120 GeoKey (https://github.com/opengeospatial/geotiff/issues/78)
- JPEG2000 drivers: read/write support for coordinate epoch in GeoJP2, GMLJP2 and GMLJP2v2
- MEM raster: support coordinate epoch
- VRT raster and PAM .aux.xml: support coordinate epoch as coordinateEpoch attribute of SRS element
- ogrinfo: report coordinate epoch when present
- ogr2ogr: add a -a_coord_epoch {epoch} switch
- gdalinfo: report coordinate epoch when present
- gdal_translate: add a -a_coord_epoch {epoch} switch

Documented in https://github.com/rouault/gdal/blob/coordinate_epoch/gdal/doc/source/user/coordinate_epoch.rst

Discussion for PostGIS started at https://trac.osgeo.org/postgis/ticket/4911


jratike80:
Is there some default epoch if it is not explicitly set? What to do and where to insert epoch in GeoJSON that it fundamentally EPSG:4326?

rouault:
> Is there some default epoch if it is not explicitly set?

No default, only the (advanced geodesist grade) user can know what the coordinate epoch is.

>  where to insert epoch in GeoJSON that it fundamentally EPSG:4326?

Good question. Could be a "coordinate_epoch" member at the FeatureCollection level.

rouault:
> > where to insert epoch in GeoJSON that it fundamentally EPSG:4326?
> 
> Good question. Could be a "coordinate_epoch" member at the FeatureCollection level.

@sgillies Opinion ?

nyalldawson:
> This is done by adding an extra coordinate_epoch column to our special gpkg_ogr_contents table

Why not place this in gpkg_metadata? As far as my understanding of the metadata extension this would be correct use of that table

rouault:
> Why not place this in gpkg_metadata? As far as my understanding of the metadata extension this would be correct use of that table

good point, that will be clearer

nyalldawson:
Reading over the changes -- this doesn't actually affect transformations at all yet, does it?

rouault:
> Reading over the changes -- this doesn't actually affect transformations at all yet, does it?

no indeed, this is preliminary work. We need datasets to be able to capture the information first. What we do with it is less urgent

nyalldawson:
Great, it's a good step forward!!

gdal/ogr/ogrsf_frmts/gpkg/ogrgeopackagetablelayer.cpp
```diff
@@ -933,6 +933,39 @@ OGRErr OGRGeoPackageTableLayer::ReadTableDefinition()
...
+                                m_pszTableName);
+
+                            SQLResult oResult;
+                            err = SQLQuery(m_poDS->GetDB(), pszSQL, &oResult);
+                            sqlite3_free(pszSQL);
+                            if ( err == OGRERR_NONE && oResult.nRowCount == 1 )
+                            {
+                                const char *pszMetadata = SQLResultGetValue(&oResult, 0, 0);
+                                if( pszMetadata &&
+                                    STARTS_WITH_CI(pszMetadata, "coordinate_epoch=") )
```
nyalldawson, line 956: Is this required? It should be already covered by the SQL `md.metadata LIKE 'coordinate_epoch=%%'` clause already

rouault: I'm kind of paranoid, but yes that isn't needed unless sqlite3 has bugs. Changing it into an assertion

nyalldawson: > unless sqlite3 has bugs

they don't have any, that's why they don't need a bug tracker :upside_down_face:

rouault:
> Reading over the changes -- this doesn't actually affect transformations at all yet, does it?

it does now with https://github.com/OSGeo/gdal/pull/3810/commits/9212bac30ce3aa9af8e45b682e07226316bfe680

rouault:
CC @bjornharrtell for the 3 FlatGeobuf commits

bjornharrtell:
I'm ambivalent on if it's too open minded to just have a wkt string in flageobuf or if the version should be explicit. I prefer explicit but fail to understand which actual versions of wkt exists which is really the only reason I did not make it explicit in the first place. To make version of specification (and other things) anything else than *very* clear is beyond me, it should be an obvious priority to make sure of before anything else...

rouault:
> I'm ambivalent on if it's too open minded to just have a wkt string in flageobuf or if the version should be explicit.

If that was your concern, WKT_2018 is a deprecated alias to WKT_2019. They refer to the same spec.

bjornharrtell:
> > I'm ambivalent on if it's too open minded to just have a wkt string in flageobuf or if the version should be explicit.
> 
> If that was your concern, WKT2_2018 is a deprecated alias to WKT2_2019. They refer to the same spec.

That in it self is not in any way reducing my confusion about projection WKT versions. (i admit I can only appreciate extremely simple and extract definitions)

rouault:
> confusion about projection WKT versions

https://proj.org/development/reference/cpp/cpp_general.html#general_doc_1standards should be of help.

gdal/ogr/ogrct.cpp
```diff
@@ -2061,6 +2073,14 @@ int OGRProjCT::TransformWithErrorCodes(
 /*      Select dynamically the best transformation for the data, if     */
 /*      needed.                                                         */
 /* -------------------------------------------------------------------- */
+    const double dfDefaultTime =
```
nyalldawson, line 2076: I don't fully understand this logic -- why do we take the destination epoch and assign it to the input points in some cases?

rouault: yes, that's a bit tricky, but PROJ can currently only do staticCRS -> dynamicCRS or dynamicCRS -> staticCRS transformations, and in either case, the coordinate epoch of the dynamicCRS must be provided as the input time. Adding that as a comment in the code

nyalldawson: Ah right -- so ultimately for proper support proj would need support for specific epoch -> specific epoch transforms then? I'm wondering if a user facing warning shouldn't be shown if both source and target have epochs set, since the results would be misleading...

rouault: > proj would need support for specific epoch -> specific epoch transforms then?

yes, that would also require to have the proj.db / EPSG to incorporate such transforms. The only ones currently are velocity grids for the Canada NAD83(CSRS)v7.  Plate motion by themselves are not really incorporated. If you want to do ATRF2014 @epoch1 to ATRF2014 @epoch2, you can emulate that by going:
ATRF2014 @epoch1 -> GDA2020 and GDA2020 -> ATRF2014 @epoch2 . Basically this is equivalent of using the Helmert transformation of ATRF2014 <--> GDA2020 with a delta_time = epoch2 - epoch1

> I'm wondering if a user facing warning shouldn't be shown if both source and target have epochs set,

will add

gdal/doc/source/user/coordinate_epoch.rst
```diff
@@ -0,0 +1,291 @@
...
+
+
+Support in utilities
+--------------------
+
+:program:`gdalinfo` and :program:`ogrinfo` report the coordinate epoch, when
+attached to a dataset/layer SRS.
+
+:program:`gdal_translate` and :program:`ogr2ogr` have a ``-a_coord_epoch`` option to be used
+toger with ``-a_srs``, and otherwise preserve the coordinate epoch in the output SRS
```
nyalldawson, line 287: "together"

gdal/doc/source/user/coordinate_epoch.rst
```diff
@@ -0,0 +1,291 @@
...
+        <!-- coordinateEpoch=2021.3 -->
+
+
+JPEG2000
+++++++++
+
+GeoJP2
+******
+
+GeoJP2 boxes use the above mentionned GeoTIFF encoding.
```
nyalldawson, line 158: "mentioned"

gdal/frmts/ecw/ecwdataset.cpp
```diff
@@ -3655,6 +3658,7 @@ void GDALRegister_JP2ECW()
     poDriver->SetMetadataItem( GDAL_DMD_HELPTOPIC, "drivers/raster/jp2ecw.html" );
     poDriver->SetMetadataItem( GDAL_DMD_EXTENSION, "jp2" );
     poDriver->SetMetadataItem( GDAL_DCAP_VIRTUALIO, "YES" );
+    poDriver->SetMetadataItem( GDAL_DCAP_COORDINATE_EPOCH, "YES" );
```
nyalldawson, line 3661: ECW isn't mentioned in coordinate_epoch.rst

gdal/frmts/mrsid/mrsiddataset.cpp
```diff
@@ -3610,6 +3610,8 @@ void GDALRegister_MrSID()
         /* for JP2CreateCopy() */
     poDriver->SetMetadataItem( GDAL_DCAP_VIRTUALIO, "YES" );
 #endif
+    poDriver->SetMetadataItem( GDAL_DCAP_COORDINATE_EPOCH, "YES" );
```
nyalldawson, line 3613: SID also is missing from coordinate_epoch.rst

nyalldawson:
Looking great to me :+1: !

gdal/frmts/mrsid/mrsiddataset.cpp
jratike80: Also, this deals with JP2MrSID, not SID.

gdt:
It doesn't really bear on this patch, but I don't follow the discussion about "what is the default if there is no epoch".    AIUI, Each dynamic CRS has a reference epoch and coordinates in that CRS presented without an epoch should be presumed to be of that epoch.   Yes, most people don't know what they are, but I think proj does and does the right thing w.r.t. this definition.  But this patch doesn't need to define semantics for formats it doesn't specify...

rouault:
> AIUI, Each dynamic CRS has a reference epoch and coordinates in that CRS presented without an epoch should be presumed to be of that epoch

Nothing in this pull request or PROJ does the above assumption. The frame reference epoch of a dynamic CRS/datum is completely ignored, except to expose it as metadata. Let's consider ITRF2014. Its frame reference epoch is 2010.0. But if you take the ITRF2014 <--> GDA2020 time dependent Helmert transformation (modelling the Australian plate drift), its central epoch is 2020.0. And the coordinate epoch of the coordinate to transform could be yet another date...

> But this patch doesn't need to define semantics for formats it doesn't specify...

Not completely sure to understand the above sentence. This patch mainly adds (a for-now GDAL specific) way of encoding the coordinate epoch in formats that don't have a standardized way of storing it.

gdt:
> Nothing in this pull request or PROJ does the above assumption. The frame reference epoch of a dynamic CRS/datum is completely ignored, except to expose it as metadata. Let's consider ITRF2014. Its frame reference epoch is 2010.0. But if you take the ITRF2014 <--> GDA2020 time dependent Helmert transformation (modelling the Australian plate drift), its central epoch is 2020.0. And the coordinate epoch of the coordinate to transform could be yet another date...

That is surprising to me - but me being confused about that is out of scope.
 
> > But this patch doesn't need to define semantics for formats it doesn't specify...
> 
> Not completely sure to understand the above sentence. This patch mainly adds (a for-now GDAL specific) way of encoding the coordinate epoch in formats that don't have a standardized way of storing it.

What I meant is that this patch is adding a way to have an epoch associated with a dCRS, so it needs to define the semantics of what a (dCRS, epoch)  pair means, and I think it does.    The existence of that definition then raises the issue of "What is the meaning of a dCRS without an associated epoch?"  That definition should already exist, and maybe it does, but it's not really the place of this pull request to establish it.

(Above, I think you're saying there are no clear semantics and various transforms treat dCRS values as being in different epochs, and not necessarily the associated reference epoch.)

rouault:
>  The existence of that definition then raises the issue of "What is the meaning of a dCRS without an associated epoch?" That definition should already exist, and maybe it does, but it's not really the place of this pull request to establish it.

Coordinates expressed in a dCRS without a coordinate epoch are under-specified. For most transformations involving that dCRS, which generally have time-dependent terms, this means, in the way PROJ currently implements that, those time-dependent terms are ignored (or equivalently, the coordinate epoch is assumed to be the central epoch of the transformation). But this is mostly an implementation detail (if we were very strict, we could potentially error out)

tobwen:
So datasets can have one epoch only, even for multi-temporal data? I once worked with long-term data series that had a date of their acquisition per row. So I would have to either break them up into individual layers or transform them all into a "stable" or consented epoch. Can GDAL get a function to split records those records?

rouault:
> So datasets can have one epoch only, even for multi-temporal data? I once worked with long-term data series that had a date of their acquisition per row

the date of acquisition is not necessarily the coordinate epoch. Some reprocessing against a 'reference' epoch could potentially have occurred.

> So I would have to either break them up into individual layers or transform them all into a "stable" or consented epoch.

Yes, the scope of this work is limited to one coordinate epoch per layer.

> Can GDAL get a function to split records those records?

If you have a distinct coordinate epoch per row, this wouldn't be practical. Otherwise with some scripting you could do that.

rouault:
Update for GeoPackage: GPKG vector/raster: store/retrieve coordinate epoch as the value of the ``epoch`` column of the ``gpkg_spatial_ref_sys`` table (see https://github.com/opengeospatial/geopackage/issues/599 and https://github.com/opengeospatial/geopackage/pull/600)

jratike80:
FYI, a new OGC Features and Geometries JSON SWG https://www.ogc.org/projects/groups/featgeojsonswg starts soon. One of the aims is to add crs support into some GeoJSON-alike and JSON and why not support for epochs as well.

rouault:
Update for GeoTIFF: during the OGC GeoTIFF WG meeting on June 16th, the proposal to add a CoordinateEpochGeoKey was accepted to be in the scope of a future GeoTIFF 1.2 spec. It will likely take months to materialize due to the usual slow pace of such processes, but I think we can use it.

rouault:
Update for GeoPackage: https://github.com/opengeospatial/geopackage/pull/600#issuecomment-863542594 was approved and merged in the master specfication

rouault:
Superseded by https://github.com/OSGeo/gdal/pull/4011

PR #3812, opened 2021-05-07
GPKG: accept CreateField with a column with the fid column name and type=real, width=20, precision=0

rouault:
See https://github.com/qgis/QGIS/issues/25795
and https://github.com/qgis/QGIS/pull/43118


nirvn:
@rouault , for the record, arcmap will transform fid field to a double when saving as shapefile too.

PR #3825, opened 2021-05-09
Csharp : Add GetFieldAsBinary methods  - fixes #3041

### LINKED ISSUE #3041: OGR.Feature.GetFieldAsBinary is unavailable in C# bindings
I'm expecting to use the OGR library to read binary data from a field in the FileGDB via the OGR.Feature.GetFieldAsBinary method. 

I am attempting to do this via the C# bindings provided for GDAL 3.1.3 (released 2020/09/01) on Windows 10 obtained from www.gisinternals.com.

I am unfamiliar with the GDAL source and C/SWIG in general, but would like to understand if this function can be easily enabled. It appears to get excluded somewhere around here I think:
https://github.com/OSGeo/gdal/blob/b51dc28ab7da7c84c1cac3477560dbc9ba9f5c9e/gdal/swig/include/ogr.i#L1567

runette:
## What does this PR do?

fixes #3041

- Opens up to C# the SWIG method `OGRErr GetFieldAsBinary( int id, int *nLen, char **pBuf)` on `Feature`
- Adds the overload `public byte[] GetFieldAsBinary(int id)`
- Add tests to `createdata.cs` to test the new method.

NOTE - this is the first time I have created a PR based on SWIG - so probably @szekerest or @bjornharrtell should take a close look. However, @srweal has been using the prototype of this change in a real environment with success.

This is related to #3041 

## Tasklist

 - [x] ADD YOUR TASKS HERE
 - [x] Add test case(s)
 - [x] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:

### DIFF
--- a/gdal/swig/include/ogr.i
+++ b/gdal/swig/include/ogr.i
@@ -1626,7 +1626,6 @@ public:
 #endif
 #endif
 
-#ifndef SWIGCSHARP
 #ifdef SWIGJAVA
 %apply (GByte* outBytes) {GByte*};
   GByte* GetFieldAsBinary(int id, int *nLen, char **pBuf) {
@@ -1652,6 +1651,15 @@ public:
       }
   }
 %clear GByte*;
+#elif defined(SWIGCSHARP)
+%apply (void *buffer_ptr) {char **};
+  OGRErr GetFieldAsBinary( int id, int *nLen, char **pBuf) {
+    GByte* pabyBlob = OGR_F_GetFieldAsBinary(self, id, nLen);
+    *pBuf = (char*)malloc(*nLen);
+    memcpy(*pBuf, pabyBlob, *nLen);
+    return OGRERR_NONE;
+  }
+%clear (char **);
 #else
   OGRErr GetFieldAsBinary( int id, int *nLen, char **pBuf) {
     GByte* pabyBlob = OGR_F_GetFieldAsBinary(self, id, nLen);
@@ -1679,7 +1687,6 @@ public:
 #endif
 #endif /* SWIGJAVA */
 
-#endif /* SWIGCSHARP */
 
   /* ---- IsFieldSet --------------------------- */
   bool IsFieldSet(int id) {
--- a/gdal/swig/csharp/apps/createdata.cs
+++ b/gdal/swig/csharp/apps/createdata.cs
@@ -249,7 +249,8 @@ class CreateData {
 			if (feat.IsFieldSet(iField))
 			{
 				Console.WriteLine(feat.GetFieldAsString(iField));
-				if (ft == FieldType.OFTString && feat.GetFieldAsBinary( iField ).Length <= 0)
+				if (ft == FieldType.OFTString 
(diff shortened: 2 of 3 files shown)


gdal/swig/include/ogr.i
```diff
@@ -1652,6 +1651,15 @@ public:
       }
   }
 %clear GByte*;
+#elif defined(SWIGCSHARP)
+%apply (void *buffer_ptr) {char **};
+  OGRErr GetFieldAsBinary( int id, int *nLen, char **pBuf) {
+    GByte* pabyBlob = OGR_F_GetFieldAsBinary(self, id, nLen);
+    *pBuf = (char*)malloc(*nLen);
```
rouault, line 1658: I believe this will leak.  You'd likely need something similar to the ``%typemap(freearg) (int *nLen, char **pBuf )`` of swig/include/java/typemaps_java.i to free it

And I guess you could probably avoid the manual public byte[] GetFieldAsBinary(int id) override by having an output typemap that would return a byte[], but I'm not familiar with C# swig typemaps to provide more guidance

runette: Ok - back to the drawing board. I will look at that

srweal: Apologies, I've missed following up this conversation since having success with the code @runette provided. I don't have the technical skills to fix this myself, but it seems close to being complete, apart from the 'suspected' memory leak issue. Is that something a more experienced OSGeo contributor could quickly fix up to get this included - possibly @szekerest or @bjornharrtell?

srweal:
@runette, were you able to progress this after the 'memory leak' comments? Would be good to reopen and get it across the line, as I think your initial contribution is not far off being good to go. It certainly works well for me in the current form.

runette:
Sorry. Distracted by other work. I will get to it soon

abspro:
Any update?

runette:
Not yet - I need to sit down sometime and focus on fixing the problem.

spietraszek:
Has anything moved forward?

runette:
To be honest - I had totally forgotten about that :(

runette:
Unfortunately - I am sailing for the next couple of weeks so cannot do
anything but one does feel this should probably be knocked out. I will try
when I get back

spietraszek:
I encountered an issue where C#'s `GetFieldAsString` failed to properly decode Hebrew text (Windows-1255 encoding). Instead of the correct text, I was getting a string full of replacement characters (�).

I initially considered using `GetFieldAsBinary` as a workaround to decode the text manually. However, I found a different solution by creating a C++/CLI bridge project that uses GDAL directly:

```cpp
// filepath: GDALOps.h
namespace GDALOps
{
    public ref class Ops
    {
    public:
        static String^ ProcessFeatureFromManaged(IntPtr featurePtr, int i)
        {
            OGRFeature* poFeature = static_cast<OGRFeature*>(featurePtr.ToPointer());
            auto result = poFeature->GetFieldAsString(i);
            return gcnew String(result);
        }
    };
}
```

Interestingly, this approach works perfectly for handling the Hebrew text encoding.

PR #3829, opened 2021-05-10
RFC 82

abellgithub:
## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:

### DIFF
--- a/gdal/doc/source/development/rfc/rfc82_memory_recommendations.rst
+++ b/gdal/doc/source/development/rfc/rfc82_memory_recommendations.rst
@@ -32,7 +32,8 @@ When GDAL came into being the C++ standard library either didn't exist or wasn't
 by many vendors. In addition, compilers were less sophisticated and tools to deal with
 memory issues were costly and difficult to use.
 
-Today, all relevant C++ compilers support the C++ standard through C++11 fully. Most support
+Today, all relevant C++ compilers support the C++ standard through C++11 fully and GDAL
+uses C++11 features (see RFC 68) . Most support
 more recent standards as well. This allows the standard libraries to be used where
 previously C-style memory allocation was the norm. This allows GDAL code to be written with
 using a RAII model rather than a malloc/free (new/delete) model in most cases. Because RAII
@@ -60,6 +61,10 @@ acceptable memory handling for changes to existing code and new additions.
 The point is to specify how memory management
 and handling in GDAL should be done if it were to be have been done from scratch in 2021.
 
+Note that some discussion surrounding these issues has taken place in the past. For context
+see this mailing list history, some of which is out of date:
+https://lists.osgeo.org/pipermail/gdal-dev/2016-May/044298.html
+
 Recommendations
 ---------------
 
@@ -77,7 +82,9 @@ Questions
       ownership requirements that are difficult.  Should the API be maintained in
       perpetuity? Should alternative versions of functions be created that help eliminate
       memory errors for users and simplify handling of data structures internally?
-    - Does acceptance of this RFC eliminate the need for RFC 19?
+    - Does acceptance of this RFC supercede RFC 19?
+
+
 
 Voting History
 --------------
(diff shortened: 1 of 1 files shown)


gdal/doc/source/development/rfc/rfc82_memory_recommendations.rst
```diff
@@ -0,0 +1,85 @@
...
+and handling in GDAL should be done if it were to be have been done from scratch in 2021.
+
+Recommendations
+---------------
+
+    - Use the C++ standard library where possible for buffer creation and general
+      data handling.
+    - Where practical, use self-created RAII types where the C++ standard library is
+      insufficient.
+    - Allow memory errors to be thrown as std::bad_alloc exceptions.
```
rouault, line 70: This would need to be more precisely defined. Currently users (either internal to GDAL, or external) of the GDAL C++ API will never receive a C++ exception, consequently their code might not be ready for it. In some coding styles, C++ exceptions are banned (https://google.github.io/styleguide/cppguide.html#Exceptions)
Obviously all GDAL C API should catch any C++ exceptions.
Basically allowing C++ exceptions requires a careful inspection of all code paths to where it can be propagated to make sure they use themselves RAII to avoid memory leaks or other type of unfreed resources (file pointers)

schwehr:
@abellgithub Thanks for working on this!  

It would be good to have a section(s) that cover alternatives and the history of the topic.  I'd like to at least have a reference to stack size as it is an issue for me.  That's why I wrote [this note](https://lists.osgeo.org/pipermail/gdal-dev/2016-May/044298.html) that introduced [Use vector<T>(length, initial_value) for local blocks of storage](https://goo.gl/vuA3D6).  It generated a fair bit of discussion then.  There are plenty of ways to do this without vector with modern C++.

Please also address the C++ language version issue covered in [RFC 68](https://gdal.org/development/rfc/rfc68_cplusplus11.html), even if it's as simple as stating "no change"

gdal/doc/source/development/rfc/rfc82_memory_recommendations.rst
abellgithub: I would normally agree, but this is for all practical purposes a fatal exception WRT any GDAL processing, which isn't much different from what can happen now, which is that `abort()` is called.

rouault: > which isn't much different from what can happen now, which is that `abort()` is called.

Memory allocation failures in GDAL, at least the larger ones normally done with VSIMalloc()/VSICalloc() (not talking about the small ones done by CPLMalloc()/CPLCalloc()) fail gracefully.

abellgithub: I'm sorry but I don't understand. What does that mean to "fail gracefully"?

rouault: > What does that mean to "fail gracefully"?

VSIxxxxx() functions related to memory allocations return a nullptr in case of failure, and the caller can exit properly, doing necessary cleanups. Like in https://github.com/OSGeo/gdal/blob/da5ad8ed4020daac5cd784ab27c5acaa899e3d90/gdal/alg/gdalwarpkernel.cpp#L5832

abellgithub:
@schwehr : I've never done one of these before and don't really know the process. Would you like to add the history and some reference to your stack concerns? I don't know the best way to do this.

abellgithub:
I guess I'm just not sure how your concerns about stack size should be folded into this. If you want to send me a paragraph or whatever I can add it.

schwehr:
> I guess I'm just not sure how your concerns about stack size should be folded into this. If you want to send me a paragraph or whatever I can add it.

How about just having a "See also" with "prior discussions on memory allocation"?  The stack size part isn't that interesting.  It's general discussion about how to handle memory allocation in GDAL.

PR #3830, opened 2021-05-10
Fix mapping of mapinfo symbol numbers to corresponding OGR symbol IDs

### LINKED ISSUE #3826: Incorrect mapping of mapinfo point symbols to ogr-sym ids
The current mapping of mapinfo marker symbol number to ogr style id used at https://github.com/OSGeo/gdal/blob/master/gdal/ogr/ogrsf_frmts/mitab/mitab_feature.cpp#L8925 appears incorrect.

The screenshot below shows a rendering in MapInfo (10.5) with the labels showing the "field1" values:

image

Querying the tab gives this output:

```
OGRFeature(marker_styles):2
  Field1 (String) = 2
  Style = SYMBOL(a:0,c:#5050ff,s:36pt,id:"mapinfo-sym-32,ogr-sym-6")
  POINT (-1.337792 0.688963)

OGRFeature(marker_styles):3
  Field1 (String) = 3
  Style = SYMBOL(a:45,c:#ff5820,s:36pt,id:"mapinfo-sym-33,ogr-sym-6")
  POINT (-1.13378 0.698997)

...
````

So feature 2 (with mapinfo symbol number 32), which is shown in mapinfo as a square is being mapped to "ogr-sym-6: triangle (not filled)". The rest of the mappings show similar errors, and the results are all over the place.

Here's how the layer renders in QGIS when using the ogr-sym identifiers and rendering as per the illustration on https://gdal.org/user/ogr_feature_style.html:

image

I'm happy to fix this mapping, but I'm unsure if there's any historical context I'm missing here or if the mapping has been incorrect ever since it was first introduced in 2000...

nyalldawson:
Fixes #3826


abellgithub:
This seems to beg for some sort of data-based mapping. Like an array of pairs or a map.

nyalldawson:
@abellgithub 
> This seems to beg for some sort of data-based mapping. Like an array of pairs or a map.

I'm on the fence with this. I can see how it would be useful for readability, but on the other hand this is potentially performance-critical code (and is unlikely to be touched again for the next 10+ years!). So I'd lean toward the current approach...

PR #3834, opened 2021-05-11
`gdallocationinfo` fixes + more tests

idanmiara:
## What does this PR do?

* gdal/swig/python/gdal-utils/osgeo_utils/auxiliary/util.py - add `ovr_only: bool = False` parameter to `_open_ds`
* gdal/swig/python/gdal-utils/osgeo_utils/auxiliary/osr_util.py - change `transform_points()` to return None as it makes an inline replacement
* gdal/swig/python/gdal-utils/osgeo_utils/samples/gdallocationinfo.py - fix inline_xy_replacement; fix unscaling; fix formatting; use faster `BandRasterIONumPy` instead of `band.ReadAsArray`; add optional parameters `return_ovr_pixel_line`, `transform_round_digits`
* autotest/pyscripts/test_gdallocationinfo_py.py - add another test

### DIFF
--- a/gdal/swig/python/gdal-utils/osgeo_utils/auxiliary/util.py
+++ b/gdal/swig/python/gdal-utils/osgeo_utils/auxiliary/util.py
@@ -32,8 +32,9 @@
 import os
 from numbers import Real
 from typing import Optional, Union, Sequence, Tuple, Dict, Any, Iterator, List
+from warnings import warn
 
-from osgeo import gdal
+from osgeo import gdal, __version__ as gdal_version_str
 from osgeo_utils.auxiliary.base import get_extension, is_path_like, PathLikeOrStr, enum_to_str, OptionalBoolStr, \
     is_true, \
     MaybeSequence, T
@@ -41,6 +42,7 @@ from osgeo_utils.auxiliary.base import get_extension, is_path_like, PathLikeOrSt
 PathOrDS = Union[PathLikeOrStr, gdal.Dataset]
 DataTypeOrStr = Union[str, int]
 CreationOptions = Optional[Dict[str, Any]]
+gdal_version = tuple(int(s) for s in str(gdal_version_str).split('.') if s.isdigit())[:3]
 
 
 def DoesDriverHandleExtension(drv: gdal.Driver, ext: str) -> bool:
@@ -216,6 +218,7 @@ class OpenDS:
         filename: PathLikeOrStr,
         access_mode=gdal.GA_ReadOnly,
         ovr_idx: Optional[Union[int, float]] = None,
+        ovr_only: bool = False,
         open_options: Optional[Union[Dict[str, str], Sequence[str]]] = None,
         logger=None,
     ):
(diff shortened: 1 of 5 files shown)


gdal/swig/python/gdal-utils/osgeo_utils/auxiliary/util.py
```diff
@@ -226,8 +229,16 @@ def _open_ds(
...
-        if ovr_idx > 0:
-            open_options["OVERVIEW_LEVEL"] = ovr_idx - 1  # gdal overview 0 is the first overview (after the base layer)
+        # gdal overview 0 is the first overview (after the base layer)
+        if ovr_idx == 0:
+            if ovr_only:
+                if gdal_version >= (3, 3):
+                    open_options["OVERVIEW_LEVEL"] = 'NONE'
+                else:
+                    raise Exception('You asked to not expose overviews, Which is not supported in your gdal version, '
+                                    'please your gdal version to gdal >= 3.3 or do not ask to hide overviews')
```
rouault, line 239:
```suggestion
                                    'please update your gdal version to gdal >= 3.3 or do not ask to hide overviews')
```

gdal/swig/python/gdal-utils/osgeo_utils/auxiliary/util.py
```diff
@@ -361,3 +374,25 @@ def get_scales_and_offsets(bands: MaybeSequence[gdal.Band]) -> Tuple[bool, Maybe
         scales, offsets = scales[0], offsets[0]
     return is_scaled, scales, offsets
 
+
+def set_scales_and_offsets(bands: Union[PathOrDS, MaybeSequence[gdal.Band]],
```
rouault, line 378: this method is not used by this PR

rouault (commented):
is it material for 3.3 ?

gdal/swig/python/gdal-utils/osgeo_utils/auxiliary/util.py
idanmiara: Right, I think I found another bug in related to `gdallocationinfo` and `scale` but I wanted to dig deeper in a new PR.
Would you suggest I'll remove it for now?

idanmiara:
> is it material for 3.3 ?

I'm think so. I'm trying to hunt bugs before adding new functionality for 3.4.

gdal/swig/python/gdal-utils/osgeo_utils/auxiliary/util.py
rouault: > Would you suggest I'll remove it for now?

yes please

idanmiara:
I think there is a bug in the cpp version of `gdallocationinfo` with regards to overviews. 
Once this PR is approved I will port this test to the cpp version to compare the results.

idanmiara:
I've implemented a function that extracts the intermediate points between two points using `pyproj.geod.inv_intermediate` 
https://github.com/pyproj4/pyproj/pull/841
Then extract the location info i.e. DTM height, and combining the results.

Considering that it depends on `pyproj`, do you think it would be a good idea to put it as a sample in gdal_utils?

rouault:
> and combining the results.

what combining means exactly here ?

> Considering that it depends on pyproj, do you think it would be a good idea to put it as a sample in gdal_utils?

I guess that's OK, but the pyproj PR doesn't seem to be merged yet AFAICS

idanmiara:
> > and combining the results.
> 
> what combining means exactly here ?
> 

It returns a 2d numpy array of the result triplets lon,lat,alt. 

> > Considering that it depends on pyproj, do you think it would be a good idea to put it as a sample in gdal_utils?
> 
> I guess that's OK, but the pyproj PR doesn't seem to be merged yet AFAICS

Right. Soon :)

gdal/swig/python/gdal-utils/osgeo_utils/auxiliary/util.py
idanmiara: done

idanmiara:
Thanks @rouault !

PR #3838, opened 2021-05-14
util.py - `get_bands(ovr_idx)` to support also negative `ovr_idx` and resolution

idanmiara:
## What does this PR do?

Without this PR `get_bands` only accept non-negative `ovr_idx: int`.
This PR adds the option to supply a negative `ovr_idx` and also `ovr_res` (as in `get_ovr_idx`)
For instance, this allow the user to use also these inputs to `gdallocationinfo`.

## Related PRs

https://github.com/OSGeo/gdal/pull/3608

### DIFF
--- a/gdal/swig/python/gdal-utils/osgeo_utils/auxiliary/util.py
+++ b/gdal/swig/python/gdal-utils/osgeo_utils/auxiliary/util.py
@@ -157,6 +157,12 @@ def get_ovr_idx(filename_or_ds: PathOrDS,
                 ovr_idx: Optional[int] = None,
                 ovr_res: Optional[Union[int, float]] = None) -> int:
     """
+    This function uses a different convention than the GDAL API itself:
+    * ovr_idx = 0 means the full resolution image (GDAL API: no OVERVIEW_LEVEL)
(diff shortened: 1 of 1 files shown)


gdal/swig/python/gdal-utils/osgeo_utils/auxiliary/util.py
```diff
@@ -354,6 +354,7 @@ def get_bands(filename_or_ds: PathOrDS, band_nums: Optional[MaybeSequence[int]]
         if band is None:
             raise Exception(f'Could not get band {band_num} from file {filename_or_ds}')
         if ovr_idx:
+            ovr_idx = get_ovr_idx(ds, ovr_idx)
```
rouault, line 357: Reviewing what get_over_idx() does and its use just below, I realize it uses a different convention than the GDAL API itself, with 0 meaning the full resolution image, 1 the first overview level, etc. The doc of get_ovr_idx() should be enhanced to underline that.  And the doc of this method too.

idanmiara: Good point. thanks, done.

PR #3849, opened 2021-05-18
mention RASTERIO_EXTRA_ARG_CURRENT_VERSION

mmomtchev:
## What does this PR do?
 
A very small doxygen PR - no code affected

### DIFF
--- a/gdal/gcore/gdal.h
+++ b/gdal/gcore/gdal.h
@@ -153,8 +153,8 @@ typedef enum
 } GDALRIOResampleAlg;
 
 /* NOTE to developers: only add members, and if so edit INIT_RASTERIO_EXTRA_ARG */
-/* and INIT_RASTERIO_EXTRA_ARG */
-/** Structure to pass extra arguments to RasterIO() method
+/** Structure to pass extra arguments to RasterIO() method,
+ * must be initialized with INIT_RASTERIO_EXTRA_ARG
   * @since GDAL 2.0
   */
 typedef struct


gdal/gcore/gdal.h
```diff
@@ -159,7 +159,8 @@ typedef enum
   */
 typedef struct
 {
-    /*! Version of structure (to allow future extensions of the structure) */
+    /*! Version of structure (to allow future extensions of the structure),
```
rouault, line 162: I'm wondering if the doc of the structure itself, at line 157, shoudn't be rather changed to suggest using INIT_RASTERIO_EXTRA_ARG(&sExtraArg)

mmomtchev: I just found it independently (and the hard way) :smile:

mmomtchev: Does resampling when writing works? Or is it only when reading?

rouault: > Does resampling when writing works? Or is it only when reading?

only nearest resampling is supported on writing

mmomtchev: @rouault I just spent 2 hours trying to find the problem with resampling when writing... probably mention it too? :smile:

rouault: > probably mention it too?

Yes, this is very valuable input! Perhaps in the doc of GDALDataset and GDALRasterBand IRasterIO() method ?

mmomtchev: @rouault , maybe in the extra structure because this is where you look. I don't know what is supposed to work? Simple nearest neighbor oversampling appears to work? Or it is not intentional?

mmomtchev: No, in fact it is just image replication, since I had a pattern it was creating the illusion of oversampling

rouault: > No, in fact it is just image replication, since I had a pattern it was creating the illusion of oversampling

On the write side, nearest oversampling = image replication

PR #3851, opened 2021-05-18
gdal-utils/osgeo_utils/auxiliary/base.py - fix `num()`

idanmiara:
This PR fixes the `num` auxiliary function. 
Before this PR `num(f: float)` returned `int(f)`
i.e. num(42.5)` returned 42 (instead of 42.5)

Tests added for this case.


idanmiara:
Thanks @rouault

PR #3857, opened 2021-05-20
OGRCT: Export novel OGRCoordinateTransformation functions to C API

drons:

### DIFF
--- a/gdal/ogr/ogrct.cpp
+++ b/gdal/ogr/ogrct.cpp
@@ -947,7 +947,10 @@ OCTNewCoordinateTransformationEx(
  * Clone transformation object.
  *
  * This is the same as the C++ function OGRCreateCoordinateTransformation::Clone
-
+ *
+ * @return handle to transformation's clone or NULL on error, 
+ *         must be freed with OCTDestroyCoordinateTransformation
+ *
  * @since GDAL 3.4
  */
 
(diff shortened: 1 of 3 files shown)


gdal/ogr/ogrct.cpp
```diff
@@ -960,6 +960,66 @@ OCTClone(OGRCoordinateTransformationH hTransform)
...
 }
 
+/************************************************************************/
+/*                             OCTGetSourceCS()                         */
+/************************************************************************/
+
+/**
+ * Transformation's source coordinate system reference.
+ *
+ * This is the same as the C++ function OGRCreateCoordinateTransformation::GetSourceCS
```
rouault, line 970: It should be mentioned that the returned object might be null, and is owned by hTransform and should not be freed.

gdal/ogr/ogrct.cpp
```diff
@@ -960,6 +960,66 @@ OCTClone(OGRCoordinateTransformationH hTransform)
...
+}
+
+/************************************************************************/
+/*                             OCTGetTargetCS()                         */
+/************************************************************************/
+
+/**
+ * Transformation's target coordinate system reference.
+ *
+ * This is the same as the C++ function OGRCreateCoordinateTransformation::GetTargetCS
```
rouault, line 990: It should be mentioned that the returned object might be null, and is owned by hTransform and should not be freed.

gdal/ogr/ogrct.cpp
```diff
@@ -960,6 +960,66 @@ OCTClone(OGRCoordinateTransformationH hTransform)
...
+}
+
+/************************************************************************/
+/*                             OCTGetInverse()                          */
+/************************************************************************/
+
+/**
+ * Inverse transformation object.
+ *
+ * This is the same as the C++ function OGRCreateCoordinateTransformation::GetInverse
```
rouault, line 1010: It should be mentioned that the returned object might be null, and should be freed with OCTDestroyCoordinateTransformation()

rouault:
I'd like to see minimal testing of this as well

drons:
Thanks!

PR #3867, opened 2021-05-21
DatasetPool:  expose the current size in private C interface

andrew-hardin:
## What does this PR do?
This exposes the current size of the dataset pool as part of the private C interface. This enables a use case that estimates the current memory usage of a VRT given the number of open datasets in the pool.

It may not seem like much, but this lets me track ~500MiB of otherwise unaccountable heap allocations (for my specific use case, it's related to the zstd decompression context within an open TIFF dataset).

None.

## Tasklist

 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS: Ubuntu 20.04
* Compiler: GCC 9.3

### DIFF
--- a/gdal/gcore/gdal_priv.h
+++ b/gdal/gcore/gdal_priv.h
@@ -2809,6 +2809,7 @@ void GDALNullifyOpenDatasetsList();
 CPLMutex** GDALGetphDMMutex();
 CPLMutex** GDALGetphDLMutex();
 void GDALNullifyProxyPoolSingleton();
+int GDALDatasetPoolGetCurrentSize();
 void GDALSetResponsiblePIDForCurrentThread(GIntBig responsiblePID);
 GIntBig GDALGetResponsiblePIDForCurrentThread();
 
--- a/gdal/gcore/gdalproxypool.cpp
+++ b/gdal/gcore/gdalproxypool.cpp
@@ -139,6 +139,7 @@ class GDALDatasetPool
 
         static void PreventDestroy();
         static void ForceDestroy();
+        static int GetCurrentSize();
 };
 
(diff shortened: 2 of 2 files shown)


rouault:
doesn't GDALGetCacheUsed64() do the job ?

andrew-hardin:
Not entirely. `GDALGetCacheUsed64()` only provides the size of the block cache - it doesn't include per-dataset heap allocations.

Maybe it would help if described out my use case in more detail. I have a large VRT composed entirely of zstd compressed tiffs. Each opened tiff holds ~2MiB of heap  (traced back to `ZSTDState`). With `GDAL_MAX_DATASET_POOL_SIZE=250`, that's up to 500MiB of heap that's unaccounted for. Exposing `GDALDatasetPoolGetCurrentSize()` improves how accurately I can estimate the current memory used by GDAL.

rouault:
> Each opened tiff holds ~2MiB of heap (traced back to `ZSTDState`).

I see. I'd say we should aim to consume just 2 MiB for *all* datasets in the pool by reusing a single ZSTD_DStream* object, which should be possible when reading whole tiles. Will need changes in both GDAL and libtiff. Are you ZSTD files single band or with interleave=band  ? (otherwise you'd get also a buffer at the GTiffDataset level for each strip/tile)

andrew-hardin:
Below are the specs for each tile in the VRT. Some of it may look familiar - it's 1x1 degree raster terrain data (similar to SRTM/DTED).
```
datatype = int16
bands = 1
xsize/ysize = 3601
compress = ZSTD
zstd_level = 9
predictor = 2
```

Issue #3870, opened 2021-05-21
Support importing GeoJSON fragments in non-null terminated strings


mmomtchev:
Currently `OGRGeometryFactory::createFromGeoJson` supports importing only from null-terminated strings.

The underlying library call `CPLJSONDocument::LoadMemory` supports also an optional `length` parameter.

This could either be a new form of `createFromGeoJson` or an optional parameter.

mmomtchev:
Ok, I got an error because I was not casting `size_t` to `int` but nothing about `size_t` = -1 :smile:

PR #3873, opened 2021-05-21
createFromGeoJson: support non-null terminated #3870

mmomtchev:
## What does this PR do?

Add an optional size argument to `OGRGeometryFactory::createFromGeojson` for non-null terminated strings

## Tasklist

## Environment

* OS: Ubuntu 20.04
* Compiler: gcc 9

### DIFF
--- a/gdal/ogr/ogr_geometry.h
+++ b/gdal/ogr/ogr_geometry.h
@@ -3253,7 +3253,7 @@ class CPL_DLL OGRGeometryFactory
     static OGRGeometry *createFromGML( const char * );
     static OGRGeometry *createFromGEOS( GEOSContextHandle_t hGEOSCtxt,
                                         GEOSGeom );
-    static OGRGeometry *createFromGeoJson( const char *);
+    static OGRGeometry *createFromGeoJson( const char *, int = -1 );
     static OGRGeometry *createFromGeoJson( const CPLJSONObject &oJSONObject );
 
     static void   destroyGeometry( OGRGeometry * );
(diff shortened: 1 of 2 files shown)


gdal/ogr/ogr_geometry.h
```diff
@@ -3253,7 +3253,7 @@ class CPL_DLL OGRGeometryFactory
     static OGRGeometry *createFromGML( const char * );
     static OGRGeometry *createFromGEOS( GEOSContextHandle_t hGEOSCtxt,
                                         GEOSGeom );
-    static OGRGeometry *createFromGeoJson( const char *);
+    static OGRGeometry *createFromGeoJson( const char *, size_t = -1 );
```
rouault, line 3256: Should rather be a int since that's what json_tokener_parse_ex() ultimately accepts

gdal/ogr/ogrgeometryfactory.cpp
```diff
@@ -6087,13 +6087,17 @@ OGRCurve* OGRGeometryFactory::curveFromLineString(
...
  * @return a geometry on success, or NULL on error.
  * @since GDAL 2.3
  */
-OGRGeometry* OGRGeometryFactory::createFromGeoJson( const char *pszJsonString )
+OGRGeometry* OGRGeometryFactory::createFromGeoJson(
+    const char *pszJsonString, size_t nSize )
 {
     CPLJSONDocument oDocument;
-    if( !oDocument.LoadMemory( reinterpret_cast<const GByte*>(pszJsonString)) )
+    if( !oDocument.LoadMemory( reinterpret_cast<const GByte*>(pszJsonString),
```
rouault, line 6099: this is a bit paranoiac, but in case someone on 64-bit would provide a > 2 GB buffer, add a ``(nSize == static_cast<size_t>(-1) || nSize < static_cast<size_t>(std::numeric_limits<int>::max()))`` check

PR #3881, opened 2021-05-25
Lint: make SQLResult as C++ class that has a destructor

### LINKED ISSUE #3811: Lint: make SQLResult as C++ class that has a destructor
The SQLResult struct in ogr/ogrsf_frmts/sqlite/ogrsqliteutility.h/.cpp should rather be a C++ class that has a destructor that frees memory instead of having to explicitly call SQLResultFree()

julienr:
By having a destructor, this allows removing the explicit calls to SQLResultFree

Closes #3811

## What does this PR do?

This addresses the lint improvement of #3811 by turning SQLResultFree from a `struct` with an explicit `SQLResultFree` function to a `class` with a proper destructor.

Note that there are a bunch more free-standing accessor functions (e.g. `SQLResultGetValue`) that could be turned into getters and setters. Happy to tackle this in another PR, but I wanted to do it step by step.

- #3811

## Tasklist

 - [x] Review
 - [x] Adjust for comments
   - [x] Remove `Reset`, refactor `SQLQuery`
   - [x] C++11 inline initializer instead of constructor
   - [x] Non-virtual destructor
   - [x] Deleted copy constructor and assignment operator on `SQLResult`


gdal/ogr/ogrsf_frmts/sqlite/ogrsqliteutility.cpp
```diff
@@ -65,23 +87,13 @@ OGRErr SQLCommand(sqlite3 * poDb, const char * pszSQL)
...
-}
-
 OGRErr SQLQuery(sqlite3 * poDb, const char * pszSQL, SQLResult * poResult)
 {
     CPLAssert( poDb != nullptr );
     CPLAssert( pszSQL != nullptr );
     CPLAssert( poResult != nullptr );
 
-    SQLResultInit(poResult);
+    poResult->Reset();
```
julienr, line 96: This is the reason to have a Reset function besides constructor/destructor in SQLResult.

I guess we could remove this Reset function if `SQLQuery` was returning a SQLResult instead of taking an existing one. Happy to look into that if you feel that would be better.

rouault: Great that you tackled this ! Yes, having SQLQuery return a SQLResult would make things more robust

gdal/ogr/ogrsf_frmts/sqlite/ogrsqliteutility.h
```diff
@@ -37,25 +37,29 @@
...
-} SQLResult;
+    public:
+        char** papszResult;
+        int nRowCount;
+        int nColCount;
+        char *pszErrMsg;
+        int rc;
+
+        SQLResult ();
+        virtual ~SQLResult ();
```
rouault, line 50: I don't think the destructor needs to be virtual as there's little reason that this class should be derived. Actually making it final would be clearer.
And having deleted copy constructor and assignment operators would also be good as we have pointer members in it.

gdal/ogr/ogrsf_frmts/sqlite/ogrsqliteutility.cpp
```diff
@@ -41,6 +41,28 @@
 
 CPL_CVSID("$Id$")
 
+
+SQLResult::SQLResult ()
+    : papszResult(nullptr), nRowCount(0), nColCount(0), pszErrMsg(nullptr), rc(0) {}
```
rouault, line 46: I would use just C++11 inline member variable declaration. That  makes it more obvious that every member is properly initialized

gdal/ogr/ogrsf_frmts/sqlite/ogrsqliteutility.cpp
julienr: Thanks, I'll work on refactoring SQLQuery then !

I have one design question before jumping in: `SQLQuery` can fail and if we can use C++11, I feel the natural way would be to to return a `std::unique_ptr<SQLResult>` that can be nullptr in case of error :

`std::unique_ptr<SQLResult> SQLQuery(sqlite3* poDB, const char* pszSQL)`

This means we loose the error code though (but currently, the only error possibly returned is `OGRERR_FAILURE`, so we aren't loosing much).

If we want to keep the error code, the alternative would be to take a unique_ptr by reference and modify it to store the result if the function returns `OGRERR_NONE`. I'm personally not a big fan of this as it kind of complexifies things I feel because now the caller has to check the returned error code but would also in theory has to check the pointer is actually not null.

`OGRErr SQLQuery(sqlite3* poDB, const char* pszSQL, std::unique_ptr<SQLResult>& poResult)`

Any guidelines about this ? I see the `std::unique_ptr<T>&` pattern used in some places (e.g. `GetCompoundDataType` in netcdfmultidim.cpp), but not sure if its intentional or not.

rouault: > std::unique_ptr<SQLResult> SQLQuery(sqlite3* poDB, const char* pszSQL)

That signature looks good

rouault:
Build issue in debug builds:
```
 ogrsqliteutility.cpp:130:16: error: use of undeclared identifier 'poResult'
    CPLAssert( poResult != nullptr );
```

gdal/ogr/ogrsf_frmts/gpkg/ogrgeopackagetablelayer.cpp
```diff
@@ -797,39 +790,34 @@ OGRErr OGRGeoPackageTableLayer::ReadTableDefinition()
...
+                oResultGeomCols = SQLQuery(poDb, pszSQL);
                 sqlite3_free(pszSQL);
             }
 
             /* gpkg_geometry_columns query has to work */
             /* gpkg_geometry_columns.table_name is supposed to be unique */
-            if ( err != OGRERR_NONE || oResultGeomCols.nRowCount != 1 )
+            if ( !oResultGeomCols || oResultGeomCols->RowCount() != 1 )
             {
-                if ( err != OGRERR_NONE )
```
julienr, line 808: Small behavior change: I've removed this if because it's not possible anymore to get the error message since SQLQuery returns null on error right now. But anyway, SQLQuery already calls `CPLError` with the error message, so I believe this was printing the error message twice

gdal/ogr/ogrsf_frmts/gpkg/ogrgeopackagetablelayer.cpp
```diff
@@ -846,27 +834,23 @@ OGRErr OGRGeoPackageTableLayer::ReadTableDefinition()
...
     /*  1|name|varchar|0||0 */
     char* pszSQL = sqlite3_mprintf("pragma table_info('%q')", m_pszTableName);
-    err = SQLQuery(poDb, pszSQL, &oResultTable);
+    auto oResultTable = SQLQuery(poDb, pszSQL);
     sqlite3_free(pszSQL);
 
-    if ( err != OGRERR_NONE || oResultTable.nRowCount == 0 )
+    if ( !oResultTable || oResultTable->RowCount() == 0 )
     {
-        if( oResultTable.pszErrMsg != nullptr )
```
julienr, line 854: Same comment regarding small behavior change, but I think it's harmless.

gdal/ogr/ogrsf_frmts/sqlite/ogrsqliteutility.h
```diff
@@ -36,26 +36,32 @@
...
+        SQLResult(char** papszResult, int nRow, int nCol);
+        ~SQLResult ();
+
+        int         RowCount() const;
+        int         ColCount() const;
+        void        LimitRowCount(int nLimit);
+
+        const char* GetValue(int iColumnNum, int iRowNum) const;
+        int         GetValueAsInteger(int iColNum, int iRowNum) const;
+    private:
```
julienr, line 53: While I was at it, I made the class more encapsulated and also removed the error message member variable since those are really only used in `SQLQuery`. So now,  a `SQLResult` will always hold the result of a successful query.

gdal/ogr/ogrsf_frmts/gpkg/ogrgeopackagetablelayer.cpp
```diff
@@ -4431,14 +4394,14 @@ OGRErr OGRGeoPackageTableLayer::RecreateTable(const CPLString& osColumnsForCreat
...
-    OGRErr eErr = SQLQuery(hDB, pszSQL, &oTriggers);
+    OGRErr eErr = OGRERR_NONE;
+    auto oTriggers = SQLQuery(hDB, pszSQL);
     sqlite3_free(pszSQL);
 
 /* -------------------------------------------------------------------- */
 /*      Make a temporary table with new content.                        */
 /* -------------------------------------------------------------------- */
-    if( eErr == OGRERR_NONE )
+    if( oTriggers )
```
rouault, line 4404: eErr shoud be set to OGRERR_FAILURE when oTriggers == null

gdal/ogr/ogrsf_frmts/sqlite/ogrsqliteutility.h
```diff
@@ -36,26 +36,32 @@
...
-    int nRowCount;
-    int nColCount;
-    char *pszErrMsg;
-    int rc;
-} SQLResult;
+    public:
+        SQLResult(char** papszResult, int nRow, int nCol);
+        ~SQLResult ();
+
+        int         RowCount() const;
```
rouault, line 47: a inline definition for RowCount() and ColCount() would be more appropriate

gdal/ogr/ogrsf_frmts/gpkg/ogrgeopackagetablelayer.cpp
julienr: Good catch, fixed it

gdal/ogr/ogrsf_frmts/sqlite/ogrsqliteutility.h
julienr: True, fixed

rouault:
```
ogrgeopackagetablelayer.cpp: In member function ‘OGRErr OGRGeoPackageTableLayer::RecreateTable(const CPLString&, const CPLString&)’:

ogrgeopackagetablelayer.cpp:4449:44: error: null pointer dereference [-Werror=null-dereference]

     for( int i = 0; i < oTriggers->RowCount() && eErr == OGRERR_NONE; i++)

                         ~~~~~~~~~~~~~~~~~~~^~
```

PR #3889, opened 2021-05-26
Doc: add link to recording how Frank prounounces GDAL

rouault:


jmckenna:
@gfenoy you're famous :)

PR #3899, opened 2021-05-28
MRF: Fix JPEG Huffman optimize flag logic

lucianpls:
Had the opposite of the intended effect

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:

### DIFF
--- a/gdal/frmts/mrf/JPEG_band.cpp
+++ b/gdal/frmts/mrf/JPEG_band.cpp
@@ -679,7 +679,7 @@ JPEG_Band::JPEG_Band( MRFDataset *pDS, const ILImage &image,
     }
 
     if( GDT_Byte == image.dt )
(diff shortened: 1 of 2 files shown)


gdal/frmts/mrf/JPEG_band.cpp
```diff
@@ -679,7 +679,7 @@ JPEG_Band::JPEG_Band( MRFDataset *pDS, const ILImage &image,
     }
 
     if( GDT_Byte == image.dt )
-        codec.optimize = GetOptlist().FetchBoolean("OPTIMIZE", FALSE) != FALSE;
+        codec.optimize = (true == GetOptlist().FetchBoolean("OPTIMIZE", FALSE));
```
rouault, line 682: can you explain why your change changes behaviour ? To me, the version before and after this change are equivalent.

Issue #3885, opened 2021-05-25
OGR Access MDB Driver appears unable to parse certain numbers


Saijin-Naib:
## Expected behavior and actual behavior.
For some odd reason QGIS does not properly read some records from some columns of a vanilla/aspatial Access MDB when joined against a GPKG with mixed spatial/aspatial data.

## Steps to reproduce the problem.
Have GPKG with geometries and some tabular data joined against a completely vanilla/aspatial MDB
Note that some records in QGIS are not parsed/loaded correctly (For example, CAMP_Length)
Verify that these records can be recalled properly outside of QGIS/OGRinfo

## Operating system
OS Name	Microsoft Windows 10 Pro
Version	10.0.19043 Build 19043

## GDAL version and provenance
GDAL 3.3.0 from OSGeo4W-NG

## Further Information
QGIS view:  
![image](https://user-images.githubusercontent.com/19295950/119404253-8dd76600-bcad-11eb-8533-534a3049596a.png)
Note how many segments have a CAMP_Length of 0. This is not valid. Some records in the dataset are NULL, but they show properly as NULL.

Compare to the view in CAMP-RS:  
![image](https://user-images.githubusercontent.com/19295950/119404336-a778ad80-bcad-11eb-8674-ac6a1208b03e.png)

Compare to the view in MDBPlus:  
![image](https://user-images.githubusercontent.com/19295950/119404431-cbd48a00-bcad-11eb-9a6f-2d61641bbe1a.png)

```
PS C:\OSGeo4W64\bin> .\ogrinfo.exe -fields=yes -sql "SELECT * FROM Inventory" Q:\User\Brett_Carlock\Projects\QGIS\Data\CAMP-RS\CAMP-RS.mdb > stdout 2> stderr
```

It looks like OGR isn't reading some of the values properly, at least according to stderr. I guess these are the ones that read out as 0 when they should not be.

It is tricky to pick out in the stdout because BEGIN_MP of 0 is totally valid, especially for a segment that begins that roadway (the milepoint of that roadway would normally begin at 0.0mi and increase from there). However, END_MP and LENGTH can NEVER be 0, so any records that OGR parses with those as 0 are erroneous. NULL is perfectly acceptable.

[stdout.txt](https://github.com/qgis/QGIS/files/6539643/stdout.txt)
[stderr.txt](https://github.com/qgis/QGIS/files/6539644/stderr.txt)

**Additional context**
Same dataset as Issue #43287 
[...]

PR #3914, opened 2021-06-01
Fix incorrect ODBC type mapping for a number of SQL types

nyalldawson:
Followup 729374778


rouault:
is it related to #3885 ?

nyalldawson:
I don't think it actually causes any reported issues.. I just noticed it and it's clearly wrong

PR #3923, opened 2021-06-01
Python import: fix FileNotFoundError when PATH has nonexistent paths (fixes #3898)

### LINKED ISSUE #3898: USE_PATH_FOR_GDAL_PYTHON=YES causes FileNotFoundError on Windows
#### Expected behavior:
Setting the env variable `USE_PATH_FOR_GDAL_PYTHON=YES` as suggested by an import error message should not cause another error. Where `os.add_dll_directory(p)` is called, it should handle the case where `p` is not found.

#### Actual behavior:
Setting the env variable `USE_PATH_FOR_GDAL_PYTHON=YES` causes `from osgeo import gdal` to fail with a different error:
```
  File "demo.py", line 1, in <module>
    from osgeo import gdal
  File "C:\Miniconda\envs\windows-latest-test-env\lib\site-packages\osgeo\__init__.py", line 10, in <module>
    os.add_dll_directory(p)
  File "C:\Miniconda\envs\windows-latest-test-env\lib\os.py", line 1109, in add_dll_directory
    cookie = nt._add_dll_directory(path)
FileNotFoundError: [WinError 3] The system cannot find the path specified: 'C:\\Miniconda\\envs\\windows-latest-test-env\\Library\\mingw-w64\\bin'
```
I'm not sure why a nonexistent path is on the PATH, but since it's happening in a fresh Windows instance on Github Actions, I think GDAL should handle this possibility and just skip the path. I'd be happy to put in a PR for this to add something like:
```
try:
    os.add_dll_directory(p)
except FileNotFoundError:
    continue
```

demo.py:
```
from osgeo import gdal
```

```
$ conda install gdal==3.3.0
$ export USE_PATH_FOR_GDAL_PYTHON=YES
$ python demo.py
```

See [this Github Actions run](https://github.com/emlys/invest/runs/2686856073) for a minimal example in action.

## Operating system

Windows 10.0.17763

3.2.2 and 3.3.0 from conda-forge

emlys:
## What does this PR do?
Prevent FileNotFoundError that can happen when importing from osgeo on Windows, python>=3.8.

I didn't see anywhere obvious to put a test case for this specific import behavior, let me know if that should go somewhere.

Fixes #3898 

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:

### DIFF
--- a/.github/workflows/windows_build.yml
+++ b/.github/workflows/windows_build.yml
@@ -249,7 +249,11 @@ jobs:
             }
             $env:SDK_PREFIX="$env:GITHUB_WORKSPACE\sdk\$env:SDK"
             $env:SDK_BIN="$env:SDK_PREFIX\bin"
-            $env:PATH="$env:GITHUB_WORKSPACE\gdal;$env:GITHUB_WORKSPACE\gdal\apps;$env:SDK_PREFIX\dll;$env:SDK_PREFIX;$env:SDK_BIN;$env:PATH"
+            # Set USE_PATH_FOR_GDAL_PYTHON=YES and include a fake path in the PATH
+            # to test robustness to nonexistent paths (only relevant for python>=3.8)
+            # See https://github.com/OSGeo/gdal/issues/3898
+            $env:USE_PATH_FOR_GDAL_PYTHON="YES"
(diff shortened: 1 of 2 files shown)


rouault:
Regarding tests, you might edit https://github.com/OSGeo/gdal/blob/master/.github/workflows/windows_build.yml#L252 to add a dummy path (and a comment explaining this is to test we're robust) , as well as adding a line to define USE_PATH_FOR_GDAL_PYTHON=YES . This will not be used currently as we're on python 3.7 for now, but at some later point where we upgrade our python version, this will be triggered

rouault:
manually backported to 3.3

PR #3931, opened 2021-06-03
Doc: update gdaltindex output formats

JakobMiksch:
## What does this PR do?

- it updates the documentation of [gdaltindex](https://gdal.org/programs/gdaltindex.html#gdaltindex). 
`gdaltindex` meanwhile can create all OGR-supported datasets and is not limited to Shapefile anymore. 
- it also updates the basic example for using GeoPackage instead of Shapefile

Please let me know if `OGR-supported dataset` is the correct wording in this case. Otherwise I am happy to correct.

- not found

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [x] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/gdal/doc/source/programs/gdaltindex.rst
+++ b/gdal/doc/source/programs/gdaltindex.rst
@@ -89,14 +89,20 @@ tileindex.
 Examples
 --------
 
-- Produce a GeoPackage (``index.gpkg``) with a record for every
-  image that the utility found in the ``raster_files`` folder. Each record holds
+- Produce a shapefile (``doq_index.shp``) with a record for every
+  image that the utility found in the ``doq`` folder. Each record holds
   information that points to the location of the image and also a bounding rectangle
   shape showing the bounds of the image:
 
 ::
 
-    gdaltindex index.gpkg raster_files/*.tif
+    gdaltindex doq_index.shp doq/*.tif
+
+- Perform the same command as before, but now we create a GeoPacakge instead of a Shapefile. 
+
+::
+
+    gdaltindex -f GPKG doq_index.gpkg doq/*.tif
 
 - The :option:`-t_srs` option can also be used to transform all input rasters
   into the same output projection:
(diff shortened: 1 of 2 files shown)


gdal/doc/source/programs/gdaltindex.rst
```diff
@@ -89,14 +89,14 @@ tileindex.
...
-  image that the utility found in the ``doq`` folder. Each record holds
+- Produce a GeoPackage (``index.gpkg``) with a record for every
+  image that the utility found in the ``raster_files`` folder. Each record holds
   information that points to the location of the image and also a bounding rectangle
   shape showing the bounds of the image:
 
 ::
 
-    gdaltindex doq_index.shp doq/*.tif
+    gdaltindex index.gpkg raster_files/*.tif
```
jratike80, line 99: I would rather use an example `gdaltindex -f GPKG index.gpkg raster_files/*.tif`. GDAL does guess the output format from the extension but it does not hurt to be explicit.

jratike80:
In the existing documentation there is an introduction in the Examples part:

> Examples
> Produce a shapefile (doq_index.shp) with a record for every image that the utility found in the doq folder. Each record holds information that points to the location of the image and also a bounding rectangle shape showing the bounds of the image:

The three examples are all related to the introduction: Basic usage, with -srs, with optfile. Perhaps it would be better to keep those and add a new item with its own description?

gdal/doc/source/programs/gdaltindex.rst
JakobMiksch: thanks for the comment. I applied your suggestion in 1343146

JakobMiksch:
@jratike80 Thanks for your review. I reverted the example to its previous state and additionally added an example for the GeoPackage. The new example uses the `-f GPKG` option as suggested in your inline comment.

Issue #3913, opened 2021-05-31
NoData in multiband images may produce wrong embedded overviews


jjimenezshaw:
## Expected behavior and actual behavior.
When using NoData value in multiple bands (but in pixel at different locations), the "average" overview computation includes the NoData value in the computation of the valid pixels.

The expected behavior is not including the NoData value in the overviews average computation.

If the overviews are generated as external ovr file, it is not happening.

## Steps to reproduce the problem.
Generate an image with overviews this way (`test.asc` and `test2.asc` just swap the nodata pixel):
```
$ cat test.asc
ncols        2
nrows        2
xllcorner    440720.000000000000
yllcorner    3750120.000000000000
cellsize     60.000000000000
NODATA_value  -10000
 0.5 1.0
 2.5 -10000

$ cat test2.asc
ncols        2
nrows        2
xllcorner    440720.000000000000
yllcorner    3750120.000000000000
cellsize     60.000000000000
NODATA_value  -10000
 -10000 1.0
 2.5 0.5

$ cat tmp.vrt
<VRTDataset rasterXSize="2" rasterYSize="2">
  <GeoTransform>  4.4072000000000000e+05,  6.0000000000000000e+01,  0.0000000000000000e+00,  3.7513200000000000e+06,  0.0000000000000000e+00, -6.0000000000000000e+01</GeoTransform>
  <VRTRasterBand dataType="Float32" band="1">
    <NoDataValue>-10000</NoDataValue>
    <ComplexSource>
      <SourceFilename relativeToVRT="1">test.asc</SourceFilename>
      <SourceBand>1</SourceBand>
      <NODATA>-10000</NODATA>
    </ComplexSource>
  </VRTRasterBand>
  <VRTRasterBand dataType="Float32" band="2">
    <NoDataValue>-10000</NoDataValue>
    <ComplexSource>
      <SourceFilename relativeToVRT="1">test2.asc</SourceFilename>
      <SourceBand>1</SourceBand>
      <NODATA>-10000</NODATA>
    </ComplexSource>
  </VRTRasterBand>
</VRTDataset>

$ gdal_translate tmp.vrt test.tif

$ gdaladdo -r average test.tif 2
```

Check that the pixel value in the second band on the overview has a value using the wrong mask:
```
$ gdal_translate test.tif /vsistdout/ -of aaigrid -b 2 -outsize 1 1
ncols        1
nrows        1
xllcorner    440720.000000000000
yllcorner    3751200.000000000000
[...]

PR #3932, opened 2021-06-03
Nodata multiband interpolated overviews

jjimenezshaw:
## What does this PR do?
nodata mask may be different per band. The method `GDALRegenerateOverviewsMultiBand` was taking the first band as a mask for all the bands. This PR fixes that problem.

Fixes issue #3913 

## Tasklist

 - [x] Add test case(s)
 - [ ] Review
 - [x] Adjust for comments

## Side effects
There is a unit test modified. After looking (visually) at the result of this test in master and in this branch, it looks better with this branch. I am curious how a change in the nodata mask affects a file without nodata value.
EDIT: this change is gone. I leave this in the description to understand the conversation.

### DIFF
--- a/gdal/gcore/overview.cpp
+++ b/gdal/gcore/overview.cpp
@@ -3996,7 +3996,7 @@ GDALRegenerateOverviews( GDALRasterBandH hSrcBand,
     // In case the mask made be computed from another band of the dataset,
     // we can't use cascaded generation, as the computation of the overviews
     // of the band used for the mask band may not have yet occurred (#3033).
-    if( (STARTS_WITH_CI(pszResampling, "AVER") |
+    if( (STARTS_WITH_CI(pszResampling, "AVER") ||
          STARTS_WITH_CI(pszResampling, "GAUSS") ||
          EQUAL(pszResampling, "RMS") ||
          EQUAL(pszResampling, "CUBIC") ||
@@ -4776,7 +4776,7 @@ GDALRegenerateOverviewsMultiBand( int nBands, GDALRasterBand** papoSrcBands,
         struct OvrJob
         {
             // Buffers to free when job is finished
-            std::shared_ptr<PointerHolder> oSrcMaskBufferHolder{};
+            std::unique_ptr<PointerHolder> oSrcMaskBufferHolder{};
             std::unique_ptr<PointerHolder> oSrcBufferHolder{};
             std::unique_ptr<PointerHolder> oDstBufferHolder{};
 
@@ -4894,7 +4894,7 @@ GDALRegenerateOverviewsMultiBand( int nBands, GDALRasterBand** papoSrcBands,
         std::list<std::unique_ptr<OvrJob>> jobList;
 
         std::vector<void*> apaChunk(nBands);
-        GByte* pabyChunkNoDataMask = nullptr;
+        std::vector<GByte*> apabyChunkNoDataMask(nBands);
 
         int nDstYOff = 0;
         // Iterate on destination overview, block by block.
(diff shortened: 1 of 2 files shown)


rouault:
> I am curious how a change in the nodata mask affects a file without nodata value.

For that unit test, could you check on that case if GetMaskBand() uses the alpha band of the previous overview level ? It might perhaps mis-identify it.

jjimenezshaw:
> > I am curious how a change in the nodata mask affects a file without nodata value.
> 
> For that unit test, could you check on that case if GetMaskBand() uses the alpha band of the previous overview level ? It might perhaps mis-identify it.

About why is it getting into the code is due to `GetMaskFlags()` in the code below, that returns `6` (`GMF_PER_DATASET` | `GMF_ALPHA`)
```
    const bool bUseNoDataMask =
        !STARTS_WITH_CI(pszResampling, "NEAR") &&
        (bIsMask || (papoSrcBands[0]->GetMaskFlags() & GMF_ALL_VALID) == 0);
```

About what is using `GetMaskBand()`, I am not sure. After debuging it seems reasonable what it is doing.

Comparing the outputs, the alpha band of the first overview is already different. The version of this branch is softer in the borders of the image. The output from master is very rough, not what you would expect from a cubic interpolation.

gdal/gcore/overview.cpp
```diff
@@ -5049,28 +5049,20 @@ GDALRegenerateOverviewsMultiBand( int nBands, GDALRasterBand** papoSrcBands,
...
-                    eErr = poMaskBand->RasterIO(
-                        GF_Read,
-                        nChunkXOffQueried, nChunkYOffQueried,
-                        nChunkXSizeQueried, nChunkYSizeQueried,
-                        pabyChunkNoDataMask,
-                        nChunkXSizeQueried, nChunkYSizeQueried,
-                        GDT_Byte, 0, 0, nullptr );
+                    if( bUseNoDataMask && eErr == CE_None )
+                    {
+                        auto poMaskBand = bIsMask ? poSrcBand : poSrcBand->GetMaskBand();
```
rouault, line 5055: could you change that logic to something like
```suggestion
                        auto poMaskBand = bIsMask || poSrcBand->GetColorInterpretation() == GCI_AlphaBand ? poSrcBand : poSrcBand->GetMaskBand();
```
I guess this should lead to the previous result. The fact that we are rough on the edges between alpha=0 and != 0 was a wanted behaviour to avoid "expending" the validity area when computing overviews.

jjimenezshaw: Doing that the test ovr_32 leads to the previous result (well, using `papoSrcBands[iBand]` instead of `poSrcBand`. Looks like the color interpretation is not well defined in `papapoOverviewBands`... and I think I had the same symptoms last week in COG).

If I understand correctly, it is using the alpha band itself to mask the alpha band. Then combining alpha band (with only 0 and 255) and "nodata value"=-10000 produces that values that are 0 in the main image, become -10000 in the overviews (instead of stay as 0). I am not sure if that is what we want. What about `auto poMaskBand = bIsMask || (!pabHasNoData[iBand] && papoSrcBands[iBand]->GetColorInterpretation() == GCI_AlphaBand) ? poSrcBand : poSrcBand->GetMaskBand();` ?

About the test ovr_32 creating the overviews for `stefan_full_rgba.tif`, the output is really rough with master.
You can compare the first overview in this images:
Master:  
master
This branch (without filtering with GetColorInterpretation): 
branch

rouault: > What about `auto poMaskBand = bIsMask || (!pabHasNoData[iBand] && papoSrcBands[iBand]->GetColorInterpretation() == GCI_AlphaBand) ? poSrcBand : poSrcBand->GetMaskBand();` ?

sound plausible, although I'm starting to be a bit lost :-) A comment in the code explaining the scenario you mention would be useful. I guess the GetMaskBand() logic could have a special case when both nodata and alpha are present, but I'm not sure we want to get into that level of complication.

> About the test ovr_32 creating the overviews for stefan_full_rgba.tif, the output is really rough with master.

yeah, for a real alpha channel like this image with not just 0 and 255 values, this doesn't look good. The current behavior was rather tuned for an alpha channel that is only 0 and 255, so just masking. Perhaps we could select the appropriate behavior by analyzing the content of the mask if it comes from an alpha band : is it just 0 and 255, or does it include intermediate values ? But that should be done in a separate pull request as I'd like that one to be the minimum to fix the broken case to be backportable.

jjimenezshaw: Yes, some things better for a later PR. I squashed the commits in one, including the GetColorInterpretation usage.

PR #3978, opened 2021-06-14
Fix 3977 iso 8601 dates in filter

### LINKED ISSUE #3977: Date filter does not work with IN operator and ISO-8601 dates
See: https://github.com/qgis/QGIS/issues/43651#issuecomment-860577285

```
$ cat test.csv
1,"2021/06/14"

$ cat test.csvt
Integer,Date

$ ogrinfo test.csv -where "field_2 = '2021-06-14'" -al -q

Layer name: test
OGRFeature(test):1
  field_1 (Integer) = 1
  field_2 (Date) = 2021/06/14
```

but this fails:
```
ogrinfo test.csv -where "field_2 IN ('2021-06-14')" -al -q
```

elpaso:
Fixes #3977

### DIFF
--- a/autotest/ogr/ogr_shape.py
+++ b/autotest/ogr/ogr_shape.py
@@ -4944,6 +4944,17 @@ def test_ogr_in_date_filter():
     lyr.SetAttributeFilter("date_minus IN ('1960/12/31', '1950/12/31')")
     _ogr_in_date_filter_check([0, 1])
 
+    lyr.SetAttributeFilter("date_minus IN ('2020/12/31', '2020/12/31')")
+    _ogr_in_date_filter_check([])
+
+    lyr.SetAttributeFilter("date_minus IN ('2020-12-31', '2020-12-31')")
+    _ogr_in_date_filter_check([])
+
+    lyr.SetAttributeFilter("date_slash IN ('2020/12/31', '2020/12/31')")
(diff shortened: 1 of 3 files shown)


autotest/ogr/ogr_shape.py
```diff
@@ -4894,6 +4894,69 @@ def test_ogr_shape_write_multipolygon_z_non_finite():
     ogr.GetDriverByName('ESRI Shapefile').DeleteDataSource('/vsimem/test.shp')
 
 ###############################################################################
+# Test that date fields stored as ISO-8601 can be used with IN operator
+
+def test_ogr_in_date_filter():
```
rouault, line 4899: As this is not shapefile specific, should rather be moved into autotest/ogr/ogr_sql_rfc28.py

autotest/ogr/ogr_shape.py
```diff
@@ -4894,6 +4894,69 @@ def test_ogr_shape_write_multipolygon_z_non_finite():
...
+    """Test that date fields stored as ISO-8601 can be used with IN operator"""
+
+    ds = ogr.GetDriverByName('ESRI Shapefile').CreateDataSource(
```
rouault, line 4902: You can just create a in-memory dataset:
    ds = ogr.GetDriverByName('MEM').CreateDataSource('')

PR #3986, opened 2021-06-15
Removed linux commands in makefile.vc

C41eb:
makefile.vc was calling linux utilites: "rm", "cp" and "zip" inside of the windows makefile. This was causing an error when I was building the windows build on my local machine. I have updated the section to check if the $(DISTDIR) exists and if not make the directory and then copy the gdal build artefacts into the $(DISTDIR).


rouault:
I suspect this target was meant to be used with cygwin tools installed. Your change lacks the copying of the public header files

C41eb:
Okay no worries, I will update to include the public header files

PR #3988, opened 2021-06-15
Changed setting of LIBS to append $saved_LIBS instead of $LIBS in configure

C41eb:
In the configure script for sqlite it seems that when setting LIBS after checking for SQLITE, in some paths the variable $LIBS get set to "". Subsequently when setting LIBS up it then appends this empty variable, which causes SQLITE to fail durng configuration.

### DIFF
--- a/gdal/configure
+++ b/gdal/configure
@@ -24765,7 +24765,7 @@ if ${ac_cv_lib_sqlite3_sqlite3_open+:} false; then :
   $as_echo_n "(cached) " >&6
 else
   ac_check_lib_save_LIBS=$LIBS
-LIBS="-lsqlite3 -L$ac_sqlite3_path/lib $LIBS"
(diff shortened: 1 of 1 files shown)


rouault:
I'm not sure about those changes. The source file is gdal/m4/ax_lib_sqlite3.m4. You chould do changes in it, run ``autoconf`` and commit both gdal/m4/ax_lib_sqlite3.m4 and gdal/configure

C41eb:
Yes it seems to of broken the MacOS build. I shall look into changing the source file, thanks.

rouault:
> Yes it seems to of broken the MacOS build

no, that's just a flake due to network issues in some tests

C41eb:
Closing as further investigation has shown that this was a problem with build environment rather than the configure itself.

PR #3991, opened 2021-06-15
RMS downsampling by factor of 2: speed-up SSE2 impl for UInt16 case with values >= 16384

rouault:


rouault:
Dropping the idea for now: in addition to getting to making sure we get consistent results in all code paths of UInt16 (pure C, SSE2 when values < 16384, SSE2 when values >= 16384), we also need to make sure we get consistent results with the Byte case, which this breaks...

Issue #3984, opened 2021-06-15
GTiff  signedbyte : zero values written instead of negative nodata value for tiles/strips entirely at the nodata value


sgillies:
While working on write support for int8 arrays in rasterio https://github.com/mapbox/rasterio/issues/2180 I have found (I believe) that GDAL's IO functions depend on image structure metadata values being made persistent in a driver-specific way. For example, I can call GDALCreate with `driver=MEM` and `PIXELTYPE=SIGNEDBYTE` (in creation options) and then write a signed int buffer directly to the new dataset. But if I call GDALCreate with `driver=GTiff` and `PIXELTYPE=SIGNEDBYTE` (in creation options) I find that I have to close the dataset and reopen it in update mode before I can write a signed int buffer. So, I'm trying to switch to the later in rasterio. It does fix the issue reported in #2180.

However, the change has destabilized other parts of rasterio. Color interpretation is one. For GeoTIFFs, the effect of `PHOTOMETRIC=CMYK` is lost when a dataset created with `driver=GTiff` and `PHOTOMETRIC=CMYK` is re-opened in update mode.

Is there a driver-independent best practice for creating a new dataset and writing data to it? Is there a simple split into creation-only and other formats? Or is it very driver specific?

rouault:
> But if I call GDALCreate with `driver=GTiff` and `PIXELTYPE=SIGNEDBYTE` (in creation options) I find that I have to close the dataset and reopen it in update mode before I can write a signed int buffer.

Sounds like a bug of the GTiff driver. You shouldn't have to close & re-open

> For GeoTIFFs, the effect of PHOTOMETRIC=CMYK is lost when a dataset created with driver=GTiff and PHOTOMETRIC=CMYK is re-opened in update mode.

CMYK handling is particular indeed. On re-opening, one uses the TIFF RGBA interface for such exotic colorspaces

> Is there a driver-independent best practice for creating a new dataset and writing data to it? 

Directly write to it is the recommended way (pending bugs like the one you spotted with signedbyte in TIFF)

rouault:
> Sounds like a bug of the GTiff driver. You shouldn't have to close & re-open

[...]

PR #3999, opened 2021-06-16
Add a test of writing a signedbyte dataset

sgillies:
See #3984. To try to reproduce https://github.com/mapbox/rasterio/issues/2180.

### DIFF
--- a/autotest/gcore/numpy_rw.py
+++ b/autotest/gcore/numpy_rw.py
@@ -222,7 +222,7 @@ def test_numpy_rw_10():
 def test_numpy_rw_10_bis():
     """Attempt to reproduce https://github.com/mapbox/rasterio/issues/2180"""
     ds = gdal.GetDriverByName('GTiff').Create('/vsimem/signed8.tif', 2, 1, options=['PIXELTYPE=SIGNEDBYTE'])
-    ar = numpy.array([[-1, -1]], dtype=numpy.int8)
+    ar = numpy.array([[-1, -1]], dtype=numpy.int8).astype(np.uint8)
     ds.GetRasterBand(1).WriteArray(ar)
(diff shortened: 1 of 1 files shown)


autotest/gcore/numpy_rw.py
```diff
@@ -218,6 +218,22 @@ def test_numpy_rw_10():
...
+    ds.GetRasterBand(1).WriteArray(ar)
+    ds.GetRasterBand(1).SetNoDataValue(-1)
+    ds = None
+
+    ds = gdal.Open('/vsimem/signed8.tif')
+    ar2 = ds.ReadAsArray()
+    ds = None
+    gdal.Unlink('/vsimem/signed8.tif')
+
+    assert ar2[0][0] == -1 and ar2[0][1] == -1
```
sgillies, line 235: @rouault the key difference from this to `test_numpy_rw_10` is that we set a nodata value and all array cells written to the dataset have that value (in int8).

sgillies: I've checked that `ar = numpy.array([[-1, -1]], dtype=numpy.int8).astype(numpy.uint8)` does what GDAL requires. All the items of `ar` are `255`.

sgillies: I think I'm able to pin it down to a nodata issue. If we write an array that has all nodata values, there is an error. See https://github.com/mapbox/rasterio/commit/fb065a0f8e437156ccd101c01dd533e96d8c5139#diff-53e13383b76f631d209604d01e8dd23b11ebf681f3a5c930f1a9b38f507bcc44R8.

rouault:
Reproducer incorporated in fix of https://github.com/OSGeo/gdal/pull/4001

Issue #4004, opened 2021-06-17
Exporting features with curve geometries to GeoJSON silently drops them


mmomtchev:
## Current behavior
Exporting a feature to GeoJSON currently produces a debug message and silently sets the geometry to `NULL` without setting last error, ie this is not considered an event worth mentioning to the end user

```cpp
char *text = geom->exportToJson()
```
`text` will be `NULL` and `CPLGetLastErrorMsg` will still return the previous error message encountered

## Suggested new behavior

The return value stays `NULL` but a `CPLError` is reported

## Implications

~~Currently `ogr2ogr` silently drops curve geometries when producing GeoJSON output.~~
`ogr2ogr` doesn't use exportToJson

mmomtchev:
https://github.com/contra/node-gdal-next/issues/13

rouault:
> Currently `ogr2ogr` silently drops curve geometries when producing GeoJSON output.

I don't confirm this. ogr2ogr / OGR core will linearize curve geometry. See:
```
$ cat test.csv
id,WKT
1,"CIRCULARSTRING(0 0,1 1,2 0)"

$ ogr2ogr -f geojson /vsistdout/ test.csv
{
"type": "FeatureCollection",
"name": "test",
"features": [
[...]

PR #4006, opened 2021-06-17
Error msg when geometry is not supported in GeoJSON

mmomtchev:
Transform CPLDebug into CPLError

Rationale:
This is a serious error condition that causes the function to fail and to set a null geometry for the feature
Without it, a NULL geometry is returned and there is no `LastErrorMsg`

It doesn't break any unit tests

Refs: #4004

### DIFF
--- a/gdal/ogr/ogrsf_frmts/geojson/ogrgeojsonwriter.cpp
+++ b/gdal/ogr/ogrsf_frmts/geojson/ogrgeojsonwriter.cpp
@@ -1079,9 +1079,9 @@ json_object* OGRGeoJSONWriteGeometry( const OGRGeometry* poGeometry,
                                             oOptions );
         else
         {
-            CPLDebug( "GeoJSON",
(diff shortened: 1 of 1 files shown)


gdal/ogr/ogrsf_frmts/geojson/ogrgeojsonwriter.cpp
```diff
@@ -1079,9 +1079,9 @@ json_object* OGRGeoJSONWriteGeometry( const OGRGeometry* poGeometry,
                                             oOptions );
         else
         {
-            CPLDebug( "GeoJSON",
-                      "Unsupported geometry type detected. "
-                      "Feature gets NULL geometry assigned." );
+            CPLError(CE_Failure, CPLE_NotSupported,
+                     "Unsupported GeoJSON geometry type detected. "
```
rouault, line 1083:
```suggestion
                     "OGR geometry type unsupported as a GeoJSON geometry detected. "
```

PR #4026, opened 2021-06-23
CEOS2 ASF Compability

kdesjard:
These changes add support for reading GCP's from ERS-1, JERS-1 and Radarsat-1 CEOS Level-1 products from the Alaska Satellite Facility (ASF).  The ASF L1 products have leader files with different identifiers and formats than the standard RSI spec.


gdal/frmts/ceos2/sar_ceosdataset.cpp
```diff
@@ -1558,6 +1639,14 @@ void SAR_CEOSDataset::ScanForGCPs()
         return;
     }
 
+    /* ASF L1 products do not have valid data
+       in the lat/long first/mid/last fields */
+    const char *pszValue = GetMetadataItem("CEOS_FACILITY");
+    if(strncmp((char *)pszValue,"ASF",3) == 0) {
```
rouault, line 1645:
```suggestion
    if(strncmp(pszValue,"ASF",3) == 0) {
```

gdal/frmts/ceos2/sar_ceosdataset.cpp
```diff
@@ -1513,29 +1563,60 @@ int SAR_CEOSDataset::ScanForMapProjection()
 
     for( int i = 0; i < nGCPCount; i++ )
     {
-        char         szId[32];
+        char         szId[34];
```
rouault, line 1566: I don't think this change is needed
```suggestion
        char         szId[32];
```

gdal/frmts/ceos2/sar_ceosdataset.cpp
```diff
@@ -1482,24 +1502,54 @@ int SAR_CEOSDataset::ScanForMapProjection()
...
+            !STARTS_WITH_CI(szField, "Ground") ) {
+            return FALSE;
+        } else {
+            GCPFieldSize = 17;
+            GCPOffset = 157;
+        }
+    }
+
+    char FieldSize[4];
+    snprintf(FieldSize,4,"A%d",GCPFieldSize);
```
rouault, line 1550:
```suggestion
    snprintf(FieldSize,sizeof(FieldSize),"A%d",GCPFieldSize);
```

rouault:
There are tabulation characters that have been added and should be removed

kdesjard:
Is it expected for me to convert all tabs that have been there for years?

rouault:
> Is it expected for me to convert all tabs that have been there for years?

The script only checks the tabulations in .cpp files, and there shouldn't be any in master

rouault:
Are there public datasets that could be used to extend autotest/gdrivers/sar_ceos.py ?

kdesjard:
I can provide a sample product that would be 66MB. Would that be acceptable?

rouault:
> I can provide a sample product that would be 66MB. Would that be acceptable?

not to include in this repository, but we could possibly put it on http://download.osgeo.org/gdal/data/

kdesjard:
yes, that was understood. How shall I upload the file?  I will work on the sar_ceos tests as well.

rouault:
> How shall I upload the file?

I'll upload it on your behalf if you provide a download link to it

kdesjard:
https://data.eodms-sgdot.nrcan-rncan.gc.ca/public/carts/20ba8a31-7b23-4c6f-a658-dcbb1229948a/

rouault:
> https://data.eodms-sgdot.nrcan-rncan.gc.ca/public/carts/20ba8a31-7b23-4c6f-a658-dcbb1229948a/

What is the license of the dataset ?

rouault:
https://github.com/OSGeo/gdal/edit//master/gdal/doc/source/drivers/raster/sar_ceos.rst would perhaps also need an update

kdesjard:
The dataset uses the Open Government License:

 https://open.canada.ca/en/open-government-licence-canada

rouault:
The files have been uploaded to http://download.osgeo.org/gdal/data/sar_ceos/R1_26161_FN1_F164.
One option for testing would be to truncate the file to a few kilobytes, the minimum sufficient for the Open() method to recognize it. If that's possible then incorporating that truncated file in autotest/gdrivers/data/sar_ceos/ would be the best option.

kdesjard:
Re: truncation, I will look into

kdesjard:
Uploaded a modified .D file to the same URL.  Only the .D and .L files are needed

rouault:
> Uploaded a modified .D file to the same URL. Only the .D and .L files are needed

given their modest size, you can directly incorporate them in a to-be-created autotest/gdrivers/data/sar_ceos/ directory in your branch

rouault:
``import pytest`` at line 34 of sar_ceos.py should be removed as no longer used

gdal/frmts/ceos2/sar_ceosdataset.cpp
```diff
@@ -1558,6 +1643,17 @@ void SAR_CEOSDataset::ScanForGCPs()
         return;
     }
 
+    /* ASF L1 products do not have valid data
+       in the lat/long first/mid/last fields */
+    const char *pszValue = GetMetadataItem("CEOS_FACILITY");
+    if(pszValue == nullptr)
+        return;
```
rouault, line 1650: I don't know anything about that format, but isn't this return too radical ? Shouldn't just the non-null check be adding to the following strncmp() ?

kdesjard: apologies, fixed

PR #4030, opened 2021-06-23
OGRSpatialReference: Support GetLinearUnits for ellipsoidal and spherical coordinate systems that have a height axis

IanLilleyT:
## What does this PR do?
This PR adds ellipsoidal and spherical coordinate system support for `OGRSpatialReference::GetLinearUnits`, as long as they have a third axis for height.

Before this PR, if you called `SetLinearUnits` on an ellipsoidal/spherical coordinate system with a non-`1.0` value it would store the units correctly, but if you followed that with a call to `GetLinearUnits` it would return `1.0`. However, this wasn't a problem with `GetTargetLinearUnits` because it would return the cached result.

I tested with the following code:

```
#include <gdal.h>
#include <ogr_spatialref.h>
#include <stdio.h>

int main() {
    {
        OGRSpatialReference srs;
        OGRErr err0 = srs.SetFromUserInput("EPSG:4979");
        OGRErr err1 = srs.SetLinearUnits("unnamed", 0.5);
        const double units = srs.GetLinearUnits();
        printf("Value: %f\n", units);
    }

    {
        OGRSpatialReference srs;
        OGRErr err0 = srs.SetFromUserInput("EPSG:4979");
        OGRErr err1 = srs.SetLinearUnits("unnamed", 0.5);
        const double units = srs.GetTargetLinearUnits("GEOGCS");
        printf("Value: %f\n", units);
    }
}
```

Before this PR: the first code block prints `1.0`. The second prints `0.5`.
After this PR: they both print `0.5`.

There's no mention of `GEOGCS` in the `GetLinearUnits` or `SetLinearUnits` docs, so I'm not sure if this omission is intentional. Thanks for reviewing and let me know if there is anything I should change.

None

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:

### DIFF
--- a/autotest/osr/osr_basic.py
+++ b/autotest/osr/osr_basic.py
@@ -1331,6 +1331,20 @@ def test_osr_basic_25():
     assert sr2 is None
 
 ###############################################################################
+# EPSG 4979, but overridden to be in Feet.
+
+
+def test_osr_basic_26():
+
+    srs = osr.SpatialReference()
+
+    srs.ImportFromEPSG(4979)
+
+    srs.SetLinearUnits('Foot', 0.3048)
+
+    assert srs.GetLinearUnits() == 0.3048
+
+###############################################################################
 # Test corner cases of osr.SetGeocCS()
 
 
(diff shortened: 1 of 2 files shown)


rouault (commented):
It could be useful to add in the doc of the method that for a Geographic 3D CRS, this will return the unit of the vertical axis / ellipsoidal height.
A test exercising this method on EPSG:4979 in autotest/osr/osr_basic.py would be useful

IanLilleyT:
I updated the PR with a test and added more details to the doc. Let me know if anything needs to be tweaked. Thanks!

PR #4043, opened 2021-06-28
GPKG add error message when prepare fails

elpaso:


rouault:
In which circumstance did you get an error during preparation ?

elpaso:
While testing https://github.com/qgis/QGIS/issues/43959#issuecomment-869718996


The error reported before my patch was:

```
OGR error creating feature -4: failed to prepare SQL: INSERT INTO "testmn" ( "fid", "geometry", "id") VALUES (?, ?, ?)
```


The error reported after my patch is:
```
OGR error creating feature -4: failed to prepare SQL: INSERT INTO "testmn" ( "fid", "geometry", "id") VALUES (?, ?, ?) - foreign key mismatch - "adressen_projekte" referencing "testmn"
```

I think it's a useful piece of information to be passed to the client code.

PR #4048, opened 2021-06-30
Also copy RPC metadata if it exists when creating a derived dataset

underchemist:
## What does this PR do?
if `RPC` metadata domain exists in the source dataset, ensure it is copied over to the derived dataset. Not sure what other domains would be appropriate to also pull into derived datasets in general.
## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [x] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:

### DIFF
--- a/gdal/frmts/derived/deriveddataset.cpp
+++ b/gdal/frmts/derived/deriveddataset.cpp
@@ -139,18 +139,11 @@ GDALDataset * DerivedDataset::Open( GDALOpenInfo * poOpenInfo )
     // Transfer metadata
     poDS->SetMetadata(poTmpDS->GetMetadata());
 
-    char ** papszMDList = poTmpDS->GetMetadataDomainList();
-
-    for( char** papszMDListIter = papszMDList;
-            papszMDListIter && *papszMDListIter;
-            ++papszMDListIter )
+    char** papszRPC = poTmpDS->GetMetadata("RPC");
+    if( papszRPC )
     {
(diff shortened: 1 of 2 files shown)


gdal/frmts/derived/deriveddataset.cpp
```diff
@@ -139,6 +139,19 @@ GDALDataset * DerivedDataset::Open( GDALOpenInfo * poOpenInfo )
     // Transfer metadata
     poDS->SetMetadata(poTmpDS->GetMetadata());
 
+    char ** papszMDList = poTmpDS->GetMetadataDomainList();
```
rouault, line 142: this could be simplified as:
```
CSLConstList papszRPC = poTmpDS->GetMetadata("RPC");
if( papszRPC )
   poDS->SetMetadata(papszRPC, "RPC");
```

underchemist: Sorry, I get
```
C:\Users\yanns\git\gdal\gdal\frmts\derived\deriveddataset.cpp(145): error C2664: 'CPLErr VRTDataset::SetMetadata(char **,const char *)': cannot convert argument 1 from 'CSLConstList' to 'char **'
```

rouault: ah right, forgot that SetMetadata() had a non const strict signature. Use char** instead of CSLConstList then

PR #4049, opened 2021-07-01
VRT: Support named arguments to C++ derived band pixel functions

dbaston:
## What does this PR do?

This commit introduces a new pixel function signature,
GDALDerivedPixelFuncWithArgs, that extends GDALDerivedPixelFunc with an
additional argument of named string arguments. These arguments can be
specified using the PixelFunctionArguments tag already used for pixel
functions defined in Python. It is the responsibility of the pixel
function to check the validity of its own arguments.

Not aware of any.

## Tasklist

 - [X] Add test case(s)
 - [x] Add documentation
 - [ ] Review
 - [x] Adjust for comments

### DIFF
--- a/gdal/frmts/vrt/pixelfunctions.cpp
+++ b/gdal/frmts/vrt/pixelfunctions.cpp
@@ -876,7 +876,7 @@ static CPLErr dB2PowPixelFunc( void **papoSources, int nSources, void *pData,
 static CPLErr PowPixelFunc( void **papoSources, int nSources, void *pData,
                                int nXSize, int nYSize,
                                GDALDataType eSrcType, GDALDataType eBufType,
-                               int nPixelSpace, int nLineSpace, char **papszArgs ) {
+                               int nPixelSpace, int nLineSpace, CSLConstList papszArgs ) {
     /* ---- Init ---- */
     if( nSources != 1 ) return CE_Failure;
     if( GDALDataTypeIsComplex( eSrcType ) ) return CE_Failure;
--- a/gdal/frmts/vrt/pixelfunctions.cpp
+++ b/gdal/frmts/vrt/pixelfunctions.cpp
@@ -976,7 +976,7 @@ CPLErr GDALRegisterDefaultPixelFunc()
     GDALAddDerivedBandPixelFunc("dB", DBPixelFunc);
     GDALAddDerivedBandPixelFunc("dB2amp", dB2AmpPixelFunc);
     GDALAddDerivedBandPixelFunc("dB2pow", dB2PowPixelFunc);
-    GDALAddDerivedBandPixelFuncWithArgs("pow", PowPixelFunc);
+    GDALAddDerivedBandPixelFuncWithArgs("pow", PowPixelFunc, nullptr);
 
     return CE_None;
 }
--- a/gdal/frmts/vrt/pixelfunctions.cpp
+++ b/gdal/frmts/vrt/pixelfunctions.cpp
@@ -882,11 +882,19 @@ static CPLErr PowPixelFunc( void **papoSources, int nSources, void *pData,
     if( GDALDataTypeIsComplex( eSrcType ) ) return CE_Failure;
 
     const char *pszPower = CSLFetchNameValue(papszArgs, "power");
-    if ( pszPower == nullptr ) return CE_Failure;
+    if ( pszPower == nullptr )
+    {
+        CPLError(CE_Failure, CPLE_AppDefined, "Missing pixel function argument: power");
+        return CE_Failure;
+    }
 
     char *end = nullptr;
     double power = std::strtod(pszPower, &end);
-    if ( end == pszPower ) return CE_Failure;
+    if ( end == pszPower )
+    {
+        CPLError(CE_Failure, CPLE_AppDefined, "Failed to parse pixel function argument: power");
+        return CE_Failure;
+    }
 
     /* ---- Set pixels ---- */
     for( int iLine = 0, ii = 0; iLine < nYSize; ++iLine ) {
--- a/gdal/frmts/vrt/pixelfunctions.cpp
+++ b/gdal/frmts/vrt/pixelfunctions.cpp
@@ -873,6 +873,40 @@ static CPLErr dB2PowPixelFunc( void **papoSources, int nSources, void *pData,
                               nPixelSpace, nLineSpace, 10.0, 10.0);
 }  // dB2PowPixelFunc
 
+static CPLErr PowPixelFunc( void **papoSources, int nSources, void *pData,
+                               int nXSize, int nYSize,
+                               GDALDataType eSrcType, GDALDataType eBufType,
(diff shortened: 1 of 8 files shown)


gdal/frmts/vrt/pixelfunctions.cpp
```diff
@@ -873,6 +873,40 @@ static CPLErr dB2PowPixelFunc( void **papoSources, int nSources, void *pData,
                               nPixelSpace, nLineSpace, 10.0, 10.0);
 }  // dB2PowPixelFunc
 
+static CPLErr PowPixelFunc( void **papoSources, int nSources, void *pData,
+                               int nXSize, int nYSize,
+                               GDALDataType eSrcType, GDALDataType eBufType,
+                               int nPixelSpace, int nLineSpace, char **papszArgs ) {
```
rouault, line 879:
```suggestion
                               int nPixelSpace, int nLineSpace, CSLConstList papszArgs ) {
```

gdal/frmts/vrt/vrtdataset.h
```diff
@@ -761,6 +762,8 @@ class CPL_DLL VRTDerivedRasterBand CPL_NON_FINAL: public VRTSourcedRasterBand
     char *pszFuncName;
     GDALDataType eSourceTransferType;
 
+    using PixelFunc = std::function<CPLErr(void**, int, void*, int, int, GDALDataType, GDALDataType, int, int, char**)>;
```
rouault, line 765:
```suggestion
    using PixelFunc = std::function<CPLErr(void**, int, void*, int, int, GDALDataType, GDALDataType, int, int, CSLConstList)>;
```

gdal/frmts/vrt/vrtderivedrasterband.cpp
```diff
@@ -245,6 +245,44 @@ GDALAddDerivedBandPixelFunc( const char *pszFuncName,
       return CE_None;
     }
 
+    osMapPixelFunction[pszFuncName] = [pfnNewFunction](void **papoSources, int nSources, void *pData,
+                                         int nBufXSize, int nBufYSize,
+                                         GDALDataType eSrcType, GDALDataType eBufType,
+                                         int nPixelSpace, int nLineSpace, char** papszFunctionArgs) {
```
rouault, line 251:
```suggestion
                                         int nPixelSpace, int nLineSpace, CSLConstList papszFunctionArgs) {
```

gdal/gcore/gdal.h
```diff
@@ -915,6 +915,15 @@ typedef CPLErr
...
                         int nPixelSpace, int nLineSpace);
 
+/** Type of functions to pass to GDALAddDerivedBandPixelFuncWithArgs.
+ * @since GDAL 3.4 */
+typedef CPLErr
+(*GDALDerivedPixelFuncWithArgs)(void **papoSources, int nSources, void *pData,
+                                int nBufXSize, int nBufYSize,
+                                GDALDataType eSrcType, GDALDataType eBufType,
+                                int nPixelSpace, int nLineSpace,
+                                char** papszFunctionArgs);
```
rouault, line 925:
```suggestion
                                CSLConstList papszFunctionArgs);
```

gdal/frmts/vrt/vrtderivedrasterband.cpp
```diff
@@ -245,6 +245,44 @@ GDALAddDerivedBandPixelFunc( const char *pszFuncName,
...
+ * Derived bands are stored with only the name of the pixel function
+ * that it will apply, and if a pixel function matching the name is not
+ * found the IRasterIO() call will do nothing.
+ *
+ * @param pszFuncName Name used to access pixel function
+ * @param pfnNewFunction Pixel function associated with name.  An
+ *  existing pixel function registered with the same name will be
+ *  replaced with the new one.
+ *
+ * @return CE_None, invalid (NULL) parameters are currently ignored.
```
rouault, line 274:
```suggestion
 * @return CE_None, invalid (NULL) parameters are currently ignored.
 * @since GDAL 3.4
```

rouault (commented):
gdal/doc/source/drivers/raster/vrt.rst should be enhanced to mention this new capability in general for C functions, and mention the ``pow`` function with its named argument in particular

gdal/frmts/vrt/pixelfunctions.cpp
```diff
@@ -873,6 +873,40 @@ static CPLErr dB2PowPixelFunc( void **papoSources, int nSources, void *pData,
...
+    /* ---- Init ---- */
+    if( nSources != 1 ) return CE_Failure;
+    if( GDALDataTypeIsComplex( eSrcType ) ) return CE_Failure;
+
+    const char *pszPower = CSLFetchNameValue(papszArgs, "power");
+    if ( pszPower == nullptr ) return CE_Failure;
```
rouault, line 885: an explicit CPLError(CE_Failure, CPLE_AppDefined, ...) error message would be appropriate

rouault:
Loud thinking (I don't mean this is something that should be implemented): I was wondering if we shouldn't have a way during the registration of a new pixel function that can take named arguments a specification of those argument names, which arguments are legal, compulsory / optional and their description (e.g using some custom XML similar to how we declare creation options in driver). But that would only be useful if we had a method to return the registered functions, so that a UI could propose that to users.

dbaston:
> I was wondering if we shouldn't have a way during the registration of a new pixel function that can take named arguments a specification of those argument names

Maybe I should add an argument (`const char* pszMetadata` ?) to `GDALAddDerviedBandPixelFuncWithArgs` to leave a hook for such a feature in the future?

rouault:
> Maybe I should add an argument (`const char* pszMetadata` ?) to `GDALAddDerviedBandPixelFuncWithArgs` to leave a hook for such a feature in the future?

Yes, makes sense

mmomtchev:
@dbaston What are your plans for `pszMetadata`? I will probably hijack it to implement a context data structure, because with the current implementation I can't create dynamic functions that perform multiple operations with a single entry point - for example calling JS or ExprTk or something else

rouault:
> What was the initial idea?

What I mentioned in https://github.com/OSGeo/gdal/pull/4049#issuecomment-872090006. Serialized XML like driver creation options to document extra arguments that pixel functions can take

mmomtchev:
Ok, you mean enumerating the `PixelFunctionArguments` in a way similar to this: <https://github.com/OSGeo/gdal/blob/6e6aff451dbcde450f051bff2f2e75ce6a4a3e6f/frmts/netcdf/netcdfdataset.cpp#L9750>
And eventually, also the number and type of mandatory pixel arguments.
Ok, I don't see any elegant way of combining those two features.

I see 3 options for the dynamic functions:
* as GDAL internally uses `std::function`, there is a very elegant solution with capturing lambdas, but it won't have plain C support
* an artificial built-in `kwarg` with the function name will achieve the same without changing any APIs
* or, an additional classical `void*` to be passed back to the function

PR #4053, opened 2021-07-01
gdal_polygonize.py: fix output driver guessing

rouault:


idanmiara:
@rouault I'll do it tomorrow :)

idanmiara (approved):
Looks good. 

I would assign `out = 'tmp/out.geojson'` and use it instead of hardcoding it multiple times.
Same applies also for the other tests in this unit.

I see that you used `gdal.Unlink` instead of `drv.DeleteDataSource`, you might want to do the same with tests 1, 3.

Thank you for asking me to review :)

PR #4074, opened 2021-07-06
GTiff: set color interpretation on overview bands (fixes #3939)

### LINKED ISSUE #3939: Second band in overviews corrupted in multiband + MINISBLACK
In the example below, the second band of the overviews get corrupted. This happens when there is more than 1 overview. During the creation of the overviews the second band is assumed to be Alpha band, while it is a normal band.

I think the problem is that the method `GDALRegenerateCascadingOverviews` is using the previous overview as base band:
```C++
        GDALRasterBand *poBaseBand = poSrcBand;
        if( i != 0 )
            poBaseBand = papoOvrBands[i-1];
```
to call later to `GDALRegenerateOverviews`.

The overviews themselves have the color interpretation "deduced" from the tiff tags, not using the actual values of the band. In this case, due to `PHOTOMETRIC=MINISBLACK` it is assuming that the second band is alpha in the constructor of the `GTiffRasterBand` class:
```C++
            if( nBand > nBaseSamples
                && nBand-nBaseSamples-1 < count
                && (v[nBand-nBaseSamples-1] == EXTRASAMPLE_ASSOCALPHA
                    || v[nBand-nBaseSamples-1] == EXTRASAMPLE_UNASSALPHA) )
                m_eBandInterp = GCI_AlphaBand;
            else
                m_eBandInterp = GCI_Undefined;
```

As a possible solution I added
```C++
  poOverview->SetColorInterpretation(poBand->GetColorInterpretation());
```
while creating the overviews just before calling to `GDALRegenerateOverviews` in the function `GTiffDataset::IBuildOverviews`, in https://github.com/OSGeo/gdal/blob/master/gdal/frmts/gtiff/geotiff.cpp#L10815 (after `SetNoDataValue`)

However, I tried the same in the code before calling `GDALRegenerateOverviewsMultiBand` https://github.com/OSGeo/gdal/blob/master/gdal/frmts/gtiff/geotiff.cpp#L10744 (I had that problem also fixing that method in #3932, but I used another workaround) and the unit tests never finish (infinite loop?)

While `SetNoDataValue` does minimal changes, `SetColorInterpretation` does deeper changes, like
```C++
    m_poGDS->m_bNeedsRewrite = true;
    m_poGDS->m_bMetadataChanged = true;
```
That seems to have consequences later.
So I do not know if that is the proper solution.

(I am afraid of editing the constructor of `GTiffRasterBand`. It may impact many things).

If you change the inverleave to `PIXEL`, it uses another workflow (via `GDALRegenerateOverviewsMultiBand` ) that works after #3932

How can I fix this?

Kind of unit test

```Python
def test_tiff_ovr_nodata_multiband_interleave_band():

    numpy = pytest.importorskip('numpy')
    nodatavalue = -10000
    data = numpy.array([
        [0.5, 0.2, 0.5, 0.2], 
        [0.2, nodatavalue, 0.2, nodatavalue], 
        [0.5, 0.2, nodatavalue, nodatavalue], 
        [0.2, nodatavalue, nodatavalue, nodatavalue]])
    numbands = 5
    
    temp_path = '/vsimem/test.tif'
    ds = gdal.GetDriverByName('GTiff').Create(
        temp_path, 4, 4, numbands, gdal.GDT_Float32, 
        options=['INTERLEAVE=BAND', 'PHOTOMETRIC=MINISBLACK', 'ALPHA=YES'])
    for i in range(1, numbands):
        ds.GetRasterBand(i).SetColorInterpretation(gdal.GCI_GreenBand)
        ds.GetRasterBand(i).SetNoDataValue(nodatavalue)
        ds.GetRasterBand(i).WriteArray(data)

    ds.GetRasterBand(numbands).SetColorInterpretation(gdal.GCI_AlphaBand)
    ds.GetRasterBand(numbands).SetNoDataValue(nodatavalue)
    ds.GetRasterBand(numbands).WriteArray(numpy.array([
        [255, 255, 255, 255], 
        [255, 255, 255, 255], 
        [255, 0, 0, 0], 
        [0, 0, 0, 0]]))

    ds.FlushCache()
    ds.BuildOverviews('AVERAGE', overviewlist=[2, 4])
    ds.FlushCache()

    assert ds.GetRasterBand(1).GetOverviewCount() == 2, 'Overview could not be generated'

    for i in range(1, numbands):
        pix = ds.GetRasterBand(i).GetOverview(1).ReadAsArray(win_xsize=1, win_ysize=1) 
        assert abs(pix[0,0] - 0.3) < 0.01, 'Error in band ' + str(i)

    pix = ds.GetRasterBand(numbands).GetOverview(1).ReadAsArray(win_xsize=1, win_ysize=1) 
    assert pix[0,0] == 255, 'Error in alpha band '
    ds = None
```

## Operating system

Ubuntu 20.04 64 bit

master

rouault:


tbonfort:
Out of curiosity, what's the use case for this? Shouldn't the colorinterpretation only be defined at the main ifd level and be inherited by overviews?

tbonfort:
> Out of curiosity, what's the use case for this? Shouldn't the colorinterpretation only be defined at the main ifd level and be inherited by overviews?

Ah sorry, I hadn't seen the link to the underlying issue

rouault:
> Shouldn't the colorinterpretation only be defined at the main ifd level and be inherited by overviews?

that's more or less what is done in this commit (with extra complication for external overviews where you need to serialize the color interpretation of the first level of overview)

PR #4079, opened 2021-07-07
Geosversion

mmomtchev:
## What does this PR do?

Add `OGRGetGEOSVersion()` with identical semantics to `OSRGetPROJVersion()`

## Tasklist

 - [x] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/gdal/ogr/ogr_api.cpp
+++ b/gdal/ogr/ogr_api.cpp
@@ -32,7 +32,6 @@
 #include "ogr_api.h"
 
 #include <cstddef>
-#include <regex>
 
 #include "cpl_error.h"
 #include "ogr_geometry.h"
@@ -67,7 +66,7 @@ bool OGRGetGEOSVersion(int *pnMajor, int *pnMinor, int *pnPatch) {
     return TRUE;
 }
 #else
-bool OGRGetGEOSVersion(int *pnMajor, int *pnMinor, int *pnPatch) {
+bool OGRGetGEOSVersion(int *, int *, int *) {
     return FALSE;
 }
 #endif
--- a/gdal/ogr/ogr_api.cpp
+++ b/gdal/ogr/ogr_api.cpp
@@ -55,28 +55,20 @@ static bool bNonLinearGeometriesEnabled = true;
  * @since GDAL 3.4.0
  */
 #ifdef HAVE_GEOS
-void OGRGetGEOSVersion(int *pnMajor, int *pnMinor, int *pnPatch) {
-    std::string verstring = GEOSversion();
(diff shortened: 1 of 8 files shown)


gdal/ogr/ogr_api.cpp
```diff
@@ -32,14 +32,53 @@
...
 
+/************************************************************************/
+/*                         OGRGetGEOSVersion()                          */
+/************************************************************************/
+
+/** \brief Get the GEOS version
+ *
+ * @param pnMajor Pointer to major version number, or NULL
+ * @param pnMinor Pointer to minor version number, or NULL
+ * @param pnPatch Pointer to patch version number, or NULL
```
rouault, line 53:
```suggestion
 * @param pnPatch Pointer to patch version number, or NULL
 * @return TRUE if GDAL is built against GEOS
```
and make changes accordingly

gdal/ogr/ogr_api.cpp
```diff
@@ -32,14 +32,53 @@
...
+ * @since GDAL 3.4.0
+ */
+#ifdef HAVE_GEOS
+void OGRGetGEOSVersion(int *pnMajor, int *pnMinor, int *pnPatch) {
+    std::string verstring = GEOSversion();
+    auto tokenizer = std::regex("\\.");
```
rouault, line 59: can use use instead to avoid std::regex ?
``const CPLStringList aosTokens(CSLTokenizeString2(verstring.c_str(), ".", 0))``

gdal/ogr/ogr_api.cpp
```diff
@@ -32,14 +32,46 @@
...
+        *pnMajor = std::stoi(aosTokens[0]);
+    if (pnMinor && aosTokens.size() > 1)
+        *pnMinor = std::stoi(aosTokens[1]);
+    if (pnPatch && aosTokens.size() > 2)
+        *pnPatch = std::stoi(aosTokens[2]);
+    return TRUE;
+}
+#else
+bool OGRGetGEOSVersion(int *pnMajor, int *pnMinor, int *pnPatch) {
+    return FALSE;
```
rouault, line 71: Setting the pointers to 0 as you did previously would seem appropriate

gdal/ogr/ogr_api.cpp
```diff
@@ -32,14 +32,46 @@
 #include "ogr_api.h"
 
 #include <cstddef>
+#include <regex>
```
rouault, line 35: Can be removed

PR #4090, opened 2021-07-09
VRT: Add interpolation pixel functions

dbaston:
## What does this PR do?

This PR adds pixel functions to interpolate between a set of raster sources representing evenly-spaced positions along a 3rd dimension such as time or elevation. Interpolation can be either piecewise linear or piecewise exponential. Example applications include linearly interpolating among hourly temperature grids or exponentially interpolating among decadal population grids.

None

## Tasklist

 - [X] Add test case(s)
 - [X] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:

### DIFF
--- a/gdal/doc/source/drivers/raster/vrt.rst
+++ b/gdal/doc/source/drivers/raster/vrt.rst
@@ -267,7 +267,7 @@ The allowed subelements for VRTRasterBand are :
     <Category>Soybeans</Category>
   </CategoryNames>
 
-- **SimpleSource**: The SimpleSource_ indicates that raster data should be read from a separate dataset, indicating the dataset, and band to be read from, and how the data should map into this bands raster space.
+- **SimpleSource**: The SimpleSource_ indicates that raster data should be read from a separate dataset, indicating the dataset, and band to be read from, and how the data should map into this band's raster space.
 
 - **AveragedSource**: The AveragedSource is derived from the SimpleSource and shares the same properties except that it uses an averaging resampling instead of a nearest neighbour algorithm as in SimpleSource, when the size of the destination rectangle is not the same as the size of the source rectangle. Note: a more general mechanism to specify resampling algorithms can be used. See above paragraph about the 'resampling' attribute.
 
@@ -293,7 +293,7 @@ filename should be interpreted as relative to the .vrt file (value is 1)
 or not relative to the .vrt file (value is 0).  The default is 0.
 
 Some characteristics of the source band can be specified in the optional
-SourceProperties tag to enable the VRT driver to differ the opening of the source
+SourceProperties tag to enable the VRT driver to defer the opening of the source
 dataset until it really needs to read data from it. This is particularly useful
 when building VRTs with a big number of source datasets. The needed parameters are the
 raster dimensions, the size of the blocks and the data type. If the SourceProperties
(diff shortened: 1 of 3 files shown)


gdal/frmts/vrt/pixelfunctions.cpp
```diff
@@ -915,6 +924,79 @@ static CPLErr PowPixelFunc( void **papoSources, int nSources, void *pData,
...
+// a value for x.
+static std::size_t intervalLeft(double x0, double dx, std::size_t nx, double x)
+{
+    if (x < x0) {
+        return 0;
+    }
+
+    std::size_t n = static_cast<std::size_t>((x - x0) / dx);
+
+    if (n >= nx) {
```
rouault, line 938: shouldn't that be 
```suggestion
    if (n >= nx - 1) {
```

gdal/frmts/vrt/pixelfunctions.cpp
```diff
@@ -915,6 +924,79 @@ static CPLErr PowPixelFunc( void **papoSources, int nSources, void *pData,
...
+// Given nx intervals spaced by dx and beginning at x0, return the the index of
+// the lower bound of the interval that should be used to interpolate/extrapolate
+// a value for x.
+static std::size_t intervalLeft(double x0, double dx, std::size_t nx, double x)
+{
+    if (x < x0) {
+        return 0;
+    }
+
+    std::size_t n = static_cast<std::size_t>((x - x0) / dx);
```
rouault, line 936: there's a slight chance that the cast might be invalid in case of dx being very close to 0

gdal/doc/source/drivers/raster/vrt.rst
```diff
@@ -796,24 +796,95 @@ Default Pixel Functions
...
+     - 1
+     - -
+     - extract imaginary part from a single raster band (0 for non-complex)
+   * - **intensity**
+     - 1
+     - -
+     - computes the intensity ``Re(x*conj(x))`` of a single raster band (real or complex)
+   * - **interpolate_exp**
+     - >= 2
+     - ``x0``, ``dx``, ``x``
```
rouault, line 846: My mind had some trouble with the x naming, confusing it with the horizontal axis. Perhaps ``t`` would make it clearer than this is not interpolation along an axis of the 2D raster plan ? (if so, the change should be made in the code too)

gdal/frmts/vrt/pixelfunctions.cpp
```diff
@@ -915,6 +924,79 @@ static CPLErr PowPixelFunc( void **papoSources, int nSources, void *pData,
...
+    }
+
+    double dfX0;
+    if ( FetchDoubleArg(papszArgs, "x0", &dfX0) == CE_Failure ) return CE_Failure;
+
+    double dfX;
+    if ( FetchDoubleArg(papszArgs, "x", &dfX) == CE_Failure ) return CE_Failure;
+
+    double dfDx;
+    if ( FetchDoubleArg(papszArgs, "dx", &dfDx) == CE_Failure ) return CE_Failure;
```
rouault, line 974: dx should likely be tested against 0

gdal/frmts/vrt/pixelfunctions.cpp
dbaston: I can check for that and handle it. Are you opposed to pixel functions throwing exceptions that are caught and converted to CPL errors at the single call site? I think that would keep the error handling cleaner.

gdal/frmts/vrt/pixelfunctions.cpp
dbaston: Yes, thank you

gdal/frmts/vrt/pixelfunctions.cpp
rouault: >  that are caught and converted to CPL errors at the single call site?

sounds good

gdal/doc/source/drivers/raster/vrt.rst
dbaston: Switched to `t` throughout

Issue #4087, opened 2021-07-08
VERSION file confuses Xcode 12.5


mathstuf:
The compilation on macOS with Xcode 12.5 fails because the standard library has a `#include <version>` in the `<cstddef>` header. This ends up selecting the `VERSION` file in the source tree and doesn't work (as one might expect). Unfortunately, case-insensitivity is not something that Apple is likely to change anytime soon. I think just renaming the file to be `VERSION.gdal` or something to avoid this confusion.

rouault:
Can you should the compilation line(s) that fail ? There's no reason that the root folder is included during compilation, so we should likely fix that

mathstuf:
It seems it comes from an `-I.` which, with the `Ninja` generator is the top-level of the build tree.

```
[...]

PR #4091, opened 2021-07-09
#include local files with -iquote instead of -I

tbonfort:
## What does this PR do?
Replaces all occurences of `-I` in unix build with `-iquote` when referencing
our own directories. Prevents local files overriding system includes, notably
on case-insensitive filesystems.

#4087

https://bugs.llvm.org/show_bug.cgi?id=42540

https://gcc.gnu.org/onlinedocs/gcc/Directory-Options.html

## Environment

* OS: Unixes


tbonfort:
ready for review/merge

rouault:
Thanks. Tested locally with success. I let this briefly open and will merge if nobody has comments

rouault:
The Intel C++ compiler seems to support -iquote as well, so we're good to go :
https://software.intel.com/content/www/us/en/develop/documentation/cpp-compiler-developer-guide-and-reference/top/compiler-reference/compiler-options/compiler-option-details/preprocessor-options/iquote.html

PR #4097, opened 2021-07-09
Add an option to disable the WMS/WMTS cache

mmomtchev:
## What does this PR do?

Add an option to disable the WMS/WMTS cache

For very dynamic data or for environments without access to a file system

## Tasklist

 - [x] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments


autotest/gdrivers/wms.py
```diff
@@ -371,6 +371,15 @@ def test_wms_8():
     gdal.SetConfigOption("GDAL_DEFAULT_WMS_CACHE_PATH", None)
     assert cs == 0, 'cs != 0'
 
+    # Test with GDAL_DEFAULT_WMS_CACHE_TYPE
+    # Now, we should not read from the cache anymore
+    gdal.SetConfigOption("GDAL_DEFAULT_WMS_CACHE_TYPE", "none")
```
rouault, line 376: ``
with gdaltest.config_option("GDAL_DEFAULT_WMS_CACHE_TYPE", "none"):
     ...
`` is better now

rouault (commented):
On reflection, I'm wondering if a GDAL_ENABLE_WMS_CACHE=YES/NO config option wouldn't be better rather than hacking the wms cache type

mmomtchev:
Equivalent to the presence/absence of a `Cache` XML element?

rouault:
> Equivalent to the presence/absence of a `Cache` XML element?

I'd say GDAL_ENABLE_WMS_CACHE=NO would overload an implicit or explicit Cache XML element

gdal/frmts/wms/gdalwmsdataset.cpp
```diff
@@ -245,8 +245,9 @@ CPLErr GDALWMSDataset::Initialize(CPLXMLNode *config, char **l_papszOpenOptions)
     }
 
     if (ret == CE_None) {
+        const char *pszEnableCache = CPLGetConfigOption("GDAL_ENABLE_WMS_CACHE", "YES");
         CPLXMLNode *cache_node = CPLGetXMLNode(config, "Cache");
-        if (cache_node != nullptr) {
+        if (cache_node != nullptr && !EQUAL(pszEnableCache, "NO")) {
```
rouault, line 250:
```suggestion
        if (cache_node != nullptr && CPLTestBool(pszEnableCache)) {
```

PR #4119, opened 2021-07-15
XLS: Set OLCStringsAsUTF8 capability to TRUE

agiudiceandrea:
## What does this PR do?

Sets the OLCStringsAsUTF8 capability to TRUE for XLS layers.

See https://lists.osgeo.org/pipermail/qgis-developer/2021-July/063855.html

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/gdal/ogr/ogrsf_frmts/xls/ogrxlslayer.cpp
+++ b/gdal/ogr/ogrsf_frmts/xls/ogrxlslayer.cpp
@@ -340,6 +340,8 @@ int OGRXLSLayer::TestCapability( const char * pszCap )
 {
     if( EQUAL(pszCap, OLCFastFeatureCount) )
         return m_poAttrQuery == nullptr /* && m_poFilterGeom == NULL */;
+    else if( EQUAL(pszCap, OLCStringsAsUTF8) )
+        return TRUE;
 
     return FALSE;
 }


nyalldawson:
I  don't think that's completely correct -- looking at the freexl documentation it seems that only some xls file versions are utf8, and others have a codepage indicating the encoding which needs to be read from
the xls metadata:

**Any BIFF version from BIFF2 to BIFF5 simply supports CodePage based character encoding, i.e. each character simply requires 8 bits to be represented (single byte). Correct representation of characters requires knowing which one CodePage table has to be applied. This can be determined from the workbook or worksheet metadata (it is the CODEPAGE record).**

**BIFF8 is much more sophisticated, since any text string is usually encoded as Unicode in UTF-16 Little Endian
[UTF-16LE] format. This encoding is a multi-byte encoding (two bytes are required to represent a single character), but being universal no character table is required.**

nyalldawson:
(guess I should have left the follow up conversation to the mailing list thread 😆)

rouault:
The recoding to UTF-8 from the source code page is done in the convert_to_utf8 function of https://www.gaia-gis.it/fossil/freexl/file?name=src/freexl.c&ci=tip

PR #4135, opened 2021-07-25
Add GDAL_DCAP_MULTIPLE_VECTOR_LAYERS capability

nyalldawson:
Capability set by drivers for formats which support
multiple vector layers.

Note: some GDAL drivers expose "virtual" layer support
while the underlying formats themselves do not. This
capability is only set for drivers of formats which have a native
concept of multiple vector layers (such as GeoPackage).

Refs https://github.com/qgis/QGIS/pull/44291#discussion_r674390840


rouault:
Other drivers that could expose the capability (likely missing some): ODS, GML, GMLAS, Memory, PDF, PostgreSQL, OCI, OGDI, NGW, MySQL, MongoDBv3, MSSQLSpatial

It would be nice to expose the capability item in swig/include/ogr.i

nyalldawson:
@rouault  done

PR #4136, opened 2021-07-26
gdal_calc.py - fix `hideNoData` for input: `NoDataValue!='none'`, `hideNoData=True`

idanmiara:
## What does this PR do?

gdal_calc.py - fix `hideNoData` for input: `NoDataValue!='none'`, `hideNoData=True`

This fix is for the case that the user wants to set ndv to the output ds and also wants the calc operation to ignore the input ndv while performing the calculation itself.

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments


rouault:
@idanmiara I let you do the manual backport PR

idanmiara:
> @idanmiara I let you do the manual backport PR

Thanks!

PR #4138, opened 2021-07-26
Add STACIT (Spatio-Temporal Asset Catalog Items) raster driver 

rouault:
This driver supports opening JSON files, or usually the result of a remote query, that are the ``items`` link of a
[STAC Collection](https://github.com/radiantearth/stac-api-spec/blob/master/stac-spec/collection-spec/collection-spec.md), and whose items also implement the [Projection Extension Specification](https://github.com/stac-extensions/projection/). It builds a virtual mosaic from the items.

On top of https://github.com/OSGeo/gdal/pull/4130

### DIFF
--- a/autotest/cpp/test_cpl.cpp
+++ b/autotest/cpp/test_cpl.cpp
@@ -45,6 +45,7 @@
 #include "cpl_http.h"
 #include "cpl_auto_close.h"
 #include "cpl_minixml.h"
+#include "cpl_quad_tree.h"
 #include "cpl_worker_thread_pool.h"
 
 #include <fstream>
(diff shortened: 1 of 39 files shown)


matthewhanson:
@rouault  This is great! just got one piece of feedback.  In the cases that an EPSG code is not available, the projection extension has a `proj:wkt2` field (also a proj:projjson` field). Could this be updated to respect `wkt2` if `proj:epsg` is null?

The new Landsat C2 Albers scenes will specify wkt2 because there's no valid EPSG for WGS-84 CONUS Albers.   The full catalog has not been made publicly available yet but an example is here:

https://74kin890wi.execute-api.us-west-2.amazonaws.com/st/collections/landsat-c2l2alb-st/items/LE07_L2SP_012031_20201207_20210103_02_A1_ST

rouault:
> `Could this be updated to respect`wkt2`if`proj:epsg` is null?

Thanks for the valuable feedback. Base support for that was a bit in place but untested and it required some adjustment . I also fixed the support for proj:transform to accept just 6 values instead of 9. I couldn't fully test the support for WKT2 because st-usgs-landsat-dds.qa-lsdsdpas.chs.usgs.gov is a unknown host, but that should be hopefully be OK

matthewhanson:
awesome thanks for the great work @rouault 

Yeah, unfortunately the data in that catalog isn't available right now. 

I was wondering about the geotransform - noticing some people have 9 elements in their STAC Items due to using rasterio dataset.transform as is, but I've been truncating mine to the 6 elements but there's still likely some catalogs out there using 9.  Will this support either 6 or 9?

rouault:
> Will this support either 6 or 9?

yes, both variants are now supported

alexgleith:
This is fantastic, @rouault! I hope it pushes more people and organisations towards a common standard implementation of STAC :-)

cpaulik:
This looks really cool. 

Would this also support overviews as a regular VRT does?

rouault:
> Would this also support overviews as a regular VRT does?

STACIT datasets will not expose explicit overviews, as this would require opening each source to see if it has overviews. But if you downsample the STACIT dataset and the sources have overviews, they will be used

cpaulik:
Thanks for the clarification.

Does that also mean that a STAC Collection with a [collection level asset](https://github.com/radiantearth/stac-api-spec/blob/master/stac-spec/collection-spec/collection-spec.md#assets) which has the [Asset Role](https://github.com/radiantearth/stac-api-spec/blob/master/stac-spec/item-spec/item-spec.md#asset-role-types) `overview` which would link to e.g. an `.ovr` file would not be used?

rouault:
> Does that also mean that a STAC Collection with a [collection level asset](https://github.com/radiantearth/stac-api-spec/blob/master/stac-spec/collection-spec/collection-spec.md#assets) which has the [Asset Role](https://github.com/radiantearth/stac-api-spec/blob/master/stac-spec/item-spec/item-spec.md#asset-role-types) `overview` which would link to e.g. an `.ovr` file would not be used?

This is currently not implemented indeed. I guess it could be (the driver would have to accept to open documents that are of type Collection, but that's a minor point). Examples would be welcome. But I'd see that more as a further enhancement for a dedicated ticket

cpaulik:
Thank you for the answers.

> Examples would be welcome. But I'd see that more as a further enhancement for a dedicated ticket

More than fair. I do hope to soon have time to test replacing our current VRT + OVR setup by STAC. Then we will probably run into this again.

PR #4139, opened 2021-07-26
Enable JPEG-in-TIFF for 12 bit samples (for internal libtiff), whatever libjpeg is used for 8 bit sample support

rouault:
Previously we required that the libjpeg used for 8 bit sample support
had the libjpeg-6b ABI. The changes done in libtiff/tif_jpeg.c
and tif_jpeg_12.c now remove this constraint, and it is thus now
possible to have JPEG 12 bit support in a build that uses libjpeg-turbo
with IJG v8 ABI for 8-bit.

### DIFF
--- a/gdal/configure
+++ b/gdal/configure
@@ -31787,38 +31787,22 @@ if test "$with_jpeg12" =  no ; then
     { $as_echo "$as_me:${as_lineno-$LINENO}: result: disabled by user" >&5
 $as_echo "disabled by user" >&6; }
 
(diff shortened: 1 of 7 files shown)


rouault:
Note: I've also submitted a pull request to libjpeg-turbo (https://github.com/libjpeg-turbo/libjpeg-turbo/pull/540) so that libjpeg-turbo 12 bit builds (which don't use SIMD accelerated code paths) can have their symbols suffixed, like our internal patched libjpeg-6b 12-bit. If it was accepted, then GDAL build scripts could be enhanced to use that external libjpeg-turbo 12bit

PR #4153, opened 2021-07-29
MSG: fix race condition in satellite number lookup

jdroenner:
## What does this PR do?
This PR fixes a race condition in the MSG driver.

The MSG driver looks for a matching "prologue"-file when calling `MSGDataset::Open`. The filename depends on the satellite number. There are only 4 satellites and all of them are tried until an existing prologue is found. The satellite number is stored as a static variable `iCurrentSatellite` and each call to `Open` triggers a loop that modifies this variable. If multiple threads try to open a MSG dataset, the state of `iCurrentSatellite` is modified in all threads.

This PR changes `iCurrentSatellite` to be a class variable and adds a static hint `iCurrentSatelliteHint` which provides a start value for the lookup. Additionally, the lookup logic (loop) was simplified.

## Environment
To use the MSG driver the _PublicDecompWT_  code from EUMETSAT is required. It is available at: https://gitlab.eumetsat.int/open-source/PublicDecompWT

### DIFF
--- a/gdal/frmts/msg/msgdataset.cpp
+++ b/gdal/frmts/msg/msgdataset.cpp
@@ -53,7 +53,7 @@ const double MSGDataset::rA[12] = {-1, -1, -1, 0.9959, 0.9963, 0.9991, 0.9996, 0
 const double MSGDataset::rB[12] = {-1, -1, -1, 3.471, 2.219, 0.485, 0.181, 0.060, 0.627, 0.397, 0.576, -1};
 const int MSGDataset::iCentralPixelVIS_IR = 1856; // center pixel VIS and IR
 const int MSGDataset::iCentralPixelHRV = 5566; // center pixel HRV
-int MSGDataset::iCurrentSatellite = 1; // satellite number 1,2,3,4 for MSG1, MSG2, MSG3 and MSG4
+int MSGDataset::iCurrentSatelliteHint = 1;     // satellite number hint 1,2,3,4 for MSG1, MSG2, MSG3 and MSG4
 const char *MSGDataset::metadataDomain = "msg"; // the metadata domain
 
 #define MAX_SATELLITES 4
--- a/gdal/frmts/msg/msgdataset.cpp
+++ b/gdal/frmts/msg/msgdataset.cpp
@@ -151,34 +151,26 @@ GDALDataset *MSGDataset::Open( GDALOpenInfo * poOpenInfo )
 /* -------------------------------------------------------------------- */
     Prologue pp;
 
+    int iCurrentSatellite = iCurrentSatelliteHint; // Start with the hint. It is nice to have but don't rely on it...
+
     std::string sPrologueFileName = command.sPrologueFileName(iCurrentSatellite, 1);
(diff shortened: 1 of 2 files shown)


gdal/frmts/msg/msgdataset.cpp
```diff
@@ -200,9 +192,15 @@ GDALDataset *MSGDataset::Open( GDALOpenInfo * poOpenInfo )
...
     poDS->command = command; // copy it
 
-/* -------------------------------------------------------------------- */
-/*      Capture raster size from MSG prologue and submit it to GDAL     */
-/* -------------------------------------------------------------------- */
+    /* -------------------------------------------------------------------- */
+    /*      Set the current satellite for this DS                           */
+    /* -------------------------------------------------------------------- */
+
+    poDS->iCurrentSatellite = iCurrentSatellite; // copy it
```
ChristianBeilschmidt, line 199: Can't you just use `poDS->iCurrentSatellite` starting from 154?

gdal/frmts/msg/msgdataset.cpp
```diff
@@ -345,31 +343,22 @@ MSGRasterBand::MSGRasterBand( MSGDataset *poDSIn, int nBandIn )
...
-      if (!fPrologueExists) // assume missing prologue file, keep original satellite number
-      {
-        poDSIn->iCurrentSatellite = 1 + poDSIn->iCurrentSatellite % MAX_SATELLITES;
-        sPrologueFileName = poDSIn->command.sPrologueFileName(poDSIn->iCurrentSatellite, nBand);
-      }
+      ++iTries;
     }
 
-    iSatellite = poDSIn->iCurrentSatellite; // From here on, the satellite that corresponds to this band is settled to the current satellite
+    iSatellite = iCurrentSatellite; // From here on, the satellite that corresponds to this band is settled to the current satellite
```
ChristianBeilschmidt, line 361: Same for here, starting in 346. I guess before someone wanted to avoid race conditions (but it didn't work).

gdal/frmts/msg/msgdataset.cpp
```diff
@@ -345,31 +343,22 @@ MSGRasterBand::MSGRasterBand( MSGDataset *poDSIn, int nBandIn )
...
+    int iCurrentSatellite = poDSIn->iCurrentSatellite; // Start with the satellite selected by the dataset
+
+    std::string sPrologueFileName = poDSIn->command.sPrologueFileName(iCurrentSatellite, 1);
     bool fPrologueExists = (access(sPrologueFileName.c_str(), 0) == 0);
 
-    // Make sure we're testing for MSG1,2,3 or 4 exactly once, start with the most recently used, and remember it in the static member for the next round.
-    if (!fPrologueExists)
+    // Make sure we're testing for MSG1,2,3 or 4 exactly once, start with the one of the dataset
+    int iTries = 1;
+    while (!fPrologueExists && (iTries < MAX_SATELLITES))
```
ChristianBeilschmidt, line 353: Isn't this function exactly the same as in the dataset case?
We could extract it and call in both times.

ChristianBeilschmidt (commented):
This looks super awesome, I've got nothing to complain.

PR #4159, opened 2021-08-01
Downloading Gdal 3.3.1 using Homebrew and proj@7

OldSubSailor:
During the downloading process, this warning appears.
Warning: proj@7 has been deprecated because it is not supported upstream!
proj@7 is listed as a dependency, however, I know nothing about these matters other than searches indicate that Gdal is installed on my iMac (Big Sur)
% brew info gdal
gdal: stable 3.3.1 (bottled), HEAD
Geospatial Data Abstraction Library
However, it appears not to be working, and a direct call from python says that the the module is missing. 
I have reported it to Homebrew, and they say to submit a pull request to gdal.  
I feel like I am caught in a Catch 22. 
I don't think that this is a support request, but a concern developer intervention.


rouault:
This is not appropriate as a pull request, and actually not even as a bug report here. This is to report to the homebrew maintainers for GDAL

OldSubSailor:
I said in the pull request that I had and that they said to report it to gdal

jef-n:
> I said in the pull request I had reported it to homebrew and that they said to report it to gdal.

See https://github.com/Homebrew/discussions/discussions/367#discussioncomment-1116033 - by telling you to make a pull requests they implied that you know how to fix the issue, do a GDAL branch, change what needs changing and then ask the project to merge that (which is what a [pull request](https://docs.github.com/en/github/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-pull-requests) is).

BTW maybe you just need to change ```import gdal``` to ```import osgeo.gdal```

PR #4151, opened 2021-07-29
color-palette: fix read/write txt file with ndv + get_raster_min_max fix

idanmiara:
## What does this PR do?

* gdal-utils/osgeo_utils/auxiliary/color_palette.py - fix read/write txt file with ndv, add some type hints
* autotest/pyscripts/test_gdal_utils.py - add test_utils_color_files_nv and fix typo in filename
* autotest/utilities/data/color_paletted_red_green_0-1-nv.txt - rename to reflect the contents
* autotest/utilities/data/colro_pseudocolor_spectral_0-100.[txt|qml] - fix filename typo colro -> color

* gdal-utils/osgeo_utils/auxiliary/color_table.py - color_table_from_color_palette to raise appropriate Exception

* gdal-utils/osgeo_utils/auxiliary/util.py - `get_raster_min_max` - use the dedicated `bnd.ComputeRasterMinMax()` function

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments
 - [ ] All CI builds and checks have passed

PR #4167, opened 2021-08-05
util.py - fix get_raster_min_max() and add a test

idanmiara:
## What does this PR do?

util.py - fix get_raster_min_max() and add a test.

On a side note... the docstring `"""ComputeRasterMinMax(Band self, int approx_ok=0)"""` is somewhat misleading as it suggests that `approx_ok` is the name of the argument, but apparently it does not accept it as a named argument but only as a positional argument. 

https://github.com/OSGeo/gdal/pull/4151 (I should have added the test there... :)

### DIFF
--- a/autotest/pyscripts/test_gdal_utils.py
+++ b/autotest/pyscripts/test_gdal_utils.py
@@ -129,6 +129,12 @@ def test_utils_py_1():
     ds_list = None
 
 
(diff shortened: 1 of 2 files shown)


rouault:
This needs to be backported right ?

idanmiara:
Thanks @rouault!

Issue #4158, opened 2021-08-01
Implement some form of fail-safe mechanism when deleting vsimem files


mmomtchev:
## Expected behavior and actual behavior.

Currently it is all too easy to delete an open vsimem file by calling `VSIGetMemFileBuffer` with `bUnlinkAndSeize=TRUE`
Ideally, it should return `EBUSY` if there is an open dataset

## Steps to reproduce the problem.

The problem is evident in the Node.js bindings where it is (usually) up to the garbage collector to close open datasets

## GDAL version and provenance

3.3.1

mmomtchev:
I am volunteering, but I wonder if this should be limited to vsimem and `VSIGetMemFileBuffer`

rouault:
I don't think this would be a good idea. /vsimem/ aims at honoring POSIX file system policies, and unlinking an open file is something that is definitely permitted (and wishable) in POSIX.

mmomtchev:
Yes but you don't get a kernel panic when closing that file
How about failing further operations gracefully?

rouault:
> Yes but you don't get a kernel panic when closing that file
> How about failing further operations gracefully?

I understand there's a sequence of calls that present some issue for you. What is it exactly ?

mmomtchev:
Not really, I just got burned by a code that did not explicitly close a Dataset obtained by `CreateCopy`. Just like in Python, closing in JS is optional - the garbage collector does it for you. If you always call `close`, there will be no problem. But without it - you get a silent error followed by random memory corruption. Normally, in the safe care bearers world of JS and Python programmers, these should be impossible.

mmomtchev:
@rouault, there is an undocumented `gdal.VSIGetMemFileBuffer_unsafe` in the Python bindings?

rouault:
> there is an undocumented `gdal.VSIGetMemFileBuffer_unsafe` in the Python bindings? What is unsafe about it?

I see this was added in fc81f28ae04b97bc3a14e30ecf53a105eb1065fe. @tzickel could you comment about what is unsafe ?

tzickel:
It's unsafe that the function returns a Python memoryview object to a piece of raw memory which the lifetime isn't controlled by Python's memory management (so if the user isn't careful it might result in memory corruption).

mmomtchev:
@tzickel, Ok, so it is precisely the same problem as the Node.js bindings
[...]

PR #4179, opened 2021-08-06
Delay destroying VSIMemFile on bUnlinkAndSeize

mmomtchev:
## What does this PR do?

It delays destroying the `VSIMemFile` object on `bUnlinkAndSeize` by reusing the same mechanism as `VSIMemFile::Unlink`

https://github.com/OSGeo/gdal/issues/4158

## Tasklist

 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

Creating an unit test requires that the undocumented Python API `gdal.VSIGetMemFileBuffer_unsafe` is replaced by a public `gdal.VSIGetMemFileBuffer` that will use this

### DIFF
--- a/gdal/port/cpl_vsi_mem.cpp
+++ b/gdal/port/cpl_vsi_mem.cpp
@@ -1034,8 +1034,11 @@ GByte *VSIGetMemFileBuffer( const char *pszFilename,
         CPLDebug("VSIMEM", "VSIGetMemFileBuffer() %s: ref_count=%d (before)",
                  poFile->osFilename.c_str(), poFile->nRefCount);
 #endif
-        CPLAtomicDec(&(poFile->nRefCount));
-        delete poFile;
+        poFile->pabyData = nullptr;
+        poFile->nLength = 0;
+        poFile->nAllocLength = 0;
+        if (CPLAtomicDec(&(poFile->nRefCount)) == 0)
+            delete poFile;
     }
 
     return pabyData;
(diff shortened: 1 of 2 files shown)


gdal/port/cpl_vsi_mem.cpp
```diff
@@ -1034,8 +1034,11 @@ GByte *VSIGetMemFileBuffer( const char *pszFilename,
         CPLDebug("VSIMEM", "VSIGetMemFileBuffer() %s: ref_count=%d (before)",
                  poFile->osFilename.c_str(), poFile->nRefCount);
 #endif
-        CPLAtomicDec(&(poFile->nRefCount));
-        delete poFile;
+        poFile->pabyData = nullptr;
```
rouault, line 1037: I believe those 3 lines are useless and could cause memory leaks.

rouault: Forget my above comment.

rouault:
A test in autotest/cpp/test_cpl.cpp doing something like the following could be added:
```
VSILFILE* fp = VSIFOpenL("/vsimem/foo", "wb");
VSIFWriteL("test", 1, 1, fp);
CPLFree(VSIGetMemFileBuffer("/vsimem/foo", nullptr, true));
VSIFCloseL(fp);
```

mmomtchev:
@rouault, hopefully all drivers should behave correctly when they cannot read/write when closing and still close the dataset/free the memory

rouault:
> hopefully all drivers should behave correctly when they cannot read/write when closing and still close the dataset/free the memory

hard to predict how they'll behave, but that fix is definitely the right thing to do

PR #4197, opened 2021-08-11
Fix Vertical Units check in GetProjection()

C41eb:
The GetProjection() function should ignore values that are within the Private User Implementations range rather than throwing an exception. In fact, it does ignore values that are set to a user-defined unit (32767)

### DIFF
--- a/gdal/frmts/gtiff/gt_wkt_srs.cpp
+++ b/gdal/frmts/gtiff/gt_wkt_srs.cpp
@@ -1366,7 +1366,7 @@ OGRSpatialReferenceH GTIFGetOGISDefnAsOSR( GTIF *hGTIF, GTIFDefn * psDefn )
 /* -------------------------------------------------------------------- */
 /*      Set the vertical units.                                         */
 /* -------------------------------------------------------------------- */
-        if( verticalUnits > 0 && verticalUnits != KvUserDefined
+        if( verticalUnits > 0 && verticalUnits < KvUserDefined
             && verticalUnits != 9001 )
         {
             char szCode[12];


gdal/frmts/gtiff/gt_wkt_srs.cpp
```diff
@@ -1366,7 +1366,7 @@ OGRSpatialReferenceH GTIFGetOGISDefnAsOSR( GTIF *hGTIF, GTIFDefn * psDefn )
 /* -------------------------------------------------------------------- */
 /*      Set the vertical units.                                         */
 /* -------------------------------------------------------------------- */
-        if( verticalUnits > 0 && verticalUnits != KvUserDefined
+        if( verticalUnits > 0 && verticalUnits < KvUserDefined
```
rouault, line 1369: if verticalUnits > KvUserDefined, this will now go to the else branch that will assign a metre vertical unit, which is likely incorrect. I'm not sure which behaviour would be appropriate, perhaps not set the vertical CRS at all, and emit a warning ? Can you share a file that triggers this ?

C41eb: I have attached a file with the VerticalUnitsGeoKey set to a private usere implementation value of 32771.

 The primary concern is that with the current implementation throws an exception when it should not.
Furthermore, in the affected files, the VerticalCSTypeGeoKey and VerticalDatumGeoKey are both set to User-Defined, so I think that no vertical information is going to be set in the spatial reference in these cases. 

Maybe as you said not setting the vertical CRS at all could be the correct resolution
[VerticalUnitsGeoKey_32771.zip](https://github.com/OSGeo/gdal/files/6972501/VerticalUnitsGeoKey_32771.zip)

rouault: I've added a commit to your branch to ignore vertical information when one of the tag has a value in the private user range

PR #4218, opened 2021-08-12
ogrdxflayer.cpp: use std::unique_ptr

rouault:


gdal/ogr/ogrsf_frmts/dxf/ogrdxflayer.cpp
```diff
@@ -459,7 +459,7 @@ OGRDXFFeature *OGRDXFLayer::TranslateMTEXT()
 {
     char szLineBuf[512];
     int nCode = 0;
-    OGRDXFFeature *poFeature = new OGRDXFFeature( poFeatureDefn );
+    auto poFeature = std::unique_ptr<OGRDXFFeature>(new OGRDXFFeature( poFeatureDefn ));
```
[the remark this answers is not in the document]
rouault: > Is there a reason not to use `std::make_unique`?

yes, this is C++14, and our base requirement is C++11 currently. The (optional) PDF and MongoDB driver have higher requirements due to the underlying libraries requiring C++14 themselves.

gdal/ogr/ogrsf_frmts/dxf/ogrdxflayer.cpp
```diff
@@ -2190,21 +2143,20 @@ OGRDXFFeature *OGRDXFLayer::TranslateSPLINE()
 /*      Use the helper function to check the input data and insert      */
 /*      the spline.                                                     */
 /* -------------------------------------------------------------------- */
-    OGRLineString *poLS = InsertSplineWithChecks( nDegree,
-        adfControlPoints, nControlPoints, adfKnots, nKnots, adfWeights );
+    auto poLS = std::unique_ptr<OGRLineString>(InsertSplineWithChecks( nDegree,
+        adfControlPoints, nControlPoints, adfKnots, nKnots, adfWeights ));
```
atlight, line 2147: This function is defined in this file so it would make as much sense to change it to return `unique_ptr`. But maybe you want to do that in another commit.

atlight (approved):

gdal/ogr/ogrsf_frmts/dxf/ogrdxflayer.cpp
rouault: yeah, that can be done later. I just wanted to fix a few things on which cppcheck stumbled upon (false positives), and did similar changes through that file.

rouault:
follow-up in https://github.com/OSGeo/gdal/pull/4226 adding a cpl::make_unique<> C++11 compatible equivalent of std::make_unique

PR #4255, opened 2021-08-18
[pgeo] Always promote polygon or line layers to multi-type geometries

nyalldawson:
The Personal Geodatabase format has a similar approach to multi-part handling as Shapefiles, where polygon and multipolygon
geometries or line and multiline geometries will co-exist in a layer reported as just polygon or line type respectively.

To handle this in a predictable way for clients we always promote the polygon/line types to multitypes, and
correspondingly ALWAYS return multi polygon/line geometry objects for features (even if strictly speaking the
original feature had a polygon/line geometry object)

Also update the driver docs to reflect that evanmiller's mdbtools fork is now the official, maintained upstream


rouault:
> which always reports these layers as of single geometry
> type (even though they may contain mixed single/multi type geometries)

We didn't dare to change the behavior of the shapefile driver but for PGeo, I'd think we could probably just advertize the Multi types (ie LineString would be exposed as MultiLineString and Polygon as MultiPolygon.  Point and MultiPoint are likely different types if that's similar to shapefile) as the layer geometry type *and* promote the returned geometries consistently with that.

nyalldawson:
Ok, sounds good to me! I'll update accordingly

nyalldawson:
@rouault ok, updated

gdal/ogr/ogrsf_frmts/pgeo/ogrpgeolayer.cpp
```diff
@@ -296,6 +296,27 @@ OGRFeature *OGRPGeoLayer::GetNextRawFeature()
 
         if( poGeom != nullptr && OGRERR_NONE == err )
         {
+            // always promote polygon/linestring geometries to multipolygon/multilinestring,
+            // so that the geometry types returned for the layer are predictable and match
+            // the advertised layer geometry type. See more details in OGRPGeoTableLayer::Initialize
+            const OGRwkbGeometryType eFlattenType = wkbFlatten(poGeom->getGeometryType());
+            if( eFlattenType == wkbPolygon )
```
rouault, line 303: all below could likely be simplified to:
```suggestion
            if( eFlattenType == wkbPolygon || eFlattenType == wkbCurvePolygon || eFlatenType == wkbLineString || eFlattenType == wkbCompoundCurve )
            {
                poGeom = OGRGeometryFactory::forceTo(poGeom, OGR_GT_GetCollection(poGeom->getGeometryType()));
            }            
```

but regarding wkbCurvePolygon and wkbCompoundCurve, we should linearize them as a  MultiSurface/MultiCurve isn't a valid MultiPolygon/MultiLineString. The alternative would be to declare those generic wbkMultiSurface/wkbMultiCurve types as the layer geometry types. hum...

autotest/ogr/ogr_pgeo.py
```diff
@@ -286,7 +286,101 @@ def test_ogr_pgeo_8():
...
+
+def test_ogr_pgeo_9():
+    if ogrtest.pgeo_ds is None:
+        pytest.skip()
+
+    if ogrtest.active_driver.GetName() != 'PGeo':
+        # MDB driver doesn't report non-spatial tables
+        pytest.skip()
+
+    ogrtest.pgeo_ds = ogr.Open('data/pgeo/mixed_types.mdb')
```
rouault, line 301: the practice of assigning dataset handles to a ogrtest. member should not be perpetuated. It makes potentially the tests interdependent which should be avoided Just use a ``ds`` local variable

autotest/ogr/ogr_pgeo.py
```diff
@@ -286,7 +286,101 @@ def test_ogr_pgeo_8():
...
+
+def test_ogr_pgeo_10():
+    if ogrtest.pgeo_ds is None:
+        pytest.skip()
+
+    if ogrtest.active_driver.GetName() != 'PGeo':
+        # MDB driver doesn't report non-spatial tables
+        pytest.skip()
+
+    ogrtest.pgeo_ds = ogr.Open('data/pgeo/mixed_types.mdb')
```
rouault, line 349: same remark as above

autotest/ogr/ogr_pgeo.py
```diff
@@ -286,7 +286,101 @@ def test_ogr_pgeo_8():
...
+    if ogrtest.pgeo_ds is None:
+        pytest.skip('could not open DB. Driver probably misconfigured')
+
+    polygon_layer = ogrtest.pgeo_ds.GetLayerByName('polygons')
+    assert polygon_layer.GetGeomType() == ogr.wkbMultiPolygon
+
+    feat = polygon_layer.GetNextFeature()
+    if feat.GetField('pk') != 1:
```
rouault, line 309: you can probably remove testing for the pk field here. it is not relevant for the purpose of the test. applies to similar below instances

gdal/ogr/ogrsf_frmts/pgeo/ogrpgeolayer.cpp
nyalldawson: For reference this logic was (mostly) copied from the openfilegdb driver, so any changes should also be applied there

autotest/ogr/ogr_pgeo.py
nyalldawson: My intention there was to make the test easier to debug if (for whatever reason) the ordering of features changes

gdal/ogr/ogrsf_frmts/pgeo/ogrpgeolayer.cpp
rouault: > For reference this logic was (mostly) copied from the openfilegdb driver, so any changes should also be applied there

I see... so yes, we have the potential flaw I underlined. I don't see a good solution to this. Linearizing is a pity, but advertizing the too-generic MultiSurface / MultiCurve could be an issue. I'd suggest to use my above proposed code simplification (with a comment about the issue with CurvePolygon/CompoundCurve), which hopefully should do the same as your code, and port it to the openfilegdb driver (where those CurvePolygon / CompoundCurve are tested as far as I can see)

nyalldawson: Actually it's not possible to use curved geometries in a personal geodatabase, so I'll clean up that logic

autotest/ogr/ogr_pgeo.py
rouault: ok, perhaps just a ``assert feat.GetField('pk') == 1`` would be a good compromise of expressiveness vs verbosity (I like succint tests :-))

PR #4264, opened 2021-08-19
[dwg] Make Closed AcDb2dPolyline and AcDb3dPolyline to Closed OGRLineString

GISerliang:
When dwg files contain closed AcDb2dPolyline or AcDb3dPolyline, ogr read it to unclosed line because closed AcDb2dPolyline or AcDb3dPolyline end point not equal to start point, only use isClosed function show it was closed.

### DIFF
--- a/gdal/ogr/ogrsf_frmts/dwg/ogrdwglayer.cpp
+++ b/gdal/ogr/ogrsf_frmts/dwg/ogrdwglayer.cpp
@@ -725,6 +725,11 @@ OGRFeature *OGRDWGLayer::Translate2DPOLYLINE( OdDbEntityPtr poEntity )
         poIter->step();
     }
 
+    if (poPL->isClosed())
+    {
+        poLS->addPoint(poLS->getX(0), poLS->getY(0), poLS->getZ(0));
+    }
+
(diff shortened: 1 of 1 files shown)


gdal/ogr/ogrsf_frmts/dwg/ogrdwglayer.cpp
```diff
@@ -725,6 +725,11 @@ OGRFeature *OGRDWGLayer::Translate2DPOLYLINE( OdDbEntityPtr poEntity )
         poIter->step();
     }
 
+    if (poPL->isClosed())
```
rouault, line 728: For more safety
```suggestion
    if (poPL->isClosed() && !poLS->IsEmpty())
```

gdal/ogr/ogrsf_frmts/dwg/ogrdwglayer.cpp
```diff
@@ -758,6 +763,11 @@ OGRFeature *OGRDWGLayer::Translate3DPOLYLINE( OdDbEntityPtr poEntity )
...
```
rouault, line 766:
```suggestion
    if (poPL->isClosed() && !poLS->IsEmpty())
```

PR #4265, opened 2021-08-19
NITF: write support for multiple images, and definition in nitf_spec.xml of DES for reading

rouault:


rouault:
CC @pl-nfwarmerdam

PR #4281, opened 2021-08-21
[geomedia, walk] Check sql tables for required tables before trying to query them when trying to open a .mdb dataset

nyalldawson:
This avoids noisy warnings raised by mdbtools odbc driver when trying to open an mdb files which isn't a geomedia/walk database,
as mdbtools dumps a lot of debug warnings when a query is attempted against a table which doesn't exist

### DIFF
--- a/gdal/ogr/ogrsf_frmts/geomedia/ogrgeomediadatasource.cpp
+++ b/gdal/ogr/ogrsf_frmts/geomedia/ogrgeomediadatasource.cpp
@@ -140,6 +140,27 @@ int OGRGeomediaDataSource::Open( const char * pszNewName, int bUpdate,
         }
     }
 
+    // check for GAliasTable table
+    {
+        bool bFoundGAliasTable = false;
+        CPLODBCStatement oTableList( &oSession );
+        if( oTableList.GetTables() )
+        {
+            while( oTableList.Fetch() )
+            {
+                const CPLString osTableName = CPLString( oTableList.GetColData(2, "") );
+                const CPLString osLCTableName(CPLString(osTableName).tolower());
+                if( osLCTableName == "galiastable" )
+                {
+                    bFoundGAliasTable = true;
+                    break;
+                }
+            }
+        }
+        if (!bFoundGAliasTable )
+            return FALSE;
+    }
+
     pszName = CPLStrdup( pszNewName );
 
     bDSUpdate = bUpdate;
--- a/gdal/ogr/ogrsf_frmts/walk/ogrwalkdatasource.cpp
+++ b/gdal/ogr/ogrsf_frmts/walk/ogrwalkdatasource.cpp
@@ -91,6 +91,27 @@ int OGRWalkDataSource::Open( const char * pszNewName, int /* bUpdate */ )
         }
     }
 
+    // check for WalkLayers table
+    {
+        bool bFoundWalkLayersTable = false;
+        CPLODBCStatement oTableList( &oSession );
+        if( oTableList.GetTables() )
+        {
+            while( oTableList.Fetch() )
+            {
+                const CPLString osTableName = CPLString( oTableList.GetColData(2, "") );
+                const CPLString osLCTableName(CPLString(osTableName).tolower());
+                if( osLCTableName == "walklayers" )
+                {
+                    bFoundWalkLayersTable = true;
+                    break;
+                }
+            }
+        }
+        if (!bFoundWalkLayersTable )
+            return FALSE;
+    }
+
     pszName = CPLStrdup( pszNewName );
 
 /* -------------------------------------------------------------------- */


gdal/ogr/ogrsf_frmts/geomedia/ogrgeomediadatasource.cpp
```diff
@@ -140,6 +140,28 @@ int OGRGeomediaDataSource::Open( const char * pszNewName, int bUpdate,
...
 
+    // check for GAliasTable table
+    bool bFoundGAliasTable = false;
+    {
+        CPLODBCStatement oTableList( &oSession );
+        if( oTableList.GetTables() )
+        {
+            while( oTableList.Fetch() )
+            {
+                const CPLString osTableName = CPLString( oTableList.GetColData(2) );
```
rouault, line 151: looking at ogr/ogrsf_frmts/odbc/ogrodbcdatasource.cpp, I see a null ptr check for the return of .GetColData(2). Assigning a nullptr to a CPLString / std::string is undefined behaviour. So perhaps pass the "" as default value

```suggestion
                const CPLString osTableName( oTableList.GetColData(2, "") );
```

gdal/ogr/ogrsf_frmts/geomedia/ogrgeomediadatasource.cpp
```diff
@@ -140,6 +140,28 @@ int OGRGeomediaDataSource::Open( const char * pszNewName, int bUpdate,
...
+                const CPLString osLCTableName(CPLString(osTableName).tolower());
+
+                if( !osLCTableName.empty() && osLCTableName == "galiastable" )
```
rouault, line 154: can be likely simplified as (was the empty() check to be slighly more performant ? but in that case I believe that empty string is highly unlikely, so that will not be necessary)
```suggestion
                if( osLCTableName == "galiastable" )
```

gdal/ogr/ogrsf_frmts/geomedia/ogrgeomediadatasource.cpp
```diff
@@ -140,6 +140,28 @@ int OGRGeomediaDataSource::Open( const char * pszNewName, int bUpdate,
         }
     }
 
+    // check for GAliasTable table
+    bool bFoundGAliasTable = false;
```
rouault, line 144: this variable can be moved within the block

gdal/ogr/ogrsf_frmts/walk/ogrwalkdatasource.cpp
```diff
@@ -91,6 +91,27 @@ int OGRWalkDataSource::Open( const char * pszNewName, int /* bUpdate */ )
         }
     }
 
+    // check for WalkLayers table
+    bool bFoundWalkLayersTable = false;
```
rouault, line 95: can be moved within the block

gdal/ogr/ogrsf_frmts/walk/ogrwalkdatasource.cpp
```diff
@@ -91,6 +91,27 @@ int OGRWalkDataSource::Open( const char * pszNewName, int /* bUpdate */ )
...
+    {
+        CPLODBCStatement oTableList( &oSession );
+        if( oTableList.GetTables() )
+        {
+            while( oTableList.Fetch() )
+            {
+                const CPLString osTableName = CPLString( oTableList.GetColData(2) );
```
rouault, line 102:
```suggestion
                const CPLString osTableName( oTableList.GetColData(2, "") );
```

gdal/ogr/ogrsf_frmts/walk/ogrwalkdatasource.cpp
```diff
@@ -91,6 +91,27 @@ int OGRWalkDataSource::Open( const char * pszNewName, int /* bUpdate */ )
...
+                const CPLString osLCTableName(CPLString(osTableName).tolower());
+                if( !osLCTableName.empty() && osLCTableName == "walklayers" )
```
rouault, line 104:
```suggestion
                if( osLCTableName == "walklayers" )
```

rouault:
Note: if you're interested in silencing mdbtools stderr messages, in OGRPGeoDataSource::CountStarWorking(), I have a hack for that for Linux (could probably work on other POSIX systems). If we wanted to generalize that use, we should probably move that to CPL as a ``CPLStderrSilencer`` class.

nyalldawson:
> Note: if you're interested in silencing mdbtools stderr messages, in OGRPGeoDataSource::CountStarWorking(), I have a hack for that for Linux (could probably work on other POSIX systems). If we wanted to generalize that use, we should probably move that to CPL as a CPLStderrSilencer class.

Thanks for the pointer! I'll leave it for now as I'm finding the warnings helpful, but will swing back round to this later

PR #4283, opened 2021-08-21
[pgeo] Add more internal tables to block list

nyalldawson:
These tables are present in the "Autodesk Test.mdb" test file


gdal/ogr/ogrsf_frmts/pgeo/ogrpgeodatasource.cpp
```diff
@@ -232,17 +232,45 @@ int OGRPGeoDataSource::Open( const char * pszNewName, int bUpdate,
                         && osLCTableName != "msysaces"
                         && osLCTableName != "msysqueries"
                         && osLCTableName != "msysrelationships"
+                        && osLCTableName != "gdb_annosymbols"
```
rouault, line 235: This list begins to be seriously painful to maintain. Why not just do ``&& !(osLCTableName.size() >= 4 && osLCTableName.substr(0, 4) == "gdb_")`` ?

PR #4290, opened 2021-08-22
Split pgeo and mdb driver test suites

nyalldawson:
It's too messy to try to handle both drivers in a single test suite, as the behaviour for the PGeo driver is differing more and more
from the MDB driver.

### DIFF
--- a/autotest/ogr/ogr_mdb.py
+++ b/autotest/ogr/ogr_mdb.py
@@ -1,0 +1,214 @@
+#!/usr/bin/env pytest
+# -*- coding: utf-8 -*-
+###############################################################################
+# $Id$
+#
+# Project:  GDAL/OGR Test Suite
+# Purpose:  Test read functionality for OGR MDB driver.
+# Author:   Even Rouault <even dot rouault at spatialys.com>
+#
+###############################################################################
+# Copyright (c) 2010-2012, Even Rouault <even dot rouault at spatialys.com>
+#
+# Permission is hereby granted, free of charge, to any person obtaining a
+# copy of this software and associated documentation files (the "Software"),
+# to deal in the Software without restriction, including without limitation
+# the rights to use, copy, modify, merge, publish, distribute, sublicense,
+# and/or sell copies of the Software, and to permit persons to whom the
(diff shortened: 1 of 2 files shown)


autotest/ogr/ogr_mdb.py
```diff
@@ -0,0 +1,216 @@
...
+    if driver is not None:
+        driver.Register()
+    else:
+        pytest.skip("MDB driver not available", allow_module_level=True)
+
+    # remove PGeo driver
+    ogrtest.pgeo_driver = ogr.GetDriverByName('PGeo')
+    if ogrtest.pgeo_driver is not None:
+        ogrtest.pgeo_driver.Deregister()
+
```
rouault, line 53: the cleanest way would be to add a ``yield`` just at the end of setup_driver() and followed by the content of the test_ogr_mdb_cleanup() (reregistration of the pgeo_driver, which you would need to attach to the ogrtest). See ogr_kml.py for a similar example
The advantage of this is that it could allow those tests to run with the --random-order switch of pytest

autotest/ogr/ogr_pgeo.py
```diff
@@ -38,57 +38,24 @@
...
+    if driver is not None:
+        driver.Register()
     else:
-        switch_driver('MDB', 'PGeo')
+        pytest.skip("PGeo driver not available", allow_module_level=True)
 
+    # remove mdb driver
+    ogrtest.mdb_driver = ogr.GetDriverByName('MDB')
+    if ogrtest.mdb_driver is not None:
+        ogrtest.mdb_driver.Deregister()
```
rouault, line 52: similar remark as above regarding yield + cleanup to incorporate here

rouault (commented):
Nice initiative!

autotest/ogr/ogr_mdb.py
nyalldawson: I'm confused why the pgeo driver would need to be attached to ogrtest in that setup -- couldn't it remain as a local variable only?

rouault: oops, sorry I forgot the negation in my sentence and actually meant that it can remain a local variable

PR #4307, opened 2021-08-24
OpenFileGDB: fix reading of raster fields

rouault:


gdal/ogr/ogrsf_frmts/openfilegdb/filegdbtable.cpp
```diff
@@ -1051,6 +1075,7 @@ int FileGDBTable::Open(const char* pszFilename,
             {
                 /* Always one byte at end ? */
```
nyalldawson, line 1076: This comment is outdated now, isn't it?

nyalldawson (approved):

nyalldawson:
Nice! Another part of the puzzle solved 😁

gdal/ogr/ogrsf_frmts/openfilegdb/filegdbtable.cpp
rouault: indeed

PR #4354, opened 2021-08-27
CI: fedora: add mdbtools-odbc so that PGeo driver can work, add default path location for driver on Fedora

rouault:


nyalldawson:
@rouault

> on Ubuntu 20.04 with one of your recent change (I observed this yesterday, before you initialized the MDB driver once), I now get "CRITICAL can't find column with internal id XXX in index YYY" traces on stderr.

FYI this is fixed upstream now

PR #4365, opened 2021-08-27
Prototype deprecated driver note in documentation

nyalldawson:
## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:

### DIFF
--- a/gdal/doc/source/_extensions/driverproperties.py
+++ b/gdal/doc/source/_extensions/driverproperties.py
@@ -1,7 +1,8 @@
 # From https://www.sphinx-doc.org/en/master/development/tutorials/todo.html
 
-import sphinx.locale
 import docutils.statemachine
+import sphinx.locale
+
 sphinx.locale.admonitionlabels['shortname'] = ''
 sphinx.locale.admonitionlabels['built_in_by_default'] = ''  # 'Built-in by default'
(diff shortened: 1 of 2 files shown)


gdal/doc/source/_extensions/driverproperties.py
```diff
@@ -270,6 +284,26 @@ def run(self):
...
 
+class DeprecatedDriverDirective(Directive):
+
+    # this enables content in the directive
+    has_content = True
+
+    def run(self):
+
+        env_variable = self.content[0]
+        self.content = docutils.statemachine.StringList(["This driver is considered for removal in GDAL 3.5. You are invited " \
```
rouault, line 295: couldn't we make the version a parameter of the directive too ? I guess we'll also have to deprecate other drivers after 3.5. Otherwise looks good !

nyalldawson: Ok, see what you think of the update...

nyalldawson:
Here's how it looks now:

image

gdal/doc/source/_extensions/driverproperties.py
rouault: excellent!

Issue #4357, opened 2021-08-27
Missing Identify() method in ODBC and PGeo drivers


rouault:
CC @nyalldawson 
Due to recent changes in those drivers, the following debug messages are now emitted:
GDAL: Driver ODBC that defines GDAL_DMD_OPENOPTIONLIST must also implement Identify(), so that it can be used
GDAL: Driver PGeo that defines GDAL_DMD_OPENOPTIONLIST must also implement Identify(), so that it can be used

PR #4376, opened 2021-08-28
[pgeo] Add identify method

nyalldawson:
Refs #4357


rouault (commented):
Could you simplify the Open() method to remove the redundant tests with Identify and just do an initial
```
// The method might return -1 when it is undecided
if (OGRPGeoDriverIdentify(poOpenInfo) == FALSE)
   return nullptr;
```

nyalldawson:
@rouault done

PR #4394, opened 2021-08-30
configure: add tirpc include path when available when building with OGDI support (fixes https://github.com/libogdi/ogdi/issues/20)

rouault:
@sebastic  what about that ?

### DIFF
--- a/gdal/configure
+++ b/gdal/configure
@@ -796,6 +796,8 @@ SOSI_ENABLED
 X_FME_HOME
 OGDI_INCLUDE
 HAVE_OGDI
+RPC_LIBS
+RPC_CFLAGS
 OGDI_LIBS
 OGDI_CFLAGS
 GIF_SETTING
@@ -1255,6 +1257,8 @@ PQ_CFLAGS
 PQ_LIBS
 OGDI_CFLAGS
 OGDI_LIBS
+RPC_CFLAGS
+RPC_LIBS
 MONGOCXXV3_CFLAGS
 MONGOCXXV3_LIBS
 HDF5_CFLAGS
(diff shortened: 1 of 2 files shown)


sebastic:
This works. But shouldn't `-I/usr/include/tirpc -ltirpc` be set in `ogdi-config` when it's built with TI RPC?

rouault:
> But shouldn't -I/usr/include/tirpc -ltirpc be set in ogdi-config when it's built with TI RPC?

could be appropriate indeed but GDAL's configure doesn't use it

sebastic:
GDAL's configure does use pkg-config, which contains the tirpc flags:
```
# cat /usr/share/pkgconfig/ogdi.pc 
prefix=/usr
exec_prefix=/usr
libdir=/usr/lib
includedir=${prefix}/include

Name: ogdi
Description: Open Geographic Datastore Interface
Version: 4.1.0
Cflags: -I${includedir}/ogdi -I/usr/include/tirpc
Libs: -L${libdir} -logdi
Libs.private: -lexpat -lz -ltirpc

```

rouault:
> GDAL's configure does use pkg-config, which contains the tirpc flags:

ah indeed, so why doesn't that work ? That should be set in OGDI_INCLUDE in GDALmake.opt

rouault:
> ah indeed, so why doesn't that work ? That should be set in OGDI_INCLUDE in GDALmake.opt

in a debian:unstable Docker image, I've just successfully (manually) built gdal-3.3.2RC1 against the packaged ogdi. The generated GDALmake.opt contains
OGDI_INCLUDE    =       -I/usr/include/ogdi -I/usr/include/tirpc
and this is passed when building the driver

sebastic:
> in a debian:unstable Docker image, I've just successfully (manually) built gdal-3.3.2RC1 against the packaged ogdi.

That must be without the DODS driver because libdap is currently broken due to the RPC change in glibc too ([Debian Bug #993016](https://bugs.debian.org/993016)).

The CFLAGS for OGDI aren't used, the `C compiler` & `C++ compiler` flags don't include them unlike they do with the changes from this issue.

Full buildlog: [gdal_3.3.2~rc1+dfsg-1_amd64.txt](https://github.com/OSGeo/gdal/files/7081245/gdal_3.3.2.rc1%2Bdfsg-1_amd64.txt)

PR #4400, opened 2021-08-31
DIMAP: expose virtual overviews

tbonfort:
The dynamically created VRT that wraps the underlying datasets does not expose any overviews.
This PR adds this capability, although I'm not sure wether this is a hack or the correct way
to do so.

Before (cancelled):
```
time /opt/bin/gdalwarp  -ts 1248 4835 DIM_PHR1B_PMS_201911041624143_SEN_5886713101.XML og.tif
Creating output file that is 1248P x 4835L.
Processing DIM_PHR1B_PMS_201911041624143_SEN_5886713101.XML [1/1] : 0^C
real	0m20,575s
user	2m36,811s
sys	0m0,748s
```

After:
```
time /opt/bin/gdalwarp  -ts 1248 4835 DIM_PHR1B_PMS_201911041624143_SEN_5886713101.XML ng.tif --config VRT_VIRTUAL_OVERVIEWS YES
Creating output file that is 1248P x 4835L.
Processing DIM_PHR1B_PMS_201911041624143_SEN_5886713101.XML [1/1] : 0...10...20...30...40...50...60...70...80...90...100 - done.

real	0m2,342s
user	0m13,045s
sys	0m0,226s
```

### DIFF
--- a/gdal/frmts/dimap/dimapdataset.cpp
+++ b/gdal/frmts/dimap/dimapdataset.cpp
@@ -36,6 +36,7 @@
 #include "mdreader/reader_pleiades.h"
 #include "vrtdataset.h"
 #include <map>
+#include <algorithm>
 
 CPL_CVSID("$Id$")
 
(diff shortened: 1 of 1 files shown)


gdal/frmts/dimap/dimapdataset.cpp
```diff
@@ -1333,6 +1334,21 @@ int DIMAPDataset::ReadImageInformation2()
...
+/*      Expose Overviews if available                                   */
+/* -------------------------------------------------------------------- */
+    auto poSrcBandFirstImage = poImageDS->GetRasterBand(1);
+    const int nSrcOverviews = std::min(30, poSrcBandFirstImage->GetOverviewCount());
+    if(nSrcOverviews>0) {
+        std::unique_ptr<int[]> ovrLevels(new int[nSrcOverviews]);
+        int iLvl=2;
+        for(int i=0; i<nSrcOverviews; i++) {
+            ovrLevels[i]=iLvl;
+            iLvl*=2;
```
rouault, line 1347: Actually for nSrcOverviews == 30, we would overvflow int at the last iteration...
```suggestion
        int iLvl=1;
        for(int i=0; i<nSrcOverviews; i++) {
            iLvl*=2;
            ovrLevels[i]=iLvl;
```

tbonfort:
This needs to be re-worked, as vrt->IBuildOverviews only kicks in correctly in our case if VRT_VIRTUAL_OVERVIEWS is set

rouault:
> This needs to be re-worked, as VRTDataset::IBuildOverviews only kicks in correctly in our case if VRT_VIRTUAL_OVERVIEWS is set

ah indeed... You'll want to add
CPLConfigOptionSetter oSetter("VRT_VIRTUAL_OVERVIEWS", "YES", false);

tbonfort:
> > This needs to be re-worked, as VRTDataset::IBuildOverviews only kicks in correctly in our case if VRT_VIRTUAL_OVERVIEWS is set
> 
> ah indeed... You'll want to add
> CPLConfigOptionSetter oSetter("VRT_VIRTUAL_OVERVIEWS", "YES", false);

Thanks for the CPLConfigOptionSetter tip, I was starting to fiddle with saving/restoring the config options manually. Confirming correct behavior with this fix.

rouault:
could be good to add a test that checks the presence of overviews

Issue #4399, opened 2021-08-31
autotest/gdrivers/rmf.py::test_rmf_31e randomly fails


rouault:
e.g in https://github.com/OSGeo/gdal/pull/4398/checks?check_run_id=3473089379
The random nature of the failure is likely due to this tests using random numbers. I strongly suspect this is due to the delta encoding of the values with a variable integer strategy in the RMFDem compressor, and for some deltas inappropriately being classified. Or using one of the reserved value.

rouault:
CC @drons

drons:
> CC @drons

Ok, I'll check it out.

rouault:
Seems to be fixed

PR #4405, opened 2021-09-01
RMF: Rework RMF delta-compression test (#4399)

drons:


rouault:
doesn't that just hide a underlying issue with some particular input values ?

rouault:
> This is "fast fix". I'll try the old test again to dig deeper.

ok, I suspect particular delta values close to the transitions like -(1 << 16) or -(1 << 24) (or perhaps with full int range)

drons:
It's seems, that algorithm does not support out of 32-bit deltas.

But I still need to look with a fresh mind.

PR #4411, opened 2021-09-01
GDALOpenInfo(): fix issue when passing a directory on netBSD and potentially other BSDs

rouault:


gdt:
Code wise that seems ok, but I think the comments should be different.   Currently it reads as if BSDs do something odd, which they aren't, given historical UNIX norms.   What's going on is that the code path is relying on behavior not guaranteed by POSIX, and that this is an optimization on a limited set of platforms known to have that behavior and which are CI tested to ensure that if this non-required behavior changes we'd catch it.

PR #4442, opened 2021-09-06
FlatGeobuf: min align to 8 bytes

bjornharrtell:
I have recently discovered a weakness which can be fixed by forcing buffers to be 8 byte aligned, see https://github.com/flatgeobuf/flatgeobuf/issues/140.

The weakness does not affect any known implementation of FlatGeobuf readers but potentially hinders an optimized path in that if the whole file already in memory it cannot be zero-copy accessed unless properly aligned.

So I'm proposing to do this change, which is entirely safe. I also propose that I use the last byte of the FlatGeobuf magic byte as a fully backwards patch level version to indicate if buffers have been created with this proper alignment. This means I want to relax the magicbytes check to essentially ignore the last byte. ~~I might do that in this PR too...~~ (done)


bjornharrtell:
Hah I must have been psychic because I in fact only verify the first four bytes of the magicbytes which is great because that means the implementation has been forward compatible with any future patch level from the start. This bit:

https://github.com/OSGeo/gdal/blob/8b47e07e21a4baf0368456d28440ac13579449d6/gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobufdataset.cpp#L53-L58

bjornharrtell:
The two failing checks seem unrelated to this PR.

I will follow up this PR with a patch level bump in magicbytes when I have confirmed the buffers produced can be zero copy read without alignment issues.

bjornharrtell:
@rouault please merge if no objections

Issue #4407, opened 2021-09-01
autotest: refactor so that test steps can be run independently


rouault:
Currently a number of test files have interdependent steps. It would be good to make them independent. Using ``pytest --random-order (https://pypi.org/project/pytest-random-order/)`` can be for example used to check that

DFEvans:
Some lists below of failures I observe when running in random order - test files with failures, and a longer list of individual failures within those files. #4421 fixes the single example I found where changing the test order causes a segmentation fault.

I also found that when running with random order that Pytest does not exit at the end. On giving a keyboard interrupt, the traceback shows it was stuck trying to join a thread, but gave no indication of where that thread was spawned from.

<details>
<summary>Files with failures</summary>

```
alg/contour.py
alg/warp.py
gcore/hfa_write.py
gcore/misc.py
gcore/numpy_rw_multidim.py
gcore/pam.py
gcore/tiff_ovr.py
gcore/tiff_read.py
gcore/tiff_write.py
gcore/vsiaz.py
gcore/vsigs.py
gcore/vsihdfs.py
gcore/vsioss.py
gcore/vsiswift.py
gcore/vsiwebhdfs.py
gdrivers/arg.py
gdrivers/ecrgtoc.py
gdrivers/eedai.py
gdrivers/fast.py
gdrivers/gdalhttp.py
gdrivers/gif.py
gdrivers/gpkg.py
gdrivers/grass.py
gdrivers/hfa.py
gdrivers/kea.py
gdrivers/mbtiles.py
gdrivers/netcdf.py
gdrivers/ngw.py
gdrivers/postgisraster.py
gdrivers/rda.py
gdrivers/rl2.py
gdrivers/srtmhgt.py
gdrivers/test_validate_jp2.py
gdrivers/vrtpansharpen.py
gdrivers/vrtrawlink.py
gdrivers/webp.py
gdrivers/wmts.py
gnm/gnm_test.py
ogr/ogr_basic_test.py
ogr/ogr_elasticsearch.py
ogr/ogr_gml_read.py
ogr/ogr_gpkg.py
ogr/ogr_ingres.py
ogr/ogr_mem.py
ogr/ogr_mitab.py
ogr/ogr_mongodbv3.py
ogr/ogr_mssqlspatial.py
ogr/ogr_mvt.py
ogr/ogr_mysql.py
ogr/ogr_ngw.py
ogr/ogr_oapif.py
ogr/ogr_ogdi.py
[...]

PR #4444, opened 2021-09-07
Remove vsis3 test order dependency

DFEvans:
## What does this PR do?

Refactor gcore/vsis3.py to remove dependency on test order.

Refactor to apply environment variable overrides only during the test using now-standard test fixtures, rather than altering the global GDAL environment.

I apologise in advance that there are a _ridiculous_ number of changes to appease flake8 - in retrospect, I wish I had done these first. If it's simply too much (and I suspect it is), options are:
* Merge without the linting changes, which are in the final commit, and submit a separate PR for them (will the build pipeline permit this?)
* Tell me to go away and lint the whole file first in one PR, and then do the dependency changes

#4407 

## Tasklist

 - [x] Refactor code
 - [x] Linting for flake8
 - [x] Review
 - [x] Adjust for comments

### DIFF
--- a/autotest/gcore/vsis3.py
+++ b/autotest/gcore/vsis3.py
@@ -64,8 +64,7 @@ def aws_test_config():
         'AWS_REQUEST_PAYER': '',
         'AWS_DEFAULT_REGION': 'us-east-1',
         'AWS_DEFAULT_PROFILE': 'default',
-        'AWS_PROFILE': 'default',
-        'AWS_NO_SIGN_REQUEST': 'NO'
+        'AWS_PROFILE': 'default'
     }
 
     with gdaltest.config_options(options):
@@ -4346,8 +4345,6 @@ def test_vsis3_read_credentials_config_file_non_default_profile(
         # runner unless it is set to 'myprofile' here.
         'AWS_DEFAULT_PROFILE': 'myprofile'
     }
-
-    options = {}
 
     os_aws = tmpdir.mkdir(".aws")
 
--- a/autotest/gcore/vsis3.py
+++ b/autotest/gcore/vsis3.py
@@ -46,15 +46,27 @@ def open_for_read(uri):
     """
     return gdal.VSIFOpenExL(uri, 'rb', 1)
 
-###############################################################################
 
-
-def test_vsis3_init():
-
-    gdaltest.aws_vars = {}
-    for var in ('AWS_SECRET_ACCESS_KEY', 'AWS_ACCESS_KEY_ID', 'AWS_TIMESTAMP', 'AWS_HTTPS', 'AWS_VIRTUAL_HOSTING', 'AWS_S3_ENDPOINT', 'AWS_REQUEST_PAYER', 'AWS_DEFAULT_REGION', 'AWS_DEFAULT_PROFILE', 'AWS_PROFILE', 'AWS_NO_SIGN_REQUEST'):
-        gdaltest.aws_vars[var] = gdal.GetConfigOption(var)
-        if gdaltest.aws_vars[var] is not None:
+@pytest.fixture()
+def initialise_aws_s3():
+    vars_to_restore = {}
+    aws_vars = (
+        'AWS_SECRET_ACCESS_KEY',
+        'AWS_ACCESS_KEY_ID',
+        'AWS_TIMESTAMP',
+        'AWS_HTTPS',
+        'AWS_VIRTUAL_HOSTING',
+        'AWS_S3_ENDPOINT',
+        'AWS_REQUEST_PAYER',
+        'AWS_DEFAULT_REGION',
+        'AWS_DEFAULT_PROFILE',
+        'AWS_PROFILE',
+        'AWS_NO_SIGN_REQUEST'
+    )
+    for var in aws_vars:
+        value = gdal.GetConfigOption(var)
+        if value is not None:
+            vars_to_restore[var] = value
             gdal.SetConfigOption(var, "")
 
     # To avoid user AWS credentials in ~/.aws/credentials and ~/.aws/config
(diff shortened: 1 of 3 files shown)


rouault:
I can see some slowdown in the fact of starting and stopping the webserver for each test case: ~ 21.5 s now vs ~ 12.5 before. Couldn't the webserver be started once for all in a ``@pytest.fixture(autouse=True, scope='module')`` function ?

DFEvans:
I'll have a look - I think I initially had that, undid it while working through the various issues, but those were mostly (entirely?) down to how the environment variables were being set.

DFEvans:
The runtimes should now be back to what they were before.

Looking at the test runtimes, I'm surprised that a couple of the local webserver tests are among the slowest. For example, `test_vsis3_random_write` takes over a second to run, with almost all of that time going into the call to `gdal.VSIFCloseL`. Is there a simple change that can be made to improve that?

rouault:
> For example, `test_vsis3_random_write` takes over a second to run, with almost all of that time going into the call to `gdal.VSIFCloseL`. Is there a simple change that can be made to improve that?

ah, setting the CPL_CURL_VERBOSE=YES environment variable, I see the following traces
```
> PUT /random_write/test.bin HTTP/1.1
Host: 127.0.0.1:8080
Accept: */*
x-amz-date: 20150101T000000Z
x-amz-content-sha256: 2c26b46b68ffc68ff99b453c1d30413413422d706483bfa0f98a5e886266e7ae
Authorization: AWS4-HMAC-SHA256 Credential=AWS_ACCESS_KEY_ID/20150101/us-east-1/s3/aws4_request,SignedHeaders=host;x-amz-content-sha256;x-amz-date,Signature=ea1dcde6e84df1f5c722685ab953c475c44a85b6968c7d4d1aee11ec4bc4888c
Expect: 100-continue
Content-Length: 3

* Done waiting for 100-continue
* We are completely uploaded and fine
* Mark bundle as not supporting multiuse
* HTTP 1.0, assume close after body
< HTTP/1.0 200 OK
< Server: BaseHTTP/0.6 Python/3.8.10
< Date: Wed, 08 Sep 2021 08:28:52 GMT
< Content-Length: 0
< 
* Closing connection 0
```
So we send a PUT request with a Expect: 100-continue header, but our server emulation doesn't emit it and just emits the HTTP/1.0 200 OK. The "Done waiting for 100-continue" must be a 1 s delay in curl in that situation. Probably that improving webserver.py/_process_req_resp() to emit the HTTP/1.1 100 Continue would fix this. I see in test_vsis3_4() that we have a custom handler function that does this.

DFEvans:
The Appveyor build failure has me confused. I can see it's failing on some tests modified by this PR, but I don't know what the important difference between the failing Appveyor build and the passing Github Windows builds is likely to be.

rouault:
> The Appveyor build failure has me confused. I can see it's failing on some tests modified by this PR, but I don't know what the important difference between the failing Appveyor build and the passing Github Windows builds is likely to be.

I don't know either. I don't recall having seen flakes in those tests before (we have sometimes flakes on test_vsis3_sync_multithreaded_download or test_vsis3_sync_multithreaded_download_chunk_size)

rouault:
I've restarted the job and it still fails. One hypothesis would be that you modified the formatting of the XML response to be a multiline string, and perhaps git clone on that config uses CR-LF when checking it out files ? But there are other similar instances in other tests and they don't fail

DFEvans:
I've pushed a change to `test_vsis3_write_multipart_retry` to change the XML to avoid newlines - we'll see if that changes it.

I had thought the Appveyor builds only started failing after the HTTP 100 change I made in webserver.py, so my next option is to try undoing that and see if it builds. The change I made could result in tests receiving HTTP 100 when they didn't before (if GDAL is sending `Expect` in the header), and perhaps there's a chance that it's revealed a very odd corner case.

rouault:
> The change I made could result in tests receiving HTTP 100 when they didn't before (if GDAL is sending `Expect: 100-continue` in the header), and perhaps there's a chance that it's revealed a very odd corner case.

yeah, that's a more likely cause. Actually I see that in master when we emit manually 100 continue, we do it in HTTP/1.1. But by default the webserver likely use HTTP/1.0 by default. Perhaps adding request.protocol_version = 'HTTP/1.1' in webserver.py would help ? (but I would expect the curl version used by the appveyor and github workflows to be the same)
(I also see that in some places we have explicit send_header('Connection', 'close'). Unfortunately I failed to note why this was needed...)

DFEvans:
The latest changes to set responses to HTTP 1.1 has caused a wide range of very odd test failures that differ by system. I've reverted this PR back to the state it was in before any changes regarding "100 Continue" responses, so it simply includes the test order dependency changes and linting. Pushing for a ~2s reduction in test runtime can wait for another day!

autotest/gcore/vsis3.py
```diff
@@ -3070,35 +4309,45 @@ def test_vsis3_read_credentials_config_file():
...
+        # runner unless it is set to 'myprofile' here.
+        'AWS_DEFAULT_PROFILE': 'myprofile'
+    }
 
-    gdal.SetConfigOption('AWS_SECRET_ACCESS_KEY', '')
-    gdal.SetConfigOption('AWS_ACCESS_KEY_ID', '')
-    gdal.SetConfigOption('CPL_AWS_CREDENTIALS_FILE', None)
-    gdal.SetConfigOption('AWS_CONFIG_FILE', None)
-    gdal.SetConfigOption('AWS_PROFILE', 'myprofile')
+    options = {}
```
rouault, line 4350: this line should be removed as it cancels the above non-empty options

rouault:
Thanks!

PR #4480, opened 2021-09-13
Maintenance: Setup flake8 config file, update code_checks paths

DFEvans:
## What does this PR do?
The new autotest/setup.cfg defines the Flake8 checks to skip,
and allowing further configuration in future. This makes it easier
for contributors to run the same Flake8 check as the code-checks
CI run, which is much less stringent than unconfigured Flake8.

Update the code_checks.yml file to use this configuration, and
update some of the paths inspected to match the current repo
layout.

### DIFF
--- a/autotest/setup.cfg
+++ b/autotest/setup.cfg
@@ -1,0 +1,65 @@
+[flake8]
+ignore = 
+    E101  # Indentation contains mised spaces and tabs
+    E111  # Indentation is not a multiple of 4
+    E114  # Indentation is not a multiple of 4 (comment)
+    E115  # Expected an indented block (comment)
+    E117  # Over-indented comment
+    E121  # Continuation line under-indented for hanging indent
+    E122  # Continuation line missing indentation or outdented
+    E123  # Closing bracket does not match indentation of opening bracket
+    E124  # Closing bracket does not match visual indentation
+    E125  # Continuation line with same indent as next logical line
+    E126  # Continuation line over-indented for hanging indent
+    E127  # Continuation line over-indented
+    E128  # Continuation line under-indented
+    E131  # Continuation line unaligned for handing indent
+    E201  # Whitespace after bracket
+    E202  # Whitespace before bracket
+    E203  # Whitespace before ','
+    E211  # Whitespace before bracket
+    E221  # Multiple spaces before operator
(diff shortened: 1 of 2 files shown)


autotest/setup.cfg
```diff
@@ -0,0 +1,65 @@
...
+    E703  # Statement ends with a semicolon
+    E711  # Comparison to None should be 'if cond is None'
+    E712  # Comparison to False should be 'if cond is False' or 'if not cond'
+    E713  # Test for membership should be 'not in'
+    E721  # Do not compare types, use isinstance()
+    E722  # Do not use bare 'except'
+    E731  # Do not assign a lambda expression, use a def
+    E741  # Ambiguous variable name 'l'
+    E291  # Trailing whitespace
+    F401  # Module imported but unused
```
rouault, line 53: F401 is currently selected, and should not be in the ignore list. Same for F403, F405, F811, F841

DFEvans: Ah, I created the list by assuming any current failures were meant to be accepted, but changing the paths in code_checks.yml fired off a number of warnings in gdal/swig/python/osgeo/.

Should all files in gdal/swig/python/osgeo/ be linted? The old code_checks.yml pointed to a pair of nonexistent directories in there - utils/ and utils/auxiliary.

DFEvans: Ah, the files auto-generated by Swig definitely shouldn't be linted.

rouault: > The old code_checks.yml pointed to a pair of nonexistent directories in there - utils/ and utils/auxiliary.

there has been some file moving in last months, and code_checks.yml has obviously failed to catch those moves. I guess @idanmiara could help  if needed in linting the issues that have been introduced lately

rouault: > Ah, the files auto-generated by Swig definitely shouldn't be linted.

agreed

idanmiara: > > The old code_checks.yml pointed to a pair of nonexistent directories in there - utils/ and utils/auxiliary.
> 
> there has been some file moving in last months, and code_checks.yml has obviously failed to catch those moves. I guess @idanmiara could help if needed in linting the issues that have been introduced lately

Sure, please ping me if you need help with that.

PR #4484, opened 2021-09-14
Test - Refactor gdalhttp.py to remove dependency on test order

DFEvans:
Refactor gdalhttp.py to remove dependency on test order (#4407 )

Also refactor code to reduce duplication, and ensure all tests have a
timeout. I was finding that test_http_4, known to be flakey, was hanging
for me due to a lack of any timeout.

### DIFF
--- a/autotest/gdrivers/gdalhttp.py
+++ b/autotest/gdrivers/gdalhttp.py
@@ -46,127 +46,87 @@ except ImportError:
     sys.path.append('../pymod')
     import gdaltest
 
+
+pytestmark = pytest.mark.require_driver('HTTP')
+
+
+@pytest.fixture(autouse=True, scope="module")
+def disable_dods_driver():
+    dods_drv = gdal.GetDriverByName('DODS')
+    if dods_drv is not None:
+        dods_drv.Deregister()
+    yield
+    if dods_drv is not None:
+        dods_drv.Register()
+    dods_drv = None
+
+
+@pytest.fixture(autouse=True)
+def set_http_timeout():
+    with gdaltest.config_option('GDAL_HTTP_TIMEOUT', '5'):
+        yield
+
+
+def skip_if_unreachable(url, try_read=False):
+    conn = gdaltest.gdalurlopen(url, timeout=4)
+    if conn is None:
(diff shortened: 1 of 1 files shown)


autotest/gdrivers/gdalhttp.py
```diff
@@ -46,127 +46,87 @@
...
     ret = tst.testOpen()
     if ret == 'fail':
-        conn = gdaltest.gdalurlopen('https://raw.githubusercontent.com/OSGeo/gdal/release/3.1/autotest/gcore/data/byte.tif')
-        if conn is None:
-            pytest.skip('cannot open URL')
-        conn.close()
-
-    return ret
+        skip_if_unreachable(url)
+        pytest.fail()
```
DFEvans, line 92: Tests 1 and 2 previously ended with `return ret`, but I've not been able to find any evidence that pytest would parse the result of `ret`. I've changed them to explicitly fail, in line with the other tests.

DFEvans:
Another oddity is Test 5, with the odd flow being:

```
url = 'https://raw.githubusercontent.com/OSGeo/gdal/release/3.1/autotest/gdrivers/data/s4103.blx'
ds = gdal.Open(url)
filename = ds.GetDescription()
ds = None
with pytest.raises(OSError):
    os.stat(filename)
    pytest.fail('file %s should have been removed' % filename)
```

My impression is that this is meant to be checking that some local temporary file is deleted. However, ds.Description() returns the original URL, not a local filepath so the os.stat() will never detect a temporary file that hasn't been deleted. I'd guess that ds.GetDescription() used to return a temporary file path; is there a way to get that file path for this test?

rouault:
> I'd guess that ds.GetDescription() used to return a temporary file path;

yes, before 850dc88185c

> is there a way to get that file path for this test?

hum, when dowloading http://example.com/some.thing, on Unix a /tmp/smoe.thing file is created (see logic in frmts/wcs/httpdriver.cpp). But only for drivers that don't support the VSI virtual file system API. At the time the test was written, this was the case of the BLX driver, but since then it has been ported to VSI VFS, so the test is no longer relevant. I'd say just drop this test case

PR #4495, opened 2021-09-15
sqlite/ogrsqlitelayer.cpp: Add missing SQL data types to sqlite driver

ShapovalovKL:
## What does this PR do?

Add missing SQL datatypes (from https://www.sqlite.org/datatype3.html) to sqlite driver field type detection algorithm (BuildFeatureDefn). Now all types from sqlite docs can be recognised by gdal

Main problem was with floating-point fields. Only FLOAT and DECIMAL field types was supported.
If first row contains null in floating-point field and field type (in sqlite metadata) REAL, DOUBLE, DOUBLE PRECISION, NUMERIC data in field is processed as string
As a result floating-point fields occasionally may have string output type in ogr2ogr and data in them is rounded to 15 characters (for string fields gdal uses sqlite3_column_text which calls sqlite printf('%!.15g') for REAL storage class)

To reproduce this behaviour create sqlite database with commands bellow and convert it to another format with ogr2ogr:
create table test(col1 double);
insert into test(col1) values (null);
insert into test(col1) values (1.0000000000000002);

### DIFF
--- a/gdal/ogr/ogrsf_frmts/sqlite/ogrsqlitelayer.cpp
+++ b/gdal/ogr/ogrsf_frmts/sqlite/ogrsqlitelayer.cpp
@@ -342,7 +342,7 @@ void OGRSQLiteLayer::BuildFeatureDefn( const char *pszLayerName,
         OGRFieldType eFieldType = OFTString;
         if (pszDeclType != nullptr)
         {
-            if (EQUAL(pszDeclType, "INTEGER_BOOLEAN"))
+            if (EQUAL(pszDeclType, "INTEGER_BOOLEAN") || EQUAL(pszDeclType, "BOOLEAN"))
             {
                 oField.SetType(OFTInteger);
                 oField.SetSubType(OFSTBoolean);
@@ -373,11 +373,17 @@ void OGRSQLiteLayer::BuildFeatureDefn( const char *pszLayerName,
             {
                 oField.SetType(OFTStringList);
             }
-            else if (EQUAL(pszDeclType, "BIGINT") || EQUAL(pszDeclType, "INT8"))
+            else if (EQUAL(pszDeclType, "BIGINT") || EQUAL(pszDeclType, "INT8") || 
+                     EQUAL(pszDeclType, "UNSIGNED BIG INT"))
             {
                 oField.SetType(OFTInteger64);
             }
-            else if (STARTS_WITH_CI(pszDeclType, "INTEGER"))
+            else if (STARTS_WITH_CI(pszDeclType, "INTEGER") ||
+                     EQUAL(pszDeclType, "INT") ||
+                     EQUAL(pszDeclType, "TINYINT") ||
+                     EQUAL(pszDeclType, "SMALLINT") ||
+                     EQUAL(pszDeclType, "MEDIUMINT") ||
+                     EQUAL(pszDeclType, "INT2"))
             {
                 oField.SetType(OFTInteger);
             }
@@ -387,7 +393,11 @@ void OGRSQLiteLayer::BuildFeatureDefn( const char *pszLayerName,
                 oField.SetSubType(OFSTFloat32);
             }
             else if (EQUAL(pszDeclType, "FLOAT") ||
-                     EQUAL(pszDeclType, "DECIMAL"))
+                     EQUAL(pszDeclType, "DECIMAL") ||
+                     EQUAL(pszDeclType, "REAL") ||
+                     EQUAL(pszDeclType, "DOUBLE") ||
+                     EQUAL(pszDeclType, "DOUBLE PRECISION") ||
+                     EQUAL(pszDeclType, "NUMERIC"))
             {
                 oField.SetType(OFTReal);
             }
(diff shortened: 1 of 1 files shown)


jratike80:
I am not sure if the table 3.1.1. Affinity Name Examples is meant to be read so that SQLite clients should read all those examples literally. And even then wouldn't it be better to use the same logic than in 3.1. Determination Of Column Affinity?

SQLite does not require any data type, and it is possible to insert strings into int fiels etc.  Perhaps the problem is rather in the null value on the first row. Maybe GDAL, if is finds an unknown or non existing data type, should look for the first not null row? But if all rows are nulls then GDAL would read the whole database and all in vain.

An example about how GDAL may drop data when it tries to handle SQLite as strictly typed database.

```
create table test(col1 float);
insert into test(col1) values (1.0002);
insert into test(col1) values ('foo');
```
Table "test" is a valid SQLite table but GDAL drops some data:

```
Layer name: test
Geometry: None
Feature Count: 2
Layer SRS WKT:
(unknown)
FID Column = rowid
col1: Real (0.0)
OGRFeature(test):1
  col1 (Real) = 1.0002

Table test is a valid SQLite table but GDAL drops some data:

Warning 1: Value 'foo' of field test.col1 parsed incompletely to real 0.
OGRFeature(test):2
  col1 (Real) = 0
```

I do not mean to say that your PR is not good, just that things may not be as simple as they first appear.

ShapovalovKL:
This PR adds support for most common SQL data types with minimal side effects and loss of backward compatibility. Users creating tables in sqlite will likely choose one of types in "3.1.1. Affinity Name Examples", becose there is no other source for supportered sqlite types.

Logic from "3.1. Determination Of Column Affinity" can't be implemented completely because of custom gdal types (INTEGER_BOOLEAN, INTEGER_INT16, INTEGER_OR_TEXT, JSONINTEGERLIST, JSONINTEGER64LIST, JSONREALLIST, JSONSTRINGLIST, FLOAT_FLOAT32)
For example "1) If the declared type contains the string "INT" then it is assigned INTEGER affinity" will make JSONINTEGERLIST integer.
All this custom gdal types must be added to this logic as an exception, as a result gdal will have its own custom logic.
Plus this solution will not backward compatible, because users may have their own data-types witch for example contains "INT" which currently is treated as string.

Yes. In my opinion logic in BuildFeatureDefn is't very good, but it's hard to implement any changes to it without loss of backward compatibility
Good algorithm in my opinion should:
1) Use data in rows to detect type ONLY if user explicitly requested it
Basicly, using data from table can lead to unexpected behaviour. For example in our data we have set of similar fields, and only few of them with data, and we expect what all fields will be processed in the same way.
create table test(pos1 double,pos2 double);
insert into test(pos1,pos2) values (1,null);
insert into test(pos1,pos2) values (2,null);
2) Use fixed (documented) list of sqlite types from metadata with documented mapping to gdal types (or Logic from "3.1. Determination Of Column Affinity"). 
3) Allow user to explicitly specify which type gdal should use for each field. Sqlite driver could have dataset open option with layer and user defined types (repeated for every layer):
-oo COLUMN_TYPES=layer1:filed1:auto,filed2:DateTime -oo COLUMN_TYPES=layer2:filed1:Integer,filed2:Integer64
4) If field type is unknown (no explicitly user type and unknown type in metadata) show an error and exit
5) Add options which controls what happens if column data type doesn't match stored data
-oo DATA_TYPE_MATCHING=strict|convertible
-oo DATA_TYPE_MISMATCH_ERROR_LEVEL=error|warning|info|debug|ignore
-oo DATA_TYPE_MISMATCH_ACTION=use_parser_value|set_null|set_value:value

gdal/ogr/ogrsf_frmts/sqlite/ogrsqlitelayer.cpp
```diff
@@ -373,11 +373,17 @@ void OGRSQLiteLayer::BuildFeatureDefn( const char *pszLayerName,
...
             {
                 oField.SetType(OFTInteger64);
             }
-            else if (STARTS_WITH_CI(pszDeclType, "INTEGER"))
+            else if (STARTS_WITH_CI(pszDeclType, "INTEGER") ||
+                     EQUAL(pszDeclType, "INT") ||
+                     EQUAL(pszDeclType, "TINYINT") ||
+                     EQUAL(pszDeclType, "SMALLINT") ||
+                     EQUAL(pszDeclType, "MEDIUMINT") ||
+                     EQUAL(pszDeclType, "INT2"))
```
rouault, line 386: as a small improvement, when INT2, oField.SetSubType(OFSTInt16) could be set to add a hint this is a short integer.

jratike80:
I agree that logic of 3.1 is not a solution and adding synonyms should have minimal side effects.

I think that `4 If field type is unknown (no explicitly user type and unknown type in metadata) show an error and exit` should not be implemented that strictly. I believe that users who create tables themselves with SQLite often skip chapter 3.1.1 and do not define any data type at all because in SQLite context it is not necessary. Or maybe it is just me. This has a connection to 1 `Use data in rows to detect type ONLY if user explicitly requested it`. I would say that "detect type automatically from data if it can't be detected from the metadata".

I am not sure if the following tests are useful. Maybe they give a hint that SQLite users should not rely too much on the literal expression of the data type because SQLite engine may change it within the same affinity.

```
CREATE TABLE 'oj_affinity' ( "ogc_fid" INTEGER PRIMARY KEY AUTOINCREMENT, 'int' INTEGER, 'double' FLOAT, 'text' VARCHAR, "GEOMETRY" GEOMETRY);

create table copy as select * from oj_affinity;

All the datatype definitions have been changed into SQLite defaults

SELECT sql FROM sqlite_master WHERE type = 'table' AND name = 'test'

CREATE TABLE copy(ogc_fid INT,int INT,double REAL,text TEXT, GEOMETRY NUM).

```
Test with a view:

```
create view viewtest as select * from oj_affinity;

PRAGMA table_info("viewtest")

Column definitions are the ones used in the source table:
INTEGER, INTEGER, FLOAT, VARCHAR
```

There could also be one more option, to have a utility for sanitizing the type names. For example your problem could have been solved by sanitizing the DOUBLE into FLOAT.  We almost have such utility already, namely ogr2ogr but it cannot handle your case where the null attribute leads to wrong guess about the data type.

-oo DATA_TYPE_SANITIZER=warning|fix

Issue #4341, opened 2021-08-26
autoreconf fails: configure.ac: error: required file 'config.rpath' not found


sebastic:
## Expected behavior and actual behavior.

Building the gdal package in Debian unstable fails:
```
   dh_autoreconf
libtoolize: putting auxiliary files in '.'.
libtoolize: copying file './ltmain.sh'
libtoolize: putting macros in AC_CONFIG_MACRO_DIRS, 'm4'.
libtoolize: copying file 'm4/libtool.m4'
libtoolize: copying file 'm4/ltoptions.m4'
libtoolize: copying file 'm4/ltsugar.m4'
libtoolize: copying file 'm4/ltversion.m4'
libtoolize: copying file 'm4/lt~obsolete.m4'
libtoolize: Consider adding '-I m4' to ACLOCAL_AMFLAGS in Makefile.am.
libtoolize: 'AC_PROG_RANLIB' is rendered obsolete by 'LT_INIT'
configure.ac:49: warning: The macro `AC_TRY_COMPILE' is obsolete.
configure.ac:49: You should run autoupdate.
./lib/autoconf/general.m4:2847: AC_TRY_COMPILE is expanded from...
lib/m4sugar/m4sh.m4:692: _AS_IF_ELSE is expanded from...
lib/m4sugar/m4sh.m4:699: AS_IF is expanded from...
./lib/autoconf/general.m4:2249: AC_CACHE_VAL is expanded from...
./lib/autoconf/general.m4:2270: AC_CACHE_CHECK is expanded from...
m4/ax_cflags_warn_all.m4:26: AX_CFLAGS_WARN_ALL is expanded from...
configure.ac:49: the top level
configure.ac:50: warning: The macro `AC_TRY_COMPILE' is obsolete.
configure.ac:50: You should run autoupdate.
./lib/autoconf/general.m4:2847: AC_TRY_COMPILE is expanded from...
lib/m4sugar/m4sh.m4:692: _AS_IF_ELSE is expanded from...
lib/m4sugar/m4sh.m4:699: AS_IF is expanded from...
./lib/autoconf/general.m4:2249: AC_CACHE_VAL is expanded from...
./lib/autoconf/general.m4:2270: AC_CACHE_CHECK is expanded from...
m4/ax_cflags_warn_all.m4:71: AX_CXXFLAGS_WARN_ALL is expanded from...
configure.ac:50: the top level
[...]

PR #4499, opened 2021-09-16
Generate `./configure` with `./autogen.sh` during build process.

### LINKED ISSUE #3561: Do not put configure in git
Having to regenerate it each time configure.ac is modified is a bad practice and not intuitive for new contributors
Will require running autogen.sh to be run in the CI configurations, Docker, etc.

cmarmo:
## What does this PR do?
This pull request remove `./configure` and `aclocal.m4` from git.
They are now produced by `autogen.sh` during the building process.

Fixes #3561

## Failing github Actions
 - [x] Conda: if I understand correctly the recipe for the build should be updated on [conda-forge](https://github.com/conda-forge/gdal-feedstock/blob/master/recipe/build.sh)? 
- [x] Fedora Rawhide: a segmentation fault in the tests that I can't reproduce... if anyone has any idea... :(


cmarmo:
Removing the WIP because after reading the ref in #4341 I can't find another solution...
Thanks for considering this PR anyway if it can be useful.

rouault:
Regarding the Conda builds, I suspect that adding the generation of configure in gdal/ci/travis/conda/setup.sh should fix the build, as the Conda build copies the whole tree to do the build

the gdal/mkgdaldist.sh script should also be updated to generate configure, since we want it to be included in the official tarballs

>     * Fedora Rawhide: a segmentation fault in the tests that I can't reproduce... if anyone has any idea... :(

I've seen that in other PRs recently, so unrelated to your changes

rouault:
gdal/fuzzers/build.sh also needs to be updated

cmarmo:
>     * [x]  Conda: if I understand correctly the recipe for the build should be updated on [conda-forge](https://github.com/conda-forge/gdal-feedstock/blob/master/recipe/build.sh)?

Thanks @rouault for your suggestion: I've put the `autogen.sh` in the conda setup and the build succeeded. However I'm wondering if the conda-forge recipe is meant to work only on the gdal dist (containing the `./configure`)  or also with the git repository: in this second case  changing the recipe would be the appropriate way to build. What is the best way to made them aware of the change (if the PR will be accepted)?

>     * [x]  Fedora Rawhide: a segmentation fault in the tests that I can't reproduce... if anyone has any idea... :(

Thanks... it went away as you said.

rouault:
>  if the conda-forge recipe is meant to work only on the gdal dist (containing the `./configure`) or also with the git repository: in this second case changing the recipe would be the appropriate way to build. What is the best way to made them aware of the change (if the PR will be accepted)?

yes, once this PR is merged, a PR to https://github.com/conda-forge/gdal-feedstock will be appropriate.

I was wondering also if we shouldn't had a rule in gdal/GNUmakefile like :
```
configure: configure.ac
    ./autogen.sh
```
so that it is regenerated for developers ?

cmarmo:
> I was wondering also if we shouldn't had a rule in gdal/GNUmakefile like :
> 
> ```
> configure: configure.ac
>     ./autogen.sh
> ```
> 
> so that it is regenerated for developers ?

To be honest, once documented I don't think typing
```
$ ./autogen.sh
```
is more time-consuming than
```
$ make configure
```
and having one single option is less confusing ... at least for me....

Also in the procedure "autogen -> configure -> make" I don't like to have `make` creating `configure` ...
But I guess it's personal taste... so you are the maintainer, if you really like the idea I will add it... :)

BTW, should I squash all my messy commits? Thanks!

rouault:
> is more time-consuming than
> 
> ```
> $ make configure
> ```

ah, and I actually meant to make that configure target as a dependency of the default build target. I believe that's what automagically done by full autoconf/automake projects, but GDAL only uses automake. The idea is that you manually bootstrap with ./autogen.sh the first time you build the repository, and then what that's done, changes in configure.ac cause configure to be automatically regenerated (and ./configure automatically called). That's more a wish item than a hard requirement (as our manual makefiles that lack dependency rules will always cause the build experience to be painful), if that turns out to be too complicated.

> BTW, should I squash all my messy commits? Thanks!

I'll do that with the Squash and merge github button

cmarmo:
>  I believe that's what automagically done by full autoconf/automake projects, but GDAL only uses automake. The idea is that you manually bootstrap with ./autogen.sh the first time you build the repository, and then what that's done, changes in configure.ac cause configure to be automatically regenerated (and ./configure automatically called).

I think it could be worth it to improve `./autogen.sh` in the long term, this will also make easier packaging for distributions.

rouault:
> I think it could be worth it to improve `./autogen.sh` in the long term, this will also make easier packaging for distributions.

to be honest, the long (and hopefully not so long) term solution will be to wipe off completely current build systems to use cmake. I would have suggested not to do anything regarding that ticket if you had asked before,  but as you were 90% finished already...

cmarmo:
> to be honest, the long (and hopefully not so long) term solution will be to wipe off completely current build systems to use cmake. I would have suggested not to do anything regarding that ticket if you had asked before, but as you were 90% finished already...

Glad to know! :)

cmarmo:
Since cmake is the perspective, should this be closed or merged soon? Thanks!

rouault:
I'm merging this . Thanks!

PR #4501, opened 2021-09-17
swig/python/GNUmakefile: no longer used distutils to determine site_package_dir

### LINKED ISSUE #4334: DEP: distutils is deprecated [python]
https://github.com/OSGeo/gdal/blob/a0ba5634552735b957377b8bba6c95b2ec8bd96f/gdal/swig/python/setup.py#L16-L19

https://www.python.org/dev/peps/pep-0632/

rouault:
The fix isn't pretty. I basically copied and trimmed /usr/lib/python3.8/distutils/sysconfig.py
as I couldn't find a direct replacement for that function.

Fixes #4334

CC @snowman2 in case you have a suggestion for something more elegant

### DIFF
--- a/gdal/swig/python/trimmedsysconfig.py
+++ b/gdal/swig/python/trimmedsysconfig.py
@@ -1,0 +1,75 @@
+# Trimmed version of /usr/lib/python3.8/distutils/sysconfig.py
+
+"""Provide access to Python's configuration information.  The specific
+configuration variables available depend heavily on the platform and
+configuration.  The values may be retrieved using
+get_config_var(name), and the list of variables is available via
+get_config_vars().keys().  Additional convenience functions are also
+available.
+
+Written by:   Fred L. Drake, Jr.
+Email:        <fdrake@acm.org>
+"""
+
+import os
+import sys
+
+# These are needed in a couple of spots, so just compute them once.
+PREFIX = os.path.normpath(sys.prefix)
+EXEC_PREFIX = os.path.normpath(sys.exec_prefix)
+BASE_PREFIX = os.path.normpath(sys.base_prefix)
+BASE_EXEC_PREFIX = os.path.normpath(sys.base_exec_prefix)
+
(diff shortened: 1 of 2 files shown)


gdal/swig/python/trimmedsysconfig.py
```diff
@@ -0,0 +1,75 @@
...
+import os
+import sys
+
+# These are needed in a couple of spots, so just compute them once.
+PREFIX = os.path.normpath(sys.prefix)
+EXEC_PREFIX = os.path.normpath(sys.exec_prefix)
+BASE_PREFIX = os.path.normpath(sys.base_prefix)
+BASE_EXEC_PREFIX = os.path.normpath(sys.base_exec_prefix)
+
+def get_python_version():
```
snowman2, line 23: `from sysconfig import get_python_version`. Requires  Python 3+

gdal/swig/python/trimmedsysconfig.py
```diff
@@ -0,0 +1,75 @@
...
+    """Return a string containing the major and minor Python version,
+    leaving off the patchlevel.  Sample return values could be '1.5'
+    or '2.2'.
+    """
+    return '%d.%d' % sys.version_info[:2]
+
+
+
+def get_python_lib(plat_specific=0, standard_lib=0, prefix=None):
```
snowman2, line 32: Since GDAL already provides the prefix, some of the code in this function could be removed.

rouault: > Since GDAL already provides the prefix, some of the code in this function could be removed.

yeah, but I wanted to remain as close as the original distutils logic
(follow-up fix for https://github.com/OSGeo/gdal/issues/4516 to come)

gdal/swig/python/trimmedsysconfig.py
rouault: ok

PR #4505, opened 2021-09-17
Consider CompoundCRS for CityGML

mugwort-rc:
## What does this PR do?

Currently, it is not possible to correctly read the coordinates of the CityGML with compound CRS (such as [EPSG:6697](https://epsg.io/6697) = EPSG:6668+EPSG:6695) defined.

In the JGD2011 specification, the geometry is serialized in Lat/Lon order, but the order is reversed when a composite CRS is specified.

This happens because the `OGRSpatialReference::EPSGTreatsAsLatLong()` does not work in the case of `d->m_pjType == PJ_TYPE_COMPOUND_CRS`.

This can also be a problem with other Lat/Lon ordered coordinate systems (e.g. WGS84) that return `PJ_TYPE_COMPOUND_CRS`.

I added the handling of the CompoundCRS case to the `OGRSpatialReference::EPSGTreatsAsLatLong()`.

And I added a test case.

## Tasklist

 - [x] Consider CompoundCRS
 - [x] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:


autotest/ogr/ogr_gml_citygml_read.py
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env pytest
```
rouault, line 1: It would be good also to add a new test case in autotest/osr/osr_epsg.py that just tests the behavior of EPSGTreatsAsLatLong() on EPSG:6697

autotest/ogr/ogr_gml_citygml_read.py
```diff
@@ -0,0 +1,67 @@
...
+
+    # open CityGML file
+    gml = ogr.Open(_citygml_dir + 'compound_crs.gml')
+
+    # check number of layers
+    assert gml.GetLayerCount() == 1, 'Wrong layer count'
+
+    lyr = gml.GetLayer(0)
+    sr = osr.SpatialReference()
+    sr.ImportFromEPSG(6668)  # JGD2011
```
rouault, line 60: I'm a bit confused why this isn't EPSG:6697, given the content of the GML file ?

autotest/ogr/data/gml_citygml/compound_crs.gml
```diff
@@ -0,0 +1,106 @@
..."http://www.opengis.net/citygml/transportation/2.0" xmlns:wtr="http://www.opengis.net/citygml/waterbody/2.0" xmlns:brid="http://www.opengis.net/citygml/bridge/2.0" xsi:schemaLocation="http://www.kantei.go.jp/jp/singi/tiiki/toshisaisei/itoshisaisei/iur/uro/1.4 http://www.kantei.go.jp/jp/singi/tiiki/toshisaisei/itoshisaisei/iur/schemas/uro/1.4/urbanObject.xsd http://www.opengis.net/citygml/2.0 http://schemas.opengis.net/citygml/2.0/cityGMLBase.xsd http://www.opengis.net/citygml/landuse/2.0 http://schemas.opengis.net/citygml/landuse/2.0/landUse.xsd http://www.opengis.net/citygml/building/2.0 http://schemas.opengis.net/citygml/building/2.0/building.xsd http://www.opengis.net/citygml/transportation/2.0 http://schemas.opengis.net/citygml/transportation/2.0/transportation.xsd http://www.opengis.net/citygml/generics/2.0 http://schemas.opengis.net/citygml/generics/2.0/generics.xsd http://www.opengis.net/citygml/cityobjectgroup/2.0 http://schemas.opengis.net/citygml/cityobjectgroup/2.0/cityObjectGroup.xsd http://www.opengis.net/gml http://schemas.opengis.net/gml/3.1.1/base/gml.xsd http://www.opengis.net/citygml/appearance/2.0 http://schemas.opengis.net/citygml/appearance/2.0/appearance.xsd">
```
rouault, line 1: I would just put that file in autotest/ogr/data/gml (to try to keep a 1:1 match between directories and drivers) and call it citygml_compound_crs.gml

gdal/ogr/ogrspatialreference.cpp
```diff
@@ -11206,25 +11206,64 @@ int OGRSpatialReference::EPSGTreatsAsLatLong() const
...
-    d->undoDemoteFromBoundCRS();
-
-    if( cs )
+    if ( d->m_pjType == PJ_TYPE_COMPOUND_CRS )
     {
-        const char* pszDirection = nullptr;
-        if( proj_cs_get_axis_info(
-            d->getPROJContext(), cs, 0, nullptr, nullptr, &pszDirection,
-            nullptr, nullptr, nullptr, nullptr) )
+        for( int i = 0; i < 2; i++ )
```
rouault, line 11211: the loop is unneeded. the horizontal coordinate system will always be at index 0 for a compound one

autotest/ogr/ogr_gml_citygml_read.py
```diff
@@ -0,0 +1,67 @@
...
+###############################################################################
+# Test reading CityGML files
+###############################################################################
+
+_citygml_dir = 'data/gml_citygml/'
+
+###############################################################################
+# Test reading CityGML of Project PLATEAU
+
+def test_gml_read_compound_crs_lat_long():
```
rouault, line 50: and I would just move that test in the existing autotest/ogr/ogr_gml_read.py

autotest/ogr/ogr_gml_citygml_read.py
mugwort-rc: Sure. I added.

autotest/ogr/data/gml_citygml/compound_crs.gml
mugwort-rc: OK. I renamed it.

autotest/ogr/ogr_gml_citygml_read.py
mugwort-rc: OK. I moved it.

autotest/ogr/ogr_gml_citygml_read.py
mugwort-rc: I was worried about this too.
Is this another bug?

rouault: > Is this another bug?

not sure to remember if this is intended or not. I'd say probably not intended. there must be something somewhere that only keeps the horizontal part of the CRS. If you want to address that, that could be dealt as a follow-up fix after this PR is merged

gdal/ogr/ogrspatialreference.cpp
mugwort-rc: OK.

autotest/ogr/ogr_gml_citygml_read.py
mugwort-rc: I'll look into this later.

rouault:
Manually backported to 3.3

mugwort-rc:
Thank you!

PR #4509, opened 2021-09-19
SAFE (Sentinel1): multiple improvements related to SLC/calibration

rouault:
Contributed by Esri


rouault:
@oarcher / @agrouaze  As I see you had interest in that driver and I'm not a specialist of that domain, if you could give that a try to this pull request, that would be welcome

oarcher:
@agrouaze, what we can do is to compare the xsar calibration with this new GDAL slc calibration. Do you have some time to do this ?

agrouaze:
I can do the comparison, but not this week.

rouault:
> I can do the comparison, but not this week.

ok, I'll merge this then. But do not hesitate to provide feedback when you've a chance

PR #4520, opened 2021-09-20
Configure removal followup

rouault:
Follow up to #4499

### DIFF
--- a/gdal/.gitignore
+++ b/gdal/.gitignore
@@ -5,6 +5,8 @@ detectsse.dSYM
 gdal.pc
 config.log
 config.status
+configure
+aclocal.m4
 libtool
 cpl_config.h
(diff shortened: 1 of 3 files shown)


gdal/.gitignore
```diff
@@ -5,6 +5,7 @@ detectsse.dSYM
 gdal.pc
 config.log
 config.status
+configure
```
cmarmo, line 8: Sorry for bothering you again.... I've added `gdal/configure` in `.gitignore` in the gdal root directory... may I ask why this is needed here? Thanks

rouault: ah I missed that. this is indeed redundant. As the gdal/.gitignore is more comprehensive, I'm going to move the additions from toplevel .gitignore into gdal/.gitignore

Issue #4524, opened 2021-09-20
For GRIB2, generate an OGRSpatialReference with proj4 +lon_wrap


mmomtchev:
## Expected behavior and actual behavior.

GRIB2s are supposed to have their longitudes in [0,360]
This is (AFAIK) not possible with WKT
GDAL currently generates a normal `OGRSpatialReference` object which does not take this into account
Ideally, it should generate an object with the proj4 `lon_wrap` extension

## Steps to reproduce the problem.

```
gdallocationinfo -wgs84 /vsis3/noaa-gfs-bdp-pds/gfs.20210918/06/atmos/gfs.t06z.pgrb2.0p25.f000 -60.874 14.463
```
(if you are reading this at a later date, replace the date string)

## Operating system

All

## GDAL version and provenance

3.4.0-git

rouault:
> Ideally, it should generate an object with the proj4 `lon_wrap` extension

I wouldn't say it's ideal, as lon_wrap is a non-standard hack that doesn't fit into the ISO 19111 model.

If the issue is only with gdallocationinfo -wgs84, some logic could be added into it to adjust the longitudes
If it is more general, then the GRIB driver could potentially be modified to cut the raster into two parts, switch them and have something that fits (or is very close to) the [-180,180] range. That behavior should probably be controlled by an open option.

mmomtchev:
This would be far more complex, but it will also solve the problem of translating to GeoTIFF - one would expect that once a layer is converted to GeoTIFF it becomes a normal raster with the greenwich meridian in the middle
Currently, if a GRIBv2 band is translated to GeoTIFF, it becomes a very weird 0-360 band that must be converted with gdalwarp:
<https://gis.stackexchange.com/questions/37790/reprojecting-raster-from-0-360-to-180-180-with-cutting-180-meridian-using-gdalw>

PR #4526, opened 2021-09-21
GRIBv2 SplitAndSwap for transparent 0-360 to -180-180 translation

mmomtchev:
This is an implementation of #4524

It adds Split&Swap, a GRIBv2 reading mode, to be enabled only on regular LATLON grids to transparently rewrap the raster around the prime meridian instead of the antemeridian

It also reimplements the longitude translation for all LATLON grids - regular or rotated - emitting warnings when the translation is not possible and keeping the old behavior in these cases

I didn't touch the multidimensional raster model which does not use a geotransform anyway


gdal/frmts/grib/gribdataset.cpp
```diff
@@ -2287,15 +2309,34 @@ void GRIBDataset::SetGribMetaData(grib_MetaData *meta)
...
 
-        // Longitude origin of GRIB files is sometimes funny. Try to shift as close
-        // as possible to the traditional [-180,180] longitude range
-        // See https://trac.osgeo.org/gdal/ticket/7103
-        if( ((rMinX >= 179 && rPixelSizeX * meta->gds.Nx > 10) || rMinX >= 180) &&
+        // GRIB2 files have longitudes in the [0-360] range
+        // Shift them to the traditional [-180,180] longitude range
+        // See https://github.com/OSGeo/gdal/issues/4524
+        if (meta->GribVersion == 2 &&
+            (rMinX + rPixelSizeX >= 180 || rMaxX >= 180) &&
```
rouault, line 2316: I would change the rMaxX >= 180 to a more relaxed one. Perhaps rMaxX - rPixelSizeX >= 180

gdal/frmts/grib/gribdataset.cpp
```diff
@@ -2287,15 +2309,34 @@ void GRIBDataset::SetGribMetaData(grib_MetaData *meta)
...
         {
-            CPLDebug("GRIB", "Adjusting longitude origin from %f to %f",
-                     rMinX - rPixelSizeX / 2, rMinX - rPixelSizeX / 2 - 360 );
-            rMinX -= 360;
+            if (rPixelSizeX * nRasterXSize > 360)
+                CPLError(CE_Warning, CPLE_NotSupported,
+                    "Cannot properly handle GRIB2 files with overlaps and 0-360 longitudes");
+            else if (rPixelSizeX * nRasterXSize == 360 && meta->gds.projType == GS3_LATLON)
+            {
+                nSplitAndSwap = static_cast<int>(ceil(180 / rPixelSizeX));
```
rouault, line 2324: I don't think that formula is correct in the general case. It should probably be
```suggestion
                nSplitAndSwap = static_cast<int>(ceil((180 - rMinX) / rPixelSizeX));
```
and should be conditionalized to 180 >= rMinX

gdal/frmts/grib/gribdataset.cpp
```diff
@@ -2287,15 +2309,34 @@ void GRIBDataset::SetGribMetaData(grib_MetaData *meta)
...
+                // Find the first row number east of the antimeridian
+                CPLDebug("GRIB", "Enabling Split&Swap mode, antimeridian is at %d", nSplitAndSwap);
+                rMinX = -180;
```
rouault, line 2327: if rMinX was 0.5, then the new rMinX should be -180.5 or -179.5

gdal/frmts/grib/gribdataset.cpp
```diff
@@ -2287,15 +2309,34 @@ void GRIBDataset::SetGribMetaData(grib_MetaData *meta)
...
+        // See https://github.com/OSGeo/gdal/issues/4524
+        if (meta->GribVersion == 2 &&
+            (rMinX + rPixelSizeX >= 180 || rMaxX >= 180) && 
             CPLTestBool(CPLGetConfigOption("GRIB_ADJUST_LONGITUDE_RANGE", "YES")) )
         {
-            CPLDebug("GRIB", "Adjusting longitude origin from %f to %f",
-                     rMinX - rPixelSizeX / 2, rMinX - rPixelSizeX / 2 - 360 );
-            rMinX -= 360;
+            if (rPixelSizeX * nRasterXSize > 360)
+                CPLError(CE_Warning, CPLE_NotSupported,
```
rouault, line 2320: I would probably just emit a debug message.

gdal/frmts/grib/gribdataset.cpp
```diff
@@ -826,9 +838,17 @@ CPLErr GRIBRasterBand::IReadBlock( int /* nBlockXOff */,
 
     const int nCopyWords = std::min(nRasterXSize, nGribDataXSize);
 
+    // Split&Swap: transparent rewrap around the prime meridian instead of the antimeridian
```
rouault, line 841: The case where nRasterXSize != nGribDataXSize is a messy one currently (the driver truncates the exposed raster. it should rather use a subdataset mechanism), and I've the feeling that split&swap strategy will produce complete garbage. If we can detect in the SetGribMetaData() function that nRasterXSize != nGribDataXSize, we should disable the split&swap strategy

gdal/frmts/grib/gribdataset.h
```diff
@@ -109,6 +109,7 @@ class GRIBDataset final: public GDALPamDataset
     GIntBig nCachedBytes;
     GIntBig nCachedBytesThreshold;
     int bCacheOnlyOneBand;
+    int nSplitAndSwap;
```
rouault, line 112: A comment explaining the semantics of this variable, and perhaps renaming it to nsplitAndSwapColumn, would be welcome

gdal/frmts/grib/gribdataset.cpp
```diff
@@ -2287,15 +2309,34 @@ void GRIBDataset::SetGribMetaData(grib_MetaData *meta)
...
             rPixelSizeY = meta->gds.Dy;
 
-        // Longitude origin of GRIB files is sometimes funny. Try to shift as close
-        // as possible to the traditional [-180,180] longitude range
-        // See https://trac.osgeo.org/gdal/ticket/7103
-        if( ((rMinX >= 179 && rPixelSizeX * meta->gds.Nx > 10) || rMinX >= 180) &&
+        // GRIB2 files have longitudes in the [0-360] range
+        // Shift them to the traditional [-180,180] longitude range
+        // See https://github.com/OSGeo/gdal/issues/4524
+        if (meta->GribVersion == 2 &&
```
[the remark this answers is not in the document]
mmomtchev: No, it is one of the major changes of GRIBv1 vs GRIBv2

gdal/frmts/grib/gribdataset.cpp
mmomtchev: It is conditionalized (these are nested ifs) and in this special case I set rMinX = -180 hardcoded + the half pixel adjustment a few lines below
This is in fact the special case (which is in fact the most common case) of a GRIBv2 that covers the whole world - this one I translate to -180 to 180

But if I am to make antimeridian-spanning GeoTIFFs, this needs generalization

gdal/frmts/grib/gribdataset.cpp
mmomtchev: Yes, I was wondering about that comparison

gdal/frmts/grib/gribdataset.cpp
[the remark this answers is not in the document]
rouault: maybe, but does that hurt to remove to remove that GRIB2 check ? There's nothing in the below logic that is GRIB2 specific as far as I can tell

gdal/frmts/grib/gribdataset.cpp
rouault: > It is conditionalized (these are nested ifs)

I may miss something, but it seems to be that if we have for example rMinx = 90 and rMaxX = 90 + 360, we would go here, hence we cannot assume that rMinX = 0. Or if we want to limit to that case, it should be a condition of that branch

gdal/frmts/grib/gribdataset.cpp
```diff
@@ -824,11 +836,22 @@ CPLErr GRIBRasterBand::IReadBlock( int /* nBlockXOff */,
     if( nBlockYOff >= nGribDataYSize )  // Off image?
         return CE_None;
 
+    if ( nRasterXSize != nGribDataXSize && poGDS->nSplitAndSwapColumn != 0)
```
rouault, line 839: could that check be moved to where nSplitAndSwapColumn is set ? Otherwise the georeferencing exposed will be inconsistant with the raster returned

gdal/frmts/grib/gribdataset.cpp
rouault: I believe my above point is still valid and not addressed

mmomtchev: Unless I am missing something, when this statement is evaluated, those hold true:
```
rMinX == 0
rMinX + rPixelSizeX >= 180 || rMaxX - rPixelSizeX >= 180
```

rouault: can you point where rMinX == 0 is checked ?

mmomtchev: I check(ed) for `rPixelSizeX * nRasterXSize == 360` which (normally) would mean `rMinX == 0` and then I was setting it to -180 hardcoded value
But indeed, maybe there could have been an (invalid) GRIB going from lets says 2 to 362 that was not going to work with that - but that GRIB would have been broken anyway

mmomtchev:
@rouault I will need to find unit tests for all the partial cases because there is a problem with this

gdal/frmts/grib/gribdataset.cpp
rouault: > The problem is that the GRIB is read sequentially so it could easily be the last rasterband that will trigger this condition

hum, that's a good point... anyway handing of such files is a bit broken

> do you happen to have a GRIB file with this?

9d629afc0b983fcec70a9c5a85c92db429ed4996 points to https://trac.osgeo.org/gdal/ticket/3246

rouault:
Is it ready for merge ?

mmomtchev:
No, still looking for a solution for that case where a single subset has a different size - the file from the old bug does not have world coverage so it still works - but there might be other cases where it won't

gdal/frmts/grib/gribdataset.cpp
mmomtchev: @rouault in fact, no matter what one does, the georeferencing of a band that doesn't match its dataset size will be incorrect
In the file from the old bug there is only one band that has a very different shape from the dataset (28x21 vs 103x78)
I wonder what you are supposed to do with this band - one can still retrieve the data (I will disable split&swap only for this band) but there won't be a valid geotransform anyway

mmomtchev: @rouault I think this dataset is corrupted, there is no other explanation
This is a sea swell grib and the weird band contains the U-wind, while the V-wind is in a normal bind
There is absolutely no reason for the U-wind to have a different shape from the V-wind
U and V are the weather science way of expressing the wind (separately on the NS and EW axis) instead of a speed and angle so that interpolation and resampling can work in a mathematically natural way
I have no explanation for this other that the dataset is corrupted

mmomtchev: especially since the values contained for the V wind look like wind, while U is total garbage

rouault: beyond the brokenness of that particular example, we should ideally isolate in separate subdatasets GRIB messages that have different sizes/georeferencing, but that might be tricky to do in a performant way since currently we get the actual size in LoadData(). I guess that at least the size could be obtained in a cheaper way without decompressing the grid, but not sure we have a degrib API for that. Anyway this is a bit out of topic of this PR. One issue at a time :-)

PR #4545, opened 2021-09-24
CPL logging mechanism: fix writing non-ASCII UTF-8 filenames on Windows (fixes #4492)

### LINKED ISSUE #4492: CPL_LOG does not handle unicode filenames correctly

I was just going through the file handling in a consuming Windows application, and found a call site that sets the GDAL property CPL_LOG. Looking into the current GDAL master source code, this config option is used in CPLDefaultErrorHandler() as an input to fopen().

Since the C stdlib functions use the Windows code page settings, it seems to me that the log file path can easily get mangled in the process. 

I assume, setting CPL_LOG to something with proper Unicode would do.

## Operating system

Any modern Windows.

Code is still present in master.

rouault:


rouault:
Manually cherry-picked to 3.3

PR #4547, opened 2021-09-25
MEMDataset::Open functions adds support for the 'GEOTRANSFORM' options

sotex:
MEM: Open functions adds support for the 'GEOTRANSFORM' options

When using `"MEM:::DATAPOINTER=342343408,PIXELS=100,LINES=100,BANDS=3,DATATYPE=Byte,PIXELOFFSET=3,LINEOFFSET=300,BANDOFFSET=1"` to create a memory data When creating a memory data set, it supports passing in spatial coordinate information.

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS: linux / windows 7 later
* Compiler: gcc 8.3 / msvc 14

### DIFF
--- a/gdal/doc/source/drivers/raster/mem.rst
+++ b/gdal/doc/source/drivers/raster/mem.rst
@@ -32,14 +32,16 @@ For example:
 ::
 
      MEM:::DATAPOINTER=342343408,PIXELS=100,LINES=100,BANDS=3,DATATYPE=Byte,
-          PIXELOFFSET=3,LINEOFFSET=300,BANDOFFSET=1
+          PIXELOFFSET=3,LINEOFFSET=300,BANDOFFSET=1,
+          GEOTRANSFORM=1.166396e+02/1.861068e-05/0.000000e+00/3.627969e+01/0.000000e+00/-1.861068e-05
 
 or
 
 ::
 
      MEM:::DATAPOINTER=0x1467BEF0,PIXELS=100,LINES=100,BANDS=3,DATATYPE=Byte,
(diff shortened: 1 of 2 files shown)


rouault:
Why do you need that ? You could just call SetGeoTransform() on the dataset returned by GDALOpen()

sotex:
> Why do you need that ? You could just call SetGeoTransform() on the dataset returned by GDALOpen()

Yes, this is sufficient for most situations.

I ran into a situation where an image processing library only supports passing in an image `filename`, which needs to be with coordinates. When the pixel data of the image is already in memory, then you need to use `GDALCreate` to generate a `MEMDataset` with path `/mem/filename` in order to share it, **but this will make a copy**. If you just use a path like `MEM:::DATAPOINTER=342343408,PIXELS=100,LINES=100,BANDS=3,DATATYPE=Byte,PIXELOFFSET=3,LINEOFFSET=300,BANDOFFSET=1`, then coordinate information cannot be passed in.

rouault:
ok, thanks for the rationale. I'll take your patch as it is non intrusive.

Another possibility would be to use a on-the-fly VRT (https://gdal.org/drivers/raster/vrt.html) wrapping a MEM::: source. You would need to set the VRT_ALLOW_MEM_DRIVER configuration option / environment variable to YES, since opening a MEM::: source from VRT is disabled by default due to security concerns.

Issue #4483, opened 2021-09-14
WCS performing wrong request when dealing with two equal-length negative numbers


albertoparadisllop:
#### Expected behavior

When performing a WCS Request on an EPSG:4326 coverage, when the minimum requested extent and the coverage's minimum extent are both negative and have the same number of characters, for example, `-5` and `-9`, the maximum of the two will be used when performing the request: `-5`

#### Actual behavior

When performing the request mentioned earlier, the minimum of the two is used instead. For `-5` and `-9`, `-9` is used.
Any attempt to request a coverage with extents as describer earlier results in the following error:
```
ERROR 1: Returned tile does not match expected configuration.
Got 6266x288 instead of 288x288.
```
## Steps to reproduce the problem.

`gdal_translate -projwin -0.15, 39.5, 0.15, 39.2 -oo NoGridAxisSwap "WCS:https://<server_url>?version=2.0.1&coverage=<coverage_name>" out.tiff`

Where coverage `<coverage_name>` in `<server_url>` has a minimum longitude of X, where `-10 < X < -0.15`

## Operating system

Ubuntu 20.04.1 LTS 64 bit

## GDAL version

GDAL 3.1.4

pixeledbrain:
I suspect it might have to do with function  [`CompareNumbers` in wcsutils](https://github.com/OSGeo/gdal/blob/master/gdal/frmts/wcs/wcsutils.cpp#L47-L88), as it seems to do a string comparison of both inputs to determine the result, so for `-1` and `-5`, for example, string-wise, `-5` will be bigger.

rouault:
CC @ajolma

ajolma:
Can you give me a working server url and coverage name?

albertoparadisllop:
> Can you give me a working server url and coverage name?

I can't give you the one I used, as it's a company server, however, I found this public one in which I was able to recreate the issue:

WCS Server: `https://servicios.idee.es/wcs-inspire/mdt?version=2.0.1&service=WCS`
Coverage: `Elevacion4326_1000`

To recreate the issue, the requested projwin must have a minimum longitude of the same *number of digits* as the extent's minimum longitude. I recreated it requesting the following projwin: `[-10.28, 39.58, -9.226, 38.84]`

ajolma:
I really don't understand/remember why the CompareNumbers is implemented in such a complex way. If I replace its contents with 
    double ad = std::stod(a);
    double bd = std::stod(b);
    if (ad < bd) {
[...]

PR #4550, opened 2021-09-26
Replace CompareNumbers with in-place numeric comparisons.

ajolma:
## What does this PR do?

Fixes(?) the CompareNumbers fct in wcsutils.cpp

#4483

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:

### DIFF
--- a/gdal/frmts/wcs/wcsutils.cpp
+++ b/gdal/frmts/wcs/wcsutils.cpp
@@ -32,6 +32,7 @@
 #include "ogr_spatialref.h"
 #include "wcsutils.h"
 #include <algorithm>
+#include <string>
 
 #define DIGITS "0123456789"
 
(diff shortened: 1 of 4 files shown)


ajolma:
The Fedora test failure seems to be due to differences between numerical accuracy in generated URLs (a bit stupid test).

gdal/frmts/wcs/wcsutils.cpp
```diff
@@ -46,45 +47,15 @@ void Swap(double &a, double &b)
...
-            a_p = a_p + "0";
-        }
-    } else if (d > 0) {
-        for (int i = 0; i < d; ++i) {
-            b_p = b_p + "0";
-        }
-    }
-    c = a_p.compare(b_p);
-    if (c < 0) {
+    double ad = std::stod(a);
```
rouault, line 50: if a or b might not be representable as double (which might be the case when looking at the callers where we trust the value of some XML element), std::stod() can throw an exception. Perhaps use CPLAtof(a.c_str()) instead.
Actually this functions seems overkill. It would be best replaced by more straighforward comparisons in the 4 call places where it is used (which would save double to string conversions)

rouault:
Perhaps the line "if gdaltest.is_travis_branch('s390x') or gdaltest.is_travis_branch('graviton2') or gdaltest.is_travis_branch('ubuntu_2004')" in wcs.py is no longer needed ?

PR #4552, opened 2021-09-26
GeoJSON: expose field names in a more consistent orders when some feature lacks some fields (refs qgis/QGIS#45139)

rouault:
Detailed rationale of the fix: http://erouault.blogspot.com/2021/09/order-of-geojson-properties-and-graph.html

### DIFF
--- a/gdal/ogr/ogrsf_frmts/geojson/directedacyclicgraph.hpp
+++ b/gdal/ogr/ogrsf_frmts/geojson/directedacyclicgraph.hpp
@@ -1,0 +1,253 @@
+/******************************************************************************
+ *
+ * Project:  GDAL
+ * Purpose:  Implementation of topologic sorting over a directed acyclic graph
+ * Author:   Even Rouault
+ *
+ ******************************************************************************
+ * Copyright (c) 2021, Even Rouault <even dot rouault at spatialys dot com>
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a
+ * copy of this software and associated documentation files (the "Software"),
+ * to deal in the Software without restriction, including without limitation
+ * the rights to use, copy, modify, merge, publish, distribute, sublicense,
+ * and/or sell copies of the Software, and to permit persons to whom the
+ * Software is furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included
+ * in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+ * OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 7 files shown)


gdal/ogr/ogrsf_frmts/geojson/directedacyclicgraph.hpp
```diff
@@ -0,0 +1,253 @@
...
+        return "node i unknown";
+    }
+    if( nodes.find(j) == nodes.end() )
+    {
+        return "node j unknown";
+    }
+
+    if( isTherePathFromTo(j, i) )
+    {
+        return "can't add edge: this would cause a cycle";
```
Komzpa, line 130: should you drop both instead and sort these two by some other means? this line will keep the attribute order feature-order-dependant still.

rouault: good point, but what would you drop if we have edges A->B and B->C, and we now want to insert C->A ?

Komzpa: all three as they can't be treated as definitions? but you can't do that on this same pass as then next feature will push them back. you generally need to build cyclic graph and then cut off cycles explicitly.

other strategies that come to mind:
 - push the edges into a set first, order it alphabetically as tuples, push into current dag. will keep alphabetical ordering in case of conflicts.
 - count number of occurences for each edge, sort by that then alphabetically, push into current dag. will keep most common order in case of conflicts.
  - say "it will be this way" and do nothing, maybe put a note. current one still works. :)

Komzpa: I guess first option can also be simplified to "edge with from<to beats all conflicting edges with from>to". So in your example C->A is dropped in any order of insertion.

rouault: I believe I will keep suggested improvements (thanks for them) for a next time. Conflicting order is a less common case, and cannot be generated with GDAL.

gdal/ogr/ogrsf_frmts/geojson/ogrgeojsonreader.cpp
```diff
@@ -2108,19 +2181,43 @@ bool OGRGeoJSONBaseReader::GenerateFeatureDefn( OGRLayer* poLayer,
...
+                                             it.key, it.val,
                                              bFlattenNestedAttributes_,
                                              chNestedAttributeSeparator_,
                                              bArrayAsString_,
                                              bDateAsString_,
                                              aoSetUndeterminedTypeFields_);
+            for( int idx: anCurFieldIndices )
+            {
+                dag.addNode(idx, apoFieldDefn[idx]->GetNameRef());
+                if( nPrevFieldIdx != -1 )
```
Komzpa, line 2215: if you ever decide to do edge-counting, the n^2 loop with all pairs and not just adjacent neighbors will get more stable voting.

rouault: I've added a commit with a comment in the code pointing to that PR if we need those enhancements

PR #4558, opened 2021-09-27
libmdbodbc package is not present on Debian

faridcher:
With the new mdbtools, Ogr works directly on the mdb files as in Windows.

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:

### DIFF
--- a/gdal/doc/source/drivers/vector/pgeo.rst
+++ b/gdal/doc/source/drivers/vector/pgeo.rst
@@ -87,13 +87,20 @@ details <http://trac.osgeo.org/gdal/wiki/mdbtools>`__
 
 Prerequisites
 ~~~~~~~~~~~~~
+.. role:: bash(code)
+   :language: bash
 
 #. Install `unixODBC <http://www.unixodbc.org>`__ >= 2.2.11
 #. Install MDB Tools. The official upstream of MDB Tools is maintained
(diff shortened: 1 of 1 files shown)


rouault:
You still need to create a odbcinst.ini configuration file to point to the libmdbodbc. so, right ? Looking at https://packages.debian.org/sid/amd64/odbc-mdbtools/filelist, I see it is installed in /usr/lib/x86_64-linux-gnu/odbc/libmdbodbc.so which I suspect to not be looked for by default

faridcher:
With changing the default `/etc/odbcinst.ini`, `~/odbcinst.ini` is created once `ogrinfo sample.mdb` is run and is filled with:

```
[Microsoft Access Driver (*.mdb)]
Driver=/usr/lib/x86_64-linux-gnu/odbc/libmdbodbc.so
FileUsage=1
UsageCount=3
```

I tried to update /etc/odbcinst.ini with the above content to avoid it's creation in the home directory but it persists.

rouault:
my mistak:  /usr/lib/x86_64-linux-gnu/odbc is actually looked for in recent GDAL versions. And since 279a37525abc9b4ccc44ec3c40f95772817ad3ca, the content of /etc/odbcinst.ini should be used if there's a mention of the Access driver in it.

PR #4573, opened 2021-09-30
Add support for JPEG-XL codec when using internal libtiff

tbonfort:
This PR adds support for JPEG-XL as a supported codec to the GTiff driver.

https://jpegxl.info/
https://jpeg.org/jpegxl/
https://github.com/libjxl/libjxl

The jpeg-xl codestream is frozen. The API exposed by the libjxl reference implementation is not yet feature-complete and may require some future changes to this patch. I will make sure the supported gdal branches compile correctly with libjxl API until it finalizes.

original patch by @rouault


gdal/doc/source/drivers/raster/gtiff.rst
```diff
@@ -474,6 +476,16 @@ Creation Options
    tile/strip with lossless WebP compression, so you may want to increase the
    BLOCKYSIZE value for strip layout.
 
+-  **JXL_LOSSLESS=YES/NO**: Set whether JPEG-XL compression should be lossless
+   (YES, default) or lossy (NO)
+
+-  **JXL_EFFORT=[1-9]**: Level of effort for JPEG-XL compression.
```
vincentsarago, line 482: @tbonfort what do you think about `JXL_LEVEL` to be somehow similar to other compression schema options (ZLEVEL, ZSTD_LEVEL)?

gdal/doc/source/drivers/raster/gtiff.rst
```diff
@@ -474,6 +476,16 @@ Creation Options
    tile/strip with lossless WebP compression, so you may want to increase the
    BLOCKYSIZE value for strip layout.
 
+-  **JXL_LOSSLESS=YES/NO**: Set whether JPEG-XL compression should be lossless
```
vincentsarago, line 479: same as 👇 For webp we use `TRUE/FALSE` 🤷‍♂️

rouault: > For webp we use `TRUE/FALSE` man_shrugging

TRUE/FALSE, YES/NO, 1/0, ON/OFF all work for those boolean options

gdal/doc/source/drivers/raster/gtiff.rst
tbonfort: I agree it would be more consistent with the existing. However the existing is already inconsistent (i.e. a higher WEBP_LEVEL will give you a heavier image, but a higher ZSTD_LEVEL will give you a lighter one). The jxl ecosystem has already committed to using the "effort" keyword for this (with cute names:  1:lightning 2:thunder 3:falcon
 * 4:cheetah 5:hare 6:wombat 7:squirrel 8:kitten 9:tortoise), I'd be inclined to follow that in gdal

.github/workflows/ubuntu_20.04/build.sh
```diff
@@ -48,6 +48,7 @@ cd /build/gdal
     --with-tiledb \
     --with-crypto \
     --with-ecw=/opt/libecwj2-3.3 \
+    --with-jxl
```
rouault, line 51:
```suggestion
    --with-jxl \
```

gdal/frmts/gtiff/tif_jxl.c
```diff
@@ -0,0 +1,798 @@
...
+
+        if(td->td_planarconfig == PLANARCONFIG_SEPARATE) {
+            format.num_channels = 1;
+            basic_info.num_color_channels = 1;
+            basic_info.num_extra_channels = 0;
+            basic_info.alpha_bits = 0;
+            basic_info.alpha_exponent_bits=0;
+        } else {
+            format.num_channels = td->td_samplesperpixel;
+            switch(td->td_samplesperpixel) {
```
rouault, line 512: it is a bit dangerous to imply that 2 or 4 mean that the last channel is alpha. It would better to test ```td->td_extrasamples > 0 && td->td_sampleinfo[td->td_extrasamples-1] == EXTRASAMPLE_UNASSALPHA``` . This whole switch() could then be removed.  But perhaps some other checks would be needed if the JXL codec has limitations related to number of color bands and alpha channel

tbonfort: Correct. For the lossy case, libjxl expects G, G+A, RGB or RGB+A, and this should probably be tested. For the lossless case, we can be less regardant as the bands will not be modified and can thus represent anything.
So, in the lossy case, we should also be checking that bands 1,2,3 are r,g,b, is that something we have available at that point in the code? if not, it doesn't make much sense to be checking that band 4 is alpha, and the documentation should state that lossy should only be applied to G, G+A, RGB or RGB+A

rouault: > So, in the lossy case, we should also be checking that bands 1,2,3 are r,g,b, is that something we have available at that point in the code?

TIFF doesn't have a per-band colorimetric information unfortunately (that's a GDAL addition in its GDAL_METADATA XML tag), but you can safely check td->td_photometric == PHOTOMETRIC_RGB that will be valid for R,G,B and R,G,B,A.   And for grey you can test PHOTOMETRIC_MINISBLACK

tbonfort: So we have three options:
- document that lossy should only be used for G, G+A, RGB or RGB+A and trust the user to have read and understood that
- emit a warning (with an option to silence it? how?)
- bail out with an error (with an option to silence it? how?)

any preferences ?

rouault: I would go for documenting + emit a warning without any option to silence

rouault: actually, what happens if you go outside of G, G+A, RGB or RGB+A  in lossy ? if that's pure garbage, perhaps emit an error rather than a warning

tbonfort: I haven't really used lossy. A quick test shows it's mostly garbage

tbonfort: AGRB: Screenshot from 2021-09-30 17-51-42
AGRB after compression (distance=1.2): Screenshot from 2021-09-30 17-51-58
)

rouault: let's go for a TIFFErrorExt() then

tbonfort: addressed in 5cf3bbd, is that OK ?

gdal/frmts/gtiff/tif_jxl.c
```diff
@@ -0,0 +1,826 @@
...
+            TIFFErrorExt(tif->tif_clientdata, module,
+                "JXL: Only up to 4 samples are supported for Planarconfig=Contig");
+            return 0;
+        }
+
+        if( !sp->lossless && td->td_planarconfig == PLANARCONFIG_CONTIG )
+        {
+            switch(td->td_samplesperpixel) {
+                case 2:
+                    if(td->td_sampleinfo[1] != EXTRASAMPLE_UNASSALPHA) {
```
rouault, line 419: td_sampleinfo[] is not necessarily td_samplesperpixel large:
```suggestion
                    if( !(td->td_extrasamples > 0 && td->td_sampleinfo[td->td_extrasamples-1] == EXTRASAMPLE_UNASSALPHA) ) {
```

gdal/frmts/gtiff/tif_jxl.c
```diff
@@ -0,0 +1,826 @@
...
+                case 3:
+                    if( td->td_photometric != PHOTOMETRIC_RGB ) {
+                        TIFFErrorExt(tif->tif_clientdata, module,
+                            "JXL: For 3-band images in lossy mode bands must be RGB");
+                        return 0;
+                    }
+                    break;
+                case 4:
+                    if( td->td_photometric != PHOTOMETRIC_RGB ||
+                        td->td_sampleinfo[3] != EXTRASAMPLE_UNASSALPHA) {
```
rouault, line 434:
```suggestion
                        !(td->td_extrasamples > 0 && td->td_sampleinfo[td->td_extrasamples-1] == EXTRASAMPLE_UNASSALPHA)) {
```

rouault:
@tbonfort is it good to go ?

tbonfort:
@rouault : :+1: (the travis failure seems unrelated)

tbonfort:
many thanks @rouault for the implementation and the review :fireworks:

gdal/frmts/gtiff/tif_jxl.c
tbonfort: trying this out right now, this doesn't seem correct after all. When converting a 4 band rgba webp to lossy jxl, I have td->td_extrasamples set to 0. ColorInterp=Alpha is correctly set on the 4th webp band.

rouault: is PHOTOMETRIC_RGB set ? this might be a  non 8-bit image. I'm not sure the logic to automatically set PHOTOMETRIC_RGB  works currently for non 8-bit.

tbonfort: td_photometric is set to PHOTOMETRIC_RGB, and the data is 8 bit

gdal/frmts/gtiff/tif_jxl.c
```diff
@@ -0,0 +1,829 @@
...
+        if (JXL_ENC_SUCCESS != JxlEncoderSetBasicInfo(enc, &basic_info))
+        {
+            TIFFErrorExt(tif->tif_clientdata, module,
+                         "JxlEncoderSetBasicInfo() failed");
+            JxlEncoderDestroy(enc);
+            return 0;
+        }
+
+        JxlColorEncoding color_encoding = {};
+        JxlColorEncodingSetToLinearSRGB(&color_encoding, /*is_gray*/
```
tbonfort, line 609: @tbonfort this needs to be set to `JxlColorEncodingSetToSRGB`

gdal/frmts/gtiff/tif_jxl.c
tbonfort: actually, with the fix at https://github.com/OSGeo/gdal/pull/4573/files#r720322668 , the encoded output seems quite correct while testing with BRGA

PR #4585, opened 2021-10-02
Adds docutils to doc/requirements.txt

mmomtchev:
## What does this PR do?

Adds docutils to `doc/requirements.txt` because for some reason `sphinx` does not pull it automatically


rouault:
This file has not been updated in ages. Wondering if we shouldn't remove the pin to particular versions ? The sphinx one is really old compared to what  we use in the actual build of the online site done with doc/.azure-pipelines.yml which uses the osgeo/proj-docs docker image (https://github.com/OSGeo/PROJ/blob/master/docs/docbuild/Dockerfile , regenerated regularly I believe)

mmomtchev:
Oh, its the version pinning in fact

PR #4587, opened 2021-10-02
pct2rgb: fix parsing arguments (release/3.3)

agiudiceandrea:
## What does this PR do?

Fixes an oversight in the pct2rgb.py arguments parsing logic, introduced with https://github.com/OSGeo/gdal/commit/33374e4545101c36baa50d1434d095c964d53817 (https://github.com/OSGeo/gdal/pull/3131), which prevents pct2rgb.py provided by GDAL 3.3.x to properly work.

It seems the logic in master branch is totally different, and I'm not sure if the issue is also present in master.

Issue https://github.com/qgis/QGIS/issues/45344 reported for QGIS 3.20.3 on Windows (GDAL 3.3.1).

### DIFF
--- a/gdal/swig/python/gdal-utils/osgeo_utils/pct2rgb.py
+++ b/gdal/swig/python/gdal-utils/osgeo_utils/pct2rgb.py
@@ -68,7 +68,7 @@ def main(argv):
             i = i + 1
             driver = argv[i]
 
-        if arg == '-ct':
+        elif arg == '-ct':
             i = i + 1
             pct_filename = argv[i]
 


rouault:
> It seems the logic in master branch is totally different, and I'm not sure if the issue is also present in master.

I don't think it is present in master. master uses a more robust way of specifying and parsing arguments

agiudiceandrea:
Thanks @rouault!
However, it seems the `-ct` option of pct2rgb.py introduced with 33374e4 (#3131) is not documented. Isn't it?

rouault:
> t seems the `-ct` option of pct2rgb.py introduced with [33374e4](https://github.com/OSGeo/gdal/commit/33374e4545101c36baa50d1434d095c964d53817) (#3131) is not documented. Isn't it?

indeed, doc is missing

PR #4590, opened 2021-10-04
RFC 84: Migrating build systems to CMake

rouault:
Rendered document at https://github.com/rouault/gdal/blob/rfc84_cmake/gdal/doc/source/development/rfc/rfc84_cmake.rst


gdt:
This is missing a requirements section, that states what the new build system is expected to do before it is considered ready, i.e. preconditions for removing autoconf.   I think that includes at least
 * build system works on all systems where the autoconf build is known to work (even if not in CI)
 * supports objdir builds
 * supports cross builds
 * cmakefiles do not contain "ifdef FooOS", except possibly to add a workaround instead of assuming POSIX when FooOS is known not to meet posix.    For any optional things, they need to have feature tests.  (autoconf culture is to use feature tests, and in packaging various things I have come across many many OS-specific ifdefs that grow to a long list.)
 * there's been a formal release with autoconf and cmake where cmake meets the above requirements, as verified by packager feedback.

PR #4595, opened 2021-10-05
GTiff: do not try to export a DerivedGeographic CRS to GeoTIFF, but fallback to PAM

rouault:


rouault:
manually cherry-picked to 3.3 per fd4a9f83330552d7da53a022f5f791e0bb65f4f8

PR #4597, opened 2021-10-05
Adding a test for retry of range requests

sgillies:
I don't yet understand webserver.SequentialHandler well enough to be sure it handles these properly or if I'm using it properly.

### DIFF
--- a/autotest/gcore/vsicurl.py
+++ b/autotest/gcore/vsicurl.py
@@ -507,6 +507,29 @@ def test_vsicurl_test_retry():
         assert '429' in error_msg
 
 
+def test_vsicurl_test_range_retry():
+
+    if gdaltest.webserver_port == 0:
+        pytest.skip()
+
+    handler = webserver.SequentialHandler()
+    handler.add('GET', '/test_retry/', 404)
+    handler.add('HEAD', '/test_retry/test.txt', 200, {'Content-Length': '6'})
+    handler.add('GET', '/test_retry/test.txt', 502, {'Content-Range': 'bytes 3-5/6'})
+    handler.add('GET', '/test_retry/test.txt', 429, {'Content-Range': 'bytes 3-5/6'})
+    handler.add('GET', '/test_retry/test.txt', 200, {'Content-Range': 'bytes 3-5/6'}, 'bar')
+    with webserver.install_http_handler(handler):
+        f = gdal.VSIFOpenL('/vsicurl?max_retry=2&retry_delay=0.01&url=http://localhost:%d/test_retry/test.txt' % gdaltest.webserver_port, 'rb')
+        assert f is not None
+        gdal.ErrorReset()
+        with gdaltest.error_handler():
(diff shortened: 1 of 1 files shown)


rouault:
Note: It would be best to push the branch in your fork, to avoid builds being done as a branch and a PR here

autotest/gcore/vsicurl.py
```diff
@@ -507,6 +507,29 @@ def test_vsicurl_test_retry():
...
 
+def test_vsicurl_test_range_retry():
+
+    if gdaltest.webserver_port == 0:
+        pytest.skip()
+
+    handler = webserver.SequentialHandler()
+    handler.add('GET', '/test_retry/', 404)
+    handler.add('HEAD', '/test_retry/test.txt', 200, {'Content-Length': '6'})
+    handler.add('GET', '/test_retry/test.txt', 502, {'Content-Range': 'bytes 3-5/6'})
```
rouault, line 518: due to the chunk caching logic which tries to 16384 bytes at once, I bet the Content-Range would be 'bytes 0-5/6'

sgillies:
I really ought to teach myself how to run the autotests locally :frowning_face:

rouault:
> I really ought to teach myself how to run the autotests locally frowning_face

https://github.com/OSGeo/gdal/blob/master/CONTRIBUTING.md should hopefully help

sgillies:
I've tripped over the vsi cache left over from a previous test. I'll use different URLs. And close this PR so we don't get double builds.

rouault:
> I've tripped over the vsi cache left over from a previous test.

you can use gdal.VSICurlClearCache()

sgillies:
@rouault thanks for the help, I appreciate it! Not a huge priority, I'm only digging in GDAL because I'm unable to get the vsicurl machinery to retry 503 responses to range requests within ChunkAndWarpMulti/Image in my rasterio tests. And I don't have any other data that confirms that range requests do get retried, and lots of reports of intermittent failures. Specifically, here is a test server that I've written for rasterio that returns a 503 error if we make a range request for a "bad byte" https://github.com/mapbox/rasterio/pull/2305/files#diff-44e05952ec6123426b8022744c1b6d5e5467c03eee0c9f20527ff973cbc21cdfR1949. Even if I specify retries like at https://github.com/mapbox/rasterio/pull/2305/files#diff-6ac3f79fc25d95cd1e3d51da53a4b21b939437392578a35ae8cd6d5366ca5485R88 I don't see a retry in the server logs.

rouault:
> I don't see a retry in the server logs.

Just by looking at the code, I don't see any obvious error in your test or in GDAL's code

sgillies:
It turns out that my "broken" server was broken :) I was calling send_error too late. Having fixed that I can confirm that range requests are retried and I'm seeing more confirmation in a project at work. Thanks for the feedback @rouault !

rouault:
I've filed https://github.com/OSGeo/gdal/issues/4610 so that we have a note to add such a test

PR #4615, opened 2021-10-07
JPIPKAK: support jpip nitf subtarget

ad960009:
skip metadata boxes error return

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:

### DIFF
--- a/gdal/frmts/jpipkak/jpipkakdataset.cpp
+++ b/gdal/frmts/jpipkak/jpipkakdataset.cpp
@@ -759,7 +759,7 @@ int JPIPKAKDataset::Initialize(const char* pszDatasetName, int bReinitializing )
 
     bool bSubtarget = false;
     if(osURL.ifind("?subtarget=") == std::string::npos && osURL.ifind("&subtarget=") == std::string::npos)
-	    bSubtarget = true;
+        bSubtarget = true;
 
     GDALJP2Metadata oJP2Geo;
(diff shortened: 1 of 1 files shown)


ad960009:
> I think I was the original author of this driver and I was doing something similar. When using JPIP against any file format, e.g. GRIB, NITF or something else it is the role of the server to present this as a JPIP stream not the client to realize it is a NITF. The subtargets of a NITF should be the `codestream=n` part of the JPIP spec.
> 
> Is `subtarget` a revision of the JPIP spec? If not I think it should map onto the codestream parameter.

Sorry I'm weak English.
so I can't understand your opinion perfectly.
I presume your opinion is "subtarget is not jpip spec".
I found "subtarget" in "ISO-15444-9 C.2.3 Sub-target" 
 and "https://gwg.nga.mil/misb/docs/rp/RP0811.pdf"'s 6.2.1 Client Request Fields

I'm Sorry If I misunderstand your opinion.
or need more implementation give me more opinion.

ad960009:
@normanb and @rouault  Thanks for review I will remove this merge request.

@normanb I have a question with url handle
current code not handle url with parameter.
how about line 505~510 in this patch.

PR #4629, opened 2021-10-09
Initial implementation

mmomtchev:
## What does this PR do?

Implements a wrapper `GDALRingDataset` that returns `GDALRingRasterBand`s for the underlying dataset that support continuous reading from an endless band over the longitudes.

It has two functions:
Allows for transparent extraction of windows crossing the antimeridian

Allows for fully functional georeferencing on datasets using [0-360] longitudes

## Tasklist

 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments


mmomtchev:
This feature has an easier implementation on top of the VRT driver

PR #4653, opened 2021-10-14
GDALDataset::FlushCache() and GDALRasterBand::FlushCache(): add a bool bAtClosing argument (fixes #4652)

### LINKED ISSUE #4652: MarkSuppressOnClose() does not prevent FlushCache() on close
GDAL 3.3.2

ENVI dataset

I observed that even when a dataset is marked with MarkSuppressOnClose(), FlushCache() will still be called internally at closing.

The problem for me is that I was expecting to drop the dataset quickly (to start a new one), without having to wait for a long sync with disk, since the files will be deleted anyway.

So:
-is this expected behaviour ?
-is there a way to close a dataset and drop everything that was not flushed yet to quickly get rid of the dataset ?

rouault:
and set it to true when calling them from dataset/band destructors.
The base implementation use it as a hint to not flush dirty blocks
on datasets marked for deletion at closing.

### DIFF
--- a/gdal/frmts/mem/memdataset.cpp
+++ b/gdal/frmts/mem/memdataset.cpp
@@ -783,7 +783,6 @@ MEMDataset::MEMDataset() :
     adfGeoTransform[4] = 0.0;
     adfGeoTransform[5] = -1.0;
     DisableReadWriteMutex();
-    MarkSuppressOnClose();
 }
 
 /************************************************************************/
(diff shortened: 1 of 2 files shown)


chacha21:
Is it safe to call FlushCache(true) at any time (not only when closing a dataset) to "drop the cache", in the context of reusing a dataset without considering any previous data ?

rouault:
> Is it safe to call FlushCache(true) at any time (not only when closing a dataset) to "drop the cache", in the context of a temporary dataset (marked as suppress on close)

it should

Issue #4667, opened 2021-10-16
doc: Add more margin to the search input on the website


shuuji3:
## Expected behavior and actual behavior.

GDAL website https://gdal.org/ has a search input in the left sidebar. But I feel the search input needs a larger margin at the bottom of the input.

I will create a PR for this change.

### Current margin

<img width="517" alt="Screen Shot 2021-10-15 at 22 57 52" src="https://user-images.githubusercontent.com/1425259/137580572-2e89a968-a454-47ae-a080-cf90407b7ec2.png">

### Expected margin

<img width="480" alt="Screen Shot 2021-10-15 at 22 58 06" src="https://user-images.githubusercontent.com/1425259/137580576-b3155dbd-ee8a-4bdc-a88c-39330064c2db.png">

## Steps to reproduce the problem.

- Visit the website https://gdal.org/
- See the layout of the search input on the left sidebar

## Operating system

N/A

## GDAL version and provenance

N/A

PR #4668, opened 2021-10-16
doc: Modify bottom margins of the search input

shuuji3:
## What does this PR do?

This PR resolves the issue (#4667) about bottom margins of the search input.

There are three CSS properties in `gdal.css` related to the bottom margins which I removed in this commit:

```css
	margin-bottom: unset;
	padding-bottom: unset;
	background-color: #28728d;
```

`margin-bottom: unset;` and `padding-bottom: unset;` causes the issue. This removes the bottom margin which added by the default stylesheet.

`background-color` is not needed here because the exact same property and value is defined just after these lines.

### Screenshots

**Before**
Screen Shot 2021-10-15 at 22 57 52

**After**
Screen Shot 2021-10-15 at 22 58 06

doc: Add more margin to the search input on the website #4667

## Tasklist

 - [x] Fix the `gdal.css`
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS: macOS 11.6
* Compiler: N/A

### DIFF
--- a/gdal/doc/source/gdal_rtd/static/css/gdal.css
+++ b/gdal/doc/source/gdal_rtd/static/css/gdal.css
@@ -197,13 +197,6 @@ a.reference.external::after
 	/* use 100% width */
 }
 
-.wy-side-nav-search
-{
-	margin-bottom: unset;
-	padding-bottom: unset;
-	background-color: #28728d;
-}
-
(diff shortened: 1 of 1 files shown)


rouault:
thanks

PR #4685, opened 2021-10-20
configure.ac: use again AC_PROG_CC_C99 for autoconf <= 2.69 ...

rouault:
to avoid issues on older systems where the C compiler doesn't default to C99 or newer


tbonfort:
out of curiosity, which system(s) does this concern ?

rouault:
> which system(s) does this concern ?

not one I tested directly on, but from the logs I got, it looked like some CentOS/RHEL 7 one.

PR #4686, opened 2021-10-21
[filegdb] Port away from deprecated driver API

nyalldawson:


nyalldawson:
(unrelated test failure)

PR #4697, opened 2021-10-25
[odbc] Always request numeric column values as numeric types

### LINKED ISSUE #3885: OGR Access MDB Driver appears unable to parse certain numbers

nyalldawson:
The MS Access ODBC driver can return incorrect values for REAL
columns requested using char types.

Fixes #3885


nyalldawson:
@rouault I've reworked this now, so that the numeric column values are stored natively as a double, and so there's no change in behaviour for the existing GetColData methods. (I.e. ODBC drivers need to be explicitly adapted to retrieve field values as doubles)

nyalldawson:
@rouault I suspect these changes are bumping up into a bug in the SQL Server ODBC driver -- possibly this https://stackoverflow.com/questions/6242895/numeric-value-out-of-range-error-when-using-sqlfetch-with-sql-c-float .

I'm wondering what you think would be the best way to handle this. Options I can think of are:

- Having a setting on the CPLODBCStatement to set whether numeric values should always be retrieved as string, and setting this from the SQL Server driver
- Do the opposite, and make retrieval as numeric values opt-in. Use this for pgeo/odbc drivers but not sql server.
- Try to detect automatically in CPLODBCStatement if the problematic driver is in use. (But I can't think of a reliable way to do this, given that a valid DSN string might not include the driver name at all)

What's your thoughts?

rouault:
> What's your thoughts?

I don't have much to be honest. Your current approach looks good on the paper, but looking at test results (crash on Windows on ogr_pgeo.py, or test failure in MSSQLSpatial on Ubuntu 18.04),  it seems we hit bugs in ODBC drivers if we ask a column for both as string and double/float. So perhaps the "make retrieval as numeric values opt-in. Use this for pgeo/odbc drivers but not sql server" ?

nyalldawson:
@rouault I've reworked this so that the new behaviour is opt-in, and only enabled for MDB files and the Pgeo driver... let's see what the CI says!

ogr/ogrsf_frmts/odbc/ogrodbclayer.cpp
```diff
@@ -260,17 +260,36 @@ OGRFeature *OGRODBCLayer::GetNextRawFeature()
...
-        const char *pszValue = poStmt->GetColData( iSrcField );
-
-        if( pszValue == nullptr )
-            poFeature->SetFieldNull( iField );
-        else if( poFeature->GetFieldDefnRef(iField)->GetType() == OFTBinary )
-            poFeature->SetField( iField,
-                                 poStmt->GetColDataLength(iSrcField),
-                                 (GByte *) pszValue );
+
+        if ( eType == OFTReal && poStmt->Flags() & CPLODBCStatement::Flag::RetrieveNumericColumnsAsDouble )
```
rouault, line 266: just to avoid relying on operator precedence rules:
```suggestion
        if ( eType == OFTReal && (poStmt->Flags() & CPLODBCStatement::Flag::RetrieveNumericColumnsAsDouble) )
```

port/cpl_odbc.cpp
```diff
@@ -1183,9 +1200,47 @@ int CPLODBCStatement::Fetch( int nOrientation, int nOffset )
 
         char szWrkData[513] = {};
 
+        // If RetrieveNumericColumnsAsDouble flag is set, then read numeric columns using numeric data types and populate native double column values array.
+        // This allows retrieval of the original numeric value as a double via GetColDataAsDouble with risk of loss of precision.
```
rouault, line 1204: I assume this is:
```suggestion
        // This allows retrieval of the original numeric value as a double via GetColDataAsDouble without risk of loss of precision.
```

rouault:
rebasing on master will fix the failure in the build-mac cmake CI target (due to https://github.com/OSGeo/gdal/pull/5156 being applied)

Saijin-Naib:
You all are the best, seriously. Thank you!

PR #4760, opened 2021-11-03
CMake: MacOSX framework builds improvements

rouault:
- Fix INSTALL_NAME_DIR so that binaries find it
- Fixes to install resource and header files
- Fixes for setting the plugin directory in the driver manager
    
However plugin loading fails
(~/install-gdal/bin/gdalinfo --format JP2OpenJPEG --debug on) with:
```
ERROR 1: dlopen(/Users/even/install-gdal/Library/Frameworks/gdal.framework/Versions/3.4/PlugIns/gdal_JP2OpenJPEG.dylib, 1): Library not loaded: @rpath/gdal.framework/Versions/3.4/gdal
   Referenced from: /Users/even/install-gdal/Library/Frameworks/gdal.framework/Versions/3.4/PlugIns/gdal_JP2OpenJPEG.dylib
Reason: image not found
```


PeterPetrik:
RPATHS works a bit differently on macOS than in linux. I assume that `@rpath/gdal.framework/Versions/3.4/gdal` exists. 

As a workaround, you can set `DYLD_FRAMEWORK_PATH` or `DYLD_FALLBACK_FRAMEWORK_PATH` to `/Users/even/install-gdal/Library/Frameworks/` in env var, and it should work.

As a correct solution, each executable & each plugin should have inserted correct RPATH so you do not need to set env. variables. To check which RPATHS binary has:

`otool -l /Users/even/install-gdal/Library/Frameworks/gdal.framework/Versions/3.4/PlugIns/gdal_JP2OpenJPEG.dylib | grep RPATH`

for example in Qt distribution

```
otool -l /opt/Qt/5.15.2/clang_64/bin/qtplugininfo  | grep RPATH -A 2
          cmd LC_RPATH
      cmdsize 48
         path @executable_path/../Frameworks (offset 12)
```

to instert / modify rpath in the binary use command line tool `install_name_tool`

CMAKE does some of those automatically, it is always best to check what CMAKE did :)

or alternatively to rpath you can use `@ load path` or so https://wincent.com/wiki/%40executable_path%2C_%40load_path_and_%40rpath

klaussnd:
You might want to use relative RPATHs to make the package relocatable, you can use `@loader_path` for that, like
```
set(CMAKE_INSTALL_RPATH "@loader_path/../PlugIns")
```
to point to `../PlugIns` relative to the library.

PR #4768, opened 2021-11-04
GPKG: fix GDAL_DMD_HELPTOPIC

agiudiceandrea:
## What does this PR do?

Changes GDAL_DMD_HELPTOPIC value from non existent drivers/vector/geopackage.html to drivers/vector/gpkg.html

The GDAL_DMD_HELPTOPIC was changed from drv_geopackage.html to drivers/vector/geopackage.html with https://github.com/OSGeo/gdal/commit/542a11742b2ae809b6e2eed4ad35275976738bbb (https://github.com/OSGeo/gdal/pull/2992) since 3.2.0RC1.

A redirection from https://gdal.org/drivers/vector/geopackage.html to https://gdal.org/drivers/vector/gpkg.html should be set in order to workaround the current issue in QGIS.

https://github.com/qgis/QGIS/issues/45875


rouault:
> A redirection from https://gdal.org/drivers/vector/geopackage.html to https://gdal.org/drivers/vector/gpkg.html should be set in order to workaround the current issue in QGIS.

you can set it up in doc/source/_extensions/redirects.py

agiudiceandrea:
@rouault I hope this is the correct way to add the redirection...

agiudiceandrea:
@rouault it seems the redirection is incorrect!

PR #4774, opened 2021-11-05
Change GetFieldAsBinary to use VSIMalloc.

C41eb:
Running the tests with a custom allocator caused crashes as GetFieldAsBinary was using malloc for allocation but in the swig generated python/extensions/ogr_wrap.cpp file at line 15401 you can see that the memory is freed using VSIFree which does not match the allocation.

This change fixes the ogr.i file however I am not sure of the process to regenerate the *_wrap.cpp files using swig as using the makefile.vc in gdal/swig creates quite a different file from what is checked in. However I can add these to the pull request if required. 

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:


rouault:
I've added a commit with the regenerated files

PR #4778, opened 2021-11-05
CMake: add JXL support

rouault:


rouault:
@lucianpls I let you investigate brunsli cmake support for MRF ?

PR #4784, opened 2021-11-05
[WIP] geoloc arrays: removing artifacts (fixes #4707)

### LINKED ISSUE #4707: gdalwarp setting pixels outside target image bounds when using geolocation arrays

It seems that gdalwarp is setting pixel values outside the target image bounding when using geolocation arrays.

I've marked the source raster with bounding rectangles to check how these were being reprojected, upper left corner is for instance:

image

The target raster correctly computes the corners but seems to replicate the source's border pixel value outside the target image boundary. Figure below shows the target top left corner, a similar behavior happens for all corners:

image

Execute in [raw data](https://drive.google.com/file/d/1RU8vzhyNNFBPnWv3O3dUcFRWkEMfEtXe/view?usp=sharing) `issue` directory:

`gdalwarp -of GTiff -geoloc -s_srs EPSG:4326 -t_srs EPSG:3031 -dstnodata 0 -tr 3.5 3.5 -r near geoloc.vrt result.tif`

## Operating system

Tested against docker images `gdal:ubuntu-small-latest released 2021/10/07` and 
`gdal:ubuntu-small-3.3.2`.

See above.

fredliporace:
## What does this PR do?

Changes the geolocation array procedures to remove the artifacts described in #4707, more specifically:

* Change the logic used to compute the backmap resolution.
* Execute the backmap gap filling routines only if required.

**Work in progress, I do not consider the artifacts to be removed yet**: I created this as an opportunity to discuss the proposed solutions and also, since this is my first PR to the repo, get some feedback on its requirements.

Issue #4707

## Tasklist

 - [ ] Validate the expected behavior for the GapFilling algorithm used to fill backmap holes and fix it if required.
 - [ ] Check test cases against proposed changes

## Environment

This is probably not env specific.

* OS: CentOS8
* Compiler: g++ 9.3.0


alg/gdalgeoloc.cpp
```diff
@@ -231,13 +231,17 @@ static bool GeoLocGenerateBackMap( GDALGeoLocTransformInfo *psTransform )
...
+/*      X and Y resolutions are at least OVERSAMPLE_FACTOR higher than */
+/*      the source.                                                    */
+/* ------------------------------------------------------------------- */
+    // Higher bound for pixel size
+    const double nPixelSizeBound = std::min((psTransform->dfMaxX - psTransform->dfMinX) / nXSize,
+                                            (psTransform->dfMaxY - psTransform->dfMinY) / nYSize) / OVERSAMPLE_FACTOR;
+    const double dfTargetPixels =
+      (psTransform->dfMaxX - psTransform->dfMinX) *
+      (psTransform->dfMaxY - psTransform->dfMinY) / (nPixelSizeBound * nPixelSizeBound);
     const double dfPixelSize = sqrt(
```
fredliporace, line 245: My test case source resolution is (4096, 1000) and the backmap resolution computed by the original algorithm was (4843,846) - the number of pixels in the backmap increased but the resolution along Y axis decreased. The proposed change make sure that this does not happen - but potentially increases the backmap size.

alg/gdalgeoloc.cpp
```diff
@@ -445,102 +449,134 @@ static bool GeoLocGenerateBackMap( GDALGeoLocTransformInfo *psTransform )
         }
     }
 
-    // Fill holes in backmap
-    auto poMEMDS = std::unique_ptr<GDALDataset>(
-          MEMDataset::Create( "",
-                              static_cast<int>(nBMXSize),
-                              static_cast<int>(nBMYSize),
-                              0, GDT_Float32, nullptr ));
+    // Check if backmap has holes
```
fredliporace, line 452: I'll remove this logic from the PR after fixing the GapFilling procedure.

rouault:
@fredliporace  you might want to try if https://github.com/OSGeo/gdal/pull/5520 works for you

fredliporace:
I'm closing since #5520 solves the issue.

Issue #4301, opened 2021-08-23
`ogrinfo` mangles duplicate columns


lnicola:
## Expected behavior and actual behavior.

At least with the `PostgreSQL/PostGIS` driver, `ogr2ogr` and `ogrinfo` don't seem to play well with duplicate source column names.

## Steps to reproduce the problem.

First, create a PostGIS database and two simple tables:

```
# create table t1(id int not null primary key, geom geometry not null);
CREATE TABLE
# insert into t1 values (1, 'point (10 10)'), (2, 'point (20 20)'), (3, 'point (30 30)');
INSERT 0 3
# create table t2(id int not null references t1(id), attr int not null);
CREATE TABLE
# insert into t2 values (1, 100), (2, 200), (3, 300);
INSERT 0 3
```

Then try to query them using `ogrinfo`:

```
$ docker run --rm -v /etc/passwd:/etc/passwd -v /etc/group:/etc/group -v /var/run/postgresql:/var/run/postgresql -u $(id -u):$(id -g) osgeo/gdal:ubuntu-full-3.3.1 ogrinfo -sql 'select t1.*, t2.* from t1 inner join t2 using (id)' 'PG:dbname=mydb'
INFO: Open of `PG:dbname=mydb'
      using driver `PostgreSQL' successful.

Layer name: sql_statement
Geometry: Unknown (any)
Feature Count: 3
Extent: (10.000000, 10.000000) - (30.000000, 30.000000)
Layer SRS WKT:
(unknown)
Geometry Column NOT NULL = geom
id: Integer (0.0) NOT NULL
id: Integer (0.0)
attr: Integer (0.0) NOT NULL
OGRFeature(sql_statement):0
  id (Integer) = 1
  attr (Integer) = 100
  POINT (10 10)

OGRFeature(sql_statement):1
  id (Integer) = 2
  attr (Integer) = 200
  POINT (20 20)

OGRFeature(sql_statement):2
  id (Integer) = 3
  attr (Integer) = 300
  POINT (30 30)

```

Notice how the second `id` column shows up as nullable, even though it's not. But that doesn't matter, let's proceed and format the query more nicely (there's an extra newline at the start):

```
$ docker run --rm -v /etc/passwd:/etc/passwd -v /etc/group:/etc/group -v /var/run/postgresql:/var/run/postgresql -u $(id -u):$(id -g) osgeo/gdal:ubuntu-full-3.3.1 ogrinfo -sql '
[...]

PR #4787, opened 2021-11-07
PG: skip all leading whitespace in SQL statements.

direvus:
Previously the logic to skip leading whitespace only skipped over the
space character (0x20), but for SQL all leading whitespace characters
can be skipped.  Use the built-in std::isspace function to test for
this, rather than testing for the space character specifically.

The previous logic led to differences in the behaviour of queries,
because the logic to detect a 'SELECT' statement failed to detect valid
SELECT queries that began with whitespace other than 0x20.

Refs #4301

## What does this PR do?

Skip all leading whitespace, instead of just spaces, in SQL statements for the PostgreSQL driver.

Related to #4301.

## Tasklist

 - [x] Build and test locally
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS: MacOS 11.6
* Compiler: Xcode

### DIFF
--- a/ogr/ogrsf_frmts/pg/ogrpgdatasource.cpp
+++ b/ogr/ogrsf_frmts/pg/ogrpgdatasource.cpp
@@ -2925,8 +2925,8 @@ OGRLayer * OGRPGDataSource::ExecuteSQL( const char *pszSQLCommand,
                                         const char *pszDialect )
 
 {
-    /* Skip leading spaces */
-    while(*pszSQLCommand == ' ')
+    /* Skip leading whitespace characters */
+    while(std::isspace(static_cast<unsigned char>(*pszSQLCommand)))
         pszSQLCommand ++;
 
     FlushCache(false);
--- a/ogr/ogrsf_frmts/pg/ogrpgdatasource.cpp
+++ b/ogr/ogrsf_frmts/pg/ogrpgdatasource.cpp
@@ -32,6 +32,7 @@
 #include "cpl_conv.h"
 #include "cpl_string.h"
 #include "cpl_hash_set.h"
+#include <cctype>
 #include <set>
 
 #define PQexec this_is_an_error


lnicola:
Adding an `#include <cctype>` might fix the Windows build on AppVeyor.

rcoup:
Should we be skipping SQL comments too? 

```
-- my query
SELECT ...
```

direvus:
> 
> Should we be skipping SQL comments too?
> 

I don't dislike that idea, but it's certainly a lot more effort.  There's double-dash comments and also slash-star comments to worry about. I might have a tinker with it and see how I go.

jratike80:
Comments are skipped already if SQL statements are saved into file.

```
gdaltest.sql
============
-- my query
SELECT * from states limit 1

ogrinfo PG:"host=localhost port=5432 dbname=db user=user password=pw" -sql @gdaltest.sql

INFO: Open of ...
      using driver `PostgreSQL' successful.

Layer name: sql_statement
Geometry (wkb_geometry): Unknown (any)
Geometry (geom2): Unknown (any)
Feature Count: 1
```

There seems to be something to improve also in the @file syntax with the statement ending ";"

```
gdaltest.sql
============
-- my query
SELECT * from states limit 1;

ERROR 1: ERROR:  syntax error at or near ";"
LINE 1: SELECT count(*) FROM (SELECT * from states limit 1; ) AS ogr...

```

direvus:
> 
> There seems to be something to improve also in the @file syntax with the statement ending ";"
> 
> ```
> gdaltest.sql
> ============
> -- my query
> SELECT * from states limit 1;
> 
> ERROR 1: ERROR:  syntax error at or near ";"
> LINE 1: SELECT count(*) FROM (SELECT * from states limit 1; ) AS ogr...
> ```

@jratike80  that does look like a bug to me, but I don't think it's much related to this PR.  Can I recommend that you raise a fresh issue for that?  I'm interesting in trying to fix it, feel free to mention me on the issue.

direvus:
> 
> 
> Adding an `#include <cctype>` might fix the Windows build on AppVeyor.

Thanks for the tip, I've pushed a commit with that change.

PR #4797, opened 2021-11-08
Update FindECW.cmake

spezialspezial:
Allows setting ECW_ROOT as environment variable like it's already working with OpenEXR_ROOT, BLOSC_ROOT, BOOST_ROOT

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:

### DIFF
--- a/cmake/modules/packages/FindECW.cmake
+++ b/cmake/modules/packages/FindECW.cmake
@@ -16,6 +16,10 @@
 # Copyright (C) 2017,2018 Hiroshi Miura
 ################################################################################
 
+if(NOT ECW_ROOT AND DEFINED ENV{ECW_ROOT})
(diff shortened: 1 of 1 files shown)


rouault:
I'm not sure why support for env variables exists in other modules, but isn't CMake variable support already good enough ?

spezialspezial:
I guess it's historical because of different build systems other than cmake

spezialspezial:
Yes, would be enough, on the other hand it doesn't hurt and helps people coming over from configure.

spezialspezial:
Nice, thanks

PR #4809, opened 2021-11-10
gdalinfo --build: report PROJ version of build and runtime

rouault:


jratike80:
Great thing, many questions in gis.stackexhange about that. Latest ones often with a connection to proj.db because proj and proj.db versions do not match and table schemas in proj.db have changed.

PR #4812, opened 2021-11-10
Oracle: fix CMake support and fix compiler warnings

rouault:
CC @fechen123  . Regarding CMake, I *build* tested on Linux with OCI 12.1 and Windows with 21.3. It would hurn't if you could give that runtime testing too


fechen123:
Not sure what I need to do, use cmake to build gdal myself and test the GeoRaster driver?

rouault:
> use cmake to build gdal myself and test the GeoRaster driver?

yes that's the idea :-) Checkout that PR first :
```
git fetch origin pull/4812/head:pr_4812
git checkout pr_4812
```

fechen123:
OK, tomorrow is a US holiday. I will try the test if I have time this Friday.

fechen123:
My cmake couldn't find the PROJ library. What is the standard way to specify the path to these libraries in GDAL build? 

Could NOT find PROJ (missing: PROJ_LIBRARY PROJ_INCLUDE_DIR) (Required is
  at least version "6.0")

rouault:
There are CMake build instructions at https://gdal.org/build_hints.html#cmake-general-configure-options, and more specifically for PROJ: https://gdal.org/build_hints.html#proj
Normally ``cmake .. -DCMAKE_PREFIX_PATH=/path/to/proj-install-root`` should do it. You need to point to a PROJ >= 6 installation, like for the autoconf bulids.

fechen123:
Got the following warnings, not sure which one is ok and which one is not. I specified sqlite3 but it was not picked up. Also I thought geotiff is internal in GDAL, why does it still require libtiff?

-- The following RECOMMENDED packages have not been found:

 *TIFF (required version >= 4.0), support for the Tag Image File Format (TIFF)., <https://libtiff.gitlab.io/libtiff/>
   gdal_GTIFF: GeoTIFF image format
   gdal_INGR: Intergraph Raster Format
   gdal_CALS: CALS type 1 driver
 *SQLite3
   Enable SQLite3 support (used by SQLite/Spatialite, GPKG, Rasterlite, MBTiles, etc.)
 *GEOS
   Geometry Engine - Open Source (GDAL core dependency)

CMake Warning at gdal.cmake:630 (message):
  CMake builds are considered only EXPERIMENTAL for now.  Do not use them for
  production.
  CMakeLists.txt:270 (include)


-- Configuring done
CMake Warning at gdal.cmake:223 (add_library):
  Cannot generate a safe runtime search path for target GDAL because files in
  some directories may conflict with libraries in implicit directories:

    runtime library [libexpat.so.1] in /usr/lib64 may be hidden by files in:
      /ade/fechen_sdo_fc7/oracle/lib

  Some of these libraries may not be found correctly.
  CMakeLists.txt:270 (include)


-- Generating done
CMake Warning:
  Manually-specified variables were not used by the project:

    SQLITE3_INCLUDE_DIR
    SQLITE3_LIBRARY

rouault:
> I specified sqlite3 but it was not picked up

ah currently the case must be SQLite3_INCLUDE_DIR and SQLite3_LIBRARY (the doc is wrong. I'll fix it. or make accept upper case). But your SQLite build is in a system location it should be normally be picked up automatically.

> Also I thought geotiff is internal in GDAL, why does it still require libtiff?

geotiff and libtiff are 2 separate libs, and can be used as external or internal. Either way shouldn't matter for Oracle testing purposes

> Cannot generate a safe runtime search path for target GDAL because files in
> some directories may conflict with libraries in implicit directories:
> runtime library [libexpat.so.1] in /usr/lib64 may be hidden by files in:
>  /ade/fechen_sdo_fc7/oracle/lib

I've never seen that one. It seems there would be libexpat.so.1 in  /ade/fechen_sdo_fc7/oracle/lib.  You can probably? ignore that warning, but I've not yet experimented such situation with cmake

fechen123:
How to configure for C99 compiler? Thanks.

rouault:
> How to configure for C99 compiler? Thanks.

https://github.com/OSGeo/gdal/pull/4821 should hopefully fix it. You'll need to manually apply this to this branch

fechen123:
I have tested the GeoRaster driver. It is working fine. Thanks.

nyalldawson (approved):

PR #4813, opened 2021-11-11
CMake: fix detection of SQLite3 extra symbols when linking with a SQLite3 static lib (fixes #4811)

### LINKED ISSUE #4811: check_symbol_exists in FindSQLite3.cmake fails for static library
Hi, I'm trying to link in a static sqlite but the check in FindSQLite3.cmake

https://github.com/OSGeo/gdal/blob/eae5c2449d0691b4dc13e69adb9206effddadbd1/cmake/modules/packages/FindSQLite3.cmake#L61-L62

doesn't detect the symbols in the object's table.

-- Looking for sqlite3_column_table_name - not found
-- Looking for sqlite3_rtree_query_callback - not found

Right now I'm forcing this to pass using

-DSQLite3_HAS_COLUMN_METADATA=yes \
-DSQLite3_HAS_RTREE=yes \

but I'm not sure that is how it's supposed to work.

rouault:


rouault:
all failures related to download.osgeo.org not being responsive today

PR #4816, opened 2021-11-12
rdb: enable building rdb driver via CMake

tnixeu:
## What does this PR do?

Updates CMake files in order to enable building the rdb driver.

I did not add anything to `doc/source/build_hints.rst` because the only build requirement is, that the rdb can be found by CMake.
Instead I have updated the information in `doc/source/drivers/raster/rdb.rst`.

## Tasklist

 - [x] Update CMake files in order to build rdb driver
 - [x] Updated documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

I currently only compiled the update on Ubuntu focal. I could add downloading and building with the rdblib in the CI steps, because the rdblib can be downloaded from a server.


rouault:
Thanks. CMakeLists.txt files need to be linted with cmake-format. You can install it with "pip install cmake-format", and then run "./scripts/reformat.sh" (you'll need for example to add a newline or a whitespace so that the file appears to be modified for git)

Issue #4792, opened 2021-11-08
C# bindings apps broken in CI


rouault:
Since today both AppVeyor and github action workflow Windows are broken, due to a change external to GDAL (something with NuGet ??)
https://github.com/OSGeo/gdal/runs/4139684148?check_suite_focus=true
https://ci.appveyor.com/project/OSGeo/gdal/builds/41445624

The failure occurs with the GDALRead test application:
```
dotnet add GDALRead.csproj reference ..\gdal\gdal_csharp.csproj
Reference `..\gdal\gdal_csharp.csproj` added to the project.
	dotnet add GDALRead.csproj package System.Drawing.Common
  Determining projects to restore...
C:\Program Files\dotnet\sdk\5.0.402\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.EolTargetFrameworks.targets(28,5): warning NETSDK1138: The target framework 'netcoreapp2.1' is out of support and will not receive security updates in the future. Please refer to https://aka.ms/dotnet-core-support for more information about the support policy. [D:\a\gdal\gdal\swig\csharp\apps\GDALRead.csproj]
  Writing C:\Users\runneradmin\AppData\Local\Temp\tmp754A.tmp
info : Adding PackageReference for package 'System.Drawing.Common' into project 'GDALRead.csproj'.
info :   GET https://api.nuget.org/v3/registration5-gz-semver2/system.drawing.common/index.json
info :   OK https://api.nuget.org/v3/registration5-gz-semver2/system.drawing.common/index.json 248ms
info : Restoring packages for D:\a\gdal\gdal\swig\csharp\apps\GDALRead.csproj...
info :   GET https://api.nuget.org/v3-flatcontainer/system.drawing.common/index.json
info :   OK https://api.nuget.org/v3-flatcontainer/system.drawing.common/index.json 46ms
info :   GET https://api.nuget.org/v3-flatcontainer/system.drawing.common/6.0.0/system.drawing.common.6.0.0.nupkg
info :   OK https://api.nuget.org/v3-flatcontainer/system.drawing.common/6.0.0/system.drawing.common.6.0.0.nupkg 12ms
info : Installed System.Drawing.Common 6.0.0 from https://api.nuget.org/v3/index.json with content hash NfuoKUiP2nUWwKZN6twGqXioIe1zVD0RIj2t976A+czLHr2nY454RwwXs6JU9Htc6mwqL6Dn/nEL3dpVf2jOhg==.
[...]

PR #4828, opened 2021-11-14
Re-target C# bindings apps to netcoreapp3.1, fix broken build (#4792)

szekerest:


rouault:
Could you also revert 90dab96f23b67f84c129467baec3786f48e9ca96 ?

rouault:
ok, so on appveyor (Visual Studio 2015 build image), we get this error
```
C:\Program Files\dotnet\sdk\2.2.402\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.TargetFrameworkInference.targets(137,5): error NETSDK1045: The current .NET SDK does not support targeting .NET Core 3.1.
```
I'm not sure if it is possible to upgrade it ? Otherwise we may still disable C# bindings building on that CI configuration, and just rely on the GitHub action one

szekerest:
@rouault According to https://www.appveyor.com/docs/windows-images-software/ netcoreapp3.x is only supported with Visual Studio 2019 or above, so I've set it to netcoreapp2.2 (the latest supported version for VS2015+). Also modified the system.drawing.common package version to v 5.0.3 because the 6.0.0 version is only supported with netcoreapp3.x

PR #4835, opened 2021-11-15
Add support for comments in ERS files

mplough-kobold:
## What does this PR do?

This PR adds support for comments in ERS file headers.  A comment starts with a `#` character and continues to the end of the line.  ER Mapper documentation describes comments in this fashion and we've recently encountered files in the wild that contain comments.  Since comments may appear at the start of the file, ERS file identification logic is modified to accommodate presence of such comments.

The `srtm.ers` test case is modified to exercise comment logic and all ERS test cases are still passing.

### DIFF
--- a/frmts/ers/ersdataset.cpp
+++ b/frmts/ers/ersdataset.cpp
@@ -828,11 +828,18 @@ int ERSDataset::Identify( GDALOpenInfo * poOpenInfo )
 /* -------------------------------------------------------------------- */
 /*      We assume the user selects the .ers file.                       */
 /* -------------------------------------------------------------------- */
-    if( poOpenInfo->nHeaderBytes < 15
-        || !STARTS_WITH_CI((const char *) poOpenInfo->pabyHeader, "DatasetHeader ") )
-        return FALSE;
+    if( poOpenInfo->nHeaderBytes >= 15
+        && STARTS_WITH_CI((const char *) poOpenInfo->pabyHeader, "DatasetHeader ") )
+        return TRUE;
 
-    return TRUE;
+/* -------------------------------------------------------------------- */
+/*      We assume the user selects the .ers file.                       */
+/* -------------------------------------------------------------------- */
+    if( poOpenInfo->nHeaderBytes >= 1
+        && STARTS_WITH_CI((const char *) poOpenInfo->pabyHeader, "#") )
+        return TRUE;
+
+    return FALSE;
 }
 
 /************************************************************************/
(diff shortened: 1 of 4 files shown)


frmts/ers/ersdataset.cpp
```diff
@@ -828,11 +828,18 @@ int ERSDataset::Identify( GDALOpenInfo * poOpenInfo )
...
+    if( poOpenInfo->nHeaderBytes >= 15
+        && STARTS_WITH_CI((const char *) poOpenInfo->pabyHeader, "DatasetHeader ") )
+        return TRUE;
+
+/* -------------------------------------------------------------------- */
+/*      We assume the user selects the .ers file.                       */
+/* -------------------------------------------------------------------- */
+    if( poOpenInfo->nHeaderBytes >= 1
+        && STARTS_WITH_CI((const char *) poOpenInfo->pabyHeader, "#") )
+        return TRUE;
```
rouault, line 840: We cannot just identify a ERS dataset if its first character is #. This is not specific enough . I'd suggest just to look for the presence of "DatasetHeader " anywhere in the starting bytes of the file:

```suggestion
    if( poOpenInfo->nHeaderBytes >= 15
        && strstr((const char *) poOpenInfo->pabyHeader, "DatasetHeader ") != nullptr )
        return TRUE;
```

mplough-kobold: Stand by for a tweak to deal with the `data/ers/caseinsensitive.ers` file...

mplough-kobold: When I re-ran the tests before, I re-ran them without _re-compiling_ -- have been dealing with too much Python lately.  Sorry about that.  

I modified the file type detection to use `CPLString`.  Since we need to a case-insensitive search of the header for the `DatasetHeader ` string, neither the `STARTS_WITH_CI` macro nor `strstr` are powerful enough.  Instantiating a `CPLString` allows us to easily use the `ifind` method, which is powerful enough for this search.

frmts/ers/ersdataset.cpp
```diff
@@ -816,23 +816,20 @@ int ERSDataset::Identify( GDALOpenInfo * poOpenInfo )
 /* -------------------------------------------------------------------- */
 /*      We assume the user selects the .ers file.                       */
 /* -------------------------------------------------------------------- */
-    if( poOpenInfo->nHeaderBytes > 15
-        && STARTS_WITH_CI((const char *) poOpenInfo->pabyHeader, "Algorithm Begin") )
+    CPLString osHeader((const char *)poOpenInfo->pabyHeader, poOpenInfo->nHeaderBytes);
+
+    if( osHeader.ifind( "Algorithm " ) != std::string::npos )
```
rouault, line 821: why removing the " Begin" ?
The check could potentially now trigger more easily on unrelated datasets. I'm not familiar with that format, but if those algorithm ERS files have also a "DatasetHeader " string in them, perhaps this check could be moved in the below check ? Otherwise we may want to add additional conditions

mplough-kobold: Agreed, the ` Begin` should stay and I've put it back in here.  This will preserve the existing behavior as much as possible, while still allowing comments to occur even at the start of algorithm files.

PR #4839, opened 2021-11-16
Create FindOpenEXR.cmake

spezialspezial:
## What does this PR do?

It is now possible to just specify OpenEXR_ROOT for Versions < 3. Versions >= 3 need an additional Imath_ROOT as Imath is separate now. This should work without an explicitly user supplied PKG_CONFIG_PATH. Tested on openexr-2.5 and openexr-3.1.

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

CMake 3.16, Ninja, openexr-2.5 and openexr-3.1

* OS: Linux
* Compiler: GCC 9.3

### DIFF
--- a/cmake/modules/packages/FindOpenEXR.cmake
+++ b/cmake/modules/packages/FindOpenEXR.cmake
@@ -14,47 +14,71 @@
 #    OpenEXR_INCLUDE_DIRS
 #    OpenEXR_LIBRARIES
 #
+
+if(OpenEXR_ROOT)
+  list(APPEND OpenEXR_INC_HINTS ${OpenEXR_ROOT}/include/OpenEXR)
+  list(APPEND OpenEXR_LIB_HINTS ${OpenEXR_ROOT}/lib)
+  list(APPEND CMAKE_PREFIX_PATH ${OpenEXR_ROOT})
+endif()
+
+if(Imath_ROOT)
+  list(APPEND Imath_INC_HINTS ${Imath_ROOT}/include/Imath)
+  list(APPEND Imath_LIB_HINTS ${Imath_ROOT}/lib)
+  list(APPEND CMAKE_PREFIX_PATH ${Imath_ROOT})
+endif()
+
 find_package(PkgConfig QUIET)
 if (PKG_CONFIG_FOUND)
-    pkg_check_modules(PC_OpenEXR QUIET OpenEXR)
+  pkg_check_modules(PC_OpenEXR QUIET OpenEXR)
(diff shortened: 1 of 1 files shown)


rouault:
Thanks. Would you mind documenting the user variables in doc/source/build_hints.rst, in particular the distinct OpenEXR_ROOT and Imath_ROOT, in a new OpenEXR subsection similar to the existing ones for other libraries ?

spezialspezial:
Ok, I'll check out the doc section.

spezialspezial:
One more thing. Behavior should be exactly the same as before if OpenEXR_ROOT is not defined.

spezialspezial:
Autsch, some builds fail. I think the escape for the inner quotes in string(REGEX MATCH "define[ \t]+OPENEXR_VERSION_STRING[ \t]+\"([0-9]+(.[0-9]+)?(.[0-9]+)?)\".*$" _ ${txt}) is platform specific. I'll look into it.

spezialspezial:
That also means it would try to build an external EXR 3 with EXR 2 linkage in the original FindOpenEXR.cmake as the OpenEXR_VERSION_STRING is empty with no PkgConfig.

spezialspezial:
Hmm, can't find an OpenEXR install for the failing build workflows. Even if this is the problem I would still try to fail more gracefully.

cmake/modules/packages/FindOpenEXR.cmake
```diff
@@ -14,47 +14,71 @@
...
-          HINTS ${PC_OpenEXR_INCLUDE_DIRS}
+          HINTS ${OpenEXR_INC_HINTS}
           PATH_SUFFIXES OpenEXR)
 find_path(Imath_INCLUDE_DIR
           NAMES ImathMatrix.h
-          HINTS ${PC_OpenEXR_INCLUDE_DIRS}
+          HINTS ${Imath_INC_HINTS} ${OpenEXR_INCLUDE_DIR}
           PATH_SUFFIXES Imath)
 
+if(NOT OpenEXR_VERSION_STRING)
```
rouault, line 49: Should avoid the failures on builds that don't have OpenEXR available
```suggestion
if(OpenEXR_INCLUDE_DIR AND NOT OpenEXR_VERSION_STRING)
```

rouault:
superseded per https://github.com/OSGeo/gdal/pull/4840

PR #4840, opened 2021-11-17
Adding documentation regarding build hints for external library OpenEXR

spezialspezial:
If this works out we could consider making it a cache entry.

### DIFF
--- a/cmake/modules/packages/FindOpenEXR.cmake
+++ b/cmake/modules/packages/FindOpenEXR.cmake
@@ -14,47 +14,71 @@
 #    OpenEXR_INCLUDE_DIRS
 #    OpenEXR_LIBRARIES
 #
+
+if(OpenEXR_ROOT)
+  list(APPEND OpenEXR_INC_HINTS ${OpenEXR_ROOT}/include/OpenEXR)
+  list(APPEND OpenEXR_LIB_HINTS ${OpenEXR_ROOT}/lib)
+  list(APPEND CMAKE_PREFIX_PATH ${OpenEXR_ROOT})
+endif()
+
+if(Imath_ROOT)
+  list(APPEND Imath_INC_HINTS ${Imath_ROOT}/include/Imath)
+  list(APPEND Imath_LIB_HINTS ${Imath_ROOT}/lib)
+  list(APPEND CMAKE_PREFIX_PATH ${Imath_ROOT})
+endif()
+
 find_package(PkgConfig QUIET)
 if (PKG_CONFIG_FOUND)
-    pkg_check_modules(PC_OpenEXR QUIET OpenEXR)
+  pkg_check_modules(PC_OpenEXR QUIET OpenEXR)
+  if(PC_OpenEXR_FOUND)
+    list(APPEND OpenEXR_LIB_HINTS ${PC_OpenEXR_LIBRARY_DIRS})
+    list(APPEND OpenEXR_INC_HINTS ${PC_OpenEXR_INCLUDE_DIRS})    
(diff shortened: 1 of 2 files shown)


doc/source/build_hints.rst
```diff
@@ -315,12 +315,35 @@ TIFF
...
+
+or
+
+	Get real specific and set 
+	``OpenEXR_INCLUDE_DIR``, ``Imath_INCLUDE_DIR``, 
+	``OpenEXR_LIBRARY``, ``OpenEXR_UTIL_LIBRARY``, 
+	``OpenEXR_HALF_LIBRARY``, ``OpenEXR_IEX_LIBRARY`` 
+	explicitly
+
+
```
rouault, line 346:
```suggestion
OpenEXR
*******

Specify ``OpenEXR_ROOT`` variable pointing to the parent directory of 
/lib and /include subdirectories, i.e. /DEV/lib/openexr-3.0.
For OpenEXR >= 3 additionally specify ``Imath_ROOT`` as this is a 
separate library now, i.e. /DEV/lib/imath-3.1.3

or

Specify root directory adding to the ``CMAKE_PREFIX_PATH`` variable to find OpenEXR's pkgconfig. 
For example -DCMAKE_PREFIX_PATH=/DEV/lib/openexr-3.0;/DEV/lib/imath-3.1.3

or

Get real specific and set 
``OpenEXR_INCLUDE_DIR``, ``Imath_INCLUDE_DIR``, 
``OpenEXR_LIBRARY``, ``OpenEXR_UTIL_LIBRARY``, 
``OpenEXR_HALF_LIBRARY``, ``OpenEXR_IEX_LIBRARY`` 
explicitly


```

cmake/modules/packages/FindOpenEXR.cmake
```diff
@@ -14,47 +14,71 @@
...
-          HINTS ${PC_OpenEXR_INCLUDE_DIRS}
+          HINTS ${OpenEXR_INC_HINTS}
           PATH_SUFFIXES OpenEXR)
 find_path(Imath_INCLUDE_DIR
           NAMES ImathMatrix.h
-          HINTS ${PC_OpenEXR_INCLUDE_DIRS}
+          HINTS ${Imath_INC_HINTS} ${OpenEXR_INCLUDE_DIR}
           PATH_SUFFIXES Imath)
 
+if(NOT OpenEXR_VERSION_STRING)
```
rouault, line 49:
```suggestion
if(OpenEXR_INCLUDE_DIR AND NOT OpenEXR_VERSION_STRING)
```

rouault:
> If this works out we could consider making it a cache entry.

OpenEXR_ROOT and Imath_ROOT ? For OpenEXR_ROOT, this should be automatically done by cmake >= 3.13 (so you might want to conditionalize it similarly as FindSPATIALITE.cmake). For Imath_ROOT, yes, we'd need a manual statement.
Feel free to add a commit to this PR

rouault:
Actually, I'm merging this PR as it. You can open a new follow-up PR with any further enhancements

spezialspezial:
> For OpenEXR_ROOT, this should be automatically done by cmake >= 3.13 

Possibly my fault but could not get this to work. Maybe because of the non standard includes location `openexr/include/OpenEXR`?

rouault:
> Possibly my fault but could not get this to work

I just meant that the declaration of the cmake variable OpenEXR_ROOT should be automatic as the module is called FindOpenEXR. Its value will need to be setup by the user to an appropriate value if it is not a system location

PR #4851, opened 2021-11-18
CMake: iconv on OSX

hobu:
Check for iconv_open availability in libiconv on OSX due to how things could be built on that platform in regard to `LIBICONV_PLUG`

### DIFF
--- a/port/CMakeLists.txt
+++ b/port/CMakeLists.txt
@@ -178,16 +178,15 @@ if (GDAL_USE_ICONV)
   if (Iconv_IS_BUILT_IN)
     target_sources(cpl PRIVATE cpl_recode_iconv.cpp)
   else ()
-      if (APPLE)
-          # linkage against OSX's system iconv is going to require
-          # LIBICONV_PLUG to be defined, but other self-built situations
-          # like CondaForge or Homebrew might not need this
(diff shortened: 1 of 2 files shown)


port/CMakeLists.txt
```diff
@@ -178,9 +178,15 @@ if (GDAL_USE_ICONV)
...
     target_sources(cpl PRIVATE cpl_recode_iconv.cpp)
   else ()
-    if (APPLE)
-      target_compile_definitions(cpl PRIVATE -DLIBICONV_PLUG)
-    endif ()
+      if (APPLE)
+          # linkage against OSX's system iconv is going to require
+          # LIBICONV_PLUG to be defined, but other self-built situations
+          # like CondaForge or Homebrew might not need this
+          check_symbol_exists(_iconv_close "iconv.h" HAVE_ICONV_CLOSE)
```
rouault, line 185: how does it know in which library to look for the symbol ? I would expect ``set(CMAKE_REQUIRED_LIBRARIES ${Iconv_LIBRARY})`` to be set before

rouault:
./scripts/reformat.sh needs to be run to lint CMakeLists.txt files (just add some newline at the end in tport/CMakeLists.txt, so that git sees a change in it)

Issue #4762, opened 2021-11-04
Suboptimal cache flush policy when processing datasets


chacha21:
I do not understand some of the design choices of  GDALRasterBlock::Internalize()

Consider the following scenario  :
I create dataset1 and dataset2
I fill dataset1
I read dataset1 and write it to dataset2 while performing some processing of the data in the temporary buffer

Please note that the processing is local spatially but requires all the bands for each pixel; that's why the dataset will be browsed line by line to handle all bands of each pixel on a row

To better see the problem, the cache limit is set slight above one dataset total size

What happens is that the current cache policy will spend a lot of time flushing/recalling the blocks of dataset2 instead of making room by dropping cached blocks of dataset1, resulting in very inefficient I/O.
(I found the comment `//In this first pass, only discard dirty blocks of this dataset...` explaining why, but it just does not fit my usage)
Instead of dropping "oldest blocks", the choice is made to drop blocks of the current dataset.

Did I miss something or is this a real problem ?

```
  GDALDriver* enviDriver = !gdalDriverManager ? 0 : gdalDriverManager->GetDriverByName("ENVI");

  const int spatialWidth = 1024;
  const int spatialHeight = 1024;
  const int bandsCount = 256;
  const GDALDataType dataType = GDT_UInt16;
  typedef unsigned __int16 pixel_t;
  std::vector<pixel_t> buffer(spatialWidth*bandsCount);

  GDALSetCacheMax64(spatialWidth*spatialHeight*bandsCount*sizeof(pixel_t)+10*bandsCount*sizeof(pixel_t));

  GDALDataset* srcDataset = (GDALDataset*)GDALCreate(enviDriver, "src.bil", spatialWidth, spatialHeight, bandsCount, dataType, 0);
  srcDataset->MarkSuppressOnClose();

  GDALDataset* dstDataset = (GDALDataset*)GDALCreate(enviDriver, "dst.bil", spatialWidth, spatialHeight, bandsCount, dataType, 0);
  dstDataset->MarkSuppressOnClose();

  //initialize srcDataset
  for(int row = 0 ; row<spatialHeight ; ++row)
  {
    std::iota(buffer.begin(), buffer.end(), static_cast<pixel_t>(row*bandsCount));
    srcDataset->RasterIO(GDALRWFlag::GF_Write, 0, row, spatialWidth, 1, 
      buffer.data(), spatialWidth, 1, dataType,
      bandsCount, 0,
      bandsCount*sizeof(pixel_t),
[...]

PR #4852, opened 2021-11-18
added GDALCacheFlushHelperHandler

chacha21:
By default, when a dataset being processed needs to allocate a new cache block
and the global cache is already full, it is not allowed to flush the cache of
other datasets to make room in the global cache, in order to prevent problems
in multithreading scenarios.
The proposed callback of this PR asks the user whether it is safe to
flush some dataset or not, in order to give a chance to do so and free some
cache blocks. By doing so, the pressure on the cache might be released and avoid
I/O congestion. It is up to the user to ensure that by answering "yes" in a
multithreading scenario, flushing another dataset will not induce unexpected results.
This callback cannot be installed on different thread local storages, since its
purpose is to get a global answer regarding the user's known current thread states


rouault:
As far as I'm concerned, I don't think this is appropriate for upstream inclusion as discussed in #4762. This is a workaround, hard to use safely and that will cludge the API

chacha21:
I am ok with your position.
I am really using that feature, though, that's why I am keeping the branch in touch with origin/master
Do you want me to delete the pull request so that it won't raise more notifications ?

rouault:
I'm closing the PR indeed

Issue #4817, opened 2021-11-12
Add libtirpc to new cmake system - DODS and OGDI drivers fail to build on Ubuntu21.10


andrew-aitchison:
This is a similar issue to
https://github.com/OSGeo/gdal/pull/4394#issuecomment-908892817

## Expected behavior and actual behavior.
Build from git head using cmake.
Expected all drivers to build.
Actual: OGDI and DODS (raster and vector) drivers fail to build because rpc/types.h is not found.

[ 45%] Building CXX object frmts/dods/CMakeFiles/gdal_DODS.dir/dodsdataset2.cpp.o
In file included from /usr/include/libdap/XDRUtils.h:37,
                 from /usr/include/libdap/Sequence.h:50,
                 from /home/werdna/git/gdal/frmts/dods/libdap_headers.h:50,
                 from /home/werdna/git/gdal/frmts/dods/dodsdataset2.cpp:38:
/usr/include/libdap/xdr-datatypes.h:16:10: fatal error: rpc/types.h: No such file or directory
   16 | #include <rpc/types.h>
      |          ^~~~~~~~~~~~~
compilation terminated.

[ 71%] Building CXX object ogr/ogrsf_frmts/dods/CMakeFiles/ogr_DODS.dir/ogrdodsdatasource.cpp.o
In file included from /usr/include/libdap/XDRUtils.h:37,
                 from /usr/include/libdap/Sequence.h:50,
                 from /home/werdna/git/gdal/ogr/ogrsf_frmts/dods/libdap_headers.h:52,
                 from /home/werdna/git/gdal/ogr/ogrsf_frmts/dods/ogr_dods.h:44,
                 from /home/werdna/git/gdal/ogr/ogrsf_frmts/dods/ogrdodsdatasource.cpp:30:
/usr/include/libdap/xdr-datatypes.h:16:10: fatal error: rpc/types.h: No such file or directory
   16 | #include <rpc/types.h>
      |          ^~~~~~~~~~~~~
compilation terminated.

[ 82%] Building CXX object ogr/ogrsf_frmts/ogdi/CMakeFiles/ogr_OGDI.dir/ogrogdidatasource.cpp.o
In file included from /home/werdna/git/gdal/ogr/ogrsf_frmts/ogdi/ogrogdi.h:40,
                 from /home/werdna/git/gdal/ogr/ogrsf_frmts/ogdi/ogrogdidatasource.cpp:30:
[...]

PR #4861, opened 2021-11-20
Fixes cmake build of OGDI driver - issue #4817

andrew-aitchison:
Fixes cmake build of OGDI driver - issue #4817
We now use pkgconfig to find tirpc headers.

This does not fix the DODS driver since from issue #4817:
DODS support will be dropped in GDAL 3.5, so you can leave that out and build with cmake .. -DGDAL_USE_DAP=OFF to disable it.

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS: Ubuntu 21.10
* Compiler: g++ v11.2


cmake/modules/packages/FindOGDI.cmake
```diff
@@ -12,28 +12,50 @@
...
-find_path(OGDI_INCLUDE_DIR ecs.h PATH_SUFFIXES ogdi)
+find_package(PkgConfig QUIET)
+if(PKG_CONFIG_FOUND)
+  pkg_check_modules(PC_OGDI QUIET OGDI)
+  set(OGDI_VERSION_STRING ${PC_OGDI_VERSION})
+  set(OGDI_INCLUDE_DIRS ${PC_OGDI_INCLUDE_DIRS})
+endif()
+
+find_path(OGDI_INCLUDE_DIR
+          NAMES ecs.h rpc/rpc.h
```
rouault, line 23: do we need to look for rpc/rpc.h ? This is an implementation detail. GDAL just includes ecs.h

cmake/modules/packages/FindOGDI.cmake
```diff
@@ -12,28 +12,50 @@
...
+          HINTS ${PC_OGDI_INCLUDE_DIRS})
```
rouault, line 24: shouldn't the PATH_SUFFIXES ogdi should be kept (for non-pkgconfig cases)

cmake/modules/packages/FindOGDI.cmake
andrew-aitchison: Possibly. The original error was

[100%] Building CXX object ogr/ogrsf_frmts/ogdi/CMakeFiles/ogr_OGDI.dir/ogrogdidatasource.cpp.o
In file included from /home/werdna/git/gdal/ogr/ogrsf_frmts/ogdi/ogrogdidatasource.cpp:30:
In file included from /home/werdna/git/gdal/ogr/ogrsf_frmts/ogdi/ogrogdi.h:40:
/usr/include/ogdi/ecs.h:9:10: fatal error: 'rpc/rpc.h' file not found
#include <rpc/rpc.h>
         ^~~~~~~~~~~
1 error generated.

The actual change in Ubuntu is that there are now two transport-independent RPC libraries  tirpc and ntirpc, with headers moved from /usr/include/rpc/ to /usr/include/tirpc/rpc/ and /usr/include/ntirpc/rpc/
It is not clear to me that without pkgconfig, cmake would find rpc/rpc.h

How do I generate a non-pkgconfig, cmake build ?

cmake/modules/packages/FindOGDI.cmake
andrew-aitchison: Sounds reasonable. I'll test it when I test the non-pkgconfig, cmake build.

cmake/modules/packages/FindOGDI.cmake
rouault: > It is not clear to me that without pkgconfig, cmake would find rpc/rpc.h

indeed, but I suspect that OGDI_INCLUDE_DIR only includes the directory of the first file found, here ecs.h ? (genuinely asking, as I don't have the target env. immediately available)

> How do I generate a non-pkgconfig, cmake build ?

I would just rename /usr/share/pkgconfig/ogdi.pc to ogdi.pc.disabled. But is there a ogdi.pc file at all ??

rouault:
why closing it ?

andrew-aitchison:
Sorry.
Trying to resync my local and github repos.
New pull request coming soon.

rouault:
> Trying to resync my local and github repos. New pull request coming soon.

ok, no problem. I'd just note that  git rebase + git push -f   or git merge can be used to update an existing pull request too.

PR #4866, opened 2021-11-21
Update BAG 2.0 Extract URLs

sebastic:
Kadaster is not good at preserving links when changing their website, links to their site need to be updated periodically.

Using archive.org might be a good idea for stable links.


rouault:
thanks

PR #4871, opened 2021-11-21
CMake: avoid warning messages for rdb and TileDB

rouault:
Fixes warnings like
```
CMake Warning at cmake/helpers/CheckDependentLibraries.cmake:35 (find_package):
  By not providing "Findrdb.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "rdb", but
  CMake did not find one.

  Could not find a package configuration file provided by "rdb" with any of
  the following names:

    rdbConfig.cmake
    rdb-config.cmake

  Add the installation prefix of "rdb" to CMAKE_PREFIX_PATH or set "rdb_DIR"
  to a directory containing one of the above files.  If "rdb" provides a
  separate development package or SDK, be sure it has been installed.
```

@tnixeu Could you check that this pull request still work with the rdb driver ? On Linux, I got the above warning due to different of case FindRDB.cmake vs 'rdb' in find_package() call. Renaming the file to Findrdb.cmake seemed to solve it, but there's no need to create a wrapper file if calling ``find_package(rdb CONFIG)``

### DIFF
--- a/cmake/helpers/CheckDependentLibraries.cmake
+++ b/cmake/helpers/CheckDependentLibraries.cmake
@@ -22,14 +22,16 @@ include(CheckSymbolExists)
 # GDAL_USE_ is OFF.
 # Accept a RECOMMENDED option
 macro(gdal_check_package name purpose)
-    set(_options CAN_DISABLE RECOMMENDED DISABLED_BY_DEFAULT)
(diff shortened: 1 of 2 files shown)


tnixeu:
I have tried it and it the rdb driver works with the changes of this PR

Issue #4911, opened 2021-11-28
Doc : misleading GDALDriver::CreateCopy() documentation


chacha21:
Let's try to `GDALDriver::CreateCopy()` from an ENVI dataset created with LINE interleaving (thanks to `INTERLEAVE=BIL` creation option)
`dst_ds = enviDriver->CreateCopy(<some dst path>, src_ds_envi_bil, TRUE, 0, 0, 0);`

The dst_ds will be an ENVI with BSQ (BAND) interleaving (the default for this driver)
So it seems that the same creation options must be used to get the same data layout.

I found the doc misleading because :

- it does not mention the importance of the creation options
- it let me believe that bStrict=TRUE would keep the interleaving

rouault:
Perhaps you could suggest a pull request with proposed doc improvements ?

chacha21:
:+ I am always afraid to pull something wrong because of a misunderstanding, but I can do this one

rouault:
You could also have a CreateCopy() implementation in the ENVI driver that would fetch the interleaving from the source dataset and pass it to Create(), and then fallback to DefaultCreateCopy().  EHdrDataset::CreateCopy() could be inspirational for that. Actually it might be possible to have a general logic in DefaultCreateCopy(), that would look if a INTERLEAVE=BIL/BIP/BSQ creation option exists in the output driver, and if a INTERLEAVE metadata item exists in the source dataset

chacha21:
> that would look if a INTERLEAVE=BIL/BIP/BSQ creation option exists in the output driver, and if a INTERLEAVE metadata item exists in the source dataset
Since the syntax for interleaving might differ between drivers (MEM:PIXEL/BAND and ENVI:BIP/BIL/BSQ, for instance) it seems difficult to find a robust generic handling

rouault:
> Since the syntax for `INTERLEAVE=...` might differ between drivers (MEM:PIXEL/BAND and ENVI:BIP/BIL/BSQ, for instance) it seems difficult to find a robust generic handling

that must be about the 2 variations possible. By looking at the possible values in the GDAL_DMD_CREATIONOPTIONLIST, one can figure it out

chacha21:
I am currently investigating, and I can't find a way to get the interleave information from a MEM dataset. There is no associated metadata in my different tests.
[...]

PR #4913, opened 2021-11-29
Smarter GDALDriver::CreateCopy()

chacha21:
Aims #4911

Better doc to explain how interleaving is handled by CreateCopy()

Added a system that tries to automatically match the src data interleaving by default, if not specified otherwise.

That might break backward compatibility for people that were relying on the default driver data interleaving when using CreateCopy() with no creation options.

~~The system is "fragile" regarding the different syntaxes for INTERLEAVE (BAND/PIXEL/LINE or BIP/BI/BSQ) and is currently hard coding BIP/BIL/BSQ for ENVI drivers only.~~
UPDATE : an XML parsing is performed to explore the declared INTERLEAVE syntaxes in the current driver creation options

### DIFF
--- a/gcore/gdaldriver.cpp
+++ b/gcore/gdaldriver.cpp
@@ -1011,7 +1011,7 @@ GDALDataset *GDALDriver::CreateCopy( const char * pszFilename,
             (CSLFindString(interleavesCSL, "PIXEL") >= 0) ? "PIXEL" :
             (CSLFindString(interleavesCSL, "BIP") >= 0) ? "BIP" :
             nullptr;
-        const char* dstInterleave = !srcInterleave ? 0 :
+        const char* dstInterleave = !srcInterleave ? nullptr :
             EQUAL(srcInterleave, "BAND") ? dstInterleaveBand :
             EQUAL(srcInterleave, "LINE") ? dstInterleaveLine :
             EQUAL(srcInterleave, "PIXEL") ? dstInterleavePixel :
--- a/gcore/gdaldriver.cpp
+++ b/gcore/gdaldriver.cpp
@@ -964,7 +964,7 @@ GDALDataset *GDALDriver::CreateCopy( const char * pszFilename,
 /*      If no INTERLEAVE creation option is given, we will try to add   */
 /*      one that matches the current srcDS interleaving                 */
 /* -------------------------------------------------------------------- */
-    char** papszOptionsWithInterleave = 0;
+    char** papszOptionsWithInterleave = nullptr;
     const bool bHasInterleaveCreationOptions =
          (CSLFetchNameValue(papszOptions, "INTERLEAVE") != nullptr);
     if (!bHasInterleaveCreationOptions && poSrcDS)
@@ -972,30 +972,25 @@ GDALDataset *GDALDriver::CreateCopy( const char * pszFilename,
         const char* srcInterleave = poSrcDS->GetMetadataItem("INTERLEAVE", "IMAGE_STRUCTURE");
 
         //look for INTERLEAVE values of the driver
-        char** interleavesCSL = 0;
+        char** interleavesCSL = nullptr;
         const char *pszOptionList = this->GetMetadataItem( GDAL_DMD_CREATIONOPTIONLIST );
-        CPLXMLNode* xmlNode = CPLParseXMLString(pszOptionList);
-        for(CPLXMLNode* child = !xmlNode ? 0 : xmlNode->psChild ; child != 0 ; child = child->psNext)
+        CPLXMLNode* xmlNode = !pszOptionList ? nullptr : CPLParseXMLString(pszOptionList);
+        for(CPLXMLNode* child = !xmlNode ? nullptr : xmlNode->psChild ; child != nullptr ; child = child->psNext)
         {
             if ((child->eType == CXT_Element) && EQUAL(child->pszValue, "Option"))
             {
-                CPLXMLNode* nameAttribute = CPLGetXMLNode(child, "name");
-                if (nameAttribute != nullptr)
+                const char* nameAttribute = CPLGetXMLValue(child, "name", nullptr);
+                const bool isInterleaveAttribute = nameAttribute && EQUAL(nameAttribute, "INTERLEAVE");
+                if (isInterleaveAttribute)
                 {
-                    const bool isInterleaveAttribute = nameAttribute->psChild &&
-                        (nameAttribute->psChild->eType == CXT_Text) &&
-                        EQUAL(nameAttribute->psChild->pszValue, "INTERLEAVE");
-                    if (isInterleaveAttribute)
+                    for(CPLXMLNode* optionChild = child->psChild ; optionChild != nullptr ; optionChild = optionChild->psNext)
                     {
-                        for(CPLXMLNode* optionChild = child->psChild ; optionChild != 0 ; optionChild = optionChild->psNext)
+                        if ((optionChild->eType == CXT_Element) && EQUAL(optionChild->pszValue, "Value"))
                         {
-                            if ((optionChild->eType == CXT_Element) && EQUAL(optionChild->pszValue, "Value"))
+                            CPLXMLNode* optionChildValue = optionChild->psChild;
+                            if (optionChildValue && (optionChildValue->eType == CXT_Text))
                             {
-                                CPLXMLNode* optionChildValue = optionChild->psChild;
-                                if (optionChildValue && (optionChildValue->eType == CXT_Text))
-                                {
-                                    interleavesCSL = CSLAddString(interleavesCSL, optionChildValue->pszValue);
-                                }
+                                interleavesCSL = CSLAddString(interleavesCSL, optionChildValue->pszValue);
                             }
                         }
                     }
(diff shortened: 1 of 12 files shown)


gcore/gdaldriver.cpp
```diff
@@ -949,16 +960,56 @@ GDALDataset *GDALDriver::CreateCopy( const char * pszFilename,
     if( pfnProgress == nullptr )
         pfnProgress = GDALDummyProgress;
 
+/* -------------------------------------------------------------------- */
+/*      If no INTERLEAVE creation option is given, we will try to add   */
+/*      one that matches the current srcDS interleaving                 */
+/* -------------------------------------------------------------------- */
+    char** papszOptionsSmart = 0;
```
rouault, line 967: to modify options I would suggest you use the existing pattern of char** papszOptionsToDelete = nullptr;   

char** papszOptionsToDelete = nullptr;   
if (we need to modify options )
{
    papszOptionsToDelete = CSLDuplicate(papszOptions);
    // alter papszOptionsToDelete with CSLSetNameValue()
    papszOptions = papszOptionsToDelete;
}

Note: you'll need to add a "if ( papszOptionsToDelete == nullptr )" before "papszOptionsToDelete = CSLDuplicate(papszOptions);" in the branch if( iIdxQuietDeleteOnCreateCopy >= 0 )   (it was the first one to alter the options, so didn't need to test that)

gcore/gdaldriver.cpp
```diff
@@ -949,16 +960,56 @@ GDALDataset *GDALDriver::CreateCopy( const char * pszFilename,
...
+    const bool bHasInterleaveCreationOptions =
+         (CSLFetchNameValue(papszOptions, "INTERLEAVE") != 0);
```
rouault, line 969:
```suggestion
         (CSLFetchNameValue(papszOptions, "INTERLEAVE") != nullptr);
```

gcore/gdaldriver.cpp
```diff
@@ -949,16 +960,56 @@ GDALDataset *GDALDriver::CreateCopy( const char * pszFilename,
...
+    if (!bHasInterleaveCreationOptions && poSrcDS)
+    {
+        const char* srcInterleave = poSrcDS->GetMetadataItem("INTERLEAVE", "IMAGE_STRUCTURE");
+        const bool isENVIDriver = !strcmpi(this->GetDescription(), "ENVI");
```
rouault, line 973: replace !strcmpi() uses by EQUAL(), which is our portability macro

rouault: I'm a big unconfortable with explicitly testing for the ENVI driver. It lacks generality. My thoughts would have rather to parse the XML of GDAL_DMD_CREATIONOPTIONLIST metadata item of the target driver, see if it has a INTERLEAVE option, and what its values are.  If they are PIXEL/BAND  or BSQ/BIP/BIL, you can do the remapping. Otherwise you give up (potentially with a CPLDebug() message so that we know either the driver options must be standardized, or we have to add a new case)

gcore/gdaldriver.cpp
```diff
@@ -949,16 +960,56 @@ GDALDataset *GDALDriver::CreateCopy( const char * pszFilename,
...
+        char** dstInterleaveOptions = (char**)CPLCalloc(1, sizeof(char*));
+        dstInterleaveOptions = CSLSetNameValue(dstInterleaveOptions, "INTERLEAVE",
+            !srcInterleave ?
+              isENVIDriver ? "BSQ" : "BAND" :
+            !isENVIDriver ? srcInterleave :
+            !strcmpi(srcInterleave, "BAND") ? "BSQ" :
+            !strcmpi(srcInterleave, "PIXEL") ? "BIP" :
+            !strcmpi(srcInterleave, "LINE") ? "BIL" :
+            srcInterleave);
+        const bool bIsDstInterleaveSupported = (GDALValidateCreationOptions(
```
rouault, line 983: the issue with GDALValidateCreationOptions() is that it will emit warnings in the standard error if the output driver doesn't support the INTERLEAVE creation options, or with values diffferent than the ones we provide. My above suggestion would avoid this.

gcore/gdaldriver.cpp
```diff
@@ -949,6 +960,82 @@ GDALDataset *GDALDriver::CreateCopy( const char * pszFilename,
     if( pfnProgress == nullptr )
         pfnProgress = GDALDummyProgress;
 
+/* -------------------------------------------------------------------- */
+/*      If no INTERLEAVE creation option is given, we will try to add   */
+/*      one that matches the current srcDS interleaving                 */
+/* -------------------------------------------------------------------- */
+    char** papszOptionsWithInterleave = 0;
```
rouault, line 967: Please use nullptr everywhere you use 0 for a null pointer

gcore/gdaldriver.cpp
```diff
@@ -949,6 +960,82 @@ GDALDataset *GDALDriver::CreateCopy( const char * pszFilename,
...
+    const bool bHasInterleaveCreationOptions =
+         (CSLFetchNameValue(papszOptions, "INTERLEAVE") != nullptr);
+    if (!bHasInterleaveCreationOptions && poSrcDS)
+    {
+        const char* srcInterleave = poSrcDS->GetMetadataItem("INTERLEAVE", "IMAGE_STRUCTURE");
+
+        //look for INTERLEAVE values of the driver
+        char** interleavesCSL = 0;
+        const char *pszOptionList = this->GetMetadataItem( GDAL_DMD_CREATIONOPTIONLIST );
+        CPLXMLNode* xmlNode = CPLParseXMLString(pszOptionList);
```
rouault, line 977:
```suggestion
        CPLXMLNode* xmlNode = pszOptionList ? CPLParseXMLString(pszOptionList) : nullptr;
```

gcore/gdaldriver.cpp
```diff
@@ -949,6 +960,82 @@ GDALDataset *GDALDriver::CreateCopy( const char * pszFilename,
...
+        for(CPLXMLNode* child = !xmlNode ? 0 : xmlNode->psChild ; child != 0 ; child = child->psNext)
+        {
+            if ((child->eType == CXT_Element) && EQUAL(child->pszValue, "Option"))
+            {
+                CPLXMLNode* nameAttribute = CPLGetXMLNode(child, "name");
```
rouault, line 982: you can simplify by using  CPLGetXMLValue(child, "name", nullptr)

gcore/gdaldriver.cpp
```diff
@@ -949,6 +960,82 @@ GDALDataset *GDALDriver::CreateCopy( const char * pszFilename,
...
+            EQUAL(srcInterleave, "BAND") ? dstInterleaveBand :
+            EQUAL(srcInterleave, "LINE") ? dstInterleaveLine :
+            EQUAL(srcInterleave, "PIXEL") ? dstInterleavePixel :
+            nullptr;
+        CSLDestroy(interleavesCSL);
+
+        char** dstInterleaveOptions = !dstInterleave ? 0 :
+          CSLSetNameValue(0, "INTERLEAVE", dstInterleave);
+        const bool bIsDstInterleaveSupported = !dstInterleaveOptions ? FALSE :
+          (GDALValidateCreationOptions(this, dstInterleaveOptions) == TRUE);
```
rouault, line 1029: we no longer need to call GDALValidateCreationOptions()

gcore/gdaldriver.cpp
```diff
@@ -949,6 +960,82 @@ GDALDataset *GDALDriver::CreateCopy( const char * pszFilename,
...
+        if (bIsDstInterleaveSupported)
+        {
+            papszOptionsWithInterleave = CSLDuplicate(papszOptions);
+            papszOptionsWithInterleave = CSLMerge(papszOptionsWithInterleave, dstInterleaveOptions);
```
rouault, line 1033: instead of calling CSLMerge(), you could just do
papszOptionsWithInterleave = CSLSetNameValue(papszOptionsWithInterleave, "INTERLEAVE", dstInterleave);

rouault (commented):
Some testing for that in autotest/gdrivers/envi.py would be appropriate

gcore/gdaldriver.cpp
```diff
@@ -949,6 +960,72 @@ GDALDataset *GDALDriver::CreateCopy( const char * pszFilename,
...
+            nullptr;
+        const char* dstInterleaveLine =
+            (CSLFindString(interleavesCSL, "LINE") >= 0) ? "LINE" :
+            (CSLFindString(interleavesCSL, "BIL") >= 0) ? "BIL" :
+            nullptr;
+        const char* dstInterleavePixel =
+            (CSLFindString(interleavesCSL, "PIXEL") >= 0) ? "PIXEL" :
+            (CSLFindString(interleavesCSL, "BIP") >= 0) ? "BIP" :
+            nullptr;
+        const char* dstInterleave = !srcInterleave ? 0 :
```
rouault, line 1014:
```suggestion
        const char* dstInterleave = !srcInterleave ? nullptr :
```

rouault:
Tests seem to crash on
gdrivers/envi.py::test_envi_writing_interleaving_larger_file[bip]

chacha21:
But is CreateCopy ever called by this test ? Could it be a bug from another copmmit from upstream/master ?

rouault:
> But is CreateCopy ever called by this test ? 

It might be another test step, but at least that was the last one to appear before the crash. Just try to run ``pytest autotest/gdrivers/envi.py``

> Could it be a bug from another copmmit from upstream/master ?

no, all green on latest master: https://github.com/OSGeo/gdal/commit/59591e36d7939b1712e08e16cff3b5679700342a

chacha21:
I just can't make a working pytest environment under Windows for custom gdal builds. I don't have the competence, it's driving me insane, I will have to give up :-(

rouault:
> I just can't make a working pytest environment under Windows for custom gdal builds

If you use cmake builds, you can run the autotests with ctest : https://github.com/OSGeo/gdal/blob/master/.github/workflows/cmake_builds.yml#L174   (provided you've installed the autotest/requirements.txt) . You can restrict to autotest/gdrivers with ctest -V -R  autotest_gdrivers

rouault:
I'm continuing this pull request with a followup commit

chacha21:
Thank you, I really could not find time recently to continue myself.

PR #4930, opened 2021-12-02
MITAB: fix parsing .mid files with newline character in string field (fixes #4789)

### LINKED ISSUE #4789: MITAB: Can't read MID files containing strings with quoted newlines

A customer supplied some MIF data to us, which has MID files where the attribute data contains windows newline characters (`\r\n`)

GDAL fails to read the records which contain the newlines:

```
ERROR 6: Error during reading Record.
```

I guess this issue is [here](https://github.com/OSGeo/gdal/blob/1645ac2d9ee2dae57f99519ab318e02061820e5e/ogr/ogrsf_frmts/mitab/mitab_middatafile.cpp#L160) in the source code. I wonder could this just call the CSV driver, since this MID file appears to be an identical format to a fairly normal CSV file?

Here's a minimal MIF and MID file that cause the error (I removed/replaced proprietary customer data but retained the newlines):

MIF: 
```
Version   410
Charset "WindowsLatin1"
Delimiter ","
Index 1,2
CoordSys Earth Projection 8, 12, "m", 153, 0, 0.9996, 500000, 10000000 Bounds (-7745874.38492, 1999.40969607) (8745874.38492, 19998000.5903)
Columns 3
  Name Char(51)
  Notes Char(100)
  Awesome Logical
Data

Pline 3
482281.14 6738543.72
482278.29 6738548.02
482276.11 6738551.17
    Pen (4,104,8421504) 
```

MID: 
```
"NEWNAME ROAD","NAME CHANGED FROM OLDNAME ROAD  11-05-01

Oldname Rd (No.123) added 11-05-01",F
```

## Operating system

MacOS 11.6

GDAL 3.3.2, released 2021/09/01

rouault:
@craigds I'm somehow nervous to backport this to the 3.4 branch. Do you have a strong need for that or keeping that as master-only material would be OK ?


craigds:
I think it's probably fine on master; we can backport ourselves on our fork if we need to.

Thanks for fixing this! I was thinking I'd have to implement it myself but got a bit lost trying to think about it and so it went on the Big Long List Of Future Projects

PR #4937, opened 2021-12-03
gdal2tiles: making --no-kml option takes effect fixes (fixes #4936)

### LINKED ISSUE #4936: gdal2tiles: --no-kml option does not takes effect

Actual behavior: kml file are generated for each tile even though --no-kml option is set.
Expected behavior: no kml files are generated with --no-kml option.

```bash
gdal2tiles.py -p geodetic -z 0-2 --no-kml autotest/gdrivers/data/small_world.tif /tmp/small_world_tiles
```

## Operating system

ArchLinux

GDAL 3.4.0dev-55e83bde78-dirty, released 2021/12/02

kikitte:
## What does this PR do?
The user setting of generating kml file or not should not be modified, and the current behavior is generating kml file for each tile even though --no-kml option is set. 
https://github.com/OSGeo/gdal/blob/59591e36d7939b1712e08e16cff3b5679700342a/swig/python/gdal-utils/osgeo_utils/gdal2tiles.py#L1777-L1781

Automatic generation of kml file is disabled by default, but some tests rely on the incorrect behavior which is described above, so modification to the tests is required.
https://github.com/OSGeo/gdal/blob/59591e36d7939b1712e08e16cff3b5679700342a/swig/python/gdal-utils/osgeo_utils/gdal2tiles.py#L1415-L1418

### DIFF
--- a/autotest/pyscripts/test_gdal2tiles.py
+++ b/autotest/pyscripts/test_gdal2tiles.py
@@ -356,7 +356,7 @@ def test_gdal2tiles_py_profile_raster():
     test_py_scripts.run_py_script_as_external_script(
         script_path,
         'gdal2tiles',
-        '-q -p raster -z 0-1 '+test_py_scripts.get_data_path('gdrivers')+'small_world.tif tmp/out_gdal2tiles_smallworld')
+        '-q -p raster -z 0-1 --force-kml '+test_py_scripts.get_data_path('gdrivers')+'small_world.tif tmp/out_gdal2tiles_smallworld')
 
     if sys.platform != 'win32':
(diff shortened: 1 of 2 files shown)


rouault:
That PR will change the behaviour of the utility. Currently if the SRS is EPSG:4326 and nothing is specifed, then KML is assumed, which is reasaonble. The fix should rather make sure that wen --no-kml is specified it doesn't output KML.  This will likely require to change the option parser to have separate variables, or have the kml option be a tri-state (None=automatic choice depending on the CRS, False=disable, True=force)

kikitte:
I think the tri-state may be the best solution, which will be less influenced compared to using another new option to achieve to goal.

PR #4942, opened 2021-12-04
TopoJSON: fix duplicate 'id' field, and other potential issues when reading fiels (3.4.0 regression)

rouault:
Fix issue raised on https://lists.osgeo.org/pipermail/gdal-dev/2021-December/055090.html


simoneves:
Thank you. Will this make it into 3.4.1?

rouault:
> Thank you. Will this make it into 3.4.1?

yes

PR #4944, opened 2021-12-04
PCIDSK: fix write heap-buffer-overflow. Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=41993

rouault:


vcunat:
Nit: the coding style seems very inconsistent with the surrounding code.

PR #4945, opened 2021-12-04
Selafin: fix time step count when none dataset in the file

vcloarec:
SELAFIN files, without dataset but only with the mesh frame, can exist, especially coming from editors as  Bluekenue, Basemesh SalomeHydro, or more recently QGIS.

When opening such file by drag/drop such files in QGIS, opening the file is tested with OGR driver, which returns a very big (and bad) layer count. That make QGIS crashes or being freeze.

This little PR fixes this issue. In such case, the file can't be opened by OGR, but it doesn't consider anymore bad layer count.

### DIFF
--- a/ogr/ogrsf_frmts/selafin/io_selafin.cpp
+++ b/ogr/ogrsf_frmts/selafin/io_selafin.cpp
@@ -634,7 +634,7 @@ namespace Selafin {
             delete poHeader;
             return nullptr;
         }
-        vsi_l_offset nStepsBig = (poHeader->nFileSize-nPos)/(poHeader->getPosition(1)-nPos);
+        vsi_l_offset nStepsBig = poHeader->nVar != 0 ? (poHeader->nFileSize-nPos)/(poHeader->getPosition(1)-nPos) : 0;
         if( nStepsBig > INT_MAX )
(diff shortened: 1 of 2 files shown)


rouault:
It wouldn't hurt to add some validation of the layer count, I guess around line 417 of ogrselafindatasource.cpp. Potentially before the static_cast<int>(poRange.getSize()) if getSize() can return > INT_MAX. Perhaps have a OGR_SELAFIN_MAX_LAYER_COUNT configuration option with a default to 1000 ?

vcloarec:
validation of layer count added.

> Perhaps have a OGR_SELAFIN_MAX_LAYER_COUNT configuration option with a default to 1000 ?

Not sure we can supposed there are less than 1000 time step in the file... There could be more.

rouault:
> Not sure we can supposed there are less than 1000 time step in the file... There could be more.

yeah, hence the idea of the runtime configurable option that the user can raise if needed. Maybe 1000 is too small, but we can't reasonably support an arbitrary number. One issue is that we preallocate the papoLayers array for all potential layers, but when date filtering is specified, we can have much less. Anyway, that can be something for later improvements.

rouault:
(no need to worry about the Conda failures. they are unrelated)

PR #4966, opened 2021-12-09
COG: tune a bit weight of the various stages in progression report

rouault:


rouault:
would probably require a better testing of the various weights of the different steps. dropping that for now

PR #4979, opened 2021-12-11
MSSQL: do not set field width when reading smallint/int/bigint/float/…

### LINKED ISSUE #3345: Integer schema columns converted to numeric (decimal) columns with OGRSQL dialect

I expect to be able to translate a table from MSSQLServer to PostGIS with as close to the same schema as reasonably possible. 

The actual behavior is that a column of type `smallint` in mssql translates to a `numeric(5)` in postgis and `int` translates to a `numeric(10)`. Numeric field types imply decimal types. Postgresql has the [equivalent types](https://www.postgresql.org/docs/9.1/datatype-numeric.html) but I assume the dialect isn't able to infer from the sql query? 

One issue created by this is that when using this data in a service that serializes these types to json or other formats, the output is different. `10` becomes `10.0` because numeric is interpreted as a decimal. However insignificant the `.0`, typed deserializers will break on the int vs decimal property type.

Create a table with a `smallint`, `int`, or `bigint` and run code similar to [this](https://github.com/agrc/open-sgid/blob/main/src/cloudb/main.py#L256-L277) to convert from mssql to postgis.

```py
sql = 'SELECT "small_int_field", "int_field", "big_int_field" FROM "schema_of.the_table_created_above"'

options = [
  '-f',
  'PostgreSQL',
  '-dialect',
  'OGRSQL',
  '-sql',
  sql,
  'OVERWRITE=YES'
]

pg_options = gdal.VectorTranslateOptions(options=options)
result = gdal.VectorTranslate(postgis_connection, mssql_connection, options=pg_options)
```

## Operating system

Windows Server 2016 Standard

3.2.1 from conda

rouault:
…real columns, and correctly roundtrip smallint/Int16 (fixes #3345)

### DIFF
--- a/autotest/ogr/ogr_mssqlspatial.py
+++ b/autotest/ogr/ogr_mssqlspatial.py
@@ -321,33 +321,190 @@ def test_ogr_mssqlspatial_create_feature_in_unregistered_table():
     feature.Destroy()
 
 ###############################################################################
-# Test UUID datatype support
+# Test all datatype support
 
-def test_ogr_mssqlspatial_uuid():
+def test_ogr_mssqlspatial_datatypes():
     if gdaltest.mssqlspatial_ds is None:
         pytest.skip()
 
-    lyr = gdaltest.mssqlspatial_ds.CreateLayer('test_ogr_mssql_uuid')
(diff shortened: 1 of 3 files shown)


rouault:
CC @bjornharrtell @szekerest

bjornharrtell:
LGTM.

PR #4995, opened 2021-12-14
CMake: add/fix CSharp support

rouault:
Note that it is currently disabled on the CMake Windows CI because we
get the following error:
```
Microsoft (R) Visual C# 2005 Compiler version 8.00.50727.9031
for Microsoft (R) Windows (R) 2005 Framework version 2.0.50727
Copyright (C) Microsoft Corporation 2001-2005. All rights reserved.
d:\a\gdal\gdal\swig\csharp\AssemblyInfo.cs(98,12): error CS0246: The type or namespace name 'SecurityRules' could not be found (are you missing a using directive or an assembly reference?) [D:\a\gdal\gdal\build\swig\csharp\csharp_samples.vcxproj]
```

Things works fine locally for me with Microsoft (R) Visual C# Compiler version 3.9.0-6.21124.20 (db94f4cc)

CC @szekerest  / @bjornharrtell  w.r.t above issue. I haven't also ported swig/csharp/makefile.vc logic related to using dotnet rather than mono. I hope any of will be able to pursue on this.

### DIFF
--- a/.github/workflows/cmake_builds.yml
+++ b/.github/workflows/cmake_builds.yml
@@ -115,8 +115,9 @@ jobs:
       #  libxml2 disabled because currently causes a 'Imported target "LibXml2::LibXml2" includes non-existent path "/mingw64/include/libxml2" in its INTERFACE_INCLUDE_DIRECTORIES'
       # Disable mySQL since C:/mysql/lib/mysqlclient.lib (unrelated to msys) is found, which causes linking issues
       # Disable Python bindings because of 'ValueError: filename D:/a/gdal/gdal/swig/python/osgeo/__init__.py does not start with the input_base_dir D:/a/gdal/gdal/swig/python/osgeo/\' when running lib2to3
+      # BUILD_CSHARP_BINDINGS=OFF because we get "AssemblyInfo.cs(98,12): error CS0246: The type or namespace name 'SecurityRules' could not be found (are you missing a using directive or an assembly reference?)" with Microsoft (R) Visual C# 2005 Compiler version 8.00.50727.9031 for Microsoft (R) Windows (R) 2005 Framework version 2.0.50727
       - name: Configure
-        run: cmake -DCMAKE_BUILD_TYPE=release -G "${env:generator}" "-DCMAKE_C_COMPILER_LAUNCHER=c:\tools\msys64\mingw64\bin\ccache.exe" "-DCMAKE_CXX_COMPILER_LAUNCHER=c:\tools\msys64\mingw64\bin\ccache.exe" "-DCMAKE_PREFIX_PATH=C:\tools\msys64\mingw64" "-DCMAKE_UNITY_BUILD=${env:CMAKE_UNITY_BUILD}" -S . -B "build" -DLIBXML2_INCLUDE_DIR=C:/tools/msys64/mingw64/include -DGDAL_USE_LIBXML2:BOOL=OFF -DGDAL_USE_MYSQL:BOOL=OFF -DBUILD_PYTHON_BINDINGS:BOOL=OFF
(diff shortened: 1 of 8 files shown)


rouault:
@runette you may also be interested in this. Looking at https://gdal.org/api/csharp/csharp_conda.html, I see this mentions "On Mac and Linux, the SWIG files are built as *_wrap in line with the windows versions. This means that there are no .config files.", and indeed when porting the existing GNUmakefile/makefile.vc of swig/csharp to CMake, I noticed this difference too. I'm not sure why it exists at all (@szekerest ?) and if it is something desirable. So in doubt I ported it (with a slight difference in naming in the generated .so file since it was easier to code it), but maybe we shouldn't ?  (remainder: I'm illiterate in the C# world!)

runette:
Config files have been removed from the latter versions of .NET because
they are a security risk.

Therefore - a solution without config files is better.

I am not totally sure why the non-windows assemblies are given a different
name anyway. I think there was a historic reason but it is not needed now.

runette:
BTW - I have the makefiles to create the version without config files - you
can find it here

https://github.com/conda-forge/gdal-csharp-feedstock/blob/master/recipe/makefile.vc

rouault:
> Therefore - a solution without config files is better.

I've added commits to remove config files for Unix too

runette:
I am adding a comment here just for future auditability - because what I said above due to poor wording can be interpreted as 100% correct or just wrong :). No action needed.

"config files" have not been removed from .NET! The use that GDAL made of the config files has been - more accurately I should have said the "DllMap" functionality has been removed from .NET.

This has been true of dotnet itself in existing versions and is also now true of mono as well - see this post https://www.mono-project.com/news/2020/08/24/native-loader-net5/

runette:
I have been trying to get this version to build locally so that I can have a look at the results - with no luck so far. 

However, I have a question for @rouault : in the previous version it was possible to run the swig build standalone on top of pre- built gdal binaries. That was very useful. It looks to me like that is not possible with the Cmake files?

rouault:
> I have been trying to get this version to build locally so that I can have a look at the results - with no luck so far.

When running cmake, do you see a "Find Mono: ...." line somewhere in the report ?

>  in the previous version it was possible to run the swig build standalone on top of pre- built gdal binaries. That was very useful. 

I didn't know this was possible. Remainder: I'm not the C# maintainer :-) 

> It looks to me like that is not possible with the Cmake files?

Probably not in the current state indeed. Contributions welcome

runette:
> > I have been trying to get this version to build locally so that I can have a look at the results - with no luck so far.
> 
> When running cmake, do you see a "Find Mono: ...." line somewhere in the report ?

The Swig and Mono configuration all looks good but on both windows and mac - it has a problem with the file `cpl_config.h` - although the error is at a different point on each architecture. Is there an address I can DM you on to discuss? It wanted to get it working to test the result.

> I didn't know this was possible. Remainder: I'm not the C# maintainer :-)

Understood - although it is not a C# thing - it was a make thing. The make files had an entry point to just run the build on the csharp targets (as described in https://gdal.org/api/csharp/csharp_compile.html. I am not experienced in cmake - but if I can get a working build environment I will give it some thought.

rouault:
> Is there an address I can DM you on to discuss?

IRC / the Slack channel could do

runette:
For information:

I have created a solution to the above problem by converting the windows build to use `dotnet` (which is what the existing makefile was doing). The Solution is in this branch and not completed ... 

BRANCH : https://github.com/runette/gdal/tree/csharp-fix
Cmake file : https://github.com/runette/gdal/blob/csharp-fix/swig/csharp/CMakeLists.txt

I have made it more modern and generic by :

1 Defaulting all platforms to `dotnet` consistently - which should work eventually. There is a switch to force the use of MONO and that is still what is being used in the Linux and Mac builds because there is a NuGet bug at the moment ... If `dotnet` not found then {win only} `csc.exe` is looked for (not sure this step is needed these days but I have not removed it) and then MONO I looked for.

2 I have defaulted the .NET version (for dotnet builds) for both libraries and sample apps to net5.0. This creates some future proofing. The version used in each case is (separately) still settable by a command line switch.

It is not working yet - but it is "mostly working" . Still working out a number of bugs. You can look in the actions for that repo to see the latest runs - I am having to debug mostly in GH actions since I cannot seem to get reliable local build environments (and I cannot dedicate a machine or two).

PR #4996, opened 2021-12-14
OGR GMT: allow writing to /vsistdout/ (fixes #4993)

### LINKED ISSUE #4993: Allow output to stdout when using e.g. GMT/OGR
What is the rationale that output to `stdout` (`/vsistdout/`) is not possible when using e.g. the GMT/OGR driver?

```
$ ogr2ogr -sql "select * from prlAreaCurrent" -F GMT /vsistdout/ prlAreaCurrent.gpkg
# @VGMT1.0 @GPOLYGON
# REGION_STUB                                                             
# @Je4230
# @Jp"+proj=longlat +ellps=intl +no_defs"
# @Jw"GEOGCS[\"ED50\",DATUM[\"European_Datum_1950\",SPHEROID[\"International 1924\",6378388,297,AUTHORITY[\"EPSG\",\"7022\"]],AUTHORITY[\"EPSG\",\"6230\"]],PRIMEM[\"Greenwich\",0,AUTHORITY[\"EPSG\",\"8901\"]],UNIT[\"degree\",0.0174532925199433,AUTHORITY[\"EPSG\",\"9122\"]],AXIS[\"Latitude\",NORTH],AXIS[\"Longitude\",EAST],AUTHORITY[\"EPSG\",\"4230\"]]"
ERROR 6: Read or update mode not supported on /vsistdout
translation from sql statement.
```

This is basically a script, dumping all the data in the geopackage file to a GMT/OGR file.

It would be very beneficial if this was possible.
I'm using the GMT/OGR driver to output data in a script, generating a movie with > 20.000 frames. If `ogr2ogr` allowed me to send output to `stdout` I would not need to write > 20.000 temporary files to disk, but instead pipe directly to the plotting utility.

Also something, not quite related, that I found interesting:

```
$ time ogr2ogr -dialect indirect_sqlite -sql "select st_buffer(st_transform(geom, 32632),4000) as geomerfds from prlAreaCurrent WHERE geom IS NOT NULL" -F GMT temp.gmt prlAreaCurrent.gpkg

real 0m0.916s
user 0m0.881s
sys 0m0.032s
```
```
$ time ogr2ogr -sql "select st_buffer(st_transform(geom, 32632),4000) as geomerfds from prlAreaCurrent WHERE geom IS NOT NULL" -F GMT temp2.gmt prlAreaCurrent.gpkg

real 0m6.888s
user 0m6.271s
sys 0m0.617s
```

Using the `-dialect indirect_sqlite` is much faster (7.5x faster) than using the _native SQL dialect_ (as I would assume is being used as per description in https://gdal.org/user/sql_sqlite_dialect.html). On the man page, it also says that using `-dialect indirect_sqlite` should be avoided : _This should be used only when necessary, since going through the virtual table mechanism might affect performance._ The opposite seems to be the case. Is this due to better plumbing in the `gdal` machinery, and/or outdated documentation?

rouault:
CC @joa-quim Can you confirm that a OGR GMT file that lacks the "# @R" region header will work properly ? (see https://github.com/OSGeo/gdal/issues/4993#issuecomment-993450875)

### DIFF
--- a/autotest/ogr/ogr_gmt.py
+++ b/autotest/ogr/ogr_gmt.py
@@ -213,3 +213,18 @@ def test_ogr_gmt_coord_only():
         lyr = ds.GetLayer(0)
         f = lyr.GetNextFeature()
         assert not ogrtest.check_feature_geometry(f, 'POINT Z (1 2 3)'), f.GetGeometryRef().ExportToIsoWkt()
+
+
+###############################################################################
+# Test writing to stdout
+
+
+def test_ogr_gmt_write_stdout():
+
+    gdal.VectorTranslate("/vsistdout_redirect//vsimem/test.gmt", "data/poly.shp", format="GMT")
+    ds = ogr.Open("/vsimem/test.gmt")
+    lyr = ds.GetLayer(0)
+    assert lyr.GetGeomType() == ogr.wkbPolygon
+    assert lyr.GetFeatureCount() == 10
+    ds = None
+    gdal.Unlink("/vsimem/test.gmt")
--- a/doc/source/drivers/vector/gmt.rst
+++ b/doc/source/drivers/vector/gmt.rst
@@ -35,3 +35,6 @@ features to existing files, but update of existing features is not
 supported. Each layer is created as a separate .gmt file. If a name that
 ends with .gmt is not given, then the GMT driver will take the layer
 name and add the ".gmt" extension.
+
+Writing to /dev/stdout or /vsistdout/ is supported since GDAL 3.5.0 (note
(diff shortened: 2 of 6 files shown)


joa-quim:
No, ``@R`` is not mandatory. For example because we can set the bounding box in another previous GMT command. It may happen that to convert from OGR_GMT to some other OGR format the ``@R`` may be necessary (but just guessing because I don't have much experience with this  format).

As a test I removed the ``@R`` line from the test file [line1.gmt](https://github.com/GenericMappingTools/gmt/blob/master/test/psxy/line1.gmt) and the test that uses it produced the same result.

PR #5024, opened 2021-12-21
fix/workaround setuptools 60.0 and Debian --install-layout 

rouault:
setuptools 60.0 defaults to using its internal copy of distutils
(see https://github.com/pypa/setuptools/blob/main/CHANGES.rst#v6000),
and thus the patched Debian's distutils is not used, which makes the
--install-layout option unavailable.
So force SETUPTOOLS_USE_DISTUTILS=stdlib environment variable when we
use --install-layout


rouault:
@sebastic  Do you know what is the plan of Debian regarding setuptools 60.0 which defaults to using its internal copy of distutils
(see https://github.com/pypa/setuptools/blob/main/CHANGES.rst#v6000), and thus the patched Debian's distutils is not used ? For now I'm setting SETUPTOOLS_USE_DISTUTILS=stdlib environment variable which works around the issue.
Another thing I noticed is that recent setuptools version display "SetuptoolsDeprecationWarning: setup.py install is deprecated. Use build and pip and other standards-based tools.", so I guess we'll have to transition at some point to something else for the install target of the Python bindings. Are you aware what other projects do ?

sebastic:
> Do you know what is the plan of Debian regarding setuptools 60.0 which defaults to using its internal copy of distutils
(see https://github.com/pypa/setuptools/blob/main/CHANGES.rst#v6000), and thus the patched Debian's distutils is not used ?

No, I'm not in the loop of the maintainers plans for this. Try asking [him](https://qa.debian.org/developer.php?email=doko%40debian.org) directly.

> Another thing I noticed is that recent setuptools version display "SetuptoolsDeprecationWarning: setup.py install is deprecated. Use build and pip and other standards-based tools.", so I guess we'll have to transition at some point to something else for the install target of the Python bindings. Are you aware what other projects do ?

`pyproject.toml` ([PEP-517](https://www.python.org/dev/peps/pep-0517/)) seems to be the new popular kid on the block.

`pybuild`, the defacto standard build tool for Python packages in Debian, recently got [support](https://lists.debian.org/debian-devel/2021/12/msg00213.html) for that.

rouault:
@snowman2 Any hints w.r.t above discussion (setuptools deprecating setup.py install)

rouault:
CC @idanmiara  too

snowman2:
I think that you need to use `python -m pip install .` instead of `python setup.py install`

PR #5046, opened 2021-12-27
CMAKE : Csharp Build Process working on Win / Mac / Linux

runette:
## What does this PR do?

Fixes the problem with building the C# bindings on Windows.

- Builds by default using `dotnet`. If it cannot find `dotnet` it looks for mono.
- Adds a compiler switch to allow the use of mono to be forced
- Compiles by default both libraries and executables when using `dotnet` to be `net5.0` but there are compiler switches to allow other versions to be used.
- if built using `dotnet`, adds NuGET packages for the bindings using an OSGeo namespace
- adds install commands for the `xxx_csharp.dll`s and the sample binaries
- adds targets to allow the bindings to be built on an existing installation of GDAL
- the name of the sample program `ogrinfo.exe` has been renamed `OgrInfo.exe` since it appears Cmake does not like having multiple targets with the same name. This is consistent with the other sample binary names and to keep consistency `createdata.exe` has been renamed `CreateData.exe`
- add two extra compile tests for building under Linux with Mono and for building standalone
- Compiles successfully for Windows (including MSYS2). Linux and MacOS using dotnet and mono (but the mono and windows combination has not been tested)
- Adds documentation
- Adds the complete set of tests from the legacy scripts

#4995


swig/csharp/CMakeLists.txt
```diff
@@ -9,225 +9,440 @@ set(GDAL_CSHARP_INSTALL_DIR
...
-function (gdal_csharp_dll)
+set(CSHARP_LIBRARY_VERSION
+    "net5.0"
+    CACHE STRING ".NET version to be used for libraries")
+set(CSHARP_APPLICATION_VERSION
+    "net5.0"
+    CACHE STRING ".NET version to be used for the sample Applicatons")
+set(CSHARP_CONFIG
+    "RELEASE"
+    CACHE STRING "Config to be used to compile the C# atrefacts > RELEASE|CONFIG")
```
rouault, line 20:
```suggestion
    CACHE STRING "Config to be used to compile the C# artefacts > RELEASE|CONFIG")
```

cmake/modules/thirdparty/FindCSharp.cmake
```diff
@@ -33,16 +34,28 @@ if( NOT ${CSHARP_PLATFORM} MATCHES "x86|x64|anycpu|itanium" )
...
-  find_package( DotNetFrameworkSdk )
-  if( NOT CSHARP_DOTNET_FOUND )
-    find_package( Mono )
-  endif( )
-else( UNIX )
+if( CSHARP_MONO )
   find_package( Mono )
-endif( )
+else()
+  find_package( Dotnet )
```
rouault, line 40: what is the difference between the find_package(Dotnet) and find_package( DotNetFrameworkSdk ) ? Should we really keep both, or one can supersede the other one ?

rouault (commented):
Thanks for tackling that ! A few comments/questions below + https://github.com/runette/gdal/pull/3

cmake/modules/thirdparty/FindCSharp.cmake
runette: FindDotnet is looking for `dotnet.exe` and FindDotNetFrameworkSdk is looking for `csc.exe`. `dotnet.exe` is the modern version and covers Net5.0 and compatible earlier versions - like .NET Standard 2.0 and 2.1 etc. I believe that there if you have .NET Framework you would still need to look for `csc.exe` - so I have left it in. It is largely a legacy thing - since I don't think anyone does actually compile the bindings with .NET Framework.

swig/csharp/CMakeLists.txt
runette: fixed

cmake/modules/thirdparty/FindCSharp.cmake
rouault: > It is largely a legacy thing - since I don't think anyone does actually compile the bindings with .NET Framework.

@szekerest @bjornharrtell opinons ? i'd say that if it is a legacy thing, we could remove support from that if that can simplify things

bjornharrtell: In my opinion it is fine to drop .NET Framework. Since .NET 5 I cannot imagine any reason to look back regardless of platform etc.

runette: That looks like a quorum to drop .NET Framework support and csc.exe. Give me a nod and I will push a commit through to do that.

bjornharrtell:
@runette if I remember correctly you do not use nuget packages right? But perhaps this PR could possibly help create proper nuget packages so that forks like https://github.com/MaxRev-Dev/gdal.netcore could be avoided?

runette:
> @runette if I remember correctly you do not use nuget packages right? But perhaps this PR could possibly help create proper nuget packages so that forks like https://github.com/MaxRev-Dev/gdal.netcore could be avoided?

@bjornharrtell  I actually have had the same thought and have done a little work in that direction!

I only don't use NuGET because my projects in C# are all Unity based and Unity does not use NuGET - strangely it uses npm :). But - with NET 5.0 based dotnet compilation on all platforms I see scope for a relatively easy toolchain to create nuget packages.

Lets get the basic cmake compile scripts working and accepted and come back to that thought...

cmake/modules/thirdparty/FindCSharp.cmake
runette: In the absence of further comment - I have gone ahead and removed .NET Framework

swig/csharp/CMakeLists.txt
```diff
@@ -1,4 +1,10 @@
-include(GdalStandardIncludes)
+option(GDAL_CSHARP_ONLY "Compile C# bindings on existing GDAL installation" OFF)
+
+if (GDAL_CSHARP_ONLY)
+  find_package(GDAL REQUIRED)
+else (GDAL_CSHARP_ONLY)
```
rouault, line 5: else (foo) / endif (foo) are deprecated practices in modern CMake. just use else() / endif(). Other occurences below

swig/csharp/CMakeLists.txt
```diff
@@ -9,226 +15,626 @@ set(GDAL_CSHARP_INSTALL_DIR
...
+          ReadXML.exe
+          GDALInfo.exe
+          GDALOverviews.exe
+          GDALCreateCopy.exe
+          GDALGetHistogram.exe
+          GDALTest.exe)
 
-if (CSHARP_INTERPRETER OR WIN32)
+# set up the tests
+if (CSHARP_INTERPRETER
```
rouault, line 635: parentheses around member of the OR expression would help: (CSHARP_INTERPRETER OR WIN32) AND NOT GDAL_CSHARP_ONLY

.github/workflows/cmake_builds.yml
```diff
@@ -74,6 +74,64 @@ jobs:
         $GITHUB_WORKSPACE/install-gdal/bin/gdalinfo --version
         PYTHONPATH=$GITHUB_WORKSPACE/install-gdal/lib/python3/dist-packages python3 -c "from osgeo import gdal;print(gdal.VersionInfo(None))"
 
+  build-linux-ubuntu-focal-using-mono:
```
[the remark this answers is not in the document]
runette: Excellent Idea - I will do that

swig/csharp/CMakeLists.txt
runette: Sorry - old habit. I will change in the next commit

swig/csharp/CMakeLists.txt
```diff
@@ -9,226 +15,624 @@ set(GDAL_CSHARP_INSTALL_DIR
...
+      ${_VERSION_STRING}
+      OUTPUT_PATH
+      ${_CSHARP_PROJ_PATH}
+      DEPENDS
+      ${_deps}
+      PACKAGE
+      ${_CSHARP_SYSTEM_DEPENDS})
+
+  else ()
+    message("BUILDING : " ${_CSHARP_TARGET} " : " ${CSHARP_COMPILER} ${CSC_OPTIONS} /out:${_CSHARP_TARGET})
```
rouault, line 186: could be removed or commented to avoid too verbose log

swig/csharp/CMakeLists.txt
```diff
@@ -9,226 +15,624 @@ set(GDAL_CSHARP_INSTALL_DIR
...
+      ${_VERSION_STRING}
+      OUTPUT_PATH
+      ${_GBCS_PROJ_PATH_NATIVE}
+      DEPENDS
+      csharp_binding
+      PACKAGE
+      ${_GBCS_SYSTEM_DEPENDS})
+
+  else ()
+    message(
```
rouault, line 363: could be removed or commented to avoid too verbose log

swig/csharp/CMakeLists.txt
```diff
@@ -247,37 +651,4 @@ if (CSHARP_INTERPRETER OR WIN32)
     set(TEST_ENV "${NEW_TEST_ENV}")
   endif ()
 
-  add_test(NAME csharp_createdata COMMAND ${CSHARP_INTERPRETER} createdata.exe
```
rouault, line 250: the tests should be added back

swig/csharp/CMakeLists.txt
runette: Made into --verbose only

swig/csharp/CMakeLists.txt
runette: made into --verbose only

swig/csharp/CMakeLists.txt
runette: Added the complete set of tests from the legacy scripts and tested and working on Win / Mac / Linux

runette:
@rouault -- there seems to be some error in the CI tests that has nothing to do with any subsystem that I have touched and did not happen in my fork using the same test  - see https://github.com/OSGeo/gdal/runs/4861157262?check_suite_focus=true. Could this be transient?

Other than that - as far I am concerned this PR is now working. I would propose that it should be merged.

swig/csharp/CMakeLists.txt
```diff
@@ -9,275 +15,753 @@ set(GDAL_CSHARP_INSTALL_DIR
...
+  NAME csharp_ogrlayeralgresult
+  WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+  COMMAND ${CSHARP_INTERPRETER} OgrInfo/OgrInfo Data/intersectionresult.shp)
+set_property(TEST csharp_ogrlayeralgresult PROPERTY ENVIRONMENT "${TEST_ENV}")
+set_property(TEST csharp_ogrlayeralgresult PROPERTY DEPENDS csharp_ogrlayeralg)
+add_test(
+  NAME csharp_gdalwarp
+  WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+  COMMAND ${CSHARP_INTERPRETER} GDALWarp/GDALWarp Data/testwarp.tif "-of GTiff -t_srs EPSG:32645 -overwrite"
+          "../../autotest/gcore/data/byte.tif")
```
rouault, line 760: if people build from a tarball, the autotest directory will not be present. You could just copy byte.tif into the swig/csharp/Data directory.  EPSG:32645 is not super relevant for that dataset. EPSG:32611 would be better. And the -overwrite flag is only used for the gdalwarp binary, not for the gdal.Warp() API

swig/csharp/CMakeLists.txt
```diff
@@ -9,275 +15,753 @@ set(GDAL_CSHARP_INSTALL_DIR
...
+
+# set up the tests
+
+include(GdalSetTestEnv)
+gdal_set_test_env(TEST_ENV)
+
+add_test(
+  NAME csharp_createdata
+  WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+  COMMAND ${CSHARP_INTERPRETER} CreateData/CreateData Data pointlayer)
```
rouault, line 694: I'm wondering how the Data subdirectory is found, as it is normally in the source tree and here the working directory is the binary directory. Is there some copying somewhere ? I can't see under the swig/csharp directory of my build tree.

Another issue I have on Ubuntu 20:04 with mcs --version == Mono C# compiler version 6.8.0.105 is that tests fail to run (the version on GitHub actions seems to be a non-stock one: "Found the following C# Mono versions: 6.12.0")
The errors I get are like:
```
test 16
      Start 16: csharp_createdata

16: Test command: /usr/bin/mono "CreateData/CreateData" "Data" "pointlayer"
16: Environment variables: 
16:  LD_LIBRARY_PATH=/home/even/gdal/gdal/build_cmake:/home/even/idb_client/lib/esql:/home/even/FileGDB_API-RHEL7-64gcc83/lib:/home/even/install-mongocxx-3.6.5/lib:/home/even/gdal/gdal/build_cmake:/home/even/spatialite/install-libspatialite-5.0.1/lib:/home/even/MrSID_DSDK-9.5.4.4709-rhel6.x86-64.gcc531/Raster_DSDK/lib:/home/even/MrSID_DSDK-9.5.4.4709-rhel6.x86-64.gcc531/Lidar_DSDK/lib:/home/even/kakadu/v8_2-00679N:/home/even/instantclient_12_1:
16:  GDAL_DRIVER_PATH=/home/even/gdal/gdal/build_cmake/gdalplugins
16:  GDAL_DATA=/home/even/gdal/gdal/data
16: Test timeout computed to be: 1500
16: Cannot open assembly 'CreateData/CreateData': No such file or directory.
 1/13 Test #16: csharp_createdata ................***Failed    0.01 sec
```
And indeed there's no CreateData/CreateData file, but an empty CreateData directory:
```
even@even-ThinkPad-P15v-Gen-1:~/gdal/gdal/build_cmake$ ll swig/csharp/CreateData.exe 
-rwxrwxr-x 1 even even 7168 janv. 19 11:29 swig/csharp/CreateData.exe*
even@even-ThinkPad-P15v-Gen-1:~/gdal/gdal/build_cmake$ ll swig/csharp/CreateData
total 8
drwxrwxr-x  2 even even 4096 janv. 19 11:28 ./
drwxrwxr-x 29 even even 4096 janv. 19 11:29 ../
```

rouault (commented):
Thanks for the latest updates. A few remarks below. (I've restarted the failed build. it was a network flake)

swig/csharp/CMakeLists.txt
runette: The Data directory for these tests is (and was in the legacy scripts) created by the CreateData executable in the working directory.

That is also - for information - why I needed to add dependencies to the tests - they need to depend on CreateData completing and given the multiple process nature of cmake and ctest there were problems with some tests starting before Data was created.

runette: The problem with Mono -- I think I know what that is. The directory structure is slightly different when building with Mono. I need to add something there.

swig/csharp/CMakeLists.txt
runette: I think that might be a deeper question. Given the way that CMAKE works - when you build the bindings standalone although it is only building the csharp targets, it is still generating all targets and all tests. So - if you build from a tarball - I don't actually think you can just run a naked  `ctest` anyway - I think you would have to only run the individual tests you want anyway - e.g. `ctest -R "^csharp.*"`. My opinion is - since building stand-alone is not the core workflow - this should be documented and possibly raised as an issue for more investigation but should not be a blocker.

As for the details of the warp function - this is just what we used before. I would guess that it was created a number of years ago originally :) 

I would suggest - at least for the time being - that we could remove this test. I feel we have enough other tests and it would avoid any problems with this test.

rouault: > I would suggest - at least for the time being - that we could remove this test. I feel we have enough other tests and it would avoid any problems with this test.

Fine for me. Or just comment it out with a note regarding the reason

swig/csharp/CMakeLists.txt
runette: Mono Tests fixed and working and ctest running on the Mono test build

swig/csharp/CMakeLists.txt
runette: I commented out the test and then changed the test standalone build that builds on a conda libgdal distribution extended to run the ctest tests successfully

rouault:
I confirm the charp mono tests work with stock Ubuntu 20.04. Thanks for all the work on this!

runette:
> I confirm the csharp mono tests work with stock Ubuntu 20.04. Thanks for all the work on this!

Thanks

Just a note - I will be most out of contact and certainly not able to run builds from 16 Feb for two months - so I would like to get this to the point of being merged before that.

szekerest:
@rouault Looks like this addition is in a good shape to get it included, right?

rouault:
Merged. Thanks a lot !

PR #5058, opened 2022-01-04
GRIB2: use CSV tables instead of hard-coded tables

rouault:
- mode degrib hard-coded tables (Table 4.2 that describes the parameter and its units) to .csv files in the data/ directory 
- merge their content with CSV files from WMO (https://github.com/wmo-im/GRIB2)


rouault:
CC @tomkralidis

efucile:
@rouault excellent that you get the tables from the WMO GRIB2 repo. I have a suggestion for an improvement. We release new tables every six months. To keep track of which table you are merging it is better to get them from the tagged release rather than master and make the version number accessible to users. Latest release is here https://github.com/wmo-im/GRIB2/releases/tag/v28.1

rouault:
> o keep track of which table you are merging it is better to get them from the tagged release rather than master and make the version number accessible to users

done in follow-up commit

PR #5059, opened 2022-01-04
configure/CMake: add selection of qhull_r vs qhullstatic_r

rouault:
CC @SpaceIm

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -4706,7 +4706,6 @@ elif test "$with_qhull" = "yes" -o "$with_qhull" = "" ; then
         AC_MSG_ERROR([--with-qhull requested, but library not found!])
       else
         QHULL_SETTING=internal
-        QHULL_INCLUDE_SUBDIR_IS_LIBQHULL=1
       fi
     fi
   fi
--- a/configure.ac
+++ b/configure.ac
@@ -4677,21 +4677,37 @@ if test "$with_qhull" = "no" ; then
 
 elif test "$with_qhull" = "yes" -o "$with_qhull" = "" ; then
 
-  AC_CHECK_HEADERS([libqhull_r/libqhull_r.h])
-  if test "$ac_cv_header_libqhull_r_libqhull_r_h" = "yes"; then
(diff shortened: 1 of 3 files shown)


SpaceIm:
Regarding the CMake part, do you try to find Qhull config file first? If not, it would be a more reliable approach.

Basically something like this:

- include a module file like `external-qhull.cmake` (not FindQHULL to avoid conflicts with config files).
- in this module file, find first config file (prefer over pkg-config because it doesn't require pkg-config installed), then pkg-config files, then fallback to manual discovery of libs (fragile approach, this is why it should be the last attempt).

rouault:
> Regarding the CMake part, do you try to find QHull config file first?

you mean the Qhull cmake config file ? No, I just see there's one. The current state of things is that it uses pkg-config first, and then fallbacks to manual discovery.

Your suggestions sound reasonable, but I'm not sure I want to go myself in that level of complexity (none of the other libraries use such a complex approach). I'll let you follow up with a complementary PR if you are interested in.

configure.ac
```diff
@@ -4677,21 +4677,37 @@ if test "$with_qhull" = "no" ; then
...
+
+  PKG_PROG_PKG_CONFIG([0.21])
+  PKG_CHECK_MODULES([QHULL],[$with_qhull_pkgname], [HAVE_QHULL=yes], [HAVE_QHULL=no])
+
+  dnl First try with pkg-config
+  if test "$HAVE_QHULL" = "yes" ; then
+    QHULL_SETTING=external
+    CFLAGS="$CFLAGS $QHULL_CFLAGS"
+    LIBS="$LIBS $QHULL_LIBS"
+  else
```
SpaceIm, line 4694: `QHULL_INCLUDE_SUBDIR_IS_LIBQHULL` check logic is missing if non-reentrant.

rouault: > `QHULL_INCLUDE_SUBDIR_IS_LIBQHULL` check logic is missing if non-reentrant.

non-reentrant qhull is no longer supported in master. I've just removed the remaining trace of QHULL_INCLUDE_SUBDIR_IS_LIBQHULL in configure.ac which is no longer needed

Issue #5060, opened 2022-01-04
FlatGeobuf; CRS WKT risk of wrong encoding


bjornharrtell:
I believe https://github.com/OSGeo/gdal/blob/7b5075b67201a1f6945b9e80d83364addf390184/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp#L353 can result in a string with some system or source specific encoding.

(All strings in FlatGeobuf are mandated to be UTF-8 and other encodings can cause issues)

rouault:
if the CRS wasn't instantiated with something like proj_create(, ... "EPSG:XXXX"), but from a user-provided string, then PROJ will return in whatever encoding the user provided the string. Perhaps the issue is due to QGIS saving a .prj file with a non-UTF8 encoding ?
You could possibly use CPLIsUTF8() to check the string is UTF8 and then remove any char > 127 (we might have a generic CPLForceUTF8() function to do that)

rouault:
Fixed per https://github.com/OSGeo/gdal/pull/5062

PR #5062, opened 2022-01-04
FlatGeobuf: If CRS WKT detected to be non UTF-8 force it

bjornharrtell:
ref https://github.com/OSGeo/gdal/issues/5060


bjornharrtell:
@rouault I think this is ready to be squash merged.

bjornharrtell:
And just when I wrote that I realized I had the conditional the wrong way... fixed.

PR #5063, opened 2022-01-04
HDF5: support libhdf5 1.13.0 (fixes #5061)

### LINKED ISSUE #5061: Building v3.4.1 against HDF5 v1.13.0 fails
Building GDAL v3.4.1 from source against HDF5 v1.13.0 (also built from source) fails because `const H5FD_class_t` contains additional members that are not initialized in `frmts/hdf5/hdf5vfl.h`.

I was trying this on Windows with MinGW-w64 11.2.0 (UCRT), but I don't think this depends on the platform.

Adding the missing fields like this allawed the build to continue:
```
patch -ulbf frmts/hdf5/hdf5vfl.h << EOF
@@ -73,2 +73,3 @@
 static const H5FD_class_t HDF5_vsil_g = {
+    H5FD_VFD_DEFAULT,           /* value */
     "vsil",                     /* name */
@@ -106,2 +107,4 @@
     nullptr,                    /* unlock */
+    nullptr,                    /* del */
+    nullptr,                    /* ctl */
     H5FD_FLMAP_DICHOTOMY        /* fl_map */
EOF
```
However I'm not sure these are the right values to use, and they should be guarded by `#ifdef`/`#endif` to ensure compatibility with older versions of HDF5.

rouault:


rouault:
Waiting a bit for feedback from help@hdfgroup.org to get an official H5FD_class_value_t for our VFD plugin

rouault:
> Waiting a bit for feedback from [help@hdfgroup.org](mailto:help@hdfgroup.org) to get an official H5FD_class_value_t for our VFD plugin

they assigned value 513 to us

PR #5074, opened 2022-01-08
Doc: Use EPSG:4326 with PG geography

jratike80:


rouault:
Putting that on hold since resolving https://github.com/OSGeo/gdal/issues/5075 would cause it to be amended

rouault:
superseded per https://github.com/OSGeo/gdal/pull/5079

PR #5078, opened 2022-01-09
Python bindings: replace unsafe use of tempfile.mktemp() by mkstemp()…

rouault:
… (fixes https://huntr.dev/bounties/59e93871-266a-40be-9aad-4576d3dc48f2/)

CC @idanmiara I'm not sure if all those code paths are tested

### DIFF
--- a/swig/python/gdal-utils/osgeo_utils/auxiliary/color_palette.py
+++ b/swig/python/gdal-utils/osgeo_utils/auxiliary/color_palette.py
@@ -257,12 +257,15 @@ class ColorPalette:
         return s
 
     def write_file(self, color_filename: Optional[PathLikeOrStr] = None, with_ndv: bool = True):
+        tmp_fd = None
         if color_filename is None:
-            color_filename = tempfile.mktemp(suffix='.txt')
+            tmp_fd, color_filename = tempfile.mkstemp(suffix='.txt')
         os.makedirs(os.path.dirname(color_filename), exist_ok=True)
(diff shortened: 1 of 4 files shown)


idanmiara:
> … (fixes https://huntr.dev/bounties/59e93871-266a-40be-9aad-4576d3dc48f2/)
> 
> CC @idanmiara I'm not sure if all those code paths are tested

The report in the link is not public.

I'm looking now at the tests.

rouault:
> The report in the link is not public.

they just point to the fact that tempfile.mktemp() is deprecated and potentially insecure, and mkstemp() should be used instead. The security risk in the GDAL context is probably low in practice, but better use the recommended approach.

idanmiara:
> > The report in the link is not public.
> 
> they just point to the fact that tempfile.mktemp() is deprecated and potentially insecure, and mkstemp() should be used instead. The security risk in the GDAL context is probably low in practice, but better use the recommended approach.

Thanks for this clarification.

idanmiara:
3/4 of the effected functions are tested.
`extent_util.make_temp_vrt` is not tested (it is used in gdal_calc if `extent not in [Extent.IGNORE, Extent.FAIL]`).
I shall add tests for this functionality in another commit. I have some previous commitments to fulfill first.

rouault:
@idanmiara is it good for merging or do you want to have a deeper look ?

idanmiara:
Yes, went over it. Looks good to me. Thanks!

PR #5092, opened 2022-01-11
CMake: add support for external LERC library (fixes #5088)

### LINKED ISSUE #5088: CMake: add support for external liblerc

rouault:


rouault:
@lucianpls any plan to modify the MRF driver to be able to use external liblerc V2 ? Currently LERC_band.cpp uses the Lerc2.h header which is not available in the headers installed by external liblerc

PR #5094, opened 2022-01-11
CMake: Update CheckDependentLibraries.cmake to fix zlib name typo

lucianpls:
## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:


rouault:
thanks

lucianpls:
@rouault 
The test can't fail since the variable gets set one line above

rouault:
> The test can't fail since the variable gets set one line above

not necessarily as this is a CACHE variable not FORCEd. So if you do "cmake .. -DGDAL_USE_ZLIB=OFF -DGDAL_USE_ZLIB_INTERNAL=OFF", the error triggers (I just verified it)

PR #5112, opened 2022-01-14
CMake: add support for libhdfs (for /vsihdfs)

rouault:
CC @jamesmcclain in case you've still interest in this.

I've managed to make the /vsihdfs/ build against hadoop-3.3.1 (https://dlcdn.apache.org/hadoop/common/hadoop-3.3.1/hadoop-3.3.1.tar.gz) on Ubuntu 20.04. It crashes at runtime (a bit unfortunate, but the robustness issue is likely outside of our control in libhdfs itself) but I suspect this is just a usage issue from myself, likely a wrong CLASSPATH
```
$ CLASSPATH=/home/even/hadoop-3.3.1/share/hadoop/common/lib:/home/even/hadoop-3.3.1/share/hadoop/hdfs:/home/even/hadoop-3.3.1/share/hadoop/hdfs/lib gdalinfo /vsihdfs/file:autotest/gcore/data/byte.tif
could not find method getRootCauseMessage from class (null) with signature (Ljava/lang/Throwable;)Ljava/lang/String;
could not find method getStackTrace from class (null) with signature (Ljava/lang/Throwable;)Ljava/lang/String;
FileSystem: loadFileSystems failed error:
(unable to get root cause for java.lang.NoClassDefFoundError)
(unable to get stack trace for java.lang.NoClassDefFoundError)
getJNIEnv: getGlobalJNIEnv failed
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x0000000000000000, pid=426256, tid=0x00007fa5f65f1f40
#
# JRE version: OpenJDK Runtime Environment (8.0_312-b07) (build 1.8.0_312-8u312-b07-0ubuntu1~20.04-b07)
# Java VM: OpenJDK 64-Bit Server VM (25.312-b07 mixed mode linux-amd64 compressed oops)
# Problematic frame:
# C  0x0000000000000000
#
# Failed to write core dump. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
#
# An error report file with more information is saved as:
# /home/even/gdal/gdal/build_cmake/hs_err_pid426256.log
#
# If you would like to submit a bug report, please visit:
#   http://bugreport.java.com/bugreport/crash.jsp
# The crash happened outside the Java Virtual Machine in native code.
# See problematic frame for where to report the bug.
#
Abandon (core dumped)
```


jamesmcclain:
@rouault Thanks notifying me about this.  Yes, this code is still of interest to me -- please let me know if I can be of any service.

rouault:
> Thanks notifying me about this. Yes, this code is still of interest to me -- please let me know if I can be of any service.

if you can confirm that you manage to make /vsihdfs work with CMake in master, that would be welcome. Perhaps with documentation enhancements to explain what is required to make runtime work as I suspect the issue I hit above wasn't specific to the CMake build (I haven't tried with autoconf though)

jamesmcclain:
> > Thanks notifying me about this. Yes, this code is still of interest to me -- please let me know if I can be of any service.
> 
> if you can confirm that you manage to make /vsihdfs work with CMake in master, that would be welcome. Perhaps with documentation enhancements to explain what is required to make runtime work as I suspect the issue I hit above wasn't specific to the CMake build (I haven't tried with autoconf though)

@rouault  Hello, sorry for the long delay, I have been a bit busy.  I made a PR https://github.com/OSGeo/gdal/pull/5254 to address some issues that I encountered when trying to use `cmake` to build this code.  Thank you.

PR #3449, opened 2021-02-08
JP2Grok : add new driver for Grok JPEG 2000  library

boxerab:
[Grok](https://github.com/GrokImageCompression/grok) is an open source JPEG 2000 library, licensed under the AGPL. It was forked from OpenJPEG in 2016. 

## Features that may be of interest to the geospatial community

1. support for reading TLM and PLT markers, for fast random access into large tiled or single-tiled images
1. support for memory mapped files, which helps speed up random access decode for large images 
1. internal tile cache that allows efficient reading of multiple tiles from single Grok codec object
1. support for new Part 15 of the JPEG 2000 standard (HTJ2K) which promises up to **10x** speedup
of encode and decode 

## Performance

1. Speed: Currently ~60% as fast as Kakadu test binaries, in a wide variety of different workflows. 
1. Memory: Memory usage is much higher than Kakadu, but there is an effort underway this year to significantly decrease memory usage.

## Distribution

Grok is available [in Debian](https://tracker.debian.org/pkg/libgrokj2k) and Homebrew

## Testing

1. extensive `ctest`-based test suite
1. part of [oss-fuzz](https://bugs.chromium.org/p/oss-fuzz/issues/list?sort=-opened&can=1&q=proj:grok) project

## Supported Platforms

1. Linux, Windows, OSX
1. x86_64, arm64

#### Compress

Test [image](https://github.com/GrokImageCompression/grok-test-data/blob/master/input/nonregression/Bretagne2.ppm) lossless- compressed with many small tiles.

Here are timings and memory usage for Kakadu, OpenJPEG and Grok on my 24 core / 48 thread Threadripper 
running Linux kernel 5.8. Both OpenJPEG and Grok are built in Release mode.

tl;dr: Kakadu is 2X faster than Grok, which is 5X faster than OpenJPEG.

```
Command being timed: "kdu_compress -i /home/aaron/src/grok-test-data/input/nonregression/Bretagne2.ppm -o /home/aaron/temp/Bretagne2_1.j2k Stiles={127,127} Creversible=yes"
	User time (seconds): 2.34
	System time (seconds): 0.04
	Percent of CPU this job got: 1377%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:00.17
	Average shared text size (kbytes): 0
	Average unshared data size (kbytes): 0
	Average stack size (kbytes): 0
	Average total size (kbytes): 0
	Maximum resident set size (kbytes): 34608
	Average resident set size (kbytes): 0
[...]

PR #5117, opened 2022-01-15
JP2Grok : add new driver for Grok JPEG 2000 library - Version 2

boxerab:
Re-opening this PR https://github.com/OSGeo/gdal/pull/3449 due to renewed interest in the driver, and also because
latest library release is now faster than Kakadu for at least one common geospatial workflow : decompress large multi-tiled 12 bit JP2. I have pasted in the overview from the old PR:

///////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////

[Grok](https://github.com/GrokImageCompression/grok) is an open source JPEG 2000 library, licensed under the AGPL. It was forked from OpenJPEG in 2016. 

## Features that may be of interest to the geospatial community

1. support for reading TLM and PLT markers, for fast random access into large tiled or single-tiled images
1. support for memory mapped files, which helps speed up random access decode for large images 
1. internal tile cache that allows efficient reading of multiple tiles from single Grok codec object
1. support for new Part 15 of the JPEG 2000 standard (HTJ2K) which promises up to 10x speedup
of encode and decode 

### Who asked for the driver and why?

Grok library developer wrote the driver to allow gdal users to make us of this library - one of only two actively developed open source JPEG 2000 toolkits.

### Read / Write / ReadWrite ?

Read and Write

### Raster / Vector / Raster&Vector?

Raster

### Driver name info

  * Short name : JP2Grok

  * Long name : JPEG-2000 driver based on Grok library

  * Description: JP2Grok

### Should this be compiled in by default, optional build, or plugin?

plugin

### Does this driver require other drivers be built?

No

### Is there a standard(s) that the driver is working to implement?

Part 1 and Part 15 of ISO/IEC 15444

### How feature complete will the driver be?

Currently it matches features of existing JP2OpenJPEG driver. 

### Support for the following features :

Should be answer with something like on of these: None, value, [my, list, of, values], Yes, No

  * DMD_MIMETYPE: 
  
image/jp2

  * DMD_EXTENSIONS:  

jp2 j2k

  * DMD_CONNECTION_PREFIX

  * GDAL_DMD_OPENOPTIONLIST
```
"<OpenOptionList>"
"   <Option name='1BIT_ALPHA_PROMOTION' type='boolean' description='Whether a 1-bit alpha channel should be promoted to 8-bit' default='YES'/>"
"   <Option name='OPEN_REMOTE_GML' type='boolean' description='Whether to load remote vector layers referenced by a link in a GMLJP2 v2 box' default='NO'/>"
"   <Option name='GEOREF_SOURCES' type='string' description='Comma separated list made with values INTERNAL/GMLJP2/GEOJP2/WORLDFILE/PAM/NONE that describe the priority order for georeferencing' default='PAM,GEOJP2,GMLJP2,WORLDFILE'/>"
"   <Option name='USE_TILE_AS_BLOCK' type='boolean' description='Whether to always use the JPEG-2000 block size as the GDAL block size' default='NO'/>"
"</OpenOptionList>" );
```

  * GDAL_DMD_SUBDATASETS

  * GDAL_DCAP_OPEN : Yes

  * DCAP_CREATE : Yes

  * GDAL_DCAP_VIRTUALIO: Yes

### What external libraries does this driver depend on?

[Grok JPEG 2000 library](https://github.com/GrokImageCompression/grok)

### Are there version requirements for the libraries?

Yes, it requires at least version 9.6.0 of Grok JPEG 2000 library

### What licenses do the libraries have?

Freely available under AGPL license

### Does the driver require a binary SDK?

No

### Any external services required for the driver?

No

### Any external service providers required for the driver?

No

### Testing plan

This driver has an autotest script in `autotest/gdriver` folder

### Maintenance plan?

1. improve block decoding performance : currently multiple threads are launched and each thread parses the JPEG 2000
file and extracts a block. As Grok has an internal tile cache, and tile decoding is multi threaded, performance will improve with 
a single decode of multiple tiles.
1. Add support for Part 15 read/write

### Who / what org is responsible for upkeep on this new driver?

Grok library maintainer will be responsible for upkeep

### Where to report bugs?

gdal gitub issue tracker

### Draft help text

see `gdal/doc/source/drivers/raster/jp2grok.rst` in PR for details

### What are the conditions for removing the driver?

Driver may be removed if Grok library is no longer maintained - no more bug fixes and security updates

### Does the driver already have a draft implementation?

Yes

### Who implemented it? Or who will implement this driver?  Does an implementer(s) need to be found?

Grok library maintainer has already implemented the driver

### Is there funding needed?

No

### Who will review the code / PR?

TBD


boxerab:
cc @pl-nfwarmerdam @warmerdam @jratike80 @tbonfort @idanmiara @normanb @rouault

EmDevys:
From some tests, it seems the Grok branch of OpenJPEG provides superior level of performance vs OpenJPEG. 
For Geospatial imagery, grok is presently agnostic of GeoTIFF (Geo TIFF tags) and GMLJP2 or GeoJP2, and therefore there is an interest to handle the Geo-enabling of Grok, presumably via GDAL.
I am not able to interfere in the debate on Grok/OpenJPEG, both being of interest to me. By the way OpenJPEG also includes a JPIP driver, and OpenJPEG was up to now the best opensource J2K driver. From a user perspective, and as it seems Grok is based on a fork from OpenJPEG, it would seem of interest for the JPEG 2000 users point of view to share improvements (I guess this is the way opensource works) and synchronize evolution for the best.
In all cases, I would recommend 
1-to provide a GDAL encapsulation of Grok, 
2- and if this is possible to synchronize / harmonize the evolution of the Grok and OpenJPEG (for the best of opensource usage of JPEG 2000).

boxerab:
@EmDevys thanks for your comments. I agree - it would be good to give user more choice in gdal for JPEG 2000.
As the Jasper library is scheduled for removal in version 3.5, this new driver could take its place. 
As for harmonizing the two projects, this is quite unlikely after 6 years of divergent development - the two are written in different languages and have different licenses. Sharing ideas is certainly possible, on the other hand.

rouault:
The project is discussing a RFC about policies for substantial code additions: https://github.com/OSGeo/gdal/pull/5128

frmts/grok/grokdataset.cpp
```diff
@@ -0,0 +1,4173 @@
...
+                return nullptr;
+            }
+            nRead += nToRead;
+        }
+
+        VSIFCloseL(fpSrc);
+    }
+    else
+    {
+        JP2GrokFile __sJP2GrokFile;
```
rouault: https://en.cppreference.com/w/cpp/language/identifiers: the identifiers with a double underscore anywhere are reserved;

frmts/grok/grokdataset.cpp
```diff
@@ -0,0 +1,4173 @@
+/******************************************************************************
```
rouault: Apart from Grok namespacing instead of openjpeg, a side-to-side diff with frmts/openjpeg/openjpegdataset.cpp shows that this file is at > 90% identical to it. We are speaking of 4173 lines for grokdataset.cpp and 4334 lines for openjpegdataset.cpp. To the best of my knowledge, this is a unique situation in the project. I'm not sure what to do in that situation:
- sigh, sigh and sigh again and accept that sad situation
- ask Grok to be relicensed into the original BSD 2 license and rip off the openjpeg driver
- use C++ templating to have one common implementation, or C++ inheritance with virtual methods
- looser variation of the above: extract parts that are independent from the underlying library so they can be reused by both. It looks like JP2 box formatting could be a candidate for that.

boxerab:
@rouault thanks for the review - it will take me a few days to respond. But I definitely agree that we should unify the openjpeg and grok drivers to avoid code duplication.

frmts/grok/grokdataset.cpp
boxerab: Latest version now has common code refactored into base `JP2Dataset` class. This moves ~25% of the code into the base class. As the remaining methods have codec-specific code interleaved with more generic code, I do not think it is worth the effort at the moment to split it any further. Of course this can be done in the future if it makes sense. Let me know what you think.

rouault:
This would need to be rebased on top of latest master

misev:
It's a pity this work didn't get merged. I'm sure many will benefit from support for a faster open-source JPEG 2000 driver in GDAL. Is there still any plan to finalize it?

PR #5128, opened 2022-01-18
Add RFC 85: Policy regarding substantial code additions

rouault:
Rendered document at https://github.com/rouault/gdal/blob/rfc_85/doc/source/development/rfc/rfc85_policy_code_additions.rst

### DIFF
--- a/doc/source/development/rfc/rfc85_policy_code_additions.rst
+++ b/doc/source/development/rfc/rfc85_policy_code_additions.rst
@@ -8,7 +8,7 @@ RFC 85: Policy regarding substantial code additions
 Authors: Howard Butler, Even Rouault
 ======== ==================================================
 Started: 2022-01-17
-Status:  Development
+Status:  Adopted
 ======== ==================================================
 
 Summary
@@ -65,12 +65,12 @@ Policy
 - Contributors of significant code additions are expected to participate in the
   day-to-day life of the GDAL project, and need to monitor closely the communication
   channels of the project: issue tracker, mailing lists, etc.
-  
+
 - Maintainers are expected to support their contributions by triaging bug reports,
   reviewing related pull requests and RFCs, making functional enhancements, testing
   releases, and improving documentation, tests, and infrastructure.
 
-- In addition, maintainers are expected to respond in a timely manner to wider 
+- In addition, maintainers are expected to respond in a timely manner to wider
   project changes (CI, build scripts, upgrade of dependencies, build tools,
   documentation, etc.) as it pertains to their contributions.
 
@@ -86,5 +86,4 @@ Policy
 Voting history
 --------------
 
-TBD
-
++1 from PSC members MateuszL, KurtS, TamasS, SeanG, HowardB, JukkaR and EvenR.
--- a/doc/source/development/rfc/rfc85_policy_code_additions.rst
+++ b/doc/source/development/rfc/rfc85_policy_code_additions.rst
@@ -66,7 +66,7 @@ Policy
   day-to-day life of the GDAL project, and need to monitor closely the communication
   channels of the project: issue tracker, mailing lists, etc.
   
-- Maintainers are expected to supporting their contributions by triaging bug reports,
+- Maintainers are expected to support their contributions by triaging bug reports,
   reviewing related pull requests and RFCs, making functional enhancements, testing
   releases, and improving documentation, tests, and infrastructure.
 
--- a/doc/source/development/rfc/rfc85_policy_code_additions.rst
+++ b/doc/source/development/rfc/rfc85_policy_code_additions.rst
@@ -25,7 +25,7 @@ The GDAL project has historically been quite open to new code additions, includi
 drivers that rely on proprietary licensed and/or closed-source SDKs (called
 "binary SDK" in the rest of this document). This approach is part
 of its strengths, but also comes at a price, with the software becoming larger
-and larger, with sometimes contributions being abandoned by their authors without
+and larger, and with contributions sometimes being abandoned by their authors without
 new maintainers.
 
 Being a free & open-source project, the GDAL project will apply stronger scrutiny to
--- a/doc/source/development/rfc/rfc85_policy_code_additions.rst
+++ b/doc/source/development/rfc/rfc85_policy_code_additions.rst
@@ -1,0 +1,85 @@
+.. _rfc-85:
+
+===========================================================
+RFC 85: Policy regarding substantial code additions
+===========================================================
+
+======== ==================================================
+Authors: Howard Butler, Even Rouault
+======== ==================================================
+Started: 2022-01-17
+Status:  Development
+======== ==================================================
+
+Summary
+-------
+
+This document describes the policies that the GDAL project will apply to assess
+substantial code additions, typically new drivers, in particular coming for new
+contributors to the project.
+
+Motivation
+----------
+
+The GDAL project has historically been quite open to new code additions, including
+drivers that rely on proprietary licensed and/or closed-source SDKs (called
+"binary SDK" in the rest of this document). This approach is part
+of its strengths, but also comes at a price, with the software being larger
+and larger, with sometimes contributions being abandoned by their authors without
+new maintainers.
+
+Being a free & open-source project, the GDAL project will apply stronger scrutiny to
+code that require a binary SDK and encourage contributors to submit code additions
+that do not depend on such binary SDK or license such SDK under free & open-source
+license terms.
+
+Policy
+------
+
+- Drivers require a designated responsible contact, tracked in
+  https://github.com/OSGeo/gdal/wiki/Maintainers-per-sub-system
+
+- Contributions should follow other RFCs describing development rules, and come
+  with test scripts and sufficient documentation, covering usage and build instructions.
+
+- If the driver require a binary SDK not downloadable without cost, or that requires
+  a complicated registration process, the GDAL team is unlikely to support
+  driver inclusion.
+
+- If the binary SDK is no longer current, or modernized to work with latest
+  compilers and GDAL, the driver can be dropped. This rule also applies to open-source
+  dependencies that are no longer maintained or are superseded by other alternatives.
+
+- If the driver has unaddressed bugs, is breaking continuous integration (CI),
+  has caused CI bits to be disabled, or has not caught up to API modifications
+  requiring updates (extremely rare), the driver will be dropped from build
+  scripts, and thus will not be built without significant user intervention.
+
+- If critical/blocking issues reported in a driver are not addressed within a
+  2 month time-frame, it can be dropped from the tree entirely by a designated
+  GDAL maintainer for all releases going forward.
+
+- Contributors of significant code additions are expected to participate to the
+  day-to-day life of the project, and need to monitor closely the communication
+  channels of the project: issue tracker, mailing lists, etc. They are expected at
+  a minimum to respond in a timely manner to changes that affect the whole project:
+  CI, build scripts,upgrade of dependencies & build tools, documentation etc.
+  They are also expected to help sharing the burden of maintenance and participate
+  to the global health of the project by doing pull request reviews, tackling general
+  bug reports, functional enhancements, documentation and infrastructure enhancements,
+  etc.
+
+- New contributions may require a significant review effort from a GDAL committer (ie
+  someone with direct modification rights to the source repository). While the
+  project has now access to funded maintainers, it must be understood that they might not be
+  immediately available to do reviews of significant code additions. Contributors
+  may contract GDAL committers to have such task done within a more predictable timeline.
+
+- The above rules are not exhaustive. The PSC reserves the right to reject a proposed
+  code addition depending on its particular nature and other contextual elements.
+
+Voting history
+--------------
+
+TBD
+
--- a/doc/source/development/rfc/rfc85_policy_code_additions.rst
+++ b/doc/source/development/rfc/rfc85_policy_code_additions.rst
@@ -24,7 +24,7 @@ Motivation
 The GDAL project has historically been quite open to new code additions, including
 drivers that rely on proprietary licensed and/or closed-source SDKs (called
 "binary SDK" in the rest of this document). This approach is part
-of its strengths, but also comes at a price, with the software being larger
+of its strengths, but also comes at a price, with the software becoming larger
 and larger, with sometimes contributions being abandoned by their authors without
 new maintainers.
 
--- a/doc/source/development/rfc/rfc85_policy_code_additions.rst
+++ b/doc/source/development/rfc/rfc85_policy_code_additions.rst
@@ -63,7 +63,7 @@ Policy
   day-to-day life of the project, and need to monitor closely the communication
   channels of the project: issue tracker, mailing lists, etc. They are expected at
   a minimum to respond in a timely manner to changes that affect the whole project:
-  CI, build scripts,upgrade of dependencies & build tools, documentation etc.
+  CI, build scripts, upgrade of dependencies & build tools, documentation etc.
   They are also expected to help sharing the burden of maintenance and participate
   to the global health of the project by doing pull request reviews, tackling general
   bug reports, functional enhancements, documentation and infrastructure enhancements,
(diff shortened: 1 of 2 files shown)


doc/source/development/rfc/rfc85_policy_code_additions.rst
```diff
@@ -0,0 +1,85 @@
...
+
+- Contributors of significant code additions are expected to participate to the
+  day-to-day life of the project, and need to monitor closely the communication
+  channels of the project: issue tracker, mailing lists, etc. They are expected at
+  a minimum to respond in a timely manner to changes that affect the whole project:
+  CI, build scripts,upgrade of dependencies & build tools, documentation etc.
+  They are also expected to help sharing the burden of maintenance and participate
+  to the global health of the project by doing pull request reviews, tackling general
+  bug reports, functional enhancements, documentation and infrastructure enhancements,
+  etc.
```
rcoup, line 70: If someone is a dedicated & diligent maintainer for the WizBang driver, do we really _expect_ them to also become a maintainer of the entire GDAL project?

doc/source/development/rfc/rfc85_policy_code_additions.rst
```diff
@@ -0,0 +1,85 @@
...
+- Drivers require a designated responsible contact, tracked in
+  https://github.com/OSGeo/gdal/wiki/Maintainers-per-sub-system
+
+- Contributions should follow other RFCs describing development rules, and come
+  with test scripts and sufficient documentation, covering usage and build instructions.
+
+- If the driver require a binary SDK not downloadable without cost, or that requires
+  a complicated registration process, the GDAL team is unlikely to support
+  driver inclusion.
+
```
rcoup, line 48:
```suggestion
- GDAL supports multiple operating systems, and new drivers should support as many as
  is practical. At a minimum, drivers should work with the latest and OS-vendor supported
  releases of operating systems the driver is documented to work with.

```

My wording here should cover things like: "my driver/library only works on RHEL7/macOS 10.9/Windows 8" without necessarily requiring everything to work on macOS/Android/Windows/32bit Linux/etc

rcoup (commented):
Should this be targeted more specifically at *drivers*?

Most of the policy items here are driver-specific, and non-driver contributions will likely have a different set of considerations that should be applied.

doc/source/development/rfc/rfc85_policy_code_additions.rst
jratike80: I do not read that so that they are expected to maintain the _entire_ project. Rather to do a little bit extra to keep the GDAL ship that is carrying their load afloat.

rouault:
> Should this be targeted more specifically at drivers?

I don't know if we should restrict to drivers, even if its the main reason for this RFC. This could as well apply to a new utility added, or a new subcomponent relatively independent from others (like GMN). I like that we stay vague and convey a general spirit otherwise some people may start to exploit holes.

doc/source/development/rfc/rfc85_policy_code_additions.rst
rouault: > I do not read that so that they are expected to maintain the entire project. Rather to do a little bit extra to keep the GDAL ship that is carrying their load afloat.

Exactly. The project can't work if people wear blinders and only care about the part that they contributed to without considering the project as a whole

szekerest:
Should we also have a policy what are the conditions of an externally contributed code will be incubated to be maintained by a paid maintainer of GDAL? For example drivers with large community interest should not necessarily die just because the original authors/maintainers are getting out of scope.

doc/source/development/rfc/rfc85_policy_code_additions.rst
```diff
@@ -0,0 +1,88 @@
...
+- Contributors of significant code additions are expected to participate to the
+  day-to-day life of the project, and need to monitor closely the communication
+  channels of the project: issue tracker, mailing lists, etc. They are expected at
+  a minimum to respond in a timely manner to changes that affect the whole project:
+  CI, build scripts, upgrade of dependencies & build tools, documentation etc.
+  They are also expected to help sharing the burden of maintenance and participate
+  to the global health of the project by doing pull request reviews, tackling general
+  bug reports, functional enhancements, documentation and infrastructure enhancements,
+  etc.
```
rcoup, line 73: How about something like this?

```suggestion
- Contributors of significant code additions are expected to participate in the
  day-to-day life of the GDAL project, and need to monitor closely the communication
  channels of the project: issue tracker, mailing lists, etc.
  
- Maintainers are expected to supporting their contributions by triaging bug reports,
  reviewing related pull requests and RFCs, making functional enhancements, testing
  releases, and improving documentation, tests, and infrastructure.

- In addition, maintainers are expected to respond in a timely manner to wider 
  project changes (CI, build scripts, upgrade of dependencies, build tools,
  documentation, etc.) as it pertains to their contributions.
```

doc/source/development/rfc/rfc85_policy_code_additions.rst
rcoup: @jratike80 that's exactly my general expectation of what we would want, but that's not how it read to me. I've added some different wording below as a suggestion.

szekerest:
> A driver can go to a state where there is no longer any known maintainer (if you look at the list of maintainers, you can see that we have a lot of them). This doesn't mean it will be immediately killed, but it will be in obvious candidate for removal if it becomes a pain and GDAL paid maintainers (maintainers by default) don't manage to keep it in life. Feel free to propose a formulation for that if you think it needs to be formalized

I think we should make the decision which drivers are being maintained by the paid maintainer(s) at the moment and that should also be added to the maintainers document. Each of these should be assigned to a specific maintainer or just "GDAL team" as the maintainer. Then the policy could look something like:

"Each driver or code portion that has no maintainer(s) assigned, will be scheduled to be removed in one of the future release of GDAL according to the decision of the PSC."

doc/source/development/rfc/rfc85_policy_code_additions.rst
```diff
@@ -0,0 +1,88 @@
...
+
+- Drivers require a designated responsible contact, tracked in
+  https://github.com/OSGeo/gdal/wiki/Maintainers-per-sub-system
+
+- Contributions should follow other RFCs describing development rules, and come
+  with test scripts and sufficient documentation, covering usage and build instructions.
+
+- If the driver require a binary SDK not downloadable without cost, or that requires
+  a complicated registration process, the GDAL team is unlikely to support
+  driver inclusion.
```
rcoup, line 47: From @gdt on gdal-dev:

```suggestion
- If the driver require a binary SDK not downloadable without cost, or that requires
  a complicated registration process, the GDAL team is unlikely to support
  driver inclusion. In addition, drivers depending on binary SDKs will default to off
  in the build system so that users who build GDAL will not accidentally create
  proprietary software. 
```

Though I think at the moment we do actually automatically build some drivers if we auto-detect their presence?

rcoup: Perhaps related, do we want something like this? 

```diff
+ - Any GDAL committer who requests it is expected to be provided with free access to binary SDKs
+   for the purposes of GDAL development.
```

(Bearing in mind that if I create an OSS driver for say, Kakadu or ODA as a licensee, I don't necessarily have the ability to fulfil this part. But it should be a consideration for the PSC)

doc/source/development/rfc/rfc85_policy_code_additions.rst
rouault: > Maintainers are expected to supporting their contributions

this formulation seems to restrict the scope of their contribution only to their main area of interest. This is the minimum we can expect, but I believe we loose the "participate to the global health of the project" which is an important point. GDAL as a working project is not just the addition of juxtaposed drivers

doc/source/development/rfc/rfc85_policy_code_additions.rst
rouault: I've just responsed to Greg by email : "Sorry , but no. If people have put proprietary SDKs in default findable locations (which generally requires efforts since they generally have non standard layouts), they will be used by default, at least in the new CMake build. That makes things much more consistent. People can still disable them (see https://gdal.org/build_hints.html#cmake-package-dependent-options and https://gdal.org/build_hints.html#selection-of-drivers). People who build GDAL should look at the licensing of all dependencies. It might be well possible to create a GDAL build with only free components with incompatible licenses, who knows. Or maybe people want a build that is globally permissive licensed, and thus they must be careful of not including copyleft components. There was a RFC attempt at addressing this, but this was abandoned as being a too complex issue. "
Anyway build system considerations are out of the scope of this RFC

rouault: > ```diff
> ```diff
> + - Any GDAL committer who requests it is expected to be provided with free access to binary SDKs
> +   for the purposes of GDAL development.
> ```

> (Bearing in mind that if I create an OSS driver for say, Kakadu or ODA as a licensee, I don't necessarily have the ability to fulfil this part. But it should be a consideration for the PSC)
> ```

for the reason you indicate, it is difficult to make it a requirement. It is clear that drivers that depend on SDKs that aren't freely available are quite prone to rot if they don't have a dedicated maintainer.

rouault:
> I think we should make the decision which drivers are being maintained by the paid maintainer(s) at the moment and that should also be added to the maintainers document. Each of these should be assigned to a specific maintainer or just "GDAL team" as the maintainer. Then the policy could look something like:
> "Each driver or code portion that has no maintainer(s) assigned, will be scheduled to be removed in one of the future release of GDAL according to the decision of the PSC."

I believe we're going to struggle to decide which driver without an assigned maintainer would deserve to be assigned to the "GDAL team" or go to the fully abandoned state. This can be a discussion for a later revision of this RFC

doc/source/development/rfc/rfc85_policy_code_additions.rst
rcoup: Sure, but if I came along and said "I've written a GDAL driver in my spare time for LibJPEG3000, I'm happy to maintain it in GDAL" should I be _expected_ to be fixing bugs in my spare time for drivers I have zero interest in or knowledge about? Of course, my help will always be welcome when I have time available, but "no, we won't accept the LibJPEG3000 driver unless you're committing to help maintain all of GDAL" doesn't seem like the message we want here?

Bearing in mind GDAL is now in a position where the wider project funds maintainers for a chunk of undirected maintenance, triage, core development, release-management, and other work. Obviously it hasn't been like that until very recently, and it still doesn't cover everything.

But maybe this RFC is more targeted at companies? In which case, very rough wording but how about:

```
- If the new contribution is commercial, the GDAL project expects the company will
  contribute either employee time, money, or both to the overall project health. This
  means either paying contributor(s) to work on improvements and maintenance
  for the wider project, or becoming a GDAL project sponsor.
```

hobu: > But maybe this RFC is more targeted at companies?

It is targeted at (mostly) companies that wish to extend the reach of their binary SDKs through GDAL's very wide and valuable software distribution platform. There have been multiple cases where organizations have wanted to make a big "contribution" of code that benefits themselves and their communities without reciprocity of maintenance or responsibility to the GDAL community. 

This RFC is about laying down some of the rules of the road so that the PSC has the standing to toss out stuff that doesn't conform to the community's expectations. The expectations this RFC lays out, frankly, are not so challenging either – basically they have to take care of the thing and not simply provide a white elephant to the project that it must now either maintain or disappoint a customer/community by removing. 

The alternative, which is not desired because it is arbitrary and picks favorites, is to gate keep stuff on the way in. I think we want to give organizations a shot, but I also want to back up the PSC with some enforceability when it's clear that the entity's attention has waned or it or the community no longer has the resources to support the thing.

rcoup: > It is targeted at (mostly) companies that wish to extend the reach of their binary SDKs through GDAL's very wide and valuable software distribution platform

Let's spell that out clearly then: "_RFC 85: Policy regarding commercial drivers & binary SDKs_". The project can do whatever we want, we don't have to be "fair" or couch stuff in vague terms so it appears to apply to everyone. Holding companies to a higher standard of expectations than individuals is absolutely fine.

> This RFC is about laying down some of the rules of the road so that the PSC has the standing to toss out stuff that doesn't conform to the community's expectations.

The PSC _always_ has the standing to toss out stuff. If they decide to boot a driver because the company is absent, that's what they decide - there's no recourse/due-process or some appeal to a higher court. "Nobody here has have the enthusiasm to maintain it... bye! :wave:" I'm not saying that there shouldn't be clear/written expectations, but if they only apply to companies let's say so.

rouault: > How about something like this?

Giving up some of my idealism, I'll take it as such. Thanks for helping shape this RFC.

rcoup: I quite liked my "you should be sponsoring the project" expectation from https://github.com/OSGeo/gdal/pull/5128#discussion_r786896901 😝 

(disclosure: work is a commercial GDAL sponsor)

doc/source/development/rfc/rfc85_policy_code_additions.rst
```diff
@@ -0,0 +1,90 @@
...
+substantial code additions, typically new drivers, in particular coming for new
+contributors to the project.
+
+Motivation
+----------
+
+The GDAL project has historically been quite open to new code additions, including
+drivers that rely on proprietary licensed and/or closed-source SDKs (called
+"binary SDK" in the rest of this document). This approach is part
+of its strengths, but also comes at a price, with the software being larger
```
rcoup, line 27:
```suggestion
of its strengths, but also comes at a price, with the software becoming larger
```

doc/source/development/rfc/rfc85_policy_code_additions.rst
```diff
@@ -0,0 +1,90 @@
...
+and larger, with sometimes contributions being abandoned by their authors without
```
rcoup, line 28:
```suggestion
and larger, and with contributions sometimes being abandoned by their authors without
```

doc/source/development/rfc/rfc85_policy_code_additions.rst
```diff
@@ -0,0 +1,90 @@
...
+new maintainers.
+
+Being a free & open-source project, the GDAL project will apply stronger scrutiny to
+code that require a binary SDK and encourage contributors to submit code additions
```
rcoup, line 32:
```suggestion
code that requires a binary SDK and encourage contributors to submit code additions
```

doc/source/development/rfc/rfc85_policy_code_additions.rst
```diff
@@ -0,0 +1,90 @@
...
+
+- If critical/blocking issues reported in a driver are not addressed within a
+  2 month time-frame, it can be dropped from the tree entirely by a designated
+  GDAL maintainer for all releases going forward.
+
+- Contributors of significant code additions are expected to participate in the
+  day-to-day life of the GDAL project, and need to monitor closely the communication
+  channels of the project: issue tracker, mailing lists, etc.
+  
+- Maintainers are expected to supporting their contributions by triaging bug reports,
```
rcoup, line 69:
```suggestion
- Maintainers are expected to support their contributions by triaging bug reports,
```

rcoup (commented):
Some more minor grammar fixes.

doc/source/development/rfc/rfc85_policy_code_additions.rst
rouault: > I quite liked my "you should be sponsoring the project" expectation

I've no opposition to it. @hobu ?

hobu: Something to the effect of this would do well: *RFC 85: PSC policy regarding binary SDKs*

Buying sponsorship is appreciated, but I don't think that has to be the expectation at all. The expectation is those who provided this thing, or their designates, need to answer the phone when issues come up. Otherwise, they have successfully externalized their user and software support costs onto the project. I think GDAL has been too lenient on this behavior in the past, and now that the project is so large, it needs to reevaluate.

> The PSC always has the standing to toss out stuff. 

Of course, but every act of doing so without some kind of policy that provides a rationale before the fact feels arbitrary. One person's junk and all that. The purpose of this RFC is to say, "yes, go ahead, but if you start externalizing costs onto the project, we can drop it without much notice." As you say, always true, but in this case much more explicit.

rouault: > RFC 85: PSC policy regarding binary SDKs

For me, the scope of the RFC goes beyond binary SDKs. They are the typical example of the most problematic situations, but a large/complex driver, without any dependency or with open source dependencies, and whose maintainer would go away could also be a problem if the core team doesn't have the domain expertise

PR #5132, opened 2022-01-19
Bad tag + change in RPC structure

Lunyxis:
Hi, 

Last time I used the wrong tag to get the coeffs (image to ground instead of ground to image).
It has been fix.
But the bad new, it is that Airbus changed the RPC nomenclature, and one tag has disappeared. (Global_RFM). I've added an if/else condition to work around this modification. I hope it is the last modification from Airbus DS Geo. This modification has been validated locally on the new RPC file.


gcore/mdreader/reader_pleiades.cpp
```diff
@@ -422,7 +431,7 @@ char** GDALMDReaderPleiades::LoadRPCXmlFile()
             }
             else {
                  pszValue = CSLFetchNameValue(papszRawRPCList,
-                 CPLSPrintf("ImagetoGround_Values.%s_%d", apszRPCTXT20ValItemsLonLat[i], j));
+                 CPLSPrintf("GroundtoImage_Values.%s_%d", apszRPCTXT20ValItems[i], j));
```
rouault, line 434: Is the change from apszRPCTXT20ValItemsLonLat back to apszRPCTXT20ValItems intended ? If so, the apszRPCTXT20ValItemsLonLat array is now no longer used (compilers complain about that) and should be removed

Lunyxis: Yes it is intended. Thanks for the explanation. It is my first time with C/C++.. shoud i remove the variable i've added in the gdal_mdreader.h? Because they are not used anymore.. but can be useful in the future

rouault: I believe it is better to just remove apszRPCTXT20ValItemsLonLat from gcore/mdreader/reader_pleiades.cpp

Issue #5103, opened 2022-01-13
[DWG] : Block Attributes exposition  should not take visibility into account for data conversion


lchevallier69:
The DWG driver should not filter attributes by visibility, in fact the current visibility in the DWG file should not implies the existence or not of a specific attribute, i think the visibility is only for display, and we loose a lot of informations when we dump a DWG. @jgrocha do you agree with that ? perhaps a new parameter could be set to ignore visibility

## Steps to reproduce the problem.

ogrinfo -al --config DWG_ATTRIBUTES TRUE --config DWG_INLINE_BLOCKS FALSE attribs2.dwg

## Operating system

All

## GDAL version and provenance

Master

lchevallier69:
[attribs2.zip](https://github.com/OSGeo/gdal/files/7861362/attribs2.zip)
here a sample file

jgrocha:
### Easy to implement

It is ok for me to ignore the visibility flag and output all attributes by default.

### Easy to implement

It is ok for me to ignore the visibility flag based on a configuration flag like `DWG_BLOCK_ATTRIBUTES`.

### More difficult to implement

If we want a more detailed information about block attributes, we could output for each attribute the value, the prompt and the visibility. Something like: 
```javascript
"ARM_NUMERO": { "value": "002", "prompt": "Numéro de l'armoire", "visibible": "true"}, "ARM_REF": { ... }, ...
```
Any option is good for me.

The example was taken from your DWG file.
![Capture](https://user-images.githubusercontent.com/163681/149305959-4fb52566-b6c2-4be9-ae15-e1aaf5d53798.PNG)

PR #5135, opened 2022-01-19
[DWG] :  add DWG_ALL_ATTRIBUTES to get all attributes #5103

lchevallier69:
display all attribute without taking visibility into account
## What does this PR do?
add a new config parameter DWG_ALL_ATTRIBUTES to ignore or expose unvisible attributes
default value is set to TRUE, not only visible attributes should be exposed, visibility is only for display
#5103


autotest/ogr/ogr_dwg.py
```diff
@@ -177,4 +177,70 @@ def test_ogr_dwg_4():
...
+        layer = ds.GetLayer( 'entities' )
+        defn = layer.GetLayerDefn()
+
+        assert defn.GetFieldCount() == 30, \
+            ('did not get expected number of fields in defn. got %d'
+                                % defn.GetFieldCount())
+
+        ds = None
+
+def test_ogr_dwg_5():
```
rouault, line 213:
```suggestion
def test_ogr_dwg_6():
```

rouault:
due to the duplicated test_ogr_dwg_5(), that I've committed, the first test_ogr_dwg_5 fails with:
```
>           assert defn.GetFieldCount() == 30, \
                ('did not get expected number of fields in defn. got %d'
                                    % defn.GetFieldCount())
E           AssertionError: did not get expected number of fields in defn. got 32
E           assert 32 == 30
E            +  where 32 = <bound method FeatureDefn.GetFieldCount of <osgeo.ogr.FeatureDefn; proxy of <Swig Object of type 'OGRFeatureDefnShadow *' at 0x7f045c1ea9f0> >>()
E            +    where <bound method FeatureDefn.GetFieldCount of <osgeo.ogr.FeatureDefn; proxy of <Swig Object of type 'OGRFeatureDefnShadow *' at 0x7f045c1ea9f0> >> = <osgeo.ogr.FeatureDefn; proxy of <Swig Object of type 'OGRFeatureDefnShadow *' at 0x7f045c1ea9f0> >.GetFieldCount
```

But as far as I can test_ogr_dwg_5 and test_ogr_dwg_6 do the same thing, but in a reversed order. Some cleanup needed...

lchevallier69:
ok my mistake, test_ogr_dwg_5 may do 2 tests => one with the new parameter FALSE and one with no parameter to chack default value wich is TRUE. This tests contains two different config value. Duplicate code is in fact a merge branch problem form my internal feature branch. I Recheck
Thanks

rouault:
@lchevallier69 any progress on finalizing this PR ?

lchevallier69:
Yep, tests are clean nox

rouault:
/__w/gdal/gdal/doc/source/drivers/vector/dwg.rst:73: WARNING: Title underline too short.
/__w/gdal/gdal/doc/source/drivers/vector/dwg.rst:77: ERROR: Unexpected indentation.

Issue #5131, opened 2022-01-18
Add support for 'rate' and 'Creversible' compression parameter into JP2KAK driver


mdhsl:
It would be useful to add support for the 'rate' compression parameter to the creation option of the JP2KAK driver options. 

Some examples:

- compressed data-rate 1 bit/pixel:
kdu_compress -i image.pgm -o out.j2c Qfactor=85 -rate 1.0)

- 3 embedded bit-rates
kdu_compress -i image.pgm -o out.j2c -rate 1.0,0.5,0.25

- compression with a progressive lossy to lossless code-stream having 4 layers
kdu_compress -i image.pgm -o out.j2c Creversible=yes -rate -,1,0.5,0.25

It would be useful to add support for the 'Creversible' compression parameter to the creation option of the JP2KAK driver options.
It is actually automatically computed depending on the Quality value. This should be able to be set manually

mdhsl:
Update the issue to add CReevrsible support option

rouault:
> Updated the issue to add CReversible support option

already implemented as documented in https://gdal.org/drivers/raster/jp2kak.html : "QUALITY=n: Set the compressed size ratio as a percentage of the size of the uncompressed image.[...]. A value of 100 will result in use of the lossless compression algorithm"

mdhsl:
In case of use of Quality parameter. But, In case of using RATE parameter, we would decide if we want to make it lossless such like kdu_compress parameter:
```sh
kdu_compress -i image.tif -o out.jp2 Creversible=yes -rate -,1,0.5
```

Quality option is kind of simplification of potential KDU option we have. Specifying Rate is more accurate and can be associated to Creversible option.

Does it make sense?

rouault:
> Does it make sense?

yes, but to avoid option proliferation, I imagined you could potentially use both QUALITY=100 (which implies CReversible=yes) and the RATE option if specified should include a lossless layer (which must be the "-" I guess). I don't have any firm opinion on this though. This will up to anyone that will be potentially willing to tackle this (you ?) to decide.

mdhsl:
Yes it is an idea. If so, the documentation has to be clear about how the Reversible option is set to yes and in which case.

For me, but I could be wrong, the reversibility is not linked to the quality factor.

From the Kakadu documentation:

"Irreversible matrix decorrelation transforms are corresponding points from each input component to the transform are treated as a vector which is multiplied by an M × N matrix, with M the number of output components and N the number of input components to the block."

[...]

PR #5138, opened 2022-01-20
JP2KAK: Add support for Creversible & RATE creation option 

mdhsl:
## What does this PR do?

This PR add two creation parameters for the JP2KAK driver:
- RATE
- Creversible

We can specify one or more bit-rates, expressed in terms of the ratio between  the total number of compressed bits (including headers) and the product of the largest horizontal and  vertical image component dimensions.

This is related to issue #5131

## Tasklist

 - [X] Compare compression against kdu_compress command
 - [X] Add documentation
 - [X] Adjust for comments

## Environment

* OS: Linux, kernel 5.15.11-arch2-1
* Compiler: GCC 11.1.0


frmts/jp2kak/jp2kakdataset.cpp
```diff
@@ -2215,25 +2215,33 @@ JP2KAKCreateCopy( const char * pszFilename, GDALDataset *poSrcDS,
         return nullptr;
     }
 
+    const bool bReversible =
+            CSLFetchNameValue(papszOptions,"Creversible") != nullptr
+            ? CPLFetchBool( papszOptions, "Creversible", false )
+            : (dfQuality < 99.5);
```
rouault, line 2221: test in wrong direction
```suggestion
            : (dfQuality >= 99.5);
```

frmts/jp2kak/jp2kakdataset.cpp
```diff
@@ -2243,10 +2251,76 @@ JP2KAKCreateCopy( const char * pszFilename, GDALDataset *poSrcDS,
 
         CPLDebug("JP2KAK", "layer_bytes[] = %g\n",
                  static_cast<double>(layer_bytes[layer_count - 1]));
-    }
-    else
-    {
-        bReversible = true;
+    } else {
+        // Use RATE option.
+        char** dRateStr = CSLTokenizeStringComplex(CSLFetchNameValue(papszOptions,"RATE"),
```
rouault, line 2256: naming convention should be papszRate instead of dRateStr, and you'll need to CSLDestroy() it after use.
Other option
CPLStringList aosRate(CSLTokenizeStringComplex(CSLFetchNameValue(papszOptions,"RATE"),",", FALSE, FALSE ));
and it will get automatically freed.

frmts/jp2kak/jp2kakdataset.cpp
```diff
@@ -2243,10 +2251,76 @@ JP2KAKCreateCopy( const char * pszFilename, GDALDataset *poSrcDS,
...
+                                                   ",", FALSE, FALSE );
+        int rate_count = CSLCount( dRateStr );
+        if(rate_count <= 0) {
+            CPLError(CE_Fatal, CPLE_AppDefined,
```
rouault, line 2260: CE_Fatal kills the process. It should not be used. You should cleanly error out from this situation

frmts/jp2kak/jp2kakdataset.cpp
```diff
@@ -2243,10 +2251,76 @@ JP2KAKCreateCopy( const char * pszFilename, GDALDataset *poSrcDS,
...
+                     "RATE argument must be followed by a string identifying one or more bit-rates, separated by commas.\n");
+        } else if(bLayerOpt && (rate_count != layer_count || layer_count < rate_count) && rate_count > 2) {
+            // compare NUMBER of layers defined by LAYERS or CLayers and the number of layers defined by RATE
+            // and is not 1 or 2
+            //  If Clayers is not used, the number of layers is set to the number of rates specified here.
+            //  If Clayers is used to specify an actual number of quality layers, one of the following must be true:
+            //  1) the number of rates specified here is identical to the specified number of layers;
+            //  or 2) one or two rates are specified using this argument.
+            CPLError(CE_Fatal, CPLE_AppDefined,
```
rouault, line 2269: same regarding CE_Fatal

frmts/jp2kak/jp2kakdataset.cpp
```diff
@@ -2243,10 +2251,76 @@ JP2KAKCreateCopy( const char * pszFilename, GDALDataset *poSrcDS,
...
-    {
-        bReversible = true;
+    } else {
+        // Use RATE option.
+        char** dRateStr = CSLTokenizeStringComplex(CSLFetchNameValue(papszOptions,"RATE"),
+                                                   ",", FALSE, FALSE );
+        int rate_count = CSLCount( dRateStr );
+        if(rate_count <= 0) {
+            CPLError(CE_Fatal, CPLE_AppDefined,
+                     "RATE argument must be followed by a string identifying one or more bit-rates, separated by commas.\n");
```
rouault, line 2261: termination \n not needed
```suggestion
                     "RATE argument must be followed by a string identifying one or more bit-rates, separated by commas.");
```

frmts/jp2kak/jp2kakdataset.cpp
```diff
@@ -2243,10 +2251,76 @@ JP2KAKCreateCopy( const char * pszFilename, GDALDataset *poSrcDS,
...
+            // and is not 1 or 2
+            //  If Clayers is not used, the number of layers is set to the number of rates specified here.
+            //  If Clayers is used to specify an actual number of quality layers, one of the following must be true:
+            //  1) the number of rates specified here is identical to the specified number of layers;
+            //  or 2) one or two rates are specified using this argument.
+            CPLError(CE_Fatal, CPLE_AppDefined,
+                     "The relationship between the number of bit-rates "
+                     "specified by the \"RATE\" argument and the number of quality layers "
+                     "explicitly specified via \"Clayers\" does not conform to the rules "
+                     "supplied in the description of the \"RATE\" argument\n");
```
rouault, line 2273: termination \n not needed
```suggestion
                     "supplied in the description of the \"RATE\" argument");
```

frmts/jp2kak/jp2kakdataset.cpp
```diff
@@ -2869,6 +2942,24 @@ void GDALRegister_JP2KAK()
...
+"           to specify an actual number of quality layers, one of the "
+"           following must be true: 1) the number of rates specified here is "
+"           identical to the specified number of layers; or 2) one or two "
+"           rates are specified using this argument.  When two rates are "
+"           specified, the number of layers must be 2 or more and intervening "
+"           layers will be assigned roughly logarithmically spaced bit-rates. "
+"           When only one rate is specified, an internal heuristic determines "
+"           a lower bound and logarithmically spaces the layer rates over the "
+"           range. The rates have to be in ASC order.'/>"
+"   <Option name='Creversible' type='string'/>"
```
rouault, line 2962:
```suggestion
"   <Option name='Creversible' type='boolean'/>"
```

frmts/jp2kak/jp2kakdataset.cpp
```diff
@@ -2869,6 +2942,24 @@ void GDALRegister_JP2KAK()
 "   <Option name='FLUSH' type='boolean' />"
 "   <Option name='NBITS' type='int' description="
 "'BITS (precision) for sub-byte files (1-7), sub-uint16 (9-15)'/>"
+"   <Option name='RATE' type='string' description='One or more bit-rates, expressed in terms of the ratio between"
```
rouault, line 2945: the description is untypically long here. This longer description should be moved to the Creation Options section of doc/source/drivers/raster/jp2kak.rst

rouault (commented):
A few comments. The tests in autotest/gdrivers/jp2kak.py should be extended to test the new options

frmts/jp2kak/jp2kakdataset.cpp
mdhsl: Fix in b0bc8297008c5f7899a093e32aa0164377d4c183

frmts/jp2kak/jp2kakdataset.cpp
mdhsl: Fix in e134d20603ee19d0cd8af6db4d42345e6b34f866, 91764013b5f0600cd41dc84516e939039684b56d

frmts/jp2kak/jp2kakdataset.cpp
mdhsl: Fix in 1e0a4f0561cc4b01ce861a8f933b799babdfff8b

frmts/jp2kak/jp2kakdataset.cpp
mdhsl: Fix in 1e0a4f0561cc4b01ce861a8f933b799babdfff8b

frmts/jp2kak/jp2kakdataset.cpp
mdhsl: Fix in b9892f387184d438931da1ea1e69899388b78db3

frmts/jp2kak/jp2kakdataset.cpp
mdhsl: Fix in 1e0a4f0561cc4b01ce861a8f933b799babdfff8b

frmts/jp2kak/jp2kakdataset.cpp
mdhsl: Fix in 7ec84bcd0aea12104456695e6ce03386c52c7455

mdhsl:
Thank you for reviewing. I have taken your remarks into account and made the necessary changes.

frmts/jp2kak/jp2kakdataset.cpp
rouault: You have forgotten to commit the changes in doc/source/drivers/raster/jp2kak.rst to describe the 2 new options.

mdhsl: Fix in c2275ddf75382210b2393b457b25ec6fee8f0752

frmts/jp2kak/jp2kakdataset.cpp
```diff
@@ -2215,25 +2215,33 @@ JP2KAKCreateCopy( const char * pszFilename, GDALDataset *poSrcDS,
...
-    if( dfQuality < 99.5 )
-    {
+    // Override Quality parameter if RATE option is used
+    // TODO: highest quality that can
+    //       be achieved with a quality factor of QUALITY and
+    //       also introducing some visual weighting to the rate-distortion
+    //       optimization objective associated with all N quality layers.
+    //       That means: combines both QUALITY and RATE
+    if(CSLFetchNameValue(papszOptions,"RATE") == nullptr) {
+        layer_bytes.resize(layer_count);
```
rouault, line 2236: there should be a if (!bReversible) test protecting lines 2236 to 2252. The current code breaks the tests in autotest/gdrivers/jp2kak.py that use QUALITY=100

frmts/jp2kak/jp2kakdataset.cpp
```diff
@@ -2243,10 +2251,90 @@ JP2KAKCreateCopy( const char * pszFilename, GDALDataset *poSrcDS,
...
+            // E.g RATE=1,3,4,5 LAYERS=3 => not valid
+            if(rate_count > 2 && rate_count != layer_count) {
+                bValid = false;
+            }
+
+            if(!bValid) {
+                CPLError(CE_Failure, CPLE_IllegalArg,
+                         "The relationship between the number of bit-rates "
+                         "specified by the \"RATE\" argument and the number of quality layers "
+                         "explicitly specified via \"Clayers\" does not conform to the rules "
```
rouault, line 2288:
```suggestion
                         "explicitly specified via \"LAYERS\" does not conform to the rules "
```

frmts/jp2kak/jp2kakdataset.cpp
mdhsl: You are right. The logic is a bit different since we will execute this block of code only if the RATE parameter is not passed. 
Thus, we will evaluate via the formula line 2237 the size of each layer.
If the RATE parameter is passed, we calculate the size according to it. 
The problem is that we must not execute this block of code even if bReevrsible && RATE != nullptr. So if we want to use the old logic, we have to test the quality and the existence of the RATE parameter because it is possible to specify Creversible=no with a quality >= 99.5, and even with a RATE parameter defined.

My proposal:

```cpp
 if(CSLFetchNameValue(papszOptions,"RATE") == nullptr && !bReversible) {
```

rouault: > if(CSLFetchNameValue(papszOptions,"RATE") == nullptr && !bReversible) {

I don't think that will work, because if using just -o QUALITY=100 for example, that would go to the else block "use RATE option". The check if (!bReversible) must be inside the then clause of " if(CSLFetchNameValue(papszOptions,"RATE") == nullptr )"

rouault: > In that case, we should execute the second block.

ok, shouldn't the logic be ?
```
const char* pszRate = CSLFetchNameValue(papszOptions,"RATE") ;
if( pszRate != nullptr )
{
    // do what is in the current else clause
}
else if( !bReversible )
{
     // do what is in the current then clause
}
```

It would be good if you could add some testing of the the new options in autotest/gdrivers/jp2kak.py. It might be possible to the dal.GetJPEG2000Structure() to check that the effect is correct . For an example of how to use that method, see jp2openjpeg_test_codeblock() in autotest/gdrivers/jp2openjpeg.py

mdhsl: Yes I think your proposal works fine. I will make the changes.
For the tests, I am not familar with pytest, I tried to run the pytest using the documentation https://github.com/OSGeo/gdal/blob/master/CONTRIBUTING.md  but many tests fail or are skipped.

Using this particular command:

```sh
pytest autotest/gdrivers/jp2kak.py
```

All tests fails because it does not find JP2KAK:

```sh
        gdaltest.jp2kak_drv = gdal.GetDriverByName('JP2KAK')
>       assert gdaltest.jp2kak_drv is not None
E       assert None is not None
E        +  where None = gdaltest.jp2kak_drv

```
Any help would be appreciated

rouault: > All tests fails because it does not find JP2KAK:

I assume ``gdalinfo --format JP2KAK`` works ?
then try ``python -c "from osgeo import gdal; print(gdal.GetDriverByName('JP2KAK'))"``
if that returns None, then it means that either you have not built the Python bindings (cd swig/python && make) and/or your PYTHONPATH doesn't point to them . They are typically in a path like swig/python/build/lib.linux-x86_64-3.8

mdhsl: Yes, that fixes the problem. Thx. I will now work on tests and push changes

mdhsl: For the testing part, I tried to use what you mentioned but I cannot get the interesting part with the current implementation.
Here what I want to test `Kdu-Layer-Info: log_2{Delta-D(squared-error)/Delta-L(bytes)`:

```sh
 [0
 'Field'
 [2
 'name'
 [1
 'COM']]
 [2
 'type'
 [1
 'string']]
 [2
 'size'
 [1
 '12']]
 [1
 'Kakadu-vs8.1']]]
 [0
 'Marker'
 [2
 'name'
 [1
 'COM']]
 [2
 'offset'
 [1
 '248']]
 [2
 'length'
 [1
 '281']]
 [0
 'Field'
 [2
 'name'
 [1
 'Rcom']]
 [2
 'type'
 [1
 'uint16']]
 [2
 'description'
 [1
 'LATIN1']]
 [1
 '1']]
 [0
 'Field'
 [2
 'name'
 [1
 'COM']]
 [2
 'type'
 [1
 'string']]
 [2
 'size'
 [1
 '275']]
 [1
 'Kdu-Layer-Info: log_2{Delta-D(squared-error)/Delta-L(bytes)}
 L(bytes)\n  -6.9
  4.8e+02\n  -8.1
  5.0e+02\n  -9.3
  5.3e+02\n -10.5
  5.7e+02\n -11.6
  6.2e+02\n -12.8
  6.7e+02\n -14.0
  7.5e+02\n -15.1
  9.3e+02\n -16.3
  1.2e+03\n -17.5
  1.6e+03\n -18.7
  2.2e+03\n -21.3
  4.6e+03\n']]]
```

If I have correctly understand how it works, I have to use that structure:

```py
get_element_val(find_element_with_name(NODE1, "Field", "COM"))
```
But I am getting `Kakadu-vs8.1`, and thats normal. Either I am wrong on how to use the parser, or we have to evolve the parser to iterate on the different node instead of getting the first one.

mdhsl: Fix condition logic in e475326f4201df20765df66943cc250a50ab00f3

mdhsl: Add tests in e8e1dc25cf54aad9074e86cc89177e17d7052f94.
Note, the GDalStructure does not return a approximate value of the size of each layer. 
For this reason, I preferred to test the string containing these approximate values. I didn't find a way to get them otherwise because these values are passed directly to the KDU compressor

autotest/gdrivers/jp2kak.py
```diff
@@ -506,3 +507,294 @@ def test_jp2kak_image_origin_not_zero():
...
+
+def test_jp2openjpeg_test_multi_rate():
+    src_ds = gdal.Open('../gcore/data/stefan_full_rgba.tif')
+    gdaltest.jp2kak_drv.CreateCopy('/vsimem/jp2kak_22.jp2', src_ds, options=['RATE=1.5,2,3,4', 'LAYERS=4'])
+    node = gdal.GetJPEG2000Structure('/vsimem/jp2kak_22.jp2', ['ALL=YES'])
+
+    arr = []
+    find_elements_with_name(node, "Field", "COM", arr)
+
+    # Approximation of [4556, 6075, 9112, 12150] = floor(dfPixelsTotal * currentBitRate * 0.125F)
```
rouault, line 522: woo, thanks for having tried that! I didn't image something as evolved as what you tried . Unfortunately all those tests based on the content of the COM marker are too fragile. They break with the Kakadu version I've at hand. I would just suggest getting the COD.SGcod_NumLayers layers and checking it. That's not very in-depth, but better than nothing.

frmts/jp2kak/jp2kakdataset.cpp
mdhsl: Fix in dccef0b7037c363fe253bbdb99186f2e4b5c8e11.

Some words, I restricted the tests to test only the size of the layers calculated according to the parameter passed in RATE. Indeed, depending on the way the RATE parameter is used, the way the layers are calculated differs. For example, if we have 4 LAYERS and we pass only RATE=1.5, the first 3 layers will be 0 and the last one will be calculated by floor(dfPixelsTotal * currentBitRate * 0.125F).

The layers with a value of 0 are automatically calculated by KDU afterwards. So what is testable are the layers calculated by the formula on the GDAL driver side.

So, I restricted the tests to compare only this value. I tested with Kakadu 7.10 and Kakadu 8.1 and the tests pass. I didn't test with kakadu 6.x but I think it's better to focus on the most recent versions but it probably works very well with version 6.

autotest/gdrivers/jp2kak.py
```diff
@@ -36,7 +36,7 @@
 import gdaltest
 import pytest
 
-pytestmark = pytest.mark.require_driver('JP2KAK')
+# pytestmark = pytest.mark.require_driver('JP2KAK')
```
rouault, line 39:
```suggestion
pytestmark = pytest.mark.require_driver('JP2KAK')
```

PR #5147, opened 2022-01-21
SAGA: allow setting nodata

johanvdw:
Quite a straighforward change.

I tested locally using
```
import rasterio

rst = '/home/johan/git/saga-gis/saga-gis/src/accessories/python/test_data/test.sdat'
rst_out = '/tmp/test.sdat'

with rasterio.open(rst) as src:
    arr = src.read()[0]
    profile = src.profile

profile["nodata"] = 56
print(profile)
arr = np.where(arr == 256, 0, arr).astype("int16")

with rasterio.open(rst_out, "w", **profile) as dst:
    dst.write(arr, 1)
```
and asserted that /tmp/test.sgrd contains 

NODATA_VALUE	= 56.000000

I tried adding a test to the gdal test suite, but I'm missing how to save the file correctly:

 ```
##############################################################################
 # Test setnodata
 def test_saga_9():
 
     gdal_type = gdal.GDT_Float64
 
     ds = gdal.GetDriverByName('SAGA').Create('tmp/test9.sdat', 2, 2, 1, gdal_type)
     ds = None
 
     ds = gdal.Open('tmp/test9.sdat', gdal.GA_Update)
 
     ds.GetRasterBand(1).SetNoDataValue(56)
     nd = ds.GetRasterBand(1).GetNoDataValue()
     assert nd == 56 # ok
 
     ds = None
 
     with open('tmp/test9.sgrd', 'r') as f:
         print(f.read()) # will still contain -99999 as nodata value
         assert 0==1
     try:
         os.remove('tmp/test9.sgrd')
         os.remove('tmp/test9.sdat')
     except OSError:
         pass
```

### DIFF
--- a/autotest/gdrivers/saga.py
+++ b/autotest/gdrivers/saga.py
@@ -201,7 +201,8 @@ def test_saga_9():
     ds = None
 
     ds = gdal.Open('tmp/test9.sdat')
-    ret = ds.GetRasterBand(1).SetNoDataValue(56)
+    with gdaltest.error_handler():
+        ret = ds.GetRasterBand(1).SetNoDataValue(56)
     assert ret == gdalconst.CE_Failure
     # make sure nodata value is not changed
     assert ds.GetRasterBand(1).GetNoDataValue() == -99999
--- a/autotest/gdrivers/saga.py
+++ b/autotest/gdrivers/saga.py
@@ -96,7 +96,7 @@ def test_saga_4():
             check_minmax = 1
         tst.testCreateCopy(new_filename='tmp/test4.sdat', check_minmax=check_minmax)
 
-    
+
 ###############################################################################
 # Test Create() for various data types
 
@@ -119,7 +119,7 @@ def test_saga_5():
             check_minmax = 1
         tst.testCreate(new_filename='tmp/test5.sdat', out_bands=1, check_minmax=check_minmax)
 
-    
+
 ###############################################################################
 # Test creating empty datasets and check that nodata values are properly written
 
@@ -161,7 +161,7 @@ def test_saga_6():
     except OSError:
         pass
 
-    
+
 ###############################################################################
 # Test /vsimem
 
@@ -191,5 +191,31 @@ def test_saga_8():
     PARAMETER["false_northing",0],
     UNIT["Meter",1]]""")
 
+##############################################################################
+# Test setnodata
+def test_saga_9():
+
+    gdal_type = gdal.GDT_Float64
+
+    ds = gdal.GetDriverByName('SAGA').Create('tmp/test9.sdat', 2, 2, 1, gdal_type)
+    ds = None
+
+    ds = gdal.Open('tmp/test9.sdat', gdal.GA_Update)
+
+    ds.GetRasterBand(1).SetNoDataValue(56)
+    nd = ds.GetRasterBand(1).GetNoDataValue()
+    # ds.SetGeoTransform(ds.GetGeoTransform())
+    assert nd == 56
+
+    ds = None
+
+    with open('tmp/test9.sgrd', 'r') as f:
+        header_string = f.read()
+        assert "NODATA_VALUE\t= 56.000000" in header_string
+    try:
+        os.remove('tmp/test9.sgrd')
+        os.remove('tmp/test9.sdat')
+    except OSError:
+        pass
 
 
--- a/autotest/gdrivers/saga.py
+++ b/autotest/gdrivers/saga.py
@@ -31,7 +31,7 @@
 
 import os
 from osgeo import gdal
-
+from osgeo import gdalconst
 
 import gdaltest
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by rouault** [comment]
The driver would probably need to have a m_HeaderDirty flag that would be set by the setter methods that affect the header (SetNoDataValue() and SetGeoTransform() as far as I can see), and the header would be written in the dataset destructor when the flag is set. Currently it seems the header is only written on SetGeoTransform() invokation https://github.com/OSGeo/gdal/blob/dc2b1e83a705c2cadb5ce78d1fabeeee57b95d78/frmts/saga/sagadataset.cpp#L791

File: frmts/saga/sagadataset.cpp [hunk 1]
```diff
@@ -72,6 +72,7 @@ class SAGADataset final: public GDALPamDataset
                                double dfZFactor, bool bTopToBottom );
     VSILFILE *fp;
     char     *pszProjection;
+    bool headerDirty=0;
```
rouault, line 75:
```suggestion
    bool headerDirty=false;
```

frmts/saga/sagadataset.cpp
```diff
@@ -304,6 +306,18 @@ double SAGARasterBand::GetNoDataValue( int * pbSuccess )
...
 
+/************************************************************************/
+/*                           SetNoDataValue()                           */
+/************************************************************************/
+
+CPLErr SAGARasterBand::SetNoDataValue(double dfNoData)
+{
+    m_NoData = dfNoData;
+    SAGADataset * poSAGADS = static_cast<SAGADataset *>(poDS);
+    poSAGADS->headerDirty=1;
```
rouault, line 317:
```suggestion
    poSAGADS->headerDirty=true;
```
Please also poSAGADS->eAccess == GA_Update at the beginning (similar to what is done in SetGeoTransform())

The CreateCopy() implementation also needs to call that method

frmts/saga/sagadataset.cpp
```diff
@@ -773,25 +797,15 @@ CPLErr SAGADataset::SetGeoTransform( double *padfGeoTransform )
...
-        poGRB->m_Rows = nRasterYSize;
-    }
-
-    return eErr;
+    poGRB->m_Xmin = dfMinX;
+    poGRB->m_Ymin = dfMinY;
+    poGRB->m_Cellsize = padfGeoTransform[1];
+    poGRB->m_Cols = nRasterXSize;
+    poGRB->m_Rows = nRasterYSize;
+    headerDirty = 1;
```
rouault, line 806:
```suggestion
    headerDirty = true;
```

frmts/saga/sagadataset.cpp
```diff
@@ -773,25 +797,15 @@ CPLErr SAGADataset::SetGeoTransform( double *padfGeoTransform )
...
-        poGRB->m_Cellsize = padfGeoTransform[1];
-        poGRB->m_Cols = nRasterXSize;
-        poGRB->m_Rows = nRasterYSize;
-    }
-
-    return eErr;
+    poGRB->m_Xmin = dfMinX;
+    poGRB->m_Ymin = dfMinY;
+    poGRB->m_Cellsize = padfGeoTransform[1];
+    poGRB->m_Cols = nRasterXSize;
```
rouault, line 804: I know this was there before but setting m_nCols and m_nRows in this method is weird. My understanding of the Create() and CreateCopy() implementation is that they should already be set (when Create() calls Open() at the end, which will read the serialized initial header)

johanvdw: I had a look, and actually m_nCols and m_nRows are not used at all. I will remove them.

frmts/saga/sagadataset.cpp
rouault: > The no data value of the raster band is already copied in CreateCopy
ok, I missed that it was setting it through the creation option (I thought that calling SetNoDataValue() would be now needed). So disregard that part of my comment

johanvdw: It is also a bit weird that the projection is written when calling SetProjection . Perhaps this should also toggle the headerDirty bit and be saved only when closing the dataset to make it consistent with header changes?

rouault: > It is also a bit weird that the projection is written when calling SetProjection . Perhaps this should also toggle the headerDirty bit and be saved only when closing the dataset to make it consistent with header changes?

up to you. As it writes another file than the .sgrd header, this isn't strictly needed (technically if defered writing of the .prj file is done, this should be done through another flag that headerDirty)

autotest/gdrivers/saga.py
```diff
@@ -191,5 +191,37 @@ def test_saga_8():
...
+# Test setnodata
+def test_saga_9():
+
+    gdal_type = gdal.GDT_Float64
+
+    ds = gdal.GetDriverByName('SAGA').Create('tmp/test9.sdat', 2, 2, 1, gdal_type)
+    ds = None
+
+    ds = gdal.Open('tmp/test9.sdat')
+    ret = ds.GetRasterBand(1).SetNoDataValue(56)
```
rouault, line 204: you may want to run that call under so that when running pytest in verbose mode (--capture=no -ra -vv), no error message is logged
```suggestion
    with gdaltest.error_handler():
        ret = ds.GetRasterBand(1).SetNoDataValue(56)
```

frmts/saga/sagadataset.cpp
johanvdw: Okay, I'll leave it as is in that case.

Issue #1368, opened 2019-03-17
Add build support for .NET core


szekerest:
It should be an additional option for the C# builds

szekerest:
@rouault .Net Standard is considered to be a subset of .Net core and any other framework implementations. By targeting .NET Standard, you can build libraries that you can share across all your .NET apps, no matter on which .NET implementation or OS they run. This is basically for class libraries (like for gdal_csharp.dll and ogr_csharp.dll). The console applications should target .NET Core as a minimum requirement.

For more info see: https://msdn.microsoft.com/en-us/magazine/mt842506.aspx

evilpilaf:
Is there any progress on this?

flensrocker:
This gets a little bit urgent.

Spatial support was added to EF Core 2.2, which is EOL - and Asp.Net Core 3.1 (the current LTS release) is without support for the classic .Net Framework.

So, going back to 2.1 (the last LTS) is not an option.

Being stuck to the 2.2 release is not good, since there was recently a security patch on the 2.1 release, which will not get ported to 2.2.

Porting this library to .Net Standard 2.0 should be the way to go.

jsmith-pathway:
@szekerest  Will this be in one of the upcoming releases?

BrannonKing:
Batchfile code I'm using to generate the .NET Standard 2.0 files:
```batch
rem Build the old wrap and csharp DLLs
cd swig\csharp
nmake -f makefile.vc interface MSVC_VER=1920 WIN64=1
nmake -f makefile.vc ogr_dir gdal_dir osr_dir const_dir gdal_csharp MSVC_VER=1920 WIN64=1
nmake -f makefile.vc install MSVC_VER=1920 WIN64=1

rem Remove the .NET Framework files
del "%GDAL_HOME%\csharp\*_csharp.dll"

rem Now make the .netstandard wrappers
setlocal ENABLEDELAYEDEXPANSION
set PROJECT_FILE_DEF=^<Project Sdk="Microsoft.NET.Sdk"^>^<PropertyGroup^>^<TargetFramework^>netstandard2.0^</TargetFramework^>^<AssemblyVersion^>3.1.2.0^</AssemblyVersion^>^</PropertyGroup^>^</Project^>
echo !PROJECT_FILE_DEF! > const\gdalconst_csharp.csproj
echo !PROJECT_FILE_DEF! > osr\osr_csharp.csproj
[...]

PR #5148, opened 2022-01-21
CSharp: Switch default platform target to AnyCPU #1368

szekerest:
## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:

### DIFF
--- a/swig/csharp/makefile.vc
+++ b/swig/csharp/makefile.vc
@@ -71,10 +71,11 @@ template.csproj:
     echo ^<Project Sdk="Microsoft.NET.Sdk"^> > $@
     echo ^<PropertyGroup^> >> $@
     echo ^<TargetFramework^>$(NETSTANDARD)^</TargetFramework^> >> $@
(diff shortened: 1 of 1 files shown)


rouault:
Is it appropriate for backport to 3.4 ?

szekerest:
> Is it appropriate for backport to 3.4 ?

Yes, that would be cool

runette:
Note that this change will not be reflected in the CMake builds - since they use a different template file. I will submit a very small PR that reflects this in the CMake script

szekerest:
@runette Thank you for your contribution the current makefile based approach still used by the gisinternals builds. 
I haven't tested this cmake version so far, but I think specifying the runtime identifier when compiling the gdal assemblies will invalidate the anycpu setting. Also we should target netstandard2.x by default instead of netcore 5.0. in order to keep supporting netframework 4.6.1 or above with that single build.

PR #5152, opened 2022-01-24
Add GetFieldDomainNames method to retrieve all field domain names which are available for a dataset

nyalldawson:
~~Implemented for OpenFileGDB driver only for now~~

~~The FileGDB and GPKG drivers will require special implementations of this, which will be done shortly. For now I'm after general feedback on the API.~~

Implemented for all drivers which currently support field domains

### DIFF
--- a/gcore/gdaldataset.cpp
+++ b/gcore/gdaldataset.cpp
@@ -8437,6 +8437,62 @@ void GDALDatasetClearStatistics(GDALDatasetH hDS)
     GDALDataset::FromHandle(hDS)->ClearStatistics();
 }
 
+
+/************************************************************************/
+/*                        GetFieldDomainNames()                         */
+/************************************************************************/
+
+/** Returns a list of the names of all field domains stored in the dataset.
+ *
+ * @note The default implementation assumes that drivers fully populate m_oMapFieldDomains
+ * when opening a dataset. If this assumption is incorrect then
+ * a specialized implementation of GetFieldDomainNames() must be implemented.
+ *
+ * @param papszOptions Driver specific options determining how attributes
+ * should be retrieved. Pass nullptr for default behavior.
+ *
+ * @return list of field domain names
+ * @since GDAL 3.5
+ */
+std::vector<std::string> GDALDataset::GetFieldDomainNames(CPL_UNUSED CSLConstList papszOptions) const{
+
+    std::vector<std::string> names;
+    names.reserve( m_oMapFieldDomains.size() );
+    for(const auto& it : m_oMapFieldDomains ) {
+        names.emplace_back( it.first );
+    }
+    return names;
+}
+
+
+/************************************************************************/
+/*                      GDALDatasetGetFieldDomainNames()                */
+/************************************************************************/
+
+/** Returns a list of the names of all field domains stored in the dataset.
+ *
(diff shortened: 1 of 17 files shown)


gcore/gdaldataset.cpp
```diff
@@ -8437,6 +8437,58 @@ void GDALDatasetClearStatistics(GDALDatasetH hDS)
...
+
+/** Returns a list of the names of all field domains stored in the dataset.
+ *
+ * @param papszOptions Driver specific options determining how attributes
+ * should be retrieved. Pass nullptr for default behavior.
+ *
+ * @return list of field domain names
+ * @since GDAL 3.5
+ */
+std::vector<std::string> GDALDataset::GetFieldDomainNames(CSLConstList) const{
```
rouault, line 8453: to make Doxygen happy
```suggestion
std::vector<std::string> GDALDataset::GetFieldDomainNames(CPL_UNUSED CSLConstList papszOptions) const{
```

gcore/gdaldataset.cpp
```diff
@@ -8437,6 +8437,58 @@ void GDALDatasetClearStatistics(GDALDatasetH hDS)
     GDALDataset::FromHandle(hDS)->ClearStatistics();
 }
 
+
+/************************************************************************/
+/*                        GetFieldDomains()                              */
```
rouault, line 8442:
```suggestion
/*                      GetFieldDomainNames()                            */
```

rouault (commented):
A note in the Doxygen comments of GDALDataset::AddFieldDomain() and GDALDataset::GetFieldDomain() that drivers should make sure to update  m_oMapFieldDomains for the default implementation of GetFieldDomainNames() to
work, otherwise they need to make a specialized implementation of it, could be helpful

rouault:
rebasing on master will fix the failure in the build-mac cmake CI target (due to https://github.com/OSGeo/gdal/pull/5156 being applied)

nyalldawson:
@rouault I've now implemented for all drivers which support field domains

ogr/ogrsf_frmts/gpkg/ogrgeopackagedatasource.cpp
```diff
@@ -7176,6 +7176,50 @@ const char* GDALGeoPackageDataset::GetGeometryTypeString(OGRwkbGeometryType eTyp
...
+        CPLError(CE_Warning, CPLE_AppDefined,
+                 "Number of rows returned for field domain names has been "
+                 "truncated.");
+    }
+    else if( oResultTable->RowCount() > 0 )
+    {
+        oDomainNamesList.reserve(oResultTable->RowCount());
+        for ( int i = 0; i < oResultTable->RowCount(); i++ )
+        {
+            const char *pszConstraintName = oResultTable->GetValue(0, i);
```
rouault, line 7215: pszConstraintName should be tested against nullptr (even if the official schema from the GPKG spec would have a not null constraint, nothing guarantee that the database we open respect it)

nyalldawson: Fixed, thanks

nyalldawson:
Thanks @rouault !

Issue #5153, opened 2022-01-24
CSV to geopackage: -dsco GEOMETRY_NAME=geom does not work


anbj:
I'm trying to convert a .csv file to gpkg with the *-dsco GEOMETRY_NAME=geom* option, but this returns a warning:
```
$ ogr2ogr -F GPKG -makevalid -a_srs EPSG:4230 -oo KEEP_GEOM_COLUMNS=NO -dsco GEOMETRY_NAME=geom gis.gpkg Area.csv -oo GEOM_POSSIBLE_NAMES=AreaWKT

Warning 6: driver GPKG does not support creation option GEOMETRY_NAME
```

* Should not this be possible? Ref. https://gdal.org/drivers/vector/gpkg.html#layer-creation-options: *GEOMETRY_NAME: Column to use for the geometry column. Default to “geom”. Note: This option was called GEOMETRY_COLUMN in releases before GDAL 2*

* When I omit *-dsco GEOMETRY_NAME=geom* the geometry column gets the same name as the input wkt-field. In the example above, the geometry field is named *AreaWKT*. Not the default, *geom* as indicated in the docs.

jratike80:
Bug tracker is meant for reporting bugs. Ask questions on gdal-dev mailing list. If discussion reveals a bug the issue into GitHub can be created later.

`-dsco` : datastore creation option, `-lco` : layer creation option

anbj:
Thanks jratike80.

PR #5163, opened 2022-01-25
gpkg.rst: improve wording of GEOMETRY_NAME description

anbj:
Reading the description of *Layer creation options* and *GEOMETRY_NAME*, it seems that the default name given to the geometry column is *geom*. This is the case *if it is not named in the data source*. If it is named, the geometry name in the new data source is inherited from the source data.

See also #5153.

### DIFF
--- a/doc/source/drivers/vector/gpkg.rst
+++ b/doc/source/drivers/vector/gpkg.rst
@@ -217,8 +217,8 @@ Layer Creation Options
 ~~~~~~~~~~~~~~~~~~~~~~
 
 -  **GEOMETRY_NAME**: Column to use for the geometry column. Default to
-   "geom". Note: This option was called GEOMETRY_COLUMN in releases before
-   GDAL 2
+   "geom" if it is not named in the source data.
+   Note: This option was called GEOMETRY_COLUMN in releases before GDAL 2
 -  **GEOMETRY_NULLABLE**: Whether the values of the
    geometry column can be NULL. Can be set to NO so that geometry is
    required. Default to "YES"


rouault:
> Reading the description of _Layer creation options_ and _GEOMETRY_NAME_, it seems that the default name given to the geometry column is _geom_. This is the case _if it is not named in the data source_. If it is named, the geometry name in the new data source is inherited from the source data.

hum, this is actually true and false. I assume you're speaking about what happens when using ogr2ogr, since ogr2ogr has logic to set GEOMETRY_NAME from the source geometry column (this is true for all other drivers that have a GEOMETRY_NAME layer creation optoin), if the user doesn't explicitly set -lco GEOMETRY_NAME=XXXX. So your proposed change should reflect that and/or update https://gdal.org/programs/ogr2ogr.html to explain that

anbj:
Well, to summarize my experience;

* I have a csv-file with a geometry column called *AreaWKT*.
* I want to convert this csv-file to a geopackage.
* Reading the docs about [layer creation options](https://gdal.org/drivers/vector/gpkg.html#layer-creation-options) and *GEOMETRY_NAME*, my interpretation is that the geometry column in the geopackage will be called *geom*.
* To my surprise, the geometry column is not called *geom*, but has the same name as the WKT column in the csv-file.

The docs says about *GEOMETRY_NAME*; ``Default to “geom”``. But as I understand it, ths is *only* when the column in the data source has no name. Is this correct?

rouault:
> The docs say about _GEOMETRY_NAME_; `Default to “geom”`. But as I understand it, this is _only_ when the column in the data source has no name. Is this correct?

The doc of the driver only speaks about the behaviou of driver when the method OGRLayer::CreateLayer() is used. So when you just use that method, then the default value for the geometry column is "geom".
But when you use the ogr2ogr utility, it automatically sets the GEOMETRY_NAME layer creation option to the source geometry name if it exists and the user doesn't have explicitly passed -lco GEOMETRY_NAME.
So the doc of the driver is technically correct, as ogr2ogr sets GEOMETRY_NAME in the above mentioned situation. So the doc enhancement you suggested should be amended to reflect that there's a ogr2ogr specific behaviour

anbj:
Thanks for explaining.

I'm a bit confused though; I thought that the doc of the driver (all drivers really) referred mainly *to* the usage of **ogr2ogr**. If **ogr2ogr** comes with a *specific behaviour*, what then is the *normal*/*ordinary* behaviour?

rouault:
> I thought that the doc of the driver (all drivers really) referred mainly _to_ the usage of **ogr2ogr**. If **ogr2ogr** comes with a _specific behaviour_, 

ogr2ogr is just a user of the OGR API. But other software use the OGR API without ogr2ogr (like the "Save as vector file" function of QGIS). The driver documentation is about the raw usage of the CreateLayer() API, not ogr2ogr specifics. The raw usage of CreateLayer() can't apply ogr2ogr behavior, since the target driver doesn't know anything about the source, because sometimes there is no source dataset.

> what then is the _normal_ / _ordinary_ / _other_ behaviour?

Well, the behaviour described in the driver doc currently :-)  If you do in Python
```
ds = ogr.GetDriverByName('GPKG').CreateDataSource('foo.gpkg')
lyr = ds.CreateLayer('foo')
```
then, the geometry name will be the default one, that is geom.
ogr2ogr specific behaviour is similar to doing:
```
ds = ogr.GetDriverByName('GPKG').CreateDataSource('foo.gpkg')
lyr = ds.CreateLayer('foo', options = ['GEOMETRY_NAME=' + source_geometry_name)
```

anbj:
Thanks alot, rouault, for your time and answer. Very appreciated.

I guess this is a classic case of ignorance shining through! Closing this.

rouault:
rather than closing, I'd have appreciated if you could amend your initial proposal to reflect the precisions I gave. This would be a better investment of your time, my time, and other people who may have similar question.

anbj:
Will do. Should I amend to the description of **GEOMETRY_NAME* (as this PR originally did) or create a paragraph? Or should it be added to the **ogr2ogr** doc?

rouault:
> Should I amend to the description of *_GEOMETRY_NAME_ (as this PR originally did) or create a paragraph?

Adding in the description of *_GEOMETRY_NAME_ is fine .

>  Or should it be added to the ogr2ogr doc?

I'd say that there could be a paragraph in there too about automatically overloaded layer creation options. Those are GEOMETRY_NAME, GEOMETRY_NULLABLE (set to NO if the source layer has a non-nullable geometry column) and FID (set to the name of the source FID column, if it is not empty)

anbj:
Will do when time permits.

rouault:
@anbj remainder that this PR is still open

anbj:
Yes, thanks for reminding me. I'll try to get it done.

rouault:
this PR is stalled

anbj:
Stalled, but not forgotten.

anbj:
Would ypu prefer to close this, and have me open a new PR when I get it done?

PR #5165, opened 2022-01-25
CMake: use Poppler_ for variable names

rouault:
to be consistent with other FindPoppler.cmake modules like https://github.com/pfultz2/extra-cmake-modules/blob/master/find-modules/FindPoppler.cmake or https://api.kde.org/ecm/find-module/FindPoppler.html

### DIFF
--- a/cmake/modules/packages/FindPoppler.cmake
+++ b/cmake/modules/packages/FindPoppler.cmake
@@ -19,7 +19,7 @@ Copyright (c) 2015, Alex Richardson <arichardson.kde@gmail.com>
 IMPORTED Targets
 ^^^^^^^^^^^^^^^^
 
-This module defines :prop_tgt:`IMPORTED` target ``POPPLER::Poppler``, if
+This module defines :prop_tgt:`IMPORTED` target ``Poppler::Poppler``, if
 curl has been found.
 
 Result Variables
@@ -27,16 +27,16 @@ Result Variables
 
 This module defines the following variables:
 
-``POPPLER_FOUND``
+``Poppler_FOUND``
   True if Poppler found.
 
-``POPPLER_INCLUDE_DIRS``
+``Poppler_INCLUDE_DIRS``
   where to find poppler/poppler-config.h, etc.
 
-``POPPLER_LIBRARIES``
+``Poppler_LIBRARIES``
   List of libraries when using Poppler.
 
-``POPPLER_VERSION_STRING``
+``Poppler_VERSION_STRING``
   The version of Poppler found.
 #]=======================================================================]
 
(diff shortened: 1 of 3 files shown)


frmts/pdf/CMakeLists.txt
```diff
@@ -21,19 +21,19 @@ target_include_directories(gdal_PDF PRIVATE ${GDAL_RASTER_FORMAT_SOURCE_DIR}/vrt
                                             ${GDAL_VECTOR_FORMAT_SOURCE_DIR}/mem)
 
 if (GDAL_USE_POPPLER)
-  target_include_directories(gdal_PDF PRIVATE ${POPPLER_INCLUDE_DIRS} ${POPPLER_INCLUDE_DIRS}/../)
-  gdal_target_link_libraries(TARGET gdal_PDF LIBRARIES POPPLER::Poppler)
-  string(REGEX REPLACE "([0-9]+)\\.([0-9]+).*" "\\1" POPPLER_VERSION_MAJOR ${POPPLER_VERSION_STRING})
-  string(REGEX REPLACE "([0-9]+)\\.([0-9]+).*" "\\2" POPPLER_VERSION_MINOR ${POPPLER_VERSION_STRING})
+  target_include_directories(gdal_PDF PRIVATE ${Poppler_INCLUDE_DIRS} ${Poppler_INCLUDE_DIRS}/../)
```
dg0yt, line 24: This line doesn't look healthy, given that `${Poppler_INCLUDE_DIRS}` might be a list of multiple paths. The var must contain all required paths. If it doesn't either the Find modules is broken, or the consuming source code. 
(If it is needed for transition, it formally still takes a loop to do this correctly.)

cmake/modules/packages/FindPoppler.cmake
```diff
@@ -71,82 +71,82 @@ set(Poppler_Qt4_header "poppler-qt4.h")
...
   if(_nextcomp GREATER -1)
-    find_path(POPPLER_${_comp}_INCLUDE_DIR
+    find_path(Poppler_${_comp}_INCLUDE_DIR
               NAMES ${Poppler_${_comp}_header}
               PATH_SUFFIXES poppler
               HINTS ${PC_Poppler_${_comp}_INCLUDE_DIRS})
-    find_library(POPPLER_${_comp}_LIBRARY
+    find_library(Poppler_${_comp}_LIBRARY
             NAMES ${Poppler_${_comp}_lib}
             HINTS ${PC_Poppler_${_comp}_LIBRARY_DIRS})
```
dg0yt, line 86: This doesn't have to be handled in this PR:
AFAICS, `PC_Poppler_${_comp}_INCLUDE_DIRS}` and `${PC_Poppler_${_comp}_LIBRARY_DIRS}` won't be initialized in this file. I assume that the component pc files were meant to be loaded in another loop with, the `PC_Poppler_${_comp}` prefix.

frmts/pdf/CMakeLists.txt
rouault: yeah I agree this is not super clean currently. The issue here is that we must also cope with the source code that was meant for the autoconf & nmake build systems that assume that the include will be ${prefix}/include (the #includes are "poppler/foo.h"), while the CMake module returns ${prefix}/include/poppler in Poppler_INCLUDE_DIR
When we don't need to support the historical biuld systems, we'll be able to modify the source code and just use ${Poppler_INCLUDE_DIRS} 
Anyway in practice FindPoppler.cmake only returns a single path (unless the user would specify several ones, but that doesn't make sense)

cmake/modules/packages/FindPoppler.cmake
rouault: maybe... I didn't write this.  For GDAL purposes, we don't use the components functionality when looking for Poppler

frmts/pdf/CMakeLists.txt
rouault: tracked as https://github.com/OSGeo/gdal/issues/5170

PR #5166, opened 2022-01-25
CMake CSharp: fix dependency rules so that a second 'make' without ..

rouault:
…any change doesn't result in .dll and sample files to be regenerated

CC @runette


runette:
@rouault thanks - I was trying to work out how to fix that.

rouault:
> I was trying to work out how to fix that.

Note that I haven't checked if there's a similar issue for the dotnet builds

PR #5167, opened 2022-01-25
/vsis3/: make sure file properties of /vsis3_streaming/foo are invalidated when /vsis3/foo ones are

rouault:
This fixes scenarios with using repeated VSISync() that may use
/vsis3_streaming/ to download file, and /vsis3/ to upload it.

### DIFF
--- a/port/cpl_vsil_curl.cpp
+++ b/port/cpl_vsil_curl.cpp
@@ -2782,11 +2782,16 @@ VSICurlFilesystemHandlerBase::GetCachedFileProp( const char* pszURL,
                                              FileProp& oFileProp )
 {
     CPLMutexHolder oHolder( &hMutex );
-
-    return oCacheFileProp.tryGet(std::string(pszURL), oFileProp) &&
-            // Let a chance to use new auth parameters
-           !(oFileProp.eExists == EXIST_NO &&
-             gnGenerationAuthParameters != oFileProp.nGenerationAuthParameters);
+    bool inCache;
+    if( oCacheFileProp.tryGet(std::string(pszURL), inCache) )
+    {
+        if( VSICURLGetCachedFileProp(pszURL, oFileProp) )
+        {
+            return true;
+        }
+        oCacheFileProp.remove(std::string(pszURL));
+    }
+    return false;
 }
 
 /************************************************************************/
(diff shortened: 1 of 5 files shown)


port/cpl_vsil_curl.cpp
```diff
@@ -4880,6 +4893,84 @@ int VSICurlParseUnixPermissions(const char* pszPermissions)
...
+/************************************************************************/
+/*               VSICURLInvalidateCachedFilePropPrefix()                */
+/************************************************************************/
+
+void VSICURLInvalidateCachedFilePropPrefix( const char* pszURL )
+{
+    std::lock_guard<std::mutex> oLock(oCacheFilePropMutex);
+    if( poCacheFileProp != nullptr )
+    {
+        std::list<std::string> keysToRemove;
```
nyalldawson, line 4950: why are you using a list here instead of vector?

port/cpl_vsil_curl.cpp
```diff
@@ -4880,6 +4893,84 @@ int VSICurlParseUnixPermissions(const char* pszPermissions)
...
+        const size_t nURLSize = strlen(pszURL);
+        auto lambda = [&keysToRemove, &pszURL, nURLSize](
+            const lru11::KeyValuePair<std::string, FileProp>& kv)
+        {
+            if( strncmp(kv.key.c_str(), pszURL, nURLSize) == 0 )
+                keysToRemove.push_back(kv.key);
+        };
+        poCacheFileProp->cwalk(lambda);
+        for( auto& key: keysToRemove )
```
nyalldawson, line 4959: is the intention of separating the visiting from the removal here to avoid modifying the props while they are being visited?

port/cpl_vsil_curl.cpp
rouault: I guess both could make sense (we just need iterative access, not random access, so std::list is not inappropriate). The lru11::Cache implementation in port/cpl_mem_cache.h (that is a third party originally) uses a std::list internally too

port/cpl_vsil_curl.cpp
rouault: yes, I'm not completely sure that removing while visiting is safe here.

PR #5174, opened 2022-01-26
Add driver metadata for GDAL_DCAP_FIELD_DOMAINS and GDAL_DMD_CREATION_FIELD_DOMAIN_TYPES

nyalldawson:
Reflects if the driver supports reading field domains, and lists
supported domain types for creation respectively


frmts/mem/memdataset.cpp
```diff
@@ -2483,6 +2483,9 @@ void GDALRegister_MEM()
 "   </Option>"
 "</CreationOptionList>" );
 
+    poDriver->SetMetadataItem( GDAL_DCAP_FIELD_DOMAINS, "YES" );
```
rouault, line 2486: oops I merged this, but I now realize that this part is incorrect. This should be moved to the ogr/ogrsf_frmts/mem/ driver.  I'll do the fix in a follow up commit

rouault: Done in 7da4d8b0b0

nyalldawson: Thanks!

PR #5186, opened 2022-01-27
Add a ConfigUser.cmake that can host optional settings.

joa-quim:
The ConfigUser.cmake file is included by the root CMakeList.txt and by default it has only commented lines

Add also a rule in gdal.cmake to optionally let users set the shared lib name. That control is set in the ConfigUser.cmake file.


CMakeLists.txt
```diff
@@ -87,6 +87,9 @@ endif ()
 project(gdal LANGUAGES C CXX)
 include(CTest)
 
+# Include defaults
+include ("${CMAKE_SOURCE_DIR}/cmake/ConfigUser.cmake")
```
rouault, line 91: Is it a common / best-practice to do that ? Are other projects doing such thing ?
And the use of CMAKE_SOURCE_DIR is incorrect here as it breaks adding GDAL as a sub project. This should rather be CMAKE_CURRENT_SOURCE_DIR or PROJECT_SOURCE_DIR.

gdal.cmake
```diff
@@ -256,6 +256,14 @@ function (gdal_add_private_link_libraries)
 endfunction (gdal_add_private_link_libraries)
 
 add_library(${GDAL_LIB_TARGET_NAME} gcore/gdal.h)
+
+# If a shared lib renaming has been set in ConfigUser.cmake
+if (GDAL_SHAREDLIB_OUTPUT_NAME)
```
rouault, line 261: it should be documented in build_hints.rst

CMakeLists.txt
rouault: actually I see that this need should be covered by the -C switch of CMake:
```
       -C <initial-cache>
              Pre-load a script to populate the cache.

              When  CMake  is  first  run in an empty build tree, it creates a CMakeCache.txt file and populates it with customizable settings for the project.  This option may be used to specify a
              file from which to load cache entries before the first pass through the project’s CMake listfiles.  The loaded entries take priority over the project’s default values.  The given file
              should be a CMake script containing set() commands that use the CACHE option, not a cache-format file.
```
Using that rather cmake/ConfigUser.cmake would avoid having in version control a file that users could modify and thus accidentally commit if they are GDAL developers.

The https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html newer functionality also seems close to that

gdal.cmake
```diff
@@ -256,6 +256,14 @@ function (gdal_add_private_link_libraries)
...
+  set_target_properties(GDAL PROPERTIES RUNTIME_OUTPUT_NAME ${GDAL_SHAREDLIB_OUTPUT_NAME})
```
rouault, line 262:
```suggestion
  set_target_properties(${GDAL_LIB_TARGET_NAME} PROPERTIES RUNTIME_OUTPUT_NAME ${GDAL_SHAREDLIB_OUTPUT_NAME})
```

But actually the if() else () thing could be avoided if you do
```
set(GDAL_LIB_OUTPUT_NAME "gdal" CACHE STRING "Name of the GDAL library")
set_target_properties(${GDAL_LIB_TARGET_NAME} PROPERTIES OUTPUT_NAME ${GDAL_LIB_OUTPUT_NAME})
```
GDAL_LIB_OUTPUT_NAME is a more general name, since it can also apply to static builds

gdal.cmake
joa-quim: It is but it refers ``cmake/ConfigUser.cmake`` so that needs to be changed after first point is solved.

CMakeLists.txt
joa-quim: GMT and NetCDF use it, though NetCDF picked it from GMT and put the ``ConfigUser.cmake`` under version control, whilst GMT says is up to the user to create that file (and it has cmake logic to create an empty one to not error).

I'm fine with any solution that makes it simpler and also don't like much the idea of putting a file under version control and then tell the users to modify it at their own need.
Perhaps you can improve this bit yourself.

rouault:
could you drop the ConfigUser.cmake part from this pull request and just keep the change in  gdal.cmake ?
I believe a hint in build_hints.rst to using thee -C switch of CMake might be enough

joa-quim:
Not for my needs. The ``ConfigUser`` file is were I set the location of all dependencies. 
Just the start of mine ...

but I understand that those are my needs and I can keep my local changes like I do with all other packages that I build. 

```
if (WIN32)
	set (BITAGE 64)
	# Detect if we are building a 32 or 64 bits version
	if (CMAKE_SIZEOF_VOID_P EQUAL 8)
		set(BITAGE 64)
	endif ()
	# Uncomment to have the dll named other than the default 'gdal.dll'
	set (GDAL_SHAREDLIB_OUTPUT_NAME gdal_w${BITAGE})

	set(BLOSC_INCLUDE_DIR C:/programs/compa_libs/c-blosc/compileds/VC14_${BITAGE}/include)
	set(BLOSC_LIBRARY C:/programs/compa_libs/c-blosc/compileds/VC14_${BITAGE}/lib/blosc.lib)

	set(CFITSIO_INCLUDE_DIR C:/programs/compa_libs/cfits335/compileds/VC14_${BITAGE}/include)
	set(CFITSIO_LIBRARY C:/programs/compa_libs/cfits335/compileds/VC14_${BITAGE}/lib/cfitsio.lib)
...
```

joa-quim:
We could do in root ``CMakeList.txt`` and so the optional ``ConfigUser`` would be out of version control
```
if (EXISTS ConfigUser.cmake)
    include (ConfigUser.cmake)
endif ()
```

rouault:
> We could do in root `CMakeList.txt` and so the optional `ConfigUser` would be out of version control

that seems a better solution.
Can you also run ./scripts/reformat.sh to make sure the files are properly formatted ? (the "Lint CMakeLists.txt" job complains)

gdal.cmake
joa-quim: Had to revert this because it I did the suggestion wrong or it doesn't work.

rouault:
> Revert to previous if() else() solution because last one didn't work.

why ? I suspect it might work but in your ConfigUser.cmake  you need to set variables with the CACHE option (as suggested in the description of the -C option of cmake : "should be a CMake script containing set() commands that use the CACHE option").  The first set( ) with the CACHE option wins over following ones

For example if I run cmake on a CMakeLists.txt with
```
cmake_minimum_required(VERSION 3.10)
project(test)

# simulate what is set in ConfigUser.cmake
set(FOO "foo" CACHE STRING "")

set(FOO "bar" CACHE STRING "long description")

message(${FOO})
```
it outputs "foo",

but if I just do
```
cmake_minimum_required(VERSION 3.10)
project(test)

# simulate what is set in ConfigUser.cmake
set(FOO "foo" )

set(FOO "bar" CACHE STRING "long description")

message(${FOO})
```
it outputs "bar" on the first execution, but "foo" on next ones if I don't remove the cache.
This is one of the oddities of CMake variables.

So  ConfigUser.cmake if really an alternative to cmake -C ConfigUser.cmake

CMakeLists.txt
```diff
@@ -87,6 +87,11 @@ endif ()
 project(gdal LANGUAGES C CXX)
 include(CTest)
 
+# Include defaults
+if (EXISTS "${PROJECT_SOURCE_DIR}/ConfigUser.cmake")
+  include(ConfigUser.cmake)
```
dg0yt, line 92: Apart from using different paths: there is an `OPTIONAL` keyword for `include` so the `if` is not needed.

gdal.cmake
```diff
@@ -256,6 +256,14 @@ function (gdal_add_private_link_libraries)
 endfunction (gdal_add_private_link_libraries)
 
 add_library(${GDAL_LIB_TARGET_NAME} gcore/gdal.h)
+
+# If a shared lib renaming has been set in ConfigUser.cmake
+if (GDAL_LIB_OUTPUT_NAME)
+  set_target_properties(${GDAL_LIB_TARGET_NAME} PROPERTIES RUNTIME_OUTPUT_NAME ${GDAL_LIB_OUTPUT_NAME})
+else ()
+  set_target_properties(${GDAL_LIB_TARGET_NAME} PROPERTIES OUTPUT_NAME "gdal")
+endif ()
```
dg0yt, line 265: Don't use `if`.
Just use a cache variable initialized with `"gdal"`.

dg0yt (commented):
I don't see the added value of `ConfigUser.cmake`. It seems to me that it is broken by design: It makes it impossible to overwrite default values from the command line or a cmake GUI. This can create additional support requests.

gdal.cmake
dg0yt: And note that it uses different properties...

joa-quim:
If I do
```
cmake -G "Ninja" -DCMAKE_CXX_COMPILER=cl -DCMAKE_C_COMPILER=cl -C ../ConfigUser.cmake ..
loading initial cache file ../ConfigUser.cmake
...
  Could NOT find PROJ (missing: PROJ_LIBRARY PROJ_INCLUDE_DIR) (Required is
  at least version "6.0")
```
That is, none of the settings in ``ConfigUser`` were set (it found no dependencies).

rouault:
> That is, none of the settings in `ConfigUser` were set (it found no dependencies).

did you remove CMakeCache.txt before ?

joa-quim:
Ok, after using the CACHE option in all settings it now works ... sort of.
A problem, though a different one but it has now became more acute, is that now I have linking problems.

```
[1411/1467] Linking CXX shared library gdal_w64d.dll
FAILED: gdal_w64d.dll GDALd.lib
cmd.exe /C "cd . && C:\programs\CMake\bin\cmake.exe -E vs_link_dll --intdir=CMakeFiles\GDAL.dir --rc=C:\PROGRA~2\WI3CF2~1\10\bin\10.0.19041.0\x64\rc.exe --mt=C:\PROGRA~2\WI3CF2~1\10\bin\10.0.19041.0\x64\mt.exe --manifests  -- "C:\PROGRA~1\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.30.30705\bin\Hostx64\x64\link.exe" /nologo @CMakeFiles\GDAL.rsp  /out:gdal_w64d.dll /implib:GDALd.lib /pdb:gdal_w64d.pdb /dll /version:34.0 /machine:x64 /debug /INCREMENTAL  && cd ."
LINK Pass 1: command "C:\PROGRA~1\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.30.30705\bin\Hostx64\x64\link.exe /nologo @CMakeFiles\GDAL.rsp /out:gdal_w64d.dll /implib:GDALd.lib /pdb:gdal_w64d.pdb /dll /version:34.0 /machine:x64 /debug /INCREMENTAL /MANIFEST /MANIFESTFILE:CMakeFiles\GDAL.dir/intermediate.manifest CMakeFiles\GDAL.dir/manifest.res" failed (exit code 1120) with the following output:
   Creating library GDALd.lib and object GDALd.exp
tif_ojpeg.c.obj : error LNK2019: unresolved external symbol jpeg_std_error referenced in function gdal_OJPEGWriteHeaderInfo
```
I had them before and thought it was due to ``OpenJpeg`` but now I removed that dependency (as well as ``turbojpeg`` and they happen all the times. Before, with some incomprehensible jpeg *sets & unsets* I was able to build ... sometime. But now, never. But as I said this seems a different issue.

joa-quim:
Yes, that fixed it. Meanwhile I got back to the linking errors but that was fixed when I repeated the cmake command before running ``ninja``
```
cmake -G "Ninja" -DCMAKE_CXX_COMPILER=cl -DCMAKE_C_COMPILER=cl -C ../ConfigUser.cmake ..
```

rouault:
> Yes, that fixed it. Meanwhile I got back to the linking errors but that was fixed when I repeated the cmake command before running `ninja`

intersesting. Perhaps you should save CMakeCache.txt in the case where it works, and in the case where it doesn't work, and compare. There might be an issue with a cache vs non-cache variable. Do you use set( .... CACHE ...) in your ConfigUser.cmake ?

joa-quim:
This is more obscure. I had added back the OpenJpeg and turbojpeg dependencies and with them the above trick did not work. Removing those dependencies, however, made it work again. The differences in the ``CMakeCache.txt`` don't look important
```
# When it worked
//No help, variable specified on the command line.
CMAKE_CXX_COMPILER:UNINITIALIZED=cl
```
vs
```
# When it didn't
//CXX compiler
CMAKE_CXX_COMPILER:STRING=C:/Program Files/Microsoft Visual Studio/2022/Community/VC/Tools/MSVC/14.30.30705/bin/Hostx64/x64/cl.exe
```
But the ``build.ninja`` files differ significantly and it places referring to JPEG. I'm attaching them for you to see.
[build.ninja.zip](https://github.com/OSGeo/gdal/files/7959977/build.ninja.zip)

rouault:
> I had added back the OpenJpeg and turbojpeg dependencies and with them the above trick did not work. Removing those dependencies, however, made it work again. 

OpenJPEG is completely unrelated to the issue.
turbojpeg is clearly more related : perhaps you have run into the ""libjpeg-turbo" != "TurboJPEG"" issue of https://libjpeg-turbo.org/About/TurboJPEG. GDAL expects a library that has the libjpeg ABI, not the turbojpeg one.

The differences in the `CMakeCache.txt` don't look important

Was it the only difference really ?

What is the output of ``grep JPEG CMakeCache.txt  | grep -iv OPENJPEG``

> But the build.ninja files differ significantly and in places referring to JPEG. 

in the working case, it used the internal libjpeg12, where in the non-working case it was disabled. What control this behaviour is the cache variable GDAL_JPEG12_SUPPORTED which defaults to ON. That said I've just done a build with it to OFF, and it works too

joa-quim:
> turbojpeg is clearly more related : perhaps you have run into the ""libjpeg-turbo" != "TurboJPEG"" issue of https://libjpeg-turbo.org/About/TurboJPEG. GDAL expects a library that has the libjpeg ABI, not the turbojpeg one.

The problem occurs also when I'm not using turbojpeg, so not this.

> Was it the only difference really ?

Yes, in two instances.

> What is the output of grep JPEG CMakeCache.txt | grep -iv OPENJPEG

```
grep JPEG CMakeCache.txt | grep -iv OPENJPEG
//Whether to enable JPEG2000 support with MrSID SDK
//Set ON to build JPEG image format format
GDAL_ENABLE_FRMT_JPEG:BOOL=ON
GDAL_JPEG12_SUPPORTED:BOOL=ON
// GDAL_USE_JPEG)
GDAL_USE_LIBJPEG_INTERNAL:BOOL=ON
JPEG_INCLUDE_DIR:PATH=JPEG_INCLUDE_DIR-NOTFOUND
JPEG_LIBRARY_DEBUG:FILEPATH=JPEG_LIBRARY_DEBUG-NOTFOUND
JPEG_LIBRARY_RELEASE:FILEPATH=JPEG_LIBRARY_RELEASE-NOTFOUND
RENAME_INTERNAL_LIBJPEG_SYMBOLS:BOOL=ON
//ADVANCED property for variable: JPEG_INCLUDE_DIR
JPEG_INCLUDE_DIR-ADVANCED:INTERNAL=1
//ADVANCED property for variable: JPEG_LIBRARY_DEBUG
JPEG_LIBRARY_DEBUG-ADVANCED:INTERNAL=1
//ADVANCED property for variable: JPEG_LIBRARY_RELEASE
JPEG_LIBRARY_RELEASE-ADVANCED:INTERNAL=1
```

The same since the diffs were only those mentioned in the other post.
```
grep JPEG CMakeCache_NOWORK.txt | grep -iv OPENJPEG
//Whether to enable JPEG2000 support with MrSID SDK
//Set ON to build JPEG image format format
GDAL_ENABLE_FRMT_JPEG:BOOL=ON
GDAL_JPEG12_SUPPORTED:BOOL=ON
// GDAL_USE_JPEG)
GDAL_USE_LIBJPEG_INTERNAL:BOOL=ON
JPEG_INCLUDE_DIR:PATH=JPEG_INCLUDE_DIR-NOTFOUND
JPEG_LIBRARY_DEBUG:FILEPATH=JPEG_LIBRARY_DEBUG-NOTFOUND
JPEG_LIBRARY_RELEASE:FILEPATH=JPEG_LIBRARY_RELEASE-NOTFOUND
RENAME_INTERNAL_LIBJPEG_SYMBOLS:BOOL=ON
//ADVANCED property for variable: JPEG_INCLUDE_DIR
JPEG_INCLUDE_DIR-ADVANCED:INTERNAL=1
//ADVANCED property for variable: JPEG_LIBRARY_DEBUG
JPEG_LIBRARY_DEBUG-ADVANCED:INTERNAL=1
//ADVANCED property for variable: JPEG_LIBRARY_RELEASE
JPEG_LIBRARY_RELEASE-ADVANCED:INTERNAL=1
```

> in the working case, it used the internal libjpeg12, where in the non-working case it was disabled.

But I haven't enable or disabled anything in between.

rouault:
I can't really make sense of why you get an error, but the error message you got
> tif_ojpeg.c.obj : error LNK2019: unresolved external symbol jpeg_std_error referenced in function gdal_OJPEGWriteHeaderInfo

is interesting since it shows that when building tif_ojpeg.c it tried to use symbol jpeg_std_error. But with GDAL_USE_LIBJPEG_INTERNAL=ON and RENAME_INTERNAL_LIBJPEG_SYMBOLS=ON, it should instead tried to use gdal_jpeg_std_error (gdal_ prefix before symbol name). So it looks like if there was a stale tif_ojpeg.c.obj object built with another configuration with an external libjpeg that was tried to be linked into a build with internal libjpeg. Not sure if it is an issue in how we capture dependencies in our CMakeLists.txt files or if it is an issue of CMake w.r.t Ninja build files.

joa-quim:
Will try later with non-ninja build ... but that is significantly slower.

doc/source/build_hints.rst
```diff
@@ -54,6 +54,11 @@ You can unset existing cached variables, by using the -U switch of cmake, for ex
 
     cmake .. -UGDAL_USE_*
 
+You can assemble dependency settings in a file ``ConfigUser.cmake`` and use it with the -C option.
+The file contains set() commands that use the CACHE option. You can set for example a different name
+for the shared lib, *e.g.* ``set (GDAL_LIB_OUTPUT_NAME gdal_x64 CACHE STRING "" FORCE)``
```
rouault, line 59:
```suggestion
for the shared lib, *e.g.* ``set (GDAL_LIB_OUTPUT_NAME gdal_x64 CACHE STRING "" FORCE)``::
```

dg0yt:
> > Note also that don't know why but the lib name is upper case, _i.e._, `gdal_w64d.dll, gdal_w64d.pdb, GDALd.exp, GDALd.lib`
> 
> Perhaps change RUNTIME_OUTPUT_NAME back to OUTPUT_NAME My understanding of https://cmake.org/cmake/help/latest/prop_tgt/RUNTIME_OUTPUT_NAME.html and https://cmake.org/cmake/help/latest/prop_tgt/LIBRARY_OUTPUT_NAME.html is that OUTPUT_NAME overrides RUNTIME_OUTPUT_NAME and LIBRARY_OUTPUT_NAME

No, it is `OUTPUT_NAME` which is overridden, possibly even per config.

`RUNTIME_OUTPUT_NAME` is for the run-time part, i.e. the DLL on Windows.
`LIBRARY_OUTPUT_NAME` ist for the link-time part, i.e. the import lib.
`ARCHIVE_OUTPUT_NAME` is for a static library.
`OUTPUT_NAME` is the used when the other properties are not set.
And the target name is the default for `OUTPUT_NAME`, so `GDAL` here.
The names can be decorated with PREFIX and POSTFIX stuff.

Think twice before changing the (`LIBRARY_`/`ARCHIVE_`)`OUTPUT_NAME`. This is what is usually looked for with `find_library(... NAMES gdal)`, or linked to with `-lgdal`.

Last not least, the current CMake build system puts the  `OUTPUT_NAME` into the pc file. So changing `LIBRARY_OUTPUT_NAME`, `ARCHIVE_OUTPUT_NAME`, prefixes and postfixes breaks pkg-config.

dg0yt:
> But the `build.ninja` files differ significantly and in places referring to JPEG. I'm attaching them for you to see.
> [build.ninja.zip](https://github.com/OSGeo/gdal/files/7959977/build.ninja.zip)

The working file compiles with `-DJPEG_DUAL_MODE_8_12` and `-IC:\programs\compa_libs\gdal_J\build\frmts\jpeg\libjpeg12`.

rouault:
> No, it is `OUTPUT_NAME` which is overridden,

Yeah, was badly formulated. I meant that it is used as the default value for the other settings.

> Last not least, the current CMake build system puts the OUTPUT_NAME into the pc file. So changing LIBRARY_OUTPUT_NAME, ARCHIVE_OUTPUT_NAME, prefixes and postfixes breaks pkg-config.

What would be the appropriate property to fetch to take into account all that ?

joa-quim:
Tried with ``nmake`` but same failure 
```
...
tif_jpeg.c.obj : error LNK2019: unresolved external symbol jpeg_has_multiple_scans referenced in function TIFFjpeg_has_multiple_scans
tif_jpeg.c.obj : error LNK2019: unresolved external symbol jpeg_abort referenced in function TIFFjpeg_error_exit
tif_jpeg.c.obj : error LNK2019: unresolved external symbol jpeg_resync_to_restart referenced in function TIFFjpeg_data_src
gdal_w64d.dll : fatal error LNK1120: 21 unresolved externals
NMAKE : fatal error U1077: 'C:\programs\CMake\bin\cmake.exe' : return code '0xffffffff'
```
and also rerunning
```
cmake -G "NMake Makefiles" -DCMAKE_CXX_COMPILER=cl -DCMAKE_C_COMPILER=cl -C ../confas.cmake ..
```
finished the build on the second time

rouault:
> finished the build on the second time

weird. could you share a minimum reproducer for this, ie starting from a fresh clone + minimum set of external dependencies + cmake invocation ?

joa-quim:
Hmm, let's see if these are enough. I have a somewhat older version of my dependencies in

http://fct-gmt.ualg.pt/gmtdependswin/

(not sure if the PROJ one is updated enough to build current GDAL but if not I can update it later)

Pick a few of them, trim and adapt my ``ConfigUser.cmake`` (attached) 

In the ``build`` dir of the fork I used to make the PR (https://github.com/joa-quim/gdal)
```
cmake -G "Ninja" -DCMAKE_CXX_COMPILER=cl -DCMAKE_C_COMPILER=cl -C ../ConfigUser.cmake ..
ninja
```
[ConfigUser.zip](https://github.com/OSGeo/gdal/files/7961685/ConfigUser.zip)

dg0yt:
> > Last not least, the current CMake build system puts the OUTPUT_NAME into the pc file. So changing LIBRARY_OUTPUT_NAME, ARCHIVE_OUTPUT_NAME, prefixes and postfixes breaks pkg-config.
> 
> What would be the appropriate property to fetch to take into account all that ?

There is no such single property. Generating useful pkgconfig data out of CMake is just painful.

PR #5191, opened 2022-01-29
Update FindPCRE2.cmake

dg0yt:
## What does this PR do?

Fixes pcre2 lib for static and debug configurations on windows.
When building pcre2 with CMake and MSVC, the static archives are suffixed `-static`, and for WIN32, there is a debug postfix `d`.
https://github.com/PhilipHazel/pcre2/blob/111cd470b55d0c06bfb157c8f4521907d3d95fdb/CMakeLists.txt#L692-L694
https://github.com/PhilipHazel/pcre2/blob/111cd470b55d0c06bfb157c8f4521907d3d95fdb/CMakeLists.txt#L515-L517

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/cmake/modules/packages/FindPCRE2.cmake
+++ b/cmake/modules/packages/FindPCRE2.cmake
@@ -9,7 +9,7 @@
 # Find the native PCRE2 headers and libraries.
 
 find_path(PCRE2_INCLUDE_DIR NAMES pcre2.h)
-find_library(PCRE2_LIBRARY NAMES pcre2-8)
+find_library(PCRE2_LIBRARY NAMES pcre2-8 pcre2-8d pcre2-8-static pcre2-8-staticd NAMES_PER_DIR)
 include(FindPackageHandleStandardArgs)
 find_package_handle_standard_args(PCRE2
                                   FOUND_VAR PCRE2_FOUND


cmake/modules/packages/FindPCRE2.cmake
```diff
@@ -9,7 +9,7 @@
 # Find the native PCRE2 headers and libraries.
 
 find_path(PCRE2_INCLUDE_DIR NAMES pcre2.h)
-find_library(PCRE2_LIBRARY NAMES pcre2-8)
+find_library(PCRE2_LIBRARY NAMES pcre2-8 pcre2-8d pcre2-8-static pcre2-8-staticd NAMES_PER_DIR)
```
rouault, line 12: shouldn't we rather use a logic like in cmake/modules/packages/FindGeoTIFF.cmake to separate finding release and debug dependencies ?

dg0yt: - For a complete story, you have to do this with every pacakage. Even when debug and release use the same library name, they are not the same file (path).
- Unfortunately, even some CMake modules don't do it yet.
- Targets can deal with with different configs much better than variables, using per-config properties.
- The `<Pkg>_LIBRARY` and `<Pkg>_LIBRARIES` variables are a minefield. The common pattern is to use `select_library_configurations` which uses the keywords `optimized` and `debug` to merge the variants. 
  However, consumers sometimes assume that `<Pkg>_LIBRARY` is a single file path, typically in using it with `get_filename_component`. 
  In gdal, I fell over this issue with expat: Because `FindEXPAT.cmake` is lacking multiconfig support, vcpkg initializes `EXPAT_LIBRARY` using `select_library_configurations`. But gdal tries to take get the lib name for writing the pkg-config file... (WIP: https://github.com/microsoft/vcpkg/blob/a53d65aa9308e93d44b8fcaf2c51324bcda9cb26/ports/gdal/generate-config.patch)

Personally I still have no use for multi-config, but I must deal with it when contributing to vcpkg. In vcpkg, debug and release are built separately as single-config, and merged in the install step after the build. That's why this change is sufficient at gdal build time, for now. Things may look different when a static build of gdal is actually used in multi-config build...

rouault: ok, I'll merge this as it

PR #5200, opened 2022-01-30
CMake: implement documented GDAL_USE_TIFF switch, rename variables controlling driver activation

rouault:
and other cleanups

also add:
- add GDAL_USE_EXTERNAL_LIBS=OFF option to disable all external libs
- rename GDAL_ENABLE_FRMT_ variables to GDAL_ENABLE_DRIVER_ and OGR_ENABLE_ to OGR_ENABLE_DRIVER_


dg0yt:
I have some remarks, but it will take a moment to write it down.

doc/source/build_hints.rst
```diff
@@ -720,6 +720,12 @@ recommended to get best performance.
 See https://cmake.org/cmake/help/latest/module/FindJPEG.html for more details
 on how the library is detected.
 
+.. note::
+
+    Whe using libjpeg-turbo, JPEG_LIBRARY[_RELEASE/_DEBUG] should point to a
```
joa-quim, line 725: ``When``

cmake/helpers/CheckDependentLibraries.cmake
```diff
@@ -136,35 +165,19 @@ gdal_check_package(LibXml2 "Read and write XML formats" CAN_DISABLE)
...
-    message(FATAL_ERROR "GDAL_USE_ZLIB or GDAL_USE_ZLIB_INTERNAL must be set to ON")
-  endif ()
-else ()
-  set(GDAL_USE_ZLIB_INTERNAL
-      OFF
-      CACHE BOOL "Use internal zlib copy (if set to ON, has precedence over GDAL_USE_ZLIB)")
+invert_on_off(GDAL_USE_ZLIB NOT_GDAL_USE_ZLIB)
+set(GDAL_USE_ZLIB_INTERNAL
+    ${NOT_GDAL_USE_ZLIB}
+    CACHE BOOL "Use internal zlib copy (if set to ON, has precedence over GDAL_USE_ZLIB)")
```
dg0yt, line 173: I would generally prefer `option` for all the boolean variables: 
~~~
option(GDAL_USE_ZLIB_INTERNAL "${NOT_GDAL_USE_ZLIB}"
  "Use internal zlib copy (if set to ON, has precedence over GDAL_USE_ZLIB)")
~~~

cmake/helpers/CheckDependentLibraries.cmake
```diff
@@ -173,114 +186,79 @@ endif ()
...
-endif ()
+gdal_check_package(TIFF "Support for the Tag Image File Format (TIFF)." VERSION 4.0 CAN_DISABLE)
 set_package_properties(
   TIFF PROPERTIES
   URL "https://libtiff.gitlab.io/libtiff/"
-  DESCRIPTION "support for the Tag Image File Format (TIFF)."
+  DESCRIPTION "Support for the Tag Image File Format (TIFF)."
   TYPE RECOMMENDED)
+invert_on_off(GDAL_USE_TIFF NOT_GDAL_USE_TIFF)
+set(GDAL_USE_LIBTIFF_INTERNAL
```
dg0yt, line 196: I would prefer if the variable was named `GDAL_USE_TIFF_INTERNAL`, similar to `GDAL_USE_TIFF`.

cmake/helpers/CheckDependentLibraries.cmake
```diff
@@ -173,114 +186,79 @@ endif ()
...
-      CACHE BOOL "Use internal libgeotiff copy (if set to ON, has precedence over GDAL_USE_GEOTIFF)")
-  if (NOT GDAL_USE_LIBGEOTIFF_INTERNAL)
-    message(FATAL_ERROR "GDAL_USE_GEOTIFF or GDAL_USE_LIBGEOTIFF_INTERNAL must be set to ON")
-  endif ()
-else ()
-  set(GDAL_USE_LIBGEOTIFF_INTERNAL
-      OFF
-      CACHE BOOL "Use internal libgeotiff copy (if set to ON, has precedence over GDAL_USE_GEOTIFF)")
+invert_on_off(GDAL_USE_GEOTIFF NOT_GDAL_USE_GEOTIFF)
+set(GDAL_USE_LIBGEOTIFF_INTERNAL
```
dg0yt, line 209: I would prefer if the variable was named `GDAL_USE_GEOTIFF_INTERNAL`, similar to `GDAL_USE_GEOTIFF`.

cmake/helpers/CheckDependentLibraries.cmake
```diff
@@ -173,114 +186,79 @@ endif ()
...
-  set(GDAL_USE_LIBPNG_INTERNAL
-      ON
-      CACHE BOOL "Use internal libpng copy (if set to ON, has precedence over GDAL_USE_PNG)")
-else ()
-  set(GDAL_USE_LIBPNG_INTERNAL
-      OFF
-      CACHE BOOL "Use internal libpng copy (if set to ON, has precedence over GDAL_USE_PNG)")
-endif ()
+invert_on_off(GDAL_USE_PNG NOT_GDAL_USE_PNG)
+set(GDAL_USE_LIBPNG_INTERNAL
```
dg0yt, line 218: I would prefer if the variable was named `GDAL_USE_PNG_INTERNAL`, similar to `GDAL_USE_PNG`.

cmake/helpers/CheckDependentLibraries.cmake
```diff
@@ -173,114 +186,79 @@ endif ()
...
-      OFF
-      CACHE BOOL "Use internal libjpeg copy (if set to ON, has precedence over GDAL_USE_JPEG)")
+if (GDAL_USE_JPEG AND (JPEG_LIBRARY MATCHES ".*turbojpeg\.(so|lib)"))
+  message(
+    FATAL_ERROR
+      "JPEG_LIBRARY should point to a library with libjpeg ABI, not TurboJPEG. See https://libjpeg-turbo.org/About/TurboJPEG for the difference"
+    )
 endif ()
+invert_on_off(GDAL_USE_JPEG NOT_GDAL_USE_JPEG)
+set(GDAL_USE_LIBJPEG_INTERNAL
```
dg0yt, line 230: I would prefer if the variable was named `GDAL_USE_JPEG_INTERNAL`, similar to `GDAL_USE_JPEG`.

cmake/helpers/CheckDependentLibraries.cmake
```diff
@@ -173,114 +186,79 @@ endif ()
...
-  set(GDAL_USE_GIFLIB_INTERNAL
-      ON
-      CACHE BOOL "Use internal giflib copy (if set to ON, has precedence over GDAL_USE_GIF)")
-else ()
-  set(GDAL_USE_GIFLIB_INTERNAL
-      OFF
-      CACHE BOOL "Use internal giflib copy (if set to ON, has precedence over GDAL_USE_GIF)")
-endif ()
+invert_on_off(GDAL_USE_GIF NOT_GDAL_USE_GIF)
+set(GDAL_USE_GIFLIB_INTERNAL
```
dg0yt, line 236: I would prefer if the variable was named `GDAL_USE_GIF_INTERNAL`, similar to `GDAL_USE_GIF`.

cmake/helpers/CheckDependentLibraries.cmake
```diff
@@ -173,114 +186,79 @@ endif ()
...
-      CACHE BOOL "Use internal libjson-c copy (if set to ON, has precedence over GDAL_USE_JSONC)")
-  if (NOT GDAL_USE_LIBJSONC_INTERNAL)
-    message(FATAL_ERROR "GDAL_USE_JSONC or GDAL_USE_LIBJSONC_INTERNAL must be set to ON")
-  endif ()
-else ()
-  set(GDAL_USE_LIBJSONC_INTERNAL
-      OFF
-      CACHE BOOL "Use internal libjson-c copy (if set to ON, has precedence over GDAL_USE_JSONC)")
+invert_on_off(GDAL_USE_JSONC NOT_GDAL_USE_JSONC)
+set(GDAL_USE_LIBJSONC_INTERNAL
```
dg0yt, line 242: I would prefer if the variable was named `GDAL_USE_JSONC_INTERNAL`, similar to `GDAL_USE_JSONC`.

cmake/helpers/CheckDependentLibraries.cmake
```diff
@@ -298,15 +276,10 @@ option(GDAL_USE_LIBLERCV1_INTERNAL "Set ON to build mrf driver with internal lib
...
-  set(GDAL_USE_LIBLERC_INTERNAL
-      ON
-      CACHE BOOL "Use internal liblerc copy (if set to ON, has precedence over GDAL_USE_LERC)")
-else ()
-  set(GDAL_USE_LIBLERC_INTERNAL
-      OFF
-      CACHE BOOL "Use internal liblerc copy (if set to ON, has precedence over GDAL_USE_LERC)")
-endif ()
+invert_on_off(GDAL_USE_LERC NOT_GDAL_USE_LERC)
+set(GDAL_USE_LIBLERC_INTERNAL
```
dg0yt, line 280: I would prefer if the variable was named `GDAL_USE_LERC_INTERNAL`, similar to `GDAL_USE_LERC`.

dg0yt (commented):
After some thought, no substantial feedback from my side.

However, after the ZLIB/LIBZ inconsistency already created some problems, and with regard to grouping in CMake GUI tools, I would urge to name input variables for enabling internal libraries with the same prefix as input variables for the externals libs.

Given the common pattern, this might even become an option to `gdal_check_package`, or for clean code, become a function on its own (e.g. `gdal_internal_library`).

sgillies:
@rouault I'm going to get confused between CMake and runtime configuration options that use the same naming scheme. How about a prefix other than "GDAL_"?

cmake/helpers/CheckDependentLibraries.cmake
rouault: done

cmake/helpers/CheckDependentLibraries.cmake
rouault: done

cmake/helpers/CheckDependentLibraries.cmake
rouault: done

cmake/helpers/CheckDependentLibraries.cmake
rouault: done

cmake/helpers/CheckDependentLibraries.cmake
rouault: done

cmake/helpers/CheckDependentLibraries.cmake
rouault: done

cmake/helpers/CheckDependentLibraries.cmake
rouault: done

cmake/helpers/CheckDependentLibraries.cmake
rouault: done

rouault:
> I'm going to get confused between CMake and runtime configuration options that use the same naming scheme. How about a prefix other than "GDAL_"?

what would you suggest ? The GDAL_ prefix here helps to distinguish them from standard CMAKE_ variables or other standard ones without the CMAKE_ prefix (like the BUILD_SHARED_LIBS, BUILD_TESTING ones)

sgillies:
@rouault I don't know what to suggest. Anything other than `GDAL_` seems to go against the grain of CMake, yes? Maybe this is just the fun of using CMake, symbols that appear to be related in name but actually are not and depend on the context.

cmake/helpers/CheckDependentLibraries.cmake
```diff
@@ -14,24 +14,31 @@ include(FeatureSummary)
 include(DefineFindPackage2)
 include(CheckSymbolExists)
 
+option(
+  GDAL_USE_EXTERNAL_LIBS
+  "Whether detected external libraries should be used befault. This should be set before CMakeCache.txt is created." ON)
```
nyalldawson, line 19:
```suggestion
  "Whether detected external libraries should be used by default. This should be set before CMakeCache.txt is created." ON)
```

rouault: done

PR #5202, opened 2022-01-31
ILI IMDReader: fix various potential crashes on invalid output (CID 1469239, 1469241, 1469243, 1469244, 1469245)

rouault:
CC @pka

### DIFF
--- a/ogr/ogrsf_frmts/ili/imdreader.cpp
+++ b/ogr/ogrsf_frmts/ili/imdreader.cpp
@@ -39,10 +39,10 @@ CPL_CVSID("$Id$")
 
 typedef std::map<CPLString,CPLXMLNode*> StrNodeMap;
 typedef std::vector<CPLXMLNode*> NodeVector;
-typedef std::map<CPLXMLNode*,int> NodeCountMap;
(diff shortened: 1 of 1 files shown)


ogr/ogrsf_frmts/ili/imdreader.cpp
```diff
@@ -369,6 +396,23 @@ void ImdReader::ReadModel(const char *pszFilename) {
     NodeCountMap oAxisCount;
     NodeVector oArcLineTypes;
 
+    const auto TidLookup = [&oTidLookup](const char* pszKey)
```
nyalldawson, line 399: isn't this identical to the TidLookup function defined above?

nyalldawson: Ahh i see the distinction -- one is using a member lookup and the other the local captured one. (Renaming the member to have m prefix would help readability here!)

nyalldawson (approved):

ogr/ogrsf_frmts/ili/imdreader.cpp
rouault: > one is using a member lookup and the other the local captured one

The member lookup is in another class than the lambda one.

PR #5207, opened 2022-01-31
GPKG: add a NOLOCK=YES option to open a file without any lock (for read-only access

rouault:
helps fixing qgis/qgis#23991, but requires QGIS changes as well

### DIFF
--- a/autotest/ogr/ogr_gpkg.py
+++ b/autotest/ogr/ogr_gpkg.py
@@ -4154,6 +4154,64 @@ def test_ogr_gpkg_wal():
     gdal.Unlink(filename + '-wal')
     gdal.Unlink(filename + '-shm')
 
+###############################################################################
+# Test NOLOCK open option
+
+
+def test_ogr_gpkg_nolock():
+
+    def get_nolock(ds):
+        sql_lyr = ds.ExecuteSQL('SELECT nolock', dialect='DEBUG')
+        f = sql_lyr.GetNextFeature()
+        res = True if f[0] == 1 else False
+        ds.ReleaseResultSet(sql_lyr)
+        return res
+
+    # needs to be a real file
+    filename = 'tmp/test_ogr_gpkg_nolock.gpkg'
+
+    ds = gdaltest.gpkg_dr.CreateDataSource(filename)
+    lyr = ds.CreateLayer('foo')
+    f = ogr.Feature(lyr.GetLayerDefn())
+    lyr.CreateFeature(f)
+    f = None
+    ds = None
+
+    ds = gdal.OpenEx(filename, gdal.OF_VECTOR, open_options=['NOLOCK=YES'])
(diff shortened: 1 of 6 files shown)


nyalldawson (approved):

nyalldawson:
This is exciting!

The code looks good to me, and given that it's opt in it's safe to merge.

That said, boy it's hard to find any discussion/documentation about this option anywhere, and whether there's any hidden nasties it exposes. :open_mouth: ! But I did see that grass was investigating it here too: https://github.com/OSGeo/grass/pull/1667 . Was this PR a result of the GRASS change?

rouault:
> Was this PR a result of the GRASS change?

no, this is an independant finding.

I experimented with the following python script to explore locking issues:
```python
import os
import sqlite3

if os.path.exists('example.db'):
    os.unlink('example.db')

con = sqlite3.connect('example.db')
con.execute("CREATE TABLE foo(bar TEXT)")
con.execute("INSERT INTO foo VALUES ('1')")
con.commit()
con.close()

con = sqlite3.connect('example.db')
#con = sqlite3.connect('file:example.db?nolock=1', uri=True)
cur = con.cursor()
#cur.execute("PRAGMA journal_mode=WAL")
cur.execute("SELECT * FROM foo")

con2 = sqlite3.connect('example.db', timeout=0.5, isolation_level=None)
# The following will timeout unless WAL is enabled before the SELECT
# or con is opened in nolock=1 mode
con2.execute("INSERT INTO foo VALUES ('3')")

print(cur.fetchall())
```

The potential downside I'm aware of of the combination of this PR + the corresponding QGIS one is that read-only connections (opened in nolock mode) could potentially see corrupted content (or have sqlite returns an error) if reading while a file is modified. But that is restricted to connections before the file is edited, since when a file is edited in QGIS, WAL is enabled, and future read-only connections will cooperate with WAL and don't use nolock mode (since sqlite actually refuses to open a nolock connection on a WAL file as demonstrated by below snippet:)

```python
import os
import sqlite3

if os.path.exists('example.db'):
    os.unlink('example.db')

con = sqlite3.connect('example.db')
con.execute("PRAGMA journal_mode=WAL")
con.execute("CREATE TABLE foo(bar TEXT)")
con.execute("INSERT INTO foo VALUES ('1')")
con.commit()
con.close()

con2 = sqlite3.connect('file:example.db?nolock=1', uri=True)
cur2 = con2.cursor()
cur2.execute("SELECT * FROM foo") # exception: sqlite3.OperationalError: unable to open database file
```

nyalldawson:
> The potential downside I'm aware of of the combination of this PR + the corresponding QGIS one is that read-only connections (opened in nolock mode) could potentially see corrupted content (or have sqlite returns an error) if reading while a file is modified.

That sounds like a perfectly reasonable trade off to me.

PR #5208, opened 2022-01-31
CMake: fix installing drivers.ini in gdalplugins directory of build tree

rouault:


rouault:
Remaining test failures pre-existed this PR . 023ac422dc temporarily disables tests on build-windows-conda while this is investigated. This PR actually helps, so merging it

PR #5209, opened 2022-01-31
GTiff: fix exposing WEBP_LOSSLESS option

rouault:


rouault:
d8b3824 backported to 3.4

PR #5212, opened 2022-02-01
CMake: make it use CMake config files in priority to find PROJ

### LINKED ISSUE #5210: CMake builds and PROJ >= 8.2.0 static builds on Windows
It seems our CMake conf can't work with PROJ >= 8.2.0 static builds on Windows , due to missing PROJ_DLL= definition that is required since https://github.com/OSGeo/PROJ/commit/6e9b324ab7bf5909df7e68409e060282db14fa54

rouault:
This fixes building against a static build of PROJ >= 8.2.0 on Windows

Fixes #5210


alg/CMakeLists.txt
```diff
@@ -71,8 +71,9 @@ if (GDAL_USE_OPENCL)
   target_sources(alg PRIVATE gdalwarpkernel_opencl.h gdalwarpkernel_opencl.cpp)
 endif ()
 
-target_include_directories(alg PRIVATE ${PROJ_INCLUDE_DIRS})
-gdal_add_private_link_libraries(PROJ::PROJ)
+target_include_directories(alg PRIVATE $<TARGET_PROPERTY:PROJ::proj,INTERFACE_INCLUDE_DIRECTORIES>)
+target_compile_definitions(alg PRIVATE $<TARGET_PROPERTY:PROJ::proj,INTERFACE_COMPILE_DEFINITIONS>)
+gdal_add_private_link_libraries(PROJ::proj)
```
dg0yt, line 76: Why are the extra `target_..` needed (here and in other spots)? 
A key point of imported target is to carry such transititive usage requiremens just by linking them.

cmake/helpers/CheckDependentLibraries.cmake
```diff
@@ -195,7 +195,11 @@ if (GDAL_USE_CRYPTOPP)
   option(CRYPTOPP_USE_ONLY_CRYPTODLL_ALG "Use Only cryptoDLL alg. only work on dynamic DLL" OFF)
 endif ()
 
-find_package(PROJ 6.0 REQUIRED)
+# First check with CMake config files, and then fallback to the FindPROJ module.
+find_package(PROJ 8.0 CONFIG)
+if (NOT TARGET PROJ::proj)
```
dg0yt, line 200: The key indicator of success is `PROJ_FOUND`.

alg/CMakeLists.txt
rouault: > Why are the extra target_.. needed (here and in other spots)?
> A key point of imported target is to carry such transititive usage requiremens just by linking them.

If I remember well my past experiments, that's related to the fact that "alg" is an object library and that our minimum CMake is 3.10 and that there are restrictions on versions < 3.12 or 3.13 related to the linking of an object library.

For example with 3.10, if I do:
target_link_libraries(alg PRIVATE PROJ::proj)

I get
CMake Error at alg/CMakeLists.txt:78 (target_link_libraries):
  Object library target "alg" may not link to anything.

whereas it works with 3.16

But using the gdal_target_link_libraries() function used in drivers actually solves this here for this non-driver component
gdal_target_link_libraries(TARGET alg LIBRARIES PROJ::proj)

I'll switch to that

cmake/helpers/CheckDependentLibraries.cmake
rouault: done

PR #5222, opened 2022-02-02
Add page for GDAL Python Utilities

maphew:
## What does this PR do?

As per "[gdal-dev] gdal-utils: About GDAL Python Utilities" thread, this adds an overview and intro page for the gdal-utils sub project.

## Tasklist

 - [x] Add documentation


rouault:
CC @idanmiara

doc/source/user/gdal_python_utilities.rst
```diff
@@ -0,0 +1,62 @@
...
+GDAL Python Utilities
+================================================================================
+
+The GDAL python utilities are included with GDAL. If you've installed
+GDAL you already have them. However you may want to use a newer or older
+version of the utilities without changing GDAL. This is where
+**gdal-utils** comes in.
+
+**gdal-utils**: is the GDAL Python Utilities *distribution*. This is
+what you install. It's home page is https://pypi.org/project/gdal-utils/
```
rouault, line 13:
```suggestion
what you install. Its home page is https://pypi.org/project/gdal-utils/
```

doc/source/user/gdal_python_utilities.rst
```diff
@@ -0,0 +1,62 @@
...
+. Install it with ``pip install gdal-utils`` and it will up or downgrade
```
rouault, line 14: The "it will up or downgrade ..." part. is a bit mysterious on reading  Is it just refering to pip standard behavior ? However what could be mentioned is the interaction with the "gdal" pip package.  gdal-utils and gdal pip packages both provide the GDAL utilities so depending in the order they are installed, overwriting could occur. as mentionned in the RFC

doc/source/user/gdal_python_utilities.rst
```diff
@@ -0,0 +1,62 @@
...
+Commonly used utilities include:
+
+-  gdal_merge
+-  gdal_edit
+-  gdal_calc
+-  ogrmerge
+
+For the full list see
+`Programs <https://gdal.org/programs/index.html#programs>`__ on the GDAL
+website and note the ones that end in ``.py``.
```
rouault, line 30: better use Sphinx anchors (see https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#role-ref).   The programs page has the ``programs`` anchor (see first line of https://raw.githubusercontent.com/OSGeo/gdal/master/doc/source/programs/index.rst)
```suggestion
:ref:`Programs <programs>` and note the ones that end in ``.py``.
```

doc/source/user/gdal_python_utilities.rst
```diff
@@ -0,0 +1,62 @@
...
+
+Developers
+----------
+
+Read the gdal-utils project charter, `RFC-78 <https://gdal.org/development/rfc/rfc78_gdal_utils_package.html>`__.
```
rouault, line 35: Same as above. using Sphinx anchors:
```suggestion
Read the :ref:`gdal-utils project charter <rfc-78>`.
```

doc/source/user/gdal_python_utilities.rst
```diff
@@ -0,0 +1,62 @@
...
+Clone or download the gdal project: https://github.com/OSGeo/gdal/
+
+In your IDE set gdal-utils as the root folder,
+`.../swig/python/gdal-utils <https://github.com/OSGeo/gdal/tree/master/swig/python/gdal-utils/>`__.
+
+   **./osgeo_utils** - contains the Programs (those scripts that have
+   launch wrappers created by pip and added to PYTHONHOME/Scripts)
+
+   **./osgeo_utils/samples** - working python scripts but not typically
+   avaialble in path (run them with ``python
```
rouault, line 46:
```suggestion
   available in path (run them with ``python
```

doc/source/user/gdal_python_utilities.rst
```diff
@@ -0,0 +1,62 @@
...
+   path/to/samples/something.py``)
+
+Discuss via `gdal-dev mailing
+list <https://lists.osgeo.org/mailman/listinfo/gdal-dev>`__.
+
```
rouault, line 51: Not wrong, but I would just drop this. We could put it on every page...

```suggestion
```

doc/source/user/gdal_python_utilities.rst
maphew: Yeah I struggle with this sentence. The goal is trying to communicate that "whichever is last most recenty installed wins" without adding as many sentences as in the RFC. Maybe it's best just to leave it as simply _"Install with ``pip install gdal-utils``"_ as the concept of up or downgrading is already introduced in the paragraph above.

doc/source/user/gdal_python_utilities.rst
maphew: thanks!

idanmiara:
Looks good @maphew, thanks!
You could also add comments about windows wrapper batch files as discussed in the following thread:
https://www.mail-archive.com/gdal-dev@lists.osgeo.org/msg37290.html

maphew:
@idanmiara there's a couple wrapper approached discussed in that thread. Any particular one you're thinking of? Or just include both?

PR #5226, opened 2022-02-03
OGR Doc: Describe the use of multiple debug prefixes

zstadler:
## What does this PR do?
Improve the documentation of the [Common options for vector programs](https://gdal.org/programs/vector_common_options.html#common-options-for-vector-programs) by adding a description how multiple categories can be selected with the [`--debug` option](https://gdal.org/programs/vector_common_options.html#cmdoption-debug)

No

## Tasklist

 - [x] Add documentation
 - [x] Review
 - [x] Adjust for comments

### DIFF
--- a/doc/source/programs/vector_common_options.rst
+++ b/doc/source/programs/vector_common_options.rst
@@ -51,6 +51,7 @@ All GDAL OGR command line programs support the following common options.
     A value of OFF will disable all debug messages.
     Another value will select only debug messages containing
     that string in the debug prefix code.
+    Multiple prefixes need to be separated by `|`.
 
 .. option:: --help-general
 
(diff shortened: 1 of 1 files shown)


doc/source/programs/vector_common_options.rst
```diff
@@ -51,6 +51,7 @@ All GDAL OGR command line programs support the following common options.
     A value of OFF will disable all debug messages.
     Another value will select only debug messages containing
     that string in the debug prefix code.
+    Multiple prefixes need to be separated by `|`.
```
rouault, line 54: Not really. The doc of the CPLDebug() function mentions "... Otherwise only messages whose category appears
 somewhere within the CPL_DEBUG value are displayed (as determined by strstr())."  . The sentence "Another value will select only debug messages containing hat string in the debug prefix code." should rather be amended to reflect that.

zstadler: For future reference, the before-mentioned docs is available at

https://github.com/OSGeo/gdal/blob/2cb831638ea63c3cf38c148b46b4887869231f3d/port/cpl_error.cpp#L579-L602

zstadler: The above has been coped and rephrased

zstadler:
@rouault Please check-out the updated phrasing.

PR #5234, opened 2022-02-04
Warped VRT opening: do not open the source dataset with GDAL_OF_SHARED

rouault:
This makes it easier to possible to open multiple times the same warped
VRT and then distribute each warped VRT handle to different threads.


EvertEt:
Minor: It looks like it's `GDAL_OF_SHARED` that's removed, not `GDAL_OF_RASTER`

rouault:
> It looks like it's `GDAL_OF_SHARED` that's removed, not `GDAL_OF_RASTER`

indeed, commit message and PR title amended

PR #5235, opened 2022-02-04
GDALAutoCreateWarpedVRT(): clarify source dataset ownership with example

rouault:


sgillies:
@vincentsarago can you think about what this means for rasterio's WarpedVRT? There may be some implications in https://github.com/OSGeo/gdal/pull/5234, too.

vincentsarago:
@sgillies no idea.

rouault:
> can you think about what this means for rasterio's WarpedVRT?

my understanding of the example
```
    >>> with rasterio.open('tests/data/RGB.byte.tif') as src:
    ...     with WarpedVRT(src, crs='EPSG:3857') as vrt:
    ...         data = vrt.read()
    """
```
is that the source dataset has a longer scope of existence than the warped VRT, so you're in the "Traditonal nested calls" case, where you close first the warped VRT and then the source dataset.

PR #5244, opened 2022-02-06
Fix LICENSE.txt symlink

dg0yt:
Hits me on every `grep`...


rouault:
I bet we can just remove that file now (the .github/LICENSE.TXT symlink). It was useful at the time where LICENSE.TXT was in the gdal/ subdirectory and github wouldn't find it automatically

rouault:
I've just deleted the .github/LICENSE.TXT symlink

PR #5250, opened 2022-02-06
CMake: tidy linking of apps

rouault:


dg0yt:
Somewhat related:
I'm struggling with new linking issues related to recent changes to `gdal_add_target_libraries`. I try to understand the added value of this function. All I found is that CMake < 3.12 does not support object libraries (such as `ogr`) with `target_link_libraries`, making it harder to carry usage requirements. And there are even public usage requirement for json and geos headers.

rouault:
> I'm struggling with new linking issues related to recent changes to `gdal_add_target_libraries`. 

like ?

> I try to understand the added value of this function. All I found is that CMake < 3.12 does not support object libraries (such as `ogr`) with `target_link_libraries`, making it harder to carry usage requirements.

Yes, I believe that's the main purpose of that function.

> And there are even public usage requirement for json and geos headers.

Can you be more explicit ?

dg0yt:
Disclaimer: This reflects the WIP packaging attempt with CMake for vcpkg. It is more a feasibility study than best practice.

> > I'm struggling with new linking issues related to recent changes to `gdal_add_target_libraries`.
> 
> like ?

The find modules in GDAL cannot fully deal with all the variety of debug+release config, import or debug suffixes, transitive usage requirements. But there are config files. (GDAL now uses that for PROJ, but there is also json-c, geos, expat, zstd.) To avoid heavy patching, my initial successful attempt created ALIAS targets in a CMAKE_PROJECT_INCLUDE. Now this doesn't work anymore because the buildsystems cannot get the target properties from the alias. And even with an IMPORTED library, the properties wouldn't be visible because they would be behind an INTERFACE_LINK_LIBRARIES step.

> > And there are even public usage requirement for json and geos headers.
> 
> Can you be more explicit ?

Probably this is resolved. The relevant headers  `include/cpl_json_header.h` and `include/ogr_geos.h` don't  seem to be installed anymore. While building HEAD, I was looking at old file lists from the name build.

dg0yt:
> `include/ogr_geos.h`

`apps/ogrlineref.cpp` is using it and caused build errors.
But if this header is not public API, it is probably better to solved it just for this app.

rouault:
> The find modules in GDAL cannot fully deal with all the variety of debug+release config, import or debug suffixes, transitive usage requirements. But there are config files. (GDAL now uses that for PROJ, but there is also json-c, geos, expat, zstd.) 

The approach of using config files when possible seem reasonable to me. Would probably be doable for recent GEOS CMake builds. For others, I'm not sure if they provide CMake config files.

> To avoid heavy patching, my initial successful attempt created ALIAS targets in a CMAKE_PROJECT_INCLUDE. Now this doesn't work anymore because the buildsystems cannot get the target properties from the alias. And even with an IMPORTED library, the properties wouldn't be visible because they would be behind an INTERFACE_LINK_LIBRARIES step.

hum still struggling to fully understand your issues due to my unsufficient CMake mastering. I'm open to suggestions/fixes... Would making gdal_target_link_libraries redirect to standard target_link_libraries if CMake >= 3.12 help solving issues ? We make CMake 3.10 the baseline for "basic" builds, but we can conditionally use CMake features of higher versions if needed.

dg0yt:
> > 
> 
> The approach of using config files when possible seem reasonable to me. Would probably be doable for recent GEOS CMake builds. For others, I'm not sure if they provide CMake config files.

The barrier is the non standard `gdal_check_package` function - the config files have different package names and target names. This is what I do for geos:

~~~cmake
if(GDAL_USE_GEOS)
    find_package(GEOS CONFIG REQUIRED)
    add_library(GEOS::GEOS ALIAS GEOS::geos_c)
endif()
~~~

The side effect of the alias is that CMake puts GEOS::geos_c into the exported config. I still have to make the gdal config load the geos config. But I don't need to deal with target names again.

> hum still struggling to fully understand your issues due to my unsufficient CMake mastering. I'm open to suggestions/fixes... Would making gdal_target_link_libraries redirect to standard target_link_libraries if CMake >= 3.12 help solving issues ? We make CMake 3.10 the baseline for "basic" builds, but we can conditionally use CMake features of higher versions if needed.

Basically trying to rebuild what cmake is doing is often incomplete, with regard to build types or to transitive requirements. (This is also a problem for creating proper pc files.)

I will create PRs, I just jumped onto this PR because it seemed to touch related problems.

rouault:
> ```cmake
>     add_library(GEOS::GEOS ALIAS GEOS::geos_c)
> ```

ok, so the proper solution here would be do replace all GEOS::GEOS by GEOS::geos_c everywhere ? Or if we anticipate that a dependency might have varying target name (that would be odd though), use it through a ${GEOS_TARGET_NAME} ?

>  I just jumped onto this PR because it seemed to touch related problems.

ok, does this PR seems to be in the right direction and OK for merge ?

apps/CMakeLists.txt
```diff
@@ -134,18 +134,13 @@ if (BUILD_APPS)
 
     # Now external libs
     if (GDAL_USE_GEOS)
-      target_include_directories(${UTILCMD} PRIVATE ${GEOS_INCLUDE_DIRS})
+      target_include_directories(${UTILCMD} PRIVATE $<TARGET_PROPERTY:GEOS::GEOS,INTERFACE_INCLUDE_DIRECTORIES>)
```
dg0yt, line 137: This is the spot which I don't like. 
- If the GEOS include directory would be really needed for all users fo lib gdal, it should already come as a public link library with lib gdal.
- If the GEOS include directory is needed for all apps here, I think it should be `target_link_libraries(${UTILCMD} PRIVATE GEOS::GEOS)`.
- If it is just `ogrlineref.cpp`, `target_link_libraries(...) should be called directly for the app which uses this source file.
- If it is just `ogrlineref.cpp`, it might be entirely obsolete after reviewing that file. Does the file use more than the GEOS version macros? Does GDAL still support older versions? (I really don't know.)

apps/CMakeLists.txt
```diff
@@ -134,18 +134,13 @@ if (BUILD_APPS)
...
       target_compile_definitions(${UTILCMD} PRIVATE -DHAVE_GEOS=1)
     endif ()
     if (NOT GDAL_USE_JSONC_INTERNAL)
-      target_include_directories(${UTILCMD} PRIVATE ${JSONC_INCLUDE_DIRS})
-      target_link_libraries(${UTILCMD} PRIVATE ${JSONC_LIBRARIES})
+      target_link_libraries(${UTILCMD} PRIVATE JSONC::JSONC)
```
dg0yt, line 141: Basically, the same considerations apply as for GEOS, minus the `ogrlineref` aspect.
But if the include directory isn't needed at all, JSONC should already come as a private link library with lib gdal.

apps/CMakeLists.txt
rouault: GEOS 3.2 is really antiquated for GDAL master. I believe the HAVE_GEOS_PROJECT in this file could just be replaced by a simple HAVE_GEOS that we could pass with a target_compile_options, and that would remove the need for including ogr_geos.h.  It seems this is the only utility needed explicit use of GEOS

apps/CMakeLists.txt
rouault: it seems only the "lib" part of the apps/ needs JSONC, not the binary part. So this part could be removed.

ok, I'll improve this PR with that and the above consideration

rouault: done

dg0yt (approved):
LGTM.

PR #5254, opened 2022-02-06
cmake: Changes to Allow vsihdfs to Build on Ubuntu

jamesmcclain:
## What does this PR do?

This PR add some changes to allow the `vsihdfs` virtual filesystem to build on Ubuntu.  On a freshly-created Ubuntu VM (Lubuntu 21.10), I encountered problems related to the fact that `libhdfs` depends on `libjvm`, so when linking against the former it is required to link against the latter (at least on Ubuntu).

I have made some changes to `findHDFS.cmake` to (try to) force it to look for `libjvm` and link against it.  These changes succeeded for me, although I understand that they might not be appropriate for other systems (I am guessing, but it seems as though MacOS does not behave the same way).  This PR should be consider **work in progress** (and/or closable without offence) because of the fact (1) that it was not test on Windows or MacOS and (2) the fact that the idea itself may not have been correctly implemented (`cmake` is fairly new to me).

I am also happy to keep working on it in response to review comments if there is interest.

See also: https://github.com/OSGeo/gdal/pull/5112

## Tasklist

 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

Tested in the following environment

* OS:  Lubuntu 21.10
* Compiler: gcc 11.2.0
* cmake: 3.18.4
* openjdk 18-ea 2022-03-15

Testing steps taken:

```bash
export JAVA_HOME=/usr/lib/jvm/java-18-openjdk-amd64/
export LD_LIBRARY_PATH="/usr/lib/jvm/java-18-openjdk-amd64/lib/server/"
mkdir build ; cd build
cmake -DCMAKE_BUILD_TYPE=Release -DHDFS_INCLUDE_DIR=$HOME/local/hadoop-3.3.1/include \
 -DHDFS_LIBRARY=$HOME/local/hadoop-3.3.1/lib/native/libhdfs.so -DGDAL_USE_HDFS=ON \
 -DCMAKE_INSTALL_PREFIX=$HOME/local/gdal -DCMAKE_LIBRARY_PATH="/usr/lib/jvm/java-18-openjdk-amd64/lib/server/" \
 -DBUILD_PYTHON_BINDINGS=OFF \
 ..
make
make install
```
then
```bash
export JAVA_HOME=/usr/lib/jvm/java-18-openjdk-amd64/
export CLASSPATH=$($HOME/Desktop/hadoop-3.3.1/bin/hadoop classpath --glob)
...
~/local/gdal/bin/gdalinfo /vsihdfs/file:/tmp/sample.tif
```
and the last command succeeded.

I should also note that I had to disable the python bindings in order for this to succeed, but that might be unrelated.


cmake/modules/packages/FindHDFS.cmake
```diff
@@ -24,23 +24,38 @@ The following vars are set if HDFS (the native library provided with Hadoop) is
 
 include(FindPackageHandleStandardArgs)
 
+find_library(JAVA_JVM_LIBRARY NAMES jvm PATH_SUFFIXES native)
```
rouault, line 27: it seems to me that you should rather use https://cmake.org/cmake/help/latest/module/FindJNI.html and
do a find_package(JNI) to get this variable

cmake/modules/packages/FindHDFS.cmake
```diff
@@ -24,23 +24,38 @@ The following vars are set if HDFS (the native library provided with Hadoop) is
...
 find_path(HDFS_INCLUDE_DIR NAMES hdfs.h)
 find_library(HDFS_LIBRARY NAMES hdfs PATH_SUFFIXES native)
 
+mark_as_advanced(JAVA_JVM_LIBRARY)
```
rouault, line 31: useless if using find_package(JNI)

cmake/modules/packages/FindHDFS.cmake
```diff
@@ -24,23 +24,38 @@ The following vars are set if HDFS (the native library provided with Hadoop) is
...
 mark_as_advanced(HDFS_INCLUDE_DIR HDFS_LIBRARY)
+
+find_package_handle_standard_args(JVM
```
rouault, line 34: useless if using find_package(JNI)

cmake/modules/packages/FindHDFS.cmake
```diff
@@ -24,23 +24,38 @@ The following vars are set if HDFS (the native library provided with Hadoop) is
...
+                                  FOUND_VAR JVM_FOUND
+                                  REQUIRED_VARS JAVA_JVM_LIBRARY)
+
 find_package_handle_standard_args(HDFS
                                   FOUND_VAR HDFS_FOUND
-                                  REQUIRED_VARS HDFS_LIBRARY HDFS_INCLUDE_DIR)
+                                  REQUIRED_VARS JVM_FOUND HDFS_LIBRARY HDFS_INCLUDE_DIR)
```
rouault, line 40: this find_package_handle_standard_args() should rather be conditionalized by if(JNI_FOUND) if using find_package(JNI)

cmake/modules/packages/FindHDFS.cmake
```diff
@@ -24,23 +24,38 @@ The following vars are set if HDFS (the native library provided with Hadoop) is
...
 
 if(HDFS_FOUND)
     set(HDFS_LIBRARIES ${HDFS_LIBRARY})
     set(HDFS_INCLUDE_DIRS ${HDFS_INCLUDE_DIR})
 
     if(NOT TARGET HDFS::HDFS)
+
+        add_library(JVM::JVM UNKNOWN IMPORTED)
```
rouault, line 48: to be adapted with the variables of https://cmake.org/cmake/help/latest/module/FindJNI.html

rouault:
> so when linking against the former it is required to link against the latter (at least on Ubuntu).

don't you like a change in port/CMakeLists.txt to link against JVM::JVM in addition to HDFS::HDFS ?

jamesmcclain:
> > so when linking against the former it is required to link against the latter (at least on Ubuntu).
> 
> don't you need a change in port/CMakeLists.txt to link against JVM::JVM in addition to HDFS::HDFS ?

That is probably the correct thing to do, but I did not.  I think the addition of the second `add_library` (the one for `libjvm`) causes that library to be linked whenever `libhdfs` is.

rouault:
> t. I think the addition of the second `add_library` (the one for `libjvm`) causes that library to be linked whenever `libhdfs` is

no, it won't unless I'm missing something. It just defines an imported library, but this shouldn't make it to be linked, if you don't add it explicitly as a dependency to another library.
In that state, I'd say those changes should have no effect on the final linking.

jamesmcclain:
> > t. I think the addition of the second `add_library` (the one for `libjvm`) causes that library to be linked whenever `libhdfs` is
> 
> no, it won't unless I'm missing something. It just defines an imported library, but this shouldn't make it to be linked, if you don't add it explicitly as a dependency to another library. In that state, I'd say those changes should have no effect on the final linking.

Okay, that makes sense.  I'll try to return to this shortly.

jamesmcclain:
Okay, I have updated `FindHDFS.cmake` to to use `FindJNI`, now.  I have also updated `port/CMakeLists.txt` to encode a dependency on `libjvm` if it has been found and `libhdfs` is a dependency.

An additional step to solve the problem that I encountered when testing the build would be to emit a warning or error about `libjvm` not being found when `libhdfs` is a dependency on the systems where the latter depends on the former.

rouault:
thanks

jamesmcclain:
> thanks

You are welcome, thank you for looping me in and reviewing the PR.

PR #5257, opened 2022-02-07
[FEATURE] Add (initial) support Int64 and UInt64 raster data types

### LINKED ISSUE #3325: ENH: Add GDT_Int64 & GDT_UInt64 Data Types
Forward from:
- https://trac.osgeo.org/gdal/ticket/6879
- https://github.com/mapbox/rasterio/issues/2072
- https://github.com/Toblerity/Fiona/issues/365

This comes from use cases I have seen recently:
- https://github.com/corteva/rioxarray/issues/195
- https://github.com/corteva/geocube/discussions/52

Bringing this up again in case there is interest in adding this, but was forgotten.

rouault:
Fixes #3325

- Add GDT_Int64 and GDT_UInt64 data types and handle them in MEM, GTiff, netCDF and Zarr drivers
- Adapt gdal_rasterize to be able to burn a OFTInteger64 vector field into a GDT_Int64 raster losslessly
- Add methods to set/get nodata value as Int64/UInt64
    
    - GDALRasterBand:
        virtual int64_t GetNoDataValueAsInt64( int *pbSuccess = nullptr );
        virtual uint64_t GetNoDataValueAsUInt64( int *pbSuccess = nullptr );
        virtual CPLErr SetNoDataValueAsInt64( int64_t nNoData );
        virtual CPLErr SetNoDataValueAsUInt64( uint64_t nNoData );
    
      and corresponding C API and Python bindings
    
    - GDALMDArray:
        int64_t GetNoDataValueAsInt64(bool* pbHasNoData = nullptr) const;
        uint64_t GetNoDataValueAsUInt64(bool* pbHasNoData = nullptr) const;
        bool SetNoDataValue(int64_t nNoData);
        bool SetNoDataValue(uint64_t nNoData);
    
      and corresponding C API and Python bindings
    
    Adapt:
    - gdalinfo
    - gdal_translate
    - gdal_create
    - gdal_edit.py
    - MEM, GTIFF, VRT, netCDF and ZARR drivers

Besides that Int64 and UInt64 support might be a bit rough due to a lot of remaining places inside GDAL using double as the pivot data type. So in particular, if the range of values in the raster exceed what double authorize to store losslessly an integer, you won't get exact results from warping and overview, min and max values from statistics, etc etc etc etc etc etc etc


rouault:
CC @snowman2

snowman2:
Thanks @rouault!

> Besides that Int64 and UInt64 support might be a bit rough due to a lot of remaining places inside GDAL using double as the pivot data type.

What are your thoughts on raising an error in the scenario that the Int64 nodata value cannot be safely converted to a double when the pivot is necessary? This idea came from this check rasterio has on data types: https://github.com/rasterio/rasterio/blob/master/rasterio/dtypes.py#L149-L174

rouault:
> What are your thoughts on raising an error in the scenario that the Int64/UInt64 nodata value cannot be safely converted to a double when the pivot is necessary?

already implemented (as a warning) in https://github.com/OSGeo/gdal/pull/5257/files#diff-8ec18576e00ae96f5d026b3cb557e526eb2f6553cb84e9323b131d3c8df2a876R4308 ( GDALGetNoDataValueCastToDouble()  methods )

snowman2:
> already implemented (as a warning) in https://github.com/OSGeo/gdal/pull/5257/files#diff-8ec18576e00ae96f5d026b3cb557e526eb2f6553cb84e9323b131d3c8df2a876R4308 ( GDALGetNoDataValueCastToDouble() methods )

Nice! It is a big diff, so I am probably missing quite a bit. That is definitely helpful and headed in the right direction.

My current opinion is that in this scenario an error would be a better fit as warnings are often lost in logs when a process successfully completes. An error would force the user to update their nodata value (or use a mask) before running the operation and would prevent creating corrupted datasets. However, since you are implementing it, you get the final say :smile: .

snowman2:
Thanks @rouault 👍

PR #5265, opened 2022-02-08
Spatialite: remove support for libspatialite < 4.1.2 (fixes #4244)

### LINKED ISSUE #4244: Cleanup: remove support for spatialite < 4.1.2
spatialite < 4.1.2 is now ancient, and as we have a number of SPATIALITE_412_OR_LATER tests in the code base, it is time to clean that up

rouault:


rouault:
The failure in the appveyor build is "interesting". So our nmake config didn't define SPATIALITE_412_OR_LATER at all, despite using libspatialite 5.0.1 now. So it seems that with those builds using the "new" spatialite initialization has some side effect we didn't have before. But that's true only for that build. The CMake Windows conda build using libspatialite 5.0.1 too pass the failing test OK

dg0yt:
> The failure in the appveyor build is "interesting". So our nmake config didn't define SPATIALITE_412_OR_LATER at all, despite using libspatialite 5.0.1 now.

Is there a mix of gdal versions, with `c:\hostedtoolcache\windows\python\3.7.9\x86\lib\site-packages\gdal-3.3.0-py3.7-win32.egg\osgeo\gdal.py` next to that error?

rouault:
> Is there a mix of gdal versions, with `c:\hostedtoolcache\windows\python\3.7.9\x86\lib\site-packages\gdal-3.3.0-py3.7-win32.egg\osgeo\gdal.py` next to that error?

no, that message is expected given that in master swig/python/setup.py has still gdal_version=3.3.0 and that we run "python setup.py install". The issue here must be more subtle, perhaps related how spatialite is built in the gistinternals SDK we use

Issue #3555, opened 2021-03-08
GDAL 3.5: remove deprecated drivers and Perl bindings


rouault:
See https://lists.osgeo.org/pipermail/gdal-dev/2021-March/053590.html .
Except that ZMAP was removed from deprecation per #4086
Maybe except DODS: ticket about DODS driver: https://github.com/OSGeo/gdal/issues/5173
Add MDB to the list: https://github.com/OSGeo/gdal/issues/4311

idanmiara:
Not that I use the Perl bindings, but I didn't see any correspondence for that in the attached email.
Are they going to be deprecated ?

rouault:
See https://github.com/OSGeo/gdal/pull/3541

rouault:
from from https://github.com/OSGeo/gdal/issues/5291 it seems GMT raster is still in use..

PR #5269, opened 2022-02-09
Remove deprecated drivers planned for removal in 3.5

rouault:
Refs #3555

Elements (code, doc, tests) from those removed drivers transfered to the OSGeo/gdal-extra-drivers repository per https://github.com/OSGeo/gdal-extra-drivers/pull/1


rouault:
from https://github.com/OSGeo/gdal/issues/5291 it seems GMT raster is still in use..

normanb:
👍 as the original author of the cloudant driver I had no requests for any changes or indication of its use for some time now.

hobu (approved):
IMO we want one more approval besides myself on this.

PR #5271, opened 2022-02-09
Use CMake config and targets for dependencies

dg0yt:
## What does this PR do?

Adds declaration of known CMake config and targets to `gdal_check_package`. Cache variables can be used to change the defaults.

Removes ignoring the `COMPONENTS` argument when `CONFIG` is given.

Update `gdal_check_package` documentation.

Puts corresponding `find_dependency` calls to the exported CMake config (mandatory at least for static linking).

## Tasklist

 - [x] Add `find_depdendency` for targets created from Find modules.
 - [ ] Add test case(s) -> #5283
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

# Notes

Manually tested with vcpkg toolchain (x64-linux).

The key changes are `gdal.cmake`.

The new cache variables could be used for every `<Pkg>` when using `<Pkg>_LIBRARIES` for linking. This would help in dealing with "unofficial exports" as created by vcpkg without needing to patch gdal. However, I didn't want to clutter this PR with link libraries changes in this early state, before feedback.


rouault:
That looks good to me. There's just one minor thing I noticed in the output of the build-linux-ubuntu-focal job : https://github.com/OSGeo/gdal/runs/5121594632?check_suite_focus=true
In the "The following OPTIONAL packages have not been found:" section, it reports "expat". And in the "The following RECOMMENDED packages have been found:", it reports "EXPAT".
 Likely because Expat wasn't found through the config way but through the FindExpat way:
```
2022-02-09T07:41:45.2268644Z -- Could NOT find expat (missing: expat_DIR)
2022-02-09T07:41:45.2847441Z -- Found EXPAT: /usr/lib/x86_64-linux-gnu/libexpat.so (found version "2.2.9") 
```
So this is logical, but a bit confusing for end users not aware of the subtleties of how packages are discovered. Same issue with GeoTIFF, json-c  Probably the hacks around https://github.com/OSGeo/gdal/blob/master/gdal.cmake#L724 should be extended

dg0yt:
Feature summary: 
- We need to deal both with configs that aren't found and with Find modules which are found in addition to configs due to transitive lookups.
  In the context of alternatives, I will try to find an adequate representation of `gdal_check_package`.
- I would prefer to handle this mostly in `gdal_check_package`, either by modifying the properties directly or by recording additional properties which are read just before feature summary.

rouault:
> * I would prefer to handle this mostly in `gdal_check_package`, either by modifying the properties directly or by recording additional properties which are read just before feature summary.

sounds better indeed

dg0yt:
I moved a little bit close to `find_package` behaviour by just using the `NAMES` and `CONFIG` options that it provides. This resolves the new feature summary quirk.

(I do have other feature summary comments/ideas but they are for separate PRs.)

PR #5273, opened 2022-02-09
CMake: make get_dep_libs() work with several library names passed to a _LIBRARY variable (fixes #5272), and also fix variable inversion for INTERFACE_LIBRARY

### LINKED ISSUE #5272: [CMake] Multiple libraries for one dependency
I noticed a small "bug". If a dependency consists of several library files, the function `split_library_to_cflags` does not work anymore.

This can happen if, for example, GEOS is linked statically. If you specify `-D GEOS_LIBRARY=D:\GEOS\lib\geos.lib;D:\GEOS\lib\geos_c.lib` as CMake parameter, you get an error message:
```
CMake Error at cmake/modules/thirdparty/SplitLibraryToCFlags.cmake:5 (get_filename_component):
  get_filename_component unknown component
  D:/GEOS/lib/geos_c.lib
```

A patch would be for example
```
diff --git a/cmake/modules/thirdparty/SplitLibraryToCFlags.cmake b/cmake/modules/thirdparty/SplitLibraryToCFlags.cmake
index 8d60902071..929e71f99a 100644
--- a/cmake/modules/thirdparty/SplitLibraryToCFlags.cmake
+++ b/cmake/modules/thirdparty/SplitLibraryToCFlags.cmake
@@ -1,7 +1,8 @@

-function(split_library_to_cflags _lib _result)
-    if(_lib)
+function(split_library_to_cflags _libs _result)
+    if(_libs)
         set(RESULT)
+               foreach(_lib ${_libs})
         get_filename_component(_lib_name ${_lib} NAME_WE)
         if(_lib_name STREQUAL "")
         else()
@@ -26,6 +27,7 @@ function(split_library_to_cflags _lib _result)
                 endif()
             endif()
         endif()
+               endforeach()
         set(${_result} "${RESULT}" PARENT_SCOPE)
     endif()
 endfunction()
```

rouault:
Patch by @OgreTransporter


dg0yt (commented):
Note that this requires double quotes to pass a list to `_libs`, inviting errors.
The alternative is a signature
~~~cmake
function(split_library_to_cflags _result)
~~~
with libs taken from `${ARGN}`


Tested as:
~~~
function(split_library_to_cflags _libs _result)
	foreach(_lib IN LISTS _libs)
		message(STATUS "${_lib}")
	endforeach()
	message(STATUS "--")
endfunction()
function(split_library_to_cflags2 _result)
	foreach(_lib IN LISTS ARGN)
		message(STATUS "${_lib}")
	endforeach()
	message(STATUS "--")
endfunction()

set(foo "A B" C "D;E")
split_library_to_cflags(${foo} bar)
split_library_to_cflags("${foo}" bar)
split_library_to_cflags2(bar ${foo})
split_library_to_cflags2(bar "${foo}")
~~~

rouault:
> Note that this requires double quotes to pass a list to `_libs`, inviting errors.

The only callers to this are in https://github.com/OSGeo/gdal/blob/8e8cdf32a4bdeaa7d23bb0b33189f39e64092cd8/cmake/helpers/GenerateConfig.cmake and pass the input double quoted

PR #5278, opened 2022-02-09
CMake: use the policy_max option of cmake_minimum_required() to avoid turning on manually policies

rouault:


CMakeLists.txt
```diff
@@ -1,89 +1,6 @@
 # CMake4GDAL project is distributed under MIT license. See accompanying file LICENSE.txt.
-cmake_minimum_required(VERSION 3.10)
+cmake_minimum_required(VERSION 3.10...3.23)
```
dg0yt, line 2: The max version came in CMake 3.12. 
But let us wait for CI. It should fail in the Ubuntu 18.04 CMake build. That's why I made it use CMake 3.10.2.

rouault: https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html mentions "If the running version of CMake is older than 3.12, the extra ... dots will be seen as version component separators, resulting in the ...<max> part being ignored and preserving the pre-3.12 behavior of basing policies on <min>.". I gave it a try locally with 3.10.2

PR #5283, opened 2022-02-10
CMake CI: Test post-install usage, add static build

dg0yt:
## What does this PR do?

Adds post-install tests with pkg-config, gdal-config and CMake to the CMake CI build for Linux and macOS.
Adds a static build for Linux.
Adds fixes for pkg-config, gdal-config and CMake config export.
Doesn't yet fix the exported CMake config for static linkage. Therefore the result of the static CMake post-install test is inverted: 
post-install test failure <=> CI success.

#5271 adds some `find_dependency` calls to the exported CMake config. The static post-install with CMake will fail until all dependencies are covered. E.g. https://github.com/dg0yt/gdal/runs/5139890930?check_suite_focus=true

## Tasklist

 - [x] Add test case(s)
 - [x] Fix missing dependencies in exported config.
 - [ ] Review
 - [ ] Adjust for comments


dg0yt:
Apart from the static linkage errors, the post-install test failure with CMake on macOS is interesting. It fails to find `odbc.h`. But it works in the pkg-config test. Maybe because `c++` (with pkg-config) is some wrapper from homebrew which implicitly adds the homebrew path, while CMake uses `c++` from XCode with absolute path. 

Now how is ODBC supposed to be used with CMake? `cpl_odbc.h` needs `odbc.h`, but this is not a system header. So an extra include path is needed to use `cpl_odbc.h`. Alternatives:
- Ignore. User is responsible for `find_package(ODBC)` in addition to `find_package(GDAL)`. 
- Add an unconditional `find_dependency()` to the GDAL CMake config. However, ODBC becomes a dependency for users which don't need ODBC. This could perhaps be mitigated by using `find_package` without `REQUIRED`.
- Move the `find_dependency()` into an optional component which the user must request.
- Move the `find_dependency()` into a separate CMake config which the user must request.

dg0yt:
With the static CMake post-install test turned into an "expected fail", this should be mergeable.
The missing find_dependency calls in expected config can be resolved later PRs, in particular by adding more `TARGETS` to `gdal_check_package` if #5271 is merged.

rouault:
Now that #5271 is merged, perhaps this one should be rebased on top of master to ensure both work fine together ?

rouault:
I just gave it a try locally on a Linux shared build to see the gdal.pc / gdal-config it generates.
I see a "/home/even/proj/install-proj-master/lib/libproj.so.22.2.1" in Libs.private . My PROJ master is found through CMake config file. Other dependencies are split under "-L/path -llib" form, sometimes not in a consecutive way, for example I have "L/home/even/install-xerces-c-master/lib ... lots of other things -lxerces-c"
```
Libs.private: -ldl -lodbcinst -lcryptopp -lssl -lcrypto -llz4 -lblosc -ldeflate -llzma -lwebp -larmadillo -lqhull_r -lSFCGAL -lgmpxx -L/home/even/install-xerces-c-master/lib -L/home/even/install-jpeg-xl/lib -ljxl -ljpeg -lzstd -lgif -lz -L/home/even/MrSID_DSDK-9.5.4.4709-rhel6.x86-64.gcc531/Raster_DSDK/lib -lltidsdk -lpcre2-8 -L/home/even/spatialite/install-libspatialite-5.0.1/lib -lspatialite -L/home/even/spatialite/install-librasterlite2-1.1.0-beta1/lib -lrasterlite2 -lexpat -lsqlite3 -lcurl -lxerces-c -lodbc -lxml2 -lgeos_c /home/even/proj/install-proj-master/lib/libproj.so.22.2.1 -ljson-c -lpthread -lstdc++ -lm
```
Not sure if that matters?

dg0yt:
I should add some more comments. It will take a few minutes to write down.

dg0yt:
Apart from your observation (`/home/even/proj/install-proj-master/lib/libproj.so.22.2.1`), there is `/opt/microsoft/msodbcsql17/lib64/libmsodbcsql-17.8.so.1.1` and `/usr/lib/x86_64-linux-gnu/libgeos_c.so.1.16.0` in CI.

With `find_library`, we got the file names which matched pairs of known library prefix and suffix (`.../libgeos_c.so`). Now the config files (used since #5212, #5271) record the actual library file location (`.../libgeos_c.so.1.16.0`). I tried to keep the behaviour as much as is for shared builds, but I didn't expect the latter. There is a change in this PR related to the heuristics for splitting a library argument into `-L... -l...`:
- Before, the filename was simply taken "without ending": 
  `libgeos_c.so.1.16.0` became `libgeos_c`.
  In a second step, the `lib` prefix was removed unconditionally.
  Downsides:
  - It would remove any ending, e.g. `libfoo.bar.so...` became `libfoo`. (I don't know if there are valid lib names with `.`.)
  - It would remove the `lib` prefix even for MSVC where it is required part of the library name if used. (DEFECT)
- With this PR, the library name is taken from what is between the respective pair prefix and suffix for either shared, import or static libraries.
  - This works with `libgeos_c.so`, `libgeosd.lib`, or `libgeos.dll.a`.
  - This doesn't work with `libgeos_c.so.1.16.0` or similar. (CHANGE) 
    However, the code leaves the full path unchanged, so that the linker can find the object. (NO DEFECT)

This points to a significant difference how configuration is exported and used:
- With CMake config, user projects will link using a full path to the library files. (FTR the export has SONAME properties in addition to LOCATION properties.)
- With pkg-conf and gdal-config, we are used to get lib path parameters (`-L`) and lib name parameter (`-l`) separately.
  However, this implies a new build-time library search process based on the lib name
  - with shared libraries normally preferred over static libraries
  - with actual search path order depending on the user project and environment
  - with no additional control that the selected library for linking matches the version which provided the config.

Ideally, the selected library exactly matches the imported config. This is easily ensured by *absolute file paths*. That's why I am unsure whether I should modify the code to restore the original splitting behaviour. Is there any benefit from the splitted flags, or should it be used even less?

(Relocatable packages are achieved by replacing the literal build-time paths with expressions based on a prefix derived from the location of the config file. CMake uses `_IMPORT_PREFIX`, pkg-config offers `${pcfiledir}`, and `gdal-config` might use an expression based on `dirname -- "$0"`.)

rouault:
Thanks for the detailed notes ! The new behavior is probably fine. I was just wondering what having the full name of a library would mean regarding potential bugfixes releases of that library that would have a different full name. But a quick test of linking a test.cpp against "libgdal.so.30.0.0" for example shows that the name embeded in the binary is "libgdal.so.30". So things looks good to me

sebastic:
> I wonder if Linux distributions have additional reasons to prefer -L... -l... in pkg-config. Maybe @sebastic can comment on this?

I doubt Linux distributions where the driving force behind the `gdal.pc` content. It seems to follow the example of other projects providing pkg-config files.

`-L` is only required if the library is not installed any of the standard system directories, it's likely included in `gdal.pc` to make it work when `./configure --libdir` or `cmake -DCMAKE_INSTALL_LIBDIR` is used for with a non standard path.

cmake/modules/thirdparty/SelectImportedConfig.cmake
```diff
@@ -1,29 +0,0 @@
...
-    get_target_property(_imported_conf ${target} MAP_IMPORTED_CONFIG_${CMAKE_BUILD_TYPE})
-    if (NOT _imported_conf)
-        # Get available imported configurations by examining target properties
-        get_target_property(_imported_conf ${target} IMPORTED_CONFIGURATIONS)
-        # Find the imported configuration that we prefer.
-        # We do this by making list of configurations in order of preference,
-        # starting with ${CMAKE_BUILD_TYPE} and ending with the first imported_conf
-        set(_preferred_confs ${CMAKE_BUILD_TYPE})
-        list(GET _imported_conf 0 _fallback_conf)
-        list(APPEND _preferred_confs RELWITHDEBINFO RELEASE DEBUG ${_fallback_conf})
```
dg0yt, line 14: There is no longer such a project-defined list of preferred fallback configs.
But unless property `MAP_IMPORTED_CONFIG_<CONFIG>` is overwritten by the imported CMake config file, the user can simply set `CMAKE_MAP_IMPORTED_CONFIG_<CONFIG>` to his preferred list.

cmake/modules/thirdparty/SplitLibraryToCFlags.cmake
```diff
@@ -1,31 +0,0 @@
...
-        else()
-            string(REGEX REPLACE "^lib" "" _lib_name ${_lib_name})
-            get_filename_component(_lib_dir ${_lib} PATH)
-            if("${_lib_dir}" STREQUAL "")
-                if(_lib_name MATCHES "-l.*")
-                    # e.g "-lm"
-                    set(RESULT "${_lib_name}")
-                else()
-                    # e.g. "dl"
-                    set(RESULT "${CMAKE_LINK_LIBRARY_FLAG}${_lib_name}")
```
dg0yt, line 16: Before this PR, this would turn `-pthread` into `-l-pthread`. (DEFECT)

cmake/modules/thirdparty/SplitLibraryToCFlags.cmake
```diff
@@ -1,31 +0,0 @@
-
-function(split_library_to_cflags _lib _result)
-    if(_lib)
-        set(RESULT)
-        get_filename_component(_lib_name ${_lib} NAME_WE)
-        if(_lib_name STREQUAL "")
-        else()
-            string(REGEX REPLACE "^lib" "" _lib_name ${_lib_name})
```
dg0yt, line 8: Before this PR, this was unsuitable for MSVC which doesn't use the `lib` prefix.

gdal.cmake
```diff
@@ -670,18 +670,13 @@ if (NOT GDAL_ENABLE_MACOSX_FRAMEWORK)
...
-  generate_config(${GDAL_LIB_TARGET_NAME} "gdal_private_link_libraries" ${GDAL_CMAKE_TEMPLATE_PATH}/gdal-config.in
-                  ${PROJECT_BINARY_DIR}/apps/gdal-config)
+  # Generate gdal-config utility command and pkg-config module gdal.pc
+  include(GdalGenerateConfig)
+  gdal_generate_config(
+    TARGET "${GDAL_LIB_TARGET_NAME}"
+    GLOBAL_PROPERTY "gdal_private_link_libraries"
+    GDAL_CONFIG "${PROJECT_BINARY_DIR}/apps/gdal-config"
+    PKG_CONFIG "${CMAKE_CURRENT_BINARY_DIR}/gdal.pc")
   add_custom_target(gdal_config ALL DEPENDS ${PROJECT_BINARY_DIR}/apps/gdal-config)
```
dg0yt, line 680: Can this target be removed? If you lost the file, you can re-run cmake.

cmake/helpers/GdalGenerateConfig.cmake
```diff
@@ -0,0 +1,202 @@
...
+    if(NOT MSVC AND CMAKE_THREAD_LIBS_INIT)
+        list(APPEND libs ${CMAKE_THREAD_LIBS_INIT})
+    endif()
+    list(APPEND libs ${CMAKE_CXX_IMPLICIT_LINK_LIBRARIES})
+    list(REMOVE_ITEM libs ${CMAKE_C_IMPLICIT_LINK_LIBRARIES})
+    gdal_get_lflags(CONFIG_DEP_LIBS ${libs})
+    if(NOT BUILD_SHARED_LIBS)
+        # Make `--libs` simply work, even for static builds.
+        string(APPEND CONFIG_LIBS " ${CONFIG_DEP_LIBS}")
+        set(CONFIG_DEP_LIBS "")
```
dg0yt, line 197: This changes the behaviour for `gdal-config` and `pkg-config` for static linkage: The user only needs to use `--libs` regardles of GDAL linkage. For static linkage, `gdal-config --dep-libs` returns an empty lib.

cmake/helpers/GenerateConfig.cmake
```diff
@@ -1,90 +0,0 @@
...
-
-    # dep-libs
-    get_dep_libs(${_target} ${_link} CONFIG_DEP_LIBS)
-
-    get_filename_component(_output_dir ${_output} DIRECTORY)
-    get_filename_component(_output_name ${_output} NAME)
-    configure_file(${_template} ${_output_dir}/tmp/${_output_name} @ONLY)
-    file(COPY ${_output_dir}/tmp/${_output_name}
-         DESTINATION ${_output_dir}
-         FILE_PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE GROUP_READ GROUP_EXECUTE WORLD_READ WORLD_EXECUTE
```
dg0yt, line 87: I dropped the chmod part, so the execute bits are missing in the binary directory. But they are set in the install step.

gdal.cmake
rouault: > Can this target be removed?

seems fine to remove it

cmake/helpers/GenerateConfig.cmake
rouault: > I dropped the chmod part, so the execute bits are missing in the binary directory. But they are set in the install step.

seems fine to me. (For reference, when I compare to the autoconf build, they have a apps/gdal-config and apps/gdal-config-inst script. The later is the one installed, and the former is aimed at being used for using GDAL in its build tree. I don't see having a gdal-config that works for the build tree as a super compelling feature. )

gdal.cmake
dg0yt: Done.

gdal.cmake
```diff
@@ -670,49 +670,24 @@ if (NOT GDAL_ENABLE_MACOSX_FRAMEWORK)
...
   install(
     PROGRAMS ${PROJECT_BINARY_DIR}/apps/gdal-config
     DESTINATION ${CMAKE_INSTALL_BINDIR}
-    PERMISSIONS OWNER_READ
-                OWNER_WRITE
-                OWNER_EXECUTE
-                GROUP_READ
-                GROUP_EXECUTE
-                WORLD_READ
-                WORLD_EXECUTE
```
dg0yt, line 695: Removed the explicit permissions: they are implied by `install(PROGRAM ...)`.

dg0yt:
> I wonder if Linux distributions have additional reasons to prefer `-L... -l...` in pkg-config.

I think I found the answer now: 
The autotools convention is to build/install both shared and static libs. With `-L... -l...`, the same pc file can be used for either `--static` or shared usage.

However, this doesn't work for Windows where DLLs are used via import libraries which are special static libraries. There have to be different lib names. (In MinGW, it is mitigated by the specific `dll.a` import library suffix.)

Now CMake doesn't support the simultaneous build of static and shared library (for the same target).

So whether splitting into `-L... -l...` is still needed in the future depends on whether there will be (Linux and similar) installations of both static and shared lib where the actual linkage is chosen via a `static` parameter to `pkg-config` or `gdal-config`.

PR #5282, opened 2022-02-10
Entry points for gdal-utils package

maphew:
## What does this PR do?

Modifies `setup.py` so that when `pip install gdal-utils` is run all of the .py files in `./scripts` are turned into native platform executable scripts that call the .py file using the appropriate interpreter. Although developed to improve the experience on Windows the implementation is cross platform (due to upstream magic by setuptools and/or distutils) and also is better for linux (c.f. https://click.palletsprojects.com/en/8.0.x/setuptools/?highlight=entry_points).

Benefits for the user is that after install the gdal python utility scripts are available from the command line immediately, just like all the first class binary utiltiies, e.g. gdalinfo, gdal_translate, and without needing to use .py extension with the name. Gdal edit is just `gdal_edit`.

## What are related issues/pull requests?

gdal-utils: improve command line in Windows and ditch .bat wrappers #5281 

## Tasklist

 - [x] implement and test locally
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments
 - [ ] All CI builds and checks have passed

## Environment

* OS: Microsoft Windows [Version 10.0.19042.804]
* Compiler: n/a

Issue #5281, opened 2022-02-10
gdal-utils: improve command line in Windows and ditch .bat wrappers


maphew:
## Expected behavior and actual behavior.

To avoid having to call every script on the command line like this: `python path\to\gdal\Scripts\gdal_edit.py` on Windows after installing gdal-utils one myse either generate batchfile wrappers or change the PATHEXT variable and ASSOC  & FILETYPE commands. 

This method is well travelled enough that there is both a batch file helper script courtesy of Osgeo4w for some years as well as python utility for more easily creating the wrappers (see https://www.mail-archive.com/gdal-dev@lists.osgeo.org/msg37290.html). These help alleviate but do not remove the pain.

I believe using `entry_points` and  `console_scripts` in `setup.py` instead of the current `scripts` method can resolve this in a solid way that fixes Windows without impacting other platforms. When I've gone through some more testing I will issue a pull request.

## Steps to reproduce the problem.

n/a

## Operating system

Windows

## GDAL version and provenance

All

_cc @idanmiara_

jratike80:
cc @jef-n

PR #100, opened 2016-02-28
netCDF: fix implementation of GetScale/Offset to not always return pb…

jdemaria:
…Success=true

PR #5296, opened 2022-02-11
Entry points for gdal-utils package

maphew:
## What does this PR do?
_this PR supercedes #5282, in which I couldn't figure out how to cleanly rebase so I created this new branch and cherry picked the relevant commit._

This PR modifies `setup.py` so that when `pip install gdal-utils` is run all of the .py files in `./scripts` are turned into native platform executable scripts that call the .py file using the appropriate interpreter. Although developed to improve the experience on Windows the implementation is cross platform (due to upstream magic by setuptools and/or distutils) and also is better for linux (c.f. https://click.palletsprojects.com/en/8.0.x/setuptools/?highlight=entry_points).

Benefits for the user is that after install the gdal python utility scripts are available from the command line immediately, just like all the first class binary utiltiies, e.g. gdalinfo, gdal_translate, and without needing to use .py extension with the name. Gdal edit is just `gdal_edit`.

gdal-utils: improve command line in Windows and ditch .bat wrappers #5281 

## Tasklist
_updated Sun 27 Mar 2022 11:55:11 PM UTC_

 - [x] implement and test locally
 - [x] fix `gdal_merge` (_"TypeError: not a sequence")
 - [x] Add test case(s)
   - [x] Insert a new step after the "Standalone Python bindings build " step in `.github/workflows/cmake_builds.yml` which tests the "gdal" package of swig/python/setup.py
    - [x] testing gdal2tiles multiprocessing on Windows with python >= 3.8  to make sure the issue fixed/workarounded by [0bdb2c8](https://github.com/OSGeo/gdal/commit/0bdb2c84a555d71c93257c2c3adb9d6fe4e7c644) doesn't reappear.
    - [x] Add tests to ensure both modules AND scripts work correctly ([ref](https://github.com/OSGeo/gdal/pull/5296#issuecomment-1074467042)
 - [x] Add documentation
   - [x] add a note that people who make binary distributions for gdal should also modify their setup file accordingly.
 - [x] Review
 - [x] Adjust for comments
   - [x] <s>put auxiliary functions (define_entry_points) in a separate module so it can be imported from both setup files?</s> No. It's really short function so the overhead of pulling it out isn't warranted. Solve with comments.
   - [x] we still see from distutils.errors import CompileError in has_flag and also in some comments. Can this be dealt with here? -- Fixed with #5297 

## Environment

* OS: Microsoft Windows [Version 10.0.19042.804]
* Compiler: n/a


maphew:
Interesting that the ubuntu failure is on only one of the scipts, `gdal2tiles`.  Thanks for pointing out the other (and primary!) means of installing the scripts. I'll study and see if I see a path for not causing problems.

rouault:
> Interesting that the ubuntu failure is on only one of the scipts, `gdal2tiles`.

maybe it just stops at the first encountered issue ? (just assuming)

idanmiara:
> ```
> ERROR: Invalid script entry point: <ExportEntry gdal2tiles = scripts.gdal2tiles:None []> for req: gdal-utils==3.3.0.7 - A callable suffix is required. Cf https://packaging.python.org/specifications/entry-points/#use-for-scripts for more information.
> ```

I didn't test this but I have an assumption based on the error message:

Maybe it has something to do with that in the entry points no function name was defined, i.e. `:main`.
I think I've added in all the scripts a main function that the script calls:
```
if __name__ == '__main__':
    sys.exit(main(sys.argv))
```

If I'm understanding correctly the explanation here:
https://packaging.python.org/en/latest/specifications/entry-points/#use-for-scripts
Maybe we should modify each script's main function to accept argv=None as follows:

```
def main(argv=None):
    if argv is None:
        argv = sys.argv
    ...
```

rouault:
Depending on what the wrapper script generates as code, testing gdal2tiles multiprocessing on Windows with python >= 3.8  would be advised to make sure the issue fixed/workarounded by 0bdb2c84a555d71c93257c2c3adb9d6fe4e7c644 doesn't reappear.

maphew:
In c3d0105 I moved the `__spec__ = None` block into the osgeo/gdal2tiles.py script instead of the to be deprecated wrapper to see if that would help. 

I tried using the recipe in https://github.com/OSGeo/gdal/pull/5296#issuecomment-1037141210. It failed but at a different point and with _"error: Microsoft Visual C++ 14.0 or greater is required."_ so someone with a linux machine will have to repeat to see if this change does anything useful.

idanmiara:
In case gdal2tiles handling is further complicate things, and as gdal2tiles
is the only multiprocessing script AFAIK and thus need special care, I
would suggest to skip it in this PR (by explicitly exclude it in the setup)
and maybe to tackle it in a dedicated PR.

jcphill:
Just a heads-up that I just made gdal2tiles startup more complex by adding mpi4py support.

jcphill:
> ```
> ERROR: Invalid script entry point: <ExportEntry gdal2tiles = scripts.gdal2tiles:None []> for req: gdal-utils==3.3.0.7 - A callable suffix is required. Cf https://packaging.python.org/specifications/entry-points/#use-for-scripts for more information.
> ```
I suspect that you're seeing this error on gdal2tiles merely because gdal2tiles.py is the first file in the directory, that you would see the same error for the other files, and that the real problem is that it doesn't like entry points that are just a module name without :function at the end (even though these are valid according to the spec).

maphew:
both those ideas can be tested by excluding gdal2tiles from console_scripts, so I'll do that and... we'll see!

jcphill:
Just so I understand, why do these wrapper scripts import * and issue a deprecation warning? Are the wrapper scripts supposed to be loadable as modules?

maphew:
Regardless of 2f2c688 outcome, courtesy of @jcphill qyestion I'm now thinking that using `./scripts` to drive console_scripts isn't a good idea. Scripts is deprecated and (someday) will be going away. Since this project is a new thing it's better to be forward looking and work from `./osgeo` instead. That will be my next focus.

idanmiara:
It imports * to be backwards compatible with old dependent scripts that
were written before gdal-utils was introduced. People then added the
scripts dir to the path and imported the script as a module.
The depreciation warning is to draw their attention for the much more
straight forward way of importing it.

jcphill:
> It imports * to be backwards compatible with old dependent scripts that were written before gdal-utils was introduced. People then added the scripts dir to the path and imported the script as a module. The depreciation warning is to draw their attention for the much more straight forward way of importing it.

Since almost all of the wrapper scripts don't check if __name__ == '__main__' before calling sys.exit(main(sys.argv)) I don't think you can import them as modules without main and exit being called.

jcphill:
https://github.com/OSGeo/gdal/pull/5359 reverts all of the mpi4py changes to the gdal2tiles.py script and running mpiexec on a gdal2tiles entry point should just work.

jcphill:
https://github.com/OSGeo/gdal/pull/5367 removes the last of the uniqueness from the gdal2tiles wrapper script.
The standard `if __name__ == '__main__'` guard is still required for multiprocessing on Windows and MacOS.

maphew:
@rouault when you get a chance would you please re-run the test you outlined in https://github.com/OSGeo/gdal/pull/5296#issuecomment-1037141210? I've not yet been successful in getting a functioning gdal in ubuntu build and it looks like I have some ways to go before I will! Thanks.

rouault:
> when you get a chance would you please re-run the test you outlined in [#5296 (comment)](https://github.com/OSGeo/gdal/pull/5296#issuecomment-1037141210)?

the wheel installs, but the wrapper scripts without extensions that get installed in bin don't work whereas the ones with .py work:
```

(myvenv) even@even-ThinkPad-P15v-Gen-1:~/gdal/gdal/build_cmake$ gdal_edit
  File "/home/even/gdal/gdal/build_cmake/myvenv/bin/gdal_edit", line 5, in <module>
    from scripts.gdal_edit import main
ModuleNotFoundError: No module named 'scripts.gdal_edit'

(myvenv) even@even-ThinkPad-P15v-Gen-1:~/gdal/gdal/build_cmake$ cat /home/even/gdal/gdal/build_cmake/myvenv/bin/gdal_edit
#!/home/even/gdal/gdal/build_cmake/myvenv/bin/python3
# -*- coding: utf-8 -*-
import re
import sys
from scripts.gdal_edit import main
if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw|\.exe)?$', '', sys.argv[0])
    sys.exit(main())

(myvenv) even@even-ThinkPad-P15v-Gen-1:~/gdal/gdal/build_cmake$ which gdal_edit.py
/home/even/gdal/gdal/build_cmake/myvenv/bin/gdal_edit.py

(myvenv) even@even-ThinkPad-P15v-Gen-1:gdal_edit.py
Usage: gdal_edit [--help-general] [-ro] [-a_srs srs_def]
                 [-a_ullr ulx uly lrx lry] [-a_ulurll ulx uly urx ury llx lly]
                 [-tr xres yres] [-unsetgt] [-unsetrpc] [-a_nodata value] [-unsetnodata]
                 [-offset value] [-scale value] [-units value]
                 [-colorinterp_X red|green|blue|alpha|gray|undefined]*
                 [-unsetstats] [-stats] [-approx_stats]
                 [-setstats min max mean stddev]
                 [-gcp pixel line easting northing [elevation]]*
                 [-unsetmd] [-oo NAME=VALUE]* [-mo "META-TAG=VALUE"]*  datasetname

Edit in place various information of an existing GDAL dataset.

```

maphew:
I finally understand why the _console_scripts_ are broken and have identified a solution. However I don't know if using it is a good idea as it adds yet one more layer of indirection.

-----
While handy `pip install --editable .` is not a reliable means of testing console_scripts top to bottom. My early tests were successful because editable mode scripts pass sys.argv() all the way though in a way wheel installed scripts do not.

The function called by console_scripts **must not** have any arguments. This was the critical insight, courtesy of @metatoaster on Stack Overflow ([ref](https://stackoverflow.com/questions/40753946/python-setup-tools-console-scripts-with-arguments)).

The workaround is to add an argumentless function to the script that then calls `main(argv)`, and then point console_scripts at that function:

gdal_edit.py tail:
~~~
def main(argv):
    return gdal_edit(argv)

def console():
    main(sys.argv)

if __name__ == '__main__':
    sys.exit(main(sys.argv))
 ~~~

setup.py:
~~~
# old: console_scripts.append([f'{name} = osgeo_utils.{name}:main'])
console_scripts.append([f'{name} = osgeo_utils.{name}:console'])
~~~
----

So this works, but as I said the extra indirection makes a bit uncomfortable. What would be the consequence of rewriting main to be optional?
~~~
def main(argv=sys.argv):
    return gdal_edit(argv)
~~~

Or:
~~~
def main():
    return gdal_edit(sys.argv)
~~~
?

idanmiara:
Hi,

I think that the optional parameter is a clean and good solution.

def main(argv=sys.argv):
    return gdal_edit(argv)

metatoaster:
Since I've been summoned I figured I might leave a few cents:

> While handy `pip install --editable .` is not a reliable means of testing console_scripts top to bottom. My early tests were successful because editable mode scripts pass sys.argv() all the way though in a way wheel installed scripts do not.

It might be possibly more to do with the `scripts` namespace not be packaged and installed with the wheel, and your working environment still got the old versions of the scripts lying around that was installed through the `scripts` attribute for `setup.py` (which might need to be be regenerated through `pip install -e .` after modifying the `setup.py` in place, I don't have access to your editable install so I can only guess here).

Even if the generated scripts done through `console_scripts` are different through `pip install -e .` (which is effectively `python setup.py develop`) and the one done through a wheel install (that bypasses `setup.py` completely since it's already built; incidentally that only results in the modification of the `#!` part), the resulting action by the "generated binary" should be identical.

> setup.py:
> 
> ```
> # old: console_scripts.append([f'{name} = osgeo_utils.{name}:main'])
> console_scripts.append([f'{name} = osgeo_utils.{name}:console'])
> ```

Yes, refer to the `osgeo_utils.` namespace, instead of `scripts.` [what it is now](https://github.com/OSGeo/gdal/pull/5296/files#diff-d50366bcc7b0dbcaf44cf1a37f744b2736a9dba42af1e173ff8968bac8f17b08R54) (quoting for posterity) should fix the issue

```python
        console_scripts.append([f'{name} = scripts.{name}:main'])
```

Also using the optional argv (i.e. `main(argv=sys.argv)` would match your intended usage, so for that I also updated the stackoverflow answer.

maphew:
@metatoaster thank you for taking the time to come and share your insight. I believe you're right, that the inconsistencies between editable and wheel installs were because of legacy file remnants still in place with repeated cycles of going back forth between the two. I tend to forget to ensure the `*.egg-info` and `build` dirs are removed in between, and I wonder if there might something in PYTHONHOME/Lib/site-packages and Scripts sometimes as well.

@idanmiara Ok, I will update all the scripts to use `main(argv=sys.argv)`, and then we'll see what the CI has to say. ;-)

...

A note for future troubleshooters: the key exploration that got me pointed in the right direction was discovering the .exe file installed in PTYHONHOME\Scripts can be unpacked by 7zip, which reveals a `__main__.py` file. The line numbers reported in the traceback correspond to this file, which was using `main()` with no args. Playing with this __main__.py let me reproduce and fix the error in place (adding sys.argv), and then it was searching for how to feed this through setup.py that led me to led me to the Stack Overflow answer (you can't do that).

maphew:
Regarding the CI failures:
1.  https://github.com/OSGeo/gdal/runs/5560131996?check_suite_focus=true _"W: Conflicting distribution: https://packages.microsoft.com/ubuntu/20.04/prod focal InRelease (expected focal but got testing)"_ I don't understand how this can be from this PR. I've merged in upstream master and haven't touched anything in `.github` that I'm aware of. Yet: CI on master is not failing.
2. https://github.com/OSGeo/gdal/runs/5560132338?check_suite_focus=true - same as No.1

3. https://github.com/OSGeo/gdal/runs/5560131863?check_suite_focus=true - Shellcheck fail. This one is my code but I don't know what to do about it. Figure out how to use absolute path, which will change from machine to machine, in the line `source ./bin/activate` ? It's not the best test script anyway. It should run one or gdal scripts  instead of especting an use to do it interactively, and remove the venv when done.

This manual test process worked for me in a Gitpod machine from this branch:
https://gitpod.io/#https://github.com/maphew/gdal/tree/pr-5281-redux

~~~
# install gdal binaries, and git
sudo add-apt-repository -y ppa:ubuntugis/ubuntugis-unstable
sudo apt-get update
sudo apt-get install gdal-bin libgdal-dev git

# install gdal-utilties
cd swig/python/gdal-utils
python setup.py bdist_wheel
pip install dist/gdal_utils-3.3.0.7-py3-none-any.whl

# ensure gdal source code tree installed gdal binaries are the same version
git checkout v`gdal-config --version`

# run python script tests
cd ../../../autotest/
pip install -r requirements.txt
pytest pyscripts/

#100%!! :-) 
~~~
image

rouault:
> Regarding the CI failures:

for 1 and 2), hopefully transient issue on github VM workers (I got a similar error in another PR). I've restarted those jobs

maphew:
@rouault Re: 
>Another point to consider is that the main "gdal" package (swig/python/scripts) also provides the scripts, so if we modify the way of installing them in "gdal-utils" we must also do the same way in "gdal" so that everything overwrites itself"

Would that be an additional step in this PR or in a new one?

maphew:
@jcphill On Windows after pip installing this PR running `pytest pyscripts\test_gdal2tiles.py` is 100% successful so I believe this confirms the multiprocessing problems above are confirmed resolved? Looking at git blame I seem to be credited with some changes I don't think I made (due to synchronizing with master, maybe I'm not doing that the best way). Please verify `gdaltiles.py` doesn't have anything left from me.

~~~
» pytest pyscripts\test_gdal2tiles.py                                                          
Test session starts (platform: win32, Python 3.10.2, pytest 7.1.0, pytest-sugar 0.9.4)         
rootdir: C:\Users\mhwilkie\code\gdal\autotest, configfile: pytest.ini                          
plugins: env-0.6.2, sugar-0.9.4                                                                
collecting ...                                                                                 
 pyscripts/test_gdal2tiles.py ✓                                                   7% ▊         
 pyscripts/test_gdal2tiles.py ✓✓                                                 14% █▌        
 pyscripts/test_gdal2tiles.py ✓✓✓                                                21% ██▎       
 pyscripts/test_gdal2tiles.py ✓✓✓✓                                               29% ██▉       
 pyscripts/test_gdal2tiles.py ✓✓✓✓✓                                              36% ███▋      
 pyscripts/test_gdal2tiles.py ✓✓✓✓✓✓                                             43% ████▍     
 pyscripts/test_gdal2tiles.py ✓✓✓✓✓✓✓                                            50% █████     
 pyscripts/test_gdal2tiles.py ✓✓✓✓✓✓✓✓                                           57% █████▊    
 pyscripts/test_gdal2tiles.py ✓✓✓✓✓✓✓✓✓                                          64% ██████▌   
 pyscripts/test_gdal2tiles.py ✓✓✓✓✓✓✓✓✓✓                                         71% ███████▎  
 pyscripts/test_gdal2tiles.py ✓✓✓✓✓✓✓✓✓✓✓                                        79% ███████▉  
 pyscripts/test_gdal2tiles.py ✓✓✓✓✓✓✓✓✓✓✓✓                                       86% ████████▋ 
 pyscripts/test_gdal2tiles.py ✓✓✓✓✓✓✓✓✓✓✓✓✓                                      93% █████████▍
 pyscripts/test_gdal2tiles.py ✓✓✓✓✓✓✓✓✓✓✓✓✓✓                                    100% ██████████
~~~

maphew:
In `cmake_builds.yml` I added a new step after "Standalone Python bindings build" to test this package, commit [f53261c](https://github.com/OSGeo/gdal/pull/5296/commits/f53261cb1bed2a7a05b3a80208ddc530274ce46b). Looking at tbe two steps together, it seems to me the wheel install test can be taken out of "Standalone Python bindings". Thoughts? 

The substantive differences in the new step are: not usng `.py` extension for `gdal_edit` (because _console_scripts_ took care of that for us), and adding a `pip uninstall` stage.

swig/python/gdal-utils/osgeo_utils/gdal2tiles.py
```diff
@@ -3393,5 +3393,11 @@ def submain(argv: List[str], pool=None, pool_size=0) -> int:
 
 # vim: set tabstop=4 shiftwidth=4 expandtab:
 
+# Running main() must be protected that way due to use of multiprocessing on Windows:
+# https://docs.python.org/3/library/multiprocessing.html#the-spawn-and-forkserver-start-methods
 if __name__ == '__main__':
+    # Trick inspired from https://stackoverflow.com/questions/45720153/python-multiprocessing-error-attributeerror-module-main-has-no-attribute
+    # and https://bugs.python.org/issue42949
```
jcphill, line 3400: These changes duplicate code at line 3292.

jcphill:
@maphew I think your changes to gdal2tiles.py are redundant as __spec__ is set elsewhere, but please test on Windows without them.

maphew:
@idanmiara Re:
>add a note that people who make binary distributions for gdal should also modify their setup file accordingly.

I've updated the Readme for this and other things. Please review and let me know what you think.

maphew:
wierd, Github is telling me at https://github.com/OSGeo/gdal/commit/dbaee3d3129944a875ea061a8ddff62dd7af4f92 is not on this branch, but [here](https://github.com/maphew/gdal/tree/pr-5281-redux) and on my local machine it is? Maybe  it will fix itself on next commit. --> yep, seems to be so.

@rouault the cmake_build.yml build step I added seems to be working. After you confirm my logic is correct and I'll remove the disabled parts.

There is a bdist_wheel error reported in "Standalone Python" step but it's not being picked up by CI as a failure. It's also occuring in master: https://github.com/OSGeo/gdal/runs/5587012389?check_suite_focus=true#step:15:600.

swig/python/gdal-utils/README.rst
```diff
@@ -1,23 +1,21 @@
...
 
 Please also refer to the osgeo documentation and to the `GDAL API Tutorial`_.
 
 Currently, The the gdal package also includes the utils package,
 but in the future they might be available only in this gdal-utils package.
 
-Because we don't test the utils against a different versions of gdal -
-If you want to get the latest version of the utils, we do recommend you to upgrade
-to the latest version of gdal and get the latest version of both.
+Versioning of GDAL and gdal-utils are independent from each other, e.g.
```
rouault, line 15: I would keep the original text: there's no way we can guarantee that a version of gdal-utils more recent than libgdal or its python bindings will work (as this is not tested and given that "what is not tested is broken" is a good mantra). In most cases, it might seem to work, but for some options, gdal-utils can use newer API or depend on bug fixes.

rouault:
> After you confirm my logic is correct and I'll remove the disabled parts.

looks good

> There is a bdist_wheel error reported in "Standalone Python" step 

that error is "normal" as far as I can see, at least it results in a correct package being produced. I remember having seen similar errors in other Python modules that compile files.

swig/python/gdal-utils/README.rst
maphew: Thanks for that feedback. I re-wrote this paragraph to accommodate and added a new section on how to test compability across versions. It's soft spot is we assume Python Osgeo module is already installed and working. 

(I'm hazy on how this works in Ubuntu, on the relationship between `python3-gdal` apt-get package and PyPi.org `gdal` package. Somehow both seem to be necessary but the latter overrides part of the former, and requires it? Or something.)

rouault: > (I'm hazy on how this works in Ubuntu, on the relationship between python3-gdal apt-get package and PyPi.org gdal package. Somehow both seem to be necessary but the latter overrides part of the former, and requires it? Or something.)

No, you can't use either. If both are used, generally the python import paths rules will use the pip module.

maphew: > No, you can't use either. If both are used, generally the python import paths rules will use the pip module.

Sorry I'm having trouble parsing this. Can you expand?  Or maybe my fuzziness on apt vs pip packages doesn't need to be resolved right now or for this PR and can wait. Is the Readme in current state okay?

rouault: sorry I meant that you can use either the python3-gdal apt-get package or the pypi pip package (but for that later one, you need to make sure to use the same version as the native libgdal installed on the system).

rouault: > Is the Readme in current state okay?

yes

.github/workflows/cmake_builds.yml
```diff
@@ -119,13 +119,11 @@ jobs:
         cmake  ${CMAKE_OPTIONS} -DCSHARP_MONO=ON -DCMAKE_INSTALL_PREFIX=$GITHUB_WORKSPACE/install-gdal -DCMAKE_UNITY_BUILD=${CMAKE_UNITY_BUILD} -DCMAKE_C_FLAGS=-Werror -DCMAKE_CXX_FLAGS=-Werror ..
         cmake --build $GITHUB_WORKSPACE/superbuild/build --target install -- -j$(nproc)
         ctest -V -R "^csharp.*"
-    - name: Standalone Python bindings build and gdal-utils package
+    - name: Standalone Python bindings build <s>and gdal-utils package</s>
```
rouault, line 122: I guess this should be
```suggestion
    - name: Standalone Python bindings build
```

maphew: Oops, that one slipped through, thanks.

maphew:
### Current status
Provided that
 1. GDAL binaries are installed (e.g. after `apt-get install gdal-bin libgdal-dev` or `conda install gdal`)
 2. GDAL Python bindings are installed (e.g. after `pip install gdal`)
 
  - `pip install gdal-utils` will replace the python utilities from No.2, and
  - add the scripts to PYTHONHOME\Scripts on Linux and Windows, and 
  - the scripts are known to work.
  - CI is happy

So from my perspective this PR is complete. 

The additional work mentioned in the thread that I think can be done as separate efforts:

- move the auxiliary functions (define_entry_points) into a separate module
- adapt the approach used in this PR to the parent GDAL python package (`swig/python/setup.py`) (and possibly splitting the two packages entirely).

Agreed? Have I missed something?

rouault:
> So from my perspective this PR is complete.

looks good to me too. I'm happy to merge as soon as you remove the Draft flag

> * move the auxiliary functions (define_entry_points) into a separate module

That seems a bit overkill given it is just a few lines long. Just adding a comment to ask the content to be kept in sync should be enough.

> adapt the approach used in this PR to the parent GDAL python package (swig/python/setup.py)

Yes, and in swig/python/setup.py.in as well.  Both co-exist because swig/python/setup.py.in is for autoconf/nmake and  swig/python/setup.py.in  is used by CMake. Ultimately we'll keep only swig/python/setup.py.in .

jcphill:
Sorry to only catch this at the last minute, buy does gdal2tiles.py also need main() to have sys.argv as its default argument?

maphew:
> Sorry to only catch this at the last minute, buy does gdal2tiles.py also need main() to have sys.argv as its default argument?

@jcphill The short answer is yes, I think so.  The main thing is that _console_scripts_ requires main() to be called with no arguments, even though in practice it actually does pass arguments. I solved this in the other scripts by making argv optional:

```
def main(argv=sys.argv):
    return gdal_edit(argv)

if __name__ == '__main__':
    sys.exit(main(sys.argv))
```

However I've never seen the construct `def main(argv: List[str]) -> int:` gdal2tiles is using before and I don't understand what it's doing. So there might be something else that works too, or work better.

> Sorry to only catch this at the last minute, 

I'm not!! The question revealed a significant problem in how I've been testing. `pytest pyscripts/gdal2tiles` gives a glowing 100% success while the simple `gdal2tiles` on the command line fails with `TypeError: main() missing 1 required positional argument: 'argv'`. I've been proving the gdal-util _modules_ are still working after this change _but not_ that the scripts are. Eeep.

jcphill:
Those are function annotations (https://peps.python.org/pep-3107/). Based on some web searches I think you would specify the default argument after the type annotation, i.e.:
`def main(argv: List[str] = sys.argv) -> int:`

maphew:
@jcphill that seems to work. Commit 2b52658 and testing on Ubuntu 20 via Gitpod is successful for me.

Testing command line util arguments are passed:
```
gitpod /workspace/gdal (pr-5281-redux) $ gdal2tiles in out
Usage: gdal2tiles.py [options] input_file [output]

gdal2tiles.py: error: The provided input file in does not exist or is not a file
```

Autotest suite:
```
cd autotest
pytest pyscripts/test_gdal2tiles.py
# 14 passed
pytest pyscripts/gdal2tiles
# 41 passed
```

Also thanks for the study link :)

maphew:
> IThe question revealed a significant problem in how I've been testing. `pytest pyscripts/gdal2tiles` gives a glowing 100% success while the simple `gdal2tiles` on the command line fails with `TypeError: main() missing 1 required positional argument: 'argv'`. I've been proving the gdal-util _modules_ are still working after this change _but not_ that the scripts are. Eeep.

@idanmiara, @rouault I think I've built a test structure that can address the problem Jim found with gdal2tiles and can be applied to all the gdal-utils scripts, see `autotest/pyscripts/test_gdal_utils_cli.py`. This my first time writing pytest tests, so a close look and feedback is warranted. ;-)  I followed the header of format of other scripts regarding license declaration, copyright etc. I'm not set on this (I know some projects need copyright assignment to the primary maintainer).

autotest/pyscripts/test_gdal_utils_cli.py
```diff
@@ -0,0 +1,86 @@
...
+# python 'site-packages' and 'Scripts')
+installed = subprocess.run([sys.executable, '-m', 'pip', 'show', 'gdal-utils'])
+if not installed.returncode == 0:
+    pytest.skip("The 'gdal-utils' package is not installed.", allow_module_level=True)
+
+
+@pytest.mark.parametrize("input,want", [
+    pytest.param("gdalcompare", {
+        "returncode": 1,
+        "stdout": "Usage: gdalcompare.py [-sds] <golden_file> <new_file>\n",
```
rouault, line 65: relying on an exact string is going to be fragile when new options are tested. I would just check that stdout or stderr starts with 'Usage:'

maphew: I like that idea. I've spent the last several sessions figuring out how to generate and then read stdout/err from files in pytest. I learned a lot! but still didn't get to the finish line (c.f. a93a1ed). I'm not sorry for the field trip, as it's stretched and extended my skills, but I do want to finish and take time look at the flowers later. ;-)

maphew: hmmm. It's not going to be as straightforward as seeking for 'Usage'. Every program is different in what they return. Arguably they should be harmonized to all use a common syntax. It would also be good to have them settle on a single one of stdout or stderr instead of back and forth like they are now.

jcphill: An alternate approach would be checking that the stderr output does not contain "^Traceback" (or whatever else is common when exiting due to uncaught exceptions) and that something was printed to either stderr or stdout.

maphew: Since the test using the external data files approach is working and done I've committed it. That can work until the usage messages are standardized, which I think is worthwhile on it's merit.

Better yet, this test script has proofed it's worth! `gdal_merge` is broken in this branch of _gdal-utils_.

maphew:
Something in the pip install process breaks `gdal_merge`:

```
» gdal_merge
  File "C:\tools\miniconda3\envs\gdal\lib\runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\tools\miniconda3\envs\gdal\lib\runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "C:\tools\miniconda3\envs\gdal\Scripts\gdal_merge.exe\__main__.py", line 7, in <module>
  File "C:\tools\miniconda3\envs\gdal\lib\site-packages\osgeo_utils\gdal_merge.py", line 323, in main
    argv = gdal.GeneralCmdLineProcessor(argv)
  File "C:\tools\miniconda3\envs\gdal\lib\site-packages\osgeo\gdal.py", line 4596, in GeneralCmdLineProcessor
    return _gdal.GeneralCmdLineProcessor(*args)
TypeError: not a sequence
```

maphew:
The gdal_merge issue was simple: it just hadn't been updated yet to use the pattern:
```
def main(argv=sys.argv):
    return gdal_merge(argv)

if __name__ == '__main__':
    sys.exit(main(sys.argv))
```

.gitpod.yml
```diff
@@ -0,0 +1,37 @@
+# Also see https://www.gitpod.io/docs/config-gitpod-file
```
rouault, line 1: this file should not go in this pull request

autotest/pyscripts/test_gdal_utils_cli.py
```diff
@@ -0,0 +1,135 @@
...
+   when called with no parameters.
+
+    The expected results must be in pre-generated in files named:
+
+        cli_outs/scriptname.stdout  # normal
+        cli_outs/scriptname.stderr  # errors
+
+    These output files can be generated with:
+
+        gdalcompare 1>gdalcompare.stdout 2>gdalcompare.stderr
```
rouault, line 88: I really don't like having all those .stderr / .stdout files. This is going to be an extra hurdle for contributors.
Looking at the output it seems that almost all have "usage:" / "Usage:" in the stdout/stderr. And for the few ones that don't (gdal_proximity, gdal_sieve, gdalmove, ogrmerge.py), it seems that checking that the name of the script and " <" or " [" is present should be enough. Or it might actually be faster to edit them so they output "usage:" as well.

.gitpod.yml
maphew: an oversight! I'll pull it.

autotest/pyscripts/test_gdal_utils_cli.py
maphew: I agree. I was just hesitant to add something like that to this PR; I don't want to overstep my bounds. I'll do this now though.

maphew: Done in fa37316.  It was a lot easier than I envisioned. (I'd noticed in my earlier scans that the usage message is coming from some place hidden to me for many of the scripts. I thought that meant extended edits would be needed. It turns out all of the hidden ones already start with _usage_ so I didn't have to touch them. Yay!)

maphew:
I believe everything has been resolved and the PR is ready for final review. _(Take with salt. I've thought this before ! ;-)_. The two CI fails don't appear to be related to this work.

rouault:
squashed and merged

Issue #5298, opened 2022-02-12
Add division to VRT Default Pixel Functions


lpinner:
Suggest to add division (`div`) to VRT default pixel functions [[1](https://gdal.org/drivers/raster/vrt.html#default-pixel-functions), [2](https://github.com/OSGeo/gdal/blob/master/frmts/vrt/pixelfunctions.cpp)].

Would be useful where a python pixel function can't be used for whatever reason (e.g. performance, possibly locked down corporate environments) and avoids a complex `<SourceFilename etc...> <![CDATA[vrt xml string]]></SourceFilename>` inlining of VRT XML to combine `mul` and `inv` functions [[3](https://gis.stackexchange.com/a/423557/2856)]

avalentino:
Dear @lpinner thanks for the interesting pointer to [3](https://gis.stackexchange.com/questions/423154/gdal-vrt-how-to-create-rgb-composite-from-two-input-geotiff-files-input1-inp/423557#423557), probably we should describe that trick in the user documentation or at lest have a link to the discussion.

Regarding the addition of the "div" pixel function I do not remember why it was not included in the base set of pixel functions at the beginning [1](https://trac.osgeo.org/gdal/ticket/3367).
I was planning to submit a PR to add some pixel functions (to improve conversion form/to dB and conversion form amp+phase to complex).
I could add also the "div" implementation to the PR.

@rouault do you have comments7recommendations?

rouault:
> @rouault do you have comments7recommendations?

that sounds good

PR #5308, opened 2022-02-14
Improvements to pixelfunctions

avalentino:
## What does this PR do?

* new pixelfunction ``div`` for division between raster bands
* new ``polar`` pixel function to create a complex raster band form amplitude and phase
* possibility to add a scalar value (provided as optional parameter) to the result of ``sum`` pixel function
* add the possibility to multiply by a scalar value (provided as optional parameter) to the result of ``mul`` pixel function
* add the possibility to multiply by a scalar value (provided as optional parameter) to the result of ``inv`` pixel function
* add the possibility to convert form intensity to dB using the existing ``dB`` pixel function by providing a (new) optional parameter via VRT file
* new ``exp`` pixel function to compute ``base ** ( fact * x )`` (default values: ``base = e`` (Euler's number), ``fact=1.``). This is not the recommended way to perform conversion form dB.
* deprecated ``db2amp`` and ``db2pow`` pixel function

~~The last point on~~
~~* one hand fixes an asymmetry in terms of features w.r.t. the inverse conversion for which ``dB2amp`` and ``dB2pow`` pixel functions were already available~~
~~* on the other hand introduces an asymmetry on the way the feature is provided:~~
~~- one pixel function with optional parameter for the direct conversion to dB~~
~~- two specific pixel functions for the in verse conversion from dB~~

~~Comments form the reviewers are very welcome in this specific point.~~

* #5298

## Tasklist

 - [x] Add test case(s)
 - [x] Add documentation
 - [x] Decide if it is necessary to keep symmetry in pixel functions for conversion from/do dB
 - [x] Review
 - [x] Adjust for comments


doc/source/drivers/raster/vrt.rst
```diff
@@ -856,8 +860,8 @@ GDAL provides a set of default pixel functions that can be used without writing
      - interpolate a value at time (or position) ``t`` given input sources beginning at ``t0`` with spacing ``dt`` using linear interpolation
    * - **inv**
      - 1
-     - -
-     - inverse (1./x). Note: no check is performed on zero division
+     - 1 (optional)
```
rouault, line 863: The "Note: no check is performed on zero division" is outdated (0 is actually checked). Similar for the div case

frmts/vrt/pixelfunctions.cpp
```diff
@@ -1086,6 +1164,8 @@ CPLErr InterpolatePixelFunc( void **papoSources, int nSources, void *pData,
  * - "sum": sum 2 or more raster bands
  * - "diff": computes the difference between 2 raster bands (b1 - b2)
  * - "mul": multiply 2 or more raster bands
+ * - "div": divide one rasted band by another (b1 / b2).
+ *          Note: no check is performed on zero division
```
rouault, line 1168:
```suggestion
 * - "div": divide one rasted band by another (b1 / b2).
```

doc/source/drivers/raster/vrt.rst
avalentino: done

frmts/vrt/pixelfunctions.cpp
avalentino: done

avalentino:
Thanks @rouault.

Just 2 questions:

* should we deprecate ``dB2amp`` and ``dB2pow`` in favour of a new ``linear`` pixwl function with parameter (for symmetry with ``dB``)?
* do you have a better name to suggest for the name of the parameter of the pixel function ``polar``? Currently it is ``amplitude_type``

rouault:
> should we deprecate dB2amp and dB2pow in favour of a new linear pixwl function with parameter (for symmetry with dB)?

'linear' sounds odd to me given the formula of those functions. That would rather be an extension of the existing "pow" to accept a "k" factor

> do you have a better name to suggest for the name of the parameter of the pixel function polar?

not really

avalentino:
>> should we deprecate dB2amp and dB2pow in favour of a new linear pixel function with parameter (for symmetry with dB)?
>
> 'linear' sounds odd to me given the formula of those functions. That would rather be an extension of the existing "pow" to accept a "k" factor

Currently ``pow`` implements ``b1 ** (power)`` where ``power`` is the pixel function parameter.

``dB2amp`` and ``dB2pow`` are based on top of the ``PowPixelFuncHelper`` function that implements ``base ** (b1 / fact)`` where ``base`` and ``fact`` are parameters.
The idea is to expose directly the functionality provided by ``PowPixelFuncHelper`` exploiting pixel function arguments.

Unfortunately we cannot use the ``pow`` name fro the registration because it is already taken.

Maybe we could use ``pow10`` (assuming 10 as default value for ``base``) or something like ``xpow``, ``powx`` or ``exp`` for "exponent".

``dB2amp`` and ``dB2pow`` of course would be kept as deprecated alternatives.

rouault:
> Maybe we could use `pow10` (assuming 10 as default value for `base`) or something like `xpow`, `powx` or `exp` for "exponent".

`exp` sounds reasonable

I was lately thinking it would be nice to add (simple) expression capabilities, to be able to have formulas like `(IR - R)/(IR +R)` (when sources would have a name), or `sum(SOURCES)` or `k * SOURCES[0] + SOURCES[1]` where SOURCES represents the list of sources and k a function argument, that would make it possible to chain in a natural way all the base operations we have. Most of the parsing logic of formula parser in the ODS driver could likely be borrowed. I'm not sure if/when I'll be able to tackle that myself

frmts/vrt/pixelfunctions.cpp
```diff
@@ -932,14 +1008,34 @@ static CPLErr PowPixelFuncHelper( void **papoSources, int nSources, void *pData,
...
+                            int nPixelSpace, int nLineSpace,
+                            CSLConstList papszArgs )
+{
+    double dfBase = 2.7182818284590452353602874713526624;
+    double dfFact = 1.;
+
+    if ( FetchDoubleArg(papszArgs, "base", &dfBase, &dfBase ) != CE_None )
+        return CE_Failure;
+
+    if ( FetchDoubleArg(papszArgs, "fact", &dfFact, &dfFact ) != CE_None )
```
rouault, line 1025: dfFact != 0 should be tested

avalentino: Good point, it is now fixed.

By the way we could also implement ``base ^ ( fact * x )`` instead of ``base ^ ( x / fact )``

avalentino:
> I was lately thinking it would be nice to add (simple) expression capabilities, to be able to have formulas like (IR - R)/(IR +R) (when sources would have a name), or sum(SOURCES) or k * SOURCES[0] + SOURCES[1] where SOURCES represents the list of sources and k a function argument, that would make it possible to chain in a natural way all the base operations we have. Most of the parsing logic of formula parser in the ODS driver could likely be borrowed. I'm not sure if/when I'll be able to tackle that myself

Indeed It would be interesting.
I will give a look to the ODS driver to understand how to formula parser works.

frmts/vrt/pixelfunctions.cpp
rouault: > By the way we could also implement base ^ ( fact * x )

yes, I was also wondering if that wouldn't be more generic

avalentino: done

avalentino:
> I will give a look to the ODS driver to understand how to formula parser works.

@rouault sorry I don't think that I will be able to look at formula parser this week.
I assume that it would be a separate PR in any case, correct?

rouault:
> I assume that it would be a separate PR in any case, correct?

yes, definitely! that's not a 5-minute job at all.

Can you drop the commit for the -j auto option, and just rebase on top of latest master where there is a fix for that doc issue ?

rouault:
can you update the text in the PR introduction (at least the "exp" part needs to be updated) so I can use it as the commit message once I squash the commits ?

avalentino:
> can you update the text in the PR introduction (at least the "exp" part needs to be updated) so I can use it as the commit message once I squash the commits ?

done

avalentino:
Thanks @rouault

PR #5309, opened 2022-02-14
gdal2tiles: support mpi4py mult-node parallelism

jcphill:
Allow launching gdal2tiles.py across hosts using mpi4py, e.g.,

  mpiexec -n $NRANKS python -m mpi4py.futures gdal2tiles.py ...

This usage model avoids dynamic process management as described at
https://mpi4py.readthedocs.io/en/stable/mpi4py.futures.html#command-line

Works by detecting if mpi4py.futures module has been imported, otherwise
supports single-host multiprocessing exactly as before.  Creates temp
directory in output directory to ensure access from all hosts.  Also
only creates tile directories once, on rank 0, for better performance.

Since there is no standard way to detect how many how many ranks are
launched per host the mpi4py path assumes that the user has set the
GDAL_CACHEMAX environment variable appropriately and does not reduce it.

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:

### DIFF
--- a/swig/python/gdal-utils/osgeo_utils/gdal2tiles.py
+++ b/swig/python/gdal-utils/osgeo_utils/gdal2tiles.py
@@ -48,7 +48,11 @@ import sys
 import tempfile
 import threading
 from functools import partial
-from multiprocessing import Pool
+if 'mpi4py.futures' in sys.modules:
+    from mpi4py.futures import MPIPoolExecutor
+    from mpi4py import MPI
+else:
+    from multiprocessing import Pool
 from typing import List, NoReturn, Tuple, Optional, Any
 from uuid import uuid4
 from xml.etree import ElementTree
@@ -1263,9 +1267,6 @@ def create_overview_tile(base_tz: int, base_tiles: List[Tuple[int, int]], output
             print("Tile generation skipped because of --resume")
         return
 
-    # Create directories for the tile
-    os.makedirs(os.path.dirname(tilefilename), exist_ok=True)
-
     if usable_base_tiles:
         scale_query_to_tile(dsquery, dstile, tile_driver, options,
                             tilefilename=tilefilename)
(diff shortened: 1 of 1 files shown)


swig/python/gdal-utils/osgeo_utils/gdal2tiles.py
```diff
@@ -48,7 +48,11 @@
 import tempfile
 import threading
 from functools import partial
-from multiprocessing import Pool
+if 'mpi4py.futures' in sys.modules:
```
rouault, line 51: I'm completely ignorant about mpi4py.futures. Just wondering if there would be use cases where mpi4py.futures would be in sys.module and where using that feature wouldn't be needed, e.g someone using gdal2tiles not as a script but as a function, in a Python session where mpi4py.futures would be loaded ? That's maybe unlikely ?

Otherwise it would be nice to document that new capability in doc/source/programs/gdal2tiles.rst

jcphill: It seems very unlikely that someone calling osgeo_utils.gdal2tiles.main() directly, which is undocumented, would also be using the mpi4py.futures module (which is only likely on an HPC cluster) but not want gdal2tiles to also use the mpi4py.futures module. That said, I can refactor the module to check for an "--mpi" flag and then also modify the script to load mpi4py.futures internally so that rather than "mpiexec ... python -m mpi4py.futures -m osgeo_utils.gdal2tiles ..." the launch command becomes "mpiexec ... gdal2tiles.py --mpi ...".

rouault: > That said, I can refactor the module to check for an "--mpi" flag and then also modify the script to load mpi4py.futures internally so that rather than "mpiexec ... python -m mpi4py.futures -m osgeo_utils.gdal2tiles ..." the launch command becomes "mpiexec ... gdal2tiles.py --mpi ...".

that sounds good

rouault:
@jcphill  is it good to merge ?

PR #5319, opened 2022-02-16
VSICurl: Print response code for failed range requests

julienr:
## What does this PR do?
Improves the error message when a VSI request fails. This is to simplify debugging such issues as knowing the response code can help understand what's the underlying cause.

Example output with the fix:

```
$ gdal_translate -srcwin 10 10 2000 2000 /vsigs/foobar/raster.tif out.tif
Input file size is 170239, 79702
ERROR 1: Request for 1859354131-1859361062 failed with response_code=42
ERROR 1: Request for 1861843881-1861850812 failed with response_code=42
ERROR 1: Request for 1865380401-1865387332 failed with response_code=42
ERROR 1: Request for 1869606905-1869613836 failed with response_code=42
```

### DIFF
--- a/port/cpl_vsil_curl.cpp
+++ b/port/cpl_vsil_curl.cpp
@@ -2118,7 +2118,7 @@ int VSICurlHandle::ReadMultiRange( int const nRanges, void ** const ppData,
                     asWriteFuncHeaderData[iReq].nEndOffset);
 
             CPLError(CE_Failure, CPLE_AppDefined,
-                     "Request for %s failed", rangeStr);
+                     "Request for %s failed with response_code=%ld", rangeStr, response_code);
             nRet = -1;
         }
         else if( nRet == 0 )


port/cpl_vsil_curl.cpp
```diff
@@ -2118,7 +2118,7 @@ int VSICurlHandle::ReadMultiRange( int const nRanges, void ** const ppData,
                     asWriteFuncHeaderData[iReq].nEndOffset);
 
             CPLError(CE_Failure, CPLE_AppDefined,
-                     "Request for %s failed", rangeStr);
+                     "Request for %s failed with response_code=%ld", rangeStr, response_code);
```
rouault, line 2121: you could potentially use asCurlErrors[iRange].szCurlErrBuf[0] to get an error message. I guess 42 must be some curl error

julienr: I should have given more context: I stumbled upon this while trying to debug a transient error on a server where I can't easily update GDAL and I figured that for next time, it would be nice to have the response code in the logs.

But the 42 is coming from me using a local proxy to inject error codes and exercise the error codepath. [Here is the python script](https://gist.github.com/julienr/fe7825cbfbd5a6797aedc79383733a37) I'm using to proxy requests from GDAL to google cloud storage and inject errors.

Now, thanks for the suggestion to look into `asCurlErrors[iRange].szCurlErrBuf`. If I'm adding this to the log, I'm getting the byte range as error message:
```
ERROR 1: _TIFFPartialReadStripArray:Cannot read offset/size for strile around ~3007
ERROR 1: Request for 1969439661-1969448324 failed with response_code=403, error_message=1969439661-1969448324
```

I'm not a curl expert, but reading the doc, I feel it is only valid to use it if the last curl call returned something else than `CURLE_OK` and you can use it to diagnose curl error.

I think there are two categories of errors that can happen:
- curl errors, that is when a curl function returns something else than `CURLE_OK` (connection issue, SSL issue)
- non-2xx http response,

So here I'm trying to address the second (non-2xx http response). In that case, it feels we'd almost want to include the response body in the log because for example google cloud storage will return some XML error message if the response code is 403:
```
?xml version='1.0' encoding='UTF-8'?><Error><Code>AccessDenied</Code><Message>Access denied.</Message><Details>Anonymous caller does not have storage.objects.list access to the Google Cloud Storage bucket.</Details></Error>
```

That could be heavy though and it looks like it could introduce new failure modes while reading this response (what if the error response is super long?) - so I'm not sure it's a good idea.

rouault: > so I'm not sure it's a good idea.

ok I'm fine with the error code only

PR #5323, opened 2022-02-16
GTiff: remove limitation to 32,000 bytes when writing the GDAL metadata tag (fixes #4116)

### LINKED ISSUE #4116: GeoTiff metadata limitation to 32KB
Using gdal_translate to convert a raster, with a lot of metadata, to GeoTiff, a Warning appears:
`Metadata exceeding 32000 bytes cannot be written into GeoTIFF. Transferred to PAM instead.`
The result of gdal_translate operation is a .tiff file (without the metadata) and a .xml file (with the metadata).

rouault:
This limitation came from the initial revision of metadata writing, in e86e2db8d5ddd5d716384f5dd49ab2b435e3879b,
20 years ago. There's no limitation in the TIFF spec nor in libtiff
itself, to text tag data larger than 32 KB, so just remove it.

### DIFF
--- a/autotest/gcore/tiff_write.py
+++ b/autotest/gcore/tiff_write.py
@@ -1257,55 +1257,24 @@ def test_tiff_write_34():
     gdaltest.tiff_drv.Delete('tmp/tw_34.tif')
 
 ###############################################################################
-# Test fallback from internal storage of Geotiff metadata to PAM storage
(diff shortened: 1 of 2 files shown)


rouault:
CC @warmerdam if you happen to remember the reason for the original limitation to 32,000 bytes for GDAL TIFF metadata... ?

PR #5326, opened 2022-02-16
CMake: avoid error when importing broken expat::expat target on Ubuntu 21.10 (fixes #5324)

### LINKED ISSUE #5324: CMake build on Ubuntu 21.10 failing on expat

cmake .

cd builddir
cmake -B . -S ~/git/gdal

## Expected behavior: 
Output ends with
-- Configuring done
-- Generating done

## Actual behavior: 
Output ends with (blank lines reduced):

-- Configuring done
CMake Error in CMakeLists.txt:
  Imported target "expat::expat" includes non-existent path
    "/usr/lib/include"
  in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:
  * The path was deleted, renamed, or moved to another location.
  * An install or uninstall procedure did not complete successfully.
  * The installation package was faulty and references files it does not provide.

CMake Error in CMakeLists.txt:
  Imported target "expat::expat" includes non-existent path
    "/usr/lib/include"
  in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:
  * The path was deleted, renamed, or moved to another location.
  * An install or uninstall procedure did not complete successfully.
  * The installation package was faulty and references files it does not provide.

-- Generating done
CMake Generate step failed.  Build files cannot be regenerated correctly.

Yes, the error is reported twice.

## Operating system
Ubuntu 21.10 64 bit

## GDAL version 
Checked out from github.com/OSGeo/gdal master on 16 Feb 2022.

## Comments
Ubuntu 21.10 (and presumably the LTS version 22.04 due in April)
has a new(ish) version of expat which suffers from
   https://github.com/libexpat/libexpat/issues/501
  
The expat developer hartwork recognised the issue in July 2021 but said:
  Confirming for --includedir= and (for some cases) of --libdir=.  A clean universal fix is not trivial and the issue can be worked 
  around on packaging level (downstream) rather easily, so it's not very high priority right now.

Not sure whether Hartwork means the OS or package level and it is not obvious to me where to look to fix this.

I note that since github.com/OSGeo/gdal/issues/5178 the docker version has an even more recent version of expat.

rouault:


dg0yt:
I must admit that I dislike various aspects of the automatic formatting. It fails to make good use of both dimensions simultaneously.

Is the original issue a particular (upstream) expat version, or just the Ubuntu package?
FTR, for limited cases, there is the option to overwrite the expected targets via the cache variable, i.e.
`-DGDAL_CHECK_PACKAGE_EXPAT_TARGETS=EXPAT::EXPAT`
Similar for the config names.
So it is not mandatory to handle every broken case.

rouault:
> I must admit that I dislike various aspects of the automatic formatting
agreed that it doesn't fit very optimal, anyway...

> Is the original issue a particular (upstream) expat version, or just the Ubuntu package?

I've tried with the latest expat git and couldn't reproduce, so it seems to be (mostly) Ubuntu specific. But I think that we should try, when possible like here, to handle those broken cases in a default configuration, especially for popular environments, otherwise we'll get a lot of support activities to explain people workarounds.

dg0yt:
Okay. 

- I would recommend a "Clean Code" style even for CMake: Create a test function with a clear interface. Similar to `gdal_check_package_target`. Even more now when only two commands fit on the screen ;-)
- I think that the check needs to become part of `gdal_check_package_target`: This is where we already iterate over the list of expected targets. Just don't accept a target here, and you don't need another loop.

rouault:
> * Create a function with a clear interface.

done

dg0yt:
I'm still unhappy with the formatting changes. As it is here now, it also blurs git history and it blurs the core change of this commit.
 And the formatting will probably change again if we add the function signatures to clang-format configuration. I currently don't touch this script, but it would need to resolve merge conflict if it would target other WIP.

rouault:
> I'm still unhappy with the formatting changes

oh well. Disabled in master (15715baaca97e884e20bb428e8a0e68f8e272686)

I've reworked that PR to only include the relevant changes without the (de)formatting

PR #5341, opened 2022-02-17
CMake: fix handling of geotiff-config.cmake

rouault:
Windows Conda builds with the MrSID drivers failed because it couldn't
find the geotiff headers. Other drivers need it but must include the
Conda include directories because of other packages, hence it was
unnoticed.

geotiff-config.cmake doesn't define an explicit CMake target but only
CMake variables.

CC @dg0yt

### DIFF
--- a/cmake/helpers/CheckDependentLibraries.cmake
+++ b/cmake/helpers/CheckDependentLibraries.cmake
@@ -33,6 +33,15 @@ function (gdal_check_target_is_valid target res_var)
         return()
       endif()
     endforeach()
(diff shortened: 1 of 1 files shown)


dg0yt:
Okay. 

`geotiff-config.cmake` seems to focus on providing the variables generally used with Find Modules. Apart from the include dirs, the target `geotiff_library` also lacks a proper name suffix, and doesn't use targets for its interface link libraries. This needs to be improved in libgeotiff.

I hope that there are not too many more surprises left, because we use only a few configs now. Another observation: In CI on Linux, cmake now finds zstd::zst_shared from `linuxbrew` intead of the system's zstd. This has an effect on the include dirs and on the rpath.

PR #5350, opened 2022-02-19
CI: Revise mingw cmake build

dg0yt:
## What does this PR do?

- Uses msys2/setup-msys2@v2 to setup the Mingw64 CMake CI environment, including numpy.
  Mingw + python setup time: 4 min instead of 15 min
- Configures GH Actions cache to actually enable `ccache` effects. 
  Build step with empty cache: 26 min 
  Build step with full cache: 1 min 
  Cache size: 0.03 GB
- Adds post-install test with `pkg-config` for Mingw64.

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/.github/workflows/cmake_builds.yml
+++ b/.github/workflows/cmake_builds.yml
@@ -134,6 +134,7 @@ jobs:
     runs-on: windows-2022
     env:
       generator: MinGW Makefiles
+      cache-name: cmake-mingw64
     defaults:
       run:
         shell: msys2 {0}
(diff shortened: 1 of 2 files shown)


.github/workflows/cmake_builds.yml
```diff
@@ -134,49 +134,79 @@ jobs:
...
+            -DWERROR_DEV_FLAG=-Werror=dev \
+            -DCMAKE_C_COMPILER_LAUNCHER=ccache \
+            -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
+            -DCMAKE_PREFIX_PATH=/mingw64 \
+            -DCMAKE_INSTALL_PREFIX=$PWD/install-gdal \
+            "-DCMAKE_UNITY_BUILD=${CMAKE_UNITY_BUILD}" \
+            -DLIBXML2_INCLUDE_DIR=/mingw64/include -DGDAL_USE_LIBXML2:BOOL=OFF \
+            -DGDAL_USE_MYSQL:BOOL=OFF \
+            -DGDAL_ENABLE_DRIVER_SAR_CEOS:BOOL=OFF \
+            -DOGR_ENABLE_DRIVER_MSSQLSPATIAL:BOOL=OFF \
```
rouault, line 207: should no longer be needed with the fixes I done
```suggestion
```

dg0yt: This was left over, indeed. At least it is no longer hidden in a single super long line.

PR #5338, opened 2022-02-17
CMake: do not generate gdal.pc and gdal-config on MSVC builds

rouault:
With CMake 3.21.3 on a Windows MSVC Conda-based build, I get the
following warnings:
```
CMake Warning at cmake/helpers/GdalGenerateConfig.cmake:47 (message):
  Dropping unsupported generator expression: '$<PLATFORM_ID:Windows>'
Call Stack (most recent call first):
  cmake/helpers/GdalGenerateConfig.cmake:27 (gdal_evaluate_link_genex)
  cmake/helpers/GdalGenerateConfig.cmake:7 (gdal_resolve_link_genex)
  cmake/helpers/GdalGenerateConfig.cmake:119 (gdal_flatten_link_libraries)
  cmake/helpers/GdalGenerateConfig.cmake:193 (gdal_get_lflags)
  gdal.cmake:689 (gdal_generate_config)
  CMakeLists.txt:206 (include)

CMake Warning at cmake/helpers/GdalGenerateConfig.cmake:47 (message):
  Dropping unsupported generator expression: '$<NOT:>'
Call Stack (most recent call first):
  cmake/helpers/GdalGenerateConfig.cmake:27 (gdal_evaluate_link_genex)
  cmake/helpers/GdalGenerateConfig.cmake:7 (gdal_resolve_link_genex)
  cmake/helpers/GdalGenerateConfig.cmake:119 (gdal_flatten_link_libraries)
  cmake/helpers/GdalGenerateConfig.cmake:193 (gdal_get_lflags)
  gdal.cmake:689 (gdal_generate_config)
  CMakeLists.txt:206 (include)

CMake Warning at cmake/helpers/GdalGenerateConfig.cmake:47 (message):
  Dropping unsupported generator expression: '$<:>'
Call Stack (most recent call first):
  cmake/helpers/GdalGenerateConfig.cmake:27 (gdal_evaluate_link_genex)
  cmake/helpers/GdalGenerateConfig.cmake:7 (gdal_resolve_link_genex)
  cmake/helpers/GdalGenerateConfig.cmake:119 (gdal_flatten_link_libraries)
  cmake/helpers/GdalGenerateConfig.cmake:193 (gdal_get_lflags)
  gdal.cmake:689 (gdal_generate_config)
  CMakeLists.txt:206 (include)
```

and the resulting gdal.pc is not super helpful:
```
CONFIG_VERSION=3.5.0dev
CONFIG_INST_PREFIX=C:/Program Files/gdal
[...]

PR #5351, opened 2022-02-19
CMake: Update configuration export

dg0yt:
## What does this PR do?

- Extends the configuration export to handle more generator expressions and target properties.
  - Fixes warning about unsupported generator expressions on Windows.
  - Fixes the selection of per-config properties.
  - Uses import lib property when set.
  - Reduces the use of regex for parsing generator expressions.
- Makes the splitting of libs into path flag and libname (`-L/osgeo -lproj`) optional (`ON` by default except for Window).
- Allows splitting of libs even for the full versioned shared object (`/usr/lib/geos_c.so.16.0.0`).
- Adds quotes when needed to deal with whitespace (`C:\Program Files\...`).
- Prints the generated `gdal.pc` in CI for manual validation. (Should be replaced by a post-install test.)
- Modifies Ubuntu CMake CI to avoid finding package config files from non-system locations. (Happened with zstd from linuxbrew.) Config mode uses the `PATH` environment variable which is full of additional locations on GH actions. This is different from the behaviour of find modules.

Replaces #5338 with regard to silencing the warnings.
I don't want to drop `gdal.pc` file because it should be usable with `pkg-config --msvc-syntax`.
`gdal-config` might be dropped for Windows - or generally for the CMake build, with CMake and pkg-config as successors. (I'm aware that `gdal-config` only needs `sh`.)

## Tasklist

 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments


cmake/helpers/GdalGenerateConfig.cmake
```diff
@@ -21,54 +41,95 @@ endfunction()
...
 # Resolve a single generator expressions for the purpose of exporting linker flags.
-function(gdal_evaluate_link_genex _result _genex)
-    if(_genex MATCHES "^\\\$<(LINK_ONLY|1):(.*)>\$")
-        set(_genex "${CMAKE_MATCH_2}")
-    elseif(_genex MATCHES "^\\\$<([^:]*COMPILER_ID):([^>]*)>\$")
-        string(REPLACE "," ";" compiler_ids "${CMAKE_MATCH_2}")
-        if("${CMAKE_${CMAKE_MATCH_1}}" IN_LIST compiler_ids)
-            set(_genex 1)
+function(gdal_evaluate_link_genex _result _keyword _params)
+    message(STATUS "xxx '${_keyword}' '${_params}'")
```
dg0yt, line 55: To be removed.

cmake/helpers/GdalGenerateConfig.cmake
```diff
@@ -1,3 +1,23 @@
+if (NOT WINDOWS OR MINGW)
+    set(split_libs_default ON)
+else ()
+    set(split_libs_default OFF)
+endif ()
+option(GDAL_SPLIT_EXPORTED_LIBS "Split library path and name on export" ${split_libs_default})
```
dg0yt, line 6: I didn't apply this option to the gdal lib itself: 
It would require a little bit more code to do correctly (static/shared/import...). But it would make sense with regard to consistency.

cmake/helpers/GdalGenerateConfig.cmake
```diff
@@ -84,9 +144,12 @@ function(gdal_split_library_to_lflags _lib_flag _path_flag _input)
         if(_pos EQUAL "-1")
             continue()
         endif()
+        get_filename_component(_lib_dir "${_input}" DIRECTORY)
         string(SUBSTRING "${_lib_name}" 0 "${_pos}" _name_we)
         if(NOT "${_name_we}${_suffix}" STREQUAL _lib_name)
-            continue()
+            if(NOT _suffix STREQUAL ".so" OR NOT EXISTS "${_lib_dir}/${_name_we}${_suffix}")
```
dg0yt, line 150: With the information from https://lists.osgeo.org/pipermail/proj/2022-February/010551.html, it might be more obvious why splitting can be a bad idea: We really don't know if `-lfoo` resolves to `/original/libfoo.so.1.2.3` for the linker and to `libfoo.so.<SOVERSION>` being added to lib gdal.

rouault: > We really don't know if `-lfoo`

I don't think that splitting is a bad idea. When building, -lfoo resolves to finding libfoo.so, which is the development oriented file, and there must be a single one at a time in the directories you link to, and it is a symlink to the actual library libfoo.so.SONAME, and that will be the one kept in the linked library. So if you have a sane environment, linking against /path/libfoo.so.1.2.3, /path/libfoo.so.1 or -Lpath -lfoo will end up in the same result.

cmake/helpers/GdalGenerateConfig.cmake
```diff
@@ -1,3 +1,23 @@
+if (NOT WINDOWS OR MINGW)
```
rouault, line 1: It doesn't look like WINDOWS is a CMake variable, and when looking at that line, I had a second of doubt about operator priority, so I'd suggest:
```suggestion
if ((NOT WIN32) OR MINGW)
```

dg0yt: You are right about `WIN32`. This explains why I didn't see the left-over message output... I really need a few days of holidays ;-)
But there is no need to worry about operator precedence here, it is okay without extra markup.

cmake/helpers/GdalGenerateConfig.cmake
dg0yt: IMO it is not so easy. 
- `-lfoo` prefers shared over static (unless building with `--static`). I acknowledge that static is less relevant but it should be noted nevertheless.
- > there must be a single one at a time in the directories you link to
  
  There can be only one `libfoo.so` in a single directory, but there can be one in *each directory*. In fact there must be one in each prefix *for development*: `-lfoo` is common in pkg-config files, and `libfoo.so` is also what `find_library` looks for.

- `-lfoo` must be complemented by `-L/bar`, but `-L/bar`'s effect is not scoped to `-lfoo`.

  While we can safely append to the explicit search path, the explicit path will be before the implicit search path AFAIU. One might suddenly get a different `libz.so` from `-lfoo`. 
  Basically, you already pointed to the biggest source of trouble in another comment: `find_package` looking directories derived from `PATH`. This is somewhat okay for consistent linking in CMake, with absolute paths. But it gets dangerous when we start to take the lib's directory separately and add it to the linker search path. One could say this is not how CMake config is meant to be used. 
  It may perhaps be summarized as: Don't use `-L` where CMake wouldn't use `-L`.
  
  (Even the current version of `FindPkgConfig.cmake` resolves the given `Libs` into absolute paths for use in the imported target's property.)

rouault: I agree with your observations. I'm not sure if you are asking for specific feedback on what to do ?

rouault:
@dg0yt do you have plans to compete this PR ?

dg0yt:
@rouault Thanks for the reminder. I had my holidays, but came back to a different world. I will need a few more days.

cmake/helpers/GdalGenerateConfig.cmake
dg0yt: Done.

dg0yt:
Tested on x64 linux with vcpkg (release+debug). More tests started in vcpkg CI.

rouault:
@dg0yt is this PR ready for merge ?

dg0yt:
I think it is ready now. I just had to wait for CI to finish.
- pc files passed manual inspection in my local build (Ubuntu 18.04, static linkage for `Debug` and `Release`).
- *pc files passed post-install usage test in GDAL CI* (Ubuntu 20.04, dynamic and static linkage). The log shows the libs implicitly:
  `cc -o test_c test_c.o -L/home/runner/work/gdal/gdal/install-gdal-static/lib -lgdal -ldl -lodbcinst -lcryptopp -lssl -lcrypto -llzma -ldeflate -llz4 -larmadillo -lqhull_r -lSFCGAL -lgmpxx -ltiff -lheif -lzstd -lpoppler -lpodofo -lpdfium -ljpeg -llcms2 -lpng -lgif -lnetcdf -lblosc -lcfitsio -lwebp -lz -lopenjp2 -lCharLS -lIlmImf -lIlmImfUtil -lHalf -lIex -ldap -ldapclient -lkmlbase -lkmldom -lkmlengine -lfyba -lfygm -lfyut -lpcre2-8 -lspatialite -lsqlite3 -lcurl -lxerces-c -lmysqlclient /opt/microsoft/msodbcsql17/lib64/libmsodbcsql-17.9.so.1.1 -lodbc -lfreexl -logdi -lexpat -lxml2 -lgeos_c -lproj -ljson-c -lpthread -lstdc++ -lm` 
  `Testing expected version ... passed`
- pc files can be inspected for manually for the Windows builds in GDAL CI: 
  `Libs.private:  C:/Miniconda/envs/gdalenv/Library/lib/iconv.lib C:/Miniconda/envs/gdalenv/Library/lib/charset.lib C:/Miniconda/envs/gdalenv/Library/lib/libssl.lib C:/Miniconda/envs/gdalenv/Library/lib/libcrypto.lib crypt32 C:/Miniconda/envs/gdalenv/Library/lib/liblzma.lib C:/Miniconda/envs/gdalenv/Library/lib/libdeflate.lib C:/Miniconda/envs/gdalenv/Library/lib/liblz4.lib C:/Miniconda/envs/gdalenv/Library/lib/blosc.lib wbemuuid C:/Miniconda/envs/gdalenv/Library/lib/armadillo.lib C:/Miniconda/envs/gdalenv/Library/lib/mkl_intel_lp64_dll.lib "C:/Program Files/LLVM/lib/libiomp5md.lib" C:/Miniconda/envs/gdalenv/Library/lib/mkl_intel_thread_dll.lib C:/Miniconda/envs/gdalenv/Library/lib/mkl_core_dll.lib C:/Miniconda/envs/gdalenv/Library/lib/qhull_r.lib C:/Miniconda/envs/gdalenv/Library/lib/geotiff_i.lib C:/Miniconda/envs/gdalenv/Library/lib/tiff.lib ws2_32 C:/Miniconda/envs/gdalenv/Library/lib/jpeg.lib C:/Miniconda/envs/gdalenv/Library/lib/libpng.lib C:/Miniconda/envs/gdalenv/Library/lib/Lerc.lib C:/Miniconda/envs/gdalenv/Library/lib/zstd.lib C:/Miniconda/envs/gdalenv/Library/lib/z.lib C:/Miniconda/envs/gdalenv/Library/lib/pcre2-8.lib C:/Miniconda/envs/gdalenv/Library/lib/spatialite_i.lib C:/Miniconda/envs/gdalenv/Library/lib/libexpat.lib C:/Miniconda/envs/gdalenv/Library/lib/sqlite3.lib C:/Miniconda/envs/gdalenv/Library/lib/libcurl.lib C:/Miniconda/envs/gdalenv/Library/lib/xerces-c_3.lib "C:/Program Files (x86)/Windows Kits/10/Lib/10.0.22000.0/um/x64/odbccp32.lib" "C:/Program Files (x86)/Windows Kits/10/Lib/10.0.22000.0/um/x64/WS2_32.Lib" legacy_stdio_definitions.lib odbccp32.lib C:/Miniconda/envs/gdalenv/Library/lib/xml2.lib C:/Miniconda/envs/gdalenv/Library/lib/geos_c.lib psapi C:/Miniconda/envs/gdalenv/Library/lib/proj.lib C:/Miniconda/envs/gdalenv/Library/lib/json-c.lib`

PR #5354, opened 2022-02-20
vrt.rst: expand VRT acronym

neteler:
## What does this PR do?

Explain what "VRT" stands for.


jratike80:
I have been thinking the VRT comes from ViRTual and explains the name of the vector driver https://gdal.org/drivers/vector/vrt.html by the same.

rouault:
> I have been thinking the VRT comes from ViRTual

I also believe this comes from ViRTual

neteler:
I see - then there is probably no need for this change.

PR #5360, opened 2022-02-21
HFA: Fix "Pulkovo 1942" datum write to IMG files

drons:
Default `spheroid.tab` from `Erdas imagine` defines "Pulkovo_1942" datum as "Pulkovo 1942" (without "_" symbol), so we need to translate these names on read/write for correct interaction with `Erdas`.


drons:
Thanks!

PR #5363, opened 2022-02-21
CMake: fix error in gdal_generate_config() when CMAKE_C_IMPLICIT_LINK_LIBRARIES is empty

rouault:
I'm not sure how this can happen exactly but this was reported to me by
a Windows user.

FYI @dg0yt


dg0yt:
What error did occur? Which version of CMake?

I can only test CMake 3.20.1 today:
~~~cmake
set(a "")
set(b "")
list(REMOVE_ITEM a ${b})
list(REMOVE_ITEM a ${unset})
list(REMOVE_ITEM unset ${unset})
~~~
..no problems

rouault:
> What error did occur?

"list sub-command REMOVE_ITEM requires two or more arguments"

> Which version of CMake?

cmake 3.18.1

anyway it should be harmless to merge this, right ?

dg0yt:
Great... May be it would be shorter and more obvious to add `CMake_3.19_quirk` to the list of removed items, instead of adding `if`?

dg0yt:
Oh, merges while I was commenting...

PR #5367, opened 2022-02-22
gdal2tiles: move workaround out of wrapper script

jcphill:
An incompatibility between pdb and multiprocessing using the
'spawn' start method was being worked around by setting
__spec__ = None in the gdal2tiles wrapper script.

Moved workaround into the gdal2tiles module and made more general by
fixing __spec__ regardless of where the __main__ module is defined.


rouault:
did you try this with python 3.8 or above on Windows, using regular multiprocessing, and using the launcher script for the wrapper script which is in ${PYTHONHOME}\Scripts\gdal2tiles.py ?

rouault:
> and can confirm that this change preserves the fix.

ok cool

PR #5371, opened 2022-02-22
Pixelfn metadata

mmomtchev:
## What does this PR do?

  * Implements metadata for C++ pixel functions
  * Adds two new types of arguments, to be specified in the metadata: builtins (currently `NoData`, `scale` and `offset`) and constants
  * Adds two new builtin functions: `replace_nodata` and `scale`

## Tasklist

 - [x] Add test case(s)
 - [x] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/autotest/cpp/data/pixelfn.vrt
+++ b/autotest/cpp/data/pixelfn.vrt
@@ -4,7 +4,7 @@
     <PixelFunctionType>custom</PixelFunctionType>
     <SourceTransferType>Float64</SourceTransferType>
     <SimpleSource>
-      <SourceFilename relativeToVRT="0">data/float32.tif</SourceFilename>
+      <SourceFilename relativeToVRT="1">float32.tif</SourceFilename>
     </SimpleSource>
   </VRTRasterBand>
 </VRTDataset>
--- a/autotest/cpp/data/pixelfn.vrt
+++ b/autotest/cpp/data/pixelfn.vrt
@@ -1,0 +1,10 @@
+<VRTDataset rasterXSize="20" rasterYSize="20">
+  <VRTRasterBand dataType="Float64" band="1" subClass="VRTDerivedRasterBand">
+    <Description>CustomPixelFn</Description>
+    <PixelFunctionType>custom</PixelFunctionType>
+    <SourceTransferType>Float64</SourceTransferType>
+    <SimpleSource>
+      <SourceFilename relativeToVRT="0">data/float32.tif</SourceFilename>
+    </SimpleSource>
+  </VRTRasterBand>
+</VRTDataset>
--- a/autotest/cpp/test_gdal_pixelfn.cpp
+++ b/autotest/cpp/test_gdal_pixelfn.cpp
@@ -106,7 +106,7 @@ CPLErr CustomPixelFuncWithMetadata( void **papoSources, int nSources, void *pDat
     {
         for( int iCol = 0; iCol < nXSize; ++iCol, ++ii )
         {
-            const double dfPixVal = GetSrcVal(papoSources[0], eSrcType, ii);
+            const double dfPixVal = GetSrcVal(papoSources[0], eSrcType, ii) * 2;
             GDALCopyWords(
                     &dfPixVal, GDT_Float64, 0,
                     static_cast<GByte *>(pData) + static_cast<GSpacing>(nLineSpace) * iLine +
@@ -134,7 +134,7 @@ CPLErr CustomPixelFunc( void **papoSources, int nSources, void *pData,
     {
         for( int iCol = 0; iCol < nXSize; ++iCol, ++ii )
         {
-            const double dfPixVal = GetSrcVal(papoSources[0], eSrcType, ii);
+            const double dfPixVal = GetSrcVal(papoSources[0], eSrcType, ii) * 3;
             GDALCopyWords(
                     &dfPixVal, GDT_Float64, 0,
                     static_cast<GByte *>(pData) + static_cast<GSpacing>(nLineSpace) * iLine +
@@ -160,7 +160,7 @@ CPLErr CustomPixelFuncNoArgs( void **papoSources, int nSources, void *pData,
     {
         for( int iCol = 0; iCol < nXSize; ++iCol, ++ii )
         {
-            const double dfPixVal = GetSrcVal(papoSources[0], eSrcType, ii);
+            const double dfPixVal = GetSrcVal(papoSources[0], eSrcType, ii) * 4;
             GDALCopyWords(
                     &dfPixVal, GDT_Float64, 0,
                     static_cast<GByte *>(pData) + static_cast<GSpacing>(nLineSpace) * iLine +
@@ -204,7 +204,7 @@ namespace tut
         float buf[20 * 20];
         CPL_IGNORE_RET_VAL(GDALRasterIO(band, GF_Read, 0, 0, 20, 20, buf, 20, 20, GDT_Float32, 0, 0));
 
-        ensure_equals("Read wrong data", buf[0], 107);
+        ensure_equals("Read wrong data", buf[0], 107 * 2);
 
         GDALClose(ds);
     }
@@ -229,7 +229,7 @@ namespace tut
         CPL_IGNORE_RET_VAL(GDALRasterIO(
           band, GF_Read, 0, 0, 20, 20, buf, 20, 20, GDT_Float32, 0, 0));
 
-        ensure_equals("Read wrong data", buf[0], 107);
+        ensure_equals("Read wrong data", buf[0], 107 * 3);
 
         GDALClose(ds);
     }
@@ -254,7 +254,7 @@ namespace tut
         CPL_IGNORE_RET_VAL(GDALRasterIO(
           band, GF_Read, 0, 0, 20, 20, buf, 20, 20, GDT_Float32, 0, 0));
 
-        ensure_equals("Read wrong data", buf[0], 107);
+        ensure_equals("Read wrong data", buf[0], 107 * 4);
 
         GDALClose(ds);
     }
--- a/autotest/cpp/test_gdal_pixelfn.cpp
+++ b/autotest/cpp/test_gdal_pixelfn.cpp
@@ -32,15 +32,24 @@
 #include "gdal_alg.h"
 #include "gdal_priv.h"
 #include "gdal.h"
+#include "../frmts/vrt/vrtdataset.h"
 
 #include <sstream>
 #include <string>
 #include <vector>
 
+CPLErr CustomPixelFuncWithMetadata( void **papoSources, int nSources, void *pData,
+                            int nXSize, int nYSize,
+                            GDALDataType eSrcType, GDALDataType eBufType,
+                            int nPixelSpace, int nLineSpace, CSLConstList papszArgs );
 CPLErr CustomPixelFunc( void **papoSources, int nSources, void *pData,
                             int nXSize, int nYSize,
                             GDALDataType eSrcType, GDALDataType eBufType,
                             int nPixelSpace, int nLineSpace, CSLConstList papszArgs );
+CPLErr CustomPixelFuncNoArgs( void **papoSources, int nSources, void *pData,
+                            int nXSize, int nYSize,
+                            GDALDataType eSrcType, GDALDataType eBufType,
+                            int nPixelSpace, int nLineSpace);
 
 template<typename T>
 inline double
@@ -78,7 +87,7 @@ GetSrcVal(const void* pSource, GDALDataType eSrcType, T ii)
     return 0;
 }
 
-CPLErr CustomPixelFunc( void **papoSources, int nSources, void *pData,
+CPLErr CustomPixelFuncWithMetadata( void **papoSources, int nSources, void *pData,
                             int nXSize, int nYSize,
                             GDALDataType eSrcType, GDALDataType eBufType,
                             int nPixelSpace, int nLineSpace, CSLConstList papszArgs ) {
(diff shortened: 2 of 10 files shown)


autotest/cpp/data/pixelfn.vrt
```diff
@@ -0,0 +1,10 @@
+<VRTDataset rasterXSize="20" rasterYSize="20">
+  <VRTRasterBand dataType="Float64" band="1" subClass="VRTDerivedRasterBand">
+    <Description>CustomPixelFn</Description>
+    <PixelFunctionType>custom</PixelFunctionType>
+    <SourceTransferType>Float64</SourceTransferType>
+    <SimpleSource>
+      <SourceFilename relativeToVRT="0">data/float32.tif</SourceFilename>
```
rouault, line 7:
```suggestion
      <SourceFilename relativeToVRT="1">float32.tif</SourceFilename>
```

autotest/cpp/test_gdal_pixelfn.cpp
```diff
@@ -0,0 +1,262 @@
...
+ * DEALINGS IN THE SOFTWARE.
+ ****************************************************************************/
+
+#include "gdal_unit_test.h"
+
+#include "cpl_string.h"
+#include "gdal_alg.h"
+#include "gdal_priv.h"
+#include "gdal.h"
+#include "../frmts/vrt/vrtdataset.h"
```
rouault, line 35: like without effect as this would resolve to autotest/frmts/vrt/vrtdataset.h

frmts/vrt/pixelfunctions.cpp
```diff
@@ -1165,6 +1222,99 @@ CPLErr InterpolatePixelFunc( void **papoSources, int nSources, void *pData,
     return CE_None;
 }
 
+static const char pszNanPixelFuncMetadata[] =
+"<PixelFunctionArgumentsList>"
+"   <Argument type='builtin' value='NoData'>"
+"   </Argument>"
+"</PixelFunctionArgumentsList>";
+
+static CPLErr NanPixelFunc( void **papoSources, int nSources, void *pData,
```
rouault, line 1231: a more generic function would be ReplaceNoData() that would take a OutputNoData value that could be nan

frmts/vrt/pixelfunctions.cpp
```diff
@@ -1165,6 +1222,99 @@ CPLErr InterpolatePixelFunc( void **papoSources, int nSources, void *pData,
...
+    if ( FetchDoubleArg(papszArgs, "NoData", &dfNoData) != CE_None ) return CE_Failure;
+
+    /* ---- Set pixels ---- */
+    size_t ii = 0;
+    for( int iLine = 0; iLine < nYSize; ++iLine )
+    {
+        for( int iCol = 0; iCol < nXSize; ++iCol, ++ii )
+        {
+            double dfPixVal = GetSrcVal(papoSources[0], eSrcType, ii);
+            if (dfPixVal == dfNoData) dfPixVal = NAN;
```
rouault, line 1260:
```suggestion
            if (dfPixVal == dfNoData || std::isnan(dfPixVal)) dfPixVal = ... outputNoData ...;
```

frmts/vrt/pixelfunctions.cpp
```diff
@@ -1165,6 +1222,99 @@ CPLErr InterpolatePixelFunc( void **papoSources, int nSources, void *pData,
...
+                               int nPixelSpace, int nLineSpace, CSLConstList papszArgs ) {
+    /* ---- Init ---- */
+    if( nSources != 1 ) return CE_Failure;
+    if (GDALDataTypeIsComplex(eSrcType))
+    {
+        CPLError(
+          CE_Failure, CPLE_AppDefined, "nan cannot convert complex data types");
+        return CE_Failure;
+    }
+    if (!GDALDataTypeIsFloating(eBufType))
```
rouault, line 1243: to remove if using something more generic

frmts/vrt/vrtderivedrasterband.cpp
```diff
@@ -743,6 +747,73 @@ bool VRTDerivedRasterBand::InitializePython()
     return true;
 }
 
+CPLErr
+VRTDerivedRasterBand::GetPixelFunctionArguments(
+  const CPLString& osMetadata,
+  std::vector<std::pair<CPLString, CPLString>>& oAdditionalArgs)
+{
+
+    CPLXMLNode* oArgs = CPLParseXMLString(osMetadata);
```
rouault, line 756: to avoid memory leak:
```suggestion
    auto poArgs = CPLXMLTreeCloser(CPLParseXMLString(osMetadata));
```

frmts/vrt/vrtderivedrasterband.cpp
```diff
@@ -743,6 +747,73 @@ bool VRTDerivedRasterBand::InitializePython()
...
+    if (oArgs != nullptr && oArgs->eType == CXT_Element &&
+        !strncmp(oArgs->pszValue,
```
rouault, line 758: why not just a plain strcmp() ? (applies to below too)

frmts/vrt/vrtderivedrasterband.cpp
```diff
@@ -743,6 +747,73 @@ bool VRTDerivedRasterBand::InitializePython()
...
+                 "PixelFunctionArgumentsList",
+                 strlen("PixelFunctionArgumentsList")))
+    {
+        for (CPLXMLNode* psIter = oArgs->psChild; psIter != nullptr;
+             psIter = psIter->psNext)
+        {
+            if (psIter->eType == CXT_Element &&
+                !strncmp(psIter->pszValue, "Argument", strlen("Argument")))
+            {
+                CPLString szName, szType, szValue;
```
rouault, line 768:
```suggestion
                CPLString osName, osType, osValue;
```

frmts/vrt/vrtderivedrasterband.cpp
```diff
@@ -743,6 +747,73 @@ bool VRTDerivedRasterBand::InitializePython()
...
+                                 "PixelFunction builtin %s not supported",
+                                 szValue.c_str());
+                        return CE_Failure;
+                    }
+                    // Should we signal the user that he is using a builtin
+                    // that depends on a value that is not set on the raster?
+                    // Maybe we should allow generalized use of scale/offset
+                    // even if they are left with their default 1/0 values
+                    oAdditionalArgs.push_back(std::pair<CPLString, CPLString>(
+                      szValue, CPLSPrintf("%lf", dfVal)));
```
rouault, line 804: to get full double accuracy:
```suggestion
                      szValue, CPLSPrintf("%.18g", dfVal)));
```

frmts/vrt/vrtderivedrasterband.cpp
```diff
@@ -743,6 +747,73 @@ bool VRTDerivedRasterBand::InitializePython()
...
+                                 CPLE_NotSupported,
+                                 "PixelFunction builtin %s not supported",
+                                 szValue.c_str());
+                        return CE_Failure;
+                    }
+                    // Should we signal the user that he is using a builtin
+                    // that depends on a value that is not set on the raster?
+                    // Maybe we should allow generalized use of scale/offset
+                    // even if they are left with their default 1/0 values
+                    oAdditionalArgs.push_back(std::pair<CPLString, CPLString>(
```
rouault, line 803: we should definitely not pass values if success != TRUE

frmts/vrt/vrtderivedrasterband.cpp
```diff
@@ -876,6 +948,15 @@ CPLErr VRTDerivedRasterBand::IRasterIO( GDALRWFlag eRWFlag,
                     this->pszFuncName) ;
             return CE_Failure;
         }
+
+        if (poPixelFunc->second != "")
+        {
+            if ((GetPixelFunctionArguments(poPixelFunc->second,
```
rouault, line 954: extraneous parenthesis

autotest/cpp/test_gdal_pixelfn.cpp
mmomtchev: It does build without it, isn't this relative to the directory you are building from?

frmts/vrt/pixelfunctions.cpp
mmomtchev: No more `CPLIsNan`?

autotest/cpp/test_gdal_pixelfn.cpp
rouault: > isn't this relative to the directory you are building from?

I would have said relative to file, but in any case the build directory is autotest/cpp, so that's equivalent. Actually, this only works because there are include paths like -I/home/even/gdal/gdal/gcore which make things eventually work.

mmomtchev: I couldn't make it work with `vrt/vrtdataset.h` - what do you suggest?

frmts/vrt/vrtderivedrasterband.cpp
mmomtchev: No, in fact it was because of `CPL_IGNORE_RET_VAL` in the test

autotest/cpp/test_gdal_pixelfn.cpp
rouault: let's keep it that way ( #include "../../frmts/vrt/vrtdataset.h" would also work )

autotest/cpp/test_gdal_pixelfn.cpp
```diff
@@ -0,0 +1,262 @@
...
+                            GDALDataType eSrcType, GDALDataType eBufType,
+                            int nPixelSpace, int nLineSpace, CSLConstList papszArgs );
+CPLErr CustomPixelFuncNoArgs( void **papoSources, int nSources, void *pData,
+                            int nXSize, int nYSize,
+                            GDALDataType eSrcType, GDALDataType eBufType,
+                            int nPixelSpace, int nLineSpace);
+
+template<typename T>
+inline double
+GetSrcVal(const void* pSource, GDALDataType eSrcType, T ii)
```
avalentino, line 56: This is the duplicate of a function in ``pixelfunctions.cpp``.
Maybe we could move it to a common header ``pixelfunctions.hpp``.

frmts/vrt/pixelfunctions.cpp
```diff
@@ -1165,6 +1212,99 @@ CPLErr InterpolatePixelFunc( void **papoSources, int nSources, void *pData,
...
+"   <Argument type='builtin' value='scale' />"
+"</PixelFunctionArgumentsList>";
+
+static CPLErr ScalePixelFunc( void **papoSources, int nSources, void *pData,
+                               int nXSize, int nYSize,
+                               GDALDataType eSrcType, GDALDataType eBufType,
+                               int nPixelSpace, int nLineSpace, CSLConstList papszArgs ) {
+    /* ---- Init ---- */
+    if( nSources != 1 ) return CE_Failure;
+    if (GDALDataTypeIsComplex(eSrcType))
```
avalentino, line 1277: In principle scale and offset could also be applied to a complex band, even if we have the limitation that ``offset`` an d``scale`` factors con be only real numbers.

avalentino (approved):
LGTM
Maybe we could add some example for the builtin functions in the documentation.

autotest/cpp/test_gdal_pixelfn.cpp
mmomtchev: Yes, but in this case it is to be considered public API and to be properly referenced.

I also suggest that both the OSGeo and OpenJS foundations finally realize and come to terms with the situation, as accidently running down a pedestrian is obviously a problem, but finishing him off is a huge gamble and rarely has the intended result.

avalentino: > Yes, but in this case it is to be considered public API and to be properly referenced.

I don't think that we are forced to consider it a public API to to install the new ``pixelfunctions.hpp``.

rouault: > to install the new pixelfunctions.hpp.

Actually we have the SRCVAL() macro in gdal.h thta does a similar job

mmomtchev: Yes,  I saw it, I was thinking of a way to merge both.
As this is performance-critical, I am also thinking of a single switch, performed once per pixel function call, and then using a lambda to cast each pixel.

rouault: actually SRCVAL() is the historical way of doing things, but for internal pixel functions, I introduced the template in 9048844951bd8b31f179f709126a1c3fb0cd4cc1 (to fix a cppcheck warning)
More than this switch() if you're concerned about performance, the GDALCopyWords() function call is certainly much more heavy weight. The GDALCopyWord() template function (in gcore/gdal_priv_templates.hpp) could potentially be used.
But then you need to have a template on the source and target data types.

mmomtchev: Templated functions wins for every compiler:
https://gist.github.com/mmomtchev/ae0251f0d400db625b9d33069237a547

-O3 everywhere

g++ 9.3:
templated function: 520 ms
macro: 606 ms
function pointer: 628 ms
lambda: 786 ms

clang 12:
templated function: 512 ms
macro: 512 ms
function pointer: 608 ms
lambda: 822 ms

intel compiler 2022:
templated function: 469 ms
macro: 470 ms
function pointer: 630 ms
lambda: 785 ms

rouault:
Note: the CMake build-mac failure is unrelated and I've just fixed in master per https://github.com/OSGeo/gdal/commit/df7e86c30d0b77731c06a86707cdcb5932957267

rouault:
Looks goot to me. Anything  remaining on your side @mmomtchev  ?

frmts/vrt/pixelfunctions.cpp
mmomtchev: I will push this in its current form - I have never used complex numbers Datasets and I can't evaluate if these make sense or not

PR #5384, opened 2022-02-24
CI: fix and enable msys2 tests

rouault:


.github/workflows/cmake_builds.yml
```diff
@@ -173,25 +173,27 @@ jobs:
       run:
         shell: msys2 {0}
     steps:
+      # To avoid git clone to mess with the line endings of GDAL autotest data
+      # files that look like text, but should be handled as binary content
+      - name: Set git core.autocrlf to false
+        shell: bash -l {0}
```
dg0yt, line 179: This ...

.github/workflows/cmake_builds.yml
```diff
@@ -235,18 +236,23 @@ jobs:
...
-      #- name: test
-      #  run: cmake --build build --target quicktest
-      #  working-directory: ${{ github.workspace }}
-      #  env:
-      #    PROJ_LIB: "C:\\tools\\msys64\\mingw64\\share\\proj"
+      - name: test
+        run: cmake --build build --target quicktest
+        working-directory: ${{ github.workspace }}
+      - name: test (with ctest)
+        shell: bash -l {0}
```
dg0yt, line 248: ... and this will use a different `bash` and msys runtime than the other commands (which run in a msys2 wrapper). I'm concerned that this may be fragile.

rouault: > I'm concerned that this may be fragile.

for the git part, this was necessary otherwise there were wrong line endings (I suppose that setting the options of the git under the msys2 shell had no effect on the actions/checkout@v2)
for the tests, this is indeed not intended / a copy&paste from another config. I'll issue a PR to remove it

PR #5386, opened 2022-02-25
MRF: Lerc refactor for external libLerc

lucianpls:
## What does this PR do?

Changed MRF LERC to use libLerc C API

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:


lucianpls:
@rouault 

I'm not sure why the link with libLerc is failing in this case, can you please take a look?

Thanks!

rouault:
Looking at the build log of the failed Conda windows build, one sees:
```
cl   /nologo /MP2 /MD /EHsc /Ox /FC /D_CRT_SECURE_NO_DEPRECATE /D_CRT_NONSTDC_NO_DEPRECATE /DNDEBUG /W4 /wd4127 /wd4251 /wd4275 /wd4786 /wd4100 /wd4245 /wd4206 /wd4351 /wd4611  /DHAVE_SSE_AT_COMPILE_TIME /DHAVE_SSSE3_AT_COMPILE_TIME -I..\..\port -I..\..\ogr -I..\..\gcore  -I..\..\alg -I..\..\ogr\ogrsf_frmts  -I..\..\gnm -I..\..\gnm\gnm_frmts -I..\..\apps /DHAVE_AVX_AT_COMPILE_TIME -I%PREFIX%\Library\include -I../zlib -I%PREFIX%\Library\include -DZSTD_SUPPORT -DLERC -ILERCV1 -I../../third_party/LercLib  -I%PREFIX%\Library\include  -DGNM_ENABLED  -DGDAL_COMPILATION -DNOMINMAX /c LERC_Band.cpp Packer_RLE.cpp Tif_band.cpp PNG_band.cpp JPEG_band.cpp JPNG_band.cpp mrf_band.cpp mrf_overview.cpp mrf_util.cpp marfa_dataset.cpp 
```
One can notice that we have several ``-I%PREFIX%\Library\include`` inclusions before -I../../third_party/LercLib  . But Conda-Forge now ships with external LERC, so in %PREFIX%\Library\include there are the Lerc2 headers. So we build against them, using not the gdal_ LERC prefixed symbols, but at runtime the build recipee doesn't link against that external LercLib.
The likely fix is to edit in frmts/mrf/makefile.vc all lines that prepend include directories to EXTRAFLAGS, like ``EXTRAFLAGS =    -I$(JPEGDIR) $(EXTRAFLAGS)`` (but it is not the only one), to make sure instead that they append them (and also move the EXTRAFLAGS = $(EXTRAFLAGS) $(ZSTD_CFLAGS) line after the Lerc ones), so that -I../../third_party/LercLib is first.

rouault:
in cmake/helpers/CheckDependentLibraries.cmake, you should also remove the DISABLED_BY_DEFAULT in
```
# DISABLED_BY_DEFAULT since it prevents MRF Lerc support (see frmts/mrf/CMakeLists.txt)
gdal_check_package(LERC "Enable LERC (external)" CAN_DISABLE DISABLED_BY_DEFAULT)
```

rouault:
> in cmake/helpers/CheckDependentLibraries.cmake, you should also remove the DISABLED_BY_DEFAULT in

and doc/source/build_hints.rst should be amended with the following patch:
```patch
diff --git a/doc/source/build_hints.rst b/doc/source/build_hints.rst
index 23cd959a69..c13cc9dc2f 100644
--- a/doc/source/build_hints.rst
+++ b/doc/source/build_hints.rst
@@ -863,13 +863,6 @@ which supports rapid encoding and decoding for any pixel type (not just RGB or B
 Users set the maximum compression error per pixel while encoding, so the precision
 of the original input image is preserved (within user defined error bounds).
 
-.. warning::
-
-    Use of the external LERC library is not recommended, as it cannot be used
-    by the :ref:`raster.marfa` driver currently (that one requires the internal
-    LERC copy). The external LERC Library can only be used by the internal libtiff,
-    which can also use the internal LERC copy.
-
 
 .. option:: LERC_INCLUDE_DIR
 
@@ -881,12 +874,12 @@ of the original input image is preserved (within user defined error bounds).
 
 .. option:: GDAL_USE_LERC=ON/OFF
 
-    Control whether to use LERC (V2). Defaults to *OFF* when LERC (V2) is found.
+    Control whether to use LERC (V2). Defaults to ON when LERC (V2) is found.
 
 .. option:: GDAL_USE_LERC_INTERNAL=ON/OFF
 
-    Control whether to use the LERC (V2) internal library. Defaults to ON,
-    unless GDAL_USE_LERC is set to ON.
+    Control whether to use the LERC (V2) internal library. Defaults to ON when
+    external LERC (V2) library is not found.
 
 
 LERCV1
```

lucianpls:
@rouault 
Should we consider removing libLerc source from gdal at some later time?

rouault:
> Should we consider removing libLerc source from gdal at some later time?

maybe... its external packaging history is still recent though

rouault:
actually for things to be reliable, a structure similar to what is done in frmts/mrf/CMakeLists.txt should be done: first include all the internal libraries, and then all the external ones.
The failure in the fedora build has been fixed in master, and can be solved by a rebase on top of it.

frmts/mrf/mrf_util.cpp
```diff
@@ -537,6 +537,8 @@ void GDALRegister_mrf() {
 #if defined(LERC)
         "       LERC_PREC - numeric, set LERC precision, defaults to 0.5 for int and 0.001 for float\n"
         "       V1 - boolean, enable LERC V1 (older) format\n"
+        "       L2_VER - numeric, 2+, encode Lerc for a specific version, defaults to library default\n"
+        "                for single band or INTERLEAVE=BAND, LERC2_VER defaults to 2"
```
rouault, line 541:
```suggestion
        "                for single band or INTERLEAVE=BAND, L2_VER defaults to 2"
```

frmts/mrf/mrf_util.cpp
```diff
@@ -537,6 +537,8 @@ void GDALRegister_mrf() {
 #if defined(LERC)
         "       LERC_PREC - numeric, set LERC precision, defaults to 0.5 for int and 0.001 for float\n"
         "       V1 - boolean, enable LERC V1 (older) format\n"
+        "       L2_VER - numeric, 2+, encode Lerc for a specific version, defaults to library default\n"
```
rouault, line 540: L2_VER is the minor Lerc 2.X version number, right ?  Perhaps this should be more explicit

PR #5390, opened 2022-02-25
Add VSISetCredential() to set /vsi[s3|gs|az|...] credentials for a given file prefix

rouault:
which makes it easier if using the same virtual file system but
with different credentials (e.g. different credentials for bucket
"/vsis3/foo" and "/vsis3/bar")

### DIFF
--- a/autotest/gcore/vsicredential.py
+++ b/autotest/gcore/vsicredential.py
@@ -1,0 +1,70 @@
+#!/usr/bin/env pytest
+###############################################################################
+# $Id$
+#
+# Project:  GDAL/OGR Test Suite
+# Purpose:  Test VSI credentials
+# Author:   Even Rouault <even dot rouault at spatialys dot com>
+#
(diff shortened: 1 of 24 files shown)


sgillies:
@rouault I read the note at https://github.com/OSGeo/gdal/pull/5390/files#diff-fcdb54fe21947da57f39b0a3ad50acabda307d6be20a2af65e886b49286c4421R155 but would like to hear more about the use case for this as I expect it to cross over into rasterio. Are you seeing that people want to have VRTs with source files that are in a mix of buckets or clouds?

rouault:
> Are you seeing that people want to have VRTs with source files that are in a mix of buckets or clouds?

Not necessarily VRTs, but people doing some processing with GDAL and having different buckets with the same cloud provider but with different credentials. Up to now, they have to use CPLSetConfigOption() each time to reset them when they switch between files.

PR #5397, opened 2022-02-28
ADD facility to search points included in search ellipse for GDALDataMetrics (minimum, maximum, range, count, average_distance, average_distance_pts) in the already implemented quadtree data structure.

FeU-aKlos:
## What does this PR do?
This PR facilitates the opportunity to perform  the search which points lie within the search ellipse by utilizing the quadtree datastructure. If quadtree is utilized or not is decided as for GDALGridNearestNeighbor (nPoints > 100 && radius1==radius2 && searchradius>0)


alg/gdalgrid.cpp
```diff
@@ -2646,10 +2912,134 @@ CPLErr ParseAlgorithmAndOptions( const char *pszAlgorithm,
             break;
         }
         case GGA_MetricMinimum:
+        {
```
rouault, line 2915: that block as well as for the other metric constants seem to be a copy&paste of the generic one that already existed and applied to all of them. Why is that needed ?

alg/gdalgrid.cpp
```diff
@@ -1901,6 +2149,9 @@ GDALGridContextCreate( GDALGridAlgorithm eAlgorithm, const void *poOptions,
             memcpy(poOptionsNew, poOptions, sizeof(GDALGridDataMetricsOptions));
 
             pfnGDALGridMethod = GDALGridDataMetricMinimum;
+            bCreateQuadTree = (nPoints > 100 &&
```
rouault, line 2152: I believe we should also add dfAngle == 0 as a condition here and in the below case (including for the nearest neighbour case), as the formulas in the quadtree case don't take into account rotation.

alg/gdalgrid.cpp
FeU-aKlos: @rouault Good point. Did not see the default in the switch statement. Is indeed not needed!

alg/gdalgrid.cpp
FeU-aKlos: @rouault Yes, of course, thsi would make absolute sense.

alg/gdalgrid.cpp
FeU-aKlos: what about changing the switch statement so that the default case contains everything of the current GGA_MetricMinimum  case? Otherwise, the GGA_Metric* cases would include stuff like power, smoothing and max_points, which are not needed in those cases.

rouault: > what about changing the switch statement 

I would just revert those changes. All the GGA_Metric* cases will go to the "case GGA_MetricAverageDistancePts:" which does the right thing

FeU-aKlos (commented):
code has been changed regarding to the feedback of @rouault

FeU-aKlos (commented):
Incorporated feedback

rouault:
this cause changes (regressions ?) in test cases

FeU-aKlos:
Do I understand correct, that this should not be the case? Basically, the quadtree data structure should make the search of neighboring points more efficient. But why should the result be different?

FeU-aKlos:
> this cause changes (regressions ?) in test cases

sorry, that is my first contribution to a project on github, should I do the adjustment of the test cases in a new pull request? Is it okay to adjust the errors in the asserts in the test functions and simply replace the reference data or do I have to write completely new tests ?

rouault:
> in a new pull request?

no, changes must be queued in that pull request. But you need to investigate if you added a bug, or if there's a good reason to adjust the existing test cases.

as you've added optimized branches, we'd want to make sure that both the optimized and exisitng non-optimized branches are tested, which will require new test cases

FeU-aKlos (commented):
REFACTOR the if condition for each metric and nn. Checks are done preliminary while checking, if quadtree should be created so that most checks only need to be performed once. Remove CPL_UNUSED from hExtraParamsIn in datametric functions.
FIX dfSearchRadius bug. Was before equal to  the value set in hExtraParamsIn, not its set to dfRadius1.

FeU-aKlos (commented):
REMOVE unnecessary bracket

FeU-aKlos:
> > in a new pull request?
> 
> no, changes must be queued in that pull request. But you need to investigate if you added a bug, or if there's a good reason to adjust the existing test cases.
> 
> as you've added optimized branches, we'd want to make sure that both the optimized and exisitng non-optimized branches are tested, which will require new test cases

Was my fault.. seems to work now. Thanks for your suggestions. Checking test_gdal_grid.py, I figure out, that both branches are evaluated for: maximum, range, count, average_distance, nearest. For minimum and average_distance_pts new tests would be required, if I did not miss anything ...

FeU-aKlos:
> > in a new pull request?
> 
> no, changes must be queued in that pull request. But you need to investigate if you added a bug, or if there's a good reason to adjust the existing test cases.
> 
> as you've added optimized branches, we'd want to make sure that both the optimized and exisitng non-optimized branches are tested, which will require new test cases

would it be up to me, to provide new tests? or should they be written from someone else, to make sure, that the tests are not too optimistic?

rouault:
> would it be up to me, to provide new tests? or should they be written from someone else, to make sure, that the tests are not too optimistic?

it would be nice indeed that you provide new tests, at least that exercise the code you modified/added (bonus point for other existing untested code paths, but obviously we can't require that)

rouault:
I was thinking that to test the case of nPoints <= 100 vs nPoints > 100, instead of hardcoding this 100 value, you could use a  "const int nPointCountTreshold = atoi(CPLGetConfigOption("GDAL_GRID_POINT_COUNT_THRESHOLD", "100"));"  , and then in autotest, you would just have to use a pattern like
```
@pytest.mark.parametrize("use_quadtree", [True, False])
def test_mytest(use_quadtree):
    with gdaltest.config_option('GDAL_GRID_POINT_COUNT_THRESHOLD', '0' if use_quadtree else '1000000000'):
          ....
```

bnordgren:
Thanks for this PR! I am curious if the omission of the MovingAverage and InverseDistanceToAPower is intentional, an oversight, or just not needed for the moment?  It seems weird to have the same point-search-algorithm in all the functions except two (?) In any case, my main use case would be MovingAverage, so I'd be in line right behind you to add the QuadTree to that algorithm as long as there's no conscious "for-cause" decision to exclude it from the QuadTree goodness. 

Just spitballing here, but how hard would it be (and what would the performance impact be) to implement each of these point search algorithms once (so there's only two), allowing the caller to specify an accumulator class that implements the arithmetic algorithm (average, average distance, count, max, min, range, etc.) It seems that this would be a strategy that would make code maintenance a bit less tedious down the road. 

I entirely accept that this PR carries forward the existing design decisions for the gdal_grid family of algorithms, so refactoring the code as above may be completely out of scope. :) 

Thanks again for your work on this!!

FeU-aKlos:
> Thanks for this PR! I am curious if the omission of the MovingAverage and InverseDistanceToAPower is intentional, an oversight, or just not needed for the moment? It seems weird to have the same point-search-algorithm in all the functions except two (?) In any case, my main use case would be MovingAverage, so I'd be in line right behind you to add the QuadTree to that algorithm as long as there's no conscious "for-cause" decision to exclude it from the QuadTree goodness.
> 
> Just spitballing here, but how hard would it be (and what would the performance impact be) to implement each of these point search algorithms once (so there's only two), allowing the caller to specify an accumulator class that implements the arithmetic algorithm (average, average distance, count, max, min, range, etc.) It seems that this would be a strategy that would make code maintenance a bit less tedious down the road.
> 
> I entirely accept that this PR carries forward the existing design decisions for the gdal_grid family of algorithms, so refactoring the code as above may be completely out of scope. :)
> 
> Thanks again for your work on this!!

thanks for your comments. The InverseDistanceToAPower algorithm is already implemented with quadtree search capabilities (GDALGridInverseDistanceToAPowerNearestNeighbor).  Can be utilized with invdistnn.

There was no reason, for skipping the moving average algorithm. Just have overseen this algorithm. The adjustment is straightforward and should be implemented for the moving average algorithm in this PR too. 

Thanks for your suggestions regarding the refactorization.

bnordgren:
Thanks for updating MovingAverage as well. I tried it today and the speedup is _unreal_!  

I did have to set the bCreateQuadTree flag in order to turn on the quadtree option. Also, I changed all the static casts for the parameter objects on the "Metric" algorithms to match the rest of the code. It's not causing a problem now, but could cause a strange regression in the future if someone changes the parameters for Nearest Neighbor but not Metrics (or vice versa). Perhaps you can cherry-pick my change from this commit in my fork? c744522823808f2d079093104ace011e87a3a015

If not, it may be easier to just replicate the changes in your source tree.

bnordgren:
Nevermind about cherry picking, I seem to have figured out how to send a pull request to your fork.

FeU-aKlos:
> Nevermind about cherry picking, I seem to have figured out how to send a pull request to your fork.

great work! thanks for improving the PR!

FeU-aKlos:
> I was thinking that to test the case of nPoints <= 100 vs nPoints > 100, instead of hardcoding this 100 value, you could use a "const int nPointCountTreshold = atoi(CPLGetConfigOption("GDAL_GRID_POINT_COUNT_THRESHOLD", "100"));" , and then in autotest, you would just have to use a pattern like
> 
> ```
> @pytest.mark.parametrize("use_quadtree", [True, False])
> def test_mytest(use_quadtree):
>     with gdaltest.config_option('GDAL_GRID_POINT_COUNT_THRESHOLD', '0' if use_quadtree else '1000000000'):
>           ....
> ```

I implemented new test cases as you suggested. Do I have to do something else? Or implement more tests? What would be the next steps?

PR #5415, opened 2022-03-07
DIMAP: avoid warning when extracting metadata from unsupported band_id

tbonfort:
PNEO `Radiometric_Data.Radiometric_Calibration.Instrument_Calibration.Band_Measurement_List.*` has changed the nomenclature of `BAND_ID` :cry: and is now referenced as `R`,`G`,`B`,`NIR`,`RE`,`DB` instead of the previous `B0`,`B1`, etc...

This pull request silences the warning that is raised when `BAND_ID` does not match `B[0-9]+` . A cleaner solution mapping the band's alphabetic id with it's band number is desirable for a future release.

### DIFF
--- a/frmts/dimap/dimapdataset.cpp
+++ b/frmts/dimap/dimapdataset.cpp
@@ -1588,11 +1588,14 @@ int DIMAPDataset::ReadImageInformation2()
                         if( EQUAL(psTag->pszValue, "BAND_ID") )
                         {
                             // BAND_ID is: B0, B1, .... P
-                            if( !EQUAL(psTag->psChild->pszValue, "P") )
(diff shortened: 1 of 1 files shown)


tbonfort:
@rouault does this PR have a chance to make it into the 3.4.2 release you announced this morning or is too late already?

rouault:
> does this PR have a chance to make it into the 3.4.2 release you announced this morning or is too late already?

that would require a RC2 at that point.

rouault:
> If however an RC2 was needed for another reason, could we slip this one in ?

sure. This is going to be merged in the 3.4 branch in the follow up backport PR.

PR #5422, opened 2022-03-07
CMake: bring back capability of finding plugins in non ABI versioned /lib/gdalplugins directory

### LINKED ISSUE #5421: master/cmake: plugin path
After using a master build from today, my gdal_xxx.so plugin located in /usr/local/lib/gdalplugins is not loaded automatically anymore. If I move it to /usr/local/lib/gdalplugins/3.5 it loads correctly. From the docs at https://gdal.org/api/gdaldriver_cpp.html#_CPPv4N17GDALDriverManager15AutoLoadDriversEv I don't see anything documenting this change. Is this a known issue or the expected behavior ?

rouault:
Default installation path remains $prefix/lib/gdalplugins/$major.$minor (fixes #5421)


tbonfort:
Hi Even, I don't understand how this patch is supposed to be used. After having applied it plugins are still not registered when located in /usr/local/lib/gdalplugins. I'm fine with breaking backwards compatibility and requiring that future versions of gdal expect their plugins to be in $prefix/lib/gdalplugins/major.minor and can submit a PR to document this (I think this would require the next version to be a 4.0 however)

rouault:
> After having applied it plugins are still not registered when located in /usr/local/lib/gdalplugins.

you'll likely need to run "cmake .. -UINSTALL_PLUGIN_DIR_BASE -UINSTALL_PLUGIN_DIR" to clear existing variables.

tbonfort:
```
cmake .. -UINSTALL_PLUGIN_DIR_BASE -UINSTALL_PLUGIN_DIR
cmake .. -DGDAL_USE_JXL=ON -DGDAL_USE_KDU=OFF -DGDAL_USE_TIFF_INTERNAL=ON -DCMAKE_BUILD_TYPE=RelWithDebugInfo -DBUILD_PYTHON_BINDINGS=ON
make clean
make -j8
sudo make install
```

=> plugin in /usr/local/lib/gdalplugins/ is not loaded, plugin in /usr/local/lib/gdalplugins/3.5/ is loaded

rouault:
> plugin in /usr/local/lib/gdalplugins/ is not loaded, plugin in /usr/local/lib/gdalplugins/3.5/ is loaded

note that if there's a  /usr/local/lib/gdalplugins/3.5 directory, even empty, plugins will be looked for only in it.

tbonfort:
OK, that makes sense, but given the fact that the  /usr/local/lib/gdalplugins/3.5/ is created at install time, there's a backwards compatibility issue here that needs to show up in the docs and migration guide, and probably require a major version bump. I'll submit a doc PR for that

rouault:
> given the fact that the /usr/local/lib/gdalplugins/3.5/ is created at install time,

I'm looking at that. We shouldn't create it if there are no plugins

rouault:
> I'm looking at that. We shouldn't create it if there are no plugins

I've actually reverted 69d0cd488375da23b524fead8d87c1ee98e5c4fd, and in 2e03ac304f3bd47557128d3a0002b245f06b1c4b set INSTALL_PLUGIN_DIR to an unversioned directory. Distributions like Debian I believe that use versioned plugin directories can just set INSTALL_PLUGIN_DIR at cmake time.

tbonfort:
imho, I think the current behavior, where plugins in gdalplugins/ are ignored or not depending on whether gdalplugins/x.y is present or not, and gdalplugins/x.y is present or not depending on whether some drivers have been configured as plugins or not, is prone to much confusion. I would find it clearer to stop supporting plugins in gdalplugins/ and document clearly in the migration guide that they are expected to be found in gdalplugins/x.y

tbonfort:
> > I'm looking at that. We shouldn't create it if there are no plugins
> 
> I've actually reverted [69d0cd4](https://github.com/OSGeo/gdal/commit/69d0cd488375da23b524fead8d87c1ee98e5c4fd), and in [2e03ac3](https://github.com/OSGeo/gdal/commit/2e03ac304f3bd47557128d3a0002b245f06b1c4b) set INSTALL_PLUGIN_DIR to an unversioned directory. Distributions like Debian I believe that use versioned plugin directories can just set INSTALL_PLUGIN_DIR at cmake time.

:+1:  that's fine with me

PR #5426, opened 2022-03-08
Assorted set of fixes related to CMake Android cross compilation

rouault:


gdal.cmake
```diff
@@ -197,12 +197,15 @@ endif ()
...
 if (CMAKE_C_FLAGS)
-  string(REPLACE "-Werror" " " CMAKE_C_FLAGS ${CMAKE_C_FLAGS})
+  string(REGEX REPLACE "-Werror$" " " CMAKE_C_FLAGS ${CMAKE_C_FLAGS})
+  string(REGEX REPLACE "-Werror " " " CMAKE_C_FLAGS ${CMAKE_C_FLAGS})
   string(REPLACE "/WX" " " CMAKE_C_FLAGS ${CMAKE_C_FLAGS})
 endif ()
 if (CMAKE_CXX_FLAGS)
-  string(REPLACE "-Werror" " " CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS})
+  string(REGEX REPLACE "-Werror$" " " CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS})
+  string(REGEX REPLACE "-Werror " " " CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS})
```
dg0yt, line 208: I would prefer quotes on the input, and an extra space for a single simple `REPLACE`:
~~~cmake
string(REPLACE "-Werror " " " CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ")
~~~
(`CMAKE_<LANG>_FLAGS` is a string, not a CMake list.)

rouault: done

.github/workflows/android_cmake/start.sh
```diff
@@ -0,0 +1,97 @@
...
+
+# build sqlite3
+wget -q https://sqlite.org/2022/sqlite-autoconf-3370200.tar.gz
+tar xzf sqlite-autoconf-3370200.tar.gz
+cd sqlite-autoconf-3370200
+CC="ccache $NDK_TOOLCHAIN/bin/aarch64-linux-android24-clang" ./configure \
+  --prefix=/tmp/install --host=x86_64-linux-android24
+make -j3
+make install
+cd ..
```
akontsevich, line 43: Hi @rouault, I've successfully built gdal, proj and sqlite3 same way as You did here. However when I run android application with these libraries I get this error:
```
AndroidRuntime: java.lang.UnsatisfiedLinkError: dlopen failed: library "libsqlite3.so.0" not found: needed by /data/app/~~X1wOeY5-90cImKIsrkJNMQ==/com.myapp.mycompany-_udxwh7uKcyKcoksBRfj7g==/lib/arm64/libgdal.so in namespace classloader-namespace
```
while `libsqlite3.so` is deployed to the APK bundle at the same location as `libgdal.so`: `/lib/arm64-v8a`. Any ideas what is the problem and how to solve?

akontsevich: I know this, and this is the 2nd question: why it asks for `libsqlite3.so.0` and not for `libsqlite3.so`. May be some wrong or missed `cmake` build arguments? How to solve the situation?

dg0yt: I would start with checking the logs and installed files from the sqlite3 build. Does it properly use the Android naming conventions?

But first, everything is targeting arm64, but the sqlite build is configured for `--host=x86_64-linux-android24` 💥

akontsevich: no, in my case it is configured for `aarch64-linux-android24`.

akontsevich: Adding `LDFLAGS="-avoid-version"` to `./configure` [solves](https://stackoverflow.com/a/12067374/630169) the issue. No error now, however application crashes, not sure it is connected with GDAL build however.

.github/workflows/android_cmake/start.sh
```diff
@@ -0,0 +1,97 @@
...
+
+ccache -M 1G
+ccache -s
+
+# build sqlite3
+wget -q https://sqlite.org/2022/sqlite-autoconf-3370200.tar.gz
+tar xzf sqlite-autoconf-3370200.tar.gz
+cd sqlite-autoconf-3370200
+CC="ccache $NDK_TOOLCHAIN/bin/aarch64-linux-android24-clang" ./configure \
+  --prefix=/tmp/install --host=x86_64-linux-android24
```
akontsevich, line 40:
```suggestion
CC="ccache $NDK_TOOLCHAIN/bin/aarch64-linux-android24-clang" ./configure \
  LDFLAGS="-avoid-version" --prefix=/tmp/install --host=x86_64-linux-android24
```

rouault: maybe you just needed to deploy libsqlite3.so.0 ?

akontsevich: Not possible, Android build system does not allow this - only without version soname. `LDFLAGS="-avoid-version"` like above works - produces only 1 .so file.

akontsevich:
For compatibility reasons need to build PROJ and GDAL with NDKr21. PROJ build fine with the same commands as NDKr23 but GDAL always produces errors:
```
/android/ndk/21.3.6528147/toolchains/llvm/prebuilt/linux-x86_64/lib/gcc/aarch64-linux-android/4.9.x/../../../../aarch64-linux-android/bin/ld: warning: libproj.so, needed by ../libgdal.so, not found (try using -rpath or -rpath-link)                                                                                                                               
/android/ndk/21.3.6528147/toolchains/llvm/prebuilt/linux-x86_64/lib/gcc/aarch64-linux-android/4.9.x/../../../../aarch64-linux-android/bin/ld: warning: libsqlite3.so, needed by ../libgdal.so, not found (try using -rpath or -rpath-link)  
```
While NDK r23 builds fine but can't use with an app built with 21th one. Also with NDK 21 GDAL does not recognize sqlite build with rtree, so forced to use `ACCEPT_MISSING_SQLITE3_RTREE:BOOL=ON` flag.

@rouault any ideas why it fails with 21? Any fix or workaround available? Downgrade GDAL/PROJ versions (trying latest for now)?

akontsevich:
So we solved above errors adding `-rpath-link` flags to the `cmake` **CMAKE_CXX_FLAGS** variable when building GDAL:

    cmake -DCMAKE_BUILD_TYPE=Release \
          -DCMAKE_SYSTEM_NAME=Android \
          -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TOOLCHAIN_FILE} \
          -DANDROID_ABI=${ABI} \
          -DANDROID_NATIVE_API_LEVEL=${MIN_SDK_VERSION} \
          -DSQLITE3_INCLUDE_DIR=${SQLITE_INSTALL_DIR}/include \
          -DSQLITE3_LIBRARY=${SQLITE_INSTALL_DIR}/lib/libsqlite3.so \
          -DCMAKE_CXX_FLAGS="-Wl,-rpath-link=${PROJ_INSTALL_DIR}/lib \
                             -Wl,-rpath-link=${SQLITE_INSTALL_DIR}/lib \
                             -Wl,-rpath-link=${NDK_TOOLCHAIN}/sysroot/usr/lib/${HOST}/${MIN_SDK_VERSION}" \
          -DACCEPT_MISSING_SQLITE3_RTREE:BOOL=ON -DBUILD_TESTING=OFF \
          -DGDAL_USE_GEOTIFF_INTERNAL=ON -DGDAL_USE_TIFF_INTERNAL=ON -DGDAL_USE_ZLIB_INTERNAL=ON \
          -DPROJ_INCLUDE_DIR=${PROJ_INSTALL_DIR}/include -DPROJ_LIBRARY_RELEASE=${PROJ_INSTALL_DIR}/lib/libproj.so .. && \
    cmake --build . -j $(nproc) && \
    cmake --install . --prefix "${GDAL_INSTALL_DIR}"

So it is possible to build with NDK 21 as well! Thanks!

PR #5435, opened 2022-03-10
Updated install instructions for Python bindings

ChrisBarker-NOAA:
I didn't change much, but this was very out of date -- `easy_install` is not a thing anymore.

And I added a section on conda, because that is by far the easiest way to get GDAL on Windows and Mac systems, and maybe even LInux.


avalentino (approved):
LGTM

PR #5436, opened 2022-03-10
Make cpl_aws_win32.cpp compilation conditioned by atlbase.h availability

rouault:
refs https://github.com/OSGeo/gdal/pull/3022#issuecomment-1063681408


rouault:
CC @dg0yt

dg0yt:
Thanks. I will test this on top of the 3.4.2 release, in order to enable for arm64-windows and uwp builds (if no other blockers).

dg0yt:
This needs more work, but I can prepare a PR.

PR #5437, opened 2022-03-10
CMake: fix error related to ZSTD when building against Conda on Linux

rouault:


rouault:
Issue reported to conda-forge zstd-feedstock in https://github.com/conda-forge/zstd-feedstock/issues/58

PR #5440, opened 2022-03-12
Fix build without atlbase.h

dg0yt:
## What does this PR do?

Fixes undefined symbols when building without atlbase.h on windows.

Amends #5436.

---

Wish list:
For packaging, I would prefer to control "AWS EC2 host detection" as a feature, at least with CMake.
(This is possible with nmake now via `HAVE_ATLBASE_H`.)

### DIFF
--- a/port/cpl_aws.cpp
+++ b/port/cpl_aws.cpp
@@ -43,7 +43,7 @@ CPL_CVSID("$Id$")
 // #define DEBUG_VERBOSE 1
 
 #ifdef WIN32
-#if defined(_MSC_VER)
+#if defined(HAVE_ATLBASE_H)
 bool CPLFetchWindowsProductUUID(CPLString &osStr); // defined in cpl_aws_win32.cpp
 #endif
 const char* CPLGetWineVersion(); // defined in cpl_vsil_win32.cpp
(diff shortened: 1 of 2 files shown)


rouault:
> I would prefer to control "AWS EC2 host detection" as a feature, at least with CMake.

why would that be necessary if all the requirements to compile this file are met ?

dg0yt:
> > I would prefer to control "AWS EC2 host detection" as a feature, at least with CMake.
> 
> why would that be necessary if all the requirements to compile this file are met ?

For packaging, I want to actively control if a feature is present. I don't want features to be present which were not intented to be present (but enabled by autodetection), and I don't want features to be missing which were intented to be present (but disabled without error by autodetection). 
Otherwise users might rely on a behaviour which was not intented.

rouault:
> to actively control if a feature is present

Personally I'd consider this AWS EC2 host detection as a "micro-feature". It wouldn't really make sense to present this has an optional feature, whereas larger features can't be disabled. For example if Curl is detected,  currently we build all VSI network file systems (/vsicurl/, /vsis3/, /vsigs/, etc.). One could potentially imagine to have a flag to compile each separately, but then it becomes difficult to test that each potential combination builds. Before GDAL 2, it was theoretically possible to build GDAL without OGR support, but this got broken regularly because of cross references between the 2 parts, so it was removed.

dg0yt:
IIUC there is a fallback to a network-based approach, so there is no observable difference?

rouault:
> so there is no observable difference?

Yes, CPLFetchWindowsProductUUID() avoids doing a useless network request to know if we run on a EC2 instance in case other authentication methods to AWS S3 haven't succeeded so far (see https://gdal.org/user/virtual_file_systems.html#vsis3-aws-s3-files)

PR #5450, opened 2022-03-14
Fix and extend dependencies in GDALConfig.cmake

dg0yt:
## What does this PR do?

For static builds, GDALConfig.cmake must load all dependency packages needed to link to the imported GDAL library.
This PR fixes some simple errors, installs the custom find modules, and adds some individual improvements.

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments

## Environment

Triggered by, and tested with, PDAL in vcpkg, x64-linux.
Failed CI: https://dev.azure.com/vcpkg/public/_build/results?buildId=68759&view=results
New CI (just started): https://dev.azure.com/vcpkg/public/_build/results?buildId=68788&view=results


cmake/helpers/CheckDependentLibraries.cmake
```diff
@@ -567,7 +575,10 @@ define_find_package2(KEA libkea/KEACommon.h kea)
 gdal_check_package(KEA "Enable KEA driver" CAN_DISABLE)
 
 gdal_check_package(ECW "Enable ECW driver" CAN_DISABLE)
-gdal_check_package(NetCDF "Enable netCDF driver" CAN_DISABLE)
+gdal_check_package(NetCDF "Enable netCDF driver" CAN_DISABLE
+  NAMES netCDF
+  TARGETS netCDF::netcdf NETCDF::netCDF
```
rouault, line 580: this breaks netCDF NC4 detection (NETCDF_HAS_NC4). Perhaps the logic at https://github.com/OSGeo/gdal/blob/master/cmake/modules/packages/FindNetCDF.cmake#L118 should be moved here ?

dg0yt: Is NETCDF_HAS_NC4 needed in global scope? 
I would prefer to move it to `frmts/netcdf/CMakeLists.txt`.

dg0yt: Or just use variables from netCDFConfig.cmake:
~~~cmake
# Features:
#
set(netCDF_HAS_NC2 yes)
set(netCDF_HAS_NC4 yes)
set(netCDF_HAS_HDF4 no)
set(netCDF_HAS_HDF5 yes)
set(netCDF_HAS_PNETCDF no)
set(netCDF_HAS_PARALLEL no)
set(netCDF_HAS_DAP yes)
set(netCDF_HAS_DAP2 yes)
set(netCDF_HAS_DAP4 yes)
set(netCDF_HAS_DISKLESS yes)
set(netCDF_HAS_MMAP no)
set(netCDF_HAS_JNA no)
~~~

rouault: > I would prefer to move it to `frmts/netcdf/CMakeLists.txt`.

it should be fine there.

> Or just use variables from netCDFConfig.cmake:

yes we could harmonize the naming of our variable NETCDF_HAS_NC4 with the one of netCDFConfig.cmake.

rouault:
The "mingw32-make[2]: *** No rule to make target 'D:/a/_temp/msys64/mingw64/lib/libszip.dll.a', needed by 'libgdal.dll'.  Stop." error in the build-windows-msys2-mingw  seems to be specific of this PR (likely related to netCDF since szip is a dependency of HDF5).
The crash in tests for build-mac is unrelated as it happens in another PR (seems related to libgeotiff 1.7.1 upgrade in homebrew)

rouault:
> I can either disable netcdf cmake config for MSYS2, or patch the installed cmake config

ideally we'd want someone trying that build to do a plain "cmake .." without bumping into that issue which leave them puzzled. Is there a way we could detect during cmake time that the netcdf config is broken and just revert to using cmake/modules/packages/FindNetCDF.cmake ?

dg0yt:
> > I can either disable netcdf cmake config for MSYS2, or patch the installed cmake config
> 
> ideally we'd want someone trying that build to do a plain "cmake .." without bumping into that issue which leave them puzzled. Is there a way we could detect during cmake time that the netcdf config is broken and just revert to using cmake/modules/packages/FindNetCDF.cmake ?

I'm not sure what is the best approach. 
- The find module approach is unable to properly deal with transitive usage requirements. As usual, the issue is hidden when building with dynamic linkage. Implementing dependency detection in find modules doesn't scale well. The netCDF config avoids the problem.
- The msys2 netcdf config bug is outside gdal. I wouldn't want to add workarounds for every potential outside bug. (I remember the libgeotiff config issue.) This wouldn't scale well. However, there could be a document of known issues.

I would suggest/do the following:
- Keep the changes in `frmts/netcdf` (no problem, just enabling netcdf config compatibility).
- Move the netcdf gdal_check_package change to a separate draft PR. Add the reason for draft state there.
- Add an issue for the problem with using static builds using netcdf, and the workaround. (`-DGDAL_CHECK_PACKAGE_NetCDF_NAMES=netCDF -DGDAL_CHECK_PACKAGE_NetCDF_TARGETS=netCDF::netcdf`)

So user don't face a breaking change in the build system, and users stumbling over the linkage problem find a known issue and solutions.

rouault:
Looks good to me

dg0yt:
You can merge at this point.

PR #5453, opened 2022-03-15
CMake: Use netcdf config

### LINKED ISSUE #5452: CMake build cannot properly link to a static build of netcdf
(Split from https://github.com/OSGeo/gdal/pull/5450#issuecomment-1067605904)

When linking against a static build of netcdf, it is necessary to link the libraries which are needed by netcdf. For example, netcdf may need hdf5_hl. (GDAL only uses hdf5.) `FindNetCDF.cmake` in GDAL doesn't handle the usage requirements.

Build against static dependencies, e.g. from static triplets in vcpgk.

## Operating system

any

`master`

## Proposed solution

Detect and use export cmake config from netcdf

## Workaround

After merge of #5450: Enable cmake config import via command line parameters:
~~~
cmake [...] -DGDAL_CHECK_PACKAGE_NetCDF_NAMES=netCDF
~~~

dg0yt:
## What does this PR do?

This PR enables using netcdf's exported CMake config.
This allows to handle transitive usage requirements, and it saves expensive tests for feature detection.
Fixes #5452.

Depends on #5450.
~Blocked by https://github.com/msys2/MINGW-packages/issues/10993~

## Tasklist

 - [x] Wait for fix to msys2 netcdf package.
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/cmake/helpers/CheckDependentLibraries.cmake
+++ b/cmake/helpers/CheckDependentLibraries.cmake
@@ -582,7 +582,7 @@ gdal_check_package(KEA "Enable KEA driver" CAN_DISABLE)
 
 gdal_check_package(ECW "Enable ECW driver" CAN_DISABLE)
 gdal_check_package(NetCDF "Enable netCDF driver" CAN_DISABLE
-  # NAMES netCDF # Cf. https://github.com/OSGeo/gdal/pull/5453
+  NAMES netCDF
   TARGETS netCDF::netcdf NETCDF::netCDF)
 gdal_check_package(OGDI "Enable ogr_OGDI driver" CAN_DISABLE)
 # OpenCL warping gives different results than the ones expected by autotest, so disable it by default even if found.


rouault:
(not sure why there's now an issue with the Linux of Windows builds ? "CMake Error at gdal/gdal.cmake:264 (add_library):  Target "GDAL" links to target "NETCDF::netCDF" but the target was not found.  Perhaps a find_package() call is missing for an IMPORTED target, or an ALIAS target is missing?"

For dependencies with known issues in the CMake config file, it might be more prudent to *not* define the GDAL_CHECK_PACKAGE_xxxxxx_NAMES and GDAL_CHECK_PACKAGE_xxxx_TARGETS variables and let that responsibility to the person issuing the build (in your case your calling vcpkg scripts). Or if you can detect that this is a build using vcpkg dependencies, define them conditionally in gdal.cmake

Static builds are not the main use case of GDAL, so I'd not like people doing shared builds to hit issues (and thus cause extra support activity) they would normally hit otherwise.

rouault:
> Basically this means to not add `NAMES` arguments to the corresponding `gdal_check_package` call.
> 
> But we can allow the user to opt-in by passing the input variable `GDAL_CHECK_PACKAGE_xxxxxx_NAMES`.

yes, I agree with that.
The question here is what to do with netCDF. If https://github.com/msys2/MINGW-packages/issues/10993 is effectively now solved, then perhaps try the approach of your PR.

rouault:
gdal_target_link_libraries(gdal_netCDF PRIVATE NETCDF::netCDF) in frmts/netcdf/CMakeLists.txt will have to be changed to gdal_target_link_libraries(gdal_netCDF PRIVATE ${netCDF_TARGET}) once https://github.com/OSGeo/gdal/pull/5455 is merged

rouault:
> gdal_target_link_libraries(gdal_netCDF PRIVATE NETCDF::netCDF) in frmts/netcdf/CMakeLists.txt will have to be changed to gdal_target_link_libraries(gdal_netCDF PRIVATE ${netCDF_TARGET}) once #5455 is merged

actually I've had to make that change in #5455 since master was now using NetCDF_LIBRARIES. sorry for the noise and confusion

rouault:
It seems that homebrew has also a broken netCDF config file...
```
make[2]: *** No rule to make target `/usr/local/lib/libsz.dylib', needed by `libgdal.30.0.0.dylib'.  Stop.
```

dg0yt:
> It seems that homebrew has also a broken netCDF config file...
> 
> ```
> make[2]: *** No rule to make target `/usr/local/lib/libsz.dylib', needed by `libgdal.30.0.0.dylib'.  Stop.
> ```

Search for szip in the "Install dependencies" step:
~~~
Warning: szip has been deprecated because it has an incompatible license. Check out `libaec` instead!
...
==> Pouring szip--2.1.1_1.big_sur.bottle.tar.gz
==> Caveats
szip is keg-only, which means it was not symlinked into /usr/local,
because szip conflicts with libaec, which provides an open-source replacement.

For compilers to find szip you may need to set:
  export LDFLAGS="-L/usr/local/opt/szip/lib"
  export CPPFLAGS="-I/usr/local/opt/szip/include"
~~~

dg0yt:
... and postgis still pulls in szip. Not sure if szip and libaec can coexist at linktime or runtime.

rouault:
> postgis still pulls in szip.

postgis ?  or maybe postgis because it pulls gdal which pulls ...
But ultimately szip is a dependecy of HDF5 as shown by:
```
Found HDF5: /usr/local/Cellar/hdf5/1.13.0/lib/libhdf5.dylib;/usr/local/opt/szip/lib/libsz.dylib;/Applications/Xcode_12.4.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.1.sdk/usr/lib/libz.tbd;/Applications/Xcode_12.4.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.1.sdk/usr/lib/libdl.tbd;/Applications/Xcode_12.4.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.1.sdk/usr/lib/libm.tbd;/usr/local/Cellar/hdf5/1.13.0/lib/libhdf5_cpp.dylib;/usr/local/Cellar/hdf5/1.13.0/lib/libhdf5.dylib;/usr/local/opt/szip/lib/libsz.dylib;/Applications/Xcode_12.4.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.1.sdk/usr/lib/libz.tbd;/Applications/Xcode_12.4.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.1.sdk/usr/lib/libdl.tbd;/Applications/Xcode_12.4.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.1.sdk/usr/lib/libm.tbd 
```
However the build now succeeds...

As it seems we've hit a few issues with the netCDF config file , wouldn't it be more prudent to have "NAME netCDF" only set when BUILD_SHARED_LIBS=OFF ? The FindNetCDF.cmake file is probably good enough to handle the shared libs case and  users could still manually set GDAL_CHECK_PACKAGE_NetCDF_NAMES manually if really needed.

dg0yt:
> As it seems we've hit a few issues with the netCDF config file , wouldn't it be more prudent to have "NAME netCDF" only set when BUILD_SHARED_LIBS=OFF ? The FindNetCDF.cmake file is probably good enough to handle the shared libs case and users could still manually set GDAL_CHECK_PACKAGE_NetCDF_NAMES manually if really needed.

I would prefer to not add more complexity to the build system. We can hold back the netcdf `NAMES` if you think it is too early. But we can already align the target names. (I would create a new PR, and reduce this PR to the `NAMES` change only, or close it.)

If you want some global control, we could add a GDAL_CHECK_PACKAGE_PREFER_CONFIG (default off) option to control search order, similar to CMAKE_FIND_PACKAGE_PREFER_CONFIG.

dg0yt:
> reopen this pull request.

No such button...

PR #5455, opened 2022-03-15
Fix CMake Mac build issue with ZSTD from homebrew

rouault:


rouault:
CC @dg0yt  for the second commit: I have hit a compilation error on Mac homebrew in the MRF driver related to ZSTD. The issue was with "gdal_target_link_libraries(gdal_MRF PRIVATE ${ZSTD_LIBRARIES})" where ZSTD_LIBRARIES pointed only to the library name, and thus the magic in gdal_target_link_libraries() that finds the interface include directories from the target name didn't work, since ZSTD_LIBRARIES was not a target name.

rouault:
(First commit fixing CMake Mac homebrew crashes cherry-picked to master)

dg0yt:
> CC @dg0yt for the second commit: I have hit a compilation error on Mac homebrew in the MRF driver related to ZSTD. The issue was with "gdal_target_link_libraries(gdal_MRF PRIVATE ${ZSTD_LIBRARIES})" where ZSTD_LIBRARIES pointed only to the library name, and thus the magic in gdal_target_link_libraries() that finds the interface include directories from the target name didn't work, since ZSTD_LIBRARIES was not a target name.

This sounds like the libgeotiff pattern...

.github/workflows/cmake_builds.yml
```diff
@@ -443,15 +443,17 @@ jobs:
...
         cmake --version
     - name: Install dependency
+      # libgeotiff 1.7.1 at time of writing links to proj@7, and not proj (9.0.0)
+      # so in order to avoid issue at runtime install proj@7 and not proj
       run: |
         brew update
         brew outdated
         brew install postgresql || brew link postgresql
-        brew install freexl libxml2 libspatialite geos proj openjpeg giflib szip postgis poppler doxygen unixodbc
+        brew install freexl libxml2 libspatialite geos proj@7 libgeotiff openjpeg giflib szip postgis poppler doxygen unixodbc
```
dg0yt, line 452: Recently was a related on discussion on one of the mailinglists.

cmake/helpers/CheckDependentLibraries.cmake
```diff
@@ -59,6 +59,7 @@ function(gdal_check_package_target name)
     if(TARGET ${target})
       gdal_check_target_is_valid(${target} _is_valid)
       if (_is_valid)
+        set(${name}_TARGET "${target}" PARENT_SCOPE)
         set(${name}_INCLUDE_DIRS "" PARENT_SCOPE)
         set(${name}_LIBRARIES "${target}" PARENT_SCOPE)
```
dg0yt, line 64: Basically, `${name}_TARGET` would now take over what `${name}_LIBRARIES` was trying to achieve.
(Would `${name}_LIBRARIES` still be used?)

Note that `${name}_LIBRARIES` is also mentioned in the documentation a few lines later.

dg0yt (commented):
The key problem before this PR is that `GDAL_CHECK_PACKAGE_${name}_NAMES` and `GDAL_CHECK_PACKAGE_${name}_TARGETS` can only be used if the find module's target name is not hard coded into the link commands.

cmake/helpers/CheckDependentLibraries.cmake
rouault: > The key problem before this PR is that GDAL_CHECK_PACKAGE_${name}_NAMES and GDAL_CHECK_PACKAGE_${name}_TARGETS can only be used if the find module's target name is not hard coded into the link commands.

not sure to understand what you mean and if that has implications on this PR

> Basically, `${name}_TARGET` would now take over what `${name}_LIBRARIES` was trying to achieve.

yes, but making it more explicit that we intend to use a target, to get from it both include_dirs and libs.

> (Would ${name}_LIBRARIES still be used?)

no, I should remove that and the mention in the doc you pointed.

rouault: PR amended to remove setting ${name}_LIBRARIES and ${name}_INCLUDE

dg0yt: > > The key problem before this PR is that GDAL_CHECK_PACKAGE_${name}_NAMES and GDAL_CHECK_PACKAGE_${name}_TARGETS can only be used if the find module's target name is not hard coded into the link commands.
> 
> not sure to understand what you mean and if that has implications on this PR

This PR only touches the locations where there already was `<Pkg>_LIBRARIES`. To fully enable user config opt-in via the variables, `<Pkg>_TARGET` would need to be used for every `<Pkg>`. But I think it is okay to add this incrementally.

(I considered a `<Pkg>_TARGET` variable as part of my PR which introduced config support, but it would have made that PR more complex. And I don't want to touch the Find modules too much.)

rouault: > To fully enable user config opt-in via the variables, <Pkg>_TARGET would need to be used for every <Pkg>. But I think it is okay to add this incrementally.

I doubt we have a better choice. The different drivers need to refer to a target name to be able to get both the include dirs and libraries. If we make the target names configurable with GDAL_CHECK_PACKAGE_xxx_TARGETS then we need to change the hard-coded target name (that comes from the FindXXXX.cmake files) to the <Pkg>_TARGET approach I'm taking.

Issue #5399, opened 2022-03-02
gdalwarp offset on reprojection from epsg:28992 to epsg:25831


freekvw:
## Expected behavior and actual behavior.
Reprojecting some tiles of the Dutch AHN3 elevation model results in offsets in the output of >100 meters.

## Steps to reproduce the problem.

* Download & unpack e.g. tile https://download.pdok.nl/rws/ahn3/v1_0/5m_dsm/R5_46BN2.ZIP
* `gdalwarp -t_srs EPSG:25831 R5_46BN2.TIF R5_46BN2-25831.TIF`

Now open both files in e.g. QGIS and observe the offset of approximately 30 m east and 110 m north.

## Operating system

Reproducible on various Linux variants, including Ubuntu 20.04.

## GDAL version and provenance

Various, including a fresh build of master using proj 8.2.0.

## Further details
Specifying `-ct` to gdalwarp makes the issue go away. E.g.:
```
$ gdalwarp -t_srs EPSG:25831 -ct '+proj=pipeline +step +inv +proj=sterea +lat_0=52.1561605555556 +lon_0=5.38763888888889         +k=0.9999079 +x_0=155000 +y_0=463000 +ellps=bessel   +step +proj=push +v_3 +step +proj=cart +ellps=bessel   +step +proj=helmert +x=565.7381 +y=50.4018 +z=465.2904 +rx=0.395025981036064 +ry=-0.330772431242031 +rz=1.87607329462821 +s=4.07244  +convention=coordinate_frame +step +inv +proj=cart +ellps=GRS80 +step +proj=pop +v_3 +step +proj=utm +zone=31 +ellps=GRS80' R5_46BN2.TIF R5_46BN2-25831-2.TIF
$ gdalinfo R5_46BN2-25831.TIF | grep Origin
Origin = (704565.326535094529390,5744584.640074593946338)
$ gdalinfo R5_46BN2-25831-2.TIF | grep Origin
Origin = (704533.696954668965191,5744477.634767718613148)
```

jratike80:
The QGIS test means just that when is does on-the-fly re-projection it does it in another way than what  gdalwarp was using. Gdalwarp probably uses the best method that Proj finds and is using gridshift file

```
projinfo -s epsg:28992 -t epsg:25831
Candidate operations found: 2
-------------------------------------
Operation No. 1:

[...]

PR #5456, opened 2022-03-15
GTIFF SRS reading: avoid warnings due to previous fix for #5399 and other fixes

rouault:
Avoid erroneous new warnings
"The definition of geographic CRS EPSG:%d got from GeoTIFF keys "
"is not the same as the one from the EPSG registry, "
"which may cause issues during reprojection operations. "

- do not massage datum name to WKT1 (avoid warnings with datum names
  that have spaces)
- do not try to compare when the datum is dynamic
- do not try to compare when the code is of a geographic 3D CRS.

Other issues fixed during investigation:
- Fix handling of prime meridian with units != degree
- Internal libgeotiff: fix import of dynamic datum.


rouault:
manually backported to 3.4 per a22aa06a279824047f079b85643b4499600551c2

PR #5459, opened 2022-03-17
Fix exported CMake config for static linkage

dg0yt:
## What does this PR do?

This PR adds the missing pieces for making the exported CMake config work for a static build of GDAL.
The corresponding postinstall test is no longer expected to fail.
(However, there might still be feature configurations which don't work without additional changes.

In particular, the PR adds the export of dependencies based on `find_package2`.

The PR also fixes the link language for CryptoPP and SQLite.

The find modules now provide provide enough information even for C-only projects and for CXX-only projects. 
However, without CXX enabled (read: without the CXX linker), linking to static GDAL on Linux fails due to unresoved C++ standard library symbols. AFAIU the only effective fix is enabling CXX in the user project. This is now also done in the postinstall test project, and a warning is added to the exported config.

## Tasklist

 - [ ] Review
 - [x] Adjust for comments


cmake/modules/3.16/FindPostgreSQL.cmake
```diff
@@ -10,6 +10,8 @@ Find the PostgreSQL installation.
 IMPORTED Targets
 ^^^^^^^^^^^^^^^^
 
+.. versionadded:: 3.14
```
rouault, line 13: Ah good to see that the newest upstream version of FindPostgreSQL.cmake handles this cleanly, but looking at the history, the addition of the server component dates to commit https://github.com/Kitware/CMake/commit/8b066f1a65466ae2dc542b097f877795f386f942 which went to 3.20.  So this file should be move to a 3.20 subdirectory with the appropriate changes to deal with that version.

dg0yt: Good catch. I will deal with this tonight.

rouault:
Unrelated CI failure in ubuntu_18_04_32bit / addressed per https://github.com/OSGeo/gdal/pull/5466

PR #5460, opened 2022-03-17
CI: Cancel PR and branch workflows on new push

dg0yt:
## What does this PR do?

This change to the workflows cancel CI runs for a branch or PR when a new push is made to the same target. Typically, this happens when a contributor notices a defect after one push, locally or by a fast CI jobs (e.g. CMake Lint) and pushes a fix while the long-running workflows are still in progress.

This a common behaviour in other CIs, but needs extra activation in GH Actions. It frees runners from obsolete workload, making them available for running new jobs.

Upsides: This also affects the additional workflows in forks.

Downsides: A notification is sent for each canceled workflow. GDAL has many of them...

References: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency

Disclaimer: I'm not an expert on GH Actions.

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments

Unbenannt

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -8,6 +8,10 @@ on:
         paths-ignore:
             - 'doc/**'
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
+  cancel-in-progress: true
+
 jobs:
 
   android_build:
--- a/.github/workflows/android_cmake.yml
+++ b/.github/workflows/android_cmake.yml
@@ -8,6 +8,10 @@ on:
         paths-ignore:
             - 'doc/**'
 
+concurrency:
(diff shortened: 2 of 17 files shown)


rouault:
> A notification is sent for each canceled workflow.

what is the form of the notification: email (I'd hope not), ... ?

dg0yt:
> email

👿 Of course. That's why I mention it explicitly. You probably use e-mail filter rules already...

rouault:
looking a bit, I see that in https://github.com/settings/notifications for my account, it seems that notifications for actions are disabled (in the "Actions" tab), so I'd expect that to apply to to cancel jobs:

image

Perhaps in your account, this is enabled ?

dg0yt:
You are right, I do have "Email" enabled.
I also have "Send notifications for failed workflows only". So canceled = not successful = failed => notification.

PR #5501, opened 2022-03-24
CMake: use CMAKE_INSTALL_LIBDIR in INSTALL_PLUGIN_DIR

rouault:


gdal.cmake
```diff
@@ -353,7 +353,7 @@ if (GDAL_ENABLE_MACOSX_FRAMEWORK)
 else ()
   include(GNUInstallDirs)
   set(INSTALL_PLUGIN_DIR
-      "lib/gdalplugins"
+      "lib${LIB_SUFFIX}/gdalplugins"
```
dg0yt, line 356: With `GNUInstallDirs` ahead, shouldn't this be:
~~~cmake
   "${CMAKE_INSTALL_LIBDIR}/gdal_plugins"
~~~

rouault: @manisandro opinion w.r.t above suggestion of using ${CMAKE_INSTALL_LIBDIR} ?

manisandro: Works for me!

rouault: PR amended

PR #5503, opened 2022-03-24
GDAL GRASS driver: add rpath to Makefile.in

metzm:
## What does this PR do?
Add `-Wl,-rpath,<path>` to Makefile.in of the GDAL GRASS driver. This is supposed to be a generic solution replacing the [Debian patch](https://salsa.debian.org/debian-gis-team/gdal-grass/-/commit/0192b0418e04d79678bdd5c4c377b8cece5e0939).

The reason for this PR, as well as the Debian patch, is that GRASS shared libraries are in a non-standard location that is usually not in the common list of directories with shared libraries, also not in `/etc/ld.so.conf.d/*` and not in `LD_LIBRARY_PATH`.

### DIFF
--- a/frmts/grass/pkg/Makefile.in
+++ b/frmts/grass/pkg/Makefile.in
@@ -34,10 +34,10 @@ distclean: clean
 
 
 $(GLIBNAME):	grass.o
-	$(LD_SHARED) $(LDFLAGS) grass.o $(LIBS) -o $(GLIBNAME)
+	$(LD_SHARED) $(LDFLAGS) grass.o $(LIBS) -o $(GLIBNAME) -Wl,-rpath,"@GRASS_GISBASE@/lib"
 
 $(OLIBNAME):	ogrgrassdriver.o ogrgrassdatasource.o ogrgrasslayer.o
-	$(LD_SHARED) $(LDFLAGS) ogrgrassdriver.o ogrgrassdatasource.o ogrgrasslayer.o $(LIBS) -o $(OLIBNAME)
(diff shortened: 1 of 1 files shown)


rouault:
CC @sebastic

metzm:
I see. In this case this PR, just like the Debian patch, does not help.

A solution, as long as GRASS libraries are in non-standard directories, would be to create something like `/etc/ld.so.conf.d/grass.conf` if possible, but this is beyond my Makefile skills. Simply adding
```
echo "@GRASS_GISBASE@/lib" > /etc/ld.so.conf.d/grass.conf
```
to the `install:` section is probably not save.

sebastic:
`/usr/bin/grass` sets `LD_LIBRARY_PATH` to find the libraries in non-standard paths, if the GRASS build sets `RUNPATH` on the binaries and libraries the `LD_LIBRARY_PATH` work around won't be needed any more either.

metzm:
With [GRASS PR 2280](https://github.com/OSGeo/grass/pull/2280) and this PR, the GDAL-GRASS plugin works without any further modifications to `LD_LIBRARY_PATH` or `/etc/ld.so.conf.d/`.

neteler:
@rouault @sebastic can this PR be merged?

sebastic:
It's more useful now that GRASS sets RUNPATH.

neteler:
@rouault Can this please be added to the upcoming GDAL 3.4.3? Thanks

rouault:
ok I'm doing it, but it should be the responsibility of contributors to ensure they have backported their changes ahead of release time (tagging their pull requests with the "backport release/XXXX" label)

neteler:
Thanks and sorry about not being aware of this procedure so far.

PR #5505, opened 2022-03-25
SENTINEL2: identify zipped S2 datasets from file content

marisn:
Previous code was identifying zipped S2 datasets based only on their file names. Thus renaming a S2 dataset containing zip file would make it invisible to glda. This commit introduces extra check for presence of format specific metadata files inside a zip to determine if it is a S2 containing zip file. Primary detection is still done by file name thus speeding up the most common case when file name indicates on its content.

I was not able to run black on Python code and clang-format on C++ code as existing code base has not been reformatted.

### DIFF
--- a/autotest/gdrivers/sentinel2.py
+++ b/autotest/gdrivers/sentinel2.py
@@ -2417,57 +2417,52 @@ def test_sentinel2_l1c_safe_compact_3():
 
 def test_sentinel2_zipped():
     # S2 L1C
-    if not sys.platform.startswith("win"):
-        zipname = str(uuid.uuid4()) + ".zip"
-        with tempfile.TemporaryDirectory() as tmpdir:
-            zipwpath = os.path.join(tmpdir, zipname)
-            _zip_a_dir(zipwpath, "data/sentinel2/fake_l1c/S2A_OPER_PRD_MSIL1C.SAFE/")
-            assert os.path.exists(zipwpath)
-            ds = gdal.Open(zipwpath)
-            assert ds is not None
+    zipname = str(uuid.uuid4()) + ".zip"
+    with tempfile.TemporaryDirectory() as tmpdir:
+        zipwpath = os.path.join(tmpdir, zipname)
+        _zip_a_dir(zipwpath, "data/sentinel2/fake_l1c/S2A_OPER_PRD_MSIL1C.SAFE/")
+        assert os.path.exists(zipwpath)
+        ds = gdal.Open(zipwpath)
+        assert ds is not None
 
    # S2 L1B
-    if not sys.platform.startswith("win"):
-        zipname = str(uuid.uuid4()) + ".zip"
-        with tempfile.TemporaryDirectory() as tmpdir:
-            zipwpath = os.path.join(tmpdir, zipname)
-            _zip_a_dir(zipwpath, "data/sentinel2/fake_l1b/S2B_OPER_PRD_MSIL1B.SAFE/")
-            assert os.path.exists(zipwpath)
-            ds = gdal.Open(zipwpath)
-            assert ds is not None
+    zipname = str(uuid.uuid4()) + ".zip"
+    with tempfile.TemporaryDirectory() as tmpdir:
+        zipwpath = os.path.join(tmpdir, zipname)
+        _zip_a_dir(zipwpath, "data/sentinel2/fake_l1b/S2B_OPER_PRD_MSIL1B.SAFE/")
+        assert os.path.exists(zipwpath)
+        ds = gdal.Open(zipwpath)
+        assert ds is not None
 
     # S2 L1A
-    if not sys.platform.startswith("win"):
-        zipname = str(uuid.uuid4()) + ".zip"
-        with tempfile.TemporaryDirectory() as tmpdir:
-            zipwpath = os.path.join(tmpdir, zipname)
-            _zip_a_dir(zipwpath, "data/sentinel2/fake_l2a/S2A_USER_PRD_MSIL2A.SAFE/")
-            assert os.path.exists(zipwpath)
-            ds = gdal.Open(zipwpath)
-            assert ds is not None
+    zipname = str(uuid.uuid4()) + ".zip"
+    with tempfile.TemporaryDirectory() as tmpdir:
+        zipwpath = os.path.join(tmpdir, zipname)
+        _zip_a_dir(zipwpath, "data/sentinel2/fake_l2a/S2A_USER_PRD_MSIL2A.SAFE/")
+        assert os.path.exists(zipwpath)
+        ds = gdal.Open(zipwpath)
+        assert ds is not None
 
     # S2 L2A
-    if not sys.platform.startswith("win"):
-        zipname = str(uuid.uuid4()) + ".zip"
-        with tempfile.TemporaryDirectory() as tmpdir:
-            zipwpath = os.path.join(tmpdir, zipname)
-            _zip_a_dir(
-                zipwpath,
-                "data/sentinel2/fake_l2a_MSIL2A/S2A_MSIL2A_20180818T094031_N0208_R036_T34VFJ_20180818T120345.SAFE/",
-            )
-            assert os.path.exists(zipwpath)
-            ds = gdal.Open(zipwpath)
-            assert ds is not None
+    zipname = str(uuid.uuid4()) + ".zip"
+    with tempfile.TemporaryDirectory() as tmpdir:
+        zipwpath = os.path.join(tmpdir, zipname)
+        _zip_a_dir(
+            zipwpath,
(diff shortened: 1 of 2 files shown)


autotest/gdrivers/sentinel2.py
```diff
@@ -2413,4 +2412,62 @@ def test_sentinel2_l1c_safe_compact_3():
     assert band.DataType == gdal.GDT_Byte
 
 
+###############################################################################
+# Test opening zipped versions of S2 datasets
+
+def test_sentinel2_zipped():
+    # S2 L1C
+    if not sys.platform.startswith("win"):
```
rouault, line 2420: I don't think this if platform != win check is necessary with your new code. It was necessary for the previous code that used a "sh" invokation

frmts/sentinel2/sentinel2dataset.cpp
```diff
@@ -438,6 +440,33 @@ int SENTINEL2Dataset::Identify( GDALOpenInfo *poOpenInfo )
         strstr(pszHeader, "User_Product_Level-2A" ) != nullptr )
         return TRUE;
 
+    /* Last effort – look inside of a zip file for metadata file with
+     * matching name. */
+    if ( EQUAL(CPLGetExtension(pszJustFilename), "zip") )
```
rouault, line 445: I'm not super enthusiastic about having to list the content of arbitrary zip files in a Identify() method. Wouldn't it be an option for your use case that the user prefixes the zip filename with a "SENTINEL2:" prefix ? (you'd need to make change accordingly to detect that prefix and strip it out in Open())

frmts/sentinel2/sentinel2dataset.cpp
```diff
@@ -438,6 +440,33 @@ int SENTINEL2Dataset::Identify( GDALOpenInfo *poOpenInfo )
...
+    {
+        CPLString osFilename(poOpenInfo->pszFilename);
+        if( strncmp(osFilename, "/vsizip/", strlen("/vsizip/")) != 0 )
+            osFilename = "/vsizip/" + osFilename;
+
+        auto psDir = VSIOpenDir(osFilename.c_str(), 1, nullptr);
```
rouault, line 451: psDir must be freed with VSICloseDir(). Same in the Open() method

frmts/sentinel2/sentinel2dataset.cpp
```diff
@@ -438,6 +440,33 @@ int SENTINEL2Dataset::Identify( GDALOpenInfo *poOpenInfo )
...
+            const char* pszInsideFilename = CPLGetFilename(psEntry->pszName);
+            if ( VSI_ISREG(psEntry->nMode) && (
+                STARTS_WITH_CI(pszInsideFilename, "MTD_MSIL1C") ||
+                STARTS_WITH_CI(pszInsideFilename, "MTD_MSIL2A") ||
+                STARTS_WITH_CI(pszInsideFilename, "S2A_OPER_MTD_SAFL1B") ||
+                STARTS_WITH_CI(pszInsideFilename, "S2B_OPER_MTD_SAFL1B") ||
+                STARTS_WITH_CI(pszInsideFilename, "S2A_OPER_MTD_SAFL1C") ||
+                STARTS_WITH_CI(pszInsideFilename, "S2B_OPER_MTD_SAFL1C") ||
+                STARTS_WITH_CI(pszInsideFilename, "S2A_USER_MTD_SAFL2A") ||
+                STARTS_WITH_CI(pszInsideFilename, "S2B_USER_MTD_SAFL2A")
```
rouault, line 463: To avoid copy&paste, that list of filename prefixes should be moved to a dedicated function, that could be used by Open() as well

autotest/gdrivers/sentinel2.py
marisn: Good point. Thanks.

frmts/sentinel2/sentinel2dataset.cpp
marisn: Thanks!

frmts/sentinel2/sentinel2dataset.cpp
marisn: Done. Is it good enough?

frmts/sentinel2/sentinel2dataset.cpp
marisn: I can see your point but that would not work with drag&drop approach taken by QGIS.

frmts/sentinel2/sentinel2dataset.cpp
```diff
@@ -143,6 +143,7 @@ static const SENTINEL2_L2A_BandDescription asL2ABandDesc[] =
 
 #define NB_L2A_BANDS (sizeof(asL2ABandDesc)/sizeof(asL2ABandDesc[0]))
 
+bool SENTINEL2isMetadataFile(const char* pszInsideFilename);
```
rouault, line 146:
```suggestion
static bool SENTINEL2isMetadataFile(const char* pszInsideFilename);
```

frmts/sentinel2/sentinel2dataset.cpp
```diff
@@ -594,9 +618,51 @@ GDALDataset *SENTINEL2Dataset::Open( GDALOpenInfo * poOpenInfo )
...
+    }
+
     return nullptr;
 }
 
+/************************************************************************/
+/*                        SENTINEL2isMetadataFile()                     */
+/************************************************************************/
+
+bool SENTINEL2isMetadataFile(const char* pszInsideFilename)
```
rouault, line 652:
```suggestion
static bool SENTINEL2isMetadataFile(const char* pszInsideFilename)
```

frmts/sentinel2/sentinel2dataset.cpp
rouault: I'm thinking to an alternative that should work and avoid doing VSIOpenDir() for unrelated .zip
If you open a .zip file with Sentinel2 content in a binary editor, you can see in the first bytes of the file, the name of the first file inside the zip. So it should be possible to look into the buffer poOpenInfo->pabyHeader of size = poOpenInfo->nHeaderBytes if it contains one of the substrings tested by SENTINEL2isMetadataFile(). If so, then use the below approach to confirm. Otherwise just give up.

frmts/sentinel2/sentinel2dataset.cpp
```diff
@@ -438,6 +441,27 @@ int SENTINEL2Dataset::Identify( GDALOpenInfo *poOpenInfo )
...
+    /* Last effort – look inside of a zip file for metadata file with
+     * matching name. */
+    if ( EQUAL(CPLGetExtension(pszJustFilename), "zip") )
+    {
+        CPLString osFilename(poOpenInfo->pszFilename);
+        if( strncmp(osFilename, "/vsizip/", strlen("/vsizip/")) != 0 )
+            osFilename = "/vsizip/" + osFilename;
+
+        auto psDir = VSIOpenDir(osFilename.c_str(), 1, nullptr);
+        while ( const VSIDIREntry* psEntry = VSIGetNextDirEntry(psDir) )
```
rouault, line 453: This causes a crash in autotest/gdrivers/jpeg.py::test_jpeg_27_max_memory  , that tries to open a '/vsisubfile/146,/vsizip/../gcore/data/eofloop_valid_huff.tif.zip' file .
The issue is that putting a '/vsizip/' prefix in front of that filename results in something that can't be opened by VSIOpenDir().  psDir should thus be tested against nullptr before being used. Applies here and at line 629 as well.

frmts/sentinel2/sentinel2dataset.cpp
marisn: You are right. I implemented a new version that uses only header for identification and unzipped version only when opening.

frmts/sentinel2/sentinel2dataset.cpp
marisn: Thanks.

frmts/sentinel2/sentinel2dataset.cpp
marisn: Unfortunately on Debian Sid I'm unable to run full test suite :-(
I reshaped my code according to your suggestions – should be better now.

frmts/sentinel2/sentinel2dataset.cpp
```diff
@@ -594,9 +600,71 @@ GDALDataset *SENTINEL2Dataset::Open( GDALOpenInfo * poOpenInfo )
...
+    }
+
     return nullptr;
 }
 
+/************************************************************************/
+/*                        SENTINEL2isZipped()                           */
+/************************************************************************/
+
+static bool SENTINEL2isZipped(const char* pszHeader)
```
rouault, line 645: the value of poOpenInfo->nHeaderBytes should be passed as an extra argument ot this fi,ctop,, so it can check that its value is >= 34 + 15  (in practice the buffer should be at least 1024 bytes large, but it is dangerous to rely on this implementation detail)

Issue #5504, opened 2022-03-25
Feature Request: Add support for named NEATLINE extraction from Adobe ISO32000 style Geospatial PDF 


pascalscheffers:
The pdf driver already supports the extraction of named NEATLINE in pdfs following the best practices as outlined in 08-139r3_GeoPDF_Encoding_Best_Practice_Version_2.2.pdf (that's the comment in pdfdataset.cpp, line 4781).

I've implemented the same functionality for Adobe ISO32000 style Geospatial PDF and will create a pull request to integrate this shortly. This format output by Esri ArcMap. 

I have used the same NEATLINE command line option to specify the name.
I have added two CPLDebug lines to debug the new behaviour. One outputs the names of the viewports as they occur in the dictionary. The second one outputs the index of the viewport name if it was selected.
If no NEATLINE was requested or there is no match, the largest viewport is selected as before.

rouault:
fixed per https://github.com/OSGeo/gdal/pull/5507

PR #5507, opened 2022-03-25
PDF Driver: Named NEATLINE extraction from ISO32000 style Geospatial …

pascalscheffers:
…PDF (implements #5504)

## What does this PR do?
This PR implements feature request #5504. Nothing more or less. ISO32000 style GeoPDFs are currently handled such that the viewport which covers the largest area is automatically selected. This typically would be the main map of a pdf with a main map and one or more insets. Other use cases we've seen have multiple unrelated maps on one pdf page. Provided the areas are named, this PR allows extraction the named region using gdal_translate, getting information from it with gdalinfo, etc.

It has been tested extensively with PDFs published by the US National Geospatial-Inteligence Agency on Ubuntu 21.10 and MacOS 12 (via a custom brew.sh formula)

I have a backport for 3.4 available if desired.

Test PDFs in this format are available from the FAA at, for example, https://www.faa.gov/air_traffic/flight_info/aeronav/digital_products/ifr/#lowHighAlaska

Chart name 'ELAK1' (Alaska) has a main map and multiple insets. Other examples are in ELUS23 (Miami) and ELUS34

None.

## Tasklist

 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:

### DIFF
--- a/frmts/pdf/pdfdataset.cpp
+++ b/frmts/pdf/pdfdataset.cpp
@@ -6256,7 +6256,11 @@ int PDFDataset::ParseVP(GDALPDFObject* poVP, double dfMediaBoxWidth, double dfMe
 /* -------------------------------------------------------------------- */
 /*      Find the largest BBox                                           */
 /* -------------------------------------------------------------------- */
+    const char* pszNeatlineToSelect =
+        GetOption(papszOpenOptions, "NEATLINE", "Map Layers");
+
     int iLargest = 0;
+    int iRequestedVP = -1;
     double dfLargestArea = 0;
 
     for(i=0;i<nLength;i++)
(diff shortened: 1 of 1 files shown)


pascalscheffers:
Edited comment above to include sample PDFs from the FAA. These can be downloaded and freely shared with very few restrictions.

rouault:
This is more like a new feature, so master only / 3.5.0. Anyway 3.5.0 should be released closely after the next 3.4 patch release

pascalscheffers:
Thank you for the quick integration. We really appreciate it!

PR #5508, opened 2022-03-25
GeoEye metadata reader: avoid potential write stack-buffer overflow on long basename files (fixes #5506)

### LINKED ISSUE #5506: overflow in reader_geo_eye ?
```c
if( nBaseNameLen > 511 )
        return;

    // get _metadata.txt file

    // split file name by _rgb_ or _pan_
    char szMetadataName[512] = {0};
    size_t i;
    for(i = 0; i < nBaseNameLen; i++)
    {
        szMetadataName[i] = osBaseName[i];
        if(STARTS_WITH_CI(osBaseName.c_str() + i, "_rgb_") || STARTS_WITH_CI(osBaseName.c_str() + i, "_pan_"))
        {
            break;
        }
    }

    // form metadata file name
    CPLStrlcpy(szMetadataName + i, "_metadata.txt", 14);
    CPLString osIMDSourceFilename = CPLFormFilename( osDirName,
                                                        szMetadataName, nullptr );
    if (CPLCheckForFile(&osIMDSourceFilename[0], papszSiblingFiles))
    {
        m_osIMDSourceFilename = osIMDSourceFilename;
    }
    else
    {
        CPLStrlcpy(szMetadataName + i, "_METADATA.TXT", 14);
        osIMDSourceFilename = CPLFormFilename( osDirName, szMetadataName, nullptr );
        if (CPLCheckForFile(&osIMDSourceFilename[0], papszSiblingFiles))
        {
            m_osIMDSourceFilename = osIMDSourceFilename;
        }
    }

```
in airbusgeo/godal#84 a stack smashing error backtracable by gdb to this function occurs. It seems as though a 512 byte buffer is allocated, then a check that the input length is less than 511, and then an append of 14 to the 512 buffer which would overflow if the input is longer than 498.

rouault:


tbonfort:
somewhat offtopic, but I would have expected ossfuzz to catch this one very quickly, any idea why this is not the case?

rouault:
> somewhat offtopic, but I would have expected ossfuzz to catch this one very quickly, any idea why this is not the case?

yes, I've also wondered. I believe the reason is that the bug here was related to a particular pattern of filenames: basename close to 511 bytes, but less than it, with a _rgb_ / _pan_ suffix close to the end. Hence it requires we actually fuzz filenames passed to GDALOpen(), and that the GeoEye MDReader is triggered (which is only for the GTiff driver). But as far as I can see & remember, our fuzzers rely on fixed filenames, things like "/vsimem/test".
We would need a new specialized fuzzer whose fuzzed content contains both the filename (actually something to pass after a "/vsimem/ prefix) and the file content. Filed as https://github.com/OSGeo/gdal/issues/5519

PR #5510, opened 2022-03-25
DOC: add warning about GDAL_PAM_ENABLED=OFF

tbonfort:
## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:

### DIFF
--- a/gcore/gdalpamdataset.cpp
+++ b/gcore/gdalpamdataset.cpp
@@ -67,7 +67,7 @@ CPL_CVSID("$Id$")
  * GDAL_PAM_ENABLED configuration option (via CPLSetConfigOption(), or the
  * environment) to the value of YES (resp. NO). Note: The default value is
  * build dependent and defaults to YES in Windows and Unix builds. Warning:
- * Setting this option to OFF may have unwanted side-effects on drivers
+ * For GDAL < 3.5, setting this option to OFF may have unwanted side-effects on drivers
  * that rely on PAM functionality.
  *
  * <h3>PAM Proxy Files</h3>
--- a/gcore/gdalpamdataset.cpp
+++ b/gcore/gdalpamdataset.cpp
@@ -66,7 +66,9 @@ CPL_CVSID("$Id$")
  * PAM support can be enabled (resp. disabled) in GDAL by setting the
  * GDAL_PAM_ENABLED configuration option (via CPLSetConfigOption(), or the
  * environment) to the value of YES (resp. NO). Note: The default value is
- * build dependent and defaults to YES in Windows and Unix builds.
+ * build dependent and defaults to YES in Windows and Unix builds. Warning:
+ * Setting this option to OFF may have unwanted side-effects on drivers
+ * that rely on PAM functionality.
  *
  * <h3>PAM Proxy Files</h3>
  *


jratike80:
I suppose that you know some exact case, would you mind to share that?

tbonfort:
@jratike80 my example was a warning/failure of the DIMAP driver to set the band's colorinterpretation on Open()

jratike80:
You had turned it OFF because you believed that you do not need PAM and got surprised? I am asking because some other user will have the same problem one day and then I may be ready to help.

gcore/gdalpamdataset.cpp
```diff
@@ -66,7 +66,9 @@ CPL_CVSID("$Id$")
  * PAM support can be enabled (resp. disabled) in GDAL by setting the
  * GDAL_PAM_ENABLED configuration option (via CPLSetConfigOption(), or the
  * environment) to the value of YES (resp. NO). Note: The default value is
- * build dependent and defaults to YES in Windows and Unix builds.
+ * build dependent and defaults to YES in Windows and Unix builds. Warning:
+ * Setting this option to OFF may have unwanted side-effects on drivers
```
rouault, line 70:
```suggestion
 * For GDAL < 3.5, setting this option to OFF may have unwanted side-effects on drivers
```
(cf fix of https://github.com/OSGeo/gdal/pull/5511)

tbonfort:
I had turned it off to avoid spurious .aux.xml files being added . I saved myself the time of doing an rm -f *.aux.xml in exchange for half a day of debugging the dimap/pam code to find why I was now getting warnings :)

jratike80:
I guessed something like that. An opposite example is to clean GeoTIFF tags with `-co PROFILE=BASELINE` and wonder why nothing changed because the extra metadata got written into aux.xml.

gcore/gdalpamdataset.cpp
jratike80: Why not to say that the side effect is "CPLError() to be emitted"?

tbonfort: > 

because the side effect is also (for example in my case) that the band's colorinterpretation was not being set

jratike80: I think that anybody who disables PAM has considered that it does not contain any useful data so I guess that colorinterpretation is stored into DIMAP XML metadata and it was skipped because aux.xml was not found.

rouault: > so I guess that colorinterpretation is stored into DIMAP XML metadata and it was skipped because aux.xml was not found.

that's more complicated than that. A number of drivers have GDALMyDriverNameDataset / GDALMyDriverNameRasterBand classes that extend the base GDALPamDataset / GDALPamRasterBand classes, in case loading / saving .aux.xml is needed (e.g. the generic function of saving statistics, as used by QGIS, in .aux.xml depend on the drivers extending those classes). But the drivers may just do very basic things like someBand->SetColorInterpretation(GCI_RedBand) during dataset opening, that do not require any serialization capabilities at all. Except that before the fixed queued in #5511, that wouldn't work when PAM is disabled.

tbonfort: No, that's not exactly the case. The colorinterpretation is written in the dimap xml, but set programmatically in a PamDataset by the dimap driver at GDALOpen() time (https://github.com/OSGeo/gdal/blob/af52706a15cdc145fc4b740a5ee29229cca2412b/frmts/dimap/dimapdataset.cpp#L1388). This causes an error when GDAL_PAM_ENABLED is set to OFF, whether or not the aux.xml file is present or not.

PR #5511, opened 2022-03-25
Avoid CPLError() on number of drivers when PAM is disabled

rouault:
When GDAL is built without PAM support or the GDAL_PAM_ENABLED
configuration option is set to NO, a number of drivers will misbehave.
e.g drivers whose dataset/band classes extend GDALPamDataset/RasterBand,
and when methods like SetDescription(), SetColorInterpretation() are
called. Before that change, this would cause a CPLError() to be emitted.

With that change, when PAM is disabled, only the loading and writing of
.aux.xml files is disabled, but setters on PAM dataset/rasterband will
work.

CC @tbonfort

### DIFF
--- a/autotest/gcore/pam.py
+++ b/autotest/gcore/pam.py
@@ -41,13 +41,26 @@ from osgeo import osr
 import pytest
 
 ###############################################################################
+@pytest.fixture(autouse=True, scope='module')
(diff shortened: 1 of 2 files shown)


tbonfort:
thanks @rouault , fix confirmed in the dimap case

PR #5528, opened 2022-03-28
Remove deprecated SWIG Perl bindings (refs #3555) 

rouault:
As warned, they were planed for removal in 3.5. The alternative documented in https://gdal.org/api/index.html#gdal-ogr-in-other-languages is https://metacpan.org/dist/Geo-GDAL-FFI

CC @ajolma


ajolma:
👍

PR #5534, opened 2022-03-29
fix S57 features updates more than 255 times

maitaivw:
## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:

### DIFF
--- a/ogr/ogrsf_frmts/s57/s57reader.cpp
+++ b/ogr/ogrsf_frmts/s57/s57reader.cpp
@@ -2686,8 +2686,8 @@ bool S57Reader::ApplyRecordUpdate( DDFRecord *poTarget, DDFRecord *poUpdate )
     if( poRVER_SFD == nullptr )
         return false;
 
-    unsigned char *pnRVER
-        = (unsigned char *) poKey->GetSubfieldData( poRVER_SFD, nullptr, 0 );
+    unsigned int *pnRVER
+        = (unsigned int *) poKey->GetSubfieldData( poRVER_SFD, nullptr, 0 );
 
     *pnRVER += 1;
 


rouault:
I've pushed an alternate pull request in https://github.com/OSGeo/gdal/pull/5546 . This one is actually not correct since the RVER field is according to the spec a uint16 value, not a int32 one. #5546 asserts that the field is of the correct format before proceeding, and also handles potential big-endian hosts.
I've also added logic to check EDTN/UPDN of update files.

maitaivw:
Thanks, I was ready to push the EDTN/UPDN logic after this one, I will revert to yours and test. I have another pending change, but I will ask on the maillist before because it is not very clear.

rouault:
> I will revert to yours and test.

yes I'll appreciate your testing of https://github.com/OSGeo/gdal/pull/5546. I did my tests on the following charts: US5PA13M, US5ME51M, US2MI80M

maitaivw:
HI Even,

I have tested a bit and I have some remarks:

Concerning the RVER fix allowing more than 255 updates for a feature, it works meaning that subsequent updates are not rejected anymore. The bad thing is that RVER attribute at OGRFeature level is not updated, it is only at DDFRecord level.

Concerning the change for EDTN, I don't understand why you test a difference only if not equal to 0. According to https://iho.int/uploads/user/pubs/standards/s-57/20ApB1.pdf page 18 it says:

In order to delete a data set, an update cell file is created, containing only the Data Set General
Information record with the "Data Set Identifier" [DSID] field. The "Edition Number" [EDTN] subfield must
be set to 0. This message is only used to cancel a base cell file.

So for me an update with EDTN=0 means the whole chart should not be used, and therefore GDAL should refuse to open it. You can find examples at US5NY1KM, US5RI23M and US5RI24M.

--Philippe

rouault:
> I have tested a bit and I have some remarks:

thanks for your testing. Your remarks make sense (I have some hesitation if GDAL should refuse to open when a cancel update is detected, but probably yes indeed), but I don't think I'll have time in the near future to implement myself your suggested enhancements. It looks to me #5546 is a step in the right direction / does not regress ? if so, I'd go for merging it, so that it can serve as a base for further enhancements.

maitaivw:
Yes it does not regress and is an improvement. The caller can eventually check the edition number and decide that the chart is canceled if EDTN=0.

PR #5542, opened 2022-03-29
Fix out of bounds read in CPLRecodeFromWCharIconV()

stefanuhrig:
If the number of bytes per character was different from the size of
wchar_t, the size of the input was not computed correctly.

### DIFF
--- a/port/cpl_recode_iconv.cpp
+++ b/port/cpl_recode_iconv.cpp
@@ -256,7 +256,7 @@ char *CPLRecodeFromWCharIconv( const wchar_t *pwszSource,
         reinterpret_cast<char*>(pszIconvSrcBuf));
 
     /* iconv expects a number of bytes, not characters */
-    nSrcLen *= sizeof(wchar_t);
+    nSrcLen *= nTargetCharWidth;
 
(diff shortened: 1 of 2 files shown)


port/cpl_recode_iconv.cpp
```diff
@@ -256,7 +256,7 @@ char *CPLRecodeFromWCharIconv( const wchar_t *pwszSource,
         reinterpret_cast<char*>(pszIconvSrcBuf));
 
     /* iconv expects a number of bytes, not characters */
-    nSrcLen *= sizeof(wchar_t);
+    nSrcLen *= nTargetCharWidth;
```
rouault, line 259: I believe that lines 280 and 281 should also be modified to:
```
                nSrcLen -= nTargetCharWidth;
                pszSrcBuf += nTargetCharWidth;
```

stefanuhrig: Good catch. There is also the corner case that the destination buffer might already be full when the terminating null character is written.

rouault:
Backport PR in https://github.com/OSGeo/gdal/pull/5551

PR #5543, opened 2022-03-29
ogr2ogr: fix cutting of geometries in projected coordinates intersecting antimeridian, in some configurations

rouault:


vincentsarago:
🙏 thanks for the fix @rouault

PR #5547, opened 2022-03-30
gdalbuildvrt: fix 3.4.1 regression when using -b to select less bands than source dataset

rouault:
Fixes issue introduced by bbdced448d964d62917ece73b2afbd2be6067584
and reported on https://lists.osgeo.org/pipermail/gdal-dev/2022-March/055662.html


jratike80:
The definition is not accurate but I hope that the fix is. The issue happens with RGB image also with:
`-b 2 -b 3` and `-b 1 -b 3`

rouault:
> The definition is not accurate

finding a short and accurate commit title is sometimes more difficult/longer than the fix :-)
Yes the fix was correct for the other cases. I've just enhanced the regression test to demonstrate it

PR #5550, opened 2022-03-30
FlatGeobuf: in SPATIAL_INDEX=NO mode, write a meaningful featureCount and extent in the header

rouault:
CC @bjornharrtell 

2 commits: the first one is a bugfix, the second one an improvement:
- fix GetFeature() when featuresCount != 0 and indexNodeSize == 0, such as files produced by the next commit
- in SPATIAL_INDEX=NO mode, write a meaningful featureCount and extent in the header. Previously featureCount=0 and extent=null where written. Writing meaningful values is a bit tricky since it involves first creating a dummy header of the right size, and overriding it at file closing.

### DIFF
--- a/autotest/ogr/ogr_flatgeobuf.py
+++ b/autotest/ogr/ogr_flatgeobuf.py
@@ -435,6 +435,10 @@ def test_ogr_flatgeobuf_mixed():
     destDS = ogr.Open('/vsimem/test.fgb')
     srcLyr = srcDS.GetLayer(0)
     destLyr = destDS.GetLayer(0)
+    assert destLyr.TestCapability(ogr.OLCFastFeatureCount)
+    assert destLyr.TestCapability(ogr.OLCFastGetExtent)
+    assert destLyr.GetFeatureCount(force=0) == srcLyr.GetFeatureCount()
+    assert destLyr.GetExtent(force=0) == srcLyr.GetExtent()
     ogrtest.compare_layers(srcLyr, destLyr)
 
     ogr.GetDriverByName('FlatGeobuf').DeleteDataSource('/vsimem/test.fgb')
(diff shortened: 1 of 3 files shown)


bjornharrtell:
My favorite use case for FlatGeobuf without index is that it can be written sequentially and as such support read/write streaming without waiting at either end. Perhaps that's not a use case where GDAL is fitting well into but somehow I view non-indexed FlatGeobuf with known feature count as kind of an abnormality so do we really need to support that (and have it being the default behaviour?).

rouault:
> My favorite use case for FlatGeobuf without index is that it can be written sequentially and as such support read/write streaming without waiting at either end.

agreed

> Perhaps that's not a use case where GDAL is fitting well into but somehow I view non-indexed FlatGeobuf with known feature count as kind of an abnormality so do we really need to support that (and have it being the default behaviour?).

that's a bit like a shapefile or GeoPackage: writing a spatial index is the nominal case, but when not doing it, there's no reason to have missing feature count and extent, and thus a suboptimal user experience. What triggered this is that I tried to create a big FlatGeobuf file without spatial index to compare with Arrow/Parquet, but I was surprised that "ogrinfo my.fgb -al -so" wasn't instant. And I discovered then that it had to compute the feature count and extent.

bjornharrtell:
@rouault well, in my case it would be suboptimal as I want sequential write only? So I would say it depends on use case if it's desired to get a feature count or not in the non-indexed variant.

rouault:
bugfix commit https://github.com/OSGeo/gdal/pull/5550/commits/eb930d948389e3431161445061d85762fd0fbc34 backported in 3.4 per d6a80b75e3574ffec00f23dcd8cc75017b8699fb

PR #5554, opened 2022-03-31
GPKG: fix opening files in NOLOCK=YES mode that contains '?' or '#' (fixes qgis/QGIS#48024)

rouault:


agiudiceandrea:
May this patch be able to fix also https://github.com/qgis/QGIS/issues/48003?

rouault:
> May this patch be able to fix also [qgis/QGIS#48003](https://github.com/qgis/QGIS/issues/48003)?

hum, I'm not sure at all... and don't have the setup to check. I'm not sure if SQLite URIs are compatible of Windows network file share names (I have strong doubts given that https://www.sqlite.org/uri.html mentions "Convert all sequences of two or more "/" characters into a single "/" character.")

rouault:
> hum, I'm not sure at all... and don't have the setup to check. I'm not sure if SQLite URIs are compatible of Windows network file share names (I have strong doubts given that https://www.sqlite.org/uri.html mentions "Convert all sequences of two or more "/" characters into a single "/" character.")

Potential QGIS side fix for that issue with network shares: https://github.com/qgis/QGIS/pull/48059#issuecomment-1086485947

PR #5564, opened 2022-04-02
FITS: fix non-conformant use of &v[0] that crashes with clang 14 -O2

rouault:
Fixes crash on fedora:rawhide


rouault:
backported to 3.4

PR #5594, opened 2022-04-07
CMake: add a GDAL_USE_INTERNAL_LIBS=ON/OFF/WHEN_NO_EXTERNAL variable

rouault:
From updated build_hints page:
```

Some of the GDAL dependencies (GEOTIFF, GIF, JPEG, JSONC, LERC, OPENCAD, PNG, QHULL, TIFF, ZLIB)
have a copy of their source code inside the GDAL source code tree. It is possible
to enable this internal copy by setting the GDAL_USE_<Packagename_in_upper_case>_INTERNAL:BOOL=ON
variable. When set, this has precedence over the external library that may be
detected. The behavior can also be globally controlled with the following variable:

.. option:: GDAL_USE_INTERNAL_LIBS=ON/OFF/WHEN_NO_EXTERNAL

     Control how internal libraries should be used.
     If set to ON, they will always be used.
     If set to OFF, they will never be used (unless individually enabled with
     GDAL_USE_<Packagename_in_upper_case>_INTERNAL:BOOL=ON)
     If set to WHEN_NO_EXTERNAL (default value), they will be used only if no
     corresponding external library is found and enabled.
     This option should be set before CMakeCache.txt is created.
```

Also add at the end of cmake phase, a summary of internal libs used. For
example:

```
-- Internal libraries enabled:

 * TIFF internal library enabled
 * GEOTIFF internal library enabled
 * PNG internal library enabled
 * GIF internal library enabled
 * OPENCAD internal library enabled
 * LERC internal library enabled
```

### DIFF
--- a/cmake/helpers/CheckDependentLibraries.cmake
+++ b/cmake/helpers/CheckDependentLibraries.cmake
@@ -19,6 +19,15 @@ option(
   "Whether detected external libraries should be used by default. This should be set before CMakeCache.txt is created."
   ON)
 
+set(GDAL_USE_INTERNAL_LIBS_ALLOWED_VALUES ON OFF WHEN_NO_EXTERNAL)
+set(
+  GDAL_USE_INTERNAL_LIBS WHEN_NO_EXTERNAL
+  CACHE STRING "Control how internal libraries should be used by default. This should be set before CMakeCache.txt is created.")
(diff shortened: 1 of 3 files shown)


mloskot (approved):
Thanks!

PR #5603, opened 2022-04-09
CMake: add support for PostgreSQL 14 ...

rouault:
... and avoid copy&paste related to CMake versions when installing a
static build.

I've submitted in https://gitlab.kitware.com/cmake/cmake/-/merge_requests/7158
the addition of PG 14 so I've anticipated the updated module will be
available in CMake 3.24

Original credits to @jmckenna / https://github.com/MapServer/MapServer/pull/6509

### DIFF
--- a/cmake/helpers/CheckDependentLibraries.cmake
+++ b/cmake/helpers/CheckDependentLibraries.cmake
@@ -582,7 +582,10 @@ gdal_check_package(NetCDF "Enable netCDF driver" CAN_DISABLE
 gdal_check_package(OGDI "Enable ogr_OGDI driver" CAN_DISABLE)
 # OpenCL warping gives different results than the ones expected by autotest, so disable it by default even if found.
 gdal_check_package(OpenCL "Enable OpenCL (may be used for warping)" DISABLED_BY_DEFAULT)
+
+set(PostgreSQL_ADDITIONAL_VERSIONS "14" CACHE STRING "Additional PostgreSQL versions to check")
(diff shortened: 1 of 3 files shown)


cmake/modules/3.24/FindPostgreSQL.cmake
```diff
@@ -102,7 +102,7 @@ set(PostgreSQL_ROOT_DIR_MESSAGE "Set the PostgreSQL_ROOT system variable to wher
 
 
 set(PostgreSQL_KNOWN_VERSIONS ${PostgreSQL_ADDITIONAL_VERSIONS}
-    "13" "12" "11" "10" "9.6" "9.5" "9.4" "9.3" "9.2" "9.1" "9.0" "8.4" "8.3" "8.2" "8.1" "8.0")
+    "14" "13" "12" "11" "10" "9.6" "9.5" "9.4" "9.3" "9.2" "9.1" "9.0" "8.4" "8.3" "8.2" "8.1" "8.0")
```
dg0yt, line 105: If there is no other change to this module, I would propose to leave this find module unchanged, in directory `3.20`, and add missing versions via `set(PostgreSQL_ADDITIONAL_VERSIONS "14" CACHE STRING "...")`, just before calling `gdal_check_package`/`find_package`.

dg0yt (changes requested):

cmake/modules/3.24/FindPostgreSQL.cmake
rouault: good point. PR amended

PR #5605, opened 2022-04-10
CMake: write_basic_package_version_file(): use COMPATIBILITY SameMinorVersion for CMake >= 3.10.1

rouault:


rouault:
arg, the comment was misleading. The minimum version for SameMinorVersion is 3.11

tbonfort:
fix confirmed, thanks @rouault

PR #5610, opened 2022-04-13
Improve configuration options documentation + small improvements

theroggy:
- Mainly improvements to the documentation of configuration options, typically 
  - add :decl_configoption: they should be present everywhere where applicable now.
  - if there is a specific section about configuration option for the driver, add a link to the configuration options page so the usage instructions are easier to find.
- Sometimes some small improvements to formatting like adding headers,... were done
- In the GPKG driver documentation added a link to "Performance hints" section in the SQLITE driver page as the performance hints are applicable to both drivers.

### DIFF
--- a/doc/source/drivers/vector/amigocloud.rst
+++ b/doc/source/drivers/vector/amigocloud.rst
@@ -55,15 +55,6 @@ For example: **"AmigoCloud:1234 datasets"**
     | 5551      | points
     | 5552      | lines
 
-Configuration options
----------------------
-
-The following configuration options are available :
-
--  AMIGOCLOUD_API_URL: defaults to https://www.amigocloud.com/api/v1.
-   Can be used to point to another server.
--  AMIGOCLOUD_API_KEY: see following paragraph.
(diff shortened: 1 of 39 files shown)


doc/source/drivers/vector/amigocloud.rst
```diff
@@ -121,6 +112,16 @@ The following layer creation options are available:
 -  **GEOMETRY_NULLABLE**\ =YES/NO: Whether the values of the geometry
    column can be NULL. Defaults to YES.
 
+Configuration options
+---------------------
+
+The following :doc:`configuration options <../../user/configoptions>` are
```
rouault, line 118: Would be best to link it to it with :ref: to avoid hardcoding the path . The same remark applies to a number of other similar places:
```suggestion
The following :ref:`configuration options <configoptions>` are 
```

theroggy:
@rouault I changed the :doc: on all locations to :ref:. That's indeed a lot cleaner maintence-wise.

PR #5612, opened 2022-04-14
slight improvement on GDALSuggestedWarpOutput2

desertstsung:
and TODO mentioned in L444 doesn't need to do anything, IMO


alg/gdaltransformer.cpp
```diff
@@ -437,15 +437,17 @@ GDALSuggestedWarpOutput2( GDALDatasetH hSrcDS,
         static_cast<double>(std::min(nInYSize, nInXSize)) / N_PIXELSTEP + 0.5);
     if( nSteps < 20 )
         nSteps = 20;
-    nSteps = std::min(nSteps, 100);
+    else if( nSteps > 100 )
+        nSteps = 100;
+    int nStepsPlusOne = nSteps + 1;
```
rouault, line 442: this should be moved after the "retry:" label as nSteps can be modified at line 469 before a goto retry

alg/gdaltransformer.cpp
```diff
@@ -467,51 +469,56 @@ GDALSuggestedWarpOutput2( GDALDatasetH hSrcDS,
...
             nSteps = 20;
+            nStepsPlusOne = 21;
             goto retry;
         }
         return CE_Failure;
     }
 
     padfY = padfX + nSampleMax;
-    padfZ = padfX + nSampleMax * 2;
+    padfZ = padfX + (nSampleMax<<1);
```
rouault, line 479: unnecessary/hard-to-read change. compilers are smart enough to optimize a multiplication by two with a bitshift. no need to make humans aware of that trick

alg/gdaltransformer.cpp
```diff
@@ -467,51 +469,56 @@ GDALSuggestedWarpOutput2( GDALDatasetH hSrcDS,
...
     padfYRevert = padfXRevert + nSampleMax;
-    padfZRevert = padfXRevert + nSampleMax * 2;
+    padfZRevert = padfXRevert + (nSampleMax<<1);
```
rouault, line 481: no need to bitshift

alg/gdaltransformer.cpp
```diff
@@ -467,51 +469,56 @@ GDALSuggestedWarpOutput2( GDALDatasetH hSrcDS,
...
 
     // Take N_STEPS steps.
-    for( int iStep = 0; iStep <= nSteps; iStep++ )
+    double dfRatio = 0.0;
+    int iStep  = 0;
+    int iStep2 = 0;
```
rouault, line 486: the scope of those variables should be kept inside the loop to minimize the number of things in the outer scope. compilers are smart enough to fold variables into registers

alg/gdaltransformer.cpp
```diff
@@ -467,51 +469,56 @@ GDALSuggestedWarpOutput2( GDALDatasetH hSrcDS,
...
-        padfY[3 * (nSteps + 1) + iStep] = dfRatio * nInYSize;
-        padfZ[3 * (nSteps + 1) + iStep] = 0.0;
+        iStep2 += nStepsPlusOne;
+        padfX[iStep2] = nInXSize;
+        padfY[iStep2] = dfRatio * nInYSize;
+        padfZ[iStep2] = 0.0;
     }
 
-    int nSamplePoints = 4 * (nSteps + 1);
+    int nSamplePoints = nStepsPlusOne<<2;
```
rouault, line 515: unneeded bitshift

alg/gdaltransformer.cpp
```diff
@@ -527,18 +534,18 @@ GDALSuggestedWarpOutput2( GDALDatasetH hSrcDS,
     constexpr int SIGN_FINAL_UNINIT = -2;
     constexpr int SIGN_FINAL_INVALID = 0;
     int iSignDiscontinuity = SIGN_FINAL_UNINIT;
+    int nFailedCount = 0;
+    int iSignArray[2] = {-1, 1};
```
rouault, line 538:
```suggestion
    const int iSignArray[2] = {-1, 1};
```

alg/gdaltransformer.cpp
```diff
@@ -527,18 +534,18 @@ GDALSuggestedWarpOutput2( GDALDatasetH hSrcDS,
...
-        else
+        if( pabSuccess[i] )
         {
             // Fix for https://trac.osgeo.org/gdal/ticket/7243
             // where echo "-2050000.000 2050000.000" |
             //              gdaltransform -s_srs EPSG:3411 -t_srs EPSG:4326
             // gives "-180 63.691332898492"
             // but we would rather like 180
-            if( iSignDiscontinuity == 1 || iSignDiscontinuity == -1 )
+            if( iSignDiscontinuity&0x01 )
```
rouault, line 548: too smart change, and relies on the complement-to-two representation of signed integers

alg/gdaltransformer.cpp
```diff
@@ -547,9 +554,9 @@ GDALSuggestedWarpOutput2( GDALDatasetH hSrcDS,
                     iSignDiscontinuity = SIGN_FINAL_INVALID;
                 }
             }
-            else if( iSignDiscontinuity == SIGN_FINAL_UNINIT )
+            else if( iSignDiscontinuity )
```
rouault, line 557: please revert that change. it is too smart and relies on the actual value of SIGN_FINAL_UNINIT

alg/gdaltransformer.cpp
```diff
@@ -564,21 +571,30 @@ GDALSuggestedWarpOutput2( GDALDatasetH hSrcDS,
...
-                                         iSignDiscontinuity * 180.0 };
-                    double ayTemp[2] = { padfY[i], padfY[i] };
-                    double azTemp[2] = { padfZ[i], padfZ[i] };
-                    int abSuccess[2] = {FALSE, FALSE};
+                    axTemp[0] = iSignDiscontinuity * -180.0;
+                    axTemp[1] = iSignDiscontinuity *  180.0;
+                    ayTemp[0] = padfY[i];
+                    ayTemp[1] = padfY[i];
+                    azTemp[0] = padfZ[i];
+                    azTemp[1] = padfZ[i];
```
rouault, line 597: please revert those changes.

alg/gdaltransformer.cpp
```diff
@@ -595,18 +611,20 @@ GDALSuggestedWarpOutput2( GDALDatasetH hSrcDS,
...
-        if( !pfnTransformer(pTransformArg, TRUE, nSamplePoints,
+        if( pfnTransformer(pTransformArg, TRUE, nSamplePoints,
                             padfXRevert, padfYRevert, padfZRevert, pabSuccess) )
         {
-            nFailedCount = 1;
-        }
-        else
-        {
+            double dfExpectedX = 0.0;
+            double dfExpectedY = 0.0;
```
rouault, line 627: please revert those changes. best is to keep variables in the most nested possible scope

alg/gdaltransformer.cpp
```diff
@@ -615,62 +633,60 @@ GDALSuggestedWarpOutput2( GDALDatasetH hSrcDS,
...
-                {
-                    dfExpectedX   = nInXSize;
-                    dfExpectedY   = dfRatio * nInYSize;
+                    break;
+                case 2:
+                    dfExpectedY = dfRatio * nInYSize;
+                    break;
+                default:
+                    dfExpectedX = nInXSize;
+                    dfExpectedY = dfRatio * nInYSize;
```
rouault, line 656: I'm a bit skeptical about using switch() here.
At the very least, the order of the case should be reorderd into the logical order 0, 1, 2, default. And then you can remove the // avoid -Wimplicit-fallthrough comment

alg/gdaltransformer.cpp
```diff
@@ -615,62 +633,60 @@ GDALSuggestedWarpOutput2( GDALDatasetH hSrcDS,
...
 
         // Take N_STEPS steps.
-        for( int iStep = 0; iStep <= nSteps; iStep++ )
+        for( iStep = 0; iStep <= nSteps; iStep++ )
         {
-            double dfRatio = (iStep == nSteps) ? 1.0 : iStep * dfStep;
+            dfRatio = (iStep == nSteps) ? 1.0 : iStep * dfStep;
 
-            for( int iStep2 = 0; iStep2 <= nSteps; iStep2++ )
+            for( iStep2 = 0; iStep2 <= nSteps; iStep2++ )
```
rouault, line 689: variables should be declared in the most nested scope possible

alg/gdaltransformer.cpp
```diff
@@ -907,10 +921,10 @@ GDALSuggestedWarpOutput2( GDALDatasetH hSrcDS,
...
                     yOut = y;
                 }
-                dfMinXOut = std::min(dfMinXOut, xOut);
-                dfMinYOut = std::min(dfMinYOut, yOut);
-                dfMaxXOut = std::max(dfMaxXOut, xOut);
-                dfMaxYOut = std::max(dfMaxYOut, yOut);
+                if (xOut < dfMinXOut) dfMinXOut = xOut;
+                if (yOut < dfMinYOut) dfMinYOut = yOut;
+                if (xOut > dfMaxXOut) dfMaxXOut = xOut;
+                if (yOut > dfMaxYOut) dfMaxYOut = yOut;
```
rouault, line 927: I don't see the point of that change

alg/gdaltransformer.cpp
```diff
@@ -1022,8 +1036,8 @@ GDALSuggestedWarpOutput2( GDALDatasetH hSrcDS,
     const double dfPixels = (dfMaxXOut - dfMinXOut) / dfPixelSize;
     const double dfLines = (dfMaxYOut - dfMinYOut) / dfPixelSize;
 
-    const int knIntMax = std::numeric_limits<int>::max();
-    if( dfPixels > knIntMax - 1 || dfLines > knIntMax - 1 )
+    const int knIntMax = std::numeric_limits<int>::max() - 1;
+    if( dfPixels > knIntMax || dfLines > knIntMax )
```
rouault, line 1040:
```suggestion
    const int knIntMaxMinusOne = std::numeric_limits<int>::max() - 1;
    if( dfPixels > knIntMaxMinusOne || dfLines > knIntMaxMinusOne )
```

alg/gdaltransformer.cpp
```diff
@@ -1049,10 +1063,10 @@ GDALSuggestedWarpOutput2( GDALDatasetH hSrcDS,
 /*      image. If so, adjust the x pixel size a bit in the hope it will */
 /*      fit.                                                            */
 /* -------------------------------------------------------------------- */
-    for( const auto& dfRatio : adfRatioArray )
+    for( const auto& dfiRatio : adfRatioArray )
```
rouault, line 1066: improper variable renaming. same for line 1091

rouault (changes requested):
To be honest, I'm a bit skeptical about a number of the proposed changes. Some of them make the code less readable / more fragile. They appear to have been done in the hope of making the code faster, but I assume that compilers do a better job than us at those micro-optimizations.

desertstsung:
Thanks for your sincere advice, and there's a lot I'll go on learning :)

alg/gdaltransformer.cpp
desertstsung: OK, `if else` is more human readable than `switch`, I'll revert this change

desertstsung:
I've revert the unnecessary changes, please have a check.

rouault:
It causes test failures. See https://ci.appveyor.com/project/OSGeo/gdal/builds/43262687 . So there must be a change in behaviour somewhere in the code changes

desertstsung:
Fixed

rouault (approved):

PR #5617, opened 2022-04-16
Add layer renaming capacity

rouault:
- Add OLCRename layer capability
- Add virtual method: OGRErr OGRLayer::Rename(const char* pszNewName)
- Add C API: OGRErr OGR_L_Rename(OGRLayerH, const char* pszNewName)
- Add SWIG bindings: ogr.OLCRename and Layer.Rename()
- Implemented in Shapefile, GPKG (already there, mapped to new API), FileGDB, PG and NGW (already there, mapped to new API) drivers

```
/** Rename layer.
 *
 * This operation is implemented only by layers that expose the OLCRename capability.
 *
 * This operation will fail if a layer with the new name already exists.
 *
 * On success, GetDescription() and GetLayerDefn()->GetName() will return
 * pszNewName.
 *
 * Renaming the layer may interrupt current feature iteration.
 *
 * @param hLayer     Layer to rename.
 * @param pszNewName New layer name. Must not be NULL.
 * @return OGRERR_NONE in case of success
 */
```


nyalldawson (approved):
I think it could also be worthwhile to add a corresponding driver metadata for "this driver supports renaming", so that clients can determine this capability quickly in advance.

rouault:
> I think it could also be worthwhile to add a corresponding driver metadata for "this driver supports renaming", so that clients can determine this capability quickly in advance.

In which use case, you'd see this useful  ? (I'd want to use "duplicating' capability flags when possible) The scenario I imagined would be a GIS client that  would propose a contextual menu when right-clicking on a layer in some browser and decide if it can propose a "Rename" item in it, but at that point it has access to the layer object and can use OLCRename. Or if it offers a "Rename layer" dialog where in the dialog you begin by selecting the layer, it is sufficient to call TestCapability(OLCRename) on any layer before populating the list of layers or enabling access to the functionality.

nyalldawson:
I was thinking from the perspective of the qgis file browser, which doesn't have a layer object available upfront. A layer object could of course be created to test this, but it'd have to be in update mode and to me it seems a bit excessive/risky to open an updateable layer just to test whether an edit of operation (which the user hasn't yet opted to do) is theoretically possible.

rouault:
> I was thinking from the perspective of the qgis file browser, which doesn't have a layer object available upfront.

makes sense. GDAL_DCAP_RENAME_LAYERS capability added

nyalldawson:
Thanks Even!

PR #5620, opened 2022-04-16
FileGDB: add support for writing field domains

rouault:


nyalldawson (approved):

nyalldawson:
The DMD_CREATION_FIELD_DOMAIN_TYPES driver metadata should also be set

rouault:
> The DMD_CREATION_FIELD_DOMAIN_TYPES driver metadata should also be set

good point. Just added

PR #5624, opened 2022-04-17
OGRGeometryFactory::forceTo(): fix potential crash (fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=46784)

rouault:


rouault:
probably doesn't need backport, as the behavior is slightly different in 3.4

PR #5625, opened 2022-04-17
HANA: Coverity scan related fixes

rouault:
CC @mrylov

### DIFF
--- a/ogr/ogrsf_frmts/hana/ogrhanafeaturereader.cpp
+++ b/ogr/ogrsf_frmts/hana/ogrhanafeaturereader.cpp
@@ -63,10 +63,16 @@ odbc::String CreateStringFromValues(
 template<typename T>
 T castInt(int value)
 {
-    if (!(std::numeric_limits<T>::min() >= value
-          || value <= std::numeric_limits<T>::max()))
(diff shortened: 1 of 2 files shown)


ogr/ogrsf_frmts/hana/ogrhanafeaturereader.cpp
```diff
@@ -63,12 +63,18 @@ odbc::String CreateStringFromValues(
 template<typename T>
 T castInt(int value)
 {
-    if (!(std::numeric_limits<T>::min() >= value
-          || value <= std::numeric_limits<T>::max()))
+    if (!(value >= std::numeric_limits<T>::min() &&
+          value <= std::numeric_limits<T>::max()))
```
nyalldawson, line 67: Wouldn't this be a bit clearer? (Or is the ! condition used to handle some corner case with nan values or suchlike?)

```suggestion
    if ( value < std::numeric_limits<T>::min() ||
          value > std::numeric_limits<T>::max() )
```

rouault: changed to your suggestion. It is only used with T=int8_t and int16_t

nyalldawson (approved):

PR #5631, opened 2022-04-19
Various optimizations for access to OGRFeature, and apply it to Parquet driver

rouault:
With those optimizations, "ogr2ogr out.parquet nz-buildings-outlines.parquet" goes from 19.7 seconds down to 12.6 sec

### DIFF
--- a/apps/ogr2ogr_lib.cpp
+++ b/apps/ogr2ogr_lib.cpp
@@ -4401,7 +4401,7 @@ std::unique_ptr<TargetLayerInfo> SetupTargetLayer::Setup(OGRLayer* poSrcLayer,
         psInfo->m_oMapDomainToKV[poDomain] = std::move(oMapCodeValue);
     }
 
-    // Detect if we can directly passed the source feature to the CreateFeature()
+    // Detect if we can directly pass the source feature to the CreateFeature()
     // method of the target layer, without doing any copying of field content.
     psInfo->m_bCanAvoidSetFrom = false;
     if( !m_bExplodeCollections && iSrcZField == -1 )
--- a/apps/ogr2ogr_lib.cpp
+++ b/apps/ogr2ogr_lib.cpp
@@ -388,6 +388,7 @@ struct TargetLayerInfo
     int          m_iRequestedSrcGeomField = -1;
     bool         m_bPreserveFID = false;
     const char  *m_pszCTPipeline = nullptr;
+    bool         m_bCanAvoidSetFrom = false;
 };
 
 struct AssociatedLayers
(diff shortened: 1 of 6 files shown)


apps/ogr2ogr_lib.cpp
```diff
@@ -4400,6 +4401,48 @@ std::unique_ptr<TargetLayerInfo> SetupTargetLayer::Setup(OGRLayer* poSrcLayer,
         psInfo->m_oMapDomainToKV[poDomain] = std::move(oMapCodeValue);
     }
 
+    // Detect if we can directly passed the source feature to the CreateFeature()
```
nyalldawson, line 4404:
```suggestion
    // Detect if we can directly pass the source feature to the CreateFeature()
```

apps/ogr2ogr_lib.cpp
```diff
@@ -4400,6 +4401,48 @@ std::unique_ptr<TargetLayerInfo> SetupTargetLayer::Setup(OGRLayer* poSrcLayer,
...
+            psInfo->m_bCanAvoidSetFrom = false;
+        }
+        else
+        {
+            for( int i = 0; i < nSrcFieldCount; ++i )
+            {
+                auto poSrcFieldDefn = poSrcFDefn->GetFieldDefn(i);
+                auto poDstFieldDefn = poDstFDefn->GetFieldDefn(i);
+                if( poSrcFieldDefn->GetType() != poDstFieldDefn->GetType() ||
+                    psInfo->m_anMap[i] != i )
```
nyalldawson, line 4423: Do we need to consider field truncation here if the destination field has a size limit smaller than the source? (Or will that be handled automatically elsewhere at the driver level?)

apps/ogr2ogr_lib.cpp
```diff
@@ -4400,6 +4401,48 @@ std::unique_ptr<TargetLayerInfo> SetupTargetLayer::Setup(OGRLayer* poSrcLayer,
         psInfo->m_oMapDomainToKV[poDomain] = std::move(oMapCodeValue);
     }
 
+    // Detect if we can directly passed the source feature to the CreateFeature()
+    // method of the target layer, without doing any copying of field content.
+    psInfo->m_bCanAvoidSetFrom = false;
+    if( !m_bExplodeCollections && iSrcZField == -1 )
```
nyalldawson, line 4407: I might be missing something (the diff is really hard to read here unfortunately), but don't we also need to consider other settings like m_poClipSrc ?

ogr/ogrutils.cpp
```diff
@@ -1364,16 +1363,19 @@ char* OGRGetXMLDateTime(const OGRField* psField, bool bAlwaysMillisecond)
                      (TZFlag > 100) ? '+' : '-', TZHour, TZMinute);
     }
 
+    // sizeof() includes null terminator. +6 is to make -Wformat-truncation= happy
+    constexpr size_t nMaxSize = sizeof("YYYY-MM-DDThh:mm:ss.sss+hh:mm")+6;
```
nyalldawson, line 1367: Possibly stupid consideration, unrelated to this PR: Do you think we'll ever encounter a "YYYY-MM-DDThh:mm:ss.sss+-03:00" format string in the wild (with '+-' literal)? (Eg. if some external app which is generating the strings has the "+" character hardcoded in their string format and then feeds in a negative offset value)

apps/ogr2ogr_lib.cpp
rouault: This consideration is not even taken into account by the non-optimized OGRFeature::SetFrom() method. At the OGRFieldDefn level, field width semantics is also probably a bit fuzzy currently. Some drivers might interpret it as a number of bytes and others as a number of UTF-8 characters. Some drivers can also happily store a string that is larger the indicated field size.

apps/ogr2ogr_lib.cpp
rouault: Good point. I now remember I was a bit perplexed by this because there is also m_poClipSrc being taken into account later in https://github.com/OSGeo/gdal/blob/5427fe95881c19adc67e8887f7d74efe86788e0c/apps/ogr2ogr_lib.cpp#L4983. It looked like the logic at https://github.com/OSGeo/gdal/blob/5427fe95881c19adc67e8887f7d74efe86788e0c/apps/ogr2ogr_lib.cpp#L4829 might be redundant, but I now believe that I understand it is for the case where someone would use -clipsrc to filter a geometry to and produce a a-spatial layer. But this situation will cause nSrcGeomFieldCount (>=1) != nDstGeomFieldCount (=0), which causes m_bCanAvoidSetFrom to be set to false. So I believe we're good as it.

ogr/ogrutils.cpp
rouault: > Possibly stupid consideration, unrelated to this PR: Do you think we'll ever encounter a "YYYY-MM-DDThh:mm:ss.sss+-03:00" format string in the wild (with '+-' literal)?

I don't see a compelling reason to take into account this situation, at least not until proven that there's a major data source that would have such bug in its data

PR #5642, opened 2022-04-21
ogr2ogr: make 'ogr2ogr [-f PostgreSQL] PG:dbname=.... source [srclayer] -lco OVERWRITE=YES' work as if -overwrite was specified (fixes #5640)

### LINKED ISSUE #5640: ogr2ogr -lco OVERWRITE doesn't work on PostgreSQL public schema
See https://gis.stackexchange.com/questions/429260/why-ogr2ogr-lco-overwrite-doesnt-work-on-postgresql-public-schemas/429283

Overwriting a PostgreSQL layer with layer creation option `-lco OVERWRITE=YES` does not work if the table is in the "public" schema.

```
ogr2ogr -f "PostgreSQL" PG:"host=localhost port=5432 dbname=dbname user=user password=password" -nlt geometry -lco geometry_name=geometry -nln overwritetest test.shp -lco overwrite=yes
Warning 1: Layer creation options ignored since an existing layer is
         being appended to.
```

Command does work if table in in some other database schema.  The layer creation option has obviously worked before because it appears in an example on the driver manual page https://gdal.org/drivers/vector/pg.html

> In this example we merge tiger line data from two different directories of tiger files into one table. Note that the second invocation uses -append and no OVERWRITE=yes.
> 
> ogr2ogr -f PostgreSQL PG:dbname=warmerda tiger_michigan \
>      -lco OVERWRITE=yes CompleteChain
> ogr2ogr -update -append -f PostgreSQL PG:dbname=warmerda tiger_ohio \
>      CompleteChain

rouault:
I'm not completely sure it is a good idea.
As noted in the code, this makes work again something that was documented in GDAL 1.1.8 and got broken at GDAL 1.3.2 as far as I can see. The alternative would be to change the doc of the PG driver.

### DIFF
--- a/apps/ogr2ogr_lib.cpp
+++ b/apps/ogr2ogr_lib.cpp
@@ -2440,6 +2440,28 @@ GDALDatasetH GDALVectorTranslate( const char *pszDest, GDALDatasetH hDstDS, int 
         }
     }
 
+    // Some syntaxic sugar to make "ogr2ogr [-f PostgreSQL] PG:dbname=.... source [srclayer] -lco OVERWRITE=YES"
+    // work like "ogr2ogr -overwrite PG:dbname=.... source [srclayer]"
+    // The former syntax used to work at GDAL 1.1.8 time when it was documented
+    // in the PG driver, but was broken starting with GDAL 1.3.2
+    // (https://github.com/OSGeo/gdal/commit/29c108a6c9f651dfebae6d1313ba0e707a77c1aa)
+    // This could probably be generalized to other drivers that support the
+    // OVERWRITE layer creation option, but we'd need to make sure that they
+    // just do a DeleteLayer() call. The CARTO driver is an exception regarding that.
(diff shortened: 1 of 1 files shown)


jratike80:
The PG driver doc tries to deal with the other case with other-than-public schema but I do not understand the last part of the documentation. How `-nln` is supposed to be used for overwriting and where is the "below example"?

_SCHEMA: Set name of schema for new table. Using the same layer name in different schemas is supported, but not in the public schema and others. Note that using the -overwrite option of ogr2ogr and -lco SCHEMA= option at the same time will not work, as the ogr2ogr utility will not understand that the existing layer must be destroyed in the specified schema. Use the -nln option of ogr2ogr instead, or better the active_schema connection string. See below example._

rouault:
> How `-nln` is supposed to be used for overwriting and where is the "below example"?

I guess this must be the one at bottom of https://gdal.org/drivers/vector/pg_advanced.html#vector-pg-advanced

PR #5648, opened 2022-04-23
GDALWarp utility improvement 

desertstsung:
Mainly coding style change;
and I also change CONTRIBUTING.md:81 due to my personal failure when adding fork using HTTPS.

### DIFF
--- a/apps/gdalwarp_bin.cpp
+++ b/apps/gdalwarp_bin.cpp
@@ -46,18 +46,18 @@ CPL_CVSID("$Id$")
 
 static int GDALExit( int nCode )
 {
-  const char  *pszDebug = CPLGetConfigOption("CPL_DEBUG",nullptr);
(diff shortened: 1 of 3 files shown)


apps/gdalwarp_bin.cpp
```diff
@@ -361,8 +375,6 @@ MAIN_START(argc, argv)
 
     GDALDumpOpenDatasets( stderr );
 
-    GDALDestroyDriverManager();
```
rouault, line 364: why removing this line ?

desertstsung:
I check out prototype of OGRCleanupAll from ogr/ogrsf_frmts/generic/ogrsfdriverregistrar.cpp:86 and found GDALDestroyDriverManager will be called when calling OGRCleanupAll.
In apps/gdalwarp_bin.cpp:main, OGRCleanupAll is called after GDALDestroyDriverManager, so I remove GDALDestroyDriverManager.

PR #5654, opened 2022-04-24
NITF: RPF.toc Skip plausibility check for Overviews and Legends; disable some checks for polar zones

adegwerthAirbus:
## What does this PR do?
Skips the plausibility check for Legends and Overlays during reading of the RPF-TOC.

rpftocdataset.cpp: 
For polar zone datasets use always the values from the dataset and not the values from the .toc file, because they do not always match.

rpftocfile.cpp:
In RPFTOCReadFromBuffer skip the plausibility checks for Legends and Overviews, because Legends and Overlays sometimes have bounding values that are equal (e.g. NW==SE).

Sample BoundingRect for Polar zone:
image

Sample BoundingRect for Overview:
image

none

## Tasklist

 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:


rouault (commented):
It seems to me that only the toc->entries[i].seLong < toc->entries[i].nwLong ||  toc->entries[i].nwLat < toc->entries[i].seLat check should be disabled for zone D and Z. The checks on nHorizFrames and nVertFrames should at the very least kept to avoid a potential integer overflow in the test static_cast<int>(toc->entries[i].nHorizFrames * toc->entries[i].nVertFrames) > atoi(CPLGetConfigOption("RPFTOC_MAX_FRAME_COUNT", "1000000")

adegwerthAirbus:
You're right. I will revise this and try to split this big "if" into 2 "if"s to make it more understandable.

PR #5678, opened 2022-05-05
rfc31_ogr_64.rst: correct option name to ADJUST_TYPE

spencermathews:
**RFC 31** makes reference to a `DETECT_TYPE` open option, but this option does not appear to exist. The name as implemented by the Shapefile driver is actually `ADJUST_TYPE`.

Reference:
* [ogrshapelayer.cpp](https://github.com/OSGeo/gdal/blob/08713ef0092be807434a945897682db4a92a07e6/ogr/ogrsf_frmts/shape/ogrshapelayer.cpp)
* [ogrshapedriver.cpp](https://github.com/OSGeo/gdal/blob/08713ef0092be807434a945897682db4a92a07e6/ogr/ogrsf_frmts/shape/ogrshapedriver.cpp)

### DIFF
--- a/doc/source/development/rfc/rfc31_ogr_64.rst
+++ b/doc/source/development/rfc/rfc31_ogr_64.rst
@@ -220,7 +220,7 @@ list of changes is :
    OFTInteger64. Above they will be treated as OFTReal. In previous GDAL
    versions, Integer fields were created with a default with of 10, and
    thus will be now read as OFTInteger64. An open option,
-   DETECT_TYPE=YES, can be specified so as OGR does a full scan of the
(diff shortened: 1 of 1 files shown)


jratike80:
I think that this small edit is OK in this case but generally a RFC document is an accepted plan and it is normal that the real implementation has some differences. Unfortunately it is also common that the RCF must be used as a user manual because that's all we have.

spencermathews:
That's understandable. Perhaps an addendum could be added to the Implementation section of accepted RFC documents to point out what is changed or missing in the real implementation?

Issue #5618, opened 2022-04-16
CI Conda builds fail


rouault:
The "Conda ubuntu-latest" CI job  fails since https://github.com/OSGeo/gdal/runs/5976324461?check_suite_focus=true with:
```
#
# To activate this environment, use
#
#     $ conda activate test
#
# To deactivate an active environment, use
#
#     $ conda deactivate

gdalinfo: error while loading shared libraries: libtiledb.so.2.2: cannot open shared object file: No such file or directory
```

Looking a bit above in the log, one can guess that https://github.com/OSGeo/gdal/blob/15e22bffe5ba8241f705d47bb95a741d938b4c33/ci/travis/conda/compile.sh#L22, instead of installing the gdal package we just built (and from a cursory review of the logs, the build of this package looks to be OK), installs an older gdal-3.4.1 from conda-forge (hence the likely mix of dependencies with tiledb, but the tiledb error looks more like a symptom that the root cause)

The latest successful build was just one day before ( https://github.com/OSGeo/gdal/runs/5963395509?check_suite_focus=true ) and did install the custom GDAL package. Nothing has changed in GDAL itself between those 2 points.
@hobu any idea ?

Calsia:
`ImportError: libtiledb.so.2.2: cannot open shared object file: No such file or directory`, i meet this.
but i found there is libtiledb.so.2.8 in my conda env path. i guess it maybe the points. but i don't know what should i do

rouault:
Conda builds work again (without any change on our side): https://github.com/OSGeo/gdal/actions/runs/2476351036
Closing

hjxwhy:
> `ImportError: libtiledb.so.2.2: cannot open shared object file: No such file or directory`, i meet this. but i found there is libtiledb.so.2.8 in my conda env path. i guess it maybe the points. but i don't know what should i do

Hi, Have you solved this problem? I also meet this.

rouault:
> Have you solved this problem?

the problem has solved itself without doing anything. Perhaps update / clean / recreate your Conda env

PR #5682, opened 2022-05-06
Fix Conda builds (fixes #5618)

rouault:


rouault:
no longer needed. Conda builds work again: https://github.com/OSGeo/gdal/actions/runs/2476351036

PR #5692, opened 2022-05-08
Add read/write raster JPEGXL driver (using libjxl)

rouault:
The JPEG-XL format is supported for reading, and batch writing (CreateCopy()), but  not update in place.
    
The driver supports reading and writing:
- georeferencing: encoded as a GeoJP2 UUID box within a JUMBF box.
- XMP in the xml:XMP metadata domain
- EXIF in the EXIF metadata domain
- color profile in the COLOR_PROFILE metadata domain.


normanb:
@rouault are we enforcing a convention for jpeg-xl or is there a reference here for defining geospatial metadata in JPEG-XL?

* georeferencing: encoded as a GeoJP2 UUID box within a JUMBF box.

I don't object to it, just wondering if we are enforcing a convention.

rouault:
> or is there a reference here for defining geospatial metadata in JPEG-XL?

Excellent question. No, to my knowledge, geospatial in JPEG-XL is mostly a unexplored area. I was a bit hesitant what to do. This solution seemed the less invasive to me, as it uses a UUID box inside a JUMBF box, which is something documented in the JUMBF specification (I mean using a "generic" UUID box, not specifically the GeoJP2 one). Interestingly, I found that exiftool can decode such files, as it uses recursive exploration of JPEG2000-like boxes (like JUMBF) and knows GeoJP2 boxes.

```
$ ~/exiftool/exiftool byte.jxl
ExifTool Version Number         : 12.41
File Name                       : byte.jxl
Directory                       : .
File Size                       : 719 bytes
File Modification Date/Time     : 2022:05:08 21:14:15+02:00
File Access Date/Time           : 2022:05:08 21:14:17+02:00
File Inode Change Date/Time     : 2022:05:12 00:25:15+02:00
File Permissions                : -rw-rw-r--
File Type                       : JXL
File Type Extension             : jxl
MIME Type                       : image/jxl
Major Brand                     : JPEG XL Image (.JXL)
Minor Version                   : 0.0.0
Compatible Brands               : jxl
JUMD Type                       : (uuid)-0011-0010-800000aa00389b71
JUMD Label                      : GeoJP2 box
Exif Byte Order                 : Little-endian (Intel, II)
Image Width                     : 1
Image Height                    : 1
Bits Per Sample                 : 8
Compression                     : Uncompressed
Photometric Interpretation      : BlackIsZero
Strip Offsets                   : 136
Samples Per Pixel               : 1
Rows Per Strip                  : 1
Strip Byte Counts               : 1
Planar Configuration            : Chunky
Pixel Scale                     : 60 60 0
Model Tie Point                 : 0 0 0 440720 3751320 0
Geo Tiff Version                : 1.1.0
GT Model Type                   : Projected
GT Raster Type                  : Pixel Is Area
GT Citation                     : NAD27 / UTM zone 11N
Geog Citation                   : NAD27
Geog Angular Units              : Angular Degree
Projected CS Type               : NAD27 UTM zone 11N
Proj Linear Units               : Linear Meter
Image Size                      : 1x1
Megapixels                      : 0.000001
```


Another solution would have been to use GMLJP2, but that requires using "asoc" boxes, the GML stuff references JPEG2000, etc, and in the end GMLJP2 is a super verbose standard with little practical advantages. That said, the reader side of the driver should be ready to such beasts as it used the generic JPEG2000 classes to read the georeferencing.

normanb:
@rouault Thanks for the detail, I was reading through the JUMBF notes as well, great that the exiftool reads this. A UUID box containing a degenerative geotiff (as per geojp2) works well and is easy to read / write by other solutions.

PR #5702, opened 2022-05-10
ESAT: Fix a case of freeing memory in use.

augu8346:


frmts/envisat/EnvisatFile.c
```diff
@@ -358,8 +358,6 @@ int EnvisatFile_Open( EnvisatFile **self_ptr,
         return FAILURE;
     }
 
-    CPLFree( sph_data );
```
augu8346, line 361: _ds_data_ is the strstr result from _sph_data_. Freeing _sph_data prematurely means _ds_data_ is not valid no more. This is being used later on. Bug verified using Application Verifier in windows.

PR #5710, opened 2022-05-12
KEA: add 64bit image types

gillins:
This PR adds support for the `GDT_Int64` and `GDT_UInt64` image types. The underlying `libkea` library always had support for 64 bit types but with this PR these types can now be passed through to it.

Any chance a 64 bit int type may be added to the RasterAttributeTable in future? This is another area where `libkea` does have support and would be nice to access from GDAL...

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:

### DIFF
--- a/frmts/kea/keadataset.cpp
+++ b/frmts/kea/keadataset.cpp
@@ -59,11 +59,17 @@ GDALDataType KEA_to_GDAL_Type( kealib::KEADataType ekeaType )
         case kealib::kea_32int:
             egdalType = GDT_Int32;
             break;
+        case kealib::kea_64int:
+            egdalType = GDT_Int64;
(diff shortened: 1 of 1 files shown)


rouault:
> Any chance a 64 bit int type may be added to the RasterAttributeTable in future?

PR welcome as one says

gillins:
>>    Any chance a 64 bit int type may be added to the RasterAttributeTable in future?

> PR welcome as one says

Ok we'll see if we can come up with something... cc: @neilflood @petescarth

PR #5720, opened 2022-05-12
[Backport 3.5] vsicurl: add NEGOTIATE to proxy auth methods

t0b3:
**vsicurl: add NEGOTIATE to proxy auth methods**

see https://github.com/curl/curl/blob/68784d73c3f8bc0d14502b4e1e930b1741c41699/docs/cmdline-opts/proxy-negotiate.d

deprecate CURLAUTH_GSSNEGOTIATE
see https://github.com/curl/curl/commit/e38ba430142153e95e2b0abe08a31d626a5c5864


t0b3:
@rouault can we add this to 3.5? would be very useful for some corp network setups.

davidoesch:
would be really helpful for users behind proxies

bernspi:
must have for corporate users behind proxies

Issue #5721, opened 2022-05-13
autotest failure with hana driver


rouault:
@mrylov  Can you promptly address this ?

https://github.com/OSGeo/gdal/runs/6421890811?check_suite_focus=true

```
______________________ ERROR at setup of test_ogr_hana_1 _______________________

    @pytest.fixture(scope="module", autouse=True)
    def setup_driver():
        driver = ogr.GetDriverByName('HANA')
        if driver is None:
            pytest.skip("HANA driver not available", allow_module_level=True)
    
        conn = create_connection()
    
>       uid = execute_sql_scalar(conn, "SELECT REPLACE(CURRENT_UTCDATE, '-', '') || '_' || BINTOHEX(SYSUUID) FROM DUMMY;")

/build/autotest/ogr/ogr_hana.py:57: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

conn = <hdbcli.dbapi.Connection object at 0x7f7155e33d00>
sql = "SELECT REPLACE(CURRENT_UTCDATE, '-', '') || '_' || BINTOHEX(SYSUUID) FROM DUMMY;"

    def execute_sql_scalar(conn, sql):
        cursor = conn.cursor()
        assert cursor
>       cursor.execute(sql)
E       hdbcli.dbapi.Error: (414, 'user is forced to change password: alter password required for user GDALCI')
```

mrylov:
Thank you @rouault. The fix is in #5722.

PR #5722, opened 2022-05-13
Fix issue #5721

mrylov:
This change fixes autotest failure in HANA driver #5721.


rouault:
Does this password need regular refresh ?

mrylov:
I hope not. We disabled that.

mrylov:
@rouault, I've added another related change.

PR #5725, opened 2022-05-13
gcore/mdreader/reader_pleiades.cpp : read FIRST_COL for RPC (fixes #5716)

### LINKED ISSUE #5716: DIMAPV3 (PNEO) RPC pixel convention
pleiades reader has been updated to support PNEO products (#5090). In DIMAP V2 Pléiades pixel convention is assumed to be [[1,1] at the center of first pixel ](https://github.com/OSGeo/gdal/blob/master/gcore/mdreader/reader_pleiades.cpp#L394). ADS Customer Technical Support just confirm that PNEO convention is [0,0] at center of first pixel (which is coherent with RPC validity domain in [example](https://github.com/OSGeo/gdal/blob/master/autotest/gdrivers/data/jpeg2000/RPC_md_pneo.XML#L32)). Thus i suspect one pixel shift in RPC handling.

## Operating system

Ubuntu 20.04 64 bit

3.0

Lunyxis:
Since DIMAP v3 (PNEO),  center of upper left pixel is 0,0. 
To be robust to further modification in this value, the value of the "center of upper left pixel" is loaded from the "FIRST_COL" in "ImagetoGround_Validity_Domain".


rouault (commented):
Perhaps unit test need changes / addition to test this ?

Lunyxis:
Hmmm, as the implementation has been done to be sure that no failure can occurs if the targeted tag does not exists, I do not know what to test.
For the moment, we only test in "_test_jp2metadata" that the right metadata domain exists, and that mandatory keys in the metadata "IMAGERY" and "RPC" are present.

Lunyxis:
I'm open to suggestions.

rouault:
> and that mandatory keys in the metadata "IMAGERY" and "RPC" are present.

the values of the metadata keys for RPC could be tested

Lunyxis:
test added on some values. 
In our PNEO case, the values that change are "SAMP_OFF" and "LINE_OFF"

gcore/mdreader/reader_pleiades.cpp
```diff
@@ -345,6 +345,24 @@ char** GDALMDReaderPleiades::LoadRPCXmlFile()
         return nullptr;
     }
 
+    // search Image to Ground Validity (since DIMAP v3)
+    CPLXMLNode* pValidityNode = CPLSearchXMLNode(pNode, "=ImagetoGround_Validity_Domain");
+
+    double firstCol = 1.0;
+    if( pValidityNode != nullptr )
+    {
+        char** papszValidity = ReadXMLToList(pValidityNode->psChild, m_papszValidity);
```
rouault, line 354:
```suggestion
        char** papszValidity = ReadXMLToList(pValidityNode->psChild, nullptr);
```

jguinet:
Sorry for my late comment, I think it is not a good solution to find  fisrt col using "ImagetoGround_Validity_Domain", because this validity domain is the sampling ROI used for RPC computation, not the first pixel convention.

Lunyxis:
And you do not think that it is linked ? I'm not expert in geometry,  but the bayesian in me finds this coincidence curious :)

Issue #5718, opened 2022-05-12
Security issue in flatgeobuf driver in 3.6.0dev


rouault:
See https://github.com/OSGeo/gdal/pull/5693#issuecomment-1125472576

rouault:
> but meanwhile, do you think this is serious enough to warrant that I should revert the upgrade?

no, this is OK for master. But if within let's say 2 weeks this isn't fixed, a revert would be indeed appropriate.

PR #5726, opened 2022-05-13
FlatGeobuf: Upgrade flatbuffers again

bjornharrtell:
Intends to fix https://github.com/OSGeo/gdal/issues/5718 which has been fixed upstream by https://github.com/google/flatbuffers/pull/7304.


rouault:
ossfuzz has verified the fix

Issue #5659, opened 2022-04-30
Problem building on macOS when using PDFium


leo-costa:
GDAL 3.5.0 from: `b3784f`

The linker fails on creating libgdal.dylib saying `ld: library not found for -llcms2`.

```
  cmake \
      -D CMAKE_BUILD_TYPE=Release \
      -D CMAKE_INSTALL_PREFIX="out/gdal" \
      -D BUILD_PYTHON_BINDINGS=OFF \
      -D BUILD_JAVA_BINDINGS=OFF \
      -D BUILD_CSHARP_BINDINGS=OFF \
      -D BUILD_TESTING=OFF \
      -D GDAL_USE_INTERNAL_LIBS=ON \
      -D GDAL_USE_CURL=OFF \
      -D GDAL_USE_EXPAT=OFF \
      -D GDAL_USE_ICONV=OFF \
      -D GDAL_USE_LIBLZMA=OFF \
      -D GDAL_USE_LIBXML2=OFF \
      -D GDAL_USE_OPENJPEG=OFF \
      -D GDAL_USE_PCRE=OFF \
      -D GDAL_USE_PCRE2=OFF \
      -D GDAL_USE_WEBP=OFF \
      -D GDAL_USE_ZSTD=OFF \
      -D GDAL_USE_PDFIUM=ON \
      -D PDFIUM_INCLUDE_DIR=`pwd`/pdfium_build_gdal_3_5/install/include/pdfium \
      -D PDFIUM_LIBRARY=`pwd`/pdfium_build_gdal_3_5/install/lib/libpdfium.a \
      -D SQLite3_INCLUDE_DIR=`pwd`/out/sqlite/include \
      -D SQLite3_LIBRARY=`pwd`/out/sqlite/lib/libsqlite3.a \
      -D PROJ_INCLUDE_DIR=`pwd`/out/proj/include \
      -D PROJ_LIBRARY=`pwd`/out/proj/lib/libproj.a \
      -G Ninja -S . -B build && \
      cmake --build build && cmake --install build
```

Just to preface the problem, these are the knowns:
 - Everything compiles successfully when `GDAL_USE_PDFIUM=OFF`
 - Have used different generator, same problem.
 - Successfully build PDFium (GDAL modded version) for macOS [from this repo](https://github.com/rouault/pdfium_build_gdal_3_5).
 - Successfully Build SQL from almagamation file.
 - Successfully Build PROJ version 6.3.2.
[...]

PR #5727, opened 2022-05-14
CMake: fix linking with Pdfium on Mac (fixs #5659)

rouault:


rouault:
@leo-costa little nudge: can you test this pull request ?

PR #5751, opened 2022-05-18
Add a OGRLayer::AlterGeomFieldDefn() method

rouault:
to alter properties of a geometry field: name, geometry type, nullable flag and SRS.
Similarly to the existing AlterFieldDefn() for regular fields.
Actual capabilities depend on drivers
Implement in MEM, Shape, GPKG and PostGIS drivers


ogr/ogrsf_frmts/mem/ogrmemlayer.cpp
```diff
@@ -819,6 +820,74 @@ OGRErr OGRMemLayer::AlterFieldDefn( int iField, OGRFieldDefn *poNewFieldDefn,
...
+
+    auto poFieldDefn = m_poFeatureDefn->GetGeomFieldDefn(iGeomField);
+
+    if( nFlagsIn & ALTER_GEOM_FIELD_DEFN_NAME_FLAG )
+        poFieldDefn->SetName(poNewGeomFieldDefn->GetNameRef());
+    if( nFlagsIn & ALTER_GEOM_FIELD_DEFN_TYPE_FLAG )
+    {
+        if( poNewGeomFieldDefn->GetType() == wkbNone )
+            return OGRERR_FAILURE;
+        poFieldDefn->SetType(poNewGeomFieldDefn->GetType());
```
nyalldawson, line 848: I'm guessing it's purely operator error if there's existing features with incompatible geometries compared to the new type, right?

nyalldawson: Ah, just saw the dox note below -- :+1:

ogr/ogrsf_frmts/ogrsf_frmts.dox
```diff
@@ -2392,6 +2398,75 @@ definition must be taken into account.
...
+You must use this to alter the definition of an existing geometry field of a real layer.
+Internally the OGRFeatureDefn for the layer will be updated
+to reflect the altered field.  Applications should never modify the OGRFeatureDefn
+used by a layer directly.
+
+Note that altering the SRS does *not* cause coordinate reprojection to occur:
+this is simply a modification of the layer metadata (correcting a wrong SRS
+definition)
+
+This method should not be called while there are feature objects in existence that
```
nyalldawson, line 2415: This caveat makes the above note regarding reprojection redundant, but I'd still err for caution and leave it. For the same reason, I'm wondering if a note along the lines of "no modification to existing geometries will ever be performed, so this method cannot be used to eg promote single part geometries to their multipart equivalents".

nyalldawson:
Looks good to me! One thing I'd potentially like to see is the ability to query whether a particular driver has "in theory" support for the various operations (rename field, change type, change srs, epoch). This would allow clients to determine what operations to expose to users in advance, without giving them the impression that an operation is supported until the try to apply it.  (This would also avoid the need to open an editable layer just to test the AlterGeomFieldDefn capability and decide what actions to expose to a user in a GUI)

ogr/ogrsf_frmts/ogrsf_frmts.dox
rouault: good idea. added

rouault:
>  One thing I'd potentially like to see is the ability to query whether a particular driver has "in theory" support

see last commit: "Add a GDAL_DMD_ALTER_GEOM_FIELD_DEFN_FLAGS driver metadata item"

nyalldawson (approved):
Thanks Even!

PR #5758, opened 2022-05-19
LVBAG: scan only .xml files when passed a directory

rouault:
... and when reading from /vsi network file systems, give up after 10
probed files, if none was detected as a LVBAG file.

Otherwise this could be really sluggish when GDAL tries to open a remote
directory with lots of files, and that it is not a LVBAG dataset

CC @yorickdewid Does that sound good to you ?


yorickdewid:
Seems fair

PR #5804, opened 2022-05-28
RMF: Fix FlushCache with NoData block

drons:


frmts/rmf/rmfdataset.cpp
```diff
@@ -1052,19 +1052,21 @@ void RMFDataset::FlushCache(bool bAtClosing)
...
+    if( bAtClosing && eRMFType == RMFT_MTW )
     {
         GDALRasterBand *poBand = GetRasterBand(1);
 
         if( poBand )
         {
             poBand->ComputeRasterMinMax( FALSE, sHeader.adfElevMinMax );
+            // ComputeRasterMinMax can setup error in case of dataset full
+            // from NoData values, but it  makes no sense here.
+            CPLErrorReset();
```
rouault, line 1064: to get silent errors and avoid changing the error set, you could do:
```suggestion
            // ComputeRasterMinMax can setup error in case of dataset full
            // from NoData values, but it  makes no sense here.
            CPLErrorStateBackuper oErrorStateBackuper;
            CPLErrorHandlerPusher oErrorHandler(CPLQuietErrorHandler);
            poBand->ComputeRasterMinMax( FALSE, sHeader.adfElevMinMax );
```

But in the case of a dataset full of nodata values, adfElevMinMax[] will be {0, 0}. Is that what you want to write in the header ? Perhaps you should override them with the nodata value ? (but not sure if it is legal to write it in adfElevMinMax[] though ?). No strong opinion here, just wanted to raise that point

drons: I think setting the min and max values to 0 is a good choice for an empty dataset. In most cases, the NoData value is INT_MIN for MTW datasets. If we set min-max to INT_MIN, client applications will get confused.

drons:
Thanks!

PR #5812, opened 2022-05-29
SAFE: for non-SLC products, do not report subdatasets that can't be opened

rouault:
Fixes issue reported in https://lists.osgeo.org/pipermail/gdal-dev/2022-May/055882.html

With that fix, the only reported subdatasets now are:
```
  SUBDATASET_1_NAME=SENTINEL1_CALIB:UNCALIB:S1A_IW_GRDH_1SDV_20220529T045822_20220529T045847_043421_052F5A_01B5.SAFE/manifest.safe:IW_VH:AMPLITUDE
  SUBDATASET_1_DESC=Single band with IW swath and VH polarization and UNCALIB calibration
  SUBDATASET_2_NAME=SENTINEL1_CALIB:UNCALIB:S1A_IW_GRDH_1SDV_20220529T045822_20220529T045847_043421_052F5A_01B5.SAFE/manifest.safe:IW_VV:AMPLITUDE
  SUBDATASET_2_DESC=Single band with IW swath and VV polarization and UNCALIB calibration
  SUBDATASET_3_NAME=SENTINEL1_CALIB:UNCALIB:S1A_IW_GRDH_1SDV_20220529T045822_20220529T045847_043421_052F5A_01B5.SAFE/manifest.safe:IW:AMPLITUDE
  SUBDATASET_3_DESC=IW swath with all polarizations (VH+VV) as bands and UNCALIB calibration
```

### DIFF
--- a/frmts/safe/safedataset.cpp
+++ b/frmts/safe/safedataset.cpp
@@ -1470,26 +1470,27 @@ GDALDataset *SAFEDataset::Open( GDALOpenInfo * poOpenInfo )
                     osSubDS2 += pol;
                     //Create single band or multiband complex SubDataset
                     int i = 0;
-                    for( i = 0; i < 3; i++ )
-                    {
-                      CPLString osCalibTemp = aosCalibrationValues[i];
-                      SAFEDataset::AddSubDataset(
-                          poDS.get(), iSubDS + i,
-                          CPLSPrintf("SENTINEL1_CALIB:%s:%s:%s_%s:%s",
-                                     osCalibTemp.c_str(), osMDFilename.c_str(),
-                                     osSubDS1.c_str(), pol.c_str(),
-                                     aosDataUnitValues[2].c_str()),
-                          CPLSPrintf("Single band with %s swath and %s "
-                                     "polarization and %s calibration",
-                                     osSubDS1.c_str(), pol.c_str(),
-                                     osCalibTemp.c_str()));
-                    }
-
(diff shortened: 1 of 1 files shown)


sgofferj:
@rouault Thanks for the fast reaction! However, I'd like to point out that there **is** calibration data also in the GRDH products, at least, corresponding XML files are present and SNAP does offer calibrated intensity bands also from GRDH products.

Here's the files which e.g. are in the annotation/calibration folder of S1A_IW_GRDH_1SDV_20220529T045822_20220529T045847_043421_052F5A_01B5.SAFE (UUID: c2380d61-207e-46c6-9ade-5cb1a0abdcd2)
[calibration.zip](https://github.com/OSGeo/gdal/files/8796113/calibration.zip)

rouault:
> I'd like to point out that there **is** calibration data also in the GRDH products, at least, corresponding XML files are present and SNAP does offer calibrated intensity bands also from GRDH products

maybe. I'm not familiar with calibration process. If the calibration process is the same as SLC, it might be relatively straightforward to adapt the driver re-using the SLC logic, otherwise it would need more involved changes

sgofferj:
OK. Unfortunately, that's "a little" above my skills with regards to C++ coding. How would I proceed? Would I open an issue with a feature request/suggestion?

rouault:
> Would I open an issue with a feature request/suggestion?

you may (but there's no guarantee someone will tackle it)

PR #5824, opened 2022-05-31
MRF QB3

lucianpls:
## What does this PR do?

Add [QB3](https://github.com/lucianpls/QB3) compression to MRF

QB3 is a new raster specific lossless compression for integer types, lightweight, very fast and very good.

## Tasklist

 - [x] ADD YOUR TASKS HERE
 - [x] Add test case(s)
 - [x] Add documentation
 - [x] Review
 - [x] Adjust for comments

## Environment

* OS:
* Compiler:


cmake/modules/packages/FindQB3.cmake
```diff
@@ -0,0 +1,60 @@
...
+
+This module will set the following variables if found:
+
+``QB3_INCLUDE_DIR`` - where to find QB3 public headers
+``QB3_LIBRARY`` - the library to link against
+``QB3_FOUND`` - TRUE if found
+
+#]=======================================================================]
+
+find_package(libQB3 CONFIG)
```
dg0yt, line 30: If upstream provides a CMake config file, that package name and its targets shall be used directly, and this `FindQB3.cmake` is obsolete.

dg0yt (changes requested):

rouault:
- You could likely drop autoconf & nmake build support as they will be soon removed from master
- Is there some benchmarking of https://github.com/lucianpls/QB3 in terms of speed, compression ratio, speed/compression_ratio compared to other lossless image codecs (PNG, lossless WEBP, etc.), or maybe generic purpose methods such as LZ4

cmake/modules/packages/FindQB3.cmake
lucianpls: I think I got it.

lucianpls:
@rouault

That's why I added timing debug statements to MRF, to make it easier to compare with other formats. :)

Most formats read fast enough, the bigger differences are in resources used during compression (CPU and memory).
As far as I can tell QB3 is faster than anything smaller and smaller than anything faster, so it has it's role. It works great for MRF pass-through caching for example, or for transient compression during transmission.
For byte RGB I see about 250MB/sec both for compression and decompression (compression is marginally faster). In general it is slightly smaller than PNG, but likely to be larger than WebP and others that take a while to compressed (JP2, JPEGXL...). It is faster and sometimes smaller than ZSTD, even at level 1, faster and smaller than LERC. For larger sized integers it gets progressively better and faster, since it is data type aware and not dealing with just a string of bytes.
In addition, algorithm complexity is very small, it is plain, portable C++ and has almost no memory overhead (uses a few KB worth of tables for faster encoding).

frmts/mrf/mrf_util.cpp
```diff
@@ -457,8 +468,8 @@ void XMLSetAttributeVal(CPLXMLNode *parent, const char*pszName, std::vector<doub
         if (val != values[i])
             single_val = false;
         value.append(PrintDouble(values[i]) + " ");
+        value.resize(value.size() - 1); // Cut the last space
```
rouault, line 471: this change is suspicious

lucianpls:
@dg0yt 
QB3 is 2D raster specific, comparing it with generic lossless compressions doesn't make sense.

rouault:
> QB3 is 2D raster specific, comparing it with generic lossless compressions doesn't make sense.

some minimum benchmark with figures comparing QB3 performance to other lossless codecs would still be appreciated.
You mention that compression is slightly faster than decompression: that's quite unusual, isn't it ?

lucianpls:
@rouault 
I'm working on a paper on QB3. 

Indeed, it is usually faster to decompress than to compress. With QB3 it's the other way around, not by much. This is because it is very fast to start with, there are only a few CPU instructions per data value, every change has a measurable impact. The QB3 encoder has very few choices to make, the encoding speed doesn't depend much on the input data.
I think the main reason decompression is slightly slower because it is a long, uninterrupted dependency chain, every value depends on the immediately preceding one, which isn't the case during compression, so the CPU stalls more often.
Under 12 bits per value tables (static) are used for both encoding and decoding, but the decoding tables are larger due to variable symbol size. Above 12bits per value the speed flips the other way (faster decoding), there is more code that can be executed asynchronously. Hand assembly or vectorization is a possibility.
Lastly, the encoder assumes the output buffer to be large enough to handle the worst case encoded size, which is bound. The decoder is more careful, it checks all the time that it doesn't read past the end of the input, especially if the stream is corrupt or malicious.
___
I really don't want to start or get pulled into a flame war, benchmark is always a hot subject, but here you go:
For a sample of values, I'm working with Mars HiRISE right now, which starts as JP2 (almost)lossless, 10bit single band.
The specific sample is ESP_011261_1960_COLOR.JP2, which is 35326x111350, a slanted swath with surrounding black area.
The source is 1'571'162'296 bytes in JP2. Takes a long while to read with OpenJPEG, (14m real, 22m CPU).

QB3 size is: 1'778'298'865 (full resolution only). So it's not as small as JP2 (yet it is fully lossless).
Compression time is: 11.535s, Decompression time: 13.687s.

The closest algorithm in MRF is ZSTD at the lowest compression level (fastest).
Size: 1'823'961'772
Compression time:14.280s, Decompression time: 10.579s.

MRF w/LERC
Size: 2'059'365'440
Compression time: 17.210s, Decompression time: 8.473s

_____
Another sample, the Kodak 8bit RGB true color images, from http://r0k.us/graphics/kodak/
Size comparison, sorted by input image size:
image

PNG folder size (du -ks): 15076

8bit RGB isn't exactly the QB3 strength, but it's still good. In bulk, figures for all 24 images, converted with gdal_translate:
QB3:
Conversion time (includes reading PNG): .56s, folder size (du -ks): 14208
Compression to QB3 takes about 7ms per image

That includes the MRF tiling and multiple file (block) penalty, I tried MRF w/PNG encoding (Z level 8):
time: 5.5s, folder size: 15820
Compression to PNG (in MRF) takes about 200ms per image (standard zlib)

Webp lossless: time 22.29s, folder size: 11188

MRF-ZSTD Level 1:
time: .55s, folder size: 17604

MRF-LERC: (Huffman for 8bit data, not LERC algorithm)
time: .76s, folder size: 19016

rouault:
@lucianpls  Thanks for the data points

autotest/gdrivers/mrf.py would need to be extended to test this new codec.

lucianpls:
That means adding libQB3 build from github during GDAL CI, is that OK?
Otherwise the tests won't be exercised.

rouault:
> That means adding libQB3 build from github during GDAL CI, is that OK?
> Otherwise the tests won't be exercised.

Having tests that someone correctly setup can run is a first step.
As far as adding them in CI, you could either add a build step as for https://github.com/OSGeo/gdal/blob/master/.github/workflows/cmake_builds.yml#L99 , or generate a Linux x86_64 prebuild binary like in https://github.com/OSGeo/gdal/blob/master/.github/workflows/cmake_builds.yml#L117

cmake/helpers/CheckDependentLibraries.cmake
```diff
@@ -417,6 +417,8 @@ gdal_internal_library(LERC)
 
 gdal_check_package(BRUNSLI "Enable BRUNSLI for JPEG packing in MRF" CAN_DISABLE RECOMMENDED)
 
+gdal_check_package(libQB3 "Enable QB3 compression in MRF" CAN_DISABLE RECOMMENDED)
```
rouault, line 420: we'd need to have it mentioned in doc/source/build_hints.rst

rouault (commented):
Gave a quick local try.
There's a memory leak on the encoder site:
```
$ valgrind --leak-check=full gdal_translate autotest/gcore/data/byte.tif byte.mrf -co compress=qb3

==1400051== 432 bytes in 1 blocks are definitely lost in loss record 8 of 35
==1400051==    at 0x483BE63: operator new(unsigned long) (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==1400051==    by 0x755BCB9: qb3_create_encoder (in /home/even/install-qb3/lib/libQB3.so)
==1400051==    by 0x4FEDEC7: GDAL_MRF::QB3_Band::Compress(GDAL_MRF::buf_mgr&, GDAL_MRF::buf_mgr&) (QB3_band.cpp:30)
==1400051==    by 0x4FDFF8E: GDAL_MRF::MRFRasterBand::IWriteBlock(int, int, void*) (mrf_band.cpp:1077)
==1400051==    by 0x52D2E1F: Write (gdalrasterblock.cpp:849)
==1400051==    by 0x52D2E1F: GDALRasterBlock::Write() (gdalrasterblock.cpp:835)
==1400051==    by 0x5311927: GDALArrayBandBlockCache::FlushBlock(int, int, int) (gdalarraybandblockcache.cpp:436)
==1400051==    by 0x53119C4: GDALArrayBandBlockCache::FlushCache() (gdalarraybandblockcache.cpp:275)
==1400051==    by 0x52AD565: GDALDataset::FlushCache(bool) (gdaldataset.cpp:446)
==1400051==    by 0x52DA170: GDALPamDataset::FlushCache(bool) (gdalpamdataset.cpp:178)
==1400051==    by 0x539636D: GDALTranslateFlush(void*) (gdal_translate_lib.cpp:507)
==1400051==    by 0x539D1A1: GDALTranslate (gdal_translate_lib.cpp:1243)
==1400051==    by 0x10A99A: main (gdal_translate_bin.cpp:365)
==1400051== 
```

lucianpls:
> Gave a quick local try. There's a memory leak on the encoder site:
> 
> ```
> $ valgrind --leak-check=full gdal_translate autotest/gcore/data/byte.tif byte.mrf -co compress=qb3
> 
> ==1400051== 432 bytes in 1 blocks are definitely lost in loss record 8 of 35
> ==1400051==    at 0x483BE63: operator new(unsigned long) (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
> ==1400051==    by 0x755BCB9: qb3_create_encoder (in /home/even/install-qb3/lib/libQB3.so)
> ==1400051==    by 0x4FEDEC7: GDAL_MRF::QB3_Band::Compress(GDAL_MRF::buf_mgr&, GDAL_MRF::buf_mgr&) (QB3_band.cpp:30)
> ==1400051==    by 0x4FDFF8E: GDAL_MRF::MRFRasterBand::IWriteBlock(int, int, void*) (mrf_band.cpp:1077)
> ==1400051==    by 0x52D2E1F: Write (gdalrasterblock.cpp:849)
> ==1400051==    by 0x52D2E1F: GDALRasterBlock::Write() (gdalrasterblock.cpp:835)
> ==1400051==    by 0x5311927: GDALArrayBandBlockCache::FlushBlock(int, int, int) (gdalarraybandblockcache.cpp:436)
> ==1400051==    by 0x53119C4: GDALArrayBandBlockCache::FlushCache() (gdalarraybandblockcache.cpp:275)
> ==1400051==    by 0x52AD565: GDALDataset::FlushCache(bool) (gdaldataset.cpp:446)
> ==1400051==    by 0x52DA170: GDALPamDataset::FlushCache(bool) (gdalpamdataset.cpp:178)
> ==1400051==    by 0x539636D: GDALTranslateFlush(void*) (gdal_translate_lib.cpp:507)
> ==1400051==    by 0x539D1A1: GDALTranslate (gdal_translate_lib.cpp:1243)
> ==1400051==    by 0x10A99A: main (gdal_translate_bin.cpp:365)
> ==1400051== 
> ```

Ah, yeah, really bad, I kept postponing this issue and then I forgot to finish it.

doc/source/build_hints.rst
```diff
@@ -265,6 +266,17 @@ by the :ref:`raster.marfa` driver.
     Control whether to use BRUNSLI. Defaults to ON when Brunsli is found.
 
 
+QB3
```
rouault, line 269: the library are sorted by alphabetic order

PR #5829, opened 2022-05-31
Arrow: fix extension name key in field metadata

jorisvandenbossche:
## What does this PR do?

The current field metadata key for the extension name is using a double colon, but that should be a single colon, see https://arrow.apache.org/docs/format/Columnar.html#extension-types

Noticed this while testing to read GDAL produced files with pyarrow with extension types registered for those "geoarrow.<geometry_type>" types (https://github.com/jorisvandenbossche/python-geoarrow/)

## Tasklist

 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

Does this require a test case? 
I see that this key is actually checked for when reading, but since it also checks in the metadata for the known geometry columns (and its encoding), this ARROW:extension:name key is not required I suppose.

### DIFF
--- a/autotest/ogr/ogr_arrow.py
+++ b/autotest/ogr/ogr_arrow.py
@@ -367,3 +367,27 @@ def test_ogr_arrow_coordinate_epoch(write_gdal_footer):
     ds = None
 
     gdal.Unlink(outfilename)
+
+
+###############################################################################
+# Test that Arrow extension type is recognized as geometry column
+# if "geo" metadata is absent
+
+
+def test_ogr_arrow_extension_type():
+
+    outfilename = '/vsimem/out.feather'
+    with gdaltest.config_options({'OGR_ARROW_WRITE_GDAL_FOOTER': 'NO',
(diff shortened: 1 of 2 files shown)


rouault:
> Does this require a test case?
> I see that this key is actually checked for when reading, but since it also checks in the metadata for the known geometry columns (and its encoding), this ARROW:extension:name key is not required I suppose.

One way of checking the fix has effect is to add a new test case in autotest/ogr/ogr_arrow.py and generate a GeoArrow file with the following configuration options set:
- OGR_ARROW_WRITE_GDAL_FOOTER=NO
- OGR_ARROW_WRITE_GEO=NO
which will prevent both the "geo" and "gdal:geo" metadata to be written. 
And open it back and check that the geometry column is correctly detected.
I can help writing it if needed

jorisvandenbossche:
Thanks for those pointers. Did a first attempt to add a test based on that (and verified that the test failed before the patch)

PR #5830, opened 2022-06-01
Add RFC 86: Column-oriented read API for vector layers

rouault:
This RFC describes the addition of new methods to the `OGRLayer` class to retrieve batches of features with a column-oriented memory layout, that suits formats that have that organization or downstream consumers that expect data to be presented
in such a way, in particular the [Apache Arrow](https://arrow.apache.org/docs/), [Pandas](https://pandas.pydata.org/) / [GeoPandas](https://geopandas.org/) ecosystem, R spatial packages, and many modern (data analytics focused) databases / engines which are column oriented (eg Snowflake, Google BigQuery, ..)

(Semi-)rendered version of the RFC at: https://github.com/rouault/gdal/blob/rfc_86/doc/source/development/rfc/rfc86_column_oriented_api.rst

### DIFF
--- a/doc/source/development/rfc/rfc86_column_oriented_api.rst
+++ b/doc/source/development/rfc/rfc86_column_oriented_api.rst
@@ -20,7 +20,8 @@ batches of features with a column oriented memory layout, that suits formats tha
 have that organization or downstream consumers that expect data to be presented
 in such a way, in particular the `Apache Arrow <https://arrow.apache.org/docs/>`_,
 `Pandas <https://pandas.pydata.org/>`_ / `GeoPandas <https://geopandas.org/>`_
-ecosystem.
+ecosystem, R spatial packages, and many modern (data analytics focused)
+databases / engines which are column oriented (eg Snowflake, Google BigQuery, ..)
 
 Motivation
 ----------
@@ -157,6 +158,9 @@ The following virtual methods are added to the OGRLayer class:
   and GetNextRecordBatch() is not recommended, as the behaviour will be unspecified
   (but it should not crash).
 
+  Options passed to this method should be identical between calls, and the same as
+  the ones provided to GetRecordBatchSchema().
+
   When this method returns true, and the array is no longer needed, it must
   be released with the following procedure, to take into account that it might
   have been released by other code, as documented in the Arrow C data
@@ -177,15 +181,17 @@ Other remarks
 -------------
 
 Using directly (as a producer or a consumer), the ArrowArray is admitedly not
-trivial, and requires good intimacy with the Arrow C data interface specification,
-to know, in which buffer of an array, data is to be read, which data type void*
-buffers should be cast to, how to use buffers that contain null/not_null information,
-how to use offset buffers for data types of type List, etc.
+trivial, and requires good intimacy with the Arrow C data interface and columnar
+array specifications, to know, in which buffer of an array, data is to be read,
+which data type void* buffers should be cast to, how to use buffers that contain
+null/not_null information, how to use offset buffers for data types of type List, etc.
+
 For the consuming side, the new API will be best used with the (Py)Arrow, Pandas,
 GeoPandas, Numpy libraries which offer easier and safer access to record batches.
 The study of the gdal_array._RecordBatchAsNumpy() method added to the SWIG Python
 bindings can give a good hint of how to use an ArrowArray object, in conjunction
-with the associated ArrowSchema.
+with the associated ArrowSchema. DuckDB is also another example of using the ArrowArray
+inferface: https://github.com/duckdb/duckdb/blob/master/src/common/types/data_chunk.cpp
 
 It is not expected that most drivers will have a dedicated implementation of
 GetNextRecordBatch(). Implementing it requires a non-trivial effort, and
@@ -195,6 +201,11 @@ total time (I/O + shuffling).
 
 Potential future work, no in the scope of this RFC, could be the addition of a
 column-oriented method to write new features, a WriteRecordBatch() method.
+
+The use of the `Arrow C stream interface <https://arrow.apache.org/docs/format/CStreamInterface.html>`_,
+which associates the functionality of GetRecordBatchSchema() and
+GetNextRecordBatch() in a same C structure has been considered, but not kept as
+this interface is currently marked as experimental and not guaranteed to be ABI stable.
 
 Impacted drivers
 ----------------
@@ -249,8 +260,10 @@ dataset with 3.3 millions features, with 13 fields each (2 fields of type Intege
 :ref:`rfc-86-bench-ogr-py`, :ref:`rfc-86-bench-fiona` and :ref:`rfc-86-bench-ogr-cpp`
 have similar functionality: iterating over features with GetNextFeature().
 
+:ref:`rfc-86-bench-pyogrio-raw` does a little more by building Arrow arrays.
+
 :ref:`rfc-86-bench-pyogrio`, :ref:`rfc-86-bench-geopandas` and :ref:`rfc-86-bench-ogr-to-geopandas`
-have similar functionality: building a GeoPandas GeoDataFrame
+have all similar functionality: building a GeoPandas GeoDataFrame
 
 :ref:`rfc-86-bench-ogr-batch-cpp` can be used to measure the raw performance of the
 proposed GetNextRecordBatch() API.
@@ -263,6 +276,7 @@ proposed GetNextRecordBatch() API.
 bench_ogr.cpp                             6.6
 bench_ogr.py                              71
 bench_fiona.py                            68
+bench_pyogrio_raw.py                      40
 bench_pyogrio.py                          108
 bench_geopandas.py                        232
 bench_ogr_batch.cpp (driver impl.)        4.5
@@ -284,6 +298,7 @@ using GetNextFeature() underneath, is used.
 bench_ogr.cpp                             6.4
 bench_ogr.py                              72
 bench_fiona.py                            70
+bench_pyogrio_raw.py                      46
 bench_pyogrio.py                          115
 bench_geopandas.py                        228
 bench_ogr_batch.cpp (driver impl.)        1.6
@@ -302,6 +317,7 @@ Note: Fiona slightly modified to accept Parquet driver as a recognized one.
 bench_ogr.cpp                             17.7
 bench_ogr.py                              81
 bench_fiona.py                            81
+bench_pyogrio_raw.py                      58
 bench_pyogrio.py                          120
 bench_geopandas.py                        258
 bench_ogr_batch.cpp (driver impl.)        N/A
@@ -477,13 +493,29 @@ expose them as GeoJSON features holded by a Python dictionary.
         for f in features:
             pass
 
+.. _rfc-86-bench-pyogrio-raw:
+
+bench_pyogrio_raw.py
+++++++++++++++++++++
+
+Use of the pyogrio Python library which uses the OGR C GetNextFeature() underneath to
+expose a layer as a set of Arrow arrays.
+
+.. code-block:: python
+
+    import sys
+    from pyogrio.raw import read
+
+    read(sys.argv[1])
+
+
 .. _rfc-86-bench-pyogrio:
 
 bench_pyogrio.py
 ++++++++++++++++
 
 Use of the pyogrio Python library which uses the OGR C GetNextFeature() underneath to
-expose a layer as GeoPandas GeoDataFrame.
+expose a layer as GeoPandas GeoDataFrame (which involves parsing WKB as GEOS objects)
 
 .. code-block:: python
 
--- a/doc/source/development/rfc/rfc86_column_oriented_api.rst
+++ b/doc/source/development/rfc/rfc86_column_oriented_api.rst
@@ -1,0 +1,600 @@
+.. _rfc-86:
+
+=============================================================
+RFC 86: Column-oriented API for vector layers
+=============================================================
+
+============== =============================================
+Author:        Even Rouault
+Contact:       even.rouault @ spatialys.com
+Started:       2022-May-24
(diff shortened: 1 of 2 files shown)


doc/source/development/rfc/rfc86_column_oriented_api.rst
```diff
@@ -0,0 +1,679 @@
...
+with the associated ArrowSchema. DuckDB is also another example of using the ArrowArray
+inferface: https://github.com/duckdb/duckdb/blob/master/src/common/types/data_chunk.cpp
+
+It is not expected that most drivers will have a dedicated implementation of
+GetArrowStream() or its callbacks. Implementing it requires a non-trivial effort, and
+significant gains are to be expected only for those for which I/O is very fast,
+and thus in-memory shuffling of data takes a substantial time relatively to the
+total time (I/O + shuffling).
+
+Potential future work, no in the scope of this RFC, could be the addition of a
```
snowman2, line 243:
```suggestion
Potential future work, not in the scope of this RFC, could be the addition of a
```

jorisvandenbossche:
AFAIU _all_ drivers can produce this, they only might not have a specialized implementation (so falling back to the generic implementation using the existing `GetNextFeature` underneath). 
In the current POC of Even, he also included a specialized implementation for FlatGeoBuf (besides Parquet/Arrow, see [Impacted Drivers](https://github.com/rouault/gdal/blob/rfc_86/doc/source/development/rfc/rfc86_column_oriented_api.rst#impacted-drivers) section in the RFC), which gives a 2-3x speedup compared to the generic implementation of this API according for the included benchmark. And in this benchmark, even this generic implementation of the proposed API gives a speed-up compared to the current fiona/pyogrio bindings to get this data in Python memory.

You are correct that if you just want to consume an Arrow file, it might still be faster / easier to use an arrow library to read that file. But so personally for me (and for geopandas), this proposal is mostly exciting for _all other formats_ (geopandas will probably keep a specialized `read_parquet` that does not go through GDAL but uses pyarrow directly):

* If you need to the data in a columnar layout (as we do for reading the data into a geopandas GeoDataFrame, but the same is true for R sf, columnar databases, ..), it is both more efficient and more convenient that the GetNextFeature iteration happens in GDAL than in the bindings, even if the file format you are reading is row-oriented. 
* For some file formats (even row-oriented ones, like FlatGeoBuf in the POC), GDAL can include a specialized driver-specific implementation of the proposed API, providing an additional speed-up.

mrylov:
@rouault, unfortunately there is no optimized API in HANA for column oriented retrieval.

rouault:
Updated to reflect the fact that the implementation has now a GeoPackage  specialized implementation, which is extremely fast !

doc/source/development/rfc/rfc86_column_oriented_api.rst
```diff
@@ -0,0 +1,692 @@
...
+        //
+        if( stream.get_next(&stream, &array) != 0 ||
+            array.release == nullptr )
+        {
+            break;
+        }
+        // Do something useful
+        array.release(&array);
+    }
+    stream.release(&stream);
```
nyalldawson, line 161: Why does this method require the explicit object pointer argument?

doc/source/development/rfc/rfc86_column_oriented_api.rst
```diff
@@ -0,0 +1,692 @@
...
+  The starting address of buffers allocated by the
+  default implementation is aligned on 64-byte boundaries.
+
+  The default implementation outputs geometries as WKB in a binary field,
+  whose corresponding entry in the schema is marked with the metadata item
+  ``ARROW:extension:name`` set to ``WKB``. Specialized implementations may output
+  by default other formats (particularly the Arrow driver that can return geometries
+  encoded according to the GeoArrow specification (using a list of coordinates).
+  The GEOMETRY_ENCODING=WKB option can be passed to force the use of WKB (through
+  the default implementation)
```
nyalldawson, line 200: One thing I'm not super-clear on here -- if the requested geometry format matches the underlying dataset's geometry format, does that mean there'll be absolutely **no** parsing/re-interpretation of the geometry on GDAL's side?

nyalldawson:
Another question -- it's not mentioned anywhere how this would interact with any filters applied to a layer, eg via OGR_L_SetAttributeFilter/OGR_L_SetSpatialFilter. I'm assuming that OGR level filters will be incompatible with the columnar api?

doc/source/development/rfc/rfc86_column_oriented_api.rst
rouault: The release callbacks may need to use the private_data member of the ArrowArray and ArrowArrayStream structures (those are C structures, not C++ objects)

doc/source/development/rfc/rfc86_column_oriented_api.rst
rouault: > does that mean there'll be absolutely no parsing/re-interpretation of the geometry on GDAL's side?

yes, this is typically the case for GeoParquet that uses WKB as the storage format. And also for GeoPackage where only the GeoPackage geometry header is stripped from the SQlite blob to get a WKB geometry. No OGRGeometry object is instanciated then.

rouault:
> Another question -- it's not mentioned anywhere how this would interact with any filters applied to a layer, eg via OGR_L_SetAttributeFilter/OGR_L_SetSpatialFilter. I'm assuming that OGR level filters will be incompatible with the columnar api?

"  The method may take into account ignored fields set with SetIgnoredFields() (the
  default implementation does), and should take into account filters set with
  SetSpatialFilter() and SetAttributeFilter(). Note however that specialized implementations
  may fallback to the default (slower) implementation when filters are set."

doc/source/development/rfc/rfc86_column_oriented_api.rst
```diff
@@ -0,0 +1,692 @@
...
+  of up to 65,536 features (configurable with the MAX_FEATURES_IN_BATCH=num option).
+  The starting address of buffers allocated by the
+  default implementation is aligned on 64-byte boundaries.
+
+  The default implementation outputs geometries as WKB in a binary field,
+  whose corresponding entry in the schema is marked with the metadata item
+  ``ARROW:extension:name`` set to ``WKB``. Specialized implementations may output
+  by default other formats (particularly the Arrow driver that can return geometries
+  encoded according to the GeoArrow specification (using a list of coordinates).
+  The GEOMETRY_ENCODING=WKB option can be passed to force the use of WKB (through
```
jorisvandenbossche, line 199: Could it in theory (later) also be possible to enable that you can specify `GEOMETRY_ENCODING=GEOARROW` to force the use of this encoding, even if the file itself is not using it?

rouault: yes one could imagine such an extension

doc/source/development/rfc/rfc86_column_oriented_api.rst
rouault: I should precise that when the default implementation of OGRLayer is used, as GetNextFeature() is used underneath and returns a OGRGeometry, if the underlying format would happen to use WKB for storage, you'd have:  WKB --> OGRGeometry (in GetNextFeature() -> WKB (in OGRLayer::GetNextArrowArray()). That would for example be the case for the PostGIS driver currently (with EWKB as the source encoding)

doc/source/development/rfc/rfc86_column_oriented_api.rst
```diff
@@ -0,0 +1,692 @@
...
+  state (the method will ignore any initial content).
+
+  The default implementation uses GetNextFeature() internally to retrieve batches
+  of up to 65,536 features (configurable with the MAX_FEATURES_IN_BATCH=num option).
+  The starting address of buffers allocated by the
+  default implementation is aligned on 64-byte boundaries.
+
+  The default implementation outputs geometries as WKB in a binary field,
+  whose corresponding entry in the schema is marked with the metadata item
+  ``ARROW:extension:name`` set to ``WKB``. Specialized implementations may output
```
jorisvandenbossche, line 196: The general recommendation at https://arrow.apache.org/docs/dev/format/Columnar.html#extension-types is to "namespace" the name value. I don't know if "WKB" is an abbreviation that is also used in other context, but so that is something to consider here.

One option it to use the "geoarrow" namespace (eg "geoarrow.wkb"); we are also considering to keep this as an option in the geo-arrow-spec (current PR to list those names is at https://github.com/geopandas/geo-arrow-spec/pull/22.

On the one hand that keeps it consistent with the other encodings ("geoarrow.point", "geoarrow.linestring", ..), but on the other hand this can also be confusing if we are using "geoarrow" for both the nested-list-array based encodings as for a binary one like WKB (I know that @kylebarron has mentioned this confusion before).

Using something like "gdal.wkb" is also an option, but on the other hand the representation itself (WKB) is not really GDAL specific, so it might be nice to have a more general name.

jorisvandenbossche: (this is actually also something that is relevant for the existing Arrow driver, so it might be worth to have this conversation in a separate issue)

jratike80: It might be "ogc.wkb" but some parties think that the axis order in WKB should depend on CRS  https://bugs.mysql.com/bug.php?id=88869. So perhaps gdal.wkb is better, or ogc.wkb-subversion-as-in-the-gpkg-standard

> [K11] The axis order in WKB is always (x,y{,z}{,m}) where x is easting or longitude, y is northing
> or latitude, z is optional elevation and m is optional measure.

rouault: The coordinate order and the WKB encoding are orthogonal topics (GDAL has its own mechanics to possibly define an for the override the CRS axis order through the DataAxisToSRSAxisMapping concept: https://gdal.org/tutorials/osr_api_tut.html#crs-and-axis-order)
==> "ogc.wkb" or "iso.wkb" could be good candidates

rouault: changed to "ogc.wkb"

doc/source/development/rfc/rfc86_column_oriented_api.rst
nyalldawson: Hmm, ok. Do you think it's worth having a gdal specific wrapper around those structures so that we can provide a nicer public api which hides this kind of detail?

doc/source/development/rfc/rfc86_column_oriented_api.rst
nyalldawson: (I realise this is off-topic, sorry!)

> yes, this is typically the case for GeoParquet that uses WKB as the storage format. And also for GeoPackage where only the GeoPackage geometry header is stripped from the SQlite blob to get a WKB geometry.

What's your thoughts regarding having a similar short-cut approach for the non-column oriented apis? I.e. being able to read features from a gpkg where the geometry is left as just the wkb binary content, and no interpretation is made on gdal's side. In the gpkg case this could be a good optimisation opportunity for some clients, in that they could request the wkb content directly if they are going to be constructing their own geometry representation of the features (ie gpkg -> wkb -> client geometry classes ) and want to avoid the extra representation (gpkg -> wkb -> ogr -> client geometry )

nyalldawson:
@rouault 
> The method may take into account ignored fields set with SetIgnoredFields() (the
default implementation does), and should take into account filters set with
SetSpatialFilter() and SetAttributeFilter(). Note however that specialized implementations
may fallback to the default (slower) implementation when filters are set.

Gosh, how did I miss that?! :facepalm:

doc/source/development/rfc/rfc86_column_oriented_api.rst
rouault: > Do you think it's worth having a gdal specific wrapper around those structures so that we can provide a nicer public api which hides this kind of detail?

The Arrow C stream/array interfaces defines a standard, and I'd want to avoid defining something GDAL specific on top of it. Anything that would make it more friendly to consume it would apply to any software that would produce Arrow C streams/arrays, so that doesn't really belong to GDAL specificially.  @paleolimbot  raised a discussion about that in https://lists.apache.org/thread/6vt7btpbrkm554otxqz7oxzdlpfwkhwc . He has prototyped
https://github.com/paleolimbot/gpkg/tree/master/src/arrow-hpp for the writing side. Not sure if there's an equivalent on the reading side (except using the arrow-cpp library itself to ingest Arrow C arrays and expose corresponding C++ objects, but that's an heavy dependency)

doc/source/development/rfc/rfc86_column_oriented_api.rst
rouault: That's a good idea, but I'm not sure how to implement this in a generic option. I guess this could be an open option implemented in the drivers where that would make sense that would turn the geometry column as a regular Binary field. I'm not sure how to inform client of which column(s) has the geometry. maybe some new field in the OGRGeomFieldDefn class.

doc/source/development/rfc/rfc86_column_oriented_api.rst
paleolimbot: Yes, there's general agreement from the Arrow end of things that there should be a library to make what you are doing here easier (both writing and reading). What you are doing here has been very helpful to push that forward in the discussions that we're having about this...I'm hoping to convert the mailing list discussion into a PR in the next few days although it will be some time before it's ready to use. It will basically be the arrow-hpp thing that Even linked to but using C idioms.

I separated out some C++ helpers here: https://github.com/paleolimbot/geoarrow-cpp , which could eventually help on the Geo-specific reading side (on pause at the moment while the lower-level Arrow library gets sorted).

All of these are some months away from being useful here, and reading Even's implementations has been incredibly helpful both in writing code and in demonstrating that people actually want to do this kind of thing.

PR #5840, opened 2022-06-01
Add a GDALRasterBand::GetSuggestedBlockAccessPattern() method

### LINKED ISSUE #5589: Very long process time of this PDF

Seems very specific to this PDF.
In fact I don't know if it crash or not but the process time is very long, I kill QGIS after 1 or 2 minutes.

Downstream issue https://github.com/qgis/QGIS/issues/48103

Can be labeled as very low priority 😉 

1. Try to import the following PDF
[Plan Topo 1983 modifie en 1984 de Pelgros a St Amand 1de2.pdf](https://github.com/qgis/QGIS/files/8428226/Plan.Topo.1983.modifie.en.1984.de.Pelgros.a.St.Amand.1de2.pdf)

## Operating system and GDAL provenance

Confirmed on Windows 10: 
using QGIS 3.24.1 (qgis 3.24.1-2) and QGIS 3.22.5 (qgis-ltr 3.22.5-2) from OSGeo4W v2 (with GDAL 3.4.2)   
and also using QGIS 3.16.16 (qgis-ltr 3.16.16-1) from OSGeo4W v2 (with GDAL 3.4.1)  
and from OSGeo4W v1 (with GDAL 3.1.4).

rouault:
Fixes #5589

```cpp
/**
 * \brief Return the suggested/most efficient access pattern to blocks
 *        (for read operations).
 *
 * While all GDAL drivers have to expose a block size, not all can guarantee
 * efficient random access (GSBAP_RANDOM) to any block.
 * Some drivers for example decompress sequentially a compressed stream from
 * top raster to bottom (GSBAP_TOP_TO_BOTTOM), in which
 * case best performance will be achieved while reading blocks in that order.
 * (accessing blocks in random access in such rasters typically causes the
 * decoding to be re-initialized from the start if accessing blocks in
 * a non-sequential order)
 *
 * The base implementation returns GSBAP_UNKNOWN, which can also be explicitly
 * returned by drivers that expose a somewhat artificial block size, because they can
 * extract any part of a raster, but in a rather inefficient way.
 *
 * The GSBAP_LARGEST_CHUNK_POSSIBLE value can be combined as a logical bitmask
 * with other enumeration values (GSBAP_UNKNOWN, GSBAP_RANDOM, GSBAP_TOP_TO_BOTTOM,
 * GSBAP_BOTTOM_TO_TOP). When a driver sets this flag, the most efficient strategy
 * is to read as many pixels as possible in the less RasterIO() operations.
 *
 * The return of this method is for example used to determine the swath size used by
 * GDALDatasetCopyWholeRaster() and GDALRasterBandCopyWholeRaster().
 *
 * @since GDAL 3.6
 */

### DIFF
--- a/frmts/gtiff/geotiff.cpp
+++ b/frmts/gtiff/geotiff.cpp
@@ -1331,6 +1331,9 @@ public:
     virtual CPLErr IReadBlock( int, int, void * ) override;
     virtual CPLErr IWriteBlock( int, int, void * ) override;
 
+    virtual GDALSuggestedBlockAccessPattern GetSuggestedBlockAccessPattern()
+        const override { return GSBAP_RANDOM; }
+
     virtual int IGetDataCoverageStatus( int nXOff, int nYOff,
                                         int nXSize, int nYSize,
                                         int nMaskFlagStop,
--- a/frmts/jpeg/jpgdataset.h
+++ b/frmts/jpeg/jpgdataset.h
@@ -361,6 +361,9 @@ class JPGRasterBand final: public GDALPamRasterBand
     virtual CPLErr IReadBlock( int, int, void * ) override;
     virtual GDALColorInterp GetColorInterpretation() override;
 
+    virtual GDALSuggestedBlockAccessPattern GetSuggestedBlockAccessPattern()
+        const override { return GSBAP_TOP_TO_BOTTOM; }
+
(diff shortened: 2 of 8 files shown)


nyalldawson (approved):
Nice!

PR #5849, opened 2022-06-02
OGR: example of API extension - add upsert operation

TheDorkKnight:
## What does this PR do?

This submits for consideration a pattern by which GDAL could expand its current set of supported feature operations (`Create`, `Set`, and `Delete`, ie. `INSERT`, `UPDATE`, and `DELETE` in SQL parlance).

As an example, it adds a method for `UPSERT` (ie. if record does not exist then `INSERT`, else `UPDATE`) and implements support for that method in the MongoDBv3 driver.

I don't expect that you should simply accept the pull request, but rather I'm hoping to demonstrate the general pattern and then continue discussion with GDAL project maintainers.

### SetFeature
The current `SetFeature` implementations for various drivers often have subtly different functionality. These usually either:
 - *Replace* the existing feature with the contents of the incoming `OGRFeature`
 - *Update* just the fields that exist on the incoming `OGRFeature`

Expanding the set of supported operations would allow an opportunity to disambiguate those cases.

### DeleteFeature
The current `DeleteFeature` implementation only allows an integer primary key. However, many formats supported by GDAL actually use string primary keys (eg. Elasticsearch, MongoDB).

Expanding the set of supported operations would allow an opportunity to provide a string-keyed overload or alternative.

### (Stretch Goal) Multiple Keys
It would be useful to allow update or delete operations to match multiple key fields on an incoming `OGRFeature` instead of just the primary key. For example, you could have an API like:

```c++
// Updates all existing features that match the values of each key field
OGRErr UpdateFeature(OGRFeature* poFeature, const int* panKeyFields, int nKeyFieldCount);
```

## Tasklist

 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/autotest/ogr/ogr_mongodbv3.py
+++ b/autotest/ogr/ogr_mongodbv3.py
@@ -730,6 +730,33 @@ def test_ogr_mongodbv3_2():
     gdal.PopErrorHandler()
     assert ret != 0
 
+    # Upsert when feature does not exist
+    gdal.PushErrorHandler()
+    ret = lyr.UpsertFeature(f)
+    gdal.PopErrorHandler()
+    assert ret != 0
+
+    gdal.PushErrorHandler()
+    ret = lyr.DeleteFeature(1)
+    gdal.PopErrorHandler()
+    assert ret != 0
+
+    # Upsert when feature already exists
+    gdal.PushErrorHandler()
+    ret = lyr.CreateFeature(f)
+    gdal.PopErrorHandler()
+    assert ret != 0
+
+    gdal.PushErrorHandler()
+    ret = lyr.UpsertFeature(f)
+    gdal.PopErrorHandler()
+    assert ret != 0
+
+    gdal.PushErrorHandler()
+    ret = lyr.DeleteFeature(1)
+    gdal.PopErrorHandler()
+    assert ret != 0
+
 ###############################################################################
 # test_ogrsf
 
--- a/ogr/ogr_api.h
+++ b/ogr/ogr_api.h
@@ -660,6 +660,7 @@ OGRFeatureH CPL_DLL OGR_L_GetFeature( OGRLayerH, GIntBig )  CPL_WARN_UNUSED_RESU
 OGRErr CPL_DLL OGR_L_SetFeature( OGRLayerH, OGRFeatureH ) CPL_WARN_UNUSED_RESULT;
 OGRErr CPL_DLL OGR_L_CreateFeature( OGRLayerH, OGRFeatureH ) CPL_WARN_UNUSED_RESULT;
 OGRErr CPL_DLL OGR_L_DeleteFeature( OGRLayerH, GIntBig ) CPL_WARN_UNUSED_RESULT;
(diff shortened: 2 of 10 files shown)


jratike80:
The string-id is somehow dealt with at least WFS driver https://github.com/OSGeo/gdal/blob/53364c89d9514e0f5964c2842ba6b602148b7079/doc/source/drivers/vector/wfs.rst. 

Upsert was new to me so this wiki page was interesting to read https://wiki.postgresql.org/wiki/UPSERT.
Upsert as a feature might be nice to have when a dataset should be updated and the data provider does not know which features the user already has. But user may have extra features that should be deleted and upsert does not resolve that issue.

rouault:
- The IUpsertFeature() proposal is a reasonable one. We'd want to add a OLCUpsertFeature capability for layers to advise its availability
- I was wondering if a default implementation that would do "if GetFeature(feature->GetFID()) SetFeature(feature) else CreateFeature(feature)" would make sense, but probably not given that it wouldn't be atomic and not all drivers support properly CreateFeature() with the FID set. And that wouldn't make sense for drivers were the FID is a "fake" one
- Updating the Memory driver to have an implementation of it that is always present would make sense.
- Regarding the update/replace subtelty, that must be mostly with SQL based drivers where UpdateFeature() is implemented as a SQL "UPDATE ... SET ..." statement, when passing a feature with unset fields (to be opposed to null fields). Such unset fields aren't listed in the SET list (I can actually see this was mentioned in https://gdal.org/development/rfc/rfc67_nullfieldvalues.html#drivers). But other drivers would likely consider unset as null. Having two different methods for update vs replace (or a single one with a flag specifying the intended behaviour) would be nice but certainly challenging to implement consistently over all the drivers.
- Regarding multiple columns for the primary key and/or non-numeric single primary key, this is a harder topic given how entangled in the API & design of OGRLayer the FID integer column concept is. I can see that some drivers  (sqlite, postgreSQL, mongoDB etc) could automatically detect which columns form the primary key and report it in a GetPrimaryKeyColumns() method. I'm not clear if there would be uses cases where it would probably be the responsibility of the user to declare which columns for the primary key.
- Regarding the comment "Updates all existing features that match the values of each key field" in the UpdateFeature() with multiple fields, I'm not sure to understand. If the user specifies fields that define a primary key, then at most one single feature would be updated, not several.

TheDorkKnight:
> The string-id is somehow dealt with at least WFS driver https://github.com/OSGeo/gdal/blob/53364c89d9514e0f5964c2842ba6b602148b7079/doc/source/drivers/vector/wfs.rst.

@jratike80 I assume we're talking about `OGRErr DeleteFromFilter( CPLString osOGCFilter );`? Because otherwise the problem of `DeleteFeature( GIntBig nFID );` remains, in that `nFID` is unsuitable to carry a string ID value.

In any case, I think something like a generic call and capability to `OLCDeleteFeature` (something like `OLCDeleteFeatureWithStringId`) would solve this problem both for WFS and for a variety of other formats such as MongoDB and Elasticsearch (and the deprecated CouchDB/Cloudant drivers too).

rouault:
superseded per #6199

PR #5854, opened 2022-06-03
VFK: Support to UTF-8 VFK data

isax-sro:
## What does this PR do?

Support for UTF-8 encoding. The support for old encodings and possibly new ones should also work.

## Environment

* OS: Debian
* Compiler: gcc

### DIFF
--- a/gdal/ogr/ogrsf_frmts/vfk/vfkdatablock.cpp
+++ b/gdal/ogr/ogrsf_frmts/vfk/vfkdatablock.cpp
@@ -179,7 +179,7 @@ int IVFKDataBlock::AddProperty(const char *pszName, const char *pszType)
         pszType = "T30";
 
     VFKPropertyDefn *poNewProperty = new VFKPropertyDefn(pszName, pszType,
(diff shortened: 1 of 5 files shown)


rouault:
CC @landam

rouault:
Doxygen / documentation issue:
```
 vfkpropertydefn.cpp:44:10: error: parameter 'bEncoding' not found in the function declaration [-Werror,-Wdocumentation]
  \param bEncoding encoding (only for "text" type)
         ^~~~~~~~~
```

isax-sro:
Can you point me to some place how to compile GDAL from github code? The configure file is missing so standard way
`
./configure
./make && make install
`
is not possible.

rouault:
> The configure file is missing

run "./autogen.sh" before to generate it.

by the way, you should re-target your pull request against the master branch of GDAL, which is where changes should go first

isax-sro:
Thanks for configure file help.
How to rebuild GDAL if there are changes in header files? It seems to me I have to do `make clean` and start fresh new build with `make` that takes so long. Any help?

PR #5871, opened 2022-06-05
GPKG: preliminary non-user-visible support for Related Tables Extension

rouault:
(https://docs.opengeospatial.org/is/18-000/18-000.html)

There is non-user-visible change related to that (as we don't have any
API to report relationships between table), but this brings:
- recognizing the extension as known
- take it into account during column renaming and table renaming/deletion
  to avoid corruption of the database.
- make the validate_gpkg.py validate the use of the extension

CC @nyalldawson


nyalldawson (approved):

nyalldawson:
LGTM!

nyalldawson:
(fwiw the relation api is one of my next work items, so that's coming soon too)

PR #5883, opened 2022-06-07
VFK: Support for UTF-8

isax-sro:
## What does this PR do?

Support to UTF-8 data

## Environment

* OS: Ubuntu 20.04
* Compiler: gcc

### DIFF
--- a/ogr/ogrsf_frmts/vfk/vfkreader.cpp
+++ b/ogr/ogrsf_frmts/vfk/vfkreader.cpp
@@ -155,6 +155,65 @@ char *VFKReader::ReadLine()
 }
 
 /*!
+  \brief Load text encoding from header (&HENCODING)
+
+  Called from VFKReader::ReadDataBlocks()
+*/
+void VFKReader::ReadEncoding()
+{
+    VSIFSeekL(m_poFD, 0, SEEK_SET);
+    char *pszLine = nullptr;
+    while ((pszLine = ReadLine()) != nullptr) {
+        if (strlen(pszLine) < 2 || pszLine[0] != '&') {
+            CPLFree(pszLine);
+            continue;
+        }
+        if (pszLine[1] == 'B' ||
+            (pszLine[1] == 'K' && strlen(pszLine) == 2)) {
+            /* 'B' record closes the header section */
+            /* 'K' record is end of file */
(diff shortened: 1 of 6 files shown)


rouault:
Does this need to be backported to 3.5 branch ?

ogr/ogrsf_frmts/vfk/vfkreader.cpp
```diff
@@ -154,6 +154,65 @@ char *VFKReader::ReadLine()
...
+            break;
+        }
+        if (pszLine[1] != 'H') {
+            /* (not) 'H' header */
+            CPLFree(pszLine);
+            continue;
+        }
+
+        char *poKey = pszLine + 2; /* &H */
+        char *poValue = poKey;
```
rouault, line 185: Naming convention: this should pszKey and pszValue

isax-sro:
Hello,
I think it would be very useful as from the 14th April 2022 data from 
state institution are exported only in UTF-8. But Martin Landa should 
confirm this.

Michal

nyalldawson:
The layer should also return true for the OLCStringsAsUTF8 capability now, shouldn't it?

rouault:
note that if this needs to go in 3.5.1 (@landam  is that needed ? or is your previous commit in the driver sufficient?), this PR should be finalized to take into account above comments before June 27th

landam (approved):

landam:
> note that if this needs to go in 3.5.1 (@landam is that needed ? or is your previous commit in the driver sufficient?), this PR should be finalized to take into account above comments before June 27th

This PR also contains important fix for octal representation of the currenncy sign. @rouault Therefore I think it should be backported to 3.5 branch.

rouault:
My and Nyall's comment need to be taken into account before being able to merge this

landam:
> My and Nyall's comment need to be taken into account before being able to merge this

@isax-sro Please change your PR accordingly.

isax-sro:
Hello, what do you mean "change your PR accordingly"? Any hint? I am not 
familiar with github/GDAL workflow.

landam:
> nevermind: as time is ticking and I want to issue a release candidate today, I've just pushed myself the fixes

Thanks @rouault !

PR #5892, opened 2022-06-08
OpenFileGDB: fix reading some spatial index .spx files whose value_count field is 0 (fixes #5888)

### LINKED ISSUE #5888: OpenFileGDB driver fails to retrieve features when zoomed in if .spx files are present, but works fine if they are removed.

I have a data folder in Esri File Geodatabase format containing .spx files. It has a cluster of features in South America and another cluster in Scandinavia. 

The features can be read via the OpenFileGDB driver, but only in zoomed-out scales, for example a world map, and when zooming in further, the features are no longer retrieved. 

Actually, the zoomed-in scale isn't the only trigger. There are some scales where you _can_ see the features as long as all they are all inside the map window, but if you then pan so that one or a few features goes outside the map window, they will all disappear. 

However, if all .spx index files are removed from the folder, the features are retrieved in any scale and panning. 

Drag and drop the data folder onto QGIS version 3.22.0, and choose to open all three layers. At startup, you should be able to see some features in South America and some features in Scandinavia. After zooming in or panning too much, the features will no longer be visible. 

The data folder comes from one of our customers and may be used only to reproduce the problem. So I don't want to attach the data here in public, but I can send it by mail. 

## Operating system

Windows 10. 

I have noticed the bug in our own software that uses GDAL version 3.3.2. I can also reproduce it with QGIS version 3.22.0-Białowieża which also uses GDAL version 3.3.2.

rouault:


nyalldawson:
Is this a corrupt file? Or something we missed in the reverse engineering?

nyalldawson (approved):

PR #5901, opened 2022-06-09
FileGDB/OpenFileGDB: handle Shape_Area/Shape_Length fields

rouault:
FileGDB and OpenFileGDB:
- on reading, when parsing the XML layer definition, use the
  AreaFieldName and LengthFieldName attributes to set the default value
  of the corresponding OGR fields to the driver specific value of
  "FILEGEODATABASE_SHAPE_AREA" / "FILEGEODATABASE_SHAPE_LENGTH"

FileGDB:
- on writing, add a CREATE_SHAPE_AREA_AND_LENGTH_FIELDS=YES/NO creation
  option to automatically create the Shape_Area and Shape_Length special
  fields
- make ogr2ogr automatically set that option when the source layer has
  the Shape_Area and Shape_Length special fields

### DIFF
--- a/apps/ogr2ogr_lib.cpp
+++ b/apps/ogr2ogr_lib.cpp
@@ -3863,6 +3863,28 @@ std::unique_ptr<TargetLayerInfo> SetupTargetLayer::Setup(OGRLayer* poSrcLayer,
             CPLDebug("GDALVectorTranslate", "Transferring layer NATIVE_DATA");
         }
 
+        // For FileGeodatabase, automatically set CREATE_SHAPE_AREA_AND_LENGTH_FIELDS=YES
+        // creation option if the source layer has a Shape_Area/Shape_Length field
+        if( papszDestCreationOptions &&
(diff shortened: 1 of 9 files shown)


nyalldawson (approved):

nyalldawson:
Code looks good 👍

What's the intended use case for this change? Is it required to make compliant gdbs? (Or, more specifically -- should we be auto populating these special fields whenever features are written?)

rouault:
> What's the intended use case for this change? Is it required to make compliant gdbs? (Or, more specifically -- should we be auto populating these special fields whenever features are written?)

I couldn't find from online Esri docs if those fields are required, although there are several mentions of them. I don't think they are required (otherwise people would have complained that the GDBs we create with the SDK aren't interoperable), but they seem quite frequent (I suspect most workflows from Esri software result in them being automatically created). When the <AreaFieldName>Shape_Area</AreaFieldName> declaration is found in the XML definition (same for Shape_Length), the SDK will automatically populate their value on feature creation from the geometry (and override whatever value the user might have put in it).
The main interest of this will be for the write side of the OpenFileGDB driver where we'll have to emulate that behaviour when editing an existing table with those special fields

PR #5915, opened 2022-06-12
Fix multiple build and runtime issues on architectures with "capability pointers"

rouault:
Architectures such as ARM Morello or RISC V CHERI (see https://www.cl.cam.ac.uk/research/security/ctsrd/cheri/) use 128 bit pointers that have extra protection against wrong usage. So in particular on those architectures sizeof(void*) == 16 and sizeof(size_t) == 8. This breaks a couple assumption in code
Another more tricky aspect is that a valid pointer cannot be used from the string representation of its address (it crashes at runtime, due to being considered invalid). So basically all uses of CPLPrintPointer() / DATAPOINTER with the MEM driver, and other more confidential usages in a few drivers, are all broken. Fortunately they can be avoided.

### DIFF
--- a/alg/gdalcutline.cpp
+++ b/alg/gdalcutline.cpp
@@ -40,6 +40,8 @@
 #include "cpl_string.h"
 #include "gdal.h"
 #include "gdal_alg.h"
+#include "gdal_priv.h"
+#include "memdataset.h"
 #include "ogr_api.h"
 #include "ogr_core.h"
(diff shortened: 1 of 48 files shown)


nyalldawson (approved):
Looks good (up to [a0460c1](https://github.com/OSGeo/gdal/pull/5915/commits/a0460c10e051fe4af770933a3e60f1cdf583a1c5))

PR #5921, opened 2022-06-13
OGR SQL dialect: improve awkward grammar

spencermathews:
## What does this PR do?

Make the phrasing of the special fields section in the OGR SQL dialect docs less awkward and grammatically correct.


rouault:
Thanks

PR #5927, opened 2022-06-15
KEA: add support for 64 bit nodata functions

gillins:
## What does this PR do?

Adds support for (u)int64 int nodata functions to the KEA driver. KEA supports reading and writing of the nodata value as any data type so we use this functionality to save and retrieve the nodata as (u)int64.

This PR also addresses finding `libkea` on Windows. Because of the somewhat non standard naming kea uses on Windows (is called `libkea.lib` there rather than `kea.lib` as `cmake` expects) we need to explicitly add `libkea` so it is found.

Support for 64 bit KEA images was added in https://github.com/OSGeo/gdal/pull/5710.

@rouault are there any other areas of driver functionality that have been recently changed to support 64 bit?

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:

### DIFF
--- a/cmake/helpers/CheckDependentLibraries.cmake
+++ b/cmake/helpers/CheckDependentLibraries.cmake
@@ -576,7 +576,7 @@ gdal_check_package(GEOS "Geometry Engine - Open Source (GDAL core dependency)" R
 )
 gdal_check_package(HDF4 "Enable HDF4 driver" CAN_DISABLE)
 
-define_find_package2(KEA libkea/KEACommon.h kea)
+define_find_package2(KEA libkea/KEACommon.h kea;libkea)
 gdal_check_package(KEA "Enable KEA driver" CAN_DISABLE)
 
 gdal_check_package(ECW "Enable ECW driver" CAN_DISABLE)
--- a/frmts/kea/keaband.cpp
+++ b/frmts/kea/keaband.cpp
@@ -558,6 +558,44 @@ double KEARasterBand::GetNoDataValue(int *pbSuccess)
     }
 }
 
+int64_t KEARasterBand::GetNoDataValueAsInt64(int *pbSuccess)
+{
+    try
+    {
+        int64_t nVal;
+        this->m_pImageIO->getNoDataValue(this->nBand, &nVal, kealib::kea_64int);
+        if( pbSuccess != nullptr )
+            *pbSuccess = 1;
+
+        return nVal;
+    }
+    catch (const kealib::KEAIOException &)
+    {
+        if( pbSuccess != nullptr )
+            *pbSuccess = 0;
+        return -1;
+    }
+}
+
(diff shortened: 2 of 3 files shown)


cmake/helpers/CheckDependentLibraries.cmake
```diff
@@ -576,7 +576,7 @@ gdal_check_package(GEOS "Geometry Engine - Open Source (GDAL core dependency)" R
 )
 gdal_check_package(HDF4 "Enable HDF4 driver" CAN_DISABLE)
 
-define_find_package2(KEA libkea/KEACommon.h kea)
+define_find_package2(KEA libkea/KEACommon.h kea;libkea)
```
rouault, line 579: why/when is that needed ? according to https://cmake.org/cmake/help/latest/command/find_library.html "Each library name given to the NAMES option is first considered as a library file name and then considered with platform-specific prefixes (e.g. lib) and suffixes (e.g. .so)."

frmts/kea/keaband.cpp
```diff
@@ -558,6 +558,44 @@ double KEARasterBand::GetNoDataValue(int *pbSuccess)
...
+        if( pbSuccess != nullptr )
+            *pbSuccess = 1;
+
+        return nVal;
+    }
+    catch (const kealib::KEAIOException &)
+    {
+        if( pbSuccess != nullptr )
+            *pbSuccess = 0;
+        return -1;
```
rouault, line 595: to avoid confusion with a negative value in a signed type.
```suggestion
        return std::numeric_limits<uint64_t>::max();
```
you'll also need to add a ``#include <limits>``

rouault (commented):
Some added tests in autotest/gdrivers/kea.py might be useful

rouault:
> are there any other areas of driver functionality that have been recently changed to support 64 bit?

I don't think so

cmake/helpers/CheckDependentLibraries.cmake
gillins: On Windows AFAICT `lib` isn't one of the prefixes that is tried but the library is called `libkea.lib` on Windows.

frmts/kea/keaband.cpp
gillins: Good call, I've just committed this change.

cmake/helpers/CheckDependentLibraries.cmake
dg0yt: @rouault The file name is: 
`libkea.lib`.
But the tested patterns for Windows are:
`kea` (the "name")
`kea.lib` (name with prefix and suffix).
(This is a problem when Windows libs try to simulate POSIX name style.

PR #5931, opened 2022-06-15
[Backport release/3.4] OGR: Backport fixes for `Is3D` & `IsMeasured` from #5477 to 3.4

evetion:
## What does this PR do?
Backports fixes for `Is3D` & `IsMeasured` from #5477 to 3.4

The GeoParquet support introduced in #5477, also fixed two smaller bugs in `Is3D` and `IsMeasured`, namely that they don't return a boolean as documented, but the result of a bitflag and operation.

This change was breaking for packages that worked around this bug, when 3.5 was introduced in downstream packages, such as GDAL.jl 1.4 in https://github.com/yeesian/ArchGDAL.jl/pull/307.

By backporting, we can align the behaviour of downstream supported versions.

## Tasklist

 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/gdal/ogr/ogr_geometry.h
+++ b/gdal/ogr/ogr_geometry.h
@@ -402,9 +402,9 @@ class CPL_DLL OGRGeometry
     virtual OGRGeometry* Normalize() const;
     virtual OGRBoolean  IsSimple() const;
     /*! Returns whether the geometry has a Z component. */
-    OGRBoolean  Is3D() const { return flags & OGR_G_3D; }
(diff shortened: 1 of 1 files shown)


rouault:
note that there won't be any 3.4.x official releases anymore at that point

rouault:
actually if the change that went in 3.5.0 breaks users (I wouldn't have expected it, but that's something that can happen for a feature release), it is *not* appropriate for backport in a stable branch like 3.4, so I believe this PR shouldn't be merged.

PR #5933, opened 2022-06-16
Multidim API: significantly enhance performance of reading transposed…

rouault:
… arrays for netCDF/HDF5

Using the netCDF/HDF5 APIs to read a slice with strides that express a
transposed view yield to extremely poor/unusable performance. This fixes
this by using temporary memory to read in a contiguous buffer in a
row-major order, and then do the transposition to the final buffer.


rouault:
(CMake build-mac failure unrelated)

PR #5935, opened 2022-06-17
KEA: add tests for 64 bit funcionality

gillins:
## What does this PR do?

Adds some tests for 64 bit functionality for the KEA driver. It also fixes a missing space in `keadriver.cpp`.

https://github.com/OSGeo/gdal/pull/5927

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:


gillins:
I'm not sure the `autotest/gcore/data/gtiff/int64.tif` etc files are correct for this. Should we create a `autotest/gcore/data/int64.tif` instead with the same data as the other tiff files in `autotest/gcore/data`?

rouault:
> I'm not sure the `autotest/gcore/data/gtiff/int64.tif` etc files are correct for this

yes, I'd prefer you use autotest/gcore/data/gtiff/int64.tif instead of creating a new file. Historically all data files were under autotest/gcore/data without any driver-based subdirectory, which became messy (actually that was especially for the autotest/gdrivers/data directory). I've re-organized that a bit, but gave up for those byte, int16, etc. existing TIFF files that are used in so many places in the .py files

gillins:
> yes, I'd prefer you use autotest/gcore/data/gtiff/int64.tif instead of creating a new file.

It turns out the files in autotest/gcore/data are 20x20 images with some random looking data with a known checksum. The autotest/gcore/data/gtiff/int64.tif file is a 2x1 file with a different checksum to the others. Happy to go with whatever you think, but maybe we should be moving the larger files into the `gtiff` folder instead?

rouault:
>      It turns out the files in autotest/gcore/data are 20x20 images with some random looking data with a known checksum. The autotest/gcore/data/gtiff/int64.tif file is a 2x1 file with a different checksum to the others.

ah true. I now remember I've created the int64 and uint64 test files with pixel values that illustrate their specific range of values, whereas other files have all the same content.

> Happy to go with whatever you think, but maybe we should be moving the larger files into the `gtiff` folder instead?

that would be nice, but not in that pull request, as you're going to have to edit several tens of files to reflect the path change

gillins:
> that would be nice, but not in that pull request, as you're going to have to edit several tens of files to reflect the path change

By my count there are 442 .py files using images in `gcore/data`... Are you happy to merge this one and look at moving files around later?

rouault:
> Are you happy to merge this one and look at moving files around later?

ok. My slight worry is that we have now a autotest/gcore/data/[u]int16.tif and autotest/gcore/data/gtiff/[u]int16.tif, and they are not the same, which is confusing. I guess renaming the later files to autotest/gcore/data/gtiff/[u]int16_full_range.tif or something like that (they must be referenced in very few files) would probably be better. But can be done in a later pull request

PR #5969, opened 2022-06-23
Fix 'union' and 'intersect' descriptions in gdal_calc.py documentation

skualos:
The definitions of the `union` and `intersect` parameters of the `extent` option seemed to be reversed in the gdal_calc.py documentation

### DIFF
--- a/doc/source/programs/gdal_calc.rst
+++ b/doc/source/programs/gdal_calc.rst
@@ -108,9 +108,9 @@ but no projection checking is performed (unless projectionCheck option is used).
     
     ``fail`` - the dimensions and the extent (bounds) of the rasters must agree, otherwise the operation will fail.
     
-    ``union`` - the extent (bounds) of the output will be the minimal rectangle that contains all the input extents.
+    ``intersect`` - the extent (bounds) of the output will be the minimal rectangle that contains all the input extents.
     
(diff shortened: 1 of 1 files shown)


jratike80:
I am not sure. If this is the situation
image
then isn't the union case this
image
but the intersect case this
image

rouault:
@idanmiara can you check this proposed change ? (I'm not sure it is. the current description seems consistent with the behaviour I'd expect from the naming of the options)

stephane-rds:
The functions work as expected. And their description is actually also right.
Apologies for this.
The words `maximal` and `minimal` led me to a wrong interpretation, while not reading it too carefully.
If anyone can reject the PR, I can't find a way to remove it.

skualos:
The functions work as expected. And their description is also actually right.
Apologies for this. The words `minimal` and `maximal` led me to a wrong interpretation while not reading it too carefully.
Closing PR

idanmiara:
> The functions work as expected. And their description is also actually right. Apologies for this. The words `minimal` and `maximal` led me to a wrong interpretation while not reading it too carefully. Closing PR

Yes, correct. But I see your point. Do you want to rephrase the definition?

lnicola:
TBH, I think the current phrasing is fine.

PR #6050, opened 2022-07-09
Fix compiling for ios

m-kuhn:
"Darwin" matches osx only

### DIFF
--- a/cmake/helpers/configure.cmake
+++ b/cmake/helpers/configure.cmake
@@ -180,7 +180,7 @@ else ()
     endif ()
   endif ()
 
-  if (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
+  if (APPLE)
     if (NOT VSI_FTELL64 AND NOT VSI_FSEEK64)
       set(VSI_FTELL64 "ftello")
       set(VSI_FSEEK64 "fseeko")


rouault:
Did you test on a recent enough master ? https://github.com/OSGeo/gdal/pull/6024 address iOS build issues in a different way

m-kuhn:
I tested 3.5.2 only, patching with https://github.com/OSGeo/gdal/pull/6024 now

m-kuhn:
I can confirm https://github.com/OSGeo/gdal/pull/6024 works, but this one looks simpler and I'm not sure https://github.com/OSGeo/gdal/pull/6024 has other advantages?

rouault:
> I can confirm #6024 works, but this one looks simpler and I'm not sure #6024 has other advantages?

It was found by the reporter of what lead to #6024 that your approach helped, but was not sufficient (there were compilation errors with statsvfs64 for example). See thread https://lists.osgeo.org/pipermail/gdal-dev/2022-July/thread.html#56026 . Perhaps there was some oddity in the build environment of the reporter...

m-kuhn:
In this case let's stick to https://github.com/OSGeo/gdal/pull/6024

PR #6080, opened 2022-07-16
DXF: Prevent buffer from sometimes splitting CRLF newlines in MLEADER entities / Support files between 2 GB and 4 GB in size

atlight:


rouault:
is there a reason to limit to 4 GB files ? If using vsi_l_offset to store file offsets, we could address arbitrarily large files

PR #6089, opened 2022-07-19
Add new driver metadata capability flags

nyalldawson:
Adds driver metadata to determine on a driver level:

- create layer capability
- delete layer capability
- create/delete field capability
- reorder field capability
- supported alter field definition flags
- curve/measured geometry support

Allows these capabilities to be determined at a driver level


swig/include/gdalconst.i
```diff
@@ -205,11 +205,19 @@
 %constant char *DCAP_RASTER     = GDAL_DCAP_RASTER;
 %constant char *DCAP_VECTOR     = GDAL_DCAP_VECTOR;
 %constant char *DCAP_GNM     = GDAL_DCAP_GNM;
+%constant char *DCAP_CREATE_LAYER = GDAL_DCAP_CREATE_LAYER;
```
rouault, line 208: those constants should also be ported in the #else clause (after line 233) for the Java and C# bindings

ogr/ogrsf_frmts/filegdb/FGdbLayer.cpp
```diff
@@ -4014,6 +4014,9 @@ int FGdbLayer::TestCapability( const char* pszCap )
     else if( EQUAL(pszCap,OLCIgnoreFields) )
         return TRUE;
 
+    else if( EQUAL(pszCap,OLCCurveGeometries) )
```
rouault, line 4017: I don't think this is appropriate for the FileGDB driver. It uses internally OGRWriteToShapeBin() which isn't ready yet to support curve geometries (it could, just that it isn't implemented).

ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobufdataset.cpp
```diff
@@ -127,6 +127,12 @@ void RegisterOGRFlatGeobuf()
     GDALDriver *poDriver = new GDALDriver();
     poDriver->SetDescription("FlatGeobuf");
     poDriver->SetMetadataItem(GDAL_DCAP_VECTOR, "YES");
+    poDriver->SetMetadataItem(GDAL_DCAP_CREATE_LAYER, "YES");
+    poDriver->SetMetadataItem(GDAL_DCAP_CREATE_FIELD, "YES");
+    poDriver->SetMetadataItem(GDAL_DCAP_DELETE_FIELD, "YES");
+    poDriver->SetMetadataItem(GDAL_DCAP_REORDER_FIELDS, "YES");
+    poDriver->SetMetadataItem(GDAL_DCAP_CURVE_GEOMETRIES, "YES");
```
rouault, line 134: to remove per above comment

ogr/ogrsf_frmts/filegdb/FGdbDriver.cpp
```diff
@@ -847,6 +847,11 @@ void RegisterOGRFileGDB()
     poDriver->SetMetadataItem( GDAL_DMD_LONGNAME,
                                 "ESRI FileGDB" );
     poDriver->SetMetadataItem( GDAL_DCAP_VECTOR, "YES");
+    poDriver->SetMetadataItem( GDAL_DCAP_DELETE_LAYER, "YES" );
+    poDriver->SetMetadataItem( GDAL_DCAP_CREATE_LAYER, "YES" );
+    poDriver->SetMetadataItem( GDAL_DCAP_CREATE_FIELD, "YES" );
+    poDriver->SetMetadataItem( GDAL_DCAP_DELETE_FIELD, "YES" );
+    poDriver->SetMetadataItem( GDAL_DCAP_REORDER_FIELDS, "YES" ); // although marked as "TBD" in FGdbLayer.cpp!
```
rouault, line 854: this line should be removed. And the 2 lines in FGdbLayer that advertize OLCReorderFields should be removed, as they are definitely not appropriate

rouault (commented):
Great (and fastidious I guess!) job.
I'm wondering if we shouldn't extent test_ogrsf to test for example that if a dataset declares ODsCCreateLayer, then its driver declares GDAL_DCAP_CREATE_LAYER. And similarly for other capabilities

nyalldawson:
Thanks for the review @rouault !

I was also thinking of a potential "supports z" driver flag to mimic the measure equivalent. Do you have any ideas for a clever way to identify these drivers? (I'll do it all by hand if not 😱)

rouault:
> I was also thinking of a potential "supports z" driver flag to mimic the measure equivalent. Do you have any ideas for a clever way to identify these drivers? (I'll do it all by hand if not scream)

No universal way I'm afraid. For drivers that explicitly deal with individual geometry types, grepping for "25D" (such as in wkbPoint25D) is a way of finding a subset. Grepping for "getZ" or "setZ" too

swig/include/gdalconst.i
nyalldawson: done

ogr/ogrsf_frmts/filegdb/FGdbLayer.cpp
nyalldawson: done

ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobufdataset.cpp
nyalldawson: @rouault  was this comment supposed to be directed at the filegdb or openfilegdb drivers install of flatgeobuf?

ogr/ogrsf_frmts/filegdb/FGdbDriver.cpp
nyalldawson: done

ogr/ogrsf_frmts/mysql/ogrmysqldriver.cpp
```diff
@@ -144,6 +144,9 @@ void RegisterOGRMySQL()
 
     poDriver->SetDescription( "MySQL" );
     poDriver->SetMetadataItem( GDAL_DCAP_VECTOR, "YES" );
+    poDriver->SetMetadataItem( GDAL_DCAP_CREATE_LAYER, "YES" );
+    poDriver->SetMetadataItem( GDAL_DCAP_DELETE_LAYER, "YES" );
+    poDriver->SetMetadataItem( GDAL_DCAP_CREATE_FIELD, "YES" );
```
rouault, line 149: MYSQL supports Z and M geometries  (I'd expect it to also support curve geometries, but that should be dealt in a separate pull request by adding OLCMeasuredGeometries and testing. (if OLCMeasuredGeometries isn't declared, the CreateFeature() logic silently linearizes geometries))

apps/test_ogrsf.cpp
```diff
@@ -401,6 +401,40 @@ static int TestDataset( GDALDriver** ppoDriver )
     poDS->GetMetadata( nullptr );
     poDS->GetMetadataItem( "", nullptr );
 
+    if ( !poDriver->GetMetadataItem( GDAL_DCAP_VECTOR ))
+    {
+        printf("FAILURE: Driver does not advertize GDAL_DCAP_VECTOR!\n" );
+        GDALClose(poDS);
```
rouault, line 407: instead of returning, just set bRet = false so that later checks can be also done in the same run. Same for below checks

apps/test_ogrsf.cpp
```diff
@@ -401,6 +401,40 @@ static int TestDataset( GDALDriver** ppoDriver )
...
+    // Test consistency of datasource capabilities and driver metadata
+    if ( poDS->TestCapability(ODsCCreateLayer) && !poDriver->GetMetadataItem( GDAL_DCAP_CREATE_LAYER ) )
+    {
+        printf("FAILURE: Dataset advertizes ODsCCreateLayer capability but driver metadata does not advertize GDAL_DCAP_CREATE_LAYER!\n" );
+        GDALClose(poDS);
+        return FALSE;
+    }
+    if ( !bReadOnly && poDriver->GetMetadataItem( GDAL_DCAP_CREATE_LAYER ) && !poDS->TestCapability(ODsCCreateLayer) )
+    {
+        printf("FAILURE: Driver advertizes GDAL_DCAP_CREATE_LAYER capability but dataset does not advertize ODsCCreateLayer!\n" );
```
rouault, line 420: I can imagine a number of drivers where that would fail: those who can support only 1 layer. That should be combined with a test of GDAL_DCAP_MULTIPLE_VECTOR_LAYERS

apps/test_ogrsf.cpp
```diff
@@ -1144,6 +1178,85 @@ static int TestBasic( OGRLayer *poLayer )
...
+
+        if ( poLayer->TestCapability(OLCCreateField) && !poDriver->GetMetadataItem( GDAL_DCAP_CREATE_FIELD ) )
+        {
+            bRet = FALSE;
+            printf("FAILURE: Layer advertizes OLCCreateField capability but driver metadata does not advertize GDAL_DCAP_CREATE_FIELD!\n" );
+        }
+        if ( bLayerShouldHaveEditCapabilities && poDriver->GetMetadataItem( GDAL_DCAP_CREATE_FIELD ) && !poLayer->TestCapability(OLCCreateField) )
+        {
+            bRet = FALSE;
+            printf("FAILURE: Driver metadata advertizes GDAL_DCAP_CREATE_FIELD but layer capability does not advertize OLCCreateField!\n" );
```
rouault, line 1212: I wouldn't be surprised to see this fail for a number of drivers. Some drivers only support CreateField() on a newly created layers, before the first feature is written

apps/test_ogrsf.cpp
```diff
@@ -1144,6 +1178,85 @@ static int TestBasic( OGRLayer *poLayer )
...
+        }
+
+        if ( poLayer->TestCapability(OLCCreateField) && !poDriver->GetMetadataItem( GDAL_DMD_CREATIONFIELDDATATYPES ) )
+        {
+            bRet = FALSE;
+            printf("FAILURE: Layer advertizes OLCCreateField capability but driver metadata does not include GDAL_DMD_CREATIONFIELDDATATYPES!\n" );
+        }
+        if ( bLayerShouldHaveEditCapabilities && poDriver->GetMetadataItem( GDAL_DMD_CREATIONFIELDDATATYPES ) && !poLayer->TestCapability(OLCCreateField) )
```
rouault, line 1220: the following would be safer
```suggestion
        if ( poDriver->GetMetadataItem( GDAL_DMD_CREATIONFIELDDATATYPES ) && !poDriver->GetMetadataItem( GDAL_DCAP_CREATE_FIELD ) )
```

apps/test_ogrsf.cpp
```diff
@@ -1144,6 +1178,85 @@ static int TestBasic( OGRLayer *poLayer )
...
+        {
+            bRet = FALSE;
+            printf("FAILURE: Driver metadata advertizes GDAL_DMD_CREATIONFIELDDATATYPES but layer capability does not advertize OLCCreateField!\n" );
```
rouault, line 1223:
```suggestion
            printf("FAILURE: Driver metadata advertizes GDAL_DMD_CREATIONFIELDDATATYPES but driver metadata does not advertize GDAL_DCAP_CREATE_FIELD!\n" );
```

gcore/gdal.h
```diff
@@ -493,6 +534,27 @@ typedef struct GDALDimensionHS* GDALDimensionH;
...
+
+/** Capability set by a driver that can support 2.5 dimension geometries (i.e. support for Z dimensions).
+ *
+ * Note: This capability is only partially supported by drivers. If present then
+ * Z dimensions are definitely supported by a driver, but the absence of this
+ * capability does not guarantee that Z dimensions are NOT supported by the driver.
+ *
+ * @since GDAL 3.6
+ */
+#define GDAL_DCAP_25D_GEOMETRIES "DCAP_25D_GEOMETRIES"
```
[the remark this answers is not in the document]
nyalldawson: done

ogr/ogr_core.h
```diff
@@ -887,6 +887,7 @@ int CPL_DLL OGRParseDate( const char *pszInput, OGRField *psOutput,
 #define OLCCreateGeomField     "CreateGeomField"    /**< Layer capability for geometry field creation */
 #define OLCCurveGeometries     "CurveGeometries"    /**< Layer capability for curve geometries support */
 #define OLCMeasuredGeometries  "MeasuredGeometries" /**< Layer capability for measured geometries support */
+#define OLC25DGeometries       "25DGeometries"      /**< Layer capability for 25D geometry support. Since GDAL 3.6. */
```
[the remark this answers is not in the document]
nyalldawson: I've gone with that for consistency.

(Another option could be "OLCZDimension" )

nyalldawson:
Note to self: fix inconsistency in constant definition names in gdalconst.i for c#/java case (some constants strip the GDAL_ prefix, others don't)

nyalldawson:
@rouault this is ready for re-review/merge

As a follow up I'm also thinking of adding a driver level metadata for indicating whether a particular driver is sensitive to multi/single parts for different geometry types. I.e. a way to differentiate e.g. gpkg which cares about multipolygons vs polygons vs shapefile which has no distinction.

Got any thoughts on that before I dig in?

nyalldawson:
(I should also ask if there's any other metadata items you've previously thought would be useful, and I'll tackle them while I'm focused on this...)

rouault:
> whether a particular driver is sensitive to multi/single parts for different geometry types. I.e. a way to differentiate e.g. gpkg which cares about multipolygons vs polygons vs shapefile which has no distinction.
> 
> Got any thoughts on that before I dig in?

for the write side of the drivers ? I can think of shapefile, (open)filegdb. Not sure which others.  And for those, this is only for linestring/multilinestring, and polygon/multipolygon. point and multipoint are dealt as separate types.

swig/include/gdalconst.i
```diff
@@ -228,26 +237,50 @@
...
 #define GDAL_DMD_CREATIONDATATYPES "DMD_CREATIONDATATYPES"
 #define GDAL_DMD_CREATIONFIELDDATATYPES "DMD_CREATIONFIELDDATATYPES"
+#define GDAL_DMD_CREATIONFIELDDATASUBTYPES "DMD_CREATIONFIELDDATASUBTYPES"
 #define GDAL_DMD_SUBDATASETS "DMD_SUBDATASETS"
 #define GDAL_DMD_CREATION_FIELD_DOMAIN_TYPES "DMD_CREATION_FIELD_DOMAIN_TYPES"
 #define GDAL_DMD_ALTER_GEOM_FIELD_DEFN_FLAGS "DMD_ALTER_GEOM_FIELD_DEFN_FLAGS"
 
 #define GDAL_DCAP_OPEN       "DCAP_OPEN"
 #define GDAL_DCAP_CREATE     "DCAP_CREATE"
+#define GDAL_DCAP_CREATE_MULTIDIMENSIONAL "DCAP_CREATE_MULTIDIMENSIONAL"
```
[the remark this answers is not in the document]
nyalldawson: > Perhaps we should go for the constants without GDAL_ for new stuff, and offer the shorter aliases for the existing ones ?

Sounds good. We could always just add aliases for ALL constants so that they all work with/without the prefix too...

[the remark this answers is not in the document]
rouault: > We could always just add aliases for ALL constants so that they all work with/without the prefix too...

yes probably better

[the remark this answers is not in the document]
nyalldawson: Ok, I've gone with exposing both variants

nyalldawson:
@rouault 
> I can think of shapefile, (open)filegdb.

Looks like it also applies to pgeo  :+1:

Issue #5680, opened 2022-05-06
Remove autoconf & nmake build systems


rouault:
- [x] Migrate CI (.github, appveyor.yml, .travis.yml) to CMake builds:
    * [x] Travis CI s390x and graviton2: https://github.com/OSGeo/gdal/pull/6040
    * [x]  mingw_w64: https://github.com/OSGeo/gdal/pull/6048
    * [x]  fedora: https://github.com/OSGeo/gdal/pull/6053
    * [x]  asan: https://github.com/OSGeo/gdal/pull/6055
    * [x] windows: https://github.com/OSGeo/gdal/pull/6056 
    * [x] code_checks: https://github.com/OSGeo/gdal/pull/6057
    * [x] ubuntu_2004: https://github.com/OSGeo/gdal/pull/6058
    * [x] ubuntu_1804 and ubuntu_1804_32bit: https://github.com/OSGeo/gdal/pull/6101
    * [x] macos:  https://github.com/OSGeo/gdal/pull/6108
    * [x] clang static analyzer: https://github.com/OSGeo/gdal/pull/6107 
    * [x]  Remove .github/workflows/android.yml: android_cmake.yml already exists: https://github.com/OSGeo/gdal/commit/678a428cac4dfcf8690ed27923e1834bc3b55661
    * [x] Remove appveyor.yml: https://github.com/OSGeo/gdal/commit/265f3e8db7057d59356331ffb0a67bd176bc5c02 
- [x] Migrate OSSFuzz fuzzers/build.sh to CMake builds: https://github.com/OSGeo/gdal/pull/6038
- [ ] Remove configure.ac, m4, GDALmake.opt.in, nmake.opt, all GNUmakefile & makefile.vc files, generate_vcxproj.bat: https://github.com/OSGeo/gdal/pull/6110
- [x] Adapt release process / mkgdaldist.sh: https://github.com/OSGeo/gdal/pull/6039
- [x] swig/python/CMakeLists.txt: apply all customizations from GNUmakefile: https://github.com/OSGeo/gdal/pull/6100

rouault:
> The `generate_vcxproj.bat` file in the repo root is `nmake`-based and will need to be updated.

--> to be removed actually

atlight:
> > to be removed actually
> 
> @atlight why sad face ? The CMake visual studio generator should do an equivalent job, or am I missing something ?

Ah, if an alternative option is available that is ok. I am not familiar with CMake at all so I was scared of losing access to such a resource.

PR #6100, opened 2022-07-20
swig/python/CMakeLists.txt: apply all customizations from GNUmakefile

rouault:
to regenerated files, and copy regenerated files into source directory

refs #5680

### DIFF
--- a/doc/source/build_hints.rst
+++ b/doc/source/build_hints.rst
@@ -1665,7 +1665,13 @@ It is used for the Python, Java and CSharp bindings.
 
 .. option:: SWIG_EXECUTABLE
 
-    Path to the SWIG executable
+    Path to the SWIG executable.
+
+    Note that setting it explictly might be needed, and that putting the
+    directory of the installed binary into the PATH might not be sufficient.
+    The reason is that when buildingfrom source, a "swig" binary will be
+    generated, but FindSWIG will prefer a "swig-4.0" binary if found elsewhere
+    in the PATH.
 
 
 TEIGHA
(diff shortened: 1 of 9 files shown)


doc/source/build_hints.rst
```diff
@@ -2014,6 +2020,11 @@ Python bindings options
     Whether Python bindings should be built. It is ON by default, but only
     effective if a Python installation is found.
 
+.. option:: SWIG_REGENERATE_PYTHON:BOOL=ON/OFF
+
+    Whether to refresh the generated SWIG Python bindings. It is OFF by default.
+    Setting it to ON is needed if modifying the SWIG interface files.
```
nyalldawson, line 2026: On thing I've been unsure about -- is this something which should be run by a contributor prior to submitting a change, or is it something which should instead be run prior to release/as required on an ad-hoc basis?

I get a lot of noise when I regenerate the bindings automatically and usually resort to just manually updating the files (or just copy/pasting the actual useful changes). 

I think it would be good to clarify the expectations somewhere...

rouault: > - is this something which should be run by a contributor prior to submitting a change

yes. Maybe at some point we will require SWIG to be able to build GDAL, but this is not yet the case.

> I get a lot of noise when I regenerate the bindings automatically 

you have to use SWIG 4.0.2.  I'm not sure how you regenerated the bindings. If that was from a "make generate" from swig/python (which will be eventually retired once we've removed autoconf support), then I wouldn't expect noise.

> I think it would be good to clarify the expectations somewhere...

the updated CONTRIBUTING.md normally gives the good hints. Do you see additions that would be useful ?

PR #6104, opened 2022-07-21
Add GDALRelationship class for describing a relationship between two tables, implement for some drivers

nyalldawson:
This PR adds a new GDALRelationship class for describing the relationship between two tables.

This class is designed to be somewhat of a superset of the relationship properties utilised by formats which are likely to gain support for the relationship API in future, specifically Geopackage (via the related tables extension), FileGeodatabase and the various database drivers (postgres/sql server/oracle/...).

This PR also implements relationship discovery in the GPKG, openfilegdb, filegdb and pgeo drivers


nyalldawson:
@rouault this is ready for review/merge

ogr/ogrsf_frmts/gpkg/ogrgeopackagedatasource.cpp
```diff
@@ -1873,6 +1879,82 @@ void GDALGeoPackageDataset::FixupWrongMedataReferenceColumnNameUpdate()
...
+    m_osMapRelationships.clear();
+
+    auto oResultTable = SQLQuery(hDB, "SELECT base_table_name, base_primary_column, "
+                                 "related_table_name, related_primary_column, relation_name, "
+                                 "mapping_table_name FROM gpkgext_relations" );
+    if ( oResultTable && oResultTable->RowCount() > 0 )
+    {
+        for(int i=0; i<oResultTable->RowCount();i++)
+        {
+            const char* pszBaseTableName = oResultTable->GetValue(0, i);
```
rouault, line 1898: for robustness, you should assume that the return value of all those GetValue() might be nullptr (corrupted DB, etc.), and thus test for it before dereferencing/passing it to other functions

ogr/ogrsf_frmts/openfilegdb/ogropenfilegdbdatasource.cpp
```diff
@@ -526,11 +527,12 @@ int OGROpenFileGDBDataSource::OpenFileGDBv10(int iGDBItems,
     if( !oTable.Open(osFilename, false) )
         return FALSE;
 
-    int iUUID = oTable.GetFieldIdx("UUID");
-    int iName = oTable.GetFieldIdx("Name");
-    int iPath = oTable.GetFieldIdx("Path");
-    int iDefinition = oTable.GetFieldIdx("Definition");
-    int iDocumentation = oTable.GetFieldIdx("Documentation");
+    const int iUUID = oTable.GetFieldIdx("UUID");
+    const int iType = oTable.GetFieldIdx("Type");
```
rouault, line 531: iType should be tested like other fields at line 536

gcore/gdal.h
```diff
@@ -283,6 +283,11 @@ typedef void *GDALRasterAttributeTableH;
 /** Opaque type used for the C bindings of the C++ GDALAsyncReader class */
 typedef void *GDALAsyncReaderH;
 
+/** Opaque type used for the C bindings of the C++ GDALRelationship class
+ *  @since QGIS 3.6
```
rouault, line 287:
```suggestion
 *  @since GDAL 3.6
```

swig/include/Relationship.i
```diff
@@ -0,0 +1,243 @@
...
+  %apply (char **options) {char **};
+  char **GetLeftTableFields() {
+      return GDALRelationshipGetLeftTableFields(self);
+  }
+  %clear (char **);
+#else
+  void GetLeftTableFields(char ***pList) {
+      *pList = GDALRelationshipGetLeftTableFields(self);
+  }
+#endif
```
rouault, line 123:
```suggestion
%apply (char **CSL) {char **};
  char **GetLeftTableFields() {
     return GDALRelationshipGetLeftTableFields(self);
  }
%clear char **;
```
similarly for other instances that return a char** that must be CSLDestroy()'ed

gcore/gdal.h
```diff
@@ -589,6 +594,12 @@ typedef struct GDALDimensionHS* GDALDimensionH;
  */
 #define GDAL_DCAP_FIELD_DOMAINS "DCAP_FIELD_DOMAINS"
 
+/** Capability set by drivers for formats which support reading table relationships.
+ *
+ * @since GDAL 3.6
+ */
+#define GDAL_DCAP_RELATIONSHIPS "DCAP_RELATIONSHIPS"
```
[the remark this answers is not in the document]
nyalldawson: I was planning to add a bunch of DMD strings which break down what create relationship options are supported. Eg one for supported cardinality types, one for general supported properties (the labels, related table type,..) etc.

gcore/gdal.h
nyalldawson: :relaxed: oops, muscle memory!

nyalldawson:
Thanks for the review @rouault  -- it's all addressed now

gcore/gdalrelationship.cpp
```diff
@@ -0,0 +1,676 @@
...
+
+#include <algorithm>
+#include <cstddef>
+#include <new>
+
+#include "cpl_atomic_ops.h"
+#include "cpl_error.h"
+#include "cpl_multiproc.h"
+
+CPL_CVSID("$Id$")
```
rouault, line 41: All that can likely be removed:
```suggestion
```

rouault:
do you think you could rework the commit history a bit (I kown I'm somewhat maniac about that), or maybe just squashing everything together ?

gcore/gdalrelationship.cpp
nyalldawson: Done

nyalldawson:
@rouault 

> do you think you could rework the commit history a bit (I kown I'm somewhat maniac about that), or maybe just squashing everything together ?

No problem! I've done that now

nyalldawson:
Failing mac test is unrelated

Issue #5489, opened 2022-03-22
Error while reading geotiff: IReadBlock failed at X offset *, Y offset *: TIFFReadEncodedTile() failed


ktheijs:
## Expected behavior and actual behavior.

I expect to be able to read a specific location of a geotiff but I'm getting an error saying "IReadBlock failed at X offset *, Y offset *: TIFFReadEncodedTile() failed.".

I have seen this issue for many years in many different versions, but I now finally have a dataset on which I can reproduce the issue. I can provide the data privately if it helps to fix the problem.

## Steps to reproduce the problem.

The issue can be reproduced in C# with the code below. This is only one specific location where I know it crashes, but there are multiple.

```
RasterIOExtraArg extraArg = new RasterIOExtraArg();
        extraArg.nVersion = 1;
        extraArg.eResampleAlg = RIOResampleAlg.GRIORA_CubicSpline;
        extraArg.dfXOff = 49265.901653834488;
        extraArg.dfYOff = 665.27952667756563;
        extraArg.dfXSize = 129.00000000002183;
        extraArg.dfYSize = 129.000000000015;
        extraArg.bFloatingPointWindowValidity = 1;

float[] byteBuffer = new float[16384];	
int[] bandMap = { 1 };	
Dataset myDataset = Gdal.Open([MY_FILENAME], Access.GA_ReadOnly)
myDataset .ReadRaster(
         xOff = 49265,
         yOff = 665,
         xSize = 130,
         ySize = 130,
         byteBuffer,
         buf_xSize = 128,
         buf_ySize = 128,
         bandCount = 1,
         bandMap = bandMap,
         pixelSpace = 0,
         lineSpace = 0,
         bandSpace = 0,
         extraArg = extraArg);
```

It can also be reproduced in QGIS by trying to get the pixel info in the same location. I can provide data privately so you can easily find the location. You should get an error like in the screenshot below.

![image](https://user-images.githubusercontent.com/12170304/159459017-eb730420-eae2-4183-b335-81b3090c2588.png)

## Operating system

Windows 11 - 64 bit (OS build 22000.556)

## GDAL version and provenance

GDAL 3.3.3 in C# from https://www.gisinternals.com/
It can also be reproduced in QGIS 3.10.8 (which uses GDAL 3.0.4)
The first time I have encountered this issue must have been more than 4 years ago so I assume it happens in all versions of GDAL.

jratike80:
[...]

PR #6105, opened 2022-07-21
GTiff: fix reading some JPEG compressed COG files through /vsicurl/

rouault:


pedros007:
Thanks, Even!  We were able to find the reproduce case for this while investigating a different read error we'd occasionally get:

```
IReadBlock failed at X offset 26, Y offset 76: TIFFReadEncodedTile() failed
```

Makes me curious if this PR might address #5489 as well?

rouault:
> Makes me curious if this PR might address #5489 as well?

no, #5489 is about a corrupted file. The fix of that pull request is specific to the network reading optimized code path

PR #6116, opened 2022-07-23
Rename PCRE2::PCRE2 to PCRE2::PCRE2-8

m-kuhn:
To avoid collisions with other packages like qtbase that link to the 16bit version with the same target name, leading to race conditions if the 8 or 16 bit version of the library is linked


dg0yt:
FTR given that pcre2 supplies multiple libraries, the CMake package could keep it is name.
It is only the naming of the `PCRE2_LIBRARY...` and `PCRE2::PCRE2` stuff that doesn't fit to the multiple libraries.
Even `PCRE2_LIBRARIES` could be set according to `COMPONENTS` passed to `find_package`.

m-kuhn:
@dg0yt is that what you had in mind?

cmake/modules/packages/FindPCRE2.cmake
```diff
@@ -9,20 +9,20 @@
...
 include(FindPackageHandleStandardArgs)
-find_package_handle_standard_args(PCRE2
-                                  FOUND_VAR PCRE2_FOUND
-                                  REQUIRED_VARS PCRE2_LIBRARY PCRE2_INCLUDE_DIR)
-mark_as_advanced(PCRE2_INCLUDE_DIR PCRE2_LIBRARY)
-if(PCRE2_FOUND)
-    set(PCRE2_LIBRARIES ${PCRE2_LIBRARY})
+find_package_handle_standard_args(PCRE2-8
+                                  FOUND_VAR PCRE2-8_FOUND
+                                  REQUIRED_VARS PCRE2-8_LIBRARY PCRE2_INCLUDE_DIR)
```
dg0yt, line 16: We probably still need a
~~~cmake
set(PCRE2_FOUND "${PCRE2-8_FOUND}")
~~~

m-kuhn: Do we need `PCRE2-8_FOUND`? Or can we replace it as `FOUND_VAR PCRE2_FOUND`?

dg0yt: No, we don't need `PCRE2-8_FOUND`. I'm just not so familiar with the options of `find_package_handle_standard_args`.

m-kuhn: Research shows:
- `FOUND_VARS` is deprecated since cmake 3.3
- ci complains about `find_package_handle_standard_args` being used with a name different from the current Find module. Latest commit should fix that.

PR #6110, opened 2022-07-21
Remove autoconf and nmake build systems (fixes #5680)

rouault:

PR #6137, opened 2022-07-29
DOC: Generate Python API documentation with sphinxcontrib-apidoc

### LINKED ISSUE #6132: DOC: intersphinx for python API
https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html

It seems to work for the C/CPP functions, but not for Python.

snowman2:
Closes #6132

image
image


rouault:
ok, so that does seem to do a similar job to epydoc, but in a more integrated way.
I guess we could just get rid of the manually added epydoc that is installed by "tar xzf /tmp/python-doc.tar.gz" in doc_build.yml , and get rid of doc/source/api/python_api_ref.rst or redirect it to your new solution

snowman2:
Will need to get the Python bindings working for the docs:
```
WARNING: autodoc: failed to import module 'osgeo'; the following exception was raised:
No module named 'osgeo'
```

rouault:
Thanks. The mkgdaldist.sh script will also need to be updated since it does a "make man", and this fails if the Python bindings aren't build (or the alternative would be to manage to have "make man" not requiring them / skipping the sphinxcontrib-apidoc paths)

rouault:
I've locally tried the HTML generated pages, and I've noticed a strange navigation issue. If I go to "file:///home/even/gdal/gdal/doc/build/html/api/index.html#python-api" (reached from "API" on home page and then "Python API"), then clicking on "Python API reference" links from the menu or the non-menu part of the page leads to file:///home/even/gdal/gdal/doc/build/html/python/index.html which is a dead link. But if I click on "Python bindings", which leas to file:///home/even/gdal/gdal/doc/build/html/api/python.html, and then click on "Python API reference", it goes to file:///home/even/gdal/gdal/doc/build/html/api/python_api_ref.html which is the working page.

snowman2:
@rouault, I see in CONTRIBUTING how to regenerate the SWIG. However, when I make a change to the `*.i` files, it doesn't seem to regenerate when I do `cmake --build .`. Is there something that needs to be tweaked to do that? Am I missing something?

snowman2:
> I've locally tried the HTML generated pages, and I've noticed a strange navigation issue....

Good catch. I can dig into that later.

doc/source/api/index.rst
```diff
@@ -83,8 +83,8 @@ API
    .. toctree::
        :maxdepth: 1
 
-       python
-       python_api_ref
```
snowman2, line 87: Removed `python_api_ref` as it wasn't entirely necessary and was causing issues for some reason.

doc/source/api/index.rst
```diff
@@ -83,8 +83,8 @@ API
    .. toctree::
        :maxdepth: 1
 
-       python
```
snowman2, line 86: Renamed `python` to `python_bindings` to avoid conflicts with the `python/` directory.

swig/python/GNUmakefile
```diff
@@ -1,7 +1,5 @@
 
 
-include ../../GDALmake.opt
```
snowman2, line 3: This file doesn't exist and causes the makefile to error.

mkgdaldist.sh
```diff
@@ -141,7 +141,7 @@ CWD=${PWD}
 #
 echo "* Generating man pages..."
 
-(cd doc; make man)
+(cd doc; SPHINXOPTS='--keep-going -j auto' make man)
```
snowman2, line 144: This makes it so the warnings aren't turned into errors and it completes successfully even without the python bindings.

snowman2:
Test failure is unrelated: `AssertionError: Failed to open dataset: http://gdal.org/gdalicon.png`

rouault:
> I see in CONTRIBUTING how to regenerate the SWIG. However, when I make a change to the `*.i` files, it doesn't seem to regenerate when I do `cmake --build .` after the I initial run. Is there something that needs to be tweaked to do that? Am I missing something?

Did you cmake -DSWIG_REGENERATE_PYTHON=ON ? That should regenerate the .py and .cpp files in swig/python after a successful build

swig/python/GNUmakefile
rouault: swig/python/GNUmakefile is going to be removed per https://github.com/OSGeo/gdal/pull/6110

snowman2:
> Did you cmake -DSWIG_REGENERATE_PYTHON=ON ? That should regenerate the .py and .cpp files in swig/python after a successful build

Yes, I did. That is how I generated the files in this PR. However, it only seems to work on the initial call of `cmake --build .`. If I modify the `*.i` files and re-run `cmake --build .` it doesn't seem like it re-generates anything.

swig/python/GNUmakefile
snowman2: Makes sense - removed it.

rouault:
> If I modify the `*.i` files and re-run `cmake --build .` it doesn't seem like it re-generates anything.

I presume we have a lack of dependency rules on the .i files included by the "main" .i files. "rm -rf swig/python" from the build directory and run cmake again is a workaround.

rouault:
Thanks!

snowman2:
Thanks 👍

Issue #6115, opened 2022-07-23
Parquet Driver: some datasets make ogrinfo (and QGIS) crash


agiudiceandrea:
## Expected behavior and actual behavior.

Using https://github.com/OSGeo/gdal/blob/master/autotest/ogr/data/parquet/test.parquet there is no issue:

```
>ogrinfo test.parquet --debug ON
GDAL: Auto register C:\OSFB9F~1\apps\gdal\lib\gdalplugins\gdal_ECW_JP2ECW.dll using GDALRegister_ECW_JP2ECW.
GDAL: Auto register C:\OSFB9F~1\apps\gdal\lib\gdalplugins\gdal_GEOR.dll using GDALRegister_GEOR.
GDAL: Auto register C:\OSFB9F~1\apps\gdal\lib\gdalplugins\gdal_HDF5.dll using GDALRegister_HDF5.
GDAL: Auto register C:\OSFB9F~1\apps\gdal\lib\gdalplugins\gdal_MrSID.dll using GDALRegister_MrSID.
GDAL: Auto register C:\OSFB9F~1\apps\gdal\lib\gdalplugins\gdal_TileDB.dll using GDALRegister_TileDB.
GDAL: Auto register C:\OSFB9F~1\apps\gdal\lib\gdalplugins\ogr_OCI.dll using RegisterOGROCI.
GDAL: Auto register C:\OSFB9F~1\apps\gdal\lib\gdalplugins\ogr_SOSI.dll using RegisterOGRSOSI.
[...]

PR #6140, opened 2022-07-30
Parquet driver: fix compression methods

agiudiceandrea:
## What does this PR do?

In [docs](https://gdal.org/drivers/vector/parquet.html#layer-creation-options): adds UNCOMPRESSED, removes BZ2 and substitutes LZ4 with LZ4_RAW.

In OGRParquetDriver: removes BZ2 and LZ4_FRAME and substitutes LZ4 with LZ4_RAW. Would be OK to also remove the LZO method (it is not supported by the C++ implementation)?

See:
https://github.com/apache/parquet-format/blob/master/Compression.md
https://github.com/apache/arrow/blob/02c8598d264c839a5b5cf3109bfd406f3b8a6ba5/cpp/src/arrow/util/compression.cc#L48-L78
https://github.com/apache/arrow/blob/02c8598d264c839a5b5cf3109bfd406f3b8a6ba5/cpp/src/parquet/types.cc#L38-L51
https://github.com/apache/arrow/blob/02c8598d264c839a5b5cf3109bfd406f3b8a6ba5/cpp/src/parquet/types.cc#L57-L70
https://github.com/OSGeo/gdal/issues/6115#issuecomment-1200104043

### DIFF
--- a/doc/source/drivers/vector/parquet.rst
+++ b/doc/source/drivers/vector/parquet.rst
@@ -38,9 +38,9 @@ The driver supports creating only a single layer in a dataset.
 Layer creation options
 ----------------------
 
-- **COMPRESSION=string**: Compression method. Can be one of ``NONE``, ``SNAPPY``,
-  ``GZIP``, ``BROTLI``, ``ZSTD``, ``LZ4``, ``BZ2``, ``LZ4_HADOOP``. Available
-  values depend on how the Parquet library was compiled.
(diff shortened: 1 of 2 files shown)


rouault:
Thanks!

> Would be OK to also remove the LZO method (it is not supported by the C++ implementation)?

we can probably leave the GDAL code like it is currently. If a future version of Arrow Parquet would support LZO, it would be automatically exposed.

rouault (approved):

PR #6143, opened 2022-08-01
Fix flake8 E275 missing whitespace after keyword warnings

nyalldawson:


rouault:
I see a number of instances where the old version before your fix was ``return' fail'``, so the space character was wrongly put in the string. That should likely be corrected, although it probably doesn't matter much now ('fail' was a special recognized string before we pytest'ify autotest), or more likely those return 'fail' should be converted to a pytest assertion

PR #6161, opened 2022-08-05
gdalwarp: do not error out with a memory allocation failure if warping…

rouault:
…from a source with RPC to a target dataset with a no-overlapping extent, but emit a warning

### DIFF
--- a/apps/gdalwarp_lib.cpp
+++ b/apps/gdalwarp_lib.cpp
@@ -2163,7 +2163,10 @@ static void SetupSkipNoSource(int iSrc,
 /*                     AdjustOutputExtentForRPC()                       */
 /************************************************************************/
 
-static void AdjustOutputExtentForRPC( GDALDatasetH hSrcDS,
+/** Returns false if there's no intersection between source extent defined
+ * by RPC and target extent.
+ */
+static bool AdjustOutputExtentForRPC( GDALDatasetH hSrcDS,
                                       GDALDatasetH hDstDS,
                                       GDALTransformerFunc pfnTransformer,
                                       void *hTransformArg,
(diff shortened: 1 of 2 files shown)


apps/gdalwarp_lib.cpp
```diff
@@ -2163,7 +2163,10 @@ static void SetupSkipNoSource(int iSrc,
 /*                     AdjustOutputExtentForRPC()                       */
 /************************************************************************/
 
-static void AdjustOutputExtentForRPC( GDALDatasetH hSrcDS,
+/** Returns false if there's no intersection bettween source extent defined
```
EvertEt, line 2166:
```suggestion
/** Returns false if there's no intersection between source extent defined
```

apps/gdalwarp_lib.cpp
```diff
@@ -2205,6 +2208,12 @@ static void AdjustOutputExtentForRPC( GDALDatasetH hSrcDS,
                     static_cast<int>(std::ceil(dfMaxX)) + nPadding - nWarpDstXOff);
                 nWarpDstYSize = std::min(nWarpDstYSize - nWarpDstYOff,
                     static_cast<int>(std::ceil(dfMaxY)) + nPadding - nWarpDstYOff);
+                if( nWarpDstXSize <= 0 || nWarpDstYSize <= 0 )
+                {
+                    CPLError(CE_Warning, CPLE_AppDefined,
+                             "No intersection between source extent defined by RPC and target extent");
+                    return false;
```
pl-nfwarmerdam, line 2215: How convinced are you that this should issue a warning?  We don't warn for other files provided to gdalwarp that do not overlap the target file do we?  Perhaps it should only be a debug message?

rouault: changed to a debug message

PR #6171, opened 2022-08-09
gdalbuildvrt: skip source datasets that just touch by an edge the area of interest, and would result in a source with 0 width/height

rouault:


apps/gdalbuildvrt_lib.cpp
```diff
@@ -208,7 +208,8 @@ static int  GetSrcDstWin(DatasetProperty* psDP,
         *pdfSrcYSize = *pdfDstYSize / dfSrcToDstYSize;
     }
 
-    return TRUE;
+    return *pdfSrcXSize >= 0 && *pdfDstXSize >= 0 &&
+           *pdfSrcYSize >= 0 && *pdfDstYSize >= 0;
```
pl-nfwarmerdam, line 212: I'm surprised this is >= 0 instead of > 0 since it was the size=0 sources that was triggering an error later.

rouault: > I'm surprised this is >= 0 instead of > 0 since it was the size=0 sources that was triggering an error later.

ah indeed, should be > 0...  will fix. This part is more a safety belt. The actual change that fixes the issue is the changes at top of the function

PR #6210, opened 2022-08-16
Codec background in GTiff driver page

hobu:
The GTiff driver page doesn't do a great job of telling folks *why* they might choose one TIFF codec over another. Now that there are so many to choose from, we should editorialize a little bit to point people in the right direction.

### DIFF
--- a/doc/source/drivers/raster/gtiff.rst
+++ b/doc/source/drivers/raster/gtiff.rst
@@ -817,6 +817,53 @@ the default behavior of the GTiff driver.
    always written. If set to NO, then the transformation will not be written in
    any situation.
 
+
+Codec Recommendations
+---------------------
+
+
+LZW
+~~~
+
+If you don't know what to chose, chose this one.
+
+DEFLATE
+~~~~~~~
+
+The most commonly supported TIFF codec, especially with older non-geo software.
+
+LERC
+~~~~
+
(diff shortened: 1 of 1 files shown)


doc/source/drivers/raster/gtiff.rst
```diff
@@ -817,6 +817,53 @@ the default behavior of the GTiff driver.
...
 
+
+Codec Recommendations
+---------------------
+
+
+LZW
+~~~
+
+If you don't know what to chose, chose this one.
```
rouault, line 828:
```suggestion
If you don't know what to choose, choose this one.
```

rouault (commented):
maybe we should indicate which ones are lossless / lossy ?

hobu:
> maybe we should indicate which ones are lossless / lossy ?

Yeah, I wasn't sure how detailed to get. Most of the codecs are both lossy and lossless depending on configuration.

tbonfort:
I'd be willing to contribute to this PR, need a couple of days to do so. I think it would make sense to include some kind of comparison, at least storage wise, between the different codecs (eg one case for rgb, one case for a dem). Including compression/decompression speeds would also be helpfull but much more difficult to compare across various platforms/hardware.

rouault:
I'm going to merge this at it. Follow up contributions warmly welcome!

PR #6265, opened 2022-08-28
gdalinfo: add a STAC section to `gdalinfo -json` output

gadomski:
## What does this PR do?

`gdalinfo -json` output now includes a "stac" key, which is a dictionary of commonly-used STAC extension data.

STAC extensions supported:
- projection: https://github.com/stac-extensions/projection
- raster: https://github.com/stac-extensions/raster
- eo: https://github.com/stac-extensions/eo (only description)

All of the STAC fields are duplicated from other data already included in the JSON output. At some point in the future, it may make sense to output STAC Items from `-json` by default or provide a `-stac` flag; however, that would be a more significant change and would require some additional decisions (e.g. how to handle item ids, asset structure, etc).

No changes are made to the existing keys in the JSON output, so this should not affect existing usage.

EDIT: example output provided in a comment below.

The eo:band section is only for the "description" field, which should probably be moved to the raster extension in STAC (nothing to do with this PR).

None.

## Tasklist

 - [x] Add test case(s)
 - [x] Add documentation
 - [ ] Review
 - [x] Adjust for comments

## Environment

* OS: developed on macos
* Compiler: developed with clang 12.0.0

## CCs

@hobu @matthewhanson

### DIFF
--- a/apps/gdalinfo_lib.cpp
+++ b/apps/gdalinfo_lib.cpp
@@ -666,20 +666,12 @@ char *GDALInfo( GDALDatasetH hDataset, const GDALInfoOptions *psOptions )
             json_object_put(poMetadata);
 
         // Include eo:cloud_cover in stac output
-        char **papszMetadata = GDALGetMetadata(hDataset, "IMAGERY");
-        if( papszMetadata != nullptr && *papszMetadata != nullptr ) {
-            json_object * poValue = nullptr;
-            for (int i = 0; papszMetadata[i] != nullptr; ++i) {
-                char *pszKey = nullptr;
-                const char *pszValue =
-                    CPLParseNameValue( papszMetadata[i], &pszKey );
-                if( pszKey && strcmp(pszKey, "CLOUDCOVER") == 0 )
-                {
-                    poValue = json_object_new_int( atoi(pszValue) );
-                    json_object_object_add( poStac, "eo:cloud_cover", poValue );
-                    CPLFree( pszKey );
-                }
-            }
+        const char* pszCloudCover = GDALGetMetadataItem(hDataset, "CLOUDCOVER", "IMAGERY");
+        json_object * poValue = nullptr;
+        if( pszCloudCover )
+        {
+            poValue = json_object_new_int(atoi(pszCloudCover));
+            json_object_object_add(poStac, "eo:cloud_cover", poValue);
         }
     }
 
--- a/apps/gdalinfo_lib.cpp
+++ b/apps/gdalinfo_lib.cpp
@@ -254,6 +254,10 @@ char *GDALInfo( GDALDatasetH hDataset, const GDALInfoOptions *psOptions )
     json_object *poJsonObject = nullptr;
     json_object *poBands = nullptr;
     json_object *poMetadata = nullptr;
+    json_object *poStac = nullptr;
+    json_object *poStacProperties = nullptr;
+    json_object *poStacRasterBands = nullptr;
+    json_object *poStacEOBands = nullptr;
 
     const bool bJson = psOptions->eFormat == GDALINFO_FORMAT_JSON;
 
@@ -272,6 +276,10 @@ char *GDALInfo( GDALDatasetH hDataset, const GDALInfoOptions *psOptions )
         poJsonObject = json_object_new_object();
         poBands = json_object_new_array();
         poMetadata = json_object_new_object();
+        poStac = json_object_new_object();
+        poStacProperties = json_object_new_object();
+        poStacRasterBands = json_object_new_array();
+        poStacEOBands = json_object_new_array();
 
         json_object_object_add(poJsonObject, "description", poDescription);
         json_object_object_add(poJsonObject, "driverShortName",
@@ -344,7 +352,11 @@ char *GDALInfo( GDALDatasetH hDataset, const GDALInfoOptions *psOptions )
 
         json_object_array_add(poSize, poSizeX);
         json_object_array_add(poSize, poSizeY);
+
+        json_object *poStacSize = nullptr;
+        json_object_deep_copy(poSize, &poStacSize, nullptr);
         json_object_object_add(poJsonObject, "size", poSize);
+        json_object_object_add(poStacProperties, "proj:shape", poStacSize);
     }
     else
     {
@@ -382,7 +394,26 @@ char *GDALInfo( GDALDatasetH hDataset, const GDALInfoOptions *psOptions )
         if( bJson )
         {
             json_object *poWkt = json_object_new_string(pszPrettyWkt);
+            if (strcmp(psOptions->pszWKTFormat, "WKT2") == 0)
+            {
+                json_object *poStacWkt = nullptr;
+                json_object_deep_copy(poWkt, &poStacWkt, nullptr);
+                json_object_object_add(poStacProperties, "proj:wkt2", poStacWkt);
+            }
             json_object_object_add(poCoordinateSystem, "wkt", poWkt);
+
+            const char *epsg = OSRGetAuthorityCode(hSRS, nullptr);
+            if (epsg)
+            {
+                json_object *poEPSG = json_object_new_int64(atoi(epsg));
+                json_object_object_add(poStacProperties, "proj:epsg", poEPSG);
+            }
+            char * pszProjJson = nullptr;
+            OGRErr result = OSRExportToPROJJSON(hSRS, &pszProjJson, nullptr);
+            if (result == OGRERR_NONE) {
+                json_object *poStacProjJson = json_tokener_parse(pszProjJson);
+                json_object_object_add(poStacProperties, "proj:projjson", poStacProjJson);
+            }
 
             json_object* poAxisMapping = json_object_new_array();
             for( int i = 0; i < nAxesCount; i++ )
@@ -462,18 +493,28 @@ char *GDALInfo( GDALDatasetH hDataset, const GDALInfoOptions *psOptions )
         if( bJson )
         {
             json_object *poGeoTransform = json_object_new_array();
+            // Deep copy wasn't working on the array, for some reason, so we
+            // build the geotransform STAC array at the same time.
+            json_object *poStacGeoTransform = json_object_new_array();
 
             for( int i = 0; i < 6; i++ )
             {
                 json_object *poGeoTransformCoefficient =
                     json_object_new_double_with_precision(adfGeoTransform[i],
                                                           16);
+                json_object *poStacGeoTransformCoefficient =
+                    json_object_new_double_with_precision(adfGeoTransform[i],
+                                                          16);
+
                 json_object_array_add(poGeoTransform,
                                       poGeoTransformCoefficient);
+                json_object_array_add(poStacGeoTransform,
+                                      poStacGeoTransformCoefficient);
             }
 
             json_object_object_add(poJsonObject, "geoTransform",
                                    poGeoTransform);
+            json_object_object_add(poStacProperties, "proj:transform", poStacGeoTransform);
         }
         else
         {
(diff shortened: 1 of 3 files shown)


apps/gdalinfo_lib.cpp
```diff
@@ -440,7 +463,10 @@ char *GDALInfo( GDALDatasetH hDataset, const GDALInfoOptions *psOptions )
             if( bJson )
             {
                 json_object *proj4 = json_object_new_string(pszProj4);
+                json_object *poStacProj4 = nullptr;
+                json_object_deep_copy(proj4, &poStacProj4, nullptr);
                 json_object_object_add(poCoordinateSystem, "proj4", proj4);
+                json_object_object_add(poStac, "proj:projjson", poStacProj4);
```
hobu, line 469: `proj:projjson` should be https://proj.org/specifications/projjson.html not proj4 strings

hobu (changes requested):
looks good to me. Just need to fix up proj4 vs projjson

apps/gdalinfo_lib.cpp
gadomski: Yup will fix, thanks...

matthewhanson:
This looks like a great start to getting some STAC creation capability in GDAL.  

A clarifying question - I don't see the projjson in the example provided above, would the intention be that the output would provide all the fields, even if redundant?  As in, would it output wkt2, projjson, and epsg (if an epsg code existed)?   For this purpose where it's simply just outputting STAC fields I think it makes sense to do so, but if one were to actually create STAC Items at some point I'd rather not see such redundancy.

Related question, perhaps this should be another ticket, but does a STACIT write option (Create/Copy) make sense at all?  This is similar to a VRT in that it's not actually writing any data, just pointing to other existing data and the ability copy or create a STAC Item from scratch seems rather useful.

gadomski:
I've updated this PR to address @hobu's comment (https://github.com/OSGeo/gdal/pull/6265#discussion_r956721591) and slightly changed the structure of the output object. The STAC information is now under `stacItem`, and all non-band information is now under `properties`, e.g.:

```shell
$ gdalinfo -json autotest/gcore/data/byte.tif | jq  .stacItem
{
  "properties": {
    "proj:shape": [
      20,
      20
    ],
    "proj:wkt2": "PROJCRS[...snip...]",
    "proj:epsg": 26711,
    "proj:projjson": {
      ... snip ...
    },
    "proj:transform": [
      440720,
      60,
      0,
      3751320,
      0,
      -60
    ]
  },
  "raster:bands": [
    {
      "data_type": "uint8",
      "stats": {
        "minimum": 74,
        "maximum": 255,
        "mean": 126.765,
        "stddev": 22.928
      }
    }
  ],
  "eo:bands": [
    {}
  ]
}
```

> As in, would it output wkt2, projjson, and epsg (if an epsg code existed)?

Yup, right now I'm outputting wkt2 (if it's already there -- if the user specified WKT1 on the command line the WKT is excluded), projjson (as an object), and EPSG code (if it exists).

> if one were to actually create STAC Items at some point I'd rather not see such redundancy.

Agreed. In my head, we'll eventually make `gdalinfo -stac` which does exactly this. There's some complexity about multi-file situations, asset metadata, etc, which is why I think it makes sense to start here with this PR.

> does a STACIT write option (Create/Copy) make sense at all? 

I don't know enough about the STACIT driver and its use-cases to comment at this point, I'll leave this to others for now.

apps/gdalinfo_lib.cpp
```diff
@@ -382,8 +394,27 @@ char *GDALInfo( GDALDatasetH hDataset, const GDALInfoOptions *psOptions )
...
+                json_object *poStacWkt = nullptr;
+                json_object_deep_copy(poWkt, &poStacWkt, nullptr);
+                json_object_object_add(poStacProperties, "proj:wkt2", poStacWkt);
+            }
             json_object_object_add(poCoordinateSystem, "wkt", poWkt);
 
+            const char *epsg = OSRGetAuthorityCode(hSRS, nullptr);
+            if (epsg)
+            {
+                json_object *poEPSG = json_object_new_int64(atoi(epsg));
```
rouault, line 408: there can be codes in other namespaces than EPSG:, so:
```suggestion
            const char *pszAuthCode = OSRGetAuthorityCode(hSRS, nullptr);
            const char *pszAuthName = OSRGetAuthorityName(hSRS, nullptr);
            if (pszAuthCode && pszAuthName && EQUAL(pszAuthName, "EPSG"))
            {
                json_object *poEPSG = json_object_new_int64(atoi(pszAuthCode));
```

apps/gdalinfo_lib.cpp
```diff
@@ -382,8 +394,27 @@ char *GDALInfo( GDALDatasetH hDataset, const GDALInfoOptions *psOptions )
...
+                json_object_object_add(poStacProperties, "proj:epsg", poEPSG);
+            }
+            char * pszProjJson = nullptr;
+            OGRErr result = OSRExportToPROJJSON(hSRS, &pszProjJson, nullptr);
+            if (result == OGRERR_NONE) {
+                json_object *poStacProjJson = json_tokener_parse(pszProjJson);
+                json_object_object_add(poStacProperties, "proj:projjson", poStacProjJson);
```
rouault, line 415:
```suggestion
                json_object_object_add(poStacProperties, "proj:projjson", poStacProjJson);
                CPLFree(pszProjJson);
```

apps/gdalinfo_lib.cpp
```diff
@@ -382,8 +394,27 @@ char *GDALInfo( GDALDatasetH hDataset, const GDALInfoOptions *psOptions )
...
             json_object_object_add(poCoordinateSystem, "wkt", poWkt);
 
+            const char *epsg = OSRGetAuthorityCode(hSRS, nullptr);
+            if (epsg)
+            {
+                json_object *poEPSG = json_object_new_int64(atoi(epsg));
+                json_object_object_add(poStacProperties, "proj:epsg", poEPSG);
+            }
+            char * pszProjJson = nullptr;
+            OGRErr result = OSRExportToPROJJSON(hSRS, &pszProjJson, nullptr);
```
rouault, line 412:
```suggestion
            CPLPushErrorHandler(CPLQuietErrorHandler); // PROJJSON requires PROJ >= 6.2
            OGRErr result = OSRExportToPROJJSON(hSRS, &pszProjJson, nullptr);
            CPLPopErrorHandler();
```

autotest/utilities/test_gdalinfo.py
```diff
@@ -1109,3 +1109,44 @@ def test_gdalinfo_if_option():
...
+    assert "stac" in data
+    stac = data["stac"]
+
+    assert "properties" in stac
+    properties = stac["properties"]
+    assert properties["proj:shape"] == [20, 20]
+
+    assert properties["proj:wkt2"].startswith("PROJCRS")
+    assert properties["proj:epsg"] == 26711
+    assert isinstance(properties["proj:projjson"], dict)
```
rouault, line 1138:
```suggestion
    from osgeo import osr
    if osr.GetPROJVersionMajor() >= 7:
        assert isinstance(properties["proj:projjson"], dict)
```

autotest/utilities/test_gdalinfo.py
```diff
@@ -1109,3 +1109,44 @@ def test_gdalinfo_if_option():
...
+
+def test_gdalinfo_stac_json():
+    if test_cli_utilities.get_gdalinfo_path() is None:
+        pytest.skip()
+
+    ret, _ = gdaltest.runexternal_out_and_err(
+        test_cli_utilities.get_gdalinfo_path()
+        + " -json -proj4 -stats -hist ../gcore/data/byte.tif",
+        encoding="UTF-8",
+    )
```
rouault, line 1126: computing statistics directly on reference files has side effects on other tests, so:
```suggestion
    tmpfilename = "tmp/test_gdalinfo_stac_json.tif"
    shutil.copy("../gcore/data/byte.tif", tmpfilename)
    ret, _ = gdaltest.runexternal_out_and_err(
        test_cli_utilities.get_gdalinfo_path()
        + " -json -proj4 -stats -hist " + tmpfilename,
        encoding="UTF-8",
    )
    gdal.GetDriverByName("GTiff").Delete(tmpfilename)
```

rouault (commented):
Perhaps worth editing doc/source/programs/gdalinfo.rst to mention that starting with GDAL 3.6, STAC information is included in JSON output, and that the -proj4 -stats -hist flags have influence on what is included

rouault:
on datasets with all-empty GDALRasterBand::GetDescription() return, we get a rather dummy "eo:bands": [ {}, {}, .... {} ] . I suppose it is not an issue by itself, right ? (a thought I had is that we could perhaps not include eo:bands if we have only dummy information, but that's perhaps more complication than needed)
Seeing https://github.com/stac-extensions/eo#eocloud_cover, I just wanted to point that in the IMAGERY metadata domain (see https://gdal.org/user/raster_data_model.html#imagery-domain-remote-sensing) there's a CLOUDCOVER metadata domain item that could be used (when present) to fill it. Can be tested with autotest/gcore/data/md_dg.tif sample test data (among others. This is also available in sidecar files of GeoTIFF files of other vendors)

gadomski:
> on datasets with all-empty GDALRasterBand::GetDescription() return, we get a rather dummy "eo:bands": [ {}, {}, .... {} ] . I suppose it is not an issue by itself, right ? (a thought I had is that we could perhaps not include eo:bands if we have only dummy information, but that's perhaps more complication than needed)

Yeah, I went back and forth on this. I ended up keeping it because we're generally "including everything", e.g. multiple `proj:` CRS definitions. My thought it that it doesn't hurt to include, and downstream consumers can choose to drop if empty.

If we end up creating real STAC items at some point, we'll want to exclude empty `eo:bands`.

> Seeing https://github.com/stac-extensions/eo#eocloud_cover, I just wanted to point that in the IMAGERY metadata domain (see https://gdal.org/user/raster_data_model.html#imagery-domain-remote-sensing) there's a CLOUDCOVER metadata domain item that could be used (when present) to fill it. Can be tested with autotest/gcore/data/md_dg.tif sample test data (among others. This is also available in sidecar files of GeoTIFF files of other vendors)

This is good information, and would be very helpful I think. I'll update this PR to include.

apps/gdalinfo_lib.cpp
```diff
@@ -621,6 +664,23 @@ char *GDALInfo( GDALDatasetH hDataset, const GDALInfoOptions *psOptions )
...
+                const char *pszValue =
+                    CPLParseNameValue( papszMetadata[i], &pszKey );
+                if( pszKey && strcmp(pszKey, "CLOUDCOVER") == 0 )
+                {
+                    poValue = json_object_new_int( atoi(pszValue) );
+                    json_object_object_add( poStac, "eo:cloud_cover", poValue );
+                    CPLFree( pszKey );
+                }
+            }
+        }
```
rouault, line 683: can likely be simplified as (untested):
```suggestion
        const char* pszCloudCover = GDALGetMetadataItem(hDataset, "CLOUDCOVER", "IMAGERY");
        if( pszCloudCover )
        {
            poValue = json_object_new_int( atoi(pszCloudCover) );
            json_object_object_add( poStac, "eo:cloud_cover", poValue );
        }
```

gadomski: Done, thanks. I naively messed up the order of the second and third arguments `GDALGetMetadataItem` on my first try :-).

rouault:
merged. thanks!

rouault:
> doesn't seems to be valid as per the `eo` stac extension.

ah right there's a "At least one of the fields must be specified." requirement. Maybe add a automatically generated "description" of the form "Band XX of color interpretation YY" when there's none coming from GDALGetDescription() ? @gadomski do you want to tackle that ?

vincentsarago:
@rouault so in (rio-stac)[] (pr coming soon) what I'll do is:
- set name to `b{index}` 
- set description to either the description metadata or the color interpretation value 

```
      "eo:bands": [
        {
          "name": "b1",
          "description": "gray"
        },
        {
          "name": "b2",
          "description": "undefined"
        },
        {
          "name": "b3",
          "description": "undefined"
        },
        {
          "name": "b4",
          "description": "undefined"
        }
      ]
```

gadomski:
> @gadomski do you want to tackle that ?

Yup, I can take a crack.

PR #6268, opened 2022-08-29
HANA: LaunderName returns wrong value for non-ASCII characters

mrylov:
This PR fixes the following issues in HANA driver:
 - LaunderName function doesn't properly work with non-ASCII characters
 - Geometry column name in OGRHanaDataSource::ICreateLayer is not laundered
 - OGRHanaTableLayer::CreateField function must use Unicode variant of odbc::Connection::prepareStatement

### DIFF
--- a/ogr/ogrsf_frmts/hana/ogrhanadatasource.cpp
+++ b/ogr/ogrsf_frmts/hana/ogrhanadatasource.cpp
@@ -804,7 +804,7 @@ int OGRHanaDataSource::FindLayerByName(const char* name)
 CPLString OGRHanaDataSource::FindSchemaName(const char* objectName)
 {
     auto getSchemaName = [&](const char* sql) {
-        odbc::PreparedStatementRef stmt = conn_->prepareStatement(sql);
+        odbc::PreparedStatementRef stmt = PrepareStatement(sql);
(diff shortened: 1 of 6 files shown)


ogr/ogrsf_frmts/hana/ogrhanadatasource.cpp
```diff
@@ -1305,6 +1300,72 @@ void OGRHanaDataSource::InitializeLayers(
...
+        if ((c & 0x80) == 0x00)
+            return 1;
+        if ((c & 0xE0) == 0xC0)
+            return 2;
+        if ((c & 0xF0) == 0xE0)
+            return 3;
+        if ((c & 0xF8) == 0xF0)
+            return 4;
+
+        throw std::runtime_error("Invalid UTF-8 sequence");
```
rouault, line 1330: are you sure that all calling code paths catch that exception ?

mrylov: My assumption was that we can rely on a check in line 1311.
If this check is reliable than we should never reach line 1330.

rouault: > My assumption was that we can rely on a check in line 1311.

ah right.

PR #6302, opened 2022-09-02
Add vcpkg installation instructions

LilyWangLL:
Gdal is available as a port in [vcpkg](https://github.com/microsoft/vcpkg), a C++ library manager that simplifies installation for gdal and other project dependencies. Documenting the install process here will help users get started by providing a single set of commands to build gdal, ready to be included in their projects.

We also test whether our library ports build in various configurations (dynamic, static) on various platforms (OSX, Linux, Windows: x86, x64, UWP, ARM) to keep a wide coverage for users.

I'm a maintainer for vcpkg, and [here](https://github.com/microsoft/vcpkg/blob/master/ports/gdal/portfile.cmake) is what the port script looks like. We try to keep the library maintained as close as possible to the original library.

### DIFF
--- a/doc/source/download.rst
+++ b/doc/source/download.rst
@@ -256,6 +256,23 @@ available at https://anaconda.org/conda-forge/gdal.
 
     conda install -c conda-forge gdal
 
+Vcpkg
+................................................................................
+
+The gdal port in vcpkg is kept up to date by Microsoft team members and community contributors.
+The url of vcpkg is: https://github.com/Microsoft/vcpkg .
+You can download and install gdal using the vcpkg dependency manager:
+
+::
+
+    git clone https://github.com/Microsoft/vcpkg.git
+    cd vcpkg
+    ./bootstrap-vcpkg.sh  # ./bootstrap-vcpkg.bat for Windows
+    ./vcpkg integrate install
+    ./vcpkg install gdal
+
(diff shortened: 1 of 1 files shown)


doc/source/download.rst
```diff
@@ -256,6 +256,23 @@ available at https://anaconda.org/conda-forge/gdal.
...
+The gdal port in vcpkg is kept up to date by Microsoft team members and community contributors.
+The url of vcpkg is: https://github.com/Microsoft/vcpkg .
+You can download and install gdal using the vcpkg dependency manager:
+
+::
+
+    git clone https://github.com/Microsoft/vcpkg.git
+    cd vcpkg
+    ./bootstrap-vcpkg.sh  # ./bootstrap-vcpkg.bat for Windows
+    ./vcpkg integrate install
```
dg0yt, line 271: AFAICT this command has no effect outside Window/MSVS. What users really need is using the CMake toolchain file provided by vcpkg.
FTR vcpkg will build gdal and all dependencies from source, not just install binary packages.

dg0yt:
> We also test whether our library ports build in various configurations (dynamic, static) on various platforms (OSX, Linux, Windows: x86, x64, UWP, ARM) to keep a wide coverage for users.

It should be noted that vcpkg CI mostly builds libraries. It doesn't run any tests.

PR #6312, opened 2022-09-04
COSAR: handle version 2 files that contain half-foat samples (fixes #6289)

### LINKED ISSUE #6289: TerraSAR COSAR v2 reading as incorrect data type

I expect to be able to correctly read TerraSAR captures that are using the COSAR v2 format (from TanDEM-X CoSSC files).
Instead, the imagery that is read has incorrect I and Q values (likely incorrectly assumed integers rather than half-precision floats).

- Use GDAL to read a TerraSAR image of COSAR v2 format.

Unfortunately I cannot find any public example dataset for this.

I have only been using GDAL through Orfeo Toolbox, so I have only tested this with GDAL v3.2.2.0 for Windows.

## What I believe the issue is

GDAL is reading COSAR v2 imagery as integers rather than half-precision floats.
See: https://tandemx-science.dlr.de/pdfs/TD-GS-PS-3028_TanDEM-X-Experimental-Product-Description_1.2.pdf
To fix this it would mean checking for the COSAR version when reading and change datatype accordingly.
I initially encountered this error with another library which has implemented a fix here:
https://github.com/ngageoint/sarpy/issues/339

rouault:


lsma604:
I have successfully built GDAL with these changes and can confirm that COSAR version 2 files are read correctly.

PR #6313, opened 2022-09-04
Python bindings: use %feature(shadow) to override Python wrappers (fixes #6304)

### LINKED ISSUE #6304: Python bindings: use feature("shadow") to override Python wrapper
When wanting to customize the behaviour of a Python wrapper method, up to now, we used to override it in a %extend{ %pythoncode %{ %} } section. The issue is that in the .py file the method was defined twice, with the override being luckily used, but that is a bit confusing. Using %feature("shadow") is cleaner in that it only emits our override. Cf https://github.com/OSGeo/gdal/pull/6303/commits/f3ef33abd19e7ead546e970f61fb8ff6379e02f7#diff-8fec665d73368152f062f1b0c99c142a495009547d2c5bc47c1891c77b635a93R676

rouault:
When wanting to customize the behaviour of a Python wrapper method, up to now,
we used to override it in a %extend{ %pythoncode %{ %} } section.
The issue is that in the .py file the method was defined twice, with the override
being luckily used, but that is a bit confusing. Using %feature("shadow")
is cleaner in that it only emits our override.


swig/include/python/ogr_python.i
```diff
@@ -103,20 +103,44 @@
...
+        DeleteLayer(DataSource self, layer_name_or_index) -> OGRErr
+
+        Delete the indicated layer from the datasource.
+
+        If this method is supported the ODsCDeleteLayer capability will test
+        TRUE on the OGRDataSource.
+
+        Parameters
+        -----------
+        self:
```
snowman2, line 120: In python, we don't usually document `self`.

swig/include/python/ogr_python.i
```diff
@@ -103,20 +103,44 @@
...
+            handle to the datasource
+        layer_name_or_index:
```
snowman2, line 122:
```suggestion
        value: Union[str, int]
```

swig/include/python/ogr_python.i
```diff
@@ -103,20 +103,44 @@
...
+            index or name of the layer to delete.
+
+        Returns
+        -------
+        OGRErr:
+            OGRERR_NONE on success, or OGRERR_UNSUPPORTED_OPERATION if deleting
```
snowman2, line 128:
```suggestion
            :const:`osgeo.ogr.OGRERR_NONE` on success, or :const:`osgeo.ogr.OGRERR_UNSUPPORTED_OPERATION` if deleting
```

swig/include/python/ogr_python.i
```diff
@@ -103,20 +103,44 @@
             return self.GetLayerByIndex(iLayer)
         else:
             raise TypeError("Input %s is not of String or Int type" % type(iLayer))
+  }
+
+%feature("shadow") DeleteLayer %{
+    def DeleteLayer(self, value) -> "OGRErr":
+        """
+        DeleteLayer(DataSource self, layer_name_or_index) -> OGRErr
+
```
snowman2, line 112: For more details: :c:func:`OGR_DS_DeleteLayer`

swig/include/python/ogr_python.i
```diff
@@ -103,20 +103,44 @@
             return self.GetLayerByIndex(iLayer)
         else:
             raise TypeError("Input %s is not of String or Int type" % type(iLayer))
+  }
+
+%feature("shadow") DeleteLayer %{
+    def DeleteLayer(self, value) -> "OGRErr":
+        """
+        DeleteLayer(DataSource self, layer_name_or_index) -> OGRErr
```
snowman2, line 111:
```suggestion
        DeleteLayer(DataSource self, value) -> OGRErr
```

swig/include/python/ogr_python.i
```diff
@@ -103,20 +103,41 @@
...
+        """
+        DeleteLayer(DataSource self, value) -> OGRErr
+
+        Delete the indicated layer from the datasource.
+
+        For more details: :c:func:`OGR_DS_DeleteLayer`
+
+        Parameters
+        -----------
+            value: str | int
```
snowman2, line 119: Looks like it needs to be dedented here.

rouault: thanks

PR #6322, opened 2022-09-06
CAD/libopencad: avoid use of invalid memory/crash on corrupted files.…

rouault:
… Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=51031

CC @BishopGIS


BishopGIS:
Thanks Even (@rouault) - port changes to libopencad. Tests passed.

PR #6326, opened 2022-09-06
Configuration file: add a ignore-env-vars=yes setting

### LINKED ISSUE #6280: Environment variable to prevent GDAL from using environment variables
~For better or worse, a lot of GDAL's behavior is controlled via environment variables. Credentials, file system layout, application behavior, and available memory are all things that are articulated and controlled via environment variables. Environment variables are really useful for sidestepping application assumptions and altering GDAL's behavior from outside an application. Depending on the environment GDAL is running in, we may not want to allow that.~ 

~The solution to having too many environment variables is to add another, `GDAL_CONFIG_FILE`, that points to a file that GDAL uses to get all of its environment variable information.~

GDAL has `GDAL_CONFIG_FILE` as of 3.3. This ticket is about proposing another environment variable that prevents any other variables being provided via environment, and only being allowed via config file.

rouault:
in a [directives] leading section to disable environment
variables to be used as configuration options (fixes #6280)

### DIFF
--- a/autotest/cpp/test_cpl.cpp
+++ b/autotest/cpp/test_cpl.cpp
@@ -3820,7 +3820,7 @@ namespace tut
         ensure( CPLGetConfigOption("SOME_ENV_VAR_FOR_TEST_CPL_61", nullptr) != nullptr );
 
         VSILFILE* fp = VSIFOpenL("/vsimem/.gdal/gdalrc", "wb");
-        VSIFPrintfL(fp, "[general]\n");
+        VSIFPrintfL(fp, "[directives]\n");
         VSIFPrintfL(fp, "ignore-env-vars=yes\n");
         VSIFPrintfL(fp, "[configoptions]\n");
         VSIFPrintfL(fp, "CONFIG_OPTION_FOR_TEST_CPL_61=BAR\n");
@@ -3837,7 +3837,7 @@ namespace tut
 
         // Reset ignore-env-vars=no
         fp = VSIFOpenL("/vsimem/.gdal/gdalrc", "wb");
-        VSIFPrintfL(fp, "[general]\n");
+        VSIFPrintfL(fp, "[directives]\n");
         VSIFPrintfL(fp, "ignore-env-vars=no\n");
         VSIFPrintfL(fp, "[configoptions]\n");
         VSIFPrintfL(fp, "SOME_ENV_VAR_FOR_TEST_CPL_61=BAR\n");
--- a/autotest/cpp/test_cpl.cpp
+++ b/autotest/cpp/test_cpl.cpp
@@ -3810,6 +3810,50 @@ namespace tut
 #endif
     }
 
+    // Test ignore-env-vars = yes of configuration file
+    template<>
+    template<>
+    void object::test<61>()
+    {
+        char szEnvVar[] = "SOME_ENV_VAR_FOR_TEST_CPL_61=FOO";
+        putenv(szEnvVar);
+        ensure( CPLGetConfigOption("SOME_ENV_VAR_FOR_TEST_CPL_61", nullptr) != nullptr );
+
+        VSILFILE* fp = VSIFOpenL("/vsimem/.gdal/gdalrc", "wb");
+        VSIFPrintfL(fp, "[general]\n");
+        VSIFPrintfL(fp, "ignore-env-vars=yes\n");
+        VSIFPrintfL(fp, "[configoptions]\n");
+        VSIFPrintfL(fp, "CONFIG_OPTION_FOR_TEST_CPL_61=BAR\n");
+        VSIFCloseL(fp);
+
+        // Load configuration file
+        CPLLoadConfigOptionsFromFile("/vsimem/.gdal/gdalrc", false);
+
+        // Check that reading configuration option works
+        ensure( EQUAL(CPLGetConfigOption("CONFIG_OPTION_FOR_TEST_CPL_61", ""), "BAR") );
+
+        // Check that environment variables are not read as configuration options
+        ensure( CPLGetConfigOption("SOME_ENV_VAR_FOR_TEST_CPL_61", nullptr) == nullptr );
+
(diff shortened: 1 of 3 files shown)


sgillies:
@rouault @hobu doesn't this go against the common convention in software that environment variables ultimately override file configuration?

rouault:
> doesn't this go against the common convention in software that environment variables ultimately override file configuration?

that remains the default behavior (Actually I messed up my initial commit message. Fixing it to "Configuration file: add a ignore-env-vars=yes setting")

You can turn that off with the new ignore-env-vars=yes setting in the configuration file. The alternative was to have a environment variable telling not to trust other environment variables (GDAL_IGNORE_ENV_VARS=YES), but that sounded a bit like "I'm not a lyer" paradox.

@hobu 's motivation for such an option is a way to make sure nothing in your environment leaks in subtle way to help having more reproducible workflows.

sgillies:
I'm concerned that we're trading one problem for another: users not knowing that ignore-env-vars=yes is set in their config and pulling their hair out when environment variables have no effect. Isn't the answer to more reproducible deployments of GDAL better done at another level? "Just" deploy containers or VMs with config that's kept under revision control, for example.

rouault:
> users not knowing that ignore-env-vars=yes is set in their config

We could possibly add an explicit debug message when ignore-env-vars=yes : CPLDebug("GDAL", "Environment variables *not* taken into account as GDAL configuration options due to ignore-env-vars=yes setting in configuration file")

hobu:
> Isn't the answer to more reproducible deployments of GDAL better done at another level?

Admittedly, this particular request might be kind of special, but it is not only for better reproducibility. It's also to attempt to lessen the attack surface that GDAL provides with all of its innards just hanging out in the environment that can be poked and inspected at will. 

GDAL has already taken on sticking stuff like credentials in environment variables and we can't simply back away from it. In aggressive logging situations, this is leakage that can have undesirable consequences. My thought is to give people with the need the ability to go an extra mile to shut off greedy environment variable consumption. 

> We could possibly add an explicit debug message when ignore-env-vars=yes

I like this idea. If you're doing config files AND environment variables, you're going to get noisy logs in this situation. If you don't want noisy logs, don't set the variables and use the config like you intended 😄

jratike80:
> users not knowing that ignore-env-vars=yes is set in their config and pulling their hair out when environment variables have no effect

On the other hand, users not knowing that the env var that they set may have an an effect on a bunch of existing scripts and processes can make other people to pull their hair out.

How I think is that the config file is like a shortcut for giving one or more `--config key value` parameters on the command line - at least when the location of the config file is given as `--config GDAL_CONFIG_FILE my_config_file_001`. In that case I would be surprised if config file / config options would not override environment.
What actually happens now if there is an env var that sets for example GDAL_HTTP_PROXY. `--config GDAL_HTTP_PROXY another_proxy` given on a command line overrides the env, but how about giving `--config GDAL_CONFIG_FILE file_that_contains_GDAL_HTTP_PROXY`? Without this new options the env wins, right?

rouault:
> but how about giving `--config GDAL_CONFIG_FILE file_that_contains_GDAL_HTTP_PROXY`? Without this new options the env wins, right?

yes, just tested something similar

rouault:
with latest commit:

with gdalrc containing
```
[general]
ignore-env-vars = yes
[configoptions]
GDAL_DATA=foo
```
and GDAL_DATA and GDAL_DRIVER_PATH set as environment variables
```
$ gdalinfo byte.tif --config GDAL_CONFIG_FILE gdalrc --debug on
CPL: Loading configuration from gdalrc
CPL: Setting configuration option GDAL_DATA=foo
CPL: Ignoring environment variable GDAL_DATA=/home/even/gdal/gdal/data because of ignore-env-vars=yes setting in configuration file
CPL: Ignoring environment variable GDAL_DRIVER_PATH=/home/even/gdal/gdal/build_cmake/gdalplugins because of ignore-env-vars=yes setting in configuration file
```

with gdalrc containing
```
[configoptions]
GDAL_DATA=foo
```
and GDAL_DATA and GDAL_DRIVER_PATH set as environment variables
```
$ gdalinfo byte.tif --config GDAL_CONFIG_FILE gdalrc --debug on
CPL: Loading configuration from gdalrc
CPL: Ignoring configuration option GDAL_DATA=foo from configuration file as it is already set as an environment variable
```

sgillies:
I don't love `ignore-env-vars = yes`. @hobu it doesn't prevent the environment from leaking, for example. And I think it's part of a trend of GDAL becoming harder to explain and reason about. But am not going to lay down in front of the train.

To be helpful: `[configoptions]` is more readable as `[config-options]` (or underscore version). `[general]` might become `[directives]` or something? Seeing as the one thing in there has an out-sized effect on the makeup of GDAL's environment.

hobu:
> @hobu it doesn't prevent the environment from leaking, for example.

No, the user would still have to put in the work to move everything to config variables. With ignore-env-vars=yes set and all GDAL environment variables set in the config file, there's no way those things leak in a dump of the environment (like a logger calling `env`). Something in-process could still ask GDAL for its environment variables (taken from the config in this configuration).

> And I think it's part of a trend of GDAL becoming harder to explain and reason about. But am not going to lay down in front of the train.

GDAL's use of environment variables to configure itself and drive behavior has been a longstanding "leak" in the abstraction. We have environment variables that provide credentials. Other ones tell GDAL internals to behave this way or another. Behavior deprecation is driven by them too. They have thread-local behavior sensitivity as well. It's a closet that is so wedged full of stuff that no one wants to open it unless they have to. It helps that https://gdal.org/user/configoptions.html exists now and you can at least see what is in that closet.

I want a door lock on the thing. When I'm using Lambda or Azure Functions or Kube I want the ability to lock down my function so that none of my function's behavior could be altered by someone external to me sticking a GDAL-sensitive environment variable onto my runtime environment. That behavior does not need to be a default, and it is reasonable that I have to do extra things to enable it.

PR #6347, opened 2022-09-13
Fix build on Ubuntu 18.04 with PDF driver due to broken Poppler security update

rouault:
Workaround https://bugs.launchpad.net/ubuntu/+source/poppler/+bug/1989434 which is an issue with broken security update on Ubuntu 18.04 of poppler 0.62.0-2ubuntu2.13
If we detect that gmem.h includes GooCheckedOps.h but that one is not installed, create a dummy one in a temporary directory To be removed once https://bugs.launchpad.net/ubuntu/+source/poppler/+bug/1989434 is fixed

### DIFF
--- a/frmts/pdf/CMakeLists.txt
+++ b/frmts/pdf/CMakeLists.txt
@@ -34,6 +34,25 @@ if (GDAL_USE_POPPLER)
   endif ()
   target_compile_definitions(gdal_PDF PRIVATE -DHAVE_POPPLER -DPOPPLER_MAJOR_VERSION=${Poppler_VERSION_MAJOR}
                                               -DPOPPLER_MINOR_VERSION=${Poppler_VERSION_MINOR})
+
+  # Workaround https://bugs.launchpad.net/ubuntu/+source/poppler/+bug/1989434
+  # which is an issue with broken security update on Ubuntu 18.04 of
+  # poppler 0.62.0-2ubuntu2.13
+  # If we detect that gmem.h includes GooCheckedOps.h but that one is not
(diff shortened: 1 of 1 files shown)


frmts/pdf/CMakeLists.txt
```diff
@@ -34,6 +34,25 @@ if (GDAL_USE_POPPLER)
...
+
+  # Workaround https://bugs.launchpad.net/ubuntu/+source/poppler/+bug/1989434
+  # which is an issue with broken security update on Ubuntu 18.04 of
+  # poppler 0.62.0-2ubuntu2.13
+  # If we detect that gmem.h includes GooCheckedOps.h but that one is not
+  # installed, create a dummy one in a temporary directory
+  # To be removed once https://bugs.launchpad.net/ubuntu/+source/poppler/+bug/1989434 is fixed
+  if ("${CMAKE_SYSTEM_NAME}" STREQUAL "Linux" AND
+      EXISTS /usr/include/poppler/goo/gmem.h AND
+      NOT EXISTS /usr/include/poppler/goo/GooCheckedOps.h)
```
dg0yt, line 46: This tests files which may or may not be related to the actual `Poppler_INCLUDE_DIRS`.
(Why build the latest and greatest GDAL with an outdated poppler? I do build on Ubuntu 18.04, with vcpkg. GDAL should ignore the system version if I build with another poppler installation.)

rouault: good point. adressed in https://github.com/OSGeo/gdal/pull/6364

PR #6364, opened 2022-09-14
CMake: Poppler: tighten fix of PR #6347 for broken 18.04 package

rouault:
addressed comment https://github.com/OSGeo/gdal/pull/6347/files#r971059533


dg0yt:
LGTM. Eventual, `Poppler_INCLUDE_DIRS` should be considered a list everywhere, but if it detects the broken configuration, the current condition is okay.

PR #6393, opened 2022-09-19
GTIFF+JXL: support more than 4 bands in INTERLEAVE=PIXEL mode

tbonfort:


tbonfort:
@rouault I would like to robustify this code a bit more. Unfortunately I am unable to access the band color interpretation from tif_jxl.c , which is where the decision needs to be made to encode each band as an extra channel rather than packed rgb, and whether to also embed the alpha channel. Do you have any recommendations as to how/where I could access that info?

frmts/gtiff/geotiff.cpp
```diff
@@ -18797,6 +18797,40 @@ GTiffDataset::CreateCopy( const char * pszFilename, GDALDataset *poSrcDS,
...
 
+#ifdef HAVE_JXL_notused
+/* -------------------------------------------------------------------- */
+/*      If the output is lossy JXL compressed, and the input is not     */
+/*      RGB(A) make sure we note that.                                  */
+/* -------------------------------------------------------------------- */
+
+    if( l_nCompression == COMPRESSION_JXL )
+    {
+        bool l_bJXLLossless = GTiffGetJXLLossless(papszOptions);
```
tbonfort, line 18808: TODO: this check should only happen for INTERLEAVE=PIXEL

rouault:
> Unfortunately I am unable to access the band color interpretation from tif_jxl.c , which is where the decision needs to be made to encode each band as an extra channel rather than packed rgb, and whether to also embed the alpha channel.

If you want to handle the situation of bands in order R, G, B, channel4, channel5, channel6, ... where channel4/5/6 are alpha or non alpha, you can use the uint16_t tif->tif_dir.td_sampleinfo[] array, when present, which has normally tif->tif_dir.td_extrasamples values which should be equal to tif->tif_dir.td_samplesperpixel - _TIFFGetMaxColorChannels(tif->tif_dir.td_photometric).  _TIFFGetMaxColorChannels() returns 3 for photometric=RGB
That is td_sampleinfo[0] contains a value EXTRASAMPLE_UNSPECIFIED/EXTRASAMPLE_ASSOCALPHA/EXTRASAMPLE_UNASSALPHA for channel4, td_sampleinfo[1] for channel5, etc.

Similarly for photometric=MinIsBlack, _TIFFGetMaxColorChannels() returns 1 and td_sampleinfo[] starts at channel 2.

if you would want to handle situations like NIR, R, G, B, channel5, channel6, ..., then GDAL will use photometric=MinIsBlack, and the above would still work (ie td_sampleinfo[] will start at band 2).

rouault:
> From my understanding of
> 
> https://github.com/OSGeo/gdal/blob/da8e335e3eb5441ae4288767f59c36936a1bca5f/frmts/gtiff/geotiff.cpp#L18767

hum it looks to me to be rather a defect / historical behavior of CreateCopy() that should likely be fixed.

The driver SetColorInterpretation() works properly when called on a newly created file:
```
even@even-ThinkPad-P15v-Gen-1:~/gdal/gdal/build_cmake$ python
Python 3.8.10 (default, Jun 22 2022, 20:18:18) 
[GCC 9.4.0] on linux
>>> from osgeo import gdal
>>> ds = gdal.GetDriverByName('GTiff').Create('test.tif', 1, 1, 5)
>>> ds.GetRasterBand(4).SetColorInterpretation(gdal.GCI_AlphaBand)
0
>>> ds = None
>>> 
even@even-ThinkPad-P15v-Gen-1:~/gdal/gdal/build_cmake$ gdalinfo test.tif
Driver: GTiff/GeoTIFF
Files: test.tif
Size is 1, 1
Image Structure Metadata:
  INTERLEAVE=PIXEL
Corner Coordinates:
Upper Left  (    0.0,    0.0)
Lower Left  (    0.0,    1.0)
Upper Right (    1.0,    0.0)
Lower Right (    1.0,    1.0)
Center      (    0.5,    0.5)
Band 1 Block=1x1 Type=Byte, ColorInterp=Gray
Band 2 Block=1x1 Type=Byte, ColorInterp=Undefined
Band 3 Block=1x1 Type=Byte, ColorInterp=Undefined
Band 4 Block=1x1 Type=Byte, ColorInterp=Alpha
Band 5 Block=1x1 Type=Byte, ColorInterp=Undefined
even@even-ThinkPad-P15v-Gen-1:~/gdal/gdal/build_cmake$ tiffdump test.tif
test.tif:
Magic: 0x4949 <little-endian> Version: 0x2a <ClassicTIFF>
Directory 0: offset 8 (0x8) next 0 (0)
ImageWidth (256) SHORT (3) 1<1>
ImageLength (257) SHORT (3) 1<1>
BitsPerSample (258) SHORT (3) 5<8 8 8 8 8>
Compression (259) SHORT (3) 1<1>
Photometric (262) SHORT (3) 1<1>
StripOffsets (273) LONG (4) 1<186>
SamplesPerPixel (277) SHORT (3) 1<5>
RowsPerStrip (278) SHORT (3) 1<1>
StripByteCounts (279) LONG (4) 1<5>
PlanarConfig (284) SHORT (3) 1<1>
ExtraSamples (338) SHORT (3) 4<0 0 2 0>
SampleFormat (339) SHORT (3) 5<1 1 1 1 1>
```

tbonfort:
> hum it looks to me to be rather a defect / historical behavior of CreateCopy() that should likely be fixed

proposed fix in ffcde32

tbonfort:
ffcde32 fix moved to its own PR: #6395

rouault:
unrelated build issue in build-linux-ubuntu-focal will hopefully be fixed per https://github.com/OSGeo/gdal/pull/6402

autotest/gcore/tiff_write.py
```diff
@@ -10256,5 +10257,72 @@ def test_tiff_write_createcopy_alpha_not_in_last_band(options):
...
+    if md["DMD_CREATIONOPTIONLIST"].find("JXL") == -1:
+        pytest.skip()
+
+    
+    src_ds = gdal.GetDriverByName("MEM").Create("", 64, 64, 6)
+    for b in range(6):
+        bnd = src_ds.GetRasterBand(b+1)
+        data = numpy.random.bytes(64*64)
+        bnd.WriteRaster(0,0,64,64,data)
+        #bnd.Fill(b+1)
```
tbonfort, line 10275: @rouault if I use `bnd.Fill` instead of numpy.random + writeraster, the checksum of the band will always be 0. I would prefer to not have to import numpy here and rely on the checksum varying depending on the fill value, is that possible?

rouault: weird that checksum is 0. maybe try to add bnd.FlushCache() after bnd.Fill() ? But that shouldn't be necessary

rouault: Works for me without FlushCache():
```
$ python -c "from osgeo import gdal; ds = gdal.GetDriverByName('MEM').Create('', 64, 64); ds.GetRasterBand(1).Fill(1); print(ds.GetRasterBand(1).Checksum())"
4096
```

rouault:
for automatic local run of black, flake8 etc, see https://github.com/OSGeo/gdal/blob/master/CONTRIBUTING.md#build-gdal-cmake:
```
Setup pre-commit:

python -m pip install pre-commit
pre-commit install
```

tbonfort:
Ready for review. I still plan on investigating if/how we can support non-default NBITS/td_bitspersample with JXL (which would be required in order to properly support lossy 16bit data that doesn't span [0-65535]), but will open another issue/pr for that.

frmts/gtiff/tif_jxl.c
```diff
@@ -326,6 +411,31 @@ JXLPreDecode(TIFF* tif, uint16_t s)
...
+        if( nFirstExtraChannel < info.num_extra_channels ){
+            //first reorder the main buffer
+            int nMainChannels = bAlphaEmbedded?info.num_color_channels+1:info.num_color_channels;
+            int mainPixSize = nMainChannels*nBytesPerSample;
+            int fullPixSize = td->td_samplesperpixel*nBytesPerSample;
+            unsigned int outOff = sp->uncompressed_size - fullPixSize;
+            int inOff = main_buffer_size - mainPixSize;
+            for(;inOff>=0;inOff-=mainPixSize,outOff-=fullPixSize)
+            {
+                memcpy(sp->uncompressed_buffer+outOff,sp->uncompressed_buffer+inOff,mainPixSize);
```
rouault, line 423: there might be some perf hits to have memcpy() in a loop with the size argument being a variable and not a constant. Given that mainPixSize will generally be in (1, 2, 3, 4, 6, 8) range   ( nbands = [1,3,4] * nBytesPerSample = [1,2] ) ), it might be worth to have  particularized for loops with mainPixSize hardcoded to those values (and the current loop as a fallback)
Same for below loop for the extra channels

frmts/gtiff/tif_jxl.c
```diff
@@ -599,23 +721,105 @@ JXLPostEncode(TIFF* tif)
...
+#ifdef HAVE_JxlExtraChannels
+        if (td->td_planarconfig == PLANARCONFIG_CONTIG &&
+            (basic_info.num_extra_channels > 1 ||
+             (basic_info.num_extra_channels == 1 && !bAlphaEmbedded)))
+        {
+            main_size = (sp->uncompressed_size / td->td_samplesperpixel);
+            int nMainChannels = basic_info.num_color_channels;
+            if(bAlphaEmbedded) nMainChannels++;
+            main_size *= nMainChannels;
+            main_buffer = _TIFFmalloc(main_size);
```
rouault, line 746: should be tested against null

frmts/gtiff/tif_jxl.c
```diff
@@ -599,23 +721,105 @@ JXLPostEncode(TIFF* tif)
...
+            extra_channel_info.exponent_bits_per_sample = basic_info.exponent_bits_per_sample;
+            for(int iChannel=nMainChannels; iChannel<td->td_samplesperpixel; iChannel++)
+            {
+                if (JXL_ENC_SUCCESS != JxlEncoderSetExtraChannelInfo(
+                    enc,
+                    (bAlphaEmbedded)?iChannel-nMainChannels+1:iChannel-nMainChannels, &extra_channel_info))
+                {
+                    TIFFErrorExt(tif->tif_clientdata, module,
+                                 "JxlEncoderSetExtraChannelInfo(%d) failed",iChannel);
+                    JxlEncoderDestroy(enc);
```
rouault, line 767:
```suggestion
                    JxlEncoderDestroy(enc);
                    _TIFFfree(main_buffer);
```

frmts/gtiff/tif_jxl.c
```diff
@@ -599,23 +721,105 @@ JXLPostEncode(TIFF* tif)
...
+        if (td->td_planarconfig == PLANARCONFIG_CONTIG &&
+            (basic_info.num_extra_channels > 1 ||
+             (basic_info.num_extra_channels == 1 && !bAlphaEmbedded)))
+        {
+            main_size = (sp->uncompressed_size / td->td_samplesperpixel);
+            int nMainChannels = basic_info.num_color_channels;
+            if(bAlphaEmbedded) nMainChannels++;
+            main_size *= nMainChannels;
+            main_buffer = _TIFFmalloc(main_size);
+            int outChunkSize = td->td_bitspersample / 8 * nMainChannels;
```
rouault, line 747:
```suggestion
            int outChunkSize = nBytesPerSample * nMainChannels;
```

frmts/gtiff/tif_jxl.c
```diff
@@ -599,23 +721,105 @@ JXLPostEncode(TIFF* tif)
...
         }
+
+#ifdef HAVE_JxlExtraChannels
+        if (td->td_planarconfig == PLANARCONFIG_CONTIG &&
+            (basic_info.num_extra_channels > 1 ||
+             (basic_info.num_extra_channels == 1 && !bAlphaEmbedded)))
+        {
+            int nMainChannels = basic_info.num_color_channels;
+            if(bAlphaEmbedded) nMainChannels++;
+            int extra_channel_size = (sp->uncompressed_size / td->td_samplesperpixel);
```
rouault, line 795: this is actually the total number of samples per pixel. should probably be
```suggestion
            int extra_channel_size = td->td_samplesperpixel - nMainChannels;
```

frmts/gtiff/tif_jxl.c
```diff
@@ -599,23 +721,105 @@ JXLPostEncode(TIFF* tif)
...
+            uint8_t *extra_channel_buffer = _TIFFmalloc(extra_channel_size);
```
rouault, line 796: should be tested against NULL

frmts/gtiff/tif_jxl.c
```diff
@@ -599,23 +721,105 @@ JXLPostEncode(TIFF* tif)
...
+            main_size = (sp->uncompressed_size / td->td_samplesperpixel);
+            int nMainChannels = basic_info.num_color_channels;
+            if(bAlphaEmbedded) nMainChannels++;
+            main_size *= nMainChannels;
+            main_buffer = _TIFFmalloc(main_size);
+            int outChunkSize = td->td_bitspersample / 8 * nMainChannels;
+            int inStep = nBytesPerSample * td->td_samplesperpixel;
+            uint8_t *cur_outbuffer=main_buffer;
+            uint8_t *cur_inbuffer=sp->uncompressed_buffer;
+            for( ; cur_outbuffer-main_buffer<main_size ; cur_outbuffer+=outChunkSize, cur_inbuffer+=inStep  ) {
```
rouault, line 751: loop could be particularized for outChunckSize in [1,2,3,4,6,8]

frmts/gtiff/tif_jxl.c
```diff
@@ -599,23 +721,105 @@ JXLPostEncode(TIFF* tif)
...
+            if(bAlphaEmbedded) nMainChannels++;
+            int extra_channel_size = (sp->uncompressed_size / td->td_samplesperpixel);
+            uint8_t *extra_channel_buffer = _TIFFmalloc(extra_channel_size);
+            int inStep = nBytesPerSample * td->td_samplesperpixel;
+            int outStep = nBytesPerSample;
+            for(int iChannel=nMainChannels; iChannel<td->td_samplesperpixel; iChannel++)
+            {
+                uint8_t *cur_outbuffer = extra_channel_buffer;
+                uint8_t *cur_inbuffer = sp->uncompressed_buffer+iChannel*outStep;
+                for (; cur_outbuffer - extra_channel_buffer < extra_channel_size; cur_outbuffer += outStep, cur_inbuffer += inStep)
```
rouault, line 803: loop could be particularized for outStep in [1,2]

frmts/gtiff/tif_jxl.c
tbonfort: extra_channel_size is the number of bytes needed to hold a single channel

rouault: makes sense. ignore my comment

tbonfort:
I'm inclined to defer implementing your remaining review comments concerning memcpy optimizations until we've worked out how to cleanly support nbits. We can easily add them later once the code has stabilised

PR #6395, opened 2022-09-19
GTiff: fix marking alpha channels that are not the last one

tbonfort:


rouault:
@tbonfort I've added a commit with a new test case

tbonfort:
thanks Even, I also added tests to ensure the TIFFTAG_EXTRASAMPLES tag is correctly set

PR #6398, opened 2022-09-20
gdal_target_interfaces must recurse

dg0yt:
## What does this PR do?

This PR fixes build problems which appeared in vcpkg. `gdal_target_interfaces` is a helper function which is used to implement an alternative to plain `target_link_libraries`. Similar to that CMake function, it must recurse into `INTERFACE_LINK_LIBRARIES` in order to (more) properly pass transitive usage requirements, or builds may fail.

This PR doesn't consider generator expressions, but it is usefull enough as is. 

vcpkg use case: To fix static linkage, instead of taking pkg-config files only for hinting, a patch in vcpkg lets `pkg_check_modules` create imported targets *with all usage requirements*, and then adds a interface library with the name used in gdal. This is most beneficial for dependencies with configuration-dependent static dependencies (freexl, openjpeg, poppler, spatialite). Cf.  
https://github.com/microsoft/vcpkg/blob/master/ports/gdal/find-link-libraries.patch

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments


rouault:
> https://github.com/microsoft/vcpkg/blob/master/ports/gdal/find-link-libraries.patch

wouldn't make that one make sense to be upstreamed too ? I guess some extra precautions would be needed, like checking that the _LIBRARY and _INCLUDE_DIR variable haven't been explicitly set before using the pkgconfig output

dg0yt:
> wouldn't make that one make sense to be upstreamed too ? I guess some extra precautions would be needed, like checking that the _LIBRARY and _INCLUDE_DIR variable haven't been explicitly set before using the pkgconfig output

Yes. It is the extra precautions which prevented a higher priority in my personal backlog.
- The `IMPORTED_TARGET` option exists since CMake 3.6. So this is okay for GDAL. The quality probably improves with newer releases.
- There is the risk of invalid pc files. In vcpkg, I used them already even for nmake builds, and it took a while to get them all right... But OTOH, this is a problem mostly for static builds, and they would be broken already now.
   As another mitigation, we might check a `GDAL_DISABLE_PKGCONFIG_<Module>` variable.

PR #6406, opened 2022-09-20
DOC: Adding --tmscompatible option to gdal2tiles.py documentation

MCBama:
`--tmscompatible` has been a valid option for a while and was already described in the `gdal2tiles.py` help section but was not explicitly defined in the online documentation. This simply corrects that oversight.


rouault:
thanks

PR #6412, opened 2022-09-21
JPEGXL & JXL-in-TIFF: preserve Alpha color interpretation when the Alpha band does not immediately follow color bands

rouault:
(e.g. R,G,B,undefined,Alpha)

+ JXL-in-TIFF: check on reading consistency between JXL extra channel info and ExtraSamples TIFF tag

### DIFF
--- a/frmts/gtiff/tif_jxl.c
+++ b/frmts/gtiff/tif_jxl.c
@@ -737,9 +737,9 @@ JXLPostEncode(TIFF* tif)
         
         uint8_t *main_buffer = sp->uncompressed_buffer;
         unsigned int main_size = sp->uncompressed_size;
-        int nBytesPerSample = GetJXLDataTypeSize(format.data_type);
 
 #ifdef HAVE_JxlExtraChannels
+        int nBytesPerSample = GetJXLDataTypeSize(format.data_type);
         if (td->td_planarconfig == PLANARCONFIG_CONTIG &&
             (basic_info.num_extra_channels > 1 ||
              (basic_info.num_extra_channels == 1 && !bAlphaEmbedded)))
--- a/frmts/gtiff/tif_jxl.c
+++ b/frmts/gtiff/tif_jxl.c
@@ -335,7 +335,34 @@ JXLPreDecode(TIFF* tif, uint16_t s)
         format.data_type = jxlDataType;
         format.endianness = JXL_NATIVE_ENDIAN;
         format.align = 0;
-        int bAlphaEmbedded = (info.alpha_bits!=0);
+        // alpha_bits is set even for a gray, gray, Alpha, gray, gray
+        // or for R, G, B, undefined, Alpha
+        // Probably a defect of libjxl: https://github.com/libjxl/libjxl/issues/1773
+        // So for num_color_channels==3, num_color_channels > 1 and
+        // alpha_bits != 0, get information of the first extra channel to
+        // check if it is alpha, to detect R, G, B, Alpha, undefined.
+        // Note: there's no difference in the codestream if writing RGBAU
+        // as num_channels == 3 with 2 extra channels the first one being
+        // explicitly set to alpha, or with num_channels == 4.
+        int bAlphaEmbedded = 0;
+        if( info.alpha_bits !=0 )
+        {
+            if( (info.num_color_channels==3 || info.num_color_channels==1) &&
+                (info.num_extra_channels==1) )
+            {
+                bAlphaEmbedded = 1;
+            }
+            else if( info.num_color_channels==3 && info.num_extra_channels > 1 )
+            {
+                JxlExtraChannelInfo extra_channel_info;
+                memset(&extra_channel_info, 0, sizeof(extra_channel_info));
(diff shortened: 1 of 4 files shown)


rouault:
CC @tbonfort

tbonfort:
please let me review this before merging the tif part

frmts/gtiff/tif_jxl.c
```diff
@@ -335,7 +335,15 @@ JXLPreDecode(TIFF* tif, uint16_t s)
         format.data_type = jxlDataType;
         format.endianness = JXL_NATIVE_ENDIAN;
         format.align = 0;
-        int bAlphaEmbedded = (info.alpha_bits!=0);
+        // alpha_bits is set even for a gray, gray, Alpha, gray, gray
+        // or for R, G, B, undefined, Alpha
+        // It is only reliable if num_extra_channels == 1
+        // Probably a defect of libjxl
+        // Anyway it is safe to read R,G,B,A,undefined as RGB interleaved
+        // and Alpha and undefined as separate
```
tbonfort, line 343: Do you have any pointers to this defect ?
From my testing when creating a jxl file like this:
`gdal_translate  -b 1 -b 2 -b 3 -b 4 -b 5 src.tif  -colorinterp red,green,blue,alpha,undefined -co TILED=YES -co COMPRESS=JXL rgbai.tif`
The produced jxl codestream has embedded alpha, but does not pass this added test. Here is the corresponding basic_info:
```bits_per_sample = 16,
num_color_channels = 3,
num_extra_channels = 2, 
alpha_bits = 16,   
alpha_premultiplied = 0
```

frmts/gtiff/tif_jxl.c
```diff
@@ -369,6 +380,58 @@ JXLPreDecode(TIFF* tif, uint16_t s)
                     _TIFFfree(extra_channel_buffer);
                     return 0;
                 }
+
```
tbonfort, line 383: I was worried that these extra tests might raise warnings on tifjxl files created before we began supporting more than 4 bands, as in that case the jxl codestream had marked the 4th band as alpha whereas that might not have been the case in the tif's ExtraSample[0]. As it happens this series of validations is not activated for this configuration, as `nFirstExtraChannel < info.num_extra_channels` will be false.
I'd like to add such an "old style" jxl tif to the autotests, how can I write a test to ensure that `TIFFWarningExt` was not called?

frmts/gtiff/tif_jxl.c
rouault: > Do you have any pointers to this defect ?

I've just reported it as https://github.com/libjxl/libjxl/issues/1773 and added an extra test to detect RGBAxxxx in an explicit way and be able to use the interleaved decoding for the first 4 bands

frmts/gtiff/tif_jxl.c
rouault: > how can I write a test to ensure that TIFFWarningExt was not called?

similarly to how I've enhanced the test you added yesterday
```
                    gdal.ErrorReset()
                    got_cs = ds.GetRasterBand(idx + 1).Checksum()
                    assert gdal.GetLastErrorMsg() == ""
```

frmts/gtiff/tif_jxl.c
```diff
@@ -335,7 +335,29 @@ JXLPreDecode(TIFF* tif, uint16_t s)
...
+        // So for num_color_channels==3, num_color_channels > 1 and
+        // alpha_bits != 0, get information of the first extra channel to
+        // check if it is alpha, to detect R, G, B, Alpha, undefined.
+        // Note: there's no difference in the codestream if writing RGBAU
+        // as num_channels == 3 with 2 extra channels the first one being
+        // explicitly set to alpha, or with num_channels == 4.
+        int bAlphaEmbedded = (info.alpha_bits!=0) &&
+                             (info.num_color_channels==3 || info.num_color_channels==1) &&
+                             (info.num_extra_channels==1);
+        if( info.num_color_channels==3 && info.num_color_channels > 1 &&
```
tbonfort, line 350:
```suggestion
        if( info.num_color_channels==3 && info.num_extra_channels > 1 &&
```

tbonfort:
From the discussion about how band ordering in jxl should not be interpreted by index but rather by name, which implies that the jxl encoder /may/ be allowed to reorder the bands if we set their type to something other than JXL_CHANNEL_OPTIONAL, I'm not overly confident that we should be setting/checking that type in the tif codec, as the source of truth for the band interpretation is the tif header, not the jxl metadata. What do you think?

rouault:
@tbonfort Given the last answer in the libjxl ticket "-Would the encoder be allowed to reorder bands ? - I don't think the encoder should be allowed to do that, and the current libjxl encoder certainly doesn't.", it seems to me that this PR should be safe. That said, I can disable the part in tif_jxl.c that checks the consistency of the TIFF ExtraSample tag with the type of the libjxl extra channels. It seems to me that giving an hint to the encoder that a channel is alpha should help it to do potentially a better job.

tbonfort:
@rouault agreed. LGTM once my suggestion correcting a typo is applied.

frmts/gtiff/tif_jxl.c
rouault: done

PR #6415, opened 2022-09-22
gdal_rasterize: add an ALL_INTERSECTED option (fixes #6414)

### LINKED ISSUE #6414: gdal_rasterize - Missing an "ALL_INTERSECTED" option

In gdal_rasterize, ALL_TOUCHED option includes neighboor pixels on the right and the bottom for segments which are already on the raster grid. It's fine, because these pixels are "touched", especially in case of line rasterization, but it's not a perfect conservative method for polygon as new pixels are included. An ALL_INTERSECTED option should be added for this special case.

Data : [extract_data.zip](https://github.com/OSGeo/gdal/files/9624710/extract_data.zip)

I produce masks on raster (ie: shadow), but these masks doesn't always fit the entity (origin_mask.shp)

image

It's easier to modify vector data, so I vectorize my mask and move vertex using QGIS (tweaked_mask.shp)

image

And I want to snap my mask to the raster grid so I use gdal_rasterize. I want to keep every pixel that are intersected by the polygon. I tested the ALL_TOUCHED option :

gdal_rasterize -burn 1 -tr 10. 10. -a_nodata 0.0 -te 709680.0 1612100.0 711020.0 1613440.0 -at 'extract_data/tweaked_mask.shp'  'extract_data/mask_at.tif'

But for all segments which are already snapped on the grid, this option burn pixels on the right (for vertical segment) and the bottom (for horizontal pixel):

image

mask_at.tif

----

I expect to get : 

image

mask_ai.tif

## Operating system

Ubuntu 22.04

GDAL 3.4.1, released 2021/12/27 from http://fr.archive.ubuntu.com/ubuntu/

SebastienPeillet:
## What does this PR do?

Add an ALL_INTERSECTED option in gdal_rasterize with a `-ai` argument.

Using a vector entity snapped on the output grid, the ALL_TOUCHED option includes pixels on the right and the bottom  (see #6414). The new option allow a conservative rasterization method without this consequence.

I copied `GDALdllImageLineAllTouched()` in a new function `GDALdllImageLineAllIntersected()` but there is only a few changes, tell me if you think it is better to only add an argument to `GDALdllImageLineAllTouched()` to distinguish the two methods.

## Tasklist

 - [x] Add test case(s)
 - [x] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/alg/llrasterize.cpp
+++ b/alg/llrasterize.cpp
@@ -504,10 +504,11 @@ GDALdllImageLineAllTouched( int nRasterXSize, int nRasterYSize,
             // Special case for horizontal lines.
             if( floor(dfY) == floor(dfYEnd) || fabs(dfY - dfYEnd) < .01 )
             {
-                double dYOffset = std::abs(std::fmod(dfY, 1.));
-                double dYEndOffset = std::abs(std::fmod(dfYEnd, 1.));
-                if ( bIntersectOnly && abs(dYOffset - .5) > .49 &&  abs(dYEndOffset - .5) > .49)
-                    continue;
+                if ( bIntersectOnly )
+                {
+                    if( std::abs(dfY - std::round(dfY)) < 0.01 && std::abs(dfYEnd - std::round(dfYEnd)) < 0.01 )
+                        continue;
+                }
 
                 if( dfXEnd < dfX )
                 {
--- a/alg/llrasterize.cpp
+++ b/alg/llrasterize.cpp
@@ -379,7 +379,7 @@ GDALdllImageLineAllTouched( int nRasterXSize, int nRasterYSize,
                             const double *padfX, const double *padfY,
                             const double *padfVariant,
                             llPointFunc pfnPointFunc, void *pCBData,
-                            int bAvoidBurningSamePoints )
+                            int bAvoidBurningSamePoints, int bIntersectOnly)
 
 {
     if( !nPartCount )
@@ -429,6 +429,11 @@ GDALdllImageLineAllTouched( int nRasterXSize, int nRasterYSize,
             // Special case for vertical lines.
             if( floor(dfX) == floor(dfXEnd) || fabs(dfX - dfXEnd) < .01 )
             {
+                double dXOffset = std::abs(std::fmod(dfX, 1.));
+                double dXEndOffset = std::abs(std::fmod(dfXEnd, 1.));
+                if ( bIntersectOnly && std::abs(dXOffset - .5) > .49 &&  std::abs(dXEndOffset - .5) > .49)
+                    continue;
+
                 if( dfYEnd < dfY )
                 {
                     std::swap(dfY, dfYEnd );
@@ -498,6 +503,11 @@ GDALdllImageLineAllTouched( int nRasterXSize, int nRasterYSize,
             // Special case for horizontal lines.
             if( floor(dfY) == floor(dfYEnd) || fabs(dfY - dfYEnd) < .01 )
             {
+                double dYOffset = std::abs(std::fmod(dfY, 1.));
+                double dYEndOffset = std::abs(std::fmod(dfYEnd, 1.));
+                if ( bIntersectOnly && abs(dYOffset - .5) > .49 &&  abs(dYEndOffset - .5) > .49)
+                    continue;
+
                 if( dfXEnd < dfX )
                 {
                     std::swap(dfX, dfXEnd);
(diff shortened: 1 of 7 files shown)


rouault:
> Add an ALL_INTERSECTED option in gdal_rasterize with a `-ai` argument.

I believe the intent of ALL_TOUCHED (probably not very well named if one has https://en.wikipedia.org/wiki/DE-9IM in mind, but I believe here it was not intended to be that strict) was to be exactly what you do with ALL_INTERSECTED, which at first sight is juts a small variant of ALL_TOUCHED that is less sensitive to numerical inaccuracies, right ? I'd say just modify ALL_TOUCHED behaviour to have the behaviour you intend. There's no practical point in burning a pixel if the intersection between that pixel and the geometry is a geometry of area equal (within some tiny threshold, like 1/1000 of the pixel size) to 0.

Would be good to add a very simple test case for that in autotest/alg/rasterize.py , like the geometry being a square perfectly aligned on a raster pixel

SebastienPeillet:
@rouault Thank you for the quick answer !

> I believe the intent of ALL_TOUCHED was to be exactly what you do with ALL_INTERSECTED

For polygons I guess it is, but for polylines I'm not sure.

I converted polygon in polyline and also un-snap one vertex to test the `-ai` method. When I try to rasterize it, here is the result I get with `-at` method:

image

and with the `-ai` method:

image

In my opinion, users expect the first result when he rasterizes a polyline, whether it is snapped to the grid or not.

Or we keep `-at` for polyline and the variant for polygon ?

-----

> juts a small variant of ALL_TOUCHED that is less sensitive to numerical inaccuracies

Yes, from what I understand about GDALdllImageLineAllTouched(), I observed that if the transformed coordinates padfX and padfY have no decimal parts, so the vertex is snapped on the grid (tell me if I'm wrong, I probably didn't test every configuration) So I changed the condition (https://github.com/OSGeo/gdal/pull/6415/files#diff-70cc44010b08e7f321b62bd2413cc1e243292667f17185363105f3a03b3456f6R442):

```c
if( iX < 0 || iX >= nRasterXSize )
```

into

```c
if (( iX < 0 || iX >= nRasterXSize ) || (std::fmod(dfX, 1.) == 0. && std::fmod(dfXEnd, 1.) == 0.)) // for vertical line
```

and the same for horizontal line with Y coordinate.

-----

> Would be good to add a very simple test case for that in autotest/alg/rasterize.py , like the geometry being a square perfectly aligned on a raster pixel

Of course, also depending of what we decide about polyline rasterization (cf. above).

rouault:
> In my opinion, users expect the first result when he rasterizes a polyline, whether it is snapped to the grid or not.

When the line is snapped to the grid, this is tricky for a polyline. Currently we only rasterize pixels below and at the right of the line. That seems rather arbitrary. I'm wondering if we shouldn't rasterize both sides. But we might also keep the current behaviour for lines as it is probably acceptable as it is currently, and have a boolean flag to the GDALdllImageLineAllTouched() method to decide if we rasterize for lines and polygons (and for polygons use your variation to avoid being too "greedy")

> I observed that if the transformed coordinates padfX and padfY have no decimal parts, so the vertex is snapped on the grid (tell me if I'm wrong, I probably didn't test every configuration) 

That would be surprising. padfX/padfY should just reflect the input that was provided by the user.

rouault:
@SebastienPeillet any update on this ?

apps/gdal_rasterize_bin.cpp
```diff
@@ -42,7 +42,7 @@ static void Usage(const char* pszErrorMsg = nullptr)
 
 {
     printf(
-        "Usage: gdal_rasterize [-b band]* [-i] [-at]\n"
+        "Usage: gdal_rasterize [-b band]* [-i] [-at] [-ai]\n"
```
rouault, line 45:
```suggestion
        "Usage: gdal_rasterize [-b band]* [-i] [-at]\n"
```

autotest/alg/rasterize.py
```diff
@@ -444,6 +444,48 @@ def test_rasterize_6():
...
+
+    # Create a memory raster to rasterize into.
+    print(gdal)
+    target_ds = gdal.GetDriverByName("MEM").Create("", 12, 12, 1, gdal.GDT_Byte)
+    target_ds.SetGeoTransform((0, 1, 0, 12, 0, -1))
+    target_ds.SetProjection(sr_wkt)
+
+    # Create a memory layer to rasterize from.
+
+    snapped_square_ds = ogr.Open("data/square.csv")
```
rouault, line 464: the file hasn't been committed

alg/llrasterize.cpp
```diff
@@ -429,6 +429,11 @@ GDALdllImageLineAllTouched( int nRasterXSize, int nRasterYSize,
             // Special case for vertical lines.
             if( floor(dfX) == floor(dfXEnd) || fabs(dfX - dfXEnd) < .01 )
             {
+                double dXOffset = std::abs(std::fmod(dfX, 1.));
+                double dXEndOffset = std::abs(std::fmod(dfXEnd, 1.));
+                if ( bIntersectOnly && std::abs(dXOffset - .5) > .49 &&  std::abs(dXEndOffset - .5) > .49)
+                    continue;
```
rouault, line 435: to save some computations when bIntersectOnly == false
```suggestion
                if ( bIntersectOnly )
                {
                    double dXOffset = std::abs(std::fmod(dfX, 1.));
                    double dXEndOffset = std::abs(std::fmod(dfXEnd, 1.));
                    if( std::abs(dXOffset - .5) > .49 &&  std::abs(dXEndOffset - .5) > .49)
                        continue;
                }
```

but couldn't that be just rewritten as
```suggestion
                if ( bIntersectOnly )
                {
                    if( std::abs(dfX - std::round(dfX)) < 0.01 && std::abs(dfXEnd - std::round(dfXEnd)) < 0.01 )
                        continue;
                }
```

alg/llrasterize.cpp
```diff
@@ -498,6 +503,11 @@ GDALdllImageLineAllTouched( int nRasterXSize, int nRasterYSize,
             // Special case for horizontal lines.
             if( floor(dfY) == floor(dfYEnd) || fabs(dfY - dfYEnd) < .01 )
             {
+                double dYOffset = std::abs(std::fmod(dfY, 1.));
+                double dYEndOffset = std::abs(std::fmod(dfYEnd, 1.));
+                if ( bIntersectOnly && abs(dYOffset - .5) > .49 &&  abs(dYEndOffset - .5) > .49)
+                    continue;
```
rouault, line 509: to save some computations when bIntersectOnly == false
```suggestion
                if ( bIntersectOnly )
                {
                    double dYOffset = std::abs(std::fmod(dfY, 1.));
                    double dYEndOffset = std::abs(std::fmod(dfYEnd, 1.));
                    if( std::abs(dYOffset - .5) > .49 &&  std::abs(dYEndOffset - .5) > .49)
                        continue;
                }
```

but couldn't that be just rewritten as
```suggestion
                if ( bIntersectOnly )
                {
                    if( std::abs(dfY - std::round(dfY)) < 0.01 && std::abs(dfYEnd - std::round(dfYEnd)) < 0.01 )
                        continue;
                }
```

autotest/alg/rasterize.py
SebastienPeillet: Indeed, I add it

alg/llrasterize.cpp
SebastienPeillet: Indeed, I don't know why I went to this complicated way :)

SebastienPeillet:
> @SebastienPeillet any update on this ?

I made the changes to only keep one function `GDALdllImageLineAllTouched` and drop `GDALdllImageLineAllIntersected`. And I investigated on the following remark :

> That would be surprising. padfX/padfY should just reflect the input that was provided by the user.

When I debug, my input is always transformed [on this line](https://github.com/OSGeo/gdal/pull/6415/files#diff-e296988311bd30bde5d7bf4267478fde1bf41eff7ba988cae918e56c26e71e44R571) :

I have `aPointX`, a vector of x coordinate that reflect the input (EPSG:32620)
<details>
<summary>aPointX (before)</summary>

> 		aPointX	<103 éléments>	std::vector<double, std::allocator<double> >
> 				710090	double
> 				710280	double
> 				710280.9056772852	double   // the un-snapped point above
> 				710290	double
> 				710290	double
> 				710300	double
>                              etc....

</details>

and after transform : 

<details>
<summary>aPointX (after)</summary>

> 		aPointX	<103 éléments>	std::vector<double, std::allocator<double> >
> 				41	double
> 				60	double
> 				60.09056772851909	double   // the un-snapped point above
> 				61	double
> 				61	double
> 				62	double
>                              etc....

</details>

And I have similar behaviour with data in EPSG:4326 except that coordinates are not fully rounded because of the precision of the projection : 

<details>
<summary>aPointX (before)</summary>

> 				-61.04972977579104	double
> 				-61.04701595362687	double
> 				-61.04701595362687	double
> 				-61.04757743407463	double
> 				-61.04757743407463	double
> 				-61.04870039497015	double
> 				-61.04870039497015	double
> 				-61.04804533444776	double
> 				-61.04804533444776	double
> 				-61.04926187541791	double
> 				-61.04926187541791	double
> 				-61.04972977579104	double
> 				-61.04972977579104	double

</details>

and after 

<details>
<summary>aPointX (after)</summary>

> 		aPointX	<13 éléments>	std::vector<double, std::allocator<double> >
> 				43	double
> 				71.99999999988358	double
> 				71.99999999988358	double
> 				66	double
> 				66	double
> 				54	double
> 				54	double
> 				60.99999999988358	double
> 				60.99999999988358	double
> 				47.99999999988358	double
> 				47.99999999988358	double
> 				43	double
> 				43	double

</details>

Based on this, I supposed that data was translated on pixel coodinates but maybe it was only a coincidence. Futhermore in [llrasterize.cpp](https://github.com/OSGeo/gdal/pull/6415/files#diff-70cc44010b08e7f321b62bd2413cc1e243292667f17185363105f3a03b3456f6R444), I guess `floor` function was used to retrieve pixel coord in integer to burn value after ([here](https://github.com/OSGeo/gdal/pull/6415/files#diff-70cc44010b08e7f321b62bd2413cc1e243292667f17185363105f3a03b3456f6R470-R475)).

Again, I'm not fully aware of GDAL API, so I'm not sure if my modification applies to all use cases (projection, transform etc) but until then I didn't meet a case where it didn't work.

rouault:
python script formatting issue
Install pre-commit: pip install precommit
Apply it to repository (to be done once): pre-commit install
Manually fix files (not needed afterwards, since it will be done automatically at git commit time): pre-commit run --all-files

SebastienPeillet:
@rouault thank you for all your reviews ! :)

PR #6422, opened 2022-09-23
GTIFF+JXL: add test to check handling of files created before #6393

tbonfort:


autotest/gcore/tiff_read.py
```diff
@@ -4573,3 +4573,18 @@ def test_tiff_read_unhandled_codec_unknown_name():
...
+
+
+###############################################################################
+# Test reading a 4 band jxl tiff created before 3.6 where alpha vs undefined
+# channel handling was not explicitely handled (#6393)
+
+def test_tiff_jxl_read_for_files_created_before_6393():
+    gdal.ErrorReset()
+    with gdaltest.error_handler():
+        ds = gdal.Open("data/jxl-rgbi.tif")
```
rouault, line 4585: Could you also move it to autotest/gcore/data/gtiff/jxl-rgbi.tif . We try to avoid adding new stuff into autotest/gcore/data/ itself.
Same for be below file, which hasn't been committed

PR #6431, opened 2022-09-25
libopencad: avoid casting an invalid value as ResolutionUnit

rouault:
(fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=51826)


rouault:
CC @BishopGIS

BishopGIS:
@rouault Thanks Even. Merged to lib_opencad. All unit tests passed successfully.

PR #6438, opened 2022-09-27
GTiff: add multi-threaded read capabilities

rouault:
when doing a RasterIO() request that intersects multiple tiles/strips,
and that the NUM_THREADS open option or GDAL_NUM_THREADS configuration
option is defined.

Limitations:
- not used bands with number of bits != 8, 16, 32, 64, 128

Benchmark results:

- 3-band Byte tiled 10,000 x 5,000 large file:

```
gdal_translate autotest/gdrivers/data/small_world.tif small_world_none.tif -co tiled=yes -r cubic -outsize 10000 0
gdal_translate autotest/gdrivers/data/small_world.tif small_world_lzw.tif -co tiled=yes -co compress=lzw -co predictor=2 -r cubic -outsize 10000 0
gdal_translate autotest/gdrivers/data/small_world.tif small_world_deflate.tif -co tiled=yes -co compress=deflate -co predictor=2 -r cubic -outsize 10000 0
gdal_translate autotest/gdrivers/data/small_world.tif small_world_jpeg.tif -co tiled=yes -co compress=jpeg -co photometric=ycbcr -r cubic -outsize 10000 0
gdal_translate autotest/gdrivers/data/small_world.tif small_world_webp_lossy.tif -co tiled=yes -co compress=webp -r cubic -outsize 10000 0
gdal_translate autotest/gdrivers/data/small_world.tif small_world_webp_lossless.tif -co tiled=yes -co compress=webp -co webp_lossless=yes -r cubic -outsize 10000 0
gdal_translate autotest/gdrivers/data/small_world.tif small_world_jxl_lossy.tif -co tiled=yes -co compress=jxl -co jxl_lossless=no -r cubic -outsize 10000 0
gdal_translate autotest/gdrivers/data/small_world.tif small_world_jxl_lossless.tif -co tiled=yes -co compress=jxl -r cubic -outsize 10000 0
```

Measured with Intel(R) Core(TM) i7-10750H CPU @ 2.60GHz (6 cores hyper-threaded => 12 vCPUs)
```
time GDAL_NUM_THREADS=12 hyperfine --warmup 1 'python -c "from osgeo import gdal; ds = gdal.Open(\"small_world_none.tif\"); x = ds.ReadRaster()"'
```

```
File                        single-threaded    12 threads   speed-up  size
small_world_none            210 ms             112 ms       x 1.9     157 MB
small_world_deflate         310 ms             137 ms       x 2.3      11 MB
small_world_jpeg            312 ms             138 ms       x 2.3       1.1 MB
small_world_lzw             390 ms             142 ms       x 2.7      12 MB
small_world_webp_lossy      411 ms             148 ms       x 2.8       0.5 MB
small_world_webp_lossless   558 ms             157 ms       x 3.6       7.7 MB
small_world_jxl_lossy       1415 ms            368 ms       x 3.8       0.7 MB
small_world_jxl_lossless    6810 ms            1368 ms      x 5.0       8.6 MB
```

- 11977 x 8745 x 8 bands UInt16 tiled deflate compressed file

```
File                        single-threaded    12 threads   speedup   size
big_deflate_8bands_uint16   5700 ms            1109 ms      x 5.1     990 MB
```

- [updated] above file served locally with HTTP mongoose server

```
File                        single-threaded    12 threads   speedup   size
big_deflate_8bands_uint16   6800 ms            1500 ms      x 4.5    990 MB
```


tbonfort:
> not used when /vsicurl/ multi-range read capabilities are used

I did not find any references to vsicurl itself in the diff, does this mean that there is no multi-threaded read if VSIHasOptimizedReadMultiRange is true ?

rouault:
> does this mean that there is no multi-threaded read if VSIHasOptimizedReadMultiRange is true ?

yes, my phrasing was to make things simpler to understand.  Could potentially be done, but would require some extra specific code path. It is also not entirely clear if fetching all block data should be done in a single time like currently, or if each thread should fetch the data for the block it processes (which would require duplicating the VSI handle)

tbonfort:
I'm not sure to understand the "which would require duplicating the VSI handle" part of your reply. In this PR's current state, the VSIHandle is not duplicated, and the underlying Reads happen sequentially ?

For the VSIHasOptimizedReadMultiRange=true case, have you considered an option to use VSIFReadL instead of  VSIFReadMultiRangeL in this case in order to force this concurrent decoding to happen? (just throwing out ideas, I'm not certain it would help)

tbonfort:
> parallel single-range fetching from each decoding task

Given that VSIFReadMultiRangeL supposedly doesn't update the current offset (as returned by ftell) it _should_ be possible to call it concurrently (each time with a single chunk) without duplicating the VSIHandle :-P (More seriously, I don't know if that would work correctly with the vsicurl handlers, but should work out of the box with the vsiplugin et. al. backends)

rouault:
> it _should_ be possible to call it concurrently (each time with a single chunk) without duplicating the VSIHandle

VSICurlHandle::ReadMultiRange() cannot be used concurrently on the same handle currently: some GDAL specific code should be put under a mutex.
The best would probably to have a PRead() implementation for it.
Regarding vsiplugin, there's no guarantee that the read_multi_range callback implementation can be used concurrently. Same here: a PRead() implementation would be needed to make it reliable.
Or have a IsReadMultiRangeThreadSafe() method.

tbonfort:
A more subtle implementation to address my previous comment could be to extend the largefile API with a VSISupportsConcurrentRangeReads() method, which if true would activate the "parallel single-range fetching from each decoding task" code path. My gut feeling is that that will lead to better performance than the other 2 options.

tbonfort:
Sorry for bothering you @rouault with a somewhat unrelated question, I'm testing out this support alongside airbusgeo/godal#101 , after having swapped the order of the two tests in geotiff.cpp line 2392 in order to bypass the HasOptimizedReadMultiRange() test. My issue is that inside MultiThreadedRead() the `sContext.poHandle->HasPRead()` is being handled by the default vsi handler instead of the one I have installed. What am I doing wrong?

rouault:
> What am I doing wrong?

I don't see anything obviously wrong in your code. Are you sure you're building against the updated GDAL headers ?

tbonfort:
> I don't see anything obviously wrong in your code. Are you sure you're building against the updated GDAL headers ?

Yes I'm quite sure. If I build against the old headers I get
```godal.cpp:1308:24: error: ‘bool cpl::VSIGoHandle::HasPRead() const’ marked ‘override’, but does not override
 1308 |                   bool HasPRead() const override { /*printf("called\n");*/return true; }
```

rouault:
You've modified GTiffDataset::IRasterIO( ), but perhaps your code rather use GTiffRasterBand::IRasterIO() which would need a similar modification around line 5669 ?

tbonfort:
> You've modified GTiffDataset::IRasterIO( ), but perhaps your code rather use GTiffRasterBand::IRasterIO() which would need a similar modification around line 5669 ?

I don't think this is the case, I've been running this code in gdb and I only hit the breakpoint on line 3047, not on line 5669. Stepping into HasPread() of line 3050, I get to the cpl_vsil.cpp implementation of HasPread

tbonfort:
After much head banging, I found the issue that can be fixed with
```
diff --git a/port/cpl_vsil_cache.cpp b/port/cpl_vsil_cache.cpp
index a182162a83..45646e52c1 100644
--- a/port/cpl_vsil_cache.cpp
+++ b/port/cpl_vsil_cache.cpp
@@ -137,6 +137,9 @@ class VSICachedFile final : public VSIVirtualHandle
     int Close() override;
     void *GetNativeFileDescriptor() override
         { return poBase->GetNativeFileDescriptor(); }
+
+    bool HasPRead() const override;
+    size_t    PRead( void* /*pBuffer*/, size_t /* nSize */, vsi_l_offset /*nOffset*/ ) const override;
 };
 
@@ -542,6 +545,14 @@ int VSICachedFile::Flush()
 {
     return 0;
 }
+    
+bool VSICachedFile::HasPRead() const {
+    return poBase->HasPRead();
+}
+
+size_t VSICachedFile::PRead( void* pBuffer, size_t nSize, vsi_l_offset nOffset ) const {
+    return poBase->PRead(pBuffer, nSize, nOffset);
+}
 
 //! @endcond

```

rouault:
Update: extended to support /vsicurl/-like, or more precisely any file declaring HasOptimizedReadMultiRange() and HasPRead()

tbonfort:
LGTM. In the case of a remote cog file, I believe there is room for improvement still by emitting one pread request for all consecutive tiles in a scanline, which would cut down the number of request to roughly sqrt of how many are done now in the general case. That might be better treated in a subsequent pr.

PR #6450, opened 2022-09-28
Zarr: fix `ZSTD_LEVEL` range in Zarr docs

lnicola:
## What does this PR do?

`ZSTD_LEVEL` defaults to 13, which is outside of the 1-9 range. I'm not sure if we accept 1-19 or 1-22 here, but GTif uses 1-19 and I can't find the code where this is actually used.

## Tasklist

 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/doc/source/drivers/raster/zarr.rst
+++ b/doc/source/drivers/raster/zarr.rst
@@ -338,7 +338,7 @@ with ``ARRAY:`` using :program:`gdalmdimtranslate`):
 - **LZMA_DELTA=integer** : Delta distance in byte. Only used when COMPRESS=LZMA.
   Defaults to 1.
 
-- **ZSTD_LEVEL=integer** [1-9]: ZSTD compression level. Only used when COMPRESS=ZSTD.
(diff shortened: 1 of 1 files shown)


doc/source/drivers/raster/zarr.rst
```diff
@@ -338,7 +338,7 @@ with ``ARRAY:`` using :program:`gdalmdimtranslate`):
 - **LZMA_DELTA=integer** : Delta distance in byte. Only used when COMPRESS=LZMA.
   Defaults to 1.
 
-- **ZSTD_LEVEL=integer** [1-9]: ZSTD compression level. Only used when COMPRESS=ZSTD.
+- **ZSTD_LEVEL=integer** [1-19]: ZSTD compression level. Only used when COMPRESS=ZSTD.
```
rouault, line 341:
```suggestion
- **ZSTD_LEVEL=integer** [1-22]: ZSTD compression level. Only used when COMPRESS=ZSTD.
```

rouault:
The max is 22. I've fixed it by amending your commit. Thanks

PR #6455, opened 2022-09-29
JPEG2000 drivers: report compression reversibility

rouault:
- JPEG2000 drivers: report a COMPRESSION_REVERSIBILITY=LOSSLESS/LOSSLESS (possibly)/LOSSY metadata item in IMAGE_STRUCTURE domain
- JP2OpenJPEG: for reversible compression, write a hint in the COM marker if the compression is lossy or not, and use it on reading
- NITF: report COMPRESSION_REVERSIBILITY for JPEG2000 compressed images

### DIFF
--- a/autotest/gdrivers/jp2openjpeg.py
+++ b/autotest/gdrivers/jp2openjpeg.py
@@ -95,9 +95,14 @@ def test_jp2openjpeg_3():
     ds = gdal.Open("data/jpeg2000/int16.jp2")
     ds_ref = gdal.Open("data/int16.tif")
 
+    # 9x7 wavelet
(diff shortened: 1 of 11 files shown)


jratike80:
What is reported if the JPEG2000 file is created with another software and does not contain that metadata?  Maybe "LOSSY" if 9x7 is used and "POSSIBLY LOSSLESS" if it was 5x3. Or "POSSIBLY LOSSY" in the latter case.

rouault:
> Maybe "LOSSY" if 9x7 is used and "POSSIBLY LOSSLESS" if it was 5x3.

if 9x7, "LOSSY"
If 5x3, "LOSSLESS (possilby)" (normally if you use 5x3, you would include all quality layers, so the likelihood it is lossless is higher than lossy).

PR #6470, opened 2022-10-04
Extend BuildOverviews() to support options

### LINKED ISSUE #6467: Adding options to build overviews function
Hi, 
I want to add overviews to a raster using build overviews, where the options/config can change between rasters (like different compression type, compression quality, ...).
If I use the command line I can use the gdalaldo and set there the options I want, such as:
`gdaladdo --config GDAL_NUM_THREADS 4 --config COMPRESS_OVERVIEW ZSTD geotif.tif 2 4 8 16 32`

But if I use the c#/ python swig wrapper I can use the BuildOverviews function like this:
`gdal.SetConfigOption("COMPRESS_OVERVIEW", "DEFLATE")` 
`Image.BuildOverviews("AVERAGE", [2,4,8,16,32,64, 128, 256])` 
From my understating the gdal.SetConfigOption is a global parameter, meaning that that the next time I will want to built an overview it will use the same config option unless I set it otherwise.

This structure makes it really difficult to built generic code that can handle multiple config options, for example having a function that take as input the overview options and use it to add overviews to the raster with this options. It will be much simpler if the BuildOverviews will have the ability to receive the overviews options/config, like there is for the translate function.

Am I missing something ? Is there a way to setConfigOption only for a specific raster ?

Thanks in advance,
Idan

I am using gdal 3.5.2

rouault:
Fixes #6467

### DIFF
--- a/MIGRATION_GUIDE.TXT
+++ b/MIGRATION_GUIDE.TXT
@@ -1,3 +1,12 @@
+MIGRATION GUIDE FROM GDAL 3.5 to GDAL 3.6
+-----------------------------------------
+
+- Out-of-tree vector drivers:
+  * GDALDataset::IBuildOverviews(): parameters panOverviewList and panBandList
+    are now of type 'const int*' (previously 'int*')
+  * GDALRasterBand::BuildOverviews(): parameter panOverviewList is now of
+    type 'const int*' (previously 'int*')
+
 MIGRATION GUIDE FROM GDAL 3.4 to GDAL 3.5
 -----------------------------------------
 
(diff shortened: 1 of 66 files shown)


idanaviv10:
Is there a way to get all possible options somehow (to validate the options exist before using them) ?

rouault:
> Is there a way to get all possible options somehow (to validate the options exist before using them) ?

unfortunately no. This would be non trivial to implement given that the relevant options may depend on if the dataset is opened in read-only/update, if USE_RRD=YES is used or not, and other driver specific logic.

idanaviv10:
It might worth to consider having a general list of parameters per driver (like the creation option in the metadata of the driver). In this case we can have some validation on the parameters, detecting parameters that are not supported at all.

idanaviv10:
Is there something missing in order to merge this branch to master ? 
Is there anything I could help with ?

PR #6471, opened 2022-10-05
DOC:Python: Make osgeo rst files static

### LINKED ISSUE #6465: New Python API doc is less user-friendly than previous doc
The recent improvements to the Python doc are welcome, but... from a usability perspective, the doc is much harder to navigate.

It's actually easier to refer to the old doc at https://web.archive.org/web/20220328010331/https://gdal.org/python/ than to try and scroll through the huge page of items at https://gdal.org/api/python/osgeo.html.

Maybe all it needs is to add everything in to the sidebar - is it just a case of increasing the TOC level or does Sphinx make you jump through more hoops to achieve this?

snowman2:
Closes #6465


rouault:
Thanks!

PR #6473, opened 2022-10-05
Make CPLODBCSession and CPLODBCStatement member variables 'protected' (fixes #6314)

### LINKED ISSUE #6314: Make CPLODBCSession and CPLODBCStatement member variables protected, not private
`CPLODBCSession` and `CPLODBCStatement` member variables are private, so it's not possible to derive from them if one wants to add some features & helper methods.

Would it be possible to make them `protected`?

## Operating system
Windows 10, Visual Studio 2019

GDAL 3.5.1-2 from OSGeo4W

fstanchina:
Not private, so it's possible to create useful derived classes.

### DIFF
--- a/port/cpl_odbc.h
+++ b/port/cpl_odbc.h
@@ -174,6 +174,7 @@ class CPL_DLL CPLODBCSession {
 
     CPL_DISALLOW_COPY_ASSIGN(CPLODBCSession)
 
+  protected:
     CPLString m_osLastError{};
     HENV      m_hEnv = nullptr;
     HDBC      m_hDBC = nullptr;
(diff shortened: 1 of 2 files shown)


port/cpl_odbc.h
```diff
@@ -174,6 +174,7 @@ class CPL_DLL CPLODBCSession {
 
     CPL_DISALLOW_COPY_ASSIGN(CPLODBCSession)
 
+  protected:
```
rouault, line 177: Can you add the following line just below protected
``//! @cond Doxygen_Suppress
``
and that one just before public:
``//! @endcond
``
to avoid Doxygen complaining about undocumented members.

Same for CPLODBCStatement class

rouault:
Hopefully the last change needed to solve
```
Preprocessing /home/runner/work/gdal/gdal/port/cpl_odbc.cpp:1571: warning: no matching class member found for 
  int CPLODBCStatement::Failed(int nResultCode)
```

``//! @cond Doxygen_Suppress`` and ``//! @endcond`` should be put around the Failed() definition in cpl_odbc.cpp itself

fstanchina:
Sorry, got sidetracked by life and forgot about your comment.

PR #6476, opened 2022-10-05
GPKG: raster overlays: Fix min max mean std_dev values when creating overviews

dirkvdb:
## What does this PR do?
Fix the values of the `min` `max` `mean` `std_dev` columns of the `gpkg_2d_gridded_tile_ancillary` table

When creating overlays for a GeoPackage containing raster tiles that have a nodata of e.g. -9999
The values of min max mean will be calculated without ignoring the nodata entries.

Reason:
In the `GDALGeoPackageDataset::CreateCopy` implementation `poDS->SetGeoTransform(adfGeoTransform);` is called before the nodata values are taken over.

`SetGeoTransform` calls `GDALGeoPackageDataset::FinalizeRasterRegistration` which in turn calls `poOvrDS->InitRaster` for each zoom level.

This is currently done before the nodata has been assigned to the parent dataset so when the overview tiles are created in `GDALGPKGMBTilesLikePseudoDataset::WriteTileInternal` `bHasNoData` will be false and de nodata values will be included in the statistics calculation.

This pull request puts the copy of the nodata values before the call to `SetGeoTransform`

None

## Tasklist

 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/ogr/ogrsf_frmts/gpkg/ogrgeopackagedatasource.cpp
+++ b/ogr/ogrsf_frmts/gpkg/ogrgeopackagedatasource.cpp
@@ -5504,6 +5504,9 @@ GDALDataset* GDALGeoPackageDataset::CreateCopy( const char *pszFilename,
         return nullptr;
     }
 
+    // Assign nodata values before the SetGeoTransform call.
+    // SetGeoTransform will trigger creation of the overview datasets for each zoom level
+    // and at that point the nodata value needs to be known.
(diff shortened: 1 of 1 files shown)


dirkvdb:
The comment is added

rouault:
Test added per https://github.com/OSGeo/gdal/pull/6480

PR #6491, opened 2022-10-09
Add OGRLayer::GetGeometryTypes() 

rouault:
This method iterates over features to retrieve their geometry types. This
is mostly useful for layers that report a wkbUnknown geometry type.
    
By default this method returns an array of nEntryCount entries with each
geometry type (in OGRGeometryTypeCounter::eGeomType) and the corresponding
number of features (in OGRGeometryTypeCounter::nCount).
Features without geometries are reported as eGeomType == wkbNone.

Optimized implementations for GPKG and PostgreSQL

### DIFF
--- a/autotest/ogr/ogr_basic_test.py
+++ b/autotest/ogr/ogr_basic_test.py
@@ -793,6 +793,101 @@ def test_ogr_basic_float32_formatting():
 
 
 ###############################################################################
+
+
+def test_ogr_basic_get_geometry_types():
+    """Test Layer.GetGeometryTypes()"""
+
+    ds = ogr.GetDriverByName("Memory").CreateDataSource("")
+    lyr = ds.CreateLayer("layer")
+    lyr.CreateField(ogr.FieldDefn("str", ogr.OFTString))
+    lyr.CreateField(ogr.FieldDefn("str2", ogr.OFTString))
+    lyr.CreateGeomField(ogr.GeomFieldDefn("geom2", ogr.wkbUnknown))
+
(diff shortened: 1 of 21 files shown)


jratike80:
Sorry that I cannot read the details from the code.

Do you stop with MIXED if just linestring and multilinestring, or polygon and multipolygon are found? They could still be handled by promoting the geometries to MULTI without a need to change the geometry field into generic GEOMETRY if no other geometry types are found?

In case of GeoPackage, do you first check that the geometry_type_name in gpkg_geometry_columns is `GEOMETRY` (means geom_type=ogr.wkbUnknown), because otherwise it should be impossible to have mixed geometrytypes?

rouault:
> Do you stop with MIXED if just linestring and multilinestring, or polygon and multipolygon are found?

yes. I had some hesitation about that, but I decided to keep things simple. It could be tricky to decide which grouping of "equivalent" geometry types make sense. In some context linestring ~= multilinestring, but circularstring, compoundcurve or multicurve should be treated differently. In other contexts, all types of lines should be considered equivalent.  Thus the user can just query all types and decide by itself which groupings make sense.

> In case of GeoPackage, do you first check that the geometry_type_name

No, the implementation of GetGeometryTypes() is really checking all features. It is the responsibility of the user to only call it when it makes sense (ie GetGeomType() == wkbUnknown).

> But on the other hand, gpkg can be invalid and the metadata and the contents of the table do not necessarily match.

That wasn't the motivation, but yes could be used for that.

PR #6499, opened 2022-10-11
SWIG: fix MSVC library name since the switch to CMake

Baljak:
Following the switch to CMake that was made in 3.5, GDAL library on Windows is no longer named `gdal_i.lib`, so this has to be reflected in SWIG.

### DIFF
--- a/swig/python/setup.py
+++ b/swig/python/setup.py
@@ -242,9 +242,6 @@ class gdal_ext(build_ext):
         elif isinstance(self.library_dirs, str) and sys.platform == 'darwin':
             self.library_dirs += ':' + ':'.join(library_dirs)
         if self.libraries is None:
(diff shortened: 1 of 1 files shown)


rouault:
the proposed fix cannot be used as it, as it would break if GDAL was still built with nmake.
In master branch / 3.6, things have been fixed
For 3.5, I'm not sure there's a better alternative than specifying manually the name of the import lib with ``python setup.py build --libraries=gdal``. Or perhaps some code that would check the content of the self.libraries directory to see if there's a gdal_i.lib or gdal.lib file ?

Baljak:
Ok, I understand, I'll see what I can do about this.

rouault:
closing this merge request as it is specific of release/3.5 which is end of life now

PR #6510, opened 2022-10-12
Docs (GTiff & COG): blocksize must be divisible by 16

maxfreu:
Hi! This tripped me up in some software that doesn't pass gdal warnings / errors. So I wondered why I got no output and apparently the reason was an incorrectly set block size. This hint aims to inform users about the need that the blocksize must be divisible by 16.

### DIFF
--- a/doc/source/drivers/raster/gtiff.rst
+++ b/doc/source/drivers/raster/gtiff.rst
@@ -370,11 +370,11 @@ Creation Options
 -  **TILED=YES**: By default striped TIFF files are created. This
    option can be used to force creation of tiled TIFF files.
 
--  **BLOCKXSIZE=n**: Sets tile width, defaults to 256.
+-  **BLOCKXSIZE=n**: Sets tile width, defaults to 256. Must be divisible by 16.
 
 -  **BLOCKYSIZE=n**: Set tile or strip height. Tile height defaults to
    256, strip height defaults to a value such that one strip is 8K or
-   less.
+   less. Must be divisible by 16.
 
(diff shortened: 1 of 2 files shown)


jratike80:
I don't know what is the truth but at least it is not that simple.
```
gdal_translate -of GTiff -co blockysize=5 stat.tif temp.tif`

gdalinfo temp.tif
Band 1 Block=12000x5 Type=Byte, ColorInterp=Red
Band 2 Block=12000x5 Type=Byte, ColorInterp=Green
Band 3 Block=12000x5 Type=Byte, ColorInterp=Blue
```

doc/source/drivers/raster/gtiff.rst
```diff
@@ -370,11 +370,11 @@ Creation Options
 -  **TILED=YES**: By default striped TIFF files are created. This
    option can be used to force creation of tiled TIFF files.
 
--  **BLOCKXSIZE=n**: Sets tile width, defaults to 256.
+-  **BLOCKXSIZE=n**: Sets tile width, defaults to 256. Must be divisible by 16.
 
 -  **BLOCKYSIZE=n**: Set tile or strip height. Tile height defaults to
    256, strip height defaults to a value such that one strip is 8K or
-   less.
+   less. Must be divisible by 16.
```
rouault, line 377:
```suggestion
   less. Must be divisible by 16 when TILED=YES.
```

rouault:
> I don't know what is the truth but at least it is not that simple.

This should indeed by divisible by 16, but only for the TILED=YES cases. I've updated the BLOCKYSIZE doc to mention that

PR #6515, opened 2022-10-12
SXF: fix floating-point division by zero, and other issues with uninitialized values

rouault:
Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=52346

CC @drons


drons:
Looks quite good

PR #6521, opened 2022-10-13
SXF: fix harmless unsigned-integer-overflow in OGRSXFDataSource::FillLayers()…

rouault:
…, and simplify code

Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=52369

CC @drons


drons:
>Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=52369

Unfortunately, permission denied for me. 

But the code has become better and clearer.

@BishopGIS Isn't it?

BishopGIS:
Fixes looks fine.

PR #6532, opened 2022-10-15
GPKG: implement upsert with INSERT OR REPLACE

rouault:
CC @beryan : digging a bit I saw there was no need to use UPSERT advanced capabilities, but that SQLite has forever supported INSERT OR REPLACE which is exactly what we want. The only tricky point is that it doesn't seem there's a reliable way of knowing if a new row has been created or just inserted, which had some impact on the way OGR book keeps feature count.


jratike80:
I wonder how this will work in practice. GPKG tables have primary key as `integer autoincrement` and I guess that GPKG clients send insert requests without FID and let the SQLite to create it. Shouldn't the REPLACE require that also the FID is included in `INSERT OR REPLACE`? But if the client knows the FID why would it create INSERT instead of UPDATE? I apologize if I have misunderstood something.

rouault:
> I guess that GPKG clients send insert requests without FID and let the SQLite to create it.

CreateFeature() which uses INSERT works both in the case where the feature has the FID unset or set. But if the client tries to insert a feature with a known FID and that this feature already exists, the operation fails.

> Shouldn't the REPLACE require that also the FID is included in `INSERT OR REPLACE`?

if UpsertFeature() is invoked on a feature without a FID, a regular INSERT is generated.
If it is invoked with a FID, then INSERT OR REPLACE includes the FID.

> But if the client knows the FID why would it create INSERT instead of UPDATE?

The client may have an up-to-date feature whose FID is meaningful to uniquely identify the feature, but does not know if there's already an older version of it in the database. UpsertFeature (here this is a very simple version of upsert where the conflict is on the FID and the resolution of the conflict is to accept all the updated fields) solves this in an atomic way

jratike80:
> The client may have an up-to-date feature whose FID is meaningful to uniquely identify the feature, but does not know if there's already an older version of it in the database. UpsertFeature (here this is a very simple version of upsert where the conflict is on the FID and the resolution of the conflict is to accept all the updated fields) solves this in an atomic way

We were studying a possibility to use https://www.sqlite.org/sqldiff.html for keeping customers GPKG databases up-to-date with minimal inserts/deletes/updates but found that it is not suitable for us, or at least it would be complicated to implement, because the meaningful unique attribute we have cannot be used as a primary key and we could not guarantee that the "integer autoincrement" PK keeps in sync with the meaningful identifiers (order of features in the GPKG table should remain always the same and PK values of deleted features must not appear again ever).

UPSERT by using a user-given unique attribute as a key instead of the GPKG PK might work if the ON CONFLICT trigger could be modified accordingly.

jratike80:
Perhaps there should be a warning in a documentation that the one who is using "insert of replace" must be very careful and understand their data and what their are doing. For example a user may have an existing gpkg and some new data in another gpkg with the same schema. Let's say it is created with ogr2ogr and `-sql "select * from features where update_date>....` They may think that it would be fine to upsert features from the new gpkg, but because in both databases the fid is a counting number starting from 1 the result of upsert would be a catastrophe.

rouault:
@jratike80 the last 2 commits should cover the use case of your https://github.com/OSGeo/gdal/pull/6532#issuecomment-1279970158 comment

jratike80:
I noticed. They look good to me and I am excited to test once I'll find some time for that.

rouault:
> They look good to me and I am excited to test once I'll find some time for that.

the "interesting" part is that I hit what I believe is a bug in SQLite which doesn't like the GeoPackage RTree _update1 trigger when doing a real UPSERT (ie "INSERT INTO ... ON CONFLICT ... DO UPDATE ..."). I've raised it https://sqlite.org/forum/forumpost/8c8de6ff91. If the SQLite team says it is expected and doesn't intend to change this behaviour, then I might have to raise the issue to the OGC GeoPackage github tracker as it could be worth updating the official _update1 with the 2 ones I temporarily insert

rouault:
The SQLite3 author answered that SQLite behaves as documented regarding this (even if this is quite unexpected). https://github.com/opengeospatial/geopackage/issues/639 raised to make the GeoPackage team aware of that issue and suggest updating the spec.
No more to update in this PR

PR #6533, opened 2022-10-16
Fix dollars to cents conversion expression

zstadler:
## What does this PR do?
Fix an incorrect example expression in the OGR SQL documentation
## Tasklist

## Environment
N/A


rouault:
thanks :-)

PR #6537, opened 2022-10-17
Add CPLMask functions to simplify bitmask handling

dbaston:
## What does this PR do?

This PR attempts to some parts of the warp code easier to read by adding functions to manipulate bit masks stored in a GUInt32*. I thought of putting this in a custom type but it seemed better to be less invasive, avoid changing signatures, and stick with GUInt32*.

None

## Tasklist

 - [X] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/port/cpl_mask.h
+++ b/port/cpl_mask.h
@@ -1,0 +1,136 @@
+/**********************************************************************
+ * $Id$
+ *
+ * Name:     cpl_string.h
+ * Project:  CPL - Common Portability Library
+ * Purpose:  Bitmask manipulation functions
+ * Author:   Daniel Baston, dbaston@gmail.com
+ *
+ **********************************************************************
+ * Copyright (c) 2022, ISciences LLC
+ *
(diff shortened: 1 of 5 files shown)


port/cpl_mask.h
```diff
@@ -0,0 +1,136 @@
+/**********************************************************************
+ * $Id$
+ *
+ * Name:     cpl_string.h
```
rouault, line 4:
```suggestion
 * Name:     cpl_mask.h
```

port/cpl_mask.h
```diff
@@ -0,0 +1,136 @@
+/**********************************************************************
+ * $Id$
```
dg0yt, line 2: Does GDAL still need to carry the CVSish `$Id`?

rouault: indeed. mass removal in https://github.com/OSGeo/gdal/pull/6540

PR #6561, opened 2022-10-20
CPLHTTPFetch(): ignore SSL/TLS errors about non-properly terminated connections

rouault:
Fixes https://lists.osgeo.org/pipermail/gdal-dev/2022-October/056330.html

Cf https://github.com/curl/curl/pull/3148 and https://github.com/curl/curl/issues/9024

### DIFF
--- a/port/cpl_http.cpp
+++ b/port/cpl_http.cpp
@@ -1384,7 +1384,8 @@ CPLHTTPResult *CPLHTTPFetchEx( const char *pszURL, CSLConstList papszOptions,
         if( strlen(szCurlErrBuf) > 0 )
         {
             bool bSkipError = false;
-
+            const char* pszContentLength =
+                CSLFetchNameValue(psResult->papszHeaders, "Content-Length");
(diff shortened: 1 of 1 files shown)


jgrocha:
When the error is different then those related to SSL, the message might not be very helpful (`"You may set the CPL_CURL_IGNORE_ERROR..."` )

Example (using a server listed on our tests):

```
ogrinfo 'https://sampleserver6.arcgisonline.com/arcgis/services/SampleWorldCities/MapServer/WFSServer?service=WFS&version=2.0.0&request=GetFeature&typename=esri:World&count=1'
ERROR 1: HTTP/2 stream 0 was not closed cleanly: INTERNAL_ERROR (err 2). You may set the CPL_CURL_IGNORE_ERROR configuration option to YES to try to ignore it.
```

It is just a comment. There is no problem is showing the message, even if it does not apply.

jgrocha:
I've checkout out this PR and it runs fine on my side.

PR #6565, opened 2022-10-21
OCI: Update driver description regarding database views.

heidivanparys:
## What does this PR do?

* Add information regarding how to read database views.
* Add two examples.
* ALL_SDO_GEOM_METADATA is a view, not a table: correct this information in the driver documentation.

N/A

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments

## Environment

N/A

### DIFF
--- a/doc/source/drivers/vector/oci.rst
+++ b/doc/source/drivers/vector/oci.rst
@@ -15,14 +15,16 @@ Oracle client libraries are available.
 When opening a database, its name should be specified in the form
 "OCI:userid/password@database_instance:table,table". The list of tables
 is optional. The database_instance portion may be omitted when accessing
-the default local database instance.
+the default local database instance. See the 
+`Oracle Help Center <https://docs.oracle.com/search/?q=oci%20driver>`_ 
+for more information about the OCI driver.
 
 If the list of tables is not provided, then all tables appearing in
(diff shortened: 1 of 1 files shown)


jratike80:
It has been a while since I used Oracle but by then we used to register the spatial views into USER_SDO_GEOM_METADATA (and from there is flows to ALL_SDO_GEOM_METADATA). Isn't that not necessary/recommended practice any more?

heidivanparys:
> It has been a while since I used Oracle but by then we used to register the spatial views into USER_SDO_GEOM_METADATA (and from there is flows to ALL_SDO_GEOM_METADATA). Isn't that not necessary/recommended practice any more?

Yes, you are right, the text is not completely correct.

It is not necessary to register views in USER_SDO_GEOM_METADATA, but some applications do require it in order to be able to read from that view, see e.g. https://community.safe.com/s/article/adding-metadata-entries-for-oracle-spatial-tables. So it is a good practice to do so.

And you must register tables in USER_SDO_GEOM_METADATA if you want to create an index or use spatial operators on the geometry column.

I created a new (draft) pull request, please let me know if you agree with the changes: https://github.com/OSGeo/gdal/pull/6588.

jratike80:
Feels good to me.

PR #6573, opened 2022-10-22
GPKG: speed-up: implement background RTree creation in bulk insertion in a new table

rouault:
We create a temporary database with only the RTree, and we insert records into it in a dedicated thread, in parallel of the main thread that inserts rows in the user table. When the layer is finalized, we just use bulk copy statements of the form
INSERT INTO rtree_xxxx_rowid/node/parent SELECT * FROM temp_rtree.my_rtree_rowid/node/parent
to copy the RTree auxiliary tables into the main database, which is a very fast operation.

With that, "ogr2ogr out.gpkg nz-building-outlines.parquet" goes from 62 seconds down to 40 seconds (35% faster)
with the dataset at https://storage.googleapis.com/open-geodata/linz-examples/nz-building-outlines.parquet

### DIFF
--- a/autotest/cpp/test_cpl.cpp
+++ b/autotest/cpp/test_cpl.cpp
@@ -54,6 +54,7 @@
 #include "cpl_quad_tree.h"
 #include "cpl_worker_thread_pool.h"
 #include "cpl_vsi_virtual.h"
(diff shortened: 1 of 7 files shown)


jratike80:
Nice thinking.

PR #6578, opened 2022-10-24
GPKG: implement a ST_Area() method...

rouault:
... without requiring libspatialite+geos, that directly computes
the area from the WKB.

Around 5 times faster than libspatialite ST_Area()

### DIFF
--- a/autotest/ogr/ogr_gpkg.py
+++ b/autotest/ogr/ogr_gpkg.py
@@ -7247,3 +7247,59 @@ def test_ogr_gpkg_background_rtree_build(filename):
     ds = None
 
     gdal.Unlink(filename)
+
+
+###############################################################################
+# Test ST_Area()
+
+
+@pytest.mark.parametrize(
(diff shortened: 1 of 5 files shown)


jratike80:
Certainly fastest this way but it would be nice to know if the RTTOPO based ST_Area in Spatialite is any faster https://www.gaia-gis.it/gaia-sins/spatialite-sql-latest.html.

rouault:
> if the RTTOPO based ST_Area in Spatialite is any faster

that would be comparing apples to oranges. The RTTOPO based ST_Area() in Spatialite is for the 2 argument ST_Area(geom, use_ellipsoid=0 or 1) variant, that uses geodesic computations on sphere/ellipsoid involving trigonometry, which are necessarily slower than the cartesian ones that involve just a few multiplications and additions. The main reason for the speed-up with just the cartesian variant is to save creation of Spatialite and GEOS objects and memory copies

PR #6583, opened 2022-10-26
HANA: properly handle special characters in connection string

mrylov:
This PR reworks the building of a connection string by following the official [documentation](https://help.sap.com/docs/SAP_HANA_CLIENT/f1b440ded6144a54ada97ff95dac7adf/7cab593774474f2f8db335710b2f5c50.html).

### DIFF
--- a/ogr/ogrsf_frmts/hana/ogrhanadatasource.cpp
+++ b/ogr/ogrsf_frmts/hana/ogrhanadatasource.cpp
@@ -45,7 +45,6 @@
 #include <sstream>
 #include <vector>
 
-
 using namespace OGRHANA;
 
 namespace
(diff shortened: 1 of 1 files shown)


ogr/ogrsf_frmts/hana/ogrhanadatasource.cpp
```diff
@@ -153,92 +152,100 @@ class OpenOptionsConstants
...
+
+    auto addParameter = [&](const char* paramName, const char* paramValue)
+    {
         if (paramValue == nullptr)
             return;
-        params.push_back(CPLString(paramName) + "=" + CPLString(paramValue));
+
+        CPLString value(paramValue);
+        bool hasSpecialChar = false;
+        for (char ch : value)
```
rouault, line 182: maybe use https://cplusplus.com/reference/string/string/find_first_of/ ?

mrylov: Good point @rouault, thank you. I've pushed a new version.

rouault:
could you please rebase on top of latest master to get the fix of https://github.com/OSGeo/gdal/pull/6584 that fixes the issue with the build-linux-ubuntu-focal CI configuration ?

PR #6585, opened 2022-10-28
Fixed building position independent static lib

JanWielemaker:
The various sub projects use inconsistent setting for the POSITION_INDEPENDENT_CODE property of the libraries, which makes it impossible to build a static lib that is position independent.

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:


JanWielemaker:
Patch is self explanatory and simple.  We want to build a static library that is later linked into a shared object.

rouault:
thanks!

PR #6603, opened 2022-11-01
CMake: better handle source, build and install directory with spaces, at least on Unix

rouault:


rouault:
> Doesn `file(TO_NATIVE_PATH...)` do any harm for `NOT WIN32`?

it adds backslashes in front of spaces on Linux, so "foo bar" becomes "foo\ bar". Admitedly this patch is not super elegant, but I fail to understand all the subtelties of where to use CMake or native paths

dg0yt:
I see. Thanks.

With CMake 3.20, there is `cmake_path(NATIVE ..)` which should replace `file(TO_NATIVE_PATH...)`. A quick test indicates that it doesn't add backslashes before spaces.
We might add a function with that signature for old versions.
~~~cmake
set(A "/tmp/foo bar")
message(STATUS "1 ${A}")
file(TO_NATIVE_PATH "${A}" B)
message(STATUS "2 ${B}")

if(NOT COMMAND cmake_path)
	function(cmake_path command)
		if(command STREQUAL "NATIVE_PATH")
			if(ARGV2 STREQUAL "NORMALIZE") # not implemented
				set(out_var "${ARGV3}")
			else()
				set(out_var "${ARGV2}")
			endif()
			if(CMAKE_HOST_WIN32)
				file(TO_NATIVE_PATH "${ARGV1}" "out_var")
			else()
				set("${out_var}" "${${ARGV1}}")
			endif()
			set("${out_var}" "${${out_var}}" PARENT_SCOPE)
		else()
			message(FATAL_ERROR "cmake_path polyfill doesn't support ${command}")
		endif()
	endfunction()
endif()

cmake_path(NATIVE_PATH A C)
message(STATUS "3 ${C}")
~~~

PR #6633, opened 2022-11-08
Add GDT_Int8 support (RFC 87)

rouault:
On reading, PIXELTYPE=SIGNEDBYTE in IMAGE_STRUCTURE metadata domain is no longer reported in drivers that used to do it.

On writing, the PIXELTYPE=SIGNEDBYTE creation option is preserved in drivers that used to support it, but is deprecated.


rouault:
> for consistency purpose it should be great to have the GDT_UInt8 as an alias of GDT_Byte witch is still ambiguous, isn't it?

I had indeed pondered about that. GDT_Byte is not strictly speaking ambiguous as it has always been documented as a unsigned 8-bit integer (if one expects the PIXELTYPE=SIGNEDBYTE hint that amends it). If we were to provide GDT_UInt8, this should by as a #define GDT_UInt8 GDT_Byte rather than as a proper enumeration value of GDALDataType, so that people who swith( eDataType ) don't have to include ``case`` statements for both GDT_Byte and GDT_UInt8.

sgillies:
Can we adhere to the example of numpy? Ubyte (same as uint8) vs byte?

rouault:
> Ubyte (same as uint8) vs byte?

what do you suggest exactly: keep GDT_Byte in the enumeration and add ``#define GDT_UByte GDT_Byte`` to create the alias ?

PR #6634, opened 2022-11-08
Add RFC 87: Signed int8 data type for raster

rouault:
Pre-rendered view of RFC text at https://github.com/rouault/gdal/blob/rfc87/doc/source/development/rfc/rfc87_signed_int8.rst

### DIFF
--- a/doc/source/development/rfc/rfc87_signed_int8.rst
+++ b/doc/source/development/rfc/rfc87_signed_int8.rst
@@ -1,0 +1,119 @@
+.. _rfc-87:
+
+=============================================================
+RFC 87: Signed int8 data type for raster
+=============================================================
+
+============== =============================================
+Author:        Even Rouault
+Contact:       even.rouault @ spatialys.com
+Started:       2022-Nov-8
+Status:        Discussion
+Target:        GDAL 3.7
+============== =============================================
+
+Summary
+-------
+
+This RFC describes the addition of a new GDT_Int8 data type, for signed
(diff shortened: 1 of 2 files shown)


doc/source/development/rfc/rfc87_signed_int8.rst
```diff
@@ -0,0 +1,119 @@
...
+:ref:`rfc-14` will be amended to point to this RFC, and mention that starting
+with GDAL 3.7.0, PIXELTYPE=SIGNEDBYTE is no longer the technical solution.
+
+Testing
+-------
+
+The test_gdal.cpp and testcopywords.cpp test files are extended to test functions
+like :cpp:func:`GDALDataTypeIsInteger`, :cpp:func:`GDALDataTypeIsFloating`,
+:cpp:func:`GDALDataTypeIsComplex`, etc. and all data type conversion combinations
+from/into GInt8.
```
mloskot, line 104: Is it worth to touch on the conversions here in the RFC? e.g. lossy, linear mapping, source min/max -> target min/max, what about preserving signed Zero when converting to unsigned channel type, etc. implications, and also on mixed channel arithmetic, if any.

rouault: not completely sure to follow all what you've in mind here, but the conversion rules applied for Int8 are consistent with the ones applied for similar types such as Int16 or Int32.

mloskot: That will do 😄

PR #6637, opened 2022-11-09
swig/python_exceptions.i: Add Context Manager for handling Python exception state

neilflood:
## Allow the state of using exceptions to be set/reset locally, within a Python `with` statement

The Python gdal/ogr/osr/gnm modules each have a global state flag determining whether or not errors raise Exceptions. It defaults to not, and can be set or unset. However, because this is global, conflicts can arise in what a section of code requires, sometimes in code from other libraries. To avoid interfering with other code, one often needs to save current state, set desired state, perform some actions, then reset to saved state. 

This PR adds a Python Context Manager class which handles this process, so that a Python `with` statement can have a desired state within that block, and not interact with the global state. 

It is added to the generic `python_exceptions.i` SWIG code, so it appears in each of the different GDAL modules.

### DIFF
--- a/autotest/gcore/basic_test.py
+++ b/autotest/gcore/basic_test.py
@@ -752,6 +752,7 @@ def test_gdal_DataTypeUnion():
 
     assert gdal.DataTypeUnion(gdal.GDT_Byte, gdal.GDT_UInt16) == gdal.GDT_UInt16
 
+
 def test_exceptionmanager():
     currentExceptionsFlag = gdal.GetUseExceptions()
     usingExceptions = currentExceptionsFlag == 1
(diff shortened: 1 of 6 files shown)


rouault (commented):
Would you mind also comitting the regnerated bindings file ?
Cf https://github.com/OSGeo/gdal/blob/master/CONTRIBUTING.md#build-swig to have SWIG 4.0.2 and also note you need to pass  -DSWIG_REGENERATE_PYTHON=ON to CMake

It would be great to have testing of this class, for example in autotest/gcore/basic_test.py

neilflood:
Thanks @rouault 
Yes, I have tested locally (Ubuntu) using -DSWIG_REGENERATE_PYTHON and swig-4.0.2, all seems fine. 

Have added the four regenerated bindings, and a simple test where you suggested. Thank you!

PR #6672, opened 2022-11-11
CMake: remove RECOMMENDED flag to BRUNSLI and QB3. Add it for CURL

rouault:
(refs https://github.com/spack/spack/pull/33856#issue-1446090634)


adamjstewart:
Here is a list of all discrepancies I noticed with recommended deps:

Recommended only by CMake: brunsli, geos, lerc, qb3, qhull

Recommended only by https://gdal.org/build_hints.html#build-requirements: curl

Recommended but seems to actually be required?: zlib, libtiff, libgeotiff, json-c

rouault:
> Recommended but seems to actually be required?: zlib, libtiff, libgeotiff, json-c

if you don't have them as external libraries, the internal libraries will be used. So yes, they are kind of required, but not necessarily as external libraries like PROJ

PR #6676, opened 2022-11-12
CMake: do not use Armadillo if it lacks LAPACK support (such as on Alpine)

rouault:
and make it a FATAL_ERROR if the user used -DGDAL_USE_ARMADILLO=ON

### DIFF
--- a/cmake/helpers/CheckDependentLibraries.cmake
+++ b/cmake/helpers/CheckDependentLibraries.cmake
@@ -540,6 +540,8 @@ define_find_package2(GTA gta/gta.h gta PKGCONFIG_NAME gta)
 gdal_check_package(GTA "Enable GTA driver" CAN_DISABLE)
 
 gdal_check_package(MRSID "MrSID raster SDK" CAN_DISABLE)
+
+set(GDAL_USE_ARMADILLO_OLD ${GDAL_USE_ARMADILLO})
 gdal_check_package(Armadillo "C++ library for linear algebra (used for TPS transformation)" CAN_DISABLE)
 if (ARMADILLO_FOUND)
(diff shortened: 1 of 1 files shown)


cmake/helpers/CheckDependentLibraries.cmake
```diff
@@ -597,6 +597,15 @@ if (ARMADILLO_FOUND)
     set(GDAL_USE_ARMADILLO CACHE BOOL OFF FORCE)
   endif ()
 
+  # LAPACK support required for arma::solve()
+  if (GDAL_USE_ARMADILLO AND EXISTS "${ARMADILLO_INCLUDE_DIRS}/armadillo_bits/config.hpp")
+      file(READ "${ARMADILLO_INCLUDE_DIRS}/armadillo_bits/config.hpp" armadillo_config)
+      if ("${armadillo_config}" MATCHES "/\\* #undef ARMA_USE_LAPACK")
+          message(WARNING "Armadillo build lacks LAPACK support. Disabling it as it cannot be used by GDAL")
```
dg0yt, line 604: "lacks" or "uses" ?  The PR title and commit message seem to be wrong.

dg0yt: I also wonder it shouldn't be a fatal error if the user explicitly enabled Armadillo.

rouault: > The PR title and commit message seem to be wrong.

indeed. changed "CMake: do not use Armadillo if it *lacks* LAPACK support (such as on Alpine)". Thanks!

rouault: > I also wonder it shouldn't be a fatal error if the user explicitly enabled Armadillo.

good idea. done

PR #6655, opened 2022-11-10
gfs: add XML schema

jechterhoff:
## What does this PR do?

It adds an XML Schema file for the gfs XML format.

## What are related issues/pull requests?

_none_

## Tasklist

 - [ ] Clarify all TBDs in the documentation annotations contained in the XML schema file
 - [ ] Review
 - [ ] Adjust for comments
 - [ ] All CI builds and checks have passed

## Environment

_irrelevant for this PR_

PR #6716, opened 2022-11-16
gml_registry.xsd: add XML schema

jechterhoff:
## What does this PR do?

It adds an XML schema for the gml registry files used by the GML driver

_none_

## Tasklist

 - [ ] Clarify all TBDs in the documentation annotations contained in the XML schema file
 - [ ] Review
 - [ ] Adjust for comments

## Environment

_irrelevant for this PR_


jechterhoff:
@rouault Here is the PR, as discussed by email.

Just as last time (in #6655), your review of the XSD as well as the documentation annotations (including the TBDs and questions contained therein) would be much appreciated.

Questions / comments:

* Are the multiplicities, i.e., minOccurs and maxOccurs, correct?
* Are the chosen element types correct?
* Documentation annotations:
  * Please check the parts marked with 'TBD', and the questions written there.
  * Please check for other parts that the documentation is correct.

data/gml_registry.xsd
```diff
@@ -0,0 +1,69 @@
...
+                  </xs:attribute>
+                  <xs:attribute name="schemaLocation" type="xs:string">
+                    <xs:annotation>
+                      <xs:documentation>XML schema file in which the feature type is defined (possibly together with other feature types). The GML driver will attempt to parse it and create layer definitions. The schema location can be given as URL (http and https), absolute file path, and relative file path (relative to the location of the registry file).</xs:documentation>
+                    </xs:annotation>
+                  </xs:attribute>
+                </xs:complexType>
+              </xs:element>
+            </xs:sequence>
+            <xs:attribute name="prefix" type="xs:string" use="required">
```
rouault, line 47: It is actually optional, otherwise the data/gml_registry.xml file doesn't validate
```suggestion
            <xs:attribute name="prefix" type="xs:string" use="optional">
```

Would you mind also editing .github/workflows/validate_xml.yml to add validation of data/gml_registry.xml

data/gml_registry.xsd
```diff
@@ -0,0 +1,69 @@
...
+                    <xs:annotation>
+                      <xs:documentation>Name of the XML element that is used to identify the feature type. Typically the XML element name of an XML encoded feature. However, it can also be a different element that can be used in combination with @elementValue to produce a match for this feature type element.</xs:documentation>
+                    </xs:annotation>
+                  </xs:attribute>
+                  <xs:attribute name="elementValue" type="xs:string">
+                    <xs:annotation>
+                      <xs:documentation/>
+                      <xs:documentation>Used in combination with @elementName. If present, a match for this feature type element is only produced if an XML element whose name equals the value of @elementName and whose value equals the value of @elementValue is found in the GML data. For examle: If namespace/@prefix = 'ex', featureType/@elementName = 'typeOfFeature', and featureType/@elementValue = 'FT', then a match is produced if &lt;ex:typeOfFeature>FT&lt;/ex:typeOfFeature> is present in the GML data.</xs:documentation>
```
rouault, line 31:
```suggestion
                      <xs:documentation>Used in combination with @elementName. If present, a match for this feature type element is only produced if an XML element whose name equals the value of @elementName and whose value equals the value of @elementValue is found in the GML data. For example: If namespace/@prefix = 'ex', featureType/@elementName = 'typeOfFeature', and featureType/@elementValue = 'FT', then a match is produced if &lt;ex:typeOfFeature&gt;FT&lt;/ex:typeOfFeature&gt; is present in the GML data.</xs:documentation>
```

data/gml_registry.xsd
```diff
@@ -0,0 +1,69 @@
...
+            </xs:sequence>
+            <xs:attribute name="prefix" type="xs:string" use="required">
+              <xs:annotation>
+                <xs:documentation>Namespace prefix assigned to the features of this namespace in actual GML data. The prefix will be looked up in the first bytes of the GML file (e.g. xmlns:abc="http://example.org/abc"). NOTE: Case is important - the parser looks for the exact string, i.e., 'ex' is treated differently than 'EX'. If actual data may use different prefixes for the same namespace URI, multiple namespace elements must be added to the gml_registry - one for each applicable prefix.
+
+TBD: Is the prefix actually required? In the gdal/data/gml_registry.xml, the namespace for the japanese schemas does not contain a prefix. Not sure what the exact behavior is when a prefix is defined vs. when it is not. If no prefix is defined, does the parser then automatically resolve namespace prefixes and only match XML elements whose namespace-uri equals the defined uri, and local-name equals featureType/@elementName (plus the test for @elementValue)? Or is the namespace ignored completely, then, and just the local-name matters?</xs:documentation>
```
rouault, line 51: the namespace prefix, when present, is used to confirm the match. If it is missing, then only the uri is used. I don't recall the rationale for that. In any case, no "smart" XML resolution is done. This is just "stupid" string comparisons. The GML driver is not really namespace aware and if you use 2 different namespace prefixes that are for the same namespace URI, it might be confused and doesn't realize that actually define the same namespace

rouault (commented):
thanks! just a few comments and suggestions

jechterhoff:
@rouault Thank you for reviewing the XSD, as well as all the suggestions and clarifications.

I'm going to update the PR, also editing .github/workflows/validate_xml.yml to add validation of data/gml_registry.xml.

Regarding the "prefix" XML attribute, I have a question, though. Omitting the prefix does not appear to work as expected. I made a test - see files in [gfs-Tests.zip](https://github.com/OSGeo/gdal/files/10028643/gfs-Tests.zip). Command line call is `ogrinfo test_combined.gml -ro -al --config GML_REGISTRY my_gml_registry.xml`. Tested with GDAL 3.5.2, released 2022/09/02 (returned by executing `gdalinfo --version` in the OSGeo4W shell).

When run, none of the schemas defined in my_gml_registry.xml are picked up. The result is what the GML driver constructs itself, based upon the content of the GML file. Not even the second namespace is picked up, even though it would match. When prefix="ex2" is added to the first namespace, then apparently a match is detected by the GML driver, and test_structured_type.gfs is applied. Can you confirm this behavior? If so, is it a bug or not (or due to the GDAL version I used for the test)? Am I missing something - which could well be the case. :-)

rouault:
> Omitting the prefix does not appear to work as expected.

ok, reading more the code, it is a bit trickier. So if you have
```
  <namespace uri="http://example.org/ex2">
    <featureType elementName="Feature" gfsSchemaLocation="test_structured_type.gfs"/>
  </namespace>
```
then this rule will only trigger if the GML file contains  "http://example.org/ex2" somewhere (presumably as the default namespace xmlns="http://example.org/ex2", but this isn't enforced) *and* there is at least ``<Feature``  ( ``<${elementName}``) somewhere

For example autotest/ogr/data/gml_jpfgd/BldA.xml starts with
```
<?xml version="1.0" encoding="utf-8"?>
<Dataset xsi:schemaLocation="http://fgd.gsi.go.jp/spec/2008/FGD_GMLSchema FGD_GMLSchema.xsd"
       xmlns:gml="http://www.opengis.net/gml/3.2"
       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
       xmlns:xlink="http://www.w3.org/1999/xlink"
       xmlns="http://fgd.gsi.go.jp/spec/2008/FGD_GMLSchema"
       gml:id="Dataset1">
  <gml:description>テストデータ</gml:description>
  <gml:name>基盤地図情報テストデータ（GML版）</gml:name>
<BldA gml:id="K17_1234567890_123456">
```
and matches
```
    <namespace uri="http://fgd.gsi.go.jp/spec/2008/FGD_GMLSchema">
        <featureType elementName="BldA"
                     gfsSchemaLocation="jpfgdgml_BldA.gfs" />
```

jechterhoff:
Ah, yes - it is exactly as you wrote. The presence of a default namespace declaration - in the root element, or any element in the path from the root up until and including the feature element - is not enforced (even though that would probably make the instance XSD invalid - but that would be a separate concern) or checked (it can even be a different uri). What matters, if no prefix is defined for a namespace, is that an XML element with the @elementName of the feature type is found in the GML data - without any prefix. 

I'll update the documentation accordingly.

rouault:
there's a remaining TBD

jechterhoff:
Well spotted. I missed that. Fixed.

Issue #6714, opened 2022-11-16
QGIS crashes when applying .qml-style on GML-layer


pathmapper:
This has been reported downstream before (https://github.com/qgis/QGIS/issues/50901) and was identified as related to GDAL.

## Expected behavior and actual behavior.

No crash is expected, but QGIS crashes when applying .qml-style on GML-layer.

## Steps to reproduce the problem.

With QGIS (tested with QGIS 3.29.0-Master):

1. Download and unzip [test_data.zip](https://github.com/qgis/QGIS/files/10018998/test_data.zip)
2. Drag'n'drop `test.gml` into QGIS
3. For this layer, load and apply style `BP_KennzeichnungsFlaeche_Flaeche.qml`
4. QGIS crashes

**No crash for GPKG**

If the same style is applied on `test.gpkg` (also included in [test_data.zip](https://github.com/qgis/QGIS/files/10018998/test_data.zip)), there's no crash.

## Operating system

Ubuntu 22.04.1 LTS

## GDAL version and provenance

3.4.1 from QGIS 3.29.0-Master (https://github.com/qgis/QGIS/commit/9b8be60f7b4bf81a953a718e192e25b40f611358)

nicogodet:
Confirmed on GDAL 3.6.0 on Windows (OSGeo4W)

elpaso:
The crash comes from the duplicated `zweckbestimmung` entry, which makes it an `OFTIntegerList` type, the expression evaluator considers it a string (because it gets converted to a string representation by the `default` entry in the switch from ` swq_expr_node *OGRFeatureFetcher`, when it comes to evaluation, again we take the default (`else`) which considers it a comparison between two strings, but the actual value to compare with is an integer and there is a null `string_value` in the expression node -> crash.

elpaso:
I'm in doubt about the expected behavior (besides the obvious fact that it must not crash): the expression is `"zweckbestimmung" = 'Naturgewalten' or  "zweckbestimmung"  = 1000`, but the GML has two values for `zweckbestimmung`:

```xml
<ogr:featureMember>
    <ogr:test gml:id="GML_b0245903-e6cc-4068-b9ed-b3ef287b5581">
[...]

PR #6717, opened 2022-11-16
SWQ: fix crash when comparing integer array fields (fixes: #6714)

elpaso:
When comparing integer array fields with integers the code represents the arrays with their string serialized form, comparing with the string representation of the integer value which is null led to the crash.

### DIFF
--- a/autotest/ogr/ogr_sql_test.py
+++ b/autotest/ogr/ogr_sql_test.py
@@ -1607,6 +1607,7 @@ def test_ogr_sql_field_names_same_case():
 ###############################################################################
 # Test no crash when comparing string with integer array
 
+
 def test_ogr_sql_string_int_array_comparison():
 
     ds = ogr.GetDriverByName("Memory").CreateDataSource("")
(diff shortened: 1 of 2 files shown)


rouault (approved):
I feel like there would probably a better fix in a step that occurs before (SWQGeneralChecker perhaps ? not sure... as this might be trickier as IntegerList if a OGRFieldType value but not a swq_field_type one, so the mismatch is bound to cause such issues), but your fix is reasonable.

rouault:
python tests linting issue: https://github.com/OSGeo/gdal/actions/runs/3481075088/jobs/5821696426

Install pre-commit: ``pip install precommit``
Apply it to repository (to be done once): ``pre-commit install``
Manually fix files (not needed afterwards, since it will be done automatically at git commit time): ``pre-commit run --all-files``

pathmapper:
@rouault just noticed that this is tagged for 3.6.1 but looks like it isn't included in 3.6.1 and 3.6.2.

rouault:
> @rouault just noticed that this is tagged for 3.6.1 but looks like it isn't included in 3.6.1 and 3.6.2.

ah good point, I guess I wanted to use the "backport release/3.6" label but accidentally used the milestone one.
I've just manually backported this in release/3.6 branch per 69d3d8a95cfb654b2cdc86e1f6252479d7a12bcb. So this will in 3.6.3

PR #6725, opened 2022-11-17
found another not renamed symbol on iOS

PeterPetrik:
on iOS with QT 6.4.1 it looks like there is another symbol clash in bundled jpeg QT lib


rouault:
manually cherry picked to 3.6 per e8541e495f

Issue #6726, opened 2022-11-17
Performance problem because of RawRasterBand::CanUseDirectIO()


chacha21:
GDAL 3.5.3
Windows 7 64b
Visual Studio 2019 (latest)

I have noticed a performance problem due to an unexpected overhead in `RawRasterBand::CanUseDirectIO()`

Short story : the call to `CPLGetConfigOption("GDAL_ONE_BIG_READ", nullptr);` is too expensive, and perhaps should be cached with some new field of `GDALRasterIOExtraArg`

Long story :
I am using a dataset with LINE interleaving. For information : spatial size 800x1000, bandsCount 734.
I am submitting 1000 elements of size 800x734 for writing (which best fits the interleaving, that's why the dataset was configured that way).
I was surprised to see a very high difference in writing performance between a "MEM" and an "ENVI" dataset.
The ENVI dataset was ~6x times slower
So, at first, I made sure to "warm up the cache" so that no cached block allocation was needed during writing, and the performance was still ~x3 slower with ENVI.
I did not understand where the overhead could com from.
After investigation, I replaced `const char *pszGDAL_ONE_BIG_READ = CPLGetConfigOption("GDAL_ONE_BIG_READ", nullptr);` with `const char *pszGDAL_ONE_BIG_READ = nullptr;`, and boom, the performance was now comparable.
I think this occurs because `RawRasterBand::CanUseDirectIO()` is inside a band-driven loop, and I have 734 bands, which is a lot.

Is it affordable to extend `GDALRasterIOExtraArg` with a cached value for `CPLGetConfigOption("GDAL_ONE_BIG_READ", nullptr);` ?

rouault:
> Is it affordable to extend `GDALRasterIOExtraArg` with a cached value for `CPLGetConfigOption("GDAL_ONE_BIG_READ", nullptr);` ?

I would rather suggest to had a member variable in RawRasterBand to cache the value of GDAL_ONE_BIG_READ (1 if CPLTestBool() is true, 0 if its is false, and -1 if the config option is not set at all) the first time it is accessed. The behavior will be slightly changed compared to currently, but I believe it is acceptable to assume that on a given dataset/band instance, the value of GDAL_ONE_BIG_READ shouldn't change between invokations of RasterIO()

rouault:
> Do you want me to PR ?

yes

chacha21:
> I would rather suggest to ~had~ add a member variable in RawRasterBand
[...]

PR #6729, opened 2022-11-17
cache GDAL_ONE_BIG_READ value for RasterIO

chacha21:
Proposal for #6726


gcore/gdaldataset.cpp
```diff
@@ -1968,6 +1969,10 @@ CPLErr GDALDataset::IRasterIO( GDALRWFlag eRWFlag,
 
     CPLAssert(nullptr != pData);
 
+    RawDataset* rawDataset = dynamic_cast<RawDataset*>(this);
```
rouault, line 1972: not pretty to put that here. That should be put in RawDataset::IRasterIO() itself

gcore/rawdataset.h
```diff
@@ -61,9 +61,11 @@ class CPL_DLL RawDataset : public GDALPamDataset
          virtual ~RawDataset() = 0;
 
     bool GetRawBinaryLayout(GDALDataset::RawBinaryLayout&) override;
-
+    void resetCache(void) {this->cachedCPLOneBigReadOption = std::make_pair(false, 0);}
```
rouault, line 64: I feel a better name than resetCache() should be found, as there is a risk of confusion with the block cache.
ClearCachedConfigOption() maybe ?

gcore/gdaldataset.cpp
chacha21: Ok, inheritance is more complex that I thought. I believed that RawDataset::IRasterIO was somwhat totally overriden

gcore/rawdataset.h
```diff
@@ -61,9 +61,11 @@ class CPL_DLL RawDataset : public GDALPamDataset
          virtual ~RawDataset() = 0;
 
     bool GetRawBinaryLayout(GDALDataset::RawBinaryLayout&) override;
-
+    void ClearCachedConfigOption(void) {this->cachedCPLOneBigReadOption = std::make_pair(false, 0);}
   private:
     CPL_DISALLOW_COPY_ASSIGN(RawDataset)
+  protected:
+    std::pair<bool, int> cachedCPLOneBigReadOption;
```
rouault, line 68: ``#include <utility>`` should be added in the include section at top of this file

PR #6733, opened 2022-11-18
Optimize can use direct io

chacha21:
Additional proposal for #6726 about thread safety


rouault:
I don't understand the thread safety argument: it is illegal to use a same dataset (or bands from a same dataset) concurrently from different threads. If it works currently, you are just lucky and it might break in any later version.
And your change is definitely not thread safe. For that you should employ some locking techniques

chacha21:
Ok, let me develop a little more :
I assumed the following scenario : 
A single RasterIO() occurs, which spans accross several bands. Some implementation is smart enough to parallelize writes in different bands. With the original version, it could occur that updating the cachedCPLOneBigReadOption pair fields in a non-atomic manner could make one thread catch an unexpected state of this member variable.
(think about updating the first field to true while the second field is not yet updated)

My "fix" consists only in updating cachedCPLOneBigReadOption with a single "int" write.

rouault:
Yes, I perfectly understood your scenario. I'm just telling you you are going against what GDAL guarantees to be safe, and that there are risks that things go wrong.

> My "fix" consists only in updating cachedCPLOneBigReadOption with a single "int" write.

I dont think ``int`` is guaranteed to be atomic on all platforms by the C++ standard (although it is probably on Intel architectures). But here as you use a union, etc, compilers could do funny things. Using std::atomic<> on a structure with the 2 members ``valid`` and ``value`` would be much safer and would clarify the usage. I could tolerate a patch that would do that (but with the warning I mentioned above about other places being potentially unsafe).

The ``cached8_t`` naming is also weird (I don't understand the 8).

chacha21:
> I dont think `int` is guaranteed to be atomic on all platforms by the C++ standard (although it is probably on Intel architectures). But here as you use a union, etc, compilers could do funny things. Using std::atomic<> on a structure with the 2 members `valid` and `value` would be much safer and would clarify the usage.

Ok. but I feel that bringing some std::atomic<> in the rawdataset header is a little invasive.
See my next commit involving CPLAtomicCompareAndExchange() instead.


> The `cached8_t` naming is also weird (I don't understand the 8).
The cached value is now a "char", so 8. The size of the packing unioned field ("all") is not mentioned.

rouault:
CachedValidValue_t would be clearer. I also feel a comment in the code itself, near the declaration of the type would be welcome to explain why there is this union

gcore/rawdataset.h
```diff
@@ -63,11 +63,18 @@ class CPL_DLL RawDataset : public GDALPamDataset
          virtual ~RawDataset() = 0;
 
     bool GetRawBinaryLayout(GDALDataset::RawBinaryLayout&) override;
-    void ClearCachedConfigOption(void) {this->cachedCPLOneBigReadOption = std::make_pair(false, 0);}
+    void ClearCachedConfigOption(void) {CPLAtomicCompareAndExchange(&this->cachedCPLOneBigReadOption.all, this->cachedCPLOneBigReadOption.all, 0);}
```
rouault, line 66: please move the implementation in gcore/rawdataset.cpp and add an explicit ``#include "cpl_atomic_ops.h"`` in it

PR #6734, opened 2022-11-18
WFS filter: use double-quotes for PropertyIsLike attributes

aborruso:
Otherwise the output xml query have both double and single quotes https://lists.osgeo.org/pipermail/gdal-dev/2022-November/056480.html

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:


jratike80:
I think that single quotes are just wrong. This example is from the "OpenGIS Filter Encoding Implementation Specification"
```
<Filter>
<PropertyIsLike wildCard="*" singleChar="#" escapeChar="!">
<PropertyName>LAST_NAME</PropertyName>
<Literal>JOHN*</Literal>
</PropertyIsLike>
</Filter>
```

rouault:
> I think that single quotes are just wrong

no, they are not: cf https://stackoverflow.com/a/6800482 .

@aborruso  As I suspected, some unit tests should be adjusted. See the end of the output of https://github.com/OSGeo/gdal/actions/runs/3497336791/jobs/5856354519 . The file to edit is autotest/ogr/ogr_wfs.py . Are you able to add a commit with the needed changes ? I'll do it otherwise

aborruso:
> I'll do it otherwise

It's better, thank you, but I looked at the file and I don't understand what to change

aborruso:
Thank you very much @rouault and all you gdal folks

PR #6752, opened 2022-11-20
S3: fix issue with EC2 IDMSv2 request failing inside Docker container…

rouault:
… with default networking

and causing later unrelated issues when opening a dataset. Cf https://lists.osgeo.org/pipermail/gdal-dev/2022-November/056488.html thread

https://community.grafana.com/t/imdsv2-is-not-working-from-docker/65944 points that the ``curl -X PUT "http://169.254.169.254/latest/api/token" -H "X-aws-ec2-metadata-token-ttl-seconds: 10`` request doesn't work inside Docker with default networking settings. ``docker run --network=host`` avoids that issue

The issue was that the failure of the EC2 IDMSv2 request set a CPLError() that was hidden, but not cleared. Hence when GDALOpen() run, after probing a first driver (VRT e.g.) that doesn't detect the dataset, the fact that the error state was in error made GDALOpen() abort without trying subsequent drivers.

So make sure to clear the error reset after EC2 HTTP requests, and when EC2 IDMSv2 fails but http://169.254.169.254/latest/meta-data succeeds, emit a debug message with a hint when we detect we are inside Docker.

### DIFF
--- a/port/cpl_aws.cpp
+++ b/port/cpl_aws.cpp
@@ -939,9 +939,46 @@ bool VSIS3HandleHelper::GetConfigurationFromEC2(bool bForceRefresh,
                     // Failure: either we are not running on EC2 (or something emulating it)
                     // or this doesn't implement yet IDMSv2
                     // Go on trying IDMSv1
+
+                    // /latest/api/token doesn't work inside a Docker container that
+                    // has no host networking.
+                    // Cf https://community.grafana.com/t/imdsv2-is-not-working-from-docker/65944
+                    if( psResult->pszErrBuf != nullptr &&
+                        strstr(psResult->pszErrBuf, "Operation timed out after") != nullptr )
+                    {
+                        aosOptions.Clear();
+                        aosOptions.SetNameValue("TIMEOUT", "1");
+                        CPLPushErrorHandler(CPLQuietErrorHandler);
+                        CPLHTTPResult* psResult2 =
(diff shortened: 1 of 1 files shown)


port/cpl_aws.cpp
```diff
@@ -939,9 +939,46 @@ bool VSIS3HandleHelper::GetConfigurationFromEC2(bool bForceRefresh,
...
+                        if( psResult2 )
+                        {
+                            if( psResult2->nStatus == 0 && psResult2->pabyData != nullptr )
+                            {
+                                VSIStatBufL sStat;
+                                if( VSIStatL("/.dockerenv", &sStat) == 0 )
+                                {
+                                    CPLDebug("AWS",
+                                             "/latest/api/token EC2 IDMSv2 request timed out, but /latest/metadata succeeded. "
+                                             "Trying with IDMSv1. "
```
pedros007, line 964: [Instance Metadata Service (IMDS) v1 is disabled on Amazon Linux 2023](https://docs.aws.amazon.com/linux/al2023/ug/compare-with-al2.html#imdsv2).  When running in a Docker container on an Amazon instance you'll need to update metadata allowing 2 HTTP Hops.  Here's one adhoc way to do that:

```
aws ec2 modify-instance-metadata-options \
    --instance-id <instance_id> \
    --http-put-response-hop-limit 2 \
    --http-endpoint enabled
```

PR #6773, opened 2022-11-22
Doc: OGR_L_GetArrowStream(): clarify lifecycle (fixes #6772)

### LINKED ISSUE #6772: Clarify lifecycle of ArrowArray and ArrowSchema structures in columnar access API
In the developer documentation for the stream interface, I see:

> Unless otherwise specified by a particular driver implementation, the ArrowArrayStream structure, and the ArrowSchema or ArrowArray objects its callbacks have returned, should no longer be used (except for potentially being released) after the [OGRLayer](https://gdal.org/doxygen/classOGRLayer.html) from which it was initialized has been destroyed (typically at dataset closing).

However, in the [stream specification](https://arrow.apache.org/docs/format/CStreamInterface.html#result-lifetimes), I see:

> The data returned by the get_schema and get_next callbacks must be released independently. Their lifetimes are not tied to that of the ArrowArrayStream.

The lifecycle of the ArrowArrayStream makes sense (and aligns with the Arrow Database Connectivity spec, which requires that a "statement" be kept alive for the lifecycle of the underlying stream); however, the assumption of an independent Array/Schema lifecycle is important. For example, many conversions from pyarrow to other Python objects never copy this object (this is also true in R, where we create objects that are backed by the ArrowArray whose data is never copied).

I think this is probably what is already being done (as judged mostly by lack of segfaults in my testing in R); however it would be nice to solidify that in the documentation if this is the case.

(Discussion came up here: https://github.com/geopandas/pyogrio/pull/155 )

rouault:


ogr/ogrsf_frmts/generic/ogrlayer.cpp
```diff
@@ -6139,11 +6139,17 @@ const char* OGRLayer::GetLastErrorArrowArrayStream(struct ArrowArrayStream*)
...
- * OGRLayer from which it was initialized has been destroyed (typically at dataset
- * closing). Furthermore, unless otherwise specified by a particular driver
+ * otherwise specified by a particular driver implementation, the get_schema(),
+ * get_next() and get_last_error() function pointers of the ArrowArrayStream
+ * structure should no longer be used after the OGRLayer, from which the
+ * ArrowArrayStream structure was initialized, has been destroyed (typically at
+ * dataset closing). The reason is that those function pointers will typically
+ * point to methods of the OGRLayer instance.
+ * However, the ArrowSchema and ArrowArray structures filled from those callbacks
+ * can potentially be used and released after the destruction of the
```
paleolimbot, line 6149:
```suggestion
 * can be used and must be released independently from the
```

ogr/ogrsf_frmts/generic/ogrlayer.cpp
```diff
@@ -6259,11 +6265,17 @@ bool OGRLayer::GetArrowStream(struct ArrowArrayStream* out_stream,
...
```
paleolimbot, line 6275:
```suggestion
 * can be used and must be released independently from the
```

paleolimbot (commented):
Thank you! Maybe just a little stronger wording to make it clear?

PR #6787, opened 2022-11-23
RFC69 C++ formatting: Add links and details to the RFC

elpaso:
Update the RFC with links and more details on the
proposed solution.

Formatted view of the RFC: https://github.com/elpaso/gdal/blob/rfc69_cplusplus_formatting-text-only/doc/source/development/rfc/rfc69_cplusplus_formatting.rst

### DIFF
--- a/doc/source/development/rfc/rfc69_cplusplus_formatting.rst
+++ b/doc/source/development/rfc/rfc69_cplusplus_formatting.rst
@@ -19,7 +19,6 @@ Status:     *Work-In-Progress*
 This RFC is based on `GEOS RFC
 4 <https://trac.osgeo.org/geos/wiki/RFC4>`__ by Mateusz Łoskot.
 
-**TODO:** Give examples of formatting.
 
 Summary
 -------
--- a/doc/source/development/rfc/rfc69_cplusplus_formatting.rst
+++ b/doc/source/development/rfc/rfc69_cplusplus_formatting.rst
@@ -7,12 +7,14 @@ RFC 69: C/C++ Code Formatting
 This document proposes and describes desired code formatting style used
 across C and C++ source code in GDAL.
 
-======== ======================================
-Author:  Kurt Schwehr
-Contact: schwehr@google.com / schwehr@gmail.com
-Started: 2017-May-04
-Status:  *Work-In-Progress*
-======== ======================================
+=========== ======================================
+Author:     Kurt Schwehr
+Contact:    schwehr@google.com / schwehr@gmail.com
+Co-Author:  Alessandro Pasotti
+Contact:    elpaso@itopen.it
+Started:    2017-May-04
+Status:     *Work-In-Progress*
+=========== ======================================
 
 This RFC is based on `GEOS RFC
 4 <https://trac.osgeo.org/geos/wiki/RFC4>`__ by Mateusz Łoskot.
@@ -76,18 +78,15 @@ development workflow as standalone tool or as one of many editor
 integrations or other bespoke utilities (eg. ``git cl format``
 [Chromium]).
 
-No automation of code reformatting is proposed. It would be treating the
-symptoms, no cause: developers not following the code formatting
-standard.
+A new pre-commit hook will be added to the current configuration in
+``.pre-commit-config.yaml`` to run [clang-format pre-commit].
 
-Although no means to enforce the default formatting style are proposed,
-currently used CI services (eg. Travis CI) may be employed as a
-post-commit safety valve - a clang-format lint failure as a compile
-break (e.g.
-`clang_format.py <https://github.com/mongodb/mongo/blob/master/buildscripts/clang_format.py>`__
-build script used by MongoDB). Alternatively, a gatekeeper may be
-installed in SVN/Git, rejecting commits with code not conforming to the
-code formatting style.
+To enforce the code formatting, a gatekeeper will be installed in CI,
+rejecting commits with code not conforming to the code formatting style
+and a brief textual hint to install or update the pre-commit hooks
+will be added to failure message.
+
+
 
 Code Formatting Rules
 ---------------------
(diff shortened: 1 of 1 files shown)


rouault:
@elpaso  After removing the "TODO: Give examples of formatting." at top, this looks good to me. You can post a Call for discussion for it on gdal-dev at your convenience

misbahkhalilaz (approved):

rouault:
@elpaso There's a quite important point to check I recall now. I sort of recall that the output of clang-format used to change at some point depending on which clang-format version you used, which is super annoying, as we don't want contributors with different versions of it to reformat over and over. Could you experiment a bit regarding that ? I believe checking with different versions of clang-format >= 10 should be enough (10 is what is used in ubuntu 20.04 which we can assume to be the lowest reasonable platform for development).

elpaso:
@rouault I've stumbled upon the issue myself but I think that pre-commit handles that nicely: it installs its own wheel of clang-format from pypy:

```bash
$ clang-format --version
clang-format version 15.0.4 (https://github.com/ssciwr/clang-format-wheel 0bdee69eb2db136302efab015caaf616d33f5430)
```

The important point is to match the versions here:
 https://github.com/elpaso/gdal/blob/rfc69_cplusplus_formatting_revival/.pre-commit-config.yaml#L31

and here:
https://github.com/elpaso/gdal/blob/rfc69_cplusplus_formatting_revival/.github/workflows/clang-format-check.yml#L25

From a user perspective, just run `pre-commit` instead of invoking the system `clang-format`.

rouault:
- Other directories to exclude: frmts/pcidsk/sdk, frmts/grib/degrib/degrib, frmts/grib/degrib/g2clib
- I've tried the https://github.com/elpaso/gdal/tree/rfc69_cplusplus_formatting_revival branch and when running "pre-commit run -a", I get 30 or so lines of error with the message ``Configuration file(s) do(es) not support Json: /home/even/gdal/rfc69_cplusplus_formatting_revival/.clang-format``, and it causes the clang-format hook to fail. It seems the reformatting has worked though, but the error status doesn't go away even when running again

elpaso:
@rouault 

See: https://github.com/elpaso/gdal/commit/4e4f982fa7f00eedfd4532b9462656fef8713a26 the JSON warning is gone.

About formatting, I would recommend for improved readability to change from `false` to `true`:

SpaceAfterCStyleCast
SpaceInEmptyParentheses
SpacesInAngles
SpacesInContainerLiterals
SpacesInCStyleCastParentheses
SpacesInParentheses
SpacesInSquareBrackets

Any though about `ColumnLimit:     80` ?  I know it will be controversial to change it but worth asking.

rouault:
> See: https://github.com/elpaso/gdal/commit/4e4f982fa7f00eedfd4532b9462656fef8713a26 the JSON warning is gone.

Great

> I would recommend for improved readability to change from `false` to `true`:

looking at https://clang.llvm.org/docs/ClangFormatStyleOptions.html, I'm not a big fan of those extra spaces. The current settings match quite well the current state of the code base

> Any though about ColumnLimit: 80 ?

I'd stick with it. FWIW I find QGIS lines way too long sometimes. Longer lines tend to encourage people to not create named variables where they really should.

rouault:
@elpaso To explore: blame ignore files. Cf https://lists.osgeo.org/pipermail/gdal-dev/2022-November/056558.html + https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view . What is not obvious if there is an automated way of adding the blame ignore file to the local git configuration, or if each user must manually register it.

elpaso:
@rouault I tested the `--ignore-revs-file .git-blame-ignore-revs` method and it works as expected. 

I am afraid that developers must individually run `git config blame.ignoreRevsFile .git-blame-ignore-revs` if they want to make it permanent.

elpaso:
Of course we may provide a `.git-blame-ignore-revs` with the big refactoring commit in it.

Would you like me to add this instructions to the RFC?

rouault:
> Of course we may provide a `.git-blame-ignore-revs` with the big refactoring commit in it.

that would be the idea

> Would you like me to add this instructions to the RFC?

Yes, mention in the RFC that we'll use that file. The instructions for "git config blame.ignoreRevsFile .git-blame-ignore-revs" will have to be added to https://gdal.org/development/dev_practices.html I guess.
Was wondering if that couldn't be automated with pre-commit ? but perhaps out-of-scope of it

rcoup:
> The instructions for "git config blame.ignoreRevsFile .git-blame-ignore-revs" will have to be added to https://gdal.org/development/dev_practices.html I guess.
> Was wondering if that couldn't be automated with pre-commit ? but perhaps out-of-scope of it

There's some work underway in Git to allow repos to provide some safe git configuration options with repos when they're cloned. Easiest thing is probably to wait for that. Unless it can go in a `scripts/setup-useful-dev-defaults.sh`.

doc/source/development/rfc/rfc69_cplusplus_formatting.rst
```diff
@@ -7,17 +7,18 @@ RFC 69: C/C++ Code Formatting
...
-Started: 2017-May-04
-Status:  *Work-In-Progress*
-======== ======================================
+=========== ======================================
+Author:     Kurt Schwehr
+Contact:    schwehr@google.com / schwehr@gmail.com
+Co-Author:  Alessandro Pasotti
+Contact:    elpaso@itopen.it
+Started:    2017-May-04
+Status:     *Work-In-Progress*
```
rouault, line 16:
```suggestion
Status:     Adpoted, implemented
```

doc/source/development/rfc/rfc69_cplusplus_formatting.rst
```diff
@@ -226,3 +222,4 @@ References
    ``clang-format`` interactive guide and builder
 -  `https://zed0.co.uk/clang-format-configurator/ <https://zed0.co.uk/clang-format-configurator/>`__
 -  `https://trac.osgeo.org/geos/wiki/RFC4 <https://trac.osgeo.org/geos/wiki/RFC4>`__
+- [git_blame_ignore] `Ignore Commits in Blame View` <Git https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view>`__
```
rouault, line 225:
```suggestion
- [git_blame_ignore] `Ignore Commits in Blame View` <Git https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view>`__

Voting history
------------------

+1 from PSC members KurtS and EvenR
```

PR #6788, opened 2022-11-24
Remove autotest/pytest.ini which is stale

rouault:
as it is now a generated file in the build autotest subdirectory.


rouault:
unrelated test failure: cf https://github.com/OSGeo/gdal/issues/6791

PR #6789, opened 2022-11-24
autotest/conftest.py: add collect_ignore_glob to ignore pymod/ files

rouault:
For some reason, it doesn't seem the ``testpaths`` setting of pytest.ini is honoured as expected when running ``pytest autotest`` from the build directory (at least on Linux)


rouault:
unrelated test failure: cf https://github.com/OSGeo/gdal/issues/6791

PR #6799, opened 2022-11-25
Add minimal support for reading 12-bit JPEG images with libjpeg-turbo 2.2dev and internal libjpeg12

rouault:


rouault:
failures with ubuntu_18_04_32bit_build are unrelated: cf https://github.com/OSGeo/gdal/pull/6800

PR #6808, opened 2022-11-27
Fixed build documentation for version 3.6.0 with cmake

igorgatis:
## What does this PR do?
Fixes build documentation for version 3.6.0 which uses cmake.


rouault:
There is no error. The CMake build system is available since GDAL 3.5.0, and required since 3.6.0

PR #6828, opened 2022-11-30
Fixes NITF driver bug that prevents adding subsequent TREs after a HEX TRE (fixes #6827)

### LINKED ISSUE #6827: Adding a Single HEX TRE to a NTF File Prevents Subsequent TREs from being Added.

Calling `gdal.Translate` with the `NTF` driver prevents subsequent TREs from being added after adding a single HEX TRE.

There is a bug in `nitffile.c` in method `NITFWriteTREsFromOptions`.  In this method there is a loop.  Within this loop, is a string compare method that accepts the length to compare.  This string compare method is used to filter out creation_options that do not have the correct prefix (in our case the prefix is `TRE=`).  The string compare length variable is initialized prior to the loop, but once a hex string is passed in, this length is increased by 4.  On the next iteration of the loop, this string comparison length is not reset to the original value so all subsequent TREs are filtered out because the string comparison fails (incorrectly).  You will see what I'm talking about in the PR.  It's a simple 2 line change.

See the linked zip below for the code and files to reproduce.  There is a README.md within the zip with instructions.

https://drive.google.com/file/d/1qgENDcI_QMqPzh-CuXeRheVecDgwV2th/view?usp=share_link

## Operating system

Ubuntu 20.04 64-bit

3.7.0dev - installed via master branch from source

Snaze:
## What does this PR do?

There was a bug in nitffile.c in method NITFWriteTREsFromOptions. In this method there is a loop. Within this loop, is a string compare method that accepts the length to compare. This string compare method is used to filter out options that do not have the correct prefix by `continue`-ing the loop (in our case the prefix is TRE=). The string compare length variable is initialized prior to the loop (in our case 4), but once a hex string is passed in, this length is increased by 4. On the next iteration of the loop, this string comparison length is not reset to the original value so all subsequent TREs are filtered out because the string comparison fails (incorrectly).  This commit fixes that issue.

https://github.com/OSGeo/gdal/issues/6827

## Tasklist

 - [X] Fixed bug in nitffile.c / NITFWriteTREsFromOptions
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS: Ubuntu 20.04
* Compiler: gcc (Ubuntu 9.4.0-1ubuntu1~20.04.1) 9.4.0 / g++ (Ubuntu 9.4.0-1ubuntu1~20.04.1) 9.4.0

### DIFF
--- a/frmts/nitf/nitffile.c
+++ b/frmts/nitf/nitffile.c
@@ -1457,7 +1457,6 @@ static int NITFWriteTREsFromOptions(
     int bIgnoreBLOCKA =
         CSLFetchNameValue(papszOptions,"BLOCKA_BLOCK_COUNT") != NULL;
     int iOption;
-    int nTREPrefixLen = (int)strlen(pszTREPrefix);
     const bool bReserveSpaceForTREOverflow =
         CSLFetchNameValue(papszOptions,"RESERVE_SPACE_FOR_TRE_OVERFLOW") != NULL;
 
@@ -1472,6 +1471,7 @@ static int NITFWriteTREsFromOptions(
         int  nContentLength;
         const char* pszSpace;
         int bIsHex = FALSE;
+        int nTREPrefixLen = (int)strlen(pszTREPrefix);
 
         if( !EQUALN(papszOptions[iOption], pszTREPrefix, nTREPrefixLen) )
             continue;


rouault (approved):

rouault:
thanks. I've added a test case for your fix in commit c96f83a8c7

scott-osk:
> thanks. I've added a test case for your fix in commit [c96f83a](https://github.com/OSGeo/gdal/commit/c96f83a8c7751889e8a06cd703d50e5cf8aa54da)

Thank you @rouault !  I'm still new to the code base so I didn't know where to put the unit tests (but I tested the change manually).  This helps for the future if I submit other PRs.  I really appreciate the help and the fast response!  Thanks again.

rouault:
For some reason the backport bot to 3.6 didn't work. I've just manually backported it per cb60ab91d47456be081638f8ef55e95adecc31cc. This will be in 3.6.3 in March

PR #6829, opened 2022-12-01
Do not advertise the KML driver as supporting feature styling

nirvn:
## What does this PR do?

Fixes what I assume might have been a wrong assumption when adding feature styling metadata to various vector drivers in this commit (https://github.com/OSGeo/gdal/commit/23e085223024d6e033d1d315ddbffe27be4b3751): contrary to LIBKML, the KML driver does not support feature styling.

@nyalldawson , as discussed.

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments


nyalldawson:
@nirvn 

So the issue here is that the KML driver has *write* support for styles, which is why it advertises that flag. We need to make the metadata more granular and indicated read/write support individually...

PR #6830, opened 2022-12-01
Add DCAP_FEATURE_STYLES_READ and DCAP_FEATURE_STYLES_WRITE capabilities

nyalldawson:
These provide clients with more granular capability information vs the existing DCAP_FEATURE_STYLES capability (which only indicates that the driver has either read and/or write support, so cannot be used to determine that eg a driver supports reading styles)

Supersedes https://github.com/OSGeo/gdal/pull/6829


nirvn:
@nyalldawson , thanks.

PR #6831, opened 2022-12-01
ogr2ogr: densify points of spatial filter specified with -spat_srs to avoid reprojection artifacts

rouault:
The logic before this fix would select points that shouldn't be selected, or reject ones that ought to be included. Added test cases represent situations that resulted in unselected points before that fix.

CC @jratike80

### DIFF
--- a/apps/ogr2ogr_lib.cpp
+++ b/apps/ogr2ogr_lib.cpp
@@ -1183,7 +1183,23 @@ void ApplySpatialFilter(OGRLayer* poLayer, OGRGeometry* poSpatialFilter,
        poSpatialFilterReprojected->assignSpatialReference(poSpatSRS);
        OGRSpatialReference* poSpatialFilterTargetSRS  = poSourceSRS ? poSourceSRS : poLayer->GetSpatialRef();
        if( poSpatialFilterTargetSRS )
+       {
+           // When transforming the spatial filter from its spat_srs to the
+           // layer SRS, make sure to densify it sufficiently to avoid issues
+           constexpr double SEGMENT_DISTANCE_METRE = 10 * 1000;
+           if( poSpatSRS->IsGeographic() )
+           {
+               const double LENGTH_OF_ONE_DEGREE = poSpatSRS->GetSemiMajor(nullptr) * M_PI / 180.0;
+               poSpatialFilterReprojected->segmentize(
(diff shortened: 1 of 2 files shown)


apps/ogr2ogr_lib.cpp
```diff
@@ -1183,7 +1183,23 @@ void ApplySpatialFilter(OGRLayer* poLayer, OGRGeometry* poSpatialFilter,
        poSpatialFilterReprojected->assignSpatialReference(poSpatSRS);
        OGRSpatialReference* poSpatialFilterTargetSRS  = poSourceSRS ? poSourceSRS : poLayer->GetSpatialRef();
        if( poSpatialFilterTargetSRS )
+       {
+           // When transforming the spatial filter from its spat_srs to the
+           // layer SRS, make sure to densify it sufficiently to avoid issues
+           constexpr double SEGMENT_DISTANCE_METRE = 10 * 1000;
```
nyalldawson, line 1189: Could proj_trans_bounds be used here instead?

rouault: > Could proj_trans_bounds be used here instead?

If we were doing only bounding box checks, yes. But we do a bit more than just bounding box filtering, and actually use the densified polygon to do intersection checks. The use case behind this fix is for example requesting a OGC API Features server (returning only features in EPSG:4326 currently) with a bounding box expressed in EPSG:3067. So the OAPIF driver will use the reprojected bounding box from EPSG:3067 to EPSG:4326, but it will also used the densified rectangle reprojected to EPSG:4326 as a post filtering step.

nyalldawson: Ah - got it. So we really need some kind of "proj_trans_bounds_to_polygon_ring", which gives us all the "raw" transformed densified box...

rouault: That would bring geometry into PROJ, which we have avoided up to now. Not sure which would be the most natural place (PROJ ? GEOS ? GDAL ?) for that. Especially given that the densification could be done with different logics: here, when the CRS is geographic, we densify in the (latitude, longitude) space as if it was cartesian, which is OK for the use case. But you could possible also want to densify along geodesic / shortest lines on the ellipsoid

nyalldawson: > That would bring geometry into PROJ

I was thinking more along the lines of just raw axis arrays -- nothing special :)

PR #6837, opened 2022-12-02
RFC89: sql logging callback

elpaso:
RFC89: SQL query logging callback

### DIFF
--- a/doc/source/development/rfc/rfc89_sql_logging_callback.rst
+++ b/doc/source/development/rfc/rfc89_sql_logging_callback.rst
@@ -45,16 +45,13 @@ make database calls each time a SQL command is sent to the backend,
 in case of prepared queries the actual SQL after parameter
 substitutions should be sent.
 
-The callback may provide additional information about the executed
-query:
+The callback will provide additional information about the executed
+query (if available):
 
 - error string message
 - number of affected/retrieved records
-
-Further research is necessary to determine if the following 
-information could be also provided:
-
 - time taken to execute the query
+
 
 Example API:
 
@@ -77,7 +74,7 @@ Example API:
     }
 
 
-The callback function will be initially used by the GeoPackage driver only.
+The callback function will be initially used by the SQLite-based drivers only (GPKG included).
 
 
 Efficiency considerations
@@ -88,6 +85,10 @@ check if the function pointer is `nullptr` and call the function if it is
 not.
 
 The cost of the a.m. check is probably negligible on most architectures.
+
+In order to catch SQLite prepare errors, a prepare function wrapper will be 
+called instead of the sqlite3 API C function, this implies the cost of
+a function call (which might be inlined if necessary).
 
 
 Backward compatibility
--- a/doc/source/development/rfc/rfc89_sql_logging_callback.rst
+++ b/doc/source/development/rfc/rfc89_sql_logging_callback.rst
@@ -131,4 +131,4 @@ A C++ test will be added to the test suite.
 Voting history
 --------------
 
-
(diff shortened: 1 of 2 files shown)


doc/source/development/rfc/rfc89_sql_logging_callback.rst
```diff
@@ -0,0 +1,116 @@
...
+in case of prepared queries the actual SQL after parameter
+substitutions should be sent.
+
+The callback will provide additional information about the executed
+query (if available):
+
+- error string message
+- number of affected/retrieved records
+- time taken to execute the query
+
```
rouault, line 54: It should be documented that the callback may be called by multiple threads in a concurrent way, and that the implementation of the callback should be robust to that (use of lock typically). That may arise in drivers that use multi-threading as an optimization implementation detail. And this is typically the case of the GeoPackage driver in its Arrow stream interface. Cf https://github.com/OSGeo/gdal/blob/master/ogr/ogrsf_frmts/gpkg/ogrgeopackagetablelayer.cpp#L7099.

Regarding the implementation in the GPKG driver, it should around https://github.com/OSGeo/gdal/blob/master/ogr/ogrsf_frmts/gpkg/ogrgeopackagetablelayer.cpp#L7121 install the callback set initially on the main dataset to the ``task->m_poDS`` internal dataset that will run in a dedicated thread.

rouault:
you need to commit the updated index.rst to point to the RFC

doc/source/development/rfc/rfc89_sql_logging_callback.rst
elpaso: I've shamelessly copy-pasted your sentence.

elpaso: About the implementation, I've installed the callback this way: https://github.com/elpaso/gdal/commit/4b1bf394e5af24e7758aa31c7162ee61f9a768dd#diff-125788fb2cb0d32e396c0af12cb5bbc19b12a006603dddc29236a3afdb9c8096R4342

rouault: > I've installed the callback this way

My point regarding the multi-threaded optimization was that, to instrument everything, the callback also needs to be propagated by the driver itself from the main dataset to the auxiliary datasets it opens on the same file, and that should be done around https://github.com/OSGeo/gdal/blob/master/ogr/ogrsf_frmts/gpkg/ogrgeopackagetablelayer.cpp#L7121

elpaso: See https://github.com/elpaso/gdal/commit/4e3f0e8eb742599320ba9e1da4eb4ae988004a77

rouault:
@elpaso You can update the "Voting history" section with "+1 from PSC members MateuszL, JukkaR, TamasS and EvenR", and change its status to "Adopted, implemented" (well, "implemented" only after the implementation PR will have been merged)

elpaso:
@rouault done, I'll make the iplementation PR after rebase and reformat, probably tomorrow.

doc/source/development/rfc/rfc89_sql_logging_callback.rst
```diff
@@ -0,0 +1,134 @@
...
+=============================================================
+RFC 89: SQL logging callback
+=============================================================
+
+============== =============================================
+Author:        Alessandro Pasotti
+Contact:       elpaso @ itopen.it
+Started:       2022-Nov-30
+Status:        Draft
```
rouault, line 11:
```suggestion
Status:        Adopted, implemented
```

PR #6858, opened 2022-12-04
[sqlite] Implement AddRelationship support

nyalldawson:
Adds support for creating one-to-many relationships by adding foreign key constraints to fields.

### DIFF
--- a/autotest/ogr/ogr_sqlite.py
+++ b/autotest/ogr/ogr_sqlite.py
@@ -3895,6 +3895,238 @@ def test_ogr_sqlite_relationships():
 
 
 ###############################################################################
+# Test support for altering relationships
+
+
+def test_ogr_sqlite_alter_relations():
+    def clone_relationship(relationship):
+        res = gdal.Relationship(
+            relationship.GetName(),
+            relationship.GetLeftTableName(),
+            relationship.GetRightTableName(),
+            relationship.GetCardinality(),
+        )
+        res.SetLeftTableFields(relationship.GetLeftTableFields())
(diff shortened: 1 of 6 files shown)


autotest/ogr/ogr_sqlite.py
```diff
@@ -3894,6 +3894,181 @@ def test_ogr_sqlite_relationships():
...
+
+    relationship.SetLeftTableFields(["o_pkey"])
+    relationship.SetRightTableFields(["dest_pkey"])
+    ds.ExecuteSQL(
+        "CREATE UNIQUE INDEX origin_table3_o_pkey_idx ON origin_table3(o_pkey)"
+    )
+
+    assert ds.AddRelationship(relationship)
+
+    assert set(ds.GetRelationshipNames()) == {
```
rouault, line 4056: before adding the relationship, it could be interesting to add a feature in the table that gets foreign key constraints added and read it back after AddRelationship() to check the table was recreated OK. And perhaps do that after dataset closing and re-opening

nyalldawson: Do you mean a feature which would violate the foreign key constraint?

rouault: no, just a regular feature, to test that field recreation logic of AddForeignKeysToTable() does its job right. Currently I guess that if for example it was bogus and would just wipe out completely the table, that would be unnoticed by the current test.

nyalldawson: Ah, gotcha :+1:

PR #6872, opened 2022-12-07
OAPIF: implement Part 2 / CRS extension, and related enhancements

rouault:
-   OAPIF: add support for OGC API Features - Part 2 CRS extension
    
    storageCrs when found is used to set the layer CRS, express the spatial
    extent in that CRS, and set the crs and bbox-crs query parameters to
    request items/features in that CRS.
    
    storageCrsCoordinateEpoch is also used when found

- OAPIF driver: add CRS/PREFERRED_CRS open options to control the active CRS

- OAPIF: add a SERVER_FEATURE_AXIS_ORDER open option

- Add OGRLayer::GetSupportedSRSList() and SetActiveSRS(), and implement in OAPIF driver

- ogrinfo: output CRS supported list

- ogr2ogr: use SetActiveSRS() when possible when -t_srs is used

- WFS: implement GetSupportedSRSList() and SetActiveSRS()

### DIFF
--- a/apps/ogr2ogr_lib.cpp
+++ b/apps/ogr2ogr_lib.cpp
@@ -391,6 +391,9 @@ struct TargetLayerInfo
     bool         m_bPreserveFID = false;
     const char  *m_pszCTPipeline = nullptr;
     bool         m_bCanAvoidSetFrom = false;
+    const char*  m_pszSpatSRSDef = nullptr;
+    OGRGeometryH m_hSpatialFilter = nullptr;
+    const char*  m_pszGeomField = nullptr;
 };
 
 struct AssociatedLayers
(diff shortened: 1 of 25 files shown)


jratike80:
The current behavior will be changed if the OAPIF server provides storageCrs. Will that prevent backporting to 3.6.x?

rouault:
> Will that prevent backporting to 3.6.x?

Yes, I was a bit hesitant about that. It could potentially affect workflows where people always expect the driver to return OGC:CRS84.

jratike80:
Yes, I understand. The motivation for the work was exactly to have some effect on workflows and prevent small changes in coordinate values due to double transformation storageCrs -> CRS84 -> storageCrs.  Sure it can affect for example automatic tests.

rouault:
Let's merge this!

PR #6896, opened 2022-12-11
Fix PG and PGDump drivers to support the TEMPORARY layer CO

palmerj:
## What does this PR do?

Currently, the PG and PGDump drivers do not correctly support the TEMPORARY layer creation options. The current code is using pg_temp_1 as the schema name, which is incorrect. Instead the [pg_temp](https://www.postgresql.org/docs/current/runtime-config-client.html) alias should be used in cases where the schema is supplied. I've fixed both the PG and PGDump drivers. The PGDump driver is for correctness with the current documentation, noting that the benefits of using the feature in PGDump are likely minimal as the driver format is likely used for data exchange rather than processing within an OGR script.

## Tasklist

 - [x] Add test case(s)
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS: All 
* Compiler: All

### DIFF
--- a/ogr/ogrsf_frmts/pgdump/ogrpgdumpdatasource.cpp
+++ b/ogr/ogrsf_frmts/pgdump/ogrpgdumpdatasource.cpp
@@ -298,6 +298,13 @@ OGRPGDumpDataSource::ICreateLayer( const char * pszLayerName,
         }
     }
 
+    const bool bTemporary = CPLFetchBool( papszOptions, "TEMPORARY", false );
+    if( bTemporary )
+    {
+        CPLFree(pszSchemaName);
+        pszSchemaName = CPLStrdup("pg_temp");
+    }
+
     if ( pszSchemaName == nullptr)
     {
         pszSchemaName = CPLStrdup("public");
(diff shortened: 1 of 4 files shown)


ogr/ogrsf_frmts/pgdump/ogrpgdumpdatasource.cpp
```diff
@@ -534,7 +538,7 @@ OGRPGDumpDataSource::ICreateLayer( const char * pszLayerName,
 
         osCommand.Printf(
                 "SELECT AddGeometryColumn('%s',%s,'%s',%d,'%s%s',%d)",
-                pszSchemaName, pszEscapedTableNameSingleQuote, pszGFldName,
+                bTemporary ? '': pszSchemaName, pszEscapedTableNameSingleQuote, pszGFldName,
```
rouault, line 541:
```suggestion
                bTemporary ? "": pszSchemaName, pszEscapedTableNameSingleQuote, pszGFldName,
```

rouault:
you need to set-up pre-commit to have .py files correctly formatted: https://gdal.org/development/dev_practices.html#commit-hooks

rouault:
Just clicked the merge button. Thanks

PR #6899, opened 2022-12-12
GPKG/FlatGeoBuf: speed-up writing of DateTime/Date values

rouault:
With that, ``ogr2ogr /vsimem/out.gpkg nz-building-outlines.gpkg -lco spatial_index=no`` goes from 28 seconds to 19.


rouault:
Note: this also fixes writing DateTime with milliseconds with a locale where the decimal point is not dot, and when spatialite is not loaded (spatialite installs LC_NUMERIC = POSIX ! for the whole process)

PR #6903, opened 2022-12-12
RFC69: C++ code formatting - config, scripts and workflows

elpaso:
This PR contains only the scripts and configurations for the C++ clang format.


rouault:
included in https://github.com/OSGeo/gdal/pull/6937

PR #6928, opened 2022-12-15
Parquet: update to GeoParquet 1.0.0-beta.1 specification (fixes #6646)

### LINKED ISSUE #6646: Update to GeoParquet 0.5.0
when it is released

rouault:
Main changes in the implementation are:
- "geometry_type" is renamed as "geometry_types" and is always an array
- "bbox" is written with zmin and zmax when there are 3D geometries
- slightly improved deduction of OGR layer geometry type when there are mixes like LineString, LineString Z, MultiLineString in geometry_types

CC @jorisvandenbossche @paleolimbot @cholmes


jorisvandenbossche:
Thanks for the update!

PR #6956, opened 2022-12-20
VRT: add BLOCKXSIZE and BLOCKYSIZE creation options

rouault:


sgillies:
@rouault what does it mean for a VRT to have blocks? How do these blocks relate to the blocks of the datasets behind the VRT?

rouault:
> what does it mean for a VRT to have blocks?

All GDAL raster bands have blocks. By default the block size for a sourced-base VRT raster band is arbitrarily set to 128x128 (for a warped VRT 512x128), unless the VRT is made of a single source and this single source is not subsetted, in which case the block size of the unique source will be set as the VRT raster band block size. This PR allows the user to more easily customize the block size if needed (definitely an advanced setting).

Most of the time (always?), RasterIO() operations will not trigger the block cache on the VRT raster band but will be forwarded to the underlying source's RasterIO(). But it is still possible for code to use GDAL block based methods (ReadBlock(), etc.) on the VRT raster band, in which case the block size might have some impacts.

rouault:
https://github.com/OSGeo/gdal/pull/6963 tries to capture part of the above explanations in the doc

sgillies:
Thanks @rouault !

> Most of the time (always?), RasterIO() operations will not trigger the block cache on the VRT raster band but will be forwarded to the underlying source's RasterIO(). But it is still possible for code to use GDAL block based methods (ReadBlock(), etc.) on the VRT raster band, in which case the block size might have some impacts.

It was never completely clear to me that VRT blocks were not cached. This is good to know.

I'm curious about applications of this advanced setting. For example, some users are fond of using Rasterio WarpedVRTs to make tile and serve raster tiles. By setting the VRT block size to 512x512, do we reduce overhead when making 512x512 tiles?

rouault:
actually, warped VRT was the trigger for this pull request.

> By setting the VRT block size to 512x512, do we reduce overhead when making 512x512 tiles?

most likely as VRTWarpedRasterBand only implements IReadBlock(), and not IRasterIO(), so it forces block access. But of course warping has lots of degenerated situations, like warping polar areas from geographic CRS to polar projection where to generate a small area in projected CRS you may need to fetch whole source raster lines, that could make that rule not be always true... I didn't know what GDAL was when the 512x128 default was selected (back to https://github.com/OSGeo/gdal/commit/95dd6441e1e28cacf28d0fe5bb01ce6c84666a34#diff-706a04014708d2e5147ab8fc9748fbecc6caacadbb09f463d813fea032344bc2R250 )  ;-) I suspect it is a compromise to make it work reasonably for both tiled and strip organized source datasets of modest width and in situations where warping doesn't distort too much.

PR #6957, opened 2022-12-20
CSHARP: Add SkiaSharp

runette:
## What does this PR do?

@bjornharrtell @szekerest 

This is a reasonably complicated explanation:

1. The C# bindings build includes a number of sample apps. These are (excellent) code exemplars and also are used in all builds to test the C# bindings - but I don't think they are intended to be used for production or that anybody does so. The Sample applications use `System.Drawing.Bitmap` to manipulate bitmaps.
2. `System.Drawing.Bitmap` uses GDI+ (at least in the non-mono builds) and the cross-platform implementations of that have always been "interesting". Microsoft has clarified the situation by saying that `System.Drawing.Common` (that implements Bitmap) is NOT supported cross-platform (see https://learn.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/6.0/system-drawing-common-windows-only) after Net6. Net6 builds using this will generate a warning message and from Net7 there will be an error message.
3. In order to future-proof the use of the sample apps as tests, this PR does the following:
- Nothing to the build of actual binding libraries
- Nothing for builds using Net5 or earlier to build the Sample Apps (except for the deletion below)
- Nothing for builds using `mcs` to build the Sample Apps (except for the deletion below)
- For builds using Net6 or later to build the sample apps, the preferred MS replacements `System.Drawing.Primitives` and  `SkiaSharp` are used with a simple and simplistic compatibility layer so that `System.Drawing.Common` types are converted to `SkiaSharp` types. This is not intended to be a complete implementation - just to make sure that the GDAL API is exercised.
- One small piece of functionality is removed from `GDALAdjustContrast.cs` for all builds. This would be very complicated to implement as a compatibility layer and does not change the extent to which the GDAL API is tested.

None

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:

### DIFF
--- a/swig/csharp/CMakeLists.txt
+++ b/swig/csharp/CMakeLists.txt
@@ -92,7 +92,11 @@ endif ()
 
 # needed because of differences in the packages
 if (DOTNET_FOUND)
-  set(CSHARP_DRAWING "System.Drawing.Common")
+  if(_CSHARP_LIBRARY_SDK_NUMBER LESS "6")
+    set(CSHARP_DRAWING "System.Drawing.Common")
+  else ()
+    set(CSHARP_DRAWING "System.Drawing.Primitives;SkiaSharp")
+  endif ()
 else ()
   set(CSHARP_DRAWING "System.Drawing")
 endif ()
@@ -415,7 +419,7 @@ function (gdal_build_csharp_sample)
   endif (_GBCS_DEPENDS)
   if (_GBCS_SYSTEM_DEPENDS)
     foreach (_dep IN LISTS _GBCS_SYSTEM_DEPENDS)
-      if (NOT _dep MATCHES "^OSGeo*.")
+      if (_dep MATCHES "^System*.")
         list(APPEND CSC_OPTIONS /r:${_dep}.dll)
       endif ()
     endforeach ()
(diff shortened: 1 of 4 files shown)


rouault:
can you rebase on top of latest master to get the CI fix for build_mac ?

bjornharrtell:
I've played around with Skia and C# and it quite good but it's a large and complex dependency. But it's only added for the sample apps right? Then it seems ok to me.

runette:
> I've played around with Skia and C# and it quite good but it's a large and complex dependency. But it's only added for the sample apps right? Then it seems ok to me.

Yes - it is just for sample apps so that they can be used as tests for the bindings.

> can you rebase on top of latest master to get the CI fix for build_mac ?

done

runette:
> > can you rebase on top of latest master to get the CI fix for build_mac ?
> 
> done

@rouault but now the Alpine build is failing ?

rouault:
> now the Alpine build is failing ?

apparently related to an upgrade of libcurl in between :-) (alpine CI is tracking the "latest" moving channel, so those things can happen). Anyway unrelated to your changes and can be ignored.

rouault:
> now the Alpine build is failing ?

fixed in latest master

PR #6970, opened 2022-12-22
ogr2ogr: make conversion from GML2 to GPKG work without explicit -lco FID=some_name_different_than_fid

rouault:


jratike80:
So what happens if the GML fid is a string?

rouault:
> So what happens if the GML fid is a string?

it is kept as an attribute, and the GPKG fid column is named ``gpkg_fid``

PR #6971, opened 2022-12-22
HTTP: set default User-Agent header to GDAL/x.y.z and …

### LINKED ISSUE #6376: Default User-Agent
I know that there are some ways to set the value of User-Agent, for example:

```
$ GDAL_HTTP_USERAGENT="gdal/3.0.4" gdalinfo /vsicurl/https://...
```

But it would be nice if GDAL could have some default value for it (ideally with the version). This is how it is done in [requests](https://github.com/psf/requests/blob/7104ad4b135daab0ed19d8e41bd469874702342b/requests/utils.py#L899):

```python
>>> from requests.utils import default_user_agent
>>> default_user_agent()
'python-requests/2.16.5'
```

rouault:
…add a CPLHTTPSetDefaultUserAgent() function (fixes #6376)

### DIFF
--- a/port/cpl_http.cpp
+++ b/port/cpl_http.cpp
@@ -494,6 +494,28 @@ static int ProcessFunction(void *p, double dltotal, double dlnow,
 #endif /* def HAVE_CURL */
 
 /************************************************************************/
+/*                     CPLHTTPSetDefaultUserAgent()                     */
+/************************************************************************/
+
+static std::string gosDefaultUserAgent;
+
+/**
+ * \brief Set the default user agent.
+ *
+ * GDAL core will by default call this method with "GDAL/x.y.z" where x.y.z
+ * is the GDAL version number (during driver initialization). Applications may
+ * override it.
+ *
+ * @param pszUserAgent String (or nullptr to cancel the default user agent)
+ *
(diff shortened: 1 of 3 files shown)


port/cpl_http.cpp
```diff
@@ -921,16 +943,19 @@ int CPLHTTPPopFetchCallback(void)
 
 // NOTE: when adding an option below, add it in asAssocEnvVarOptionName[]
 
+// clang-format off
```
elpaso, line 946: Did you try ReflowComments false in clang-format options?

rouault: > Did you try ReflowComments false in clang-format options?

I just tried ``ReflowComments: false``, but it doesn't seem to help

elpaso: Maybe also add: PenaltyBreakComment: 1000000000

rouault: done in https://github.com/OSGeo/gdal/pull/6999 . That indeed fixed the weird reformatting of the Doxygen comments. But it had unpleasant consequences, where it generated invalid code (with '\' continuation character in comments), which I've manually fixed.

PR #6974, opened 2022-12-22
Add capability to quantize geometry coordinates for higher compression efficiency

rouault:
and GPKG: add creation options to quantify coordinates
    
    -  **XY_QUANTIZED_PRECISION** =float or string. (GDAL >= 3.7)
       Desired precision for X/Y coordinate quantization.
       This parameter will influence the number of nullified
       least significant bits in the mantissa of the binary representation of
       coordinates as IEEE-754 double-precision floating-point numbers.
       This is only useful if the file is compressed.
       By default, the value is understood as expressed in the units of the CRS
       of the layer (for example, for a geographic Earth CRS, "8.9e-9" corresponds
       to a millimetric precision), unless the 'm', 'cm' or 'mm' suffixes are added
       to express respectively the value in metre, centimetre or millimetre (will work
       for a geographic or projected CRS). For example, "1 mm" for a millimetric
       precision.
    -  **Z_QUANTIZED_PRECISION** =float or string. (GDAL >= 3.7)
       Desired precision for Z coordinate quantization.
       This parameter will influence the number of nullified
       least significant bits in the mantissa of the binary representation of
       coordinates as IEEE-754 double-precision floating-point numbers.
       This is only useful if the file is compressed.
       By default, the value is understood as expressed in the units of the CRS
       of the layer, unless the 'm', 'cm' or 'mm' suffixes are added
       to express respectively the value in metre, centimetre or millimetre (will work
       for a geographic or projected CRS). For example, "1 mm" for a millimetric
       precision.
    -  **M_QUANTIZED_PRECISION** =float. (GDAL >= 3.7)
       Desired precision for M coordinate quantization.
       This parameter will influence the number of nullified
       least significant bits in the mantissa of the binary representation of
       coordinates as IEEE-754 double-precision floating-point numbers.
       This is only useful if the file is compressed.

and ogr2ogr: add -xy_quantized_prec/-z_quantized_prec/-m_quantized_prec options

On a test dataset, setting XY_QUANTIZED_PRECISION=0.001m reduces the size of the .zip of the .gpkg file from 766 MB to 667 MB (13% size decrease).


rouault:
We should probably adopt the interface of https://postgis.net/docs/ST_QuantizeCoordinates.html to specify the decimal precision and compute the binary precision from that

ogr/ogr_api.cpp
```diff
@@ -1893,3 +1893,281 @@ int OGRGetNonLinearGeometriesEnabledFlag(void)
...
+ */
+
+void OGRPrecisionOptionsSetMetricPrecision(OGRPrecisionOptions *psOptions,
+                                           OGRSpatialReferenceH hSRS,
+                                           double dfXYMetricPrecision,
+                                           double dfZMetricPrecision,
+                                           double dfMPrecision)
+{
+    const OGRSpatialReference *poSRS = OGRSpatialReference::FromHandle(hSRS);
+    if (dfXYMetricPrecision != 0)
```
elpaso, line 2014: Do we miss a check for `poSRS` ?

ogr/ogr_geometry.h
```diff
@@ -307,6 +308,134 @@ class CPL_DLL OGRDefaultConstGeometryVisitor : public IOGRConstGeometryVisitor
...
+ * @param nBitsPrecision Desired precision (number of bits after integral part)
+ * @return quantized nVal.
+ * @since GDAL 3.7
+ */
+inline uint64_t OGRQuantizeValue(uint64_t nVal,
+                                 int nBitsPrecision) CPL_WARN_UNUSED_RESULT;
+
+inline uint64_t OGRQuantizeValue(uint64_t nVal, int nBitsPrecision)
+{
+    // Extact the binary exponent from the IEEE754 representation
```
elpaso, line 367:
```suggestion
    // Extract the binary exponent from the IEEE754 representation
```

ogr/ogr_geometry.h
```diff
@@ -307,6 +308,134 @@ class CPL_DLL OGRDefaultConstGeometryVisitor : public IOGRConstGeometryVisitor
...
+
+/************************************************************************/
+/*                        OGRQuantizeXYValues()                         */
+/************************************************************************/
+
+/** Quantize XY values.
+ *
+ * @since GDAL 3.7
+ */
+template <int SPACING>
```
elpaso, line 391: Why a template and not a simple argument?

elpaso (commented):
LGTM but I admit I've got lost in the WKB bit manipulations.

ogr/ogr_geometry.h
rouault: To generate better assembly code.   ptr + variable * constant can be expressed into a single assembly mnemonic

rouault:
I'm putting this on hold because I'm wondering if that shouldn't be part of a larger "framework" to convey coordinate precision between drivers

rouault:
closing as superseded per https://github.com/OSGeo/gdal/pull/9378

PR #6985, opened 2022-12-28
gdalwarp: add -srcband and -dstband options

rouault:
.. option:: -srcband <n>

    Specify a input band number to warp (between 1 and the number of bands
    of the source dataset). This option may be repeated multiple times, to warp
    several bands of the input dataset. The alpha band should not be specified
    in the list, as it will be automatically retrieved (unless :option:`-nosrcalpha`
    is specified).
    If :option:`-srcband` is not specified, all input bands are used.

.. option:: -dstband <n>

    Specify the output band number in which to warp.
    If :option:`-srcband` is specified, there must be as many occurences of
    :option:`-dstband` as there are of :option:`-srcband`.
    The output alpha band should not be specified, as it will be automatically
    created if the input dataset has an alpha band, or if :option:`-dstalpha`
    is specified.
    If :option:`-dstband` is not specified, then
    ``-dstband 1 -dstband 2 ... -dstband N`` is assumed where N is the number
    of input bands specified explicitly or implicitly.

### DIFF
--- a/alg/gdalwarpoperation.cpp
+++ b/alg/gdalwarpoperation.cpp
@@ -896,7 +896,7 @@ void GDALWarpOperation::CollectChunkList(int nDstXOff, int nDstYOff,
                 ->AdviseRead(nSrcXOff, nSrcYOff, nSrcX2Off - nSrcXOff,
                              nSrcY2Off - nSrcYOff, nDstXSize, nDstYSize,
                              psOptions->eWorkingDataType, psOptions->nBandCount,
-                             nullptr, nullptr);
+                             psOptions->panSrcBands, nullptr);
         }
     }
 }
--- a/apps/gdalwarp_bin.cpp
+++ b/apps/gdalwarp_bin.cpp
@@ -67,6 +67,7 @@ static void Usage(const char *pszErrorMsg = nullptr)
 {
     printf(
         "Usage: gdalwarp [--help-general] [--formats]\n"
+        "    [-srcband n]* [-dstband n]*\n"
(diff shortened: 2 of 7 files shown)


rouault:
CC @sgillies w.r.t https://github.com/rasterio/rasterio/pull/2437

tbonfort:
For consistency with gdal_translate, could we replace -srcband with -b ?

rouault:
> For consistency with gdal_translate, could we replace -srcband with -b ?

yes I was contemplating that too. [gdal_proximity](https://gdal.org/programs/gdal_proximity.html) uses -srcband and -dstband, hence my current choice. Or maybe have -b alias to -srcband ?

jratike80:
-srcband and -dstband work good together as a pair and they are more explicit. The -b alias might be good because it is also used in gdalbuildvrt.
The documentation of -srcband and -dstband is somewhat obscure but unfortunately I cannot suggest how to make it more clear that -dstband is optional and needed for changing the band numbers and order from the default 1,2,3... and that the first option affects the first band selected with -srcband and so on. Maybe there could be also an usage example about how to warp bands 2 and 4 from a source image and update bands 8 and 6 of the target image with them.

tbonfort:
aliasing -srcband and -b works for me, so that gdalwarp and gdal_translate work consistently in (the vast majority in my opinion) of cases where you need to select a subset of the original bands to create a new dataset.

rouault:
-b alias added, and documentation hopefully improved, including examples

tbonfort:
thanks @rouault , looks good to me

Issue #6486, opened 2022-10-07
Python bindings: add mapFieldType option to gdal.VectorTranslate()


vianmora:
Hello :) 

I searched for some project to add new options to vector translate, but i found nothing, 
for my project, i could need the `mapFieldType` option, 

Is it possible to add it ? Or someone could help me to do it please ?  

I found [this commit](https://github.com/OSGeo/gdal/commit/1c6f9839082f81b3906048f7160fee82adc96822) I could be inspired of, but as i'm new in the open source game, i wouldn't do something bad :/ 

It also could be the moment to add all the new options that are compatible with ogr2ogr : https://gdal.org/programs/ogr2ogr.htm  
- mapFieldType
- preserve_fid
- fid
- geomfield
- gt
- clipsrs, clipsrclayer, clipsrcwhere, clipdst, clipdstsql, clipdstlayer, clipdstwhere, 
- wrapdateline
- datelineoffset
- segmentize
- unsetFid
- relaxedFieldNameMatch
- unsetDefault
- fieldTypeToString
- unsetFieldWidth
- fieldmap
- splitlistfields
- maxsubfields
- explodecollections
- zfield 
- gcp
- s_coord_epoch , t_coord_epoch , a_coord_epoch 
- nomd
- mo
- noNativeData

rouault:
> I found [this commit](https://github.com/OSGeo/gdal/commit/1c6f9839082f81b3906048f7160fee82adc96822) I could be inspired of, but as i'm new in the open source game, i wouldn't do something bad :/

Yes you could certainly take inspiration from that to add new option(s). Don't be afraid to submit a PR. This will not do any harm. See https://github.com/OSGeo/gdal/blob/master/CONTRIBUTING.md first

PR #6988, opened 2022-12-28
Python bindings: VectorTranslate: add some extra options (#6486)

theroggy:
## What does this PR do?

Add some extra options to the python binding of VectorTranslate (in VectorTranslateOptions)

Related: #6486

## Tasklist

 - [x] Add test case(s)
 - [x] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS: developed on Windows, but should be cross platform
* Compiler: used python 3.9 while developing

### DIFF
--- a/autotest/utilities/test_ogr2ogr_lib.py
+++ b/autotest/utilities/test_ogr2ogr_lib.py
@@ -1418,6 +1418,7 @@ def test_ogr2ogr_lib_options_and_args():
 # Test using simplify
 
 
+@pytest.mark.skipif(not ogrtest.have_geos(), reason="GEOS is not available")
 def test_ogr2ogr_lib_simplify():
 
     src_ds = gdal.GetDriverByName("Memory").Create("", 0, 0, 0, gdal.GDT_Unknown)
--- a/autotest/utilities/test_ogr2ogr_lib.py
+++ b/autotest/utilities/test_ogr2ogr_lib.py
@@ -480,7 +480,8 @@ def test_ogr2ogr_lib_21():
 ###############################################################################
 
 
-def test_ogr2ogr_clipsrc_no_dst_geom():
+@pytest.mark.parametrize("args_using_options", [True, False])
+def test_ogr2ogr_clipsrc_wkt_no_dst_geom(args_using_options):
 
     if not ogrtest.have_geos():
         pytest.skip()
@@ -490,9 +491,14 @@ def test_ogr2ogr_clipsrc_no_dst_geom():
 
     tmpfilename = "/vsimem/out.csv"
     wkt = "POLYGON ((479461 4764494,479461 4764196,480012 4764196,480012 4764494,479461 4764494))"
-    ds = gdal.VectorTranslate(
-        tmpfilename, "../ogr/data/poly.shp", options='-f CSV -clipsrc "%s"' % wkt
-    )
+    if args_using_options:
+        ds = gdal.VectorTranslate(
+            tmpfilename, "../ogr/data/poly.shp", options='-f CSV -clipsrc "%s"' % wkt
+        )
+    else:
+        ds = gdal.VectorTranslate(
+            tmpfilename, "../ogr/data/poly.shp", format="CSV", clipSrc=wkt
+        )
     lyr = ds.GetLayer(0)
     fc = lyr.GetFeatureCount()
     assert fc == 1
@@ -1030,7 +1036,8 @@ def test_ogr2ogr_lib_t_srs_ignored():
     gdal.PopErrorHandler()
     gdal.Unlink("/vsimem/out.txt")
     assert got_msg == [
-        "Target SRS WGS 84 / UTM zone 31N not taken into account as target driver likely implements on-the-fly reprojection to WGS 84"
+        "Target SRS WGS 84 / UTM zone 31N not taken into account as target driver "
+        "likely implements on-the-fly reprojection to WGS 84"
     ]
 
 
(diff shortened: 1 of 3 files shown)


swig/python/osgeo/gdal.py
```diff
@@ -962,25 +962,42 @@ def VectorTranslateOptions(options=None, format=None,
...
+         clipSrcWhere=None,
+         clipDst=None,
+         clipDstSQL=None,
+         clipDstLayer=None,
+         clipDstWhere=None,
+         simplifyTolerance=None,
+         segmentizeMaxDist=None,
          makeValid=False,
+         mapFieldType=None,
+         explodecollections=False,
```
rouault, line 986: camelCase to apply here and in other related places:
```suggestion
         explodeCollections=False,
```

swig/python/osgeo/gdal.py
```diff
@@ -1087,24 +1153,94 @@ def VectorTranslateOptions(options=None, format=None,
                     val += ','
                 val += item
             new_options += ['-select', val]
+        
+        if srcDatasetOpenOptions is not None:
```
rouault, line 1157: that can only work with the ogr2ogr binary, not gdal.VectorTranslate(), unless you do special processing of them when opening the dataset

swig/python/osgeo/gdal.py
```diff
@@ -1087,24 +1153,94 @@ def VectorTranslateOptions(options=None, format=None,
...
+            for opt in srcDatasetOpenOptions:
+                new_options += ['-oo', opt]
+        if dstDatasetOpenOptions is not None:
```
rouault, line 1160: same as above

autotest/utilities/test_ogr2ogr_lib.py
```diff
@@ -490,9 +491,14 @@ def test_ogr2ogr_clipsrc_no_dst_geom():
...
     wkt = "POLYGON ((479461 4764494,479461 4764196,480012 4764196,480012 4764494,479461 4764494))"
-    ds = gdal.VectorTranslate(
-        tmpfilename, "../ogr/data/poly.shp", options='-f CSV -clipsrc "%s"' % wkt
-    )
+    if args_using_options:
+        ds = gdal.VectorTranslate(
+            tmpfilename, "../ogr/data/poly.shp", options='-f CSV -clipsrc "%s"' % wkt
+        )
+    else:
+        ds = gdal.VectorTranslate(
```
rouault, line 499: To make things simpler, I would just test this branch.  The use of -clipsrc was only because there was no corresponding keyword option

swig/python/osgeo/gdal.py
```diff
@@ -962,25 +962,42 @@ def VectorTranslateOptions(options=None, format=None,
          forceNullable=False,
          emptyStrAsNull=False,
          spatFilter=None, spatSRS=None,
+         srcDatasetOpenOptions=None,
```
rouault, line 965: This file is a generated file. The master file to modify is swig/include/python/gdal_python.i , and then run cmake with -DSWIG_REGENERATE_PYTHON:BOOL=ON and with SWIG 4.0.2 to regenerate swig/python/osgeo/gdal.py

theroggy:
There are two things I'd like to raise that I noticed while writing tests:
- In test_ogr2ogr_lib_clipsrc_datasource, if VectorTranslate is executed without filter, so using the 3 rows in the clip datasource, this gives results in an error and thus an empty output datasource. I'm don't know how the clip functionality works under the hood... but I suppose this is not wanted behaviour?
- Something I noticed before, but while writing a test for it I was reminded of it: the options parameter of VectorTranslateOptions() has some weird behaviour, which is partly documented, partly not document, but feels a bit weird to me in general:
  - when you provide an "options" keyword as a string, it is parsed as a commandline string, which is OK. However, in this case all other keywords are ignored, which is a bit of a pity I think: this way you need to choose either one or the other way.
  - when you provide an "options" keyword as a list, the contents of "options" is used in combination with all "standard" keywords specified when calling VectorTranslateOptions()... which is better in my opinion, but the extra parameters are also actually added in the "options" list and thus the input variable is changed, which is a bit weird?

rouault:
> In test_ogr2ogr_lib_clipsrc_datasource, if VectorTranslate is executed without filter, so using the 3 rows in the clip datasource, this gives results in an error and thus an empty output datasource

if using clipSrc without filter, then all input features are fetched, and their geometries intersected with the clipSrc one.
what's the error eactly ?

> * when you provide an "options" keyword as a string, it is parsed as a commandline string, which is OK. However, in this case all other keywords are ignored, which is a bit of a pity I think: this way you need to choose either one or the other way.

that's intended. Could potentially be improved but you'd have to define the behaviour when there are conflicting settings set through options or keywords

> when you provide an "options" keyword as a list,

huh, wasn't aware that was possible. That's not documented usage

theroggy:
> > In test_ogr2ogr_lib_clipsrc_datasource, if VectorTranslate is executed without filter, so using the 3 rows in the clip datasource, this gives results in an error and thus an empty output datasource
> 
> if using clipSrc without filter, then all input features are fetched, and their geometries intersected with the clipSrc one. what's the error eactly ?

ERROR 1: TopologyException: Input geom 1 is invalid: Self-intersection at 1 2
translation of layer test (use -skipfailures to skip errors)

> > * when you provide an "options" keyword as a string, it is parsed as a commandline string, which is OK. However, in this case all other keywords are ignored, which is a bit of a pity I think: this way you need to choose either one or the other way.
> 
> that's intended. Could potentially be improved but you'd have to define the behaviour when there are conflicting settings set through options or keywords

True... then again, using the "options" keyword by itself isn't veery "safe" either... you can also make any combination and/or duplication of options...

> > when you provide an "options" keyword as a list,
> 
> huh, wasn't aware that was possible. That's not documented usage

I've been a happy user of it for quite some years already :-)...

rouault:
> I suppose the clipSrc features are first unioned before intersecting them with the input features?

ah, no, they are not unioned. A multipolygon is created by adding each polygon in it. But if those polygons overlap, this results in an invalid multipolygon from a OGC Simple Features point of view, which is the GEOS point of view. We should probably union though. But perhaps keep that for a later PR. Can you change your test to only have a single feature in the clip dataset, as you just need to test the clipSrc python keyword and not the underlying functionality ?

> then again, using the "options" keyword by itself isn't very "safe" either... 

yes, it is undefined behaviour if you mix both usage. I have no definite opinion on whether we should raise an exception when mixing both or try to hybrid values as you suggested.

theroggy:
> > I suppose the clipSrc features are first unioned before intersecting them with the input features?
> 
> ah, no, they are not unioned. A multipolygon is created by adding each polygon in it. But if those polygons overlap, this results in an invalid multipolygon from a OGC Simple Features point of view, which is the GEOS point of view. We should probably union though.

To get the "typical" clip behaviour the union is needed to avoid that the input features are "split" by "extra" boundaries in the layer that is clipped by. To illustrate, the yellow parcel to the bottom left would be split in thow if the clip layer (purple in this example) wouldn't be unioned first.
https://geofileops.readthedocs.io/en/stable/api/geofileops.clip.html#geofileops.clip
> But perhaps keep that for a later PR.

Yes, is definitely not in scope for this PR.

> Can you change your test to only have a single feature in the clip dataset, as you just need to test the clipSrc python keyword and not the underlying functionality ?

Sure.

> > then again, using the "options" keyword by itself isn't very "safe" either...
> 
> yes, it is undefined behaviour if you mix both usage. I have no definite opinion on whether we should raise an exception when mixing both or try to hybrid values as you suggested.

No, I mean that I supposes that even if you don't mix both, but only use the options keyword, you can specify any option double,... or am I wrong?

rouault:
> you can specify any option double,... or am I wrong?

ah like doing ``gdal.Translate("out.tif", "in.tif", options = "-b 1 -b 1")`` ? ... which is actually a perfectly fine usage to duplicate an input band. But there are definitely usages where repeating a same option doesn't make sense. This is the utility responsibility to deal with that. Most of the time, they will only take into account the last occurrence, silently ignoring the first occurences, when repetition of the option isn't documented as valid.

theroggy:
> > you can specify any option double,... or am I wrong?
> 
> ah like doing `gdal.Translate("out.tif", "in.tif", options = "-b 1 -b 1")` ? ... which is actually a perfectly fine usage to duplicate an input band. But there are definitely usages where repeating a same option doesn't make sense. This is the utility responsibility to deal with that. Most of the time, they will only take into account the last occurrence, silently ignoring the first occurences, when repetition of the option isn't documented as valid.

Yes, this is indeed what I mean... Applied to the case where a combination of the options keyword and other keywords are used in VectorTranslate the options specified using the "other" keywords will be appended at the end of the options... so based on the logic you explain, they will "win" in the case of invalid repetitions...
But, obviously the typical case I'm using it for is to be able to add options that aren't supported yet in VectorTranslateOptions, so specifying "double" isn't really the typical case.

theroggy:
The failing check doesn't seem related with my changes... a WMS check... but I'm not able to restart the failed check...

autotest/utilities/test_ogr2ogr_lib.py
```diff
@@ -1163,3 +1346,108 @@ def test_ogr2ogr_lib_fid_string_to_gpkg():
...
+        "",
+        "../ogr/data/poly.shp",
+        format="Memory",
+        selectFields=["eas_id", "prfedea"],
+        options=raw_options_list,
+    )
+
+    # Note: the raw_options_list parameter is changed: all options specified as standard
+    # arguments are also added to the list.
+    assert len(raw_options_list) == 6
```
rouault, line 1394: This is really testing an implementation detail that user shouldn't rely upon. I think this assertion should be removed. Below asserts are sufficient to check the combination is done.

rouault, line 1394:
```suggestion
```

PR #7024, opened 2023-01-05
Added link to Rust GDAL bindings in `index.rst`.

metasim:


doc/source/api/index.rst
```diff
@@ -109,6 +109,7 @@ API
       .. toctree::
        :maxdepth: 1
 
+       Rust <https://github.com/georust/gdal>
```
rouault, line 112: the list is apparently in alphabetical order, so this should be moved at the end after R

metasim: @rouault Sorry I missed that. Thanks for taking care of it!

PR #7035, opened 2023-01-08
Add a read-only /vsi7z/ virtual file system (depends on libarchive)

### LINKED ISSUE #3344: support .7z files

The ability to load layers from .zip files is great, it would be nice to be able to do the same from .7z files, too.

## Additional context
The French IGN just released a lot of their data in open data. Most of their files are compressed with 7z, and the format seems to be used a lot by other GIS providers. I guess QGIS is not supposed to be able to load data from whatever archive format, but adding support for this one would be a nice addition. (In the meantime, repacking the archives to .zip works)

## Operating system

Debian sid

3.2.1 (Debian sid archive)

rouault:
Fixes #3344

/vsi7z/ is a file handler that allows reading [7z](https://en.wikipedia.org/wiki/7z)
archives on-the-fly without decompressing them beforehand. This file system is
read-only. Directory listing and `VSIStatL` are available, similarly
to /vsizip/ or /vsitar/ file systems.

It requires GDAL to be built against [libarchive](ttps://libarchive.org/)
(and libarchive having LZMA support to be of practical use).

To point to a file inside a 7z file, the filename must be of the form
`/vsi7z/path/to/the/file.7z/path/inside/the/7z/file`, where
`path/to/the/file.7z` is relative or absolute and `path/inside/the/7z/file`
is the relative path to the file inside the archive.`

Default extensions recognized by this virtual file system are:
``7z``, ``lpk`` (Esri ArcGIS Layer Package), ``lpkx``, ``mpk`` and ``mpkx``

An alternate syntax is available so as to enable chaining and not being
dependent on those extensions, e.g.: `/vsi7z/{/path/to/the/archive}/path/inside/the/archive`.
Note that `/path/to/the/archive` may also itself use this alternate syntax.

Note that random seeking within a large compressed file will be inefficient when
backward seeking is needed (decompression will be restarted from the start of the
file). Performance will be the best in sequential reading.


port/cpl_vsil_libarchive.cpp
```diff
@@ -0,0 +1,532 @@
...
+    CPL_DISALLOW_COPY_ASSIGN(VSILibArchiveFilesystemHandler)
+
+    const std::string m_osPrefix;
+    virtual const char *GetPrefix() override
+    {
+        return m_osPrefix.c_str();
+    }
+    virtual std::vector<CPLString> GetExtensions() override
+    {
+        return {".7z", ".lpk", ".lpkx"};
```
mtrcn, line 444: Can we add map package extensions here?

mtrcn (commented):
Thanks for your effort on this improvement. I think we should also support Map Package extensions; *.mpkx and *.mpk.

On the other hand specifying relative path for map/layer packages could be tricky as they have different paths for each version such as p20, p30, p40.. so on.

It could be nice to specify version to read or passing ‘latest’ to read from what latest version available in the package so that I can specify relative path by ignoring pX0 folder.

rouault:
> On the other hand specifying relative path for map/layer packages could be tricky as they have different paths for each version such as p20, p30, p40.. so on.

The alternate syntax ``/vsi7z/{/path/to/the/archive.p20}/path/inside/the/archive`` can be used for that

port/cpl_vsil_libarchive.cpp
rouault: done

mtrcn:
> > On the other hand specifying relative path for map/layer packages could be tricky as they have different paths for each version such as p20, p30, p40.. so on.
> 
> 
> 
> The alternate syntax ``/vsi7z/{/path/to/the/archive.p20}/path/inside/the/archive`` can be used for that

It is true that you can specify px0 path if you know which ArcGIS Desktop client version was used to generate the package beforehand.

port/cpl_vsil_libarchive.cpp
mtrcn: Amazing, thanks a lot

PR #7036, opened 2023-01-08
cpl_http: Change behaviour - CUSTOMREQUEST option overrides POST request method while sending form

BishopGIS:
## What does this PR do?
Current CPLHTTPFetch function only allows to send form using POST request method.  This PR also add ability to send form using PUT or PATCH request method.

### DIFF
--- a/port/cpl_http.cpp
+++ b/port/cpl_http.cpp
@@ -2413,14 +2413,14 @@ void *CPLHTTPSetOptions(void *pcurl, const char *pszURL,
             unchecked_curl_easy_setopt(http_handle, CURLOPT_POSTFIELDS,
                                        pszPost);
         }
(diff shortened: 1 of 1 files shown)


rouault:
@BishopGIS See https://gdal.org/development/dev_practices.html#commit-hooks to have the pre-commit hooks to take into account the new code formatting rules of https://gdal.org/development/rfc/rfc69_cplusplus_formatting.html

BishopGIS:
> @BishopGIS See https://gdal.org/development/dev_practices.html#commit-hooks to have the pre-commit hooks to take into account the new code formatting rules of https://gdal.org/development/rfc/rfc69_cplusplus_formatting.html

Thanks Even. I missed this.

PR #7041, opened 2023-01-09
CMake: fix failure to lookup dependent package if transitively already found (fixes #7040)

### LINKED ISSUE #7040: Failure to lookup dependent package if transitively already found
https://github.com/OSGeo/gdal/blob/516e10a58d1a218afcbae9988956e078bf1ea769/cmake/helpers/CheckDependentLibraries.cmake#L113

is a problem and stops `gdal` to lookup `zstd` correctly if `zstd` was already looked up by `tiff`. So `tiff` calls for `zstd` -> `ZSTD_FOUND` set. gdal uses target `ZSTD:zstd` but forgot to setup that target since upstream target names differs. 

Generally don't guard `find_package(pkg)` calls with `pkg_FOUND` stuff.

rouault:
The current logic that tests ``if (NOT DEFINED ${key}_FOUND)`` is a problem and stops `gdal` to lookup `zstd` correctly if `zstd` was already looked up by `tiff`. So `tiff` calls for `zstd` -> `ZSTD_FOUND` set. gdal uses target `ZSTD:zstd` but forgot to setup that target since upstream target names differs.

Fix suggested by @Neumann-A


rouault:
CC @dg0yt

rouault:
turning as draft. probably not the correct fix

PR #7048, opened 2023-01-10
CMake: Fix integration of find_package2

### LINKED ISSUE #7040: Failure to lookup dependent package if transitively already found

dg0yt:
## What does this PR do?

find_package2 is a custom alternative to regular find_package. gdal_check_package was meant to not call find_package when find_pacakge2 returned a positive or negative result. But the previous guard also interfered with the standard CMake pattern of repeated calls to find_package.
This commit changes the control over the alternative:
- find_package2 is called exclusively iff a definition exists for the module of interest.
- Even this call can be disabled by the user via `FIND_PACKAGE2_<pk>_ENABLED` in order to enable a regular find_package call.

Fixes #7040
Closes #7041.

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments


dg0yt:
This PR is fine from my POV. But I would like point out that it draws a clear line between `find_package2` and `find_package`. This PR doesn't allow `find_package` as a fallback when `find_package2` is configured and fails. But this isn't a regression, and I assume that this is the intended design of the original code.
On the positive side, the user-controlled switching variable allows for using CMake config when available. In `gdal_check_package`, we already have a good set of variables to adapt to new configs without patching.

PR #7050, opened 2023-01-10
OpenFileGDB geometry builder: fix comment related to polygon reconstruction (fixes #7047)

### LINKED ISSUE #7047: Possible inefficiency in reading GDB tables
At https://github.com/OSGeo/gdal/blob/master/ogr/ogrsf_frmts/openfilegdb/filegdbtable.cpp#L3742 the comment states that the slow organizePolygons route is not taken by default. Yet this method will always be called, because ASSUME_INNER_RINGS_IMMEDIATELY_AFTER_OUTER_RING is never defined anywhere. 

Is this an intentional change and the comment is wrong, or should ASSUME_INNER_RINGS_IMMEDIATELY_AFTER_OUTER_RING be defined so that the slow call is avoided by default?

rouault:
@nyalldawson you're right the comment was erroneous. I've run git blame to reconstruct the history of this code. hopefully my updated comments should give a picture on why we do the slow method currently.
Have you hit perf issues in that code path ? Perhaps organizePolygons() could use GEOS in a better way (actually it tries to avoid using it as much as possible when it is not strictly needed, by doing cheap topological checks first)

### DIFF
--- a/ogr/ogrsf_frmts/openfilegdb/filegdbtable.cpp
+++ b/ogr/ogrsf_frmts/openfilegdb/filegdbtable.cpp
@@ -3735,11 +3735,27 @@ FileGDBOGRGeometryConverterImpl::GetAsGeometry(const OGRField *psField)
                 poPoly->addRingDirectly(papoRings[0]);
             }
             else
+            // Note: ASSUME_INNER_RINGS_IMMEDIATELY_AFTER_OUTER_RING is *not* defined
 #ifdef ASSUME_INNER_RINGS_IMMEDIATELY_AFTER_OUTER_RING
                 if (bUseOrganize || !(papoRings[0]->isClockwise()))
 #endif
             {
-                /* Slow method : not used by default */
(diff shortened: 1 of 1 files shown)


nyalldawson:
> Have you hit perf issues in that code path ?

Yes, but admittedly it was when using a stupidly complex layer (a vector layer created by someone who polygonised a binary raster, so it had many polygons with thousands of rings). Actually, to be more precise -- I encountered the issue when converting the layer from a GPKG to a GDB. The actual GDB feature writing was fine, but the performance issue was hit when the driver creates the spatial index and is re-reading back in the written GDB geometries. 

> Perhaps organizePolygons() could use GEOS in a better way (actually it tries to avoid using it as much as possible when it is not strictly needed, by doing cheap topological checks first)

I wondering that too... perhaps by using prepared geos geometries for the polygons when there's say > 20 polygons being organised (based on https://github.com/libgeos/geos/pull/777#issuecomment-1355974174)

In any case, the new comments and explanation are appreciated!

PR #7052, opened 2023-01-11
gdal_translate: make -colorinterp work on a source band that is a mask band

rouault:


rouault:
backported to 3.6 (except commit 7ca0cb0 which isn't appropriate for stable branch)

Issue #6976, opened 2022-12-24
When an SQLITE sql statement returns no rows, the VectorTranslate output file is often missing or invalid 


theroggy:
## Expected behavior and actual behavior.

When using VectorTranslate to run an sql statement of the 'SQLITE' dialect on an empty file, the sql statement doesn't return any rows or if the geometries returned by the query is NULL, often there is either no output file or the output file is invalid.
The expected behaviour would be that in all situations a valid output file/layer would be written, containing the columns consistent with the sql query but no rows.
In some cases it might be unclear which geometry type the output file should have, but when an output geometry type is specified I guess this should never be an issue?

There are a lot of different situations leading to different outputs, so I ended up writing a script thats lists many situations and its results. I'll post a comment with the full listing + a comment with the script.

The big lines:
- for some SQLITE sql statements, if they start with '\n', '\t' or ' ' and the input file is an empty (no rows) .shp file, no output file will be created. If the statement doesn't start like that, output is ok.
- if a simple SQLITE sql statement starts with '\n' or '\t' and the input file is an empty (no rows) .gpkg file, the output file will be invalid, if the simple sql statement doesn't start with these, the output is OK.
- when a SQLITE sql statement is ran on an empty (no rows) .gpkg file, an invalid .gpkg is created (when it is opened, an exception is thrown that the file is "not recognized as a supported file format"
- when a SQLITE sql statement is ran on a .gpkg file and it returns no rows or NULL geometries, depending on whether (output)geometrytype is specified or not another variant of invalid output file is written.
- if NULL geometries are returned when running the SQLITE sql statement on a .gpkg, in some cases VectorTranslate throws an exception. In all other cases no exception is thrown.

## Steps to reproduce the problem.

Script in following comment.

## Operating system

Windows 10.

## GDAL version and provenance

gdal 3.6.1 installed from conda-forge (build py39h3be0312_1)

theroggy:
Script to test many situations:
```
import logging
import os
from pathlib import Path
import shutil

from osgeo import gdal
import osgeo.ogr as ogr
import osgeo.osr as osr

logging.basicConfig(level=logging.DEBUG)
gdal.UseExceptions()
# gdal.ConfigurePythonLogging(logger_name="gdal", enable_debug=True)
# os.environ["CPL_LOG_ERRORS"] = "ON"


def run_test(driver, src_empty, sql_descr, sql, sql_prefix, dir="."):
    # Prepare empty test input file
    # -----------------------------
[...]

PR #7056, opened 2023-01-12
Fix ogr2ogr error + invalid gpkg output file in some cases

theroggy:
Proposal to fix some cases where ogr2ogr and VectorTranslate write invalid gpkg output files.

When -nlt (or geometrytype in VectorTranslate) is specified in combination with one of the following situations:
  - when executing an sql statement returning NULL geoms
  - when executing an sql statement doing a spatial operation (eg. ST_Buffer) on the geom that returns no rows

This results in:
- the following error being written: ERROR 1: sqlite3_exec(CREATE TABLE "belgie_empty2" ( "fid" INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL, "geom" POLYGON, "geom" TEXT)) failed: duplicate column name: geom
- the gpkg written is invalid: gives an exception when opened.

These are cases 3.i and 4.i in the following issue: #6976

The change in this PR avoids that an attribute column having the same name as the present geometry column is retained in the create table statement.

### DIFF
--- a/ogr/ogrsf_frmts/gpkg/ogrgeopackagetablelayer.cpp
+++ b/ogr/ogrsf_frmts/gpkg/ogrgeopackagetablelayer.cpp
@@ -4973,7 +4973,8 @@ CPLString OGRGeoPackageTableLayer::GetColumnsOfCreateTable(
     for (size_t i = 0; i < apoFields.size(); i++)
     {
         OGRFieldDefn *poFieldDefn = apoFields[i];
-        if ((eGType != wkbNone) && (stricmp(poFieldDefn->GetNameRef(), GetGeometryColumn())))
+        if ((eGType != wkbNone)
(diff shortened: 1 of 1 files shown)


rouault:
I'm afraid fixing that in the table creation of the GeoPackage driver is a very limited band-aid. If you would output to other formats supported by GDAL, like PostGIS, you would likely run into the same issue. We don't want to patch all GDAL output drivers with that.
The annoying part of my criticism is that I can't really point to what the clean solution would be.
So at the very least, you should insert a comment in the code itself explaining why we do this check, pointing to the original github issue, and mentioning that it is just a band-aid to a yet-to-be-determined-clean solution.

I'm not completely sure about ``stricmp(a,b)`` portability. I'd suggest using the ``!EQUAL(a, b)`` macro instead we use extensively in the code base

Regarding the formating issue, you need to setup precommit to get a commit hook that will autoreformat. See https://gdal.org/development/dev_practices.html#commit-hooks

PR #7060, opened 2023-01-13
Add read-only vector GTFS (General Transit Feed Specification) driver

rouault:


ogr/ogrsf_frmts/CMakeLists.txt
```diff
@@ -96,6 +96,8 @@ ogr_dependent_driver(cad "OpenCAD" "GDAL_USE_OPENCAD OR GDAL_USE_OPENCAD_INTERNA
 ogr_dependent_driver(parquet "Parquet" "GDAL_USE_PARQUET")
 ogr_dependent_driver(arrow "Arrow" "GDAL_USE_ARROW")
 
+ogr_dependent_driver(gtfs "GTFS" "OGR_ENABLE_DRIVER_CSV")
```
elpaso, line 99: Shouldn't it be OGR_ENABLE_DRIVER_GTFS ?

ogr/ogrsf_frmts/gtfs/ogrgtfsdriver.cpp
```diff
@@ -0,0 +1,724 @@
...
+    int m_nLatIdx = -1;
+    int m_nLonIdx = -1;
+    bool m_bIsTrips = false;
+    bool m_bPrepared = false;
+    std::map<std::string, std::pair<double, double>> m_oMapStopIdToLonLat{};
+    std::map<std::string, std::map<int, std::string>> m_oMapTripIdToStopIds{};
+
+    void PrepareTripsData();
+
+  public:
```
elpaso, line 88: I'm not sure about GDAL coding conventions, but I would expect to find the `public` interface first and an explicit `private` secion below it.

ogr/ogrsf_frmts/gtfs/ogrgtfsdriver.cpp
```diff
@@ -0,0 +1,724 @@
...
+                poFeature->GetFieldAsInteger(nStopSequenceIdx);
+            if (pszStopId && pszTripId)
+            {
+                m_oMapTripIdToStopIds[pszTripId][nStopSequence] = pszStopId;
+            }
+        }
+    }
+    catch (const std::exception &)
+    {
+        CPLError(CE_Failure, CPLE_OutOfMemory, "Not enough memory");
```
elpaso, line 264: Just wondering: is OOM the only case when the `try` above can `throw` ?

ogr/ogrsf_frmts/gtfs/ogrgtfsdriver.cpp
```diff
@@ -0,0 +1,724 @@
...
+        poFeature->SetFID(poSrcFeature->GetFID());
+        auto poSrcLayerDefn = m_poUnderlyingLayer->GetLayerDefn();
+        for (int i = 0; i < nFieldCount; ++i)
+        {
+            const auto eType = m_poFeatureDefn->GetFieldDefn(i)->GetType();
+            if (poSrcLayerDefn->GetFieldDefn(i)->GetType() == eType)
+            {
+                poFeature->SetField(i, poSrcFeature->GetRawFieldRef(i));
+            }
+            else if (eType == OFTDate)
```
elpaso, line 295: Maybe a switch? I see only date and int are actually handled and default to string.

elpaso (approved):

ogr/ogrsf_frmts/CMakeLists.txt
rouault: no, the GTFS driver requires (at runtime) the CSV driver

ogr/ogrsf_frmts/gtfs/ogrgtfsdriver.cpp
rouault: we don't have that much conventions regarding that. The pattern with private first is actually quite common in GDAL sources

ogr/ogrsf_frmts/gtfs/ogrgtfsdriver.cpp
rouault: I think so. The risk is here is too many objects beeing put in the std::map. I'll change it to specifically catch std::bad_alloc though

ogr/ogrsf_frmts/gtfs/ogrgtfsdriver.cpp
rouault: A switch construct isn't the most natural one here given that the first ``if`` doesn't test eType against a constant but against an expression

PR #7069, opened 2023-01-15
OpenFileGDB writer: reserve arrays with an exponential strategy

rouault:
Hopefully this should fix Coverity warning:
```
CID 1502589 (#4 of 4): Inefficient vector resizing with reserve.
(INEFFICIENT_RESERVE)
1. inefficient_reserve: A std::vector has its capacity adjusted with
   reserve() in a manner that is likely to have severe performance penalties.
To avoid a quadratic time penalty when using reserve(), always
increase the capacity by a multiple of its current value
```

CC @nyalldawson


nyalldawson (approved):
Looks good to me!

PR #7073, opened 2023-01-16
gdalwarp: preserve source resolution by default when no reprojection …

rouault:
…is involved (behaviour change), and add '-tr square' to use previous behavior


mdsumner:
I have a related wish item where one could override the -te provided with one aligned to the source and a given -ts (I was thinking -snap in/out/near), I'm intending to PR this at some  point but, but it occurs that -tr align might be a god option. 

I don't yet have my head around all of the auto options, I mostly use -te and -ts explicitly and do my own alignment for -te upfront, just wanted to inject this thought in case it's somehow incompatible with the current scheme/s.

rouault:
> I have a related wish item where one could override the '-te' provided with one aligned to the source

aren't you looking for https://gdal.org/programs/gdalwarp.html#cmdoption-gdalwarp-tap ?

mdsumner:
no (I had to go and check for exactly why), I want the output extent to snap based on an *approximate* input '-te' and an input '-ts', I don't want to input '-tr' for this case. I'll illustrate this properly at some point, thanks!

rouault:
> no (I had to go and check for exactly why), I want the output extent to snap based on an _approximate_ input '-te' and an input '-ts', I don't want to input '-tr' for this case

ok, I believe I got it. I believe however that probably the best solution would be to extend -tap to be usable with -te + -ts (to avoid option proliferation)

mdsumner:
ah, excellent - that's exactly the kind of advice I needed (before trying to understand the entire box of options!!). I'll do it, thank you

PR #7075, opened 2023-01-16
DIMAP: optimize performance of dataset RasterIO() (improve perf on multispectral datasets with JP2KAK underlying driver)

rouault:


tbonfort:
Thanks Even. Looking at the patch I don't understand why this would only affect JP2KAK and not OpenJPEG, could you enlighten me?

tbonfort:
Nice job @rouault :+1: 

before:
```
time gdal_translate -co BIGTIFF=YES -co COMPRESS=LZW -co BLOCKXSIZE=512 -co BLOCKYSIZE=512 -co TILED=YES -co PREDICTOR=2 -co NUM_THREADS=4 IMG_PHR1A_PMS_001/DIM_PHR1A_PMS_202203171128164_SEN_6530734101.XML kak36.tif
Input file size is 40000, 17580

real	3m36,324s
user	26m31,629s
sys	0m17,706s
```
after:
```
time gdal_translate -co BIGTIFF=YES -co COMPRESS=LZW -co BLOCKXSIZE=512 -co BLOCKYSIZE=512 -co TILED=YES -co PREDICTOR=2 -co NUM_THREADS=4 IMG_PHR1A_PMS_001/DIM_PHR1A_PMS_202203171128164_SEN_6530734101.XML kak37.tif
Input file size is 40000, 17580

real	1m42,529s
user	11m54,314s
sys	0m9,269s
```

rouault:
> Looking at the patch I don't understand why this would only affect JP2KAK and not OpenJPEG, could you enlighten me?

OpenJPEG band-level RasterIO() method had logic to fill the cache for other bands, because as there is no way in libopenjp2 to decode a band separately for another one, we'd better pre-fill the cache of other bands to avoid decompressing all bands multiple times (actually there is now a way in libopenjp2 API to ask to decode only a subset of bands, provided no multi component transform (MCT) is used, but that's not used by the JP2OpenJPEG driver)
Whereas JP2KAK band-level RasterIO() method only decodes the band on which it is called. Given that the MS products use the RPCL progression order (Resolution, Position, Component, Layers), the coefficients for the components are mixed together at a rather low level, and it likely causes more work to Kakadu to decode the same window of interest multiple times for each component than asking to decompress it once for all components. So having a dataset-level RasterIO() implementation in the DIMAP driver which will call the dataset-level VRT one and ultimately the JP2KAK dataset-level one helps Kakadu to be more efficient.

PR #7077, opened 2023-01-16
Doc: document nodata handling in gdaladdo, gdal_translate and gdalwarp

rouault:


doc/source/programs/gdal_translate.rst
```diff
@@ -329,6 +329,16 @@ resampling, and rescaling pixels in the process.
 
     The destination file name.
 
+
+Nodata / source validity mask handling during resampling
+--------------------------------------------------------
+
+Masked values, either identified trough a nodata value metadata set on the
```
lnicola, line 336: Nit:

```suggestion
Masked values, either identified through a nodata value metadata set on the
```

doc/source/programs/gdalwarp.rst
```diff
@@ -456,6 +457,32 @@ Starting with GDAL 3.1, it is possible to use as output format a driver that
...
+  is projected back to source coordinates and the source pixel containing that
+  coordinate is identified. If this source pixel is invalid, the target pixel
+  is considered as nodata.
+
+- for bilinear, cubic, cubicspline and lanczos, for each target pixel, the
+  coordinate of its center is projected back to source coordinates and the
+  source pixel is identified. Given that those resampling kernels have a non-null
+  kernel radius, it might be possible that there are invalid values in other
+  contributing source pixels. The weights used to take into account those invalid
+  values will be set to zero to ignore them.
```
lnicola, line 480: Does this mean there's no rescaling of the weights? Resampling (with gdal_translate) the following grid to 2x:

```
ncols        5
nrows        1
xllcorner    0
yllcorner    0
cellsize     2
NODATA_value -10000
1  2  -10000 3  4
```

yields

```
ncols        10
nrows        2
xllcorner    0.000000000000
yllcorner    0.000000000000
cellsize     1.000000000000
NODATA_value -10000
 1 1 2 2 2 3 3 3 4 4
 1 1 2 2 2 3 3 3 4 4
```

while `gdalwarp` produces:

```
ncols        10
nrows        2
xllcorner    0.000000000000
yllcorner    0.000000000000
cellsize     1.000000000000
NODATA_value -10000
 1 1 2 2 -10000 -10000 3 3 4 4
 1 1 2 2 -10000 -10000 3 3 4 4
```

So:
 - it looks like the weights might be rescaled in the `gdal_translate` case
 - there is a difference between `gdalwarp` and `gdal_translate` that doesn't appear to be covered in here (but that might be fine)

rouault: > it looks (I haven't computed it manually though) like the weights might be rescaled in the gdal_translate case

not sure what you mean by rescaled. They are set to null for nodata pixels both at the numerator and denominator, "obviously", to get something consistent.

I believe the difference in behaviour between gdalwarp and gdal_translate is explained by the doc (note that they are difference between gdalwarp and gdal_translate/gdaladdo). For gdalwarp, a main source pixel is selected to decide if the target pixel is going to be nodata. So here the source pixel at -10000 is going to give to be replicated in the output.  For gdal_translate, it is going to be ignored and only the neighbour pixels at 2 and 3 are taken into account.
The differences in behavior is obviously not super desirable, but I have no idea if there's a correct universal way of proceeding in such situation.

lnicola: > not sure what you mean by rescaled. They are set to null for nodata pixels both at the numerator and denominator, "obviously", to get something consistent.

Yeah, that's what I meant. Rescaled so the output isn't lower because one of the input pixels was ignored.

> I believe the difference in behaviour between gdalwarp and gdal_translate is explained by the doc (note that they are difference between gdalwarp and gdal_translate/gdaladdo).

Yeah, I probably misread it it in some way. In my reading:

 - for `gdalwarp`: the docs say that for one output pixel, there is _one_ input pixel ("the coordinate of its center is projected back to source coordinates and the source pixel is identified"), but then they talk about multiple "contributing source pixels". So the weights of the ignored pixels are set to 0, and the rest of the pixels are used.
 - for `gdal_translate`, there's multiple input pixels and in one specific case (`bicubic` with too many invalid pixels), the output pixel might be set to no data.

So unless I'm missing something, `gdal_translate` should output no data in more cases than `gdalwarp` for `bicubic`.

rouault: > For gdalwarp, a main source pixel is selected to decide if the target pixel is going to be nodata.

another way of saying ""the coordinate of its center is projected back to source coordinates and the source pixel is identified"

lnicola: I'm still somewhat unconvinced (is the "main" pixel the one `nearest` would pick?), but I'm going to stop being annoying on this topic :-).

lnicola (approved):

doc/source/programs/gdalwarp.rst
rouault: > is the "main" pixel the one `nearest` would pick?

yes, but I realize that the sentence "If this source pixel is invalid, the target pixel is considered as nodata" is missing for that case. Adding it

lnicola:
> is it possible to get the UNIFIED_SRC_NODATA=NO behavior in gdal_translate

Coming back to this, it seems that `gdal_translate` already seems to act as with `UNIFIED_SRC_NODATA=NO`. No data values in one band don't affect the values in the other bands. I can provide an example if necessary. Unfortunately, I don't know why I said above it doesn't work.

PR #7082, opened 2023-01-17
CMake: Add build option for using static Arrow/Parquet build

Maxxen:
## What does this PR do?

This PR adds CMake flags to specify that GDAL should use statically built Arrow/Parquet libraries.  

None from what I can find.

## Tasklist

 - ~~[ ] Add test case(s)~~
 - [x] Add documentation
 - ~~[ ] Updated Python API documentation (swig/include/python/docs/)~~
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS: Apple
* Compiler: AppleClang


rouault:
It could also be appropriate to add the following in cmake/helpers/CheckDependentLibraries.cmake  in the if (Arrow_FOUND) block at line 756
```
option(ARROW_STATIC "Use statically built Arrow targets" OFF)
mark_as_advanced(ARROW_STATIC)
```
and similarly for Parquet

Maxxen:
Alright! Thanks for the feedback, do you think I should add a `_STATIC` flag and documentation for the `GDAL_USE_ARROWDATASET` case too? I noticed the flag itself was missing from the docs.

rouault:
> or is that implied by `GDAL_USE_PARQUET`?

It might be good to document GDAL_USE_ARROWDATASET

Regarding a distinct _STATIC variable for ArrowDataset, I suspect it doesn't make sense to have a mix of static/dynamic for ArrowDataset and Parquet. And probably not for Arrow vs Parquet itself.  Maybe a global ARROW_LIBRARIES_STATIC variable (or some other name that captures the fact that arrow in that conext must be understood as all arrow related libraries) for all 3 libraries ?

Maxxen:
Thanks!

PR #7091, opened 2023-01-18
HDF4: avoid CMake error with Windows paths

jmckenna:
- handle cases of Windows paths such as `C:/Program Files/blah`
- prevent CMake error such as:
```
CMake Error at cmake/modules/packages/FindHDF4.cmake:127 (set_target_properties):
  set_target_properties called with incorrect number of arguments.
```
This took me a long time to debug, in the end it is a minor change, that someday will save someone else days of effort.


jmckenna:
ah, thanks for https://github.com/OSGeo/gdal/pull/7093

Issue #6819, opened 2022-11-29
Performance of ENVI BIP interleave


chacha21:
GDAL 3.5.3
Windows 7 64b
Latest Visual studio

I was surprised to discover that submitting pixel-interleaved data to a BIP envi dataset was ultimately not mapped to a `memcpy()` in `GDALRasterBand::IRasterIO()` but had to resort to some `GDALCopyWords64()` to handle different strides of input buffer and dst dataset.
I thought that using the same interleaving for the input buffer and the dataset guaranteed the best performance by avoiding storage reinterleaving.
Did I make a mistake somewhere ?

```

  GDALAllRegister();

  {
    GDALDriverManager* gdalDriverManager = GetGDALDriverManager();
    GDALDriver* enviDriver = !gdalDriverManager ? 0 : gdalDriverManager->GetDriverByName("ENVI");

    const int spatialWidth = 800;
    const int spatialHeight = 1000;
    const int bandsCount = 734;
    std::vector<float> dataLine(bandsCount*spatialWidth);//we assume PIXEL interleave

    const char* envi_options[] = {
      "INTERLEAVE=BIP",
      nullptr
    };
    const char** options = &envi_options[0];
    GDALDriver* gdalDriver = enviDriver;

    CPLErr error = CE_None;
    std::string dstFilepath = "temp.bip";
    GDALDataset* ds = gdalDriver->Create(dstFilepath.c_str(), spatialWidth, spatialHeight, bandsCount, GDT_Float32, options);
    for(int line = 0 ; line<spatialHeight ; ++line)
    {
      error = GDALDatasetRasterIO(ds, GDALRWFlag::GF_Write,
        0, line,
        spatialWidth, 1,
        dataLine.data(), spatialWidth, 1,
        GDT_Float32,
        bandsCount, nullptr, 
        static_cast<int>(bandsCount*sizeof(float)),//pixel
        static_cast<int>(bandsCount*spatialWidth*sizeof(float)),//line
        static_cast<int>(sizeof(float))//band
      );
    }
    ds->Release();
  }
```

chacha21:
Could it be that the cache block structure is inherently band-based and that BIP has no efficient backing representation ?

chacha21:
I have just tested and I don't think it is fixed for an ENVI dataset.
When submitting a spatial line as BIP, (a 2D buffer where x is bands, y is pixels), an inner loop on bands (in BlockBasedRasterIO()) will still deinterleave and not resort to a memcpy

[...]

PR #7112, opened 2023-01-23
RawDataset::RasterIO(): fix optimization of 4fd6dd24b53d7f23800e3b05773c5b559255ded6 in GA_Update scenarios (refs #6819)

### LINKED ISSUE #7111: ENVI BIP interleave failure
A bug is present in [https://github.com/OSGeo/gdal/issues/6819](https://github.com/OSGeo/gdal/issues/6819)
Reading a BIP ENVI dataset after updating it might erroneously bypass the cache

rouault:
Fixes #7111

### DIFF
--- a/autotest/gdrivers/envi.py
+++ b/autotest/gdrivers/envi.py
@@ -959,3 +959,46 @@ def test_envi_read_direct_access(byte_order):
     ds = None
 
     gdal.GetDriverByName("ENVI").Delete(filename)
+
+
+###############################################################################
+# Test direct access to BIP scanlines in GA_Update mode
+
+
+@pytest.mark.parametrize("GDAL_BAND_BLOCK_CACHE", ["HASHSET", "ARRAY"])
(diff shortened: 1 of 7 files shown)


chacha21:
Does the fix "just" flush the cache before reading from file when "bCanDirectAccessToBIPDataset" is true ?
(I have not digged into `FlushDirtyBlocks();`)
Flushing then reading from the file might be slower than just querying the cache and deinterleaving BIP data ?

rouault:
> Does the fix "just" flush the cache before reading from file when "bCanDirectAccessToBIPDataset" is true ?

yes, only dirty blocks, not all blocks

> Flushing then reading from the file might be slower than just querying the cache and deinterleaving BIP data ?

yes, it might be slower. That said I don't believe writing & reading back is a super common scenario.

rouault:
> I think that direct read access should be enabled only if there are no dirty blocks, rather than force-flushing at the risk of a performance hit

yes, make sense. Closing that pull request as superseded per https://github.com/OSGeo/gdal/pull/7114 which implements this simpler strategy

Issue #7099, opened 2023-01-19
TileDB: test failure on tiledb_read.py test with tiledb 2.14


rouault:
With the recent release of tiledb 2.14, the tiledb_read.py test crashes with the exception:
```
terminate called after throwing an instance of 'tiledb::TileDBError'
  what():  Query: Cannot set subarray; Setting a subarray on an already initialized  query is not supported.
Abandon (core dumped)
```
I've applied a urgent fix in 4ea7380dfdcf55cf4a31aa3e38c6a1f3b905080a to catch the exception and return an error. But there's still to be determined why this error triggers and fix it.
I've also temporarily disabled tiledb_read.py per 4cd4f7d6bc. This commit should be reverted once the underlying issue is fixed.
@normanb can you have a look at that please?

rouault:
I also had to disable test_tiledb_write_subdatasets() in 27eea70744

rouault:
I've also cherry-picked the disabling of the broken tests in the 3.6 branch (in 059f098ce7e5eb893baad5907e58636909cc3b8d and f416d3717b7649a982e8d99bd8475d169f63d8af) as this breaks the CI in that branch too. That way the master fix + revert once it is ready should apply to the 3.6 branch as well

PR #7128, opened 2023-01-26
tiledb rasterio

normanb:
## What does this PR do?

This PR changes the way that the TileDB driver requests data using rasterio. The reason for this change is to optimize for use within MapServer so that complete regions are requested at once and the TileDB driver internally does the parallel reads.

This PR in turn fixes https://github.com/OSGeo/gdal/issues/7099 as the TileDB `Query` is now initialized for every call.

Marked as WIP until the CI passes.

https://github.com/OSGeo/gdal/issues/7099

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/frmts/tiledb/tiledbdataset.cpp
+++ b/frmts/tiledb/tiledbdataset.cpp
@@ -2443,24 +2443,26 @@ GDALDataset *TileDBDataset::CreateCopy(const char *pszFilename,
             }
         }
 
+        CSLDestroy(papszCopyOptions);
+
         // TODO Supporting mask bands is a possible future task
         if (poDstDS != nullptr)
         {
             int nCloneFlags = GCIF_PAM_DEFAULT & ~GCIF_MASK;
             poDstDS->CloneInfo(poSrcDS, nCloneFlags);
+
+            if (poDstDS->eIndexMode == ATTRIBUTES)
+            {
+                poDstDS->FlushCache(false);
+            }
+
+            poDstDS->m_array->close();
+            poDstDS->eAccess = GA_ReadOnly;
+            poDstDS->m_array->open(TILEDB_READ);
+
+            return poDstDS.release();
         }
-
-        CSLDestroy(papszCopyOptions);
-        if (poDstDS->eIndexMode == ATTRIBUTES)
-        {
-            poDstDS->FlushCache(false);
-        }
-
-        poDstDS->m_array->close();
-        poDstDS->eAccess = GA_ReadOnly;
-        poDstDS->m_array->open(TILEDB_READ);
-
-        return poDstDS.release();
+        return nullptr;
     }
     catch (const tiledb::TileDBError &e)
     {
--- a/frmts/tiledb/tiledbdataset.cpp
+++ b/frmts/tiledb/tiledbdataset.cpp
@@ -421,7 +421,7 @@ CPLErr TileDBRasterBand::IRasterIO(GDALRWFlag eRWFlag, int nXOff, int nYOff,
                 GDALRasterBand *poAttrBand = poAttrDS->GetRasterBand(nBandIdx);
                 GDALDataType eAttrType = poAttrBand->GetRasterDataType();
                 int nBytes = GDALGetDataTypeSizeBytes(eAttrType);
-                int nValues = nBufXSize * nBufYSize;
+                size_t nValues = static_cast<size_t>(nBufXSize) * nBufYSize;
                 void *pAttrBlock = VSIMalloc(nBytes * nValues);
                 aBlocks.emplace_back(pAttrBlock, &VSIFree);
 
@@ -594,7 +594,7 @@ CPLErr TileDBDataset::IRasterIO(GDALRWFlag eRWFlag, int nXOff, int nYOff,
             (uint64_t)nYOff, (uint64_t)nYOff + nYSize - 1, (uint64_t)nXOff,
             (uint64_t)nXOff + nXSize - 1};
 
-        if ((eRWFlag == GF_Read) && ((eAccess == GA_Update) && (m_roArray)))
+        if ((eRWFlag == GF_Read) && (eAccess == GA_Update && m_roArray))
         {
             poQuery.reset(new tiledb::Query(*m_roCtx, *m_roArray));
         }
--- a/frmts/tiledb/tiledbdataset.cpp
+++ b/frmts/tiledb/tiledbdataset.cpp
@@ -36,7 +36,7 @@
 #include "tiledb_headers.h"
 
 #ifdef _MSC_VER
-#pragma warning( push )
+#pragma warning(push)
 // 'tiledb::Array::Array': was declared deprecated
 #pragma warning(disable : 4996) /* XXXX was deprecated */
 #endif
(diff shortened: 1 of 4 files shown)


rouault:
> Marked as WIP until the CI passes.

@normanb Your branch must be against an outdated master branch as github indicates conflicts. Make sure to make it against lastest master and also git revert commits 4cd4f7d6bc390bcbeb46ca9609d5bd0979d294c0 and 27eea7074495ce6b8be0a1720d9f1da4987eb5c3

Also remember to enable pre-commit for the new code formatting rules: https://gdal.org/development/dev_practices.html#commit-hooks

rouault:
- cppcheck issue: https://github.com/OSGeo/gdal/actions/runs/4020227119/jobs/6907910456
- fails to build against tiledb 2.1.6 on MacOSX: https://github.com/OSGeo/gdal/actions/runs/4020227114/jobs/6907909948
- crashes on autotest on Windows: https://github.com/OSGeo/gdal/actions/runs/4020227122/jobs/6919706787

normanb:
@rouault I have reviewed the code CI failures, it looks like on Mac the easiest thing would be to either bump the minimum required version of TIleDB for `set_buffer` or `set_data_buffer` or use `ifdef` I will use the latter.

I have spend most of the day trying to set up a windows build, what is the easiest way to add build tools to `osgeo4w` or do you have another approach?

rouault:
> t looks like on Mac the easiest thing would be to either bump the minimum required version of TIleDB for set_buffer or set_data_buffer or use ifdef I will use the latter.

Up to you if you want to bump the minimum TileDB version (if so, it should likely be done in cmake/helpers/CheckDependentLibraries.cmake) or can handle older versions with #ifdef

> I have spend most of the day trying to set up a windows build, what is the easiest way to add build tools to `osgeo4w` or do you have another approach?

I use Conda to grab all dependencies. Much easier. cf https://github.com/OSGeo/gdal/blob/master/.github/workflows/cmake_builds.yml#L424 (you just need proj and tiledb at line 428)

normanb:
The failure on Windows seems to be related to specifying `INTERLEAVE` as `ATTRIBUTES` which is unrecognized in https://github.com/OSGeo/gdal/blob/a632e3093d6dd34cdcf979a5b2a20529fe2df943/gcore/rasterio.cpp#L4625 and prevents the `delete` call. Hence the access violation as the array already exists.

Easiest thing is to disable testing attributes but as we want to test that we could add `ATTRIBUTES` as acceptable in the rasterio test or disable `ATTRIBUTES` on Windows testing.

Or I can can add a pre-test hook that checks if the array exists then it is deleted.

rouault:
> related to specifying INTERLEAVE as ATTRIBUTES which is unrecognized in rasterio.cpp and prevents the `delete` call.

I'm not following the connection between that code and the absence of a call to Delete()

>  Hence the access violation as the array already exists.

Looks like that should be caught (C++ exception?) to error out cleanly. Crashing is rarely a valid option :-)

> we could add ATTRIBUTES as acceptable in the rasterio test 

I guess this would be probably preferable (although I can't recall what INTERLEAVE=ATTRIBUTES mean). With a comment explaining this is TileDB specific.

> Or I can can add a pre-test hook that checks if the array exists then it is deleted.

It is not exceptional for tests to do things like the following at their beginning
```
if os.path.exists("tmp/bla"):
   shutil.rmtree("tmp/bla")
```

normanb:
I seem to have broken my local build environment, I will fix this in the morning (ET) and push a fix.

frmts/tiledb/tiledbdataset.cpp
```diff
@@ -646,6 +568,79 @@ TileDBDataset::~TileDBDataset()
...
+        eBufType == eDataType && nXSize == nBufXSize && nYSize == nBufYSize &&
+        nBufferDTSize > 0 && (nPixelSpace % nBufferDTSize) == 0 &&
+        (nLineSpace % nBufferDTSize) == 0)
+    {
+        std::unique_ptr<tiledb::Query> poQuery;
+        std::vector<uint64_t> oaSubarray = {
+            (uint64_t)nYOff, (uint64_t)nYOff + nYSize - 1, (uint64_t)nXOff,
+            (uint64_t)nXOff + nXSize - 1};
+
+        if ((eRWFlag == GF_Read) && ((eAccess == GA_Update) && (m_roArray)))
```
rouault, line 597: the (a) && ((b) && (c)) pattern looks suspicous to me. If that's intended the extra parenthesis around ((b) && (c)) should be removed. But I'm wondering if the first && shouldn't rather be a || ?

frmts/tiledb/tiledbdataset.cpp
```diff
@@ -579,9 +468,51 @@ CPLErr TileDBRasterBand::IWriteBlock(int nBlockXOff, int nBlockYOff,
...
+                                        nPixelSpace, nLineSpace, psExtraArg);
+}
+
+CPLErr TileDBRasterBand::IReadBlock(int nBlockXOff, int nBlockYOff,
+                                    void *pImage)
+{
+    const int nXOff = nBlockXOff * nBlockXSize;
+    const int nYOff = nBlockYOff * nBlockYSize;
+    const int nDTSize = GDALGetDataTypeSizeBytes(eDataType);
+    return IRasterIO(GF_Read, nXOff, nYOff, nBlockXSize, nBlockYSize, pImage,
```
rouault, line 486: If nRasterXSize or nRasterYSize may not be multiple of the block size, then this call is going to be wrong for the right-most or bottom-most blocks, unless this is dealt in a special way  in IRasterIO().
Same for IWriteBlock()

frmts/tiledb/tiledbdataset.cpp
```diff
@@ -298,278 +351,114 @@ TileDBRasterBand::TileDBRasterBand(TileDBDataset *poDSIn, int nBandIn,
...
-            int nXSize = nBlockXSize;
-            int nYSize = nBlockYSize;
-            if (nBlockXOff + nXSize > nRasterXSize)
-                nXSize = nRasterXSize - nBlockXOff;
-            if (nBlockYOff + nYSize > nRasterYSize)
-                nYSize = nRasterYSize - nBlockYOff;
+                GDALRasterBand *poAttrBand = poAttrDS->GetRasterBand(nBandIdx);
+                GDALDataType eAttrType = poAttrBand->GetRasterDataType();
+                int nBytes = GDALGetDataTypeSizeBytes(eAttrType);
+                int nValues = nBufXSize * nBufYSize;
```
rouault, line 424: To avoid potential overflows:
```suggestion
                size_t nValues = static_cast<size_t>(nBufXSize) * nBufYSize;
```

frmts/tiledb/tiledbdataset.cpp
normanb: I have removed the extra parenthesis. The logic is correct, the array can only be open in `read` or `write` mode, this is the case where we are doing an update and we read the data first and then apply the update.

frmts/tiledb/tiledbdataset.cpp
normanb: change made

frmts/tiledb/tiledbdataset.cpp
normanb: The array is padded to be have dimensions that are a multiple of the block size https://github.com/OSGeo/gdal/blob/e76b29a708f9cb5242e74e468ffc67a685466186/frmts/tiledb/tiledbdataset.cpp#L1922

frmts/tiledb/tiledbdataset.cpp
normanb: change made

rouault:
@normanb still a WIP or good for merge ?

normanb:
@rouault we are good to merge

PR #7149, opened 2023-01-31
Add RFC 92 text: WKB Only geometries

rouault:
==> [Rendered view](https://github.com/rouault/gdal/blob/rfc92_text/doc/source/development/rfc/rfc92_wkb_only_geometries.rst)

### DIFF
--- a/doc/source/development/rfc/rfc92_wkb_only_geometries.rst
+++ b/doc/source/development/rfc/rfc92_wkb_only_geometries.rst
@@ -1,15 +1,15 @@
 .. _rfc-92:
 
 =============================================================
-RFC 92: WKB Only geometries
+RFC 92: WKB Only geometries (on hold)
 =============================================================
 
 ============== =============================================
 Author:        Even Rouault
 Contact:       even.rouault at spatialys.com
 Started:       2023-Jan-31
-Status:        Draft / discussion
-Target:        GDAL 3.7 ?
+Status:        On hold
+Target:        N/A
 ============== =============================================
 
 Summary
@@ -18,6 +18,8 @@ Summary
 This RFC provides shortcuts to avoid instantiation of full OGRGeometry instances
 in scenarios where only the WKB representation of geometries is needed. The
 hope is to save CPU time.
+
+.. note:: This RFC is on hold.
 
 Motivation
 ----------
@@ -35,7 +37,7 @@ GDAL side:
 
 1. acquire GeoPackage geometry blob (WKB with a GeoPackage specific header) from
    the database
-2. instanciate the relevant OGRGeometry subclass instance (OGRPoint, OGRLineString,
+2. instantiate the relevant OGRGeometry subclass instance (OGRPoint, OGRLineString,
    OGRPolygon, etc.) from WKB
 3. store it in a OGRFeature
 
@@ -44,7 +46,7 @@ QGIS side:
 4. get the OGRGeometry from the OGRFeature
 5. asks for its WKB representation, build from the subclass members (not the
    original WKB of step 2.)
-6. instanciate QgsGeometry from WKB
+6. instantiate QgsGeometry from WKB
 
 One can see that if we were able to store the original WKB representation and
 get it back we could save OGRGeometry subclass object creation, destruction and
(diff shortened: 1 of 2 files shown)


doc/source/development/rfc/rfc92_wkb_only_geometries.rst
```diff
@@ -0,0 +1,228 @@
...
+
+The format of geometries of some popular formats like GeoPackage or PostGIS is
+based on WKB. Consequently the chain of processings done by QGIS reading GeoPackage
+is:
+
+GDAL side:
+
+1. acquire GeoPackage geometry blob (WKB with a GeoPackage specific header) from
+   the database
+2. intanciate the relevant OGRGeometry subclass instance (OGRPoint, OGRLineString,
```
nyalldawson, line 38:
```suggestion
2. instanciate the relevant OGRGeometry subclass instance (OGRPoint, OGRLineString,
```

doc/source/development/rfc/rfc92_wkb_only_geometries.rst
```diff
@@ -0,0 +1,228 @@
...
+   OGRPolygon, etc.) from WKB
+3. store it in a OGRFeature
+
+QGIS side:
+
+4. get the OGRGeometry from the OGRFeature
+5. asks for its WKB representation, build from the subclass members (not the
+   original WKB of step 2.)
+6. intanciate QgsGeometry from WKB
```
nyalldawson, line 47:
```suggestion
6. instanciate QgsGeometry from WKB
```

doc/source/development/rfc/rfc92_wkb_only_geometries.rst
```diff
@@ -0,0 +1,228 @@
...
+
+On can see that if we were able to store the original WKB representation and
```
nyalldawson, line 49:
```suggestion
One can see that if we were able to store the original WKB representation and
```

doc/source/development/rfc/rfc92_wkb_only_geometries.rst
```diff
@@ -0,0 +1,228 @@
...
+  instances.
+
+ogr2ogr is modified to call RequestWKBOnlyGeometries(true) on the source layer:
+
+* if the source layer advertises OLCReadWKBGeometries
+* if the target layer advertises OLCWriteWKBGeometries
+* if no command line switch requires a "materialized" geometry.
+
+So basically this is for requests like
+"ogr2ogr out.gpkg in.gpkg [layer or SQL request] [attribute filter] [spatial filter]"
```
nyalldawson, line 135: Wouldn't spatial filters often require full geometry parsing (assuming the filter can't be passed to a backend/read from an index only)

doc/source/development/rfc/rfc92_wkb_only_geometries.rst
```diff
@@ -0,0 +1,228 @@
...
+
+With spatial index creation enabled (multi-threaded), the wall clock time
+difference is within measurement noise. And for singe threaded creation, the
+WKBOnlyGeometry mode is 5% faster.
+
+Discussion
+----------
+
+Is it a good idea... ?
+
```
nyalldawson, line 196: I believe so. I've extensively profiled qgis' ogr provider feature reading in the past and the geometry parsing/conversion costs are a large percentage of the reading time which cannot be further optimised using current API.

For some datasets this is likely to result in a noticable rendering time improvement in desktop QGIS, and a definitely worthwhile improvement for QGIS server.

doc/source/development/rfc/rfc92_wkb_only_geometries.rst
```diff
@@ -0,0 +1,228 @@
...
+and that doesn't change the whole OGRGeometry API and driver implementations
+comes to mind.
+
+The scope is limited to a few drivers: GeoPackage, PostGIS (but the current
+throughput of the driver is probably not sufficient for OGRGeometry overhead to
+be noticeable), what else?
+
+Should methods of OGRWKBOnlyGeometry that cannot work without materialization
+of the real geometry return an error like done currently, or do the materialization
+on-the-fly when needed ? The motivation for erroring out is to avoid silent
```
nyalldawson, line 208: At worst these are going to be ~same cost as current code, so I don't see the cost of materializing when needed as a blocker. I'd lean toward materialization on demand make it simpler/safer for users to always just ask for WKB only and trust GDAL to do the right thing transparently.

doc/source/development/rfc/rfc92_wkb_only_geometries.rst
```diff
@@ -0,0 +1,228 @@
...
+performance issues related to materialization.
+
+Issues / pull requests
+----------------------
+
+https://github.com/OSGeo/gdal/compare/master...rouault:gdal:rfc92_implementation?expand=1
+contains a preliminary candidate implementation.
+
+Not all subtelties have been taken into account in the prototype implementation
```
nyalldawson, line 217:
```suggestion
Not all subtleties have been taken into account in the prototype implementation
```

nyalldawson:
Sounds very exciting to me! As noted above the forced conversion to/from OGR geometries has been a significant chunk of processing I've bumped into many times in the past while profiling QGIS, and I can't see a better approach to avoid this then what you've described here :+1:

doc/source/development/rfc/rfc92_wkb_only_geometries.rst
rouault: note added

doc/source/development/rfc/rfc92_wkb_only_geometries.rst
rouault: > I'd lean toward materialization on demand 

That isn't as simple as that. For example as commented in the prototype of the class, getGeometryType() cannot safely be implemented as Materialize()->getGeometryType(), because otherwise users would wrongly cast a OGRWKBOnlyGeometry* to a OGRPoint*, OGRLineString*, etc

nyalldawson: I guess it's not a trivial change, but we could introduce a `ogrgeometry_cast` implementation which has special handling for OGRWKBOnlyGeometry and correctly returns the cast materialized geometry.

rouault: The base OGRGeometry class has *non virtual* toPoint(), toPolygon(), etc. cast operators. So if there were made virtual (but this would have a small run-time cost), potentially OGRWKBOnlyGeometry could reimplement them (that would also involve that OGRWKBOnlyGeometry stores the materialized geometry, when creating it). But that assumes that all code that cast OGRGeometry* to a sub-class uses those operators, which is very likely not true as they have been introduced relatively recently (I remember having made a pass to use those operators, but I'm pretty sure it isn't exhaustive, as there's no way to spot all code paths which do those casts. And even if we could do it in the GDAL code base, that would also impact external code using GDAL C++ API).

rouault:
One interesting thing to benchmark would be to check how it is effective from QGIS. The benchmarking within OGR have no extra overhead. Within QGIS, you can expect the gains to be lower (at least in %, the absolute time gains should be the same).

PR #7156, opened 2023-02-01
CI: alpine: do pip in a venv to please Alpine's pip

rouault:


rouault:
manually backported in 1adffeb0d8

PR #7175, opened 2023-02-04
nitf: add support for COMNTA

bradh:
## What does this PR do?

Add support for a draft NITF TRE that provides UTF-8 text comments.

The curious part of this is that the TRE only has one item, and it just "the TRE length, as UTF-8 encoded characters". That is, the length is variable, but there is no field that signals the length. So we need to handle that somehow. The approach used in NITRO is to say "remaining bytes" (`NITF_TRE_GOBBLE`) in a few definitions (JITCID - not actually a real TRE, and a couple of RPF TREs). We could just say "all bytes", but the "remaining bytes" concept seems more general.

The way I've currently approached the encoding is to use the `length_var` keyword, but with a value of `*` to indicate the wildcard that means "remaining bytes". That seemed like a simple solution. I also considered using a label like `TREL` or a silly magic value that probably wouldn't appear as a TRE field name (e.g. "NOM, NOM, NOM"). We could also modify the parsing to add a new keyword i.e. instead of the logic choice between `length` or `length_var`, there would be `length`, `length_var` and `length_remaining` (which would have no argument).

The TRE isn't approved yet, so I'm putting this up early as a draft to solicit feedback.

N/A.

## Tasklist

 - [ ] Resolve remaining bytes syntax.
 - [x] Add test case(s)
 - [N/A] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

N/A.

### DIFF
--- a/autotest/gdrivers/nitf.py
+++ b/autotest/gdrivers/nitf.py
@@ -4231,6 +4231,42 @@ def test_nitf_91():
 
 
 ###############################################################################
+# Test parsing COMNTA TRE
+
+
+def test_nitf_comnta_doc():
+    tre_data = (
+        "FILE_TRE=HEX/COMNTA="
+        + hex_string("Hyperion data courtesy of the U.S. Geological Survey.")
+        + "0d0a"
+        + hex_string("Relative Spectral Response for SWIR FPA bands only.")
+        + "0d0a"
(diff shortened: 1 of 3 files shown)


rouault (approved):
Simple & efficient approach. LGTM

PR #7182, opened 2023-02-04
Add GDALMDArray::GetGridded() and GDALMDArrayGetGridded()

rouault:
Return a gridded array from scattered point data, that is from an array
whose last dimension is the indexing variable of X and Y arrays.

e.g. the zeta variable in:
```
dimensions:
        node = 303714 ;
        time = UNLIMITED ; // (1 currently)
variables:
        float lon(node) ;
                lon:long_name = "nodal longitude" ;
                lon:standard_name = "longitude" ;
                lon:units = "degrees_east" ;
        float lat(node) ;
                lat:long_name = "nodal latitude" ;
                lat:standard_name = "latitude" ;
                lat:units = "degrees_north" ;
        float zeta(time, node) ;
                zeta:long_name = "Water Surface Elevation" ;
                zeta:units = "meters" ;
                zeta:positive = "up" ;
                zeta:standard_name = "sea_surface_height_above_geoid" ;
                zeta:grid = "Bathymetry_Mesh" ;
                zeta:coordinates = "time lat lon" ;
                zeta:type = "data" ;
                zeta:location = "node" ;
```

The gridding is done in 2D, using GDALGridCreate(), on-the-fly at Read()
time, taking into account the spatial extent of the request to limit the
gridding. The results got on the whole extent or a subset of it might not be
strictly identical depending on the gridding algorithm and its radius.
Setting a radius in osGridOptions is recommended to improve performance.
For arrays which have more dimensions than the dimension of the indexing
variable of the X and Y arrays, Read() must be called on slices of the extra
dimensions (ie count[i] must be set to 1, except for the X and Y dimensions
of the array returned by this method).

### DIFF
--- a/alg/delaunay.c
+++ b/alg/delaunay.c
@@ -627,7 +627,16 @@ int GDALTriangulationFindFacetDirected(const GDALTriangulation *psDT,
         }
     }
 
-    CPLDebug("GDAL", "Using brute force lookup");
+    static int nDebugMsgCount = 0;
+    if (nDebugMsgCount <= 20)
+    {
+        CPLDebug("GDAL", "Using brute force lookup%s",
+                 (nDebugMsgCount == 20)
+                     ? " (this debug message will no longer be emitted)"
+                     : "");
+        nDebugMsgCount++;
+    }
+
     return GDALTriangulationFindFacetBruteForce(psDT, dfX, dfY,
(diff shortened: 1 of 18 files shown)


mdsumner:
whoa, this has a lot of interesting implications 🙏

PR #7185, opened 2023-02-06
Geometry WKT import: accept nan as a value, for parity with PostGIS and GEOS

rouault:


autotest/ogr/ogr_wkbwkt_geom.py
```diff
@@ -750,6 +750,34 @@ def test_ogr_wkt_multipoint_postgis():
     assert g.ExportToIsoWkt() == "MULTIPOINT Z ((1 2 3),(4 5 6))"
 
 
+###############################################################################
+# Test accepting nan whici is an extension to the WKT spec, but both
```
elpaso, line 754:
```suggestion
# Test accepting nan which is an extension to the WKT spec, but both
```

elpaso (approved):

PR #7197, opened 2023-02-07
ogr2ogr: Improve performance of -clipsrc and -clipdst

dbaston:
## What does this PR do?

This PR attempts to improve the performance of `ogr2ogr` with a `-clipsrc` or `-clipdst` argument by testing whether geometries intersect the clipping area (relatively cheap) before computing their intersection (relatively expensive.)

I tested the performance of this using a dataset of ~300,000 parcel polygons for the state of Vermont. In the best reasonable case (clipping out part of one town, or 0.1% of the total polygons) it reduces runtime by ~50%, depending on the GEOS version. In the worst case, where every polygon is within the clipping window, it increases runtime by ~1%.

It could be made more efficient with a way to cache the `GEOSGeometry*` between successive calls to GEOS functions.

None.

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/apps/ogr2ogr_lib.cpp
+++ b/apps/ogr2ogr_lib.cpp
@@ -534,6 +534,10 @@ class LayerTranslator
                   GIntBig nCountLayerFeatures, GIntBig *pnReadFeatureCount,
                   GIntBig &nTotalEventsDone, GDALProgressFunc pfnProgress,
                   void *pProgressArg, GDALVectorTranslateOptions *psOptions);
+
+  private:
+    const OGRGeometry *GetDstClipGeom(OGRSpatialReference *poGeomSRS);
+    const OGRGeometry *GetSrcClipGeom(OGRSpatialReference *poGeomSRS);
 };
 
 static OGRLayer *GetLayerAndOverwriteIfNecessary(GDALDataset *poDstDS,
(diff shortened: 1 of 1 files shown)


apps/ogr2ogr_lib.cpp
```diff
@@ -5271,51 +5275,14 @@ int LayerTranslator::Translate(OGRFeature *poFeatureIn, TargetLayerInfo *psInfo,
...
-                            ? m_poClipSrcReprojectedToSrcSRS.get()
-                            : m_poClipSrcOri);
-                    delete poStolenGeometry;
-                    poStolenGeometry = nullptr;
-                    if (poClipped == nullptr || poClipped->IsEmpty())
+                    const OGRGeometry *poClipGeom =
+                        GetSrcClipGeom(poStolenGeometry->getSpatialReference());
+
+                    if (poClipGeom != nullptr &&
+                        !poClipGeom->Intersects(poStolenGeometry))
```
rouault, line 5282: the call to Intersection() is missing

dbaston: If I understood right, it's not needed here -- this path is for the case when we're writing to a dataset with no geometry, so the clip operation is just being used for filtering.

rouault: ah, you're right

jratike80:
Could you benchmark also PostGIS with the same data? In our experiments it was so much faster that in our process we now import data into PostGIS with ogr2ogr just to be able to let is do the intersection that we launch with SQL statement inside another ogr2ogr command.

dbaston:
With the same data in PostGIS, I see a benefit is 35% when 0.1% of polygons are retained and a penalty of 8% when 100% are retained.

I think most of the benefit here comes from the bounding-box check in `OGRGeometry::Intersects`. A similar check could be added to `OGRGeometry::Intersection`, which would remove the need to convert to GEOS twice. But that would require `OGRGeometry::Intersection` to determine things like the type of the resultant empty geometry, and I think those decisions would be best left to GEOS.

rouault:
> I think most of the benefit here comes from the bounding-box check in `OGRGeometry::Intersects`

you could use OGRGeometry::getEnvelope() and OGREnvelope::Intersects()

dbaston:
> you could use OGRGeometry::getEnvelope() and OGREnvelope::Intersects()

This avoids the performance penalty in the poorly selective case, so I think it's the best option. Updated, and fixed my Hungarian notation :)

jratike80:
By looking at my notes we finally decided that selecting whole geometries was better for our use case than clipping. So this comment is rather off-topic but here is comes anyway. For the selection we used temporary PostGIS table

`ogr2ogr -f GPKG -append -update -sql "select a.* from table1 a, table2 b where ST_Intersects(a.geom,b.geom) and b.natcode=%kuntatunnus%" output.gpkg PG:"host=localhost port=5432......`

Pure ogr2ogr alternative needs to be something like below because `-clipsrc` always clips and we wanted to get unclipped geometries.

`ogr2ogr -f GeoJSON -dialect sqlite -sql 'select * from table1 as a JOIN "natcode_235.geojson"."natcode_235" as b where st_intersects(a.geometry,b.geometry)' out.json table1.json`

The test geometry that was used for selecting has about 28000 vertices and table1 has maybe 20000 polygons. The ogr2ogr way was all too slow (using GeoJSON as source data probably made it worse), but the PostGIS way proved to be usable.

I think I made a suggestion to have the "select but do not clip" as an alternative in ogr2ogr but maybe that is not so common need.

rouault:
> Pure ogr2ogr alternative needs to be something like below because `-clipsrc` always clips and we wanted to get unclipped geometries.

Totally expected, especially using GeoJSON and given that the Spatialite st_intersects() doesn't use spatial indices unless you explicitly use the RTree in addition. This is definitely a use case for which PostGIS was designed for.

PR #7198, opened 2023-02-07
JXL-in-TIFF: implement support NBITS ==12 for UInt16 data type (fixes #6409)

### LINKED ISSUE #6409: TIFF+JXL: lossy compression of 16bit data is not supported correctly when the data does not span the full 16bit spectrum
From libjxl/libjxl#1763, lossy 16bit compression is not correctly supported for data with less than 16 bits of range. Investigate whether we can use GTiff's NBIT configuration option to pad the low bits of the pixels.

rouault:
This requires libjxl post 0.8.1 / https://github.com/libjxl/libjxl/pull/1812


tbonfort:
Thanks Even, I can confirm that this implementation produces similar results to my first-draft hard-coded patch from #6409.

However this double/useless bit packing/unpacking makes me sad. I was hoping that it was just me worrying about nothing, but running some timings today shows a significant slowdown:

On a 1024x1024 tile:

- using this PR's implementation (i.e. NBITS=12):
  - compression: ~0.27sec
  - decompression: ~0.24sec

-  using my dirty implementation (NBITS=16, only the jxl codestream is marked as 12bits):
  - compression: ~0.14sec
  - decompression: ~0.10 sec

I'm not sure what the way forward is here... I'm weary of introducing a sub-par implementation (performance wise) that might bring in backwards compatibility issues later on if/when we implement a more performant one, but I'm also weary of introducing a more performant one that has known forward compatibility issues. Maybe we should leave this on hold until we find an acceptable solution?

rouault:
How do you manage to get such small compression/decompression times... ?

```
$ gdal_translate autotest/gdrivers/data/small_world.tif 1024.tif -outsize 1024 1024 -r bilinear -scale 0 255 0 4095 -co nbits=12 -ot uint16
$ time gdal_translate 1024.tif 1024_jxl.tif -co compress=jxl -co jxl_distance=1 -co jxl_lossless=no
real	0m2,031s
user	0m1,993s
$ time gdal_translate 1024.tif 1024_jxl.tif -co compress=jxl
real	0m1,587s
user	0m1,562s
$ time gdal_translate 1024.tif 1024_jxl.tif -co compress=jxl -co tiled=yes
real	0m1,159s
user	0m1,118s
$ time gdal_translate 1024.tif 1024_jxl.tif -co compress=jxl -co tiled=yes -co jxl_distance=1 -co jxl_lossless=no
real	0m0,346s
user	0m0,321s

```
I've checked that both my GDAL and libjxl are built in Release mode

tbonfort:
> How do you manage to get such small compression/decompression times... ?

I use JXL_EFFORT=3 and a single tile:

time gdal_translate -co TILED=YES -co BLOCKXSIZE=1024 -co BLOCKYSIZE=1024 -co COMPRESS=JXL -co JXL_LOSSLESS=FALSE -co JXL_EFFORT=3 -co NBITS=12 tile-in.tif tile-jxl12.tif

rouault:
ok, JXL_EFFORT=3 was the trick:
```
$ time gdal_translate -co TILED=YES -co BLOCKXSIZE=1024 -co BLOCKYSIZE=1024 -co COMPRESS=JXL -co JXL_LOSSLESS=FALSE -co JXL_EFFORT=3 -co NBITS=12 1024.tif 1024_jxl.tif

real	0m0,197s
user	0m0,161s
```

rouault:
I've just committed a big performance boost by borrowing the optimized nbits==12 bit packing/unpacking from the GTiff reader, instead of relying on the generic code. I've also only enabled nbits==12 for uint16 nbits != 16

Timing before:
```
$ hyperfine --warmup 2 'gdal_translate -co TILED=YES -co BLOCKXSIZE=1024 -co BLOCKYSIZE=1024 -co COMPRESS=JXL -co JXL_LOSSLESS=FALSE -co JXL_EFFORT=3 -co NBITS=12 1024.tif 1024_jxl.tif -q'
Benchmark #1: gdal_translate -co TILED=YES -co BLOCKXSIZE=1024 -co BLOCKYSIZE=1024 -co COMPRESS=JXL -co JXL_LOSSLESS=FALSE -co JXL_EFFORT=3 -co NBITS=12 1024.tif 1024_jxl.tif -q
  Time (mean ± σ):     184.2 ms ±   1.5 ms    [User: 142.8 ms, System: 40.8 ms]
  Range (min … max):   182.4 ms … 189.1 ms
 
$ hyperfine --warmup 2 'gdal_translate 1024_jxl.tif tmp.tif -scale 0 4095 0 255 -ot byte'
Benchmark #1: gdal_translate 1024_jxl.tif tmp.tif -scale 0 4095 0 255 -ot byte
  Time (mean ± σ):     180.7 ms ±   0.9 ms    [User: 155.7 ms, System: 24.2 ms]
  Range (min … max):   178.6 ms … 181.9 ms
```

Timing after:
```
$ hyperfine --warmup 2 'gdal_translate -co TILED=YES -co BLOCKXSIZE=1024 -co BLOCKYSIZE=1024 -co COMPRESS=JXL -co JXL_LOSSLESS=FALSE -co JXL_EFFORT=3 -co NBITS=12 1024.tif 1024_jxl.tif -q'
Benchmark #1: gdal_translate -co TILED=YES -co BLOCKXSIZE=1024 -co BLOCKYSIZE=1024 -co COMPRESS=JXL -co JXL_LOSSLESS=FALSE -co JXL_EFFORT=3 -co NBITS=12 1024.tif 1024_jxl.tif -q
  Time (mean ± σ):     110.7 ms ±   1.0 ms    [User: 72.3 ms, System: 37.8 ms]
  Range (min … max):   108.7 ms … 112.9 ms
 
$ hyperfine --warmup 2 'gdal_translate 1024_jxl.tif tmp.tif -scale 0 4095 0 255 -ot byte'
Benchmark #1: gdal_translate 1024_jxl.tif tmp.tif -scale 0 4095 0 255 -ot byte
  Time (mean ± σ):      94.0 ms ±   0.7 ms    [User: 68.4 ms, System: 25.1 ms]
  Range (min … max):    93.0 ms …  96.4 ms
 ```

tbonfort:
Thanks Even,
I'm still wondering if it is desirable to add this functionality now, and have to bear with potential backwards incompatibilities later. Understandably, (at least from what I understand) jxl's lossy mode needs to know what uint16 value corresponds to "black"  and what value corresponds to "white", for each band, in order to apply its perceptually driven compression. Those two values cannot be expressed by a single number of bits, so even if we were able to support an arbitrary number NBITS instead of only 12, that would not be enough.

My feeling is that we should close this issue for now. For the future:
- we should not rely on NBITS for setting the input data range
- we need a libjxl api that allows setting the black and white points, maybe for each band independently (I haven't checked if this api already exists or not)
- the tiff spec does have some tags to express the dynamic range with MinSampleValue, MaxSampleValue, but they are marked as informative only. I don't see how a gdal user could express those without using creation options, which I understand are undesirable to extend.

Happy to discuss this more with anyone interested if needed.

rouault:
For the black and white points, it could potentially be conveyed to libtiff through the https://www.awaresystems.be/imaging/tiff/tifftags/referenceblackwhite.html tag. But indeed we' need a GTiff creation option for that. There's no particular problem in adding one to the many existing ones...
I also agree that the NBITS=12 case is probably not optimum as it. Probably we should use a dedicated TIFF tag specific to the JXL codec to mean "codec bitdepth, but not libtiff bitdepth", and that would probably need to be a real TIFF tag encoded in the file, not just an ephemeral pseudo-tag.
Closing

PR #7211, opened 2023-02-09
ComputeRasterMinMax(), ComputeStatistics(), GetHistogram(): take into account mask band

rouault:
Previously only the nodata value was taken into account to ignore masked pixels.

CC @nyalldawson


nyalldawson (approved):
LGTM!

PR #7227, opened 2023-02-11
Alpine CI: use packaged hdf4

rouault:


rouault:
ah, it is just in the "testing" repository for now

PR #7243, opened 2023-02-13
VRT derived band: add a <SkipNonContributingSources>true</SkipNonContributingSources> optional element to discard non contributing sources (fixes #7223)

### LINKED ISSUE #7223: Input to PixelFunctionType includes all ComplexSource, regardless of DstRect

I have a large (>2000) number of rasters produced by processing large sources tile-by-tile. The rasters have some (limited) overlap, that we would like to handle in a specific way.

To this end, we are generating a VRT, and providing a (Numba) Python `PixelFunctionType` to each `VRTDataset`.

Things work well when for smaller numbers of rasters, but when the number reaches 1000, Numba crashes, due to the fact that it cannot handle tuples with more than 1000 elements.

GDAL seems to be calling the `PixelFunctionType` function with every single layer, regardless of `DstRect`, resulting in a large number of empty arrays (filled with `nodata` value) and only a few (or sometimes even no) arrays containing data.

This happens, regardless of whether I use `gdalwarp` or `gdal_translate` to generate the mosaic'ed tif.

Even though I have applied a temporary workaround by filtering the `in_arr` argument in a non-Numba function to remove all-nodata arrays, before calling the Numba code, this seems hugely suboptimal.

VRT file looks like this:
 
```
<VRTDataset rasterXSize="58786" rasterYSize="9222">
 <SRS dataAxisToSRSAxisMapping="1,2”>[…]</SRS>
 <GeoTransform> -7.9902911092382642e+06,  4.6331271652776735e+02,  0.0000000000000000e+00,  2.2618926820882889e+06,  0.0000000000000000e+00, -4.6331271652773665e+02</GeoTransform>
 <VRTRasterBand subClass="VRTDerivedRasterBand" dataType="Float32" band="1" NoDataValue="0" blockXSize="128" blockYSize="128">
   <PixelFunctionType>test.test</PixelFunctionType><PixelFunctionArguments nodata="0" /><PixelFunctionLanguage>Python</PixelFunctionLanguage>
   <NoDataValue>0</NoDataValue>
   <HideNoDataValue>1</HideNoDataValue>
   <ColorInterp>Gray</ColorInterp>
   <ComplexSource>
     <SourceFilename relativeToVRT="1">././test_vrt/0a06211d1ceccd06aff9a5a50f5e70b5.tif</SourceFilename>
     <SourceBand>1</SourceBand>
     <SourceProperties RasterXSize="568" RasterYSize="544" DataType="Float32" BlockXSize="256" BlockYSize="256" />
     <SrcRect xOff="0" yOff="0" xSize="568" ySize="544" />
     <DstRect xOff="52791" yOff="4881" xSize="568" ySize="544" />
     <NODATA>0</NODATA>
   </ComplexSource>
   <ComplexSource>
     <SourceFilename relativeToVRT="1">././test_vrt/0a120c177147927bc0bd8c766907e65f.tif</SourceFilename>
     <SourceBand>1</SourceBand>
     <SourceProperties RasterXSize="556" RasterYSize="544" DataType="Float32" BlockXSize="256" BlockYSize="256" />
     <SrcRect xOff="0" yOff="0" xSize="556" ySize="544" />
     <DstRect xOff="24236" yOff="3796" xSize="556" ySize="544" />
     <NODATA>0</NODATA>
   </ComplexSource>
[…] 
[~2000 ComplexSource elements]
```

with a Python function:

```
@nb.jit(cache=True, parallel=True, nopython=True, nogil=True)
def test(in_arr, out_arr, xoff, yoff, xsize, ysize, raster_xsize, raster_ysize, buf_radius, gt, nodata):
```

Running gdalwarp:
`gdalwarp test.vrt test.tif -co TILED=TRUE -co NUM_THREADS=ALL_CPUS --config GDAL_VRT_ENABLE_PYTHON YES`
will send tuples of length 2000 (or whatever the number of `ComplexSource` elements is), resulting in an error if the function is using Numba.

## Operating system

MacOs 13.1 (22C65)

GDAL 3.4.1, released 2021/12/27

rouault:


zedrdave:
I was able to compile and run this PR, and it already runs much faster for our [use case](https://github.com/OSGeo/gdal/issues/7223)!

One thing I noticed, however, is that it might have introduced a regression (which would probably affect a lot of existing users):

When removing my native-Python wrapper around the Numba code for the `PixelFunctionType` call (basically handing over the arguments directly to a `@nb.jit`-decorated Python function), I am getting an error in my Numba code, complaining that `in_arr` is a `Tuple()` (which apparently means Numba cannot do much with it, eg applying standard Numpy functions). Before the fix, the numba type being passed was `UniTuple` (which kept Numba happy).

I am not enough of a Numba expert to know if there's a way to cast within the code, but if this is something that can be fixed in the Python bindings, it would probably be very useful.

rouault:
> So I really can't figure out why Numba is getting a different type cast.

perhaps when numBuffers = 0 or 1 ?

zedrdave:
Indeed! I was checking for len 1, but not len 0.

Treating len 1 separately in the jit code, seems to do the trick. Not sure if there's a way to enforce a uniform type through hints in the bindings, but at least that seems solvable…

rouault:
> but not len 0.

so the issue is with len 0 ? Wondering if we shouldn't pass a single empty buffer then ? Hum, or maybe better, just fill the buffer passed to RasterIO() with nodata / 0 ?

zedrdave:
Sorry I wasn't clear:

- when len is 0: the type is not inferable and numba/numpy complains, but that's to be expected. I was already handling this with a `if len(in_arr) == 0: return`

- as it turns out, when len is 1, the type is also problematic for numpy (at least in terms of handling `in_arr` as a tuple). I managed to solve that by adding a wrapper that checks and handles len == 1 separately (so that the type inference in the default case function works).

FWIW, it seems like there is no good reason to call the function when the length is 0.

It would be nice (in our case) to avoid the unnecessary `out_arr[:, :] = in_arr[0]` for the case where len is 1, but I suspect other use cases might still need to process the array then.

PR #7250, opened 2023-02-15
Doc: building_from_source.rst: add a 'Typical build issues' paragraph, initiated with PROJ related issues (fixes #7248)

### LINKED ISSUE #7248: Error when compiling from source gdal 3.6.2
I am compiling gdal (3.6.2) from source in a container (with `almalinux:9.1`) where proj (9.1.1) is also compiled from source.

Compilation finishes without errors.
when I run `gdalinfo` I get:
`gdalinfo: error while loading shared libraries: libproj.so.22: cannot open shared object file: No such file or directory`

This is easily solved doing:
`ln -s /usr/lib64/libproj.so.25 /usr/lib64/libproj.so.22`

I think this means that `gdal` sees the libraries of `proj` without problems but for some reason is looking for a different version

I used the following `cmake` instructions

```
cmake .. \
    -DCMAKE_INSTALL_PREFIX=$installdir \
    -DPROJ_INCLUDE_DIR="${installdir}/include" \
    -DPROJ_LIBRARY="/usr/lib64/libproj.so"
cmake --build ./
cmake --build . --target install
```

Including the `-DPROJ`  instructions does not impact the behaviour described.

rouault:


doc/source/development/building_from_source.rst
```diff
@@ -2210,6 +2210,37 @@ First refer to https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.htm
 and to https://github.com/OSGeo/gdal/blob/master/.github/workflows/android_cmake/start.sh for
 an example of a build script to cross-compile from Ubuntu.
 
+
+Typical build issues
+++++++++++++++++++++
+
+How do I get PROJ ?
```
dbaston, line 2217: This may fit more naturally in `dev_environment.rst`

doc/source/development/building_from_source.rst
```diff
@@ -2210,6 +2210,37 @@ First refer to https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.htm
...
+*******************
+
+PROJ is the only required build-time dependency of GDAL that is not vendorized
+in GDAL source code tree. Consequently PROJ header and library must be available
```
dbaston, line 2221:
```suggestion
in the GDAL source code tree. Consequently, the PROJ header and library must be available
```

doc/source/development/building_from_source.rst
```diff
@@ -2210,6 +2210,37 @@ First refer to https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.htm
...
+If using a custom PROJ build (that is a PROJ build that does not come from
+a distribution channel), it can sometimes happen that this custom PROJ build
+conflicts with packaged dependencies, such as spatialite or libgeotiff, that
+themselves link to another copy of PROJ.
+
+The clean way to solve this is to rebuild from sources those other libraries
+against the custom PROJ build.
+For Linux based systems, given that C API/ABI has been preserved in the
+PROJ 6, 7, 8, 9 series, if the custom PROJ build is more recent than the
+PROJ used by those other libraries, doing aliases of the older libproj.so.XX
```
dbaston, line 2236:
```suggestion
PROJ used by those other libraries, doing aliases of the older ``libproj.so.XX``
```

doc/source/development/building_from_source.rst
```diff
@@ -2210,6 +2210,37 @@ First refer to https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.htm
...
+name to the newer libproj.so.YY (with ``ln -s``) should work, although this is
```
dbaston, line 2237:
```suggestion
name to the newer ``libproj.so.YY`` (with ``ln -s``) should work, although it is
```

doc/source/development/building_from_source.rst
```diff
@@ -2210,6 +2210,37 @@ First refer to https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.htm
...
+definitely not recommended to use such solution for production purposes.
```
dbaston, line 2238:
```suggestion
definitely not recommended to use this solution in a production environment.
```

doc/source/development/building_from_source.rst
```diff
@@ -2210,6 +2210,37 @@ First refer to https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.htm
...
+
+In any case, if ``ldd libgdal.so | grep libproj`` reports more than one line,
+crashes will occur at runtime (often at process termination).
```
dbaston, line 2241: Include possible error message(s) so that someone searching the message may find this section?

doc/source/development/building_from_source.rst
rouault: I'm not sure. dev_environment has this connotation this is for people who want to modify GDAL, but here it is problems for people who just want to build GDAL for use

dbaston: Maybe a better name for that page is "Setting up a build environment" ?

doc/source/development/building_from_source.rst
rouault: added

doc/source/development/building_from_source.rst
rouault: indeed. I'll change that in an extra commit

rouault: hum actually not, the content is rather dev oriented, with Vagrant, git, etc. Hard to organize properly! And difficult to draw the line between "user just compiling GDAL to use it because not happy with existing packages" and "developer wanting to hack in GDAL"

doc/source/development/building_from_source.rst
```diff
@@ -2210,6 +2210,38 @@ First refer to https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.htm
...
+conflicts with packaged dependencies, such as spatialite or libgeotiff, that
+themselves link to another copy of PROJ.
+
+The clean way to solve this is to rebuild from sources those other libraries
+against the custom PROJ build.
+For Linux based systems, given that C API/ABI has been preserved in the
+PROJ 6, 7, 8, 9 series, if the custom PROJ build is more recent than the
+PROJ used by those other libraries, doing aliases of the older ``libproj.so.XX``
+name to the newer ``libproj.so.YY`` (with ``ln -s``) should work, although it is
+definitely not recommended to use this solution in a production environment.
```
dg0yt, line 2238: Is the API/ABI also complete if the new newer build has less features (CURL, TIFF) than the older one?

rouault: good question. The answer is yes. TIFF presence/absence has no impact on API. And CURL one either, because networking capabilities can be present even if CURL isn't enabled, because the user can pass its own networking capable callbacks

PR #7264, opened 2023-02-17
RFC 94: Numeric fields width/precision metadata

elpaso:
Add RFC 94 text.

Note: I took 94 because there is already a PR about 93.

### DIFF
--- a/doc/source/development/rfc/index.rst
+++ b/doc/source/development/rfc/index.rst
@@ -101,4 +101,3 @@ RFC list
     rfc92_wkb_only_geometries
     rfc93_update_feature
     rfc94_field_precision_width_metadata
->>>>>>> 236190f1fd (RFC 94: Numeric fields width/precision metadata)
--- a/doc/source/development/rfc/index.rst
+++ b/doc/source/development/rfc/index.rst
@@ -100,3 +100,5 @@ RFC list
     rfc91_dataset_close
     rfc92_wkb_only_geometries
     rfc93_update_feature
+    rfc94_field_precision_width_metadata
+>>>>>>> 236190f1fd (RFC 94: Numeric fields width/precision metadata)
--- a/doc/source/development/rfc/rfc94_field_precision_width_metadata.rst
+++ b/doc/source/development/rfc/rfc94_field_precision_width_metadata.rst
@@ -34,10 +34,15 @@ There is currently no way for an application to access this information and this
 can lead to loss of width or precision while converting between formats or when the
 application decides to stay on the safe side and reduce the width reported by GDAL.
 
+Additionally, attention must be paid to the meaning of "width" and "precision":
+OGR "width" corresponds to SQL "precision" and OGR "precision" corresponds to SQL "scale".
+
+
 For reference:
 
 - https://trac.osgeo.org/gdal/ticket/6960
 - https://issues.qgis.org/issues/11755
+- https://issues.qgis.org/issues/15188#note-8
 - https://github.com/qgis/QGIS/issues/51849
 
 
--- a/doc/source/development/rfc/rfc94_field_precision_width_metadata.rst
+++ b/doc/source/development/rfc/rfc94_field_precision_width_metadata.rst
@@ -15,8 +15,8 @@ Target:        GDAL 3.7
 Summary
 -------
 
-The document proposes and describes the introduction of a new method
-to the `OGRFieldDefn` class that returns information for numeric real
+The document proposes and describes the introduction of a couple of new
+vector driver metadata that return information for numeric real
 fields about how the precision and width have been calculated by the
 driver and have to be interpreted.
 
@@ -92,74 +92,42 @@ Notes about specific drivers
 Technical details
 -----------------
 
-The implementation details are still to be defined, a possible
-approach would be to define a flag with values to define if
-the decimal place is included in the width and if the minus sign
-is included in the width.
+The vector drivers will expose a metadata entry to define if the width
+of the fields includes the decimal separator and/or the sign.
 
-Additional flag values related to the field definition can be added if
-needed.
+If the metadata entry is undefined the feature is not supported (there is no
+client-accessible width or precision constraint for numeric fields).
 
 Example API:
 
 .. code-block:: c++
 
-    /** Field width includes decimal separator */
-    #define OGR_F_WIDTH_INCLUDES_DECIMAL_SEPARATOR 0x1
-
-    /** Field width includes sign */
-    #define OGR_F_WIDTH_INCLUDES_SIGN 0x2
-
-
-    /************************************************************************/
-    /*                        OGR_Fld_GetWidthPrecisionFlags()              */
-    /************************************************************************/
-    /**
-    * \brief Returns flags representing information about how the width and
-    *        precision for this field must be interpreted.
+    /** Capability set by a vector driver that supports field width and precision.
     *
-    * This function returns zero for fields of types other than OFTReal or OFTRealList.
+    * This capability reflects that a vector driver includes the decimal separator
+    * in the field width.
     *
-    * This function is the same as the CPP method OGRFieldDefn::GetWidthPrecisionFlags().
-    *
-    * @param hDefn handle to the field definition to get width and precision flags from.
-    * @return the width and precision flags.
+    * See GDAL_DMD_NUMERIC_FIELD_WIDTH_INCLUDES_SIGN for a related capability flag.
+    * @since GDAL 3.7
     */
+    #define GDAL_DMD_NUMERIC_FIELD_WIDTH_INCLUDES_DECIMAL_SEPARATOR "DMD_NUMERIC_FIELD_WIDTH_INCLUDES_DECIMAL_SEPARATOR"
 
-    int OGR_Fld_GetWidthPrecisionFlags(OGRFieldDefnH hDefn)
-
-    {
-        return OGRFieldDefn::FromHandle(hDefn)->GetWidthPrecisionFlags();
-    }
-
-    /************************************************************************/
-    /*                        OGR_Fld_SetWidthPrecisionFlags()              */
-    /************************************************************************/
-    /**
-    * \brief Set the flags representing information about how the width and
-    *        precision for this field must be interpreted.
+    /** Capability set by a vector driver that supports field width and precision.
     *
-    * Calling this function on fields of types other than OFTReal or OFTRealList
-    * does nothing.
+    * This capability reflects that a vector driver includes the sign
+    * in the field width.
     *
-    * This function is the same as the CPP method OGRFieldDefn::SetWidthPrecisionFlags().
-    *
-    * @param hDefn handle to the field definition to set precision to.
-    * @param nFlags the new width and precision flags.
+    * See GDAL_DMD_NUMERIC_FIELD_WIDTH_INCLUDES__DECIMAL_SEPARATOR for a related capability flag.
+    * @since GDAL 3.7
     */
+    #define GDAL_DMD_NUMERIC_FIELD_WIDTH_INCLUDES_SIGN "DMD_NUMERIC_FIELD_WIDTH_INCLUDES_SIGN"
 
-    void OGR_Fld_SetWidthPrecisionFlags(OGRFieldDefnH hDefn, int nFlags)
-    {
-        return OGRFieldDefn::FromHandle(hDefn)->SetWidthPrecisionFlags(nFlags);
-    }
 
 
 Efficiency considerations
 --------------------------
 
-Field definitions will have to set an additional integer field for real and list of
-reals fields, only if the driver has any of the flags set, the default for the flags
-will be zero (no flags set).
+None.
 
 
 Backward compatibility
--- a/doc/source/development/rfc/rfc94_field_precision_width_metadata.rst
+++ b/doc/source/development/rfc/rfc94_field_precision_width_metadata.rst
@@ -41,6 +41,38 @@ For reference:
 - https://github.com/qgis/QGIS/issues/51849
 
 
+Current drivers behavior
+------------------------
+
+Here is a list of the drivers and how they behave with respect to the width and precision,
+(for databases we refer to the `NUMERIC` data type).
+
+================== ================================== =====================
+ Driver             Width Includes Decimal Separator   Width Includes Sign
+================== ================================== =====================
+ Shapefile          YES                                YES
+ MapInfo            ?                                  ?
+ PostgreSQL         NO                                 NO
+ MySQL              NO                                 NO
+ MSSQL              NO                                 NO
+ OCI                NO                                 NO
+ GPKG               N/A                                N/A
+ CSV (from .csvt)   N/A                                N/A
+ HANA               ?                                  ?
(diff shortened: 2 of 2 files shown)


doc/source/development/rfc/rfc94_field_precision_width_metadata.rst
```diff
@@ -0,0 +1,136 @@
...
+Contact:       elpaso @ itopen.it
+Started:       2023-Feb-17
+Status:        Development
+Target:        GDAL 3.7
+============== =============================================
+
+Summary
+-------
+
+The document proposes and describes the introduction of a new method
```
rouault, line 18: For creation use cases, where on wants to convert from one convention to another one, one needs to know in advance the convention that will be used by the output driver.
Perhaps have a GDAL_DMD_NUMERIC_FIELD_WITH_INCLUDES_DECIMAL_SEPARATOR driver metadata item whose value could be "YES" or "NO" (and assumed to be unknown if not set)
Similarly GDAL_DMD_NUMERIC_FIELD_WITH_INCLUDES_SIGN

doc/source/development/rfc/rfc94_field_precision_width_metadata.rst
```diff
@@ -0,0 +1,136 @@
...
+Motivation
+----------
+
+Applications (for example OGR/GDAL utils and QGIS) may require information
+about the field width and precision in order to convert between different
+formats or to validate user input on editing.
+
+Different drivers may calculate the width and precision differently by including
+or not the decimal separator and/or the minus sign into the calculated length.
+
```
rouault, line 32: We should try to establish in this RFC a table with the situation for drivers that use SetPrecision(), like Shapefile, MapInfo, PostgreSQL (are all SQL drivers the same like MySQL, MSSQL, OCI) , GPKG (this one will be N/A), CSV (from the .csvt side car), HANA (@mrylov can hopefully help us clarify the semantics), FlatGeoBuf, MEM, FileGDB, GML to see if we are not missing some subtelties

doc/source/development/rfc/rfc94_field_precision_width_metadata.rst
elpaso: I agree that that information shouldn't change within a single layer but since I do not know enough about all the involved drivers I preferred to stay on the safe side and make it a field property. 

Anyway, this can be decided at a later stage of the analysis when we are sure there aren't different field behaviors within a single layer.

doc/source/development/rfc/rfc94_field_precision_width_metadata.rst
elpaso: Good idea! I have added the table placing a `?` where I do not know or I am not sure.

doc/source/development/rfc/rfc94_field_precision_width_metadata.rst
rouault: my point in having that at a driver level was for a converter application (ogr2gr, QGIS vector file writer, etc.), that needs to create numeric fields. It needs to know *before* creating them what the convention will be to be able to add +1 or +2 to the width if needed.

doc/source/development/rfc/rfc94_field_precision_width_metadata.rst
```diff
@@ -0,0 +1,173 @@
...
+ MapInfo            ?                                  ?
+ PostgreSQL         NO                                 NO
+ MySQL              NO                                 NO
+ MSSQL              NO                                 NO
+ OCI                NO                                 NO
+ GPKG               N/A                                N/A
+ CSV (from .csvt)   N/A                                N/A
+ HANA               ?                                  ?
+ FlatGeoBuf         ?                                  ?
+ FileGDB            ?                                  ?
```
rouault, line 68: FileGDB is interesting. So looking at the code in ogr/ogrsf_frmts/filegdb/FGdbLayer.cpp, we currently ogr_width to the FileGDB <Field><Length> XML element and ogr_precision to <Field><Precision>. And we set <Field><Scale> to 0.
But there is https://desktop.arcgis.com/en/arcmap/latest/manage-data/tables/arcgis-field-data-types.htm#GUID-70728949-2016-4DB5-A723-5493B4330AE0 which mentions:
"Specifying the precision and scale allows you to restrict the range of values and number formats a field can accept, giving you greater control. For example, if you specify a float with a precision of 4 and a scale of 2, the field accepts 12.34 (or 12.3 or 12). If you enter 12.345 in the field, an error message displays, as this exceeds the maximum number of digits and decimal places allowed. On the other hand, if you specify a float with a precision of 5 and a scale of 3, the field allows you to enter 12.345, 12.34, or 12.3."

and https://resources.arcgis.com/en/help/arcobjects-java/api/arcobjects/com/esri/arcgis/geodatabase/Field.html#getPrecision():
"Precision is the number of digits in a number. For example, the number 56.78 has a precision of 4. Precision is only valid for fields that are numeric. Precision is always returned as 0 from personal or File geodatabase fields."

https://resources.arcgis.com/en/help/arcobjects-java/api/arcobjects/com/esri/arcgis/geodatabase/Field.html#getScale()
"Scale is the number of digits to the right of the decimal point in a number. For example, the number 56.78 has a scale of 2. Scale applies only to fields that are double. Scale is always returned as 0 from personal or File geodatabase fields."

and
https://resources.arcgis.com/en/help/arcobjects-java/api/arcobjects/com/esri/arcgis/geodatabase/Field.html#getLength()
"The maximum length, in bytes, for values described by the field. "

So it seems to me that the FileGDB driver needs to be corrected to not set <Precision> on writing, and ignore <Length> and <Precision> on reading for floating-point fields (<Length> for strings & binary fields is likely OK)
The OpenFileGDB one would just need to be corrected in its AlterFieldDefn() implementation to not propagate width and precision.

doc/source/development/rfc/rfc94_field_precision_width_metadata.rst
```diff
@@ -0,0 +1,173 @@
...
+ GML                ?                                  ?
```
rouault, line 69: https://www.oreilly.com/library/view/xml-schema/0596002521/re52.html
"xs:totalDigits defines the maximum number of digits of decimal and derived datatypes (both after and before the decimal point, not counting the decimal point itself). "

https://www.oreilly.com/library/view/xml-schema/0596002521/re24.html
xs:fractionDigits defines the maximum number of fractional digits (i.e., digits that are after the decimal point) of an xs:decimal datatype. 

So it seems to be NO NO

and experimenting with the result of above test.csv converted to GML with "ogr2ogr test.gml test.csv", and then editing test.gml:
```
$ ogrinfo GMLAS:test.gml -oo VALIDATE=YES
ERROR 1: test.gml:19:31 value '1.234' has '3' fraction digits which exceeds fractionDigits facet value '2'
$ ogrinfo GMLAS:test.gml -oo VALIDATE=YES
ERROR 1: test.gml:13:34 value '-1123.45' has '6' total digits which exceeds totalDigits facet value '5'
```
confirms it

doc/source/development/rfc/rfc94_field_precision_width_metadata.rst
```diff
@@ -0,0 +1,173 @@
...
+------------------------
+
+Here is a list of the drivers and how they behave with respect to the width and precision,
+(for databases we refer to the `NUMERIC` data type).
+
+================== ================================== =====================
+ Driver             Width Includes Decimal Separator   Width Includes Sign
+================== ================================== =====================
+ Shapefile          YES                                YES
+ MapInfo            ?                                  ?
```
rouault, line 59: TABDATFile::WriteDecimalField()
```
    snprintf(szFormat, sizeof(szFormat), "%%%d.%df", nWidth, nPrec);
    const char *pszVal = CPLSPrintf(szFormat, dValue);
    if (static_cast<int>(strlen(pszVal)) > nWidth)
    {
                 "Cannot format %g as a %d.%d field", dValue, nWidth, nPrec);
        return -1;
    }
```
tends to show it is the same as shapefile. The .dat file format is very similar to .dbf
Also see https://trac.osgeo.org/gdal/ticket/6392 and https://docs.safe.com/fme/html/FME_Desktop_Documentation/FME_ReadersWriters/mif/user-attributes.htm

doc/source/development/rfc/rfc94_field_precision_width_metadata.rst
```diff
@@ -0,0 +1,173 @@
...
+ MySQL              NO                                 NO
+ MSSQL              NO                                 NO
+ OCI                NO                                 NO
+ GPKG               N/A                                N/A
+ CSV (from .csvt)   N/A                                N/A
+ HANA               ?                                  ?
+ FlatGeoBuf         ?                                  ?
+ FileGDB            ?                                  ?
+ GML                ?                                  ?
+ MEM                ?                                  ?
```
rouault, line 70: MEM by itself stores floating point values in the double OGRField.Real field.
If the user asks OGRFeature::GetFieldAsString() then it will uses "%.{precision}" as shown above (but that's also true of other drivers).
So probably MEM should be in the N/A N/A category

doc/source/development/rfc/rfc94_field_precision_width_metadata.rst
elpaso: Do I get it right that for FileGDB is `N/A` ?

doc/source/development/rfc/rfc94_field_precision_width_metadata.rst
elpaso: I've changed the proposed implementation to a driver metadata entry.

doc/source/development/rfc/rfc94_field_precision_width_metadata.rst
rouault: yes FileGDB is N/A

rouault:
Open question: given that width_includes_sign and width_includes_decimal_separator are always coupled, should we merge this in a single metadata item ?

For ogr2ogr, should we, in the case of converting from width_includes_sign+width_includes_decimal_separator=NO (typically SQL drivers) to width_includes_sign+width_includes_decimal_separator=YES (shape, mapinfo), increase the width of created fields of 2 ? Not sure if that should be a opt-in or opt-out option.

elpaso:
> Open question: given that width_includes_sign and width_includes_decimal_separator are always coupled, should we merge this in a single metadata item ?

This is hard to tell, originally I was thinking at multiple flags to make it easier to extend this to other properties in the future (you never know: thousand separators?). Given the analysis we have done so far, a single bool that combines the sign and the decimal separator seems to be sufficient. I have seen mentions about to the leading `0` in an old thread but IIRC the conclusion was that it wasn't a problem.

> 
> For ogr2ogr, should we, in the case of converting from width_includes_sign+width_includes_decimal_separator=NO (typically SQL drivers) to width_includes_sign+width_includes_decimal_separator=YES (shape, mapinfo), increase the width of created fields of 2 ? Not sure if that should be a opt-in or opt-out option.

I think this is the expected behavior (no narrowing), so I'm +1 for opt-out.

doc/source/development/rfc/index.rst
```diff
@@ -100,3 +100,5 @@ RFC list
     rfc91_dataset_close
     rfc92_wkb_only_geometries
     rfc93_update_feature
+    rfc94_field_precision_width_metadata
+>>>>>>> 236190f1fd (RFC 94: Numeric fields width/precision metadata)
```
rouault, line 104:
```suggestion
```

elpaso: yeah, just noticed, I'm rebasing & force pushing

rouault:
ping @mrylov Could you help us clarify the semantics of width & precision for HANA ? Do they follow SQL conventions ?

mrylov:
> ping @mrylov Could you help us clarify the semantics of width & precision for HANA ? Do they follow SQL conventions ?

SAP HANA database follows the SQL standard notation for fixed-point decimals. It only supports DECIMAL type that is defined as follows.

DECIMAL(<p>, <s>), where <p> specifies precision, or the number of total digits (the sum of whole digits and fractional digits). <s> denotes scale, or the number of fractional digits. If a column is defined as DECIMAL(5, 4) for example, the numbers 3.14, 3.1415, 3.141592 are stored in the column as 3.1400, 3.1415, 3.1415, retaining the specified precision(5) and scale(4).

Precision can range from 1 to 38. The scale can range from 0 to <p>. If the scale is not specified, then it defaults to 0.

If precision and scale are not specified, then DECIMAL becomes a floating-point decimal number. In this case, precision and scale can vary within the range of 1 to 34 for precision and -6,111 to 6,176 for scale, depending on the stored value.

doc/source/development/rfc/rfc94_field_precision_width_metadata.rst
```diff
@@ -0,0 +1,169 @@
...
+================== ================================== =====================
+ Shapefile          YES                                YES
+ MapInfo            YES                                YES
+ PostgreSQL         NO                                 NO
+ MySQL              NO                                 NO
+ MSSQL              NO                                 NO
+ OCI                NO                                 NO
+ GPKG               N/A                                N/A
+ CSV (from .csvt)   YES                                YES
+ HANA               ?                                  ?
```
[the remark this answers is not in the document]
mrylov: Perhaps I misunderstand the proposed concept, but shouldn't this be NO/NO for HANA?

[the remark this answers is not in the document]
rouault: > Perhaps I misunderstand the proposed concept, but shouldn't this be NO/NO for HANA?

huh, obviously ! Fixing it...

doc/source/development/rfc/rfc94_field_precision_width_metadata.rst
```diff
@@ -0,0 +1,169 @@
...
+================== ================================== =====================
+ Shapefile          YES                                YES
+ MapInfo            YES                                YES
+ PostgreSQL         NO                                 NO
+ MySQL              NO                                 NO
+ MSSQL              NO                                 NO
+ OCI                NO                                 NO
+ GPKG               N/A                                N/A
+ CSV (from .csvt)   YES                                YES
+ HANA               YES                                YES
```
rouault, line 66:
```suggestion
 HANA               NO                                 NO
```

doc/source/development/rfc/rfc94_field_precision_width_metadata.rst
```diff
@@ -0,0 +1,169 @@
...
+
+=============================================================
+RFC 94: Numeric fields width/precision metadata
+=============================================================
+
+============== =============================================
+Author:        Alessandro Pasotti
+Contact:       elpaso @ itopen.it
+Started:       2023-Feb-17
+Status:        Development
```
rouault, line 11:
```suggestion
Status:        Adopted, implemented
```

doc/source/development/rfc/rfc94_field_precision_width_metadata.rst
```diff
@@ -0,0 +1,169 @@
...
+Testing
+-------
+
+A C++ test will be added to the test suite.
+
+
+Voting history
+--------------
+
+
```
rouault, line 169:
```suggestion
+1 from PSC members EvenR and JukkaR
```

PR #7265, opened 2023-02-17
TIFF JXL: add a JXL_ALPHA_DISTANCE creation option, and ALPHA_DISTANCE for standaone JPEGXL driver

rouault:
Typical use case is to generate a lossy image for the non-alpha channels and a lossless alpha channel with -co JXL_LOSSLESS=NO -co JXL_ALPHA_DISTANCE=0

Requires libjxl master (> 0.8.1)

CC @tbonfort . Do you think that's worth it ?


tbonfort:
@rouault that would be a valuable addition, thanks!

doc/source/drivers/raster/cog.rst
```diff
@@ -91,6 +91,12 @@ General creation options
    0=mathematically lossless, 1.0=visually lossless, usual range [0.5,3].
    Default is 1.0
 
+-  **JXL_ALPHA_DISTANCE=[-1,0,0.1-15]**: (GDAL >= 3.7, libjxl > 0.8.1)
```
elpaso, line 94: Can you add some more details about visually lossless and the "usual range"?

frmts/gtiff/geotiff.cpp
```diff
@@ -17385,6 +17415,18 @@ CPLErr GTiffDataset::OpenOffset(TIFF *hTIFFIn, toff_t nDirOffsetIn,
...
+                         EQUAL(pszKey, "JXL_ALPHA_DISTANCE"))
+                {
+                    const double dfVal = CPLAtof(pszValue);
+                    if (dfVal > 0 && dfVal <= 15)
+                    {
+                        m_oGTiffMDMD.SetMetadataItem(
+                            "COMPRESSION_REVERSIBILITY", "LOSSY",
+                            "IMAGE_STRUCTURE");
+                        m_fJXLAlphaDistance = static_cast<float>(dfVal);
+                    }
```
elpaso, line 17428: Does it make sense to report an out of bounds warning here?

frmts/gtiff/tif_jxl.c
```diff
@@ -1055,6 +1092,21 @@ static int JXLVSetField(TIFF *tif, uint32_t tag, va_list ap)
             return 1;
         }
 
+        case TIFFTAG_JXL_ALPHA_DISTANCE:
+        {
+            float alpha_distance = (float)va_arg(ap, double);
```
elpaso, line 1097:
```suggestion
            const float alpha_distance { static_cast<float>(va_arg(ap, double))};
```

frmts/jpegxl/jpegxl.cpp
```diff
@@ -2536,6 +2572,18 @@ GDALDataset *JPEGXLDataset::CreateCopy(const char *pszFilename,
                          "JxlEncoderSetExtraChannelName() failed");
                 return nullptr;
             }
+            if (channelType == JXL_CHANNEL_ALPHA && fAlphaDistance >= 0.0f)
+            {
+#if HAVE_JxlEncoderSetExtraChannelDistance
```
elpaso, line 2577: Shouldn't this be moved before the `if` ?

frmts/gtiff/tif_jxl.c
rouault: This is a .c file, so no ;-)

frmts/gtiff/geotiff.cpp
rouault: The generic logic that parses creation options inspects the value regarding the 'min' and 'max' value and will emit a warning.
Here it is a bit different as it parses metadata registered by GDAL itself, so it should be valid, unless someone has voluntarily corrupted it. I don't think there's a need for an extra check here

doc/source/drivers/raster/cog.rst
rouault: I've tried to improve based on the information based on the man page of "cjxl" utility (https://www.mankier.com/1/cjxl) that comes with libjxl, and the usage issue of the utility:
```
$ cjxl
[...]
-d maxError, --distance=maxError
    Max. butteraugli distance, lower = higher quality.
    0.0 = mathematically lossless. Default for already-lossy input (JPEG/GIF).
    1.0 = visually lossless. Default for other input.
    Recommended range: 0.5 .. 3.0. Allowed range: 0.0 ... 25.0.
    Mutually exclusive with --quality.
 -a maxError, --alpha_distance=maxError
    Max. butteraugli distance for the alpha channel, lower = higher quality.
    0.0 = mathematically lossless. 1.0 = visually lossless.
    Default is to use the same value as for the color image.
    Recommended range: 0.5 .. 3.0. Allowed range: 0.0 ... 25.0.
```

frmts/jpegxl/jpegxl.cpp
rouault: done

PR #7249, opened 2023-02-15
Add RFC 93 text: OGRLayer::UpdateFeature() method

rouault:
==> [rendered view](https://github.com/rouault/gdal/blob/rfc93_text/doc/source/development/rfc/rfc93_update_feature.rst)

PR #7269, opened 2023-02-17
RFC 93 implementation: OGRLayer::UpdateFeature()

rouault:
Implements https://github.com/OSGeo/gdal/pull/7249

### DIFF
--- a/ogr/ogrsf_frmts/generic/ogrunionlayer.cpp
+++ b/ogr/ogrsf_frmts/generic/ogrunionlayer.cpp
@@ -905,6 +905,108 @@ OGRErr OGRUnionLayer::IUpsertFeature(OGRFeature *poFeature)
 }
 
 /************************************************************************/
+/*                           IUpdateFeature()                           */
+/************************************************************************/
+
+OGRErr OGRUnionLayer::IUpdateFeature(OGRFeature *poFeature,
+                                     int nUpdatedFieldsCount,
+                                     const int *panUpdatedFieldsIdx,
+                                     int nUpdatedGeomFieldsCount,
+                                     const int *panUpdatedGeomFieldsIdx,
+                                     bool bUpdateStyleString)
+{
+    if (!bPreserveSrcFID)
+    {
+        CPLError(CE_Failure, CPLE_NotSupported,
+                 "UpdateFeature() not supported when PreserveSrcFID is OFF");
+        return OGRERR_FAILURE;
+    }
+
+    if (osSourceLayerFieldName.empty())
+    {
+        CPLError(CE_Failure, CPLE_NotSupported,
+                 "UpdateFeature() not supported when SourceLayerFieldName is "
+                 "not set");
+        return OGRERR_FAILURE;
+    }
+
+    if (poFeature->GetFID() == OGRNullFID)
+    {
+        CPLError(CE_Failure, CPLE_NotSupported,
+                 "UpdateFeature() not supported when FID is not set");
+        return OGRERR_FAILURE;
+    }
+
+    if (!poFeature->IsFieldSetAndNotNull(0))
+    {
+        CPLError(CE_Failure, CPLE_NotSupported,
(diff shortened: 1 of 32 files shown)


ogr/ogrsf_frmts/generic/ogrunionlayer.cpp
```diff
@@ -904,6 +904,108 @@ OGRErr OGRUnionLayer::IUpsertFeature(OGRFeature *poFeature)
...
+            std::vector<int> anSrcUpdatedFieldIdx;
+            const auto poLayerDefn = GetLayerDefn();
+            for (int j = 0; j < nUpdatedFieldsCount; ++j)
+            {
+                if (panUpdatedFieldsIdx[j] != 0)
+                {
+                    const int nNewIdx = poSrcLayerDefn->GetFieldIndex(
+                        poLayerDefn->GetFieldDefn(panUpdatedFieldsIdx[j])
+                            ->GetNameRef());
+                    if (nNewIdx >= 0)
```
elpaso, line 971: Isn't the index check already done in  https://github.com/OSGeo/gdal/pull/7269/files#diff-a6dc28611df439c6421d715cce8094c14d0fd92c2402a907c170727ae9844993R739 ?

No objection for a second check here but maybe we should log the error.

ogr/ogrsf_frmts/generic/ogrunionlayer.cpp
```diff
@@ -904,6 +904,108 @@ OGRErr OGRUnionLayer::IUpsertFeature(OGRFeature *poFeature)
...
+            std::vector<int> anSrcUpdatedGeomFieldIdx;
+            for (int j = 0; j < nUpdatedGeomFieldsCount; ++j)
+            {
+                if (panUpdatedGeomFieldsIdx[j] != 0)
+                {
+                    const int nNewIdx = poSrcLayerDefn->GetGeomFieldIndex(
+                        poLayerDefn
+                            ->GetGeomFieldDefn(panUpdatedGeomFieldsIdx[j])
+                            ->GetNameRef());
+                    if (nNewIdx >= 0)
```
elpaso, line 986: same here

ogr/ogrsf_frmts/gpkg/ogrgeopackagetablelayer.cpp
```diff
@@ -230,7 +230,8 @@ OGRErr OGRGeoPackageTableLayer::FeatureBindParameters(OGRFeature *poFeature,
     }
 
     /* Bind data values to the statement, here bind the blob for geometry */
-    if (poFeatureDefn->GetGeomFieldCount())
+    if ((nUpdatedGeomFieldsCount < 0 || nUpdatedGeomFieldsCount == 1) &&
```
elpaso, line 233: I don't understand this check.

ogr/ogrsf_frmts/generic/ogrunionlayer.cpp
rouault: > Isn't the index check already done in

No, the check in L739 only checks the validity of values of panUpdatedFieldsIdx[], which are indices related to the *union* layer. Here we check that they map to a *source* layer field, as the union layer can be "larger" than its source layers. e.g if you have layer A with fieldA and layer B with fieldB, the union layer will expose fieldA and fieldB, and if you try to update fieldB of a feature belonging to layer A, then it dosn't map to anything. The question if that should be an error is interesting. Currently for ISetFeature() we also silently discards fields that don't map to the source layer definition. I'm not totally sure of the behaviour we want here. I'd keep as current for consistency with ISetFeature(). Anyway, I'm not sure update of union layers is used by anyone. This is mostly used internally by the OGRSQL engine to implement UNION ALL, and in read-only mode.

ogr/ogrsf_frmts/gpkg/ogrgeopackagetablelayer.cpp
rouault: comment added

PR #7276, opened 2023-02-20
Use parametrize for mem test

elpaso:


autotest/gdrivers/mem.py
```diff
@@ -174,37 +174,71 @@ def test_mem_2():
...
+    free = crt.free
+    free.argtypes = [ctypes.c_void_p]
+    free.restype = None
+
+    # allocate band data array.
+    width = 50
+    height = 3
+    p = malloc(width * height * 4)
+    if p is None:
+        pytest.skip()
```
rouault, line 226: as this part is common with test_mem_2(), it could be put in a @pytest.fixture like
```
@pytext.fixture()
def mem_native_memory():
      ... everything up to here ...
      yield p
      free(p)
```

and you would use it with
```def test_geotransform(ds_definition, expected_sr, mem_native_memory):```

PR #7285, opened 2023-02-21
autotest: Symlink conftest.py into build directory

dbaston:
## What does this PR do?

This PR symlinks `conftest.py` into the build directory so that changes to that file are visible to the test suite without rerunning `cmake`.

Discussion in https://github.com/OSGeo/gdal/commit/9b3905e695acb0f9634e45acbce52500326f0f46

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/autotest/CMakeLists.txt
+++ b/autotest/CMakeLists.txt
@@ -47,30 +47,38 @@ if (Python_Interpreter_FOUND)
   set(AUTOTEST_LOG_FILE "${CMAKE_CURRENT_BINARY_DIR}/autotest.log")
   set(PYTEST_INI_HEADER_MESSAGE "This file was generated from ${GDAL_CMAKE_TEMPLATE_PATH}/pytest.ini.in using ${CMAKE_CURRENT_LIST_FILE}")
   configure_file(${GDAL_CMAKE_TEMPLATE_PATH}/pytest.ini.in ${CMAKE_CURRENT_BINARY_DIR}/pytest.ini @ONLY)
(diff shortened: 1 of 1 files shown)


autotest/CMakeLists.txt
```diff
@@ -47,37 +47,39 @@ if (Python_Interpreter_FOUND)
...
+  get_filename_component(symlink_name ${source} NAME)
+
+  if (CMAKE_VERSION VERSION_GREATER 3.14)
+    file(
+      CREATE_LINK ${source} ${dest_dir}/${symlink_name}
+      RESULT res
+      SYMBOLIC)
+    if (NOT res EQUAL 0)
+      message(STATUS "Copying content of ${source} to ${destination}")
+      execute_process(COMMAND ${CMAKE_COMMAND} -E copy ${source} ${dest_dir})
```
rouault, line 62: Not sure if that explains the failures on Windows, but trying on Linux, copy doesn't seem to work as expected when source is a directory.
Same at line 69

dbaston: Indeed, it returns success but does nothing. Just pushed a fix.

```
➜  gdal git:(conftest-symlink) ✗ mkdir /tmp/abc
➜  gdal git:(conftest-symlink) ✗ mkdir /tmp/def
➜  gdal git:(conftest-symlink) ✗ touch /tmp/abc/xyz
➜  gdal git:(conftest-symlink) ✗ cmake -E copy /tmp/abc /tmp/def/abc
➜  gdal git:(conftest-symlink) ✗ echo $?
0
➜  gdal git:(conftest-symlink) ✗ ls /tmp/def/abc
➜  gdal git:(conftest-symlink) ✗ 
```

PR #7286, opened 2023-02-21
autotest: use fixture in utilities/ and pyscripts/ to get the utility filename or skip

### LINKED ISSUE #7278: autotest: use fixture in utilities/ and pyscripts/ to get the utility filename or skip
Patterns like
```
def test_gdal_translate_35():
    if test_cli_utilities.get_gdal_translate_path() is None:
        pytest.skip()
```

should be changed to:
```
@pytest.fixture()
def gdal_translate_path():
     ... something ...
# or perhaps modify test_cli_utilities so that it exposes fixtures

def test_gdal_translate_35(gdal_translate_path):
```

rouault:
Fixes #7278

### DIFF
--- a/autotest/pymod/test_cli_utilities.py
+++ b/autotest/pymod/test_cli_utilities.py
@@ -42,45 +42,29 @@ cli_exe_path = {}
 def get_cli_utility_path_internal(cli_utility_name):
 
     if sys.platform == "win32":
-        cli_utility_name = cli_utility_name + ".exe"
+        cli_utility_name += ".exe"
 
-    # First try : in the apps directory of the GDAL source tree
-    # This is the case for the buildbot directory tree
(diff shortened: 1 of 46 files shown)


dbaston:
I was wondering if it would make sense to skip these tests only if the utilities should not be present (`BUILD_APPS=NO`, as you mentioned on #7278) rather than if they are not present (because of a build error?).

rouault:
> I was wondering if it would make sense to skip these tests only if the utilities should not be present (`BUILD_APPS=NO`, as you mentioned on #7278) rather than if they are not present (because of a build error?).

No strong opinion. This would add coupling between the build system and the tests, which might not be desirable for folks using out-of-tree build systems, but I suspect we already have coupling. I'll let that as a potential follow up.

PR #7290, opened 2023-02-22
VRT: 'expand' opt for vrt:// connection

mdsumner:
Adds 'expand' to the allowed options for a "vrt:// connection. The effect is to expand a palette image index by a 1 band dataset  RGB, RGBA, or Gray dataset. 

## Tasklist

 - [x] Add test case(s)
 - [x] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/autotest/gcore/vrt_read.py
+++ b/autotest/gcore/vrt_read.py
@@ -1456,7 +1456,7 @@ def test_vrt_protocol():
 
 
 @pytest.mark.require_driver("BMP")
-def test_vrt_protocol_bmp():
+def test_vrt_protocol_expand_option():
     ds = gdal.Open("vrt://data/8bit_pal.bmp?expand=rgb")
     assert ds.GetRasterBand(1).GetRasterColorInterpretation() == gdal.GCI_RedBand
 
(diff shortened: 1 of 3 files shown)


autotest/gcore/vrt_read.py
```diff
@@ -1454,6 +1454,9 @@ def test_vrt_protocol():
     ds = gdal.Open("vrt://data/byte_with_ovr.tif?ovr=0")
     assert ds.RasterXSize == 10
 
+    ds = gdal.Open("vrt://data/8bit_pal.bmp?expand=rgb")
```
rouault, line 1457: Move that to a new test method with a
``@pytest.mark.require_driver("BMP")``
decoration as in minimal builds the BMP driver isn't built

mdsumner: done, ty

autotest/gcore/vrt_read.py
```diff
@@ -1455,6 +1455,12 @@ def test_vrt_protocol():
     assert ds.RasterXSize == 10
 
 
+@pytest.mark.require_driver("BMP")
+def test_vrt_protocol_bmp():
```
rouault, line 1459:
```suggestion
def test_vrt_protocol_expand_option():
```

PR #7295, opened 2023-02-22
autotest: Add require_creation_option decorator

dbaston:
## What does this PR do?

In the same vein as #7286, #7288, #7274, etc. this PR adds a test decorator to indicate that a specified driver creation option is required. This should help ensure that skipped tests provide a useful and consistent message for the skip reason. (Currently there are ~5600 skips with no reason.)

I have only applied it to one test file for now. If it seems like a good idea I will apply it to the remaining test files (estimated 300-400 usages) and update this PR. 

#7286, #7288, #7274, etc.

## Tasklist

 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/autotest/gcore/tiff_write.py
+++ b/autotest/gcore/tiff_write.py
@@ -9501,12 +9501,10 @@ def test_tiff_write_jpegxl_uint16_single_band():
 # Test JXL_ALPHA_DISTANCE option
 
 
+@gdaltest.require_creation_option("GTiff", "JXL_ALPHA_DISTANCE")
 def test_tiff_write_jpegxl_alpha_distance_zero():
 
     drv = gdal.GetDriverByName("GTiff")
-    md = drv.GetMetadata()
-    if "JXL_ALPHA_DISTANCE" not in md["DMD_CREATIONOPTIONLIST"]:
-        pytest.skip("libjxl > 0.8.1 required")
 
     src_ds = gdal.Open("data/stefan_full_rgba.tif")
     filename = "/vsimem/test_tiff_write_jpegxl_alpha_distance_zero.tif"
--- a/autotest/gcore/cog.py
+++ b/autotest/gcore/cog.py
@@ -291,12 +291,8 @@ def test_cog_creation_of_overviews():
 # Test creation of overviews with a different compression method
 
 
+@gdaltest.require_creation_option("COG", "JPEG")
 def test_cog_creation_of_overviews_with_compression():
-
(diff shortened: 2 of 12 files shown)


autotest/pymod/gdaltest.py
```diff
@@ -1582,6 +1583,28 @@ def skip_on_travis():
     return False
 
 
+###############################################################################
+# Decorator to skip a test if specified creation option is not available
+
+
+def require_creation_option(driver, option):
+
+    md = gdal.GetDriverByName("GTiff").GetMetadata()
```
rouault, line 1592: Hard-coded driver name:
```suggestion
    md = gdal.GetDriverByName(driver).GetMetadata()
```
One issue I've noticed is that it seems that @pytest.mark.skipif() method decoration is evaluated before top pytestmark = pytest.mark.require_driver("driver_name"). Consequently, the gdal.GetDriverByName() above might be None. So it should be tested to skip with the reason that the driver is missing if needed

rouault (commented):
Good move!

autotest/gcore/tiff_write.py
```diff
@@ -3538,10 +3516,8 @@ def test_tiff_write_90():
 
 
 @pytest.mark.parametrize("external_ovr", [True, False])
+@gdaltest.require_creation_option("GTiff", "JPEG")
```
rouault, line 3519:
```suggestion
@gdaltest.require_creation_option("GTiff", "WEBP")
```

autotest/gcore/tiff_write.py
```diff
@@ -3599,18 +3575,15 @@ def test_tiff_write_90_webp(external_ovr):
...
```
rouault, line 3578:
```suggestion
@gdaltest.require_creation_option("GTiff", "WEBP")
```

autotest/gcore/tiff_write.py
rouault: oops ignore that one

rouault, line 3519:
```suggestion
@gdaltest.require_creation_option("GTiff", "WEBP")
```

autotest/pymod/gdaltest.py
dbaston: Thanks - fixed in 4f2b2c9c8455be08fc90de97b398fae523fe2818

dbaston:
I think this PR now covers the cases where this decorator can be easily used. I'm not sure if there is a clean solution for the cases where a test loops over/is parameterized by creation options.

Issue #4453, opened 2021-09-08
ngw tests flaky


rouault:
CC @BishopGIS 
ngw tests fail regularly those last 2 days. See https://github.com/OSGeo/gdal/pull/4451/checks?check_run_id=3547640912 for example

rouault:
another one: https://github.com/OSGeo/gdal/pull/4444/checks?check_run_id=3554652292

PR #7307, opened 2023-02-23
autotest: Mark ogr_ngw tests as slow

dbaston:
## What does this PR do?

Marks the ogr_ngw tests as slow. The slowest of these, ogr_ngw_test_sf, takes about 7 minutes to run on my machine. Several of the remaining tests take > 5 seconds to run, which is longer than other tests that are currently marked as slow (e.g., test_ogr_wfs_esri_2). But because the faster tests depend on state set by the slower tests, the entire module needs to be marked as slow.

Alternatively, or additionally, these tests could be marked with `pytest.mark.run_on_demand()`.

Based on a look at the current tests that are marked as slow, I think the testing docs could be updated to recommend a marking as slow for tests taking > 5 seconds.

#4453

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/autotest/ogr/ogr_ngw.py
+++ b/autotest/ogr/ogr_ngw.py
@@ -782,6 +782,7 @@ def test_ogr_ngw_15():
 #  Run test_ogrsf
 
 
+@pytest.mark.slow()  # 460 s
 def test_ogr_ngw_test_ogrsf():
     # FIXME: depends on previous test
     if gdaltest.ngw_ds is None:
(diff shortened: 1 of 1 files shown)


rouault:
(not related to this PR, but thinking about our today conversation about a smarter/faster CI, a very smart CI would detect that changes only affect source files of driver foo and would run only related tests, but that could get really complicated & fragile to implement. what about things touching core, etc.)

PR #7316, opened 2023-02-24
VRT 'gcp' option for vrt:// connection

mdsumner:
Adds 'gcp' (ground control point) to the allowed options for a "vrt:// connection.

## Tasklist

 - [x] Add test case(s)
 - [x] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/autotest/gcore/vrt_read.py
+++ b/autotest/gcore/vrt_read.py
@@ -1463,6 +1463,23 @@ def test_vrt_protocol():
     ds = gdal.Open("vrt://data/float32.tif?ot=Int32")
     assert ds.GetRasterBand(1).DataType == gdal.GDT_Int32
 
+    ## 5-element GCP
+    ds = gdal.Open("vrt://data/float32.tif?gcp=0,0,0,10,0")
+    assert ds.GetGCPCount() == 1
+
+    ## multiple GCPs
+    ds = gdal.Open("vrt://data/float32.tif?gcp=0,0,0,10,0&gcp=20,0,10,10,0")
+    assert ds.GetGCPCount() == 2
+
+    ## 4-element GCP also ok
+    ds = gdal.Open("vrt://data/float32.tif?gcp=0,0,0,10&gcp=20,0,10,10")
+    assert ds.GetGCPCount() == 2
+
+    try:
+        ds = gdal.Open("vrt://data/float32.tif?gcp=invalid")
+    except SyntaxError:
+        pass
+
 
 @pytest.mark.require_driver("BMP")
 def test_vrt_protocol_expand_option():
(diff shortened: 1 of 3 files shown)


nyalldawson:
Very nice! 👍

autotest/gcore/vrt_read.py
```diff
@@ -1463,6 +1463,23 @@ def test_vrt_protocol():
...
+    assert ds.GetGCPCount() == 2
+
+    ## 4-element GCP also ok
+    ds = gdal.Open("vrt://data/float32.tif?gcp=0,0,0,10&gcp=20,0,10,10")
+    assert ds.GetGCPCount() == 2
+
+    try:
+        ds = gdal.Open("vrt://data/float32.tif?gcp=invalid")
+    except SyntaxError:
+        pass
```
dbaston, line 1481:
```suggestion
    ds = gdal.Open("vrt://data/float32.tif?gcp=invalid")
    assert ds is None
```

dbaston: I don't think exceptions are enabled here, and if I turn them on I get a `RuntimeError` here.

mdsumner: ah, thank you - wasn't sure about this at all

mdsumner: why doesn't it error on open, though? that's weird to silently return null ... isn't it?

dbaston: See https://gdal.org/api/python_gotchas.html#python-bindings-do-not-raise-exceptions-unless-you-explicitly-call-useexceptions

mdsumner: ah nice, thank you 🙏

kadyb:
@mdsumner, one question. If `{stars}` doesn't support rasters with GCP reference, with this PR, I will be able to load raster as vrt with georeference?

mdsumner:
stars would possibly support it through 'st_warp or 'gdal_utils("warp", ', but possibly via read as well, I haven't tried with the sf or terra bindings yet

doc/source/drivers/raster/vrt.rst
```diff
@@ -1690,6 +1690,11 @@ modification of pixel values is done), as with (:ref:`gdal_translate`).
 The effect of the ``ot`` option (added in GDAL 3.7) is to force the output image bands to have a 
 specific data type supported by the driver as with (:ref:`gdal_translate`).
 
+The effect of the ``gcp`` option (added in GDAL 3.7) is to add the indicated ground control point 
+to the output dataset. Values are a set of numbers as per (:ref:`gdal_translate`)``pixel,line,easting,northing,elevation``.
```
rouault, line 1694:
```suggestion
to the output dataset. Values are a set of numbers as per (:ref:`gdal_translate`)``pixel,line,easting,northing[,elevation]``.  
```

PR #7317, opened 2023-02-26
polygonize: switch to The Two-Arm Chains EdgeTracing Algorithm

kikitte:
## What does this PR do?
This polygonize algorithm does a faster, memory saving, and robust polygonize job. It is an implementation of `The Two-Arm Chains EdgeTracing Algorithm` described in [Junhua Teng, Fahui Wang, Yu Liu, An Efficient Algorithm for Raster-to-Vector Data Conversion](https://doi.org/10.1080/10824000809480639).

Below is the some benefits of this algorithm compared to the original polygonize alg in GDAL:
- Fast
Tested on [OR_NLCD_2011.zip](https://oe.oregonexplorer.info/ExternalContent/SpatialDataforDownload/OR_NLCD_2011.zip), it takes **2:05.05** to finish, the original takes **12:18.47**. Testing script is `time -v gdal_polygonize.py -q nlcd_or_20111.tif nlcd_or_20111_newalg.shp` .

- Robust
This algorithm can handle many special cases gracefully, produce accurate polygon geometry without topology error. And polygons produced by this follow the right-hand rule for orientation (counterclockwise external rings, clockwise internal rings).
- Easy to extend
It is easy to extend this algorithm to do another jobs by deriving the  `PolygonReceiver` base class. For eaxmple, Finding the boundary cells of a raster.
- Less memory usage
The top memory usage is 1017932 kbytes, the original is 1246088 kbytes. 

I share it here and hope it can saving the people's time for raster to vector processing.

### DIFF
--- a/alg/CMakeLists.txt
+++ b/alg/CMakeLists.txt
@@ -31,6 +31,7 @@ add_library(
   gdalwarpoperation.cpp
   llrasterize.cpp
   polygonize.cpp
+  polygonize_polygonizer_impl.cpp
   rasterfill.cpp
   thinplatespline.cpp
   gdal_simplesurf.cpp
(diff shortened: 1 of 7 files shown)


kikitte:
I am sorry, there are some problems need to fix. When I've done, I will reopen this pr.

PR #7318, opened 2023-02-26
fix uclibc build without NPTL

ffontaine:
Check for `pthread_spin_lock` instead of `pthread_spin_lock_t` to avoid the following uclibc build failure:

```
/tmp/instance-7/output-1/build/gdal-3.5.2/port/cpl_multiproc.cpp: In function 'CPLSpinLock* CPLCreateSpinLock()': /tmp/instance-7/output-1/build/gdal-3.5.2/port/cpl_multiproc.cpp:2265:9: error: 'pthread_spin_init' was not declared in this scope; did you mean 'pthread_cond_init'?
 2265 |         pthread_spin_init(&(psSpin->spin), PTHREAD_PROCESS_PRIVATE) == 0 )
      |         ^~~~~~~~~~~~~~~~~
      |         pthread_cond_init
/tmp/instance-7/output-1/build/gdal-3.5.2/port/cpl_multiproc.cpp: In function 'int CPLAcquireSpinLock(CPLSpinLock*)':
/tmp/instance-7/output-1/build/gdal-3.5.2/port/cpl_multiproc.cpp:2283:12: error: 'pthread_spin_lock' was not declared in this scope; did you mean 'pthread_spinlock_t'?
 2283 |     return pthread_spin_lock( &(psSpin->spin) ) == 0;
      |            ^~~~~~~~~~~~~~~~~
      |            pthread_spinlock_t

```
Fixes:
 - http://autobuild.buildroot.org/results/aa2a88990a07e551c40efb0c2180768add600c4f

### DIFF
--- a/cmake/helpers/configure.cmake
+++ b/cmake/helpers/configure.cmake
@@ -110,9 +110,9 @@ else ()
     "
         #define _GNU_SOURCE
         #include <pthread.h>
-        int main() { pthread_spinlock_t spin; return 1; }
+        int main() { pthread_spinlock_t spin; return pthread_spin_lock(&spin); }
         "
-    HAVE_PTHREAD_SPINLOCK)
+    HAVE_PTHREAD_SPIN_LOCK)
 
   check_c_source_compiles(
     "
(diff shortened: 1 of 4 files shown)


rouault:
@ffontaine What is the effect of this patch ? Does it enable or disable spinlock ? I guess the later. But it is strange that pthread_spinlock_t would be available but not pthread_spin_lock() ?

ffontaine:
This change will disable spinlock if `pthread_spin_lock` is not available.
If you think that this change is not upstreamable, I can also add a dependency to NPTL to gdal on buildroot side.

rouault:
> If you think that this change is not upstreamable

no, that's fine. The use of the spinlock just brings a tiny perf improvement in scenarios with high usage of the raster block cache, but the fallback mutex implementation is OK.

PR #7326, opened 2023-02-27
netCDF: resolve variable names beyond the parent group (fixes #7325)

### LINKED ISSUE #7325: NetCDF: should resolve variable names beyond the parent group
From the source code in frmts/netcdf/netcdfdataset.cpp, resolving a variable name from a given starting group is performed following the CF logic, full description here: https://github.com/diwg/cf2/blob/master/group/cf2-group.adoc#scope.

## Expected behavior and actual behavior

Shortly, it means the variable is searched in then current group then in all the parent groups untill it is found.

Though, the resolution does not inspect groups above the parent group.

## Steps to reproduce the problem

For example, dealing with test files from sensor FCI on satellite series Meteosat Third Generation (https://www.eumetsat.int/mtg-test-data), we cannot retrieve the grid mapping associated to different variables, eg. `/data/vis_08/measured/effective_radiance` :
- the attribute grid_mapping of this variable contains `mtg_geos_projection`
- the variable `mtg_geos_projection` is located in the group `/data`
- from group `/data/vis_08/measured`, the NetCDF driver searches this variable only in the current group, its father group `/data/vis_08` and the children of this latter; as a consequence it does not find the variable
- the NetCDF driver should search in the parent groups till the root group and only then perform a width-wise descending search

## How to fix it

I am submitting a PR.

frdcms:
Fixes a bug in the resolution of a variable name from a given starting group following the CF logic.
Description of the logic here:
https://github.com/diwg/cf2/blob/master/group/cf2-group.adoc#scope

Without this fix, the resolution does not inspect groups above the parent group.


rouault:
It would be great to have a minimum reproducer test case, tested in autotest/gdrivers/netcdf.py. This can be typically done by running ncdump + ncgen on a real test case and stripping down to the minimum to get a tiny file with the desired characteristics. If that sounds too complicated, I'll have a look at that

frdcms:
I tried at the beginning to implement a unit test in a new file `autotest/cpp/test_gdal_netcdf.cpp`. Unfortunately, the build with HEAD on top of master branch failed so I gave up.

As the python API dos not expose a method to resolve a variable name from a given group, we can use a NetCDF file reproducing the structure described in the issue adressed by this PR (issue #7325) and check the `SpatialReference` associated to the variable `/data/vis_08/measured/effective_radiance` is at least not `None`.

I am adding this test file and such a test.

PR #7344, opened 2023-03-03
polygonize: using the two-arm chains edgetracing algorithm

kikitte:
## What does this PR do?
The Two-Arm Chains EdgeTracing Algorithm does a faster, memory saving, and robust polygonize job. It is described in [Junhua Teng, Fahui Wang, Yu Liu, An Efficient Algorithm for Raster-to-Vector Data Conversion](https://doi.org/10.1080/10824000809480639).

It is:

- Fast
The Two-Arm Chains EdgeTracing Algorithm is as fast as the original algorithm when processing small dataset, however it is much faster than the original algorithm when processing large dataset.  Please check the benchmarks below for detail.
- Robust
This algorithm can handle many special cases gracefully, produce 'correct' polygon geometry without topology error. And the polygons produced by this algorithm follow the right-hand rule (counterclockwise external rings, clockwise internal rings).
- Easy to extend
It is easy to extend this algorithm to do another jobs by deriving the  `PolygonReceiver` base class. For eaxmple, Finding the boundary cells of a raster.
- Less memory usage
This algorithm use less memory compare to the original in all tests. 

## Benchmarks

Test on AMD 4800h, 16GB RAM, 1T nvme ssd machine, with ArchLInux installed. GDAL 3.6.2 is used as the origin algorithm. 

#### [GDEM Colorized Map](https://asterweb.jpl.nasa.gov/images/GDEM-10km-colorized.tif)(4320x2160, 26.7 MiB)

4connected:

| Algorithm          | Command                                                      | Feature Count | Time(m:ss) | Peak Memory Usage(kbytes) |
| ------------------ | ------------------------------------------------------------ | ------------- | ---------- | ------------------------- |
| Origin             | time -v gdal_polygonize.py -q GDEM-10km-colorized.tif origin/GDEM-10km-colorized.shp | 1,742,271     | 0:10.39    | 205,384                   |
| TwoArm EdgeTracing | time -v gdal_polygonize.py -q GDEM-10km-colorized.tif twoarm/GDEM-10km-colorized.shp | 1,742,271     | 0:10.27    | 177,828                   |

8connected:

| Algorithm          | Command                                                      | Feature Count | Time(m:ss) | Peak Memory Usage(kbytes) |
| ------------------ | ------------------------------------------------------------ | ------------- | ---------- | ------------------------- |
| Origin             | time -v gdal_polygonize.py -8 -q GDEM-10km-colorized.tif origin/GDEM-10km-colorized_8c.shp | 1,585,277     | 0:10.03    | 198,956                   |
| TwoArm EdgeTracing | time -v gdal_polygonize.py -8 -q GDEM-10km-colorized.tif twoarm/GDEM-10km-colorized_8c.shp | 1,585,277     | 0:09.65    | 173,736                   |

#### random_grid.tif (5000x5000, 23.9 MiB)

This file is created with the following python script:

```python
import numpy as np
from osgeo import gdal_array

xsize = 5000
ysize = 5000
arr = np.random.randint(3, size=xsize * ysize).reshape(ysize, xsize).astype(np.int8)
gdal_array.SaveArray(arr, 'random_grid.tif')
```

4connected:

| Algorithm          | Command                                                      | Feature Count | Time(m:ss) | Peak Memory Usage(kbytes) |
| ------------------ | ------------------------------------------------------------ | ------------- | ---------- | ------------------------- |
| Origin             | time -v gdal_polygonize.py -q random_grid.tif origin/random_grid.shp | 9,274,519     | 1:06.47    | 569,468                   |
| TwoArm EdgeTracing | time -v gdal_polygonize.py -q random_grid.tif twoarm/random_grid.shp | 9,274,519     | 1:07.29    | 465,776                   |

8connected:

| Algorithm          | Command                                                      | Feature Count | Time(m:ss) | Peak Memory Usage(kbytes) |
| ------------------ | ------------------------------------------------------------ | ------------- | ---------- | ------------------------- |
| Origin             | time -v gdal_polygonize.py -8 -q random_grid.tif origin/random_grid_8c.shp | 2,706,688     | 0:56.10    | 324,648                   |
| TwoArm EdgeTracing | time -v gdal_polygonize.py -8 -q random_grid.tif twoarm/random_grid_8c.shp | 2,706,688     | 0:36.80    | 272,500                   |

#### [OR_NLCD_2011(converted to geotiff)](https://oe.oregonexplorer.info/ExternalContent/SpatialDataforDownload/OR_NLCD_2011.zip)(21959x16118, 337.6 MiB)

4connected:

| Algorithm          | Command                                                      | Feature Count | Time(m:ss) | Peak Memory Usage(kbytes) |
| ------------------ | ------------------------------------------------------------ | ------------- | ---------- | ------------------------- |
| Origin             | time -v gdal_polygonize.py -q nlcd_or_20111.tif origin/nlcd_or_20111.shp | 5,699,595     | 12:32.31   | 1,246,096                 |
| TwoArm EdgeTracing | time -v gdal_polygonize.py -q nlcd_or_20111.tif twoarm/nlcd_or_20111.shp | 5,699,595     | 1:56.53    | 1,018,524                 |

8connected:

| Algorithm          | Command                                                      | Feature Count | Time(m:ss) | Peak Memory Usage(kbytes) |
| ------------------ | ------------------------------------------------------------ | ------------- | ---------- | ------------------------- |
| Origin             | time -v gdal_polygonize.py -8 -q nlcd_or_20111.tif origin/nlcd_or_20111_8c.shp | 1,877,971     | 33:53.34   | 1,211,828                 |
| TwoArm EdgeTracing | time -v gdal_polygonize.py -8 -q nlcd_or_20111.tif twoarm/nlcd_or_20111_8c.shp | 1,877,971     | 2:47.36    | 953,976                   |


alg/polygonize.cpp
```diff
@@ -664,9 +172,6 @@ static CPLErr GDALPolygonizeT(GDALRasterBandH hSrcBand,
         eErr = GDALRasterIO(hSrcBand, GF_Read, 0, iY, nXSize, 1, panThisLineVal,
                             nXSize, 1, eDT, 0, 0);
 
-        if (eErr == CE_None && hMaskBand != nullptr)
```
rouault, line 667: Is the removal of the masking appropriate here ? Isn't there a risk of setting a different ID in the first and second pass if both don't apply the masking ?

rouault (commented):
Amazing work ! I've issue a pull request against your fork in https://github.com/kikitte/gdal/pull/1 with various non substantial improvements.
Are you connected/affiliate with the authors of the paper, or did you do the implementation just from it (I've had a look at it and although I see the relationship between your code and the paper, I also see that there are various gaps you had to fill in)

kikitte:
No, I have no any connections to the authors of this paper, and I do the implementation just from it. Yes, actually this paper is less clarity in many details(but it shows us a nice idea of r2v alg), so I filled these gaps during implementation on my own.

alg/polygonize.cpp
kikitte: This is indeed a problem, it need to mask the raster data while reading.
Since the polygonizer needs a full labeling raster as its input(via Polygonizer::processLine), it means that any zero area defined by the mask should be assigned to a unique value, but this looks like a time-consuming job to find the unique value.

kikitte: Because the most used case is to use nodata area as mask, so the nodata value serve as the unique value, and all looks ok in that case.
And I think the logic of the polygonizer can be update to take into account the "invalid polygon" concept.

rouault: Agreed that nodata is probably the main use case, but we should make sure that it works with an arbitrary mask band, not necessarily tied to nodata. I would suggest you restore the use of the mask band in the first pass, per this initial pull request, and potentially come with improvements/optimizations in a follow-up pull request. I don't think the main perf improvements of the new algorithm are related to that

kikitte: I've tried to fix this problem, it should work well but with little memory usage increasing because the polygonizer may hole the invalid polygon for a long time.

rouault: do we need an extra test case in autotest to catch those cases?

kikitte: No, no problem. It is caused by my carelessness, actually the result is correct, I don't remember why I written the wrong feature count  in the benchmark table, now I have updated it.
Yes, we'd better add a none nodata mask as a test case, it would help us to avoid this situation in the feature.

rouault: > Yes, we'd better add a none nodata mask as a test case

do you want to add that in this pull request ?

kikitte: Yes, just now I've add a test case.

autotest/alg/polygonize.py
```diff
@@ -402,3 +402,80 @@ def test_polygonize_7():
         assert (
             abs(value - dn_area_vector[key]) < pixel_area
         ), "polygonized vector area not match raster area"
+
+
+###############################################################################
+# Test a none nodata mask(a user defined mask) case to make sure the polygonized area match that mask.
+
+
+def test_polygonize_8():
```
rouault, line 411:
```suggestion
@pytest.mark.require_driver("AAIGRID")
def test_polygonize_8():
```

rouault:
Merged. Thanks again @kikitte !

PR #7353, opened 2023-03-05
gdal_polygonize.py: use transactions to speed-up writing

rouault:
Running ``gdal_polygonize.py GDEM-10km-colorized.tif out.gpkg`` goes from 2m14s to 18s on https://asterweb.jpl.nasa.gov/images/GDEM-10km-colorized.tif


elpaso (approved):

elpaso:
I assume this is already CI tested, correct?

rouault:
> I assume this is already CI tested, correct?

yes, gdal_polygonize.py is CI tested

PR #7355, opened 2023-03-05
FindCryptoPP.cmake: properly take into account _LIBRARY_RELEASE/_DEBUG (fixes #7338)

### LINKED ISSUE #7338: Wrong library import generated if providing CRYPTOPP_LIBRARY_DEBUG and CRYPTOPP_LIBRARY_RELEASE
If I providing both variables, **CRYPTOPP_LIBRARY_DEBUG** and **CRYPTOPP_LIBRARY_RELEASE**, and generate Visual Studio projects with cmake, the generated project GDAL.vcxproj contains invalid settings. The linked input libraries are generated wrong:
```
[...]
<Link>    <AdditionalDependencies>odbc32.lib;optimized%3BC:\Src\MyDlls\cryptopp\8.8\v143\lib\x64\release\cryptopp.lib%3Bdebug%3BC:\Src\MyDlls\cryptopp\8.8\v143\lib\x64\debug\cryptopp.lib;wbemuuid.lib;...</AdditionalDependencies>
</Link>
[...]
```
The libpath generated in `<gdal-base-dir>\cmake\modules\packages\FindCryptoPP.cmake`, which seems to be generated by function `select_library_configurations(CRYPTOPP)`, seems not to be treated correct for generating VS-projects.

Should set different correct formatted library-paths for debug and release projects,
- for relase-config e.g. : `C:\Src\MyDlls\cryptopp\8.8\v143\lib\x64\release\cryptopp.lib`
- for debug-config e.g. : `C:\Src\MyDlls\cryptopp\8.8\v143\lib\x64\debug\cryptopp.lib`

- build your own cryptopp
- build gdal with providing the following parameters in cmak.config:
```
set (CRYPTOPP_INCLUDE_DIR "C:\\Src\\MyDlls\\cryptopp\\8.8\\v143\\include" CACHE STRING "" FORCE)
set (CRYPTOPP_LIBRARY_RELEASE "C:\\Src\\MyDlls\\cryptopp\\8.8\\v143\\lib\\x64\\release\\cryptopp.lib" CACHE STRING "" FORCE)
set (CRYPTOPP_LIBRARY_DEBUG "C:\\Src\\MyDlls\\cryptopp\\8.8\\v143\\lib\\x64\\debug\\cryptopp.lib" CACHE STRING "" FORCE)
set (CRYPTOPP_USE_ONLY_CRYPTODLL_ALG ON CACHE BOOL "" FORCE)
set (GDAL_USE_CRYPTOPP ON CACHE BOOL "" FORCE)
```

## Operating system

Windows 11

3.6.2

rouault:
Credits to @dg0yt for hinting the fix

### DIFF
--- a/cmake/modules/packages/FindCryptoPP.cmake
+++ b/cmake/modules/packages/FindCryptoPP.cmake
@@ -78,7 +78,20 @@ if(CRYPTOPP_FOUND)
         add_library(CRYPTOPP::CRYPTOPP UNKNOWN IMPORTED)
         set_target_properties(CRYPTOPP::CRYPTOPP PROPERTIES
                               INTERFACE_INCLUDE_DIRECTORIES "${CRYPTOPP_INCLUDE_DIR}"
-                              IMPORTED_LINK_INTERFACE_LANGUAGES "CXX"
-                              IMPORTED_LOCATION "${CRYPTOPP_LIBRARY}")
+                              IMPORTED_LINK_INTERFACE_LANGUAGES "CXX")
+
+        if(CRYPTOPP_LIBRARY_RELEASE)
+            set_property(TARGET CRYPTOPP::CRYPTOPP APPEND PROPERTY
+                    IMPORTED_CONFIGURATIONS RELEASE)
+            set_target_properties(CRYPTOPP::CRYPTOPP PROPERTIES
+                    IMPORTED_LOCATION_RELEASE "${CRYPTOPP_LIBRARY_RELEASE}")
+        endif()
+
+        if(CRYPTOPP_LIBRARY_DEBUG)
+            set_property(TARGET CRYPTOPP::CRYPTOPP APPEND PROPERTY
+                    IMPORTED_CONFIGURATIONS DEBUG)
+            set_target_properties(CRYPTOPP::CRYPTOPP PROPERTIES
+                    IMPORTED_LOCATION_DEBUG "${CRYPTOPP_LIBRARY_DEBUG}")
+        endif()
    endif()
 endif()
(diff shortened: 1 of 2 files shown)


cmake/modules/packages/FindCryptoPP.cmake
```diff
@@ -77,8 +77,26 @@ if(CRYPTOPP_FOUND)
...
+        if(CRYPTOPP_LIBRARY_DEBUG)
+            set_property(TARGET CRYPTOPP::CRYPTOPP APPEND PROPERTY
+                    IMPORTED_CONFIGURATIONS DEBUG)
+            set_target_properties(CRYPTOPP::CRYPTOPP PROPERTIES
+                    IMPORTED_LOCATION_DEBUG "${CRYPTOPP_LIBRARY_DEBUG}")
+        endif()
+
+        if(NOT CRYPTOPP_LIBRARY_RELEASE AND NOT CRYPTOPP_LIBRARY_DEBUG)
+            set_target_properties(CRYPTOPP::CRYPTOPP PROPERTIES
+                    IMPORTED_LOCATION_RELEASE "${CRYPTOPP_LIBRARY}")
```
dg0yt, line 99: IMPORTED_LOCATION? (Does this case exist at all?)

rouault: > (Does this case exist at all?)

Presumably not, looking at https://cmake.org/cmake/help/latest/module/SelectLibraryConfigurations.html. I've removed that case, and the original in FindDeflate.cmake from which I copied&pasted

dg0yt: CMake's `FindZLIB.cmake` has similar code, however it doesn't even look for for _DEBUG/_RELEASE if `ZLIB_LIBRARY` is already set. I guess it is mostly backward compatibility, but it could also be single-config convenience. In any case, it is not a visible cache variable unless created by the user.

elpaso (approved):

PR #7356, opened 2023-03-05
FindPoppler.cmake: check that Poppler private headers are available (fixes #7352)

### LINKED ISSUE #7352: CMake: Poppler detection should check for Object.h private header availability
GDAL depends on Poppler being built with ENABLE_UNSTABLE_API_ABI_HEADERS=YES CMake option. Currently we successfully detect a Poppler build without it, but fail at building time since headers like Object.h aren't available. During configuration phase, we should issue an explicit warning hinting the user. And https://gdal.org/development/building_from_source.html#poppler should also underline this

rouault:


cmake/modules/packages/FindPoppler.cmake
```diff
@@ -134,6 +134,14 @@ find_package_handle_standard_args(Poppler
                                   HANDLE_COMPONENTS)
 mark_as_advanced(Poppler_INCLUDE_DIR Poppler_LIBRARY)
 
+if(Poppler_FOUND)
+  if (NOT EXISTS "${Poppler_INCLUDE_DIR}/Object.h")
+    message(WARNING "Poppler private headers not found. Make sure you build Poppler with -DENABLE_UNSTABLE_API_ABI_HEADERS")
+    unset(Poppler_FOUND)
+    unset(POPPLER_FOUND)
```
elpaso, line 141: Just curious, why is it necessary to check for upper cased?

elpaso (approved):

cmake/modules/packages/FindPoppler.cmake
rouault: cf documentation of FOUND_VAR in https://cmake.org/cmake/help/latest/module/FindPackageHandleStandardArgs.html. It as needed because CMake GDAL code later seemed to test POPPLER_FOUND

PR #7373, opened 2023-03-07
Pansharpening: require geotransform on panchromatic and multispectral bands

rouault:
Pansharpening now requires that panchromatic and multispectral bands have valid geotransform (in early versions, it was assumed in the case of missing geotransform that they covered the same geospatial extent). The undocumented VRT pansharpened MSShiftX and MSShiftY options (and the corresponding C++ GDALPansharpenOptions::dfMSShiftX and dfMSShiftY members) have been removed, due to using the inverted convention as one would expect, and being sub-par solution compared to using geotransform to correlate pixels of panchromatic and multispectral bands.

### DIFF
--- a/MIGRATION_GUIDE.TXT
+++ b/MIGRATION_GUIDE.TXT
@@ -25,6 +25,15 @@ MIGRATION GUIDE FROM GDAL 3.6 to GDAL 3.7
   drivers with write support are encouraged to implement it for error
   propagation.
 
+- Pansharpening now requires that panchromatic and multispectral bands have
+  valid geotransform (in early versions, it was assumed in the case of missing
+  geotransform that they covered the same geospatial extent).
+  The undocumented VRT pansharpened MSShiftX and MSShiftY options (and the
+  corresponding C++ GDALPansharpenOptions::dfMSShiftX and dfMSShiftY members)
+  have been removed, due to using the inverted convention as one would expect,
+  and being sub-par solution compared to using geotransform to correlate pixels
+  of panchromatic and multispectral bands.
+
 MIGRATION GUIDE FROM GDAL 3.5 to GDAL 3.6
 -----------------------------------------
 
--- a/alg/gdalpansharpen.cpp
+++ b/alg/gdalpansharpen.cpp
@@ -31,6 +31,7 @@
 #include "gdalpansharpen.h"
 
 #include <algorithm>
+#include <array>
 #include <cstddef>
 #include <cstdio>
 #include <cstdlib>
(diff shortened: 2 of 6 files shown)


rouault:
CC @tbonfort

MIGRATION_GUIDE.TXT
```diff
@@ -25,6 +25,15 @@ MIGRATION GUIDE FROM GDAL 3.6 to GDAL 3.7
   drivers with write support are encouraged to implement it for error
   propagation.
 
+- Pansharpening now requires that panchromatic and multispectral bands have
+  valid geotransform (in early versions, it was assumed in the case of missing
+  geotransform that they covered the same geospatial extent).
```
tbonfort, line 30:
```suggestion
  valid geotransform (in early versions the geotransform was not considered, i.e. it was expected that all bands covered the same geospatial extent).
```

rouault: your suggestion is actually not quite true. If the geotransforms of the pan and ms band differed by more than one pixel of the pan, then the SpatialExtentAdjustment=Union default logic would add the necessary padding so their geotransform was taken into account (but with integer rounding, so not a sub-pixel accuracy from what I can see)

rouault:
@tbonfort Did you get or will you have a chance to test that before I merge?

tbonfort:
Hi @rouault ,

I just ran a test and I believe there is an issue: I'm getting some `ERROR 1: Recursion detected` when forcing an offsetted geotransform on the ms dataset (they actually only seem to be warnings as I'm getting some seemingly correct output at the end).

This happens when I am using a vrt on the MS file to force an "incorrect" geotransform for testing.

- using pan.tif with gt(0,1,0,0,0,1) and ms.tif with gt(0,4,0,0,0,4) (i.e. the nominal case), everything works correctly.

- If I now create an msbad.vrt file with gt(-100,5,0,-100,5,0) and reference this msbad.vrt instead of ms.tif inside my pansharpening vrt, I get the recursion error. From some quick testing, setting any geotransform different that 0,4,0,0,0,4 raises the error.

- If I create and use an ms.vrt file that has the "correct" geotransform, there is no error raised.

tbonfort:
FWIW, the recursion guard getting hit is this one: https://github.com/rouault/gdal/blob/69e65c35ebe2c236f5d4b9c8867005edab32fcd7/frmts/vrt/vrtrasterband.cpp#L1342

rouault:
> I just ran a test and I believe there is an issue: I'm getting some `ERROR 1: Recursion detected`

fixed by commits added to this PR

PR #7381, opened 2023-03-08
return const OGRSpatialReference* for a few methods(fixes #7377)

### LINKED ISSUE #7377: Make OGRCoordinateTransformation::GetSourceCS/GetTargetCS() return a const OGRSpatialReference*
Cf https://github.com/PDAL/PDAL/commit/5e34b3bc1e81d85d2f2e481e4d4d49f69aec09ce#r103462358

rouault:
-  Make OGRCoordinateTransformation::GetSourceCS() and GetTargetCS() return a const OGRSpatialReference* (fixes #7377)
   
    to reflect that those objects should be considered immutable.
    
    The downside is that we have internally to
    const_cast<OGRSpatialReference*>, in particular in
    OGRGeometry::transform() when assigning GetTargetCS() to the geometry,
    as assignSpatialReference() needs to increase the reference count.

-  Make OGRGeometry::getSpatialReference() return a const OGRSpatialReference*

-  Make OGRGeomFieldDefn::GetSpatialRef() return a const OGRSpatialReference*

### DIFF
--- a/MIGRATION_GUIDE.TXT
+++ b/MIGRATION_GUIDE.TXT
@@ -37,6 +37,8 @@ MIGRATION GUIDE FROM GDAL 3.6 to GDAL 3.7
 - OGRCoordinateTransformation::GetSourceCS() and GetTargetCS() now returns
   a const OGRSpatialReference*
 
+- OGRGeometry::getSpatialReference() now returns a const OGRSpatialReference*
+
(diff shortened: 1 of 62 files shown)


rouault:
> would it be possible to mark `Reference` mutable and avoid the cast?

I considered this. The issue is that we have a GetRefCount() method, so nRefCount is not just an implementation detail and Reference() has thus visible effects which make it a bit hard to tag is a const.

PR #7389, opened 2023-03-09
OpenFileGDB: add read-only support for raster datasets

rouault:
The OpenFileGDB driver provides read access to rster layers of File
Geodatabases (.gdb directories) created by ArcGIS 10 and above. The
dataset name must be the directory/folder name, and it must end with the
.gdb extension.
    
It can also read directly zipped .gdb directories (with .gdb.zip
extension), provided they contain a .gdb directory at their first level.
    
The driver supports:
- reading CRS information
- reading geotransform
- exposing overviews
- exposing nodata mask band or nodata value
- uncompressed, LZ77, JPEG and JPEG2000 compression methods.
- exposing value attribute tables as GDAL Raster attribute tables.


rouault:
FYI @r-barnes

nyalldawson (approved):
I've already extensively tested and reviewed this branch prior to PR submission, and it's all looking/working great :partying_face:

AlexanderJuestel:
@rouault, thanks for implementing that. Could you point me to an example where raster data is accessed via gdal? I am quite new to gdal....I have installed gdal 3.7.1 and also have a .gdb file but cannot seem to get the raster layers. I can, however, list them using the fiona package.

Thanks for your help!

benfarmer05:
Cross-posting my reply on stack exchange regarding opening rasters from geodatabases using R (https://gis.stackexchange.com/a/494120/231013) just in case it is useful here as well:

------

As of June 2025, you can absolutely access raster data in a geodatabase (and without any proprietary ESRI software required)! You will need to have the latest spatial packages installed (if you need older versions, this may be less helpful for you), and in my case this worked on a Macbook using homebrew. I installed GDAL, GEOS, and PROJ (e.g., 'brew install gdal' after homebrew is set up). After this is done, make sure R is up to date, and then install terra (and sf, etc. if you want them) from source by running the below in R Studio or equivalent:

install.packages("terra", type = "source")

This will ensure that terra, etc., are linked properly with the spatial packages on your system (i.e., the homebrew installs). This step is really important; personally I have had the most sucess with doing installs from source rather than in the built-in R Studio package update manager.

Once this is done, you should be able to import the raster(s) you need from a geodatabase as follows:

gdb_path <- "/Users/your_user/Documents/your_dir/your_project.gdb"

your_raster <- rast(gdb_path, subds = "your_raster_layer")

PR #7393, opened 2023-03-10
COG: relax a bit the tolerance when computing tile number 

rouault:
(fine tune 6310c1aecb3b31365fb36daedf064b3aaf3884ca)

### DIFF
--- a/autotest/gcore/cog.py
+++ b/autotest/gcore/cog.py
@@ -581,7 +581,6 @@ def test_cog_byte_to_web_mercator_manual():
     assert ds.RasterYSize == 1024
     assert ds.GetRasterBand(1).GetMaskFlags() == gdal.GMF_ALPHA + gdal.GMF_PER_DATASET
     assert ds.GetRasterBand(1).GetBlockSize() == [256, 256]
-    gt = ds.GetGeoTransform()
     expected_gt = [
         -13149614.849955443,
         76.43702828517598,
@@ -590,9 +589,7 @@ def test_cog_byte_to_web_mercator_manual():
         0.0,
         -76.43702828517598,
     ]
-    for i in range(6):
-        if gt[i] != pytest.approx(expected_gt[i], abs=1e-10 * abs(expected_gt[i])):
-            assert False, gt
+    assert ds.GetGeoTransform() == pytest.approx(expected_gt, rel=1e-10)
     assert ds.GetRasterBand(1).Checksum() in (
         4363,
         4264,  # got on Mac at some point
@@ -602,9 +599,26 @@ def test_cog_byte_to_web_mercator_manual():
     assert ds.GetRasterBand(1).GetMaskBand().Checksum() == 4356
     assert ds.GetRasterBand(1).GetOverviewCount() == 2
     ds = None
-
     src_ds = None
+
+    # Check that we correctly round to the closest tile if input bounds are
+    # very close to its boundary (less than half a pixel)
+    filename2 = directory + "/cog2.tif"
+    eps = 0.49 * res
+    gdal.Translate(
+        filename2,
+        filename,
+        options="-of COG -co TILING_SCHEME=GoogleMapsCompatible -a_ullr %.18g %.18g %.18g %.18g"
+        % (minx - eps, maxy + eps, maxx + eps, miny - eps),
+    )
+    ds = gdal.Open(filename2)
+    assert ds.RasterXSize == 1024
+    assert ds.RasterYSize == 1024
+    assert ds.GetGeoTransform() == pytest.approx(expected_gt, rel=1e-10)
+    ds = None
+
     gdal.GetDriverByName("GTiff").Delete(filename)
+    gdal.GetDriverByName("GTiff").Delete(filename2)
     gdal.Unlink(directory)
 
 
--- a/frmts/gtiff/cogdriver.cpp
+++ b/frmts/gtiff/cogdriver.cpp
@@ -394,14 +394,16 @@ static bool COGGetWarpingCharacteristics(
         const double dfOriY =
             bInvertAxis ? tmList[0].mTopLeftX : tmList[0].mTopLeftY;
         const double dfTileExtent = dfRes * nBlockSize;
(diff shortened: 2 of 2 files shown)


blacha:
Thanks for taking a look at this! I am not sure this solves my problem, In my example COG i am still seeing a extra empty tile be created in the top and left.

image

I have a `projwin` set to 
```
-projwin 19466370.897442425 -5021607.010222964 19466982.363668706 -5022218.506449246 -projwin_srs EPSG:3857
```

I have built a debug version then I have dumped the values for these equations

```
nTLTileX: (19466370.836151096970 - -20037508.342789199203 + 0.000746455354) / 38.218514142588
nTLTileY: (20037508.342789199203 - -5021606.990566720255 + 0.000746455354) / 38.218514142588
nBRTileX: (19466982.332377377898 - -20037508.342789199203 + 0.000746455354) / 38.218514142588
nBRTileY: (20037508.342789199203 - -5022218.486793001182 + 0.000746455354) / 38.218514142588

TL: 1033631.9992007862 655679.9995052208 
BR: 1033647.9992007862 655695.9995052208
tiles: 17 x 17
```

Which implies I should be seeing tiles from X: 1033631  -> 1033648 Y: 655679 -> 655696 (17 x 17)

I am unsure how `dfMinX` is set, I would have assumed it was calculated from my source projwin but its slightly padded.

If I replace the `df{Min/Max}{X/Y}` with my input `projwin` I get a bounds of  

```
TL: 1033632.0000195341 655680.0000195329
BR: 1033648.0000195341 655696.0000195329
tiles 16 x 16 
```

I am also curious as to why the extra tiles are empty ?  (In my case they are full alpha)


For my testing I switched to a smaller tile so it was faster to rebuild, the full `gdal_translate` command: 

```
gdal_translate \
 -of COG \
 -co TILING_SCHEME=WebMercatorQuad.json \
 -co NUM_THREADS=ALL_CPUS \
 -co BIGTIFF=NO \
 -co ADD_ALPHA=YES \
 -co BLOCKSIZE=512 \
 -co WARP_RESAMPLING=bilinear \
 -co OVERVIEW_RESAMPLING=lanczos \
 -co COMPRESS=webp \
 -co ALIGNED_LEVELS=0 \
 -co QUALITY=90 \
 -co SPARSE_OK=YES \
 --config GDAL_DISABLE_READDIR_ON_OPEN EMPTY_DIR \
 -co ZOOM_LEVEL_STRATEGY=UPPER \
 -projwin 19466370.897442425 -5021607.010222964 19466982.363668706 -5022218.506449246 \
 -projwin_srs EPSG:3857\
 3131110001031210.3857.vrt 16-64602-40980.tiff
```

rouault:
Please attach the output of ``gdal_create -if 3131110001031210.3857.vrt 3131110001031210.3857_dummy.tif -co sparse_ok=yes`` (which will be a blank file but with the same georeferencing as your VRT) and WebMercatorQuad.json
The issue here is on gdal_translate that when -projwin is specified must compute the source region in terms of pixels, and thus needs to apply rounding.

rouault:
> I am also curious as to why the extra tiles are empty ? (In my case they are full alpha)

Are they really empty ? I would anticipate that they have their bottom-most line / right-most column with some content. The effect of -projwin 19466370.897442425 -5021607.010222964 19466982.363668706 -5022218.506449246 is to restrict the source dataset that the COG driver will see. In TILING_SCHEME mode, it outputs an extent on whole tiles, and thus pads with nodata when there are no corresponding source pixels (that are presented to it). The TILING_SCHEME mode of the COG driver is a bit at odds with the gdal_translate mechanics.

blacha:
Sure here is the tiff as a zip as github wont let me upload a tiff

[3131110001031210.3857_dummy.tif.zip](https://github.com/OSGeo/gdal/files/10941415/3131110001031210.3857_dummy.tif.zip)

and WMQ.json is https://github.com/linz/basemaps/blame/master/packages/geo/src/tms/google.ts which is the same as  http://schemas.opengis.net/tms/1.0/json/examples/WebMercatorQuad.json

blacha:
> > I am also curious as to why the extra tiles are empty ? (In my case they are full alpha)
> 
> Are they really empty ? I would anticipate that they have their bottom-most line / right-most column with some content. The effect of -projwin 19466370.897442425 -5021607.010222964 19466982.363668706 -5022218.506449246 is to restrict the source dataset that the COG driver will see. In TILING_SCHEME mode, it outputs an extent on whole tiles, and thus pads with nodata when there are no corresponding source pixels (that are presented to it). The TILING_SCHEME mode of the COG driver is a bit at odds with the gdal_translate mechanics.

Since the COG is sparse there are no tiles, the TileOffests for the first row is all `0`  and then every column has a `0` as the tile offset.

Here is the output tiff in a zip, made using GDAL  built from this branch.

[16-64602-40980.tiff.zip](https://github.com/OSGeo/gdal/files/10941501/16-64602-40980.tiff.zip)


Attached is the output tiff

rouault:
As I suspected the issue is with your input VRT whose bounds don't align properly enough with the TMS definition at zoom level 21

Origin = (19466215.648082926869392,-5021309.901335690170527)
Pixel Size = (0.074645535434742,-0.074645535434742)
>>> (19466215.648082926869392  - -20037508.3427892) / 0.074645535434742158
529217504.58080274
>>> (20037508.3427892 - -5021309.901335690170527 ) / 0.074645535434742158
335704179.7366732

I've updated the commit to further relax the tolerance so that it works on your use case, but you might get sub-pixel shifts. It would be better if the bounds of your gdalwarp -of VRT exactly aligned on pixels at zoom level 1 (so specify -te)

blacha:
Interesting, I would not have thought the input vrt would have much bearing on the output COG as I would have assumed the cog would have warped the image into it's needs.

I think that expanding the tolerance here might not be the best idea, minor pixel shifts can be tough to track down.

Would there be a better flow of nztm imagery to 3857 cog? 

My testing of going directly from the source imagery to  the cog without the intermediate warped vrt also gives me tile discrepancies.

rouault:
Merging, even if it doesn't fix the original issue

PR #7395, opened 2023-03-10
the original code determines two nodata with numpy.nan are not equal, …

cirrusasf:
…new code determine two nodata with numpy.nan are equal

## What does this PR do?
gdalcompare.py doe not correctly compare two nodata with numpy.nan in the two geotiff files. It thinks the two numpy.nan nodata values are different, so it counts this as one more diff_count. The PR solve s the problem.
## What are related issues/pull requests?
https://github.com/OSGeo/gdal/issues/7394
## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Updated Python API documentation (swig/include/python/docs/)
 - [ ] Review
 - [ ] Adjust for comments
 - [ ] All CI builds and checks have passed

## Environment

Provide environment details, if relevant:

* OS:
* Compiler:

PR #7396, opened 2023-03-10
[Backport release/3.6] the original code determines two nodata with numpy.nan are not equal, …

rouault:
Backport #7395
 **Authored by:** @cirrusasf

### DIFF
--- a/swig/python/gdal-utils/osgeo_utils/gdalcompare.py
+++ b/swig/python/gdal-utils/osgeo_utils/gdalcompare.py
@@ -32,6 +32,7 @@
 import filecmp
 import os
 import sys
+import numpy as np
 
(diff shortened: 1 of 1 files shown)


cirrusasf:
Thank you for the release. Does this mean that only gdal3.6 include the change or gdal >=3.6 include the change. I just update my gdal, and it is gdal 3.6.2, I didn't see the new gdalcompar.py code there.

rouault:
> I didn't see the new gdalcompare.py code there.

obviously :-) This will only appear in a future release, 3.7.0 presumably, as there won't probably any more 3.6.x releases (I just backported it in case there is one)

cirrusasf:
OK. Thank you for the info.

PR #7405, opened 2023-03-10
RawRasterBand::CanUseDirectIO(): fix reading of GDAL_ONE_BIG_READ that was broken in master

rouault:


rouault:
included in https://github.com/OSGeo/gdal/pull/7406

PR #7409, opened 2023-03-11
Add OGRGeometry::UnaryUnion() / OGR_G_UnaryUnion() and use it in ogr2ogr

rouault:


ogr/ogrgeometry.cpp
```diff
@@ -4703,6 +4711,101 @@ OGRGeometryH OGR_G_UnionCascaded(OGRGeometryH hThis)
         OGRGeometry::FromHandle(hThis)->UnionCascaded());
 }
 
+/************************************************************************/
+/*                               UnaryUnion()                           */
+/************************************************************************/
+
+/**
+ * \brief Returns the union of all components of a single geometry.
```
jratike80, line 4719: I do not quite understand what this comment means.

rouault: I've just stolen it from upstream GEOS function...  : https://libgeos.org/doxygen/3.11/geos__c_8h.html#ada5b7b39a81b722b97371736ff0b8d29

Perhaps adding a link to https://postgis.net/docs/ST_UnaryUnion.html if it makes things clearer ?

apps/ogr2ogr_lib.cpp
```diff
@@ -586,54 +586,42 @@ static std::unique_ptr<OGRGeometry> LoadGeometry(const std::string &osDS,
...
-            else if (eType == wkbMultiPolygon)
-            {
-                OGRMultiPolygon *poSrcMP = poSrcGeom->toMultiPolygon();
-                const int nGeomCount = poSrcMP->getNumGeometries();
-
-                for (int iGeom = 0; iGeom < nGeomCount; iGeom++)
-                {
-                    poMP->addGeometry(poSrcMP->getGeometryRef(iGeom));
-                }
+                auto poValid =
```
dbaston, line 606: Checking validity / attempting to autocorrect errors is a new behavior here, right? If so, it seems a good idea to add a test case.

apps/ogr2ogr_lib.cpp
```diff
@@ -586,54 +586,42 @@ static std::unique_ptr<OGRGeometry> LoadGeometry(const std::string &osDS,
...
+                    std::unique_ptr<OGRGeometry>(poSrcGeom->MakeValid());
+                if (poValid)
+                    oGC.addGeometryDirectly(poValid.release());
             }
             else
             {
-                CPLError(CE_Failure, CPLE_AppDefined,
```
dbaston, line 624: If I understand right, `LoadGeometry` is used only to load a clipping geometry. Is it important to keep the error here if the clip geometry is not a polygon? Or is this intentionally allowing clipping using lines/points?

ogr/ogrgeometry.cpp
rouault: done

apps/ogr2ogr_lib.cpp
rouault: Test added

apps/ogr2ogr_lib.cpp
rouault: Test on getCoordinateDimension() == 2 added back

PR #7436, opened 2023-03-12
autotest: add a pytest.mark.require_curl()

rouault:


dbaston:
It would be nice to land on a consistent convention of `pytest.mark.require_xyz` or `gdaltest.require_xyz`.

Some advantages I see to `gdaltest.require_xyz`:
- the implementations are self-contained, not mixed in a single function
- the implementations are discoverable by symbol navigation (IDE/ctags), rather than text search (or knowing to check `conftest.py`)
- it doesn't require you to understand how `pytest` processes markers

Some advantages to `pytest.mark.require_xyz`:
- it seems to be the way `pytest` expects you to do this
- it avoids writing nested decorators, which can be confusing
- you can use `pytest -m` to only run tests with specified markers

rouault:
> It would be nice to land on a consistent convention of pytest.mark.require_xyz or gdaltest.require_xyz.

yes I agree we should probably standardize. The main drawback of the decorator approach is that it cannot be used to exclude a whole test file with "pytestmark = pytest.mark.require_XXXXX()", which was needed here

dbaston:
> The main drawback of the decorator approach is that it cannot be used to exclude a whole test file with "pytestmark = pytest.mark.require_XXXXX()",

Hmm. I would have expected the decorator to work by applying it to a module-scope fixture, but it it appears that `pytest` ignores marks applied to fixtures. I guess using `pytest.mark.xyz` is the way to go, then.

PR #7453, opened 2023-03-15
GDAL as a SQLite3 loadable extension: disable if loaded from sqlite3 binary on Unix

rouault:
Loading libgdal.so from the sqlite3 binary on Linux/Unix often leads to a crash, due to duplication of symbols in the static copy of libsqlite3 in the sqlite3 binary and the shared version of libsqlite3.
Hence we disable loading of GDAL.
If you want to override this check, set the
OGR_SQLITE_ALLOW_LOAD_FROM_SQLITE3_BINARY environment variable to YES.


rouault:
superseded per https://github.com/OSGeo/gdal/pull/7457

PR #7457, opened 2023-03-15
GDAL as a SQLite3 loadable extension: avoid crash on Linux

rouault:


elpaso (approved):
I can only imagine how hard it was to debug this.

elpaso (approved):

PR #7462, opened 2023-03-15
GRIB: g2clib: allow negative longitudes in grid templates Lon/Lat, Rotated, Stretched, Stretched & Rotated, LAEA (fixes #7456) and Mercator

### LINKED ISSUE #7456: GDAL can't read projection of MetOffice UKV 2km model gribs, Lambert Azimuthal Equal Area grid
Trying to work with grib2 files from MetOffice UKV model with gdal, files use Lambert Azimuthal Equal Area grid 

gdal shows this error:

```
$ gdalinfo ukv2km.grib2 
Warning 1: Unable to perform coordinate transformations, so the correct projected geotransform could not be deduced from the lat/long control points.  Defaulting to ungeoreferenced.
Driver: GRIB/GRIdded Binary (.grb, .grb2)
Files: ukv2km.grib2
Size is 1042, 970
Origin = (-0.500000000000000,-0.500000000000000)
Pixel Size = (1.000000000000000,1.000000000000000)
Corner Coordinates:
Upper Left  (  -0.5000000,  -0.5000000) 
Lower Left  (      -0.500,     969.500) 
Upper Right (    1041.500,      -0.500) 
Lower Right (    1041.500,     969.500) 
Center      (     520.500,     484.500) 
Band 1 Block=1042x1 Type=Float64, ColorInterp=Undefined
  Description = 0[-] SFC="Ground or water surface"
  Metadata:
...
```

wgrib2 reports the grid like this, seems OK:

```
$ wgrib2 -grid uk2km.grib2 
1:0:grid_template=140:winds(N/S):
        Lambert Azimuthal Equal Area grid: (1042 x 970) input WE:SN output WE:SN res 48
        Lat1 44.517153 Lon1 -17.117129 Cen Lon -2.500000 Std Par 54.900000
        Dx 2000.000000 m Dy 2000.000000 m mode 48

```

## Operating system

Debian GNU/Linux 11 (bullseye)

GDAL 3.2.2 from debian stable packages

example grib file here: https://www.windguru.cz/down/ukv2km.grib2

is this gdal problem?

rouault:
It is not obvious that negative longitudes are allowed in the GRIB2 standard. https://old.wmo.int/extranet/pages/prog/www/DPS/FM92-GRIB2-11-2003.pdf mentions:
"""92.1.8 The longitude values shall be limited to the range 0 to 360 degrees inclusive. Orientation shall be east longitude positive, with only positive values being used"""

That said wgrib2 in grb2.h reads the center longitude as a signed integer:
```
#define GDS_Lambert_Az_Cen_Lon(gds)     (int4(gds+50) * 0.000001)
```

https://apps.ecmwf.int/codes/grib/format/grib2/templates/3/140 (LAEA) doesn't specify the signedness, but other templates such as https://apps.ecmwf.int/codes/grib/format/grib2/templates/3/10 (Mercator) mentions signed for longitudes...

There is no harm I can think of in allowing more liberarly signed values for longitudes, latitudes, given the range on how they are encoded, there is no risk with unsigned values that would overflow the signed range.

More templates could be edited to allow signed longitudes, but this is tedious...


rouault:
CC @edwardhartnett in case you are interested in such fixes in g2clib (seeing there's now https://github.com/NOAA-EMC/NCEPLIBS-g2c/blob/develop/src/gridtemplates.c)

PR #7471, opened 2023-03-17
Massive linting in RawRasterBand based drivers

rouault:
Initially to fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=57124 in ERS driver, but turned to be a systematic issue among all those RawRasterBand based drivers.
- GDALDataset:: add a SetBand() method that takes a std::unique_ptr<GDALRasterBand>&&
- RawRasterBand: add IsValid() and Create() methods that return a unique_ptr<RawRasterBand>
- Use std::unique_ptr<>'ification for dataset and band objects in those drivers, use RawRasterBand:IsValid()/Ceate()

### DIFF
--- a/gcore/gdal_priv.h
+++ b/gcore/gdal_priv.h
@@ -393,7 +393,7 @@ class CPL_DLL GDALDataset : public GDALMajorObject
 
     void RasterInitialize(int, int);
     void SetBand(int, GDALRasterBand *);
-    void SetBand(int nNewBand, std::unique_ptr<GDALRasterBand> &&poBand);
+    void SetBand(int nNewBand, std::unique_ptr<GDALRasterBand> poBand);
 
     GDALDefaultOverviews oOvManager{};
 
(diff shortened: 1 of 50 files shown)


gcore/gdal_priv.h
```diff
@@ -393,6 +393,7 @@ class CPL_DLL GDALDataset : public GDALMajorObject
 
     void RasterInitialize(int, int);
     void SetBand(int, GDALRasterBand *);
+    void SetBand(int nNewBand, std::unique_ptr<GDALRasterBand> &&poBand);
```
dbaston, line 396: FWIW, the ISO guideline would be to take the `unique_ptr` by value instead of rvalue reference: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f18-for-will-move-from-parameters-pass-by-x-and-stdmove-the-parameter

I have not done this in GEOS because I was not aware of the guideline when I started converting the API to unique_ptr, and I think it's important to keep it consistent. That said, I recall fixing one leak where a function took a unique_ptr by rvalue reference and then failed to move from it.

rouault: interesting. Done

PR #7473, opened 2023-03-17
Python bindings: add workaround to avoid issue with Conda-Forge osx arm64 crosscompiler builds (fixes conda-forge/gdal-feedstock#736)

rouault:


rouault:
Likely unneeded given https://github.com/conda-forge/cross-python-feedstock/pull/71

PR #7485, opened 2023-03-21
PGDump: add GEOM_COLUMN_POSITION layer creation option and allow empty FID= (fixes #7482)

### LINKED ISSUE #7482: pgdump driver puts geometry column first in table structure

Generating a postGIS dump using the ogr pgDump driver generates geometry field as the first field. 
The most common practice around seem to put geometry fieds at the end of a table's structure

## Steps  to reproduce the problem.

- export a ogr PGdump from a postgis database with geometry columns

ogr2ogr -progress --config PG_USE_COPY YES -f PGDump /tmp/dump_test.sql 'PG:service=myservice' -sql  "select * from aschema.atable limit 10 "

- audit the generated DDL ,  the

```sql
SET standard_conforming_strings = OFF;
DROP TABLE IF EXISTS "public"."sql_statement" CASCADE;
BEGIN;
CREATE TABLE "public"."sql_statement" ( "ogc_fid" SERIAL, CONSTRAINT "sql_statement_pk" PRIMARY KEY ("ogc_fid") );
SELECT AddGeometryColumn('public','sql_statement','geom',2154,'GEOMETRY',2);
CREATE INDEX "sql_statement_geom_geom_idx" ON "public"."sql_statement" USING GIST ("geom");
ALTER TABLE "public"."sql_statement" ADD COLUMN "id" VARCHAR NOT NULL;
ALTER TABLE "public"."sql_statement" ADD COLUMN "insee_arr" VARCHAR(1);
ALTER TABLE "public"."sql_statement" ADD COLUMN "insee_dep" VARCHAR(3);
ALTER TABLE "public"."sql_statement" ADD COLUMN "insee_reg" VARCHAR(2);
COPY "public"."sql_statement" ("geom", "id", "insee_arr", "insee_dep", "insee_reg") FROM STDIN;
```

GDAL 3.4.3

rouault:
-  ***GEOM_COLUMN_POSITION***\ =IMMEDIATE/END (starting with GDAL 3.7). If set to IMMEDIATE (the default), geometry columns are created in the table structure as soon as OGR is instructed to create them (which typically means that the default geometry column is created just after the FID column and before the non-spatial columns). If set to END, the geometry columns are created in the table structure after non-spatial columns.

-  **FID**: Name of the FID column to create. Defaults to 'ogc_fid'. Starting with GDAL 3.7, if set explicitly to the empty string, disables the creation of a FID column.

And while we are it, creation of spatial indexes is deferred after feature insertions.


ogr/ogrsf_frmts/pgdump/ogrpgdumplayer.cpp
```diff
@@ -391,52 +442,60 @@ OGRErr OGRPGDumpLayer::CreateFeatureViaCopy(OGRFeature *poFeature)
...
+
+                poGeometry->closeRings();
+                poGeometry->set3D(poGFldDefn->GeometryTypeFlags &
+                                  OGRGeometry::OGR_G_3D);
+                poGeometry->setMeasured(poGFldDefn->GeometryTypeFlags &
+                                        OGRGeometry::OGR_G_MEASURED);
+
+                // CheckGeomTypeCompatibility(poGeometry);
+
+                /*if (bHasWkb)
```
elpaso, line 468: leftover?

elpaso (approved):
LGTM, I couldn't check if all the occurrences of table/schema name are escaped, I assume this is done in client code.

ogr/ogrsf_frmts/pgdump/ogrpgdumplayer.cpp
```diff
@@ -391,52 +442,60 @@ OGRErr OGRPGDumpLayer::CreateFeatureViaCopy(OGRFeature *poFeature)
...
+                    pszGeom = GeometryToBYTEA( poGeometry );
+                else*/
```
rouault, line 470:
```suggestion
```

rouault:
>  I couldn't check if all the occurrences of table/schema name are escaped, I assume this is done in client code.

From a quick testing, there are some bugs in that era, but they predate this PR. Ticketed as https://github.com/OSGeo/gdal/issues/7497

PR #7492, opened 2023-03-21
PG: Add SKIP_VIEWS open option to replace PG_SKIP_VIEWS config option

dbaston:
## What does this PR do?

Adds a `SKIP_VIEWS` open option to the Postgres driver to replace the undocumented `PG_SKIP_VIEWS` config option

## Tasklist

 - [X] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/ogr/ogrsf_frmts/pg/ogr_pg.h
+++ b/ogr/ogrsf_frmts/pg/ogr_pg.h
@@ -600,6 +600,8 @@ class OGRPGDataSource final : public OGRDataSource
     int bHasLoadTables = false;
     CPLString osActiveSchema{};
     int bListAllTables = false;
+    bool m_bSkipViews = false;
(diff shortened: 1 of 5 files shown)


ogr/ogrsf_frmts/pg/ogr_pg.h
```diff
@@ -600,6 +600,8 @@ class OGRPGDataSource final : public OGRDataSource
     int bHasLoadTables = false;
     CPLString osActiveSchema{};
     int bListAllTables = false;
+    bool m_bSkipViews = false;
```
dbaston, line 603: It appeared to me that newer flags use `bool` with an `m_` prefix so this is what I did. If there's some reason `int` should be used, let me know.

PR #7494, opened 2023-03-21
SQLite3: use external sqlite3ext.h instead of custom ogrsqlite3ext.h (fixes #7444)

### LINKED ISSUE #7444: Remove ogrsqlite3ext.h
While discussing on the SQLite forums about compilation options, it appears that they found that the GDAL usage of SQLite was ill-conceived.

I drop it here in case it is something that could be inserted in the GDAL dev roadmap. (But I don't need it, it is just for information)

[https://sqlite.org/forum/forumpost/0138fd5fc9](https://sqlite.org/forum/forumpost/0138fd5fc9)

> The GDAL code basically duplicates the sqlite_api structure and "links up" names to its own version of the structure which includes the sqlite3_mutex* APIs and does not know how to handle the names not existing as exports even though they are never used.
> 
> Like I said, it is a really ill-conceived design.
> 
> The proper way to build a dynamic call trampoline is to loop though the required entry-points by name, and if the name exists in the DCSS then store the address in the trampoline. Then you let your trampoline code handle what to do with missing entries (or just explode when loading the trampoline if a required entry is missing).
> 
> The sqlite3 os modules are an example of how to load and use a trampoline properly.
>

rouault:


ogr/ogrsf_frmts/sqlite/ogrsqliteexecutesql.cpp
```diff
@@ -503,6 +456,83 @@ static void OGR2SQLITEGetPotentialLayerNames(
...
 }
 
+/************************************************************************/
+/*                   OGRSQLiteGetReferencedLayers()                     */
+/************************************************************************/
+
+std::set<LayerDesc> OGRSQLiteGetReferencedLayers(const char *pszStatement)
+{
+    /* -------------------------------------------------------------------- */
+    /*      Analysze the statement to determine which tables will be used.  */
```
elpaso, line 466:
```suggestion
    /*      Analyze the statement to determine which tables will be used.  */
```

ogr/ogrsf_frmts/sqlite/ogrsqlitevirtualogr.cpp
```diff
@@ -43,19 +43,48 @@
...
+{
+    const char *pszGeomColumn = poLayer->GetGeometryColumn();
+    if (pszGeomColumn != nullptr && !EQUAL(pszGeomColumn, ""))
+    {
+        if (poLayer->GetLayerDefn()->GetFieldIndex(pszGeomColumn) < 0)
+            return pszGeomColumn;
+    }
+
+    CPLString osGeomCol("GEOMETRY");
+    int bTry = 2;
```
elpaso, line 61:
```suggestion
    int iTry = 2;
```

isn't `b` prefix for booleans?

rouault: fixed

ogr/ogrsf_frmts/sqlite/ogrsqliteexecutesql.cpp
rouault: fixed

Issue #7379, opened 2023-03-08
Containers for GDAL build/test environments


dbaston:
I've been writing up some notes on reproducing GDAL CI failures locally and had some thoughts on simplifying the process.

The GDAL CI jobs I've looked into either:
1. start with a "stock" container such as `alpine:edge`, `fedora:rawhide`, or `ubuntu:18.04` and install build dependencies during each CI run
2. create a container with the complete build environment and save it to GitLab CI's cache for use in subsequent CI runs (the Ubuntu 20.04 run is an example)

A modification of the second method would make it a bit easier to reproduce CI failures locally, while also providing good CI performance:

3. create a container with the complete build environment and save it to a container registry for use in CI runs

My experience with other projects using this approach has been positive. The only way I've set it up is to create the build environment containers via a separate process from the CI runs themselves, such as a job that runs on a specified frequency. (This covers the case where you want to track the dev version of some dependencies.) An alternative that is closer to the current GDAL practice would be to have the CI run itself update the container if the Dockerfile that defines it has changed. This _should_ be possible (e.g., via https://github.com/moby/moby/issues/39003#issuecomment-571470958) although I have not yet made it work.

Is there any interest in restructuring the GDAL CI runs along these lines?

Maybe it is also worth thinking about harmonizing the containers used for CI (`.github/workflows/**`) and those used for distribution (`docker/**`).

rouault:
> The only way I've set it up is to create the build environment containers via a separate process from the CI runs themselves, such as a job that runs on a specified frequency.

I believe that's what QGIS does with https://github.com/qgis/QGIS/blob/master/.github/workflows/build-docker.yml . Perhaps @elpaso can confirm ?

elpaso:
> > The only way I've set it up is to create the build environment containers via a separate process from the CI runs themselves, such as a job that runs on a specified frequency.
> 
> I believe that's what QGIS does with https://github.com/qgis/QGIS/blob/master/.github/workflows/build-docker.yml . Perhaps @elpaso can confirm ?

Yeah, I believe so, I didn't set it up myself so I am not familiar with the details. I think there is also a ccache permanent storage across builds to speed compilation up (in QGIS it could take hours, GDAL is surely faster).

dbaston:
@rouault Do you want to assign this one to me?

rouault:
> Do you want to assign this one to me?

done. I've sent you a GH invitation with "triage" role, that should hopefully be good enough for you to be able to assign tickets etc

[...]

PR #7501, opened 2023-03-23
autotest: Combine Linux builds into single workflow

dbaston:
## What does this PR do?

This PR restructures the Linux CI runs hosted by GitHub Actions. The restructuring makes
two major changes:

1. A single parameterized build is used. This avoids the need for common logic
(container build, ccache management, etc.) to be copied between configurations.
This also forces some consistency about which tasks are accomplished in which
build steps (i.e., a "build" script builds but does not test, etc.) Once a
contributor is familiar with the single configuration, they should be able to
easily reproduce a CI failure in any of the other configurations.

2. The complete build environment for each configuration is stored in a Docker
image that is published to a container registry. The published containers serve
as a caching layer to avoid repeating setup steps between CI runs, and can be
pulled to a contributor's local machine to easily reproduce a CI environment.
Because the containers are only rebuilt if their associated Dockerfile changes,
CI runtime is improved somewhat and spurious failures associated with the setup
steps can be avoided.

Unresolved issues:

This does not yet test a static library build nor does it perform post-install testing, so
I have not removed the Linux build from `cmake_builds.yml`.

Configurations can share a container (e.g., "ASAN" and "Ubuntu 20.04") but the
directory structure isn't clear about whether a subdirectory of .github/workflows
refers to a container or a configuration. Probably the files should either be
reorganized, or the configurations should stop sharing containers.

#7379

## Tasklist

 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:


dbaston:
Linux CI results can be seen at https://github.com/dbaston/gdal/actions/runs/4496159426/jobs/7910557941. It appears that some adjustments would be needed to make this run under the `OSGeo` user.

rouault:
@dbaston is it still a WIP ?

dbaston:
I just added a note to the docs and squashed it to a single commit. I think it's ready to go.

PR #7513, opened 2023-03-24
gdal2tiles: use logging module instead of print() to emit verbose output (fixes #4894)

### LINKED ISSUE #4894: update prints to use logging package
I would like to use the script as part of my code.
It would be easier to manage output if, instead of prints you will use the logging package.
https://github.com/OSGeo/gdal/blob/35c07b18316b4b6d238f6d60b82c31e25662ad27/swig/python/gdal-utils/osgeo_utils/gdal2tiles.py#L3209

rouault:


elpaso (approved):
LGTM.

I noticed there is a mix of `f`, `%` and `+` to format/concat strings, it would be nice to homogenize a bit.

PR #7517, opened 2023-03-25
C# Bindings: Add missing GDAL wrappers

### LINKED ISSUE #813: gdal_csharp.dll have

For example: I expected to be able to open this raster file (with a link to
the raster file, or it as an attachment) and it returns an error message
instead.

For example: "gdalinfo myfile"

## Operating system

For example: Ubuntu 16.04 64 bit

For example: the 2.2.3 version from ubuntugis-unstable PPA

### LINKED ISSUE #1436: wrapper_GDALBuildVRT_objects in gdal_csharp.dll should take array of Dataset

I expected to be able to call Gdal.wrapper_GDALBuildVRT_objects(...) and supply an array of Datasets as the third parameter. This would bring this method in line with wrapper_GDALBuildVRT_names(...) which takes an array of filenames. 

The actual behavior is the third parameter is a SWIGTYPE_p_p_GDALDatasetShadow, which corresponds to nothing in the C# bindings.

Create any C# Gdal project, open a raster (from physical file) that contains subdatasets, create an in-memory Dataset from each subdataset, collect those Datasets in an in-memory array, attempt to call BuildVRT_objects with that Dataset array.

The purpose is to be able to get a VRT from the in-memory Dataset array without having to write intermediary files for each of the collected Datasets (and then use those filenames with the _names version of this build VRT method).

## Operating system

Windows 10, 64 bit.

GDAL, GDAL.Native, GDAL.Plugins v 2.3.3 
https://www.nuget.org/packages/gdal/2.3.3

### LINKED ISSUE #7152: SWIGTYPE_p_p_GDALDatasetShadow is missing from version 3.5.3
Hello,

I am migrating from GDAL 
<package id="GDAL" version="2.4.4" targetFramework="net452" />

to 
 <package id="GDAL" version="3.5.3" targetFramework="net48" />
 
 but the class SWIGTYPE_p_p_GDALDatasetShadow no longer exists.  
 I tried SWIGTYPE_p_p_GDALRasterBandShadow, but it is not an intPtr and cannot be used with Gdal.wrapper_GDALBuildVRT_objects
 
 Please provide a pointer to the Breaking Change documentation?

MaxRev-Dev:
Initially, there was a defined wrapper only for gdal_wrap function. Other functions were not implemented, and a custom-defined method joggling with pointers is the only option to use a utility. Some older types, like `SWIGTYPE_p_p_GDALDatasetShadow` are not required with new GDAL versions and can be replaced by high-level C# methods.

This will bring more clarity into using specified functions in .NET environment. 

Resolves #7152, resolves #813, resolves #1436
Related to https://github.com/MaxRev-Dev/gdal.netcore/issues/97

## Tasklist

 - [ ] Check wrappers usability
 - [ ] Add documentation
 - [ ] Review

### DIFF
--- a/swig/include/csharp/gdal_csharp.i
+++ b/swig/include/csharp/gdal_csharp.i
@@ -351,4 +351,76 @@ public CPLErr SetGCPs(GCP[] pGCPs, string pszGCPProjection) {
       return retval;
    }
 
+   public static Dataset BuildVRT(string dest, string[] poObjects, GDALBuildVRTOptions buidVrtAppOptions, $module.GDALProgressFuncDelegate callback, string callback_data) {
+      Dataset retval = null;
+      if (poObjects.Length <= 0)
+        throw new ArgumentException("poObjects size is small (BuildVRT)");
+
+      int intPtrSize = Marshal.SizeOf(typeof(IntPtr));
+      IntPtr nativeArray = Marshal.AllocHGlobal(poObjects.Length * intPtrSize);
+      try {
+          for (int i=0; i < poObjects.Length; i++)
+            Marshal.WriteIntPtr(nativeArray, i * intPtrSize, Dataset.getCPtr(poObjects[i]).Handle);
+
+          retval  = wrapper_GDALBuildVRT_names(dstDS, poObjects, nativeArray, buidVrtAppOptions, callback, callback_data);
+      } finally {
+          Marshal.FreeHGlobal(nativeArray);
+      }
+      return retval;
+   }
+
+   public static Dataset BuildVRT(string dest, Dataset[] poObjects, GDALBuildVRTOptions buidVrtAppOptions, $module.GDALProgressFuncDelegate callback, string callback_data) {
+      Dataset retval = null;
+      if (poObjects.Length <= 0)
+        throw new ArgumentException("poObjects size is small (BuildVRT)");
+
+      int intPtrSize = Marshal.SizeOf(typeof(IntPtr));
+      IntPtr nativeArray = Marshal.AllocHGlobal(poObjects.Length * intPtrSize);
+      try {
+          for (int i=0; i < poObjects.Length; i++)
+            Marshal.WriteIntPtr(nativeArray, i * intPtrSize, Dataset.getCPtr(poObjects[i]).Handle);
+
+          retval  = wrapper_GDALBuildVRT_objects(dest, poObjects.Length, nativeArray, buidVrtAppOptions, callback, callback_data);
+      } finally {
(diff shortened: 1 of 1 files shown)


MaxRev-Dev:
Can we double-check the bindings? Also, I'm not sure about the proper implementation of GDALMultiDimTranslate?
cc @rouault @szekerest

rouault:
@MaxRev-Dev Did you manage to build this locally ? This causes build failures on our CI
```
 /home/runner/work/gdal/gdal/superbuild/build/gdal/swig/csharp/gdal/Gdal.cs(169,81): error CS0246: The type or namespace name 'MultiDimTranslateOptions' could not be found (are you missing a using directive or an assembly reference?) [/home/runner/work/gdal/gdal/superbuild/build/gdal/swig/csharp/gdal/gdal_csharp.csproj]
/home/runner/work/gdal/gdal/superbuild/build/gdal/swig/csharp/gdal/Gdal.cs(187,82): error CS0246: The type or namespace name 'MultiDimTranslateOptions' could not be found (are you missing a using directive or an assembly reference?) [/home/runner/work/gdal/gdal/superbuild/build/gdal/swig/csharp/gdal/gdal_csharp.csproj]
```

MaxRev-Dev:
@rouault Sorry for the late reply. I have built it locally and removed other broken methods. Now it should be ready for review.

swig/include/csharp/gdal_csharp.i
```diff
@@ -351,4 +351,44 @@ public CPLErr SetGCPs(GCP[] pGCPs, string pszGCPProjection) {
...
+            Marshal.WriteIntPtr(nativeArray, i * intPtrSize, Dataset.getCPtr(poObjects[i]).Handle);
+
+          retval  = wrapper_GDALBuildVRT_objects(dest, poObjects.Length, nativeArray, buidVrtAppOptions, callback, callback_data);
+      } finally {
+          Marshal.FreeHGlobal(nativeArray);
+      }
+      return retval;
+   }
+
+   public static Dataset GDALMultiDimTranslate(string dest, Dataset[] poObjects, GDALMultiDimTranslateOptions multiDimAppOptions, $module.GDALProgressFuncDelegate callback, string callback_data) {
```
rouault, line 376: I guess that should rather be
```suggestion
   public static Dataset MultiDimTranslate(string dest, Dataset[] poObjects, GDALMultiDimTranslateOptions multiDimAppOptions, $module.GDALProgressFuncDelegate callback, string callback_data) {
```

MaxRev-Dev: I agree so that the method call won't replicate the name of the class

PR #7518, opened 2023-03-26
Updated doc for LibKML (missing LIBKML_MINIZIP_LIBRARY and LIBKML_URI…

sea-kg:
Updated development doc for LibKML 

fix missing LIBKML_MINIZIP_LIBRARY and LIBKML_URIPARSER_LIBRARY

### DIFF
--- a/doc/source/development/building_from_source.rst
+++ b/doc/source/development/building_from_source.rst
@@ -1039,6 +1039,13 @@ It can be detected with pkg-config.
 
     Control whether to use LibKML. Defaults to ON when LibKML is found.
 
+.. option:: LIBKML_MINIZIP_LIBRARY
+
+    Path to a shared or static library file for ``minizip``
+
+.. option:: LIBKML_URIPARSER_LIBRARY
+
+    Path to a shared or static library file for ``urlparser``
 
 LibLZMA
 *******


doc/source/development/building_from_source.rst
```diff
@@ -1039,6 +1039,13 @@ It can be detected with pkg-config.
 
     Control whether to use LibKML. Defaults to ON when LibKML is found.
 
+.. option:: LIBKML_MINIZIP_LIBRARY
+
+    Path to a shared or static library file for ``minizip``
+
+.. option:: LIBKML_URIPARSER_LIBRARY
+
+    Path to a shared or static library file for ``urlparser``
```
jratike80, line 1048: Which one, "uriparser" or "urlparser", is correct?

sea-kg: Yep! uriparser is correct - it's very old library.

Also you can look in here:
https://github.com/OSGeo/gdal/blob/master/cmake/modules/packages/FindLibKML.cmake#L70

jratike80: Would it be better to call it as uriparser also in the comment, even it is hard to see that current "library file for ``urlparser``" could make any issues?

sea-kg: Oh... you are ritght! my mistake

PR #7526, opened 2023-03-27
Rasterize all touched: tighten(decrease) the tolerance to consider that edge of geometries match pixel obundaries (fixes #7523)

### LINKED ISSUE #7523: New rasterize behaviour is missing pixels that should be included

There was a recent change to GDAL rasterize that meant pixels with a small overlap were no longer included. For example, we have a 5m tapped raster and when you rasterize a polygon that slightly overlaps the pixels on the left hand edge (vertices are 475439.99661332502728328m in the x direction vs pixel boundary of 475440m)

image

Previously it was rasterized to this:

```
array([[1, 1, 0],
       [1, 1, 0],
       [1, 1, 0],
       [1, 1, 1],
       [1, 1, 1]], dtype=uint8)
```

but now you get this:

```
array([[1, 1, 0],
       [0, 1, 0],
       [0, 1, 0],
       [0, 1, 1],
       [1, 1, 1]], dtype=uint8)
```

If the top left and bottom left pixels are included, the ones in between should also be included because they overlap by the same amount.

Is this a side effect of the fix for #6414?

Rasterize a polygon where two vertices slightly overlap a row/column of pixels (by a proportion less than 0.01)

## Operating system

Ubuntu 22.04 64 bit

This was working at 3.5.1 but broken at 3.6.2.

rouault:
CC @SebastienPeillet . Cf https://github.com/OSGeo/gdal/issues/7523. This PR seems to preserve the behaviour you modified per https://github.com/OSGeo/gdal/pull/6415, while fixing #7523. It is quite hard to find the appropriate epsilon to distinguish pure numerical inaccuracy from a geometry slightly touching a pixel edge...

### DIFF
--- a/alg/llrasterize.cpp
+++ b/alg/llrasterize.cpp
@@ -376,6 +376,12 @@ void GDALdllImageLineAllTouched(int nRasterXSize, int nRasterYSize,
                                 bool bIntersectOnly)
 
 {
+    // This is an epsilon to detect geometries that are aligned with pixel
+    // coordinates. Hard to find the right value. We put it to that value
+    // to satisfy the scenarios of https://github.com/OSGeo/gdal/issues/7523
+    // and https://github.com/OSGeo/gdal/issues/6414
+    constexpr double EPSILON_INTERSECT_ONLY = 1e-4;
+
     if (!nPartCount)
         return;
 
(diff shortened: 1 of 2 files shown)


SebastienPeillet:
Hello @rouault,

I builded and tested your branch on my application and I confirm it still work as expected. Thanks you for notifying !

PR #7528, opened 2023-03-28
OSRImportFromEPSG(): emit warning message about deprecated CRS substitution (fixes #7524)

### LINKED ISSUE #7524: Replacement of deprecated CRSes should be disabled by default, or at least noisy
As pointed out in #5106, GDAL will replace deprecated CRSes. This can be disabled with `OSR_USE_NON_DEPRECATED=NO`, but the fact that it happens by default, and without warning, is dangerous for scientific code that is intended to be reproducible. The user asks for one projection, silently gets something else, and can get pretty perplexed trying to find out what changed, if they even notice. (Ask me how I know.) I don't think #5108 is sufficient because it seems to only affect `gdalsrsinfo`.

I think the right thing to do here would be to always use the requested CRS, and offer a separate feature to upgrade a CRS to its non-deprecated equivalent if one exists.

rouault:
rather than just in gdalsrsinfo

### DIFF
--- a/ogr/ogrspatialreference.cpp
+++ b/ogr/ogrspatialreference.cpp
@@ -11176,8 +11176,10 @@ OGRErr OGRSpatialReference::importFromEPSGA(int nCode)
 {
     Clear();
 
+    const char *pszUseNonDeprecated =
+        CPLGetConfigOption("OSR_USE_NON_DEPRECATED", nullptr);
     const bool bUseNonDeprecated =
-        CPLTestBool(CPLGetConfigOption("OSR_USE_NON_DEPRECATED", "YES"));
+        CPLTestBool(pszUseNonDeprecated ? pszUseNonDeprecated : "YES");
     const bool bAddTOWGS84 = CPLTestBool(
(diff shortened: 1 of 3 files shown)


ogr/ogrspatialreference.cpp
```diff
@@ -11214,6 +11216,22 @@ OGRErr OGRSpatialReference::importFromEPSGA(int nCode)
...
+                    if (pszUseNonDeprecated == nullptr)
+                    {
+                        const char *pszNewAuth =
+                            proj_get_id_auth_name(nonDeprecated, 0);
+                        const char *pszNewCode =
+                            proj_get_id_code(nonDeprecated, 0);
+                        CPLError(CE_Warning, CPLE_AppDefined,
+                                 "CRS EPSG:%d is deprecated. "
+                                 "Its non-deprecated replacement %s:%s "
+                                 "will be used instead.\n"
```
sgillies, line 11228: @rouault what would you think about removing the newline here and rely on applications to format the message?

rouault: done

PR #7535, opened 2023-03-30
Expose VRT option allocation and free in public header

Ryanf55:
## What does this PR do?

This PR exposes two function calls for `gdalbuildvrt` that I would like to use in my application. According to the documentation for `struct GDALBuildVRTOptions`, one must use `GDALBuildVRTOptionsNew` and `GDALBuildVRTOptionsFree` to perform allocation and freeing. Likewise, in `GDALBuildVRTOptionsNew`, the options for that call must be allocated with `GDALBuildVRTOptionsForBinaryNew` , however that function is only part of `gdalbuildvrt_lib.cpp` and not a header file. Then, I moved the implementation to the installed library rather than the executable, that way I can link to those functions in my application. 

The end goal of this work (and likely some follow up PR's), is to treat the VRT building as a set of library calls that are used in another application. GDAL building VRT's is assuming there is a CLI, but that is not the case in my application. If there is a better approach to slowly refactoring the `gdalbuildvrt` code that would be more in line with that use case, please share. 
I know that the applications in `apps` directories are supposed to be examples, however there is a significant amount of functionality here that can be re-used in an another application if it's treated as a library. By doing this, our application benefits from improvements in the VRT builder. I believe this is in the spirit of RFC 59. 

N/A

## Tasklist

 - [ ] Approve approach to add more memory allocation to public header
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS: Ubuntu 22.04
* Compiler: G++11

### DIFF
--- a/apps/gdal_utils.h
+++ b/apps/gdal_utils.h
@@ -221,6 +221,11 @@ GDALBuildVRTOptions CPL_DLL *
 GDALBuildVRTOptionsNew(char **papszArgv,
                        GDALBuildVRTOptionsForBinary *psOptionsForBinary);
 
+GDALBuildVRTOptionsForBinary CPL_DLL *GDALBuildVRTOptionsForBinaryNew(void);
+
+void CPL_DLL GDALBuildVRTOptionsForBinaryFree(
+    GDALBuildVRTOptionsForBinary *psOptionsForBinary);
+
 void CPL_DLL GDALBuildVRTOptionsFree(GDALBuildVRTOptions *psOptions);
 
 void CPL_DLL GDALBuildVRTOptionsSetProgress(GDALBuildVRTOptions *psOptions,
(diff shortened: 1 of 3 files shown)


rouault:
GDALBuildVRTOptionsForBinaryNew() and GDALBuildVRTOptionsForBinaryFree() are not necessary in public headers. You can pass a NULL pointer to GDALBuildVRTOptionsNew() as documented:
```
 * @param psOptionsForBinary (output) may be NULL (and should generally be
 * NULL), otherwise (gdalbuildvrt_bin.cpp use case) must be allocated with
 *                           GDALBuildVRTOptionsForBinaryNew() prior to this
 * function. Will be filled with potentially present filename, open options,...
```

PR #7538, opened 2023-03-30
OSM: add a tags_format=json osmconf.ini setting and TAGS_FORMAT=JSON …

### LINKED ISSUE #7533: OSM driver: prefer JSON or other standard to HSTORE

I am a newbie to databases and GDAL. I set up a workflow around a standard SQLite or GeoPackage made out of OpenStreetMap (I tested PBF and GeoJSON sources). As described in the [OSM driver docs](https://gdal.org/drivers/vector/osm.html#other-tags-field), `ogr2ogr` made an HSTORE-formatted column `other_tags` to hold the OSM tags that do not have a dedicated column (i.e. are not mentioned in `osmconf.ini`).

As GDAL's parser `hstore_get_value()` ([docs](https://gdal.org/user/sql_sqlite_dialect.html)) does not exist in `sqlite`, I use regexes when running Python / `sqlite3` on another machine, which feels conceptually wrong and is hard or impossible to do safely. Or I could try [this](https://gis.stackexchange.com/q/429483) workaround to do `hstore_to_json` and alter the DB. Also GDAL's implementation does not support SELECT DISTINCT (#3671). I cannot even find the function in the PostgreSQL docs and I feel I'm wasting time to learn idiosyncrasies of a non-reusable tool.

What about using JSON or other standard format instead? SQLite has [native support for JSON](https://www.sqlite.org/json1.html) (become opt-out one year ago, so new deployments can be expected to have it) and the syntax and concepts are shared to some extent across [SQLite's](https://www.sqlite.org/json1.html), [MySQL's](https://dev.mysql.com/doc/refman/5.7/en/json.html) and [PostgreSQL's](https://www.postgresql.org/docs/current/functions-json.html) implementations, so I'd actually spend time to learn a standard.

rouault:
…open option, as an alternative for HSTORE formatting for other_tags/all_tags fields (fixes #7533)

### DIFF
--- a/autotest/ogr/ogr_osm.py
+++ b/autotest/ogr/ogr_osm.py
@@ -37,32 +37,32 @@ import pytest
 
 from osgeo import gdal, ogr
 
+pytestmark = pytest.mark.require_driver("OSM")
+
+
+@pytest.fixture(scope="module", autouse=True)
+def setup_driver():
+    with gdaltest.error_handler(), gdaltest.disable_exceptions():
+        ds = ogr.Open("data/osm/test.osm")
+        ogrtest.osm_drv_parse_osm = ds is not None
+        if (
+            ogrtest.osm_drv_parse_osm is None
+            and "OSM XML detected, but Expat parser not available"
+            not in gdal.GetLastErrorMsg()
+        ):
+            pytest.fail(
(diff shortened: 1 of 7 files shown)


elpaso (approved):
Nice improvement!

PR #7540, opened 2023-03-30
Allow CPLSetConfigOption('PROJ_DATA', ...) to work by making it call OSRSetPROJSearchPaths()

rouault:
People expect PROJ_DATA / PROJ_LIB to be set through configuration options, whereas they only work as environment variables or when calling OSRSetPROJSearchPaths().
Make CPLSetConfigOption('PROJ_DATA', ...) to work by making it call OSRSetPROJSearchPaths()

### DIFF
--- a/port/cpl_conv.cpp
+++ b/port/cpl_conv.cpp
@@ -1870,12 +1870,12 @@ void CPLUnsubscribeToSetConfigOption(int nId)
 }
 
 /************************************************************************/
-/*                  NotifyOtherComponentsSetConfigOptiond()          */
+/*                  NotifyOtherComponentsConfigOptionChanged()          */
 /************************************************************************/
 
-static void NotifyOtherComponentsSetConfigOptiond(const char *pszKey,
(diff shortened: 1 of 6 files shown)


port/cpl_conv.cpp
```diff
@@ -1863,6 +1935,9 @@ void CPL_STDCALL CPLSetConfigOption(const char *pszKey, const char *pszValue)
 
     g_papszConfigOptions = const_cast<volatile char **>(CSLSetNameValue(
         const_cast<char **>(g_papszConfigOptions), pszKey, pszValue));
+
+    NotifyOtherComponentsSetConfigOptiond(pszKey, pszValue,
```
elpaso, line 1939: Does the ending `d` in the method name has a special meaning?

rouault: just a copy& replace error. The intended function name is ``NotifyOtherComponentsConfigOptionChanged``. Fixed

PR #7543, opened 2023-03-31
Autotest: Assorted cleanups

dbaston:
## What does this PR do?

Unrelated cleanups to test files I've collected during the last couple of weeks. I'm trying to avoid a bunch of noise by putting these in the same PR but can split if it's preferable.

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/autotest/gcore/hdf4_read.py
+++ b/autotest/gcore/hdf4_read.py
@@ -259,7 +259,7 @@ def test_hdf4_read_online_5():
 
     # 13 MB
     gdaltest.download_or_skip(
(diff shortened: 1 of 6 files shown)


rouault:
Thanks. Merging.

Is there any recommandation where "__tracebackhide__ = True" is appropriate ?

dbaston:
> Is there any recommandation where "tracebackhide = True" is appropriate ?

It's useful when you are calling a helper function that contains `assert` or `pytest.skip`. With `__tracebackhide__ = True`, the file/line number in the error/skip message will be the line that called the helper function, not line inside the helper function. When I added `gdaltest.download_or_skip` I didn't know about it, so I was trying to approximate the effect with the `inspect` code I removed here.

PR #7558, opened 2023-04-04
GTiff: make sure that band description in PAM overrides the one coming from GDAL_METADATA tag

rouault:


frmts/gtiff/geotiff.cpp
```diff
@@ -16101,6 +16101,46 @@ void GTiffDataset::ApplyPamInfo()
...
         }
     }
+
+    for (int i = 1; i <= nBands; ++i)
+    {
+        GTiffRasterBand *poBand =
+            cpl::down_cast<GTiffRasterBand *>(GetRasterBand(i));
+
+        /* Load scale, offset and unittype from PAM if available */
+        int nHaveOffsetScale = FALSE;
```
elpaso, line 16111: Shouldn't this be a bool ? Why use `FALSE` instead of `false` ?

rouault: > Shouldn't this be a bool ?

unfortunately we cannot as GDALRasterBand::GetScale() takes a int*, not a bool*

I guess we could potentially switch the GetScale() virtual method to take a bool* , update all drivers, and provide a backward compatibility GetScale(int*) method that would call to GetScale(bool*).  Or perhaps even return a std::tuple<double, bool>.  Must ponder that a bit...

> Why use FALSE instead of false ?

yes, probably.  Will change that. MSVC 2015 used to be very picky about mixing int and bool constants, but this was more about doing "bool x = TRUE" and "int x = true;" I believe. And we have now dropped support for MSVC 2015.

PR #7575, opened 2023-04-10
OCI: improve round-tripping of EPSG CRS (fixes #7551) 

### LINKED ISSUE #7551: ogr2ogr convert shape file to oracle spatial database
when i use ogr2ogr convert shape file to oracle database ,the SRID OF USER_SDO_GEOM_METADATA is not correct. 

the sql is :

```sql
INSERT INTO USER_SDO_GEOM_METADATA VALUES ('BJROAD', 'ORA_GEOMETRY', MDSYS.SDO_DIM_ARRAY(MDSYS.SDO_DIM_ELEMENT('X',12843486.24727394,13080040.17120964,0.001),MDSYS.SDO_DIM_ELEMENT('Y',4781532.934468937,5024640.603382006,0.001)), 1000026974)
```

the SRID shoule be EPSG:3857,not 1000026974.

then i use the -a_srs EPSG:3857 it not worked.

rouault:


ogr/ogrsf_frmts/oci/ogrocidatasource.cpp
```diff
@@ -838,6 +838,26 @@ OGRSpatialReference *OGROCIDataSource::FetchSRS(int nId)
         return nullptr;
     }
 
+    /* -------------------------------------------------------------------- */
+    /*      If the WKTEXT definition is equivalent to the EPSG SRS of code  */
+    /*      nId, return that EPSG SRS.                                      */
+    /* -------------------------------------------------------------------- */
+    if (nId < 32767 && papszResult[1] != nullptr && atoi(papszResult[1]) == nId)
```
elpaso, line 845: Can we use std::numeric_limits instead of the literal?

ogr/ogrsf_frmts/oci/ogrocidatasource.cpp
```diff
@@ -900,33 +920,56 @@ int OGROCIDataSource::FetchSRSId(OGRSpatialReference *poSRS)
...
     if (pszAuthName != nullptr && pszAuthCode != nullptr)
     {
+        // If the authority is Oracle, use the code as the SRID
         if (EQUAL(pszAuthName, "Oracle") && atoi(pszAuthCode) != 0)
             return atoi(pszAuthCode);
 
+        // If the authority is EPSG, try to retrieve the record of CS_SRS
+        // whose SRID is the EPSG code. And if the WKT definition from the
+        // CS_SRS record is equivalent to the one of poSRS, use the EPSG code
+        // as the SRID. This works for a number
```
elpaso, line 944: "This works for a number" is there a missing sentence closure ?

ogr/ogrsf_frmts/oci/ogrocidatasource.cpp
rouault: > Can we use std::numeric_limits instead of the literal?

The 32767 indeed probably comes from the highest signed int16, but this is more than it is documented by EPSG as the (currently) highest number for their codes: https://en.wikipedia.org/wiki/EPSG_Geodetic_Parameter_Dataset
I'll change the hard coded value to a HIGHEST_EPSG_CRS_NUMBER constant

ogr/ogrsf_frmts/oci/ogrocidatasource.cpp
rouault: indeed. completed as "This works for a number of EPSG CRS (4326, 3857, etc.)"

PR #7579, opened 2023-04-11
[gpkg] Set field alternative names using name attribute from gpkg_data_columns table

nyalldawson:
Note that the GeoPackage specifications for the gpkg_data_columns table specify that both a field "name" and field "title" can be present. The specifications do not further elaborate on what the intended differentiation is between these values. Accordingly, we set the field alias using either the name value or the title value, depending on which has been populated in the GeoPackage (and which isn't just a direct copy of the column name)

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:

### DIFF
--- a/autotest/ogr/ogr_gpkg.py
+++ b/autotest/ogr/ogr_gpkg.py
@@ -3044,13 +3044,13 @@ def test_ogr_gpkg_34():
         """CREATE TABLE gpkg_data_columns (
   table_name TEXT NOT NULL,
   column_name TEXT NOT NULL,
-  name TEXT UNIQUE,
+  name TEXT,
   title TEXT,
   description TEXT,
   mime_type TEXT,
   constraint_name TEXT,
   CONSTRAINT pk_gdc PRIMARY KEY (table_name, column_name),
-  CONSTRAINT fk_gdc_tn FOREIGN KEY (table_name) REFERENCES gpkg_contents(table_name)
+  CONSTRAINT gdc_tn UNIQUE (table_name, name)
 )"""
     )
     ds.ExecuteSQL(
@@ -3211,13 +3211,13 @@ def test_ogr_gpkg_35():
         """CREATE TABLE gpkg_data_columns (
   table_name TEXT NOT NULL,
   column_name TEXT NOT NULL,
-  name TEXT UNIQUE,
+  name TEXT,
   title TEXT,
   description TEXT,
   mime_type TEXT,
   constraint_name TEXT,
   CONSTRAINT pk_gdc PRIMARY KEY (table_name, column_name),
-  CONSTRAINT fk_gdc_tn FOREIGN KEY (table_name) REFERENCES gpkg_contents(table_name)
+  CONSTRAINT gdc_tn UNIQUE (table_name, name)
 )"""
     )
     ds.ExecuteSQL(
@@ -3332,13 +3332,13 @@ def test_ogr_gpkg_36():
         """CREATE TABLE gpkg_data_columns (
   table_name TEXT NOT NULL,
   column_name TEXT NOT NULL,
-  name TEXT UNIQUE,
+  name TEXT,
   title TEXT,
   description TEXT,
   mime_type TEXT,
   constraint_name TEXT,
   CONSTRAINT pk_gdc PRIMARY KEY (table_name, column_name),
-  CONSTRAINT fk_gdc_tn FOREIGN KEY (table_name) REFERENCES gpkg_contents(table_name)
+  CONSTRAINT gdc_tn UNIQUE (table_name, name)
 )"""
     )
     ds.ExecuteSQL(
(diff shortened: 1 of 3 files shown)


jratike80:
By the standard the "name" field must be unique but the "title" field does not need to be, so maybe there is a risk in using titles as alternative names.
```
CREATE TABLE gpkg_data_columns (
table_name TEXT NOT NULL,
column_name TEXT NOT NULL,
name TEXT UNIQUE,
title TEXT,
description TEXT,
...
```

nyalldawson:
@jratike80 
The issue with the unique constraint on the name field is that it applies across all tables and columns. I can understand the requirement for a unique alias per table, but there's no reason to restrict aliases to have to be unique across the whole gpkg.

That's why I opted for preferring the title field (and I'll use the title field as the destination in an upcoming pr to allow modifying a fields alternative name), but for flexibility/wider compatibility we fall back to name.

jratike80:
Wouldn't is be worth filing a GeoPackage issue about the universal behavior of UNIQUE?

nyalldawson:
@jratike80 

Maybe the name column isn't supposed to be used like this at all, but rather for a "name" of the row in the gpkg_data_columns table 🤔. I think this is overengineered and can't see why we'd ever need two seperate name/title values, when there's already a column name AND description available! 

Regardless, I think gdal should be flexible here just in case other libraries/applications use name for a field alias.

rouault:
There has been a fix for the DDL of gpkg_data_columns_sql between http://www.geopackage.org/spec121/#gpkg_data_columns_sql and http://www.geopackage.org/spec130/#gpkg_data_columns_sql
This was fixed in the spec per https://github.com/opengeospatial/geopackage/pull/492/files (cf https://github.com/opengeospatial/geopackage/issues/487)
@nyalldawson  Could you fix GDALGeoPackageDataset::CreateColumnsTableAndColumnConstraintsTablesIfNecessary() to remove the UNIQUE constraint to name ?
Consequently I would also just think that we need to handle the "name" column for the alias.

> The specifications do not further elaborate on what the intended differentiation is between these values.

It seems to me name is intended to be the alias name due to the mention of "short name" in Table 24. Data Columns Table Definition:
```
name | TEXT | A human-readable identifier (e.g. short name) for the column_name content
-- | -- | --
```

For the write side, I'm not sure if we want to go into the complication of checking the DDL of an already existing gpkg_data_columns_sql table to upgrade it to the new version. Probably not worth the complication

autotest/ogr/ogr_gpkg.py
```diff
@@ -8451,3 +8451,70 @@ def test_ogr_gpkg_ogr_layer_Extent():
...
+        """CREATE TABLE gpkg_data_columns (
+  table_name TEXT NOT NULL,
+  column_name TEXT NOT NULL,
+  name TEXT UNIQUE,
+  title TEXT,
+  description TEXT,
+  mime_type TEXT,
+  constraint_name TEXT,
+  CONSTRAINT pk_gdc PRIMARY KEY (table_name, column_name),
+  CONSTRAINT fk_gdc_tn FOREIGN KEY (table_name) REFERENCES gpkg_contents(table_name)
```
rouault, line 8486:
```suggestion
  CONSTRAINT gdc_tn UNIQUE (table_name, name)
```

autotest/ogr/ogr_gpkg.py
```diff
@@ -8451,3 +8451,70 @@ def test_ogr_gpkg_ogr_layer_Extent():
...
+)"""
+    )
+    # name and title same as column name, won't be used as alternative name
+    ds.ExecuteSQL(
+        "INSERT INTO gpkg_data_columns('table_name', 'column_name', 'name') VALUES ('test', 'foo', 'foo')"
+    )
+    ds.ExecuteSQL(
+        "INSERT INTO gpkg_data_columns('table_name', 'column_name', 'title') VALUES('test', 'baz', 'baz')"
+    )
```
rouault, line 8495: That, and other places testing the baz column to be removed if we handle only name as the possibility for the alias. Cf my general remark in https://github.com/OSGeo/gdal/pull/7579#issuecomment-1502999339

```suggestion
    ds.ExecuteSQL(
        "INSERT INTO gpkg_data_columns('table_name', 'column_name', 'title') VALUES('test', 'baz', 'baz')"
    )
```

ogr/ogrsf_frmts/gpkg/ogrgeopackagetablelayer.cpp
```diff
@@ -1284,7 +1284,7 @@ OGRErr OGRGeoPackageTableLayer::ReadTableDefinition()
     // Look for sub-types such as JSON
     if (m_poDS->HasDataColumnsTable())
     {
-        pszSQL = sqlite3_mprintf("SELECT column_name, mime_type, "
+        pszSQL = sqlite3_mprintf("SELECT column_name, name, title, mime_type, "
```
rouault, line 1287: Potentially title to be removed if we only handle name

nyalldawson:
@rouault 

Thanks for the insights! All addressed now.

rouault:
bug fix commit 05327ac6f2f02c235e5de20fb0ba2a5d7b259409 backported to 3.6 branch per 84f3a9a0ff9d916c407d53ddac825038317e42cb

PR #7582, opened 2023-04-11
Add optional comment to OGRFieldDefn, read for gpkg layers

nyalldawson:
Adds the API framework for handling field comments for OGR layers

And support reading comments for GPKG fields, using "description" from gpkg_data_columns.

(Write capacity for field comments will be added to GPKG driver in a follow up PR)

### DIFF
--- a/ogr/ogr_api.h
+++ b/ogr/ogr_api.h
@@ -402,6 +402,8 @@ void CPL_DLL OGR_Fld_SetDefault(OGRFieldDefnH hDefn, const char *);
 int CPL_DLL OGR_Fld_IsDefaultDriverSpecific(OGRFieldDefnH hDefn);
 const char CPL_DLL *OGR_Fld_GetDomainName(OGRFieldDefnH hDefn);
 void CPL_DLL OGR_Fld_SetDomainName(OGRFieldDefnH hDefn, const char *);
+const char CPL_DLL *OGR_Fld_GetComment(OGRFieldDefnH hDefn);
+void CPL_DLL OGR_Fld_SetComment(OGRFieldDefnH hDefn, const char *);
 
 const char CPL_DLL *OGR_GetFieldTypeName(OGRFieldType);
 const char CPL_DLL *OGR_GetFieldSubTypeName(OGRFieldSubType);
--- a/ogr/ogrsf_frmts/gpkg/ogrgeopackagetablelayer.cpp
+++ b/ogr/ogrsf_frmts/gpkg/ogrgeopackagetablelayer.cpp
@@ -1284,10 +1284,11 @@ OGRErr OGRGeoPackageTableLayer::ReadTableDefinition()
     // Look for sub-types such as JSON
     if (m_poDS->HasDataColumnsTable())
     {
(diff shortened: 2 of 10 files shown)


ogr/ogr_api.h
```diff
@@ -402,6 +402,8 @@ void CPL_DLL OGR_Fld_SetDefault(OGRFieldDefnH hDefn, const char *);
 int CPL_DLL OGR_Fld_IsDefaultDriverSpecific(OGRFieldDefnH hDefn);
 const char CPL_DLL *OGR_Fld_GetDomainName(OGRFieldDefnH hDefn);
 void CPL_DLL OGR_Fld_SetDomainName(OGRFieldDefnH hDefn, const char *);
+const char CPL_DLL *OGR_Fld_GetComment(OGRFieldDefnH hDefn);
+void CPL_DLL OGR_Fld_SetComment(OGRFieldDefnH hDefn, const char *);
```
rouault, line 406: Binding that to SWIG would be desirable. Can be in the follow-up pull request you mention

rouault:
Note: PostgreSQL could also support that capability: https://www.postgresql.org/docs/current/sql-comment.html

ogr/ogr_api.h
nyalldawson: Done

nyalldawson:
I've added read support for gpkg in here now

ogr/ogrsf_frmts/gpkg/ogrgeopackagetablelayer.cpp
```diff
@@ -1316,6 +1317,17 @@ OGRErr OGRGeoPackageTableLayer::ReadTableDefinition()
                     }
                 }
 
+                if (const char *pszDescription =
+                        oResultTable->GetValue(4, iRecord))
+                {
+                    int iIdx = m_poFeatureDefn->GetFieldIndex(pszColumn);
```
elpaso, line 1323:
```suggestion
                    const int iIdx = m_poFeatureDefn->GetFieldIndex(pszColumn);
```

ogr/ogrsf_frmts/gpkg/ogrgeopackagetablelayer.cpp
```diff
@@ -1284,10 +1284,11 @@ OGRErr OGRGeoPackageTableLayer::ReadTableDefinition()
...
     {
-        pszSQL = sqlite3_mprintf("SELECT column_name, name, mime_type, "
-                                 "constraint_name FROM gpkg_data_columns "
-                                 "WHERE table_name = '%q'",
-                                 m_pszTableName);
+        pszSQL = sqlite3_mprintf(
+            "SELECT column_name, name, mime_type, "
+            "constraint_name, description FROM gpkg_data_columns "
+            "WHERE table_name = '%q'",
+            m_pszTableName);
```
elpaso, line 1291: Did you check if m_pszTableName may require escaping?

rouault: The '%q' formatting, specific of sqlite3_mprintf(), takes care of that: https://www.sqlite.org/printf.html#percentq

PR #7591, opened 2023-04-13
Tiger: fix compiler warning with clang 16.0.0, and tweak linux_build.yml

rouault:
CC @dbaston for the first commit in this pull request


.github/workflows/linux_build.yml
```diff
@@ -127,14 +127,23 @@ jobs:
...
 
-      - name: Pull build environment
+      - name: Set variables
         shell: bash
         run: |
           export CONTAINER_NAME_FULL="${CONTAINER_REGISTRY}/${CONTAINER_REGISTRY_USER,,}/${CONTAINER_NAME}"
+          echo "CONTAINER_NAME_FULL=${CONTAINER_NAME_FULL}" >>${GITHUB_ENV}
+
+      # Only pull an existing build container in pull requests that don't
+      # contain [skip cache] in the commit message.
```
dbaston, line 137:
```suggestion
      # Pull an existing build container for all pull requests that don't
      # contain [skip cache] in the latest commit message.
```

.github/workflows/linux_build.yml
```diff
@@ -127,14 +127,23 @@ jobs:
...
+      # That is push builds will not use an existing container, to be able
+      # to refresh rolling distributions.
+      - name: Pull build environment
```
dbaston, line 140: Rather than not pulling the build environment, what do you think about adding `--pull` to the subsequent `docker build` command? This should only rebuild the underlying environment if the base container has in fact changed.

rouault: Not totally sure to follow you, w.r.t "base container" means. If you mean the fedora:rawhide one, then the thing is that it is not updated that often, but the fedora packages themselves are updated constantly.
Anyway looking at the push in my rouault/gdal event (https://github.com/rouault/gdal/actions/runs/4689161046/jobs/8310569414), I see my current logic is defeated, even if skipping this "docker pull" step. I assume the "--cache-from" arguments cause an indirect pull, so they should be removed in the mode "refresh the build container from scratch"
I'm going to drop that commit as it isn't ready. Is it something you'd be willing to try working on?

dbaston: OK, I understand. Yes, I will work on this.

PR #7598, opened 2023-04-14
[flatgeobuf] Support reading/writing field comments in field description metadata

nyalldawson:


ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
```diff
@@ -301,7 +301,9 @@ OGRFlatGeobufLayer::writeColumns(FlatBufferBuilder &fbb)
         auto title = field->GetAlternativeNameRef();
         if (EQUAL(title, ""))
             title = nullptr;
-        const char *description = nullptr;
+        const std::string osComment = field->GetComment();
```
rouault, line 304:
```suggestion
        const std::string& osComment = field->GetComment();
```

ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp
```diff
@@ -301,7 +301,9 @@ OGRFlatGeobufLayer::writeColumns(FlatBufferBuilder &fbb)
         auto title = field->GetAlternativeNameRef();
         if (EQUAL(title, ""))
             title = nullptr;
-        const char *description = nullptr;
+        const std::string& osComment = field->GetComment();
```
rouault, line 304:
```suggestion
        const std::string &osComment = field->GetComment();
```

PR #7600, opened 2023-04-15
PG: support for getting/setting/altering field comments (fixes #7587) + PGDump: set field comments

### LINKED ISSUE #7587: PostgreSQL: add support for field comments
Use capability introduced in https://github.com/OSGeo/gdal/pull/7582 to map it with "COMMENT ON"  (https://www.postgresql.org/docs/current/sql-comment.html)

rouault:


nyalldawson (approved):

nyalldawson:
@rouault are you ok with me adding a new driver metadata value for "supported field capabilities"? We have the existing one indicating support for altering different field properties, but it's a bad fit for create only drivers which don't support field alterations....

rouault:
> We have the existing one indicating support for altering different field properties, but it's a bad fit for create only drivers which don't support field alterations....

I was actually thinking this would be a good idea when doing all this!

nyalldawson:
Ok, I'll wait till this round of related driver changes are all merged then add the new metadata in a follow up!

PR #7612, opened 2023-04-18
ERS: Add GDA2020 support for .ERS files

jwallace94:
## What does this PR do?
Adds an entry of GDA2020 and its corresponding WKT to the ecw_cs.wkt file. This allows for gdal.create to successfully write the GDA2020 datum to an output ERS file when using DATUM in the creation options. 

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Review
 - [ ] Adjust for comments

## Future Work

This is a potentially incomplete solution, as it requires using creation options to successfully write the GDA2020 datum to an ERS file, whereas a standard/established datum (such as WGS84) does not require this.


rouault:
> This is a potentially incomplete solution,

I've added extra commit to make it work fully

PR #7615, opened 2023-04-19
TileDB: Add read/write vector side

rouault:


frmts/tiledb/tiledbsparse.cpp
```diff
@@ -0,0 +1,4934 @@
...
+        // create the tiledb schema
+        // dimensions will be _x and _y, we can also add _z (2.5d)
+        // filters should be implemented as per tiledbdense
+        // set dimensions and attribute type for schema
+        // we will use row-major for now but we could use hilbert indexing
+        m_schema.reset(new tiledb::ArraySchema(*m_ctx, TILEDB_SPARSE));
+        m_schema->set_tile_order(TILEDB_ROW_MAJOR);
+        m_schema->set_cell_order(TILEDB_ROW_MAJOR);
+
+        m_schema->set_coords_filter_list(*m_filterList);
```
normanb, line 3223: Need to add `m_schema→set_offsets_filter_list(*m_filterList)` so that variable length offsets are also compressed. Additionally we need to set tile capacity as per per https://github.com/PDAL/PDAL/blob/master/plugins/tiledb/io/TileDBWriter.cpp#L620 100000 is small though, it can be 500000.

frmts/tiledb/tiledbsparse.cpp
```diff
@@ -0,0 +1,4934 @@
...
+                                    : CPLGetBasename(osLayerFilename.c_str()),
+            wkbUnknown, nullptr);
+        poLayer->m_bUpdatable = poOpenInfo->eAccess == GA_Update;
+        if (!poLayer->InitFromStorage(poDS->m_ctx.get(), nTimestamp,
+                                      poOpenInfo->papszOpenOptions))
+        {
+            return false;
+        }
+
+        poLayer->m_nBatchSize =
```
normanb, line 274: `batch_size` should be larger, `ogr2ogr -f TileDB nz_tiledb nz-building-outlines.shp -lco BATCH_SIZE=500000` gives fragment sizes that are around 150MB, it is dataset dependent but the batch size above seems reasonable.

frmts/tiledb/tiledbsparse.cpp
```diff
@@ -0,0 +1,4934 @@
...
+
+        SetReadBuffers();
+
+        // Create a loop
+        tiledb::Query::Status status;
+        uint64_t nRowCount = 0;
+        while (true)
+        {
+            // Submit query and get status
+            m_query->submit();
```
normanb, line 1625: Missing stats as per the dense implementation https://github.com/rouault/gdal/blob/tiledb_geometries_ero/frmts/tiledb/tiledbdense.cpp#L1938 it will need to wrap `m_query→submit()` and be a configurable creation/open option.

normanb (changes requested):
@rouault this looks great and @perrygeo and myself have done a functional review with `QGIS`, `GeoPandas`, `fiona`, `shapely` and `tiledb-py` in addition to `ogr2ogr` and the python OGR bindings. I have put some minor comments in the code, I also have a couple of other comments;

- [ ] bench_ogr_to_geopandas.py from https://gdal.org/development/rfc/rfc86_column_oriented_api.html crashes with the TileDB version of the NZ building outlines data.
- [ ] There is an `ArrowAdapter` in TileDB, https://github.com/TileDB-Inc/TileDB/blob/dev/tiledb/sm/cpp_api/arrowio not sure if that would make the code simpler or not (your choice)

I have a longer comment on reading an array with `X`, `Y` and `Z` specified as open options, see below for steps to recreate;

Define a basic PDAL pipeline as follows;

```
    [
      {
        "type": "readers.las",
        "filename": "https://github.com/PDAL/PDAL/blob/master/test/data/las/autzen_trim.las?raw=true"
      },
      {
        "type": "filters.stats"
      },
      {
        "type": "writers.tiledb",
        "array_name": "pc_sample_tiledb"
      }
    ]
```
```
    conda install -c conda-forge pdal
    pdal pipeline -i pipeline.json --nostream
    
    ogrinfo -al -so pc_sample_tiledb -oo DIM_X=X -oo DIM_Y=Y -oo DIM_Z=Z
```
This gets the extents but fails 

```
    Warning 1: Ignoring attribute Intensity as its type UINT16 is unsupported
    Warning 1: Ignoring attribute ReturnNumber of type UINT8, as only variable length is supported, but it has a fixed size
    Warning 1: Ignoring attribute NumberOfReturns of type UINT8, as only variable length is supported, but it has a fixed size
    Warning 1: Ignoring attribute ScanDirectionFlag of type UINT8, as only variable length is supported, but it has a fixed size
    Warning 1: Ignoring attribute EdgeOfFlightLine of type UINT8, as only variable length is supported, but it has a fixed size
    Warning 1: Ignoring attribute Classification of type UINT8, as only variable length is supported, but it has a fixed size
    Warning 1: Ignoring attribute UserData of type UINT8, as only variable length is supported, but it has a fixed size
    Warning 1: Ignoring attribute PointSourceId as its type UINT16 is unsupported
    Warning 1: Ignoring attribute Red as its type UINT16 is unsupported
    Warning 1: Ignoring attribute Green as its type UINT16 is unsupported
    Warning 1: Ignoring attribute Blue as its type UINT16 is unsupported
    
    INFO: Open of `pc_sample_tiledb'
          using driver `TileDB' successful.
    
    Layer name: pc_sample_tiledb
    Geometry: 3D Point
    Feature Count: 110000
    Extent: (636001.760000, 848935.200000) - (637179.220000, 849497.900000)
    Layer SRS WKT:
    (unknown)
    ScanAngleRank: Real(Float32) (0.0) NOT NULL
    GpsTime: Real (0.0) NOT NULL
```

frmts/tiledb/tiledbsparse.cpp
rouault: done

frmts/tiledb/tiledbsparse.cpp
rouault: done

rouault:
> * bench_ogr_to_geopandas.py from https://gdal.org/development/rfc/rfc86_column_oriented_api.html crashes with the TileDB version of the NZ building outlines data.

I could reproduce. It was not specific to the Arrow interface actually. I had a hard time to figure out what was wrong (when the partial result set returned by submit() was lower than the batch size, I growed immediately the variable size arrays and called set_buffer(). Now the resizing is done only after having used the result set before the next call to submit()) , but it is now fixed per https://github.com/OSGeo/gdal/pull/7615/commits/371ad3a6760b5d3d08e4dfd4d9b580ef63beab72

> There is an ArrowAdapter in TileDB,

yes I noticed that. I ended up not using it because it doesn't handle nullable arrays and variable list of booleans. I also needed to do handle assembling of several ArrowArray in a Arrow structure to return multiple fields, as well as doing special processing to reconstruct WKB from X, Y, [Z] arrays when there's no explicit WKB. All in all, manual handling seemed simpler.

Regarding interoperability with PDAL's datasets, I've added support for UINT8 and UINT16 data types, as well as automatically detecting dimensions "X", "Y" and "Z"
``ogrinfo pc_sample_tiledb -al`` now returns:
```
INFO: Open of `pc_sample_tiledb'
      using driver `TileDB' successful.

Layer name: pc_sample_tiledb
Geometry: 3D Point
Feature Count: 110000
Extent: (636001.760000, 848935.200000) - (637179.220000, 849497.900000)
Layer SRS WKT:
(unknown)
Intensity: Integer (0.0) NOT NULL
ReturnNumber: Integer (0.0) NOT NULL
NumberOfReturns: Integer (0.0) NOT NULL
ScanDirectionFlag: Integer (0.0) NOT NULL
EdgeOfFlightLine: Integer (0.0) NOT NULL
Classification: Integer (0.0) NOT NULL
ScanAngleRank: Real(Float32) (0.0) NOT NULL
UserData: Integer (0.0) NOT NULL
PointSourceId: Integer (0.0) NOT NULL
GpsTime: Real (0.0) NOT NULL
Red: Integer (0.0) NOT NULL
Green: Integer (0.0) NOT NULL
Blue: Integer (0.0) NOT NULL
OGRFeature(pc_sample_tiledb):1
  Intensity (Integer) = 2
  ReturnNumber (Integer) = 1
  NumberOfReturns (Integer) = 1
  ScanDirectionFlag (Integer) = 0
  EdgeOfFlightLine (Integer) = 0
  Classification (Integer) = 2
  ScanAngleRank (Real(Float32)) = -14.0
  UserData (Integer) = 126
  PointSourceId (Integer) = 7326
  GpsTime (Real) = 245380.77684157
  Red (Integer) = 96
  Green (Integer) = 97
  Blue (Integer) = 94
  POINT Z (636958.95 849283.49 410.79)
[...]
```

I believe I've addressed all your comments

normanb (approved):

normanb:
@rouault I have been building the perf test suite and running comparisons, I think I was wrong to suggest a default TILE_CAPACITY of 500_000, can we set it to the default `10_000`, it makes a huge difference in the results. The BATCH_SIZE should still be large. I can make a PR against this branch if that saves some work.

rouault:
> can we set it to the default `10_000`

done

PR #7621, opened 2023-04-20
GPKG: add ST_EnvIntersects() for faster spatial filtering when there is no spatial index

rouault:
Execution time can be ~ 1.5x faster.

Before:
```
$ time ogrinfo nz-building-outlines.gpkg -sql "select count(*) from \"nz-building-outlines\" where st_maxx(geometry)>=1500000 and st_maxy(geometry)>= 5200000 and st_minx(geometry) <=1800000 and st_miny(geometry) <= 5800000" -al -q

Layer name: SELECT
OGRFeature(SELECT):0
  count(*) (Integer) = 701300

real	0m0,909s
user	0m0,633s
sys	0m0,276s
```

After:
```
$ time ogrinfo nz-building-outlines.gpkg -sql "select count(*) from \"nz-building-outlines\" where ST_EnvelopesIntersects(geometry, 1500000, 5200000 ,1800000 ,5800000)" -al -q

Layer name: SELECT
OGRFeature(SELECT):0
  count(*) (Integer) = 701300

real	0m0,605s
user	0m0,310s
sys	0m0,290s
```

### DIFF
--- a/autotest/ogr/ogr_gpkg.py
+++ b/autotest/ogr/ogr_gpkg.py
@@ -1317,6 +1317,79 @@ def test_ogr_gpkg_SetSRID():
 
 
 ###############################################################################
+# Test ST_EnvIntersects() function
+
+
+def test_ogr_gpkg_ST_EnvIntersects():
+
+    filename = "/vsimem/test_ogr_gpkg_ST_EnvIntersects.gpkg"
+    ds = ogr.GetDriverByName("GPKG").CreateDataSource(filename)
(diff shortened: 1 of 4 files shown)


jratike80:
Is the somehow ungrammatical ST_EnvelopesIntersects alias there because SpatiaLite has it also?

rouault:
> Is the somehow ungrammatical ST_EnvelopesIntersects alias there because SpatiaLite has it also?

yes

PR #7624, opened 2023-04-20
Python bindings: make Dataset.ExecuteSQL() usable as a context manager to automatically release the layer (fixes #7459)

### LINKED ISSUE #7459: Python bindings: create a ExecuteSQL() context manager
Patterns like:
```
sql_lyr = ds.ExecuteSQL(...)
something
ds.ReleaseResultSet(sql_lyr)
```
are tedious and error-prone.
It would be better if we could something like
```
with ds.ExecuteSQL(...) as sql_lyr:
    something
```
We'd probably need a different method name for the context manager as it is not obvious we can have both behaviour with same name (cf https://discourse.techart.online/t/python-can-a-context-manager-be-used-as-a-regular-function/5485/9)

autotest could benefit from that retrofit

rouault:
```
    ExecuteSQL(self, statement, spatialFilter: ogr.Geometry = None, dialect: Optional[str] = "", keep_ref_on_ds=False) -> ogr.Layer

    Example 1
    ---------

    with ds.ExecuteSQL("SELECT * FROM layer") as lyr:
        print(lyr.GetFeatureCount())

    Example 2
    ---------

    def get_sql_lyr():
        return gdal.OpenEx("test.shp").ExecuteSQL("SELECT * FROM test", keep_ref_on_ds=True)

    with get_sql_lyr() as lyr:
        print(lyr.GetFeatureCount())
```

### DIFF
--- a/swig/include/python/ogr_python.i
+++ b/swig/include/python/ogr_python.i
@@ -181,11 +181,11 @@ def ExecuteSQL(self, statement, spatialFilter=None, dialect="", keep_ref_on_ds=F
       - or a ogr.Layer handle representing a results set from the query.
 
     Note that this ogr.Layer is in addition to the layers in the data store
-    and must be destroyed with ReleaseResultSet() before the data source is closed
+    and must be released with ReleaseResultSet() before the data source is closed
     (destroyed).
 
     Starting with GDAL 3.7, this method can also be used as a context manager,
-    as a convenient way of automatically de-allocating the returned result layer.
+    as a convenient way of automatically releasing the returned result layer.
 
     For more information on the SQL dialect supported internally by OGR
     review the OGR SQL document (:ref:`ogr_sql_sqlite_dialect`)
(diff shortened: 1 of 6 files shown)


swig/include/python/ogr_python.i
```diff
@@ -168,6 +168,100 @@ def _WarnIfUserHasNotSpecifiedIfUsingExceptions():
...
+        the dialect. The SQLITE dialect can also be used.
+    keep_ref_on_ds:
+        whether the returned layer should keep a (strong) reference on
+        the current dataset. Cf example 2 for a use case.
+
+    Returns
+    -------
+    ogr.Layer:
+        a ogr.Layer containing the results of the query, that will be
+        automatically freed when the context manager goes out of scope
```
elpaso, line 216: We should be consistent with the verb used to indicate deallocation: "free" or "destroy" or one might think that they are referring to different operations.

elpaso (approved):

swig/include/python/ogr_python.i
rouault: done. I've opted for released/releasing to align with the ReleaseResultSet() method

PR #7633, opened 2023-04-21
GPKG: minimum (read) support for non-standard multiple geometry columns per table

rouault:
Refs https://github.com/opengisch/QgisModelBaker/issues/531

If a table has several geometry coluns in its DDL and is declared as many times in a modified gpkg_geometry_columns whose uk_gc_table_name unique constraint has been modified to have (table_name, column_name) as the unique constraint, then instantiate as many OGR layers has geometry columns.

e.g with the following content
```
INSERT INTO gpkg_geometry_columns VALUES('test','poly','POLYGON',0,0,0)
INSERT INTO gpkg_geometry_columns VALUES('test','pt','POINT',0,0,0)
CREATE TABLE "test" ( "ogc_fid" INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL, "poly" POLYGON, "pt" POINT, "area" REAL, "eas_id" INTEGER, "prfedea" TEXT(16))
```

the driver will report 2 layers: "test (poly)" and "test (pt)" In "test(poly)", the "pt" geometry column will be ignored, and in similarly for "test(pt)" where the "poly" geometry column will be ignored.

This might not be fully bullet proof for edition support (e.g if adding a feature in one layer, the driver might be a bit confused if not closing and re-opening), but should be OK for read support.

The way we handle that might change in the future (e.g could be a single layer with multiple geometry columns)

### DIFF
--- a/autotest/ogr/ogr_gpkg.py
+++ b/autotest/ogr/ogr_gpkg.py
@@ -4709,34 +4709,97 @@ def test_ogr_gpkg_creation_fid():
 
 def test_ogr_gpkg_57():
 
-    if gdaltest.gpkg_dr.GetMetadataItem("ENABLE_SQL_GPKG_FORMAT") != "YES":
-        pytest.skip()
+    out_filename = "/vsimem/test_ogr_gpkg_57.gpkg"
+    ogr.GetDriverByName("GPKG").CreateDataSource(out_filename)
 
-    tmpfile = "/vsimem/tmp.gpkg.txt"
-    gdal.FileFromMemBuffer(
-        tmpfile,
-        """-- SQL GPKG
-CREATE TABLE gpkg_spatial_ref_sys (srs_name,srs_id,organization,organization_coordsys_id,definition,description);
-INSERT INTO "gpkg_spatial_ref_sys" VALUES('',0,'NONE',0,'undefined','');
-CREATE TABLE gpkg_contents (table_name,data_type,identifier,description,last_change,min_x, min_y,max_x, max_y,srs_id);
-INSERT INTO "gpkg_contents" VALUES('poly','features','poly','','',NULL,NULL,NULL,NULL,0);
-INSERT INTO "gpkg_contents" VALUES('poly','features','poly','','',NULL,NULL,NULL,NULL,0);
-CREATE TABLE gpkg_geometry_columns (table_name,column_name,geometry_type_name,srs_id,z,m);
-INSERT INTO "gpkg_geometry_columns" VALUES('poly','geom','POLYGON',0,0,0);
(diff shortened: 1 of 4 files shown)


autotest/ogr/ogr_gpkg.py
```diff
@@ -4709,34 +4709,97 @@ def test_ogr_gpkg_creation_fid():
...
+    )
+    ds.ExecuteSQL(
+        "INSERT INTO test VALUES(1,X'4750000300000000000000401f401d41000000e05e511d4100000080322c5241000000001d2d52410103000000010000001b000000000000c01a481d4100000080072d5241000000e0814b1d41000000001d2d524100000040c44b1d41000000000f2d5241000000002c4c1d41000000a0002d524100000000774d1d41000000c0072d5241000000a0c44e1d4100000080112d52410000002008501d41000000c0172d5241000000e05e511d4100000020dd2c5241000000405e511d4100000040cf2c524100000000f0501d41000000c0ba2c52410000008084501d4100000020af2c524100000040a94f1d4100000000a42c524100000080744e1d41000000a09a2c524100000040014f1d41000000c0852c5241000000e0e04d1d4100000020872c524100000040f8441d41000000e0432c5241000000c012441d4100000080322c524100000000ff431d4100000020362c5241000000004b431d4100000080552c52410000000030431d41000000c05d2c5241000000c09b421d4100000000712c524100000080d6411d4100000080912c52410000008027411d4100000040b22c5241000000401f401d4100000040d62c5241000000a043441d4100000060f02c524100000060aa461d4100000080ff2c5241000000c01a481d4100000080072d5241',X'4750000100000000010100000000000000804b1d41000000001d2d5241',NULL,170,NULL);"
```
elpaso, line 4761: is it possible to use a human readable format for the geometry (WKT?) ?

ogr/ogrsf_frmts/gpkg/ogrgeopackagetablelayer.cpp
```diff
@@ -914,6 +914,10 @@ OGRErr OGRGeoPackageTableLayer::ReadTableDefinition()
     }
 #endif
 
+    bool bHasPreexistingSingleGeomColumn =
```
elpaso, line 917:
```suggestion
    const bool bHasPreexistingSingleGeomColumn =
```

elpaso (commented):
Are you going to add this feature to the documentation?

rouault:
> Are you going to add this feature to the documentation?

I'm not sure we want to encourage too much use of this capability while this hasn't been though a bit more, hence this silent form of support

autotest/ogr/ogr_gpkg.py
rouault: Unfortunately there's no ST_GeomFromText() for a GPKG datasource

PR #7635, opened 2023-04-21
A few code cleanups in geotiff.cpp

rouault:
- early compute nBlocksPerRow/Column to avoid recomputing them at several places
- add a ComputeBlockId() helper method


rouault:
> I assume the current GeoTIFF tests already cover the changes.

hopefully yes, as the driver is one of the most tested, but I didn't formally check that all modified code paths are indeed tested.

PR #7645, opened 2023-04-26
Add API methods for retrieving alternative names and comments for dataset layers

nyalldawson:
Draft only, for feedback at this stage...

The intention is that these would be the layer-level equivalents of the OGR field alternative name / comment properties.

Eg the geopackage driver could retrieve these from the `gpkg_contents` table fields `identifier` and `description` columns respectively, and the postgres driver could retrieve table comments from which have been set via a `COMMENT ON ...` query.

### DIFF
--- a/gcore/gdal.h
+++ b/gcore/gdal.h
@@ -1204,6 +1204,8 @@ int CPL_DLL GDALDatasetGetLayerCount(GDALDatasetH);
 OGRLayerH CPL_DLL GDALDatasetGetLayer(GDALDatasetH, int);
 OGRLayerH CPL_DLL GDALDatasetGetLayerByName(GDALDatasetH, const char *);
 int CPL_DLL GDALDatasetIsLayerPrivate(GDALDatasetH, int);
+const char CPL_DLL *GDALDatasetGetLayerAlternativeName(GDALDatasetH, int);
+const char CPL_DLL *GDALDatasetGetLayerComment(GDALDatasetH, int);
 OGRErr CPL_DLL GDALDatasetDeleteLayer(GDALDatasetH, int);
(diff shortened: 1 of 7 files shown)


elpaso:
Nice!

rouault:
Why not putting those methods in the OGRLayer class ?
(timing wise, I'm not sure this will be able to make it for 3.7)

nyalldawson:
> timing wise, I'm not sure this will be able to make it for 3.7

That's no problem -- it's not client funded work, there's no urgency for this one 👍

nyalldawson:
@rouault 

Resurrecting this old work :laughing: 

> Why not putting those methods in the OGRLayer class ?

I'm thinking that raster containers could potentially require this too (eg a raster in a GPKG, or potentially a raster in a filegeodatabase)

rouault:
> I'm thinking that raster containers could potentially require this too

maybe, but GDALDataset::GetLayerAlternativeName(layer_index) doesn't look very natural to me.  OGRLayer::GetAlternativeName() would feel more natural.  One could potentially even imagine that AlternativeName and Comment should be members of OGRFeatureDefn (similarly to OGRFieldDefn), although I'm 50% 50% on doing that compared to having them at the OGRLayer level

nyalldawson:
@rouault could/should we do both? Have the low level method for raster use, and a nicer API for vector?

rouault:
> Have the low level method for raster use

what do you think exactly ?  Technically a GDALDataset, as a raster, is for a single raster layer, so that would be a GetComment() / GetAlternativeName().   For subdatasets, you need to GDALOpen() each subdataset and call GetComment()/GetAlternativeName() on it.

rouault:
closing that one as staled

PR #7653, opened 2023-04-28
DIMAP: Correctly set RPC origin for SPOT, PHR and PNEO

tbonfort:
Rework #5725 to set RPC origin from the sensor profile instead of from the ImagetoGround_Validity_Domain

cc @Lunyxis @jguinet 

For reference: #5716 and https://lists.osgeo.org/pipermail/gdal-dev/2015-June/042090.html

### DIFF
--- a/gcore/mdreader/reader_pleiades.cpp
+++ b/gcore/mdreader/reader_pleiades.cpp
@@ -360,25 +360,6 @@ char **GDALMDReaderPleiades::LoadRPCXmlFile()
         return nullptr;
     }
 
-    // search Image to Ground Validity (since DIMAP v3)
-    CPLXMLNode *pValidityNode =
-        CPLSearchXMLNode(pNode, "=ImagetoGround_Validity_Domain");
-
-    double firstCol = 1.0;
-    if (pValidityNode != nullptr)
-    {
-        char **papszValidity = ReadXMLToList(pValidityNode->psChild, nullptr);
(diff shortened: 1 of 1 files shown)


jguinet:
Good idea.  ImagetoGround_Validity_Domain can theorically starts at a non origin position.  

https://github.com/OSGeo/gdal/pull/5725#issuecomment-1146155486

rouault:
@tbonfort  anything missing for this PR to be mergeable ?

tbonfort:
@rouault I'm off work this week, this pr can be included as is if you want to include it in the 3.7 release candidates. I still have a small batch of refactorings with no functional change to add to it but they can come in a subsequent pr.

tbonfort:
ha, sorry, I just saw Im late to the game as the rc has already been released. So this pr is a candidate for rc2 if it were to exist but does not warrant an rc2 as of itself

rouault:
Merging this, and tagging it for backport for 3.7.1

rouault:
@tbonfort @Lunyxis @jguinet  A user is reporting in https://lists.osgeo.org/pipermail/gdal-dev/2023-July/057505.html that it seems that this changes causes a regression with some PNEO images whose ImagetoGround_Validity_Domain.FIRST_COL/FIRST_ROW are negative

tbonfort:
@rouault thanks, I've followed up off list with him and will report back with the outcome

Issue #6882, opened 2022-12-09
TileMatrixSet::parse() not ready for OGC 2D Tile Matrix Set v2


rouault:
Cf  https://maps.gnosis.earth/ogcapi/tileMatrixSets/WorldCRS84Quad 

Some things that might have changed since then in that particular file:
- `type` removed
- `id` -> `identifier`
- `topLeftCorner` -> `pointOfOrigin`
- `bbox` removed (it is optional, and for informative purposes only -- it should not be relied upon by clients)
- `tileMatrix` -> `tileMatrices` (I think that was already changed though)

_Originally posted by @jerstlouis in https://github.com/OSGeo/gdal/issues/6832#issuecomment-1344626351_

jerstlouis:
Sorry `identifier`->`id`

PR #7656, opened 2023-04-29
- Added fixes for the OGC API Tiles driver

doublebyte1:
The OGC API Tiles dataset was not working as expected.

A complete document was being passed to TileMatrixSet::parse, which was causing a series of issues:
- the comparisons with the well-known uris were never verified
- all the other ways of retrieving a TMS were failing and it was falling back to reading TMS from a file (which had the entire document in the filename causing buffer overflow)

The well-known URI was being compared to a word/name, rather than an uri. I added support to two of the most popular ones, but we could add more.

Submitted changes:
 - Passed uri to TileMatrixSet::parse
 - updated uris of well-known tile matrix sets

With these changes, I can read raster and vector tiles from these urls:

`gdalinfo -oo API=TILES --config CPL_DEBUG ON "OGCAPI:https://maps.gnosis.earth/ogcapi/collections/blueMarble"`
`ogrinfo -oo API=TILES --config CPL_DEBUG ON "OGCAPI:https://maps.gnosis.earth/ogcapi/collections/Daraa"`

### DIFF
--- a/frmts/ogcapi/gdalogcapidataset.cpp
+++ b/frmts/ogcapi/gdalogcapidataset.cpp
@@ -684,7 +684,7 @@ bool OGCAPIDataset::ProcessScale(CPLJSONObject &oScaleDenominator,
         dfYSize /= 2;
     }
 
-    nRasterXSize = std::max(1, static_cast<int>(0 + 5 + dfXSize));
+    nRasterXSize = std::max(1, static_cast<int>(0.5 + dfXSize));
     nRasterYSize = std::max(1, static_cast<int>(0.5 + dfYSize));
     m_adfGeoTransform[0] = dfXMin;
     m_adfGeoTransform[1] = (dfXMax - dfXMin) / nRasterXSize;
--- a/frmts/ogcapi/gdalogcapidataset.cpp
+++ b/frmts/ogcapi/gdalogcapidataset.cpp
@@ -94,6 +94,9 @@ class OGCAPIDataset final : public GDALDataset
 
     bool InitFromFile(GDALOpenInfo *poOpenInfo);
     bool InitFromURL(GDALOpenInfo *poOpenInfo);
+    bool ProcessScale(CPLJSONObject &oScaleDenominator, const double dfXMin,
+                      const double dfYMin, const double dfXMax,
+                      const double dfYMax);
     bool InitFromCollection(GDALOpenInfo *poOpenInfo, CPLJSONDocument &oDoc);
     bool Download(const CPLString &osURL, const char *pszPostContent,
                   const char *pszAccept, CPLString &osResult,
@@ -656,6 +659,44 @@ bool OGCAPIDataset::InitFromFile(GDALOpenInfo *poOpenInfo)
 }
 
 /************************************************************************/
+/*                        ProcessScale()                          */
+/************************************************************************/
+
+bool OGCAPIDataset::ProcessScale(CPLJSONObject &oScaleDenominator,
+                                 const double dfXMin, const double dfYMin,
+                                 const double dfXMax, const double dfYMax)
+
+{
+    bool bRet = FALSE;
+    double dfRes = 1e-8;  // arbitrary
+    if (oScaleDenominator.IsValid())
+    {
+        const double dfScaleDenominator = oScaleDenominator.ToDouble();
+        constexpr double HALF_CIRCUMFERENCE = 6378137 * M_PI;
+        dfRes = dfScaleDenominator / ((HALF_CIRCUMFERENCE / 180) / 0.28e-3);
+    }
+
+    double dfXSize = (dfXMax - dfXMin) / dfRes;
+    double dfYSize = (dfYMax - dfYMin) / dfRes;
+    while (dfXSize > INT_MAX || dfYSize > INT_MAX)
+    {
+        dfXSize /= 2;
+        dfYSize /= 2;
+    }
+
+    nRasterXSize = std::max(1, static_cast<int>(0 + 5 + dfXSize));
+    nRasterYSize = std::max(1, static_cast<int>(0.5 + dfYSize));
+    m_adfGeoTransform[0] = dfXMin;
+    m_adfGeoTransform[1] = (dfXMax - dfXMin) / nRasterXSize;
+    m_adfGeoTransform[3] = dfYMax;
+    m_adfGeoTransform[5] = -(dfYMax - dfYMin) / nRasterYSize;
+
+    bRet = TRUE;
+
+    return bRet;
+}
+
+/************************************************************************/
 /*                        InitFromCollection()                          */
 /************************************************************************/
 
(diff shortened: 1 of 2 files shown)


rouault:
More fixes are needed to support OGC 2D Tile Matrix Set v2: cf https://github.com/OSGeo/gdal/issues/6882
And we must still support v1 because of other uses in GDAL

gcore/tilematrixset.cpp
```diff
@@ -77,7 +77,8 @@ std::unique_ptr<TileMatrixSet> TileMatrixSet::parse(const char *fileOrDef)
     std::unique_ptr<TileMatrixSet> poTMS(new TileMatrixSet());
 
     constexpr double HALF_CIRCUMFERENCE = 6378137 * M_PI;
-    if (EQUAL(fileOrDef, "GoogleMapsCompatible"))
+    if (EQUAL(fileOrDef, "http://www.opengis.net/def/tilematrixset/OGC/1.0/GoogleCRS84Quad")
```
rouault, line 80: the old names must still be supported, otherwise it breaks other things in GDAL. Cf https://app.travis-ci.com/github/OSGeo/gdal/jobs/601224478
```
________________________ ERROR collecting gcore/cog.py _________________________

gcore/cog.py:302: in <module>
    @gdaltest.require_creation_option("COG", "JPEG")
pymod/gdaltest.py:1590: in decorator
    drv is None or option not in drv.GetMetadata()["DMD_CREATIONOPTIONLIST"],
/usr/lib/python3/dist-packages/osgeo/gdal.py:2818: in GetMetadata
    return self.GetMetadata_Dict(domain)
/usr/lib/python3/dist-packages/osgeo/gdal.py:2789: in GetMetadata_Dict
    return _gdal.MajorObject_GetMetadata_Dict(self, *args)
E   RuntimeError: Invalid tiling matrix set name
```

gcore/tilematrixset.cpp
```diff
@@ -109,7 +110,7 @@ std::unique_ptr<TileMatrixSet> TileMatrixSet::parse(const char *fileOrDef)
         return poTMS;
     }
 
-    if (EQUAL(fileOrDef, "InspireCRS84Quad"))
+    if (EQUAL(fileOrDef, "http://www.opengis.net/def/tilematrixset/OGC/1.0/WorldCRS84Quad"))
```
rouault, line 113: same as above

doublebyte1:
> "InspireCRS84Quad")

Yes. I would like to address those on a different PR.

rouault:
> I would like to address #6882 on a different PR.

sounds good. Make sure to have commit hooks installed to have correct code formatting: cf https://gdal.org/development/dev_practices.html#commit-hooks

doublebyte1:
@rouault Thank you for the feedback!  🙏🏽 

With https://github.com/OSGeo/gdal/pull/7656/commits/13360c558d9a4d9697039989beb5fd1fe25a5ebb, I was able to run successfully all the tests in the test suite:

```
100% tests passed, 0 tests failed out of 28

Label Time Summary:
quicktest    = 156.70 sec*proc (19 tests)

Total Test time (real) = 156.91 sec
Process finished with exit code 0

```

gcore/tilematrixset.cpp
```diff
@@ -77,7 +77,12 @@ std::unique_ptr<TileMatrixSet> TileMatrixSet::parse(const char *fileOrDef)
     std::unique_ptr<TileMatrixSet> poTMS(new TileMatrixSet());
 
     constexpr double HALF_CIRCUMFERENCE = 6378137 * M_PI;
-    if (EQUAL(fileOrDef, "GoogleMapsCompatible"))
+    if (EQUAL(fileOrDef, "GoogleMapsCompatible") ||
+        EQUAL(fileOrDef, "http://www.opengis.net/def/tilematrixset/OGC/1.0/"
```
jerstlouis, line 81: @doublebyte1 `GoogleCRS84Quad` does not seem correct here.
That usually refers to the same TileMatrixSet as _WorldCRS84Quad_, except with the zoom levels offset by 1, with level 0 having the latitude going from -180..180, based on the _GoogleCRS84Quad_ Well-Known Scale Set specified in [WMTS 1.0 Annex E.3](https://portal.ogc.org/files/?artifact_id=35326):

https://maps.gnosis.earth/ogcapi/collections/blueMarble/map/tiles/GoogleCRS84Quad/0/0/0.jpg

image

I'm not sure if GDAL pre-defines this TileMatrixSet, but this section here is for spherical mercator (EPSG:3857).

_GoogleCRS84Quad_ is in EPSG:4326 like the _InspireCRS84Quad_ below, except it has that extra lower resolution level.

frmts/ogcapi/gdalogcapidataset.cpp
```diff
@@ -1670,7 +1670,16 @@ bool OGCAPIDataset::InitWithTilesAPI(GDALOpenInfo *poOpenInfo,
     if (!DownloadJSon(osTilingSchemeURL.c_str(), oDoc, nullptr,
                       MEDIA_TYPE_JSON))
         return false;
-    auto tms = gdal::TileMatrixSet::parse(oDoc.SaveAsString().c_str());
+
+    const auto uri = oDoc.GetRoot().GetString("uri");
+
+    if (uri.empty())
+    {
+        CPLError(CE_Failure, CPLE_AppDefined, "Cannot parse TMS uri");
```
jerstlouis, line 1678: Not sure what the logic is, but it is totally OK for a 2D TMS to not have a URI, so this should probably not fail here?
Only registered 2DTMS have a `uri`.

Also, I would also expect the URI-matching-based identification (for registered 2DTMSs) to have happened before using the `tileMatrixSetURI` property of the tileset instead. If the `tileMatrixSetURI` was a match for a registered 2DTMS, the client would not need to download the 2DTMS definition at all (the `DownloadJSon()` call above I think).

gcore/tilematrixset.cpp
doublebyte1: https://github.com/OSGeo/gdal/pull/7656/commits/5ae4b5948557678cfa7bb57a321a8922b4634903

frmts/ogcapi/gdalogcapidataset.cpp
doublebyte1: @jerstlouis you are right. There are other ways of parsing the TMS information. I will remove this check

jerstlouis: @doublebyte1 Yes essentially it could use the URI already checked at:

https://github.com/OSGeo/gdal/blob/master/frmts/ogcapi/gdalogcapidataset.cpp#L1489

```cpp
const auto oTileMatrixSetURI = oTileset.GetString("tileMatrixSetURI");
```

and if this URI is predefined (e.g., _WebMercatorQuad_ or _WorldCRS84Quad_), then it does not need to call `DownloadJSon(osTilingSchemeURL.c_str(), ...`.

This `InitWithTilesAPI()` method is very long though, so it would benefit from being split into different methods each performing one specific high level task.

gcore/tilematrixset.cpp
jerstlouis: @doublebyte1 Yes that solves that issue.

frmts/ogcapi/gdalogcapidataset.cpp
```diff
@@ -1670,7 +1695,12 @@ bool OGCAPIDataset::InitWithTilesAPI(GDALOpenInfo *poOpenInfo,
     if (!DownloadJSon(osTilingSchemeURL.c_str(), oDoc, nullptr,
                       MEDIA_TYPE_JSON))
         return false;
-    auto tms = gdal::TileMatrixSet::parse(oDoc.SaveAsString().c_str());
+
+    // Attempts to find the uri for a well-known TMS; if it does not work, it will send the entire document
+    const auto uri = oDoc.GetRoot().GetString("uri");
+
+    auto tms = gdal::TileMatrixSet::parse(
```
doublebyte1, line 1702: @jerstlouis for the time being, I am just checking if it can get an uri, and if not it defaults to what was there before (e.g.: sending the entire document). At least this is now working for well-known uris.

frmts/ogcapi/gdalogcapidataset.cpp
```diff
@@ -655,6 +658,44 @@ bool OGCAPIDataset::InitFromFile(GDALOpenInfo *poOpenInfo)
...
+
+    double dfXSize = (dfXMax - dfXMin) / dfRes;
+    double dfYSize = (dfYMax - dfYMin) / dfRes;
+    while (dfXSize > INT_MAX || dfYSize > INT_MAX)
+    {
+        dfXSize /= 2;
+        dfYSize /= 2;
+    }
+
+    nRasterXSize = std::max(1, static_cast<int>(0 + 5 + dfXSize));
```
rouault, line 687: Not introduced by you, but just spotted this error:
```suggestion
    nRasterXSize = std::max(1, static_cast<int>(0.5 + dfXSize));
```

frmts/ogcapi/gdalogcapidataset.cpp
```diff
@@ -655,6 +658,44 @@ bool OGCAPIDataset::InitFromFile(GDALOpenInfo *poOpenInfo)
     return InitFromCollection(poOpenInfo, oDoc);
 }
 
+/************************************************************************/
+/*                        ProcessScale()                          */
+/************************************************************************/
+
+bool OGCAPIDataset::ProcessScale(CPLJSONObject &oScaleDenominator,
```
rouault, line 665: Static analyzers will likely complain about bRet being set to FALSE, then TRUE, and the return value being checked and always TRUE.
I would just change the return type to be void

doublebyte1: It makes sense. I addressed it in this commit: https://github.com/OSGeo/gdal/pull/7656/commits/ca7772ab95e20f32c177be7ed324e3bc1a3aff77

frmts/ogcapi/gdalogcapidataset.cpp
doublebyte1: @jerstlouis I got that, but I don't want to address this refactoring in this PR.

PR #7673, opened 2023-05-02
GRIB: fix missing nodata on band > 1, fixes #7649

### LINKED ISSUE #7649: GRIB reporting wrong nodata value for band > 1

The correct nodata values for all the bands in a WMO GRIB image is 9999, and gdalinfo reports it right. 

But when calling the GDAL API in Python, for example, only the nodata value of the first band is correctly, the others return None, as we can see here:

```
>>> from osgeo import gdal
>>> ds = gdal.Open('/GRIB/2023/PS.557WW_SC.U_DI.C_GP.LIS_GR.C0P09DEG_AR.GLOBAL_PA.LIS24_DD.20230412_DT.1200_DF.GR2')
>>> for b in range(1,19):
...     print(ds.GetRasterBand(b).GetNoDataValue())
...
9999.0
None
None
None
None
None
None
None
None
None
None
None
None
None
None
None
None
None
```

Download a sample file from [Fast Downloading of GRIB Files](https://www.cpc.ncep.noaa.gov/products/wesley/fast_downloading_grib.html) and run the above sequence of code.

After some investigation, I noticed that if we call GetStatistics() before calling GetNoDataValue, the return values are correct. That is not a solution, but it might be a clue to fix the code.
 
```
>>> from osgeo import gdal
>>> ds = gdal.Open('/GRIB/2023/PS.557WW_SC.U_DI.C_GP.LIS_GR.C0P09DEG_AR.GLOBAL_PA.LIS24_DD.20230412_DT.1200_DF.GR2')
>>> for b in range(1,19):
...     s = ds.GetRasterBand(b).GetStatistics(True, False)
...     print(ds.GetRasterBand(b).GetNoDataValue())
...
9999.0
9999.0
9999.0
9999.0
9999.0
9999.0
9999.0
9999.0
9999.0
9999.0
9999.0
9999.0
9999.0
9999.0
9999.0
9999.0
9999.0
9999.0
```

That might explain why gdalinfo and gdalinfo.py always report the correct nodata value. Because they both call GetStatistics() before GetNoDataValue().

## Operating system

Linux (x86_64 Redhat) and Windows (Server 2016 64).

GDAL versions 3.4 and 3.6.

On Linux GDAL was build from source and on Windows it was installed via Anaconda 3.

That issues does not occurs in GDAL 2.3, not in Linux or Windows.

elpaso:
Fixes #7649.

I tried to create a small test case by clipping the test file provided in the comment https://github.com/OSGeo/gdal/issues/7649#issuecomment-1527858222 but the resulting file does not show the issue.

### DIFF
--- a/frmts/grib/gribdataset.cpp
+++ b/frmts/grib/gribdataset.cpp
@@ -1017,8 +1017,8 @@ double GRIBRasterBand::GetNoDataValue(int *pbSuccess)
 
     if (m_Grib_MetaData->gridAttrib.f_miss == 0)
     {
-        m_bHasNoData = false;
-        m_dfNoData = 0;
+        m_bHasLookedForNoData = false;
+        FindPDSTemplate();
         if (pbSuccess)
             *pbSuccess = m_bHasNoData;
         return m_dfNoData;


rouault:
Thanks for digging into this. Your fix was almost correct, with the missing extra precaution of calling FindPDSTemplate() only for GRIB2.
When looking at that deeper, I figured out a probably better fix by enhancing degrib's ReadGrib2Record() which is called by indirectly by GRIBRasterBand::ReadGribData()
Hence closing this in favor of https://github.com/OSGeo/gdal/pull/7674

Issue #7573, opened 2023-04-10
autotest: remove gdaltest.post_reason()


rouault:
Since the switch to pytest, gdaltest.post_reason() will have no visible effect. Any left over of it should be replaced with an assertion and this method should be removed.
Depends on #7572

PR #7676, opened 2023-05-02
Allow SetAttributeFilter, SetSpatialFilter, SetSpatialFilterRect to be used as context managers

dbaston:
## What does this PR do?

Allows `SetAttributeLayer`, `SetSpatialFilter`, and `SetSpatialFilterRect` to be used as context managers. Also allows `SetSpatialFilter` to handle WKT filters.

The intent is to avoid constructions like

```
geom = ogr.CreateGeometryFromWkt(wkt)
lyr.SetSpatialFilter(geom)
geom.Destroy()
tr = ogrtest.check_features_against_list(lyr, lst)
lyr.SetSpatialFilter(None)
assert tr
```

in favor of

```
with lyr.SetSpatialFilter(wkt):
  assert ogrtest.check_features_against_list(lyr, lst)
```

To retain the traditional usage, the proxy returned by `SetAttributeLayer` and `SetSpatialFilter` can be compared to an int, e.g.:

```
assert lyr.SetSpatialFilter(geom) == gdal.CE_None
```

I originally implemented helpers in `gdaltest` to accomplish the same thing (https://github.com/dbaston/gdal/commit/9c3331582928a9d7600377e2a2cddfa804e46d36) but decided to try doing it via SWIG instead. I'm not sure if it's a good idea or creates too much clutter. If it seems positive on the whole, I can complete with docstrings/tests/handling of multiple geometry fields.

## Tasklist

 - [x] Add test case(s)
 - [x] Add documentation
 - [ ] Review
 - [x] Adjust for comments

### DIFF
--- a/autotest/ogr/ogr_sql_test.py
+++ b/autotest/ogr/ogr_sql_test.py
@@ -130,14 +130,13 @@ def test_ogr_sql_1():
     gdaltest.ds = ogr.Open("data")
     gdaltest.lyr = gdaltest.ds.GetLayerByName("poly")
 
-    gdaltest.lyr.SetAttributeFilter("eas_id < 167")
+    with gdaltest.lyr.SetAttributeFilter("eas_id < 167"):
 
-    count = gdaltest.lyr.GetFeatureCount()
-    assert count == 3, (
-        "Got wrong count with GetFeatureCount() - %d, expecting 3" % count
-    )
+        count = gdaltest.lyr.GetFeatureCount()
+        assert count == 3, (
+            "Got wrong count with GetFeatureCount() - %d, expecting 3" % count
+        )
 
-    gdaltest.lyr.SetAttributeFilter("")
     count = gdaltest.lyr.GetFeatureCount()
     assert count == 10, (
         "Got wrong count with GetFeatureCount() - %d, expecting 10" % count
@@ -1167,9 +1166,8 @@ def test_ogr_sql_41():
 def test_ogr_sql_42():
 
     lyr = gdaltest.ds.GetLayerByName("poly")
-    lyr.SetAttributeFilter("prfedea <> ''")
-    feat = lyr.GetNextFeature()
-    lyr.SetAttributeFilter(None)
+    with lyr.SetAttributeFilter("prfedea <> ''"):
+        feat = lyr.GetNextFeature()
     assert feat is not None
 
     with gdaltest.ds.ExecuteSQL("SELECT * FROM poly WHERE prfedea <> ''") as sql_lyr:
@@ -1496,20 +1494,20 @@ def test_ogr_sql_attribute_filter_on_top_of_non_forward_where_clause(dialect):
     with mem_ds.ExecuteSQL(
         "SELECT * FROM test WHERE OGR_GEOMETRY = 'POLYGON'", dialect=dialect
     ) as sql_lyr:
-        sql_lyr.SetAttributeFilter("")
-        assert sql_lyr.GetFeatureCount() == 1
+        with sql_lyr.SetAttributeFilter(""):
+            assert sql_lyr.GetFeatureCount() == 1
 
     with mem_ds.ExecuteSQL(
         "SELECT * FROM test WHERE OGR_GEOMETRY = 'POLYGON'", dialect=dialect
     ) as sql_lyr:
-        sql_lyr.SetAttributeFilter("1")
-        assert sql_lyr.GetFeatureCount() == 1
+        with sql_lyr.SetAttributeFilter("1"):
+            assert sql_lyr.GetFeatureCount() == 1
 
     with mem_ds.ExecuteSQL(
         "SELECT * FROM test WHERE OGR_GEOMETRY = 'POLYGON'", dialect=dialect
     ) as sql_lyr:
-        sql_lyr.SetAttributeFilter("0")
-        assert sql_lyr.GetFeatureCount() == 0
+        with sql_lyr.SetAttributeFilter("0"):
+            assert sql_lyr.GetFeatureCount() == 0
 
 
 ###############################################################################
--- a/swig/include/python/ogr_python.i
+++ b/swig/include/python/ogr_python.i
@@ -482,6 +482,216 @@ def ReleaseResultSet(self, sql_lyr):
 
   %}
 
+  %feature("shadow") SetAttributeFilter %{
+  def SetAttributeFilter(self, filter_string: "Optional[str]"):
+      """SetAttributeFilter(self, filter_string: Optional[str])
+
+      Set the attribute query string to be used when fetching features.
+
+      See :cpp:func:`OGRLayer::SetAttributeFilter` for more information on
+      the filter statement format.
+
+      Starting with GDAL 3.8, this method can be used as a context manager,
+      automatically disabling the attribute filter when exiting the context.
+
+      Parameters
+      ----------
+      filter_string:
+        filter statement in the format of a SQL WHERE clause
+
+      Returns
+      -------
+      a proxy object that can be used as a context manager, or can be compared to an OGRErr value.
+
+      Examples
+      --------
+
+      1. Use as a context manager:
+
+      >>> with lyr.SetAttributeFilter("FID = 5"):
+      ...     print(lyr.GetFeatureCount())
+
+      2. Used to set the filter until it is explicitly disabled.
+
+      >>> err = lyr.SetAttributeFilter("FID = 5")
+      ... try:
+      ...     assert err = gdal.CE_None
+      ... finally:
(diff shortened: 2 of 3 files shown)


swig/include/python/ogr_python.i
```diff
@@ -482,6 +482,208 @@ def ReleaseResultSet(self, sql_lyr):
...
+      filter = args[-1]
+
+      class SpatialFilterProxy:
+          def __init__(self, lyr, geom_field, filter):
+              self.lyr = lyr
+              self.geom_field = geom_field
+
+              if filter is None:
+                filter_geom = None
+              if type(filter) is str:
```
rouault, line 594:
```suggestion
              elif type(filter) is str:
```

swig/include/python/ogr_python.i
```diff
@@ -482,6 +482,208 @@ def ReleaseResultSet(self, sql_lyr):
...
+                filter_geom = CreateGeometryFromWkt(filter)
+              else:
+                filter_geom = filter
+
+              self.err = $action(lyr, geom_field, filter_geom)
+
+              if type(filter) is str:
+                 filter_geom.Destroy()
```
rouault, line 602: explicit destruction of ogr.Geometry instances is not needed (dealt by the SWIG proxy class)

```suggestion
```

swig/include/python/ogr_python.i
```diff
@@ -482,6 +482,208 @@ def ReleaseResultSet(self, sql_lyr):
...
+      -------
+      a proxy object that can be used as a context manager, or can be compared
+      to an OGRErr value.
+
+      Examples
+      --------
+
+      1. Use as a context manager:
+
+      >>> with lyr.SetSpatialFilterRect(0, 10, 10, 20, 20)"):
```
rouault, line 653:
```suggestion
      >>> with lyr.SetSpatialFilterRect(0, 10, 10, 20, 20):
```

swig/include/python/ogr_python.i
```diff
@@ -482,6 +482,214 @@ def ReleaseResultSet(self, sql_lyr):
 
   %}
 
+  %feature("shadow") SetAttributeFilter %{
+  def SetAttributeFilter(self, filter_string: "Optional[str]"):
+      """SetAttributeFilter(self, filter_string: Optional[str])
+
+      Set the attribute query string to be used when fetching features.
+
+      See :cpp:func:`OGRLayer::SetAttributeFilter` for more information on
```
dbaston, line 491: Seems a little weird to put RST stuff in the docstrings, but I'm not sure of another way to make this link.

rouault:
Cf failure on ubuntu 18.04 gcc 32 bit (actually any build with -DCMAKE_BUILD_TYPE=Debug to activate the OGR API SPY functionality): https://github.com/OSGeo/gdal/actions/runs/4866524539/jobs/8678150749?pr=7676 (
- Either you adjust data/testograpispy.py to use SetSpatialFilter() with an explicit geom index
- Or you modify the context manager to respect the number of argument of the caller (ie if called without a geometry index, you call the underlying method without a geometry index)

sgillies:
@dbaston you questioned whether this is a good idea. I don't think it is. Python's `with` is intended to make it possible to factor out standard uses of try/finally statements. So that in the case below, if `func` raises an exception, the allocated external resources can be deallocated.

```python
with setup_external_resources():
    func()
```

What is the value to a user in letting a layer's filter reset itself? It feels unusual to me. If this is being done to help make assertions against different WKT in a single test, don't do that. Break that single test up into multiple tests.

autotest/ogr/ogr_sql_test.py
```diff
@@ -130,14 +130,13 @@ def test_ogr_sql_1():
...
-    count = gdaltest.lyr.GetFeatureCount()
-    assert count == 3, (
-        "Got wrong count with GetFeatureCount() - %d, expecting 3" % count
-    )
+        count = gdaltest.lyr.GetFeatureCount()
+        assert count == 3, (
+            "Got wrong count with GetFeatureCount() - %d, expecting 3" % count
+        )
 
-    gdaltest.lyr.SetAttributeFilter("")
```
sgillies, line 140: @dbaston if test_ogr_sql_1 were split into two tests, instead of combining two cases into one, reset of the layer's attribute filter would be a non-issue.

dbaston:
> Python's with is intended to make it possible to factor out standard uses of try/finally statements

That's what motivated me to write this. I took  `ogrtest.check_features_against_list` refactored it to use `assert` instead of returning `0` or `1`(#7573), so I ended up taking code like

```
lyr.SetAttributeFilter(x)
tr = ogrtest.check_features_against_list(lyr, features)
lyr.SetAttributeFilter(None)
assert tr
```

and changing it to

```
lyr.SetAttributeFilter(x)
try:
    ogrtest.check_features_against_list(lyr, features)
finally:
    lyr.SetAttributeFilter(None)
````

and I thought a context handler would be a good way to avoid repeated try/finally blocks.

> If this is being done to help make assertions against different WKT in a single test, don't do that. Break that single test up into multiple tests.

It's being done to try to incrementally improve the existing tests. The problem is more than just multiple assertions within a single test, we also have many tests sharing the same layer reference.

**Comment by sgillies** [comment]
> The problem is more than just multiple assertions within a single test, we also have many tests sharing the same layer reference.

@dbaston I think that's better solved by making that layer reference into a fixture? No need for API changes or new documentation.

**Comment by rouault** [comment]
> and I thought a context handler would be a good way to avoid repeated try/finally blocks.

if ogrtest.check_features_against_list() is changed to use assertions, then you don't need to reset the attribute filter state before the exception is raised to exit the method. There's nothing bad in having a OGR layer / dataset destroyed with the attribute filter set at a None value.

Beyond autotest purposes, I thought this could be nice for use cases like
```
with lyr.SetAttributeFilter("type = 'road'"):
    print(lyr.GetFeatureCount())
```
where you want to install a filter in a transient way without impacting the state of the layer object for subsequent calls, but I hear @sgillies objection this might not be Python idiomatic.

Side remark: fundamentally, the main issue is that the OGRLayer object mixes both things that are management of its serialized state (setting/getting layer structure and metadata, adding/removing/etc. fields, features, etc.) that are definitely appropriate for it, and things that would probably belong more to a OGRFeatureRequest class where you could set filters and get a feature iterator. But at this stage of the project, it is probably impossible / too much of an effort to do such changes that would impact both the API and would require very significant rework of drivers.

**Comment by dbaston** [comment]
> @dbaston I think that's better solved by making that layer reference into a fixture? No need for API changes or new documentation.

Using a fixture is a better long-term solution, and helps us achieve the goal of independent tests that can be run in parallel. But it's hard to do incrementally. Making the existing tests easier to understand helps prepare for changes like removing shared objects that would affect many tests.

This doesn't have to be an API change, we could do https://github.com/dbaston/gdal/commit/9c3331582928a9d7600377e2a2cddfa804e46d36 instead. Adding it as a member function to `ogr.Layer` just makes the usage a bit less verbose.

**Comment by rouault** [comment]
> This doesn't have to be an API change, we could do [dbaston@9c33315](https://github.com/dbaston/gdal/commit/9c3331582928a9d7600377e2a2cddfa804e46d36) instead

That could a good compromise. We can do whatever we want if it stays in autotest "namespace"

**Comment by dbaston** [comment]
> Regarding test fixtures, I believe the objects they return is created just once, right? If so, then yes we are in trouble if we don't reset the state.

We can control that by setting the fixture scope (https://docs.pytest.org/en/6.2.x/fixture.html#scope-sharing-fixtures-across-classes-modules-packages-or-session). It would be nice to always use function scope to have "clean" objects. I don't have any sense of what the performance impact is over a test suite as large as this one. And we still have to deal with the tests that depend on changes made to a shared object during previous tests (trivial example: `test_ogr_shape_2` and `test_ogr_shape_3`)

**Comment by rouault** [comment]
> I don't have any sense of what the performance impact is over a test suite as large as this one

it really depends. There are situations where opening the dataset might be super cheap (most file based drivers) and others where it is not (network access, database, etc.). We clearly don't want the test suite to be 10x times longer to execute but if it is a 10% time penaly, that's OK.

> And we still have to deal with the tests that depend on changes made to a shared object during previous tests 

I guess we could do merge some of those non-dependent tests in a single one like:
```
def test_lots_of_stuff_here():

  ds = ....

  def step1():
        ....

  step1()

  def step2():
         ...

  step2()

```
There might be more elegant solutions...

dbaston:
> > This doesn't have to be an API change, we could do [dbaston@9c33315](https://github.com/dbaston/gdal/commit/9c3331582928a9d7600377e2a2cddfa804e46d36) instead
> 
> That could a good compromise. We can do whatever we want if it stays in autotest "namespace"

Would this sit better with you, @sgillies ?

sgillies:
@dbaston much better. Thanks for listening!

PR #7681, opened 2023-05-02
TileDB tile extents configuration

normanb:
## What does this PR do?

When reviewing more of the TileDB Vector driver I noticed that `TILE_EXTENT` was missing from the driver documentation so I added that and at the same time added a separate configuration for `Z` as `Z` is typically narrower than `X` or `Y`. I also removed some legacy comments in the code related to schema creation with tile sizes.

https://github.com/OSGeo/gdal/pull/7615

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments


frmts/tiledb/tiledbcommon.cpp
```diff
@@ -444,6 +444,10 @@ void GDALRegister_TileDB()
         "description='Number of non-empty cells stored in a data tile'/>"
         "   <Option name='BOUNDS' type='string' description='Specify "
         "bounds for sparse array, minx, miny, [minz,] maxx, maxy [, maxz]'/>\n"
+        "   <Option name='TILE_EXTENT' type='float' description='Specify "
```
rouault, line 447: doc/source/drivers/vector/tiledb.rst should also be updated

PR #7682, opened 2023-05-02
Doc: Remove outdated information on building Python bindings

dbaston:
## What does this PR do?

Removes outdated information on building Python bindings with autoconf.

The "Windows" section (https://gdal.org/api/python_bindings.html#windows)  also appears to be very out-of-date, but I don't have the knowledge to update it.


rouault:
> The "Windows" section (https://gdal.org/api/python_bindings.html#windows) also appears to be very out-of-date, but I don't have the knowledge to update it.

It should probably entirely removed.

The "Unix" paragraph should be renamed something like "Generalities", and the mention of distutils in it should be removed as only setuptools is now supported (distutils by itself has been removed in recent python versions)

dbaston:
> It should probably entirely removed.

Done

> The "Unix" paragraph should be renamed something like "Generalities", and the mention of distutils in it should be removed as only setuptools is now supported (distutils by itself has been removed in recent python versions)

Removed this paragraph too, since `setuptools` is only discussed to contrast with `distutils`.

PR #7685, opened 2023-05-03
autotest: Fix incomplete downloads

dbaston:
## What does this PR do?

Fixes `gdaltest.download_file` so that files are completely downloaded even if the size cannot be determined from the `content-length` header. An example file that is not correctly downloaded is  `http://www2.census.gov/geo/tiger/tiger2006se/AL/TGR01001.ZIP` in `ogr/ogr_tiger.py`.

None

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments


dbaston:
While I was in there I collapsed two levels of indentation. Whitespace-ignoring diff at https://github.com/OSGeo/gdal/pull/7685/files?w=1.

PR #7686, opened 2023-05-03
autotest: skip tests requiring unavailable downloads

dbaston:
The URLs used in these tests redirect to HTML pages that are successfully downloaded, causing the tests to fail.

With the exception of the NASA file noted in the comments, I was not able to locate new URLs or turn up anything in Wayback Machine.


rouault:
I do actually have locally some of those old files, that I preciously keep from past times. Unfortunately hard to put them online due to unknown licensing. Perhaps change the skip to skipif(not os.path.exists("tmp/cache/...."), reason="....") ?

rouault:
I've now uploaded my local cache file to http://download.osgeo.org/gdal/data/netcdf/A2016273115000.L2_LAC_OC.nc

PR #7706, opened 2023-05-07
Zarr: update V3 format to latest specification

### LINKED ISSUE #7637: Zarr V3 out of date with latest spec
The experimental Zarr V3 support is no longer up-to-date with the latest specification, and is no longer interoperable with datasets produced for example with https://github.com/alimanfoo/zarrita

Cf https://zarr-specs.readthedocs.io/en/latest/v3/core/v3.0.html#draft-changes

rouault:
Fixes #7637.
On top of PR #7684. only commits starting with "[Zarr: move code belonging to classes to dedicated files](https://github.com/OSGeo/gdal/commit/60477b4958b2633a868de687d33d5c82672954aa)" belong to this PR

- Support core V3 specification, as well as gzip, blosc, endian and transpose codecs (sharding codec excluding then)
- implement group, array, attribute and dimension renaming

This breaks backwards / upwards compatibility with previous GDAL versions: Zarr V3 datasets generated by GDAL < 3.8 can't be read by GDAL >= 3.8dev, and ZarrV3 datasets generated by GDAL >= 3.8dev can't be read by GDAL < 3.8


frmts/zarr/zarr.h
```diff
@@ -812,24 +948,478 @@ class ZarrArray final : public GDALPamMDArray
...
+{
+    CPLStringList m_aosCompressorOptions{};
+    const CPLCompressor *m_pDecompressor = nullptr;
+    const CPLCompressor *m_pCompressor = nullptr;
+
+    ZarrV3CodecGZip(const ZarrV3CodecGZip &) = delete;
+    ZarrV3CodecGZip &operator=(const ZarrV3CodecGZip &) = delete;
+
+  public:
+    static constexpr const char *NAME = "gzip";
```
elpaso, line 1149: isn't `static const` redundant if `constexpr` is defined?

frmts/zarr/zarr.h
```diff
@@ -812,24 +948,478 @@ class ZarrArray final : public GDALPamMDArray
...
+{
+    CPLStringList m_aosCompressorOptions{};
+    const CPLCompressor *m_pDecompressor = nullptr;
+    const CPLCompressor *m_pCompressor = nullptr;
+
+    ZarrV3CodecBlosc(const ZarrV3CodecBlosc &) = delete;
+    ZarrV3CodecBlosc &operator=(const ZarrV3CodecBlosc &) = delete;
+
+  public:
+    static constexpr const char *NAME = "blosc";
```
elpaso, line 1192: same, and other occurrences below

frmts/zarr/zarr.h
```diff
@@ -812,24 +948,478 @@ class ZarrArray final : public GDALPamMDArray
...
+                ZarrByteVectorQuickResize &abyDst) const override;
+};
+
+/************************************************************************/
+/*                         ZarrV3CodecTranspose                         */
+/************************************************************************/
+
+class ZarrV3CodecTranspose final : public ZarrV3Codec
+{
+    // m_anOrder, such that dest_shape[i] = source_shape[order[i]]
```
elpaso, line 1279: leftover?

frmts/zarr/zarr_v2_array.cpp
```diff
@@ -0,0 +1,1968 @@
...
+        }
+        else
+        {
+            CPLError(CE_Warning, CPLE_AppDefined,
+                     "Size of _NCZARR_ARRAY.dimrefs different from the one of "
+                     "shape");
+        }
+    }
+
+    const char *dtypeKey = "dtype";
```
elpaso, line 1639: maybe constexpr?

elpaso (approved):
I am not familiar with the format but this LGTM.

My only concern is about the rename operations implementations of the various objects, it seems to me there is quite a lot of similar or identical code: would it be possible to use a template or an abstract base class to reduce duplicated code?

autotest/gdrivers/memmultidim.py
```diff
@@ -2567,6 +2567,191 @@ def test_mem_md_resize_dim_referenced_twice_error():
...
+    )
+    subgroup_ar = subgroup.CreateMDArray(
+        "subgroup_ar", [dim], gdal.ExtendedDataType.Create(gdal.GDT_Byte)
+    )
+    subgroup_ar_attr = subgroup_ar.CreateAttribute(
+        "subgroup_ar_attr", [], gdal.ExtendedDataType.Create(gdal.GDT_Byte)
+    )
+
+    # Cannot rename root group
+    with pytest.raises(Exception):
```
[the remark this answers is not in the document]
rouault: I prefer keeping the generic Exception as I'm not sure the particular one raised currently by the bindings is appropriate, and if we changed that later that could become annoying to change the autotest

frmts/zarr/zarr.h
rouault: just a comment attempt :-)
I've clarified it to
```
    // m_anOrder is such that dest_shape[i] = source_shape[m_anOrder[i]]
    // where source_shape[] is the size of the array before the Encode() operation
    // and dest_shape[] its size after.
    // m_anOrder[] describes a bijection of [0,N-1] to [0,N-1]
```

rouault:
> My only concern is about the rename operations implementations of the various objects, it seems to me there is quite a lot of similar or identical code: would it be possible to use a template or an abstract base class to reduce duplicated code?

I've tried to reduce amount of copied&pasted code in c81d12fa7180e058a5d929a5c278044c11bd4c0a

PR #7710, opened 2023-05-08
Multidim related fixes

rouault:
- Multidim GDALGroup::CopyFrom(): modify logic to copy first indexing variables. Fixes issue with gdalmdimtranslate from Zarr to netCDF
- netCDF multidim: fix issues with data vs definition mode


rouault:
manually backported in https://github.com/OSGeo/gdal/pull/7731

PR #7736, opened 2023-05-10
COG: only update mode if the IGNORE_COG_LAYOUT_BREAK=YES open option is specified (fixes #7735)

### LINKED ISSUE #7735: gdal_edit.py breaks COG layout and warns only after the damage is done

Using `gdal_edit.py -setstats` on a cloud-optimized geotiff. I expect it to either:
- set stats
- output an error and quit
- if I'm doing a risky operation, warn me and ask if I would like to proceed

Instead, it destroys my file and then says `Warning 1: /app/media/images/M1180855200RE_cog.tif: The IFD has been rewritten at the end of the file, which breaks COG layout.`

1. Use gdal_edit.py with -setstats on a tif with COG layout

## Operating system

Windows and Ubuntu

GDAL 3.4.1

rouault:
Updating a COG file generally breaks part of the optimizations, but still produce a valid GeoTIFF file. Starting with GDAL 3.8, to avoid undesired loss of the COG characteristics, opening such a file in update mode will be rejected, unless the IGNORE_COG_LAYOUT_BREAK open option is also explicitly set to YES.

Previously, we already got a warning that optimizations were broken, but this was after the fact, which might be frustrating for the user that needs to regenerate the COG.

CC @vincentsarago good idea ? bad idea ?

### DIFF
--- a/doc/source/drivers/raster/cog.rst
+++ b/doc/source/drivers/raster/cog.rst
@@ -314,7 +314,7 @@ Update
 ------
 
 Updating a COG file generally breaks part of the optimizations, but still
-produce a valid GeoTIFF file. Starting with GDAL 3.8, to avoid undesired loss
+produces a valid GeoTIFF file. Starting with GDAL 3.8, to avoid undesired loss
 of the COG characteristics, opening such a file in update mode will be rejected,
 unless the IGNORE_COG_LAYOUT_BREAK open option is also explicitly set to YES.
 
--- a/doc/source/drivers/raster/cog.rst
+++ b/doc/source/drivers/raster/cog.rst
@@ -310,6 +310,17 @@ Reprojection related creation options
 - **ADD_ALPHA=YES/NO**: Whether an alpha band is added in case of reprojection.
   Defaults to YES.
 
+Update
+------
+
+Updating a COG file generally breaks part of the optimizations, but still
+produce a valid GeoTIFF file. Starting with GDAL 3.8, to avoid undesired loss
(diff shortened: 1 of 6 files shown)


vincentsarago:
thanks for asking @rouault 

To me, we already say in the DOCs that a COG is a `CreateCopy` only format so yes it makes sense to have something that will raise an error when trying to edit it or open it in `write` mode. 

IMO, `ALLOW_COG_UPDATE` might be a bit ambiguous and could led people to think that if you had `ALLOW_COG_UPDATE=YES` you could actually update a COG without breaking it. maybe something like `IGNORE_COG_LAYOUT_BREAK` might be more explicit, but I'm not the best at choosing english words 😄.

rouault:
open option renamed to IGNORE_COG_LAYOUT_BREAK

doc/source/drivers/raster/cog.rst
```diff
@@ -310,6 +310,17 @@ Reprojection related creation options
 - **ADD_ALPHA=YES/NO**: Whether an alpha band is added in case of reprojection.
   Defaults to YES.
 
+Update
+------
+
+Updating a COG file generally breaks part of the optimizations, but still
+produce a valid GeoTIFF file. Starting with GDAL 3.8, to avoid undesired loss
```
elpaso, line 317:
```suggestion
produces a valid GeoTIFF file. Starting with GDAL 3.8, to avoid undesired loss
```

elpaso (approved):

PR #7737, opened 2023-05-10
netCDF: optimize loading of network located dataset with unlimited dimension

rouault:
Helps for https://lists.osgeo.org/pipermail/gdal-dev/2023-May/057209.html

The time dimension is unlimited, and is the one used to determine GDAL bands. Retrieving the values of the time variable for the metadata requires scanning a lot of places within the file, which is network access unfriendly.
So:
- do not set the NETCDF_DIM_xxxx_VALUES dataset metadata item for such use case
- and defer loading of the NETCDF_DIM_xxxx band metadata item until it is really required

All that combines enables access of a given band with a more reasonable amount of HTTP access.
- 85 requests for ``gdalwarp "vrt://netCDF:\"/vsicurl/http://localhost:8080/tos_day_ACCESS1-3_rcp85_r1i1p1_20560101-20651231.nc\":tos?bands=1" out2.tif -overwrite``
- 19 requests for same invokation with GDAL_NETCDF_BOTTOMUP=NO env variable set  (the output is slightly different, but that must be a side effect of the geoloc transformer)

### DIFF
--- a/frmts/netcdf/netcdfdataset.cpp
+++ b/frmts/netcdf/netcdfdataset.cpp
@@ -9624,15 +9624,28 @@ GDALDataset *netCDFDataset::Open(GDALOpenInfo *poOpenInfo)
                                  szDimName);
                         poDS->papszMetadata = CSLSetNameValue(
                             poDS->papszMetadata, szTemp, szExtraDimDef);
-                        char *pszTemp = nullptr;
-                        if (NCDFGet1DVar(nIdxGroupID, nIdxVarID, &pszTemp) ==
-                            CE_None)
+
+                        // Retrieving data for unlimited dimensions might be
+                        // costly on network storage, so don't do it.
+                        // Each band will capture the value along the extra
+                        // dimension in its NETCDF_DIM_xxxx band metadata item
+                        // Addresses use case of
+                        // https://lists.osgeo.org/pipermail/gdal-dev/2023-May/057209.html
+                        if (VSIIsLocal(osFilenameForNCOpen.c_str()) ||
+                            !NCDFIsUnlimitedDim(poDS->eFormat ==
+                                                    NCDF_FORMAT_NC4,
+                                                cdfid, poDS->m_anDimIds[j]))
                         {
-                            snprintf(szTemp, sizeof(szTemp),
-                                     "NETCDF_DIM_%s_VALUES", szDimName);
-                            poDS->papszMetadata = CSLSetNameValue(
-                                poDS->papszMetadata, szTemp, pszTemp);
-                            CPLFree(pszTemp);
+                            char *pszTemp = nullptr;
+                            if (NCDFGet1DVar(nIdxGroupID, nIdxVarID,
+                                             &pszTemp) == CE_None)
+                            {
+                                snprintf(szTemp, sizeof(szTemp),
+                                         "NETCDF_DIM_%s_VALUES", szDimName);
+                                poDS->papszMetadata = CSLSetNameValue(
+                                    poDS->papszMetadata, szTemp, pszTemp);
+                                CPLFree(pszTemp);
+                            }
                         }
                     }
                 }
--- a/frmts/netcdf/netcdfdataset.cpp
+++ b/frmts/netcdf/netcdfdataset.cpp
@@ -234,9 +234,7 @@ class netCDFRasterBand final : public GDALPamRasterBand
     bool bSignedData;
     bool bCheckLongitude;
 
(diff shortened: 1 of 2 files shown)


frmts/netcdf/netcdfdataset.cpp
```diff
@@ -1987,14 +2055,15 @@ CPLErr netCDFRasterBand::CreateBandMetadata(const int *paDimIds,
...
+        snprintf(szName, sizeof(szName), "%s",
+                 l_poDS->papszDimName[l_poDS->m_anDimIds[panBandZPos[i]]]);
 
         char szMetaName[NC_MAX_NAME + 1 + 32];
-        snprintf(szMetaName, sizeof(szMetaName), "NETCDF_DIM_%s", szVarName);
+        snprintf(szMetaName, sizeof(szMetaName), "NETCDF_DIM_%s", szName);
 
-        int nGroupID = panExtraDimGroupIds[i];
-        int nVarID = panExtraDimVarIds[i];
+        int nGroupID = l_poDS->m_anExtraDimGroupIds[i];
```
elpaso, line 2065:
```suggestion
        const int nGroupID = l_poDS->m_anExtraDimGroupIds[i];
```

frmts/netcdf/netcdfdataset.cpp
```diff
@@ -1987,14 +2055,15 @@ CPLErr netCDFRasterBand::CreateBandMetadata(const int *paDimIds,
...
+        int nVarID = l_poDS->m_anExtraDimVarIds[i];
```
elpaso, line 2066:
```suggestion
        const int nVarID = l_poDS->m_anExtraDimVarIds[i];
```

frmts/netcdf/netcdfdataset.cpp
```diff
@@ -1960,19 +2029,18 @@ CPLErr netCDFRasterBand::CreateBandMetadata(const int *paDimIds,
...
         Sum *= panBandZLev[0];
     }
 
     // Loop over non-spatial dimensions.
-    int result = 0;
     int Taken = 0;
 
     for (int i = 0; i < nd - 2; i++)
     {
+        int result;
```
elpaso, line 2043: isn't `result` being used uninitialized if the following condition is false?

rouault: no, it is also initialized in the else branch at line 2055

frmts/netcdf/netcdfdataset.cpp
rouault: done

frmts/netcdf/netcdfdataset.cpp
rouault: done

mdsumner:
this is looking good, I'm not getting blocked from usage now :0

one side question, when I don't use the '/vsicurl' prefix on the bare url I get these vsimem truncated subdataset forms: 

```
gdalinfo https://github.com/OSGeo/gdal/raw/master/autotest/gdrivers/data/netcdf/var_with_column.nc

Files: none associated
Size is 512, 512
Subdatasets:
  SUBDATASET_1_NAME=NETCDF:"/vsimem/http_1/var_with_column.nc":"VAR:NAME"
  SUBDATASET_1_DESC=[2x2] VAR:NAME (32-bit floating-point)

```

with the prefix it's correct

```
gdalinfo /vsicurl/https://github.com/OSGeo/gdal/raw/master/autotest/gdrivers/data/netcdf/var_with_column.nc
Driver: netCDF/Network Common Data Format
Files: /vsicurl/https://github.com/OSGeo/gdal/raw/master/autotest/gdrivers/data/netcdf/var_with_column.nc
Size is 512, 512
Subdatasets:
  SUBDATASET_1_NAME=NETCDF:"/vsicurl/https://github.com/OSGeo/gdal/raw/master/autotest/gdrivers/data/netcdf/var_with_column.nc":"VAR:NAME"

```

is that just a "don't use bare url" situation?

rouault:
> ```
> 
> ```
> 
> 
>     
>   
> 
> is that just a "don't use bare url" situation?

yes, the HTTP driver is a bit of a hack, and is for one-time usage of small remote files, as it creates either a in-memory (or on-disk as fallback) temporary file with the whole content of the file before passing it to any other driver that accepts it. This doesn't fly well with subdatasets, at least in a command line usage. Within the same process, and if you keep the original dataset opened, opening the subdatasets could potentially work, but I'm not totally sure.

mdsumner:
awesome, as always much appreciated 🤘

Issue #4272, opened 2021-08-19
Block cache scalability issues


rouault:
This ticket is just to remind that the current design of the global block cache causes important thread-scalability issues, in particular when writing of datasets is involved. https://github.com/OSGeo/gdal/issues/3848 is an example where there's a severe lock contention of the block cache which makes most cores waiting for others

tbonfort:
(reposting this here after having erroneously added it to #3982)
I'm not sure if this is the place to propose this, and/or if this is already implemented, but there could also be a quick win to reduce contention in the block cache: when a "write" happens on a whole block (i.e. on a single band block for band-interleaved datasets, or on interleaved data of all bands for pixel-interleaved datasets), there is in my opinion no use for that block to get pushed into the block cache so that part could be skipped entirely.

PR #7756, opened 2023-05-12
[Perf improvement] GTiff: avoid using block cache when writing whole blocks

rouault:
This avoids less contention on the block cache mutex and saves memory
copies, particular when writing dataset-level RasterIO() with a buffer
with data for all bands, and with a pixel interleaving.

The optimization triggers if the following conditions are met:
- (xsize, ysize) = (bufxsize, bufysize), ie no resampling
- origin (xoff, yoff) of buffer to write is aligned on a tile boundary
- end (xoff+xsize, yoff+ysize) of buffer is aligned on tile boundary or
  raster right/bottom
- only for bit depth = 8, 16, 32, 64 or 128
- the block cache for the bands to considered has not been initialized
  at all
- and for dataset level write with pixel interleaving, all bands are
  written at once.

The visible side effects of this optimization are hopefully few, but one
particular case exhibited by the test_tiff_ovr_10() case is when writing
with a lossy compression and re-reading the raster without closing it in
between. Previously if the raster was small enough to fit in the block
cache, re-reading it would get the original pixel values. With this
change, we now get the values after decompression.

Using the gtiff_multi_ds_parallel_write.py program to write 2048x2048
tiled datasets in parallel from 12 threads, ZSTD compression

$ time python ../perftests/gtiff_multi_ds_parallel_write.py --nbands 1 --without-optim
real    0m17,768s
user    2m55,418s
sys     0m1,174s

$ time python ../perftests/gtiff_multi_ds_parallel_write.py --nbands 1
real    0m9,149s
user    1m34,882s
sys     0m0,597s

$ time python ../perftests/gtiff_multi_ds_parallel_write.py --nbands 3 --buffer-interleave BAND --without-optim
real    0m16,106s
user    2m42,623s
sys     0m0,834s

$ time python ../perftests/gtiff_multi_ds_parallel_write.py --nbands 3 --buffer-interleave BAND
real    0m9,013s
user    1m31,537s
sys     0m0,329s

$ time python ../perftests/gtiff_multi_ds_parallel_write.py --nbands 3 --buffer-interleave PIXEL --without-optim
real    0m41,574s
user    3m23,985s
sys     3m45,286s

$ time python ../perftests/gtiff_multi_ds_parallel_write.py --nbands 3 --buffer-interleave PIXEL
real    0m5,893s
user    0m56,796s
sys     0m0,402s

Helps for #4272
CC @tbonfort

### DIFF
--- a/autotest/gcore/tiff_ovr.py
+++ b/autotest/gcore/tiff_ovr.py
@@ -444,7 +444,7 @@ def test_tiff_ovr_10(both_endian):
 
     assert src_ds is not None, "Failed to open test dataset."
 
-    ds = gdaltest.tiff_drv.CreateCopy(
+    ds = gdal.GetDriverByName("GTiff").CreateCopy(
         "tmp/ovr10.tif", src_ds, options=["COMPRESS=JPEG", "PHOTOMETRIC=YCBCR"]
     )
     src_ds = None
@@ -463,9 +463,8 @@ def test_tiff_ovr_10(both_endian):
     ds = None
 
     assert cs in (
-        5562,
-        5635,
-        5601,  # libjpeg 9e
+        5879,
+        6050,  # libjpeg 9e
     )
 
 
@@ -942,7 +941,7 @@ def test_tiff_ovr_25(both_endian):
 
 def test_tiff_ovr_26(both_endian):
 
-    ds = gdaltest.tiff_drv.Create("tmp/ovr26.tif", 100, 100, 1)
+    ds = gdal.GetDriverByName("GTiff").Create("tmp/ovr26.tif", 100, 100, 1)
     ds.GetRasterBand(1).Fill(1)
     ds.GetRasterBand(1).FlushCache()
     ds.BuildOverviews("NEAR", overviewlist=[2])
--- a/autotest/gcore/tiff_write.py
+++ b/autotest/gcore/tiff_write.py
@@ -27,6 +27,7 @@
 # Boston, MA 02111-1307, USA.
 ###############################################################################
 
+import array
 import copy
 import math
 import os
@@ -8724,10 +8725,7 @@ def test_tiff_write_lerc_zstd_level(external_ovr):
 
 def test_tiff_write_181_xmp():
 
-    src_ds = gdal.Open("data/utmsmall.tif")
-
-    new_ds = gdaltest.tiff_drv.CreateCopy("tmp/test_181.tif", src_ds)
-    src_ds = None
+    new_ds = gdaltest.tiff_drv.Create("tmp/test_181.tif", 1, 1)
 
     xmp_ds = gdal.Open("../gdrivers/data/gtiff/byte_with_xmp.tif")
     xmp = xmp_ds.GetMetadata("xml:XMP")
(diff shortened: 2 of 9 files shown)


tbonfort:
wow! the pixel interleaved multiband case speedup is impressive, and the two other ones are also very significant. can't wait to test this out. :clap: @rouault

tbonfort:
> The visible side effects of this optimization are hopefully few, but one
> particular case exhibited by the test_tiff_ovr_10() case is when writing
> with a lossy compression and re-reading the raster without closing it in
> between. Previously if the raster was small enough to fit in the block
> cache, re-reading it would get the original pixel values. With this
> change, we now get the values after decompression.

This is good to know but not unexpected: it will happen all the time with this PR applied whereas it could happen or not beforehand depending on the fullness of the blockcache. i.e. this was not some behavior that could be relied on.

tbonfort:
@rouault I was wondering, will this new codepath get activated on a regular gdal_translate and/or gdalwarp command?

rouault:
> will this new codepath get activated on a regular gdal_translate and/or gdalwarp command?

gdal_translate, definitely yes, as shown by the change of behavior of test_tiff_ovr_10() which uses CreateCopy(), and that's what gdal_translate ultimately does.

For gdalwarp, not by default unless you specify the OPTIMIZE_SIZE=YES warping option (``-wo OPTIMIZE_SIZE=YES``). Cf https://gdal.org/api/gdalwarp_cpp.html#_CPPv4N15GDALWarpOptions16papszWarpOptionsE : "OPTIMIZE_SIZE: This defaults to FALSE, but may be set to TRUE typically when writing to a compressed dataset (GeoTIFF with COMPRESSED creation option set for example) for achieving a smaller file size. This is achieved by writing at once data aligned on full blocks of the target dataset, which avoids partial writes of compressed blocks and lost space when they are rewritten at the end of the file. However sticking to target block size may cause major processing slowdown for some particular reprojections."

This reminds me that I wanted to try improving that default behaviour of gdalwarp to enable OPTIMIZE_SIZE when it is reasonable to do so (now filed as https://github.com/OSGeo/gdal/issues/7761)

tbonfort:
so am I right in thinking that with this patch there might be some marginal gains when using gdal_translate, and noticeable ones when using gdalwarp -multi -wo optimize_size ?

rouault:
> there might be some marginal gains when using gdal_translate, and noticeable ones when using gdalwarp -multi -wo optimize_size ?

Hard to anticipate gains as those work flows are not pure I/O. gdalwarp should be generally more CPU intensive in non I/O code paths that gdalwarp, so I'm not sure the effects of this PR will be very visible. 
As you mention the -multi flag of gdalwarp, I guess you might be confused by what it does. It has generally very modest influence as it does ReadRaster() and WriteRaster() in a single thread, separated from the one(s) that execute the computational part of the warping (coordinate reprojection + resampling kernel). You won't get N threads writing to the same dataset, which isn't something by the GDAL data model anyway.

tbonfort:
LGTM: I have seen a ~50% decrease in processing time for workflows with multiple readers and a single writer

PR #7760, opened 2023-05-13
Add OGRCoordinateTransformationOptions::SetOnlyBest() / OCTCoordinateTransformationOptionsSetOnlyBest()

### LINKED ISSUE #7753: Missing grid message

In case a grid file is missing you are currently getting only a message like:
```
ERROR 1: No inverse operation
```

After long research I found that the reason was a missing grid file, which can only detected by using:
```
set PROJ_DEBUG=2
set CPL_DEBUG=ON
```

which leads to:

```
PROJ: pj_open_lib(proj.ini): call fopen(C:\dev\dependencies\proj-9.2.0/share/proj/proj.ini) - succeeded
PROJ: pj_open_lib(proj.db): call fopen(C:\dev\dependencies\proj-9.2.0/share/proj/proj.db) - succeeded
PROJ: pj_open_lib(de_adv_BETA2007.tif): call fopen(de_adv_BETA2007.tif) - failed
PROJ: pj_open_lib(BETA2007.gsb): call fopen(BETA2007.gsb) - failed
```

Expected would be a message like:
```
ERROR 1: Could not find grid file de_adv_BETA2007.tif
```

Note: I think it is enough to print the `proj_grid_name`, so `original_grid_name` and `old_proj_grid_name` are just optional additional hints as these are just fallback files.

To catch this issue and show a better message to the user, you can currently only set both environment variables and set `CPLSetErrorHandler` with a regex filter to `PROJ: pj_open_lib\(([^)]*)\): call fopen\(\1\) - failed` (untested).

I also considered abusing PROJ's networking capability for this, but that would require us to explicitly link against PROJ, which would be a bit overkill just to get a better error message.

Another approach is to extract all `proj_grid_name` from the proj.db itself to test in advance if the database matches with the proj-data package, but again this would require explicitly linking sqlite3 (apart from that it would be handy to retrieve all needed files to better test your setup if all files are required, but for that Proj would have to provide an API first)

Note: I came across the issue, as for example vcpkg does not download the proj-data archive itself when building PROJ without network capability.

Use following simplified code (and make sure in `PROJ_DATA` only the `proj.db` exists, but no TIFF file or at least no `de_adv_BETA2007.tif` in this case):

```cpp
OGRSpatialReference ogrSpatialReferenceSource;
ogrSpatialReferenceSource.SetAxisMappingStrategy(OAMS_TRADITIONAL_GIS_ORDER);
ogrSpatialReferenceSource.importFromEPSG(3396);
OGRSpatialReference ogrSpatialReferenceTarget;
ogrSpatialReferenceTarget.SetAxisMappingStrategy(OAMS_TRADITIONAL_GIS_ORDER);
ogrSpatialReferenceTarget.importFromEPSG(4978);

const auto ogrCoordinateTransformation = OGRCreateCoordinateTransformation(&ogrSpatialReferenceSource, &ogrSpatialReferenceTarget);

// x/y values from projected bounds
double x = 3563762.14;
double y = 5563059.05;
double z = 100.0;
ogrCoordinateTransformation->Transform(1, &x, &y, &z);

OGRCoordinateTransformation::DestroyCT(ogrCoordinateTransformation);
```

## Operating system

Win10

GDAL 3.6.4 + PROJ 9.2

rouault:
Fixes #7753

By default (at least in the PROJ 9.x series), PROJ may use coordinate
operations that are not the "best" if resources (typically grids) needed
to use them are missing. It will then fallback to other coordinate operations
that have a lesser accuracy, for example using Helmert transformations,
or in the absence of such operations, to ones with potential very rought
accuracy, using "ballpark" transformations
(see https://proj.org/glossary.html).

When calling this method with bOnlyBest = true, PROJ will only consider the
"best" operation, and error out (at Transform() time) if they cannot be
used.
This method may be used together with SetBallparkAllowed(false) to
only allow best operations that have a known accuracy.

Note that this method has no effect on PROJ versions before 9.2.

The default value for this option can be also set with the
PROJ_ONLY_BEST_DEFAULT environment variable, or with the "only_best_default"
setting of proj.ini. Calling SetOnlyBest() overrides such default value.


jratike80:
I read the PR twice and I am still not sure what is the default if GDAL user does nothing. I guess that it is OnlyBest=false.

rouault:
> I read the PR twice and I am still not sure what is the default if GDAL user does nothing. I guess that it is OnlyBest=false.

The GDAL default behaviour will depend on PROJ default behaviour which is currently onlyBest = false.

Can you propose some rewording that would clarify things, as the documentation of the method appears rather clear to me ?

PR #7778, opened 2023-05-16
SHPRestoreSHX(): add extra sanity checks

rouault:


rouault:
CC @agiudiceandrea

agiudiceandrea:
Thank you very much for this improvement! Is there a way to catch the SHPRestoreSHX() failure and display the emitted errors when using the "Repair shapefile" processing algorithm in QGIS https://github.com/qgis/QGIS/blob/master/src/analysis/processing/qgsalgorithmrepairshapefile.cpp#L81C6-L90?

rouault:
> Is there a way to catch the SHPRestoreSHX() failure

Perhaps calling CPLGetLastErrorType() + CPLGetLastErrorMsg(), but I haven't verified if there are not other subsequent errors that are emitted. Otherwise installing a custom error handler with CPLPushErrorHandlerEx() / CPLPopErrorHandler() is a way of catching all emitted errors, not just the last one

agiudiceandrea:
I see another little issue here https://github.com/OSGeo/gdal/blob/c92b22d02c99eae0152f49595947fb3747ddc280/ogr/ogrsf_frmts/shape/shpopen.c#L685-L698

If SHAPE_RESTORE_SHX=YES, then SHPRestoreSHX is executed. If SHPRestoreSHX fails, then SHPOpenLLEx always returns SHPLIB_NULLPTR even if the layer, after the SHX restoration, could be read by SHPOpenLL.
This could lead to a bit misleading behaviour:
```
>set SHAPE_RESTORE_SHX=YES
>ogrinfo corrupted.shp
ERROR 4: Error parsing .shp to restore .shx
ERROR 4: Error parsing .shp to restore .shx
ogrinfo failed - unable to open 'corrupted.shp'.
>set SHAPE_RESTORE_SHX=
>ogrinfo corrupted.shp
INFO: Open of `corrupted.shp'
      using driver `ESRI Shapefile' successful.
1: corrupted (Line String)
```

(Why I see two times the same error?)


I would also suggest add here https://github.com/OSGeo/gdal/blob/c92b22d02c99eae0152f49595947fb3747ddc280/ogr/ogrsf_frmts/shape/shpopen.c#L612-L613
a similar advice as the one at https://github.com/OSGeo/gdal/blob/c92b22d02c99eae0152f49595947fb3747ddc280/ogr/ogrsf_frmts/shape/shpopen.c#L417-L418

rouault:
> If SHPRestoreSHX fails, then SHPOpenLLEx always returns SHPLIB_NULLPTR

I see your point, but that seems appropriate to me, as a way to indicate something went wrong in the restoration.

> (Why I see two times the same error?)

probably because it first tries in update mode, and then in read-only mode

> I would also suggest add here 

I'm a bit ambivalent about that. Suggesting SHAPE_RESTORE_SHX when there is no .shx is totally reasonable, but when the .shx exists and there is a disagreement with the .shp, this might be a corruption in the .shp itself. If you wanted to add a hint about SHAPE_RESTORE_SHX, words of caution should be added that it will override the existing .shx, which may prevent further forensic attempts, and people should probably backup the .shx before.

PR #7791, opened 2023-05-18
OGR TileDB: fully apply spatial filter when using ArrowArray interface

rouault:


normanb:
👍 I have built this branch and checked with a larger dataset and the fix works great.

rouault:
3.7 backport in https://github.com/OSGeo/gdal/pull/7822

PR #7792, opened 2023-05-18
MSSQL: fix crash in ogr2ogr when BCP is enabled

### LINKED ISSUE #7787: MSSQL: ogr2ogr crash when using MSSQLSPATIAL_USE_BCP
The MSSQL driver may reset the feature definition when initializes BCP, `LayerTranslator::Translate` holds a raw pointer to it, causing the crash when assigning a deleted definition to a new feature (this happens on the second feature copy).

I discovered this while writing a test for BCP insert in `autotest/ogr/ogr_mssqlspatial.py`

```python
def test_ogr_mssqlspatial_bulk_insert():

    if gdaltest.mssqlspatial_ds is None:
        pytest.skip()

    source_ds = gdal.OpenEx("data/poly.shp", gdal.OF_VECTOR)

    assert source_ds

    try:
        gdaltest.mssqlspatial_ds.ExecuteSQL("DROP TABLE poly")
    except Exception as ex:
        pass

    with gdaltest.error_handler():
        res = gdal.VectorTranslate(
            gdaltest.mssqlspatial_dsname,
            "data/poly.shp",
            layerCreationOptions=[
                "OVERWRITE=YES",
                "MSSQLSPATIAL_USE_BCP=YES",
                "MSSQLSPATIAL_BCP_SIZE=2",
            ],
        )

        lyr = gdaltest.mssqlspatial_ds.GetLayerByName('poly')
        assert lyr
        assert lyr.GetFeatureCount() == 10
        del lyr

```

A possible patch to fix the issue:

```C++
diff --git a/apps/ogr2ogr_lib.cpp b/apps/ogr2ogr_lib.cpp
index d73c4b5637..2194dcc2ff 100644
--- a/apps/ogr2ogr_lib.cpp
+++ b/apps/ogr2ogr_lib.cpp
@@ -5186,7 +5186,7 @@ int LayerTranslator::Translate(OGRFeature *poFeatureIn, TargetLayerInfo *psInfo,
     const int iSrcZField = psInfo->m_iSrcZField;
     const bool bPreserveFID = psInfo->m_bPreserveFID;
     const auto poSrcFDefn = poSrcLayer->GetLayerDefn();
-    const auto poDstFDefn = poDstLayer->GetLayerDefn();
+    auto poDstFDefn = poDstLayer->GetLayerDefn();
     const int nSrcGeomFieldCount = poSrcFDefn->GetGeomFieldCount();
     const int nDstGeomFieldCount = poDstFDefn->GetGeomFieldCount();
     const bool bExplodeCollections =
@@ -5345,6 +5345,9 @@ int LayerTranslator::Translate(OGRFeature *poFeatureIn, TargetLayerInfo *psInfo,
             {
                 poDstFeature = std::move(poFeature);
                 // From now on, poFeature is null !
+                // Some drivers (MSSQL) may release the feature definition, we
+                // need to get a fresh one.
+                poDstFDefn = poDstLayer->GetLayerDefn();
                 poDstFeature->SetFDefnUnsafe(poDstFDefn);
                 poDstFeature->SetFID(nDesiredFID);
             }

```

elpaso:
Fix #7787

### DIFF
--- a/autotest/ogr/ogr_mssqlspatial.py
+++ b/autotest/ogr/ogr_mssqlspatial.py
@@ -559,6 +559,43 @@ def test_ogr_mssqlspatial_datatypes():
 #
 
 
+def test_ogr_mssqlspatial_bulk_insert():
+    """Test issue GH https://github.com/OSGeo/gdal/issues/7787"""
+
+    if gdaltest.mssqlspatial_ds is None:
+        pytest.skip()
+
+    use_bcp = "BCP" in gdal.GetDriverByName("MSSQLSpatial").GetMetadataItem(
+        gdal.DMD_LONGNAME
+    )
+
+    if not use_bcp:
+        pytest.skip()
+
+    gdal.SetConfigOption("MSSQLSPATIAL_BCP_SIZE", "2")
+    source_ds = gdal.OpenEx("data/poly.shp", gdal.OF_VECTOR)
+
+    assert source_ds
+
+    with gdaltest.error_handler():
+        ds = gdal.VectorTranslate(
+            gdaltest.mssqlspatial_dsname,
+            "data/poly.shp",
+            layerCreationOptions=[
+                "OVERWRITE=YES",
+            ],
+        )
+
+        lyr = ds.GetLayerByName("poly")
+        assert lyr
+        assert lyr.GetFeatureCount() == 10
+        del lyr
+
+
+###############################################################################
+#
+
+
 def test_ogr_mssqlspatial_cleanup():
 
     if gdaltest.mssqlspatial_ds is None:
@@ -569,6 +606,7 @@ def test_ogr_mssqlspatial_cleanup():
     gdaltest.mssqlspatial_ds = ogr.Open(gdaltest.mssqlspatial_dsname, update=1)
     gdaltest.mssqlspatial_ds.ExecuteSQL("DROP TABLE Unregistered")
     gdaltest.mssqlspatial_ds.ExecuteSQL("DROP TABLE tpoly")
+    gdaltest.mssqlspatial_ds.ExecuteSQL("DROP TABLE poly")
     gdaltest.mssqlspatial_ds.ExecuteSQL("DROP TABLE test_ogr_mssql_datatypes")
 
     gdaltest.mssqlspatial_ds = None
--- a/autotest/ogr/ogr_mssqlspatial.py
+++ b/autotest/ogr/ogr_mssqlspatial.py
@@ -572,24 +572,28 @@ def test_ogr_mssqlspatial_bulk_insert():
     if not use_bcp:
         pytest.skip("BCP not available")
 
-    gdal.SetConfigOption("MSSQLSPATIAL_BCP_SIZE", "2")
-    source_ds = gdal.OpenEx("data/poly.shp", gdal.OF_VECTOR)
+    with gdaltest.config_option("MSSQLSPATIAL_BCP_SIZE", "2"):
 
-    assert source_ds
+        source_ds = gdal.OpenEx("data/poly.shp", gdal.OF_VECTOR)
 
-    with gdaltest.error_handler():
-        ds = gdal.VectorTranslate(
-            gdaltest.mssqlspatial_dsname,
-            "data/poly.shp",
-            layerCreationOptions=[
-                "OVERWRITE=YES",
-            ],
-        )
+        assert source_ds
 
-        lyr = ds.GetLayerByName("poly")
-        assert lyr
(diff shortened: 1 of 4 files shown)


autotest/ogr/ogr_mssqlspatial.py
```diff
@@ -569,6 +606,7 @@ def test_ogr_mssqlspatial_cleanup():
     gdaltest.mssqlspatial_ds = ogr.Open(gdaltest.mssqlspatial_dsname, update=1)
     gdaltest.mssqlspatial_ds.ExecuteSQL("DROP TABLE Unregistered")
     gdaltest.mssqlspatial_ds.ExecuteSQL("DROP TABLE tpoly")
+    gdaltest.mssqlspatial_ds.ExecuteSQL("DROP TABLE poly")
```
rouault, line 609: IMHO the practice of the test_XXXX_cleanup() method should be considered as deprecated. Each test step should clean up after itself so they can be run in a isolated way.
So concretely I would put lines 580-592 inside a try block and put ``gdaltest.mssqlspatial_ds.ExecuteSQL("DROP TABLE poly")`` in the finally clause

ogr/ogrsf_frmts/mssqlspatial/ogrmssqlspatialtablelayer.cpp
```diff
@@ -1562,13 +1562,15 @@ OGRErr OGRMSSQLSpatialTableLayer::CreateFeatureBCP(OGRFeature *poFeature)
...
         /* Get the column definitions for this table. */
-        if (poFeatureDefn)
+        if (poFeatureDefn && bLayerDefnNeedsRefresh)
         {
-            /* need to re-create layer defn */
-            poFeatureDefn->Release();
-            poFeatureDefn = nullptr;
+            GetLayerDefn();
+            bLayerDefnNeedsRefresh = false;
+        }
```
rouault, line 1569: I'm not really understanding this game of why we need to refresh the layer definition. Did you experiment with *not* refreshing it ? although I suspect there must be a good reason for what it has been put in place.

But if that's needed, given that BCP mode can be ended in EndCopy(), I suspect we should also refresh the layer defn in a similar way there too. So probably that bLayerDefnNeedsRefresh should be initialized to false in the .h file, not tested at line 1565 and just set to true before calling GetLayerDefn() at line 1567.

And a similar thing to be done in EndCopy()

autotest/ogr/ogr_mssqlspatial.py
```diff
@@ -559,6 +559,43 @@ def test_ogr_mssqlspatial_datatypes():
...
+
+    if gdaltest.mssqlspatial_ds is None:
+        pytest.skip()
+
+    use_bcp = "BCP" in gdal.GetDriverByName("MSSQLSpatial").GetMetadataItem(
+        gdal.DMD_LONGNAME
+    )
+
+    if not use_bcp:
+        pytest.skip()
```
rouault, line 573:
```suggestion
        pytest.skip("BCP not available")
```

autotest/ogr/ogr_mssqlspatial.py
```diff
@@ -559,6 +559,43 @@ def test_ogr_mssqlspatial_datatypes():
...
+
+    gdal.SetConfigOption("MSSQLSPATIAL_BCP_SIZE", "2")
```
rouault, line 575: using gdal.SetConfigOption() is quite dangerous as it must be done, and in all code paths (regular & exceptions)
It is better to use the gdaltest.config_option() context manager (and indenting properly what code must be in that scope)
```suggestion
    with gdaltest.config_option("MSSQLSPATIAL_BCP_SIZE", "2"):
```

ogr/ogrsf_frmts/mssqlspatial/ogrmssqlspatialtablelayer.cpp
elpaso: > I'm not really understanding this game of why we need to refresh the layer definition. Did you experiment with _not_ refreshing it ? although I suspect there must be a good reason for what it has been put in place.
> 

It took me a while to find why, the reason is because of the side effect of GetLayerDefn that calls BuildFeatureDefn() during the layer creation, when fields were not yet added. 

> But if that's needed, given that BCP mode can be ended in EndCopy(), I suspect we should also refresh the layer defn in a similar way there too. So probably that bLayerDefnNeedsRefresh should be initialized to false in the .h file, not tested at line 1565 and just set to true before calling GetLayerDefn() at line 1567.
> 

Done.

> And a similar thing to be done in EndCopy()

That shouldn't be necessary.

Issue #7784, opened 2023-05-16
Implement ESRI Tile Package (.tpkx) format


szekerest:
This format (as described per [https://github.com/Esri/tile-package-spec](https://github.com/Esri/tile-package-spec)) is in fact an [Esri Compact Cache V2](https://github.com/Esri/raster-tiles-compactcache) in a zipped package. The bulk of the implementation is already done within the ESRIC driver, which should be modified to read the root.json metadata file in addition to the current conf.xml file. (The provided parameters are basically the same).
According to the suggested implementation the following syntax would be used.

/vsizip/{path to tpkx file}/root.json

A sample file [Usa.tpkx](https://github.com/Esri/tile-package-spec/blob/master/Sample%20TPKX%20Files/Usa.tpkx) is available for testing the implementation.

rouault:
I'm CC'ing @lucianpls since I copied him of @szekerest initial email so both of you can appropriately coordinate

szekerest:
Thanks @rouault I'm trying to commit the required patch in the next few days.

PR #7799, opened 2023-05-21
Implement ESRI Tile Package (.tpkx) support in the ESRIC driver (#7784)

szekerest:


rouault:
you still have formatting issues. Try
```
pip install pre-commit
pre-commit install
pre-commit run --all-files
```

frmts/esric/esric_dataset.cpp
```diff
@@ -291,32 +312,156 @@ CPLErr ECDataset::Initialize(CPLXMLNode *CacheInfo)
...
+                resolutions.push_back(res);
+            }
+        }
+        sort(resolutions.begin(), resolutions.end());
+        if (resolutions.empty())
+            throw CPLString("Can't parse lods");
+
+        if (OGRERR_NONE !=
+            oSRS.SetFromUserInput(CPLString().Printf(
+                "ESRI:%d", oRoot.GetInteger("spatialReference/wkid"))))
```
rouault, line 354: I'm not sure about this particular format, but with other Esri stuff (like FileGDB), you want to use the ESRI: namespace only for codes > 32767. For codes <= 32767, you have to use the EPSG: namespace

rouault (commented):
Some testing for this in autotest/gdrivers/esric.py would be great.

CC'ing @lucianpls , in particular if he has a comment on the commenting off of the ``BSZ * BSZ != u32lat(header + 4)`` check  at line 123

frmts/esric/esric_dataset.cpp
lucianpls: Indeed, the previous test should work, according to the [documentation](https://github.com/Esri/raster-tiles-compactcache/blob/master/CompactCacheV2.md).

szekerest: @lucianpls u32lat(header + 4) seems to contain zero for the tpkx format, however we could leave it in for the xml based configurations.

lucianpls: Bundles should be all the same format, this is likely a bug somewhere. How was the tile package generated?

szekerest: @lucianpls I got the files from the client, probably made with some ESRI tools. But you could also check [this example](https://github.com/Esri/tile-package-spec/tree/master) where that field is also zero:
image

What do you mean by "Bundles should be all the same format". There might probably be slight differences between the implementation applications (like in the .tpkx case). Where can we find the official documentation about that?

szekerest: > I'm not sure about this particular format, but with other Esri stuff (like FileGDB), you want to use the ESRI: namespace only for codes > 32767. For codes <= 32767, you have to use the EPSG: namespace

@rouault The corresponding documentation can be found [here](https://github.com/Esri/tile-package-spec/blob/master/docs/root.md#spatialreference). As what I can see in the root.json file, the wkid corresponds to the ESRI id (102100), while the LatestWkid contains the EPSG code (3857).

szekerest: @rouault Thanks for the hint, that makes sense. I'll incorporate that in the code.

lucianpls: > @lucianpls I got the files from the client, probably made with some ESRI tools. But you could also check [this example](https://github.com/Esri/tile-package-spec/tree/master) where that field is also zero: image
> 
> What do you mean by "Bundles should be all the same format". There might probably be slight differences between the implementation applications (like in the .tpkx case). Where can we find the official documentation about that?

Indeed, turns out esri code for building bundles is not consistent. It is best to leave the check the way you had it to start with, checking value at offset 60 for both tpkx and normal bundled cache.

rouault:
@szekerest Looks good to me, but I'd prefer if you could add at least a minimum test checking that the .json parsing works fine (probably not needed to check more than dataset metadata)

szekerest:
@rouault According to my further tests, it looks like the color interpretation of tiles in the same package may be different. Currently the esric driver doesn't support if the tile contains single band rasters with color palette. Is there an easy way to convert the paletted images to rgba within the driver?

rouault:
> Is there an easy way to convert the paletted images to rgba within the driver?

That shouldn't be too hard. You may take some inspiration from the GPKG driver. Look at ogr/ogrsf_frmts/gpkg/gdalgeopackagerasterband.cpp line 673-703

autotest/gdrivers/esric.py
```diff
@@ -119,6 +119,99 @@ def test_esric_4():
...
+        pytest.skip()
+
+    ds = gdaltest.tpkx_ds
+    b1 = ds.GetRasterBand(1)
+
+    assert (
+        ds.RasterCount == 4 and ds.RasterXSize == 8192 and ds.RasterYSize == 8192
+    ), "wrong size or band count"
+
+    assert b1.GetOverviewCount() == 5, "Wrong number of overviews"
```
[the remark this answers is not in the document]
szekerest: @rouault I'm not sure what kind of zipping parameters are actually set by ArcGIS. We might probably do something, but that might not be compatible by ArcGIS, which I cannot check.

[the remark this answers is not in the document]
szekerest: I've removed the unnecessary files from the bundle and applied a deflate compression.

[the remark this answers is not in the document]
rouault: Looking with 7z, it appears this is a zip without compression (ie it can be recompressed with ''zip -0 -r" using the infoZip utility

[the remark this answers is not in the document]
lucianpls: It has to be uncompressed. The bundles/tiles are JPEG or PNG, so they would not compress much anyhow, not in realistic cases.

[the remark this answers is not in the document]
szekerest: I understand that the bundle is zipped with no compression by default for performance reasons, but now it looks like the size of the file is more important. We should anyway require to create a package from scratch by using an arcgis tool instead of trying to minify an existing package. I have no arcgis license, so I can't really proceed with this any further.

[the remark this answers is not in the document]
rouault: > Though I don't see that commit in this PR

it's well included: https://github.com/OSGeo/gdal/pull/7799/commits/ad1f7e23a39449fe05f6a7cbb1dd709c2e7cd2c5

PR #7826, opened 2023-05-25
MSSQL: fix GEOGRAPHY clockwise vertex order

### LINKED ISSUE #1128: ogr2ogr everts polygon geographies when writing shapefile to MSSQL Server Spatial

When using ogr2ogr to convert a shapefile containing polygon features into an MSSQL Server spatial table with geography data, said polygons features are everted when written to SQL. By this I mean the output features cover every part of the globe except those described by their shapefile original equivalents. This seems to be because the ring order in the shapefile is preserved verbatim when written to SQL.

The shapefile specification (https://www.esri.com/library/whitepapers/pdfs/shapefile.pdf) specifies that the internal portion of a polygon is that to the right of an observer walking along the polygon bounds in vertex order. SQL server spatial appears to use the opposite convention (see https://docs.microsoft.com/en-us/sql/t-sql/spatial-geography/reorientobject-geography-data-type?view=sql-server-2017). Thus, preserving the point order of the bounds verbatim, instead of flipping it, causes the problem.

It is possible to fix the features after import into SQL server (using the the ReorientObject() function detailed in the link above), but it is needlessly inconvenient, especially when attempting to fix multi-polygons, for which simply calling ReorientObject() does not work. (In my experience each individual part of the multi must be re-oriented on its own and then UnionAggregated with the rest in order to reproduce the original shapefile feature).

Expected behaviour is that shapefile storage of polygon geographies is intelligently converted to SQL Sever geographies by ogr2ogr, such that converted features are identical to originals. 

ogr2ogr -f MSSQLSpatial "MSSQL:server=MyServer;database=MyExampleDatabase;trusted_connection=yes" \\path\to\my\file\Land_Parcels.shp -lco GEOMETRY_NAME=GEOM -lco GEOM_TYPE=GEOGRAPHY -nln LandParcels_Everted

Note: shapefile must be a polygon file (for the purposes of demonstration, ideally containing relatively small features which cover much less than half the Earth's area; after conversion, select a feature and observe its outer bounds; they will extend from -180 to +180 longitude and ~ -90 to +90 latitude.)

## Operating system

Windows 10 Pro 64 bit

Microsoft SQL Server 2012 (SP2-GDR) (KB3194719) - 11.0.5388.0 (X64) 
	Sep 23 2016 16:56:29 
	Copyright (c) Microsoft Corporation
	Enterprise Edition (64-bit) on Windows NT 6.1 <X64> (Build 7601: Service Pack 1)

GDAL 2.2.1, released 2017/06/23, from the OSGeo4W Network Installer (64 bit) found at  https://qgis.org/en/site/forusers/download.html

elpaso:
Fix #1128

### DIFF
--- a/autotest/ogr/ogr_mssqlspatial.py
+++ b/autotest/ogr/ogr_mssqlspatial.py
@@ -624,18 +624,19 @@ def test_ogr_mssqlspatial_geography_polygon_vertex_order():
             # Simple polygon, no inner rings
             feature = lyr.GetFeature(1)
             geometry = feature.GetGeometryRef()
-            boundary = geometry.GetBoundary()
+            boundary = geometry.GetGeometryRef(0)
             # Outer ring
             assert not boundary.IsClockwise()
 
             # Inner ring
             feature = lyr.GetFeature(13)
             geometry = feature.GetGeometryRef()
-            boundary = geometry.GetBoundary()
+            boundary = geometry.GetGeometryRef(0)
             # Outer ring
-            assert not boundary.GetGeometryRef(0).IsClockwise()
+            assert not boundary.IsClockwise()
             # Inner ring
-            assert boundary.GetGeometryRef(1).IsClockwise()
+            inner_ring = geometry.GetGeometryRef(1)
+            assert inner_ring.IsClockwise()
 
             del lyr
     finally:
--- a/autotest/ogr/ogr_mssqlspatial.py
+++ b/autotest/ogr/ogr_mssqlspatial.py
@@ -606,7 +606,7 @@ def test_ogr_mssqlspatial_geography_polygon_vertex_order():
     if gdaltest.mssqlspatial_ds is None:
         pytest.skip()
 
-    source_ds = gdal.OpenEx("data/poly.shp", gdal.OF_VECTOR)
+    source_ds = gdal.OpenEx("data/shp/testpoly.shp", gdal.OF_VECTOR)
 
     assert source_ds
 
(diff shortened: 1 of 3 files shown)


ogr/ogrsf_frmts/mssqlspatial/ogrmssqlgeometrywriter.cpp
```diff
@@ -253,6 +253,11 @@ void OGRMSSQLGeometryWriter::WritePoint(double x, double y, double z, double m)
 void OGRMSSQLGeometryWriter::WriteSimpleCurve(OGRSimpleCurve *poGeom,
                                               int iStartIndex, int nCount)
 {
+    if (nColType == MSSQLCOLTYPE_GEOGRAPHY && poGeom->isClockwise())
+    {
+        poGeom->reversePoints();
```
rouault, line 258: given that WriteSimpleCurve() can be called with iStartIndex != 0 (in the case of a compoundcurve),  reversePoints() can interfere with that. Probably the best would be to switch the starting and ending indices in the below loops and the increment step.
Furthermore doesn't the winding order depends on if the curve is an exterior one or an interior one ? and shouldn't that only be applied to curves that are rings of polygons/curvepolygons/multipolygons/multisurfaces ?

rouault: acually isClockwise() must be only called on a full closed ring. if it is called on a part of a compoundcurve, its output is garbage (cf the "Assumes that the line is closed." comment in  OGRCurve::isClockwise() doc)

elpaso: Right, I'm looking athttps://learn.microsoft.com/en-us/sql/relational-databases/spatial/polygon?view=sql-server-ver16#orientation-of-spatial-data but while it mentions clearly the left-hand-rule for the outer ring(s) is doesn't say anything about the inner rings, should I check them too and apply the opposite rule (clockwise) ?

rouault: > but while it mentions clearly the left-hand-rule for the outer ring(s) is doesn't say anything about the inner rings, should I check them too and apply the opposite rule (clockwise) ?

yes, I would expect the reverse convention to be used for inner rings

autotest/ogr/ogr_mssqlspatial.py
```diff
@@ -600,6 +600,52 @@ def test_ogr_mssqlspatial_bulk_insert():
...
+                options="-nln poly_vertex_order -s_srs EPSG:32632 -t_srs EPSG:4326 -lco OVERWRITE=YES -lco GEOM_TYPE=GEOGRAPHY",
+            )
+
+            lyr = ds.GetLayerByName("poly_vertex_order")
+            assert lyr
+
+            # Simple polygon, no inner rings
+            feature = lyr.GetFeature(1)
+            geometry = feature.GetGeometryRef()
+            boundary = geometry.GetBoundary()
```
rouault, line 627: GetBoundary() involves GEOS, and I'm not sure what it does behind the back (it could potentially modify the vertex order).
You can navigate through geometry parts with GetGeometryRef(). That is GetGeometryRef(0) on a MultiPolygon returns the first Polygon. GetGeometryRef(0) on a Polygon returns the exterior ring

autotest/ogr/ogr_mssqlspatial.py
```diff
@@ -600,6 +600,52 @@ def test_ogr_mssqlspatial_bulk_insert():
...
+            # Outer ring
+            assert not boundary.IsClockwise()
+
+            # Inner ring
+            feature = lyr.GetFeature(13)
+            geometry = feature.GetGeometryRef()
+            boundary = geometry.GetBoundary()
```
rouault, line 634: same as above

autotest/ogr/ogr_mssqlspatial.py
elpaso: Ok, I'll change it, but Isn't the fact that this test fails without the current patch the prove that GetBoundary does not mess with vertex order at least in this test case?

rouault: yes the current implementation of GetBoundary probably doesn't mess with those geometries. My remark was more to be immune against future GEOS behaviour changes and making it obvious we test against the raw geometry and not something that has underwent potential post processing

PR #7827, opened 2023-05-25
TileDB: add read/write multidimensional support

rouault:
It requires GDAL to be built and run against TileDB >= 2.15 (for older
versions, it is just disabled)

The driver supports:
- creating and reading groups and subgroups
- creating and reading multidimensional dense arrays with a numeric data type
- creating and reading numeric or string attributes in groups and arrays
- storing an indexing array of a dimension as a TileDB dimension label

The multidimensional API supports reading dense arrays created by the classic
raster API of GDAL.

CC @normanb


jp-dark:
I tested this out and it worked great for me. One optional tweak you might want to consider is using the same name for the arrays and the attributes (e.g. `time1`, `y`) rather than calling the attributes `TDB_VALUES`.

rouault:
> One optional tweak you might want to consider is using the same name for the arrays and the attributes (e.g. `time1`, `y`) rather than calling the attributes `TDB_VALUES`.

done with latest commit

PR #7832, opened 2023-05-25
GDALDataset::Open(): take into account open options for OF_SHARED mode (fixes #7824)

### LINKED ISSUE #7824: Raster overviews and GDAL_OF_SHARED

I want to have multiple overviews of a raster dataset be opened multiply at the same time, and I want each group of the same overview level use the **same block cache**. This seems impossible in a simple manner and for general formats - e.g. GTiff.

The simplest way to achieve this would seem to be using GDAL_OF_SHARED for opening all overviews. However, if GDALOpenEx finds a shared, open GDALDatasetH of the same pszFilename, then it simply returns that GDALDatasetH, ignoring papszOpenOptions. Hence, it is impossible in this manner to open different, shared overviews of the same dataset at the same time. [1]

A still simple, different way would be to open the main dataset, passing GDAL_OF_SHARED, and then call GDALCreateOverviewDataset on it, passing the wanted overview level. However, GDAL does not export a symbol for GDALCreateOverviewDataset when built as a DLL [due to missing CPL_DLL](https://github.com/OSGeo/gdal/blob/9d9d393e84d031a2dff760928ad9e28084f92e27/gcore/gdal_priv.h#L3678). [2]

The only way to achieve this goal without changes to the GDAL source seems to be to use VRTCreate, VRTAddBand, and VRTAddSimpleSource to create a VRT dataset that holds overview bands, as [WMTSDataset does](https://github.com/OSGeo/gdal/blob/9d9d393e84d031a2dff760928ad9e28084f92e27/frmts/wmts/wmtsdataset.cpp#L2322). But doing this right (even more including masks and meta data) seems elaborate and it means to duplicate code that GDAL already has.

Remedies I can think of are:
- decorate GDALCreateOverviewDataset with CPL_DLL
- make GDALOpenEx consider "OVERVIEW_LEVEL" in its papszOpenOptions also when [called with GDAL_OF_SHARED and if it finds a GDALDatasetH for pszFilename](https://github.com/OSGeo/gdal/blob/9d9d393e84d031a2dff760928ad9e28084f92e27/gcore/gdaldataset.cpp#L3413).

The code below sets CPL_DEBUG to ON and GDAL_BAND_BLOCK_CACHE to HASHSET mostly in order to trace the [creation of block caches](https://github.com/OSGeo/gdal/blob/9d9d393e84d031a2dff760928ad9e28084f92e27/gcore/gdalrasterband.cpp#L1042). For simplicity, it opens overviews of a single level only.

```
  CPLSetConfigOption("CPL_DEBUG", "ON");
  CPLSetConfigOption("GDAL_BAND_BLOCK_CACHE", "HASHSET");

  CPLStringList openOvr0;
  openOvr0.AddNameValue("OVERVIEW_LEVEL", "0only");

  int pixel = 1,
      line = 1;
  uint8_t data;

  auto rasterIo = [&](GDALDataset* dataset) {
    assert(dataset->RasterIO(GF_Read, pixel, line, 1, 1, &data, 1, 1, GDT_Byte, 1, nullptr, 0, 0, 0, nullptr) == CE_None);
  };

  // Some TIFF with internal overviews (to get the exact same CPLDebug output), e.g.:
  // https://download.osgeo.org/geotiff/samples/gdal_eg/cea.tif
  // gdaladdo -r average cea.tif
  const char* tiffFn = "cea.tif";

  GDALDataset* dsMain = GDALDataset::Open(tiffFn, GDAL_OF_RASTER | GDAL_OF_READONLY | GDAL_OF_SHARED),
             * dsOvr0_0,
             * dsOvr0_1;
  {
    // Just for illustration: open the overviews without GDAL_OF_SHARED.

    dsOvr0_0 = GDALDataset::Open(tiffFn, GDAL_OF_RASTER | GDAL_OF_READONLY, nullptr, openOvr0);
    dsOvr0_1 = GDALDataset::Open(tiffFn, GDAL_OF_RASTER | GDAL_OF_READONLY, nullptr, openOvr0);
    assert(dsMain->GetRasterXSize() > dsOvr0_0->GetRasterXSize());
    assert(dsMain != dsOvr0_0);
    assert(dsOvr0_0 != dsOvr0_1);

    // Creates a new band block cache, as expected:
    // CPLDebug: GDAL: Use hashset band block cache
    rasterIo(dsOvr0_0);

    // Creates another new band block cache, as expected:
    // CPLDebug: GDAL: Use hashset band block cache
    rasterIo(dsOvr0_1);

    dsOvr0_0->ReleaseRef();
    dsOvr0_1->ReleaseRef();
  }
  {
    // [1]

    // Returns dsMain instead of the requested overview, ignoring openOvr0!
    dsOvr0_0 = GDALDataset::Open(tiffFn, GDAL_OF_RASTER | GDAL_OF_READONLY | GDAL_OF_SHARED, nullptr, openOvr0);
    assert(dsMain == dsOvr0_0);

    dsOvr0_0->ReleaseRef();
  }

  {
    // [2]

    dsOvr0_0 = GDALCreateOverviewDataset(dsMain, 0, TRUE /*bThisLevelOnly*/);
    dsOvr0_1 = GDALCreateOverviewDataset(dsMain, 0, TRUE /*bThisLevelOnly*/);
    assert(dsOvr0_0 != dsOvr0_1);

    // Creates new, common cache:
    // CPLDebug: GDAL: Use hashset band block cache
    rasterIo(dsOvr0_0);

    // Does not create a new block cache, since fetched from the common cache:
    // CPLDebug: <nothing>
    rasterIo(dsOvr0_1);

    dsOvr0_0->ReleaseRef();
    dsOvr0_1->ReleaseRef();
  }
  dsMain->ReleaseRef();
```

## Operating system

Windows 10 64 bit

gdal-3.7.0, own build

rouault:


rouault:
too tricky/risky for 3.7 backport

Issue #1990, opened 2019-11-06
MSSQL Server geometry column not identified


steveoh:
## Expected behavior and actual behavior.

I expect a MSSQL Server geometry type to be a geometry when converting between formats.

![image](https://user-images.githubusercontent.com/325813/68328007-ad874900-008b-11ea-9f10-5f4dcfdc6ec1.png)

```
Microsoft SQL Server 2014 (SP3-GDR) (KB4505218) - 12.0.6108.1 (X64) 
	May 29 2019 20:05:27 
	Copyright (c) Microsoft Corporation
	Developer Edition (64-bit) on Windows NT 6.3 <X64> (Build 14393: ) (Hypervisor)
```

and

```
Microsoft SQL Server 2014 (SP3-CU4) (KB4500181) - 12.0.6329.1 (X64)
    Jul 20 2019 21:42:29
    Copyright (c) Microsoft Corporation
    Enterprise Edition: Core-based Licensing (64-bit) on Windows NT 6.3 <X64> (Build 9600: ) (Hypervisor)
```

I currently have the following drivers. I tried them all and decided to not specify any in particular since it made no difference.

- SQL Server
- SQL Server Native Client 11.0
- ODBC Driver 17 for SQL Server

## Steps to reproduce the problem.

create a geometry type in sql server.

```py
from osgeo import gdal

shp_options = gdal.VectorTranslateOptions(
    options=[
        '-f',
        'ESRI Shapefile',
        '-sql',
        "SELECT name,shape FROM dbo.COUNTIES",
        '-nln',
        'counties'
    ],
)

result = gdal.VectorTranslate(
    "c:\dev", "MSSQL:server=(local);database=UDEQ;trusted_connection=yes;", options=shp_options
)
```

When going to shapefile this error occurs:

```
The output driver does not natively support Binary type for field Shape. Misconversion can happen. -mapFieldType can be used to control field type conversion.
```

When going to postgis no error occurs but the shape is a `bytea` type.

A solution that I do not like for performance reasons is to rebuild the geometry object.

`GEOMETRY::STGeomFromWKB(shape.STAsBinary(), srid)`

This solves the shapefile and postgis conversions.

## Operating system

Windows Server 2016 Standard

## GDAL version and provenance

2.3.3 from conda

related to #1136

jratike80:
[...]

PR #7871, opened 2023-05-30
MSSQL: add test for geometry column identification

elpaso:
Add test for issue #1990

Also add more cases to the existing test  binary field BCP.

### DIFF
--- a/autotest/ogr/ogr_mssqlspatial.py
+++ b/autotest/ogr/ogr_mssqlspatial.py
@@ -749,9 +749,8 @@ def test_geometry_column_identification():
 
     try:
         ds = ogr.Open(gdaltest.mssqlspatial_dsname)
-        lyr = ds.ExecuteSQL("SELECT eas_id, shape FROM poly_shape")
-        assert lyr
-        f = lyr.GetFeature(1)
+        with ds.ExecuteSQL("SELECT eas_id, shape FROM poly_shape") as lyr:
+            f = lyr.GetFeature(1)
         geom = f.GetGeometryRef()
         assert geom is not None
         assert geom.GetGeometryName() == "POINT"
--- a/autotest/ogr/ogr_mssqlspatial.py
+++ b/autotest/ogr/ogr_mssqlspatial.py
@@ -659,20 +659,34 @@ def test_binary_field_bcp():
     src_lyr = ds.CreateLayer("test_binary_field", geom_type=ogr.wkbPoint)
     src_lyr.CreateField(ogr.FieldDefn("binfield", ogr.OFTBinary))
     src_lyr.CreateField(ogr.FieldDefn("strfield", ogr.OFTString))
+    src_lyr.CreateField(ogr.FieldDefn("intfield", ogr.OFTInteger))
+
     f = ogr.Feature(src_lyr.GetLayerDefn())
     f.SetGeometry(ogr.CreateGeometryFromWkt("POINT(1 2)"))
     f.SetFieldBinaryFromHexString("binfield", "007FFF007FFF")
     f["strfield"] = "some text"
+    f["intfield"] = 1
     src_lyr.CreateFeature(f)
+
     f = ogr.Feature(src_lyr.GetLayerDefn())
     f.SetGeometry(ogr.CreateGeometryFromWkt("POINT(2 3)"))
     f.SetFieldBinaryFromHexString("binfield", "")
     f["strfield"] = "some text"
+    # leave int undefined
     src_lyr.CreateFeature(f)
+
     f = ogr.Feature(src_lyr.GetLayerDefn())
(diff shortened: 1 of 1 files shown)


autotest/ogr/ogr_mssqlspatial.py
```diff
@@ -711,6 +736,49 @@ def test_binary_field_bcp():
...
+def test_geometry_column_identification():
+    """Test for issue GH #1190
+    SQL Server data source: geometry column not identified by ogr2ogr"""
+
+    if gdaltest.mssqlspatial_ds is None:
+        pytest.skip("MSSQLSpatial driver not available")
+
+    # once
+    try:
+        gdaltest.mssqlspatial_ds.ExecuteSQL("ALTER TABLE dbo.tpoly ADD shape GEOMETRY")
```
rouault, line 748: that make this test step dependent on previous ones, and affect later ones.

Overall, looking at this test & ticket, I don't think it has anything to do with ogr2ogr. It is more about checking that ds.ExecuteSQL("SELECT ... the geometry column ...") returns a geometry column. No need for gdal.VectorTranslate() to check that

elpaso: done

autotest/ogr/ogr_mssqlspatial.py
```diff
@@ -711,6 +736,35 @@ def test_binary_field_bcp():
...
+def test_geometry_column_identification():
+    """Test for issue GH #1190
+    SQL Server data source: geometry column not identified by ogr2ogr"""
+
+    if gdaltest.mssqlspatial_ds is None:
+        pytest.skip("MSSQLSpatial driver not available")
+
+    gdaltest.mssqlspatial_ds.ExecuteSQL(
+        "SELECT *, ogr_geometry.STCentroid() AS shape INTO dbo.poly_shape FROM dbo.tpoly"
+    )
```
rouault, line 748: is it useful ? The result is not tested.
```suggestion
    gdaltest.mssqlspatial_ds.ExecuteSQL(
        "SELECT *, ogr_geometry.STCentroid() AS shape INTO dbo.poly_shape FROM dbo.tpoly"
    )
```

Furthermore, the result of ExecuteSQL() (when it returns something) must be explicitly freed with ds.ReleaseResultSet(lyr_returned_by_executeSql)
A way of doing that introduced recently is to use
```
with ds.ExecuteSQL(...) as lyr:
     .. do something ..
```

autotest/ogr/ogr_mssqlspatial.py
```diff
@@ -711,6 +736,35 @@ def test_binary_field_bcp():
...
+
+    try:
+        ds = ogr.Open(gdaltest.mssqlspatial_dsname)
+        lyr = ds.ExecuteSQL("SELECT eas_id, shape FROM poly_shape")
```
rouault, line 752: cf above mention about ExecuteSQL() leaks

elpaso: I am missing something: when ds and lyr go out of scope doesn't python swig call the destructors and free memory?

Besides that when the python test ends memory will be reclaimed anyway.

autotest/ogr/ogr_mssqlspatial.py
elpaso: that's a preparation step that creates a new table with SELECT INTO since you mentioned to not alter tpoly which is used by another test case.

rouault: as sorry, I missed the INTO ... . I thought it was a simple SELECT. Does it return None or non-None ? If non-None, that should be freed

autotest/ogr/ogr_mssqlspatial.py
rouault: > when ds and lyr go out of scope doesn't python swig call the destructors and free memory?

Not for a ExecuteSQL() layer that must be explicitly freed: https://gdal.org/api/python/osgeo.gdal.html#osgeo.gdal.Dataset.ExecuteSQL  . Perhaps this could be changed but this would have implications probably on the C++ side of things, probably in all drivers that implement ExecuteSQL()

> Besides that when the python test ends memory will be reclaimed anyway.

no, the C++ layer woud be leaking.

autotest/ogr/ogr_mssqlspatial.py
elpaso: it returns None

autotest/ogr/ogr_mssqlspatial.py
```diff
@@ -711,6 +736,35 @@ def test_binary_field_bcp():
...
+        assert lyr
+        f = lyr.GetFeature(1)
```
rouault, line 754:
```suggestion
        with ds.ExecuteSQL("SELECT eas_id, shape FROM poly_shape") as lyr:
            f = lyr.GetFeature(1)
```

PR #7884, opened 2023-06-01
OpenFileGDB: open v9 rasters when layer name is explictly provided

Milek7:
Simple patch to open v9 rasters when layer name is provided, no SRS support yet.

PR #7885, opened 2023-06-01
OpenFileGDB: add support for FileGDB v9 raster

rouault:
Builds upon https://github.com/OSGeo/gdal/pull/7884

Can be tested with datasets at
https://resources.gisdata.mn.gov/pub/data/elevation/lidar/county/washington/geodatabase/

### DIFF
--- a/ogr/ogrsf_frmts/openfilegdb/gdalopenfilegdbrasterband.cpp
+++ b/ogr/ogrsf_frmts/openfilegdb/gdalopenfilegdbrasterband.cpp
@@ -444,9 +444,55 @@ bool OGROpenFileGDBDataSource::OpenRaster(const GDALOpenInfo *poOpenInfo,
         m_adfGeoTransform[5] = -dfResY;
     }
 
-    // Get SRID, and fetch WKT from GDBSpatialRefs table
+    // Two cases:
+    // - osDefinition is empty (that is FileGDB v9): find the SRS by looking
+    //   at the SRS attached to the RASTER field definition of the .gdbtable
+    //   file of the main table of the raster (that is the one without fras_XXX
+    //   prefixes)
+    // - or osDefinition is not empty (that is FileGDB v10): get SRID from the
+    //   "srid" field of the _fras_bnd table, and use that has the key to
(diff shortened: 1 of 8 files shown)


ogr/ogrsf_frmts/openfilegdb/gdalopenfilegdbrasterband.cpp
```diff
@@ -444,9 +444,46 @@ bool OGROpenFileGDBDataSource::OpenRaster(const GDALOpenInfo *poOpenInfo,
         m_adfGeoTransform[5] = -dfResY;
     }
 
-    // Get SRID, and fetch WKT from GDBSpatialRefs table
+    // Get SRID, and fetch WKT from GDBSpatialRefs table (for FileGDB v
```
nyalldawson, line 447: Is this supposed to be?

```suggestion
    // Get SRID, and fetch WKT from GDBSpatialRefs table (for FileGDB v9)
```

nyalldawson (approved):
Fantastic!

ogr/ogrsf_frmts/openfilegdb/gdalopenfilegdbrasterband.cpp
rouault: actually not :-). I've expanded the comment to:
```
    // Two cases:
    // - osDefinition is empty (that is FileGDB v9): find the SRS by looking
    //   at the SRS attached to the RASTER field definition of the .gdbtable
    //   file of the main table of the raster (that is the one without fras_XXX
    //   prefixes)
    // - or osDefinition is not empty (that is FileGDB v10): get SRID from the
    //   "srid" field of the _fras_bnd table, and use that has the key to
    //   lookup the corresponding WKT from the GDBSpatialRefs table.
    //   In some cases srid might be 0 (invalid), then we try to get it from
    //   Definition column of the GDB_Items table, stored in osDefinition
```

nyalldawson: Thanks!

PR #7886, opened 2023-06-01
Doc: Migrate remaining program notes from wiki

dbaston:
## What does this PR do?

Migrates program notes from https://trac.osgeo.org/gdal/wiki/UserDocs, except for https://trac.osgeo.org/gdal/wiki/UserDocs/Gdal2Tiles (would take me some time to understand those notes)

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/doc/source/programs/gdal_translate.rst
+++ b/doc/source/programs/gdal_translate.rst
@@ -225,8 +225,7 @@ resampling, and rescaling pixels in the process.
 .. option:: -a_srs <srs_def>
 
     Override the projection for the output file. Can be used with
-    :option:`-a_ullr` to specify the pixel coordinate and size in this
-    projection.
+    :option:`-a_ullr` to specify the extent in this projection.
 
     .. include:: options/srs_def.rst
 
--- a/doc/source/programs/gdal_translate.rst
+++ b/doc/source/programs/gdal_translate.rst
@@ -224,7 +224,9 @@ resampling, and rescaling pixels in the process.
 
 .. option:: -a_srs <srs_def>
 
-    Override the projection for the output file.
+    Override the projection for the output file. Can be used with
+    :option:`-a_ullr` to specify the pixel coordinate and size in this
+    projection.
 
     .. include:: options/srs_def.rst
 
@@ -367,3 +369,9 @@ To create a RGBA dataset from a RGB dataset with a mask
 
     gdal_translate withmask.tif rgba.tif -b 1 -b 2 -b 3 -b mask
 
+
+Subsetting using :option:`-projwin` and :option:`-outsize`:
+
+.. code-block:: bash
+
+   gdal_translate -projwin -20037500 10037500 0 0 -outsize 100 100 frmt_wms_googlemaps_tms.xml junk.png
--- a/doc/source/programs/gdalwarp.rst
+++ b/doc/source/programs/gdalwarp.rst
@@ -364,6 +364,15 @@ with control information.
     in megabytes if the value is less than 10000. For values >=10000, this is
     interpreted as bytes.
 
+    The warper will total up the memory required to hold the input and output
+    image arrays and any auxilary masking arrays and if they are larger than
+    the "warp memory" allowed it will subdivide the chunk into smaller chunks
+    and try again.
+
+    If the -wm value is very small there is some extra overhead in doing many
+    small chunks so setting it larger is better but it is a matter of
+    diminishing returns.
+
 .. option:: -multi
 
     Use multithreaded warping implementation.
@@ -481,7 +490,7 @@ The details of how it is taken into account depends on the resampling kernel:
 
 - for bilinear, cubic, cubicspline and lanczos, for each target pixel, the
   coordinate of its center is projected back to source coordinates and a
-  correspond source pixel is identified. If this source pixel is invalid, the
+  corresponding source pixel is identified. If this source pixel is invalid, the
   target pixel is considered as nodata.
   Given that those resampling kernels have a non-null kernel radius, this source
   pixel is just one among other several source pixels, and it might be possible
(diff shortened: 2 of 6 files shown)


doc/source/programs/gdalwarp.rst
```diff
@@ -493,6 +502,164 @@ The details of how it is taken into account depends on the resampling kernel:
...
+Basically, the fastest situations are nearest neighbour resampling on 8bit data
+without nodata or alpha masking in effect.
+
+.. caution::
+
+   Increasing the value of the :option:`-wm` param may lead to loss of
+   performance in certain circumstances, particularly when gdalwarping *many*
+   small rasters into a big mosaic. See http://trac.osgeo.org/gdal/ticket/3120
+   for more details
+
```
dbaston, line 604: My read of the linked ticket was that the issue is resolved and this note can be removed.

doc/source/programs/gdalwarp.rst
```diff
@@ -493,6 +502,164 @@ The details of how it is taken into account depends on the resampling kernel:
...
+   ERROR 2: Out of memory allocating 365425784 byte destination buffer.
+
+Our understanding is that 32 bit processes are frequently subject to memory
+fragmentation problems and so even in a process with - in theory - 2GB of heap
+RAM space available it is still often difficult to allocate large blocks of
+memory. In this case with :option:`-wm 500` set the main warp buffers can be
+quite large (364MB for the example) and it seems it has failed to allocate the
+buffer. Use more modest buffers (or not use the options at all) or else get a
+64 bit executable for :program:`gdalwarp`.
```
dbaston, line 618: Maybe this discussion about 32-bit systems can be removed?

doc/source/programs/gdal_translate.rst
```diff
@@ -224,7 +224,9 @@ resampling, and rescaling pixels in the process.
 
 .. option:: -a_srs <srs_def>
 
-    Override the projection for the output file.
+    Override the projection for the output file. Can be used with
+    :option:`-a_ullr` to specify the pixel coordinate and size in this
+    projection.
```
rouault, line 229: phrasing sounds odd to me, maybe instead:
```suggestion
    :option:`-a_ullr` to specify the extent in this projection.
```

doc/source/programs/gdalwarp.rst
rouault: yes.

those notes also are quite ancient, and since then, multi-threading warping is available. Increasing the warp memory makes more sense if you also add --config GDAL_NUM_THREADS ALL_CPUS
A bigger value of -wm alone hardly helps. People are always keen to increase such thumbs since they believe "the bigger the better", but I'm quite skeptical it significantly helps in most cases.

doc/source/programs/gdalwarp.rst
```diff
@@ -493,6 +502,164 @@ The details of how it is taken into account depends on the resampling kernel:
...
+   gdal_translate tempfile.vrt outfile.tif -co compress=lzw ...etc.
+
+This can also be combined this into a single operation by writing the VRT to
+``/vsistdout/`` and piping that to :program:`gdal_translate` and specifying
+``/vsistdin/`` as the input. For example:
+
+.. code-block:: bash
+
+   gdalwarp -q -t_srs EPSG:32611 -of vrt input_file.tif /vsistdout/ | gdal_translate -co compress=lzw /vsistdin/ output_file.tif
+
```
rouault, line 659: clever trick, but I'll remove that. Not totally sure it will work if people want to write tiled output. The alternatives presented before are good enough

doc/source/user/configoptions.rst
```diff
@@ -205,8 +205,16 @@ Performance and caching
...
-      This option controls the default GDAL raster block cache size. If its value
-      is small (less than 100000), it is assumed to be measured in megabytes,
+      Controls the default GDAL raster block cache size. When
+      blocks are read from disk, or written to disk, they are cached in a
+      global block cache by the :cpp:class:`GDALRasterBlock` class. Once this
+      cache exceeds :config:`GDAL_CACHEMAX` old blocks are flushed from the
+      cache.
+      This cache is mostly beneficial when needing to read or write blocks
+      several times. This could occur, for instance, in a scanline oriented
+      input file which is processed in multiple chunks (horizontally) by
```
rouault, line 215:
```suggestion
      input file which is processed in multiple rectangular chunks by
```

doc/source/programs/gdalwarp.rst
rouault: yes, that can be removed (still true for the folks on 32bit platforms, but irrelevant to most)

doc/source/programs/gdalwarp.rst
dbaston: > Increasing the warp memory makes more sense if you also add --config GDAL_NUM_THREADS ALL_CPUS

So would it be correct to say "the warp memory specified by -wm is shared among all threads, so it is especially beneficial to increase this value when running gdalwarp with GDAL_NUM_THREADS > 1" ?

Also, over on the config option page `GDAL_NUM_THREADS` has a default of `ALL_CPUS` but I see in `GWKThreadsCreate` it defaults to 1. So I guess that doc should say the default depends on the context?

rouault: > So would it be correct to say "the warp memory specified by -wm is shared among all threads, so it is especially beneficial to increase this value when running gdalwarp with GDAL_NUM_THREADS > 1" ?

sort of... actually what happens is that the number of lines of each warping chunk is split among the threads available. If the warping chunk is too small, the setup cost of the threads can be non neglectable.
Let's settle with your proposed phrasing, as I can't find something better.

> Also, over on the config option page GDAL_NUM_THREADS has a default of ALL_CPUS but I see in GWKThreadsCreate it defaults to 1. So I guess that doc should say the default depends on the context?

yes

dbaston: Thanks. Updated with 

```
The warp memory specified by :option:`-wm` is shared among all
threads, so it is especially beneficial to increase this value when running
:program:`gdalwarp` with :option:`-wo NUM_THREADS` (or its equivalent
:config:`GDAL_NUM_THREADS`) greater than 1."
```

PR #7893, opened 2023-06-04
SWIG JAVA: add Read/WriteRaster abilities for GDT_UInt64 & GDT_Int64

bdezonia:
This commit extends the ReadRaster and WriteRaster methods in the Java API by changes to the SWIG definitions in gdal_java.i.

This is my first commit and is done by inspection. The changes are straightforward. Please let me know what things I need to continue to do. Tests? If so how? Do you need some kind of documentation? Please advise. Thanks.

### DIFF
--- a/swig/include/java/gdal_java.i
+++ b/swig/include/java/gdal_java.i
@@ -254,6 +254,7 @@ static CPLErr DatasetRasterIO( GDALDatasetH hDS, GDALRWFlag eRWFlag,
 {
     if ((gdal_type == GDT_Int16 && buf_type != GDT_Int16 && buf_type != GDT_UInt16 && buf_type != GDT_CInt16) ||
         (gdal_type == GDT_Int32 && buf_type != GDT_Int32 && buf_type != GDT_UInt32 && buf_type != GDT_CInt32) ||
+        (gdal_type == GDT_Int64 && buf_type != GDT_Int64 && buf_type != GDT_UInt64) ||
         (gdal_type == GDT_Float32 && buf_type != GDT_Float32 && buf_type != GDT_CFloat32) ||
         (gdal_type == GDT_Float64 && buf_type != GDT_Float64 && buf_type != GDT_CFloat64))
   {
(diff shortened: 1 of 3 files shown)


swig/include/java/gdal_java.i
```diff
@@ -304,12 +305,14 @@ static CPLErr DatasetRasterIO( GDALDatasetH hDS, GDALRWFlag eRWFlag,
 %apply (char *regularArrayOut, long nRegularArraySizeOut) { (char *regularArrayOut, long nRegularArraySizeOut) };
 %apply (short *regularArrayOut, long nRegularArraySizeOut) { (short *regularArrayOut, long nRegularArraySizeOut) };
 %apply (int *regularArrayOut, long nRegularArraySizeOut) { (int *regularArrayOut, long nRegularArraySizeOut) };
+%apply (long *regularArrayOut, long nRegularArraySizeOut) { (long *regularArrayOut, long nRegularArraySizeOut) };
```
rouault, line 308: - The C type long is only 64bit large on non-Windows 64-bit platforms. A more portable type would be int64_t
Same remark for all other places you edited
- swig/include/java/typemaps_java.i should be modified to define the 2 new typemaps you've declared. See existing typemaps ``(int *regularArrayOut, long nRegularArraySizeOut)`` and  ``(int *regularArrayIn long nRegularArraySizeIn)``
- we'd definitely need some testing. For example in swig/java/apps/GDALTestIO.java, add a new method to write 64-bit integers in a GDT_Int64 MEM dataset and read them back

rouault (changes requested):
Cf comments

rouault:
> Do you need some kind of documentation?

Cf swig/java/javadoc.java which is a collection of the Java prototypes with javadoc. This is processed by the swig/java/add_javadoc.c program that re-inserts this javadoc in the SWIG .java generated file. This should run automatically as part as the "java_binding" CMake target if the ``javadoc`` binary is found. You can check in the CMakeCache.txt file that the GDAL_JAVA_GENERATE_JAVADOC variable is set to ON

The java related tests can be run with ``ctest -R java``

bdezonia:
OK, I have done the things you have asked. Maybe in a way you don't hate. Tests pass. I will squash everything at the point where you are happy with things.

bdezonia:
Let me know if you want my tests pulled out of the test file it is in and instead put it in its own test file.

swig/java/apps/GDALTestIO.java
```diff
@@ -200,4 +202,32 @@ public static void main(String[] args) throws InterruptedException
...
+		
+		long[] data1;
+		long[] data2;
+		
+		int xSz = 5;
+		int ySz = 2;
+		int nBands = 1;
+        Driver driver = gdal.GetDriverByName("MEM");
+        Dataset dataset = driver.Create("fred", xSz, ySz, nBands, gdalconst.GDT_Int64);
+        data1 = new long[]{1,2,3,4,5,6,7,8,9,10};
```
rouault, line 216: just to be sure there's no truncation in the process, I'll put at least one value that doesn't fit on a smaller type, like:
```suggestion
        data1 = new long[]{1,234567890123456,3,4,5,6,7,8,9,10};
```

swig/java/apps/GDALTestIO.java
```diff
@@ -200,4 +202,32 @@ public static void main(String[] args) throws InterruptedException
...
+                throw new RuntimeException("int64 write and read values are not the same "+data1[i]+" "+data2[i]);
+		}
+        data1 = new long[data2.length];
+        data2 = new long[]{10,9,8,7,6,5,4,3,2,1};
+        dataset.WriteRaster(0, 0, xSz, ySz, xSz, ySz, gdalconst.GDT_Int64, data1, new int[]{1});
+        dataset.ReadRaster(0, 0, xSz, ySz, xSz, ySz, gdalconst.GDT_Int64, data2, new int[]{1});
+        for (int i = 0; i < data1.length; i++) {
+			if (data1[i] != data2[i])
+                throw new RuntimeException("int64 write and read values are not the same "+data1[i]+" "+data2[i]);
+		}
```
rouault, line 231: I'm not sure the below tests more than what is above. Using ReadRaster()/WriteRaster() on dataset.GetRasterBand(1) would be better. Something like below (untested by me):

```suggestion
        data1 = new long[data2.length];
        data2 = new long[]{10,9,8,7,6,5,4,3,2,1};
        dataset.GetRasterBand(1).WriteRaster(0, 0, xSz, ySz, xSz, ySz, gdalconst.GDT_Int64, data1);
        dataset.GetRasterBand(1).ReadRaster(0, 0, xSz, ySz, xSz, ySz, gdalconst.GDT_Int64, data2);
        for (int i = 0; i < data1.length; i++) {
			if (data1[i] != data2[i])
                throw new RuntimeException("int64 write and read values are not the same "+data1[i]+" "+data2[i]);
		}
```

bdezonia:
I have addressed your concerns. Please let me know if you think it is ready to be squashed.

Also I am new to working on public github repositories other than my own. Please let me know if my workflow is unorthodox or if there is any project etiquette I am violating. Thanks.

rouault:
> Please let me know if you think it is ready to be squashed.

I'll just use the "squash and merge" github button.

rouault (approved):

PR #7895, opened 2023-06-04
Add missing include <memory>

drons:
Fixes the build if only `cpl_vsi_virtual.h` is included.

#### Environment

* OS: Debian 11
* Compiler: gcc 10.2.1


drons:
Thanks!

PR #7900, opened 2023-06-05
HANA: Set sessionVariable:APPLICATION in connection string

mrylov:


rouault:
@mrylov does this need to be backported?

Issue #7145, opened 2023-01-29
[Docs] OGR_ORGANIZE_POLYGONS info needs to be improved


agiudiceandrea:
## Expected behavior and actual behavior.

The `OGR_ORGANIZE_POLYGONS` configuration option is cited in the [ESRI Shapefile](https://gdal.org/drivers/vector/shapefile.html) doc page in which only the `DEFAULT` value is named.
The description of the other `OGR_ORGANIZE_POLYGONS` configuration option values is missing.

Such option is missing in the [OpenFileGDB](https://gdal.org/drivers/vector/openfilegdb.html) doc page.

rouault:
What is the motivation behind this request ? I'd rather see this as a rather obscure option that should be rarely used.

nyalldawson:
I suspect it's in response to https://twitter.com/totofiandaca/status/1619409200605696000?t=aIHn2gQtqhH-5k7VUK6aGA&s=19

rouault:
Hum, a reproducer dataset would be the first step before concluding one has to mess with OGR_ORGANIZE_POLYGONS :-)

agiudiceandrea:
Actually for the sake of completeness of the GDAL docs.
I saw [this](https://www.facebook.com/pigreco314/posts/pfbid0x3jw1YTqUHeKCBfSWKpwPoMZWN25T8z8EiFb9P9DpXDxsmHESYqttE7UygzT9Qqzl) wondering what is the reason of using OGR_ORGANIZE_POLYGONS option and then I found some pieces of information spread in various places [[1](https://github.com/OSGeo/gdal/issues/1369)] [[2](https://github.com/qgis/QGIS/pull/47705)] [[3](https://gis.stackexchange.com/questions/431106/importing-a-gdb-file-into-postgis-using-ogr2ogr)] [[4](https://gis.stackexchange.com/questions/370739/ogr2ogr-large-gdb-file-to-sqlite-warning-organizepolygons-received-a-polygon)] [[5](https://gdal.org/api/ogrgeometry_cpp.html#_CPPv4N18OGRGeometryFactory16organizePolygonsEPP11OGRGeometryiPiPPKc)], that seems missing in the docs (except in API docs page, where anyway the DEFAULT value is not named).

PR #7907, opened 2023-06-06
Doc: Improve documentation for OGR_ORGANIZE_POLYGONS

dbaston:
Ref: https://github.com/OSGeo/gdal/issues/7145

### DIFF
--- a/doc/source/user/configoptions.rst
+++ b/doc/source/user/configoptions.rst
@@ -411,6 +411,38 @@ General options
 
       Controls whether curved geometries should be approximated by linear geometries.
 
+- .. config:: OGR_ORGANIZE_POLYGONS
+     :choices: DEFAULT, SKIP, ONLY_CCW, CCW_INNER_JUST_AFTER_CW_OUTER
+
+     Defines the method used to classify polygon rings as holes or shells.
+     Although one of the options is named ``DEFAULT``, some drivers may default
+     to a different method to reduce processing by taking advantage of a
+     format's constraints. The following methods are available, in order of
+     decreasing expected runtime:
+
+     - ``DEFAULT``: perform a full analysis of the topological relationships
+       between all rings, classifying them as shells or holes and associating
+       them according to the OGC Simple Features convention. If the topological
+       analysis determines that a valid geometry cannot be constructed, the
+       result will be the same as with :config:`OGR_ORGANIZE_POLYGONS=SKIP`.
+
+     - ``ONLY_CCW``: assume that rings with clockwise orientation represent
+       shells and rings with counterclockwise orientation represent holes.
+       Perform a limited topological analysis to determine which shell contains
+       each hole. The Shapefile driver defaults to this method.
+
+     - ``CCW_INNER_JUST_AFTER_CW_OUTER``: assume that rings with clockwise
+       orientation represent shells and rings with counterclockwise orientation
+       represent holes and immediately follow the outer ring with which they are
+       associated.
+
+     - ``SKIP``: avoid attempting to classify rings as shells or holes. A
+       single geometry (Polygon/MultiPolygon/CurvePolygon/MultiSurface) will be
+       returned with all polygons as top-level polygons. If non-polygonal elements
+       are present, a GeometryCollection will be returned.
+
+
+
 -  .. config:: OGR_SQL_LIKE_AS_ILIKE
       :choices: YES, NO
       :default: NO
--- a/doc/source/drivers/vector/shapefile.rst
+++ b/doc/source/drivers/vector/shapefile.rst
@@ -441,14 +441,6 @@ available:
      the 2 GB file size limit when updating a shapefile. If nothing is set, a
      warning will be emitted when the 2 GB limit is reached.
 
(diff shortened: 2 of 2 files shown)


doc/source/user/configoptions.rst
```diff
@@ -411,6 +411,31 @@ General options
...
+       orientation, assuming that all polygons with a clockwise orientation
+       represent holes. This is the convention of the Shapefile format, whose
+       driver defaults to this method.
+
+     - ``CCW_INNER_JUST_AFTER_CW_OUTER``: Assume inner rings (holes) have
+       counter-clockwise orientation and immediately follow the
+       clockwise-oriented outer ring with which they are associated.
+
+     - ``SKIP``: avoid attempting to classify polygons as shells or holes. A
+       single polygon will be returned with all polygons as top-level polygons.
```
rouault, line 437:
```suggestion
       single multipolygon (or multisurface or geometrycollection, depending on if there are curve polygons or a mix of regular and curve polygons) will be returned with all polygons as top-level polygons.
```

dbaston: A mix of regular and curved polygons would produce a MultiSurface, no? My read was that the GeometryCollection is returned if a line/point finds its way in, or if the polygons do not form a valid multipolygon (crossing edges)

rouault: indeed, you're correct.

jratike80:
Shapefile rather follows the "CCW_INNER_JUST_AFTER_CW_OUTER" convention. Excerpt from the specification:

> The neighborhood to the right of an observer walking along
> the ring in vertex order is the neighborhood inside the polygon. Vertices of rings defining
> holes in polygons are in a counterclockwise direction. Vertices for a single, ringed
> polygon are, therefore, always in clockwise order. 

But maybe the shapefile convention is "exactly one CW ring for the outer and zero or more CCW rings for the holes in any order" because the spec says also "The order of rings in the points array is not significant.".

rouault:
From my memory, rings in shapefiles can be in a somewhat random order, so CCW_INNER_JUST_AFTER_CW_OUTER is likely to be unreliable for them (although I believe it should work fine for shapefiles generated by OGR)

rouault:
> Does it mean that that the best for shapefiles would be "CW only"?

yes, that's what is used: https://github.com/OSGeo/gdal/blob/8cbf7e1a9a4dc752417da85e78e42b3480768714/ogr/ogrsf_frmts/shape/shape2ogr.cpp#L330

doc/source/user/configoptions.rst
dbaston: Updated with "A single geometry (MultiPolygon or MultiCurve) will be returned with all polygons as top-level polygons."

dbaston:
Would it be valuable to add more tests for this code outside of `autotest/ogr/ogr_shape.py` ? If I understand right, `OGRGeometryFactory` is not accessible from Python, so the tests would need to be done in C++ ?

doc/source/user/configoptions.rst
```diff
@@ -411,6 +411,33 @@ General options
...
+
+     - ``DEFAULT``: perform a full analysis of the topological relationships
+       between all polygons, classifying them as shells or holes and associating
+       them according to the OGC Simple Features convention.
+
+     - ``ONLY_CCW``: only attempt to classify polygons with a counter-clockwise
+       orientation, assuming that all polygons with a clockwise orientation
+       represent holes. This is the convention of the Shapefile format, whose
+       driver defaults to this method.
+
```
jratike80, line 431: This is **not** the Shapefile convention, the shapefile convention is exactly reversed: Clockwise rings are the outer rings, counterclockwise rings are holes. If the shapefile driver is using this method then either the name "ONLY_CCW" or the description is misleading.

doc/source/user/configoptions.rst
rouault: MultiCurve cannot be emitted. The 4 possiblye types are:
- Polygon
- MultiPolyon
- MultiSurface
- GeometryCollection: if presence of non-polygon in input or invalid topology

doc/source/user/configoptions.rst
rouault: Jukka is right. I read the proposed doc addition too fast. The Doxygen of the function mentions "or only make it analyze counterclockwise polygons by adding METHOD=ONLY_CCW to the option list if you can assume that the outline of holes is counterclockwise defined (this is the convention for example in shapefiles, Personal Geodatabases or File Geodatabases)."

I realize this might not be crystal clear. Let's try to explain it better. In METHOD=ONLY_CCW mode, rings with clockwise orientation are assumed to be outer rings, and those with counterclockwise orientation to be inner rings. METHOD=ONLY_CCW  means that a topological analysis is only needed for those counterclockwise oriented rings to figure out to which outer ring they belong too

rouault:
> Would it be valuable to add more tests for this code outside of `autotest/ogr/ogr_shape.py` ? If I understand right, `OGRGeometryFactory` is not accessible from Python, so the tests would need to be done in C++ ?

yes potentially

PR #7912, opened 2023-06-07
autotest: Add some tests for OGRGeometryFactory::organizePolygons

dbaston:
## What does this PR do?

Adds some tests for `OGRGeometryFactory::organizePolygons`. These are not exhaustive but document the basic behavior as well as some edge cases.

I commented out asserts that I would have expected to pass but do not. There are two categories of these:

1. The `pbIsValidGeometry` parameter does not behave how I would expect. It is described as "value will be set TRUE if result is valid or FALSE otherwise", but there are a number of situations where this value is `TRUE` but the result is invalid, e.g.:

- `organizePolygons` was called with a single invalid input
- `organizePolygons` was called with a method that does not perform topological validation
- `organizePolygons` was called with a method that performs topological validation, but the invalidity is not detected because `OGR_DEBUG_ORGANIZE_POLYGONS` is set to `FALSE`
- `organizePolygons` was called with a method that performs topological validation and `OGR_DEBUG_ORGANIZE_POLYGONS=YES` but the nature of the invalid result is not detected (e.g., hole touches shell at multiple points)

I don't see `pbIsValidGeometry` being used inside GDAL except to emit an error message similar to one that would be emitted inside `organizePolygons`, so maybe it makes sense to simply change the documentation of `pbIsValidGeometry` to "value may be set to FALSE if an invalid result is detected; TRUE does not indicate that the result is valid". Or remove the parameter entirely, if it's OK to change the signature.

2. The function returns a type not among those listed in the documentation (OGRPolygon, OGRCurvePolygon, OGRMultiPolygon, OGRMultiSurface or OGRGeometryCollection). For example, calling `organizePolygons` with a single `OGRPoint` will return the `OGRPoint`.

Without the change I made on line 1633, passing a non-polygonal geometry causes an assertion failure when `OGRGeometry::toCurvePolygon()` is called. So this may indicate that nothing is relying on this behavior and support for non-polygonal inputs can be removed.

#7907

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/autotest/cpp/CMakeLists.txt
+++ b/autotest/cpp/CMakeLists.txt
@@ -79,6 +79,7 @@ add_executable(
   test_gdal_gtiff.cpp
   test_gdal_pixelfn.cpp
   test_ogr.cpp
+  test_ogr_organize_polygons.cpp
   test_ogr_geometry_stealing.cpp
   test_ogr_lgpl.cpp
   test_ogr_geos.cpp
--- a/autotest/cpp/test_ogr_organize_polygons.cpp
+++ b/autotest/cpp/test_ogr_organize_polygons.cpp
@@ -1,0 +1,304 @@
+/**********************************************************************
+ *
+ * Project:  GDAL
+ * Purpose:  Test OGRGeometryFactory::organizePolygons
+ * Author:   Daniel Baston <dbaston at gmail.com>
+ *
+ **********************************************************************
+ * Copyright (c) 2023, ISciences LLC
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a
+ * copy of this software and associated documentation files (the "Software"),
+ * to deal in the Software without restriction, including without limitation
+ * the rights to use, copy, modify, merge, publish, distribute, sublicense,
+ * and/or sell copies of the Software, and to permit persons to whom the
+ * Software is furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included
+ * in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL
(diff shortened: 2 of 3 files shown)


rouault:
> maybe it makes sense to simply change the documentation of `pbIsValidGeometry` to "value may be set to FALSE if an invalid result is detected; TRUE does not indicate that the result is valid"

The IsValid term in the parameter is probably misleading indeed, as one might assume this is GEOSISValid(), whereas it is more IsValid here is closer to "not invalid in an obvious way". So your proposal sounds good, maybe adding something like: "validity checks are limited to what is strictly needed for the algorithm to link inner rings to outer rings, and do not guarantee that OGRGeometry::IsValid() returns true"

> for example, calling organizePolygons with a single OGRPoint will return the OGRPoint.

I don't think we need the method to be robust to anything different that OGRPolygon or OGRCurvePolygon, given the name of the method ("organizePolygons") and the "@param papoPolygons array of geometry pointers - should all be OGRPolygons." .
Actually the should be ammended to mention OGRCurvePolygon as well
papoPolygons array of geometry pointers - should all be OGRPolygons

dbaston:
Thanks, I will update the function docs per your suggestion and remove testing of `pbIsValidGeometry`.

> I don't think we need the method to be robust to anything different that OGRPolygon or OGRCurvePolygon

Yes, this makes sense. Right now there is some code that attempts to handle other inputs (variable `bMixedUpGeometries`) but it never gets used because an assertion failure occurs first. Instead of getting past the assertion failure (as I did in this PR) should I remove the handling of non-polygonal types? I can also leave it alone if you'd rather limit this PR to the test additions.

rouault:
> should I remove the handling of non-polygonal types?

sounds good to me

dbaston:
OK, [0a7f56a](https://github.com/OSGeo/gdal/pull/7912/commits/0a7f56ab3fa3337c2ea5d3a9a6617a9f37004051) has `orgranizePolygons` emit an error but otherwise ignore any non-poylgonal inputs. I'm not sure if it succeeds in its goal of simplifying that function, though.

rouault:
mingw-msys2 doesn't seem to like:
```
./../ogr/ogrgeometryfactory.cpp: In static member function 'static OGRGeometry* OGRGeometryFactory::organizePolygons(OGRGeometry**, int, int*, const char**)':
../../ogr/ogrgeometryfactory.cpp:1779:37: error: unknown conversion type character 'z' in format [-Werror=format=]
 1779 |      
```

rouault:
@dbaston  still a WIP or are we good to merge ?

dbaston:
@rouault I removed the WIP label, ok with me if you want to merge it.

PR #7927, opened 2023-06-12
GML: fix 3.7.0 regression when reading files with several layers, …

### LINKED ISSUE #7925: OGR GML driver uses bbox instead of actual feature geometry
Not entirely sure if this is a GDAL issue, so please close here if it's QGIS related and I'll file it there.

## Expected behavior

OGR GML driver uses actual feature geometry.

## Actual behavior

OGR GML driver uses bbox instead of actual geometry, GFS from GDAL 3.7.0 includes the following compared to the GFS from GDAL 3.6.2:

grafik

Drag'n'drop the attached `test.gml` in QGIS. GFS from GDAL 3.6.2 and 3.7.0 also attached.

[test_data.zip](https://github.com/OSGeo/gdal/files/11723965/test_data.zip)

Result after drag'n'drop in QGIS:
grafik

## Operating system

Windows 10 Version 2009

3.7.0 installed with QGIS standalone installer 3.30.3

rouault:
…whose features have gml:boundedBy elements and the geometry element name is different between the different layers (fixes #7925)


pathmapper:
Thanks for the fix and fast response @rouault !

Issue #7572, opened 2023-04-10
autotest: make ogrtest.check_feature() and ogrtest.check_feature_geometry() use assert


rouault:
Those methods currently return 0 on success and 1 on failure, and use gdaltest.post_reason() to report the reason of the failure, which is no longer used.
They should be modified to act as assertions, and use assert internally

dbaston:
Completed in https://github.com/OSGeo/gdal/pull/7944

PR #7930, opened 2023-06-13
autotest: Add attribute/spatial filter context managers

dbaston:
## What does this PR do?

Introduces context managers for attribute and spatial filters to `gdaltest` as an alternative to the version proposed in https://github.com/OSGeo/gdal/pull/7676.

The motivation is the same as described in that PR -- allowing asserts to be performed immediately, rather than deferred until after cleanup steps are performed. This is needed for #7572.

The PR also removes deferred asserts found by the `tr = something, ..., assert tr` pattern.

#7676

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/autotest/alg/polygonize.py
+++ b/autotest/alg/polygonize.py
@@ -72,23 +72,20 @@ def test_polygonize_1(is_int_polygonize):
 
     expect = [107, 123, 115, 115, 140, 148, 123, 140, 100, 101, 102, 156, 103]
 
-    tr = ogrtest.check_features_against_list(mem_layer, "DN", expect)
+    assert ogrtest.check_features_against_list(mem_layer, "DN", expect)
 
     # check at least one geometry.
-    if tr:
-        mem_layer.SetAttributeFilter("dn = 156")
-        feat_read = mem_layer.GetNextFeature()
-        if (
-            ogrtest.check_feature_geometry(
-                feat_read,
-                "POLYGON ((440720 3751200,440720 3751020,440900 3751020,440900 3751200,440720 3751200),(440780 3751140,440840 3751140,440840 3751080,440780 3751080,440780 3751140))",
-            )
-            != 0
-        ):
-            tr = 0
-        feat_read.Destroy()
+    mem_layer.SetAttributeFilter("dn = 156")
+    feat_read = mem_layer.GetNextFeature()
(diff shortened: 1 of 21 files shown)


autotest/pymod/gdaltest.py
```diff
@@ -1938,6 +1938,45 @@ def tempfile(filename, content):
...
+
+
+###############################################################################
+# Temporarily set a spatial filter
+# Single argument is parsed as WKT or assumed to be an OGRGeometry
+# Four arguments are interpreted as bounding rectangle
+
+
+@contextlib.contextmanager
+def spatial_filter(lyr, *args):
```
rouault, line 1961: could we move that to the ogrtest module instead ?

```suggestion
###############################################################################
# Temporarily set an attribute filter


@contextlib.contextmanager
def attribute_filter(lyr, filter_txt):
    lyr.SetAttributeFilter(filter_txt)
    try:
        yield
    finally:
        lyr.SetAttributeFilter(None)


###############################################################################
# Temporarily set a spatial filter
# Single argument is parsed as WKT or assumed to be an OGRGeometry
# Four arguments are interpreted as bounding rectangle


@contextlib.contextmanager
def spatial_filter(lyr, *args):
```

autotest/pymod/gdaltest.py
```diff
@@ -1938,6 +1938,45 @@ def tempfile(filename, content):
...
+
+    if len(args) == 1:
+        if type(args[0]) is str:
+            geom = ogr.CreateGeometryFromWkt(args[0])
+            lyr.SetSpatialFilter(geom)
+            geom.Destroy()
```
rouault, line 1967: not needed:
```suggestion
```

autotest/pymod/gdaltest.py
dbaston: Yes, that seems like the more natural home. Though I had wondered whether the separation between `gdaltest` and `ogrtest` is useful at this point or if `ogrtest` could just be folded into `gdaltest`.

rouault: having both modules has the advantage of having 2 smaller files rather than a larger one

dbaston: I've moved it over.

PR #7938, opened 2023-06-14
Add read/write vector driver for MVT vector tiles in PMTiles archive

rouault:
This driver supports reading and writing [PMTiles](https://github.com/protomaps/PMTiles) datasets containing vector tiles, encoded in the MapVector Tiles (MVT) format.

PMTiles is a single-file archive format for tiled data. A PMTiles archive can be hosted on a commodity storage platform such as S3, and enables low-cost, zero-maintenance map applications that are "serverless" - free of a custom tile backend or third party provider.

Note that, similarly to the MVT and MBTiles drivvers, the PMTiles driver will make no effort of stitching together geometries for
linear or polygonal features that overlap several tiles. An application that wishes to eliminate those interrupts could potentially use the CLIP=NO open option to get larger boundaries, and use appropriate clipping graphic primitives to hide those discontinuities.

Direct conversion from MBTiles to PMTiles is possible through ogr2ogr (if using no option of ogr2ogr)

Generic write support, relying on the capabilities of the MVT driver to generate vector tiles from OGR features, is also available, using a temporary MBTiles vector dataset, that is translated to PMTiles.


nvkelso:
Awesome work, I'm very excited to use this. Doubly so for ease of use in QGIS :)

Will this support PNG and other raster formatted tiles, too, out of the gate? Or will that be a followup? Or you recommend COGs instead for that purpose?

rouault:
> Will this support PNG and other raster formatted tiles, too, out of the gate?

no

> Or will that be a followup?

not planned / funded at that stage

> Or you recommend COGs instead for that purpose?

yes, COG should do a similar job

rouault:
> Doubly so for ease of use in QGIS :)

The note I mentioned "the PMTiles driver will make no effort of stitching together geometries for linear or polygonal features that overlap several tiles" is a practical limitation for use in QGIS. You will see polygons cut at tile boundaries. Proper display would probably require a specific development in QGIS, like its native MBTiles provider, or some hint provided by GDAL and consumed by QGIS so it knows the geometries it get come from a tiled provider and thus apply proper clipping to avoid seeing tile boundaries

doc/source/drivers/vector/pmtiles.rst
```diff
@@ -0,0 +1,337 @@
...
+zero-maintenance map applications that are "serverless" - free of a custom tile
+backend or third party provider.
+
+This driver is compatible with all GDAL
+:ref:`network-based virtual file systems <network_based_file_systems>`
+
+The SRS is always the `Pseudo-Mercator <https://en.wikipedia.org/wiki/Web_Mercator_projection>`__
+(a.k.a Google Mercator) projection, EPSG:3857.
+
+The supported datasets must contain a JSON metadata document
```
bdon, line 25: This should be OK but is technically not part of the PMTiles spec yet - see [3.1 PR](https://github.com/protomaps/PMTiles/pull/123) since I have been waffling on the exact definition of `fields` here.

doc/source/drivers/vector/pmtiles.rst
```diff
@@ -0,0 +1,337 @@
...
+.. supports_georeferencing::
+
+.. supports_virtualio::
+
+Opening options
+---------------
+
+The following open options are available:
+
+-  .. oo:: CLIP
```
bdon, line 50: Does `CLIP=YES/NO` only apply to writing a pmtiles and not reading? Based on my understanding this should be:

Vector tiles are generally produced with a buffer that provides overlaps between adjacent tiles, and can be used to display them properly, **hence the default behavior of clipping.** When using vector tiles as a vector layer source, like in the OGR vector model, this padding is undesirable.

doc/source/drivers/vector/pmtiles.rst
```diff
@@ -0,0 +1,337 @@
...
+     Whether to auto-select the zoom level for vector layers according to the
+     spatial filter extent.
+     Only for display purpose.
+
+-  .. oo:: JSON_FIELD
+     :choices: YES, NO
+     :default: NO
+
+     Whether tile attributes should be serialized in a single ``json`` field
+     as JSON. This may be useful is tiles may have different attribute schemas.
```
bdon, line 81: "is tiles may have" > "if tiles have"

ogr/ogrsf_frmts/mvt/mvtutils.h
```diff
@@ -60,23 +60,30 @@
...
     "default='500000' "                                                        \
     "description='Maximum size of a tile in bytes'/>"                          \
     "  <Option name='MAX_FEATURES' scope='vector' type='unsigned int' "        \
     "min='1' default='200000' "                                                \
     "description='Maximum number of features per tile'/>"
 
+#define MVT_MBTILES_COMMON_DSCO                                                \
+    MVT_MBTILES_PMTILES_COMMON_DSCO                                            \
+    "  <Option name='COMPRESS' scope='vector' type='boolean' description="     \
+    "'Whether to deflate-compress tiles' default='YES'/>"                      \
```
bdon, line 73: In [other places](https://gdal.org/drivers/vector/mvt.html) the GDAL docs refer to DEFLATE/gzip interchangeably though the [PMTiles spec](https://github.com/protomaps/PMTiles/blob/main/spec/v3/spec.md) only defines gzip as a well-known option. The requirements are driven by the [Web DecompressionStream standard](https://developer.mozilla.org/en-US/docs/Web/API/DecompressionStream/DecompressionStream) which on most platforms does support raw `deflate` but I haven't found a compelling reason to support both.

bdon:
FYI the tile boundaries display affects QGIS for MBTiles already like this: (it doesn't seem to choose a zoom level dynamically either)
Screenshot 2023-06-19 at 14 09 50

I opened this via layer > Add Vector Layer > choosing the mbtiles on disk, which seems to use the GDAL driver and not the QGIS native one. 

Will comment about QGIS in https://github.com/protomaps/PMTiles/issues/177

jratike80:
A comment from https://docs.mapbox.com/data/tilesets/guides/vector-tiles-standards/

> A common question, when it comes to clipping is "how do renderers know which lines to connect for clipped geometry?". This is the reason Mapbox adds a buffer to vector tiles and clipped geometry. When it is time to render, Mapbox sets the canvas to the exact tile size, which sets the edges outside of the visual frame, thus the tiles all line up. So, there is no need to know which nodes are part of others for rendering purposes. Although, you could use the id field in the protobuf to store information necessary for reconstructing polygons.

QGIS does not implement similar rendering trick.

doc/source/drivers/vector/pmtiles.rst
rouault: > Does `CLIP=YES/NO` only apply to writing a pmtiles and not reading?

no, CLIP is really a reading / open option. The corresponding creation option is BUFFER documented later in the document.
I'm not totally sure I agree with your change. For something that would display tiles, they would do their rendering of the tile with the extra buffer around it, taking into account line styles etc and then clip to the actual tile extent. So if you use the OGR output for rendering, you don't want OGR to do the clipping itself. But if you want OGR to return features for possibly other purposes than rendering, the clipping is preferable.

doc/source/drivers/vector/pmtiles.rst
rouault: fixed. thanks

rouault:
> I opened this via layer > Add Vector Layer > choosing the mbtiles on disk, which seems to use the GDAL driver and not the QGIS native one.

yes, buf if you drag & drop the file, the native QGIS MBTiles vector tile renderer will be used instead

doc/source/drivers/vector/pmtiles.rst
bdon: I have harmonized this in [3.1](https://github.com/protomaps/PMTiles/blob/main/spec/v3/spec.md#metadata) with [TileJSON](https://github.com/mapbox/tilejson-spec/tree/master/3.0.0#33-vector_layers), so there's no surprises here.

PR #7944, opened 2023-06-15
autotest: Use assert in ogrtest helper functions

dbaston:
## What does this PR do?

Converts ogrtest helper functions to use `assert` instead of return codes, as described in https://github.com/OSGeo/gdal/issues/7572. Updates call sites to remove checking of return values.

https://github.com/OSGeo/gdal/issues/7572

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments


rouault:
Thanks !
This fully closes #7572 right ?

PR #7949, opened 2023-06-15
autotest: remove gdaltest.post_reason

dbaston:
## What does this PR do?

Removes `gdaltest.post_reason`, largely by moving the information it would provide into `assert` statements.

https://github.com/OSGeo/gdal/issues/7573 is resolved by this PR.

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/autotest/pyscripts/test_gdal_ls_py.py
+++ b/autotest/pyscripts/test_gdal_ls_py.py
@@ -194,7 +194,6 @@ def test_gdal_ls_py_6(script_path):
         # FIXME
         # Fails on Travis with dates at 1970-01-01 00:00
         # Looks like a 32/64bit issue with Python bindings of VSIStatL()
-        pytest.skip()
         pytest.fail(ret_str)
 
 
--- a/autotest/pyscripts/test_gdal_ls_py.py
+++ b/autotest/pyscripts/test_gdal_ls_py.py
@@ -124,11 +124,10 @@ def test_gdal_ls_py_4(script_path):
         )
         == -1
     ):
-        if gdaltest.skip_on_travis():
-            # FIXME
-            # Fails on Travis with dates at 1970-01-01 00:00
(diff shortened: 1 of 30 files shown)


rouault:
> ```
> 
> ```
> 
> 
>     
>   
> 
> which I have seen several times over the past day.

ok, looking at the log and the test, those tests checks the value of the Authorization header that has been computed only when the HTTP server is 8080. Here it is 8081 (presumably because unable to reuse one that was run on 8080 before). Easy "fix": add a skip for those tests if port != 8080.  Or potentially add alternative expected Authorization header for 8081 (by forcing the port to be 8081 to get the expected value)

autotest/pyscripts/test_gdal_ls_py.py
```diff
@@ -191,11 +190,11 @@ def test_gdal_ls_py_6(script_path):
...
-        if gdaltest.skip_on_travis():
-            # FIXME
-            # Fails on Travis with dates at 1970-01-01 00:00
-            # Looks like a 32/64bit issue with Python bindings of VSIStatL()
-            pytest.skip()
+        gdaltest.skip_on_travis()
+        # FIXME
+        # Fails on Travis with dates at 1970-01-01 00:00
+        # Looks like a 32/64bit issue with Python bindings of VSIStatL()
+        pytest.skip()
```
rouault, line 197: Looks like this is not needed (looking at similar case just above)
```suggestion
```

rouault:
> ok, looking at the log and the test, those tests checks the value of the Authorization header that has been computed only when the HTTP server is 8080. Here it is 8081

issue created at https://github.com/OSGeo/gdal/issues/7952

rouault:
> which I have seen several times over the past day.

hum, just thinking it is probably related to 0012beb917b8c906357af1590718cb3f0a2418d7, where some module would start a http server on 8080 while vsis3 also needs one

Issue #7947, opened 2023-06-15
Feature Request [GML]: Make "Use geometry in boundedBy element" an open option 


pathmapper:
What was the intention for the change in https://github.com/OSGeo/gdal/pull/7133?

If it doesn't conflict with the original intention, maybe make it an open option with `default = false`?

Came across this because there is a need to adapt an XPlanung QGIS plugin to restore the original behaviour, if this change stays like it is currently (with XPlanung-schema it happens often that there is a feature with bbox but without geoemtry).

Also, currently there might be confusion, e.g. when you drag'n drop a GML like the [test data](https://github.com/OSGeo/gdal/files/11723965/test_data.zip) from #7925 in QGIS, the import dialog shows a polygon feature for `BP_Bereich` without notifying the user that this is the bbox and not the actual feature geometry (or that the layer has no feature with geometry).

<img src="https://github.com/OSGeo/gdal/assets/20856381/3154ec06-3abb-4e2a-aa7c-f5256f26e969" width="500px">

rouault:
> What was the intention for the change in #7133?

There are some data models like https://github.com/rouault/gdal/blob/a35c3ab224f55dce7af060a1d1b7279fec24dbf8/autotest/ogr/data/gml/only_boundedby.gml which is a simplified version of some Ordnance Survey files where getting the bbox as a geometry was desirable

> If it doesn't conflict with the original intention, maybe make it an open option with default = false?

I must check with the user who requested the behaviour of #7133

rouault:
ok, so the user confirmed a "-oo USE_BBOX=YES" would work for him

Issue #2084, opened 2019-12-04
defective presentation from *.xml files


MalteEBA:
Describe the bug
If I load xml files out of this zip file (see below) in QGIS, the buildings aren't displayed correct. There are a lot of buldings missing or corrupt. I know, that this file contains all informations, cause with other programms (FME) I can extract them. I am not the only one, who has problems with xml files.
see also: https://github.com/qgis/QGIS/issues/33222

How to Reproduce
https://fbinter.stadt-berlin.de/fb/berlin/service_intern.jsp?id=a_lod1@senstadt&type=FEED
http://fbarc.stadt-berlin.de/FIS_Broker_Atom/lod1/LoD1.zip

QGIS and OS versions
3.10 and
3.4.5-Madeira

Exemplary I toke 391_5820 and 5821 + 390_5820 and 5821 -these are displayed in red. In turquoise u can see the informations, which should be displayed.
![2019-12-04 10_18_01-_Unbenanntes Projekt - QGIS](https://user-images.githubusercontent.com/58505266/70133768-0d6f1080-1687-11ea-96e9-ae5821ead92b.png)

jratike80:
Could you make a test. GMLAS driver finds 976 buildings from file lod1_391_5820.xml. How many buildings does FME find?

```
ogrinfo GMLAS:lod1_391_5820.xml -oo remove_unused_fields=yes -oo remove_unused_layers=yes building -so
INFO: Open of `GMLAS:lod1_391_5820.xml'
      using driver `GMLAS' successful.

Layer name: building
Geometry (lod1solid): PolyhedralSurface
Geometry (lod1terrainintersection): Multi Curve
Feature Count: 976
Extent (lod1solid): (391005.329000, 5820001.883000) - (392036.155000, 5821047.638000)
SRS WKT (lod1solid):
(unknown)
SRS WKT (lod1terrainintersection):
(unknown)
Geometry Column 1 = lod1solid
Geometry Column 2 = lod1terrainintersection
ogr_pkid: String (0.0) NOT NULL
id: String (0.0)
creationdate: Date (0.0)
function: String (0.0)
measuredheight_uom: String (0.0)
measuredheight: Real (0.0)
storeysaboveground: Integer (0.0)
```

AndreM84:
Hey folks,

I was wondering, why this issue is opened fo so long, since I still cannot find good manuals/tutorials for GDAL-based transformation from CityGML.
[...]

PR #7954, opened 2023-06-15
GML: add a USE_BBOX=YES open option to allow using gml:boundedBy as feature geometry

rouault:
Fixes https://github.com/OSGeo/gdal/issues/7947
    
That capability was added by default in #7133, but there are use cases (like reading XPlannung datasets) where this is desirable, so make it available only if requested with the new USE_BBOX=YES open option.

### DIFF
--- a/ogr/ogrsf_frmts/gml/gmlhandler.cpp
+++ b/ogr/ogrsf_frmts/gml/gmlhandler.cpp
@@ -1275,7 +1275,13 @@ OGRErr GMLHandler::startElementFeatureAttribute(const char *pszName,
             return startElementGeometry(pszName, nLenName, attr);
         }
     }
-    else if (nLenName == 9 && strcmp(pszName, "boundedBy") == 0)
+    else if (nLenName == 9 && strcmp(pszName, "boundedBy") == 0 &&
+             // We ignore the UseBBOX() flag for CityGML, since CityGML
+             // has elements like bldg:boundedBy, which are not a simple
+             // rectangular bbox. This is needed to read properly
+             // autotest/ogr/data/gml/citygml_lod2_713_5322.xml
+             // (this is a workaround of not being namespace aware)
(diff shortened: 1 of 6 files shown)


ogr/ogrsf_frmts/gml/gmlhandler.cpp
```diff
@@ -1275,7 +1275,8 @@ OGRErr GMLHandler::startElementFeatureAttribute(const char *pszName,
             return startElementGeometry(pszName, nLenName, attr);
         }
     }
-    else if (nLenName == 9 && strcmp(pszName, "boundedBy") == 0)
+    else if (nLenName == 9 && strcmp(pszName, "boundedBy") == 0 &&
+             (eAppSchemaType == APPSCHEMA_CITYGML || m_poReader->UseBBOX()))
```
pathmapper, line 1279: Maybe document that the open option `USE_BBOX=NO` has no effect for CityGML ?

Is this because of https://github.com/OSGeo/gdal/issues/2084?

pathmapper (commented):
Thanks a bunch!

ogr/ogrsf_frmts/gml/gmlhandler.cpp
rouault: > Maybe document that the open option `USE_BBOX=NO` has no effect for CityGML ?

I've added this comment in the code:
             // We ignore the UseBBOX() flag for CityGML, since CityGML
             // has elements like bldg:boundedBy, which are not a simple
             // rectangular bbox. This is needed to read properly
             // autotest/ogr/data/gml/citygml_lod2_713_5322.xml
             // (this is a workaround of not being namespace aware)

So I don't think this deserves a user doc addition

> Is this because of https://github.com/OSGeo/gdal/issues/2084? > 

no, because of the fix for qgis/QGIS#51647

PR #7960, opened 2023-06-16
autotest: Enhancements to ogrtest.check_feature_geometry

dbaston:
## What does this PR do?

Add "pointwise" flag to skip GEOS topological equality testing. This may be needed in some cases where it is important to assert a specific ring ordering or orientation.

Improve assertion messages by allowing pytest to format the differences between failed vertices, and printing the entire WKT for the actual and expected geometries.

Rename internal variables for clarity.

https://github.com/OSGeo/gdal/pull/7945#discussion_r1230993442

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/autotest/ogr/ogr_ogrtest.py
+++ b/autotest/ogr/ogr_ogrtest.py
@@ -1,0 +1,114 @@
+#!/usr/bin/env pytest
+# -*- coding: utf-8 -*-
+###############################################################################
+# $Id$
+#
+# Project:  GDAL/OGR Test Suite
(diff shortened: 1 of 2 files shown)


rouault:
wondering if we wouldn't need some tests for this assertion method as it is becoming non trivial ? (at least testing a few *failure* modes, as success must be tested quite extensively)

dbaston:
I've been using a script on the side. I'll bring it into pytest and add it here.

rouault:
Excellent, thanks!

Issue #7952, opened 2023-06-15
autotest: failure on some vsis3.py tests when run on port 8081 


rouault:
See https://github.com/OSGeo/gdal/pull/7949#issuecomment-1593477279 for details

PR #7965, opened 2023-06-17
Python bindings: Invalidate band refs when dataset closes

dbaston:
## What does this PR do?

    Modify gdal.Dataset to track gdal.Band objects it creates. When
    gdal.Dataset closes, extant gdal.Band objects will be invalidated. If an
    invalidated gdal.Band object is used, it will throw an exception instead
    of creating a segfault.

I think this is the approach described in https://github.com/OSGeo/gdal/pull/7697#issuecomment-1535461854. I have a little hesitation about the list of references in `Dataset` growing without bound. This could be avoided by having bands cleanup their own references in Dataset (if they go out of scope first) or by having `GetRasterBand` return the same `Band` object if it is called multiple times with the same index.

The exception thrown when an invalidated `Band` is used is uninformative (`TypeError: in method 'Band_Checksum', argument 1 of type 'GDALRasterBandShadow *'`) but I think a bad exception is better than a segfault. We could also provide our own message by `%prepend`-ing a check on the methods of `Band`.

https://github.com/OSGeo/gdal/pull/7697#issuecomment-1535461854

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/autotest/gcore/basic_test.py
+++ b/autotest/gcore/basic_test.py
@@ -771,3 +771,22 @@ def test_create_context_manager(tmp_path):
 
     ds_in = gdal.Open(fname)
     assert ds_in.GetRasterBand(1).Checksum() != 0
+
+
+def test_band_use_after_dataset_close_1():
+    ds = gdal.Open("data/byte.tif")
+    band = ds.GetRasterBand(1)
+    del ds
+
+    # Make sure "del ds" has invalidated "band" so we don't crash here
+    with pytest.raises(Exception):
+        band.Checksum()
+
+
+def test_band_use_after_dataset_close_2():
+    with gdal.Open("data/byte.tif") as ds:
+        band = ds.GetRasterBand(1)
+
+    # Make sure ds.__exit__() has invalidated "band" so we don't crash here
+    with pytest.raises(Exception):
+        band.Checksum()
--- a/autotest/ogr/ogr_basic_test.py
+++ b/autotest/ogr/ogr_basic_test.py
@@ -917,6 +917,29 @@ def test_ogr_basic_create_data_source_context_manager(tmp_path):
 
 
 ###############################################################################
+# check layer access after datasource has closed
+
+
+def test_layer_use_after_datasource_close_1():
+    with ogr.Open("data/poly.shp") as ds:
+        lyr = ds.GetLayer(0)
+
+    # Make sure ds.__exit__() has invalidated "lyr" so we don't crash here
+    with pytest.raises(Exception):
+        lyr.GetFeatureCount()
+
+
+def test_layer_use_after_datasource_close_2():
+    ds = ogr.Open("data/poly.shp")
+    lyr = ds.GetLayerByName("poly")
+
+    del ds
+    # Make sure ds.__del__() has invalidated "lyr" so we don't crash here
+    with pytest.raises(Exception):
+        lyr.GetFeatureCount()
+
+
+###############################################################################
 # cleanup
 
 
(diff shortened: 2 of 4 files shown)


rouault:
> I have a little hesitation about the list of references in `Dataset` growing without bound. This could be avoided by having bands cleanup their own references in Dataset (if they go out of scope first) or by having `GetRasterBand` return the same `Band` object if it is called multiple times with the same index.

I too. I guess making _band_references a dict of band_number -> weak_ref of Band would solve this.

dbaston:
> I too. I guess making _band_references a dict of band_number -> weak_ref of Band would solve this.

I think that also implies we must return the same `Band` object each time we call `GetRasterBand` with the same index?

rouault:
> I think that also implies we must return the same `Band` object each time we call `GetRasterBand` with the same index?

yes, I can't see why this would be a problem ?

(note, but not needed for this PR, we have a similar issue with the Band.GetMaskBand() method which returns a C++ GDALRasterBand (mask band) that is owned by the C++ GDALRasterBand on which GetMaskBand() has been called)

dbaston:
> I can't see why this would be a problem ?

I don't see a problem either, just confirming before I implement it that way.

dbaston:
The `DataSource` case is definitely trickier because we may request and delete layers either by index or by name. Instead of tracking references by either layer index or name, I tried tracking them based on the layer's location in memory. To remove references when `DeleteLayer` is called, I need to call `GetLayer` to find out what the location in memory is. It seems like the APISPY test isn't able to deal with the case where one API method calls another internally. I guess I could avoid cleaning up references on `DeleteLayer`, or modify the APISPY test to try and ignore the extra `GetLayer` call.

rouault:
> or modify the APISPY test to try and ignore the extra `GetLayer` call.

the APISPY is definitely not something "sacred". It used to be useful when debugging some bugs when using some OGR drivers with QGIS and being able to replicate what QGIS did outside of it, but I haven't used it in years.
You could definitely use some filtering on the ref_data or got_data in the failed tests to add / remove the needed extra GetLayer() calls.

dbaston:
Turns out this whole thing can be greatly simplified by `weakref.WeaksSet` : https://docs.python.org/3/library/weakref.html#weakref.WeakSet

Updated to use that now.

dbaston:
Windows conda build failing with
```
35: =========================== short test summary info ===========================
35: FAILED gcore/vsiaz.py::test_vsiaz_write_appendblob_retry - AssertionError: (4, 6)
35: assert 4 == 6
35:  +  where 4 = <webserver.SequentialHandler object at 0x00000255FB962A40>.req_count
35:  +  and   6 = len([RequestResponse(PUT, /azure/blob/myaccount/test_copy/file.bin, 502, headers={}, body=None, custom_method=None, expect...ne, custom_method=None, expected_headers={}, expected_body=None, add_content_length_header=True, unexpected_headers=[]])
35:  +    where [RequestResponse(PUT, /azure/blob/myaccount/test_copy/file.bin, 502, headers={}, body=None, custom_method=None, expect...ne, custom_method=None, expected_headers={}, expected_body=None, add_content_length_header=True, unexpected_headers=[]] = <webserver.SequentialHandler object at 0x00000255FB962A40>.req_resp
35: = 1 failed, 3362 passed, 123 skipped, 5 xfailed, 29 warnings in 180.53s (0:03:00) =
```

Ubuntu 20.04 build failing with https://github.com/OSGeo/gdal/issues/7952

rouault:
> Windows conda build failing with test_vsiaz_write_appendblob_retry

yeah, that one is an occasional flake I believe

> Ubuntu 20.04 build failing with #7952

I've just merged https://github.com/OSGeo/gdal/pull/7975

Merging this PR

Issue #5711, opened 2022-05-12
CMake build error with external shapelib


dg0yt:
## Expected behavior and actual behavior.

When trying to build using an external shapelib, I run into:

~~~
CMake Error at frmts/aigrid/CMakeLists.txt:14 (target_link_libraries):
  Error evaluating generator expression:

    $<TARGET_OBJECTS:shapelib>

  Objects of target "shapelib" referenced but no such target exists.
~~~

`ogr/ogrsf_frmts/shape/CMakeLists.txt` contains a similar reference.

I expect the external lib to be used without errors.

I guess there is more work needed: The `FindShapelib.cmake` module makes assumptions about MSVC shapelib library names which don't match the vcpkg installation.

## Steps to reproduce the problem.

I assume these CMake options are relevant: `-DGDAL_USE_SHAPELIB=ON -DGDAL_USE_INTERNAL_LIBS=OFF`

## Operating system

Ubuntu 18.04, but CMake 3.22.

## GDAL version and provenance

3.5.0rc2

dg0yt:
In addition, `GDAL_USE_SHAPELIB_INTERNAL` should take precedence over `GDAL_USE_SHAPELIB`, but `CheckDependentLibraries.cmake` overrides `GDAL_USE_SHAPELIB_INTERNAL` with `OFF` if the external shapelib is recent enough.

rouault:
Building against external shapelib is not recommended. Cf https://github.com/OSGeo/gdal/blob/c4355a360384812db4df180d1ee1a73e864ada9d/cmake/helpers/CheckDependentLibraries.cmake#L420
```
# Disable by default the use of external shapelib, as currently the SAOffset member that holds file offsets in it is a
# 'unsigned long', hence 32 bit on 32 bit platforms, whereas we can handle DBFs file > 4 GB. Internal shapelib has not
# this issue
```

rouault:
> But this is something else then _fails to build_

a good way to enforce the recommendation :-)
I don't have much desire to fix building against external shapelib, knowing that it will break some use cases. autoconf/nmake builds don't allow building against external shapelib, so this is not a regression.

> I'm not so much concerned about limitations on 32 bit systems

That would also impact Windows 64 bit builds since unsigned long is 32 bit there too.

> From a packaging POV, vendored dependencies are not recommended.
[...]

PR #7973, opened 2023-06-19
Internal Shapelib: update list of symbols to rename to avoid conflicts with external shapelib (fixes microsoft/vcpkg#32070)

rouault:


michaeladamkatz:
Please check `SASetupDefaultHooks`. See https://github.com/OSGeo/gdal/issues/5711.

rouault:
> Please check `SASetupDefaultHooks`. See #5711.

It is in the updated list of symbols to be renamed to avoid conflicts: see https://github.com/OSGeo/gdal/pull/7973/files#diff-4bc13938e1804d9d783b571ccbb36f72d2154969b6bbdfd82f4f07278a0f999eR48

PR #7996, opened 2023-06-22
GTiff driver documentation: Minimum value for JPEG_QUALITY is 10

phanuel:
The documentation states that the minimum value for JPEG_QUALITY is 1. Actually, the minimum value is 10.
Same for JPEG_QUALITY_OVERVIEW.

### DIFF
--- a/doc/source/drivers/raster/gtiff.rst
+++ b/doc/source/drivers/raster/gtiff.rst
@@ -511,12 +511,12 @@ Creation Options
       already allocated in the file). The default is FALSE.
 
 -  .. co:: JPEG_QUALITY
-      :choices: 1-100
+      :choices: 10-100
       :default: 75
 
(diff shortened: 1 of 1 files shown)


rouault:
> Actually, the minimum value is 10.

What makes you believe so ?
With libjpeg-turbo, JPEG_QUALITY=1 works. Although values below 25 are associated with "JPEGLib:Caution: quantization tables are too coarse for baseline JPEG"

How did you build GDAL / which libjpeg do you use ?

phanuel:
I've been using the JPEG driver before (min value for QUALITY is 10) and I didn't expect the GTiff driver to behave any differently from the JPEG driver, sorry. This is quite confusing though.
I Just made a test with GTiff and JPEG_QUALITY < 10 to confirm and it does work.

PR #8010, opened 2023-06-24
gdal_calc.py: make --hideNoData imply --NoDataValue=none as documented (fixes #8009)

### LINKED ISSUE #8009: gdal_calc.py: `--hideNoData` does not remove nodata

According to the documentation for `--hideNodata` 

```
The output will not have a set NoDataValue, unless you explicitly specified a specific value by setting --NoDataValue=<value>.
```

This doesn't seem to be happening.

```
$ gdalinfo foo.tiff
...
Band 1 Block=2601x1 Type=Int16, ColorInterp=Gray
  NoData Value=-1
  Unit Type: W m-2
...

$ gdal_calc.py --quiet --hideNoData  -A foo.tiff  --A_band 1 --outfile foo2.tiff --calc "A"
$ gdalinfo foo2.tiff
...
Band 1 Block=2601x1 Type=Int16, ColorInterp=Gray
  NoData Value=-32768
```

According to the documentation, I would expect foo2.tiff to not have a NoData value, yet it does (and it's different from the previous one).

The following works to truly remove NoData:

```
$ gdal_calc.py --quiet --hideNoData --NoDataValue=none -A foo.tiff  --A_band 1 --outfile foo2.tiff --calc "A"
```

but as said I would expect the `--NoDataValue=none` to not be necessary.

## Operating system

Linux

```
# gdalinfo --version
GDAL 3.8.0dev-9d9d393e84d031a2dff760928ad9e28084f92e27, released 2023/05/23
```

This is using the latest gdal docker image.

rouault:
CC @idanmiara  You introduced --hideNoData in d9e9e827d85 and documented it as implying --NoDataValue=none, but the implementation was setting the default noDataValue depending on the data type. So I've fixed the implementation to follow the documentation, which seems the most natural expected behaviour.

### DIFF
--- a/autotest/pyscripts/test_gdal_calc.py
+++ b/autotest/pyscripts/test_gdal_calc.py
@@ -436,6 +436,7 @@ def test_gdal_calc_py_9(script_path):
         kwargs = copy(common_kwargs)
         kwargs.update(inputs0)
         ds = gdal_calc.Calc(calc="a", outfile=outfile, **kwargs)
+        assert ds.GetRasterBand(1).GetNoDataValue() is None
         if return_ds:
             input_file = ds
         else:
--- a/swig/python/gdal-utils/osgeo_utils/gdal_calc.py
+++ b/swig/python/gdal-utils/osgeo_utils/gdal_calc.py
@@ -443,7 +443,7 @@ def Calc(
         if ProjectionCheck:
             myOut.SetProjection(ProjectionCheck)
 
-        if NoDataValue is None:
+        if NoDataValue is None and not hideNoData:
             myOutNDV = DefaultNDVLookup[
                 myOutType
             ]  # use the default noDataValue for this datatype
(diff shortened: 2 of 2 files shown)


idanmiara:
@rouault thanks for looking into this.
Should we consider keeping the current behavior available using another switch/parameter value?

rouault:
> Should we consider keeping the current behavior available using another switch/parameter value?

hum I'm afraid that adding a new switch would bring confusion and would be likely a bit awkward to document: nodata is already a "negative" concept, "hideNoData" is the negation of a negation...  
I'd say either we align the implementation on the doc (this PR), or we fix the documentation to align on the implementation.

idanmiara:
> > Should we consider keeping the current behavior available using another switch/parameter value?
> 
> hum I'm afraid that adding a new switch would bring confusion and would be likely a bit awkward to document: nodata is already a "negative" concept, "hideNoData" is the negation of a negation... I'd say either we align the implementation on the doc (this PR), or we fix the documentation to align on the implementation.

I went over this again and I see that that we indeed have already all options covered:
* opts.NoDataValue='none' - to indicate not setting a nodatavalue on the output file, whereas opts.NoDataValue=None already indicates to use a default nodatavalue that depends on the output data type
* opts.hideNodata - Ignores the input bands NoDataValue. By setting this setting - no special treatment will be performed on the input NoDataValue. and they will be participating in the calculation as any other value.

So no problem and thanks for your fix 👍

PR #8011, opened 2023-06-25
Fix missing quotes in RFC45 SWIG bindings

vilhelmen:
## What does this PR do?
Third code section under SWIG bindings for RFC45 is missing a closing docstring quote, making it render wrong and harder to read.

https://gdal.org/development/rfc/rfc45_virtualmem.html#swig-bindings

None

## Tasklist
None?


rouault:
thanks

PR #8017, opened 2023-06-26
CI: add a coverage job

rouault:
For push operations on master, pushes to https://gdalautotest-coverage-results.github.io/coverage_html/index.html
(was tested on my fork : https://github.com/rouault/gdal/actions/runs/5379543312/jobs/9760905648 )

Pull requests and pushes should also appear at https://coveralls.io/github/OSGeo/gdal

### DIFF
--- a/.github/workflows/coverage/build.sh
+++ b/.github/workflows/coverage/build.sh
@@ -1,0 +1,19 @@
+#!/bin/sh
+
+set -eu
+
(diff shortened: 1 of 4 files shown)


nyalldawson:
Nice work! I've struggled trying to setup this kind of thing in the past but your approach here looks great 👍

PR #8022, opened 2023-06-28
autotest: Remove some inter-test dependencies

dbaston:
## What does this PR do?

Refactors some test files to remove dependencies between test functions.

I'm wondering if it would be a good idea to mark test files with no inter-test dependencies (`pytestmark = pytest.mark.independent` ?) and then use `pytest --random-order` for these tests in one of the CI configurations.

#4407

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/autotest/gcore/colortable.py
+++ b/autotest/gcore/colortable.py
@@ -31,33 +31,25 @@
 ###############################################################################
 
 
-import gdaltest
 import pytest
 
 from osgeo import gdal
 
 ###############################################################################
-# Create a color table.
+# Create a color table and verify its contents
 
 
 def test_colortable_1():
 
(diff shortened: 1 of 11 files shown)


rouault:
> I'm wondering if it would be a good idea to mark test files with no inter-test dependencies (`pytestmark = pytest.mark.independent` ?) and then use `pytest --random-order` for these tests in one of the CI configurations.

sounds like a good idea. https://pypi.org/project/pytest-random-order/ suggests the reverse logic "To disable reordering of tests in a module or class, use pytest marker notation: pytestmark = pytest.mark.random_order(disabled=True)"  I'm not sure which proportion of our existing test files are currently random order compatible.

dbaston:
Yes, the reverse logic is definitely the way to do. I didn't realize there was a standard marker for it - thanks for point it out.

About 75% of the test files repeatedly pass in random order.

PR #8048, opened 2023-07-04
swig java binding improvements

bdezonia:
I have enhanced the swig java bindings to support MDArrays more fully than they were. I also fixed a ton of typemap bugs and omissions in existing code. And I made the API more java developer friendly and the javadocs look very sensible.

All my tests pass (GDALTestMultiDim.java) and the system suite passes although I did not test things that require pytest (as it is not found on my system).

I am hoping to have someone review the code. I have questions marked by TODOs in:
- swig/include/java/gdal_java.i
- swig/include/java/typemaps_java.i
- swig/java/apps/GDALTestMultiDim.java

Note that the existing typemap code had lots of broken code and I do not pretend to have eliminated all the bad stuff. But I have left many things unchanged so no new bugs should crop up.


bdezonia:
Note that my last changes make things work better but they are still an api change (for the FieldDefn class in this case). Before it responded to GetFieldType() (and code to enable that via a rename directive caused problems elsewhere in the java api). Now it responds to GetType() instead. I changed the ogr apps that use it. I don't know if anyone else uses it. We could be safe and extend via gdal_java.i to make FieldDefn respond to both GetType() and GetFieldType() with the later just delegating to the former. Please advise as to how you think this should best be taken care of. Thanks

bdezonia:
So how do I debug/fix the code so that the tests pass on the Windows build? I don't know why it failed and I only have a Linux dev box that I work on.

In my development I found a gotcha bug where MDArray written data was not stored. My fix in the typemap code worked on my linux box. I was just modifying some broken code in an existing typemap and I did the most basic change. Not sure why it wouldn't work on Windows. But I am exercising pieces of code that maybe never have been stretched before?

rouault:
> So how do I debug/fix the code so that the tests pass on the Windows build? I don't know why it failed and I only have a Linux dev box that I work on.

It doesn't seem to be OS specific, but a "random" bug that I can also trigger on Linux by running the test program under Valgrind:

```
(cd swig/java; valgrind --leak-check=full /usr/bin/java "-Djava.library.path=/home/even/gdal/gdal/build_cmake/swig/java" "-cp" "gdal.jar:build/apps" "GDALTestMultiDim")
[...]
java.lang.RuntimeException: data read does not match data written
	at GDALTestMultiDim.testMDArrayStuff(GDALTestMultiDim.java:223)
	at GDALTestMultiDim.main(GDALTestMultiDim.java:53)
```

I feel this pull request is becoming too large and is going hard to be manageable. I would suggest ideally:
- to have a pull request with only the GetField / GetFieldType change, along the line I suggested to avoid breaking users
- have another one with the parseInt() / parseDouble() changes
- have another one, based on the one with the GetField/GetFieldType change (so typically once it has been merged into master), with the actual MultiDimensional changes, and by avoiding all the whitespace changes and comment removal you did in typemaps_java.i

bdezonia:
Even, I will make many of these changes over the next few days. I have already started breaking things into 3 commits (as you well know).

I had one comment about the namings in %apply's in the various .i files. You had mentioned you didn't like some of the choices I made. Imagine with dimensions you have the typemap (int nList, GInt64 *pList). Now you often need to %apply it to methods that have different names for the variables. In dimension cases you often have (int nDims, GInt64* dimensions). However, if that is how the %apply is defined the java api's javadoc will show something like "long[] nDims" (since it seems to grab the first name of the pair of variables). Really we want it to be "long[] dimensions". I have been using %apply's to instead say something like (int dimensions, GInt64* sizes). The javadoc then indeed refers to "long[] dimensions". By doing this I have made the javadoc for a lot of classes a lot more clear. But I didn't think how this might have effected other language bindings. Can we come up with a solution that works well everywhere and makes the javadoc really nice? I did play with changing how typesmaps are defined by specifying them as (GInt64* dimensions, int count), changing the source code in key places to pass things in those orders, and then the javadoc looks like "long[] dimensions". But that could be a big change to make and could be bug prone translating all the typemaps and interchanging $1 with $2 in them and updating source code everywhere to make sure they get called, etc.. Do you have any thoughts here?

rouault:
> By doing this I have made the javadoc for a lot of classes a lot more clear. But I didn't think how this might have affected other language bindings

oh I understand now. That makes sense. If my memories are OK, that should also do the right thing for Python as well.

> But that could be a big change to make and could be bug prone translating all the typemaps and interchanging $1 with $2 in them and updating source code everywhere to make sure they get called, etc..

Indeed. That's a cleanup we could consider at some point, and that would indeed involve modifying typemaps of other languages. Should be mostly mechanical, but definitely material for a dedicated pull request.

jratike80:
@bdezonia, it seems that you have done a lot of good work. Much appreciated!

bdezonia:
@jratike80 thanks! I am down to debugging / fixing bugs / dealing with one tricky compile error. I hope to be done soon (in a week? in a couple weeks?)

My participation is somewhat mercenary. I have a java library (zorbage-gdal) that uses gdal to load data into formats my numeric processing library (zorbage) and viewer (zorbage-viewer) can work with. My libraries are multidimensional in nature and my current attempts at handling multidimensional raster data in zorbage-gdal are a bit clunky. Once these java binding updates are in place I hope to improve my importers substantially.

bdezonia:
So I am down to 2 bugs to iron out and I'm not sure how to proceed. Advice appreciated.

First the MDArray::Reaf()/Write() routines (probably Write()) don't seem to work on the Windows test server. I am not sure why. Even told me he can reproduce the problem in valgrind on a (linux?) dev box but when I run it on my linux laptop I can't figure out how to tell if the test code failed or succeeded or even ran. Any pointers? (I've never used valgrind before.)

The other issue is a bit pernicious. I am going to link to code below. I really cannot figure out how to write Dimension[] MDArray::GetDimensions() without leaking memory. I tried doing an int GetDimensionCount() with a GDALDimensionH GetDimension(int i) approach in C code and a pure Java approach for Dimension[] GetDimensions() that just uses those two routines. However GetDimension() in C code gets all the dimensions and returns a pointer GDALDimensionH to the one dimension of interest. However that points into the memory block of the GDALGetDimensions call. If I free the block on method exit (and access the one GDALDimensionH later) the program crashes. If I don't things run well but I know I am leaking memory. I have tried defining new typemaps etc. (not in the branch yet) and I can't work around the problem. One complication is that I can't copy GDALDimensionH contents around. GDALDimensionsH's are related to Groups. MDArray has no idea how to duplicate a GDALDimensionH and all it's contents. Maybe there is a way I don't know about. I am definitely looking for ideas on how to handle this.

Here are some relevant code links:

https://github.com/bdezonia/gdal/blob/849d3b9eae0e26990b3a3ab8dea181d604697a03/swig/include/java/gdal_java.i#L156

https://github.com/bdezonia/gdal/blob/849d3b9eae0e26990b3a3ab8dea181d604697a03/swig/include/java/gdal_java.i#L267

https://github.com/bdezonia/gdal/blob/849d3b9eae0e26990b3a3ab8dea181d604697a03/swig/include/java/gdal_java.i#L446

https://github.com/bdezonia/gdal/blob/849d3b9eae0e26990b3a3ab8dea181d604697a03/swig/include/java/typemaps_java.i#L2220

rouault:
> First the MDArray::Reaf()/Write() routines (probably Write()) don't seem to work on the Windows test server. I am not sure why. Even told me he can reproduce the problem in valgrind on a (linux?) dev box but when I run it on my linux laptop I can't figure out how to tell if the test code failed or succeeded or even ran. Any pointers? (I've never used valgrind before.)

Not sure to understand your issue is exactly. What if you immitate my above Valgrind invokation of https://github.com/OSGeo/gdal/pull/8048#issuecomment-1620048769 ?

> The other issue is a bit pernicious. I am going to link to code below. I really cannot figure out how to write Dimension[] MDArray::GetDimensions() without leaking memory.

cf perhaps my suggestion in https://github.com/OSGeo/gdal/pull/8048#discussion_r1251708537 where I nullify dims[index] = NULL before GDALReleaseDimensions()
Or otherwise bite the bullet and enable GetDimensions() for Java as well as Python, as write a typemap for "(GDALDimensionHS*** pdims, size_t* pnCount)" similar to the python one ( probably that the Java one for Geometry.GetPoints() would be a starting point)

bdezonia:
Re: valgrind I used your exact invocation code before. I'm not sure how to tell if the bug happens in the test or not.

Re: GetDimensions() I will look at your python example. I had tried the typemap with indirection idea and couldn't convince the javacode to use it.

rouault:
> I'm not sure how to tell if the bug happens in the test or not.

well, look at the output. When I tried, I got the same exception "data read does not match data written" as on the CI Windows build, which is a sign there's a bug somewhere.

bdezonia:
BTW I did not see your "dims[index] = null;" trick. That might help things a lot.

bdezonia:
OK. So I feel I have pretty much made the final commits for this branch.

The multidim test under valgrind still fails (though works fine when launched without valgrind from the command line). But at this point I seem to be spending more time trying to debug MEMAbstractMDArray::ReadWrite() than anything else. I don't think I am the person to do this and that it would be better done by someone who better understands it. I will note that both 3d and 4d examples fail. The read call sets some of the output plane's values to zero for unknown reasons.

There is also a message in valgrind that it finds a small number of bytes that are leaking. I still need to figure this out.

Can someone chime in on how to proceed with debugging the ReadWrite() code?

bdezonia:
Correction: under valgrind the amount of leaked memory is the same if I comment out my test code or not. There is no evidence the code as it is has memory leaks.

bdezonia:
Note that I made the MultiDimTest code now run tests on 1d, 2d, 3d, and 4d data. Only 1d survives the valgrind test. The tests are almost identical mainly differing in the input dimensions things that we construct from and the degree of nested loops needed for iterating read/write plane positions. Note that the 2d case in MEMAbstractMDArray::ReadWrite() is the first one to use the stack oriented code.

bdezonia:
Note that when lambdaLastDim(0) is called this subcase of data assignment is happening inside it with nCopySizeLastDim == 4:

https://github.com/OSGeo/gdal/blob/a6d03b2e86e760b32306e9fdb4961a17d33ab66d/frmts/mem/memdataset.cpp#L2063

bdezonia:
Random thought: About the two lines starting here:

https://github.com/OSGeo/gdal/blob/a6d03b2e86e760b32306e9fdb4961a17d33ab66d/frmts/mem/memdataset.cpp#L2244

Should source offset and source increment be multiplied by the size of the (internal?) type? In my testing I am passing the X and Y strides as 1. Is that a mistake when copying planes? Should the Y stride not be 1?

One thing I can test later is to make the internal data type into bytes (size of 1) instead of shorts (size of 2). If the all the tests pass through valgrind just fine after doing that then I think we will know the culprit. I will try to do so tonight.

rouault:
> Should source offset and source increment be multiplied by the size of the (internal?) type?

m_anStride[i] takes into account m_oType.GetSize() (cf MEMAbstractMDArray::Init()). Note that this is an internal detail of the MEM driver. The strides[] array passed to GDALMDArray::Read()Write() should be in number of elements, not in number of bytes

> In my testing I am passing the X and Y strides as 1. Is that a mistake when copying planes? Should the Y stride not be 1?

For a row-major contiguous buffer, the pattern should expressed that the last dimension is the fastest varying one:
stride[ndims-1] = 1
stride[ndims-2] = dims[ndims-1]->GetSize()
stride[ndims-3] = dims[ndims-1]->GetSize() * dims[ndims-2]->GetSize()
...
stride[0] = product of dims[i]->GetSize() for i = 1 to ndims-1

bdezonia:
I'm not entirely following your post. So are you saying I am passing in bad strides? Or is your code info on how strides are calculated internally?

My 4d example input to read/write routines is shown starting on these lines:

https://github.com/bdezonia/gdal/blob/0b4197910ba8233b0a1decc1a86e9dce3e278696/swig/java/apps/GDALTestMultiDim.java#L622

You can see what I am setting for steps, strides, etc.

Please let me know if I need to be passing in better strides (and/or steps) info. Thanks.

rouault:
> My 4d example input to read/write routines is shown starting on these lines:

Your strides aren't correct. They are stepping onto each other. They should be set according to the following logic (actually slightly revised from my previous suggestion, to use count[i] instead of dims[i]->GetSize() where count[i] is the number of elements requested on dimension i):
```
stride[ndims-1] = 1
stride[ndims-2] = count[ndims-1]
stride[ndims-3] = count[ndims-1]* count[ndims-2]
...
stride[0] = product of count[i] for i = 1 to ndims-1
```
Cf https://github.com/OSGeo/gdal/blob/a6d03b2e86e760b32306e9fdb4961a17d33ab66d/swig/include/python/gdal_python.i#L1267 for how this is done in Python bindings.
I would expect the Java bindings to also have a Read() variant where the user doesn't need to explicitly pass the strides values, and they would be computed for him. And possibly for other parameters:
- if start_idx not specified, assume all 0
- if count not specified, assume dims[i]->GetSize()
- if step not specified, assume all 1

rouault:
> How does that sound?

OK to me

bdezonia:
OK. So everything is now passing. I consider the branch feature complete. Please review and comment. Thanks.

bdezonia:
OK, I have no idea why the MacOS stuff keeps failing. The error messages seem to imply that maybe you are having infrastructure problems. You can see my "clean up some tabs" commit passed everything.

Anyhow I consider this code complete. I am guessing folks might be busy with the new release etc. I'm just commenting in case you saw the broken build and didn't see my "please review" comment.

rouault:
Thanks. I'll give this a review in coming days. the macos_build failure is indeed related

rouault:
@bdezonia I've pushed a few tweaks/fixes, and remove the useless typemaps as I suspected using FooHS* instead of FooH made the trick. Thanks for this work!

bdezonia:
Yeehah! Thanks! That was entirely too much work. Hahaha

Issue #8018, opened 2023-06-27
Python bindings: gdal.config_option may not restore original state


dbaston:
## Expected behavior and actual behavior.

The `gdal.config_option` context manager should restore the original state upon exiting its context. Instead, it can cause migration of a config option from global to thread-local storage and vice-versa. This leads to confusing results such as:

```
gdal.SetConfigOption("A", "1")
with gdal.config_option("A", "2"):
    pass
gdal.SetConfigOption("A", "3")
gdal.GetConfigOption("A")  # 1
```

I have a partial fix to the problem in https://github.com/dbaston/gdal/commit/e8127d008c2d19e436769bdc738a3c4bbc7c5701. But this still fails in the situation where we have a config option defined both in global and thread-local storage, because there is no `CPLGetConfigOption` variant that provides access to the global value in this case. It looks like we can get access via `CPLGetConfigOptions`, but this would require exposing that function to SWIG and I think we should avoid exposing that function because the behavioral difference with `CPLGetConfigOption` is confusing. (I would propose exposing a `gdal.GetConfigOptions()` that returns all config options applicable to the current thread.)

So, we could go with a fix like the one above, which should be sufficient for non-contrived cases in the test suite, or (more cleanly) I could add a `CPLGetGlobalConfigOption` that complements `CPLGetThreadLocalConfigOption` and allows the `gdal.config_option` context manager to fully restore the original state. Or maybe there's another solution that I'm not seeing.

rouault:
> I could add a `CPLGetGlobalConfigOption` that complements `CPLGetThreadLocalConfigOption` and allows the `gdal.config_option` context manager to fully restore the original state

sounds good to me

PR #8050, opened 2023-07-04
Python bindings: Prevent migration of config options between env vars, global and thread-local storage

dbaston:
## What does this PR do?

Fix for https://github.com/OSGeo/gdal/issues/8018

Maybe there is a better name than `CPLGetGlobalConfigOption`. I couldn't think of a word that exactly means the opposite of "thread-local".

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/autotest/cpp/test_cpl.cpp
+++ b/autotest/cpp/test_cpl.cpp
@@ -4269,8 +4269,8 @@ TEST_F(test_cpl, config_file_ignore_env_vars)
 {
     char szEnvVar[] = "SOME_ENV_VAR_FOR_TEST_CPL_61=FOO";
     putenv(szEnvVar);
-    ASSERT_TRUE(CPLGetConfigOption("SOME_ENV_VAR_FOR_TEST_CPL_61", nullptr) !=
-                nullptr);
+    ASSERT_STREQ(CPLGetConfigOption("SOME_ENV_VAR_FOR_TEST_CPL_61", nullptr),
+                 "FOO");
 
     VSILFILE *fp = VSIFOpenL("/vsimem/.gdal/gdalrc", "wb");
     VSIFPrintfL(fp, "[directives]\n");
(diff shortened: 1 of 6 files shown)


autotest/cpp/test_cpl.cpp
```diff
@@ -4301,17 +4301,41 @@ TEST_F(test_cpl, config_file_ignore_env_vars)
     // Reload configuration file
     CPLLoadConfigOptionsFromFile("/vsimem/.gdal/gdalrc", false);
 
-    // Check that environment variables are read as configuration options
-    // and override configuration options
+    // Check that environment variables are read as configuration options and override
+    // configuration options defined in a file. This behavior differs from configuration
+    // options defined outside of a file, in which case the configuration option overrides
+    // the environment variable.
```
rouault, line 4307: I've spent some minutes trying to understand both my own original comment and your update proposal, being confused before realising that this was due to the false argument at line 4302.

I guess it would be perhaps clearer to remove that comment, but amend the one at line 4301 by underlying that we pass bOverrideEnvVars = false, and thus configuration options defined in the config file will not override pre-existing environment variables.

port/cpl_conv.cpp
```diff
@@ -1808,6 +1795,29 @@ const char *CPL_STDCALL CPLGetThreadLocalConfigOption(const char *pszKey,
     return pszResult;
 }
 
+/************************************************************************/
+/*                   CPLGetGlobalConfigOption()                    */
+/************************************************************************/
+/** Same as CPLGetConfigOption() but excludes options set with
```
rouault, line 1801: some spaces missing for nice formatting

port/cpl_conv.cpp
```diff
@@ -1808,6 +1795,29 @@ const char *CPL_STDCALL CPLGetThreadLocalConfigOption(const char *pszKey,
...
+ *  CPLSetThreadLocalConfigOption() */
```
rouault, line 1802:
```suggestion
 *  CPLSetThreadLocalConfigOption().
 * This function should generally not be used by applications, which should rather
 * use CPLGetConfigOption()
 * @since 3.8
```

autotest/cpp/test_cpl.cpp
dbaston: I also missed that, thanks for pointing it out. I pulled `bOverrideEnvVars` out into a local variable, which should make it more clear what's happening (at least, it does for me)

port/cpl_conv.cpp
dbaston: fixed

port/cpl_conv.cpp
dbaston: Added. Also noted that it excludes environment variables.

PR #8055, opened 2023-07-04
Eliminate some deprecations / compiler warnings

bdezonia:
Some java deprecation warnings were getting in the way of easily seeing real compile errors during the build process. Bring that code up to date to eliminate the warnings.


rouault:
thanks

PR #8059, opened 2023-07-04
Add a compatablity method to swig bindings that is Java Only

bdezonia:
It will simplify some future development in regards to FieldDefn and upcoming changes to java typemaps file.

Note that the java typemaps file has not yet removed the %rename GetType GetFieldType code from swig/include/java/ogr_java.i

### DIFF
--- a/swig/include/ogr.i
+++ b/swig/include/ogr.i
@@ -2496,6 +2496,13 @@ public:
         OGR_Fld_SetType(self, type);
   }
 
+#ifdef SWIGJAVA
+  // Alias for backward compatibity
+  OGRFieldType GetFieldType() {
+    return OGR_Fld_GetType(self);
+  }
+#endif
+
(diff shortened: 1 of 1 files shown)


rouault:
> Note that the java typemaps file has not yet removed the %rename GetType GetFieldType code from swig/include/java/ogr_java.i

would make sense to include that removal in the PR, unless there's a strong reason not to do it

bdezonia:
Working on your request now.

Do you want me to mark the SWIGJAVA one as deprecated?

rouault:
> Do you want me to mark the SWIGJAVA one as deprecated?

GetFieldType() you mean ? I don't have a strong opinion about that. Your call. I'm not sure how you'd do that, unless you do it in swig/java/javadoc.java

bdezonia:
Good point. I have no idea how to do it either. For now I will skip it. I think the code is ready to merge.

PR #8066, opened 2023-07-06
SWIG: Add GetConfigOptions

dbaston:
## What does this PR do?

Exposes a `GetConfigOptions` function that returns the set of options obtainable with `GetConfigOption`, with the exception of environment variables. This is done by blending the results of `CPLGetConfigOptions` and `CPLGetThreadLocalConfigOptions`.

I had conflicting opinions about this. On the one hand, the `GetConfigOptions` name is similar to `GetConfigOption`, so the functions should behave similarly and return thread-local config options that shadow global config options. On the other hand, the SWIG bindings should behave similarly to the underling C API, but this would introduce an unintuitive behavior difference between `GetConfigOption` and `GetConfigOptions`. This conflict could also be resolved by changing the behavior of `CPLGetConfigOptions`. I'm not seeing real usage in the wild: https://github.com/search?q=CPLGetConfigOptions+NOT+path%3Acpl_conv.*+NOT+path%3Agdal_*.rs+NOT+path%3Atest_cpl.cpp&type=code 

https://trac.osgeo.org/gdal/ticket/7163

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/autotest/gcore/misc.py
+++ b/autotest/gcore/misc.py
@@ -942,6 +942,31 @@ def test_misc_config_context_mgrs_3():
 
 
 ###############################################################################
+# Test GetConfigOptions
+
+
+def test_misc_get_config_options():
+
+    assert gdal.GetConfigOptions() == {}
+
+    try:
+        gdal.SetConfigOption("A", "1")
+
+        assert gdal.GetConfigOptions() == {"A": "1"}
+
+        gdal.SetConfigOption("B", "2")
+
+        with gdal.config_options({"A": "9", "C": "8"}):
+            assert gdal.GetConfigOptions() == {"A": "9", "B": "2", "C": "8"}
+
+        assert gdal.GetConfigOptions() == {"A": "1", "B": "2"}
+
+    finally:
+        gdal.SetConfigOption("A", None)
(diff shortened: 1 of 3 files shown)


autotest/gcore/misc.py
```diff
@@ -941,6 +941,31 @@ def test_misc_config_context_mgrs_3():
         gdal.SetThreadLocalConfigOption("C", None)
 
 
+###############################################################################
+# Test GetConfigOptions
+
+
+def test_misc_get_config_options():
+
+    assert gdal.GetConfigOptions() == {}
```
dbaston, line 950: The rasterlite commit is needed to make this assert pass

rouault:
The suggestion function of github doesn't allow me to suggest this:
```
            CPLConfigOptionSetter oSetter("OGR_SQLITE_EXACT_EXTENT", "YES", false);
            OGR_L_GetExtent(hMetadataLyr, &oEnvelope, TRUE);
```
Note that you can't re-use the const char* returned by CPLGetThreadLocalConfigOption() as it might have been invalidated by the CPLSetThreadLocalConfigOption(). The CPLConfigOptionSetter class takes care of all this.

PR #8073, opened 2023-07-07
Python bindings: Allow passing options as dict

dbaston:
## What does this PR do?

Allows functions/methods in the Python bindings to accept options via a dictionaries.

A typemap (`char **dict`) already existed to accept a dictionary or sequence and convert it into a CSL, but it was only applied to `SetMetadata`. I updated the `char **dict` typemap to use the `CSLFromPySequence` helper, and set the `char **options` typemap to alias this.

I noticed that `CSLFromPySequence` has some special handling for UTF8 strings that was not present in the sequence-to-CSL code I replaced in `char **dict`. The dict-to-CSL pathway in `char **dict` does not have any special handling, but it appears that UTF8 option names and values are successfully transmitted to GDAL.

## Tasklist

 - [x] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/autotest/gcore/basic_test.py
+++ b/autotest/gcore/basic_test.py
@@ -534,6 +534,47 @@ def test_basic_test_16():
     assert "Invalid value for NUM_THREADS: INVALID" in gdal.GetLastErrorMsg()
 
 
+def test_basic_dict_open_options():
+
+    ds1 = gdal.Open("data/byte.tif")
+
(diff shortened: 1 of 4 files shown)


dbaston:
> The dict-to-CSL pathway in char **dict does not have any special handling, but it appears that UTF8 option names and values are successfully transmitted to GDAL.

Disregard this, the UTF8 handling is just happening inside of helper functions.

PR #8076, opened 2023-07-08
CI: linux_build.yml: build and run tests as non-root

rouault:


dbaston:
I had thought about adding similar instructions to the documentation about running these container locally. I'm curious -- was running as root causing a problem in the CI somewhere?

rouault:
> I'm curious -- was running as root causing a problem in the CI somewhere?

Running as root was a minor inconvenience as some tests try to test accessing files in read-only (e.g. test_ogr_openfilegdb_read_readonly_in_update_mode). When running as root, read-only is ignored (unless you use "chattr +i the.file", which is non portable), and thus those tests are skipped

Cf
```
root@even-ThinkPad-P15v-Gen-1:~# touch /tmp/readonly
root@even-ThinkPad-P15v-Gen-1:~# chmod 0555 /tmp/readonly 
root@even-ThinkPad-P15v-Gen-1:~# !ll
ll /tmp/readonly 
-r-xr-xr-x 1 root root 0 juil. 18 16:25 /tmp/readonly*

root@even-ThinkPad-P15v-Gen-1:~# python
Python 3.8.10 (default, May 26 2023, 14:05:08) 
[GCC 9.4.0] on linux
>>> open('/tmp/readonly', 'rb+')
<_io.BufferedRandom name='/tmp/readonly'>
>>> 
# was expecting an exception...


root@even-ThinkPad-P15v-Gen-1:~# chattr +i /tmp/readonly
root@even-ThinkPad-P15v-Gen-1:~# python
Python 3.8.10 (default, May 26 2023, 14:05:08) 
[GCC 9.4.0] on linux
>>> open('/tmp/readonly', 'rb+')
PermissionError: [Errno 1] Operation not permitted: '/tmp/readonly'
```

dbaston:
Thanks!

PR #8096, opened 2023-07-13
validate_geoparquet: avoid GitHub login via .netrc

dbaston:
## What does this PR do?

Avoids logging into GitHub via credentials in a `.netrc` file when downloading GeoParquet schemas in `validate_geoparquet.py`. This causes a test failure if the login fails or the associated token does not have certain permissions, even though the schemas being accessed are public.

None

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/swig/python/gdal-utils/osgeo_utils/samples/validate_geoparquet.py
+++ b/swig/python/gdal-utils/osgeo_utils/samples/validate_geoparquet.py
@@ -159,7 +159,7 @@ class GeoParquetValidator(object):
                     geoparquet_schemas[schema_url] = json.loads(response)
                 except Exception as e:
                     return self._error(
-                        f"Failed to read GeoParquet schema at {schema_url} as JSON. Schema content = '{response}. Exception = {repr(e)}"
+                        f"Failed to read GeoParquet schema at {schema_url} as JSON. Schema content = '{response}'. Exception = {repr(e)}"
                     )
 
             schema_j = geoparquet_schemas[schema_url]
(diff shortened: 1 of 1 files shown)


rouault:
Wondering if it might not be better to simplify all this download logic that currently goes through GDAL to use urllib like in https://github.com/OSGeo/gdal/pull/8078/files#diff-715fdaaab37506a1eef1584ca6ddf4f5f84c771c3efdf6ffeac54b308f3d52e9R128 ?

dbaston:
Yes, that's simpler. Updated.

swig/python/gdal-utils/osgeo_utils/samples/validate_geoparquet.py
```diff
@@ -145,35 +145,24 @@ def _check(self):
...
                     )
 
                 try:
-                    schema_j = json.loads(schema)
+                    geoparquet_schemas[schema_url] = json.loads(response)
                 except Exception as e:
                     return self._error(
-                        f"GeoParquet schema at {schema_url} is not valid JSON. Schema content = '{schema}'. Exception = '%s' "
-                        % (str(e))
+                        f"Failed to read GeoParquet schema at {schema_url} as JSON. Schema content = '{response}. Exception = {repr(e)}"
```
rouault, line 162:
```suggestion
                        f"Failed to read GeoParquet schema at {schema_url} as JSON. Schema content = '{response}'. Exception = {repr(e)}"
```

PR #8097, opened 2023-07-13
autotest: make utilities, pyscript tests run independently

dbaston:
## What does this PR do?

This PR makes tests under the `utilities` and `pyscripts` folders run independently, using the following types of changes:

- Pulling out generated data that is used by multiple tests into module-scoped fixtures.

- Moving cleanup steps from a dummy test into a fixture, or in some cases writing generated files to a location provided by `tmp_path` to avoid the need for a cleanup fixture. The `tmp_path` method is more invasive but probably easier for understand, so it might be worth doing this in all cases.

- Merging tests, occasionally, such as a test that writes a file followed by a test that reads the file.

-  Flagging tests as non-independent using the `random_order(disabled=True)` marker (44b24c8c656d23112bb9ff0ea3852be3ae7fdcb5).
   These will need to be addressed eventually if we want to run the tests in parallel.

While I was in there, I also split and parameterized some tests (2062341495720874bda7624fd3904ba9d237c359)

#4407

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/autotest/utilities/test_gdalwarp_lib.py
+++ b/autotest/utilities/test_gdalwarp_lib.py
@@ -40,6 +40,23 @@ import pytest
 
 from osgeo import gdal, ogr, osr
 
+
+@pytest.fixture(scope="module", autouse=True)
+def setup_and_cleanup():
+
+    yield
+
+    # We don't clean up when run in debug mode.
+    if gdal.GetConfigOption("CPL_DEBUG", "OFF") == "ON":
+        return
+
+    for i in range(2):
+        try:
+            os.remove("tmp/testgdalwarp" + str(i + 1) + ".tif")
+        except OSError:
+            pass
+
+
 ###############################################################################
 # Simple test
 
@@ -104,7 +121,12 @@ def test_gdalwarp_lib_4():
 # Test warping from GCPs without any explicit option
 
 
-def test_gdalwarp_lib_5():
+@pytest.fixture(scope="module")
+def testgdalwarp_gcp_tif(tmp_path_factory):
+
+    testgdalwarp_gcp_tif_fname = str(
+        tmp_path_factory.mktemp("tmp") / "testgdalwarp_gcp.tif"
+    )
 
     ds = gdal.Open("../gcore/data/byte.tif")
     gcpList = [
@@ -114,8 +136,19 @@ def test_gdalwarp_lib_5():
         gdal.GCP(440720.000, 3750120.000, 0, 0, 20),
     ]
     ds1 = gdal.Translate(
-        "tmp/testgdalwarp_gcp.tif", ds, outputSRS="EPSG:26711", GCPs=gcpList
+        testgdalwarp_gcp_tif_fname, ds, outputSRS="EPSG:26711", GCPs=gcpList
     )
+    del ds1
+
+    yield testgdalwarp_gcp_tif_fname
+
+
+def test_gdalwarp_lib_5(testgdalwarp_gcp_tif):
+
+    ds = gdal.Open("../gcore/data/byte.tif")
+
+    ds1 = gdal.Open(testgdalwarp_gcp_tif)
+
     dstDS = gdal.Warp("", ds1, format="MEM", tps=True)
 
     assert dstDS.GetRasterBand(1).Checksum() == 4672, "Bad checksum"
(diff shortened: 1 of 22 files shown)


hobu:
💪

dbaston:
I'm working though a couple Windows issues here:

- the `gdaltest.runexternal*` commands use `shlex.split`, which erases `'\` characters in Windows paths. The `\` characters arise when converting `pathlib` paths to strings on Windows. Removing `shlex.split` appears to break Windows command strings that contain newlines.
- some tests require a relative path between datasets but this is not possible on Windows when code is checked out to `D:\` and the temporary directory created by `tmp_path` is under `C:\`

rcoup:
> writing generated files to a location provided by tmp_path to avoid the need for a cleanup fixture. The tmp_path method is more invasive but probably easier for understand, so it might be worth doing this in all cases.

If we’re using `tmp_path` more, would it be helpful to document where pytest preserves files/how to find them.

https://github.com/pytest-dev/pytest/issues/8141 ha changed retention behaviour to keep only failed test data by default too (configurable via `tmp_path_retention_policy`)

rouault:
for test_gdal_edit_py_3 failure, changing ``-a_srs ''`` to ``-a_srs None`` can do the job

dbaston:
> for test_gdal_edit_py_3 failure, changing -a_srs '' to -a_srs None can do the job

I've made this change, though I worry similar issues may pop up in other places and cause confusing platform-specific test failures. I tried to address the root cause by patching `pathlib` to use slashes instead of backslashes:

```
@pytest.fixture(autouse=True)
def use_posix_paths(monkeypatch):

    from pathlib import PureWindowsPath

    oldstr = PureWindowsPath.__str__

    monkeypatch.setattr(PureWindowsPath, "__str__", lambda x : oldstr(x).replace('\\', '/'))
```

But this causes other failures, such as `osgeo_utils` tests that require the behavior of `pathlib` to be the same as `os.path`.

I guess I'm inclined to leave it as-is now, but maybe a better solution will come.

autotest/utilities/test_gdalwarp_lib.py
```diff
@@ -40,6 +40,23 @@
...
+    if gdal.GetConfigOption("CPL_DEBUG", "OFF") == "ON":
+        return
+
+    for i in range(2):
+        try:
+            os.remove("tmp/testgdalwarp" + str(i + 1) + ".tif")
+        except OSError:
+            pass
+
+
```
rouault, line 59: I assume this isn't needed as the temporary files aren't written in tmp/ anymore

dbaston: I haven't switched to `tmp_path` everywhere, though I could follow up with that.

rouault:
Regarding the failures on the build-windows-conda target, cf https://github.com/OSGeo/gdal/pull/8123 (once 8123 passes, please rebase on top of it or master if it is merged)

autotest/utilities/test_gdalwarp_lib.py
rouault: fine then. can indeed be done as a follow up after this PR

rcoup:
> I tried to address the root cause by patching pathlib to use slashes instead of backslashes:

Patching stdlib to behave _opposite_ to the platform's behaviour seems like a dangerous approach to solve it...

> the `gdaltest.runexternal*` commands use shlex.split, which erases `\` characters in Windows paths. The `\` characters arise when converting pathlib paths to strings on Windows. 

IIUC on Windows `subprocess.*` just re-joins a list argument anyway, so you might as well just pass the command string and avoid `shlex.split()`, which will never work right on Windows.

> Removing `shlex.split` appears to break Windows command strings that contain newlines.

Where do we have newlines in command strings? Should/are they CRLF (`\r\n`) on windows? Seems like this is probably the  problem to focus on?

dbaston:
> Patching stdlib to behave opposite to the platform's behaviour seems like a dangerous approach to solve it...

Agree - it was an option I shared from my search for the least bad solution. In this case it created more problems than it solved.

> IIUC on Windows subprocess.* just re-joins a list argument anyway, so you might as well just pass the command string and avoid shlex.split(), which will never work right on Windows.

That's what I've done here, though it caused a handful of Windows-only failures that were difficult for me to understand and reproduce.

> Where do we have newlines in command strings?

The only one that executes on CI was addressed in https://github.com/OSGeo/gdal/pull/8097/commits/a32e4ff866434b847b3c4a83673a2712695f32f7

rcoup:
ah, I had a search in master and found that code, but I hadn't spotted Even had merged this already - hence the confusion 👍

PR #8140, opened 2023-07-21
Make gdal.VectorTranslate() to return consistent output dataset for JSON, JSONFG and GPKG

### LINKED ISSUE #8033: Writing invalid GeoJSON

I want to save my result in GeoJSON format. Currently, the output file is invalid. At the end of the file, two brackets "]}" are missing.

Execute in python:
```python
gdal.VectorTranslate("./output.geojson", "./input.geojson")
```
Valid input file:
```json
{
    "type": "FeatureCollection",
    "name": "test",
    "crs": {
        "type": "name",
        "properties": {
            "name": "urn:ogc:def:crs:EPSG::25833"
        }
    },
    "features": [
        {
            "type": "Feature",
            "properties": {},
            "geometry": {
                "type": "Point",
                "coordinates": [
                    391746.3,
                    5819908.1
                ]
            }
        }
    ]
}
```
Invalid output file:
```json
{
    "type": "FeatureCollection",
    "name": "test",
    "crs": {
        "type": "name",
        "properties": {
            "name": "urn:ogc:def:crs:EPSG::25833"
        }
    },
    "features": [
        {
            "type": "Feature",
            "properties": {},
            "geometry": {
                "type": "Point",
                "coordinates": [
                    391746.299999999988358,
                    5819908.099999999627471
                ]
            }
        }
```

## Operating system

Windows 10.0.19045, 64bit
Anaconda environment, Python 3.10.10

gdal 3.6.4 py310h3ccc0e8_1 conda-forge

rouault:
Fixes #8033


dbaston:
I tried building this branch but I am seeing the same behavior as in `master`:

I added a print statement to the `__del__` method of `GDALDatasetShadow`. I am still trying to understand why it is called called twice here. In any case, executing any command (even an expression like `12345`) after calling `gdal.VectorTranslate` is what causes the flush to disk.

```
Python 3.10.6 (main, Mar 10 2023, 10:55:28) [GCC 11.3.0] on linux
>>> import os
>>> from osgeo import gdal
>>> gdal.VectorTranslate('/tmp/out.geojson', '/home/dan/dev/gdal-scratch/input.geojson')
/home/dan/dev/build-gdal-Desktop-Debug/swig/python/osgeo/gdal.py:290: FutureWarning: Neither gdal.UseExceptions() nor gdal.DontUseExceptions() has been explicitly called. In GDAL 4.0, exceptions will be enabled by default.
  warnings.warn(
called __del__()
<osgeo.gdal.Dataset; proxy of <Swig Object of type 'GDALDatasetShadow *' at 0x7f436dcc4330> >
>>> os.stat('/tmp/out.geojson')
called __del__()
os.stat_result(st_mode=33204, st_ino=11797189, st_dev=64769, st_nlink=1, st_uid=1000, st_gid=1000, st_size=0, st_atime=1689992262, st_mtime=1689992262, st_ctime=1689992262)
>>> os.stat('/tmp/out.geojson')
os.stat_result(st_mode=33204, st_ino=11797189, st_dev=64769, st_nlink=1, st_uid=1000, st_gid=1000, st_size=291, st_atime=1689992262, st_mtime=1689992276, st_ctime=1689992276)
>>> 
```

rouault:
> I tried building this branch but I am seeing the same behavior as in `master`:

ok, I've added missing Flush() calls so that the writes are actually visible to external process, as the tests involved the same process, it actually saw a consistent state

dbaston:
This works for me -- the entire file is now written when `gdal.VectorTranslate` returns.

As far as what I was seeing previously -- I learned that Python stores the result of the last statement to the variable `_`. So `GDALDatasetShadow.__del__` is called once when `gdal.VectorTranslate` returns, but the `GDALDataset` is not closed and its ownership is transferred to `_`. Executing another statement invokes `_.__del__()`, actually closing the `GDALDataset`.

PR #8146, opened 2023-07-24
WFS: Don't issue STARTINDEX if feature count is small

craigds:
## What does this PR do?

WFS datasources with no primary key cannot be naturally ordered by the server, and so using the STARTINDEX argument causes a 400 error with the response:

> Cannot do natural order without a primary key, please add it or
> specify a manual sort over existing attributes

An example is here: https://geo.irceline.be/wfs?request=GetFeature&version=2.0.0&service=WFS&TYPENAMES=bc_24hmean&COUNT=1000000&STARTINDEX=0

This change avoids issuing `STARTINDEX` if:
* the feature count is known by the time the first GetFeature request is issued
* the feature count is less than the page size.

In other words, this change allows us to use small datasets that have no primary key, while still allowing paging in larger datasets (as long as they have a primary key)

### What's not included

1. In the gdal-dev post, @rouault suggested adding a new value to the existing `OGR_WFS_PAGING_ALLOWED` (`COUNT_WITH_HITS`) in addition to the existing boolean values. I attempted to implement that but quickly came up against infinite recursion, since this makes `MakeGetFeatureURL` call `GetFeatureCount` which in turn eventually calls `MakeGetFeatureURL`. I decided to skip that complication for this PR, which is still useful in most circumstances without it. The WIP change is [here](https://github.com/OSGeo/gdal/compare/master...koordinates:gdal:wfs-disable-paging-check-with-hits) if you want to see it.
2. ~Tests.~ I've now included a test

As discussed in [gdal-dev](https://lists.osgeo.org/pipermail/gdal-dev/2023-May/057174.html) a month ago

## Tasklist

 - [x] Add test case(s)
 - [x] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:

### DIFF
--- a/autotest/ogr/ogr_wfs.py
+++ b/autotest/ogr/ogr_wfs.py
@@ -3920,9 +3920,108 @@ xsi:schemaLocation="http://foo /vsimem/wfs_endpoint?SERVICE=WFS&amp;VERSION=1.1.
         #    return 'fail'
 
 
+def test_ogr_wfs_vsimem_wfs200_with_no_primary_key(with_and_without_streaming):
+    # This server 'supports' paging, but the datasource doesn't have a primary key,
+    # so in practice doesn't actually support paging.
+    with gdal.config_options(
+        {"OGR_WFS_PAGING_ALLOWED": "ON", "OGR_WFS_PAGE_SIZE": "2"}
+    ):
+        with gdaltest.tempfile(
+            "/vsimem/wfs200_endpoint_no_pk?SERVICE=WFS&REQUEST=GetCapabilities",
+            """
+            <WFS_Capabilities version="2.0.0">
+                <OperationsMetadata>
+                    <ows:Operation name="GetFeature">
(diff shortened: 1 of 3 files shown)


rouault:
For tests, the best is to simulate a fake server by providing GetCapabilities, DescribeFeatureCount, GetFeature responses in temporary files like done in most of the tests of ogr_wfs.py.
test_ogr_wfs_vsimem_wfs110_multiple_layers_same_name_different_ns() or test_ogr_wfs_vsimem_wfs200_paging() could potentially serve as inspiration

craigds:
I have added a test that exercises the new behaviour.

craigds:
The CI failures _seem_ to be unrelated to the change I made here. Is there a known issue with CI?

rouault:
> Is there a known issue with CI?

for the Mac one, now, yes :-) Due to a version upgrade of a dependency in homebrew
The build-windows-minimum one must be a flake

PR #8155, opened 2023-07-28
API: SubdatasetInfo 

### LINKED ISSUE #7261: Add API for GDAL users to be able to extract information / transform subdataset names
Cf https://github.com/qgis/QGIS/pull/51901/files#r1109095423 for the context

Perhaps GDALGetFilenameFromSubdatasetName(), GDALModifyFilenameInSubdatasetName(), etc.

elpaso:
Fix #7261 

Initial GDALSubdatasetInfo API.


elpaso:
@rouault I'd like a general review of the API before I get through all the formats, for now the only implemented one is GPKG.

I will also probably need a list of the drivers that support subdatasets.

gcore/gdal.h
```diff
@@ -1290,6 +1290,79 @@ bool CPL_DLL GDALDatasetSetQueryLoggerFunc(
...
+ *  Opaque type used for the C bindings of the C++ GDALSubdatasetInfo class
+ *  @since GDAL 3.8
+*/
+typedef struct GDALSubdatasetInfo *GDALSubdatasetInfoH;
+
+/**
+ * @brief Returns a new GDALSubdatasetInfo object with methods to extract
+ *        and manipulate subdataset information.
+ *        If the pszFileName argument is not recognized by any driver as
+ *        a subdataset descriptor NULL is returned.
```
rouault, line 1307:
```suggestion
 *        a subdataset descriptor, NULL is returned.
```

gcore/gdal.h
```diff
@@ -1290,6 +1290,79 @@ bool CPL_DLL GDALDatasetSetQueryLoggerFunc(
...
+ * @param pszFileName           File name with subdataset information
+ * @note                        This method does not check if the subdataset actually exists.
+ * @return                      Opaque pointer to a GDALSubdatasetInfo object or NULL if no drivers accepted the file name.
+ * @since                       GDAL 3.8
+ */
+GDALSubdatasetInfoH CPL_DLL CPL_STDCALL
+GDALGetSubdatasetInfo(const char *pszFileName);
+
+/**
+ * @brief GDALSubdatasetInfoGetFileName returns the file path (or base) component of a
```
rouault, line 1318: Conventionally, the name of the function isn't contained in the brief

```suggestion
 * @brief Returns the file path (or base) component of a
```

gcore/gdal.h
```diff
@@ -1290,6 +1290,79 @@ bool CPL_DLL GDALDatasetSetQueryLoggerFunc(
...
+ * @note                        This method does not check if the subdataset actually exists.
+ * @return                      TRUE if pszFileName is recognized by at least one driver
+ *                              as a subdataset descriptor.
+ * @since                       GDAL 3.8
+ */
+bool CPL_DLL CPL_STDCALL GDALSubdatasetInfoIsSubdatasetSyntax(
+    GDALSubdatasetInfoH hInfo, const char *pszFileName);
+
+/**
+ * @brief GDALSubdatasetInfoModifyFileName replaces the base component of a
```
rouault, line 1345:
```suggestion
 * @brief Replaces the base component of a
```

gcore/gdal.h
```diff
@@ -1290,6 +1290,79 @@ bool CPL_DLL GDALDatasetSetQueryLoggerFunc(
...
+ * @note                        This method does not check if the subdataset actually exists.
+ * @return                      The original string with the old file name replaced by newFileName and the subdataset information unaltered.
+ * @since                       GDAL 3.8
+ */
+const char CPL_DLL *CPL_STDCALL GDALSubdatasetInfoModifyFileName(
+    GDALSubdatasetInfoH hInfo, const char *pszFileName,
+    const char *pszNewFileName);
+
+/**
+ * @brief GDALDestroySubdatasetInfo destroys a GDALSubdatasetInfo object.
```
rouault, line 1360:
```suggestion
 * @brief Destroys a GDALSubdatasetInfo object.
```

gcore/gdal_priv.h
```diff
@@ -60,6 +60,7 @@ class GDALRelationship;
 
 #include "gdal.h"
 #include "gdal_frmts.h"
+#include "gdalsubdatasetinfo.h"
```
rouault, line 63: this header file should be added in target_public_header() in gcore/CMakeLists.txt

gcore/gdalsubdatasetinfo.cpp
```diff
@@ -0,0 +1,96 @@
...
+ copyright            : (C) 2023 by ale
+ email                : [your-email-here]
+ ***************************************************************************
+ *                                                                         *
+ *   This program is free software; you can redistribute it and/or modify  *
+ *   it under the terms of the GNU General Public License as published by  *
+ *   the Free Software Foundation; either version 2 of the License, or     *
+ *   (at your option) any later version.                                   *
+ *                                                                         *
+ ***************************************************************************/
```
rouault, line 15: to be replaced by the MIT license

gcore/gdal.h
```diff
@@ -1290,6 +1290,79 @@ bool CPL_DLL GDALDatasetSetQueryLoggerFunc(
...
+ *        and manipulate subdataset information.
+ *        If the pszFileName argument is not recognized by any driver as
+ *        a subdataset descriptor NULL is returned.
+ *        The returned object must be freed with GDALDestroySubdatasetInfo().
+ * @param pszFileName           File name with subdataset information
+ * @note                        This method does not check if the subdataset actually exists.
+ * @return                      Opaque pointer to a GDALSubdatasetInfo object or NULL if no drivers accepted the file name.
+ * @since                       GDAL 3.8
+ */
+GDALSubdatasetInfoH CPL_DLL CPL_STDCALL
```
rouault, line 1314: Just raising this once, but the CPL_STDCALL are mostly a historical remain of the times of VB6 if I remember well. So they could be removed in the .h and .cpp

gcore/gdal.h
```diff
@@ -1290,6 +1290,79 @@ bool CPL_DLL GDALDatasetSetQueryLoggerFunc(
...
+ *        subdataset descriptor effectively stripping the information about the subdataset
+ *        and returning the "parent" dataset descriptor.
+ *        The returned string must be freed with CPLFree().
+ * @param hInfo                 Pointer to GDALSubdatasetInfo object
+ * @param pszFileName           File name with subdataset information
+ * @note                        This method does not check if the subdataset actually exists.
+ * @return                      The original string with the subdataset information removed.
+ * @since                       GDAL 3.8
+ */
+const char CPL_DLL *CPL_STDCALL GDALSubdatasetInfoGetFileName(
```
rouault, line 1328: "const char*" is for returns of non-freeable strings

```suggestion
char CPL_DLL *CPL_STDCALL GDALSubdatasetInfoGetFileName(
```

gcore/gdal.h
```diff
@@ -1290,6 +1290,79 @@ bool CPL_DLL GDALDatasetSetQueryLoggerFunc(
...
+ *        file name by keeping the subdataset information unaltered.
+ *        The returned string must be freed with CPLFree().
+ * @param hInfo                 Pointer to GDALSubdatasetInfo object
+ * @param pszFileName           File name with subdataset information
+ * @param pszNewFileName        New file name with no subdataset information
+ * @note                        This method does not check if the subdataset actually exists.
+ * @return                      The original string with the old file name replaced by newFileName and the subdataset information unaltered.
+ * @since                       GDAL 3.8
+ */
+const char CPL_DLL *CPL_STDCALL GDALSubdatasetInfoModifyFileName(
```
rouault, line 1355:
```suggestion
char CPL_DLL *CPL_STDCALL GDALSubdatasetInfoModifyFileName(
```

gcore/gdalsubdatasetinfo.cpp
```diff
@@ -0,0 +1,96 @@
...
+}
+
+const char *CPL_STDCALL GDALSubdatasetInfoGetFileName(GDALSubdatasetInfoH hInfo,
+                                                      const char *pszFileName)
+{
+    const std::string fileName{
+        hInfo->GetFilenameFromSubdatasetName(pszFileName)};
+    char *ptr = new char[fileName.size() + 1];
+    strcpy(ptr, fileName.c_str());
+    return ptr;
```
rouault, line 78:
```suggestion
    return CPLStrdup(hInfo->GetFilenameFromSubdatasetName(pszFileName).c_str());
```

gcore/gdalsubdatasetinfo.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+const char *CPL_STDCALL GDALSubdatasetInfoModifyFileName(
+    GDALSubdatasetInfoH hInfo, const char *pszFileName,
+    const char *pszNewFileName)
+{
+    const std::string fileName{
+        hInfo->ModifyFileName(pszFileName, pszNewFileName)};
+    char *ptr = new char[fileName.size() + 1];
+    strcpy(ptr, fileName.c_str());
+    return ptr;
```
rouault, line 95:
```suggestion
    return CPLStrdup(hInfo->ModifyFileName(pszFileName, pszNewFileName).c_str());
```

gcore/gdalsubdatasetinfo.h
```diff
@@ -0,0 +1,80 @@
...
+ copyright            : (C) 2023 by ale
+ email                : [your-email-here]
+ ***************************************************************************
+ *                                                                         *
+ *   This program is free software; you can redistribute it and/or modify  *
+ *   it under the terms of the GNU General Public License as published by  *
+ *   the Free Software Foundation; either version 2 of the License, or     *
+ *   (at your option) any later version.                                   *
+ *                                                                         *
+ ***************************************************************************/
```
rouault, line 15: MIT license

gcore/gdal_priv.h
```diff
@@ -1812,6 +1813,9 @@ class CPL_DLL GDALDriver : public GDALMajorObject
         CSLConstList papszVectorTranslateArguments,
         GDALProgressFunc pfnProgress, void *pProgressData) = nullptr;
 
+    /* Returns a (possibly null) pointer to the Subdataset informational function */
+    GDALSubdatasetInfo *(*pfnGetSubdatasetInfoFunc)() = nullptr;
```
rouault, line 1817: I would have make pfnGetSubdatasetInfoFunc take a const char* argument with the subdataset string. That way this saves the GDALGetSubdatasetInfo::IsSubdatasetSyntax() method, and passing the subdataset name as an argument of each method of GDALGetSubdatasetInfo
(even if that method was kept, I don't think it should be exposed in the C API, as it is mostly an implementation detail of GDALSubdatasetInfo)

ogr/ogrsf_frmts/gpkg/ogrgeopackagedriver.cpp
```diff
@@ -246,6 +246,88 @@ static int OGRGeoPackageDriverIdentify(GDALOpenInfo *poOpenInfo)
...
 
+/************************************************************************/
+/*                    OGRGeoPackageDriverGetSubdatasetInfo()            */
+/************************************************************************/
+
+struct OGRGeoPackageDriverSubdatasetInfo : public GDALSubdatasetInfo
+{
+    virtual ~OGRGeoPackageDriverSubdatasetInfo() = default;
+
+    bool IsSubdatasetSyntax(const std::string &fileName) const override
```
rouault, line 257: seeing this implementation indeed makes me thinks that my above remark about pfnGetSubdatasetInfoFunc taking the subdataset name to be probably better, as saving lines of code. The components of the subdataset could be saved as member variable(s) of the class during construction

The advantage of your approach is that it enables to deal with several subdatasets with a same GDALSubdatasetInfo instance, but I don't think that's a real use case. And the name SubdatasetInfo would make me think that this is a class whose instances know about a particular subdataset

ogr/ogrsf_frmts/gpkg/ogrgeopackagedriver.cpp
```diff
@@ -246,6 +246,88 @@ static int OGRGeoPackageDriverIdentify(GDALOpenInfo *poOpenInfo)
...
+        if (OGRGeoPackageDriverIdentify(&poOpenInfo) != TRUE)
+        {
+            return false;
+        }
+
+        // Split
+        char **papszParts{CSLTokenizeString2(fileName.c_str(), ":", 0)};
+        const int iPartsCount{CSLCount(papszParts)};
+        CSLDestroy(papszParts);
+        return iPartsCount > 2;
```
rouault, line 269: This test is too generic. It would catch "c:\foo\bar.gpkg" as a subdataset name

ogr/ogrsf_frmts/gpkg/ogrgeopackagedriver.cpp
```diff
@@ -354,6 +436,8 @@ class GDALGPKGDriver final : public GDALDriver
         InitializeCreationOptionList();
         return GDALDriver::GetMetadata(pszDomain);
     }
+
+    GDALSubdatasetInfo *GetSubdatasetInfo();
```
rouault, line 440: Probably a remain of initial tries:

```suggestion
```

swig/include/SubdatasetInfo.i
```diff
@@ -0,0 +1,89 @@
...
+class GDALSubdatasetInfoShadow {
+
+    public:
+
+%extend {
+
+        GDALSubdatasetInfoShadow(){
+            return new GDALSubdatasetInfo();
+        }
+
```
rouault, line 53: We don't need a constructor:

```suggestion
```

swig/include/SubdatasetInfo.i
```diff
@@ -0,0 +1,89 @@
...
+        ~GDALSubdatasetInfoShadow() {
+            GDALDestroySubdatasetInfo(reinterpret_cast<GDALSubdatasetInfoH>(self));
+        }
+
+        const char* GetFilenameFromSubdatasetName(const char *pszFileName)
```
rouault, line 58:
```suggestion
        retStringAndCPLFree* GetFilenameFromSubdatasetName(const char *pszFileName)
```

swig/include/SubdatasetInfo.i
```diff
@@ -0,0 +1,89 @@
...
+        {
+            return GDALSubdatasetInfoGetFileName(reinterpret_cast<GDALSubdatasetInfoH>(self), pszFileName );
+        }
+
+        const char* ModifyFileName(const char *pszFileName, const char *pszNewFileName)
```
rouault, line 63:
```suggestion
        retStringAndCPLFree* ModifyFileName(const char *pszFileName, const char *pszNewFileName)
```

swig/include/SubdatasetInfo.i
```diff
@@ -0,0 +1,89 @@
...
+        }
+
+        bool IsSubdatasetSyntax(const char *pszFileName)
+        {
+            return GDALSubdatasetInfoIsSubdatasetSyntax(reinterpret_cast<GDALSubdatasetInfoH>(self), pszFileName );
+        }
+}
+};
+
+%newobject GetSubdatasetInfo(const char *pszFileName);
```
rouault, line 75: was probably fine, but it seems we use the short form usually:
```suggestion
%newobject GetSubdatasetInfo;
```

gcore/gdal.h
elpaso: If that's the case they should be removed from all source files not just these ones.

ogr/ogrsf_frmts/gpkg/ogrgeopackagedriver.cpp
elpaso: Sounds good, I had the same doubt but reading the issue comments it wasn't totally clear which way you wanted to go.

swig/include/SubdatasetInfo.i
elpaso: I doesn't build without it.

gcore/gdal.h
rouault: Changing the call convention is an ABI break, so we cannot do that lightly. My point was that there was no need to perpetuate using CPL_STDCALL for new API.

swig/include/SubdatasetInfo.i
rouault: looking at other SWIG classes, you need to use that pattern:
```
class GDALSubdatasetInfoShadow {
private:
  GDALSubdatasetInfoShadow();
public:
%extend {
```

gcore/gdalsubdatasetinfo.cpp
```diff
@@ -0,0 +1,137 @@
...
+ *                                                                         *
+ ***************************************************************************/
+#include "gdalsubdatasetinfo.h"
+#include "gdal_priv.h"
+
+/************************************************************************/
+/*                     Subdataset informational functions               */
+/************************************************************************/
+
+GDALSubdatasetInfoH CPL_STDCALL GDALGetSubdatasetInfo(const char *pszFileName)
```
rouault, line 36:
```suggestion
GDALSubdatasetInfoH GDALGetSubdatasetInfo(const char *pszFileName)
```

gcore/gdalsubdatasetinfo.h
```diff
@@ -0,0 +1,94 @@
...
+    //! @cond Doxygen_Suppress
+  protected:
+    /**
+     * This method is called once to parse the fileName and populate the member variables.
+     * It must be reimplemented by concrete derived classes.
+     */
+    virtual void parseFileName() = 0;
+    mutable bool m_initialized = false;
+    std::string m_fileName;
+    std::string m_baseComponent;
```
rouault, line 88: The difference of semantics between those 2 is unclear to me.
I believe some comments showing how an example subdataset name is split into all those member variables would be helpful

gcore/gdalsubdatasetinfo.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+std::string GDALSubdatasetInfo::GetFileName() const
+{
+    if (!m_initialized)
+    {
+        GDALSubdatasetInfo *this_ = const_cast<GDALSubdatasetInfo *>(this);
+        this_->parseFileName();
+        m_initialized = true;
+    }
+    return m_driverPrefixComponent + ":" + m_baseComponent;
```
rouault, line 109: I would expect this to be just:
```suggestion
    return m_filename;
```
I would avoid assuming at this generic level that a ':' is used to join m_driverPrefixComponent and m_baseComponent
I guess if there is a need to return m_driverPrefixComponent + ":" + m_baseComponent that should be through a virtual method implemented in the driver

gcore/gdal.h
```diff
@@ -1290,6 +1290,71 @@ bool CPL_DLL GDALDatasetSetQueryLoggerFunc(
...
+
+/**
+ * @brief Returns the subdataset component of a subdataset file name descriptor.
+ *        The returned string must be freed with CPLFree().
+ * @param hInfo                 Pointer to GDALSubdatasetInfo object
+ * @note                        This method does not check if the subdataset actually exists.
+ * @return                      The subdataset name.
+ * @since                       GDAL 3.8
+ */
+char CPL_DLL *GDALSubdatasetInfoGetSubdatasetName(GDALSubdatasetInfoH hInfo);
```
rouault, line 1336: I would rather perhaps call it
```suggestion
char CPL_DLL *GDALSubdatasetInfoGetSubdatasetComponent(GDALSubdatasetInfoH hInfo);
```
(and with related changes at the C++ and SWIG level)

since in the terminology used else where the subdataset name is the full  "netCDF:foo.nc:my_variable"

gcore/gdalsubdatasetinfo.h
elpaso: m_fileName is the unaltered string set in the constructor, the *Component are the individual pieces of information after the parsing (split `:`) operation

elpaso:
@rouault I agree there is a little confusion in the terminology, here is my proposal to clarify and avoid ambiguity:

The complete file name (e.g. `GPKG:/path_to.gpkg:layer_name:var_name`) will be named "File name"  and the methods and variables will be renamed accordingly.

The parsed components will be named:

- driver prefix e.g. `GPKG`  -> "Prefix component"
- file path: e.g. `/path_to.gpkg` -> "Path component"
- subdataset: e.g. `layer_name` -> "Subdataset component"

So, the methods will become:

GetFileName()  -> returns the original complete file name e.g. `GPKG:/path_to.gpkg:layer_name:var_name`
GetPathComponent() ->  returns the path component e.g. `/path_to.gpkg` 
GetPrefixComponent() -> returns the prefix component e.g. `GPKG` 
GetSubdatasetComponent() -> returns the subdataset component  e.g. `layer_name`
ModifyPathComponent( newPathComponent) -> replaces the path component

Or, the shorter versions without `Component`:
GetFileName() 
GetPath()
GetPrefix()
GetSubdataset()
ModifyPath(newPath)

rouault:
> GetSubdataset()

or maybe GetSuffix() ?

elpaso:
> > GetSubdataset()
> 
> or maybe GetSuffix() ?

I don't follow: isn't subdataset more appropriate to describe it? Suffix is really generic and if we have variables it might not even be the last component in the descriptor.

rouault:
> I don't follow: isn't subdataset more appropriate to describe it? Suffix is really generic and if we have variables it might not even be the last component in the descriptor.

ok, let's keep GetSubdataset(). "There are only two hard things in Computer Science: cache invalidation and naming things. " :-)

gcore/gdalsubdatasetinfo.cpp
elpaso: Ok, my idea was to return the path to the parent dataset with the subdataset information stripped off, I thought it was useful but we can add it later if needed.

frmts/gtiff/geotiff.cpp
```diff
@@ -1132,6 +1132,84 @@ CPLString GTiffGetCompressValues(bool &bHasLZW, bool &bHasDEFLATE,
     return osCompressValues;
 }
 
+/************************************************************************/
+/*                    OGRGTiffDriverGetSubdatasetInfo()                 */
+/************************************************************************/
+
+struct OGRGTiffDriverSubdatasetInfo : public GDALSubdatasetInfo
```
rouault, line 1139: should rather be:
```suggestion
struct GDALGTiffDriverSubdatasetInfo : public GDALSubdatasetInfo
```
or maybe just:
```suggestion
struct GTiffDriverSubdatasetInfo : public GDALSubdatasetInfo
```

frmts/gtiff/geotiff.cpp
```diff
@@ -1132,6 +1132,84 @@ CPLString GTiffGetCompressValues(bool &bHasLZW, bool &bHasDEFLATE,
...
+    // GDALSubdatasetInfo interface
+  private:
+    void parseFileName() override
+    {
+        if (!STARTS_WITH_CI(m_fileName.c_str(), "GTIFF_DIR:"))
+        {
+            return;
+        }
+
+        char **papszParts{CSLTokenizeString2(m_fileName.c_str(), ":", 0)};
```
rouault, line 1156: you can use CPLStringList to avoid having to explicitly free
```suggestion
        CPLStringList aosParts(CSLTokenizeString2(m_fileName.c_str(), ":", 0));
```

frmts/gtiff/geotiff.cpp
```diff
@@ -1132,6 +1132,84 @@ CPLString GTiffGetCompressValues(bool &bHasLZW, bool &bHasDEFLATE,
...
+{
+  public:
+    explicit OGRGTiffDriverSubdatasetInfo(const std::string &fileName)
+        : GDALSubdatasetInfo(fileName)
+    {
+    }
+
+    // GDALSubdatasetInfo interface
+  private:
+    void parseFileName() override
```
rouault, line 1149: what do you think about the opportunity of using this in the GTiffDataset::OpenDir() method to avoid duplication of parsing ? (this could apply to other drivers as well)

frmts/gtiff/geotiff.cpp
```diff
@@ -1132,6 +1132,84 @@ CPLString GTiffGetCompressValues(bool &bHasLZW, bool &bHasDEFLATE,
...
+    {
+        if (!STARTS_WITH_CI(m_fileName.c_str(), "GTIFF_DIR:"))
+        {
+            return;
+        }
+
```
rouault, line 1155: There is a more esoteric supported variant:   GTIFF_DIR:off:<dir_offset>:filename
If we reused this class for GTiffDataset::OpenDir(), we'd have to support it

frmts/postgisraster/postgisrasterdataset.cpp
```diff
@@ -4109,6 +4109,86 @@ GBool PostGISRasterDataset::PolygonFromCoords(int nXOff, int nYOff,
     return true;
 }
 
+/************************************************************************/
+/*                    OGRPostGISRasterDriverGetSubdatasetInfo()            */
+/************************************************************************/
+
+struct OGRPostGISRasterDriverSubdatasetInfo : public GDALSubdatasetInfo
```
rouault, line 4116:
```suggestion
struct PostGISRasterDriverSubdatasetInfo : public GDALSubdatasetInfo
```

frmts/wms/wmsdriver.cpp
```diff
@@ -1132,6 +1132,67 @@ void WMSDeregister(CPL_UNUSED GDALDriver *d)
     };                                                                         \
     WMSRegisterMiniDriverFactory(new WMSMiniDriverFactory_##name());
 
+/************************************************************************/
+/*                    OGRWMSDriverGetSubdatasetInfo()                   */
+/************************************************************************/
+
+struct OGRWMSDriverSubdatasetInfo : public GDALSubdatasetInfo
```
rouault, line 1139:
```suggestion
struct WMSDriverSubdatasetInfo : public GDALSubdatasetInfo
```

ogr/ogrsf_frmts/gpkg/ogrgeopackagedriver.cpp
```diff
@@ -246,6 +246,86 @@ static int OGRGeoPackageDriverIdentify(GDALOpenInfo *poOpenInfo)
...
+    // GDALSubdatasetInfo interface
+  private:
+    void parseFileName() override
+    {
+        if (!STARTS_WITH_CI(m_fileName.c_str(), "GPKG:"))
+        {
+            return;
+        }
+
+        char **papszParts{CSLTokenizeString2(m_fileName.c_str(), ":", 0)};
```
rouault, line 270:
```suggestion
        CPLStringList aosParts(CSLTokenizeString2(m_fileName.c_str(), ":", 0));
```

frmts/gtiff/geotiff.cpp
```diff
@@ -1132,6 +1132,82 @@ CPLString GTiffGetCompressValues(bool &bHasLZW, bool &bHasDEFLATE,
...
+            const bool hasDriveLetter{strlen(aosParts[2]) == 1 &&
+                                      std::isalpha(aosParts[2][0])};
+
+            // Check for drive letter
+            if (iPartsCount == 4)
+            {
+                // Invalid
+                if (!hasDriveLetter)
+                {
+                    CSLDestroy(aosParts);
```
rouault, line 1173:
```suggestion
```

frmts/gtiff/geotiff.cpp
```diff
@@ -1132,6 +1132,82 @@ CPLString GTiffGetCompressValues(bool &bHasLZW, bool &bHasDEFLATE,
...
+                }
+                m_pathComponent = aosParts[2];
+                m_pathComponent.append(":");
+                m_pathComponent.append(aosParts[3]);
+            }
+            else  // count is 3
+            {
+                if (hasDriveLetter)
+                {
+                    CSLDestroy(aosParts);
```
rouault, line 1184:
```suggestion
```

ogr/ogrsf_frmts/gpkg/ogrgeopackagedriver.cpp
```diff
@@ -246,6 +246,85 @@ static int OGRGeoPackageDriverIdentify(GDALOpenInfo *poOpenInfo)
...
+            const bool hasDriveLetter{strlen(aosParts[1]) == 1 &&
+                                      std::isalpha(aosParts[1][0])};
+
+            // Check for drive letter
+            if (iPartsCount == 4)
+            {
+                // Invalid
+                if (!hasDriveLetter)
+                {
+                    CSLDestroy(aosParts);
```
rouault, line 288:
```suggestion
```

ogr/ogrsf_frmts/gpkg/ogrgeopackagedriver.cpp
```diff
@@ -246,6 +246,85 @@ static int OGRGeoPackageDriverIdentify(GDALOpenInfo *poOpenInfo)
...
+                m_pathComponent = aosParts[1];
+                m_pathComponent.append(":");
+                m_pathComponent.append(aosParts[2]);
+                subdatasetIndex++;
+            }
+            else  // count is 3
+            {
+                if (hasDriveLetter)
+                {
+                    CSLDestroy(aosParts);
```
rouault, line 300:
```suggestion
```

frmts/gtiff/geotiff.cpp
elpaso: I'll have a look, I prefer a method that does not have to open the file though, ideally this API should just manipulate strings without I/O overhead.

sgillies:
@rouault @elpaso how did we arrive at using `:` as a delimiter between dataset and subdataset name? Does this have to be extended to drivers where it doesn't exist already? I think `:` is a poor choice and leads to people putting quotes around dataset URLs within a GDAL filename.

elpaso:
> @rouault @elpaso how did we arrive at using `:` as a delimiter between dataset and subdataset name? Does this have to be extended to drivers where it doesn't exist already? I think `:` is a poor choice and leads to people putting quotes around dataset URLs within a GDAL filename.

I don't know where the colon separator comes from, but I totally agree that it's a poor choice, the QGIS counterpart QgsDatasourceUri is a little bit better engineered, for most datasources it uses URL encoding for the KVP and it allows to encode the opening options.

rouault:
> @rouault @elpaso how did we arrive at using `:` as a delimiter between dataset and subdataset name? Does this have to be extended to drivers where it doesn't exist already? I think `:` is a poor choice and leads to people putting quotes around dataset URLs within a GDAL filename.

@sgillies  Before my involvement with the project. I can see that this syntax was already used in the initial version of the HDF4 driver back in 2002:  [7c56e20797c17d2bcb3f173d652ec1fdf590d558](https://github.com/OSGeo/gdal/commit/7c56e20797c17d2bcb3f173d652ec1fdf590d558#diff-5b0f8cff7aa04486157ad10213a9a297a72afa1301576a3071f3352f86c0ef13R656)
While this is indeed not ideal, changing the subdataset syntax in existing drivers would have backward compatibility issues. I suspect a number of people have VRT files referencing the old syntax, or some code manually composing them.
The new API itself in this PR makes no assumption on the separator, so I don't think it is a step backwards, and it should be hopefully compatible with better subsdataset syntax.
That said the replacement logic in ModifyPathComponent() is fully string based currently. if at some point a driver would URI-encode the filename, that wouldn't work. But if that occured, the method could be made virtual and implemented by the driver that would deal with the URI encoding/decoding.

@elpaso  I believe to keep this PR manageable we should target the most popular file-based drivers that use subsdataset (others might follow in subsequent PR): netCDF, HDF4, HDF5, GPKG . Anything else ?
I'd be interested however to see frmts/vrt/vrtsources.cpp updated to remove the NETCDF releated hacks in favor of the new API to demonstrate its usefulness (hum I see that to remove all of them we should also tackle NITF, PDF, RASTERLITE and TILEDB)

elpaso:
> > @rouault @elpaso how did we arrive at using `:` as a delimiter between dataset and subdataset name? Does this have to be extended to drivers where it doesn't exist already? I think `:` is a poor choice and leads to people putting quotes around dataset URLs within a GDAL filename.
> 
> @sgillies Before my involvement with the project. I can see that this syntax was already used in the initial version of the HDF4 driver back in 2002: [7c56e20797c17d2bcb3f173d652ec1fdf590d558](https://github.com/OSGeo/gdal/commit/7c56e20797c17d2bcb3f173d652ec1fdf590d558#diff-5b0f8cff7aa04486157ad10213a9a297a72afa1301576a3071f3352f86c0ef13R656) While this is indeed not ideal, changing the subdataset syntax in existing drivers would have backward compatibility issues. I suspect a number of people have VRT files referencing the old syntax, or some code manually composing them. The new API itself in this PR makes no assumption on the separator, so I don't think it is a step backwards, and it should be hopefully compatible with better subsdataset syntax. That said the replacement logic in ModifyPathComponent() is fully string based currently. if at some point a driver would URI-encode the filename, that wouldn't work. But if that occured, the method could be made virtual and implemented by the driver that would deal with the URI encoding/decoding.
> 

Yeah.

> @elpaso I believe to keep this PR manageable we should target the most popular file-based drivers that use subsdataset (others might follow in subsequent PR): netCDF, HDF4, HDF5, GPKG . Anything else ? I'd be interested however to see frmts/vrt/vrtsources.cpp updated to remove the NETCDF releated hacks in favor of the new API to demonstrate its usefulness (hum I see that to remove all of them we should also tackle NITF, PDF, RASTERLITE and TILEDB)

I have actually already implemented WMS and PostGIS raster because I wanted to test the API with a DB and a webservice. I will proceed according to your priority list.

elpaso:
@rouault here is another iteration. I have removed the double quotes from the logic by storing both the original unmodified path (used only internally to modify the path) and I am storing isQuoted to determine if the new path should be quoted (if it is already quoted it is not quoted twice).

When returning the path is is always unquoted and unescaped.

I have added a few new test cases.

rouault:
@sgillies Do you think we can merge this PR or do you want the RFC you mention in https://lists.osgeo.org/pipermail/gdal-dev/2023-September/057641.html to be a pre-condition for it ?

rouault:
@elpaso Can you rebase on top of latest master and fix conflict in autotest/gdrivers/netcdf.py?

elpaso:
> @elpaso Can you rebase on top of latest master and fix conflict in autotest/gdrivers/netcdf.py?

done

autotest/gdrivers/netcdf.py
```diff
@@ -6451,3 +6451,73 @@ def test_netcdf_NASA_EMIT():
         "Y_BAND": "1",
         "Y_DATASET": 'NETCDF:"data/netcdf/fake_EMIT.nc":/location/lat',
     }
+
+
+###########################################################
+# Test gdal subdataset informational functions
+
+@pytest.mark.parametrize(
```
rouault, line 6459:
```suggestion

@pytest.mark.parametrize(
```

PR #8160, opened 2023-07-31
Make the java 2d multidim api test code exercise itself harder

bdezonia:
It looks like I had left in a very small plane size that I had used for debugging.


bdezonia:
So the build failure looks unrelated to my changes. There was some failure to access some url in the http_2 tests, I think. I think you can merge my code at any time.

PR #8173, opened 2023-08-05
Update FindSPATIALITE.cmake for SpatiaLite 5.1.0

jef-n:
Fixes detection of SpatiaLite 5.1.0.  Version string changed from "5.0, something" to "5.1.0, something"

### DIFF
--- a/cmake/modules/packages/FindSPATIALITE.cmake
+++ b/cmake/modules/packages/FindSPATIALITE.cmake
@@ -55,8 +55,8 @@ mark_as_advanced(SPATIALITE_LIBRARY SPATIALITE_INCLUDE_DIR)
 if(SPATIALITE_LIBRARY AND SPATIALITE_INCLUDE_DIR
    AND NOT SPATIALITE_VERSION_STRING)
     file(STRINGS "${SPATIALITE_INCLUDE_DIR}/spatialite.h" _spatialite_h_ver
-         REGEX "^[ \t]version[ \t]([0-9]+\\.[0-9]+),.*")
-    string(REGEX REPLACE "[ \t]version[ \t]([0-9]+\\.[0-9]+),.*" "\\1" _spatialite_h_ver ${_spatialite_h_ver})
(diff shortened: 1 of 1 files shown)


cmake/modules/packages/FindSPATIALITE.cmake
```diff
@@ -55,8 +55,8 @@ mark_as_advanced(SPATIALITE_LIBRARY SPATIALITE_INCLUDE_DIR)
 if(SPATIALITE_LIBRARY AND SPATIALITE_INCLUDE_DIR
    AND NOT SPATIALITE_VERSION_STRING)
     file(STRINGS "${SPATIALITE_INCLUDE_DIR}/spatialite.h" _spatialite_h_ver
-         REGEX "^[ \t]version[ \t]([0-9]+\\.[0-9]+),.*")
-    string(REGEX REPLACE "[ \t]version[ \t]([0-9]+\\.[0-9]+),.*" "\\1" _spatialite_h_ver ${_spatialite_h_ver})
+         REGEX "^[ \t]version[ \t]([0-9]+\\.[0-9]+)[,.].*")
+    string(REGEX REPLACE "[ \t]version[ \t]([0-9]+\\.[0-9]+)[,.].*" "\\1" _spatialite_h_ver ${_spatialite_h_ver})
```
dg0yt, line 59: Note that the last argument here is the same ticking bomb as before: 
When `file(STRINGS ...)` returns an empty value, `string(REGEX REPLACE ...)`  will see only five arguments, and raise an error:
~~~
CMake Error at cmake/modules/packages/FindSPATIALITE.cmake:67 (string):
  string sub-command REGEX, mode REPLACE needs at least 6 arguments total to
  command.
~~~
The fix is to add quotes.
(Noticed while reviewing https://github.com/microsoft/vcpkg/issues/33039.)

rouault: > The fix is to add quotes.

done in 37391454a7

PR #8185, opened 2023-08-08
autotest: Make more tests independent

dbaston:
## What does this PR do?

Continues #8097 for most of the `gcore` and `gdrivers` sets of tests.

#4407

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/autotest/gdrivers/pcidsk.py
+++ b/autotest/gdrivers/pcidsk.py
@@ -97,50 +97,48 @@ def test_pcidsk_4():
 # Test writing metadata to a newly created file.
 
 
-def test_pcidsk_5():
+def test_pcidsk_5(tmp_path):
+
+    testfile = str(tmp_path / "pcidsk_5.pix")
 
     # Create testing file.
 
-    gdaltest.pcidsk_ds = gdal.GetDriverByName("PCIDSK").Create(
-        "tmp/pcidsk_5.pix", 400, 600, 1, gdal.GDT_Byte
+    pcidsk_ds = gdal.GetDriverByName("PCIDSK").Create(
+        testfile, 400, 600, 1, gdal.GDT_Byte
     )
 
     # Write out some metadata to the default and non-default domain and
     # using the set and single methods.
 
-    gdaltest.pcidsk_ds.SetMetadata(["ABC=DEF", "GHI=JKL"])
-    gdaltest.pcidsk_ds.SetMetadataItem("XXX", "YYY")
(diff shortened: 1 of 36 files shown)


autotest/gdrivers/pcidsk.py
```diff
@@ -97,62 +97,70 @@ def test_pcidsk_4():
...
 
-    gdaltest.pcidsk_ds = gdal.GetDriverByName("PCIDSK").Create(
-        "tmp/pcidsk_5.pix", 400, 600, 1, gdal.GDT_Byte
-    )
+
+def test_pcidsk_5(pcidsk_5):
 
     # Write out some metadata to the default and non-default domain and
     # using the set and single methods.
+    pcidsk_ds = gdal.Open(pcidsk_5, gdal.GA_Update)
```
rouault, line 115: Create() + SetMetadata()  is not necessarily the same as Create() + Close() + Open_in_update_mode() + SetMetadata()  (speaking generally. I haven't looked at the details of this particular driver)
Perhaps just combine test_pcidsk_5,6 and 7 in the same function ?

autotest/gdrivers/plmosaic.py
```diff
@@ -179,6 +190,12 @@ def test_plmosaic_8():
 }""",
     )
 
+    yield
```
rouault, line 193: Not necessarily to be done in this PR, but I'm thinking that it would be cleaner if those temp /vsimem/ files would be removed after the test is finished. So presumably as a cleanup step after the yield

An alternative would be to use
```
with gdaltest.tempfile("/vsimem/foo", "content"), gdaltest.tempfile("/vsimem/bar", "content"):
    yield
```
which takes care of the cleanup.

The same applies for the valid_mosaic fixture

rouault (commented):
Thanks again!

autotest/gdrivers/plmosaic.py
dbaston: What do you think of the generic cleanup on line 56? That way, whatever mess a given test creates, subsequent tests will not be affected.

rouault: > What do you think of the generic cleanup on line 56?

ah, I thought it was a module level setup_and_cleanup() like done usually, and didn't realize that it would apply to all tests. So ignore my above remark

dbaston: Maybe it should be renamed `setup_and_cleanup_test` ?

Somewhat related: if tests are running in parallel, they share a global `/vsimem` space, right? So to avoid potential collisions there might be use for some sort of `mem_temp_path` fixture?

PR #8206, opened 2023-08-10
CI: workaround weird caching issue with fedora_rawhide builds

rouault:
CC @dbaston in case you've an idea... I've also manually deleted the image ghcr.io/osgeo/gdal-deps:fedorarawhide-master before doing a master push build, but this doesn't change anything.

### DIFF
--- a/.github/workflows/linux_build.yml
+++ b/.github/workflows/linux_build.yml
@@ -163,8 +163,19 @@ jobs:
         env:
           DOCKER_BUILDKIT: 1
         run: |
+          # FIXME: for some reason, the fedora rawhide container pushed by
+          # master job is corrupted (looks like it contains an outdated layer
+          # symlinking libssl.so.3 to an older version of the actual file),
+          # once it is pushed. But in the job that generates it,
+          # compilation & tests work fine. It looks like some weird caching
+          # issue
+          if test "${{ matrix.container }}" = "fedora_rawhide"; then
+            DOCKER_BUILD_CACHE_FROM=""
+          else
+            BUILD_ARG_INLINE_CACHE="--build-arg BUILDKIT_INLINE_CACHE=1"
+          fi
           docker build \
-            --build-arg BUILDKIT_INLINE_CACHE=1 \
+            ${BUILD_ARG_INLINE_CACHE} \
             ${DOCKER_BUILD_CACHE_FROM} \
             -t ${CONTAINER_NAME_FULL} \
             -f .github/workflows/${{ matrix.container }}/Dockerfile.ci \


rouault:
Cf https://github.com/OSGeo/gdal/actions/runs/5819155820/job/15777004256 for the job where the image was generated and if you use, you can see the broken symlink:
```
$ docker run --rm -it ghcr.io/osgeo/gdal-deps:fedorarawhide-master sh -c "ls -al /usr/lib64/libssl*"
lrwxrwxrwx 1 root root     15 Jul 27 00:00 /usr/lib64/libssl.so -> libssl.so.3.1.1
lrwxrwxrwx 1 root root     15 Jul 27 00:00 /usr/lib64/libssl.so.3 -> libssl.so.3.0.8
-rwxr-xr-x 1 root root 676288 Jul 27 00:00 /usr/lib64/libssl.so.3.1.1
-rwxr-xr-x 1 root root 430984 Aug  1 00:00 /usr/lib64/libssl3.so
```

dbaston:
Really strange. I'm not sure it's a caching issue, because if I directly pull and run the image generated by https://github.com/OSGeo/gdal/actions/runs/5819155820/job/15777004256 I see the same symlink problems locally. They're fixed by `ldconfig`.

```
➜  gdal git:(pytest-independent) ✗ docker pull ghcr.io/osgeo/gdal-deps:fedorarawhide-master
fedorarawhide-master: Pulling from osgeo/gdal-deps
Digest: sha256:f2884143896b5f938805d78ceff79d915d452b460808b684e8600298b31df80f
Status: Image is up to date for ghcr.io/osgeo/gdal-deps:fedorarawhide-master
ghcr.io/osgeo/gdal-deps:fedorarawhide-master
➜  gdal git:(pytest-independent) ✗ docker run -it --rm -v /home/dan/dev/gdal:/gdal ghcr.io/osgeo/gdal-deps:fedorarawhide-master
[root@6b84c447b26e /]# dnf5 update
dnf5: error while loading shared libraries: libcrypto.so.3: cannot open shared object file: No such file or directory
[root@6b84c447b26e /]# ldconfig
[root@6b84c447b26e /]# dnf5 update
Updating and loading repositories:
Repositories loaded.
Nothing to do.
```

rouault:
> I'm not sure it's a caching issue, because if I directly pull and run the image generated by https://github.com/OSGeo/gdal/actions/runs/5819155820/job/15777004256 I see the same symlink problems locally.

well, what makes me believe there's an issue with caching or push is that the image works correctly in the job that generates it for all the "docker run" steps that use it, between the "docker build" and "docker push" steps

PR #8214, opened 2023-08-10
OGRParseDate(): restrict valid times to HH:MM:SS(.sss) with at least …

### LINKED ISSUE #8150: False positive guessing of time format
Hi all, I have a string in the geojson with the following text "2:3".  Automatically in the conversion from geojson to ESRI shapefile the library recognizes the string as if it were in Time format.  Isn't there a way to ignore that field and tell it not to do the type conversion? 

Thank you all

rouault:
…2 figures (fixes #8150)

Also accept:
- THH:MM(:SS(.sss)?)?Z? (ISO 8601 extended format)
- THHMM(SS(.sss)?)?Z? (ISO 8601 basic format)


rouault:
not going to backport this, as could cause regressions in the stable branch to the tighten format checks

PR #8220, opened 2023-08-11
Use unaligned memory accesses unconditionally

MaskRay:
CPL_CPU_REQUIRES_ALIGNED_ACCESS is undefined for x86 which can cause -fsanitize=alignment failures.

Let's just use CPL_CPU_REQUIRES_ALIGNED_ACCESS unconditionally. Modern compilers are able to optimize memcpy.

It seems that GDALCopyWords64 performs more conditions and FileGDBIndexIterator::GetNextRow() has less efficient code with this change, but the difference unlikely matters in practice.

---

## What does this PR do?

Fix -fsanitize=alignment failures on x86 even if `MAKE_SANITIZE_HAPPY` is not defined.

### DIFF
--- a/alg/gdalwarpkernel.cpp
+++ b/alg/gdalwarpkernel.cpp
@@ -2962,13 +2962,9 @@ static bool GWKCubicResample4Sample(const GDALWarpKernel *poWK, int iBand,
 
 static CPL_INLINE __m128 XMMLoad4Values(const GByte *ptr)
 {
-#ifdef CPL_CPU_REQUIRES_ALIGNED_ACCESS
     unsigned int i;
     memcpy(&i, ptr, 4);
     __m128i xmm_i = _mm_cvtsi32_si128(s);
-#else
-    __m128i xmm_i = _mm_cvtsi32_si128(*(unsigned int *)(ptr));
-#endif
     // Zero extend 4 packed unsigned 8-bit integers in a to packed
     // 32-bit integers.
 #if __SSE4_1__
@@ -2982,13 +2978,9 @@ static CPL_INLINE __m128 XMMLoad4Values(const GByte *ptr)
 
 static CPL_INLINE __m128 XMMLoad4Values(const GUInt16 *ptr)
 {
-#ifdef CPL_CPU_REQUIRES_ALIGNED_ACCESS
     GUInt64 i;
     memcpy(&i, ptr, 8);
     __m128i xmm_i = _mm_cvtsi64_si128(s);
-#else
-    __m128i xmm_i = _mm_cvtsi64_si128(*(GUInt64 *)(ptr));
-#endif
     // Zero extend 4 packed unsigned 16-bit integers in a to packed
     // 32-bit integers.
 #if __SSE4_1__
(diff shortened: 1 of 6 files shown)


rouault:
I'd like to see some performance measurement to confirm that removing those unaligned memory accesses doesn't have measurable impact. 

- ./perftests/testperfcopywords can be used for changes in gcore/
- gdalwarp -r cubic .... for warp related changes
- and ogrinfo -where .... on a FileGDB dataset with a indexed string column (I guess the worst case would be a dataset with all rows having the same value)

dg0yt:
> Let's just use CPL_CPU_REQUIRES_ALIGNED_ACCESS unconditionally. Modern compilers are able to optimize memcpy.

I'm confused: Shoudn't the PR title / commit message use the word `aligned` instead?

rouault:
This breaks Windows builds:
D:\a\gdal\gdal\gcore\gdal_priv_templates.hpp(500,18): error C3861: '_mm_cvtsi128_si64': identifier not found [D:\a\gdal\gdal\build\gcore\gcore.vcxproj]

rouault (commented):
-

rouault (changes requested):
-

MaskRay:
> This breaks Windows builds:
> D:\a\gdal\gdal\gcore\gdal_priv_templates.hpp(500,18): error C3861: '_mm_cvtsi128_si64': identifier not found [D:\a\gdal\gdal\build\gcore\gcore.vcxproj]

Fixed by always using `_mm_storel_epi64(reinterpret_cast<__m128i *>(pDest), xmm);` for `GDALCopyXMMToInt64`.

MaskRay:
> I'm confused: Shoudn't the PR title / commit message use the word `aligned` instead?

* aligned:  p%a==0
* misaligned: p%a!=0
* unaligned: agnostic whether aligned or misaligned. 

The original macro `CPL_CPU_REQUIRES_ALIGNED_ACCESS` is actually confusing. It means that we should use unaligned access, in case some processors trap when a misaligned access is detected.

MaskRay:
> * gdalwarp

`./perftests/testperfcopywords` runs very fast and I cannot find differences w/o and w/ the change.

```
% hyperfine --warmup 2 --min-runs 10 "numactl -C 20 "apps/gdalwarp.{old,new}" ~/projects/gdal/autotest/utilities/data/w_jpeg.tiff out.tiff"
Benchmark 1: numactl -C 20 apps/gdalwarp.old ~/projects/gdal/autotest/utilities/data/w_jpeg.tiff out.tiff
  Time (mean ± σ):      24.8 ms ±   0.2 ms    [User: 16.3 ms, System: 8.4 ms]
  Range (min … max):    24.4 ms …  25.7 ms    111 runs

Benchmark 2: numactl -C 20 apps/gdalwarp.new ~/projects/gdal/autotest/utilities/data/w_jpeg.tiff out.tiff
  Time (mean ± σ):      24.8 ms ±   0.3 ms    [User: 15.6 ms, System: 9.1 ms]
  Range (min … max):    23.4 ms …  25.7 ms    110 runs

Summary
  'numactl -C 20 apps/gdalwarp.old ~/projects/gdal/autotest/utilities/data/w_jpeg.tiff out.tiff' ran
    1.00 ± 0.02 times faster than 'numactl -C 20 apps/gdalwarp.new ~/projects/gdal/autotest/utilities/data/w_jpeg.tiff out.tiff'
```

> and ogrinfo -where .... on a FileGDB dataset with a indexed string column (I guess the worst case would be a dataset with all rows having the same value)

I don't know how to test it...

rouault:
after my own testing, I can't spot any significant differences in performance (if there are, there are within the measurement noise)
Merging

rouault (approved):

PR #8222, opened 2023-08-14
autotest: Add a VSIFile wrapper class

dbaston:
## What does this PR do?

This PR adds a `gdal.vsi_open()` function that returns a file-like object, allowing virtual files to be used as "regular" Python files in some cases.  For example, a zipped CSV can be read using the Python `csv` module:

```
with gdal.vsi_open(f'/vsizip/prime_meridian.csv') as f:
    records = [x for x in csv.DictReader(f)]

assert records[2]['INFORMATION_SOURCE'] == "Institut Geographique National (IGN), Paris"
```

Still need to correctly handle different linebreak/OS combinations, ensure errors are raised when accessing a closed file, handle append mode (or raise error), and probably other things.

## Tasklist

 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/autotest/pymod/gdaltest.py
+++ b/autotest/pymod/gdaltest.py
@@ -2134,7 +2134,6 @@ class VSIFile:
             raise OSError(gdal.VSIGetLastErrorMsg())
 
         self._closed = False
-        self._buffer = None
 
     def __enter__(self):
         return self
--- a/autotest/pymod/gdaltest.py
+++ b/autotest/pymod/gdaltest.py
@@ -2116,3 +2116,80 @@ def reopen(ds, update=False, open_options=None):
         allowed_drivers=[ds_drv.GetDescription()],
         open_options=open_options,
     )
+
+
+# VSIFile helper class
+
+
+class VSIFile:
+    def __init__(self, path, mode, encoding="utf-8"):
+        self._path = path
+        self._mode = mode
+
+        self._binary = "b" in mode
+        self._encoding = encoding
+
+        self._fp = gdal.VSIFOpenExL(self._path, self._mode, True)
+        if self._fp is None:
+            raise OSError(gdal.VSIGetLastErrorMsg())
+
+        self._closed = False
+        self._buffer = None
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, *args):
+        self.close()
+
+    def __iter__(self):
+        return self
+
+    def __next__(self):
+        line = gdal.CPLReadLineL(self._fp)
+        if line is None:
+            raise StopIteration
+        if self._binary:
+            return line.encode()
+        return line
+
+    def close(self):
+        if self._closed:
+            return
+
+        self._closed = True
+        gdal.VSIFCloseL(self._fp)
+
+    def read(self, size=-1):
+        if size == -1:
+            pos = self.tell()
+            self.seek(0, 2)
+            size = self.tell()
+            self.seek(pos)
+
+        raw = gdal.VSIFReadL(1, size, self._fp)
+
+        if self._binary:
+            return bytes(raw)
+        else:
+            return raw.decode(self._encoding)
+
+    def write(self, x):
+
+        if self._binary:
+            assert type(x) in (bytes, bytearray, memoryview)
+        else:
+            assert type(x) is str
+            x = x.encode(self._encoding)
+
+        gdal.VSIFWriteL(x, 1, len(x), self._fp)
+
+    def seek(self, offset, whence=0):
+        gdal.VSIFSeekL(self._fp, offset, whence)
+
+    def tell(self):
+        return gdal.VSIFTellL(self._fp)
+
+
+def vsi_open(path, mode="r"):
+    return VSIFile(path, mode)
--- a/autotest/pymod/gdaltest.py
+++ b/autotest/pymod/gdaltest.py
@@ -2182,10 +2182,17 @@ class VSIFile:
             assert type(x) is str
             x = x.encode(self._encoding)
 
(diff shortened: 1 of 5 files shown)


swig/include/python/gdal_python.i
```diff
@@ -3730,6 +3730,105 @@ def ApplyVerticalShiftGrid(*args, **kwargs):
...
+        return self
+
+    def __exit__(self, *args):
+        self.close()
+
+    def __iter__(self):
+        return self
+
+    def __next__(self):
+        chunk_size = 1024
```
rouault, line 3756: probably that CPLReadLineL() could be used to avoid this reimplementatoin

rcoup:
If it's helpful, here's an implementation we have been using at Koordinates for a while, with a set of tests: feel free to crib anything that's useful. I've licensed it as MIT. Any questions most welcome.

https://gist.github.com/rcoup/5f4b49149ca26084eed4dccfd1b12895

- was written quite a while ago — I would *guess* it could be simpler today incorporating some stdlib updates in recent Python versions and/or other bugfixes.
- Inheriting from `io.RawIOBase` means it can fit under/into the standard python buffered & text IO wrapper classes
- I've removed/replaced/inlined use of a handful of internal library functions & test setup. If there's anything you get stuck on  please ask.

dbaston:
Fantastic, thank you!

dbaston:
I have not forgotten about this. I'm leaning towards moving it to `gdaltest` to focus on internal use for now.

swig/include/python/gdal_python.i
dbaston: done

autotest/pymod/gdaltest.py
```diff
@@ -2116,3 +2116,80 @@ def reopen(ds, update=False, open_options=None):
...
+        if self._binary:
+            assert type(x) in (bytes, bytearray, memoryview)
+        else:
+            assert type(x) is str
+            x = x.encode(self._encoding)
+
+        gdal.VSIFWriteL(x, 1, len(x), self._fp)
+
+    def seek(self, offset, whence=0):
+        gdal.VSIFSeekL(self._fp, offset, whence)
```
rouault, line 2188: we need to be careful here. Due to VSIFSeekL() taking only unsigned offsets,  negative seeking with SEEK_CUR must be emulated doing VSIFSeekL(fp, VSIFTellL(fp) + offset, SEEK_SET)

dbaston: Where should the emulation happen?

- `VSIFile.seek`
- `gdal.VSIFSeekL` (via `%pythonprepend%`)
- `gdal.VSIFSeekL` (via a new wrapper function in `cpl_i`. The signature here already uses signed offsets, so it seems reasonable that it should handle them correctly)

I'm still scratching my head a bit as to why the tests using negative seek work correctly as-is.

rouault: > * `gdal.VSIFSeekL` (via a new wrapper function in `cpl_i`. The signature here already uses signed offsets, so it seems reasonable that it should handle them correctly)

sounds good here. By the way, I've just realized yesterday, that is is not needed to use a "wrapper_foo" + %rename tricks. You can directly create a inline function with the target name. cf https://github.com/OSGeo/gdal/pull/10050/commits/5bcbc5c05486bf02e0bdb212ca4dc2f6670c2188#diff-56ee4d91557b7ba46f850e46abbdc73e617c981c662a47a27d00eaef97136061R972

> I'm still scratching my head a bit as to why the tests using negative seek work correctly as-is.

I suppose that the negative offsets get turned into unsigned values within SWIG .cpp code, and then unsigned overflow on additions within the virtual file system implementations accidentally lead to the correct result. But we'd better not rely on that. In particular the CI sanitize target and our configuration of OSSFuzz errors out on unsigned integer overflows, as they are most of the time bugs.

dbaston: Added code to handle negative values in afe7ca8. I think the `wrapper_` and `%rename` are needed in this case because the function name exposed by SWIG is the same as the underlying function name.

autotest/pymod/gdaltest.py
```diff
@@ -2116,3 +2116,80 @@ def reopen(ds, update=False, open_options=None):
...
+
+    def write(self, x):
+
+        if self._binary:
+            assert type(x) in (bytes, bytearray, memoryview)
+        else:
+            assert type(x) is str
+            x = x.encode(self._encoding)
+
+        gdal.VSIFWriteL(x, 1, len(x), self._fp)
```
rouault, line 2185: we should probably assert the returned value to be len(x) ?

dbaston: That makes sense, and to be consistent, we should check the return code of  `VSIFSeekL` and throw an exception on failure?

rouault: > we should check the return code of `VSIFSeekL` and throw an exception on failure?

sounds good

autotest/pymod/gdaltest.py
```diff
@@ -2116,3 +2116,87 @@ def reopen(ds, update=False, open_options=None):
...
+
+        self._binary = "b" in mode
+        self._encoding = encoding
+
+        self._fp = gdal.VSIFOpenExL(self._path, self._mode, True)
+        if self._fp is None:
+            raise OSError(gdal.VSIGetLastErrorMsg())
+
+        self._closed = False
+        self._buffer = None
```
rouault, line 2137: looks to be unused:
```suggestion
```

PR #7588, opened 2023-04-12
CI: build_mac: fix issue with libkml

rouault:
Recent builds fails with
```
/usr/local/Cellar/libkml/1.3.0_1/include/kml/base/xml_element.h:31:10: fatal error: 'boost/intrusive_ptr.hpp' file not found
```

This is due to this change in the libkml recipee
https://github.com/Homebrew/homebrew-core/pull/128140 which makes boost a build-only dependency.
Explicit installation of boost is now needed to use libkml headers

PR #8226, opened 2023-08-15
Doc: Document common MacOS CMake build issues

tbarsballe:
## Preamble

Long-time GDAL user, first-time (potential) contributor.

I recently updated my local GDAL build from 3.3.2 to 3.7.0, and encountered a number of issues building with CMake on MacOS. I discovered basically all of these issues had fixes or workarounds in the [cmake_builds.yml](https://github.com/OSGeo/gdal/blob/master/.github/workflows/cmake_builds.yml#L552) CI config, but these workarounds didn't seem to be documented anywhere. After applying these, I was able to get a successful build.

## What does this PR do?

This PR aims to add a section with information about common CMake build failures on MacOS, and their workarounds, to the documentation. For consistency, I've also moved the preexisting Windows guideline to a similar section. 

 - https://github.com/OSGeo/gdal/pull/7588
 - [**8891395**: CI: fix build-mac](https://github.com/OSGeo/gdal/commit/889139563a4869cbfea212621f25a81c33cc3a92)
 - [**369f0c3**: CI: cmake_builds.yml: build-mac: disable Arrow/Parquet](https://github.com/OSGeo/gdal/commit/369f0c3da15e0a72efd66496e161684c0063dca2)

## Tasklist

 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS: MacOS 13.5
* Compiler: AppleClang 14.0.3.14030022

### DIFF
--- a/doc/source/development/building_from_source.rst
+++ b/doc/source/development/building_from_source.rst
@@ -47,12 +47,6 @@ From the build directory you can now configure CMake, build and install the bina
     and is thus not recommended. It is also not supported on Windows multi-configuration
     generator (such as VisualStudio).
 
-`
-On Windows, one may need to specify generator:
-
-.. code-block:: bash
-
-    cmake -G "Visual Studio 15 2017" ..
 
 If a dependency is installed in a custom location, specify the
 paths to the include directory and the library:
@@ -89,6 +83,41 @@ for the shared lib, *e.g.* ``set (GDAL_LIB_OUTPUT_NAME gdal_x64 CACHE STRING "" 
     you may try removing CMakeCache.txt to start from a clean state.
 
 Refer to :ref:`using_gdal_in_cmake` for how to use GDAL in a CMake project.
+
+Building on Windows
++++++++++++++++++++
+
+On Windows, one may need to specify generator:
+
+.. code-block:: bash
+
+    cmake -G "Visual Studio 15 2017" ..
+
+
+Building on MacOS
++++++++++++++++++
+
+On MacOS, there are a couple of libraries that do not function properly when the GDAL build requirements are installed using Homebrew.
+
+The `Apache Arrow <https://arrow.apache.org/docs/index.html>`_ library included in the current distribution of CMake is broken, and causes a detection issue. In order to build GDAL successfuly, you must first remove it:
+
+.. code-block:: bash
+
+    rm -rf /usr/local/lib/cmake/Arrow
+
+And then set the appropriate options to exclude Arrow from the build:
+
+.. code-block:: bash
+
+    cmake -DGDAL_USE_ARROW=OFF -DGDAL_USE_PARQUET=OFF ..
+
+
+Similarly, recent versions of Homebrew no longer bundle `Boost <https://www.boost.org/>`_ with libkml, causing a failure to find Boost headers. You should either install Boost manually or disable libkml when building on MacOS:
+
+.. code-block:: bash
+
+    cmake -DGDAL_USE_LIBKML=OFF ..
+
 
 CMake general configure options
 +++++++++++++++++++++++++++++++


doc/source/development/building_from_source.rst
```diff
@@ -90,6 +84,41 @@ for the shared lib, *e.g.* ``set (GDAL_LIB_OUTPUT_NAME gdal_x64 CACHE STRING ""
...
+
+.. code-block:: bash
+
+    rm -rf /usr/local/lib/cmake/Arrow
+
+And then set the appropriate options to exclude Arrow from the build:
+
+.. code-block:: bash
+
+    cmake -DGDAL_USE_ARROW=OFF -DGDAL_USE_PARQUET=OFF ..
```
dg0yt, line 112: Instead of removing arrow cmake config files, it should be enough to use `-DCMAKE_DISABLE_FIND_PACKAGE_Arrow=ON`.
Related analysis: https://github.com/microsoft/vcpkg/issues/33193

OTOH arrow in vcpkg should work also for macOS.

doc/source/development/building_from_source.rst
```diff
@@ -90,6 +84,41 @@ for the shared lib, *e.g.* ``set (GDAL_LIB_OUTPUT_NAME gdal_x64 CACHE STRING ""
...
+
+
+Similarly, recent versions of Homebrew no longer bundle `Boost <https://www.boost.org/>`_ with libkml, causing a failure to find Boost headers. You should either install Boost manually or disable libkml when building on MacOS:
+
+.. code-block:: bash
+
+    cmake -DGDAL_USE_LIBKML=OFF ..
```
dg0yt, line 119: Similar: Disable find_package to work around the issue. And proabably not broken in vcpkg.

dg0yt (changes requested):
Maybe gdal should generally avoid `find_package` for disabled dependencies.
In vcpkg, users recently stumbled over brew libspatialite 5.1.0 breaking gdal despite libspatialite being disabled, https://github.com/microsoft/vcpkg/issues/33039.

rouault:
> Maybe gdal should generally avoid `find_package` for disabled dependencies.

The issue is that we can't tell the difference between a user having explicitly disabled a dependency, and a previous run of CMake having failed to detect the dependency and thus setting the GDAL_USE_xxxx variable to NO. In the latter case, we'd want that if the user installs the dependency afterwards and run CMake again to see the status message "XXXX has been found, but is disabled by default. Enable it by setting GDAL_USE_XXXX=ON"

doc/source/development/building_from_source.rst
tbarsballe: I can confirm this works - after restoring my Arrow config files (and confirming I once again got the original find_package error without the flag), I can successfully build while the Arrow config is present if that flag is included. Thanks for the suggestion, it's much cleaner.

I've updated this section of the docs accordingly.

doc/source/development/building_from_source.rst
rouault: I don't have time/the environment to investigate that, but indeed this is weird. I'm not sure why the build recipee of GDAL in homebrew would work with boost + libkml installed, and not manual GDAL builds. Or maybe because the boost headers are in a non standard location that their build script adds an include for ?? Maybe the authors of the libkml boost change in homebrew could bring some light on this

dg0yt: > See discusion here: [Homebrew/homebrew-core#128083](https://github.com/Homebrew/homebrew-core/pull/128083)
> It looks like Boost is being intentionally excluded.

Hm, AFAIU they were concerned about the size of (their!) boost package and decided to focus on runtime usage.

GDAL could update its `FindLibKML.cmake` to ensure boost is available.

> Though from other discussions on [this PR](https://github.com/OSGeo/gdal/pull/7588), it looks like Boost fails even when installed manually, so there may be something else going on. Whether that's an issue with brew, boost, or gdal, I'm unsure.

https://github.com/OSGeo/gdal/pull/7588#issuecomment-1681626135. Not a gdal issue.

dg0yt:
I realize that this is a downside of cmake vs. pkg-config: Broken CMake code from optional external files can break the build, broken. optional pc files can't.

> In the latter case, we'd want that if the user installs the dependency afterwards and run CMake again to see the status message "XXXX has been found, but is disabled by default. Enable it by setting GDAL_USE_XXXX=ON"

Even when configuring with `GDAL_USE_EXTERNAL_LIBS=OFF`? 

I wonder if the tiny change in one status message is really noticed in the noise of other messages (which may have the same "found, but disabled" pattern, for other development packages already installed on the users system.

rouault:
> In the latter case, we'd want that if the user installs the dependency afterwards and run CMake again to see the status message "XXXX has been found, but is disabled by default. Enable it by setting GDAL_USE_XXXX=ON"

sorry I was wrong. This message is only displayed for the very few dependencies that are not enabled by default, even when detected (actually just external shapelib now)

> Even when configuring with GDAL_USE_EXTERNAL_LIBS=OFF?

When GDAL_USE_EXTERNAL_LIBS=OFF, and a component is found but GDAL_USE_XXXX is not true, then there's this message: "${key} has been found, but is disabled due to GDAL_USE_EXTERNAL_LIBS=OFF. Enable it by setting GDAL_USE_${key}=ON". I believe the rationale was to help users to do minimal builds with just the components they need

doc/source/development/building_from_source.rst
```diff
@@ -90,6 +84,35 @@ for the shared lib, *e.g.* ``set (GDAL_LIB_OUTPUT_NAME gdal_x64 CACHE STRING ""
...
+
+    cmake -G "Visual Studio 15 2017" ..
+
+
+Building on MacOS
++++++++++++++++++
+
+On MacOS, there are a couple of libraries that do not function properly when the GDAL build requirements are installed using Homebrew.
+
+The `Apache Arrow <https://arrow.apache.org/docs/index.html>`_ library included in the current distribution of CMake is broken, and causes a detection issue. In order to build GDAL successfuly, configure CMake to not find the Arrow package:
```
dg0yt, line 102: Which Apache Arrow library is included in the current distribution of CMake? Was this meant to refer to Homebrew instead of CMake?

rouault: > Was this meant to refer to Homebrew instead of CMake?

surely, fixed

PR #8231, opened 2023-08-16
Python bindings: Use OGRFeature::DumpReadableAsString for __repr__

dbaston:
## What does this PR do?

Make the Python `Feature.__repr__` use `DumpReadableAsString`. This is convenient for interactive use and also lets pytest generate assertion failure messages, like

```
>       assert feat1 == feat2
E       assert OGRFeature(rfc35_test):1\n  intfield (Integer) = (null)\n  foo5 (String) = foo0\n  bar10 (String) = (null)\n  baz15 (String) = (null)\n  baw20 (String) = (null)\n\n == OGRFeature(rfc35_test):2\n  intfield (Integer) = (null)\n  foo5 (String) = foo1\n  bar10 (String) = bar1\n  baz15 (String) = (null)\n  baw20 (String) = (null)\n\n
E         Full diff:
E         - OGRFeature(rfc35_test):2
E         ?                        ^
E         + OGRFeature(rfc35_test):1
E         ?                        ^
E             intfield (Integer) = (null)
E         -   foo5 (String) = foo1
E         ?                      ^
E         +   foo5 (String) = foo0
E         ?                      ^
E         -   bar10 (String) = bar1
E         ?                    ^^^^
E         +   bar10 (String) = (null)
E         ?                    ^^^^^^
E             baz15 (String) = (null)
E             baw20 (String) = (null)

/home/dan/dev/build-gdal-Desktop-Debug/autotest/ogr/ogr_rfc35.py:363: AssertionError
```

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:

### DIFF
--- a/ogr/ogrfeature.cpp
+++ b/ogr/ogrfeature.cpp
@@ -5676,6 +5676,40 @@ void OGR_F_DumpReadable(OGRFeatureH hFeat, FILE *fpOut)
 }
 
 /************************************************************************/
+/*                         OGR_F_DumpReadableAsString()                 */
+/************************************************************************/
+
+/**
+ * \brief Dump this feature in a human readable form.
+ *
+ * This dumps the attributes, and geometry; however, it doesn't include
+ * definition information (other than field types and names), nor does
+ * it report the geometry spatial reference system.
+ *
+ * A few options can be defined to change the default dump :
+ * <ul>
+ * <li>DISPLAY_FIELDS=NO : to hide the dump of the attributes</li>
+ * <li>DISPLAY_STYLE=NO : to hide the dump of the style string</li>
+ * <li>DISPLAY_GEOMETRY=NO : to hide the dump of the geometry</li>
+ * <li>DISPLAY_GEOMETRY=SUMMARY : to get only a summary of the geometry</li>
+ * </ul>
+ *
+ * @param hFeat handle to the feature to dump.
+ * @param papszOptions NULL terminated list of options (may be NULL)
+ * @return a string with the feature representation (to be freed with CPLFree())
+ * @since GDAL 3.8
+ */
+
+char *OGR_F_DumpReadableAsString(OGRFeatureH hFeat, CSLConstList papszOptions)
+{
+    VALIDATE_POINTER1(hFeat, "OGR_F_DumpReadableAsString", nullptr);
+
+    return CPLStrdup(OGRFeature::FromHandle(hFeat)
+                         ->DumpReadableAsString(papszOptions)
+                         .c_str());
+}
+
+/************************************************************************/
 /*                               GetFID()                               */
 /************************************************************************/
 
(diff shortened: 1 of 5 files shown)


ogr/ogrfeature.cpp
```diff
@@ -5675,6 +5675,39 @@ void OGR_F_DumpReadable(OGRFeatureH hFeat, FILE *fpOut)
...
+ * A few options can be defined to change the default dump :
+ * <ul>
+ * <li>DISPLAY_FIELDS=NO : to hide the dump of the attributes</li>
+ * <li>DISPLAY_STYLE=NO : to hide the dump of the style string</li>
+ * <li>DISPLAY_GEOMETRY=NO : to hide the dump of the geometry</li>
+ * <li>DISPLAY_GEOMETRY=SUMMARY : to get only a summary of the geometry</li>
+ * </ul>
+ *
+ * @param papszOptions NULL terminated list of options (may be NULL)
+ * @return a string with the feature representation.
```
rouault, line 5698:
```suggestion
 * @return a string with the feature representation (to be freed with CPLFree())
```

ogr/ogrfeature.cpp
```diff
@@ -5675,6 +5675,39 @@ void OGR_F_DumpReadable(OGRFeatureH hFeat, FILE *fpOut)
...
+ * @since GDAL 3.8
+ */
+
+char *OGR_F_DumpReadableAsString(OGRFeatureH hFeat, CSLConstList papszOptions)
+{
+    VALIDATE_POINTER1(hFeat, "OGR_F_DumpReadable", nullptr);
```
rouault, line 5704:
```suggestion
    VALIDATE_POINTER1(hFeat, "OGR_F_DumpReadableAsString", nullptr);
```

swig/include/ogr.i
```diff
@@ -1983,6 +1983,10 @@ public:
     OGR_F_DumpReadable(self, NULL);
   }
 
+  char* DumpReadableAsString() {
```
rouault, line 1986:
```suggestion
  retStringAndCPLFree* DumpReadableAsString() {
```

swig/include/python/ogr_python.i
```diff
@@ -743,6 +743,11 @@ def ReleaseResultSet(self, sql_lyr):
 
         return self.GetGeometryRef()
 
+
+    def __repr__(self):
```
rouault, line 747: maybe we need some tiny testing of this in autotest/ogr/ogr_feature.py ?

dbaston: I actually don't see any tests for `DumpReadableAsString`, so I'll plan to expose its "options" argument in SWIG and add some tests for that as well as `__repr__` ?

rouault: > I actually don't see any tests for `DumpReadableAsString`

some of it might already be tested through test_ogrinfo.py with the --fields and --geom switches

> so I'll plan to expose its "options" argument in SWIG and add some tests for that as well as __repr__ ?

sounds good (I had only in mind minimal testing)

PR #8235, opened 2023-08-17
PDF: update to support (and require) PDFium/5952

rouault:


pathmapper:
Thanks for the PR which was also great hint for me :-)

Just built this branch successfully using also https://github.com/rouault/pdfium_build_gdal_3_8

`gdalinfo --format PDF` returns

image

PR #8238, opened 2023-08-18
Remove AWS from GDAL Sponsorship program

rouault:


geographika:
That was a scary pull request title!

hobu:
It begs the question – why does GDAL support a special file system like `/vsis3` or `/vsigs`? We don't have specific Lustre support even though a lot of GDAL's users drive big processing tasks with it. In fact, except for HDFS, the only special file system drivers that are represented for distributed types are the big clouds – Google, AWS, and S3. 

The code paths through GDAL to support these file systems is not trivial. It puts behavior and capability limitations on drivers in a non-uniform way, and it's special code to maintain as these different cloud vendors increment their APIs and deprecate their security approaches over and over again. Why do we give preferential treatment to them? Can't GDAL's users simply use FUSE?

pl-nfwarmerdam:
I do not think FUSE satisfies the same role and that these file system handlers can be a much more user friendly approach without installing extra system services, and using folks "natural" security mechanism (ie. dot files for instance).  That said, they aren't a core part of GDAL and should primarily exist when there are user/developers using them and willing to support them.  I had previously mostly attempted to address these use cases using signed urls to objects but this doesn't work well for "directory like" file sets.

mojodna:
[Mountpoint for Amazon S3](https://aws.amazon.com/blogs/storage/the-inside-story-on-mountpoint-for-amazon-s3-a-high-performance-open-source-file-client/) ([docs](https://docs.aws.amazon.com/AmazonS3/latest/userguide/mountpoint.html)) is FUSE-esque and might be equivalent for S3 (I don't know of anyone benchmarking it yet, but all indications are that it's better than other FUSE implementations out there, which can burn a lot of CPU and be pretty slow). Mountpoint has some optimizations under the hood (including use of the AWS common runtime (CRT) for highly-optimized S3 access) and looks like a filesystem, provided you don't try to do anything weird with it.

How tuned are the filesystem reads in GDAL for concurrency + ranges (the stuff where the HTTP implementation has been highly tuned)? I'm curious how sensitive they are to latency on syscalls that would be near-instantaneous on a local disk.

msmitherdc:
Um, its not FUSE-esque, it is FUSE. It uses FUSE via RUST

nyalldawson:
A thought: why not make an environment variable which must be set before the S3 etc support can be used. (A bit like when we deprecate drivers). If the environment variable isn't set, gdal will just dump an obnoxious explanatory error which states something like "this functionality is subject to removal as it is not financially supported by Amazon. If you rely on this functionality please contact your Amazon sales rep and let them know about its pending removal".

hobu:
> Is GDAL really considering removing AWS support?

Like a practical PR that excises `/vsis3` from the entire tree? Nope. Not at this time. If we did, who would you scream at? The GDAL project maintainers and developers or AWS for only participating in the community in an extractive fashion? 

Entities like AWS are the toll bridges that extract significant value from software like GDAL, but not only does the project not share in the financial reward of doing so, it generates NEGATIVE costs of maintaining the software going forward.  

This thread is more like a consideration of the fairness of the project supporting S3 relative to all the other distributed file system types, with a recognition that there is a significant maintenance, education, and complexity cost to doing so.

tomkom:
Quick question, only found out about this from @hobu 's tweet. We are using a S3 *compatible* store for COGs and COPCs, running on the OS CEPH project (university storage). Many others do. Will this impact? What are our options?

rouault:
Closing, with just removing them as sponsors. Let's hope we'll be able to revert that commit at some later point...

PR #8243, opened 2023-08-20
Doc: gdalwarp: clarify that -t_srs might still be needed when using -ct (fixes #8242)

### LINKED ISSUE #8242: gdalwarp with proj pipeline does not write CRS to header

Running this command:

```bash
gdalwarp -s_srs EPSG:32603 -t_srs EPSG:6393 source_epsg32603.tif dest_epsg6393.tif
```

And this command:

```bash
gdalwarp -ct "+proj=pipeline +step +inv +proj=utm +zone=3 +ellps=WGS84 +step +proj=aea +lat_0=50 +lon_0=-154 +lat_1=55 +lat_2=65 +x_0=0 +y_0=0 +ellps=GRS80" source_epsg32603.tif dest_epsg6393.tif
```

(with the proj pipeline passed to `-ct` gathered from this command:
```bash
projinfo -s EPSG:32603 -t EPSG:6393 -o PROJ --hide-ballpark --spatial-test intersects
```
)

Should produce the same results.

They do produce the same GeoTiff; but in the second command with the `ct` pipeline; the CRS info is not written to the `dest_*.tif` header. 

Am I misusing the `-ct` flag? Or should these equivalent operations produce the same output, including the correct CRS info in the `dest_*.tif` header when I run a `gdalinfo`?

## Operating system

macOS 13.4.1

GDAL 3.6.4, released 2023/04/17 installed via `brew install gdal`

rouault:


bpurinton:
@rouault That does work, I tested it. But I guess now I'm confused, because I thought the `ct` proj pipeline is an alternative to using the `t_srs`; or, is it the case that the `ct` option would allow a different pipeline (if `projinfo` showed a few options), overriding the `t_srs` option?

I actually tested this by adding a slight y-offset in the proj pipeline (while still including the `t_srs`). Indeed, it seems that the `ct` options did override the standard `t_srs` usage. 

In that case, perhaps my suggested change to the documentation is useful?

doc/source/programs/gdalwarp.rst
```diff
@@ -147,6 +147,8 @@ with control information.
     or a urn:ogc:def:coordinateOperation:EPSG::XXXX URN overriding the default
     transformation from the source to the target CRS. It must take into account the
     axis order of the source and target CRS.
+    When creating a new output file, using :option:`-t_srs` is still necessary
+    to have the target CRS written in the metadata of the output file.
```
bpurinton, line 151:
```suggestion
    to have the target CRS written in the metadata of the output file, but the parameters of the CoordinateOperation will override those of the standard transformation.
```

rouault: suggestion adopted

bpurinton:
LGTM, thanks for the prompt help @rouault!

PR #8246, opened 2023-08-21
NITF: add support for CSCSDB (Common Sensor Covariance Support Data) DES from GLAS/GFM SDEs

rouault:


rouault:
manually backported to 3.7

Issue #8260, opened 2023-08-23
gdal_retile.py fails when using some input with gaps.


mlinds:
## Expected behavior and actual behavior.
I expected to be able to use `gdal_retile` to retile and create pyramids of several input raster files, even if there are gaps in the input rasters. I often work with coastal data, so it is common for me to tile large piles of rasters overlapping the coastline. 

However, if none of the resulting new tiles overlap the first index shape, the script will crash and not continue building pyramids. This is because the script looks at the very first tile of the index (with name like `{rastername}_01_01.tif`).

### Illustrative example
When I try to use it on several DEMs of the netherlands, the layer 0 tiling works fine. However, at the next step it starts to build the pyramids and it raises a RuntimeError.

This is my layer 0 of retiled data, as sucessfully generated by `gdal_retile.py`

![image](https://github.com/OSGeo/gdal/assets/45837856/92873d0d-c7fd-4657-b664-3d1e0fe76d61)

Now, its time to create the pyramids. To set some values for metadata about the mosaic tile, the script looks for the tile in the upper left side, as highlighted here:

![image](https://github.com/OSGeo/gdal/assets/45837856/49146ad2-db0f-43a9-8b61-08daf10eca00)

There is no data in that index, and therefore no corresponding .tiff file. GDAL is set to use exceptions, so it raises one about not being able to find that file.

The traceback looks like:
```
Traceback (most recent call last):
  File "/home/mlinds/miniconda3/envs/default_geo_env/bin/gdal_retile.py", line 14, in <module>
    sys.exit(main(sys.argv))
             ^^^^^^^^^^^^^^
  File "/home/mlinds/miniconda3/envs/default_geo_env/lib/python3.11/site-packages/osgeo_utils/gdal_retile.py", line 1081, in main
    buildPyramid(g, minfo, dsCreatedTileIndex, g.TileWidth, g.TileHeight, g.Overlap)
  File "/home/mlinds/miniconda3/envs/default_geo_env/lib/python3.11/site-packages/osgeo_utils/gdal_retile.py", line 762, in buildPyramid
    levelMosaicInfo = mosaic_info(minfo.filename, inputDS)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mlinds/miniconda3/envs/default_geo_env/lib/python3.11/site-packages/osgeo_utils/gdal_retile.py", line 150, in __init__
    fhInputTile = self.cache.get(imgLocation)
[...]

PR #8269, opened 2023-08-24
Gdal retile non contiguous

mlinds:
## What does this PR do?
Updates `gdal_retile.py` so that it can now handle gaps between input rasters larger than the size of a single tile. Right now, if there are gaps in the input raster, there can be tiles in the tile index which don't actually have a corresponding raster file. Right now this causes an exception.

Also added some comments and rudimentary docstrings to some internal functions, just based on what I found helpful.

This is my first commit to gdal and I would appreciate some feedback. Especially regarding:

- So far I modified the default behavior of the script. Would it be better to add a flag like `--allow-gaps` ?
- I only did a few basic asserts for my test case; let me know if there need to be more

So far I have not squashed my commits, but I figured I'd do that after my PR is reviewed.

I submitted issue #8260 

## Tasklist

(Skipping documentation and updating Python API docs)

 - [x] Modify `mosaic_info` object `__init__()` method to set metadata based on the first valid raster tile it can find, instead of always the upper left raster file
 - [x] Modify  `mosaic_info.getDataSet()` method to skip missing tiles 
 - [x] Add test case
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment
Tested against the `ghcr.io/osgeo/gdal-deps:ubuntu20.04-master` image, only ran the pytest tests but they all pass. 

(likely not relevant since I have not touched any compiled code)
* OS: Ubuntu 20.04.6 LTS
* Compiler:

### DIFF
--- a/swig/python/gdal-utils/osgeo_utils/gdal_retile.py
+++ b/swig/python/gdal-utils/osgeo_utils/gdal_retile.py
@@ -524,7 +524,7 @@ def createPyramidTile(
         ]
     )
     # if -resume flag and the tile is present, add it to the index and exit function
-    if OGRDS is not None and feature_only:
+    if feature_only:
         points = dec.pointsFor(width, height)
         addFeature(g.TileIndexFieldName, OGRDS, tileName, points[0], points[1])
         return
@@ -645,7 +645,7 @@ def createTile(
     ]
 
     # if -resume flag and the tile is present, add it to the index and exit function
-    if OGRDS is not None and feature_only:
+    if feature_only:
         dec2 = AffineTransformDecorator(geotransform)
         points = dec2.pointsFor(width, height)
         addFeature(g.TileIndexFieldName, OGRDS, tilename, points[0], points[1])


rouault:
@mlinds Thanks for this PR (and the added comments in the code). I'm not super familiar with that utility, but it seems to me that the root cause could be that createPyramidTile() calls addFeature() even when there is no tile to generate.

Wouldn't doing the following change do the job and avoid your other changes ?
```
    if OGRDS is not None:
        points = dec.pointsFor(width, height)
        addFeature(g.TileIndexFieldName, OGRDS, tileName, points[0], points[1])

    if feature_only:
        return

    s_fh = levelMosaicInfo.getDataSet(
        dec.ulx, dec.uly + height * dec.scaleY, dec.ulx + width * dec.scaleX, dec.uly
    )
    if s_fh is None:
        return
```
to
```
    s_fh = levelMosaicInfo.getDataSet(
        dec.ulx, dec.uly + height * dec.scaleY, dec.ulx + width * dec.scaleX, dec.uly
    )
    if s_fh is None:
        return

    if OGRDS is not None:
        points = dec.pointsFor(width, height)
        addFeature(g.TileIndexFieldName, OGRDS, tileName, points[0], points[1])

    if feature_only:
        return
```

mlinds:
Thanks for the quick review @rouault!

Good catch, that would be a cleaner way of doing it. And a side effect would be that one is saving the tile index output, it would also not have the needless tiles, which I think is probably good behavior for most use cases. I will update it tomorrow if I have some time.

mlinds:
I had another look, and was able to implement the approach you suggested @rouault. This is definitely a cleaner way to do it. 

I had to change it in 2 places: the `createTile()` function that is used to create layer 0 tiles, and the `createPyramidTile()` function that builds the higher levels.

Now, if the `-resume` flag is used and the tile exists, it is added to the index and the function returns. Thereafter, the tile index is only added if there is valid data in that tile (i.e. the `.getDataSet()` method returns a dataset instead of `None`).

swig/python/gdal-utils/osgeo_utils/gdal_retile.py
```diff
@@ -611,12 +644,11 @@ def createTile(
         dec.scaleY,
     ]
 
-    if OGRDS is not None:
+    # if -resume flag and the tile is present, add it to the index and exit function
+    if OGRDS is not None and feature_only:
```
rouault, line 648: As far as I can see OGRDS cannot be None:

```suggestion
    if feature_only:
```

swig/python/gdal-utils/osgeo_utils/gdal_retile.py
```diff
@@ -505,19 +523,21 @@ def createPyramidTile(
             sy,
         ]
     )
-
-    if OGRDS is not None:
+    # if -resume flag and the tile is present, add it to the index and exit function
+    if OGRDS is not None and feature_only:
```
rouault, line 527: As far as I can see OGRDS cannot be None:
```suggestion
    if feature_only:
```

mlinds:
Hi @rouault, yes that is a good suggestion, I don't think `OGRDS` could end up being `None`. Thanks for making that change.

Should I go ahead and rebase it all together (including upstream commits to make the CI checks happy)?

rouault:
> Should I go ahead and rebase it all together (including upstream commits to make the CI checks happy)?

no, that's fine. I've just used the squash&merge github functionality

PR #8273, opened 2023-08-24
Python bindings: make Feature.SetField(field_idx_or_name, binary_values)

rouault:
add feature.SetFieldBinary(field_idx_or_name, binary_values)


swig/include/python/ogr_python.i
```diff
@@ -677,6 +677,28 @@ def ReleaseResultSet(self, sql_lyr):
             # For Python3 on non-UTF8 strings
             return self.GetFieldAsBinary(fld_index)
 
+    def SetFieldBinary(self, field_index_or_name, value):
+        """
+        SetFieldBinary(Feature self, field_index_or_name: int | str, value: bytes)
+
+        Set field to binary data.
+        This function currently on has an effect of OFTBinary fields.
```
dbaston, line 685:
```suggestion
        This function currently only has an effect on OFTBinary fields.
```

rouault: done

PR #8283, opened 2023-08-28
 Zarr: add MULTIBAND=YES/NO, DIM_X and DIM_Y open options (fixes #8237) 

### LINKED ISSUE #8237: gdal_translate zarr to multi-band tif
After finding an [example in the docs]((https://gdal.org/drivers/raster/zarr.html#:~:text=array%20to%20GeoTIFF%3A-,gdal_translate,-%27ZARR%3A%22my.zarr)), I found that multi-band zarr to tif is not working i.e. `gdal_translate data.zarr data.tif`.
For now the python alternative works fine, but a cli option for this is defo a nice to have.

creating a 3d zarr file from a few tifs
```
import rioxarray as rxr
import xarray as xr

s1 = rxr.open_rasterio("s1.tif", chunks=True)
s2 = rxr.open_rasterio("s2.tif", chunks=True)
ds = xr.concat([s1, s2], dim="band")
ds.to_zarr("data.zarr")
```

converting the zarr to a multi-band tif
```
ds_zarr = xr.open_zarr("data.zarr").__xarray_dataarray_variable__
ds_zarr.rio.to_raster("data.tif")
```

To handle the flexibility of zarr data, the gdal_translate could require the coordinates [x, y, band].

rouault:
and default to MULTIBAND=YES

### DIFF
--- a/frmts/zarr/zarrdriver.cpp
+++ b/frmts/zarr/zarrdriver.cpp
@@ -553,8 +553,10 @@ GDALDataset *ZarrDataset::Open(GDALOpenInfo *poOpenInfo)
 
     if (poMainArray && (bMultiband || poMainArray->GetDimensionCount() <= 2))
     {
-        auto poNewDS = std::unique_ptr<GDALDataset>(
-            poMainArray->AsClassicDataset(iXDim, iYDim));
+        // Pass papszOpenOptions for LOAD_EXTRA_DIM_METADATA_DELAY
+        auto poNewDS =
+            std::unique_ptr<GDALDataset>(poMainArray->AsClassicDataset(
+                iXDim, iYDim, poOpenInfo->papszOpenOptions));
         if (!poNewDS)
             return nullptr;
 
@@ -1581,6 +1583,10 @@ void GDALRegister_Zarr()
         "'Name or index of the X dimension (only used when MULTIBAND=YES)'/>"
         "   <Option name='DIM_Y' type='string' description="
         "'Name or index of the Y dimension (only used when MULTIBAND=YES)'/>"
+        "   <Option name='LOAD_EXTRA_DIM_METADATA_DELAY' type='string' "
+        "description="
+        "'Maximum delay in seconds allowed to set the DIM_{dimname}_VALUE band "
+        "metadata items'/>"
         "</OpenOptionList>");
 
     poDriver->SetMetadataItem(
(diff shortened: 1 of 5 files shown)


frmts/zarr/zarrdriver.cpp
```diff
@@ -1443,6 +1573,14 @@ void GDALRegister_Zarr()
         "   <Option name='CACHE_TILE_PRESENCE' type='boolean' "
         "description='Whether to establish an initial listing of present "
         "tiles' default='NO'/>"
+        "   <Option name='MULTIBAND' type='boolean' default='NO' "
+        "description='Whether to expose >= 3D arrays as GDAL multiband "
```
jratike80, line 1577: Should it be "default='YES'" like in the zarr.rst above?

rouault: good point . fixed

mdsumner:
thanks Even!  

fwiw initial test with my go-to Zarr-url is very slow

```
gdalinfo --config AWS_NO_SIGN_REQUEST YES ZARR:"/vsis3/mur-sst/zarr-v1/":/analysed_sst -oo MULTIBAND=YES

gdal_translate --config AWS_NO_SIGN_REQUEST YES ZARR:"/vsis3/mur-sst/zarr-v1/":/analysed_sst -oo MULTIBAND=YES -b 1 out.tif
```

I appreciate it could just be slow because my compute is not in the right S3 tenancy. 

(setting MULTIBAND=NO precludes that syntax as expected)

with a local source  it's fast and behaves as expected (MULTIBAND and the DIM_X/DIM_Y), I'm hoping to put more time into this to explore the various options

rouault:
> I appreciate it could just be slow because my compute is not in the right S3 tenancy.

the reason is that the ``time`` variable has shape = [6443] but chunks=[5]. Consequently, when instantiating the GDAL dataset and assign to each band a DIM_time_VALUE metadata item, we need to read 6443 / 5 chunk data files of the time variable...

mdsumner:
ah I see, thank you

rouault:
> the reason is that the `time` variable has shape = [6443] but chunks=[5]. Consequently, when instantiating the GDAL dataset and assign to each band a DIM_time_VALUE metadata item, we need to read 6443 / 5 chunk data files of the time variable...

I've just added a LOAD_EXTRA_DIM_METADATA_DELAY option to offer a configurable maximum delay, that defaults to 5 sec

mdsumner:
excellent set of fixes, this is awesome

PR #8299, opened 2023-09-01
OGRCurve::isClockwise(): fix wrong result when lowest rightmost vertex is the first one (fixes #8296)

### LINKED ISSUE #8296: OpenGDB driver issue exporting CurvePolygons with holes
## Steps to reproduce the problem

- Import the following geometry inside a PostGIS layer (schema=public, table name=test, EPSG 3857):

CurvePolygon (CompoundCurve ((1328 540, 1340 573),CircularString (1340 573, 1345 564, 1358 553),(1358 553, 1328 540)),CompoundCurve ((1334 547, 1341 551, 1338 557, 1334 547)))

- Export the layer in GDB format with the command:

ogr2ogr -f OpenFileGDB test.gdb PG:"host=localhost user=... password=... dbname=test schemas=public" -overwrite

The original geometry is valid (CurvePolygon with one hole) while the exported geometry is not valid (CurvePolygon whitout the hole and a nested shell).

## Operating system - GDAL version
Windows 10  - GDAL 3.7.1
Ubuntu 22.04 - GDAL 3.8.0dev-c032069be4-dirty

rouault:


rouault:
Backported to 3.7 per 362297dc28

PR #8301, opened 2023-09-01
SQLite dialect: when the underlying layer has a FID column name, enable the user to use it as an alias of ROWID

rouault:


aborruso:
Thank you very much!

PR #8312, opened 2023-09-02
OGRGeometry classes: add addGeometry()/addRing()/addCurve() methods accepting a std::unique_ptr

rouault:
and use them in gml2ogrgeometry.cpp

### DIFF
--- a/autotest/ogr/ogr_gml_geom.py
+++ b/autotest/ogr/ogr_gml_geom.py
@@ -1711,6 +1711,10 @@ def test_gml_invalid_geoms():
             None,
         ),
         ("<gml:TriangulatedSurface></gml:TriangulatedSurface>", None),
+        (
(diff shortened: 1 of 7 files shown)


cmoth150415:
when the return type of this file's function is switched from a plain pointer to std::unique_ptr, compiling on CentOS 7.9 with gcc version 4.8.5 results in an error regarding the correct usage of rvalue references with std::unique_ptr. It is advised that the corresponding return value should be returned using std::move.
Additionally, is it the tag that 3.8.4 can no longer be compiled with older versions of compilers like gcc 4.8.5?

rouault:
>  It is advised that the corresponding return value should be returned using std::move.

newer compilesr actually warn against using those std::move in return statements...

> is it the tag that 3.8.4 can no longer be compiled with older versions of compilers like gcc 4.8.5?

possibly. Such ancient compilers are no longer tested. GCC 4.8 has a few known non compliances with C++11, and only 4.9 is fully C++11 compliant. I believe they are more up to date build chains available for CentOS

PR #8336, opened 2023-09-05
Don't use jpeg if disabled

autoantwort:
Fixes https://github.com/microsoft/vcpkg/issues/33565

When `-DGDAL_USE_JPEG=OFF` is used and gdal is found, the target `JPEG::JPEG` exists. Based on this some checks are executed and HAVE_JPEGTURBO_DUAL_MODE_8_12 is set to `ON`.  
Then the following if gets true and gdal tries to use jpeg: 
https://github.com/OSGeo/gdal/blob/d12af24fa7496d7bf2aa850453230278dd9070e4/frmts/mrf/CMakeLists.txt#L35-L38

Since we don't want to use jpeg anyway, we can simply skip the checks (`check_c_source_compiles`).

### DIFF
--- a/cmake/helpers/CheckDependentLibraries.cmake
+++ b/cmake/helpers/CheckDependentLibraries.cmake
@@ -416,7 +416,7 @@ if (GDAL_USE_JPEG AND (JPEG_LIBRARY MATCHES ".*turbojpeg\.(so|lib)"))
       "JPEG_LIBRARY should point to a library with libjpeg ABI, not TurboJPEG. See https://libjpeg-turbo.org/About/TurboJPEG for the difference"
     )
 endif ()
-if (TARGET JPEG::JPEG)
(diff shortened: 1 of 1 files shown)


autoantwort:
The ci error seems unrelated to this PR

rouault:
> The ci error seems unrelated to this PR

yes, merging. Thanks

dg0yt:
If the problem is `HAVE_JPEGTURBO_DUAL_MODE_8_12` being set to `ON`, you must also take care of unsetting its cached value when `GDAL_USE_JPEG` is off.

rouault:
> If the problem is `HAVE_JPEGTURBO_DUAL_MODE_8_12` being set to `ON`, you must also take care of unsetting its cached value when `GDAL_USE_JPEG` is off.

Probably a "better" (at least more aligned with our practice) fix is : https://github.com/OSGeo/gdal/pull/8338

Issue #7477, opened 2023-03-18
VRT connection "vrt://<dsn>?opt" syntax progress/discussion


mdsumner:
The VRT connection 'vrt://<dsn>?opt1=<args>&opt2=<args>' syntax was introduced in 3.1 with 
option `bands`, and in play for 3.7 we currently have

`a_srs`, `a_ullr`, `ovr`, `expand`, `a_scale`, `a_offset`, `ot`, `gcp`, and `if` 

These generally match the `gdal_translate` syntax. The syntax varies, for example rather than "-b" used
multiple times there is a single `?bands=1,2,3` for the syntax, and for `a_ullr` the values are also
comma separated instead of by spaces as in the command line. 

I've listed the remaining args here, to classify them in terms of utility, and progress, as a list for myself to refer back to and possibly to trigger interest from others. 

Please let me know what you think, there's a few I'd like to still get in for the 3.7 release and I'll be focussing on the ones I think I can manage. 

When I started adding these I thought that many of these didn't make sense, but now I think they do. If you disagree or otherwise have comments I'd like to hear them. 

Thanks! 

## doesn't have utility 

- `of` , `stats`

## in progress

- `scale`,  `scale_bn`, `exponent`

## could still be done

- `strict`
- `outsize`
- `tr`
- `r`
- `unscale`
- `srcwin`   (I really want this)
- `projwin`  (I really want this)
- `a_coord_epoch`
- `a_nodata`
- `nogcp`
- `colorinterp_bn`
- `colorinterp`
- `mo`
- `q` (I'm not sure this is useful)
- `sds` (I really want this)
- `co`
- `norat`
- `noxmp`
- `oo`

rouault:
> * `sds` (I really want this)

There's no way this can work with the vrt:// syntax. -sds is a special mode for the gdal_translate binary

mdsumner:
damn, I really wanted that 😃 - appreciate the explanation of why it can't work (I tried) 🙏

rouault:
> damn, I really wanted that smiley - appreciate the explanation of why it can't work (I tried) pray

wasn't that obvious it can't work ? -sds makes a materialized copy as a geotiff file of every subdataset of the source dataset. That's totally at odds with the VRT approach that exposes a single non-materialized dataset

mdsumner:
it's not obvious 😅 I had other reasons why I figured it couldn't work

that's heinous, I need to think about what you just told me 😱

mdsumner:
[...]

PR #8343, opened 2023-09-06
VRT 'scale'  'unscale' for vrt:// connection

mdsumner:
Add 'scale' (without arguments) and  'unscale' to allowed options for a "vrt://" connection.

## Tasklist

 - [x] Add test case(s)
 - [x] Add documentation
 - [ ] Review
 - [ ] Adjust for comments


rouault:
Perhaps use scale=true / unscale=true to be more consistent with query-parameter style arguments ?

mdsumner:
ok, yes  thanks - I'll play around with it for a bit because it implies handling of the existing values for 'scale='.  Because it's different to gdal_translate itself I want to see how it feels across the remaining args too. 

should it be "truthy" in the YES/True/true/1 kind of way? Or just exactly "true" or "false", or "true" and anything else is false?   These are the kinds of questions that threw me a bit from the beginning!

rouault:
> should it be "truthy" in the YES/True/true/1 kind of way? Or just exactly "true" or "false", or "true" and anything else is false? These are the kinds of questions that threw me a bit from the beginning! Mostly it's been able to rely on value-sanitizing by the utility itself.

as we can know that it is a boolean value if there is only one token in the value, you could use CPLTestBool():
```

/**
 * Test what boolean value contained in the string.
 *
 * If pszValue is "NO", "FALSE", "OFF" or "0" will be returned false.
 * Otherwise, true will be returned.
 *
 * @param pszValue the string should be tested.
 *
 * @return true or false.
 */

bool CPLTestBool(const char *pszValue)
{
    return !(EQUAL(pszValue, "NO") || EQUAL(pszValue, "FALSE") ||
             EQUAL(pszValue, "OFF") || EQUAL(pszValue, "0"));
}
```

PR #8359, opened 2023-09-07
autotest: fix recent errors introduced in GML tests

dbaston:
## What does this PR do?

1) restores loss of `.xsd` testing that I caused with 4faf097. When copying `.gml` files into `tmp_path` to avoid side-effects from opening them in `data`, I missed some `xsd` files.
2) adds explicit testing that `.gfs` files are ignored when they are older than the corresponding `.gml` files. This was implicitly tested somewhere prior to 4faf097, but I added an explicit test for it.

I have a reasonable procedure in place to catch these kinds of regressions in the future. Maybe it makes sense to add it to the coverage build.

#8333

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/autotest/ogr/ogr_gml_read.py
+++ b/autotest/ogr/ogr_gml_read.py
@@ -32,6 +32,7 @@
 
 import os
 import shutil
+import sys
 
 import gdaltest
 import ogrtest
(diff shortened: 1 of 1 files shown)


dbaston:
> can you give a bit more detail on how it works?

Essentially just a script to run the test suite twice under two different git commits, producing an `.info` file for each commit, and then using https://github.com/capgelka/lcov-diff to subtract lines that are common to both `.info` files files, producing a `diff.info` file. If you go on to produce an HTML report from the `diff.info` file, any lines/branches shown as covered represent lines/branches that _lost_ coverage with the test suite changes.

It is working well if there are only changes to the test files between the two commits, but `lcov-diff` seems to error out if there are C++ functions that are defined in one commit but not the other. I haven't tried to resolve that.

```
#!/usr/bin/env bash

set -euxo pipefail

GDAL_ROOT=/home/dan/dev/gdal

TEST_BRANCH=1e9850c
BASE_BRANCH=20131c09c6449d6375d17f4e7c87d59764526ffb

COVERAGE_SCOPE=.

for BRANCH in ${BASE_BRANCH} ${TEST_BRANCH} ; do

    git -C ${GDAL_ROOT} checkout ${BRANCH}
    ninja

    touch fake.gcda # prevent pipe fail
    find ./ -name "*.gcda" | xargs rm

    cd autotest
    pytest || true
    cd ..

    lcov --directory ${COVERAGE_SCOPE} --capture --output-file gdal.info 2>/dev/null   
    cp gdal.info gdal_filtered.info
    lcov --remove gdal_filtered.info '/usr/*' --output-file gdal_filtered.info
    lcov --remove gdal_filtered.info '/opt/*' --output-file gdal_filtered.info
    lcov --remove gdal_filtered.info '*jpeg/libjpeg*' --output-file gdal_filtered.info
    lcov --remove gdal_filtered.info '*pcraster/libcsf*' --output-file gdal_filtered.info
    lcov --remove gdal_filtered.info '*googletest-src*' --output-file gdal_filtered.info

    mv gdal_filtered.info gdal_filtered_${BRANCH}.info
done

mkdir -p /tmp/gdal-coverage
/home/dan/.cargo/bin/lcov-diff gdal_filtered_${BASE_BRANCH}.info gdal_filtered_${TEST_BRANCH}.info -o /tmp/gdal-coverage/diff.info
rm -rf /tmp/gdal-coverage/diff_html
genhtml -o /tmp/gdal-coverage/diff_html --ignore-errors source --num-spaces 2 /tmp/gdal-coverage/diff.info >/dev/null
```

rouault:
> It is working well if there are only changes to the test files between the two commits, but `lcov-diff` seems to error out if there are C++ functions that are defined in one commit but not the other. I haven't tried to resolve that (but would need to do so if it is to be integrated into CI)

I was thinking you might perhaps just build libgdal & the bindings with the latest commit,  but run the autotest suite (at least the python part) with the latest and reference commit. That should work > 99.9 % of the cases since our changes in the C API and ABI and Python interface are mostly backward compatible (but there would still be < 0.1% of the cases where that would break, but we can probably live with that...) 
At least that would make sense only/mostly for commits related to test suite changes, and as the output would require downloading an artifact with the .zip HTML, perhaps that could be triggered only on demand through a "[diff_coverage]" label in the (last) commit message?

PR #8362, opened 2023-09-07
AAIGRID output: remove leading space on each line (fixes #8344)

### LINKED ISSUE #8344: Extra white space on first line of each converted element for ASC file
Reported here as I suspect it is an upstream possible issue rather than a QGIS bug

Using QGIS convert the attached file
[PNOA_MDT200_ETRS89_HU30_Madrid.zip](https://github.com/OSGeo/gdal/files/12536590/PNOA_MDT200_ETRS89_HU30_Madrid.zip)

Raster > Conversion > Translate > Save to file as ASC
I guess the above steps can be done via the command line directly with GDAL but have no clue on how to do that

Extra white space on first line of each converted element, check line 7  / 8 
image
image

This makes the file not readable in software not expecting this white space which apparently is the norm or maybe I am mistaken but it would be great if there was an option/switch to produce an ASC version without the whitespace rather than removing it afterwards by some other mean

## Operating system

Windows 11

GDAL 3.8.0dev-4b68c95c09 used in QGIS

rouault:
Esri spec shows an example without leading space:
https://desktop.arcgis.com/en/arcmap/10.3/manage-data/raster-and-images/esri-ascii-raster-format.htm .

Let's align on that


rouault:
> Did you take into account the space before each linefeed in the ESRI style files?

no, hopefully that doesn't matter

rouault:
> Linefeeds are not required in the specification but the whole data section can be just a single row with space separated values. I think that the idea is to remove all linefeeds and carriage returns and get such a string as a result.

ah ok, missed that. Actually that helps making the code simpler (2 line shorter :-)). Doing that then

jratike80:
> > Linefeeds are not required in the specification but the whole data section can be just a single row with space separated values. I think that the idea is to remove all linefeeds and carriage returns and get such a string as a result.
> 
> ah ok, missed that. Actually that helps making the code simpler (2 line shorter :-)). Doing that then

I believe that it is safer that way even it seems that ESRI is not so strict. I made a test by removing the ending space from an .asc file that was created by ArcGIS Pro, and it could still be opened with ArcGIS Pro.

PR #8368, opened 2023-09-08
PLScenes testing: add '?fortest=true' suffix to avoid having both a /vsimem/foo file and /vsimem/foo/ directory

rouault:


rouault:
closing as pulled in https://github.com/OSGeo/gdal/pull/8369

PR #8375, opened 2023-09-10
VRT 'a_coord_epoch' for vrt:// connection

mdsumner:
Add `a_coord_epoch` to allowed options for a "vrt://" connection.

Does this need a pytest requirement for the proj data files? 

## Tasklist

 - [x] Add test case(s)
 - [x] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/autotest/gcore/vrt_read.py
+++ b/autotest/gcore/vrt_read.py
@@ -1627,7 +1627,7 @@ def test_vrt_protocol():
 
 @pytest.mark.require_proj(7, 2)
 def test_vrt_protocol_a_coord_epoch_option():
-    ds = gdal.Open("vrt://data/byte.tif?a_coord_epoch=2021.3")
+    ds = gdal.Open("vrt://data/byte.tif?a_srs=EPSG:4326&a_coord_epoch=2021.3")
     srs = ds.GetSpatialRef()
     assert srs.IsDynamic()
     assert srs.GetCoordinateEpoch() == 2021.3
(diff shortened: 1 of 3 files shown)


autotest/gcore/vrt_read.py
```diff
@@ -1625,6 +1625,14 @@ def test_vrt_protocol():
         assert not gdal.Open("vrt://data/minfloat.tif?a_ullr=0,1,1,0&unscale&")
 
 
+@pytest.mark.require_proj(7, 2)
+def test_vrt_protocol_a_coord_epoch_option():
+    ds = gdal.Open("vrt://data/byte.tif?a_coord_epoch=2021.3")
+    srs = ds.GetSpatialRef()
+    assert srs.IsDynamic()
```
rouault, line 1632: byte.tif uses a projected CRS over NAD27 which is not a dynamic CRS. You should probably also add a_srs=EPSG:4326

mdsumner: does that affect the test though?  (humour me 😃)

rouault: well, this is the reason why it fails:
```
25: >       assert srs.IsDynamic()
25: E       AssertionError: assert False
```

mdsumner: oh, thanks yes learning but still not looking enough 🙏

PR #8387, opened 2023-09-12
MYSQL: Fix passwords with comma, fixes #8385

### LINKED ISSUE #8385: MYSQL: cannot connect to MySQL if user's password contains commas.
See downstream https://github.com/qgis/QGIS/issues/54493

elpaso:


port/cpl_string.cpp
```diff
@@ -894,6 +894,13 @@ char **CSLTokenizeString2(const char *pszString, const char *pszDelimiters,
                 }
 
                 bInString = !bInString;
+
+                if (!bInString)
+                {
+                    ++pszString;
+                    break;
+                }
```
rouault, line 902: why is this necessary? The existing logic to break when a delimiter should be sufficient. Here this breaks regular parsing of quoted CSV fields since this results in a empty token to be emitted when the delimiter is found. I guess that's the reason why in your test you needed to add the CSLT_STRIPLEADSPACES | CSLT_STRIPENDSPACES flags

Looking at ogrmysqldatasource.cpp what I can see however (but probably not related to the original issue) is that there is logic when open options are used to specify the connection parameter to build a traditional connection string, and it lacks the logic to quote values, and do quote-escaping if quotes are in the parameter value.

elpaso: > why is this necessary? 

Because when ending a quoted string (i.e. when unescaped `"` is found the token is ready and there is no need to look for the separator.

I have added a new test case, does it make sense to support this format?

- a string containing a `"` must escape the `"` and be double quoted
- a string containing a `,` must be double quoted

```c++
    {
        CPLStringList aosStringList(CSLTokenizeString2(
            R"(one=one,two_quoted="two\"quoted\"rest",three_quote="quote\"rest",four_quoted_comma="four,\"quoted\"")",
            ",", CSLT_HONOURSTRINGS));
        ASSERT_EQ(aosStringList.size(), 4);
        ASSERT_TRUE(EQUAL(aosStringList[0], "one=one"));
        ASSERT_TRUE(EQUAL(aosStringList[1], R"(two_quoted=two"quoted"rest)"));
        ASSERT_TRUE(EQUAL(aosStringList[2], R"(three_quote=quote"rest)"));
        ASSERT_TRUE(
            EQUAL(aosStringList[3], R"(four_quoted_comma=four,"quoted")"));
    }
```

> The existing logic to break when a delimiter should be sufficient. Here this breaks regular parsing of quoted CSV fields since this results in a empty token to be emitted when the delimiter is found. I guess that's the reason why in your test you needed to add the CSLT_STRIPLEADSPACES | CSLT_STRIPENDSPACES flags
> 

No, that was to test the single token prefixed by a space.

I see that it breaks CSV, I'll look into that.

> Looking at ogrmysqldatasource.cpp what I can see however (but probably not related to the original issue) is that there is logic to build a traditional connection string, when open options are used to specify the connection parameters, and it lacks the logic to quote values, and do quote-escaping if quotes are in the parameter value.

I maybe doing something wrong but I'm testing a MySQL user `qgis"user` with pwd `qgis,"rocks"`  and I've found no way to get `ogrinfo` work with any combination of quotes and escaped chars.

Do you agree that we have an issue here?

rouault: > > why is this necessary?
> 
> Because when ending a quoted string (i.e. when unescaped `"` is found the token is ready and there is no need to look for the separator.

I'm probably missing something as the code is a bit tricky, but with my understanding, it should behave correctly already for what you wanted to do (it should break just at the next iteration of the loop when it encounters the delimiter), and adding a perhaps unnecessary new branch doesn't seem the way to go . But I'll let you dig further as you have a fresher view than mine on this code.

> I've found no way to get ogrinfo work with any combination of quotes and escaped chars.

may it not be just a shell usage issue with all those quoting/escaping trouble ? Did you try through the python API with gdal.VectorInfo() for example ?

elpaso: > > I've found no way to get ogrinfo work with any combination of quotes and escaped chars.

Anyway, you were right and this works on GDAL master, I guess I was not escaping the arguments correctly.

> 
> may it not be just a shell usage issue with all those quoting/escaping trouble ? Did you try through the python API with gdal.VectorInfo() for example ?

I've stumbled upon shell quoting/escaping at first but then I started passing single quoted arguments and that bypassed the issue.

PR #8389, opened 2023-09-13
gdal_translate -ovr: properly rescale RPC (fixes #8386)

### LINKED ISSUE #8386: gdal_translate with -ovr produces incorrect RPCs
It would seem like the RPC recalculation is not done (correctly) when using gdal_translate's `-ovr` switch. i.e. the RPCs of the created file are identical to those of the input files.

reproducer:
`gdal_translate -ovr 2 tif_with_rpcs.tif output.tif`
=> output.tif has incorrect RPCs

rouault:


rouault:
manually backported to 3.7 per 9924e92de62c0176e77189a5307787550159478f

PR #8390, opened 2023-09-13
python-drivers: Support WKB geometries

rcoup:
## What does this PR do?

1. Add support for passing ISO WKB geometries from Python OGR drivers by using a bytes-like object (bytes, bytesarray, memoryview); as well as the existing WKT strings.
2. Expand the dummy python driver in tests to take options, and use them to test the geometry formats.
3. Update the documentation and examples; tidy up some formatting.

## Tasklist

 - [x] Add test case(s)
 - [x] Add documentation
 - Updated Python API documentation (swig/include/python/docs/)
 - [x] Review
 - [x] Adjust for comments

## Environment

Imports a couple more functions from the Python Stable ABI, but nothing introduced recently.
Tested on macOS with Python3.11.

### DIFF
--- a/gcore/gdalpythondriverloader.cpp
+++ b/gcore/gdalpythondriverloader.cpp
@@ -773,6 +773,8 @@ OGRFeature *PythonPluginLayer::TranslateToOGRFeature(PyObject *poObj)
     PyObject *myLongType = PyObject_Type(myLong);
     PyObject *myFloat = PyFloat_FromDouble(1.0);
     PyObject *myFloatType = PyObject_Type(myFloat);
+    PyObject *myStr = PyUnicode_FromString("");
+    PyObject *myStrType = PyObject_Type(myStr);
 
     auto poFields = PyDict_GetItemString(poObj, "fields");
     auto poGeometryFields = PyDict_GetItemString(poObj, "geometry_fields");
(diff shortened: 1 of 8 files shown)


rouault:
Interesting to see this OGR Python driver mechanism actually used

rouault:
@rcoup Some formatting issues. You need to setup pre-commit. See https://gdal.org/development/dev_practices.html#commit-hooks

gcore/gdalpythondriverloader.cpp
```diff
@@ -773,6 +773,8 @@ OGRFeature *PythonPluginLayer::TranslateToOGRFeature(PyObject *poObj)
     PyObject *myLongType = PyObject_Type(myLong);
     PyObject *myFloat = PyFloat_FromDouble(1.0);
     PyObject *myFloatType = PyObject_Type(myFloat);
+    PyObject *myStr = PyUnicode_FromString("");
+    PyObject *myStrType = PyObject_Type(myStr);
```
rouault, line 777: both should be Py_DecRef()'ed at the end of the method

rcoup:
> Interesting to see this OGR Python driver mechanism actually used

Just some swiss-army-knife application for data format conversion. Followed the tutorial, worked well.

Only thing I did run into:
```
$ ogrinfo -so -al mydata.file
GDAL: Found /opt/homebrew/bin/python3
GDAL: Which is an alias to: ../Cellar/python@3.11/3.11.5/bin/python3
GDAL: Python version from binary name: 3
GDAL: Trying libpython3.dylib
GDAL: Trying libpython3m.dylib
GDAL: Trying libpython3.8.dylib
GDAL: Trying libpython3.9.dylib
GDAL: Trying libpython3.10.dylib
GDAL: Trying libpython3.11.dylib
GDAL: Trying libpython3.12.dylib
GDAL: Trying libpython3.7m.dylib
GDAL: Trying libpython3.6m.dylib
GDAL: Trying libpython3.5m.dylib
GDAL: Trying libpython3.4m.dylib
GDAL: Trying libpython3.3.dylib
GDAL: Trying libpython3.2.dylib
ERROR 1: Cannot find python/libpython. You can set the PYTHONSO configuration option to point to the a python .so/.dll/.dylib
```
Setting `$PYTHONSO` to the actual Python binary on macOS works fine.  ala `PYTHONSO=/Library/Frameworks/Python.framework/Versions/3.10/Python` or `PYTHONSO=/opt/homebrew/bin/python3` or `PYTHONSO=/path/to/my/venv/bin/python3`

gcore/gdalpythondriverloader.cpp
rcoup: Good spot. I'd have much preferred to use:
- `PyUnicode_Check()` & friends, but they didn't load. Not sure if that's because they're not in the Stable ABI, or maybe related to my comment not using true libpython?
- `PyObject_IsInstance(o, PyUnicode_Type)` but that requires getting the `PyTypeObject` opaque struct to be defined, which was a bit beyond my Wednesday afternoon 😁

rcoup: Ah. At least some of the `Py*_Check()` functions are macos. That'd explain it :-)

PR #8399, opened 2023-09-15
Add RFC95 text: Use standard C/C++ integer types

rouault:
Partially rendered view [here](https://github.com/rouault/gdal/blob/rfc95_text/doc/source/development/rfc/rfc95_standard_int_types.rst)


doc/source/development/rfc/rfc95_standard_int_types.rst
```diff
@@ -0,0 +1,216 @@
...
+-----------------------------------
+
+Q1: What to do with the 64-bit integer changes?
+
+A1: Potential alternatives:
+
+1) Proceed with them, keep GDAL 3.8 as version number. External code will have
+   to use #ifdef if support for GDAL < 3.8 and >= 3.8 is needed.
+
+2) Same as 1), but bump GDAL version number to 4.0. Cleaner way
```
snowman2, line 158: I think doing this as a 4.0 change makes the most sense since it is a breaking change requiring changes in upstream codebases.

hyoklee (approved):

doc/source/development/rfc/rfc95_standard_int_types.rst
sgillies: @snowman2 @rouault I mentioned implications for Rasterio and Fiona on gdal-dev and will repeat myself here. I'm not sure what the necessary changes would be for them... they're Cython, not C, and the Cython project feels like conditional compilation directives was a mistake 😂 

It's not clear to me that we can do just this https://github.com/PDAL/PDAL/pull/4179/files#diff-22b766759d40394aa74c390f2ff295b1737020f1a824e103238b60efa9e2814fR36 like for PDAL.

Introducing stubs to support type changes would be a drag, hopefully that could be avoided.

doc/source/development/rfc/rfc95_standard_int_types.rst
```diff
@@ -0,0 +1,183 @@
...
+
+=============================================================
+RFC 95: Use standard C/C++ integer types
+=============================================================
+
+============== =============================================
+Author:        Even Rouault
+Contact:       even.rouault @ spatialys.com
+Started:       2023-Sep-15
+Status:        Proposed
```
rouault, line 11:
```suggestion
Status:        Proposed, but *not* adopted
```

doc/source/development/rfc/rfc95_standard_int_types.rst
```diff
@@ -0,0 +1,183 @@
+.. _rfc-95:
+
+=============================================================
+RFC 95: Use standard C/C++ integer types
```
rouault, line 4:
```suggestion
RFC 95: Use standard C/C++ integer types (proposed, *NOT* adopted)
```

doc/source/development/rfc/rfc95_standard_int_types.rst
```diff
@@ -0,0 +1,183 @@
+.. _rfc-95:
+
+=============================================================
+RFC 95: Use standard C/C++ integer types (proposed, *NOT* adopted)
+=============================================================
```
rouault, line 5:
```suggestion
==================================================================
RFC 95: Use standard C/C++ integer types (proposed, *NOT* adopted)
==================================================================
```

PR #8405, opened 2023-09-16
VRT 'eco' 'epo' for vrt:// connection

mdsumner:
Add `epo` and `eco` to allowed options for a "vrt://" connection.

- [x] Add test case(s)
 - [x] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/doc/source/drivers/raster/vrt.rst
+++ b/doc/source/drivers/raster/vrt.rst
@@ -1769,16 +1769,17 @@ the order 'gt(0),gt(1),gt(2),gt(3),gt(4),gt(5)'.
 The effect of the ``oo`` option (added in GDAL 3.8) is to set driver-specific dataset open options, multiple values are allowed. The value
 consists of string key value pairs with multiple pairs separated by commas e.g. ``oo=<key>=<val>`` or . ``oo=<key1>=<val1>,<key2>=<val2>,...``. This is applied in the same way as (:ref:`gdal_translate`).
 
-The effect of the ``unscale`` option (added in GDAL 3.8) is to apply the scale/offset metadata for the bands to convert scaled values to unscaled values. Do apply this use syntax ``unscale=true``, or ``unscale=false`` which is the default behaviour if not specified. Do consider the need for also using ``ot`` option in order to accomodate the intended output range, see more details for the same argument as with (:ref:`gdal_translate`).
+The effect of the ``unscale`` option (added in GDAL 3.8) is to apply the scale/offset metadata for the bands to convert scaled values to unscaled values. To apply this use syntax ``unscale=true``, or ``unscale=false`` (which is the default if not specified). Do consider the need for also using ``ot`` option in order to accomodate the intended output range, see more details for the same argument as with (:ref:`gdal_translate`).
 
 The effect of the ``a_coord_epoch`` option (added in GDAL 3.8) is to assign a coordinate epoch, linked with the output SRS as
 with (:ref:`gdal_translate`).
 
-The effect of the ``nogcp`` option (added in GDAL 3.8) is to not copy the GCPs in the source dataset to the output dataset (:ref:`gdal_translate`).
+The effect of the ``nogcp`` option (added in GDAL 3.8) is to not copy the GCPs in the source dataset to the output dataset (:ref:`gdal_translate`). To apply this
+use syntax ``nogcp=true``, or ``nogcp=false`` (which is the default if not specified).
 
-The effect of the ``epo`` option (added in GDAL 3.8) is that ``srcwin`` or ``projwin`` values that fall partially outside the source raster extent will be considered as an error as per (:ref:`gdal_translate`).
+The effect of the ``epo`` option (added in GDAL 3.8) is that ``srcwin`` or ``projwin`` values that fall partially outside the source raster extent will be considered as an error as per (:ref:`gdal_translate`). To apply this use syntax ``epo=true``, or ``epo=false`` (which is the default if not specified).
 
-The effect of the ``eco`` option (added in GDAL 3.8) is that ``srcwin`` or ``projwin`` values that fall completely outside the source raster extent will be considered as an error as per (:ref:`gdal_translate`).
+The effect of the ``eco`` option (added in GDAL 3.8) is that ``srcwin`` or ``projwin`` values that fall completely outside the source raster extent will be considered as an error as per (:ref:`gdal_translate`). To apply this use syntax ``eco=true``, or ``eco=false`` (which is the default if not specified).
 
 The options may be chained together separated by '&'. (Beware the need for quoting to protect
 the ampersand).
--- a/doc/source/drivers/raster/vrt.rst
+++ b/doc/source/drivers/raster/vrt.rst
@@ -1697,7 +1697,7 @@ For example:
 
 The supported options currently are ``bands``, ``a_srs``, ``a_ullr``, ``ovr``, ``expand``,
 ``a_scale``, ``a_offset``, ``ot``, ``gcp``, ``if``, ``scale``, ``exponent``, ``outsize``, ``projwin``,
-``projwin_srs``, ``tr``, ``r``, ``srcwin``, ``a_gt``, ``oo``, ``unscale``, ``a_coord_epoch``, and ``nogcp``.
+``projwin_srs``, ``tr``, ``r``, ``srcwin``, ``a_gt``, ``oo``, ``unscale``, ``a_coord_epoch``, ``nogcp``, ``epo``, and ``eco``.
 
 Other options may be added in the future.
 
@@ -1775,6 +1775,10 @@ The effect of the ``a_coord_epoch`` option (added in GDAL 3.8) is to assign a co
 with (:ref:`gdal_translate`).
 
 The effect of the ``nogcp`` option (added in GDAL 3.8) is to not copy the GCPs in the source dataset to the output dataset (:ref:`gdal_translate`).
+
+The effect of the ``epo`` option (added in GDAL 3.8) is that ``srcwin`` or ``projwin`` values that fall partially outside the source raster extent will be considered as an error as per (:ref:`gdal_translate`).
+
+The effect of the ``eco`` option (added in GDAL 3.8) is that ``srcwin`` or ``projwin`` values that fall completely outside the source raster extent will be considered as an error as per (:ref:`gdal_translate`).
 
 The options may be chained together separated by '&'. (Beware the need for quoting to protect
 the ampersand).
--- a/autotest/gcore/vrt_read.py
+++ b/autotest/gcore/vrt_read.py
@@ -1629,6 +1629,13 @@ def test_vrt_protocol():
     ds = gdal.Open("vrt://data/gcps.vrt?nogcp=true")
     assert ds.GetGCPCount() == 0
 
+    assert gdal.Open("vrt://data/byte.tif?srcwin=0,0,20,20&eco=true&epo=true")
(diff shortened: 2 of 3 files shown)


doc/source/drivers/raster/vrt.rst
```diff
@@ -1776,6 +1776,10 @@ with (:ref:`gdal_translate`).
 
 The effect of the ``nogcp`` option (added in GDAL 3.8) is to not copy the GCPs in the source dataset to the output dataset (:ref:`gdal_translate`).
 
+The effect of the ``epo`` option (added in GDAL 3.8) is that ``srcwin`` or ``projwin`` values that fall partially outside the source raster extent will be considered as an error as per (:ref:`gdal_translate`).
+
+The effect of the ``eco`` option (added in GDAL 3.8) is that ``srcwin`` or ``projwin`` values that fall completely outside the source raster extent will be considered as an error as per (:ref:`gdal_translate`).
+
```
rouault, line 1782: it is not obvious reading that (and this applies to nogcp as well) that the ``true`` value must be provided

mdsumner: thank you, fixed now.   I'm intending to do a review of the whole set for consistency and grouping, it's probably good to do that soon.

doc/source/drivers/raster/vrt.rst
```diff
@@ -1769,12 +1769,17 @@ the order 'gt(0),gt(1),gt(2),gt(3),gt(4),gt(5)'.
...
+The effect of the ``unscale`` option (added in GDAL 3.8) is to apply the scale/offset metadata for the bands to convert scaled values to unscaled values. To apply this use syntax ``unscale=true``, or ``unscale=false`` (which is the default if not specified). Do consider the need for also using ``ot`` option in order to accomodate the intended output range, see more details for the same argument as with (:ref:`gdal_translate`).
 
 The effect of the ``a_coord_epoch`` option (added in GDAL 3.8) is to assign a coordinate epoch, linked with the output SRS as
 with (:ref:`gdal_translate`).
 
-The effect of the ``nogcp`` option (added in GDAL 3.8) is to not copy the GCPs in the source dataset to the output dataset (:ref:`gdal_translate`).
+The effect of the ``nogcp`` option (added in GDAL 3.8) is to not copy the GCPs in the source dataset to the output dataset (:ref:`gdal_translate`). To apply this
+use syntax ``nogcp=true``, or ``nogcp=false`` (which is the default if not specified).
```
rouault, line 1778: I'm wondering if it wouldn't be more natural to change the logic of this option from nogcp to gcp, but that can be for a later pull request

mdsumner: I don't like changing them, though `bands` established the scene 😃 and it always worried me having commas in place of spaces, and bundling multiple instances into comma lists ...

What about other contexts, /vrtvec:// and /vrtwarp:// etc?  I wonder if the python API is a better model for option names?

It seems a good time for a review given the sds api and the broader dsn syntax discussion

rouault: > What about other contexts, /vrtvec:// and /vrtwarp:// etc? I wonder if the python API is a better model for option names?
> 
@sgillies is not a big fan of vrt:// https://lists.osgeo.org/pipermail/gdal-dev/2023-September/057646.html

mdsumner: I saw

mdsumner: 🙏

well I'm keen to see alternatives that work as well

sgillies: I said that "vrt://" was a mistake because it goes against the syntax best practices in https://datatracker.ietf.org/doc/html/rfc4395#section-2.2. Particularly regarding the `//` and GDAL-specific paths. Currently, a thing like `vrt:///viscurl/https://example.com/foo.tif` is okay with GDAL, right? But this would be hard to parse as a URL. It's not a good URL. We shouldn't expand this practice in GDAL. That's my opinion.

I'm not quibbling about the parameter encoding, which seems fine. And yes, I appreciate that a `vrt://` connection string will be shorter than its equivalent XML. I think both of these features are worth keeping.

Essentially, `vrt://` means run `gdal_translate -of VRT ...` behind the scene, right? So, why not this?

`/vsitranslate/https://example.com/foo.tif?nogcp=true&...`

We already have `/vsicurl`, which means send the rest of the connection string through curl. `/vsitranslate` would mean send the rest of the connection string through gdal_translate.

In a nutshell, I'm saying that I think it's better and clearer to extend GDAL's `/vsi` space than to invent more special URI schemes.

rouault: > `/vsitranslate` would mean send the rest of the connection string (after some parsing) through gdal_translate.

/vsi is smething you pass to VSIFOpenL() to return a VSILFILE*
vrt:// is something you pass to GDALOpen() to return a GDALDataset*
Too different kind of objects

mdsumner: excellent comments @sgillies I appreciate the details, I agree it's an odd syntax but I don't otherwise have anything to add and I'm pretty deep in becoming reliant on it now. I am worried about what *might* variants look like, because I now want this for other utilites,  especially for `ogr2ogr` and `gdalwarp`, but perhaps that becomes another arg `app=<applib name>` to require all args are "compatible with target app lib". 


`vrt://{dsn}?simplify=1&app=vectranslate` 

`vrt://{dsn}?t_srs=EPSG:3031&ts=512,0&app=warper` 

that would be pretty awesome

tbonfort: how does this play when using /vsicurl options, e.g. for `vrt:///vsicurl/https://server/object?unsafessl=yes&t_srs=epsg:4326` ?

mdsumner: it works really well, in fact you can really go overboard: 

```
 gdalinfo "vrt://NETCDF:\"/vsitar//vsicurl/https://cran.r-project.org/src/contrib/tidync_0.3.0.tar.gz/tidync/inst/extdata/oceandata/S20080012008031.L3m_MO_CHL_chlor_a_9km.nc\":chlor_a?a_srs=OGC:CRS84" -nomd

```

```
Driver: VRT/Virtual Raster
Files: NETCDF:"/vsitar//vsicurl/https://cran.r-project.org/src/contrib/tidync_0.3.0.tar.gz/tidync/inst/extdata/oceandata/S20080012008031.L3m_MO_CHL_chlor_a_9km.nc":chlor_a
Size is 4320, 2160
Coordinate System is:
GEOGCRS["WGS 84 (CRS84)",
    DATUM["World Geodetic System 1984",
        ELLIPSOID["WGS 84",6378137,298.257223563,
            LENGTHUNIT["metre",1]]],
    PRIMEM["Greenwich",0,
    CS[ellipsoidal,2],
        AXIS["geodetic longitude (Lon)",east,
            ORDER[1],
        AXIS["geodetic latitude (Lat)",north,
            ORDER[2],
    USAGE[
        SCOPE["unknown"],
        AREA["World"],
        BBOX[-90,-180,90,180]],
    ID["OGC","CRS84"]]
Origin = (-179.999994916092277,90.000002544309424)
Pixel Size = (0.083333338043927,-0.083333335689175)
Corner Coordinates:
Upper Left  (-179.9999949,  90.0000025) (179d59'59.98"W, 90d 0' 0.01"N)
Lower Left  (-179.9999949, -90.0000025) (179d59'59.98"W, 90d 0' 0.01"S)
Upper Right ( 180.0000254,  90.0000025) (180d 0' 0.09"E, 90d 0' 0.01"N)
Lower Right ( 180.0000254, -90.0000025) (180d 0' 0.09"E, 90d 0' 0.01"S)
Center      (   0.0000153,  -0.0000000) (  0d 0' 0.05"E,  0d 0' 0.00"S)
Band 1 Block=44x22 Type=Float32, ColorInterp=Undefined
  NoData Value=-32767
  Unit Type: mg m^-3

```

rouault: > how does this play when using /vsicurl options, e.g. for `vrt:///vsicurl/https://server/object?unsafessl=yes&t_srs=epsg:4326` ?

badly AFAICS.  I guess we'd need to allow a syntax like ``vrt://{/vsicurl/https://server/object?unsafessl=yes}?t_srs=epsg:4326``, similarly to what is done for /vsizip/ when your zip doesn't have a .zip extension, like ``/vsizip/{/path/to/myarchive.random_extension}/path/within/archive``
Or say that anything between `vrt://` and `?` is the base filename and needs to be URL encoded if the filename itself contains `?`

tbonfort: sorry for hijacking this issue, ths should really be a followup to @sgillies message on gdal-dev. I'm specifically curious about the example I mentionned, mixing vsicurl's `unsafessl` option with vrt://'s `t_srs` option

rouault: > want this for other utilites, especially for `ogr2ogr` and `gdalwarp`, but perhaps that becomes another arg `app=<applib name>` to require all args are "compatible with target app lib".

you might have some luck with gdalwarp since most options are compatible with a WarpedVRT. But no luck for ogr2ogr that doesn't go through a VRT mechanism.

mdsumner: oh, I see - didn't realize there were vsicurl args -  thanks

rouault: > there's no unsafessl opt so

there is. Cf https://gdal.org/user/virtual_file_systems.html#vsicurl-http-https-ftp-files-random-access. It is actually with the ``/vsicurl?unsafessl=yes&url=http://`` syntax

sgillies: > /vsi is smething you pass to VSIFOpenL() to return a VSILFILE*
vrt:// is something you pass to GDALOpen() to return a GDALDataset*
Two different kind of objects

@rouault a year ago I would have completely accepted this, but I now know much about about VSI plugins. Enough to be dangerous!

In fact, users like @mdsumner pass `/vsi` filenames to `GDALOpen()` all the time, which then calls `VSIFOpenL()`. Users rarely call `VSIFOpenL()` directly.

Just as `GDALOpen()` now opens the `vrt://` target file, calls `GDALTranslate()` on it, then opens the resulting XML and returns that VRT dataset, a `vsitranslate` handler could open the target file, call `GDALTranslate()`, and then `VSIFOpenL()` the resulting XML string. Or am I missing an important detail?

rouault: > now opens the vrt:// target file, calls GDALTranslate() on it, then opens the resulting XML and returns that VRT dataset,

Pedantic and not important note: the implementation doesn't go through XML serialization. The VRT dataset is built using the VRT API (through GDALTranslate())

> a `vsitranslate` handler could open the target file, call `GDALTranslate()`, and then `VSIFOpenL()` the resulting XML. Or am I missing an important detail?

Oh I now understand what you meant: /vsitranslate would return the XML VRT file. That could work, although that seems to be slightly convoluted to me, as I'd expect /vsi to be for existing file-like objects. Not for generating a XML VRT file describing how to alter properties of a GDAL dataset. A user could expect /vsitranslate/ to potentially return a GeoTIFF file whereas the implementation couldn't reasonably only work for a VRT output file

mdsumner: shall we take this convo to the breaking changes issue?    I think this PR can be safely merged ?

mdsumner:
thanks!

PR #8416, opened 2023-09-18
/vsi network file system: support r+ access under CPL_VSIL_USE_TEMP_FILE_FOR_RANDOM_WRITE=YES ...

rouault:
... by copying initial content of remove file locally

### DIFF
--- a/port/cpl_vsil_s3.cpp
+++ b/port/cpl_vsil_s3.cpp
@@ -656,14 +656,14 @@ class VSIS3FSHandler final : public IVSIS3LikeFSHandler
         return STARTS_WITH(pszHeaderName, "x-amz-");
     }
 
+    VSIVirtualHandle *CreateWriteHandle(const char *pszFilename,
+                                        CSLConstList papszOptions) override;
+
   public:
     explicit VSIS3FSHandler(const char *pszPrefix) : m_osPrefix(pszPrefix)
     {
     }
     ~VSIS3FSHandler() override;
-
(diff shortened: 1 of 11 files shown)


port/cpl_vsil_s3.cpp
```diff
@@ -2034,45 +2027,88 @@ int VSIS3WriteHandle::Close()
...
         }
 
-        VSIS3HandleHelper *poS3HandleHelper = VSIS3HandleHelper::BuildFromURI(
-            pszFilename + GetFSPrefix().size(), GetFSPrefix().c_str(), false);
-        if (poS3HandleHelper == nullptr)
-            return nullptr;
-        VSIS3WriteHandle *poHandle = new VSIS3WriteHandle(
-            this, pszFilename, poS3HandleHelper, false, papszOptions);
-        if (!poHandle->IsOK())
+        const std::string osTmpFilename(CPLGenerateTempFilename(nullptr));
```
pedros007, line 2074: Could a user set `CPL_TMPDIR=/vsimem/` to write the tempfile to memory with a goal of lower IO latency assuming there is enough free RAM available?

https://github.com/OSGeo/gdal/blob/944a839bc0b8847585e017792985e04aced58e6e/port/cpl_path.cpp#L1115

rouault: > Could a user set `CPL_TMPDIR=/vsimem/` to write the tempfile to memory with a goal of lower IO latency assuming there is enough free RAM available?

yes, should work

PR #8419, opened 2023-09-18
ogr2ogr: better deal when reprojecting curve geometries to a non-curve geometry type (fixes #8332)

### LINKED ISSUE #8332: GeoJSON 2008 creates invalid geometry (self-intersection)

`ogr2ogr -f GeoJSON out.geojson test.gml -t_srs "EPSG:4326"` creates a valid geometry because the source geometry is valid but it creates a self intersecting geomety.

grafik

grafik

[test_data.zip](https://github.com/OSGeo/gdal/files/12523339/test_data.zip)

1. `ogr2ogr -f GeoJSON out.geojson test.gml -t_srs "EPSG:4326"`

2. Use "Check validity" from the processing toolbox in QGIS on out.geojson.

## Additional information

1. `-lco RFC7946=YES` creates a valid geometry:
`ogr2ogr -f GeoJSON out_7946.geojson test.gml -lco RFC7946=YES`

2. Reducing the precision doesn't help: 
`ogr2ogr -f GeoJSON out_7.geojson test.gml -t_srs "EPSG:4326" -lco COORDINATE_PRECISION=7`
 
## Operating system

Windows 10 Version 1809

3.6.4 and 3.7.1 from OSGeo4W

rouault:


apps/ogr2ogr_lib.cpp
```diff
@@ -5663,46 +5680,133 @@ int LayerTranslator::Translate(OGRFeature *poFeatureIn, TargetLayerInfo *psInfo,
...
+                                std::unique_ptr<OGRGeometry>(
+                                    OGRGeometryFactory::forceTo(
+                                        poReprojectedGeom->clone(),
+                                        eTargetType));
+                            if (!IsValid(poDstGeometryTmp.get()))
+                            {
+                                CPLDebug("OGR2OGR",
+                                         "Curve geometry no longer valid after "
+                                         "reprojection: transforming it into "
+                                         "linear one before reprojecting");
```
pathmapper, line 5790: For this [data](https://share.krzn.de/alfresco/service/ecm4u/exs/kreis-viersen/share/9URfvgdO/download/d29ya3NwYWNlOi8vU3BhY2VzU3RvcmUvYzkyZmUzMmMtYzI3YS00MjIzLTkxMDgtOGU5YTRjOTQyNzIx/test2.gml) with
`ogr2ogr -f GeoJSON out2_rfc.geojson test2.gml -lco RFC7946=YES --debug on`
there are 41 invalid geometries in the GeoJSON.

I suspect this is because of the default precision of 7 for RFC7946 because with 
`ogr2ogr -f GeoJSON out2_rfc_prec_15.geojson test2.gml -lco RFC7946=YES -lco COORDINATE_PRECISION=15`
all resulting geometries are valid.

But shouldn't be there this debug output?
I only get 
```
ogr2ogr -f GeoJSON out2_rfc.geojson test2.gml -lco RFC7946=YES --debug on
GML: Using Expat reader
GML: Global SRS = urn:ogc:def:crs:EPSG::25832
GDAL: GDALOpen(test2.gml, this=0x56267279e8e0) succeeds as GML.
GDAL: QuietDelete(out2_rfc.geojson) invoking Delete()
GDAL: GDALDriver::Create(GeoJSON,out2_rfc.geojson,0,0,0,Unknown,(nil))
GML: ResetReading()
GML: ResetReading()
GDALVectorTranslate: 176666 features written in layer 'test2'
GDAL: GDALClose(test2.gml, this=0x56267279e8e0)
GDAL: GDALClose(out2_rfc.geojson, this=0x562672b12120)
GDAL: In GDALDestroy - unloading GDAL shared library.
```

autotest/utilities/data/curvepolygon_epsg_32632.csv
```diff
@@ -0,0 +1,3 @@
+_WKT_CURVEPOLYGON_EPSG_32632
```
pathmapper, line 1: Out of curiosity: The original test data is in EPSG:25832 - do you opted here for EPSG:32632 to have the same geodetic system as EPSG:4326?

pathmapper (commented):
Works great, thanks a bunch! 

Tested this fix with several datasets with a lot of curve geometries.

I only wonder about the (missing?) debug output when using `RFC7946=YES`.

autotest/utilities/data/curvepolygon_epsg_32632.csv
rouault: > Out of curiosity: The original test data is in EPSG:25832 - do you opted here for EPSG:32632 to have the same geodetic system as EPSG:4326?

yes, to make the test independent of potential datum shifts between ETRS89 and WGS 84 which could occur at some point

apps/ogr2ogr_lib.cpp
rouault: > But shouldn't be there this debug output?

not one about geometry type conversion, since this is expected given the capabilities of the driver.
That said I've added a new commit to deal with the reduced coordinate precision of GeoJSON output to run MakeValid() if the input was valid, but the output (after rounding, and before this new correction) was invalid. Works well on your test dataset

pathmapper: Nice!

Works well also for other big datasets I've tried :tada:

PR #8423, opened 2023-09-19
autotest: Make SQLite, MySQL, Postgres tests run independently

dbaston:
## What does this PR do?

This PR updates the tests for PostgreSQL, SQLite, and MySQL to run independently.
Each test runs in a clean schema / database to avoid interactions between tests.
Test tables that are used by more than one dataset can be brought in with, e.g. `@pytest.usefixtures("tpoly")`.

The Postgres tests cannot run in parallel because some of them manipulate `spatial_ref_sys`. 

#4407

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/autotest/ogr/ogr_pg.py
+++ b/autotest/ogr/ogr_pg.py
@@ -27,8 +27,9 @@
 # Boston, MA 02111-1307, USA.
 ###############################################################################
 
+import functools
 import os
-import shutil
+import re
 import sys
 import threading
 import time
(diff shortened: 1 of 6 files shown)


autotest/ogr/ogr_pg.py
```diff
@@ -1633,64 +1595,17 @@ def test_ogr_pg_29():
...
-    feat = None
-
-    ds.Destroy()
-
-
-###############################################################################
-# Duplicate all data types in PG_USE_COPY mode
-
-
-def test_ogr_pg_30():
```
dbaston, line 1645: Merged this test into `test_ogr_pg_28`.

autotest/ogr/ogr_pg.py
```diff
@@ -915,126 +1127,51 @@ def test_ogr_pg_16():
...
-
-    #
-    # Test reading 4-dim geometries normalized to OGC WKT form.
-    #
+@only_with_postgis
+@pytest.mark.usefixtures("testgeom")
+def test_ogr_pg_20(pg_ds):
 
-    layer = gdaltest.pg_ds.GetLayerByName("testgeom")
+    layer = pg_ds.GetLayerByName("testgeom")
```
rouault, line 1165: The test currently runs with "ERROR 6: SetFeature : unsupported operation on a read-only datasource." errors, that happens to be undetected:

```suggestion
    ds = reconnect(pg_ds, update=1)
    layer = ds.GetLayerByName("testgeom")
```

The PG driver is quite lax about checking the update mode, so most of this issues are undetected. I also see that test_ogr_22 would also need an updatable dataset, although it doesn't hit any code path that checks permission. And probably other tests. Perhaps there should be a pg_updatable_ds fixture, and use it more extensively?

dbaston: Having `pg_ds` be read-only is an oversight (for example, `mysql_ds` opens in update mode). Since `pg_ds` is scoped to each test, is there a use for a read-only version?

rouault: >  is there a use for a read-only version?

probably not, unless we test read-only checks, which we probably don't do in ogr_pg

jratike80: It depends on what the tests are testing. In managed production systems users tend to have minimum rights. For example Mapserver user having a layer through OGR connection should not have any other rights than SELECT on the data table, geometry_columns, spatial_ref_sys and maybe some other metadata and system tables. Developers tend to forget that there are users who operate with limited rights. However, I have not seen any reports about GDAL failing because of this.
I guess that opening a connection as read-only is not the same than opening a connection as a user who has only SELECT rights. Do we test the latter in any place? Is there any need for such tests, for example try to update without proper rights?

rouault: There are 2 things regarding rights: the update flag that is handled by GDAL (at the driver level, and the PG driver lacks checks to validate that operations that update are done with the update flag set), and possibly opening the underlying datasource in read-only (which the PG driver does not currently either. According to https://stackoverflow.com/questions/41960217/connecting-to-postgres-database-with-read-only-access,it can be done at the session level with "SET SESSION CHARACTERISTICS AS TRANSACTION READ ONLY;").
So currently opening a PG connection in read-only mode does not prevent a user from modifying the database, if he has rights at the PostgreSQL level. Which is kind of unexpected.

Issue #8422, opened 2023-09-19
expose GDALDriverManager::AutoLoadDrivers in C API?


tbonfort:
I would like to be able to register drivers compiled as plugins without having to call `GDALAllRegister()`. From the code in gdal_priv.h it would seem I can be able to do this through the C++ api, but given the name of that file I'm not sure that is the correct solution. Would it be OK to expose it in our C API? If so I can submit a PR.

rouault:
> Would it be OK to expose it in our C API? If so I can submit a PR.

sounds reasonable. But you'd probably want the function to do all of :
```
    auto poDriverManager = GetGDALDriverManager();
    // AutoLoadDrivers is a no-op if compiled with GDAL_NO_AUTOLOAD defined.
    poDriverManager->AutoLoadDrivers();

    poDriverManager->AutoLoadPythonDrivers();

    /* -------------------------------------------------------------------- */
    /*      Deregister any drivers explicitly marked as suppressed by the   */
    /*      GDAL_SKIP environment variable.                                 */
    /* -------------------------------------------------------------------- */
    poDriverManager->AutoSkipDrivers();

    poDriverManager->ReorderDrivers();
```

PR #8447, opened 2023-09-21
add GDALRegisterPlugins function to register all/only plugins

tbonfort:
implements #8422

for the naming I went with RegisterPlugins instead of AutoloadDrivers, as the
plugin mechanism can be used for registering more than just drivers (e.g. vrt
pixel functions, VSI handlers, ...)

### DIFF
--- a/autotest/cpp/test_gdal.cpp
+++ b/autotest/cpp/test_gdal.cpp
@@ -58,6 +58,12 @@ TEST_F(test_gdal, driver_manager)
     ASSERT_TRUE(nullptr != drv_mgr);
 }
 
+// Test that GDALRegisterPlugins can be called
+TEST_F(test_gdal, register_plugins)
+{
+    GDALRegisterPlugins();
+}
+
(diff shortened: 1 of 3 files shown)


tbonfort:
Some additional comments:
- not sure if and how this should be tested, as we would need to compile an out-of-tree plugin, and run the test suite in a code path where we are sure that no previous GDALAllRegister() calls have happened.
- it might be handy to add a `GDALRegisterPlugin(char *name)` to only load a single plugin and return an error if not found.
- should this also be exposed to swig?

rouault:
> * not sure if and how this should be tested,

I would just do a minimal testing of calling the function in autotest/cpp/test_gdal.cpp

> should this also be exposed to swig?

I don't see a strong need for that. The Python bindings for example automatically call GDALAllRegister()

rouault:
The failure on build-windows-conda is unrelated and can be solved by rebasing on top of latest master

PR #8450, opened 2023-09-21
ArrowArray filter improvements

rouault:
- Python bindings: GetArrowStreamAsNumPy(): fix reading fixed size list arrays that were ignoring the parent offset (affects Parquet)
- Python bindings: GetArrowStreamAsNumPy(): fix reading fixed width binary that were misusing the offset (affects Parquet)
- Python bindings: GetArrowStreamAsNumPy(): implement reading struct (s+) fields
- OGRLayer::CanPostFilterArrowArray(): allow filtering on struct fields (if the underlying type is supported), and allow compacting struct and list fields
- OGRLayer::CanPostFilterArrowArray(): allow compacting map fields

### DIFF
--- a/autotest/ogr/ogr_parquet.py
+++ b/autotest/ogr/ogr_parquet.py
@@ -2120,13 +2120,6 @@ def test_ogr_parquet_arrow_stream_fast_attribute_filter_pyarrow(
 
     ds = ogr.Open(test_file)
     lyr = ds.GetLayer(0)
-    ignored_fields = []
-    lyr_defn = lyr.GetLayerDefn()
-    for i in range(lyr_defn.GetFieldCount()):
-        fld_defn = lyr_defn.GetFieldDefn(i)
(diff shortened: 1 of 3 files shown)


hobu:
> OGRLayer::CanPostFilterArrowArray(): allow filtering on struct fields (if the underlying type is supported), and allow compacting struct and list fields
OGRLayer::CanPostFilterArrowArray(): allow compacting map fields

Are we exposing Arrow into OGR's public API?

rouault:
> Are we exposing Arrow into OGR's public API?

This is a protected method, but yes ArrowArrayStream, ArrowSchema and ArrowArray are in the public API since https://gdal.org/development/rfc/rfc86_column_oriented_api.html through the OGRLayer::GetArrowStream() public method

hobu:
ah, I hadn't realized. sorry for the noise.

PR #8454, opened 2023-09-22
SWIG: Expose GDALClose

dbaston:
## What does this PR do?

Exposes `GDALClose` through SWIG so that `DataSource` and `Dataset` objects can be explicitly closed. I named the method `Close` rather than `close` for consistency with the rest of the library. I do not expect this method to be used often. For most cases, a `with` block would be a better choice.

For `DataSource`, this is essentially a synonym of `Destroy`, except that it does not leave the object in a state where further use will cause a crash. Instead, further use will cause an unintuitive `TypeError`:

```
>>> ds.Close()
0
>>> ds.GetLayerCount()
  File "/home/dan/dev/build-gdal-Desktop-Debug/swig/python/osgeo/ogr.py", line 830, in GetLayerCount
    return _ogr.DataSource_GetLayerCount(self, *args)
TypeError: in method 'DataSource_GetLayerCount', argument 1 of type 'OGRDataSourceShadow *'
```

We can get a better error message by annotating _every single method_ with an explicit check, e.g.:

```
%feature("pythonprepend") GetLayerCount %{
    if self.this is None:
        raise ValueError("Operation on closed Dataset")
%}
```

but it appears for now (https://github.com/swig/swig/issues/2657) that there is no automated way to ask SWIG to apply this to all methods. 

## Tasklist

 - [x] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/swig/include/ogr.i
+++ b/swig/include/ogr.i
@@ -871,8 +871,13 @@ public:
   char const *name;
 %mutable;
 
+
   ~OGRDataSourceShadow() {
     OGRReleaseDataSource(self);
+  }
+
+  CPLErr Close() {
+    return GDALClose(self);
   }
 
   int GetRefCount() {
--- a/swig/include/python/ogr_python.i
+++ b/swig/include/python/ogr_python.i
@@ -82,11 +82,15 @@ def _WarnIfUserHasNotSpecifiedIfUsingExceptions():
       "Once called, self has effectively been destroyed.  Do not access. For backwards compatibility only"
       _ogr.delete_DataSource(self)
       self.thisown = 0
+      self.this = None
+      self._invalidate_layers()
 
     def Release(self):
       "Once called, self has effectively been destroyed.  Do not access. For backwards compatibility only"
       _ogr.delete_DataSource(self)
       self.thisown = 0
+      self.this = None
+      self._invalidate_layers()
 
     def Reference(self):
       "For backwards compatibility only."
--- a/swig/include/python/ogr_python.i
+++ b/swig/include/python/ogr_python.i
@@ -77,6 +77,7 @@ def _WarnIfUserHasNotSpecifiedIfUsingExceptions():
 
 %extend OGRDataSourceShadow {
   %pythoncode {
+
     def Destroy(self):
       "Once called, self has effectively been destroyed.  Do not access. For backwards compatibility only"
       _ogr.delete_DataSource(self)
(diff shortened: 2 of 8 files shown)


swig/include/ogr.i
```diff
@@ -871,10 +871,15 @@ public:
   char const *name;
 %mutable;
 
+
   ~OGRDataSourceShadow() {
     OGRReleaseDataSource(self);
   }
 
+  OGRErr Close() {
```
rouault, line 879: should return a CPLErr or some CPLErr -> OGRErr translation should be done

rouault (commented):
> I named the method `Close` rather than `close` for consistency with the rest of the library.

Wondering if we shouldn't offer both (given that it would have the expected semantics) ?  Python'ists will call close() naturally, and GDAL'ers would probably attempt Close() :-)

swig/include/python/ogr_python.i
```diff
@@ -179,6 +178,13 @@ def _WarnIfUserHasNotSpecifiedIfUsingExceptions():
     self._add_layer_ref(val)
 %}
 
+%feature("pythonappend") Close %{
```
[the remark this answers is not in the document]
dbaston: Yes, I thought I should follow up with a "make objects not crash after calling Destroy" commit. Then I started wondering if `Destroy` should raise a deprecation warning. Then I started wondering if `ogr.DataSource` should exist at all, or if `ogr.Open` could just return a `Dataset` :)

swig/include/ogr.i
dbaston: changed to `CPLErr`

dbaston:
A behavior difference between this method and a typical Python `close()` is that a Python file object still allows access to some information after calling `close()`, while a GDAL dataset does not:

```
>>> f = open('/tmp/x', 'w')
>>> f.close()
>>> f.name
'/tmp/x' 
>>> f.mode
'w'
>>> f.encoding
'UTF-8'
```

vs

```
>>> ds = gdal.OpenEx('ogr/data/poly.shp')
>>> ds.Close()
0
>>> ds.GetDescription()
  File "/home/dan/dev/build-gdal-Desktop-Debug/swig/python/osgeo/gdal.py", line 3097, in GetDescription
    return _gdal.MajorObject_GetDescription(self, *args)
TypeError: in method 'MajorObject_GetDescription', argument 1 of type 'GDALMajorObjectShadow *'
```

PR #8458, opened 2023-09-22
gdal_translate: add  -dmo option for domain-specific md

mdsumner:
Adds `-dmo` to availalable options for gdal_translate. 

Usage is like that for `mo` but with a domain prefix and colon before the key,value pair: 

```
gdal_translate [in] [out] -dmo DOMAIN:key=value
```

Was discussed on gdal-dev: https://lists.osgeo.org/pipermail/gdal-dev/2023-September/057663.html

Ugh - this includes a PR to building_from_source.rst .... I'll have to restart this. 

 I will add 'dmo' to vrt:// in a future PR. 

I think a bit more doc is needed for what domains are allowed and being driver-specific, but I need a bit more experience to write anything meaningful there. 

## Tasklist

 - [x] Add test case(s)
 - [x] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/apps/gdal_translate_lib.cpp
+++ b/apps/gdal_translate_lib.cpp
@@ -59,6 +59,8 @@
 
 static int ArgIsNumeric(const char *);
 static void AttachMetadata(GDALDatasetH, const CPLStringList &);
+static void AttachDomainMetadata(GDALDatasetH, const CPLStringList &);
+
 static void CopyBandInfo(GDALRasterBand *poSrcBand, GDALRasterBand *poDstBand,
                          int bCanCopyStatsMetadata, int bCopyScale,
                          int bCopyNoData, bool bCopyRAT,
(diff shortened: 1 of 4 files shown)


jratike80:
Can user add and edit such domain-specific metadata with gdal_edit https://gdal.org/programs/gdal_edit.html?

mdsumner:
ah no I don't think so, thanks I'll look there - and ogr2ogr too

(I'm stalled submitting this properly because I keep making rebase mistakes, but I think I've worked out why now so will fix soon)

rouault:
One thing I realize know is that we have special metadata domains whose name contains a ':' . Like "xml:foo" or "json:bar". Most of them are meant to be read-only, but nothing prevents a user from setting them

```
from osgeo import gdal
ds = gdal.GetDriverByName('GTiff').Create("test.tif", 1, 1)
ds.SetMetadata(["<bar/>"], "xml:foo")
ds = None
ds = gdal.Open('test.tif')
print(ds.GetMetadata("xml:foo"))
['<bar/>']
```

PR #8461, opened 2023-09-23
Add time zone flag property to OGRFieldDefn, and use it in ArrowStream interface

### LINKED ISSUE #8460: ArrowStream interface: support for timezones

Assume a geojson file `test_datetime_tz.geojson` with content:
```
{
"type": "FeatureCollection",
"features": [
{ "type": "Feature", "properties": { "col": "2020-01-01T09:00:00-05:00" }, "geometry": { "type": "Point", "coordinates": [ 1.0, 1.0 ] } },
{ "type": "Feature", "properties": { "col": "2020-01-01T10:00:00-05:00" }, "geometry": { "type": "Point", "coordinates": [ 2.0, 2.0 ] } }
]
}
```

and reading this with the ArrowStream interface:

```
>>> from osgeo import ogr
>>> ds = ogr.Open("test_datetime_tz.geojson")
>>> lyr = ds.GetLayer(0)
>>> stream = lyr.GetArrowStreamAsPyArrow()
>>> list(stream)[0]
<pyarrow.lib.StructArray object at 0x7fedaa11dc00>
-- is_valid: all not null
-- child 0 type: int64
  [
    0,
    1
  ]
-- child 1 type: timestamp[ms]
  [
    2020-01-01 09:00:00.000,
    2020-01-01 10:00:00.000
  ]
-- child 2 type: binary
  [
    0101000000000000000000F03F000000000000F03F,
    010100000000000000000000400000000000000040
  ]
```

We get tz-naive timestamp values in the "local" time (wall time), not the UTC-equivalent values (after applying the offset). While the "traditional" field based access with `OGR_F_GetFieldAsDateTimeEx` will give you timezone information. 

It's not necessarily clear what would be the best behaviour here. GDAL doesn't actually store or know the time "zone" (like `Europe/Brussels`), only an offset per value (which is not necessarily constant for the full column). And thus it cannot return an Arrow timestamp type with a zone as `tz`. 
It _could_ return a timestamp with a fixed offset (but that's only possible if all offsets are the same, which you only know when parsing the last value, and so this would require a first pass to check this), or return `timestamp[ms, tz=UTC]` type (this doesn't preserve the wall time, but it does preserve the actual point in time). Although I can imagine that it depends on the use case whether you would prefer getting wall-time tz-naive timestamps or tz-aware UTC timestamps.

GDAL 3.7.2 installed with conda forge

rouault:
- Add OGRFieldDefn::GetTZFlag()/SetTZFlag(), and OGR_TZFLAG_ constants
- ogrinfo: output timezone flag
- GeoJSON reading: set field TZFlag
- ArrowStream interface: set timezone on schema (when known), and do necessary conversions (fixes #8460)
- Arrow/Parquet reading: set field TZFlag

### DIFF
--- a/ogr/ogr_core.h
+++ b/ogr/ogr_core.h
@@ -866,6 +866,29 @@ typedef enum
  */
 #define OGRNullMarker -21122
 
+/** Time zone flag indicating unknown timezone. For the
+ *  OGRFieldDefn::GetTZFlag() property, this may also indicate a mix of
+ *  unknown, localtime or known time zones in the same field.
+ */
+#define OGR_TZFLAG_UNKNOWN 0
+
+/** Time zone flag indicating local time */
+#define OGR_TZFLAG_LOCALTIME 1
+
+/** Time zone flag only returned by OGRFieldDefn::GetTZFlag() to indicate
+ * that all values in the field have a known time zone (ie different from
+ * OGR_TZFLAG_UNKNOWN and OGR_TZFLAG_LOCALTIME), but it may be different among
+ * features. */
+#define OGR_TZFLAG_MIXED_TZ 2
+
+/** Time zone flag indicating UTC.
+ * Used to derived other time zone flags with the following logic:
+ * - values above 100 indicate a 15 minute increment per unit.
+ * - values under 100 indicate a 15 minute decrement per unit.
+ * For example: a value of 101 indicates UTC+00:15, a value of 102 UTC+00:30,
+ * a value of 99 UTC-00:15 ,etc. */
+#define OGR_TZFLAG_UTC 100
+
 /************************************************************************/
 /*                               OGRField                               */
 /************************************************************************/
--- a/ogr/ogrsf_frmts/generic/ogrlayerarrow.cpp
+++ b/ogr/ogrsf_frmts/generic/ogrlayerarrow.cpp
@@ -31,6 +31,7 @@
 #include "ogr_recordbatch.h"
 #include "ograrrowarrayhelper.h"
 #include "ogr_wkb.h"
+#include "ogr_p.h"
 
 #include "cpl_float.h"
 #include "cpl_time.h"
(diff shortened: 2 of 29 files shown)


jratike80:
Wouldn't it be good to test also the case when conversion into UTC is changing the the date part, not only hours and minutes?

ogr/ogr_core.h
```diff
@@ -866,6 +866,29 @@ typedef enum
...
+/** Time zone flag only returned by OGRFieldDefn::GetTZFlag() to indicate
+ * that all values in the field have a known time zone (ie different from
+ * OGR_TZFLAG_UNKNOWN and OGR_TZFLAG_LOCALTIME), but it may be different among
+ * features. */
+#define OGR_TZFLAG_MIXED_TZ 2
+
+/** Time zone flag indicating UTC.
+ * Used to derived other time zone flags with the following logic:
+ * - values above 100 indicate a 15 minute increment per unit.
+ * - values above 100 indicate a 15 minute decrement per unit.
```
jratike80, line 887: There reads "above" on both lines. Should there be "under" on line 887?

rouault: indeed, thanks. fixed

rouault:
> Wouldn't it be good to test also the case when conversion into UTC is changing the the date part, not only hours and minutes?

the time zone correction is applied on the timestamp (as a number of seconds since Unix Epoch), before any potential splitting as YYYY-MM-DDTHH:mm;ss, so change of date itself isn't super relevant.

jratike80:
> the time zone correction is applied on the timestamp (as a number of seconds since Unix Epoch), before any potential splitting as YYYY-MM-DDTHH:mm;ss, so change of date itself isn't super relevant.

Ok, then it is only the user who may get gray hair due to change of date or year, or due to February 29, but GDAL will handle all cases correctly. That's fine.

ogr/ogrsf_frmts/generic/ogrlayerarrow.cpp
```diff
@@ -318,8 +321,44 @@ int OGRLayer::GetArrowSchema(struct ArrowArrayStream *,
                 break;
 
             case OFTDateTime:
-                psChild->format = "tsm:";
+            {
+                const char *pszPrefix = "tsm:";
+                const char *pszTZOverride =
+                    m_aosArrowArrayStreamOptions.FetchNameValueDef("TIMEZONE",
+                                                                   "");
```
jorisvandenbossche, line 328: This option is only meant to be used internally (so that GPKG driver can set it)? 
(since I don't see it documented; it's probably also not really useful to expose)

rouault: yes, this was internal only. But I've decided to expose it, if user want to force a timezone if they have knownledge the driver doesn't have

PR #8466, opened 2023-09-24
Update gdal_grid.rst

jidanni:
These "ref"s to the same page that we are on just cause a mess on the man page...

### DIFF
--- a/doc/source/programs/gdal_grid.rst
+++ b/doc/source/programs/gdal_grid.rst
@@ -350,7 +350,7 @@ It has following parameters:
 Data metrics
 ------------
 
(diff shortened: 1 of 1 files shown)


jratike80:
Try to fix the ref instead of removing it. For example the ref here seems to work
https://github.com/OSGeo/gdal/blob/master/doc/source/programs/gdallocationinfo.rst?plain=1#L98C1-L98C31

rouault:
fixed per b744e4a629  using either ``:program:`` (when the utility mentions itself) or ``:ref:`` (when linking to another utility)

PR #8489, opened 2023-09-28
Doc: Update Python gotchas

dbaston:
## What does this PR do?

- Note that exceptions will be the default in GDAL 4.0
- Update discussion of relationships to note exceptions vs crashes
- Update discussion of Destroy to describe context managers and Close

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/doc/source/api/python_gotchas.rst
+++ b/doc/source/api/python_gotchas.rst
@@ -80,7 +80,7 @@ If we delete ``dataset`` and then try to use ``band``, Python will throw a confu
    >>> dataset = gdal.Open('C:\\RandomData.img')
    >>> band = dataset.GetRasterBand(1)
    >>> del dataset           # This will cause the Python garbage collector to deallocate dataset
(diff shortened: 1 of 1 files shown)


doc/source/api/python_gotchas.rst
```diff
@@ -68,24 +72,28 @@ Consider this example:
...
 
 .. code-block::
 
    >>> from osgeo import gdal
    >>> dataset = gdal.Open('C:\\RandomData.img')
    >>> band = dataset.GetRasterBand(1)
    >>> del dataset           # This will cause the Python garbage collector to deallocate dataset
-   >>> band.GetChecksum()    # This will now crash Python because the band's dataset is gone
-   < Python crashes >
+   >>> band.Checksum()       # This will now crash Python because the band's dataset is gone
```
rouault, line 83:
```suggestion
   >>> band.Checksum()       # This will throw an exception because the band's dataset is gone
```

PR #8505, opened 2023-10-02
Update the list of released PgSQL versions to the present

pramsey:
I'm not sure what the balance between this list of versions and the `PostgreSQL_ADDITIONAL_VERSIONS` configuration parameter is supposed to be, but superficially it makes sense to me that all the officially released versions should be in this list?

If there any argument against pulling this back into stable branches?


rouault:
> it makes sense to me that all the officially released versions should be in this list?

yes.
Upstream version in CMake repository should likely be updated to: https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/FindPostgreSQL.cmake?ref_type=heads

> If there any argument against pulling this back into stable branches?

no, that's fine. adding the "backport 3.7" tag

pramsey:
Reported upstream in https://gitlab.kitware.com/cmake/cmake/-/issues/25292

PR #8508, opened 2023-10-02
zarr_driver.py: fix test_zarr_read_fill_value_complex_datatype_v3() to not alter its input parameters (fixes #8507)

### LINKED ISSUE #8507: zarr test fails when run out-of-order

I would expect the following run of two tests cases to pass, but the second fails:

```
pytest --random-order gdrivers/zarr_driver.py -k read_fill_value_complex_datatype_v3 -k nodata7 --random-order-seed=460289
```

```
>       assert ar.GetNoDataValueAsRaw() == bytes(
            array.array(dtype, nodata)
        ), struct.unpack(dtype * 2, ar.GetNoDataValueAsRaw())
E       AssertionError: (1.058925634e-314, 2.0)
E       assert bytearray(b'\...x00\x00\x00@') == b'\x00\x00\x0...\x00\x00\x00@'
E         At index 2 diff: 192 != 0
E         Use -v to get more diff
```

I have tried using `tmp_vsimem` to avoid interference between the tests, but it does not resolve the problem.

Test command above

## Operating system

Pop OS 22.04

10bdf2c5193984dde9f74cde4e22a657ec655f83

rouault:


dbaston:
Thanks!

PR #8513, opened 2023-10-03
autotest: allow utilities tests to run in parallel

dbaston:
## What does this PR do?

* Applies `tmp_vsimem` fixture to tests in `utilities` to avoid collisions when running tests in parallel (`pytest -n`)
* Adds `os.PathLike` handling to some arguments that were lacking it

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/autotest/utilities/test_gdal_translate.py
+++ b/autotest/utilities/test_gdal_translate.py
@@ -965,7 +965,7 @@ def test_gdal_translate_34(gdal_translate_path, tmp_path):
 # Test various errors (missing source or dest...)
 
 
-def test_gdal_translate_35(gdal_translate_path):
+def test_gdal_translate_35(gdal_translate_path, tmp_vsimem):
 
     (_, err) = gdaltest.runexternal_out_and_err(gdal_translate_path)
     assert "No source dataset specified" in err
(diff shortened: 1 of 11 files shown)


autotest/utilities/test_gdal_translate.py
```diff
@@ -965,7 +965,7 @@ def test_gdal_translate_34(gdal_translate_path, tmp_path):
 # Test various errors (missing source or dest...)
 
 
-def test_gdal_translate_35(gdal_translate_path):
+def test_gdal_translate_35(gdal_translate_path, tmp_vsimem):
```
rouault, line 968: I don't believe tmp_vsimem here and in the below tests of that file is useful, since it is used in gdaltest.runexternal_out_and_err() which is an external process. Thus plain /vsimem/ is fine. That said, that change probably doesn't hurt

rouault (approved):

autotest/utilities/test_gdal_translate.py
dbaston: Agreed, the benefit here (and in many other changes) is just in a future `grep /vsimem/` or `grep tmp/` coming up clean.

PR #8517, opened 2023-10-04
Add a OGRLayer::WriteArrowBatch method

rouault:
This is the continuation of [RFC 86 Column-oriented read API for vector layers](https://gdal.org/development/rfc/rfc86_column_oriented_api.html), but on the writing side.

Cf https://github.com/rouault/gdal/blob/write_arrow_batch/doc/source/tutorials/vector_api_tut.rst#writing-to-ogr-using-the-arrow-c-data-interface for an introduction

3 new methods to OGRLayer are added (and their corresponding C API and SWIG Python bindings):

* OGRLayer::IsArrowSchemaSupported():

```
/** Returns whether the provided ArrowSchema is supported for writing.
 *
 * This method exists since not all drivers may support all Arrow data types.
 *
 * The ArrowSchema must be of type struct (format=+s)
 *
 * It is recommended to call this method before calling WriteArrowBatch().
 *
 * This is the same as the C function OGR_L_IsArrowSchemaSupported().
 *
 * @param schema Schema of type struct (format = '+s')
 * @param[out] osErrorMsg Reason of the failure, when this method returns false.
 * @return true if the ArrowSchema is supported for writing.
 * @since 3.8
 */
bool OGRLayer::IsArrowSchemaSupported(const struct ArrowSchema *schema,
                                      std::string &osErrorMsg) const;
```

*  OGRLayer::CreateFieldFromArrowSchema()

```
/** Creates a field from an ArrowSchema.
 *
 * This should only be used for attribute fields. Geometry fields should
 * be created with CreateGeomField(). The FID field should also not be
 * passed with this method.
 *
 * Contrary to the IsArrowSchemaSupported() and WriteArrowBatch() methods, the
 * passed schema must be for an individual field, and thus, is *not* of type
 * struct (format=+s) (unless writing a set of fields grouped together in the
 * same structure).
 *
 * This method and CreateField() are mutually exclusive in the same session.
 *
 * This method is the same as the C function OGR_L_CreateFieldFromArrowSchema().
 *
 * @param schema Schema of the field to create.
 * @param papszOptions Options. Pass nullptr currently.
 * @return true in case of success
 * @since 3.8
 */
bool OGRLayer::CreateFieldFromArrowSchema(const struct ArrowSchema *schema,
                                          CSLConstList papszOptions);
```

* OGRLayer::WriteArrowBatch() 

```
/** Writes a batch of rows from an ArrowArray.
 *
 * This is semantically close to calling CreateFeature() with multiple features
 * at once.
 *
 * The ArrowArray must be of type struct (format=+s), and its children generally
 * map to a OGR attribute or geometry field (unless they are struct themselves).
 *
 * Method IsArrowSchemaSupported() can be called to determine if the schema
 * will be supported by WriteArrowBatch().
 *
 * OGR fields for the corresponding children arrays must exist and be of a
 * compatible type. For attribute fields, they should be created with
 * CreateFieldFromArrowSchema().
 *
 * Arrays for geometry columns should be of binary or large binary type and contain WKB geometry.
 *
 * Note that the passed array may be set to a released state
 * (array->release==NULL) after this call (not by the base implementation,
 * but in specialized ones such as Parquet or Arrow for example)
 *
 * Supported options of the base implementation are:
 * <ul>
 * <li>FID=name. Name of the FID column in the array. If not provided,
 *     GetFIDColumn() is used to determine it. The special name
 *     OGRLayer::DEFAULT_ARROW_FID_NAME is also recognized if neither FID nor
 *     GetFIDColumn() are set.
 *     The corresponding ArrowArray must be of type int32 (i) or int64 (l).
 *     On input, values of the FID column are used to create the feature.
 *     On output, the values of the FID column may be set with the FID of the
 *     created feature (if the array is not released).
 * </li>
 * <li>GEOMETRY_NAME=name. Name of the geometry column. If not provided,
 *     GetGeometryColumn() is used. The special name
 *     OGRLayer::DEFAULT_ARROW_GEOMETRY_NAME is also recognized if neither
 *     GEOMETRY_NAME nor GetGeometryColumn() are set.
 *     Geometry columns are also identified if they have
 *     ARROW:extension:name=ogc.wkb as a field metadata.
 *     The corresponding ArrowArray must be of type binary (w) or large
 *     binary (W).
 * </li>
 * </ul>
 *
 * The following example demonstrates how to copy a layer from one format to
 * another one (assuming it has at most a single geometry column):
\code{.py}
    def copy_layer(src_lyr, out_filename, out_format, lcos = {}):
        stream = src_lyr.GetArrowStream()
        schema = stream.GetSchema()

        # If the source layer has a FID column and the output driver supports
        # a FID layer creation option, set it to the source FID column name.
        if src_lyr.GetFIDColumn():
            creationOptions = gdal.GetDriverByName(out_format).GetMetadataItem(
                "DS_LAYER_CREATIONOPTIONLIST"
            )
            if creationOptions and '"FID"' in creationOptions:
                lcos["FID"] = src_lyr.GetFIDColumn()

        with ogr.GetDriverByName(out_format).CreateDataSource(out_filename) as out_ds:
            if src_lyr.GetLayerDefn().GetGeomFieldCount() > 1:
                out_lyr = out_ds.CreateLayer(
                    src_lyr.GetName(), geom_type=ogr.wkbNone, options=lcos
                )
                for i in range(src_lyr.GetLayerDefn().GetGeomFieldCount()):
                    out_lyr.CreateGeomField(src_lyr.GetLayerDefn().GetGeomFieldDefn(i))
            else:
                out_lyr = out_ds.CreateLayer(
                    src_lyr.GetName(),
                    geom_type=src_lyr.GetGeomType(),
                    srs=src_lyr.GetSpatialRef(),
                    options=lcos,
                )

            success, error_msg = out_lyr.IsArrowSchemaSupported(schema)
            assert success, error_msg

            src_geom_field_names = [
                src_lyr.GetLayerDefn().GetGeomFieldDefn(i).GetName()
                for i in range(src_lyr.GetLayerDefn().GetGeomFieldCount())
            ]
            for i in range(schema.GetChildrenCount()):
                # GetArrowStream() may return "OGC_FID" for a unnamed source FID
                # column and "wkb_geometry" for a unnamed source geometry column.
                # Also test GetFIDColumn() and src_geom_field_names if they are
                # named.
                if (
                    schema.GetChild(i).GetName()
                    not in ("OGC_FID", "wkb_geometry", src_lyr.GetFIDColumn())
                    and schema.GetChild(i).GetName() not in src_geom_field_names
                ):
                    out_lyr.CreateFieldFromArrowSchema(schema.GetChild(i))

            write_options = []
            if src_lyr.GetFIDColumn():
                write_options.append("FID=" + src_lyr.GetFIDColumn())
            if (
                src_lyr.GetLayerDefn().GetGeomFieldCount() == 1
                and src_lyr.GetGeometryColumn()
            ):
                write_options.append("GEOMETRY_NAME=" + src_lyr.GetGeometryColumn())

            while True:
                array = stream.GetNextRecordBatch()
                if array is None:
                    break
                out_lyr.WriteArrowBatch(schema, array, write_options)
\endcode
 *
 * This method and CreateFeature() are mutually exclusive in the same session.
 *
 * This method is the same as the C function OGR_L_WriteArrowBatch().
 *
 * @param schema Schema of array
 * @param array Array of type struct. It may be released (array->release==NULL)
 *              after calling this method.
 * @param papszOptions Options. Null terminated list, or nullptr.
 * @return true in case of success
 * @since 3.8
 */

bool OGRLayer::WriteArrowBatch(const struct ArrowSchema *schema,
                               struct ArrowArray *array,
                               CSLConstList papszOptions);
```

Some benchmarks comparing a ogr2ogr_arrow.py utility using the above copy_layer() method using GetNextRecordBatch() + WriteArrowBatch() to traditionnal ogr2ogr using GetNextFeature() + CreateFeature():

```
$ time python ogr2ogr_arrow.py out.parquet nz-building-outlines.parquet

real    0m4,268s
user    0m4,154s
sys	    0m1,777s

$ time ogr2ogr  out.parquet nz-building-outlines.parquet

real    0m11,300s
user    0m10,907s
sys     0m1,012s

$ time pythonogr2ogr_arrow.py out.gpkg nz-building-outlines.gpkg -lco SPATIAL_INDEX=NO

real    0m13,021s
user    0m13,893s
sys     0m2,200s

$ time ogr2ogr out.gpkg nz-building-outlines.gpkg -lco SPATIAL_INDEX=NO

real    0m16,977s
user    0m15,541s
sys     0m1,415s

```

CC @jorisvandenbossche, @paleolimbot, @kylebarron

### DIFF
--- a/doc/source/tutorials/vector_api_tut.rst
+++ b/doc/source/tutorials/vector_api_tut.rst
@@ -1747,13 +1747,6 @@ Supported options of the base implementation are:
   The corresponding ArrowArray must be of type binary (w) or large
   binary (W).
 
-- USE_FALLBACK_TYPES=YES. This may be set to allow GDAL to fallback to
-  another OGRFieldType data type (typically String) when the natural OGR type
-  is not supported in creation by the driver. This implies that the layer
-  implements GetDataset() to be able to retrieve supported data types from the
-  driver. This option must be used consistently among IsArrowSchemaSupported(),
-  CreateFieldFromArrowSchema() and WriteArrowBatch().
-
 Drivers that have a specialized implementation (such as :ref:`vector.parquet`
 and :ref:`vector.arrow`) advertise the OLCFastWriteArrowBatch layer capability.
 
--- a/doc/source/tutorials/vector_api_tut.rst
+++ b/doc/source/tutorials/vector_api_tut.rst
@@ -1025,6 +1025,8 @@ of a integer field and a geometry field:
     }
 
 
+To write features by batches using an ArrowArray, consult :ref:`vector_api_tut_arrow_write`.
+
 Writing To OGR
 --------------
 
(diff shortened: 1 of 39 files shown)


paleolimbot (commented):
This is so cool! I am personally excited for how this will simplify writing to OGR from R (but I know it has many other applications). It also opens up the ability to wrap OGR in an ADBC driver (because ADBC has a bulk ingest feature that is basically writing batches in this exact form).

It may not help you at this point, but the nanoarrow library ( https://github.com/apache/arrow-nanoarrow ) implements + tests a number of the helpers here. It's basically a copy/paste of two files...I left a few notes where it might help (but obviously no pressure to use since you already have test/feature coverage of more or less the whole thing).

I may have missed checks for a non-null `schema->dictionary` here...it's important to check (otherwise you'd be ingesting an integer array that is actually a dictionary-encoded string). Again, apologies if I missed it!

rouault:
> I may have missed checks for a non-null `schema->dictionary` here...it's important to check (otherwise you'd be ingesting an integer array that is actually a dictionary-encoded string). Again, apologies if I missed it!

I actually missed dealing with dictionaries (for the generic implementation. that should work out-of-the-box for Arrow & Parquet drivers). A possibility would be to map a dictionary to a OGR coded field domain (https://gdal.org/api/ogrfeature_cpp.html#classOGRCodedFieldDomain): concept coming from ESRI file geodatabase, and that can be expressed in GeoPackage.  The only thing that I'm not sure about if its the values of the dictionary for a given field can change from one record batch to another one? (I already made that perhaps incorrect assumption in the reading side of the Arrow & Parquet drivers) If so, that would seriously complicate the implementation (the content of the coded field domain would have to be updated each time WriteArrowBatch() is called, but not all drivers support UpdateFieldDomain(), so the creation of the field domain would have to be defered...), to the point it might be simpler to "resolve" the dictionary and use the type of the dictionary as the OGR field type.

I actually looked at nanoarrow which looks very promising, but I'm undecided if I would switch. Lists and maps don't have a dedicated API, right ?

rouault:
Dictionary now handled by resolving to its values

doc/source/tutorials/vector_api_tut.rst
```diff
@@ -1662,3 +1664,173 @@ In Python :
...
+map to a OGR attribute or geometry field (unless they are struct themselves).
+
+Method :cpp:func:`OGRLayer::IsArrowSchemaSupported` can be called to determine
+if the schema will be supported by WriteArrowBatch().
+
+OGR fields for the corresponding children arrays must exist and be of a
+compatible type. For attribute fields, they should be created with
+:cpp:func:`OGRLayer::CreateFieldFromArrowSchema`.
+
+Arrays for geometry columns should be of binary type and contain WKB geometry.
```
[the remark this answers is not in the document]
rouault: both int32 and int64 offsets are supported. I've just clarified the doc

ogr/ogrsf_frmts/generic/ogrlayerarrow.cpp
```diff
@@ -30,16 +30,70 @@
...
+constexpr char ARROW_LETTER_FLOAT64 = 'g';
+constexpr char ARROW_LETTER_STRING = 'u';
+constexpr char ARROW_LETTER_LARGE_STRING = 'U';
+constexpr char ARROW_LETTER_BINARY = 'z';
+constexpr char ARROW_LETTER_LARGE_BINARY = 'Z';
+constexpr char ARROW_2ND_LETTER_LIST = 'l';
+constexpr char ARROW_2ND_LETTER_LARGE_LIST = 'L';
+static inline bool IsStructure(const char *format)
+{
+    return format[0] == '+' && format[1] == 's';
```
paleolimbot, line 63:
```suggestion
    return format[0] == '+' && format[1] == 's' && format[2] == '\0';
```

It came up in a recent discussion about adding types to the C data interface in Arrow land that consumers should check the entire string (lest it be something invalid like `"+something else"`). I think our upcoming change to the spec won't break any of these, but it's still probably a good idea to add the null terminator check for these helpers. (Apologies if you do this somewhere else that I missed!)

rouault: ok done

rouault:
This should now be pretty feature complete regarding the Arrow data types supported. There is perhaps one inconsistency: for a Arrow data type like a list of integer that translates into a OGR OFTIntegerList, if the output layer doens't natively support OFTIntegerList (GeoPackage for example), we error out unless the user specifies USE_FALLBACK_TYPES=YES in which case we store it as JSON in a OFTString(OFSTJSON) field. But for for a Arrow data type like a map which has no direct OGR type match, we automatically use a OFTString (OFSTJSON) field with the JSON representation of the map. That doesn't look very consistent. I'm hesitating between:
- making USE_FALLBACK_TYPES=YES the default, and offering the possibility to set it to NO. But the effect of NO would be restricted to the first case I mentionned (disallowing OFTIntegerList -> OFTString). Not sure if that's matches a real use case.
- making USE_FALLBACK_TYPES=YES behaviour a non-parametrable behaviour. I'm leaning towards that.
Opinions ?

rouault:
> would the representation be JSON?

yes

I've just made USE_FALLBACK_TYPES=YES behaviour a non-parametrable behaviour

rouault:
Use of those new capabilities in ogr2ogr in https://github.com/OSGeo/gdal/pull/8544

PR #8518, opened 2023-10-05
Add a GDALPamDataset::SetDerivedDatasetName() method, …

rouault:
…and use it to be able to save statistics of datasets returned by GDALMDArray::AsClassicDataset()

### DIFF
--- a/gcore/gdalpamdataset.cpp
+++ b/gcore/gdalpamdataset.cpp
@@ -137,6 +137,18 @@
  *
  *      poDS->TryLoadXML();
  * \endcode
+ *
+ * In some situations where a derived dataset (e.g. used by
+ * GDALMDArray::AsClassicDataset()) is linked to a physical file, the name of
+ * the derived dataset is set with the SetDerivedSubdatasetName() method.
+ *
+ * \code
+ *      poDS->SetDescription( poOpenInfo->pszFilename );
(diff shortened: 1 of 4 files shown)


gcore/gdalpamdataset.cpp
```diff
@@ -916,29 +943,44 @@ CPLErr GDALPamDataset::TryLoadXML(char **papszSiblingFiles)
...
+        std::string osSubNodeValue;
+        if (!psPam->osSubdatasetName.empty())
         {
-            if (psSubTree->eType != CXT_Element ||
-                !EQUAL(psSubTree->pszValue, "Subdataset"))
-                continue;
+            osSubNode = "Subdataset";
+            osSubNodeValue = psPam->osSubdatasetName;
+        }
+        else if (!psPam->osDerivedDatasetName.empty())
```
elpaso, line 955: the `else if` excludes the possibility to have both "Subdataset" and "DerivedDataset": are the options mutually exclusive?

rouault: yes ad documented above in https://github.com/OSGeo/gdal/pull/8518/files#diff-8fa35e4c010f88e38395f7a251e702945e9000d65457f200dd5efe865a780ff8R790

PR #8519, opened 2023-10-05
ESRIJSON: add support for esriFieldTypeSingle and esriFieldTypeDate data types

rouault:


ogr/ogrsf_frmts/geojson/ogresrijsonreader.cpp
```diff
@@ -228,24 +229,45 @@ bool OGRESRIJSONReader::ParseField(json_object *poObj)
...
             eFieldType = OFTReal;
+            eFieldSubType = OFSTFloat32;
         }
-        else if (EQUAL(json_object_get_string(poObjType),
-                       "esriFieldTypeSmallInteger") ||
-                 EQUAL(json_object_get_string(poObjType),
-                       "esriFieldTypeInteger"))
+        else if (EQUAL(pszObjType, "esriFieldTypeDouble"))
+        {
+            eFieldType = OFTReal;
```
elpaso, line 247: no need to define the subtype here?  Is it 64b the default?

elpaso (approved):

ogr/ogrsf_frmts/geojson/ogresrijsonreader.cpp
rouault: 64bit is the default indeed

PR #8526, opened 2023-10-08
Doc: RasterIO(): document how overviews are selected

rouault:


rouault:
> just here to say this could also mention the warper app lib? (and the cog driver specifically, as opposed to the tile server drivers ... I think)

the warp lib has a slightly different logic (I guess it could be aligned to work with the arbitrary 1.2 ratio of RasterIO() if consistency was needed). Cf lines 2732-2755 of gdalwarp_lib.cpp if you want to document it.
The COG driver is a different beast, and its doc already explains how overviews are handled

mdsumner:
> > just here to say this could also mention the warper app lib? (and the cog driver specifically, as opposed to the tile server drivers ... I think)
> 
> the warp lib has a slightly different logic (I guess it could be aligned to work with the arbitrary 1.2 ratio of RasterIO() if consistency was needed). Cf lines 2732-2755 of gdalwarp_lib.cpp if you want to document it. The COG driver is a different beast, and its doc already explains how overviews are handled

ah, excellent thank you that's helpful

elpaso (approved):

PR #8531, opened 2023-10-09
gdal_translate: when specifying -srcwin, preserve source block size in the temporary VRT if srcwin top,left is a multiple of the block size

rouault:


frmts/vrt/vrtrasterband.cpp
```diff
@@ -649,11 +649,13 @@ CPLXMLNode *VRTRasterBand::SerializeToXML(const char *pszVRTPath)
     // serialized at the dataset level.
     if (dynamic_cast<VRTWarpedRasterBand *>(this) == nullptr)
     {
-        if (nBlockXSize != 128 && nBlockXSize != nRasterXSize)
+        if (nBlockXSize != 128 &&
```
elpaso, line 652: is this change related to the other one? Or just an optimization?

elpaso (approved):
LGTM

frmts/vrt/vrtrasterband.cpp
rouault: it is kind of related indeed. If doing gdal_translate -srcwin 0 0 width 2048 on a source raster with a block height of 2048, the blockYSize=2048 would be omitted on serialization of the VRT, and when reading back, it would use the default of 128, which is not desired. The added test case tests that fix as well

PR #8538, opened 2023-10-10
CI: add benchmark test suite

rouault:
- Uses [pytest-benchmark](https://pytest-benchmark.readthedocs.io/en/latest/)
- Add a autotest/benchmark module (now with only a few tests for the GTiff and GPKG driver)
- Add a new CI benchmark configuration, which builds the current tree and a reference tree, and run the benchmark test suite in both

### DIFF
--- a/.github/workflows/linux_build.yml
+++ b/.github/workflows/linux_build.yml
@@ -90,6 +90,13 @@ jobs:
             build_script: build.sh
             test_script: test.sh
 
+          - name: Ubuntu 20.04, benchmarks
+            id: benchmarks
+            travis_branch: ubuntu_2004
+            container: ubuntu_20.04
+            build_script: build.sh
+            test_script: test.sh
+
           - name: Ubuntu 20.04, Intel compiler
             id: icc
             container: icc
@@ -267,6 +274,13 @@ jobs:
             TEST_CMD="$(pwd)/.github/workflows/${{ matrix.id }}/${{ matrix.test_script }}"
           else
             TEST_CMD="ctest -V -j $(nproc)"
+          fi
+
+          if test "${{ matrix.id }}" = "benchmarks"; then
+            if test -f /sys/devices/system/cpu/intel_pstate/no_turbo; then
+              echo "Disable TurboBoost"
+              echo 1 | sudo tee /sys/devices/system/cpu/intel_pstate/no_turbo
+            fi
           fi
 
           # For cache
--- a/autotest/benchmark/test_ogr_gpkg.py
+++ b/autotest/benchmark/test_ogr_gpkg.py
@@ -1,0 +1,80 @@
+#!/usr/bin/env pytest
+# -*- coding: utf-8 -*-
+###############################################################################
+# $Id$
+#
+# Project:  GDAL/OGR Test Suite
+# Purpose:  Benchmarking of GeoPackage driver
+# Author:   Even Rouault <even dot rouault at spatialys.com>
+#
+###############################################################################
+# Copyright (c) 2023, Even Rouault <even dot rouault at spatialys.com>
+#
+# Permission is hereby granted, free of charge, to any person obtaining a
+# copy of this software and associated documentation files (the "Software"),
+# to deal in the Software without restriction, including without limitation
+# the rights to use, copy, modify, merge, publish, distribute, sublicense,
+# and/or sell copies of the Software, and to permit persons to whom the
+# Software is furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included
+# in all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+# OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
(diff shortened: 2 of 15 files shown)


rouault:
CC @dbaston

autotest/benchmark/test_ogr_gpkg.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+@pytest.fixture()
+def source_file(tmp_vsimem):
+    filename = str(tmp_vsimem / "test.gpkg")
+    create_file(filename)
+    return filename
+
+
+def test_ogr_gpkg_spatial_index(source_file):
```
dbaston, line 70: I think bringing in the `benchmark` fixture explicitly would let you more clearly exclude the setup cost. It's not clear to me if the runtime of `source_file` is part of the timing or not.

autotest/conftest.py
```diff
@@ -271,6 +271,23 @@ def pytest_report_header(config):
         if loaded_dlls:
             gdal_header_info += "\nLoaded shared objects:\n" + loaded_dlls
 
+    has_benchmark = False
+    for arg in sys.argv:
+        if "benchmark" in arg:
```
dbaston, line 276: This works if we call `pytest benchmark` from `autotest` but not if we call `pytest test_gtiff` from `autotest/benchmark`. See https://github.com/dbaston/gdal/commit/e1a77f48586a5416fe51b6a3bab3696dfe2c982d for an alternative that also writes a warning if we benchmark a debug build.

.github/workflows/linux_build.yml
```diff
@@ -269,6 +276,13 @@ jobs:
             TEST_CMD="ctest -V -j $(nproc)"
           fi
 
+          if test "${{ matrix.id }}" = "benchmarks"; then
```
dbaston, line 279: Move into `test.sh` ?

autotest/benchmark/test_ogr_gpkg.py
rouault: > It's not clear to me if the runtime of `source_file` is part of the timing or not.

it is not, comparing timings with the somewhat more verbose below version. source_file is run once as expected

```
def test_ogr_gpkg_spatial_index(source_file, benchmark):
    @benchmark
    def do():
        ds = ogr.Open(source_file)
        lyr = ds.GetLayer(0)
        lyr.SetSpatialFilterRect(1000, 1000, 10000, 10000)
        count = 0
        for f in lyr:
            count += 1
        assert count == 10000 - 1000 + 1
```

.github/workflows/linux_build.yml
rouault: I don't think this can be done under Docker. Anyway it doesn't look like this is even hit (perhaps because we already run in a container/VM that doesn't expose turboboost setting)

autotest/conftest.py
rouault: thanks. Picked up

rouault:
https://github.com/OSGeo/gdal/pull/8538/commits/d35198a3031e41a8f24702a3ef6a6c552b5b388b fails as expected:
```
 Performance has regressed:
	test_gtiff_byte (0001_ref) - Field 'mean' has failed PercentageRegressionCheck: 100.708833475 > 5.000000000
```

rouault:
changing to use min statistics as suggest by @DFEvans on the mailing list

rouault:
@dbaston I've written more tests, and it seems we now hit unreliability of timings. Initial failed run in https://github.com/OSGeo/gdal/actions/runs/6482985166/job/17603597567 using the naive procedure (running the reference test suite, and then the one of the PR)

I've attempted to modify the test procedure (cf updated benchmark/test.sh) to do more runs to attempt mitigating this, but given See https://github.com/OSGeo/gdal/actions/runs/6485533623/job/17611807035?pr=8538
this obviously doesn't work: the first comparison run has just one test that is slightly over my 5% arbitrary criterion, but the retry has 2 tests 11% and 21% slower.

Any ideas (except increasing significantly the tolerance threshold to let's say 30% to be hopefully robust, but we won't catch up small perf regressions) or are we hitting a dead end?
Hum looking at https://pytest-benchmark.readthedocs.io/en/stable/faq.html, maybe use an alternate timer : "You could use something like time.process_time (Python 3.3+ only) as the timer. Process time [doesn’t include sleeping or waiting for I/O](https://en.wikipedia.org/wiki/CPU_time).", assuming the difference in timings result in the VM being randomly scheduled out by the VM hypervisor.

rouault:
more stable timings with --benchmark-timer=time.process_time in https://github.com/OSGeo/gdal/actions/runs/6486353872/job/17614377148?pr=8538.

rouault:
A retry failed : https://github.com/OSGeo/gdal/actions/runs/6486353872/job/17617138599?pr=8538 .  Up to 16% of slowdown on one test. Bumping tolerance to 20% ...

rouault:
6 good runs in a row:
- https://github.com/OSGeo/gdal/actions/runs/6487684534/job/17618535517?pr=8538
- https://github.com/OSGeo/gdal/actions/runs/6487684534/job/17621599661?pr=8538
- https://github.com/OSGeo/gdal/actions/runs/6487684534/job/17623767294?pr=8538
- https://github.com/OSGeo/gdal/actions/runs/6487684534/job/17624447841?pr=8538
- https://github.com/OSGeo/gdal/actions/runs/6487684534/job/17637415371?pr=8538
- https://github.com/OSGeo/gdal/actions/runs/6487684534/job/17637999428?pr=8538

rouault:
merging this

PR #8539, opened 2023-10-11
OGR2OGR: fix preserve_fid and explodecollections incompatible options

### LINKED ISSUE #8523: Error in gdal.VectorTranslate for explodecollections=True + some weird python logging behaviour

This issue raises several things that could be improved:
1. When `gdal.VectorTranslate` is used on a gpkg file containing multipolygons with `explodeCollections=True`, the default behaviour is to preserve the fid's. However, when `explodecollections` is being applied, normally is will be impossible to preserve fid's because duplicate fid's will appear. So, an error is thrown.
2. The Exception thrown doesn't contain the error needed to understand what is happening.
3. There is some strange interference happening between using gdal.UseExceptions() and enabling debug logging (e.g. by setting os.environ["CPL_DEBUG"] = "ON"). More details below.

### Code snippet
```
import os
from osgeo import gdal

# There is some weird behaviour here:
#   - If only gdal.UseExceptions() is used, an exception is thrown, but not an
#     interesting one:
#   - If gdal.UseExceptions() + os.environ["CPL_DEBUG"] = "ON", the debug logging stops
#     being written too soon and no exception is thrown.
#   - If only os.environ["CPL_DEBUG"] = "ON", the debug logging contains all necessary
#     information, but obviously no exception is thrown.

gdal.UseExceptions()
# os.environ["CPL_DEBUG"] = "ON"

src = "https://github.com/geofileops/geofileops/raw/main/tests/data/polygon-parcel.gpkg"
dst = "/vsimem/output.gpkg"
dst = "C:/Temp/output.gpkg"
options = gdal.VectorTranslateOptions(explodeCollections=True)
ds_output = gdal.VectorTranslate(srcDS=src, destNameOrDestDS=dst, options=options)
```

### When gdal.UseExceptions() is called and "CPL_DEBUG" is not set

Like this, an exception is thrown, but it doesn't contain the error needed to understand the problem. 

The error thrown is `RuntimeError: sqlite3_exec(CREATE TRIGGER "trigger_delete_feature_count_parcels" AFTER DELETE ON "parcels" BEGIN UPDATE gpkg_ogr_contents SET feature_count = feature_count - 1 WHERE lower(table_name) = lower('parcels'); END;) failed: no such table: main.parcels`

The useful error to be thrown would be `ERROR 1: failed to execute insert : UNIQUE constraint failed: parcels.fid`

```
  File "C:\Users\Gebruiker\Documents\GitHub\pysnippets\pysnippets\bug_reproducing\gdal_error_explode.py", line 7, in <module>
    ds_output = gdal.VectorTranslate(srcDS=src, destNameOrDestDS=dst, options=options)
  File "C:\Tools\mambaforge\envs\pysnippets\lib\site-packages\osgeo\gdal.py", line 1381, in VectorTranslate
    return wrapper_GDALVectorTranslateDestName(destNameOrDestDS, srcDS, opts, callback, callback_data)
  File "C:\Tools\mambaforge\envs\pysnippets\lib\site-packages\osgeo\gdal.py", line 5438, in wrapper_GDALVectorTranslateDestName
    return _gdal.wrapper_GDALVectorTranslateDestName(*args)
RuntimeError: sqlite3_exec(CREATE TRIGGER "trigger_delete_feature_count_parcels" AFTER DELETE ON "parcels" BEGIN UPDATE gpkg_ogr_contents SET feature_count = feature_count - 1 WHERE lower(table_name) = lower('parcels'); END;) failed: no such table: main.parcels
```

### When gdal.UseExceptions() is NOT called but  "CPL_DEBUG" is set to "ON"

Like this, the logging is OK, but no exception thrown (logically).

```
...
ERROR 1: failed to execute insert : UNIQUE constraint failed: parcels.fid
ERROR 1: Unable to write feature 9 from layer parcels.
translation of layer parcels (use -skipfailures to skip errors)
GPKG: 9 rows inserted into rtree_parcels_geom
GPKG: Creating insert/delete feature_count triggers
ERROR 1: sqlite3_exec(CREATE TRIGGER "trigger_insert_feature_count_parcels" AFTER INSERT ON "parcels" BEGIN UPDATE gpkg_ogr_contents SET feature_count = feature_count + 1 WHERE lower(table_name) = lower('parcels'); END;) failed: no such table: main.parcels
ERROR 1: sqlite3_exec(CREATE TRIGGER "trigger_delete_feature_count_parcels" AFTER DELETE ON "parcels" BEGIN UPDATE gpkg_ogr_contents SET feature_count = feature_count - 1 WHERE lower(table_name) = lower('parcels'); END;) failed: no such table: main.parcels
...
```

### When gdal.UseExceptions() is called AND "CPL_DEBUG" is set to "ON"

With this combo, the logging stops being written before the interesting stuff happend and no exception is thrown.

```
...
GDALVectorTranslate: Using FID=fid and -preserve_fid
GPKG: ResetStatement(SELECT m."fid", m."geom", m."OIDN", m."UIDN", m."index", m."HFDTLT", m."LBLHFDTLT", m."GEWASGROEP", m."PM", m."LBLPM", m."LENGTE", m."OPPERVL", m."DATUM" FROM "parcels" m)
Warning 1: Non-conformant content for record 1 in column DATUM, 2020-05-01T00:00:00Z, successfully parsed
GPKG: 9 rows inserted into rtree_parcels_geom
GPKG: Creating insert/delete feature_count triggers
GDAL: GDALClose(C:/Temp/output.gpkg, this=000001EFB7BAE050)
GDAL: GDALClose(https://github.com/geofileops/geofileops/raw/main/tests/data/polygon-parcel.gpkg, this=000001EFB75DB2D0)
GDAL: In GDALDestroy - unloading GDAL shared library.
```

## Operating system

Windows 10

Version 3.7.2 installed via conda-forge.

elpaso:
Fixes #8523  (partial fix)


swig/include/python/gdal_python.i
```diff
@@ -2544,7 +2545,15 @@ def VectorTranslateOptions(options=None, format=None,
         and is only an afterwards conversion.
     explodeCollections:
         produce one feature for each geometry in any kind of geometry collection in the
-        source file, applied after any -sql option.
+        source file, applied after any -sql option. This option is not compatible with
```
rouault, line 2548: those pieces of doc should be pub in ogr2ogr.rst (in ogr2ogr.rst, the hint of using SQL like "SELECT fid AS original_fid, * FROM ..." can also be added if preservation of the FID is whished)

autotest/utilities/test_ogr2ogr_lib.py
```diff
@@ -2061,3 +2061,54 @@ def test_ogr2ogr_lib_reprojection_curve_geometries_output_does_not_support_curve
...
+        )
+
+        if preserveFID:
+            with pytest.raises(RuntimeError) as excinfo:
+                gdal.VectorTranslate(srcDS=src, destNameOrDestDS=dst, options=options)
+
+            assert (
+                str(excinfo.value)
+                == "cannot use -preserve_fid and -explodecollections at the same time."
+            )
```
rouault, line 2105:
```suggestion
            with pytest.raises(RuntimeError, match="cannot use -preserve_fid and -explodecollections at the same time"):
                gdal.VectorTranslate(srcDS=src, destNameOrDestDS=dst, options=options)
```

swig/include/python/gdal_python.i
rouault: should be put in ogr2ogr.rst *as well*

rouault:
indentation issue: https://github.com/OSGeo/gdal/actions/runs/6483504803/job/17605242006?pr=8539

rouault:
manually cherry-picked to 3.7 per 515eab0b4e0f9e79168fec6b0d72333adddb50e6

PR #8544, opened 2023-10-11
ogr2ogr: use Arrow interface in reading and writing when possible

rouault:
That is when:
- the input driver reports OLCFastGetArrowStream
- and the output driver reports OLCFastWriteArrowBatch
- and the input ArrowSchema is accepted by the target layer
- and there are basically no ogr2ogr options (other than -dsco, -lco, -limit, -where and -spat).
    
Note that the efficiency in case of -where and -spat can be less than
using the classic OGRFeature strategy when the source is Parquet and a
lot of features are discarded.

With that Parquet->Parquet or GPKG->Parquet translation can be up to 3x times faster.

$ time ogr2ogr out.parquet nz-building-outlines.parquet --config OGR2OGR_USE_ARROW_API NO

real    0m11,246s
user    0m10,934s
sys     0m0,974s

$ time ogr2ogr out.parquet nz-building-outlines.parquet

real    0m4,311s
user    0m3,968s
sys     0m0,889s

$ time ogr2ogr out.parquet nz-building-outlines.gpkg --config OGR2OGR_USE_ARROW_API NO

real    0m12,120s
user    0m11,360s
sys     0m0,764s

$ time ogr2ogr out.parquet nz-building-outlines.gpkg

real    0m3,853s
user    0m5,053s
sys     0m1,028s

### DIFF
--- a/apps/ogr2ogr_lib.cpp
+++ b/apps/ogr2ogr_lib.cpp
@@ -64,6 +64,7 @@
 #include "ogr_featurestyle.h"
 #include "ogr_geometry.h"
 #include "ogr_p.h"
+#include "ogr_recordbatch.h"
 #include "ogr_spatialref.h"
 #include "ogrlayerdecorator.h"
 #include "ogrsf_frmts.h"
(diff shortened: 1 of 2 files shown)


PostholerCom:
This is huge! Thanks Even!

PR #8566, opened 2023-10-16
OAPIF: bump default limit to 1000 and honor schema from API

### LINKED ISSUE #8522: OAPIF: Increase the default page size
The current default page size in the OGC API Features driver is 10. While this default value is also defined in the OGC API Features standard, for practical use it is all too small. See discussion in the mailing list thread https://lists.osgeo.org/pipermail/gdal-dev/2023-October/057758.html

I think that the larger the page size, and so the less requests needed for reading the data, the better. For that reason the page size could be the maximum limit that is defined in the API of the service. It is of course possible that a service provider has not considered what is a reasonable max limit but just blindly used the value 10000 as it appears in the standard https://docs.ogc.org/is/17-069r4/17-069r4.html#_parameter_limit. For my mind GDAL could still trust on the information that the /api endpoint of the service delivers.

elpaso:
Fix #8522

Supports internal reference schema and remote JSON components, external YAML components are not supported.


ogr/ogrsf_frmts/wfs/ogroapifdriver.cpp
```diff
@@ -83,7 +83,8 @@ class OGROAPIFDataset final : public GDALDataset
     CPLString m_osRootURL;
     CPLString m_osUserQueryParams;
     CPLString m_osUserPwd;
-    int m_nPageSize = 10;
+    int m_nPageSize = 1000;
+    int m_bPageSizeSetFromOpenOptions = false;
```
rouault, line 87:
```suggestion
    bool m_bPageSizeSetFromOpenOptions = false;
```

ogr/ogrsf_frmts/wfs/ogroapifdriver.cpp
```diff
@@ -853,6 +874,150 @@ bool OGROAPIFDataset::LoadJSONCollections(const CPLString &osResultIn)
...
+                    }
+                }
+
+                if (!bFound)
+                {
+                    // check $ref
+                    for (const auto &param : parameters)
+                    {
+                        const auto ref{param.GetString("$ref")};
+                        if (ref.find("limit") != std::string::npos)
```
rouault, line 921: we could also test that ".yml" and ".yaml" are not in the URL, to save an external download

ogr/ogrsf_frmts/wfs/ogroapifdriver.cpp
```diff
@@ -853,6 +874,150 @@ bool OGROAPIFDataset::LoadJSONCollections(const CPLString &osResultIn)
...
+            {
+
+                const auto parameters{path.GetArray("get/parameters")};
+
+                // loop through parameters to find "limit"
+                for (const auto &para : parameters)
+                {
+                    if (para.GetName() == "limit")
+                    {
+                        bFound = true;
```
rouault, line 911: I suspect something is missing here (and in tests), because nMaximum and nDefault are only set in the below !bFound case

rouault (commented):
- doc/source/drivers/vector/oapif.rst would need to be updated
- autotest/gdrivers/ogcapi.py is also impacted by those changes

ogr/ogrsf_frmts/wfs/ogroapifdriver.cpp
elpaso: dead leftover code, the case is handled below (I've removed the `if` as well).

ogr/ogrsf_frmts/wfs/ogroapifdriver.cpp
elpaso: done

ogr/ogrsf_frmts/wfs/ogroapifdriver.cpp
elpaso: done

elpaso:
> * doc/source/drivers/vector/oapif.rst would need to be updated

Done, but I realized that the documented maximum of 10000 is not enforced in code, should we?

rouault:
> I realized that the documented maximum of 10000 is not enforced in code, should we?

I think we should just remove that part in the documentation. I don't see a strong reason to put an upper bound

jratike80:
> I don't see a strong reason to put an upper bound

The OGCFeat server is always allowed to send less features than the requested maximum limit. If the maximum is 10000 and user is using &limit=20000 the server probably sends 10000 features, but it may as well send only 875 features. And the latter may happen as well with &limit=10000. Client must just follow the "next" links. So I agree there is no strong reason to put an upper bound.

rouault:
remaining issues are cppcheck issues and gdrivers/ogcapi.py::test_ogr_ogcapi_features failure

PR #8617, opened 2023-10-26
GetArrowStream(): support a GEOMETRY_METADATA_ENCODING=GEOARROW option (fixes #8605)

### LINKED ISSUE #8605: Add layer option to return geometry columns as GeoArrow extension types when using the ArrowArrayStream interface
Currently, the behaviour we get when using the (fantastic!) `ArrowArrayStream` interface to OGR is that geometry columns are returned as WKB or the preliminary version of GeoArrow that existed when the Arrow and Parquet drivers were first implemented. The current format that well-known binary is returned has the field metadata key `ARROW:extension:name` set to `ogc.wkb`.

With [GeoArrow extension types](https://github.com/geoarrow/geoarrow/blob/main/extension-types.md), it is possible to attach the CRS to each geometry column. Clients that have registered the GeoArrow extension types (e.g., using [geoarrow-pyarrow](https://github.com/geoarrow/geoarrow-python)) can then ingest the stream without any extra steps (e.g., currently you have to do some manual inference/assignment of metadata to get the CRS to propagate through the rest of the pyarrow machinery: https://github.com/geoarrow/geoarrow-python/blob/main/geoarrow-pyarrow/src/geoarrow/pyarrow/io.py#L34-L54).

Perhaps an option like `GEOMETRY_METADATA_ENCODING=GEOARROW` would be appropriate? If `GEOARROW`, the `ARROW:extension:name` would be `geoarrow.wkb` and `ARROW:extension:metadata` would be `{"crs": <projjson CRS representation>}`.

I think the encoding of the geometry (e.g., using a serialized layout like WKB or a GeoArrow layout) is orthogonal to this (and probably a good deal more work since it involves building up `ArrowArray` and/or `ArrowSchema` objects rather than just building up JSON metadata). Potentially those could be future options of the GEOMETRY_ENCODING layer option...it would be particularly useful for points (for which the per-feature WKB deserialization overhead can be particularly egregious).

rouault:


ogr/ogrsf_frmts/generic/ogrlayerarrow.cpp
```diff
@@ -542,12 +567,36 @@ OGRLayer::CreateSchemaForWKBGeometryColumn(const OGRGeomFieldDefn *poFieldDefn,
...
+        {
+            char *pszPROJJSON = nullptr;
+            poSRS->exportToPROJJSON(&pszPROJJSON, nullptr);
+            if (pszPROJJSON)
+            {
+                osExtensionMetadata = "{\"crs\":";
+                osExtensionMetadata += pszPROJJSON;
+                osExtensionMetadata += '}';
+                CPLFree(pszPROJJSON);
+            }
```
paleolimbot, line 584: Out of curiosity, what would the `else` branch here be? The CRS is present but can't be exported to PROJJSON? If that's the case, I wonder if this should error instead of dropping the CRS?

ogr/ogrsf_frmts/generic/ogrlayerarrow.cpp
```diff
@@ -5579,7 +5671,8 @@ static bool BuildOGRFieldInfo(
                             OGRParseArrowMetadata(pabyMetadata);
                         auto oIter = oMetadata.find(ARROW_EXTENSION_NAME_KEY);
                         if (oIter != oMetadata.end() &&
-                            oIter->second == EXTENSION_NAME_WKB)
+                            (oIter->second == EXTENSION_NAME_OGC_WKB ||
+                             oIter->second == EXTENSION_NAME_GEOARROW_WKB))
```
paleolimbot, line 5675: If the `ARROW:extension:metadata` field is set with a CRS, will this be ignored? (Apologies if that's not the context here)

paleolimbot (approved):
Thank you!

It seems like it will be the case after this PR, but if a stream with a GeoArrow extension column is passed to the writer, will it end up as a geometry column with the correct CRS?

ogr/ogrsf_frmts/generic/ogrlayerarrow.cpp
rouault: warning message added in the unlikely case (for PROJ >= 6.2 which is the minimum to support PROJJSON) exportToPROJJSON() fails

ogr/ogrsf_frmts/generic/ogrlayerarrow.cpp
rouault: cf general answer

rouault:
> but if a stream with a GeoArrow extension column is passed to the writer, will it end up as a geometry column with the correct CRS?

no. as documented in WriteArrowBatch(), "OGR fields for the corresponding children arrays must exist and be of a
compatible type. [...] For geometry fields, they should be created either implicitly at CreateLayer() type (if geom_type != wkbNone), or explicitly with CreateGeomField()."

We could possibly have a CreateGeomFieldFromArrowSchema() that would call CreateGeomField() with the appropriate CRS if finding the geoarrow.wkb extension with a crs, but that would only be limited to drivers that support CreateGeomField(), which doesn't include Shapefile or GeoJSON that must know the CRS at layer creation time (mostly because there are single geometry column formats, and that it is easier from an implementation point of view to require it at creation time than later)

rouault:
> Is it worth warning or erroring if the CRS embedded in the `ARROW:extension:metadata` doesn't match or is going to be dropped?

on the writing side you mean? I guess we could warn if OGRGeomFieldDefn::GetSpatialRef() != crs provided in schema::metadata. If you want to do it, that's welcome

PR #8620, opened 2023-10-27
gdal_calc.py: restore specifc --format behaviour (master only fix, fixes #8619)

### LINKED ISSUE #8619: gdal_calc : changed meaning of --format argument
In master the meaning of the --format argument to gdal_calc has changed from the previous/documented behaviour of setting the output file format to the generic GDAL command behaviour of outputting format details. This breaks existing scripts and differs from the documented behavior.

Looks like this was introduced by https://github.com/OSGeo/gdal/commit/3c6296edfafd0e053c2c5c5e0b2034a6c3bc1fb8 , as GDALArgumentParser is now used for gdal_calc argument processing .

rouault:


nyalldawson:
Thanks!

PR #8627, opened 2023-10-28
OGRLayer::FilterGeometry(): extend optimized code path

rouault:


ogr/ogrsf_frmts/generic/ogrlayer.cpp
```diff
@@ -1530,6 +1530,72 @@ int OGRLayer::InstallFilter(OGRGeometry *poFilter)
 }
 //! @endcond
 
+/************************************************************************/
+/*                   DoesGeometryIntersectEnvelope()                    */
+/************************************************************************/
+
+static bool DoesGeometryIntersectEnvelope(const OGRGeometry *poGeometry,
```
dbaston, line 1537: I think it would be more clear to call this something like `DoesGeometryHavePointInEnvelope` because it is not fully testing intersection.

rouault: good point. done

PR #8631, opened 2023-10-29
Arrow read/write: add GDAL specific field metadata

rouault:
CC @paleolimbot

### DIFF
--- a/ogr/ogrsf_frmts/generic/ogrlayerarrow.cpp
+++ b/ogr/ogrsf_frmts/generic/ogrlayerarrow.cpp
@@ -52,6 +52,7 @@ constexpr const char *MD_GDAL_OGR_DEFAULT = "GDAL:OGR:default";
 constexpr const char *MD_GDAL_OGR_SUBTYPE = "GDAL:OGR:subtype";
 constexpr const char *MD_GDAL_OGR_WIDTH = "GDAL:OGR:width";
 constexpr const char *MD_GDAL_OGR_UNIQUE = "GDAL:OGR:unique";
+constexpr const char *MD_GDAL_OGR_DOMAIN_NAME = "GDAL:OGR:domain_name";
 
 constexpr char ARROW_LETTER_BOOLEAN = 'b';
 constexpr char ARROW_LETTER_INT8 = 'c';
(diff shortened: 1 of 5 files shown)


ogr/ogrsf_frmts/generic/ogrlayerarrow.cpp
```diff
@@ -497,6 +497,78 @@ int OGRLayer::GetArrowSchema(struct ArrowArrayStream *,
...
+            offsetMD += sizeof(int32_t);
+            memcpy(pszMetadata + offsetMD, ARROW_EXTENSION_METADATA_KEY,
+                   strlen(ARROW_EXTENSION_METADATA_KEY));
+            offsetMD += static_cast<int>(strlen(ARROW_EXTENSION_METADATA_KEY));
+            *reinterpret_cast<int32_t *>(pszMetadata + offsetMD) =
+                static_cast<int32_t>(osExtensionMetadata.size());
+            offsetMD += sizeof(int32_t);
+            memcpy(pszMetadata + offsetMD, osExtensionMetadata.c_str(),
+                   osExtensionMetadata.size());
+            offsetMD += osExtensionMetadata.size();
```
paleolimbot, line 568: If you need/want a utility for this, you could copy the relevant pieces of https://github.com/apache/arrow-nanoarrow/blob/main/src/nanoarrow/schema.c#L1416-L1525 (or vendor + use the metadata builder).

paleolimbot (commented):
I think what you might want here is regular field metadata rather than an extension type? Regular field metadata might look like:

- Key: `GDAL:OGR:alternative_name` (with associated value)
- Key: `GDAL:OGR:comment` (with associated value)
- Key: `GDAL:OGR:default` (with associated value)
- Key: `GDAL:OGR:subtype` (with associated value)
- Key: `GDAL:OGR:width` (with associated value)

Rather than

- Key: `ARROW:extension:name` = `gdal.field_metadata`
- Key: `ARROW:extension:metadata` = `{"alternative_name": ... }`

An extension type might perhaps be better suited to a GDAL type that could not be represented losslessly by an Arrow type and had to be serialized to a binary column in some way and given an extension name + metadata.

I think what you are trying to represent here is true "field metadata": the Arrow types are accurate and the values can be interpreted safely according to the type of the array...there's just some extra information for consumers who want it.

Arrow C++ silently drops unknown extension types fairly aggressively but I don't think that behaviour is guaranteed and might be different in another implementation (I might be alone but I personally wish it would warn or error by default in those cases to prevent misinterpreting true extension types).

rouault:
@paleolimbot Thanks for the hints!

paleolimbot (approved):
Looks great!

PR #8642, opened 2023-10-31
PG: do not override search_path when not needed (fixes #8641)

### LINKED ISSUE #8641: ogr utilities do not respect the PostgreSQL search_path

I expect ogr utilities (ogrinfo, ogr2ogr) to use the [schema `search_path`](https://www.postgresql.org/docs/current/ddl-schemas.html#DDL-SCHEMAS-PATH) set in the target database if a schema or search path is not otherwise specified.

The use case is controlling the schema search path at the database level rather than at the clients.  This allows the database schemas to change (eg, in a blue/green deployment pattern) without having to change client code.

This worked fine until GDAL 3.1.0.

These examples connect to a local instance of PostgreSQL 16 where `search_path` is set to `"$user", vectors, sys, users, extensions`.

Using GDAL verison <= 3.0.4, the correct `search_path` is returned (and queries to tables in those schemas are successful):
```
# ogrinfo PG:"host=localhost dbname=foo user=bar password=password" -sql "show search_path"
...
OGRFeature(sql_statement):1
  search_path (String) = "$user", vectors, sys, users, extensions
```

However, when using GDAL >= 3.1.0, the `search_path` returned is always `public, extensions`:
```
# ogrinfo PG:"host=localhost dbname=foo user=bar password=password" -sql "show search_path"
...
OGRFeature(sql_statement):1
  search_path (String) = public, extensions
```

## Operating system

Ubuntu 20.04, 22.04

Docker images
* osgeo/gdal:ubuntu-small-3.0.4 (works as expected)
* osgeo/gdal:ubuntu-small-3.1.0 (fails as noted)
* osgeo/gdal:ubuntu-small-3.6.3 (fails as noted)

rouault:


elpaso (approved):
LGTM

PR #8643, opened 2023-11-01
autotest: make more tests run independently

dbaston:
## What does this PR do?

Addresses remaining issues so that tests within individual test files exercised by the Ubuntu 20.04 CI configuration can run in random order, except for `gnm` and `utilities/gdal2tiles`.

This does _not_ mean that we can run `pytest --random-order autotest` because there are a handful of cases of inter-module interference -- where running `test_driver2.py` before `test_driver1.py` causes something in `test_driver1.py` to fail. (Examples in https://github.com/dbaston/gdal/actions/runs/6714087188/job/18246767524#step:16:3630)

#4407

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/autotest/conftest.py
+++ b/autotest/conftest.py
@@ -136,7 +136,9 @@ def pytest_collection_modifyitems(config, items):
         for mark in item.iter_markers("require_driver"):
             driver_name = mark.args[0]
             if driver_name not in drivers_checked:
-                driver = gdal.GetDriverByName(driver_name)
+                driver = gdal.GetDriverByName(driver_name) or ogr.GetDriverByName(
+                    driver_name
+                )
                 drivers_checked[driver_name] = bool(driver)
                 if driver:
                     # Store the driver on gdaltest module so test functions can assume it's there.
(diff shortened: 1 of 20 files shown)


rouault:
gcore/vsioss.py::test_vsioss_6  fails repeatdly on Fedora Rawhide and Alpine

dbaston:
> gcore/vsioss.py::test_vsioss_6 fails repeatdly on Fedora Rawhide and Alpine

Oddly, I could not reproduce this locally using latest Docker images. I "fixed" it here by changing the webserver fixture from function to module-scope (a better design anyway) but I'm not sure what was causing the failure or why it was only showing up in those environments. Maybe there is some issue with repeatedly starting and shutting down the webserver.

rouault:
> Maybe there is some issue with repeatedly starting and shutting down the webserver.

We have always add some random test failures in /vsis3, /vsioss that go away when rerunning the tests. Not sure if it is a bug in the operational code or some issues with the test infrastructure

PR #8644, opened 2023-11-01
Documentation: Update mamba installation instructions in download.rst

gregorywaynepower:
## What does this PR do?

- Makes sure mamba is installed in base conda environment
- Uses installation instructions based on their current stable version

Updates the mamba installation instructions according to [mamba's documentation for their stable release](https://mamba.readthedocs.io/en/stable/mamba-installation.html#existing-conda-install-not-recommended). I understand the caveat to the download instructions is that they are assuming that a user already has conda installed instead of starting from a fresh environment.

No issues have been opened for this. Would you like for me to include/add a section for if a user is wanting to install using `mamba` from an environment where conda isn't already installed? If so, I'd probably write up instructions for their [Miniforge3](https://github.com/conda-forge/miniforge#miniforge3) installer.

### DIFF
--- a/doc/source/download.rst
+++ b/doc/source/download.rst
@@ -315,8 +315,8 @@ GDAL master Conda builds
 GDAL master builds are available in the `gdal-master <https://anaconda.org/gdal-master/gdal>`__
 channel. They are based on dependencies from the ``conda-forge`` channel.
 
-First, install mamba into the ``base`` environment, then create a dedicated ``gdal_master_env`` 
-environment, activate the dedicated ``gdal_master_env`` environment.
+First, install mamba into the ``base`` environment, create a dedicated ``gdal_master_env`` 
+environment, and then activate the dedicated ``gdal_master_env`` environment.
 
 ::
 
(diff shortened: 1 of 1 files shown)


rouault:
why is this change needed? what's wrong with installing mamba in the new environment ? Seems to me like a matter of personal taste

gregorywaynepower:
@rouault No worries, I appreciate the questions. Installing mamba outside of a base environment can cause conflicts with dependencies between conda's default channel and the dependencies required by mamba in the conda-forge channel when it's installed _inside_ an existing conda environment, resulting in a user's existing conda environment being broken.

doc/source/download.rst
```diff
@@ -321,9 +321,9 @@ the ``mamba`` package manager.
 ::
 
     conda update -n base -c conda-forge conda
+    conda install -n base --override-channels -c conda-forge mamba 'python_abi=*=*cp*'
```
rouault, line 324: the sentence above that block should be updated to revise the order in which actions are done

PR #8648, opened 2023-11-02
Add RFC96 text: Deferred C++ plugin loading

rouault:
Partially rendered view [here](https://github.com/rouault/gdal/blob/rfc96_text/doc/source/development/rfc/rfc96_deferred_plugin_loading.rst)

### DIFF
--- a/doc/source/development/rfc/rfc96_deferred_plugin_loading.rst
+++ b/doc/source/development/rfc/rfc96_deferred_plugin_loading.rst
@@ -8,7 +8,7 @@ RFC 96: Deferred C++ plugin loading
 Author:        Even Rouault
 Contact:       even.rouault @ spatialys.com
 Started:       2023-Nov-01
-Status:        In development
+Status:        Adopted
 Target:        GDAL 3.9
 ============== =============================================
 
@@ -496,4 +496,4 @@ Related issues and PRs
 Voting history
 --------------
 
-TBD
++1 from PSC members KurtS, HowardB, JukkaR, JavierJS and EvenR
--- a/doc/source/development/rfc/rfc96_deferred_plugin_loading.rst
+++ b/doc/source/development/rfc/rfc96_deferred_plugin_loading.rst
@@ -1,7 +1,7 @@
 .. _rfc-96:
 
 ==================================================================
-RFC 96: Deferred in-tree C++ plugin loading
+RFC 96: Deferred C++ plugin loading
 ==================================================================
 
 ============== =============================================
@@ -15,10 +15,14 @@ Target:        GDAL 3.9
 Summary
 -------
 
-This RFC adds a mechanism to defer the loading of in-tree C++ plugin drivers to
+This RFC adds a mechanism to defer the loading of C++ plugin drivers to
 the point where their executable code is actually needed, and converts a number
 of relevant drivers to use that mechanism. The aim is to allow for more modular
 GDAL builds, while improving the performance of plugin loading.
+
+It mostly targets in-tree plugin drivers, but it also provides a way for
+out-of-tree plugin drivers to benefit from deferred loading capabilities,
+provided libgdal is built in a specific way
 
 Context and motivation
 ----------------------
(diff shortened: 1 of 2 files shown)


doc/source/development/rfc/rfc96_deferred_plugin_loading.rst
```diff
@@ -0,0 +1,449 @@
...
+defined in a C++ file that does not depend on any external library.
+Similarly for the GetSubdatasetInfoFunc() optional method.
+
+When loading the actual driver, the GDALPluginDriverProxy::GetRealDriver()
+method will check that all information set in its metadata is
+consistent with the actual metadata of the underlying driver, and will warn
+when there are differences.
+
+GDALDataset::Open(), Create(), CreateCopy() methods are modified to not use
+directly the pfnOpen, pfnCreate, pfnCreateCopy callbacks (that would be the ones
```
elpaso, line 158: Is this change really necessary? Wouldn't be possible to delegate the loading to the pfn* in the proxy, maybe a generic function of the proxy that will load (once) the real driver and then call the real driver's pfn* ?

elpaso (commented):
Looks good to me. 
A big improvement. 
I'd probably liked even more if  the sidecar file solution would be implemented, but  I guess that if that would work well with exposing metadata, for identify and subdataset it must necessarily be done in C++ code.

doc/source/development/rfc/rfc96_deferred_plugin_loading.rst
rouault: I acknowledge that this is not super pretty, but it is intended to be GDAL core low-level internal detail. The issue is for example that GDALDriver::Create() currently calls pfnCreate on the proxy driver, but it doesn't know that it is a proxy driver. When the proxy driver has not loaded yet the real driver, the pfnCreate callback is null... Hence this GetCreateCallback() virtual method implemented in the proxy driver that takes care of loading the real driver and setting the pfnCreate callback of the proxy with the value of the pfnCreate of the real driver.  The alternative would be that each place that accesses the callbacks checks if the driver object derives from GDALPluginDriverProxy to force it to load the real driver, but I don't find that solution prettier.

tbonfort:
LGTM given that the scope is limited to deferred loading of **in-tree** drivers, but this has limited utility in the scope of rfc85 where we try to keep large/external/proprietary drivers out of tree

I'm wondering if it would be possible to obtain the same outcome with completely out-of-tree drivers, using some kind of code-generation at cmake time:
- cmake accepts `-DADD_DEFERRED_PLUGIN_FOO=/path/to/foodriver/core.cpp` with core.cpp containing the well-known functions `FOODatasetIdentify`, `FOODriverSetCommonMetadata` and `DeclareDeferredFOOPlugin` (in the same way that for traditional plugins we expect a gdal_foo.so file to contain a `GDALRegister_foo()` function)
- add the `/path/to/foodriver/core.cpp` to the list of files to be compiled into the main libgdal.so library
- use code generation to create the `DeclareDeferredPlugins()` function that will be called by `GDALAllRegister()`:
```c
// this file is automatically generated and will be overwritten, do not edit manually
void DeclareDeferredPlugins() {
  DeclareDeferredFOOPlugin(); //from -DADD_DEFERRED_PLUGIN_FOO
  DeclareDeferredBARPlugin(); //from -DADD_DEFERRED_PLUGIN_BAR
}
```

rouault:
> I'm wondering if it would be possible to obtain the same outcome with completely out-of-tree drivers

good point. Turns out to be quite straightforward. I've added a new commit in the RFC with that capability and candidate implementation in https://github.com/rouault/gdal/commit/8f880cef1367bb1a9245697e80557e3d80cfe570

rouault:
Candidate implementation available in https://github.com/OSGeo/gdal/pull/8695

doc/source/development/rfc/rfc96_deferred_plugin_loading.rst
```diff
@@ -0,0 +1,499 @@
...
+dependencies (libparquet, etc) and that would substantially increase the size of
+the core libgdal library, or which have licenses more restrictive than the MIT
+license used by libgdal.
+
+However, there is a penalty at GDALAllRegister() time. For example, on Linux,
+it takes ~ 300 ms to complete for a build with 126 plugins, which is a substantial
+time for short lived GDAL-based processes (for example a script which would run
+gdalinfo or ogrinfo on many files). This time is entirely spent in the dlopen()
+method of the operating system and there is, to the best of our knowledge,
+nothing we can do reduce it... besides limiting the amount of dynamic loading
```
schwehr, line 53: Mising "to": "nothing we can do to reduce it"

rouault: added

PR #8663, opened 2023-11-04
Fix arrow repo URLs to reference main branch instead of master

amoeba:
## What does this PR do?

Fixes URLs in the docs to the Apache Arrow GitHub repository that reference the master branch instead of the current main branch. The master branch was renamed to main in 2022 and, while GitHub helpfully redirects outdated links, I think it would be good for these docs to reference the new branch.

None.


amoeba:
Thanks @rouault!

Issue #8668, opened 2023-11-06
[feature request] SELECT * EXCEPT (for complicated situations)


tobwen:
## tl;dr
Please add a 
- `-excludeField "field1","field2"` or
- `-sql 'SELECT * EXCEPT ("field1", "field2")'` (at least for _OGRSQL_) or
- `-select '!"field1", !"field2"'` (at least for _OGRSQL_)

## summary
I'm converting a multi-layer dataset (MVT) into GPKG. The columns/fields per layer are not equal, but contain at least one equal column, which should be dropped. Since there are more than 50 layers and up to 80 columns, it would be pretty complicated to allowlist create a VRT or similar solution.

For OGRSQL, this might be a nice addition and wouldn't mess with compatibility.

jratike80:
There is already a similar open ticket about this.  While having "except" would be handy, standard SQL does not support it and implementation might get pretty complicated. But maybe you could prototype how you would do it with plain SQL.

tobwen:
> There is already a similar open ticket about this. While having "except" would be handy, standard SQL does not support it and implementation might get pretty complicated. But maybe you could prototype how you would do it with plain SQL.

That's why I mentioned OGRSQL. Is the language tied to standard SQL? I thought you had the freedom to go other ways and implement new functions. There are currently a number of modern databases, such as DuckDB or Clickhouse, which support standard SQL but offer their own extensions. These extensions are not contradictory, as they are simply not used in standard SQL.

I am surprised that the implementation is difficult. Isn't a column label required for each vector format? The column label must be read in somewhere and then skipped by index or string comparison.

In PostgreSQL, I do this either via complex JSONb objects or by querying the table description. However, both are out of the question for most formats (SQLite is unlikely to support either).

rouault:
https://stackoverflow.com/a/58700908 points to extensions adopted by Databricks and DuckDB, that could be reasonable to borrow
```
SELECT * EXCEPT(ColumnNameX, ColumnNameY, ...) FROM tablename
SELECT * EXCLUDE(ColumnNameX, ColumnNameY, ...) FROM tablename
```

jratike80:
The other ticket is https://github.com/OSGeo/gdal/issues/854 and it seems to have a predecessor in the old GDAL trac, so this one has a long a history.
[...]

PR #8675, opened 2023-11-07
OGRSQL: Add SELECT * EXCLUDE (...)

dbaston:
## What does this PR do?

Add `SELECT * EXCLUDE (...)` as described in #8668.

Still need to address some memory leaks. 

#8668

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/autotest/ogr/ogr_sql_test.py
+++ b/autotest/ogr/ogr_sql_test.py
@@ -1527,7 +1527,7 @@ def test_ogr_sql_attribute_filter_on_top_of_non_forward_where_clause(dialect):
 # Test min/max on string fields
 
 
-def test_ogr_sql_min_max_string_field(data_ds):
+def test_ogr_sql_min_max_string_field():
 
     mem_ds = ogr.GetDriverByName("Memory").CreateDataSource("")
     mem_lyr = mem_ds.CreateLayer("test")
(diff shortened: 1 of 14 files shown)


autotest/ogr/ogr_sql_test.py
```diff
@@ -1551,3 +1551,173 @@ def test_ogr_sql_min_max_string_field(data_ds):
...
+
+        assert sql_lyr.GetLayerDefn().GetGeomFieldCount() == 2
+
+
+def test_ogr_sql_select_except_named_geometry():
+
+    mem_ds = ogr.GetDriverByName("Memory").CreateDataSource("my_ds")
+    mem_lyr = mem_ds.CreateLayer("my_layer")
+
+    mem_lyr.GetLayerDefn().GetGeomFieldDefn(0).SetName("named_geom")
```
rouault, line 1633: Changing the name (or other property) of a field or geometry field after it has been created is in the unspecified behaviour category
should rather be something along:
```suggestion
    mem_lyr = mem_ds.CreateLayer("my_layer", geom_type = ogr.wkbNone)
    mem_lyr.CreateGeomField(ogr.GeomFieldDefn("named_geom", ogr.wkbUnknown))
```

ogr/swq_select.cpp
```diff
@@ -1221,7 +1240,7 @@ CPLErr swq_select::parse(swq_field_list *field_list,
         }
     }
 
-    /* -------------------------------------------------------------------- */
+    //    /* --------------------------------------------------rc_------------------ */
```
rouault, line 1243:
```suggestion
    /* -------------------------------------------------------------------- */
```

rouault (commented):
Very cool! We'd need a bit of documentation in https://gdal.org/user/ogr_sql_dialect.html (with a note that it is available *only* in OGR SQL currently, and not in the SQLite or most other native RBDMS native SQL)

dbaston:
It looks like the tests are failing because of warnings in the generated `swq_parser.cpp`, e.g. https://github.com/OSGeo/gdal/actions/runs/6891452376/job/18746536270?pr=8675#step:8:2538 and https://github.com/OSGeo/gdal/actions/runs/6871181938/job/18687534919?pr=8675#step:5:15. I guess I should modify the generated file to avoid the warnings?

rouault:
> I guess I should modify the generated file to avoid the warnings?

that or, suppress the MSVC C4505 warning in CMakeLists.txt for that file + ignore the cppcheck warning in scripts/cppcheck.sh

rouault:
> that or, suppress the MSVC C4505 warning in CMakeLists.txt for that file + ignore the cppcheck warning in scripts/cppcheck.sh

@dbaston Just curious how you fixed those issues as I don't see any special hacks in CMakeLists.txt/cppchecks.h

dbaston:
I edited the generated `swq_parser.cpp` to remove an always-true assert and a static function that was never called.

rouault:
> I edited the generated `swq_parser.cpp` to remove an always-true assert and a static function that was never called.

ok, we'll have to redo that the next time we regenerate the file, but that might be in many years :-) so OK as a resolution

PR #8683, opened 2023-11-08
/vsicurl: add ANYSAFE & BEARER to auth methods

drons:


doc/source/user/configoptions.rst
```diff
@@ -829,6 +829,19 @@ Networking options
...
 
+-  .. config:: GDAL_GSSAPI_DELEGATION
+      :since: 3.3
+      :choices: NONE, POLICY, ALWAYS
+
+      Set allowed GSS-API delegation. Relevant only with
+      :config:`GDAL_HTTP_AUTH=NEGOTIATE`.
+
+-  .. config:: GDAL_HTTP_BEARER
+      :since: 3.8
```
rouault, line 840:
```suggestion
      :since: 3.9
```

doc/source/user/configoptions.rst
```diff
@@ -829,6 +829,19 @@ Networking options
...
+
+      Set HTTP OAuth 2.0 Bearer Access Token to use for the connection. Must be used
```
rouault, line 842: Is that enough ? I'm a bit confused by https://curl.se/libcurl/c/CURLOPT_HTTPAUTH.html which mentions "You can set the Bearer token to use with [CURLOPT_XOAUTH2_BEARER](https://curl.se/libcurl/c/CURLOPT_XOAUTH2_BEARER.html)."

drons: I'll add a test to  [vsicurl.py](https://github.com/OSGeo/gdal/blob/master/autotest/gcore/vsicurl.py) a little later to remove doubts

drons: To successful authorization we must provide both `GDAL_HTTP_AUTH=BEARER` and `GDAL_HTTP_BEARER`. If we comment out `unchecked_curl_easy_setopt(http_handle, CURLOPT_HTTPAUTH, CURLAUTH_BEARER);` than authorization will fail, even if we setup `CURLOPT_XOAUTH2_BEARER`.

drons:
Thanks!

PR #8692, opened 2023-11-10
Add workflow that signals rasterio

### LINKED ISSUE #8688: Post to rasterio/dispatches endpoint when a new GDAL tag is pushed
So that rasterio will build and test release candidates. See also https://github.com/rasterio/rasterio/issues/2960.

sgillies:
So that rasterio will immediately build against GDAL release candidates and provide some feedback. Work on the rasterio side is here https://github.com/rasterio/rasterio/pull/2967.

I'm not sure how this can be generalized to other projects. I suspect QGIS is doing something smarter and better, but I felt like it was good to learn about using these GitHub ways.

cc @rouault @dbaston @nyalldawson 

Resolves #8688.

### DIFF
--- a/.github/workflows/rasterio_dispatch.yml
+++ b/.github/workflows/rasterio_dispatch.yml
@@ -1,0 +1,21 @@
+name: Rasterio dispatch 
+on:
+  push:
+    tags:
+      - '*'
+
+jobs:
+  dispatch:
+    name: Rasterio Dispatch
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - name: POST
+        run: |
+          curl -L \
+            -X POST \
+            -H "Accept: application/vnd.github+json" \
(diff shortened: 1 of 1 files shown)


rouault:
I'm OK with that. It seems however odd to me that github doesn't provide a more straightforward way of doing that which doesn't involve doing changes in the upstream repository. github should have in theory all the information needed so a downstream project can request being triggered by an event in another github repo...

sgillies:
@rouault I expected that GitHub would have a better solution, and maybe they do in the form of installable actions. To allow one project to subscribe to events in any other public project would require a global event stream, yeah?

An alternative to this: write a rasterio cron job that used the GitHub API to search for new GDAL tags. Or implement a subscription to https://github.com/osgeo/gdal/tags.atom. I'll give that some thought. It's a little more work to track seen tags or otherwise detect new tags in the feed, but it would remove the need for secrets or a new GDAL workflow.

sgillies:
Not needed!

PR #8695, opened 2023-11-12
RFC 96 implementation: Deferred C++ plugin loading

rouault:
Implements https://github.com/OSGeo/gdal/pull/8648

- All drivers that can be built as plugins and depends on external libraries have been converted to use the deferred plugin loading capability. Plus a few other drivers that only depend on core
- The PNG, JPEG, GIF, MRF, NITF and planetary (PDS, PDS4, ISIS2, ISIS3, VICAR) drivers can also now be built as plugins (when not using the in-tree vendored libraries such as libpng, libjpeg, etc.), and use deferred plugin loading
- The "raw" drivers can also now be built as plugins, but *not* as deferred plugin loading (doable, but significant effort required)

On my full configuration with all drivers that can be built as plugins effectively built as plugins (that is those that only depend on libgdal and those that depend on third-party libraries, with -DGDAL_ENABLE_PLUGINS=ON and -DGDAL_ENABLE_PLUGINS_NO_DEPS=ON), the initialization time has dropped from ~300 ms to 70 ms. If building only as plugins drivers that depend on third-party libraries with -DGDAL_ENABLE_PLUGINS=ON (which is the config which makes most sense), it drops to 26 ms, with all 55 plugins in deferred loading mode. One of the biggest culprit identified during the conversion process was the DGNv8 / DWG driver with the gazillions shared libraries of the Teigha SDK.


gcore/gdaldrivermanager.cpp
```diff
@@ -508,6 +524,43 @@ int GDALDriverManager::RegisterDriver(GDALDriver *poDriver)
...
+    {
+        if (m_oMapRealDrivers.find(poDriver->GetDescription()) !=
+            m_oMapRealDrivers.end())
+        {
+            CPLError(
+                CE_Failure, CPLE_AppDefined,
+                "RegisterDriver() in m_bInDeferedDriverLoading: %s already "
+                "registered!",
+                poDriver->GetDescription());
+            delete poDriver;
```
elpaso, line 537: just wondering if poDriver could be automatically managed with a unique_ptr (and use std::move later).

gcore/gdaldrivermanager.cpp
```diff
@@ -1146,6 +1219,496 @@ void GDALDriverManager::ReorderDrivers()
...
+GDALPluginDriverProxy::GDALPluginDriverProxy(
+    const std::string &osPluginFileName)
+    : m_osPluginFileName(osPluginFileName)
+{
+}
+
+//! @cond Doxygen_Suppress
+GDALDriver::OpenCallback GDALPluginDriverProxy::GetOpenCallback()
+{
+    auto poRealDriver = GetRealDriver();
```
elpaso, line 1239: I'm not  big fan of macros but maybe here it could be used to reduce duplicated checks, maybe something like ENSURE_REAL_DRIVER

gcore/gdaldrivermanager.cpp
```diff
@@ -1146,6 +1219,496 @@ void GDALDriverManager::ReorderDrivers()
...
+/************************************************************************/
+/*                           GetRealDriver()                            */
+/************************************************************************/
+
+GDALDriver *GDALPluginDriverProxy::GetRealDriver()
+{
+    if (m_osPluginFullPath.empty())
+        return nullptr;
+
+    CPLMutexHolderD(&hDMMutex);
```
elpaso, line 1396: should the mutex be moved before the  m_osPluginFullPath.empty() check?

gcore/gdaldrivermanager.cpp
```diff
@@ -1146,6 +1219,496 @@ void GDALDriverManager::ReorderDrivers()
...
+    }
+
+    const char *pszGDAL_DRIVER_PATH =
+        CPLGetConfigOption("GDAL_DRIVER_PATH", nullptr);
+    if (pszGDAL_DRIVER_PATH == nullptr)
+        pszGDAL_DRIVER_PATH = CPLGetConfigOption("OGR_DRIVER_PATH", nullptr);
+
+    /* ---------------------------------------------------------------- */
+    /*      Allow applications to completely disable this search by     */
+    /*      setting the driver path to the special string "disable".    */
```
elpaso, line 1612: I didn't check but I assume this has been added to the documentation.

gcore/gdaldriver.cpp
```diff
@@ -2760,6 +2785,14 @@ CPLErr GDALDriver::SetMetadataItem(const char *pszName, const char *pszValue,
         {
             GDALMajorObject::SetMetadataItem(GDAL_DMD_EXTENSIONS, pszValue);
         }
+        /* and vice-verca if there is a single extension in GDAL_DMD_EXTENSIONS */
```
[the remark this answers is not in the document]
rouault: fixed

gcore/gdaldrivermanager.cpp
rouault: That would be good indeed, but GDALDriverManager::RegisterDriver() is an existing public method used by all drivers. Changing its signature would affect all existing in-tree and out-of-tree driver. So I'm leaving that out

gcore/gdaldrivermanager.cpp
rouault: Good suggestion. Code now looks like:
```
#define DEFINE_DRIVER_METHOD_GET_CALLBACK(method_name, output_type)            \
    GDALDriver::output_type GDALPluginDriverProxy::method_name()               \
    {                                                                          \
        auto poRealDriver = GetRealDriver();                                   \
        if (!poRealDriver)                                                     \
            return nullptr;                                                    \
        return poRealDriver->method_name();                                    \
    }

DEFINE_DRIVER_METHOD_GET_CALLBACK(GetOpenCallback, OpenCallback)
DEFINE_DRIVER_METHOD_GET_CALLBACK(GetCreateCallback, CreateCallback)
DEFINE_DRIVER_METHOD_GET_CALLBACK(GetCreateMultiDimensionalCallback,
                                  CreateMultiDimensionalCallback)
DEFINE_DRIVER_METHOD_GET_CALLBACK(GetCreateCopyCallback, CreateCopyCallback)
DEFINE_DRIVER_METHOD_GET_CALLBACK(GetDeleteCallback, DeleteCallback)
DEFINE_DRIVER_METHOD_GET_CALLBACK(GetRenameCallback, RenameCallback)
DEFINE_DRIVER_METHOD_GET_CALLBACK(GetCopyFilesCallback, CopyFilesCallback)
```

gcore/gdaldrivermanager.cpp
rouault: no, this is not needed.

I've just added the following comment:
```
    // No need to take the mutex has this member variable is not modified
    // under the mutex.
```

gcore/gdaldrivermanager.cpp
rouault: yes, this is documented in https://gdal.org/api/gdaldriver_cpp.html#_CPPv4N17GDALDriverManager15AutoLoadDriversEv . But this was missing in https://gdal.org/user/configoptions.html#driver-management. Added there too

rouault:
Following comment added in GDALOpenEx():
```
    // Loading of drivers can involve up to 2 passes:
    // - in the first pass, all drivers that are compabile of the request mode
    //   (raster/vector/etc.) are probed using their Identify() method if it
    //   exists. If the Identify() method returns FALSE, the driver is skipped.
    //   If the Identify() methods returns GDAL_IDENTIFY_UNKNOWN and that the
    //   driver is a deferred-loading plugin, it is added to the
    //   apoSecondPassDrivers list for potential later probing, and execution
    //   continues to the next driver in the list.
    //   Otherwise if Identify() returns non-FALSE, the Open() method is used.
    //   If Open() returns a non-NULL dataset, the loop stops and it is
    //   returned. Otherwise looping over remaining drivers continues.
    // - the second pass is optional, only if at least one driver was added
    //   into apoSecondPassDrivers during the first pass. It is similar
    //   to the first pass except it runs only on apoSecondPassDrivers drivers.
    //   And the Open() method of such drivers is used, causing them to be
    //   loaded for real.
```

PR #8710, opened 2023-11-14
Cxx17 tunings

rouault:
- various CMakeLists.txt: remove now useless target_compile_features(cxx_std_17)
- Doc: update requirements to C++17
- Replace uses of cpl::make_unique<> with std::make_unique<>
- cpl_conv.h: remove cpl::make_unique<> template


doc/source/faq.rst
```diff
@@ -71,7 +71,7 @@ See :ref:`software_using_gdal`:
 What compiler can I use to build GDAL/OGR?
 ++++++++++++++++++++++++++++++++++++++++++++++++
 
-GDAL/OGR can be compiled with a C++11 capable compiler.
+GDAL/OGR must be compiled with a C++17 capable compiler since GDAL 3.9 (C++11 in previous versions)
```
dbaston, line 74: For some reason I'm not seeing the "Suggest changes" button, but I would suggest

```
Build requirements are described in :ref:`build_requirements`.
```

rouault: Good suggestion. Done

> For some reason I'm not seeing the "Suggest changes" button, but I would suggest

Perhaps because you tried from a commit and not for the full diff of the full request ?

PR #8726, opened 2023-11-16
[gpkg] Fix adding field comments after alternative name

nyalldawson:
Fix constraints on gpkg_data_columns are violated when AlterFieldDefn is called in some circumstances, eg when a comment is added after an alternative name has already been set on the field.

### DIFF
--- a/autotest/ogr/ogr_gpkg.py
+++ b/autotest/ogr/ogr_gpkg.py
@@ -3474,7 +3474,9 @@ def test_ogr_gpkg_36(tmp_vsimem, tmp_path):
     dbname = tmp_vsimem / "ogr_gpkg_36.gpkg"
     ds = gdaltest.gpkg_dr.CreateDataSource(dbname)
     lyr = ds.CreateLayer("test", geom_type=ogr.wkbPolygon)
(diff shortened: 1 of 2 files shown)


nyalldawson:
Thanks @rouault , fixed

rouault:
I've seen this is is likely related to https://github.com/qgis/QGIS/pull/55304 . How does the GeoPackage driver reacts when having duplicated aliases: does it error out cleanly and return things in a consistent state?

nyalldawson:
@rouault 
> I've seen this is is likely related to https://github.com/qgis/QGIS/pull/55304

Not really related -- you can easily trigger the bug from raw gdal api by setting an alternative name, and then setting a comment in a second step. Eg from qgis if you set alias/comment via the GUI then it'll be two atomic operations like this, and the second will error out with the constraint violation.

> How does the GeoPackage driver reacts when having duplicated aliases: does it error out cleanly and return things in a consistent state?

It errors out when trying to set the second field alias, with a message about the failed constraint. It's a relatively technical error message (it's the raw constraint violation message), but it DOES give sufficient information to see what's happening. Nothing is changed, so the GPKG remains valid.

nyalldawson:
Thanks @rouault !

PR #8733, opened 2023-11-16
[RFC 97 implementation] OGRFeatureDefn, OGRFieldDefn and OGRGeomFieldDefn "sealing"

rouault:


ogr/ogrsf_frmts/geojson/ogrgeojsonreader.cpp
```diff
@@ -906,7 +911,11 @@ void OGRGeoJSONReader::ReadLayer(OGRGeoJSONDataSource *poDS,
         poSRS = new OGRSpatialReference();
         bDefaultSRS = true;
     }
-    poLayer->GetLayerDefn()->GetGeomFieldDefn(0)->SetSpatialRef(poSRS);
+    {
+        auto poGeomFieldDefn = poLayer->GetLayerDefn()->GetGeomFieldDefn(0);
+        auto oTemporaryUnsealer(poGeomFieldDefn->GetTemporaryUnsealer());
```
nyalldawson, line 916: This could potentially be made a little simpler -- I'm thinking something like

```suggestion
        whileUnsealing(poGeomFieldDefn)->SetSpatialRef(poSRS));
```

Where "whileUnsealing" is a template based function that handles the GetTemporaryUnsealer logic. Something akin to QGIS' whileBlocking method (https://github.com/qgis/QGIS/blob/7a8e82c8637b3d75b2d084f6cb85736b31662f40/src/core/qgis.h#L4305)

rouault: Excellent suggestion. Implemented in b191a0ee796018b4c6f59226934eab008a255a6a

nyalldawson (approved):

PR #8741, opened 2023-11-17
OGRLayer: Have CreateField take const OGRFieldDefn* argument

dbaston:
## What does this PR do?

Change `OGRLayer::CreateField` interface to accept a `const OGRField*`, reflecting the semantics of current implementations.

Avoids the need to call the non-const `OGRFeatureDefn::GetFieldDefn` only to pass the return value along to `OGRLayer::CreateField`.

Does not touch drivers not included in my local build (can address these if the overall approach is sound).

## Tasklist

 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/MIGRATION_GUIDE.TXT
+++ b/MIGRATION_GUIDE.TXT
@@ -1,3 +1,13 @@
+MIGRATION GUIDE FROM GDAL 3.8 to GDAL 3.9
+-----------------------------------------
+
+- Out-of-tree vector drivers:
+  * OGRLayer::CreateField() now takes a const OGRFieldDefn* instead of a
+    OGRFieldDefn*.
+  * OGRLayer::CreateGeomField() now takes a const OGRGeomFieldDefn* instead of
+    a OGRGeomFieldDefn*.
+
+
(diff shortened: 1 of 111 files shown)


rouault:
> Change `OGRLayer::CreateField` interface to accept a `const OGRField*`, reflecting the semantics of current implementations.

sounds good. We'll need a note in MIGRATION_GUIDE.TXT to mention that change to out-of-tree drivers authors.

> Does not touch drivers not included in my local build (can address these if the overall approach is sound).

OK, let me know when you're done or have too much trouble dealing with drivers you can't easily test. I'll have to do an extra pass anyway as a few drivers aren't tested in CI (like DWG/DGNv8, IDB, etc).

rouault:
For consistency, we should likely do the same with CreateGeomFieldDefn()

rouault:
@dbaston I've just added a commit to your branch to fix a build issue in the OCI driver. With that, everything builds fine on my setup

rouault:
> Thanks! It looks like IDB would also need an update?

actually no, CreateField is disabled in a #if 0 // #endif block

PR #8748, opened 2023-11-19
sqlite_rtree_bulk_load.c: define __STDC_FORMAT_MACROS

barracuda156:
## What does this PR do?

The PR fixes the build of `gdal` for systems where `__STDC_FORMAT_MACROS` has to be defined in order to use `PRI*` macros.

See also: https://trac.macports.org/ticket/68734

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS: macOS 10.6
* Compiler: gcc13

P. S. Notice that while `gcc` advises to `#include <cinttypes>`, which indeed gonna work for it, some older macOS with clang will still be left broken. Defining instead `__STDC_FORMAT_MACROS` fixes all of them. Whether `<inttypes.h>` or `<cinttypes>` is used is then inconsequential.

### DIFF
--- a/ogr/ogrsf_frmts/sqlite/sqlite_rtree_bulk_load/sqlite_rtree_bulk_load.c
+++ b/ogr/ogrsf_frmts/sqlite/sqlite_rtree_bulk_load/sqlite_rtree_bulk_load.c
@@ -26,6 +26,10 @@
  * DEALINGS IN THE SOFTWARE.
  ****************************************************************************/
 
+#ifndef __STDC_FORMAT_MACROS
+#define __STDC_FORMAT_MACROS
+#endif
+
 #include "sqlite_rtree_bulk_load.h"
 
 #include <assert.h>


rouault:
Where I'm confused is that PRI macros are used in other files such as ogr/ogrsf_frmts/generic/ogrlayerarrow.cpp and we don't define ``__STDC_FORMAT_MACROS`` in that one

barracuda156:
@rouault Maybe some header defines which is pulled in there? Or cinttypes gets pulled.
I did not search through the source (and away from machine now), but that fixed the build for me which otherwise failed.

barracuda156:
@rouault Or otherwise they sit in some optional components which Macports `gdal` does not build. In such a case we do need this define there too. (Or move it into some general header instead, to avoid per-file fix-ups?)

rouault:
> Or maybe move it into some general header instead, to avoid per-file fix-ups?

you could possibly even just set ``-DCMAKE_CXX_FLAGS="-D__STDC_FORMAT_MACROS"`` when invoking CMake

barracuda156:
@rouault Yes, I know, but that will leave it unfixed for others.

rouault:
ok, I'm merging this as it is non super intrusive

barracuda156:
@rouault Thank you!

PR #8758, opened 2023-11-20
Check availability of std::shared_mutex with configure build test

nilason:
Older systems/compilers (or combinations thereof) may not have all C++17 features implemented and checking for `__cplusplus >= 201703L` is apparently not enough.

That is the case on Mac with failing builds of MacOSX 10.7–10.11 as reported with https://trac.macports.org/ticket/68716 (resulting in `Undefined symbols for architecture x86_64:   "std::__1::__shared_mutex_base::lock_shared()" ...` build error).

The patch submitted here, where the presence and usability of `std::shared_mutex` is tested during configuration, is tested with and works for macOS 13 as well as MacOSX 10.7.

(The `check_cxx_source_compiles` code is shamelessly taken and modified from cppreference.com).

### DIFF
--- a/cmake/helpers/configure.cmake
+++ b/cmake/helpers/configure.cmake
@@ -359,6 +359,46 @@ else ()
     add_definitions(-DDONT_DEPRECATE_SPRINTF)
   endif ()
 
+  check_cxx_source_compiles(
+    "
+    #include <mutex>
+    #include <shared_mutex>
+    int main(int argc, const char * argv[]) {
+        std::shared_mutex smtx;
+        int product = 0;
+        auto writer = [&smtx, &product](int start, int end)
+        {
+            for (int i = start; i < end; ++i)
+            {
+                auto data = i;
+                {
+                    std::unique_lock<std::shared_mutex> lock(smtx);
+                    product = data;
+                }
+            }
+            smtx.lock();
+            smtx.unlock();
+        };
+        auto reader = [&smtx, &product]()
+        {
+            int data = 0;
+            std::vector<int> seen;
+            do
+            {
+                {
+                    smtx.lock_shared();
+                    data = product;
+                    smtx.unlock_shared();
+                }
+            }
+            while (1);
+        };
+        return 0;
+    }
+    "
+    HAVE_SHARED_MUTEX
+  )
+
   check_include_file("linux/userfaultfd.h" HAVE_USERFAULTFD_H)
 endif ()
 
--- a/cmake/helpers/configure.cmake
+++ b/cmake/helpers/configure.cmake
@@ -361,38 +361,11 @@ else ()
 
   check_cxx_source_compiles(
     "
-    #include <mutex>
     #include <shared_mutex>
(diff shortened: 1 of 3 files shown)


cmake/helpers/configure.cmake
```diff
@@ -359,6 +359,46 @@ else ()
...
+                {
+                    smtx.lock_shared();
+                    data = product;
+                    smtx.unlock_shared();
+                }
+            }
+            while (1);
+        };
+        return 0;
+    }
```
rouault, line 397: hopefully that can be simplified as:

```suggestion
    #include <shared_mutex>
    int main(int argc, const char * argv[]) {
        std::shared_mutex smtx;
        smtx.lock_shared();
        smtx.unlock_shared();
        return 0;
    }
```

rouault:
@nilason The new requirement is C++17. I'll accept your patch (with my simplification), but at some point, we may use new features that have no workaround.

nilason:
> @nilason The new requirement is C++17. I'll accept your patch (with my simplification), but at some point, we may use new features that have no workaround.

Absolutely understandable and I wouldn't argue on that.

(It is a MacPorts' philosophy to support the latest software even to older systems, as far as possible.)

I did notice that the in the previous CI builds `Performing Test HAVE_SHARED_MUTEX` seem to fail for a number of workflows. Please do check that nothing fails where you'd expect it to succeed, before merging.

rouault:
> I did notice that the in the previous CI builds `Performing Test HAVE_SHARED_MUTEX` seem to fail for a number of workflows.

why ones?

nilason:
After a quick look, I believe all "Linux builds" fail, but the Mac build succeeds.

nilason:
Travis' linux build succeeds.

nilason:
Travis' windows-msys2-mingw succeeds.

rouault:
> After a quick look, I believe all "Linux builds" fail, but the Mac build succeeds.

With my simplified version, they pass

cmake/helpers/configure.cmake
```diff
@@ -359,6 +359,19 @@ else ()
     add_definitions(-DDONT_DEPRECATE_SPRINTF)
   endif ()
 
+  check_cxx_source_compiles(
+    "
+    #include <shared_mutex>
+    int main(int argc, const char * argv[]) {
```
rouault, line 365:
```suggestion
    int main() {
```

nilason:
> With my simplified version, they pass

The builds passed before too. But the configuration tests didn't and don't. See configure logs (look for `Performing Test HAVE_SHARED_MUTEX`).

rouault:
> See configure logs

which ones? The ones I checked passed

nilason:
> > See configure logs
> 
> which ones? The ones I checked passed

You're right, now I checked again and they seem to be good.

I did look just after your commit and they seemingly had failed... perhaps there was a GitHub glitch showing me the old logs.

nilason:
Thanks!

PR #8772, opened 2023-11-21
Enable removal of virtual filesystems

Maxxen:
## What does this PR do?

This PR enables the removal of installed virtual filesystems. 

Not in GDAL, but over in DuckDB Spatial I've ran into multiple problems as a result of trying to integrate DuckDB's own filesystem abstractions into GDAL since we don't assume that there is a "global" filesystem. New filesystems can be attached or removed at runtime and are tied to the lifetime of a connection (of which there can be multiple simultaneously in different threads). My solution is to give every new connection its own filesystem handler with a random and unique prefix, which keeps connections from interfering with each other, but ideally I'd like to be able to remove the attached filesystem as well once a connection is closed. Hence, this patch.

## Tasklist

 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS: MacOS 13.4
* Compiler: Homebrew Clang 16.0.6

### DIFF
--- a/port/cpl_vsi.h
+++ b/port/cpl_vsi.h
@@ -719,6 +719,13 @@ void CPL_DLL VSIFreeFilesystemPluginCallbacksStruct(
 int CPL_DLL VSIInstallPluginHandler(
     const char *pszPrefix, const VSIFilesystemPluginCallbacksStruct *poCb);
 
+/**
+ * Unregister a handler previously installed with VSIInstallPluginHandler() on
+ * the given prefix.
+ * @since GDAL 3.9
+ */
+int CPL_DLL VSIRemovePluginHandler(const char *pszPrefix);
+
 /* ==================================================================== */
 /*      Time querying.                                                  */
 /* ==================================================================== */
--- a/port/cpl_vsi_virtual.h
+++ b/port/cpl_vsi_virtual.h
@@ -327,8 +327,7 @@ class CPL_DLL VSIFileManager
     static VSIFilesystemHandler *GetHandler(const char *);
     static void InstallHandler(const std::string &osPrefix,
                                VSIFilesystemHandler *);
-    /* RemoveHandler is never defined. */
-    /* static void RemoveHandler( const std::string& osPrefix ); */
+    static void RemoveHandler(const std::string& osPrefix);
 
     static char **GetPrefixes();
 };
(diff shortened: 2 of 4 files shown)


rouault:
Can you add a test for that in autotest/cpp/test_cpl.cpp at the place where VSIInstallPluginHandler() is tested ?

port/cpl_vsi.h
```diff
@@ -719,6 +719,13 @@ void CPL_DLL VSIFreeFilesystemPluginCallbacksStruct(
 int CPL_DLL VSIInstallPluginHandler(
     const char *pszPrefix, const VSIFilesystemPluginCallbacksStruct *poCb);
 
+/**
+ * Unregister a handler previously installed with VSIInstallPluginHandler() on
+ * the given prefix.
+ * @since GDAL 3.9
```
rouault, line 725:
```suggestion
 * the given prefix.
 * Note: it is generally unsafe to remove a handler while there are still file handles
 * opened managed by that handler. It is the responsibility of the caller to ensure
 * that it calls this function in a situation where it is safe to do so.
 * @since GDAL 3.9
```

Maxxen: Also: Is the `@since` versioning correct?

Maxxen:
Thanks! Will update later tonight.

port/cpl_vsi.h
rouault: > s the `@since` versioning correct?

yes

autotest/cpp/test_cpl.cpp
```diff
@@ -4197,6 +4197,67 @@ TEST_F(test_cpl, VSI_plugin_minimal_testing)
...
+
+    EXPECT_TRUE(VSIFOpenL("/vsimyplugin/test", "rb") == nullptr);
+    EXPECT_TRUE(VSIFOpenL("/vsimyplugin/i_dont_exist", "rb") == nullptr);
+
+    // Removing a non-existing handler is a no-op
+    VSIRemovePluginHandler("/vsimyplugin/");
+    VSIRemovePluginHandler("/vsifoobar/");
+}
+
+TEST_F(test_cpl, VSI_plugin_removal)
```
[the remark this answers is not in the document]
Maxxen: Alright!

PR #8804, opened 2023-11-23
Make build compatible of -DCMAKE_UNITY_BUILD=ON for faster builds

rouault:
Similar to PROJ's https://github.com/OSGeo/PROJ/pull/3962

Timings on Linux with Intel(R) Core(TM) i7-10750H CPU @ 2.60GHz, 12 vCPUS, and "make -j12" with pretty much all possible drivers:

Before:
```
debug, regular build:
real	    5m0,278s
user	    51m12,358s
sys	    4m34,644s
```

After with -DCMAKE_UNITY_BUILD=ON:
```
debug, unity:
real	    2m15,768s
user	    22m25,952s
sys	    1m43,179s

relwithdebinfo, unity:
real	    4m45,055s
user	    44m20,402s
sys	    2m9,070s
```

I've added the following note in building_from_soure.rst:
```

.. option:: CMAKE_UNITY_BUILD=OFF

    .. versionadded:: 3.9

    Default is OFF. This can be set to ON to build GDAL using the
    https://cmake.org/cmake/help/latest/variable/CMAKE_UNITY_BUILD.html feature.
    This helps speeding GDAL build times. This feature is still considered
    experimental for now, and could hide subtle bugs (we are not aware of
    any at writing time though). We don't recommend it for mission critical
    builds.
```


rouault:
Depends on RFC98 adoption

PR #8806, opened 2023-11-24
OGR: Add OGRLayer::GetExtent3D()

### LINKED ISSUE #8570: Add OGRLayer::GetExtent3D()
Motivation: https://github.com/qgis/QGIS/pull/54941#pullrequestreview-1680935143

- [x] Add generic implementation that iterates over features (like for the 2D version)
- [x] Add specialized implementation for shapefile where this can be quick as this in the .shp header
- [x] Add semi-rapid implementation for GeoPackage using a custom aggregate SQL function that just parses the GPKG geometry header of features to compute the 3D extent (cf OGRGeoPackageTableLayer::GetGeometryTypes() for an exemple). We might also consider (as a follow-up) to extent the gpkg_contents table with a zmin and zmax columns, but probably not by default as this could affect compatibility with other readers
- [x] GeoJSON could also have it for free as it scans the whole file at opening.
- [x] Add a OLCFastGetExtent3D layer capability

elpaso:
Fix #8570

## TODO:
- [ ] fast implementations for drivers that support/allow it


ogr/ogrsf_frmts/ogrsf_frmts.h
```diff
@@ -115,6 +115,8 @@ class CPL_DLL OGRLayer : public GDALMajorObject
     int InstallFilter(OGRGeometry *);
 
     OGRErr GetExtentInternal(int iGeomField, OGREnvelope *psExtent, int bForce);
+    OGRErr GetExtent3DInternal(int iGeomField, OGREnvelope3D *psExtent3D,
+                               int bForce);
```
rouault, line 119: I don't think this is needed
```suggestion
```

ogr/ogrsf_frmts/generic/ogrlayer.cpp
```diff
@@ -227,6 +227,85 @@ OGRErr OGRLayer::GetExtent(int iGeomField, OGREnvelope *psExtent, int bForce)
...
+OGRErr OGRLayer::GetExtent3D(int iGeomField, OGREnvelope3D *psExtent3D,
+                             int bForce)
+
+{
+    psExtent3D->MinX = 0.0;
+    psExtent3D->MaxX = 0.0;
+    psExtent3D->MinY = 0.0;
+    psExtent3D->MaxY = 0.0;
+    psExtent3D->MinZ = 0.0;
+    psExtent3D->MaxZ = 0.0;
```
rouault, line 239: we might want to initialize those to NaN in case there would be only 2D geometries
```suggestion
    psExtent3D->MinZ = 0.0;
    psExtent3D->MaxZ = 0.0;
```

ogr/ogrsf_frmts/generic/ogrlayer.cpp
```diff
@@ -227,6 +227,85 @@ OGRErr OGRLayer::GetExtent(int iGeomField, OGREnvelope *psExtent, int bForce)
...
+        if (poGeom == nullptr || poGeom->IsEmpty())
+        {
+            /* Do nothing */
+        }
+        else if (!bExtentSet)
+        {
+            poGeom->getEnvelope(psExtent3D);
+            if (!(CPLIsNan(psExtent3D->MinX) || CPLIsNan(psExtent3D->MinY) ||
+                  CPLIsNan(psExtent3D->MaxX) || CPLIsNan(psExtent3D->MaxY)) ||
+                CPLIsNan(psExtent3D->MinZ) || CPLIsNan(psExtent3D->MaxZ))
```
rouault, line 282: I believe the tests on X/Y being not NaN are unnecessary.  The tests on Z might be needed. It all depends if we fix OGRGeometry::getEnvelope(OGREnvelope3D*) to return NaN for MinZ/MaxZ when the geometry is 2D.

ogr/ogrsf_frmts/generic/ogrlayer.cpp
```diff
@@ -227,6 +227,85 @@ OGRErr OGRLayer::GetExtent(int iGeomField, OGREnvelope *psExtent, int bForce)
...
+            if (oEnv.MinY < psExtent3D->MinY)
+                psExtent3D->MinY = oEnv.MinY;
+            if (oEnv.MinZ < psExtent3D->MinZ)
+                psExtent3D->MinZ = oEnv.MinZ;
+            if (oEnv.MaxX > psExtent3D->MaxX)
+                psExtent3D->MaxX = oEnv.MaxX;
+            if (oEnv.MaxY > psExtent3D->MaxY)
+                psExtent3D->MaxY = oEnv.MaxY;
+            if (oEnv.MaxZ > psExtent3D->MaxZ)
+                psExtent3D->MaxZ = oEnv.MaxZ;
```
rouault, line 301: You might want to use the OGREnvelope3D::Merge() method. It might perhaps need tweeking to deal with NaN in Z components

ogr/ogrsf_frmts/ogrsf_frmts.h
elpaso: yes, leftover from a first copy-paste.

ogr/ogrsf_frmts/generic/ogrlayer.cpp
elpaso: Ok, I'll implement it to handle NaN for Z quietly. 
Client code will need to check for Z not NaN to make sure it's 3D.

ogr/ogrsf_frmts/generic/ogrlayer.cpp
elpaso: I am trying to handle NaN but before going down the rabbit hole of patching all getEnvelope() implementations (and the OGR geometry classes which initialize Z to 0) I wanted to ask you if it worth the effort.

Maybe it is better to check if geom is 3D and bail out from GetExtent3D if it's not.

elpaso: @rouault I am trying a mixed approach (keep NaN for the envelope but not dig into the geometry classes) see https://github.com/OSGeo/gdal/pull/8806/commits/9783464d62429d9df209289c4a9c80c049e6bd05#diff-65b11183cce13d2ef546cfbe7935e31cc23b6557aae65276d8b8494a3a80720fR2939 for the new behavior, does it make sense?

rouault: > I am trying a mixed approach (keep NaN for the envelope but not dig into the geometry classes) see [9783464#diff-65b11183cce13d2ef546cfbe7935e31cc23b6557aae65276d8b8494a3a80720fR2939](https://github.com/OSGeo/gdal/commit/9783464d62429d9df209289c4a9c80c049e6bd05#diff-65b11183cce13d2ef546cfbe7935e31cc23b6557aae65276d8b8494a3a80720fR2939) for the new behavior, does it make sense?

yes, that's doable. Updating OGRGeometry::getEnvelope() to set NaN for Z could also be done as a later independent step.

ogr/ogr_core.h
```diff
@@ -228,6 +228,12 @@ extern "C++"
         /** Assignment operator */
         OGREnvelope3D &operator=(const OGREnvelope3D &) = default;
 
+        /** Returns TRUE if MinZ and MaxZ are both valid numbers. */
+        int Is3D() const
```
rouault, line 232:
```suggestion
        bool Is3D() const
```

ogr/ogrsf_frmts/shape/ogrshapelayer.cpp
```diff
@@ -1674,6 +1674,17 @@ GIntBig OGRShapeLayer::GetFeatureCount(int bForce)
 
 OGRErr OGRShapeLayer::GetExtent(OGREnvelope *psExtent, int bForce)
 
+{
+    OGREnvelope3D envelope3D;
+    const OGRErr retVal{GetExtent3D(0, &envelope3D, bForce)};
+    psExtent->MinX = envelope3D.MinX;
+    psExtent->MinY = envelope3D.MinY;
+    psExtent->MaxX = envelope3D.MaxX;
+    psExtent->MaxY = envelope3D.MaxY;
```
rouault, line 1683: As OGREnvelope3D subclass OGREnvelope, I believe below should just work (hopefully...):
```suggestion
    *psExtent = envelope3D;
```

autotest/ogr/ogr_shape.py
```diff
@@ -1898,13 +1898,15 @@ def test_ogr_shape_48(tmp_vsimem):
...
-        ogr.CreateGeometryFromWkt("POLYGON((0 0 2,0 1 2,1 1 2,1 0 2,0 0 2))")
+        ogr.CreateGeometryFromWkt("POLYGON((0 0 2,0 1 1,1 1 2,1 0 2,0 0 3))")
     )
     lyr.SetFeature(feat)
     ds.ExecuteSQL("RECOMPUTE EXTENT ON ogr_shape_48")
-    # FIXME: when we have a GetExtent3D
     extent = lyr.GetExtent()
     assert extent == (0, 1, 0, 1), "did not get expected extent (4)"
+    extent3D = lyr.GetExtent3D()
+    assert lyr.TestCapability(ogr.OLCFastGetExtent)
```
rouault, line 1908: shouldn't that be ?
```suggestion
    assert lyr.TestCapability(ogr.OLCFastGetExtent3D)
```

It would be great to also test GetExtent3D() on a 2D only shapefile. Too check that we get NaN in the minZ/maxZ fields

ogr/ogrsf_frmts/shape/ogrshapelayer.cpp
elpaso: clazy not happy:
image

rouault: would be nice to add a assign operator indeed

ogr/ogr_core.h
```diff
@@ -228,6 +230,12 @@ extern "C++"
         /** Assignment operator */
         OGREnvelope3D &operator=(const OGREnvelope3D &) = default;
 
+        /** Returns TRUE if MinZ and MaxZ are both valid numbers. */
+        bool Is3D() const
+        {
+            return !std::isnan(MinZ) && !std::isnan(MaxZ);
```
rouault, line 236: you need to add ``#include <cmath>`` at line 51

rouault:
some Doxygen errors: https://github.com/OSGeo/gdal/actions/runs/7029697053/job/19127794968?pr=8806

swig/include/ogr.i
```diff
@@ -491,6 +491,7 @@ typedef void retGetPoints;
 %constant char *OLCFastSpatialFilter   = "FastSpatialFilter";
 %constant char *OLCFastFeatureCount    = "FastFeatureCount";
 %constant char *OLCFastGetExtent       = "FastGetExtent";
+%constant char *OLCFastGetExtent3D     = "FastGetExtent3D";
```
rouault, line 494: this should be also declared a bit below around line 548 for the C# bindings

elpaso:
> @elpaso besides the comment I just pasted regarding defining OLCFastGetExtent3D for the C# bindings, I believe this PR is ready for the scope we discussed, right ?

Yes, I think so. As a following step I could try to implement more efficient "OCLFast" methods for drivers that support it.

rouault:
> As a following step I could try to implement more efficient "OCLFast" methods for drivers that support it.

yes, GeoPackage and PostGIS are ones that could be interesting

PR #8813, opened 2023-11-24
CSV writer: do not quote integer fields by default (only if STRING_QUOTING=ALWAYS is specified)

rouault:
Fixes https://gis.stackexchange.com/questions/470855/integer-field-of-csv-file-always-output-as-string-when-processed-with-ogr2ogr

### DIFF
--- a/autotest/ogr/ogr_csv.py
+++ b/autotest/ogr/ogr_csv.py
@@ -2633,7 +2633,7 @@ def test_ogr_csv_string_quoting_always(tmp_vsimem):
     data = gdal.VSIFReadL(1, 10000, f).decode("ascii")
     gdal.VSIFCloseL(f)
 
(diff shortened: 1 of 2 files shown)


jratike80:
What I think was a real problem in the gis.stackexchange case was that the field was defined to be integer in the .csvt file but GDAL did not trust it. Without .csvt it is quite possible that strings like '0123' appear later in the data, but having .csvt indicates that user has taken the responsibility.

rouault:
> What I think was a real problem in the gis.stackexchange case was that the field was defined to be integer in the .csvt file but GDAL did not trust it.

no, the issue in the gis.stackexchange case was only on the writing side. Even if the field was typed as integer, it would be written as quoted.

jratike80:
Right, I can see with ogrinfo that the reading side was OK.

PR #8854, opened 2023-11-28
Also mention vector on co.rst

jidanni:
So now it has both raster and vector.


rouault:
I'm not sure this is appropriate to mention vector drivers there. -co is typically used for raster  drivers  . For vector driver drivers, the options are generally -dsco and -lco

jidanni:
Oops, from just the filename path I thought I was in a part of the tree that referred to both.

PR #8864, opened 2023-11-29
heif: add CreateCopy support (WIP)

bradh:
## What does this PR do?

Initial sketch of work to add write support to the HEIF driver.

This does not use libheif to do the writing, given the differences in what libheif can do and what GDAL can represent.

At this stage it can only do the uncompressed codec (which libheif can support, but is behind a cmake flag). A recent version of gpac is also an option

Perhaps it would be better to do a different driver. Possibly the box support could be generalised and used for other drivers (jpegxl, jpeg2000, etc).

Comments and feedback on the approach and way forward are sought.

N/A.

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:

### DIFF
--- a/autotest/gdrivers/heif.py
+++ b/autotest/gdrivers/heif.py
@@ -29,6 +29,9 @@
 # DEALINGS IN THE SOFTWARE.
 ###############################################################################
 
+import array
+import os
+
 import pytest
 
 from osgeo import gdal
@@ -158,3 +161,51 @@ def test_heif_subdatasets():
         assert gdal.Open("HEIF:1") is None
     with pytest.raises(Exception):
         assert gdal.Open("HEIF:1:") is None
+
+
+def test_heif_create():
+    drv = gdal.GetDriverByName("HEIF")
+
+    try:
+        os.remove("tmp/test_create.hif")
+    except OSError:
+        pass
+
+    ds = gdal.GetDriverByName("MEM").Create("", 300, 200, 3, gdal.GDT_Byte)
+
+    ds.GetRasterBand(1).SetRasterColorInterpretation(gdal.GCI_RedBand)
+    ds.GetRasterBand(2).SetRasterColorInterpretation(gdal.GCI_GreenBand)
+    ds.GetRasterBand(3).SetRasterColorInterpretation(gdal.GCI_BlueBand)
+
+    red_green_blue = (
+        ([0xFF] * 100 + [0x00] * 200)
+        + ([0x00] * 100 + [0xFF] * 100 + [0x00] * 100)
+        + ([0x00] * 200 + [0xFF] * 100)
+    )
+    rgb_bytes = array.array("B", red_green_blue).tobytes()
+    for line in range(100):
+        ds.WriteRaster(
+            0, line, 300, 1, rgb_bytes, buf_type=gdal.GDT_Byte, band_list=[1, 2, 3]
+        )
+    black_white = ([0xFF] * 150 + [0x00] * 150) * 3
+    black_white_bytes = array.array("B", black_white).tobytes()
+    for line in range(100):
+        ds.WriteRaster(
+            0,
+            100 + line,
+            300,
+            1,
+            black_white_bytes,
+            buf_type=gdal.GDT_Byte,
+            band_list=[1, 2, 3],
+        )
+
+    assert ds.FlushCache() == gdal.CE_None
+
+    ds = drv.CreateCopy("tmp/test_create.hif", ds, options=[])
+
+    ds = None
+
+    ds = gdal.Open("tmp/test_create.hif")
+
+    assert ds
--- a/frmts/heif/CMakeLists.txt
+++ b/frmts/heif/CMakeLists.txt
@@ -1,5 +1,17 @@
 add_gdal_driver(TARGET gdal_HEIF
-                SOURCES heifdataset.cpp
+                SOURCES heifdataset.cpp heifdataset.h 
+                        heifdatasetcreatecopy.cpp
+                        boxes/box.cpp boxes/box.h
+                        boxes/fullbox.cpp boxes/fullbox.h
+                        boxes/cmpd.cpp boxes/cmpd.h
+                        boxes/hdlr.cpp boxes/hdlr.h
+                        boxes/iloc.cpp boxes/iloc.h
+                        boxes/iprp.cpp boxes/iprp.h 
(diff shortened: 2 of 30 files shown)


rouault:
Can you give some rationale why HEIF write support is needed ? I believe someone mentioned to me recently that a group within OGC was investigating about geospatial support for that format. They should rather look at JPEGXL IMHO.
To come to this PR, I'm quite sad to see that we need to do it at the image codec level. That seems like crossing a red line to me. We should really try to leverage libheif. Can't it be convinced to add custom boxes like libjxl can do, which the JPEGXL driver uses to insert a JUMBF superbox with a GeoJP2 one inside it ? If it cannot, then it would seem reasonable to try to add that capability first in libheif. If that turned to be impossible, maybe write a non geospatial HEIF file in a temporary file, and then manually insert the geospatial boxes in it.

bradh:
> Can you give some rationale why HEIF write support is needed ? I believe someone mentioned to me recently that a group within OGC was investigating about geospatial support for that format. They should rather look at JPEGXL IMHO. To come to this PR, I'm quite sad to see that we need to do it at the image codec level. That seems like crossing a red line to me. We should really try to leverage libheif. Can't it be convinced to add custom boxes like libjxl can do, which the JPEGXL driver uses to insert a JUMBF superbox with a GeoJP2 one inside it ? If it cannot, then it would seem reasonable to try to add that capability first in libheif. If that turned to be impossible, maybe write a non geospatial HEIF file in a temporary file, and then manually insert the geospatial boxes in it.

The driver is NGA - it is designing a replacement for NITF called GIMI (see https://opengeospatial.github.io/ogcna-auto-review/23-059.html#_b95138e1-5be7-4a49-ab1f-135c6b9daf22). The differentiator is that ISOBMFF also supports image sequences and video tracks. There is planned to be a proper geospatial model (likely format is RDF encoded in KLV), but that is lagging.

I've done some work on libheif, but I don't see the maintainer (Dirk) wanting to do some things that GDAL can use (e.g. floating point for example).  In some ways libheif is an abstraction layer over HEVC, AV1 and JPEG 2000 libraries, and there is some overlap with capabilities that GDAL already has (e.g. tiling is important, but the API isn't there). 

If you are -1 on this approach, would writing via libheif likely be acceptable?

rouault:
> If you are -1 on this approach, would writing via libheif likely be acceptable?

yes, that would seem more reasonable to me. The prospect of having low level understanding of HEIF details in GDAL doesn't seem appealing to me. That could potentially mean a rather big number of tricky lines of code, and adding even more dependent libraries to GDAL. But perhaps I've not a fully informed opinion on this as what I know about HEIF is limited to using a few functions of libheif.
If the maintainer of libheif isn't open to your needs, there's the possibility of forking it, although I can see it could cause distribution issues.
This whole subject seems to fall into https://gdal.org/development/rfc/rfc85_policy_code_additions.html . It could be good that you prepare a RFC for consideration by the whole GDAL project steering committee detailing the rationale, plan & impacts.

hobu:
>  NGA - it is designing a replacement for NITF called GIMI 

This is the most NGA way forward possible – design something that hasn't been tested, using software that doesn't exist, and foisting it on GDAL so it gets convenient distribution of its frankenstein. 

I'm a strong -1 on vendoring this much special heif stuff into GDAL. You're going to leave and this stuff will rot just like NITF stuff in PDAL does. IMO the way forward is to get this stuff distributed with libheif or make your own libheif and then let users link against it.

bradh:
> > NGA - it is designing a replacement for NITF called GIMI
> 
> This is the most NGA way forward possible – design something that hasn't been tested, using software that doesn't exist, and foisting it on GDAL so it gets convenient distribution of its frankenstein.

[Not speaking for NGA - I'm trying to work with them, but I don't work for them, and definitely don't represent them in any way].

I do have some sympathy for your "NGA way forward" assessment in general, but at least for GIMI, I see genuine intent to collaborate.

The design for the file format is ISOBMFF and HEIF. There is mass adoption of those in the commercial world - mainly in modern phones. HEIF supports (and libheif implements) a range of codec options, including some that have common hardware acceleration (or GPU acceleration). The lossless / uncompressed codec is more complicated than I would have liked (at least on decode) and isn't published yet, but it being worked through MPEG and it isn't just an ISO stamp on an NGA design. 

The software mostly does exist, and we did some testing with open source geospatial during the last OGC code sprint (which NGA and DSTL kicked in some sponsorship to provide travel for developers, including me). A bunch of things were hard, and a lot of those would have been easy if GDAL already implemented more of HEIF. 

> I'm a strong -1 on vendoring this much special heif stuff into GDAL. You're going to leave and this stuff will rot just like NITF stuff in GDAL does. IMO the way forward is to get this stuff distributed with libheif or make your own libheif and then let users link against it.

Understood, if not fully agreed. I'll try the libheif option, and potentially some other ideas.

@hobu: I'll also take you up on your offer (from Baltimore) to discuss OGC and NGA.

hobu:
Just because I'm -1 doesn't mean you should close the ticket 😄 

> I do have some sympathy for your "NGA way forward" assessment in general, but at least for GIMI, I see genuine intent to collaborate.

/me fails at googling anything about "GIMI" "NGA", and/or "NITF". I think a missing piece of the story here is simply an educational one. 

> The software mostly does exist, and we did some testing with open source geospatial during the last OGC code sprint (which NGA and DSTL kicked in some sponsorship to provide travel for developers, including me). A bunch of things were hard, and a lot of those would have been easy if GDAL already implemented more of HEIF.

Everything is *always* a lot easier when GDAL implements it 😉 

The problem I have with the PR is just like NITF, GIMI expertise might be openly distributed. GDAL's NITF support gets slagged frequently because it is not as sophisticated as necessary in some instances, but the expertise and attention to keep it current just hasn't had any staying power with the project. Your +2432 line PR is a significant chunk, and the fear is that it would be orphaned code due to libheif not wanting to accept it. We definitely don't want to have a GDAL implementation be an ice floe like that unless there were some kind of support coming along with it to care for and curate it.

I also understand it is a bit of a 🐔 🥚 scenario in that a convenient GDAL usage scenario means more users, which means more use, which means more refinement and evolution of the effort. The GDAL project itself kind of lead that implementation lifecycle with COG.

PR #8866, opened 2023-11-29
OpenFileGDB: add support for Integer64, Date, Time and DateTimeWithOffset data types added in ArcGIS Pro 3.2 (fixes #8862)

### LINKED ISSUE #8862: Support bigint types in file geodatabases
As described in:

https://www.esri.com/arcgis-blog/products/arcgis-pro/data-management/whats-new-from-the-geodatabase-team-in-arcgis-pro-3-2/#64-bit-OIDs-and-the-BigInteger-Field-Data-Type

A new field type for "big integer" was introduced in the file geodatabase specification in ArcGIS Pro 3.2.

I've made a sample gdb using this field type here:
[big_int.gdb.zip](https://github.com/OSGeo/gdal/files/13494330/big_int.gdb.zip)

Currently this field type isn't handled in GDAL, and prevents opening the table.

rouault:
@nyalldawson  I've also tried to add support for Date, Time and DateTimeWithOffset . Could you create a test dataset with those types, preferably with default values for the fields  (default value for Int64 would also be cool) ?
and provided decoded value of the feature field content, as I'm not sure to have fully understood how Time (values between 0 and 1?) and DateTimeWithOffset (is the double value encoded as if it was already a UTC value and the offset is just an extra qualifier, or must it also be converted to UTC ?)  are encoded


nyalldawson:
Ok, here we go: [arcgis_pro_32_types.gdb.zip](https://github.com/OSGeo/gdal/files/13506063/arcgis_pro_32_types.gdb.zip)


The big_int table contents are:

| OBJECTID | Shape | short | long | big | float | double |
|----------|-------|-------|------|-----|-------|--------|
| 1        | PointZ| 32767 | 2147483647 | 9007199254740991 | 3.400000e+38 | 1.797693e+308 |
| 2        | PointZ| -32768| -2147483647 | -9007199254740991 | -3.400000e+38 | -1.797693e+308 |

Default value for the `big` field is 1234567890123456

There's an index `bigint_index` I've added to that field too.


The contents of `date_types` are:

| OBJECTID | SHAPE  | date            | dateonly   | time_only       | timestamp_offset        |
|----------|--------|-----------------|------------|-----------------|-------------------------|
| 1        | Point Z| 29/11/2023 1:14:15 PM | 29/11/2023 | 1:14:15.000 PM  | 29/11/2023 1:14:15.000 PM -05:00                   |
| 2        | Point Z| 31/12/2023 12:01:01 AM | 31/12/2023 | 12:01:01.000 AM | 31/12/2023 12:01:01.000 AM +10:00                   |
| 3        | Point Z| 1/01/1901 12:01:01 AM  | 1/01/1901  | 12:01:01.000 AM | 1/01/1901 12:01:01.000 AM +10:00                   |

Default values are:

| field | value |
|----|---|
| date | 1/02/2023 4:05:06 AM |
| date_only | 1/02/2023 |
| time_only | 4:05:06 AM |
| timestamp_offset | 1/02/2023 4:05:06.000 AM + 06:00 |


Then, I've included a table `date_types_high_precision`. This is similar to the `date_types` table, but the `date` field has been migrated to high precision time stamp field. See https://pro.arcgis.com/en/pro-app/latest/help/data/geodatabases/overview/arcgis-field-data-types.htm#ESRI_SECTION3_34470CE25A0B479CA8F7546B23D440A9

Values in this table are:

| OBJECTID | SHAPE  | date            | dateonly   | time_only       | timestamp_offset        |
|----------|--------|-----------------|------------|-----------------|-------------------------|
| 1        | Point Z| 29/11/2023 1:14:15.678 PM | 29/11/2023 | 1:14:15.000 PM  | 29/11/2023 1:14:15.000 PM -05:00                   |
| 2        | Point Z| 31/12/2023 12:01:01.001 AM | 31/12/2023 | 12:01:01.000 AM | 31/12/2023 12:01:01.000 AM +10:00                   |
| 3        | Point Z| 1/01/1901 12:01:01.999 AM  | 1/01/1901  | 12:01:01.000 AM | 1/01/1901 12:01:01.000 AM +10:00                   |

Default value for the `date` field in this table is `2/01/2023 4:05:06.007 AM`. Note that it's not just the values which differ in the "high precision" table, as they are displayed differently in the UI. There must be a flag somewhere to indicate that the field is "high precision" too.

There's also date_index, time_index, timestamp_offset_index present in this second table.

nyalldawson:
Something to consider is this:

```
Some client applications may not fully support this field type and are subject to change. Object ID (64-bit) and big integer (64-bit) field types in ArcGIS Pro are limited to storing values up to 53-bit safe integers. 53-bit safe integer values support a range between (-9007199254740991) and (9007199254740991), which is also written as -(2 53) -1 and (2 53) -1. Values larger than 53 bits in object ID (64-bit) and big integer fields added from external sources can be read from the ArcGIS Pro client. Attempting to write or edit numbers outside this range in ArcGIS Pro will generate an error. In other clients, values may be rounded and result in broken functionality.
```

(from https://pro.arcgis.com/en/pro-app/latest/help/data/geodatabases/overview/arcgis-field-data-types.htm#ESRI_SECTION2_8BF2454C879941258DC44AF6BB31F386)

I'm wondering if we need to expose an option to permit values outside the range usable in ArcGIS Pro, or if this is overkill and considered the operators responsibility?

rouault:
> Object ID (64-bit) and big integer (64-bit) field types in ArcGIS Pro are limited to storing values up to 53-bit safe integers

sigh... They couldn't even do the job properly... It means that they are actually using Float64 in their internal model.

> I'm wondering if we need to expose an option to permit values outside the range usable in ArcGIS Pro, or if this is overkill and considered the operators responsibility?

I'd just go for documenting that limitation

rouault:
@nyalldawson Thanks for the dataset. Very useful. I've update the PR with what I learnt from it. 

Could you test with ArcGIS Pro the attached dataset ( [converted_by_gdal_from_arcgis_pro_32_types.gdb.zip](https://github.com/OSGeo/gdal/files/13507042/converted_by_gdal_from_arcgis_pro_32_types.gdb.zip) ) resulting from: 
```
ogr2ogr converted_by_gdal_from_arcgis_pro_32_types.gdb  ../autotest/ogr/data/filegdb/arcgis_pro_32_types.gdb -lco TARGET_ARCGIS_VERSION=ARCGIS_PRO_3_2_OR_LATER
ogrinfo converted_by_gdal_from_arcgis_pro_32_types.gdb -sql "CREATE INDEX idx_date_only ON date_types(date_only)"
ogrinfo converted_by_gdal_from_arcgis_pro_32_types.gdb -sql "CREATE INDEX idx_time_only ON date_types(time_only)"
ogrinfo converted_by_gdal_from_arcgis_pro_32_types.gdb -sql "CREATE INDEX idx_tsoffset ON date_types(timestamp_offset)"
ogrinfo converted_by_gdal_from_arcgis_pro_32_types.gdb -sql "CREATE INDEX idx_big ON big_int(big)"
```

nyalldawson:
@rouault looking good! It's showing all the field values / defaults / field types as expected.

The only thing I can spot is that the "high precision" flag has been dropped from the date field in the "date_types_high_precision" table, and accordingly the decimal seconds have been lost (in both the attribute values and default value). I'm not sure how that should be handled (or even if it should at all)...

nyalldawson (approved):

rouault:
> The only thing I can spot is that the "high precision" flag has been dropped from the date field in the "date_types_high_precision" table, and accordingly the decimal seconds have been lost (in both the attribute values and default value). I'm not sure how that should be handled (or even if it should at all)...

yes, I'm aware of that. Actually, when using ARCGIS_PRO_3_2_OR_LATER, the driver always creates fields of type TimestampOffset and those don't appear to support the "high precision" flag. At least, by looking at your sample file, the flag was set only to the "date" field, not "timestamp_offset"

nyalldawson:
> Actually, when using ARCGIS_PRO_3_2_OR_LATER, the driver always creates fields of type TimestampOffset and those don't appear to support the "high precision" flag. At least, by looking at your sample file, the flag was set only to the "date" field, not "timestamp_offset"

Looks like the newer TimestampOffset type is **always** high precision. (which would make sense, the opt-in update for the older field type might be there just for backward compatibility).

rouault:
> Looks like the newer TimestampOffset type is **always** high precision

ok, I've updated the PR to write to write such fields as high precision.

nyalldawson:
Thanks @rouault ! Is this suitable for backporting? Or too much of a new feature?

rouault:
> Or too much of a new feature?

yes, slightly too substantial and non-trivial... I can't exclude that it could cause issues. Let's see if fuzzers uncover obvious issues in the coming weeks. If you can exercise that functionality a bit in the coming weeks too, perhaps we can reconsider

PR #8885, opened 2023-11-30
GDALTranslate(): avoid useless extra GDALOpen() call on a target GeoRaster (again)

rouault:


fechen123:
Verified that this fix removed the two extra GDALOpen() calls to GeoRaster dataset in gdal_translate.

PR #8899, opened 2023-12-01
OGRSpatialReference::importFromUrl(): changes to no longer use a 'Accept: application/x-ogcwkt' header

rouault:
autotest/osr/osr_url.py already tests the following URLs:
- "http://spatialreference.org/ref/epsg/4326/"
- "http://spatialreference.org/ref/epsg/4326/ogcwkt/"
- "https://spatialreference.org/ref/epsg/4326/"


rouault:
CC @jjimenezshaw

PR #8906, opened 2023-12-04
VRT: 'sd_name' and 'sd' for vrt:// connection

mdsumner:
Add `sd_name` and `sd` to allowed options for a "vrt://" connection.

## Tasklist

 - [x] Add test case(s)
 - [x] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Example

online netcdf subdataset by name or number

```bash
gdalinfo "vrt:///vsicurl/https://www.ncei.noaa.gov/data/sea-surface-temperature-optimum-interpolation/v2.1/access/avhrr/202202/oisst-avhrr-v02r01.20220218.nc?sd_name=ice"

gdalinfo "vrt:///vsicurl/https://www.ncei.noaa.gov/data/sea-surface-temperature-optimum-interpolation/v2.1/access/avhrr/202202/oisst-avhrr-v02r01.20220218.nc?sd=4"

```

### DIFF
--- a/frmts/vrt/vrtdataset.cpp
+++ b/frmts/vrt/vrtdataset.cpp
@@ -1044,8 +1044,7 @@ GDALDataset *VRTDataset::OpenVRTProtocol(const char *pszSpec)
         return nullptr;
     }
 
-    // scan for subdataset/subdataset_n in tokens, close the source dataset and reopen with subdataset if found/valid
-    // make subdataset and subdataset_n mutually exclusive
+    // scan for sd_name/sd in tokens, close the source dataset and reopen if found/valid
     bool bFound_subdataset = false;
     for (int i = 0; i < aosTokens.size(); i++)
     {
@@ -1054,31 +1053,41 @@ GDALDataset *VRTDataset::OpenVRTProtocol(const char *pszSpec)
 
         if (pszKey && pszValue)
         {
-            char **papszSubdatasets = GDALGetMetadata(poSrcDS, "SUBDATASETS");
 
-            int nSubdatasets = CSLCount(papszSubdatasets);
-
-            if (EQUAL(pszKey, "subdataset"))
+            if (EQUAL(pszKey, "sd_name"))
             {
                 if (bFound_subdataset)
                 {
                     CPLError(CE_Failure, CPLE_IllegalArg,
-                             "'subdataset' is mutually exclusive with option "
-                             "'subdataset_n'");
+                             "'sd_name' is mutually exclusive with option "
+                             "'sd'");
                     poSrcDS->ReleaseRef();
 
                     CPLFree(pszKey);
                     return nullptr;
                 }
+                char **papszSubdatasets = poSrcDS->GetMetadata("SUBDATASETS");
+                int nSubdatasets = CSLCount(papszSubdatasets);
+
                 if (nSubdatasets > 0)
                 {
                     bool bFound = false;
                     for (int j = 0; j < nSubdatasets; j += 2)
                     {
-                        char *pszSubdatasetSource =
-                            CPLStrdup(strstr(papszSubdatasets[j], "=") + 1);
+                        const std::string osSubdatasetSource(
+                            strstr(papszSubdatasets[j], "=") + 1);
+                        if (osSubdatasetSource.empty())
+                        {
+                            CPLError(CE_Failure, CPLE_IllegalArg,
+                                     "'sd_name:' failed to obtain "
+                                     "subdataset string ");
+                            poSrcDS->ReleaseRef();
+
+                            CPLFree(pszKey);
+                            return nullptr;
+                        }
                         GDALSubdatasetInfoH info =
-                            GDALGetSubdatasetInfo(pszSubdatasetSource);
+                            GDALGetSubdatasetInfo(osSubdatasetSource.c_str());
                         char *component =
                             info
                                 ? GDALSubdatasetInfoGetSubdatasetComponent(info)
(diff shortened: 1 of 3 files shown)


frmts/vrt/vrtdataset.cpp
```diff
@@ -1044,6 +1044,121 @@ GDALDataset *VRTDataset::OpenVRTProtocol(const char *pszSpec)
...
+    for (int i = 0; i < aosTokens.size(); i++)
+    {
+        char *pszKey = nullptr;
+        const char *pszValue = CPLParseNameValue(aosTokens[i], &pszKey);
+
+        if (pszKey && pszValue)
+        {
+            char **papszSubdatasets = GDALGetMetadata(poSrcDS, "SUBDATASETS");
+
+            int nSubdatasets = CSLCount(papszSubdatasets);
```
rouault, line 1059: querying subdatasets might be "costly" when not needed. So I'd move this within the if (EQUAL(pszKey, "subdataset")) and if (EQUAL(pszKey, "subdataset_n")) cases

frmts/vrt/vrtdataset.cpp
```diff
@@ -1044,6 +1044,121 @@ GDALDataset *VRTDataset::OpenVRTProtocol(const char *pszSpec)
...
+
+            if (EQUAL(pszKey, "subdataset"))
```
rouault, line 1061: perhaps ``subdataset_name`` for that one and ``subdataset_idx`` instead of ``subdataset_n``  for clarity ?

frmts/vrt/vrtdataset.cpp
```diff
@@ -1044,6 +1044,121 @@ GDALDataset *VRTDataset::OpenVRTProtocol(const char *pszSpec)
...
+    // make subdataset and subdataset_n mutually exclusive
+    bool bFound_subdataset = false;
+    for (int i = 0; i < aosTokens.size(); i++)
+    {
+        char *pszKey = nullptr;
+        const char *pszValue = CPLParseNameValue(aosTokens[i], &pszKey);
+
+        if (pszKey && pszValue)
+        {
+            char **papszSubdatasets = GDALGetMetadata(poSrcDS, "SUBDATASETS");
```
rouault, line 1057:
```suggestion
            char **papszSubdatasets = poSrcDS->GetMetadata("SUBDATASETS");
```

frmts/vrt/vrtdataset.cpp
```diff
@@ -1044,6 +1044,121 @@ GDALDataset *VRTDataset::OpenVRTProtocol(const char *pszSpec)
...
+                    CPLFree(pszKey);
+                    return nullptr;
+                }
+                if (nSubdatasets > 0)
+                {
+                    bool bFound = false;
+                    for (int j = 0; j < nSubdatasets; j += 2)
+                    {
+                        char *pszSubdatasetSource =
+                            CPLStrdup(strstr(papszSubdatasets[j], "=") + 1);
```
rouault, line 1079: slightly unsafe to assume strstr() returns a not-null value. Should be checked

frmts/vrt/vrtdataset.cpp
```diff
@@ -1044,6 +1044,121 @@ GDALDataset *VRTDataset::OpenVRTProtocol(const char *pszSpec)
...
+                                : nullptr;
+
+                        bFound = component && EQUAL(pszValue, component);
+                        bFound_subdataset = true;
+                        CPLFree(component);
+                        GDALDestroySubdatasetInfo(info);
+                        if (bFound)
+                        {
+                            poSrcDS->ReleaseRef();
+                            poSrcDS = GDALDataset::Open(
```
rouault, line 1094: the return of Open() should be checked for null, to return if it is (and doing appropriate cleanup)

frmts/vrt/vrtdataset.cpp
rouault: and instead of putting that in a C char* with CPLStrdup() / CPLFree(), you could assign to a std::string for automatic memory management

frmts/vrt/vrtdataset.cpp
```diff
@@ -1044,6 +1044,121 @@ GDALDataset *VRTDataset::OpenVRTProtocol(const char *pszSpec)
...
+                        CPLError(
+                            CE_Failure, CPLE_IllegalArg,
+                            "'subdataset_n' option should indicate a valid "
+                            "subdataset component index (starting with 1)");
+                        CPLFree(pszKey);
+                        poSrcDS->ReleaseRef();
+                        return nullptr;
+                    }
+                    char *pszSubdatasetSource = CPLStrdup(
+                        strstr(papszSubdatasets[(iSubdataset - 1) * 2], "=") +
```
rouault, line 1146: - check strstr() for null
- assign to a std::string to avoid CPLStrdup()/CPLFree()

frmts/vrt/vrtdataset.cpp
mdsumner: I find this hard, in the API it's "component" rather than "name" (but NAME in info). I prefer "subdataset" and "subdataset_i", but I'm ok with this will just ponder a while. It's not clear what the precedents are so I'll have a bit of a look.

frmts/vrt/vrtdataset.cpp
mdsumner: ok thanks!  I copied it from somewhere :)

frmts/vrt/vrtdataset.cpp
rouault: > find this hard, in the API it's "component" rather than "name" (but NAME in info).

ah, yes I missed that. I'm not too opinionated on naming...

frmts/vrt/vrtdataset.cpp
mdsumner: (from gdal_translate_bin)

frmts/vrt/vrtdataset.cpp
mdsumner: maybe it should be "sd_name" and "sd" to be consistent with gdalinfo? 

I'm still thinking ...

mdsumner: and gdalbuildvrt also has "-sd", I don't see any other cases in apps/

mdsumner: ah, missed your reply - I'm going with sd_name/sd and see if the other changes pass :)

frmts/vrt/vrtdataset.cpp
mdsumner: done

frmts/vrt/vrtdataset.cpp
mdsumner: done with other changes, thank you

frmts/vrt/vrtdataset.cpp
mdsumner: done

mdsumner:
awesome, thank you!!!

PR #8524, opened 2023-10-06
ogr_proj_p.cpp: disable pthread_atfork() optimization on MacOS (fixes #8497)

rouault:
@emlys could you try this patch on top of GDAL master or 3.7 branch and confirm it fixes your isse ?

Issue #8497, opened 2023-09-29
python seg fault with GDAL>=3.6.0 after opening many rasters in threads


emlys:
## Expected behavior and actual behavior.

I expected the following python script to run successfully, but it exits with a seg fault.

## Steps to reproduce the problem.
This error was happening in our test suite when I tried to upgrade from GDAL 3.5 (Github Actions run [here](https://github.com/emlys/invest/actions/runs/6332320462/job/17198593123)). I narrowed it down to this weird, minimal example. Even the unused numpy import is necessary to reproduce:
```python
# test.py
import threading
import subprocess

import numpy
import requests
from osgeo import gdal

gdal.GetDriverByName('GTiff').Create('test.tif', 1, 1)

for _ in range(164):  # the seg fault does not happen with fewer than 164 threads
    thread = threading.Thread(
        target=gdal.OpenEx,
        args=('test.tif',))
    thread.start()
    thread.join()

requests.get('https://example.com')
subprocess.run(['pwd'])
```
```
$ python -X dev test.py
/Users/emily/mambaforge/envs/test/lib/python3.11/site-packages/osgeo/gdal.py:287: FutureWarning: Neither gdal.UseExceptions() nor gdal.DontUseExceptions() has been explicitly called. In GDAL 4.0, exceptions will be enabled by default.
  warnings.warn(
Fatal Python error: Segmentation fault

Current thread 0x00007ff855fa7680 (most recent call first):
  File "/Users/emily/mambaforge/envs/test/lib/python3.11/subprocess.py", line 1883 in _execute_child
  File "/Users/emily/mambaforge/envs/test/lib/python3.11/subprocess.py", line 1026 in __init__
  File "/Users/emily/mambaforge/envs/test/lib/python3.11/subprocess.py", line 548 in run
  File "/Users/emily/invest/test.py", line 18 in <module>

Extension modules: numpy.core._multiarray_umath, numpy.core._multiarray_tests, numpy.linalg._umath_linalg, numpy.fft._pocketfft_internal, numpy.random._common, numpy.random.bit_generator, numpy.random._bounded_integers, numpy.random._mt19937, numpy.random.mtrand, numpy.random._philox, numpy.random._pcg64, numpy.random._sfc64, numpy.random._generator, _brotli, osgeo._gdal, osgeo._gdalconst, osgeo._ogr, osgeo._osr (total: 18)
Segmentation fault: 11
```

## Operating system

macOS Ventura 13.2.1

[...]

PR #8909, opened 2023-12-04
OSRPJContextHolder: call pthread_atfork() once for the process

rouault:
Cf https://lists.osgeo.org/pipermail/gdal-dev/2023-December/058058.html

### DIFF
--- a/ogr/ogr_proj_p.cpp
+++ b/ogr/ogr_proj_p.cpp
@@ -121,7 +121,17 @@ struct OSRPJContextHolder
 #endif
     {
 #if HAVE_PTHREAD_ATFORK
-        pthread_atfork(nullptr, nullptr, ForkOccurred);
+        static std::once_flag flag;
+        std::call_once(
+            flag,
+            []()
+            {
+                if (pthread_atfork(nullptr, nullptr, ForkOccurred) != 0)
+                {
+                    CPLError(CE_Failure, CPLE_OutOfMemory,
+                             "pthread_atfork() in ogr_proj_p failed");
+                }
+            });
 #endif
         init();
     }
--- a/ogr/ogr_proj_p.cpp
+++ b/ogr/ogr_proj_p.cpp
@@ -35,14 +35,6 @@
 
 #include "proj.h"
 
-#if defined(__MACH__) && defined(__APPLE__) && defined(HAVE_PTHREAD_ATFORK)
-// Works around a weird issue with GDAL, numpy and python threading on
-// Mac. There is definitely something not understood, but as using pthread_atfork()
-// is just an optimization, just disable it.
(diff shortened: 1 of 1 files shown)


carsten-schultz-pix4d:
Thank you for this. Do you also plan to re-enable this for Macs?

Regarding the second point that I had brought up in my email, I agree that it is probably not of practical relevance. It wasn't specifically about having multiple threads though but about having multiple instances of `OSRPJContextHolder` in any way. If the only instances are the thread_local ones defined here and if we assume that no-one is foolish enough to mix threads and forks, then it will be okay. Nevertheless I would have preferred to not have a bool flag but just a counter that is increased and that could be handled in the same way as the pid in the `!defines(HAVE_PTHREAD_ATFORK)` code.

So, thanks again.

rouault:
> Do you also plan to re-enable this for Macs?

I'm not sure this was the root cause of https://github.com/OSGeo/gdal/issues/8497 . As I can't test it directly to check, I prefer to not change that.

carsten-schultz-pix4d:
> I'm not sure this was the root cause of #8497 . As I can't test it directly to check, I prefer to not change that.

That makes sense. If I can figure out how to use the Python bindings then I will test that. Given that I see the number 164 in the report and that the limit for calls to `pthread_atfork` on x86_64 is 170, I am quite confident that it is at least related. Unfortunately the limit is higher on ARM, which will make it harder to test. (And I will *not* try to get an x86_64 Python running on my machine ;) )

rouault:
@emlys Would you be willing to test this pull request, somewhat related to #8497 you created some time ago, and disable the ``#undef HAVE_PTHREAD_ATFORK`` at line 43 of ogr_proj_p.cpp ( https://github.com/OSGeo/gdal/pull/8909/files#diff-c516e8cff204f7f45759b9972d20b8ae8113468623a1c0c594d339e7bcf9055aR43  ) on top of this to PR ? Cf above comment https://github.com/OSGeo/gdal/pull/8909#issuecomment-1839196582 that explains it is likely that the fix of this PR obsoletes the ``#undef HAVE_PTHREAD_ATFORK``  trick added in https://github.com/OSGeo/gdal/pull/8524

carsten-schultz-pix4d:
Okay, this is what I tested.

I started at v3.6.4, because that's where I happen to still be. I built with Python bindings, set all of the necessary paths....
I tested `test.py` from https://github.com/OSGeo/gdal/issues/8497. As expected, on my machine the number of iterations for which it started to fail was a bit below `16 * 1024 // 24`. (Indeed at `-7`, with the kind of failure depending on the exact value, segfault, bus error, not producing the final output.)

I then cherry-picked the commit from this PR, and tested the same. Works, problem fixed.

PR #8921, opened 2023-12-06
GetExtent3D(): better docs, and make sure that it takes into account …

rouault:
…attribute and spatial filters (in shapefile implementation)

CC @elpaso I've had a chat with @benoitdm-oslandia regarding how to integrate OGR_L_GetExtent3D() in QGIS, and it appears that guaranteing that it takes into account filters will make things simpler on QGIS side. Hence this change. From what I saw, the GeoPackage specialized impl will also apply the filter (could be appropriate to add a quick test for that), so we should be good.

### DIFF
--- a/ogr/ogr_api.h
+++ b/ogr/ogr_api.h
@@ -765,15 +765,6 @@ GIntBig CPL_DLL OGR_L_GetFeatureCount(OGRLayerH, int);
 OGRErr CPL_DLL OGR_L_GetExtent(OGRLayerH, OGREnvelope *, int);
 OGRErr CPL_DLL OGR_L_GetExtentEx(OGRLayerH, int iGeomField,
                                  OGREnvelope *psExtent, int bForce);
-/**
- * @brief OGR_L_GetExtent3D computes the 3D extent of the layer.
- * @param hLayer the layer to consider.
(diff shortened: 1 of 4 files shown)


benoitdm-oslandia:
Just a thought: you will have 2 different behaviours regarding the attribute filter between the `GetExtent` and `GetExtent3D`. 
May be you could have a extra parameter on `GetExtent` (default to false to preserve compatibility) to handle with/without attribute filter beheviour (one redirecting to the old `GetExtent` and the other one the wrapped of `GetExtent3D`)?

rouault:
> Just a thought: you will have 2 different behaviours regarding the attribute filter between the `GetExtent` and `GetExtent3D`.
> May be you could have a extra parameter on `GetExtent` (default to false to preserve compatibility) to handle with/without attribute filter beheviour (one redirecting to the old `GetExtent` and the other one the wrapped of `GetExtent3D`)?

I can see your point about consistency. That said, adding an additional argument to GetExtent() would require changes in all drivers that implement it, and is IMHO not worth the trouble. And changing the behaviour would also certainly cause interesting issues in code that rely on current behavior. So I'm inclined to defer such change. Users will be able to use GetExtent3D() as a workaround if they want to have an extent taking into account filters.

PR #8927, opened 2023-12-06
/vsicurl/: re-emit HTTP error code next times we try opening a resource that failed the first time (fixes #8922)

### LINKED ISSUE #8922: Http response errors in VSI curl VFS are only returned once

When the VSI curl virtual filesystem encounters a non-200 status code, it's intended to set an error with the status code that was returned, see here - https://github.com/OSGeo/gdal/blob/master/port/cpl_vsil_curl.cpp#L1506

Somehow this error state doesn't seem to be being cleared correctly though, as subsequent reads of the same (failed) url return a much more generic and less descriptive error message.

This Python script reproduces the issue reliably: 

```python
from osgeo import gdal

url = 'http://example.com/path/to/your/file.tif'

gdal.UseExceptions()

print("=== ATTEMPT 1 ===")

try:
    dataset = gdal.Open(f'/vsicurl/{url}')
except Exception as e:
    print(f'Error: {e}')

print("=== ATTEMPT 2 ===")

try:
    dataset = gdal.Open(f'/vsicurl/{url}')
except Exception as e:
    print(f'Error: {e}')
```

Running it using the latest ubuntu-full docker image gives:

```
root@67b6c616908d:/# gdalinfo --version
GDAL 3.9.0dev-c1369ecc4b9a36315d1e9c685f4153c23f735386, released 2023/11/28

root@67b6c616908d:/# python bug.py
=== ATTEMPT 1 ===
Error: HTTP response code: 404
=== ATTEMPT 2 ===
Error: `/vsicurl/http://example.com/path/to/your/file.tif' does not exist in the file system, and is not recognized
 as a supported dataset name.
```

As you can see, the second time the same call is made the error message gives a lot less indication of what the problem is. This is a big problem for us running it in a long-lived web server process and needing to distinguish between authentication/authorisation/other http errors.

Running with debug logs enabled gives an indication that we may only be making one http request and then mistakenly caching the result. Setting `VSI_CACHE=FALSE` made no difference though.

```
root@67b6c616908d:/# CPL_DEBUG=ON python bug.py
=== ATTEMPT 1 ===
HTTP: libcurl/7.81.0 OpenSSL/3.0.2 zlib/1.2.11 brotli/1.0.9 zstd/1.4.8 libidn2/2.3.2 libpsl/0.21.0 (+libidn2/2.3.2) libssh/0.9.6/openssl/zlib nghttp2/1.43.0 librtmp/2.3 OpenLDAP/2.5.16
VSICURL: GetFileSize(http://example.com/path/to/your/file.tif)=0  response_code=404
VSICURL: GetFileList(/vsicurl/http://example.com/path/to/your)
Error: HTTP response code: 404
=== ATTEMPT 2 ===
Error: `/vsicurl/http://example.com/path/to/your/file.tif' does not exist in the file system, and is not recognized as a supported dataset name.
```

## Operating system

Ubuntu 22.04 64-bit

Reproduced on `GDAL 3.9.0dev-c1369ecc4b9a36315d1e9c685f4153c23f735386, released 2023/11/28`, from the public container registry (https://github.com/OSGeo/gdal/pkgs/container/gdal/152385880?tag=ubuntu-full-latest)

rouault:


mrob95:
Many thanks!

PR #8932, opened 2023-12-07
gdal_footprint: write source dataset path in a 'location' field (fixes #8795)     

### LINKED ISSUE #8795: gdal_footprint: write the filename as an attribute
For making it easier to use the footprints afterwards it would be good to write the filename into the vector data as an attribute. Gdaltindex is using field "location" for that purpose by default so gdal_footprint could use the same. Gdaltindex has also an option "write_absolute_path" that is useful sometimes.

Saving the filename could be the default behavior. If there is some good reason for not saving the filename it could be skipped for example with some option like `-nopath`.

rouault:
New options:
```
.. option:: -location_field_name <field_name>

    .. versionadded:: 3.9.0

    Specifies the name of the field in the resulting vector dataset where the
    path of the input dataset will be stored. The default field name is
    "location". To prevent writing the path of the input dataset, use
    :option:`-no_location`

.. option:: -no_location

    .. versionadded:: 3.9.0

    Turns off the writing of the path of the input dataset as a field in the
    output vector dataset.

.. option:: -write_absolute_path

    .. versionadded:: 3.9.0

    Enables writing the absolute path of the input dataset. By default, the
    filename is written in the location field exactly as specified on the
    command line.
```
First commit is a bugfix that must be backported

### DIFF
--- a/apps/gdal_footprint_bin.cpp
+++ b/apps/gdal_footprint_bin.cpp
@@ -49,6 +49,8 @@ static void Usage(bool bIsError, const char *pszErrorMsg = nullptr)
             "       [-convex_hull] [-densify <value>] [-simplify <value>]\n"
             "       [-min_ring_area <value>] [-max_points <value>|unlimited]\n"
             "       [-of <ogr_format>] [-lyr_name <dst_layername>]\n"
+            "       [-location_field_name <field_name>] [-no_location]\n"
+            "       [-write_absolute_path]\n"
             "       [-dsco <name>=<value>]... [-lco <name>=<value>]... "
             "[-overwrite] [-q]\n"
             "       <src_filename> <dst_filename>\n");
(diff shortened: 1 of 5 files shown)


jratike80:
I think yes. I do not quite understand the difference between `locationFieldName=None` and `-no_location` but I guess there is a reason.
I have needed the feature during a few last weeks and my workaround was to use temporary files and ogr2ogr with sql
`-sql "select geometry as geom,'%1.tif' as location from footprint" %1.json`.

rouault:
> do not quite understand the difference between `locationFieldName=None` and `-no_location` but I guess there is a reason.

locationFieldName=None is for Python. In Python, None is different from the "None" string. So a single switch can be used to specify the field name, or disable it.
For a command line utility, we need 2 separate switches. Or we'd need to agree on a special value for the field name that would mean to disable it. That could potentially be "none"... but that would prevent someone from naming their location field "none"...

PR #8935, opened 2023-12-07
gdal_translate: add -dmo option, for domain-metadata items

mdsumner:
adds `-dmo` for adding one or more domain specific metadata items, as a complement to `-mo` which adds `metadata=value` items, we can specify `domain:metadata=value` with dmo, e.g. 

```bash
gdal_translate in.tif out.vrt -dmo DOMAIN:METATAG1=VALUE1 -dmo DOMAIN:METATAG1=VALUE
```

(note that random domains won't be reported by gdalinfo, so the effect may not be obvious without looking at the VRT file or querying the dataset). 

The motivating example is to add GEOLOCATION domain elements, e.g. 

```bash
gdal_translate in.tif out.vrt -dmo GEOLOCATION:X_DATASET=x.tif -dmo GEOLOCATION:Y_DATASET=y.tif ...
```

Discussed on the mailing list: https://lists.osgeo.org/pipermail/gdal-dev/2023-September/057665.html

Also removed outdated comments: 

https://lists.osgeo.org/pipermail/gdal-dev/2023-September/057677.html

## Tasklist

 - [x] need more test cases for the python API
 - [x] Add documentation
 - [x] Update Python API 
 - [ ] Review
 - [ ] Adjust for comments

I'm using key/valure parsing to get tokens from colon and = separators, I think this means it's not possible to have domain or key tags with colons in them .... I think that's ok, you can use the API for weirdo labels (and that needs tests added still).

### DIFF
--- a/apps/gdal_translate_lib.cpp
+++ b/apps/gdal_translate_lib.cpp
@@ -2573,24 +2573,25 @@ static void AttachDomainMetadata(GDALDatasetH hDS,
     {
 
         char *pszKey = nullptr;
-        char *pszKey2 = nullptr;
-        // parse twice because the first time picks the colon ":", second picks the "="
-        //    (so we can't have domain names with colons but we can have values like "OGC:CRS84")
-        // first key is E1 from E1:E2=E3
-        // value is E2=E3
-        // then second key is E2, second value is E3
-        const char *pszValue =
-            CPLParseNameValue(aosDomainMetadataOptions[i], &pszKey);
-        const char *pszValue2 = CPLParseNameValue(pszValue, &pszKey2);
+        char *pszDomain = nullptr;
 
-        if (pszKey && pszValue && pszKey2 && pszValue2)
+        // parse the DOMAIN:KEY=value, Remainder is KEY=value
+        const char *pszRemainder =
+            CPLParseNameValueSep(aosDomainMetadataOptions[i], &pszDomain, ':');
+
+        if (pszDomain && pszRemainder)
         {
-            CPLDebug("translate", "pszValue2: %s, pszKey2: %s, pszKey: %s",
-                     pszValue2, pszKey2, pszKey);
-            GDALSetMetadataItem(hDS, pszKey2, pszValue2, pszKey);
+
+            const char *pszValue =
+                CPLParseNameValueSep(pszRemainder, &pszKey, '=');
+            if (pszKey && pszValue)
+            {
+                GDALSetMetadataItem(hDS, pszKey, pszValue, pszDomain);
+            }
         }
         CPLFree(pszKey);
-        CPLFree(pszKey2);
+
+        CPLFree(pszDomain);
     }
 }
 
--- a/apps/gdal_translate_lib.cpp
+++ b/apps/gdal_translate_lib.cpp
@@ -58,6 +58,8 @@
 #include "vrtdataset.h"
 
 static void AttachMetadata(GDALDatasetH, const CPLStringList &);
+static void AttachDomainMetadata(GDALDatasetH, const CPLStringList &);
+
 static void CopyBandInfo(GDALRasterBand *poSrcBand, GDALRasterBand *poDstBand,
                          int bCanCopyStatsMetadata, int bCopyScale,
                          int bCopyNoData, bool bCopyRAT,
@@ -189,10 +191,11 @@ struct GDALTranslateOptions
 
     bool bHasUsedExplicitExponentBand = false;
 
-    /*! list of metadata key and value to set on the output dataset if possible.
-     *  GDALTranslateOptionsSetMetadataOptions() and
-     * GDALTranslateOptionsAddMetadataOptions() should be used */
+    /*! list of metadata key and value to set on the output dataset if possible. */
     CPLStringList aosMetadataOptions{};
+
+    /*! list of metadata key and value in a domain to set on the output dataset if possible. */
+    CPLStringList aosDomainMetadataOptions{};
 
     /*! override the projection for the output file. The SRS may be any of the
        usual GDAL/OGR forms, complete WKT, PROJ.4, EPSG:n or a file containing
@@ -1168,10 +1171,10 @@ GDALDatasetH GDALTranslate(const char *pszDest, GDALDatasetH hSrcDataset,
         psOptions->asScaleParams.empty() && psOptions->adfExponent.empty() &&
         !psOptions->bUnscale && !psOptions->bSetScale &&
         !psOptions->bSetOffset && psOptions->aosMetadataOptions.empty() &&
-        bAllBandsInOrder && psOptions->eMaskMode == MASK_AUTO &&
-        bSpatialArrangementPreserved && !psOptions->bNoGCP &&
-        psOptions->nGCPCount == 0 && !bGotBounds && !bGotGeoTransform &&
-        psOptions->osOutputSRS.empty() &&
+        psOptions->aosDomainMetadataOptions.empty() && bAllBandsInOrder &&
+        psOptions->eMaskMode == MASK_AUTO && bSpatialArrangementPreserved &&
+        !psOptions->bNoGCP && psOptions->nGCPCount == 0 && !bGotBounds &&
+        !bGotGeoTransform && psOptions->osOutputSRS.empty() &&
         psOptions->dfOutputCoordinateEpoch == 0 && !psOptions->bSetNoData &&
         !psOptions->bUnsetNoData && psOptions->nRGBExpand == 0 &&
         !psOptions->bNoRAT && psOptions->anColorInterp.empty() &&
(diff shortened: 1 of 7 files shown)


apps/gdal_translate_lib.cpp
```diff
@@ -2553,6 +2559,31 @@ static void AttachMetadata(GDALDatasetH hDS,
     }
 }
 
+/************************************************************************/
+/*                           AttachDomainMetadata()                           */
```
rouault, line 2563: extra spaces

apps/gdal_translate_lib.cpp
```diff
@@ -2553,6 +2559,31 @@ static void AttachMetadata(GDALDatasetH hDS,
...
+
+static void AttachDomainMetadata(GDALDatasetH hDS,
+                                 const CPLStringList &aosDomainMetadataOptions)
+
+{
+    const int nCount = aosDomainMetadataOptions.size();
+
+    for (int i = 0; i < nCount; i++)
+    {
+        // FIXME: catch invalid cases syntax is DOMAIN:META-TAG=value
```
rouault, line 2574: yes definitely something to do.
I would do the CSLTokenizeString2() on the pszKey, so as to be able to support  DOMAIN:FOO=BAR:WITH:COLUMN
And you should check that aosTokens.size() == 2

mdsumner: ah! I see, thanks - should I error out if it's not size 2, or just ignore it?

rouault: > should I error out if it's not size 2

would make sense to emit a CPLError(CE_Warning, CPLE_IllegalArg, "....");

mdsumner: ignore it but with a warning message of an ill-formed argument seems right

mdsumner: CPLParseNameValue parses on "=" or ":" so I need to have  closer look

rouault: > CPLParseNameValue parses on "=" or ":" so I need to have closer look

ah, good point. You'll have to use strchr(str, '=') then to do the splitting.

autotest/utilities/test_gdal_translate.py
```diff
@@ -378,6 +378,27 @@ def test_gdal_translate_13(gdal_translate_path, tmp_path):
...
+###############################################################################
+# Test -dmo option
+
+
+def test_gdal_translate_dmo_option(gdal_translate_path, tmp_path):
+
+    dst_vrt = str(tmp_path / "test_dmo.vrt")
+
+    gdaltest.runexternal(
+        f"{gdal_translate_path} -dmo NEW_DOMAIN:META-TAG=value ../gcore/data/byte.tif {dst_vrt}"
```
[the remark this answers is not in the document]
mdsumner: I'm going to have to do some homework for this python dictionary stuff, keys are unique right so it can' be 

```python
{ "domain_name" : { "key": "value" } , "domain_name" : { "key1": "value1" } }

```

but I guess a list of dictionaries is pythonic? 

```python
[ {"domain_name" : { "key": "value" }} , {"domain_name" : { "key1": "value1" }} ]
```

i'll explore what the api does now and the kinds of structures it deals with  🙏 this is good homework for me :)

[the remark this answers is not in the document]
rouault: > I'm going to have to do some homework for this python dictionary stuff, keys are unique right so it can't be
> 
> ```python
> 
> ```

I was thinking to ``{ "domain1": { "key1": "value1", "key2": "value2" }, "domain2": .... } ``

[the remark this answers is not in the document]
mdsumner: ah, lol ... right thanks!

autotest/utilities/test_gdal_translate_lib.py
```diff
@@ -1158,6 +1158,57 @@ def test_gdal_translate_lib_dict_arguments():
...
+    ## string dmo input
+    ds = gdal.Translate(
+        dst_vrt,
+        "../gcore/data/byte.tif",
+        domainMetadataOptions="NEW_DOMAIN:META-TAG=value",
+    )
+
+    md = ds.GetMetadata("NEW_DOMAIN")
+    assert "META-TAG" in md, "Did not get META-TAG"
+
```
rouault, line 1178:
```suggestion
    assert "META-TAG" in md, "Did not get META-TAG"
    assert md["META-AG"] = "value"
```

autotest/utilities/test_gdal_translate_lib.py
```diff
@@ -1158,6 +1158,57 @@ def test_gdal_translate_lib_dict_arguments():
...
+        "../gcore/data/byte.tif",
+        domainMetadataOptions=[
+            "NEW_DOMAIN:META-TAG=value",
+            "ANOTHER_NEW_DOMAIN:META2-TAG=value",
+        ],
+    )
+
+    md = ds.GetMetadata("ANOTHER_NEW_DOMAIN")
+    assert "META2-TAG" in md, "Did not get META2-TAG"
+
```
rouault, line 1193:
```suggestion
    md = ds.GetMetadata("NEW_DOMAIN")
    assert "META-TAG" in md, "Did not get META-TAG"
    assert md["META-TAG"] == "value"
    md = ds.GetMetadata("ANOTHER_NEW_DOMAIN")
    assert "META2-TAG" in md, "Did not get META2-TAG"
    assert md["META2-TAG"] == "value"

```

autotest/utilities/test_gdal_translate_lib.py
```diff
@@ -1158,6 +1158,57 @@ def test_gdal_translate_lib_dict_arguments():
...
+        "domain_2": {"key2": "value2"},
+    }
+    ds = gdal.Translate(
+        "/vsimem/dmo.vrt",
+        gdal.Open("../gcore/data/byte.tif"),
+        domainMetadataOptions=items,
+    )
+    md = ds.GetMetadata("domain_2")
+    assert "key2" in md, "Did not get key2"
+
```
rouault, line 1208:
```suggestion
    assert "key2" in md, "Did not get key2"
    assert md["key2"] == "value2"
```

autotest/utilities/test_gdal_translate_lib.py
```diff
@@ -1158,6 +1158,57 @@ def test_gdal_translate_lib_dict_arguments():
...
+    ds = None
+
```
rouault, line 1210: It would be good to add tests for invalid value of domainMetadataOptions, like "FOO" and "FOO=BAR"

mdsumner:
I got a bit lost in this, but will try to integrate the various parts. There's a problem already there with CPLParseNameValue() splitting on ":" or "=" so there'll be a follow up PR for that. 

This gives `tag` = `mag=value`, so the whole strchr thing should happen here too.  I don't think we can safely use CPLNameParseValue() at all here.  

```
gdal_translate new1.tif -mo tag:mag=value md.tif && gdalinfo md.tif  -listmdd | grep Metadata -A 10

```

Curently this is valid syntax in place of "tag=value": 

````
gdal_translate in.tif out.tif -mo tag:value

```



Should we guard against elements domain, tag, value containing ":" or "="? I don't think so, just go with these rules is my current take: 

mo: "split on first '='", error out if either of tag,value is empty. 

dmo: "split on first '='", error out if either domaintag,value empty, then "split domaintag on first ':'", error out if either of domain,tag is empty.

mdsumner:
that approach is working for me, is it permissible to add 

```cpp
const char CPL_DLL *CPLParseNameValueSep(const char *pszNameValue, char **ppszKey, char chSep);
```

to cpl_string?   so that we can do 

```c++
 const char *pszRemainder =
            CPLParseNameValueSep(aosDomainMetadataOptions[i], &pszDomain, ':');

// ... include checks
const char *pszTag = CPLParseNameValueSep(pszRemainder, &pszValue, '=');
//... include checks
```

Other ways I tried weren't successful yet ...

rouault:
> ```c++
> const char CPL_DLL *CPLParseNameValueSep(const char *pszNameValue, char **ppszKey, char chSep);
> ```

looks good to me!

mdsumner:
I've put in tests for what works and does not, but there's no error for invalid input - the same is true for '-mo': 

```
-mo will:work -mo wontwork
```

## only the third one is invalid, and silently ignored
```
-dmo DOMAIN:KEY=value -dmo GEOLOCATION:X_DATASET=x.tif -dmo THIS:wont:work -dmo THIS=WILL:work:but=weirdeh
```

It doesn't seem appropriate to close the dataset and error out in AttachDomainMetadata()` ?  Validating metadata seems like a bigger project for later given the other issues mentioned in https://github.com/OSGeo/gdal/pull/8458#issuecomment-1732274963

rouault:
> It doesn't seem appropriate to close the dataset and error out in AttachDomainMetadata()` ? Validating metadata seems like a bigger project for later

yes. We just need to make sure not to hard crash on invalid input.

mdsumner:
thanks for all the help, learnt a lot 👌

port/cpl_string.cpp
```diff
@@ -1804,6 +1804,54 @@ const char *CPLParseNameValue(const char *pszNameValue, char **ppszKey)
...
+ * Parse NAME<Sep>VALUE string into name and value components.
+ *
+ * This is derived directly from CPLParseNameValue() which will separate
+ * on '=' OR ':', here chSep is required for specifying the separator
+ * explicitly.
+ *
+ * @param pszNameValue string in "NAME=VALUE" format.
+ * @param ppszKey optional pointer though which to return the name
+ * portion.
+ * @pararm chSep required single char separator
```
rouault, line 1820:
```suggestion
 * @param chSep required single char separator
```

Issue #8938, opened 2023-12-08
Problem with FlushCache(true) to mimic a cache "drop"


chacha21:
If I remember correclty, the `bool bAtClosing = false` parameter of `GDALRasterBand::FlushCache()` has been introduced in #4652.
However, I have a scenario where it was misleading and lead me to a very unexpected and hard to find bug (my fault, but still !).

The scenario is :

1. I use a dataset D as a temporary space to store processed data that is regularly updated.
2. I don't care if everything occurs in cache, it will swap to disk as necessary, I just read to and write from D.
3. At the end of my work session, the remaining data in D is no more relevant.
4. I want to "drop" the cache of D to release resources and free cache for other uses, without paying the penalty of flushing to disk, since data is no more relevant.
5. I call D.FlushCache(true), (which has the effect of setting `m_bWriteDirtyBlocks` to `false` inside `DisableDirtyBlockWriting()`)
6. I start a new work session, reusing D as an intermediate storage
7. Since nothing reverted `m_bWriteDirtyBlocks` the cache is permanently marked as "not flushable" (that's the reason of the bug to come)
8. For some reason, the data remaining in D will be relevant that time, and I want to save it.
9. I call some `GDALDriver::CreateCopy(D)`
10. However, since the cache cannot be flushed any more, what I get is old deprecated data that was swapped to disk during previous work sessions.

The problem is in step 5 where I lied to `FlushCache()`, because I am not really closing the dataset, I just wanted to drop the cache without penalty.

Is it possible to either  restore `m_bWriteDirtyBlocks = true` after a `FlushCache(true)`, or to get a new native DropCache() feature that leaves `m_bWriteDirtyBlocks ` untouched ?

chacha21:
A proposal (not sure if it's enough, just to locate the problem)

```
CPLErr GDALRasterBand::FlushCache(bool bAtClosing)

{
    if (bAtClosing && poDS && poDS->bSuppressOnClose && poBandBlockCache)
        poBandBlockCache->DisableDirtyBlockWriting();

    CPLErr eGlobalErr = eFlushBlockErr;

    if (eFlushBlockErr != CE_None)
    {
        ReportError(
            eFlushBlockErr, CPLE_AppDefined,
            "An error occurred while writing a dirty block from FlushCache");
        eFlushBlockErr = CE_None;
    }

    if (poBandBlockCache == nullptr || !poBandBlockCache->IsInitOK())
        return eGlobalErr;

[...]

PR #8940, opened 2023-12-08
added DropCache to GDALDataset

chacha21:
DropCache() release cache blocks but does not write cache to disk. It is similar to FlushCache(true) but does not assume that the dataset will be closed.

proposal for #8938

### DIFF
--- a/autotest/cpp/test_gdal.cpp
+++ b/autotest/cpp/test_gdal.cpp
@@ -3498,7 +3498,7 @@ TEST_F(test_gdal, drop_cache)
     {
         GDALDriverManager* gdalDriverManager = GetGDALDriverManager();
         GDALDriver* enviDriver = !gdalDriverManager ? nullptr : gdalDriverManager->GetDriverByName("ENVI");
-        char* enviOptions[] = {
+        const char* enviOptions[] = {
             "SUFFIX=ADD",
             "INTERLEAVE=BIL",
             nullptr
--- a/autotest/cpp/test_gdal.cpp
+++ b/autotest/cpp/test_gdal.cpp
@@ -3497,7 +3497,9 @@ TEST_F(test_gdal, drop_cache)
     CPLErrorReset();
     {
         GDALDriverManager* gdalDriverManager = GetGDALDriverManager();
-        GDALDriver* enviDriver = !gdalDriverManager ? nullptr : gdalDriverManager->GetDriverByName("ENVI");
+        GDALDriver* enviDriver = !gdalDriverManager ?
+            nullptr :
+            gdalDriverManager->GetDriverByName("ENVI");
         const char* enviOptions[] = {
             "SUFFIX=ADD",
             "INTERLEAVE=BIL",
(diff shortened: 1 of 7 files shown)


gcore/gdalrasterband.cpp
```diff
@@ -1140,6 +1140,68 @@ CPLErr CPL_STDCALL GDALFlushRasterCache(GDALRasterBandH hBand)
...
+
+/**
+* \brief Drop raster data cache : data in cache will be lost.
+*
+* This call will recover memory used to cache data blocks for this raster
+* band, and ensure that new requests are referred to the underlying driver.
+*
+* This method is the same as the C function GDALDropRasterCache().
+*
+* @return CE_None on success.
```
rouault, line 1155:
```suggestion
* @return CE_None on success.
* @since 3.9
```

gcore/gdalrasterband.cpp
```diff
@@ -1140,6 +1140,68 @@ CPLErr CPL_STDCALL GDALFlushRasterCache(GDALRasterBandH hBand)
...
+}
+
+/************************************************************************/
+/*                        GDALDropRasterCache()                         */
+/************************************************************************/
+
+/**
+* \brief Drop raster data cache.
+*
+* @see GDALRasterBand::DropCache()
```
rouault, line 1194:
```suggestion
* @see GDALRasterBand::DropCache()
* @since 3.9
```

gcore/gdal_priv.h
```diff
@@ -537,6 +537,7 @@ class CPL_DLL GDALDataset : public GDALMajorObject
     Bands GetBands();
 
     virtual CPLErr FlushCache(bool bAtClosing = false);
+    virtual CPLErr DropCache();
```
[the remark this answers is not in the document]
chacha21: Honestly it is just copy/paste from FlushCache().
I think `virtual` is harmless and make it ready for future specific implementation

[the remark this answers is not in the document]
rouault: ok

chacha21:
> We'd need a test for that. Presumably in autotest/cpp/test_gdal.cpp (run by autotest/cpp/gdal_unit_test)

What kind of test should be written ?

chacha21:
I was not able to run the test, there's something broken with my gtests

chacha21:
Does `DropCache()` make `FlushCache(true)` obsolete ?
I mean , instead of having the `bClosing` parameter, one could instead use `DropCache()` at the beginning of a dataset destructor in the case of `MarkSuppressOnClose()`, so that every subsequent `FlushCache()` would do nothing.

rouault:
> Does `DropCache()` make `FlushCache(true)` obsolete ? I mean , instead of having the `bClosing` parameter, one could instead use `DropCache()` at the beginning of a dataset destructor in the case of `MarkSuppressOnClose()`, so that every subsequent `FlushCache()` would do nothing.

yes, that mostly obsoletes it. That said there are a few drivers that use now the bAtClosing flag to save some work (cf frmts/rmf/rmfdataset.cpp)

autotest/cpp/test_gdal.cpp
```diff
@@ -3491,4 +3491,48 @@ TEST_F(test_gdal, open_shared_open_options)
     }
 }
 
+// Test GDAL_OF_SHARED flag and open options
```
rouault, line 3494: this comment should be updated or removed

rouault:
@chacha21 are you done with this PR?

chacha21:
> @chacha21 are you done with this PR?

As long as you have no more comments, I have nothing remaining to be committed.
I just could not run the test myself (can't explain why, gtest is crashing on my machine), so I just have to trust the buildbot.

I ran the test code manually and it behaved as expected, though. (GDALChecksumImage returns 0)

PR #8951, opened 2023-12-10
-rl vs. -sql: warn now on the man page before the user attempts mixing the two

jidanni:
Not heeding the warning that I am putting on this man page will cause an "-rl is incompatible with -sql" error message.

I bet there are other man pages that need this warning too...


jratike80:
Probably `-rl` don't work or at least cannot give speed ups with anything that selects only one layer out of some multilayer datasource. Or when the datasource has only one layer.

rouault:
> Probably `-rl` don't work or at least cannot give speed ups with anything that selects only one layer out of some multilayer datasource.

-rl only makes sense when several layers as selected (hence not when -sql is used), and for some particular drivers as mentioned in the doc. OK, merging that one

PR #8975, opened 2023-12-18
FileGDB/OpenFileGDB: fix co-operation between the 2 drivers to make sure .cdf is opened with FileGDB

rouault:


geographika:
I think I ran into an error possibly fixed by this this with a recent MapServer/GDAL build - the same Mapfile and plugin setup and the following started to return the error `msInitDefaultOGROutputFormat(): General error message. OGR `FileGDB' driver does not support output.`

The MapServer check for this is `  if (!OGR_Dr_TestCapability(hDriver, ODrCCreateDataSource)) {`

```
    OUTPUTFORMAT
        NAME "FileGDB"
        DRIVER "OGR/FileGDB"
```

The OpenFileGDB however worked fine:

```
    OUTPUTFORMAT
        NAME "FileGDB"
        DRIVER "OGR/OpenFileGDB"
```

rouault:
> I think I ran into an error possibly fixed by this

I doubt so... This PR just fixes an obscure use case, and is not related to the creation code path. I've no theory why OGR_Dr_TestCapability(hDriver, ODrCCreateDataSource) would fail on the FileGDB driver. That said, at this point, you should be able to safely use the OpenFileGDB one

geographika:
Ah ok - I saw this check-in just as I ran into that error. Switching to OpenFileGDB resolved all issues. Sorry for the noise.

Issue #8980, opened 2023-12-19
MarkSuppressOnClose() : no IsSuppressOnClose() or UnMarkSuppressOnClose()


chacha21:
GDAL 3.8

Currently `GDALDataset::MarkSuppressOnClose()` just sets the internal `bSuppressOnClose`
I would need a`GDALDataset::IsSuppressOnClose()` to query that value
I would also like a`GDALDataset::UnmarkSuppressOnClose()` to revert that value

It seems trivial to implement, but before issueing a PR, do you know if there are any reason NOT to implement such functions ?
`IsSuppressOnClose()` must be harmless. But would `UnmarkSuppressOnClose()` expose to misleading scenarios ?

[edit]
For the context :
I want to reuse a dataset to store data with similar size and layout. But If the new data must be persistent, I must know if the dataset is doomed by bSuppressOnClose and should not be used for that usage.

rouault:
I don't see strong reasons not to implement those methods

PR #8981, opened 2023-12-19
added additional functions to change the bSuppressOnClose flag

chacha21:
GDALDataset::UnMarkSuppressOnClose()
GDALDataset::IsMarkSuppressOnClose()

related to #8980

### DIFF
--- a/autotest/cpp/test_gdal.cpp
+++ b/autotest/cpp/test_gdal.cpp
@@ -2533,6 +2533,35 @@ TEST_F(test_gdal, MarkSuppressOnClose)
     }
 }
 
+// Test effect of UnMarkSuppressOnClose()
+TEST_F(test_gdal, UnMarkSuppressOnClose)
+{
+    const char *pszFilename = "/vsimem/out.tif";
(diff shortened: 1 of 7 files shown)


autotest/cpp/test_gdal.cpp
```diff
@@ -2533,6 +2533,35 @@ TEST_F(test_gdal, MarkSuppressOnClose)
...
+        // All buffers have been flushed, and our dirty block should have
+        // been written hence the checksum will not be 0
+        EXPECT_NE(GDALChecksumImage(
+                      GDALRasterBand::FromHandle(poDstDS->GetRasterBand(1)), 0,
+                      0, 1, 1),
+                  0);
+    }
+    {
+        VSIStatBufL sStat;
+        EXPECT_TRUE(VSIStatL(CPLSPrintf("%s.aux.xml", pszFilename), &sStat) ==
```
rouault, line 2560: Writing metadata and creating a .aux.xml is unnecessary to test the effect of UnMarkSuppressOnClose. Just checking for the present of pszFilename should be enough.
VSIUnlink(pszFilename) should be called for cleanup.

PR #8994, opened 2023-12-23
Add specific documentation section about AWS IMDS (and fix typo)

rouault:
Credits to @pedros007 for diagnosing the issue and finding the workarounds


doc/source/user/virtual_file_systems.rst
```diff
@@ -625,7 +625,28 @@ The :config:`CPL_VSIS3_CREATE_DIR_OBJECT` configuration option can be set to NO
...
+    aws ec2 modify-instance-metadata-options \
+        --instance-id <instance_id> \
+        --http-put-response-hop-limit 2 \
+        --http-endpoint enabled
+
+Another is to set the HttpPutResponseHopLimit metadata on an AutoScalingGroup LaunchTemplate:
+- https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_InstanceMetadataOptionsRequest.html
+- https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-launchtemplate-metadataoptions.html
+
+Another possibility is to start the Docker container with host networking (``--network=host``).
```
pedros007, line 649: Docker host networking should maybe be frowned upon. It grants your container access to _all_ the ports of the host. 

>  from a security perspective, Docker containers can listen on many ports, even though they only advertise (expose) a single port. Normally this is fine as you only forward the single expected port, however if you use --net=host then you'll get all the container's ports listening on the host, even those that aren't listed in the Dockerfile. This means you will need to check the container closely (especially if it's not yours, e.g. an official one provided by a software project) to make sure you don't inadvertently expose extra services on the machine. [source](https://stackoverflow.com/a/57051970/40785)

If you had two different containers running on one host with `--network=host` I think they could access each others ports.  This could happen on large instances that a scheduler like ECS or Kubernetes can assign many containers to run on a single host.  Because of this I think increasing the HTTP Put Response Hop Limit might be a better solution.

rouault: > Docker host networking should maybe be frowned upon

thanks. I've added a note about this

PR #8995, opened 2023-12-23
VFK: Fix solving circle parcel geometry (fixes #8993)

### LINKED ISSUE #8993: VFK: Generated invalid circle shaped parcel geometries

In some cases, parcel geometries defined by three points are wrongly generated. The problem is in computing center point and farther  four points on circle in axes direction.

Import SGI and SPI file 616397.vfk and look at parcel ID 1390618203.

## Operating system
Debian 11

build from source Github master on 22.12.2023

seidlmic:
## What does this PR do?

It seems that current code generates in some situation wrong shaped parcels. Probably due to unexpected point order which is not in VFK format guaranteed.

Because it is not obvious how to fix  current algorithm it was replaces with new one based on solving 3 equations by computing determinant of system. The percentage of circle shaped parcel is very low so computing time is not so important.

The advantage of the solution with determinant is, it does not need to care about points order or their configuration. Only the value of determinant should be tested. If it is too small (the area off 3 point triangle) means invalid configuration.  If determinant is too large no reasonable size of circle in case of VFK data.

These situation should be handle in method _GetCircleCenterFrom3Points()_ but it is not clear to me how to do it directly in method.

The fix should be confirm by @landam as the author of original code.

To test the code run:
```
/usr/local/gdal/test/bin/ogr2ogr --config OGR_VFK_DB_OVERWRITE YES --config OGR_VFK_DB_NAME 616397.db --config OGR_VFK_DB_READ_ALL_BLOCKS N0 -f SHP 616397_PAR.SHP 2020-09-01_sgi/616397.vfk PAR

 /usr/local/gdal/test/bin/ogr2ogr --config OGR_VFK_DB_OVERWRITE NO --config OGR_VFK_DB_NAME 616397.db --config OGR_VFK_DB_READ_ALL_BLOCKS N0  616397_PAR.SHP 2020-09-01_spi/616397.vfk PAR
```

## Environment

[616397_sgi.zip](https://github.com/OSGeo/gdal/files/13758784/616397_sgi.zip)
[616397_spi.zip](https://github.com/OSGeo/gdal/files/13758785/616397_spi.zip)

The circle parcel ID is 1390618203


rouault:
Code formatting issues:
```If you are seeing this message in CI, reproduce locally with: `pre-commit run --all-files`.
To run `pre-commit` as part of git workflow, use `pre-commit install`. ```

Cf https://gdal.org/development/dev_practices.html#commit-hooks how to install and configure pre-commit

File: ogr/ogrsf_frmts/vfk/vfkfeature.cpp [hunk 1]
```diff
@@ -86,6 +86,116 @@ void IVFKFeature::SetFID(GIntBig nFID)
...
+}
+
+/*!
+  \brief Find circle center determined by three point
+
+  \param c_xy circle center coordinates array
+  \param x array of three x coordinates
+  \param y array of three x coordinates
+
+  \return void
```
rouault, line 112: clang doesn't like return void:

```suggestion
```

ogr/ogrsf_frmts/vfk/vfkreader.h
```diff
@@ -116,6 +116,11 @@ class VFKProperty
 /************************************************************************/
 class IVFKFeature
 {
+  private:
+    double GetDeterminatOfMatrixDim3(double [3], double [3], double [3]);
+    void GetCircleCenterFrom3Points(double [2], double [3], double [3]);
+    void AddCirclePointsToGeomString(OGRCircularString &, double, double, double);
```
rouault, line 122:
```suggestion
    static double GetDeterminatOfMatrixDim3(double [3], double [3], double [3]);
    static void GetCircleCenterFrom3Points(double [2], double [3], double [3]);
    static void AddCirclePointsToGeomString(OGRCircularString &, double, double, double);
```

rouault:
some linting issues remaining: https://github.com/OSGeo/gdal/actions/runs/7308961912/job/19916097431?pr=8995

cf my comment https://github.com/OSGeo/gdal/pull/8995#issuecomment-1868266258 how to solve them automatically

seidlmic:
I have installed pre-commit and hopefully fix clang formatting issue

rouault:
Things look good to me. Hopefully @landam can give it some review too

ogr/ogrsf_frmts/vfk/vfkfeature.cpp
```diff
@@ -86,6 +86,118 @@ void IVFKFeature::SetFID(GIntBig nFID)
...
+}
+
+/*!
+  \brief Find circle center determined by three point
+
+  \param c_xy circle center coordinates array
+  \param x array of three x coordinates
+  \param y array of three x coordinates
+*/
+void IVFKFeature::GetCircleCenterFrom3Points(double c_xy[2], double x[3],
```
rouault, line 114: @seidlmic could you check if the OGRGeometryFactory::GetCurveParameters() couldn't be used instead ? That would save code duplication

rouault:
OK, I'm merging this... There are likely potential optimizations like the one I suggested in https://github.com/OSGeo/gdal/pull/8995#pullrequestreview-1866404386 , but this should be good enough as it for now.

PR #9003, opened 2023-12-26
VSICurl and derived: replace most uses of CPLString with std::string

rouault:
and a few coverity fixes

### DIFF
--- a/ogr/utf8.h
+++ b/ogr/utf8.h
@@ -1422,9 +1422,11 @@ utf8_constexpr14_impl utf8_int32_t utf8lwrcodepoint(utf8_int32_t cp) {
     case 0x01ac:
       cp = 0x01ad;
       break;
+#if 0
     case 0x01af:
       cp = 0x01b0;
       break;
+#endif
     case 0x01b8:
       cp = 0x01b9;
       break;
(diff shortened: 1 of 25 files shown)


ogr/utf8.h
```diff
@@ -1422,9 +1422,11 @@ utf8_constexpr14_impl utf8_int32_t utf8lwrcodepoint(utf8_int32_t cp) {
     case 0x01ac:
       cp = 0x01ad;
       break;
+#if 0
```
elpaso, line 1425: leftover?

elpaso (approved):
I think that moving away from CPLString is a good move but can you explain me what is the long term strategy?

Is the CPL* a legacy of the past when the standard library wasn't mature (or even existing)?

Is it the goal to eventually drop CPL completely?

rouault:
> Is it the goal to eventually drop CPL completely?

no, I don't think so as it provides useful utilities. But using plain std::string when we don't need CPLString makes sense.

ogr/utf8.h
rouault: I had left that as a way to trace back the change w.r.t the upstream project, but I just submitted the removal upstream in https://github.com/sheredom/utf8.h/pull/121. I'll delete that code once upstream has integrated the change

rouault: done

PR #9017, opened 2024-01-03
GEOJSON: GetExtent3D "fast" implementation 

elpaso:
Followup https://github.com/OSGeo/gdal/pull/8806, adds a "fast" implementation for `GetExtent3D` and `GetExtent`

### DIFF
--- a/ogr/ogrsf_frmts/geojson/ogrgeojsonreader.cpp
+++ b/ogr/ogrsf_frmts/geojson/ogrgeojsonreader.cpp
@@ -3269,12 +3269,13 @@ bool OGRGeoJSONGetExtent3D(json_object *poObj, OGREnvelope3D *poEnvelope)
     // This function looks for "coordinates" and for "geometries" to handle
     // geometry collections.  It will recurse on itself to handle nested geometry.
     std::function<bool(json_object *, OGREnvelope3D *)> fParseGeometry;
-    fParseGeometry = [&fParseGeometry, &fParseCoords](
-                         json_object *poObj, OGREnvelope3D *poEnvelope) -> bool
+    fParseGeometry = [&fParseGeometry,
+                      &fParseCoords](json_object *poObjIn,
+                                     OGREnvelope3D *poEnvelopeIn) -> bool
     {
         // Get the "coordinates" array from the JSON object
         json_object *poObjCoords =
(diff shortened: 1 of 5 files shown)


ogr/ogrsf_frmts/geojson/ogrgeojsonreader.cpp
```diff
@@ -3157,3 +3189,171 @@ json_object *CPL_json_object_object_get(struct json_object *obj,
     json_object_object_get_ex(obj, key, &poRet);
     return poRet;
 }
+
+bool OGRGeoJSONGetExtent(json_object *poObj, OGREnvelope *poEnvelope)
```
rouault, line 3193: It seems to me that having just OGRGeoJSONGetExtent3D() should be sufficient, with slight modification to handle the 2D & 3D cases. For example by declaring dfX, dfY, dfZ variables, and doing the Merge() outside the loop to be independent of the number of ordinates in the array

ogr/ogrsf_frmts/geojson/ogrgeojsonreader.cpp
```diff
@@ -3157,3 +3189,171 @@ json_object *CPL_json_object_object_get(struct json_object *obj,
...
+    if (!poEnvelope || !poObj)
+    {
+        return false;
+    }
+
+    // Get the "coordinates" array from the JSON object
+    json_object *poObjCoords = OGRGeoJSONFindMemberByName(poObj, "coordinates");
+
+    // Return if not found or not an array
+    if (!poObjCoords || json_object_get_type(poObjCoords) != json_type_array)
```
rouault, line 3204: the GeometryCollection case where we have a "geometries" array should also be handled

ogr/ogrsf_frmts/geojson/ogrgeojsonreader.cpp
```diff
@@ -3157,3 +3189,171 @@ json_object *CPL_json_object_object_get(struct json_object *obj,
...
+    {
+        return false;
+    }
+
+    // poObjCoords can be an array of arrays, this lambda function will
+    // recursively parse the array
+    std::function<bool(json_object *, OGREnvelope *)> fParseCoords;
```
rouault, line 3211: ``#include <functional>`` missing.
(or potentially transform this lamba as a static function outside of this function scope)

PR #9018, opened 2024-01-03
GPKG: disable by default multi-threaded ArrowArray interface. Make it…

rouault:
… opt-in with the OGR_GPKG_NUM_THREADS config option

Cf https://lists.osgeo.org/pipermail/gdal-dev/2024-January/058177.html

### DIFF
--- a/autotest/ogr/ogr_gpkg.py
+++ b/autotest/ogr/ogr_gpkg.py
@@ -8092,6 +8092,46 @@ def test_ogr_gpkg_arrow_stream_numpy(tmp_vsimem):
 
 
 ###############################################################################
+# Test GetArrowStreamAsNumPy() and multi-threading
+
+
+def test_ogr_gpkg_arrow_stream_numpy_multi_threading(tmp_vsimem):
+    pytest.importorskip("osgeo.gdal_array")
+    pytest.importorskip("numpy")
+
+    filename = tmp_vsimem / "test.gpkg"
+
+    ds = gdal.GetDriverByName("GPKG").Create(filename, 0, 0, 0, gdal.GDT_Unknown)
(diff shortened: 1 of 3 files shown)


jratike80:
Out of interest, does the "consecutive feature ID numbering" support cases with missing ID numbers when some rows have been deleted from the middle of a table?

rouault:
> does the "consecutive feature ID numbering" support cases with missing ID numbers when some rows have been deleted from the middle of a table?

no, consecutive feature ID numbering means no holes and N features with id = 1, 2, .... N where N is the total number of features.

jratike80:
> no, consecutive feature ID numbering means no holes and N features with id = 1, 2, .... N where N is the total number of features.

Is that tested so that user cannot activate numthreads>1 and ArrowArray interface if that is doomed to fail?
With a valid GeoPackage it should be enough to test that `max(fid)=count(*)`

rouault:
> Is that tested so that user cannot activate numthreads>1 and ArrowArray interface if that is doomed to fail?

yes, the optimization checks that max(fid) = count(*) ( actually using the gpkg_ogr_contents special table to get count(*) in a fast way). So it should be safe to set OGR_GPKG_NUM_THREADS unconditionally, in theory. I'm not sure why the user on the mailing list hit that issue. I've never hit myself.

leofuhrmann:
User from the mailing list here, here's a GeoPackage to hopefully (?) reproduce the problem.

`ogr2ogr test.gpkg example.gpkg`
throws "ERROR 1: Worker thread task has not expected m_iStartShapeId value"

`ogr2ogr test.gpkg example.gpkg -where 1`
doesn't produce the error.

[example.gpkg.zip](https://github.com/OSGeo/gdal/files/13828934/example.gpkg.zip)

PR #9022, opened 2024-01-04
CMake: Handle cycles in CMake target dependencies

valgur:
## What does this PR do?
Fixes CMake configure step dying due to infinite recursion when cycles are present in the dependencies of targets provided by find_package(). This is encountered in third-party dependencies provided by Conan 2.0.

This is definitely more of a Conan issue, but I thought including it upstream in GDAL would not hurt and would add a bit of robustness to the build scripts. I opened a bug report for it in the Conan repo: https://github.com/conan-io/conan/issues/15386

This patch has been successfully tested in a recent Conan Center Index recipe for GDAL 3.5+: https://github.com/conan-io/conan-center-index/pull/19298

Non-GDAL PRs/issues:
- https://github.com/conan-io/conan/issues/15386
- https://github.com/conan-io/conan-center-index/pull/19298

## Tasklist

 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments


cmake/helpers/GdalDriverHelper.cmake
```diff
@@ -299,9 +299,9 @@ function(is_plugin _result _target)
     endif ()
 endfunction()
 
-function(gdal_target_interfaces _TARGET)
+function(gdal_target_interfaces _TARGET _seen_libs)
```
rouault, line 302: Some comment explaining the function arguments would be welcome.
And shouldn't https://github.com/OSGeo/gdal/blob/8fddbf591fa973b5875b8d29f433ce59036e9d53/cmake/helpers/GdalDriverHelper.cmake#L364 be modified to provide the _seen_libs argument ?
I actually do see it in the Conan-center-index pull request: https://github.com/conan-io/conan-center-index/pull/19298/files#diff-46ebe8f12bf2455c0cc64c21467d6d8991953aa4696bc7d425268239694d8ff5R16

cmake/helpers/GdalGenerateConfig.cmake
```diff
@@ -44,6 +44,11 @@ function(gdal_flatten_link_libraries _result)
         elseif(_lib STREQUAL "general")
             continue()
         endif()
+        if(_lib IN_LIST _seen_libs)
+            # Library has already been processed
+            continue()
+        endif()
+        list(APPEND _seen "${_lib}")
```
rouault, line 51: shouldn't that be
```suggestion
        list(APPEND _seen_libs "${_lib}")
```

rouault (commented):
Given my comments, I'm wondering to which extent this has actually been tested without the other Conan patches...  Which appears to heavily patch GDAL CMake scripts.

CC @dg0yt

cmake/helpers/GdalGenerateConfig.cmake
valgur: Oops, looks like I messed up when renaming it for the PR. Sorry about that!

cmake/helpers/GdalDriverHelper.cmake
valgur: Indeed, I accidentally moved it to a wrong patch when cleaning things up for this PR. I fixed it now both here and there.

valgur: Added a comment documenting the function as well now.

valgur:
I messed up when isolating some irrelevant changes for this PR. Sorry about that.

I updated the patch in the CCI PR and then re-applied it here to ensure that they are identical. I'll let you know once the build for CCI build has completed successfully, probably after a couple of hours.

valgur:
The CCI build passed with the updated patches: https://github.com/conan-io/conan-center-index/pull/19298#issuecomment-1877469405

dg0yt:
I will try to review this soon. I do have some concerns.

> The CCI build passed with the updated patches: [conan-io/conan-center-index#19298 (comment)](https://github.com/conan-io/conan-center-index/pull/19298#issuecomment-1877469405)

Does it include a POSIX configuration with static linkage and downstream usage via pkgconfig?

We have to be careful about the order of link libraries. And maybe some libraries really need to be repeated before breaking the cycle?

rouault:
(test failures on build-linux-ubuntu-focal, linux builds / ubuntu 20.04 gcc and  linux builds / ubuntu 20.04 coverage are unrelated to this PR, and could be solved by rebasing it on top of latest master)

cmake/helpers/GdalDriverHelper.cmake
```diff
@@ -361,7 +364,7 @@ function(gdal_target_link_libraries target)
             endif()
         endforeach()
     else ()
-        gdal_target_interfaces(${target} ${_DRIVER_PRIVATE})
+        gdal_target_interfaces(${target} "" ${_DRIVER_PRIVATE})
```
dg0yt, line 367: In my eyes this is an indication of a function signature which invites trouble.

cmake/helpers/GdalGenerateConfig.cmake
```diff
@@ -44,6 +44,11 @@ function(gdal_flatten_link_libraries _result)
         elseif(_lib STREQUAL "general")
             continue()
         endif()
+        if(_lib IN_LIST _seen_libs)
+            # Library has already been processed
+            continue()
+        endif()
+        list(APPEND _seen_libs "${_lib}")
```
dg0yt, line 51: Here we are in exporting linker options (for gdal.pc).

The PR cuts away later references to a lib which has already been seen. This must not be done! You don't want to preserve the first `-lm` but the last.

dg0yt (commented):
AFAICT this PR may fix the configuration failure for dependency cycles, but it leads to changes in building and consuming GDAL which may break users.

valgur:
> Does it include a POSIX configuration with static linkage and downstream usage via pkgconfig?

Yes and no. It builds for Linux and macOS, with both static and shared GDAL and/or dependencies.
It includes a small test build with CMake. GDAL tests are not run. pkgconfig is not tested, but note that Conan uses neither the CMake config modules or pkgconfig files installed by the project, but rather re-generates them from the package structure defined in the recipe. So the pkgconfig test would likely pass if the CMake build works, but probably not provide any useful info to you.

> We have to be careful about the order of link libraries. And maybe some libraries really need to be repeated before breaking the cycle?

That's a good point. Although I did not encounter this when building GDAL with nearly all libraries enabled, it is a possibility. Since any libraries that are dependencies need to come after their dependents, the filtering should really be done in reverse order to ensure correctness. Combined with the need to keep intentionally repeated libraries, this becomes hideously complex, especially when implementing it in CMake.

I feel like I should maybe close this and go back to the drawing board and rethink it. It adds more real pitfalls than it potentially solves. I will probably keep it in the Conan recipe only and exclude repeated Conan-internal targets only, which should be much safer. And, of course, hopefully this gets fixed in Conan itself.

Thanks a lot for the valuable feedback!

PR #9040, opened 2024-01-08
Modify the logic of selection of overviews for non-nearest resampling…

rouault:
…; add a GDAL_OVERVIEW_OVERSAMPLING_THRESHOLD config option

The current logic reads:
```
 * Some formats may efficiently implement decimation into a buffer by
 * reading from lower resolution overview images. The logic of the default
 * implementation in the base class GDALRasterBand is the following one. It
 * computes a target_downscaling_factor from the window of interest and buffer
 * size which is min(nXSize/nBufXSize, nYSize/nBufYSize).
 * It then walks through overviews and will select the first one whose
 * downscaling factor is greater than target_downscaling_factor / 1.2.
 *
 * Let's assume we have overviews at downscaling factors 2, 4 and 8.
 * The relationship between target_downscaling_factor and the select overview
 * level is the following one:
 *
 * target_downscaling_factor  | selected_overview
 * -------------------------  | -----------------
 * ]0,       2 / 1.2]         | full resolution band
 * ]2 / 1.2, 4 / 1.2]         | 2x downsampled band
 * ]4 / 1.2, 8 / 1.2]         | 4x downsampled band
 * ]8 / 1.2, infinity[        | 8x downsampled band
```

With this PR, is is ammended with the following complement:
```
 * Note that starting with GDAL 3.9, this 1.2 oversampling factor can be
 * modified by setting the GDAL_OVERVIEW_OVERSAMPLING_THRESHOLD configuration
 * option. Also note that starting with GDAL 3.9, when the resampling algorithm
 * specified in psExtraArg->eResampleAlg is different from GRIORA_NearestNeighbour,
 * this oversampling threshold defaults to 1. Consequently if there are overviews
 * of downscaling factor 2, 4 and 8, and that the desired downscaling factor is
 * 7.99, the overview of factor 4 will be selected for a non nearest resampling.
```

### DIFF
--- a/gcore/gdaldataset.cpp
+++ b/gcore/gdaldataset.cpp
@@ -2574,6 +2574,14 @@ CPLErr GDALDataset::ValidateRasterIOOrAdviseReadParameters(
  * ]4 / 1.2, 8 / 1.2]         | 4x downsampled band
  * ]8 / 1.2, infinity[        | 8x downsampled band
  *
+ * Note that starting with GDAL 3.9, this 1.2 oversampling factor can be
+ * modified by setting the GDAL_OVERVIEW_OVERSAMPLING_THRESHOLD configuration
+ * option. Also note that starting with GDAL 3.9, when the resampling algorithm
+ * specified in psExtraArg->eResampleAlg is different from GRIORA_NearestNeighbour,
+ * this oversampling threshold defaults to 1. Consequently if there are overviews
+ * of downscaling factor 2, 4 and 8, and that the desired downscaling factor is
+ * 7.99, the overview of factor 4 will be selected for a non nearest resampling.
+ *
  * For highest performance full resolution data access, read and write
  * on "block boundaries" as returned by GetBlockSize(), or use the
  * ReadBlock() and WriteBlock() methods.
--- a/gcore/gdaldataset.cpp
+++ b/gcore/gdaldataset.cpp
@@ -2579,7 +2579,7 @@ CPLErr GDALDataset::ValidateRasterIOOrAdviseReadParameters(
  * option. Also note that starting with GDAL 3.9, when the resampling algorithm
  * specified in psExtraArg->eResampleAlg is different from GRIORA_NearestNeighbour,
  * this oversampling threshold defaults to 1. Consequently if there are overviews
- * of downscaling factor 2, 4 and 8, and that the desired downscaling factor is
+ * of downscaling factor 2, 4 and 8, and the desired downscaling factor is
  * 7.99, the overview of factor 4 will be selected for a non nearest resampling.
  *
  * For highest performance full resolution data access, read and write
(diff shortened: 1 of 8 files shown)


gcore/gdaldataset.cpp
```diff
@@ -2574,6 +2574,14 @@ CPLErr GDALDataset::ValidateRasterIOOrAdviseReadParameters(
  * ]4 / 1.2, 8 / 1.2]         | 4x downsampled band
  * ]8 / 1.2, infinity[        | 8x downsampled band
  *
+ * Note that starting with GDAL 3.9, this 1.2 oversampling factor can be
+ * modified by setting the GDAL_OVERVIEW_OVERSAMPLING_THRESHOLD configuration
+ * option. Also note that starting with GDAL 3.9, when the resampling algorithm
+ * specified in psExtraArg->eResampleAlg is different from GRIORA_NearestNeighbour,
+ * this oversampling threshold defaults to 1. Consequently if there are overviews
+ * of downscaling factor 2, 4 and 8, and that the desired downscaling factor is
```
elpaso, line 2582:
```suggestion
 * of downscaling factor 2, 4 and 8, and  the desired downscaling factor is
```

gcore/gdalrasterband.cpp
```diff
@@ -153,6 +153,14 @@ GDALRasterBand::~GDALRasterBand()
  * ]4 / 1.2, 8 / 1.2]         | 4x downsampled band
  * ]8 / 1.2, infinity[        | 8x downsampled band
  *
+ * Note that starting with GDAL 3.9, this 1.2 oversampling factor can be
+ * modified by setting the GDAL_OVERVIEW_OVERSAMPLING_THRESHOLD configuration
+ * option. Also note that starting with GDAL 3.9, when the resampling algorithm
+ * specified in psExtraArg->eResampleAlg is different from GRIORA_NearestNeighbour,
+ * this oversampling threshold defaults to 1. Consequently if there are overviews
+ * of downscaling factor 2, 4 and 8, and that the desired downscaling factor is
```
elpaso, line 161:
```suggestion
 * of downscaling factor 2, 4 and 8, and the desired downscaling factor is
```

elpaso (approved):

tbonfort:
I am seeing qgis doing some back and forth between the full-resolution data and the first overview, while only doing a small pan. I believe this is due to some numerical instability when calculating the downscaling factor, and the default GDAL_OVERVIEW_OVERSAMPLING_THRESHOLD=1 used. 
To reproduce:
- open a raster in qgis
- right click on raster layer, select "zoom to native resolution"
- zoom out one level (one click on the mouse wheel)
- => when panning, qgis switches ,seemingly randomly, between the first overview level or the full resolution level.

This can be "fixed" by setting GDAL_OVERVIEW_OVERSAMPLING_THRESHOLD=1.01. I believe that the default value used for GDAL_OVERVIEW_OVERSAMPLING_THRESHOLD should be set to 1+epsilon in order to avoid this instability on commonly used zoom levels.

PR #9041, opened 2024-01-08
Add GDALGetOutputDriversForDatasetName()

rouault:
should help implementing https://github.com/georust/gdal/pull/510

### DIFF
--- a/gcore/gdaldriver.cpp
+++ b/gcore/gdaldriver.cpp
@@ -2810,3 +2810,165 @@ CPLErr GDALDriver::SetMetadataItem(const char *pszName, const char *pszValue,
     }
     return GDALMajorObject::SetMetadataItem(pszName, pszValue, pszDomain);
 }
+
+/************************************************************************/
+/*                   DoesDriverHandleExtension()                        */
+/************************************************************************/
+
+static bool DoesDriverHandleExtension(GDALDriverH hDriver, const char *pszExt)
+{
+    bool bRet = false;
+    const char *pszDriverExtensions =
+        GDALGetMetadataItem(hDriver, GDAL_DMD_EXTENSIONS, nullptr);
+    if (pszDriverExtensions)
+    {
+        const CPLStringList aosTokens(CSLTokenizeString(pszDriverExtensions));
+        const int nTokens = aosTokens.size();
+        for (int j = 0; j < nTokens; ++j)
+        {
+            if (EQUAL(pszExt, aosTokens[j]))
+            {
+                bRet = true;
+                break;
+            }
+        }
+    }
+    return bRet;
+}
+
+/************************************************************************/
+/*                  GDALGetOutputDriversForDatasetName()                */
+/************************************************************************/
+
+/** Return a list of driver short names that are likely candidates for the
+ * provided output file name.
+ *
+ * @param pszDestDataset Output dataset name (might not exist).
+ * @param nFlagRasterVector GDAL_OF_RASTER, GDAL_OF_VECTOR or
+ *                          binary-or'ed combination of both
+ * @param bSingleMatch Whether a single match is desired, that is to say the
+ *                     returned list will contain at most one item, which will
+ *                     be the first driver in the order they are registered to
+ *                     match the output dataset name. Note that in this mode, if
+ *                     nFlagRasterVector==GDAL_OF_RASTER and pszDestDataset has
+ *                     no extension, GTiff will be selected.
+ * @param bEmitWarning Whether a warning should be emitted when bSingleMatch is
+ *                     true and there are more than 2 candidates.
+ * @return NULL terminated list of driver short names.
+ * To be freed with CSLDestroy()
+ * @since 3.9
+ */
+char **GDALGetOutputDriversForDatasetName(const char *pszDestDataset,
+                                          int nFlagRasterVector,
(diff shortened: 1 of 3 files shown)


lnicola (approved):

gcore/gdaldriver.cpp
```diff
@@ -2810,3 +2810,162 @@ CPLErr GDALDriver::SetMetadataItem(const char *pszName, const char *pszValue,
...
+                    aosDriverNames.AddString(GDALGetDriverShortName(hDriver));
+                }
+            }
+        }
+    }
+
+    // GMT is registered before netCDF for opening reasons, but we want
+    // netCDF to be used by default for output.
+    if (EQUAL(osExt.c_str(), "nc") && aosDriverNames.size() == 2 &&
+        EQUAL(aosDriverNames[0], "GMT") && EQUAL(aosDriverNames[1], "NETCDF"))
```
lnicola, line 2924: Maybe

```suggestion
        EQUAL(aosDriverNames[0], "GMT") && EQUAL(aosDriverNames[1], "netCDF"))
```

I think the short name is `netCDF` and the comparison doesn't seem to be case-insensitive.

rouault: > and the comparison doesn't seem to be case-insensitive.

EQUAL() actually does case insensitive comparison, but taking your change.

Atreyagaurav: I actually had checked it before and saw that EQUAL macro does case-insensitive in linux/posix compliant ones and case-sensitive one in windows. I was thinking about commenting that but thought it should be fine :D

You can see that if you search for the macro definition. https://github.com/OSGeo/gdal/blob/15751e3b42af9be91e344e7cd4b26c45ed8074ed/frmts/dted/dted_api.h#L56


@lnicola that was also the reason I did, `to_uppercase()` before doing the comparisons (to mirror EQUAL), but you seem to have removed it. Which should still be fine. 

The case for ElasticSearch was changed in some versions, if that happens and the check if not case-insensitive they might fail, but I think it's unlikely.

gcore/gdaldriver.cpp
```diff
@@ -2810,3 +2810,162 @@ CPLErr GDALDriver::SetMetadataItem(const char *pszName, const char *pszValue,
...
+    bool bRet = false;
+    const char *pszDriverExtensions =
+        GDALGetMetadataItem(hDriver, GDAL_DMD_EXTENSIONS, nullptr);
+    if (pszDriverExtensions)
+    {
+        const CPLStringList aosTokens(CSLTokenizeString(pszDriverExtensions));
+        const int nTokens = aosTokens.size();
+        for (int j = 0; j < nTokens; ++j)
+        {
+            if (EQUAL(pszExt, aosTokens[j]))
```
elpaso, line 2829: Is this case sensitive?

gcore/gdaldriver.cpp
```diff
@@ -2810,3 +2810,162 @@ CPLErr GDALDriver::SetMetadataItem(const char *pszName, const char *pszValue,
...
+/*                  GDALGetOutputDriversForDatasetName()                */
+/************************************************************************/
+
+/** Return a list of driver short names that are likely candidates for the
+ * provided output file name.
+ *
+ * @param pszDestDataset Output dataset name (might not exist).
+ * @param nFlagRasterVector GDAL_OF_RASTER, GDAL_OF_VECTOR or
+ *                          binary-or'ed combination of both
+ * @param bSingleMatch Whether a single match is desired. In this mode, if
```
elpaso, line 2849: To be honest I don't understand the naming of this argument and it's doc string. What has the described behavior to do with "single match" ?

gcore/gdaldriver.cpp
```diff
@@ -2810,3 +2810,162 @@ CPLErr GDALDriver::SetMetadataItem(const char *pszName, const char *pszValue,
...
+ * To be freed with CSLDestroy()
+ * @since 3.9
+ */
+char **GDALGetOutputDriversForDatasetName(const char *pszDestDataset,
+                                          int nFlagRasterVector,
+                                          bool bSingleMatch, bool bEmitWarning)
+{
+    CPLStringList aosDriverNames;
+
+    std::string osExt = CPLGetExtension(pszDestDataset);
```
elpaso, line 2864: I have checked CPLGetExtension doc and it doesn't mention case changes. Should it be changed to lowercase for the check down below?

gcore/gdaldriver.cpp
rouault: No, EQUAL() performs case insensitive comparison

gcore/gdaldriver.cpp
rouault: > Should it be changed to lowercase for the check down below?

no, cf above comment about EQUAL() being case insensitive

gcore/gdaldriver.cpp
rouault: Description changed to
```
 * @param bSingleMatch Whether a single match is desired, that is to say the
 *                     returned list will contain at most one item, which will
 *                     be the first driver in the order they are registered to
 *                     match the output dataset name. Note that in this mode, if
 *                     nFlagRasterVector==GDAL_OF_RASTER and pszDestDataset has
 *                     no extension, GTiff will be selected.
 ```

elpaso (approved):

PR #9058, opened 2024-01-11
Restore use of gmtime_r and localtime_r; extend to ctime_r; use Windows variants too

rouault:


rouault:
CC @BishopGIS for the libopencad change

PR #9061, opened 2024-01-11
Warping: fix issues in long/lat <--> orthographic reprojection

### LINKED ISSUE #9056: Gdalwarp: data lost when warping from EPSG:4326 into orthographic
Tests data: World map from https://naciscdn.org/naturalearth/50m/raster/HYP_50M_SR_W.zip
Command:
`gdalwarp -t_srs +proj=ortho -et 0 -wo source_extra=5 HYP_50M_SR_W.tif ortho.tif`

Using -et and source_extra does not change the result. The result looks good at the first sight, but there are four areas like this with  missing data.

kuva

I think I have seen these mouse bites before, perhaps in some user mail but I could not find it righ now. This old ticket may be related https://github.com/OSGeo/gdal/issues/862

**EDIT**: I found it, it was in gis.stackexchange https://gis.stackexchange.com/questions/111320/reprojecting-natural-earth-raster-into-natearth-projection-with-gdalwarp-causing
and in the gdal-dev list, asked by myself https://lists.osgeo.org/pipermail/gdal-dev/2014-August/039899.html
Using `-wo SAMPLE_STEPS=1000` removed the mouse bites.

Warping back to EPSG:4326 gives more artifacts

kuva

Tested with GDAL 3.9.0dev, Proj 9.3.0 on Windows.

rouault:
fix #9056

### DIFF
--- a/alg/gdalwarpkernel.cpp
+++ b/alg/gdalwarpkernel.cpp
@@ -5069,6 +5069,185 @@ GWKCheckAndComputeSrcOffsets(GWKJobStruct *psJob, int *_pabSuccess, int _iDstX,
 }
 
 /************************************************************************/
+/*                   GWKOneSourceCornerFailsToReproject()               */
+/************************************************************************/
+
+static bool GWKOneSourceCornerFailsToReproject(GWKJobStruct *psJob)
+{
+    GDALWarpKernel *poWK = psJob->poWK;
+    for (int iY = 0; iY <= 1; ++iY)
+    {
+        for (int iX = 0; iX <= 1; ++iX)
+        {
+            double dfXTmp = poWK->nSrcXOff + iX * poWK->nSrcXSize;
+            double dfYTmp = poWK->nSrcYOff + iY * poWK->nSrcYSize;
+            double dfZTmp = 0;
+            int nSuccess = FALSE;
+            poWK->pfnTransformer(psJob->pTransformerArg, FALSE, 1, &dfXTmp,
+                                 &dfYTmp, &dfZTmp, &nSuccess);
+            if (!nSuccess)
+                return true;
+        }
+    }
+    return false;
+}
+
+/************************************************************************/
+/*                       GWKAdjustSrcOffsetOnEdge()                     */
+/************************************************************************/
+
+static bool GWKAdjustSrcOffsetOnEdge(GWKJobStruct *psJob,
+                                     GPtrDiff_t &iSrcOffset)
+{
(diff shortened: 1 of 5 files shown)


alg/gdalwarpkernel.cpp
```diff
@@ -5068,6 +5068,185 @@ GWKCheckAndComputeSrcOffsets(GWKJobStruct *psJob, int *_pabSuccess, int _iDstX,
...
+/************************************************************************/
+
+static bool GWKOneSourceCornerFailsToReproject(GWKJobStruct *psJob)
+{
+    GDALWarpKernel *poWK = psJob->poWK;
+    for (int iY = 0; iY <= 1; ++iY)
+    {
+        for (int iX = 0; iX <= 1; ++iX)
+        {
+            double dfXTmp = poWK->nSrcXOff + iX * poWK->nSrcXSize;
```
elpaso, line 5082:
```suggestion
            const double dfXTmp = poWK->nSrcXOff + iX * poWK->nSrcXSize;
```

alg/gdalwarpoperation.cpp
```diff
@@ -2889,6 +2831,115 @@ CPLErr GDALWarpOperation::ComputeSourceWindow(
...
+    }
+
+    // Use grid sampling as soon as a special point falls into the extent of
+    // the target raster.
+    if (!bUseGrid && psOptions->hDstDS)
+    {
+        for (const auto &xy : aDstXYSpecialPoints)
+        {
+            if (0 <= xy.first &&
+                GDALGetRasterXSize(psOptions->hDstDS) >= xy.first &&
```
elpaso, line 2925: I didn't check how costly is to call GDALGetRasterX(/Y)Size but maybe you can get the value outside of the loop.

elpaso (approved):
LGTM code-wise, I can't say much on the business logic but the tests look robust.

alg/gdalwarpkernel.cpp
rouault: it cannot be const, as it is updated by poWK->pfnTransformer()

alg/gdalwarpoperation.cpp
rouault: > I didn't check how costly is to call GDALGetRasterX(/Y)Size

not that much (this basically casts hDstDS to a GDALDataset*, and call a inline getter), and here the number of points in aDstXYSpecialPoints is just a handful, so that would have no practical effet

Issue #3184, opened 2020-11-16
i cannot find the definition of struct _CPLMutex


wise101:
i compile a c++ source code which linked GDAL library (the version is 3.1.2).   the envirment is windows 10, VS 2015 x64 . in release mode it is ok，but in debug mode there is a linker error :  pcidskd.lib(pcirasterblock.obj) : error LNK2001: Unresolved external symbols  "public: __cdecl CPLMutexHolder::CPLMutexHolder(void * *,double,char const *,int,int)" (??0CPLMutexHolder@@QEAA@PEAPEAXNPEBDHH@Z)
1>..\..\..\OutDir\Debug_x64\\PixIOd.dll : fatal error LNK1120: 1 external command that cannot be resolved

      i   have checked the related code：

`CPLMutexHolderD( &hRBMutex );`

and i check the definition of CPLMutexHolderD:
`#define CPLMutexHolderD(x)  CPLMutexHolder oHolder(x,1000.0,__FILE__,__LINE__);`

CPLMutexHolder is a class ,its contractor is ：
```
    explicit CPLMutexHolder( CPLMutex **phMutex, double dfWaitInSeconds = 1000.0,
                    const char *pszFile = __FILE__,
                    int nLine = __LINE__,
                    int nOptions = CPL_MUTEX_RECURSIVE);
```
       and i have finded definition of  CPLMutex:
```
#ifdef DEBUG
typedef struct _CPLMutex  CPLMutex;
typedef struct _CPLCond   CPLCond;
typedef struct _CPLJoinableThread CPLJoinableThread;
#else
#define CPLMutex void
#define CPLCond void
#define CPLJoinableThread void
#endif
```

    so you can see in release mode CPLMutex is void type ,in debug mode CPLMutex  is struct _CPLMutex.  the problem is i cannot find the definition of struct _CPLMutex.

rouault:
> there is a linker error : pcidskd.lib(pcirasterblock.obj)

Seeing pcidskd.lib, are you building against an external PCIDSK lib ? If so, that might be the issue if it was built with the non-debug flags of GDAL.
The issue is not that struct _CPLMutex is not exposed (it is intended to be an opaque type), but rather a mix of objects built in debug and release modes.

wise101:
> > there is a linker error : pcidskd.lib(pcirasterblock.obj)
> 
[...]

PR #9064, opened 2024-01-11
Make CPLMutex ABI independent of DEBUG vs non-DEBUG; set -DDEBUG or CMAKE_BUILD_TYPE=Debug for Windows builds as well

rouault:
Cf comment at https://github.com/OSGeo/gdal/issues/3184#issuecomment-1887707910

-   No longer alias CPLMutex, CPLCond and CPLJoinableThread to void in non-DEBUG builds
    
    but alias them to separate anonymous structs.
    
    ABI change
    
    Similar to 8d74e2459027f09252ad8d94855be3d34c7a69d7 and what was done
    for VSILFILE

- gdal.cmake: set -DDEBUG or CMAKE_BUILD_TYPE=Debug for Windows builds as well


gdal.cmake
```diff
@@ -199,6 +198,10 @@ else ()
 
 endif ()
 
+if (CMAKE_BUILD_TYPE MATCHES Debug)
+  add_definitions(-DDEBUG)
+endif ()
```
SunBlack, line 203: This command is superseded. I think this should this now (untested):
```suggestion
add_compile_options($<$<CONFIG:DEBUG>:-DDEBUG>)
```

rouault: Good suggestion. Using actually ``add_compile_definitions($<$<CONFIG:DEBUG>:DEBUG>)``, since it is a compile definition rather than a compile option

SunBlack: Ah true, I usually just add options to adjust the compiler warning levels, so I forgot there is a special command for definitions 😅 .

Btw: there are more occurrences of `add_definition` in this project, even this is maybe nothing for this PR.

PR #9075, opened 2024-01-13
OGRCoordinateTransformation::Transform(): change nCount parameter to size_t, and optimize performance of axis swapping

### LINKED ISSUE #9073: Improve performance of axis swapping in OGRProjCT::TransformWithErrorCodes
The following code is currently executed in OGRProjCT::TransformWithErrorCodes for axis swapping:

https://github.com/OSGeo/gdal/blob/4daa3198862100f27a9180a7400e294c8c01fcdc/ogr/ogrct.cpp#L2300-L2318

If you do not use custom axis mapping, there are actually only 2 variants of how the axis swapping can look:
https://github.com/OSGeo/gdal/blob/4daa3198862100f27a9180a7400e294c8c01fcdc/ogr/ogrspatialreference.cpp#L502-L519
So it is always `1, 2, 3` or `2, 1, 3`

Therefore the code could be optimized like this (untested):

```cpp
// Since we may swap the x and y pointers, but cannot tell the caller about this swap,
// we save the original pointers. The same axis swap code is executed for poSRSTarget.
// If this nullifies, we save the swap of both axes
const auto xOriginal = x;
const auto yOriginal = y;
const auto zOriginal = z;

// I am assuming here that (see also #9071):
// - mapping [0] => -1/1 or -2/2
// - mapping [1] => -1/1 or -2/2
// - mapping [2] => -3/3
// - abs(mapping[0]) != abs(mapping[1])
double xSign = (mapping[0] < 0) ? -1.0 : 1.0;
double ySign = (mapping[1] < 0) ? -1.0 : 1.0;
double zSign = (mapping[2] < 0) ? -1.0 : 1.0;
if (std::abs(mapping[0]) == 2)
{
    std::swap(xSign, ySign);
    std::swap(x, y);
}
if ((xSign < 0) || (ySign < 0) || (zSign < 0))
{
    for (int i = 0; i < nCount; i++) 
    { 
        x[i] *= xSign;
        y[i] *= ySign;
        z[i] *= zSign;
    } 
}
...
// after the same for poSRSTarget, check whether the axis swap is really necessary
if (x != xOriginal)
{
    for (int i = 0; i < nCount; i++) 
    { 
        std::swap(x[i], y[i]);
    } 
}
```

I stumbled across this because we have to copy our coordinates into helper vectors beforehand for GDAL and at this point the data may be copied again (with a lot of if-cases). As I am not yet sure whether I can simply replace `OAMS_TRADITIONAL_GIS_ORDER` by `OAMS_AUTHORITY_COMPLIANT` (due to the `bSwitchForGisFriendlyOrder` case) in code like this:

```cpp
OGRSpatialReference ogrSpatialReferenceSource;
ogrSpatialReferenceSource.SetAxisMappingStrategy(OAMS_TRADITIONAL_GIS_ORDER);
ogrSpatialReferenceSource.importFromEPSG(sourceEpsgCode);
OGRSpatialReference ogrSpatialReferenceTarget;
ogrSpatialReferenceTarget.SetAxisMappingStrategy(OAMS_TRADITIONAL_GIS_ORDER);
ogrSpatialReferenceTarget.importFromEPSG(targetEpsgCode);

ogrCoordinateTransformation = OGRCreateCoordinateTransformation(&ogrSpatialReferenceSource, &ogrSpatialReferenceTarget);

const size_t numberOfPoints = xyzMemory.size();
std::vector<double> xValues(numberOfPoints);
std::vector<double> yValues(numberOfPoints);
std::vector<double> zValues(numberOfPoints);

for (size_t i = 0; i < numberOfPoints; i++)
{
	const auto& xyz = xyzMemory[i];
	xValues[i] = xyz.x;
	yValues[i] = xyz.y;
	zValues[i] = xyz.z;
}

m_ogrCoordinateTransformation->TransformWithErrorCodes(numberOfPoints, xValues.data(), yValues.data(), zValues.data(), nullptr, nullptr)
```

Since I'm not sure if I can adjust the axis order without introducing any breaking change, I think it's safer to optimize this within GDAL so that it doesn't matter which order is used from a performance perspective.

### LINKED ISSUE #9074: Support `size_type` in `OGRCoordinateTransformation`
As GDAL requires C++ 11 and therefore it is using itself `std::vector` I would also expect to fully support code like this:

```cpp
std::vector<double> xValues(numberOfPoints);
std::vector<double> yValues(numberOfPoints);
std::vector<double> zValues(numberOfPoints);

ogrCoordinateTransformation->TransformWithErrorCodes(numberOfPoints,
                                                     xValues.data(),
                                                     yValues.data(),
                                                     zValues.data(),
                                                     nullptr,
                                                     nullptr);
```

But sadly the API still accepts only `int` instead of `size_type` (usually `size_t`). So you need a workaround for this, like this:

```cpp
std::vector<double> xValues(numberOfPoints);
std::vector<double> yValues(numberOfPoints);
std::vector<double> zValues(numberOfPoints);

size_t untransformedPoints = numberOfPoints;
constexpr size_t maxChunkSize = static_cast<size_t>(std::numeric_limits<int>::max());
while (untransformedPoints > 0)
{
    size_t chunkSize = std::min(untransformedPoints, maxChunkSize);
    size_t offset = numberOfPoints - untransformedPoints;
    m_ogrCoordinateTransformation->TransformWithErrorCodes(chunkSize,
                                                           xValues.data() + offset,
                                                           yValues.data() + offset,
                                                           zValues.data() + offset,
                                                           nullptr,
                                                           nullptr);
    untransformedPoints -= chunkSize;
}
```

I don't know if there are reasons for this, but it would be nice if there was an overload that also allowed larger amounts of data. In other words, instead of an `int`, an `uint64_t` overload would be good (if 32-bit systems are still supported, you have to be careful with the size of `size_t`). If an overload is not possible because of the C-API, you might have to choose a method with a slightly different name (e.g., `TransformWithErrorCodes64`)? Internally, the 32-bit methods could call the 64-bit method, as this should be safe.

rouault:
- OGRCoordinateTransformation::Transform(): change nCount parameter to size_t (C++ API only for now) (fixes #9074)
- OGRProjCT::TransformWithErrorCodes(): Improve performance of axis swapping (inspired by @SunBlack proposal) (fixes #9073)

### DIFF
--- a/ogr/ogr_spatialref.h
+++ b/ogr/ogr_spatialref.h
@@ -802,7 +802,8 @@ class CPL_DLL OGRCoordinateTransformation
      *
      * This method is the same as the C function OCTTransformEx().
      *
-     * @param nCount number of points to transform.
+     * @param nCount number of points to transform (`size_t` type since 3.9,
+     *               `int` in previous versions).
      * @param x array of nCount X vertices, modified in place. Should not be
      * NULL.
      * @param y array of nCount Y vertices, modified in place. Should not be
@@ -814,7 +815,7 @@ class CPL_DLL OGRCoordinateTransformation
      * @return TRUE if some or all points transform successfully, or FALSE if
      * if none transform.
      */
-    int Transform(int nCount, double *x, double *y, double *z = nullptr,
+    int Transform(size_t nCount, double *x, double *y, double *z = nullptr,
                   int *pabSuccess = nullptr);
 
     /**
@@ -822,7 +823,8 @@ class CPL_DLL OGRCoordinateTransformation
      *
      * This method is the same as the C function OCTTransform4D().
      *
-     * @param nCount number of points to transform.
+     * @param nCount number of points to transform (`size_t` type since 3.9,
+     *               `int` in previous versions).
      * @param x array of nCount X vertices, modified in place. Should not be
      * NULL.
      * @param y array of nCount Y vertices, modified in place. Should not be
@@ -835,7 +837,7 @@ class CPL_DLL OGRCoordinateTransformation
      * @return TRUE if some or all points transform successfully, or FALSE if
      * if none transform.
      */
-    virtual int Transform(int nCount, double *x, double *y, double *z,
+    virtual int Transform(size_t nCount, double *x, double *y, double *z,
                           double *t, int *pabSuccess) = 0;
 
     /**
@@ -843,7 +845,8 @@ class CPL_DLL OGRCoordinateTransformation
      *
      * This method is the same as the C function OCTTransform4DWithErrorCodes().
      *
-     * @param nCount number of points to transform.
+     * @param nCount number of points to transform (`size_t` type since 3.9,
+     *               `int` in previous versions).
      * @param x array of nCount X vertices, modified in place. Should not be
      * NULL.
      * @param y array of nCount Y vertices, modified in place. Should not be
@@ -857,7 +860,7 @@ class CPL_DLL OGRCoordinateTransformation
      * if none transform.
      * @since GDAL 3.3, and PROJ 8 to be able to use PROJ public error codes
      */
-    virtual int TransformWithErrorCodes(int nCount, double *x, double *y,
+    virtual int TransformWithErrorCodes(size_t nCount, double *x, double *y,
                                         double *z, double *t,
                                         int *panErrorCodes);
 
--- a/ogr/ogrct.cpp
+++ b/ogr/ogrct.cpp
@@ -2844,10 +2844,6 @@ int OGRProjCT::TransformWithErrorCodes(size_t nCount, double *x, double *y,
         const auto &mapping = poSRSTarget->GetDataAxisToSRSAxisMapping();
         if (mapping.size() >= 2)
         {
-            if (std::abs(mapping[0]) == 2 && std::abs(mapping[1]) == 1)
-            {
-                std::swap(x, y);
-            }
             const bool bNegateX = mapping[0] < 0;
             if (bNegateX)
             {
@@ -2870,6 +2866,11 @@ int OGRProjCT::TransformWithErrorCodes(size_t nCount, double *x, double *y,
                 {
                     z[i] = -z[i];
                 }
+            }
+
+            if (std::abs(mapping[0]) == 2 && std::abs(mapping[1]) == 1)
+            {
+                std::swap(x, y);
             }
         }
     }
--- a/ogr/ogrct.cpp
+++ b/ogr/ogrct.cpp
@@ -2308,25 +2308,43 @@ int OGRProjCT::TransformWithErrorCodes(size_t nCount, double *x, double *y,
     /* -------------------------------------------------------------------- */
     /*      Apply data axis to source CRS mapping.                          */
     /* -------------------------------------------------------------------- */
+
+    // Since we may swap the x and y pointers, but cannot tell the caller about this swap,
+    // we save the original pointer. The same axis swap code is executed for poSRSTarget.
+    // If this nullifies, we save the swap of both axes
+    const auto xOriginal = x;
+
     if (poSRSSource)
     {
         const auto &mapping = poSRSSource->GetDataAxisToSRSAxisMapping();
-        if (mapping.size() >= 2 && (mapping[0] != 1 || mapping[1] != 2))
+        if (mapping.size() >= 2)
         {
-            for (size_t i = 0; i < nCount; i++)
+            if (std::abs(mapping[0]) == 2 && std::abs(mapping[1]) == 1)
             {
-                double newX = (mapping[0] == 1)    ? x[i]
-                              : (mapping[0] == -1) ? -x[i]
-                              : (mapping[0] == 2)  ? y[i]
-                                                   : -y[i];
-                double newY = (mapping[1] == 2)    ? y[i]
-                              : (mapping[1] == -2) ? -y[i]
-                              : (mapping[1] == 1)  ? x[i]
-                                                   : -x[i];
-                x[i] = newX;
-                y[i] = newY;
-                if (z && mapping.size() >= 3 && mapping[2] == -3)
+                std::swap(x, y);
+            }
+            const bool bNegateX = mapping[0] < 0;
+            if (bNegateX)
+            {
+                for (size_t i = 0; i < nCount; i++)
+                {
+                    x[i] = -x[i];
+                }
+            }
+            const bool bNegateY = mapping[1] < 0;
+            if (bNegateY)
+            {
+                for (size_t i = 0; i < nCount; i++)
+                {
+                    y[i] = -y[i];
+                }
+            }
+            if (z && mapping.size() >= 3 && mapping[2] == -3)
+            {
+                for (size_t i = 0; i < nCount; i++)
+                {
                     z[i] = -z[i];
+                }
             }
         }
     }
@@ -2374,10 +2392,7 @@ int OGRProjCT::TransformWithErrorCodes(size_t nCount, double *x, double *y,
 
         if (m_eSourceFirstAxisOrient != OAO_East)
         {
-            for (size_t i = 0; i < nCount; i++)
-            {
-                std::swap(x[i], y[i]);
-            }
+            std::swap(x, y);
         }
 
         double y0 = y[0];
@@ -2458,10 +2473,7 @@ int OGRProjCT::TransformWithErrorCodes(size_t nCount, double *x, double *y,
 
         if (m_eTargetFirstAxisOrient != OAO_East)
         {
-            for (size_t i = 0; i < nCount; i++)
-            {
-                std::swap(x[i], y[i]);
-            }
+            std::swap(x, y);
         }
 
         bTransformDone = true;
@@ -2830,24 +2842,42 @@ int OGRProjCT::TransformWithErrorCodes(size_t nCount, double *x, double *y,
     if (poSRSTarget)
     {
         const auto &mapping = poSRSTarget->GetDataAxisToSRSAxisMapping();
-        if (mapping.size() >= 2 && (mapping[0] != 1 || mapping[1] != 2))
+        if (mapping.size() >= 2)
         {
-            for (size_t i = 0; i < nCount; i++)
+            if (std::abs(mapping[0]) == 2 && std::abs(mapping[1]) == 1)
             {
-                double newX = (mapping[0] == 1)    ? x[i]
-                              : (mapping[0] == -1) ? -x[i]
-                              : (mapping[0] == 2)  ? y[i]
-                                                   : -y[i];
-                double newY = (mapping[1] == 2)    ? y[i]
-                              : (mapping[1] == -2) ? -y[i]
-                              : (mapping[1] == 1)  ? x[i]
-                                                   : -x[i];
-                x[i] = newX;
-                y[i] = newY;
-                if (z && mapping.size() >= 3 && mapping[2] == -3)
+                std::swap(x, y);
+            }
+            const bool bNegateX = mapping[0] < 0;
+            if (bNegateX)
+            {
+                for (size_t i = 0; i < nCount; i++)
+                {
+                    x[i] = -x[i];
+                }
+            }
+            const bool bNegateY = mapping[1] < 0;
+            if (bNegateY)
+            {
+                for (size_t i = 0; i < nCount; i++)
+                {
+                    y[i] = -y[i];
+                }
+            }
+            if (z && mapping.size() >= 3 && mapping[2] == -3)
+            {
+                for (size_t i = 0; i < nCount; i++)
+                {
                     z[i] = -z[i];
+                }
             }
         }
+    }
+
+    // Check whether final "genuine" axis swap is really necessary
+    if (x != xOriginal)
+    {
+        std::swap_ranges(x, x + nCount, y);
     }
 
 #ifdef DEBUG_VERBOSE
--- a/ogr/ogrct.cpp
+++ b/ogr/ogrct.cpp
@@ -740,7 +740,7 @@ class OGRProjCT : public OGRCoordinateTransformation
 
     bool bWebMercatorToWGS84LongLat = false;
 
-    int nErrorCount = 0;
+    size_t nErrorCount = 0;
 
     double dfThreshold = 0.0;
 
@@ -815,10 +815,10 @@ class OGRProjCT : public OGRCoordinateTransformation
     const OGRSpatialReference *GetSourceCS() const override;
     const OGRSpatialReference *GetTargetCS() const override;
 
-    int Transform(int nCount, double *x, double *y, double *z, double *t,
+    int Transform(size_t nCount, double *x, double *y, double *z, double *t,
                   int *pabSuccess) override;
 
-    int TransformWithErrorCodes(int nCount, double *x, double *y, double *z,
+    int TransformWithErrorCodes(size_t nCount, double *x, double *y, double *z,
                                 double *t, int *panErrorCodes) override;
 
     int TransformBounds(const double xmin, const double ymin, const double xmax,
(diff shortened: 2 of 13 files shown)


ogr/ogr_spatialref.h
```diff
@@ -857,7 +860,7 @@ class CPL_DLL OGRCoordinateTransformation
      * if none transform.
      * @since GDAL 3.3, and PROJ 8 to be able to use PROJ public error codes
      */
```
SunBlack, line 862: Maybe add
```
@since GDAL 3.9 the type of @p nCount changed from `int` to `size_t`
```

ogr/ogrct.cpp
```diff
@@ -2325,7 +2356,7 @@ int OGRProjCT::TransformWithErrorCodes(int nCount, double *x, double *y,
     {
         if (m_eSourceFirstAxisOrient == OAO_East)
         {
-            for (int i = 0; i < nCount; i++)
+            for (size_t i = 0; i < nCount; i++)
             {
                 if (x[i] != HUGE_VAL && y[i] != HUGE_VAL)
```
elpaso, line 2361: Out of scope but I was wondering if in general we could replace HUGE_VAL with std::numeric_limit<>::infinity.

elpaso (approved):

ogr/ogrct.cpp
```diff
@@ -2817,34 +2842,55 @@ int OGRProjCT::TransformWithErrorCodes(int nCount, double *x, double *y,
...
                     z[i] = -z[i];
+                }
             }
         }
     }
 
+    // Check whether final "genuine" axis swap is really necessary
+    if (x != xOriginal)
+    {
+        for (size_t i = 0; i < nCount; i++)
```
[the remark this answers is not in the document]
SunBlack: `x`, `y` & `z` are raw pointer, so you don't have an iterator to use std methods. Not sure whether you can cast it somehow.

ogr/ogrct.cpp
```diff
@@ -2295,25 +2308,43 @@ int OGRProjCT::TransformWithErrorCodes(int nCount, double *x, double *y,
...
+            }
+            const bool bNegateY = mapping[1] < 0;
+            if (bNegateY)
+            {
+                for (size_t i = 0; i < nCount; i++)
+                {
+                    y[i] = -y[i];
+                }
+            }
+            if (z && mapping.size() >= 3 && mapping[2] == -3)
```
SunBlack, line 2342: Are there any tests for this? The revised method changes the behavior: If `1, 2, -3` was passed, nothing was done before this change. I think this was a bug, but it would be nice if there is a test.

ogr/ogrct.cpp
rouault: I keep HUGE_VAL, as a way of connecting with the convention used in PROJ source code.

ogr/ogrct.cpp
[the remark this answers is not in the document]
rouault: using ``std::swap_ranges(x, x + nCount, y);`` instead

ogr/ogrct.cpp
rouault: test added

SunBlack:
I now wrote a benchmark for this to see how good it performs (including some tests to compare):
- OldAxisSwap: Code before this PR
- DraftAxisSwap: The suggested code in this PR
- Draft2AxisSwap: Initial suggestion by my (multiplication instead of another loop)

```cpp
#include <benchmark/benchmark.h>

namespace
{

struct Points
{
	std::vector<double> x;
	std::vector<double> y;
	std::vector<double> z;
};

Points createPoints()
{
	static Points points;
	if (points.x.empty())
	{
		constexpr size_t pointCount = 100'000'000;
		points.x.resize(pointCount);
		points.y.resize(pointCount);
		points.z.resize(pointCount);
		for (size_t i = 0; i < pointCount; i++)
		{
			points.x[i] = i * 0.125;
			points.y[i] = -1.0 * (i % 2) * 0.25;
			points.z[i] = points.y[i] * -0.25;
		}
	}

	return points;
}

std::tuple<double*, double*, double*> oldAxisSwap(const std::vector<int>& mapping, size_t nCount, double* x, double* y, double* z)
{
	// if (mapping.size() >= 2 && (mapping[0] != 1 || mapping[1] != 2))
	if (mapping.size() >= 2 && (mapping[0] != 1 || mapping[1] != 2 || (mapping.size() >= 3 && mapping[2] != 3)))
	{
		for (int i = 0; i < nCount; i++)
		{
			double newX = (mapping[0] == 1)	   ? x[i]
						  : (mapping[0] == -1) ? -x[i]
						  : (mapping[0] == 2)  ? y[i]
											   : -y[i];
			double newY = (mapping[1] == 2)	   ? y[i]
						  : (mapping[1] == -2) ? -y[i]
						  : (mapping[1] == 1)  ? x[i]
											   : -x[i];
			x[i] = newX;
			y[i] = newY;
			if (z && mapping.size() >= 3 && mapping[2] == -3)
			{
				z[i] = -z[i];
			}
		}
	}

	return { x, y, z };
}

std::tuple<double*, double*, double*> draftAxisSwap(const std::vector<int>& mapping, size_t nCount, double* x, double* y, double* z)
{
	if (mapping.size() >= 2)
	{
		if (std::abs(mapping[0]) == 2 && std::abs(mapping[1]) == 1)
		{
			std::swap(x, y);
		}
		const bool bNegateX = mapping[0] < 0;
		if (bNegateX)
		{
			for (size_t i = 0; i < nCount; i++)
			{
				x[i] = -x[i];
			}
		}
		const bool bNegateY = mapping[1] < 0;
		if (bNegateY)
		{
			for (size_t i = 0; i < nCount; i++)
			{
				y[i] = -y[i];
			}
		}
		if (z && mapping.size() >= 3 && mapping[2] == -3)
		{
			for (size_t i = 0; i < nCount; i++)
			{
				z[i] = -z[i];
			}
		}
	}

	return { x, y, z };
}

std::tuple<double*, double*, double*> draft2AxisSwap(const std::vector<int>& mapping, size_t nCount, double* x, double* y, double* z)
{
	if (mapping.size() >= 2)
	{
		if (std::abs(mapping[0]) == 2 && std::abs(mapping[1]) == 1)
		{
			std::swap(x, y);
		}
		const double bNegateX = (mapping[0] < 0) ? -1.0 : 1.0;
		const double bNegateY = (mapping[1] < 0) ? -1.0 : 1.0;
		if (z && mapping.size() >= 3 && mapping[2] == -3)
		{
			for (size_t i = 0; i < nCount; i++)
			{
				x[i] = bNegateX * x[i];
				y[i] = bNegateY * y[i];
				z[i] = -z[i];
			}
		}
		else
		{
			if ((bNegateX < 0) || (bNegateY < 0))
			{
				for (size_t i = 0; i < nCount; i++)
				{
					x[i] = bNegateX * x[i];
					y[i] = bNegateY * y[i];
				}
			}
		}
	}

	return { x, y, z };
}

std::vector<std::vector<int>> generateAllMappingCombinations()
{
	std::vector<std::vector<int>> mapping;
	mapping.reserve(8);
	for (int x : { 1, -1 })
	{
		for (int y : { 2, -2 })
		{
			for (int z : { 3, -3 })
			{
				mapping.emplace_back(std::vector<int>{ x, y, z });
			}
		}
	}
	return mapping;
}

bool checkMethods()
{
	const Points points{
		{ 1.0, 2.0, 3.0 }, // x
		{ 4.0, 5.0, 6.0 }, // y
		{ 7.0, 8.0, 9.0 }, // z
	};
	constexpr size_t pointCount = 3;

	for (const auto& mapping : generateAllMappingCombinations())
	{
		Points oldAxisSwapPoints = points;
		auto oldAxisSwapX = oldAxisSwapPoints.x.data();
		auto oldAxisSwapY = oldAxisSwapPoints.y.data();
		auto oldAxisSwapZ = oldAxisSwapPoints.z.data();
		std::tie(oldAxisSwapX, oldAxisSwapY, oldAxisSwapZ) = oldAxisSwap(mapping, pointCount, oldAxisSwapX, oldAxisSwapY, oldAxisSwapZ);

		Points draftAxisSwapPoints = points;
		auto draftAxisSwapX = draftAxisSwapPoints.x.data();
		auto draftAxisSwapY = draftAxisSwapPoints.y.data();
		auto draftAxisSwapZ = draftAxisSwapPoints.z.data();
		std::tie(draftAxisSwapX, draftAxisSwapY, draftAxisSwapZ) = draftAxisSwap(mapping, pointCount, draftAxisSwapX, draftAxisSwapY, draftAxisSwapZ);

		Points draft2AxisSwapPoints = points;
		auto draft2AxisSwapX = draft2AxisSwapPoints.x.data();
		auto draft2AxisSwapY = draft2AxisSwapPoints.y.data();
		auto draft2AxisSwapZ = draft2AxisSwapPoints.z.data();
		std::tie(draft2AxisSwapX, draft2AxisSwapY, draft2AxisSwapZ) = draft2AxisSwap(mapping, pointCount, draft2AxisSwapX, draft2AxisSwapY, draft2AxisSwapZ);

		// now compare the data
		for (size_t i = 0; i < 3; i++)
		{
			assert(oldAxisSwapX[i] == draftAxisSwapX[i]);
			assert(oldAxisSwapY[i] == draftAxisSwapY[i]);
			assert(oldAxisSwapZ[i] == draftAxisSwapZ[i]);

			assert(oldAxisSwapX[i] == draft2AxisSwapX[i]);
			assert(oldAxisSwapY[i] == draft2AxisSwapY[i]);
			assert(oldAxisSwapZ[i] == draft2AxisSwapZ[i]);
		}
	}

	return false;
}

static auto dummy = checkMethods(); // only via assertions, so use debug mode for this!

}

static void BM_OldAxisSwap_3D(benchmark::State& state)
{
	std::vector<int> mapping = { static_cast<int>(state.range(0)),
								 static_cast<int>(state.range(1)),
								 static_cast<int>(state.range(2)) };
	auto points = createPoints();

	double* x = points.x.data();
	double* y = points.y.data();
	double* z = points.z.data();

	size_t nCount = points.x.size();
	for (auto _ : state)
	{
		auto result = oldAxisSwap(mapping, nCount, x, y, z);
		BENCHMARK_DONT_OPTIMIZE(result);
	}
}

BENCHMARK(BM_OldAxisSwap_3D)
	->ArgsProduct({ { 1, -1 }, { 2, -2 }, { 3, -3 } })
	->Unit(benchmark::kMillisecond);

static void BM_DraftAxisSwap_3D(benchmark::State& state)
{
	std::vector<int> mapping = { static_cast<int>(state.range(0)),
								 static_cast<int>(state.range(1)),
								 static_cast<int>(state.range(2)) };
	auto points = createPoints();

	double* x = points.x.data();
	double* y = points.y.data();
	double* z = points.z.data();

	size_t nCount = points.x.size();
	for (auto _ : state)
	{
		auto result = draftAxisSwap(mapping, nCount, x, y, z);
		BENCHMARK_DONT_OPTIMIZE(result);
	}
}

BENCHMARK(BM_DraftAxisSwap_3D)
	->ArgsProduct({ { 1, -1 }, { 2, -2 }, { 3, -3 } })
	->Unit(benchmark::kMillisecond);

static void BM_Draft2AxisSwap_3D(benchmark::State& state)
{
	std::vector<int> mapping = { static_cast<int>(state.range(0)),
								 static_cast<int>(state.range(1)),
								 static_cast<int>(state.range(2)) };
	auto points = createPoints();

	double* x = points.x.data();
	double* y = points.y.data();
	double* z = points.z.data();

	size_t nCount = points.x.size();
	for (auto _ : state)
	{
		auto result = draft2AxisSwap(mapping, nCount, x, y, z);
		BENCHMARK_DONT_OPTIMIZE(result);
	}
}

BENCHMARK(BM_Draft2AxisSwap_3D)
	->ArgsProduct({ { 1, -1 }, { 2, -2 }, { 3, -3 } })
	->Unit(benchmark::kMillisecond);

static void BM_OldAxisSwap_2D(benchmark::State& state)
{
	std::vector<int> mapping = { static_cast<int>(state.range(0)),
								 static_cast<int>(state.range(1)),
								 static_cast<int>(state.range(2)) };
	auto points = createPoints();

	double* x = points.x.data();
	double* y = points.y.data();

	size_t nCount = points.x.size();
	for (auto _ : state)
	{
		auto result = oldAxisSwap(mapping, nCount, x, y, nullptr);
		BENCHMARK_DONT_OPTIMIZE(result);
	}
}

BENCHMARK(BM_OldAxisSwap_2D)
	->ArgsProduct({ { 1, -1 }, { 2, -2 }, { 3, -3 } })
	->Unit(benchmark::kMillisecond);

static void BM_DraftAxisSwap_2D(benchmark::State& state)
{
	std::vector<int> mapping = { static_cast<int>(state.range(0)),
								 static_cast<int>(state.range(1)),
								 static_cast<int>(state.range(2)) };
	auto points = createPoints();

	double* x = points.x.data();
	double* y = points.y.data();

	size_t nCount = points.x.size();
	for (auto _ : state)
	{
		auto result = draftAxisSwap(mapping, nCount, x, y, nullptr);
		BENCHMARK_DONT_OPTIMIZE(result);
	}
}

BENCHMARK(BM_DraftAxisSwap_2D)
	->ArgsProduct({ { 1, -1 }, { 2, -2 }, { 3, -3 } })
	->Unit(benchmark::kMillisecond);

static void BM_Draft2AxisSwap_2D(benchmark::State& state)
{
	std::vector<int> mapping = { static_cast<int>(state.range(0)),
								 static_cast<int>(state.range(1)),
								 static_cast<int>(state.range(2)) };
	auto points = createPoints();

	double* x = points.x.data();
	double* y = points.y.data();

	size_t nCount = points.x.size();
	for (auto _ : state)
	{
		auto result = draft2AxisSwap(mapping, nCount, x, y, nullptr);
		BENCHMARK_DONT_OPTIMIZE(result);
	}
}

BENCHMARK(BM_Draft2AxisSwap_2D)
	->ArgsProduct({ { 1, -1 }, { 2, -2 }, { 3, -3 } })
	->Unit(benchmark::kMillisecond);
```


Results (CPU time in ms):

Mapping (3D) | OldAxisSwap | DraftAxisSwap | Draft2AxisSwap
-- | --: | --: | --:
1, 2, 3 | 5,625e-07 | 1,84152e-06 | 2,40653e-06
-1,   2, 3 | 93,75 | 28,125 | 55,3977
1, -2, 3 | 93,75 | 28,75 | 57,8125
-1,   -2, 3 | 135,417 | 57,2917 | 56,8182
1, 2, -3 | 117,188 | 29,375 | 89,2857
-1,   2, -3 | 114,583 | 59,6591 | 89,2857
1, -2, -3 | 114,583 | 58,2386 | 89,2857
-1,   -2, -3 | 106,771 | 86,8056 | 88,5417


Mapping (2D) | OldAxisSwap | DraftAxisSwap | Draft2AxisSwap
-- | --: | --: | --:
1, 2, 3 | 5,625e-07 | 1,84152e-06 | 2,04032e-06
-1, 2, 3 | 76,3889 | 28,75 | 59,6591
1,   -2, 3 | 80,3571 | 29,2969 | 56,8182
-1, -2, 3 | 91,5179 | 59,375 | 56,8182
1,   2, -3 | 63,9205 | 1,84152e-06 | 2,04032e-06
-1, 2, -3 | 107,639 | 28,75 | 56,8182
1,   -2, -3 | 78,125 | 28,6458 | 58,2386
-1, -2, -3 | 93,75 | 58,2386 | 58,2386

So it seems the current draft is the better version.

@rouault What do you think about deduplicating the code and using a helper method like I had done in the benchmarks? On this way you could also write tests just for this method.

SunBlack:
> in practice, not that easily, since that would be an internal method.

Ah okay - I haven't really looked into the tests here yet, but according to the new test here in the PR, they are written in Python, which explains why helper methods are harder to test. So yeah, the new test must be enough.

>And regarding performance, only looking at that part is probably a bit deceptive, as I would anticipate most of the time to be spent in PROJ. The fact that DraftAxisSwap performs slighlyt better than Draft2AxisSwap is due to the fact that negation is hard-coded, and probably more important because of CPU cache: there's no advantage of working with x, y, z arrays inside a loop as they are not interleaved. Whereas working with on single array in a loop is going to be much more CPU cache friendly. There's also the possibility the compiler would emit more easily SIMD instructions

Also true. I just like to test something like that, because another loop also means more if-queries, if it doesn't optimize that - and I'm a bit harmed by compute shaders (GPU), where you try to bypass if-queries. But you're right, the CPU ticks a little differently

But what do you think about duplicating this method? So to prevent that if someone touches the mapping algorithm (e.g. that you can swap the z-axis with the other axes after all), that an error occurs because only one place is touched instead of both.

rouault:
> But what do you think about duplicating this method?

Actually thinking more about it and with the symetry between the source and target SRS, I do believe there's a bug for the target SRS case, in the situation where there would be a x/y axis swap and a negation of values for one axis. In that situation, the negation of values must be done before the x/y axis swap, since the semantic is SetDataAxisToSRSAxisMapping, but here we start from the SRS axis mapping and need to go back to the data axis. So we can't just factor the method. I've added a commit fixing that and related tests

PR #9089, opened 2024-01-16
ogr2ogr: do not copy private layers (would for example break SQLite -> SQLite/GPKG)

rouault:


jratike80:
What are private layers? Could you give an example?

rouault:
> What are private layers? Could you give an example?

things like the "sqlite1_stats" table for a sqlite DB

PR #9108, opened 2024-01-19
Move CF-1 SRS convention import/export out of netCDF driver, and add capability to Zarr and TileDB to read SRS from CF-1

rouault:
- Add OGRSpatialReference::exportToCF1() and importFromCF1(), and make netCDF driver use them
- SWIG: add osr.SpatialReference.ImportFromCF1(), ExportToCF1(), ExportToCF1Units()
- gdalmdimtranslate: add -arrayoptions
- Zarr: be able to read CRS from CF1 conventions
- TileDB: be able to read datasets converted with 'tiledb-cf netcdf-convert'


snowman2:
Since the core CRS import/export logic is managed in PROJ, is this something that you think will eventually be migrated? Or, do imagine this only being implemented in GDAL SRS?

snowman2:
This may be useful to use as a reference for testing: [ref](https://github.com/pyproj4/pyproj/blob/3f13a33d92acd11ab96e75574df83d974415b3fd/test/crs/test_crs_cf.py)

rouault:
> Since the core CRS import/export logic is managed in PROJ, is this something that you think will eventually be migrated?

I don't know... Not sure why CF-1 SRS support would be in PROJ, but not ERMapper, OZI, Panorama, PCI, USGS, GML, etc.  Plus the fact that GDAL does need this, and if that was introduced in PROJ 9.4, GDAL would still need to have a copy of that functionality to work with PROJ < 9.4

snowman2:
> Not sure why CF-1 SRS support would be in PROJ, but not ERMapper, OZI, Panorama, PCI, USGS, GML, etc. 

PROJ is becoming the source of truth for CRS functionality in the open source geospatial community. Other applications could potentially depend on PROJ for CF management to prevent duplication of the logic. Though that is not guaranteed outside of pyproj. pyproj has quite a few users of the CF functionality. However, it already has the CF logic internally, so it wouldn't be a big deal to continue maintaining a copy of the logic in pyproj as well.

> Plus the fact that GDAL does need this, and if that was introduced in PROJ 9.4, GDAL would still need to have a copy of that functionality to work with PROJ < 9.4

Yes, that would have to happen for a while until GDAL's PROJ pin changes to >= 9.4. The sooner it is ported, the sooner it can be removed from GDAL if you think that direction makes sense.

rouault:
> The sooner it is ported, the sooner it can be removed from GDAL if you think that direction makes sense.

I have a hate&love (actually just the first part) relationship with whatever from the netCDF realm, so I'm not sure I want to be the one contaminating PROJ with that :-) That said, if someone wants to do that, I've no opposition.

snowman2:
Honestly, I wish they just used the `crs_wkt` and dropped the CF CRS grid_mapping stuff altogether.

rouault:
FYI @normanb regarding the tiledb enhancement

PR #9111, opened 2024-01-19
FindSQLite3.cmake: improve detection of static libsqlite3.a (fixes #9096)

### LINKED ISSUE #9096: check_symbol_exists not setting SQLite3_HAS_XXX by sqlite3.h in CMake
I'm using a self-compiled SQLite3 library to build GDAL. However, when using the CMake command for building, I encountered the following issue:
_-- Looking for sqlite3_open
-- Looking for sqlite3_open - not found
-- Looking for sqlite3_open
-- Looking for sqlite3_open - not found
-- Looking for sqlite3_open
-- Looking for sqlite3_open - not found
CMake Warning at cmake/helpers/CheckDependentLibraries.cmake:543 (message):
  /usr/local/third_party/lib/libsqlite3.a lacks the mutex extension! Access
  to SQLite3 databases from multiple threads will be unsafe
  gdal.cmake:265 (include)
  CMakeLists.txt:246 (include)_

I found that the issue originated from the cmake/modules/packages/FindSQLite3.cmake file, where statements like check_cxx_symbol_exists(xxx sqlite3.h SQLite3_HAS_xxx) are used to set the value of SQLite3_HAS_xxx. Upon examining the sqlite3.h file, I noticed relevant declarations for xxx. Additionally, the compilation command for SQLite3 was as follows: CFLAGS="-O3 -DNDEBUG -DSQLITE_ENABLE_COLUMN_METADATA=1 -DSQLITE_ENABLE_RTREE=1 -DSQLITE_THREADSAFE=1" ./configure xxxx. I would like to inquire if there is a solution to this issue.

## Operating system

CentOS Linux release 7.9.2009 (Core)，cmake version 3.27.1，gcc version 13.1.0 (GCC)

gdal 3.8.2 SQLite3 3.44.2

rouault:


rhuijben:
@rouault In proj we moved towards the standard cmake support for finding sqlite from cmake, while here we adding more customizations?

Can we do similar things in these related projects? That way projects depending on this don't accidentally rely on one or the others customizations.

rouault:
> Can we do similar things in these related projects?

I'm still looking to find my "CMake style". Based on my usual learning curve, this might well take a few extra years :-) And there's also this tension that upstream FindXXX modules don't necessarily work or provide the features we want.
The GDAL's FindSQLite3.cmake doesn't only detect SQLite3 but also detects optional features of it. Perhaps that should be moved to cmake/helpers/CheckDependentLibraries.cmake

PR #9116, opened 2024-01-21
Arrow/Parquet: add (minimum) support for libarrow 15.0

rouault:


ogr/ogrsf_frmts/arrow_common/ograrrowlayer.hpp
```diff
@@ -488,6 +488,12 @@ inline bool OGRArrowLayer::MapArrowTypeToOGR(
...
         case arrow::Type::RUN_END_ENCODED:
+#endif
+#if ARROW_VERSION_MAJOR >= 15
+        case arrow::Type::STRING_VIEW:
+        case arrow::Type::BINARY_VIEW:
+        case arrow::Type::LIST_VIEW:
+        case arrow::Type::LARGE_LIST_VIEW:
 #endif
         case arrow::Type::MAX_ID:
         {
```
jorisvandenbossche, line 499: Instead of adding every unsupported type explicitly, could you do a `default: ` fallback at the end? Then whenever Arrow adds a new type (that can always happen in the future, and that is also not considered a backwards incompatible change for the format), then the code will still work and by default it would ignore the column? (just as is done now explicitly for those types)

rouault: It is still good to be warned when new types are added, so we are warned that action might be needed...

But in a follow up commit in https://github.com/OSGeo/gdal/commit/35dcdb58a40305b0c6cd93f36852dbc2886dbe6e , I've modified the code so that if a new unknown type gets added, a field of that type gets ignored.

PR #9119, opened 2024-01-21
[Lint] PG: modernize SRS cache

rouault:


dg0yt:
https://github.com/OSGeo/gdal/compare/master...dg0yt:gdal:cache
for the same with a small `std::vector`.

ogr/ogrsf_frmts/pg/ogr_pg.h
```diff
@@ -38,6 +38,7 @@
 #include "ogrpgutility.h"
 #include "ogr_pgdump.h"
 
+#include <map>
```
dg0yt, line 41: Nitpick:
```suggestion
#include <map>
#include <memory>
```
for `std::unique_ptr`.

PR #9126, opened 2024-01-23
PG: be robust to inexact case for schema when creating a layer 'schema_name.layer_name' (fixes #9125)

### LINKED ISSUE #9125: With ogr2ogr -nln schema.layer, schema is case-sensitive and layer is case-insensitive

I use following command to conver a gdb file into postgresql:

```
ogr2ogr -f PostgreSQL $PG_CONNECTION "ne_50m_admin_0_countries.gdb" "ne_50m_admin_0_countries" -nln GLJX.SXJXZQHJX
```
it complains: ERROR:  schema "GLJX" does not exist

when I changed to this:

```
ogr2ogr -f PostgreSQL $PG_CONNECTION "ne_50m_admin_0_countries.gdb" "ne_50m_admin_0_countries" -nln gljx.SXJXZQHJX
```
it succed with creating a gljx.sxjxzqhjx table in postgresql.

I suppose ogr2ogr treate schema name as case-sensitive and layer name as case-insensitive. I think it would be better to be case-sensitive only if the schema name is quoted with "".

see above.

## Operating system

macos sonoma

GDAL 3.8.3 from homebrew

rouault:


ogr/ogrsf_frmts/pg/ogrpgdatasource.cpp
```diff
@@ -2355,6 +2376,62 @@ OGRLayer *OGRPGDataSource::GetLayer(int iLayer)
         return papoLayers[iLayer];
 }
 
+/************************************************************************/
+/*                             FindSchema()                             */
+/************************************************************************/
+
+// Check that the schema exist. If there is a single match in a case
+// insensitive way, use it. Otherwise error out if th match is not exact.
```
dbaston, line 2384:
```suggestion
// Check that the schema exists. If there is a single match in a case
// insensitive way, use it. Otherwise error out if the match is not exact.
```

ogr/ogrsf_frmts/pg/ogrpgdatasource.cpp
```diff
@@ -2355,6 +2376,62 @@ OGRLayer *OGRPGDataSource::GetLayer(int iLayer)
...
+// Return the schema name with its exact case from pg_catalog, or an empty
+// string if an error occurs.
+std::string OGRPGDataSource::FindSchema(const char *pszSchemaNameIn)
```
dbaston, line 2387: consider returning `std::optional<std::string>` for more clarity at the call site?

ogr/ogrsf_frmts/pg/ogrpgdatasource.cpp
```diff
@@ -2355,6 +2376,62 @@ OGRLayer *OGRPGDataSource::GetLayer(int iLayer)
...
+            {
+                osSchemaName = pszSchemaNameIn;
+                break;
+            }
+        }
+        if (osSchemaName.empty())
+        {
+            CPLError(CE_Failure, CPLE_AppDefined,
+                     "Several schemas exist whose name equals to \"%s\", but "
+                     "not with that case. Please use the exact case wished.",
```
dbaston, line 2425:
```suggestion
                     "Several schemas exist whose name matches \"%s\", but "
                     "not with that case. Please specify the schema name with the exact case.",
```

dbaston (approved):

ogr/ogrsf_frmts/pg/ogrpgdatasource.cpp
rouault: done

ogr/ogrsf_frmts/pg/ogrpgdatasource.cpp
rouault: done

ogr/ogrsf_frmts/pg/ogrpgdatasource.cpp
rouault: done

PR #9133, opened 2024-01-24
Python bindings: add a ogr.Layer.WriteArrow() method consuming __arrow_c_stream__ or __arrow_c_array__ interfaces

### LINKED ISSUE #9132: Python bindings: check for Arrow PyCapsule Interface in `ogr.Layer.WritePyArrow`

This is a corollary of https://github.com/OSGeo/gdal/pull/9043, which added support for the [Arrow PyCapsule Interface](https://arrow.apache.org/docs/format/CDataInterface/PyCapsuleInterface.html) for reading from a layer. This ticket is a feature request for _writing_ from objects that expose the PyCapsule Interface.

The current implementation of `ogr.Layer.WritePyArrow` uses pyarrow-specific APIs, including the `to_batches` method https://github.com/OSGeo/gdal/blob/5f9ffa3e00929caf104f290fee1515a4d2ac459b/swig/include/python/ogr_python.i#L652

and later the `_export_to_c` methods
https://github.com/OSGeo/gdal/blob/5f9ffa3e00929caf104f290fee1515a4d2ac459b/swig/include/python/ogr_python.i#L666
https://github.com/OSGeo/gdal/blob/5f9ffa3e00929caf104f290fee1515a4d2ac459b/swig/include/python/ogr_python.i#L642

With the PyCapsule Interface, _any_ arrow-based table or record batch would be supported. Not just pyarrow (v14 or higher) but also `geoarrow-c`, `geoarrow-rust`, and potentially more in the future, like `geopandas` (ref https://github.com/pandas-dev/pandas/pull/56587 for the pandas implementation).

Table constructs like `pyarrow.Table` include an [`__arrow_c_stream__()`](https://arrow.apache.org/docs/format/CDataInterface/PyCapsuleInterface.html#arrow_c_stream__) method and RecordBatch constructs like `pyarrow.RecordBatch` include an [`__arrow_c_array__`](https://arrow.apache.org/docs/format/CDataInterface/PyCapsuleInterface.html#arrow_c_array__) method that returns a struct column.

The exact changes to `WritePyArrow` would be:

- If keeping the dependency on pyarrow is desired:
    - The most minimal change would be checking for `__arrow_c_stream__` on the input value and calling `pyarrow.table()` on the input. If `__arrow_c_stream__` does not exist but `__arrow_c_array__` does exist, then call `pyarrow.record_batch()` on the input.
- If removing the dependency on pyarrow is desired:
    - Call `__arrow_c_stream__` to access the underlying stream pointer, which has a reference to the schema and an iterator for the batches. Then pass those pointers directly into `self.CreateFieldFromArrowSchema` and `self.WriteArrowBatch`.

Feature request, not a bug.

## Operating system

Feature request, not a bug.

Feature request, not a bug.

rouault:
fixes #9132

### DIFF
--- a/swig/include/python/ogr_python.i
+++ b/swig/include/python/ogr_python.i
@@ -652,10 +652,10 @@ def ReleaseResultSet(self, sql_lyr):
            obj:
                Object implementing the __arrow_c_stream__ or __arrow_c_array__ interface
 
-           requested_schema: PyCapsule, default None
+           requested_schema: PyCapsule, object implementing __arrow_c_schema__ or None. Default None
                The schema to which the stream should be casted, passed as a
                PyCapsule containing a C ArrowSchema representation of the
-               requested schema.
+               requested schema, or an object implementing the __arrow_c_schema__ interface.
 
            createFieldsFromSchema: boolean or None. Default to None
                Whether OGRLayer::CreateFieldFromArrowSchema() should be called. If None
@@ -672,6 +672,9 @@ def ReleaseResultSet(self, sql_lyr):
             createFieldsFromSchema = 1
         else:
             createFieldsFromSchema = 0
+
+        if requested_schema is not None and hasattr(requested_schema, "__arrow_c_schema__"):
+            requested_schema = requested_schema.__arrow_c_schema__()
 
         if hasattr(obj, "__arrow_c_stream__"):
             stream_capsule = obj.__arrow_c_stream__(requested_schema=requested_schema)
--- a/swig/include/python/ogr_python.i
+++ b/swig/include/python/ogr_python.i
@@ -643,8 +643,52 @@ def ReleaseResultSet(self, sql_lyr):
         return self.CreateFieldFromArrowSchema(schema, options)
 
 
+    def WriteArrow(self, obj, requested_schema=None, createFieldsFromSchema=None, options=[]):
+        """Write the content of the passed object, which must implement the
+           __arrow_c_stream__ or __arrow_c_array__ interface, into the layer.
+
+           Parameters
+           ----------
+           obj:
+               Object implementing the __arrow_c_stream__ or __arrow_c_array__ interface
+
+           requested_schema: PyCapsule, default None
+               The schema to which the stream should be casted, passed as a
+               PyCapsule containing a C ArrowSchema representation of the
+               requested schema.
+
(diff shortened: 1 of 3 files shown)


swig/include/python/ogr_python.i
```diff
@@ -643,8 +643,52 @@ def ReleaseResultSet(self, sql_lyr):
...
+
+           Parameters
+           ----------
+           obj:
+               Object implementing the __arrow_c_stream__ or __arrow_c_array__ interface
+
+           requested_schema: PyCapsule, default None
+               The schema to which the stream should be casted, passed as a
+               PyCapsule containing a C ArrowSchema representation of the
+               requested schema.
```
jorisvandenbossche, line 658: I don't think there is already much experience with this `requested_schema` keyword, but you could also accept any "schema-like" object that implements `__arrow_c_schema__` (and call that method to get the capsule before passing it to `__arrow_c_array/stream__`). 
That makes it similar to the main `obj` keyword, where the user can also pass any object implementing the protocol, instead of passing a capsule itself. 

cc @kylebarron

rouault: The implementation just passes requested_schema to ``obj.__arrow_c_stream__(requested_schema=requested_schema)`` . So if ``__arrow_c_stream__()`` also accepts an object that implements ``__arrow_c_schema__`` that will work. I just copied the official documentation from https://arrow.apache.org/docs/format/CDataInterface/PyCapsuleInterface.html#arrowstream-export

jorisvandenbossche: Yes, the dunder methods expect a capsule for the `requested_schema`, that's indeed how it is suppose to work (and is documented). 

But here the user is not calling that directly, they are using a public user-facing ``WriteArrow`` method, and so there you can do whatever you want. And what I did for other projects was to accept any schema-like object.

What I meant is essentially just adding this:

```
if hasattr(requested_schema, "__arrow_c_schema__"):
    requested_schema = requested_schema.__arrow_c_schema__()
```

before passing `requested_schema` to `obj.__arrow_c_stream__(requested_schema=requested_schema)`

rouault: ok, got it. Commit implementing that added

jorisvandenbossche: Looks good!

PR #9141, opened 2024-01-25
GPKG: fixes to make most operations compatible with PRAGMA foreign_keys=1 

### LINKED ISSUE #9135: "FOREIGN KEY constraint failed" error with the geopackage driver

I have a custom setup that use a libsqlite3 version that is build with `-DSQLITE_DEFAULT_FOREIGN_KEYS=1` (which enables foreign key constraints by default. If I now try to create a new geopackage geometry and insert data in that geometry this process fails with a "FOREIGN KEY constraint failed" error. 

It's possible to workaround this issue for libsqlite3 versions build with the mentioned feature flag by just setting `OGR_SQLITE_PRAGMA=FOREIGN_KEYS=0` (i.e disabling the foreign key check). The produced sqlite database contains only valid foreign key relations afterwards, so it just seems to be an issue in the order of which certain SQL statements are executed. I would expect that gdal uses an order that always results in valid foreign key relations instead of the other way around. If someone points to the relevant code I might consider working on a fix.

This problem can be reproduced by manually requiring sqlite to check the foreign key constraints by setting the `OGR_SQLITE_PRAGMA=FOREIGN_KEYS=1` config flag while running the driver independently from your sqlite compile flags. 
For example it can be reproduced by using `ogr2ogr`:

```
OGR_SQLITE_PRAGMA=FOREIGN_KEYS=1 ogr2ogr /tmp/test.gpkg some_file.shp            
ERROR 1: sqlite3_exec(INSERT INTO gpkg_geometry_columns (table_name,column_name,geometry_type_name,srs_id,z,m) VALUES ('some_file','geom','POINT',4326,0,0)) failed: FOREIGN KEY constraint failed
ERROR 1: Unable to write feature 0 from layer Ereignisabfrage 2022-10-11.
translation of layer some_file (use -skipfailures to skip errors)
```

## Operating system

Fedora 39

Observed with:
* GDAL 3.7.3, released 2023/10/30
* GDAL 3.8.3 (build from 654f4907abbbf6bf4226d58a8c067d134eaf3ce9)

rouault:
(fixes #9135)

### DIFF
--- a/ogr/ogrsf_frmts/gpkg/ogr_geopackage.h
+++ b/ogr/ogrsf_frmts/gpkg/ogr_geopackage.h
@@ -156,6 +156,9 @@ class GDALGeoPackageDataset final : public OGRSQLiteBaseDataSource,
     mutable int m_nHasMetadataTables = -1;  // -1 = unknown, 0 = false, 1 = true
     int m_nCreateMetadataTables = -1;  // -1 = on demand, 0 = false, 1 = true
 
+    // Set by CreateTileGriddedTable() and used by FinalizeRasterRegistration()
+    std::string m_osSQLInsertIntoGpkg2dGriddedCoverageAncillary{};
+
     CPLString m_osIdentifier{};
     bool m_bIdentifierAsCO = false;
     CPLString m_osDescription{};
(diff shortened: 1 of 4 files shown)


ogr/ogrsf_frmts/gpkg/ogr_geopackage.h
```diff
@@ -411,6 +414,44 @@ class GDALGeoPackageDataset final : public OGRSQLiteBaseDataSource,
...
+                    SQLCommand(m_poDS->GetDB(), "PRAGMA foreign_keys = 0"));
+            }
+        }
+
+        ~TemporaryForeignKeyCheckDisabler()
+        {
+            if (m_nPragmaForeignKeysOldValue)
+            {
+                CPL_IGNORE_RET_VAL(
+                    SQLCommand(m_poDS->GetDB(), "PRAGMA foreign_keys = 1"));
```
weiznich, line 438: I think there is no need for setting that value back as that is only a connection wide setting. This connection won't be reused outside the geopackage writer, right?

jratike80: Maybe it is not expensive to turn it on, and wouldn't it be better to play safe?

rouault: > This connection won't be reused outside the geopackage writer, right?

Users of the API could potentially run the GDALDataset::ExecuteSQL("PRAGMA foreign_keys = 1") , so better restore to the setting they used

elpaso (approved):

PR #9144, opened 2024-01-26
gdal_edit: add -a_coord_epoch option and update doc

dshean:
## What does this PR do?

Add new options to `gdal_edit.py` that will add, update or remove a user-defined coordinate epoch from the dataset SRS.  Updated documentation.  

https://lists.osgeo.org/pipermail/gdal-dev/2024-January/058314.html

## Tasklist

 - [x] Add test case(s)
 - [x] Add documentation
 - [ ] Review
 - [x] Adjust for comments

## Environment

* OS:
* Compiler:

### DIFF
--- a/doc/source/programs/gdal_edit.rst
+++ b/doc/source/programs/gdal_edit.rst
@@ -15,9 +15,10 @@ Synopsis
 
 .. code-block::
 
-    gdal_edit [--help] [--help-general] [-ro] [-a_srs <srs_def>]
+    gdal_edit [--help] [--help-general] [-ro] [-a_srs <srs_def>] 
             [-a_ullr <ulx> <uly> <lrx> <lry>] [-a_ulurll <ulx> <uly> <urx> <ury> <llx> <lly>]
             [-tr <xres> <yres>] [-unsetgt] [-unsetrpc] [-a_nodata <value>] [-unsetnodata]
+            [-a_coord_epoch <epoch>] [-unsetepoch]
             [-unsetstats] [-stats] [-approx_stats]
             [-setstats <min> <max> <mean> <stddev>]
             [-scale <value>] [-offset <value>] [-units <value>]
@@ -57,6 +58,18 @@ It works only with raster formats that support update access to existing dataset
     If the empty string or None is specified, then the existing
     coordinate system will be removed (for TIFF/GeoTIFF, might not be well
     supported besides that).
+
+.. option:: -a_coord_epoch <epoch>
+
+    Assign/override the coordinate epoch of the dataset, as a decimal year (e.g., 2021.3).
+
+    .. versionadded:: 3.9
+
+.. option:: -unsetepoch
+
+    Remove the coordinate epoch of the dataset.
+
+    .. versionadded:: 3.9
 
 .. option:: -a_ullr  <ulx> <uly> <lrx> <lry>
 
--- a/autotest/pyscripts/test_gdal_edit.py
+++ b/autotest/pyscripts/test_gdal_edit.py
@@ -444,3 +444,26 @@ def test_gdal_edit_py_unsetrpc(script_path):
         ds = None
     finally:
         gdal.GetDriverByName("GTiff").Delete(filename)
+
+
+###############################################################################
+# Test -a_coord_epoch
+
+
+def test_gdal_edit_py_epoch(script_path):
+
+    filename = "tmp/test_gdal_edit_py.tif"
+    shutil.copy(test_py_scripts.get_data_path("gcore") + "byte.tif", filename)
+
+    try:
+        test_py_scripts.run_py_script(
+            script_path, "gdal_edit", filename + " -a_coord_epoch 2021.3"
+        )
+
+        ds = gdal.Open(filename)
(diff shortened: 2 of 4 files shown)


dshean:
I don't think there are any tests for `gdal_edit.py`.  I tested several sequences and combinations of `-a_srs`, `-a_coord_epoch` and `-unsetepoch` for a sample GeoTiff.

I added the `-unsetepoch` option to maintain consistency with other existing options, like `-unsetnodata`.  It would be simpler to use `-a_coord_epoch 0.0` to unset, as that is what is returned by GetCoordinateEpoch() when the SRS has no epoch defined.  I expect this will evolve as support for epoch-aware CRS transformations evolves.

No changes to Python API (or associated doc).

rouault:
Thanks! Tests for gdal_edit are in autotest/pyscripts/test_gdal_edit.py

doc/source/programs/gdal_edit.rst
```diff
@@ -58,6 +59,18 @@ It works only with raster formats that support update access to existing dataset
     coordinate system will be removed (for TIFF/GeoTIFF, might not be well
     supported besides that).
 
+.. option:: -a_coord_epoch <epoch>
+
+    Assign/override the coordinate epoch of the dataset, as a decimal year (e.g., 2021.3).
+
+    .. versionadded:: 3.4
```
rouault, line 66:
```suggestion
    .. versionadded:: 3.9
```

doc/source/programs/gdal_edit.rst
```diff
@@ -58,6 +59,18 @@ It works only with raster formats that support update access to existing dataset
...
+
+.. option:: -unsetepoch
+
+    Remove the coordinate epoch of the dataset.
+
+    .. versionadded:: 3.4
```
rouault, line 72:
```suggestion
    .. versionadded:: 3.9
```

dshean:
OK, I added a simple test and updated the version number to 3.9 in the doc.

I did a rebase instead of merge before pushing changes.  Let me know if I should clean up, or if you can isolate the relevant commit [0d5fa29](https://github.com/OSGeo/gdal/pull/9144/commits/0d5fa293fdc2c3a079c9066fba7f864496fd724c)

rouault:
> I did a rebase instead of merge before pushing changes.

actually, you did the reverse ;-) I'd prefer indeed that it is a rebase with a single commit

dshean:
Sorry about that - I'm a little rusty. I reset my local branch, squashed the two relevant commits, and did a force push to my remote branch, which I think takes care of the issue.  Should now be a single commit, I think ready to merge.  Thanks!

rouault:
> Sorry about that - I'm a little rusty. I reset my local branch, squashed the two relevant commits, and did a force push to my remote branch, which I think takes care of the issue. Should now be a single commit, I think ready to merge. Thanks!

no problem. that looks good. I'm letting now CI run and see if things are in order

dshean:
One other thought - I actually like the FlatGeoBuf strategy to include `EPOCH[2016.47]` in WKT2 (https://gdal.org/user/coordinate_epoch.html#flatgeobuf).  Easy to include this when defining custom WKT2 CRS for a particular dataset, and not limited to GeoTiff format.  Any plans to support that approach in PROJ/GDAL?

Also, it might be useful to have a centralized function to convert a datetime/Timestamp to decimal year for the epoch specification.  I have my own implementation of this, and I recall seeing something in the GDAL code, but I don't think it's exposed in the Python API.

rouault:
> Any plans to support that approach in PROJ/GDAL?

well PROJ  (9.2 maybe, not sure exactly which version, but "recently") does support the COORDINATEMETADATA[] WKT support, and when this is ingested by GDAL, it feeds the EPOCH[] to the relevant member of OGRSpatialReference

> Also, it might be useful to have a centralized function to convert a datetime/Timestamp to decimal year for the epoch specification. 

I'm not sure about that. I've always seen the epoch being specified as a decimal year. I imagine that someone does the conversion, but it seems this is something generally done "before" PROJ or GDAL, mostly be geodesists

dshean:
> I've always seen the epoch being specified as a decimal year. I imagine that someone does the conversion, but it seems this is something generally done "before" PROJ or GDAL, mostly be geodesists

Yes, epoch should always be specified as decimal year, before updating metadata in PROJ/GDAL.  

I'm just thinking about best place for a convenience function to convert the more common (YYYY, MM, DD) date information to a decimal year, since that's typically what is available from filename or metadata for a dataset.  

Most people who think about these issues have an existing capability to do the conversion to decimal year, usually with their own implementation (something like this: https://stackoverflow.com/a/6451892).  

But it presents an additional hurdle for newer users who want to perform epoch-aware CRS transformations, but only have the more common YYYYMMDD timestamp of their dataset from metadata, and don't have an existing utility to convert.

Not suggesting any change is necessary - users can figure this out on their own.  Alternative might be to add an option for a utility like gdal_edit to accept a timestamp (with known formatting) and then do the conversion to decimal year before setting the epoch.  Or maybe the conversion to decimal year could be easily added to Python datetime module or pandas.Timestamp.

PR #9166, opened 2024-01-30
Doc: OpenFileGDB: document OPENFILEGDB_DEFAULT_STRING_WIDTH configuration option (fixes #9165)

### LINKED ISSUE #9165: Feature: FGDB_STRING_WIDTH configuration option in OpenFileGDB driver
According to https://github.com/OSGeo/gdal/blob/master/NEWS.md  - OGR 1.11.0 enabled FGDB_STRING_WIDTH configuration option for the FileGDB driver though it isn't noted here https://gdal.org/drivers/vector/filegdb.html#vector-filegdb.(are the drivers different between OGR and GDAL?) 

I'm wondering if the FGDB_STRING_WIDTH configuration option could be added to the OpenFileGDB driver in GDAL?  My use case is writing to gdb from R

rouault:


ogr/ogrsf_frmts/openfilegdb/ogropenfilegdblayer.cpp
```diff
@@ -664,7 +664,7 @@ int OGROpenFileGDBLayer::BuildLayerDefinition()
         // string width is 0, we pick up 65536 by default to mean unlimited
```
elpaso, line 664: this comment needs updates, maybe
```suggestion
        // string width is 0, if OPENFILEGDB_DEFAULT_STRING_WIDTH is not set we pick up 65536 by default to mean unlimited
```

elpaso:
I found it confusing that the C++ constant OPENFILEGDB_DEFAULT_STRING_WIDTH and the open option have the same name: if I read this right the first is the hard limit (and the default) and the second is a user-defined default which still needs to be less than the hard limit.

rouault:
> I find it confusing that the C++ constant OPENFILEGDB_DEFAULT_STRING_WIDTH

yes, I agree that's confusing. Renamed in an extra commit where I've also added extensive comments to explain the trick

elpaso (approved):

rouault:
(note: CIFuzz failure is unrelated, and will be addressed per https://github.com/OSGeo/gdal/pull/9181)

PR #9167, opened 2024-01-30
CI: try macOS-14 for M1 MacOS runner + bump minimum Java version to 8

rouault:


rouault:
@elpaso

We get
```
 [ RUN      ] test_ogr.GDALDatasetSetQueryLoggerFunc
Segmentation fault
```
on the macOS-14 M1 runner (https://github.com/OSGeo/gdal/actions/runs/7715287729/job/21029498239?pr=9167)
Any idea of the cause?

rouault:
> Any idea of the cause?

ok, so the reason was the this initial GDAL build linked to the system Mac SQLite3, which is a heavily patched/restricted version. There must be some fragility in our code that would be nice to fix, but hard to investigate without access to that build. I've workaround by linking to Homebrew's SQLite3 build, like was done for x64 Mac

PR #9168, opened 2024-01-31
BMP: fix reading images larger than 4GB

rouault:
This fixes a regression introduced in GDAL 3.4.0 per https://github.com/OSGeo/gdal/pull/4720


elpaso (approved):

jratike80:
I thought that BMP cannot be larger that 4 GB https://www.artwork.com/gerber/vlbv/faq/bmp_header.htm but obviously it can. Is the specification changed at some time?

rouault:
> but obviously it can

I don't know if the spec allows it but GDAL has never limited it. But I would assume most BMP viewer will struggle with such large files. GDAL writes a dummy 0xFFFFFFF value in the header field where the file size is written, and on reading it ignores it.  Cf c2b55fe89510a54a6161ed8807a20a2492bd36dd for a related past changes (but writing and reading files larger than 4 GB was older than this one)

PR #9170, opened 2024-01-31
Identification tunings in TileDB and Shapefile drivers

rouault:
Relates to https://github.com/conda-forge/tiledb-feedstock/issues/228

### DIFF
--- a/frmts/tiledb/tiledbcommon.cpp
+++ b/frmts/tiledb/tiledbcommon.cpp
@@ -133,9 +133,17 @@ int TileDBDataset::Identify(GDALOpenInfo *poOpenInfo)
             return TRUE;
         }
 
+        const bool bIsS3OrGS =
+            STARTS_WITH_CI(poOpenInfo->pszFilename, "/VSIS3/") ||
+            STARTS_WITH_CI(poOpenInfo->pszFilename, "/VSIGS/");
+        // If this is a /vsi virtual file systems, bail out, except if it is S3 or GS.
+        if (!bIsS3OrGS && STARTS_WITH(poOpenInfo->pszFilename, "/vsi"))
+        {
+            return false;
+        }
+
         if (poOpenInfo->bIsDirectory ||
-            ((STARTS_WITH_CI(poOpenInfo->pszFilename, "/VSIS3/") ||
-              STARTS_WITH_CI(poOpenInfo->pszFilename, "/VSIGS/")) &&
+            (bIsS3OrGS &&
              !EQUAL(CPLGetExtension(poOpenInfo->pszFilename), "tif")))
         {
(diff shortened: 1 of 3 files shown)


frmts/tiledb/tiledbcommon.cpp
```diff
@@ -133,9 +133,17 @@ int TileDBDataset::Identify(GDALOpenInfo *poOpenInfo)
             return TRUE;
         }
 
+        const bool bIsS3OrGS =
+            STARTS_WITH_CI(poOpenInfo->pszFilename, "/VSIS3/") ||
+            STARTS_WITH_CI(poOpenInfo->pszFilename, "/VSIGS/");
+        // If this is a /vsi virtual file systems, bail out, except if it is S3 or GS.
+        if (!bIsS3OrGS && STARTS_WITH(poOpenInfo->pszFilename, "/vsi"))
```
elpaso, line 140: is there a reason for not checking case insensitive here too? (assuming that STARTS_WITH and STARTS_WITH_CI are what I think they are regarding case sensitivity).

frmts/tiledb/tiledbdrivercore.cpp
```diff
@@ -51,10 +51,20 @@ static int TileDBDriverIdentifySimplified(GDALOpenInfo *poOpenInfo)
...
     }
 
-    if (poOpenInfo->bIsDirectory ||
-        ((STARTS_WITH_CI(poOpenInfo->pszFilename, "/VSIS3/") ||
-          STARTS_WITH_CI(poOpenInfo->pszFilename, "/VSIGS/")) &&
-         !EQUAL(CPLGetExtension(poOpenInfo->pszFilename), "tif")))
+    const bool bIsS3OrGS = STARTS_WITH_CI(poOpenInfo->pszFilename, "/VSIS3/") ||
+                           STARTS_WITH_CI(poOpenInfo->pszFilename, "/VSIGS/");
+    // If this is a /vsi virtual file systems, bail out, except if it is S3 or GS.
+    if (!bIsS3OrGS && STARTS_WITH(poOpenInfo->pszFilename, "/vsi"))
```
elpaso, line 57: ditto

frmts/tiledb/tiledbcommon.cpp
rouault: Well, the checking for VSIS3 and VSIGS should have been case sensitive from the beginning, but for some reason the original contributor did it case insensitive, so I've kept that check as it in case it was depended upon

rouault:
@normanb Is there any strong reason for the TileDB driver to support ``/VSIS3/foo`` spelled like that ? Normally GDAL only recognize the name of the virtual file system in lower case : ``/vsis3/foo``

PR #9182, opened 2024-02-01
CSV: parse header with line breaks (fixes #9172)

### LINKED ISSUE #9172: CSV files with line breaks in the header - behaviour inconsistent with other CSV readers
Suppose we have a CSV as follows:

```
Column one,Column two,"Column with a
line break",Column three,"Another
line break",Column four,Column five
1,2,3,4,5,6,7
4,5,6,7,8,9,10
7,8,9,10,11,12,13
```

Excel, Google Sheets and QGIS' "Delimited Text" function (with "First record has field names" checked) all successfully read this file:

image

However, GDAL only reads the first three fields:

```
Layer name: linebreaks
Geometry: None
Feature Count: 3
Layer SRS WKT:
(unknown)
Column one: String (0.0)
Column two: String (0.0)
Column with a: String (0.0)
```
version: `GDAL 3.9.0dev-5d6bbd00de, released 2024/01/29`

(Note, when you use Excel to create a file with line breaks in cells, those line breaks are saved as LF only, even though the line breaks between records are CRLF. However, no CSV reader I tested behaved differently when I interchanged LF and CRLF in various permutations.)

rouault:


rouault:
> Does the code search for the longest line in the CSV for finding out how many columns there are, and then find as many column names from the beginning of the file?

The implementation is at https://github.com/OSGeo/gdal/blob/85c93fd6db4f021bbccea8cb2bee209b87e8a2e9/port/cpl_csv.cpp#L637 . Basically it will read extra lines if it detects that there is an odd number of double-quotes in a line
This function is what is used already to read non-header lines where such situation can happen. There is no reason not to use if for header lines too (the CSV format itself doesn't really distinguish header from non-header, which is just a convention often used, but not always)

PR #9191, opened 2024-02-05
[WIP] GDAL_GCP C++'ification: appropriate for 3.9 ?

rouault:
Up to now, GDAL_GCP was only a C-only structure. The typical/recommended pattern to create GCPs from C is:
```
GDAL_GCP* gcps = (GDAL_GCP*)CPLMalloc( count * sizeof(GDAL_GCP) );
GDALInitGCPs( count, gcps );  // zero-initialize  pixel,line,x,y,z double members of each gcps[], and affects a CPLStrdup("") string to pszId and pszInfo members of each gcps[]
CPLFree(gcps[0].pszId); // free the initial empty string
gcps[0].pszId = CPLStrdup("foo"); // for example
GDALDeinitGCPs( count, gcps ); // free memory of gcps[i].pszId and gcps[i].pszInfo
CPLFree(gcps); // free the arra itself
```
This is "fine" for a C user, but quite tedious in a C++ context. Big users of creating GCPs are GDAL drivers themselves.

So this pull request makes GDAL_GCP is a C++ struct, when gdal.h included from C++ code, with a default allocator, a destructor and move constructor / assignment operators.

This mostly works except a few details:

- If a GDAL_GCP is allocated as a C++ object ( ie something like ``GDAL_GCP gcp``, or ``GDAL_GCP gcps[5]`` or with ``std::vector<GDAL_GCP>``, or anything that is not ``CPLMalloc()``/``VSIMalloc()``), and then passed to ``GDALInitGCPs()``, a memory leak would occur because ``pszId`` will be initialized twice. Here the fix consists in the GDAL_GCP constructor to assign it to the return of a  ``VSIGetStaticEmptyString()``  function, which returns a special static instance of an empty string, that ``VSIFree()`` explicitly recognizes and does not free. Hence no leak occurs.

- For a GDAL_GCP created as a C++ object, one could run into issues with double-free, when explicitly calling ``GDALDeinitGCPs()``, and then when the GDAL_GCP destructor automatically runs. This was easily fixed by making ``GDALDeinitGCPs()`` to nullify ``pszId`` and ``pszInfo`` members after CPLFree()'ing them.

All above are just implementation details. More annoyingly, here are the few backward incompatibilities I found with a few instances of C++ code:

- In gdalpamdataset.cpp, we kind of abused the recommended workflow mentionned at the beginning, by assigning a static empty string to pszId and pszInfo on a GDAL_GCP[] object, and not use GDALInitGCPs() / GDALDeinitGCPs(). The issue now is that when ``~GDAL_GCP()`` is called it will try to ``CPLFree()`` that.  No other solution here than removing that initialization, that is no longer needed

- In ``GDALLoadTabFile()`` and ``GDALLoadOziMapFile()``, a GDAL_GCP[] array was allocated, and then memcpy()'d onto a dynamically array allocated with ``CPLMalloc( count * sizeof(GDAL_GCP) )`` returned to the user. This code has to be changed to remove the ``memcpy()`` part and do ``cpl_malloc_gcps[i] = std::move(asGCPs[i])``

- The Sentinel2 driver used ``memset(gcps, 0, sizeof(gcps))`` to initialize a GDAL_GCP[] array. Some compilers complained about memset()'ing a non plain-old-object instance. The fix is just to remove that now useless memset().

Except those details, all other drivers compile and pass their regression tests, which makes me believe, that while being backward incompatible, this is still reasonable. The compiler should complain in most problematic situations where the code has to be changed (this was for example for the memcpy() in GDALLoadTabile/GDALLoadOziMapFile). The only undetected change at compile time was the "clever trick" of gdalpamdataset.cpp which was only detected at runtime. Hopefully such a pattern should be exceptional.
In the worse case, users can define the new ``GDAL_SUPPRESS_GCP_CPLUSPLUS`` macro before including gdal.h, which will make their code see the POD C structure.

And to re-iterate, this has no impact on C-only users

The first commit introduces the minimum changes to get things working, and the second one modernizes the HFA driver to use a ``std::vector<GDAL_GCP>`` to show the benefits of the proposed change (less magic numbers, automatic construction/destruction) . More drivers could be modernized in a similar way

### DIFF
--- a/port/cpl_vsisimple.cpp
+++ b/port/cpl_vsisimple.cpp
@@ -835,6 +835,21 @@ void *VSIRealloc(void *pData, size_t nNewSize)
 }
 
 /************************************************************************/
+/*                       VSIGetStaticEmptyString()                      */
+/************************************************************************/
+
+static char *VSI_STATIC_EMPTY_STRING = const_cast<char *>("");
+
+/** Return an empty string ("") that VSIFree() does not free.
+ *
+ * That string should not be modified.
+ */
+char *VSIGetStaticEmptyString()
+{
+    return VSI_STATIC_EMPTY_STRING;
+}
+
+/************************************************************************/
 /*                              VSIFree()                               */
 /************************************************************************/
 
@@ -843,10 +858,10 @@ void *VSIRealloc(void *pData, size_t nNewSize)
 void VSIFree(void *pData)
 
 {
-#ifdef DEBUG_VSIMALLOC
-    if (pData == nullptr)
+    if (pData == nullptr || pData == VSI_STATIC_EMPTY_STRING)
         return;
 
+#ifdef DEBUG_VSIMALLOC
     char *ptr = ((char *)pData) - 2 * sizeof(void *);
     VSICheckMarkerBegin(ptr);
     size_t nOldSize = 0;
@@ -885,8 +900,7 @@ void VSIFree(void *pData)
 #endif
 
 #else
-    if (pData != nullptr)
-        free(pData);
+    free(pData);
 #endif
 }
 
--- a/port/cpl_vsisimple.cpp
+++ b/port/cpl_vsisimple.cpp
@@ -838,7 +838,7 @@ void *VSIRealloc(void *pData, size_t nNewSize)
 /*                       VSIGetStaticEmptyString()                      */
 /************************************************************************/
 
-static char *VSI_STATIC_EMPTY_STRING = const_cast<char *>("");
+static char VSI_STATIC_EMPTY_STRING = 0;
 
 /** Return an empty string ("") that VSIFree() does not free.
  *
@@ -846,7 +846,7 @@ static char *VSI_STATIC_EMPTY_STRING = const_cast<char *>("");
  */
 char *VSIGetStaticEmptyString()
 {
-    return VSI_STATIC_EMPTY_STRING;
+    return &VSI_STATIC_EMPTY_STRING;
 }
 
 /************************************************************************/
@@ -858,7 +858,7 @@ char *VSIGetStaticEmptyString()
 void VSIFree(void *pData)
 
 {
-    if (pData == nullptr || pData == VSI_STATIC_EMPTY_STRING)
+    if (pData == nullptr || pData == VSIGetStaticEmptyString())
         return;
 
 #ifdef DEBUG_VSIMALLOC
--- a/autotest/cpp/test_gdal.cpp
+++ b/autotest/cpp/test_gdal.cpp
@@ -3597,4 +3597,72 @@ TEST_F(test_gdal, drop_cache)
     }
 }
 
+TEST_F(test_gdal, GDAL_GCP)
+{
+    // Test default constructor
+    {
+        GDAL_GCP x;
+        EXPECT_STREQ(x.pszId, "");
+        EXPECT_STREQ(x.pszInfo, "");
+        EXPECT_EQ(x.dfGCPPixel, 0.0);
+        EXPECT_EQ(x.dfGCPLine, 0.0);
+        EXPECT_EQ(x.dfGCPX, 0.0);
+        EXPECT_EQ(x.dfGCPY, 0.0);
+        EXPECT_EQ(x.dfGCPZ, 0.0);
+    }
+
+    // Test move constructor
+    {
+        GDAL_GCP x;
+        x.SetId("id");
+        x.SetInfo("info");
+        x.dfGCPPixel = 1.0;
+        x.dfGCPLine = 2.0;
+        x.dfGCPX = 3.0;
+        x.dfGCPY = 4.0;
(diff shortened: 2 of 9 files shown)


dg0yt:
This `<const_cast>(special empty string)` dance doesn't look healthy...

I assume `VSIFree` is used *much more* often than `GDAL_GCP`. Thus the extra test is quite expensive. Would it help to make the special empty string a non-static private member of `GDAL_GCP` instead of a global static? A little bit more memory for very few cases instead of a little bit more runtime for many cases.

Could `pszId` and  `pszInfo` be changed to `const char*`? This would fit well to the signatures of `SetId()` and `SetInfo()`.

rouault:
> Thus the extra test is quite expensive

I haven't measured the perf impact, but the only significant impact I would anticipate would only be in loops where you would do tons of malloc/free patterns, and that's obviously something that should be avoid be avoided upfront. I'd be very surprised that this shows up in realistic usage scenarios.

> Would it help to make the special empty string a non-static private member of `GDAL_GCP` instead of a global static?

This special static empty string that VSIFree() doesn't free is necessary if we want backward compatibility for scenarios like:
```GDAL_GCP x; GDALInitGCPs(1, &x);```.  Without the special empty string trick, the GDAL_GCP constructor would have to dynamically allocate one empty string, and GDALInitGCPs() would override it with another one, without freeing (since GDALInitGCPs() is aimed at being used on uninitialized data in its nominal usage), thus causing a memory leak.
We could fix all GDAL code to avoid doing GDALInitGCPs() on GDAL_GCP objects C++ constructed, but I've no idea of the impact on outside code.

> Could pszId and pszInfo be changed to const char*?

Definitely not for the C version of the structure. And for the C++ one, that would break all code that currently sets them directly (also hard to quantify how frequent this is outside of the GDAL codebase)

**Comment by dg0yt** [comment]
I see. I still feel a little bit worried by the mix of C, C++, and `const_cast`.

File: port/cpl_vsisimple.cpp [hunk 1]
```diff
@@ -834,6 +834,21 @@ void *VSIRealloc(void *pData, size_t nNewSize)
...
+static char *VSI_STATIC_EMPTY_STRING = const_cast<char *>("");
+
+/** Return an empty string ("") that VSIFree() does not free.
+ *
+ * That string should not be modified.
+ */
+char *VSIGetStaticEmptyString()
+{
+    return VSI_STATIC_EMPTY_STRING;
+}
```
dg0yt, line 850:
```suggestion
/** Return an empty string ("") that VSIFree() does not free.
 *
 * That string must not be modified.
 */
char *VSIGetStaticEmptyString()
{
    static char * const VSI_STATIC_EMPTY_STRING = CPLStrdup("");
    return VSI_STATIC_EMPTY_STRING;
}
```

port/cpl_vsisimple.cpp
```diff
@@ -843,10 +858,10 @@ void *VSIRealloc(void *pData, size_t nNewSize)
 void VSIFree(void *pData)
 
 {
-#ifdef DEBUG_VSIMALLOC
-    if (pData == nullptr)
+    if (pData == nullptr || pData == VSI_STATIC_EMPTY_STRING)
```
dg0yt, line 861:
```suggestion
    if (pData == nullptr || pData == VSIGetStaticEmptyString())
```

dg0yt (commented):
A single allocation. No `const_cast`.

port/cpl_vsisimple.cpp
rouault: Well, checking against VSIGetStaticEmptyString() involves a function call in which a static variable must be checked for being initialized in a thread-safe way. This will definitely have some cost compared to the approach to comparing to a static variable initialized at module loading.

dg0yt: Yes, there is a price tag, and it may prevent inlining.

It shouldn't be too hard to initialize a global-scope VSI_STATIC_EMPTY_STRING with a dynamic allocation without running in static initialization order fiasko.

But avoid entering undefined behaviour with a casting away constness from truly const memory.

(If it only wasn't too invasive to simply use nullptr instead of a empty strings...)

port/cpl_vsisimple.cpp
dg0yt:
```suggestion
static char VSI_STATIC_EMPTY_STRING = 0;

/** Return an empty string ("") that VSIFree() does not free.
 *
 * That string should not be modified.
 */
char *VSIGetStaticEmptyString()
{
    return &VSI_STATIC_EMPTY_STRING;
}
```

port/cpl_vsisimple.cpp
dg0yt: Actually nothing dynamic needed.

rouault:
Closing, as superseded per https://github.com/OSGeo/gdal/pull/9285

PR #9222, opened 2024-02-09
disable codeql checks on docs-only commits

tbonfort:


rouault:
Let's see how it goes, but I do believe I had tried that initially and CodeQL wasn't happy. I should have let a comment in the file...

tbonfort:
@rouault: marked as work in progress as I'm not confident this is the sufficient solution. I copied the same logic from build_linux.yml

rouault:
CodeQL seems happy... @tbonfort To confirm I've seen only ghosts, could you try to merge this PR in the master of your fork, push it, and then issue a PR against your fork that only modifies something under doc/  , and another one that modifies something else

tbonfort:
@rouault done:
doc only PR: https://github.com/tbonfort/gdal/pull/2
code only PR: https://github.com/tbonfort/gdal/pull/3

PR #9224, opened 2024-02-09
autotest: run Python tests directly from source tree on Windows

dbaston:
## What does this PR do?

Modifies CMake configuration so that Python tests are run directly from the source tree, rather than from a symlink or copy in the build directory. This is an experiment in removing the following "gotcha":

> (be careful on Windows, the 
content of $build_dir/autotest is copied from $source_dir/autotest each 
time "cmake" is run, so if you edit your test .py file directly in the 
build directory, be super careful of not accidentally losing your work, 
and make sure to copy its content to the source directory first. That's 
admittedly an annoying point of the current test setup on Windows, 
compared to Unix where we use symbolic links)

described in https://lists.osgeo.org/pipermail/gdal-dev/2024-February/058425.html

A downside of this change is that, because `pytest.ini` lives only in the build directory, we must always explicitly provide the config file via `pytest -c pytest.ini`. Is it an overall reduction in complexity? I don't know, am interested in feedback.

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/autotest/CMakeLists.txt
+++ b/autotest/CMakeLists.txt
@@ -49,6 +49,9 @@ if (Python_Interpreter_FOUND)
   configure_file(${GDAL_CMAKE_TEMPLATE_PATH}/pytest.ini.in ${CMAKE_CURRENT_BINARY_DIR}/pytest.ini @ONLY)
   unset(PYTEST_INI_HEADER_MESSAGE)
 
+  # Symlink test files into the build directory so that we can run tests directly from that location.
+  # This doesn't work on Windows,
+
 function (copy_file_or_dir source dest)
     if (IS_DIRECTORY ${source})
         message(STATUS "Copying contents of ${source} to ${destination}")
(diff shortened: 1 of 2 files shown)


rouault:
ah, this was mostly me not being aware of the -c option of pytest then. But perhaps we could keep the current symlink approach on non-Windows so that only Windows users have to pay the price of using an inferior operating system ;-) ?

rouault:
@dbaston do we need to keep that one opened ? I'm trying to reduce the number of staled PRs to keep the list as small as possible.

dbaston:
I updated this so that the tests are still symlinked on Linux/Mac, so I wasn't planning more work here. It can be merged or closed if we would prefer to continue copying the tests on Windows.

autotest/CMakeLists.txt
```diff
@@ -72,44 +75,43 @@ endfunction ()
...
+  else()
+      SET(python_test_path ${CMAKE_CURRENT_BINARY_DIR})
+      if (NOT "${CMAKE_BINARY_DIR}" STREQUAL "${CMAKE_SOURCE_DIR}")
+          foreach (subdir IN ITEMS
+                  pymod
+                  proj_grids
+                  cpp/data
+                  ${pytest_dirs})
+              if (SKIP_COPYING_AUTOTEST_SUBDIRS)
+                message(STATUS "Skipping copying ${CMAKE_CURRENT_SOURCE_DIR}/${subdir}")
```
rouault, line 92: That can be removed. It only made sense on Windows to sometimes speed-up things to avoid repeated copying of the directories. But here we are in the branch that only symlinks.

rouault:
> I updated this so that the tests are still symlinked on Linux/Mac, so I wasn't planning more work here. It can be merged or closed if we would prefer to continue copying the tests on Windows.

ok, sometimes I'm getting lost in what is ready for final review/merge or staled ;-) Don't hesitate to ping me if you see a PR that should be merged

Perhaps https://gdal.org/development/testing.html#running-a-subset-of-tests-using-pytest could get some updates to explain the subtelty on Windows of doing "python -m pytest -c autotest/pytest.ini ../autotest/gcore/vrt_read.py"

dbaston:
> Perhaps https://gdal.org/development/testing.html#running-a-subset-of-tests-using-pytest could get some updates to explain the subtelty on Windows of doing "python -m pytest -c autotest/pytest.ini ../autotest/gcore/vrt_read.py"

Good idea, I've done this at https://github.com/OSGeo/gdal/pull/9224/files#diff-1c899043c9c79440030b3ee7fcf72b357d3333d5070807d933f03ce80b4861edR51

doc/source/development/testing.rst
```diff
@@ -48,24 +48,34 @@ and the Python bindings:
...
+On Linux and MacOS builds, the tests are symlinked into the build directory, so this
+can be done by running the following from the build directory:
 
 .. code-block:: bash
 
-    pytest --collect-only autotest -k tiff
+    pytest autotest/gcore/vrt_read.py
 
-Running an individual test file
+On Windows, the test files remain in the source tree, but the pytest configuration file ``pytest.ini`` is only available in the build directory. To accommodate this, the above command would be  modified as follows:
```
rouault, line 59:
```suggestion
On Windows, the test files remain in the source tree, but the pytest configuration file ``pytest.ini`` is only available in the build directory. To accommodate this, the above command would be modified as follows:
```

PR #9231, opened 2024-02-13
OGC API - Map Tiles Content Negotiation

cfis:
## What does this PR do?

This commit updates OGC map tiles api to support additional image formats beyond PNG/JPG as well as enables content negotiation.

In our case, the motivating factor is we have map tiles that are stored as GeoTiffs which we would like to share via the OGC Tile API. Tiffs are one of the six predefined encodings allowed in an implementation - see https://docs.ogc.org/is/20-057/20-057.html#toc69.

As part of these changes, collections which do not specify a image type are now supported via content negotiatoin (see https://docs.ogc.org/is/20-057/20-057.html#toc28).

For example, previously this command used to fail and now succeeds:

```
gdalinfo "OGCAPI:https://test.cubewerx.com/cubewerx/cubeserv/demo/ogcapi/Daraa/collections/AgricultureSrf?f=json"
```

The reason for failure is the link with a "rel" value of "http://www.opengis.net/def/rel/ogc/1.0/tilesets-map" did not specify a type field. And thus the old code did not recognize it as JPEG or PNG and ignored it.

Finally, this commit removes the hard-coded value of l_nBands (3 for JPEG and 4 for PNG). Instead, it tries to download a tile and read the actual number of bands. It tries to read a tile at 0,0,0 and if can't falls back to 3? Perhaps that is too brittle? But the hard-coded band numbers weren't really correct, at least for JPEG, and definitely not for TIFF.

None

## Tasklist

 - [x] Verify add test cases pass
 - [x] Add test case(s)
 - [x] Updated documentation


frmts/ogcapi/gdalogcapidataset.cpp
```diff
@@ -1052,34 +1145,47 @@ bool OGCAPIDataset::InitWithMapAPI(GDALOpenInfo *poOpenInfo,
...
     if (osImageURL.empty())
     {
         CPLError(CE_Failure, CPLE_AppDefined,
-                 "Cannot find link to PNG or JPEG images");
+                 "Cannot find link to tileset items");
         return false;
     }
-    const int l_nBands = ((osImageURL == osPNG_URL) ? 4 : 3);
+
+    std::unique_ptr<GDALDataset> dataset =
```
rouault, line 1181: opening a sample tile has some cost. It would be preferrable if this could be avoided in common cases like JPEG or PNG

frmts/ogcapi/gdalogcapidataset.cpp
```diff
@@ -1877,7 +1990,15 @@ bool OGCAPIDataset::InitWithTilesAPI(GDALOpenInfo *poOpenInfo,
             CPLGetConfigOption("GDAL_WMS_MAX_CONNECTIONS", "5")));
         const char *pszTileMatrix =
             CSLFetchNameValue(poOpenInfo->papszOpenOptions, "TILEMATRIX");
-        const int l_nBands = ((osRasterURL == osPNG_URL) ? 4 : 3);
+
+        std::unique_ptr<GDALDataset> dataset =
```
rouault, line 1994: opening a sample tile has some cost. It would be preferrable if this could be avoided in common cases like JPEG or PNG

frmts/ogcapi/gdalogcapidataset.cpp
```diff
@@ -110,6 +111,11 @@ class OGCAPIDataset final : public GDALDataset
                       ", " MEDIA_TYPE_JSON,
                       CPLStringList *paosHeaders = nullptr);
 
+    std::unique_ptr<GDALDataset>
+    OpenTile(const CPLString &osURLPattern, int nMatrix, int nColumn, int nRow,
+             unsigned int nOpenTileFlags = 0, const CPLString osPrefix = {},
```
rouault, line 116:
```suggestion
             unsigned int nOpenTileFlags = 0, const CPLString &osPrefix = {},
```

frmts/ogcapi/gdalogcapidataset.cpp
```diff
@@ -587,6 +592,52 @@ bool OGCAPIDataset::DownloadJSon(const CPLString &osURL, CPLJSONDocument &oDoc,
...
 }
 
+/************************************************************************/
+/*                            OpenTile()                                */
+/************************************************************************/
+
+std::unique_ptr<GDALDataset>
+OGCAPIDataset::OpenTile(const CPLString &osURLPattern, int nMatrix, int nColumn,
+                        int nRow, unsigned int nOpenTileFlags,
+                        const CPLString osPrefix,
```
rouault, line 602:
```suggestion
                        const CPLString &osPrefix,
```

frmts/ogcapi/gdalogcapidataset.cpp
```diff
@@ -2487,18 +2588,14 @@ OGRFeature *OGCAPITiledLayer::GetNextRawFeature()
             {
                 return nullptr;
             }
-            bool bEmptyContent = false;
-            m_poUnderlyingDS.reset(OpenTile(m_nCurX, m_nCurY, bEmptyContent));
-            if (bEmptyContent)
+
+            m_poUnderlyingDS.reset(OpenTile(m_nCurX, m_nCurY));
+            if (!m_poUnderlyingDS)
```
rouault, line 2593: why is the logic that distinguishes empty content from a failure to open a tile removed?

frmts/ogcapi/gdalogcapidataset.cpp
cfis: Make sense. But isn't that only true for PNG (4 bands). JPEG images can have 1 or 3 bands. Is it ok to assume 3?

rouault: > Is it ok to assume 3?

Yes, I believe underlying code should auto-expand single bland to 3 band

jratike80: I read https://docs.ogc.org/is/20-057/20-057.html as png can have 1, 2, 3, or 4 bands (grayscale, grayscale and alpha, RGB, RGB and alpha. Are all these cases covered in the code?

jratike80:
Is there a problem if server implements this mixed-content thing

> If both “image/png” and “image/jpeg” are supported at an endpoint and are present in the HTTP “Accept” request header with the same q value, and no other supported output format is present with a higher q value, the server SHOULD return the requested image in whichever of these two formats is considered optimal (typically PNG if there’s transparency or for categorical maps where a limited number of colors is used, and JPEG otherwise).

frmts/ogcapi/gdalogcapidataset.cpp
cfis: Sorry, meant to point that out in the PR and then forgot.  The main reason was that it made for a more awkward api when moving `OpenTile` to `OGCAPIDataset` and my reading of the code was that it didn't make a difference.

So I just went back and read the original code to double check and see `bEmptyContent` used in 4 places:

1. line 2538:

```
    bool bEmptyContent = false;
    std::unique_ptr<GDALDataset> poUnderlyingDS(
        OpenTile(nX, nY, bEmptyContent));
```

`bEmptyContent` is not referenced in the rest of that method, so doesn't do anything.

2. Line 2567:

```
            bool bEmptyContent = false;
            m_poUnderlyingDS.reset(OpenTile(m_nCurX, m_nCurY, bEmptyContent));
            if (bEmptyContent || !m_poUnderlyingDS)
            {
                if (!IncrementTileIndices())
                    break;
                continue;
            }
```

In this case there is no distinction between returning a null pointer and `bEmptyContent  = true`

3. Line 2592:

```
                    bool bEmptyContent = false;
                    m_poUnderlyingDS.reset(
                        OpenTile(m_nCurX, m_nCurY, bEmptyContent));
                    if (bEmptyContent || !m_poUnderlyingDS)
                    {
                        continue;
                    }
```

In this case there is no distinction between returning a null pointer and `bEmptyContent  = true`

4. Line 2490:

```
bool bEmptyContent = false;
            m_poUnderlyingDS.reset(OpenTile(m_nCurX, m_nCurY, bEmptyContent));
            if (bEmptyContent)
            {
                if (!IncrementTileIndices())
                    return nullptr;
                continue;
            }
            if (m_poUnderlyingDS == nullptr)
            {
                return nullptr;
            }
```
In this case there is a difference. If `bEmptyContent` is true, then increment the indices, and loop again. Versus a nullptr in which case exist the method.

I collapsed those 2 together, so if a nullptr is returned then increment the indices and try again. 

The downside of that is if a server gives you an invalid url, you will check each tile before giving up. So one unsuccessful request could become many unsuccessful requests.

If that distinction is important then I can add `bEmptyContent`  back.

cfis:
@jratike80 - From what I see, the code didn't check that before so no change to that in this PR. Not sure how the code could know that, outside of the caller setting IMAGE_FORMAT to PNG_PREFERRED or JPEG_PREFERRED.

frmts/ogcapi/gdalogcapidataset.cpp
cfis: Ok, I added a new `OGCAPIDataset::FigureBands` method that assume 4 for PNG, 3 for JPEG and otherwise attempts to download a tile. If that fails, then it returns 3.

If we want to change the logic, it is now in one place.

frmts/ogcapi/gdalogcapidataset.cpp
cfis: See comment above.

frmts/ogcapi/gdalogcapidataset.cpp
cfis: Done

frmts/ogcapi/gdalogcapidataset.cpp
cfis: Done

cfis:
Formatting should be fixed.

Yeah, I did setup the precommit hooks, but they fail with the error below (but clang format does run). Googling tells me I should downgrade importlib_metadata (to a much lower version than I have) or in get a newer version of flake8....

------------

```
    self.manager = PluginManager(
                   ^^^^^^^^^^^^^^
  File "/home/cfis/.cache/pre-commit/reponfw0uzhs/py_env-python3.12/lib/python3.12/site-packages/flake8/plugins/manager.py", line 238, in __init__
    self._load_entrypoint_plugins()
  File "/home/cfis/.cache/pre-commit/reponfw0uzhs/py_env-python3.12/lib/python3.12/site-packages/flake8/plugins/manager.py", line 254, in _load_entrypoint_plugins
    eps = importlib_metadata.entry_points().get(self.namespace, ())
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'EntryPoints' object has no attribute 'get'
```

frmts/ogcapi/gdalogcapidataset.cpp
rouault: I would prefer keeping the "failed to open" vs "nothing was downloaded" distinction, because when iterating over vector tiles, we want to go on silently if a tile is missing (ie call IncrementTileIndices()), but if we download something and can't GDALOpen() it, then we should just stop

cfis: Ok. Suppose I can't just raise an exception?

rouault: > Suppose I can't just raise an exception?

we tend to avoid them in the code base

cfis:
Thanks for merging Even! I assume you saw that I added back in the empty content check.

PR #9237, opened 2024-02-16
Update gdaltransform.rst to use correct result

jidanni:


jratike80:
"Correct" is often too much promised when it comes to coordinate transformations, but my GDAL and Proj versions are rather much in line with your result.

```
gdaltransform -s_srs EPSG:28992 -t_srs EPSG:31370
177502 311865
244296.534583438 165937.283590881 0
```

jidanni:
(I actually also tried the inverse function, to reconstruct the input from the output, but these all never make the round trip intact.)

jidanni:
Actually this is pretty odd  We get different results on different machines. And round trip is impossible, even with the results that we get.

jratike80:
I believe that it is Proj that makes the work. My GDAL is using Proj version 9.3.1. And Proj is taking the definitions from the EPSG database and both CRS have had some edits not so long ago: https://epsg.org/crs_28992/Amersfoort-RD-New.html 2019 and https://epsg.org/crs_31370/BD72-Belgian-Lambert-72.html 2021. I did not check the changes but what I want to say that it is not so odd to get different results with different software versions.

jidanni:
What about on the same computer, not being able to do the round trip back using https://gdal.org/programs/gdaltransform.html#cmdoption-gdaltransform-i

PR #9243, opened 2024-02-18
gdalinfo: do not call GDALGetFileList() if -nofl is specified

rouault:


apps/gdalinfo_lib.cpp
```diff
@@ -290,30 +290,31 @@ char *GDALInfo(GDALDatasetH hDataset, const GDALInfoOptions *psOptions)
...
     {
-        if (bJson)
+        // The list of files of a raster FileGDB is not super useful and potentially
+        // super long, so omit it, unless the -json mode is enabled
+        char **papszFileList =
+            (!bJson && EQUAL(GDALGetDriverShortName(hDriver), "OpenFileGDB"))
+                ? nullptr
+                : GDALGetFileList(hDataset);
+
+        if (papszFileList == nullptr || *papszFileList == nullptr)
```
elpaso, line 302:
```suggestion
        if (!papszFileList || *papszFileList == nullptr)
```

if you are wondering why I changed the first check and not the second, I think that whenever the check is more complex than `var == nullptr` or `var != nullptr` the longer form better catches the eye.

elpaso (approved):

Issue #1108, opened 2018-11-20
Potential deadlock in `GDALAllRegister()`


artemp:
## Potential deadlock in `GDALAllRegister()`  
WARNING: ThreadSanitizer: lock-order-inversion (potential deadlock) (pid=9528)

This can be a "red herring" as sanitizer might not understanding code correctly ^ 

## Steps to reproduce the problem.

```c++
#include <iostream>
#include <gdal_version.h>
#include <gdal_priv.h>
int main()
{
    std::cerr << "GDAL" << std::endl;
    GDALAllRegister();
    std::cerr << "Done" << std::endl;
    return 0;
}
```

```bash
"g++"   -fPIC -O0 -fno-inline -Wall -g -fsanitize=thread -fno-omit-frame-pointer   -I"$(GDAL_DIR)/include" -c -o "bin/gcc-8.2.0/debug/main.o" "main.cpp"

"g++" -L"$(GDAL_DIR)/lib" -Wl,-rpath -Wl,"$(GDAL_DIR)/lib" -Wl,-rpath-link -Wl,"$(GDAL_DIR)/lib" -o "bin/gcc-8.2.0/debug/gdal-bug" -Wl,--start-group "bin/gcc-8.2.0/debug/main.o"  -Wl,-Bstatic  -Wl,-Bdynamic -lgdal -Wl,--end-group -fPIC -g -fsanitize=thread
```

Output: 

```
export TSAN_OPTIONS="second_deadlock_stack=1
./bin/gcc-8.2.0/debug/gdal-bug
```

```
GDAL
==================
WARNING: ThreadSanitizer: lock-order-inversion (potential deadlock) (pid=9528)
  Cycle in lock order graph: M55 (0x7fd0aec1d998) => M56 (0x7b1000000000) => M55

  Mutex M56 acquired here while holding mutex M55 in main thread:
    #0 pthread_mutex_lock <null> (libtsan.so.0+0x4016b)
    #1 CPLAcquireMutex /home/artem/projects/gdal/gdal/port/cpl_multiproc.cpp:1604 (libgdal.so.20+0xb9aeb3)
    #2 CPLCreateMutexInternal(bool, int) /home/artem/projects/gdal/gdal/port/cpl_multiproc.cpp:1581 (libgdal.so.20+0xb9aeb3)
    #3 __libc_start_main <null> (libc.so.6+0x2409a)

  Mutex M55 previously acquired by the same thread here:
[...]

PR #9253, opened 2024-02-19
GDALGetCacheMax64(): fix TSAN complaint (fixes #1837)

### LINKED ISSUE #1837: GDALGetCacheMax64() TSAN complaint
Super low priority...

In the I don't really care category, but TSAN complained about `GDALGetCacheMax64()`.  It doesn't like the bool check: `if( !bCacheMaxInitialized )`.  I don't think that two threads getting into this code would really be a big deal.   This would easily solve with `call_once`, but this would be the first use of call_once in GDAL.  But maybe it's not so easy since there interaction in `GDALSetCacheMax64()` for `bCacheMaxInitialized`

```c++
namespace {
std::once_flag flagSetupGDALGetCacheMax64;
}

GIntBig CPL_STDCALL GDALGetCacheMax64()
{
    std::call_once(flagSetupGDALGetCacheMax64, [](){
        // All the setup code here.
    });
    return nCacheMax;
}
```

https://en.cppreference.com/w/cpp/thread/call_once

rouault:


port/cpl_multiproc.cpp
```diff
@@ -1462,35 +1462,31 @@ int CPLCreateOrAcquireMutexEx(CPLMutex **phMutex, double dfWaitInSeconds,
...
 
 {
-    bool bSuccess = false;
-
     pthread_mutex_lock(&global_mutex);
     if (*phMutex == nullptr)
     {
         *phMutex = CPLCreateMutexInternal(true, nOptions);
-        bSuccess = *phMutex != nullptr;
+        const bool bSuccess = *phMutex != nullptr;
```
elpaso, line 1469: are the changes in this file related to the TSAN complaint mentioned above?

On a side note, is there chance we can eventually get rid of in-house mutex handling in favor of std:: library methods?

elpaso (approved):

port/cpl_multiproc.cpp
rouault: > are the changes in this file related to the TSAN complaint mentioned above?

not directly, I also incorporated that commit from #1108 because a sanitizer that would test this PR would also frown upon that issue

> On a side note, is there chance we can eventually get rid of in-house mutex handling in favor of std:: library methods?

could be a nice cleanup indeed.

PR #9255, opened 2024-02-19
VRTDerivedRasterBand: Support Int8, (U)Int64 with Python pixel functions

dbaston:
## What does this PR do?

Avoid assertion failure in when using Python pixel function with bands of int8 and (u)int64 types.

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/autotest/gdrivers/vrtderived.py
+++ b/autotest/gdrivers/vrtderived.py
@@ -1024,6 +1024,51 @@ def identity(in_ar, out_ar, xoff, yoff, xsize, ysize, raster_xsize, raster_ysize
 
 
 ###############################################################################
+
+
+@pytest.mark.parametrize("dtype", range(1, gdal.GDT_TypeCount))
+def test_vrt_derived_dtype(tmp_vsimem, dtype):
+    pytest.importorskip("numpy")
+
+    input_fname = tmp_vsimem / "input.tif"
+
+    nx = 1
+    ny = 1
+
+    with gdal.GetDriverByName("GTiff").Create(
+        input_fname, nx, ny, 1, eType=gdal.GDT_Int8
+    ) as input_ds:
+        input_ds.GetRasterBand(1).Fill(1)
+        gt = input_ds.GetGeoTransform()
+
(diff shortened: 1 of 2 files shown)


rouault:
We should probably also remove the "default:" case in the switch() to catch more easily such omission in the future

frmts/vrt/vrtderivedrasterband.cpp
```diff
@@ -108,15 +117,14 @@ static PyObject *GDALCreateNumpyArray(PyObject *pCreateArray, void *pBuffer,
             break;
         case GDT_CInt16:
         case GDT_CInt32:
-            CPLAssert(FALSE);
-            break;
```
dbaston, line 112: OK to coerce these types to CFloat32?

dbaston:
> We should probably also remove the "default:" case in the switch() 

done

frmts/vrt/vrtderivedrasterband.cpp
rouault: I don't remember the details of how this works, but it is not obvious to me that this will cause a cint16/cint32 -> cfloat32 type conversion. This might require a prior type conversion
Actually I see at lines 1208 - 1220 that we already error out on GDT_CInt16/GDT_CInt32, so the assertion here is legit

dbaston: restored it

autotest/gdrivers/vrtderived.py
```diff
@@ -1023,6 +1023,49 @@ def identity(in_ar, out_ar, xoff, yoff, xsize, ysize, raster_xsize, raster_ysize
     _validate(xml)
 
 
+###############################################################################
+
+
+@pytest.mark.parametrize("dtype", range(1, gdal.GDT_TypeCount))
+def test_vrt_derived_dtype(tmp_vsimem, dtype):
+    input_fname = tmp_vsimem / "input.tif"
```
rouault, line 1031: numpy availability should be checked (cf failure on windows minimum build)

PR #9278, opened 2024-02-22
[gpkg] Ensure that tables present in gpkgext_relations can be read

nyalldawson:
Tables which are listed in gpkgext_relations but NOT gpkg_contents should be readable as valid layers. The GeoPackage related tables extension explicitly states that tables in gpkgext_relations are not required to be listed in gpkg_contents.

### DIFF
--- a/autotest/ogr/ogr_gpkg.py
+++ b/autotest/ogr/ogr_gpkg.py
@@ -7008,6 +7008,13 @@ def test_ogr_gpkg_relations(tmp_vsimem, tmp_path):
     assert rel.GetRightMappingTableFields() == ["related_id"]
     assert rel.GetRelatedTableType() == "attributes"
 
+    # ensure that the mapping table, which is present in gpkgext_relations but
+    # NOT gpkg_contents can be opened as a layer
+    lyr = ds.GetLayer("my_mapping_table")
+    assert lyr is not None
+    assert lyr.GetLayerDefn().GetFieldDefn(0).GetName() == "base_id"
+    assert lyr.GetLayerDefn().GetFieldDefn(1).GetName() == "related_id"
+
     lyr = ds.GetLayer("a")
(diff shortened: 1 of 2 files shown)


autotest/ogr/ogr_gpkg.py
```diff
@@ -7269,6 +7276,21 @@ def get_query_row_count(query):
         )
         == 1
     )
+    # force delete from gpkg_contents, and then ensure that we CAN successfully
+    # load layers which are present ONLY in gpkgext_relations but NOT
+    # gpkg_contents (i.e. datasources which fully the Related Tables specification
```
rouault, line 7281:
```suggestion
    # gpkg_contents (i.e. datasources which follow the Related Tables specification
```

rouault (commented):
the approach seems very reasonable to me

ogr/ogrsf_frmts/gpkg/ogrgeopackagedatasource.cpp
```diff
@@ -1575,6 +1576,12 @@ int GDALGeoPackageDataset::Open(GDALOpenInfo *poOpenInfo,
                           "name = 'gpkg_geometry_columns' AND "
                           "type IN ('table', 'view')",
                           nullptr) == 1;
+        bHasGPKGExtRelations =
+            SQLGetInteger(hDB,
+                          "SELECT 1 FROM sqlite_master WHERE "
+                          "name = 'gpkgext_relations' AND "
+                          "type IN ('table', 'view')",
+                          nullptr) == 1;
```
rouault, line 1584:
```suggestion
        bHasGPKGExtRelations = HasGpkgextRelationsTable();
```

nyalldawson: Done

PR #9279, opened 2024-02-22
[gpkg] Also read relationships defined using foreign key constraints

nyalldawson:
When reading relationships, always read relationships defined using foreign key constraints regardless of whether or not the related tables extension is in use.

The related table extension only permits definition of many-to-many relationships, so there's a strong case for supporting one-to-many relationships defined outside of this extension. In fact it's what's recommended upstream: https://github.com/opengeospatial/geopackage/issues/678#issuecomment-1954402113

### DIFF
--- a/ogr/ogrsf_frmts/sqlite/ogrsqlitedatasource.cpp
+++ b/ogr/ogrsf_frmts/sqlite/ogrsqlitedatasource.cpp
@@ -284,7 +284,9 @@ std::vector<std::string> OGRSQLiteDataSource::GetRelationshipNames(
 
 {
     if (!m_bHasPopulatedRelationships)
-        LoadRelationshipsFromForeignKeys();
+    {
+        LoadRelationships();
+    }
 
     std::vector<std::string> oasNames;
     oasNames.reserve(m_osMapRelationships.size());
@@ -305,7 +307,9 @@ OGRSQLiteDataSource::GetRelationship(const std::string &name) const
 
 {
     if (!m_bHasPopulatedRelationships)
-        LoadRelationshipsFromForeignKeys();
+    {
+        LoadRelationships();
+    }
 
     auto it = m_osMapRelationships.find(name);
     if (it == m_osMapRelationships.end())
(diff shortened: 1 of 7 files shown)


ogr/ogrsf_frmts/sqlite/ogrsqlitedatasource.cpp
```diff
@@ -728,8 +732,25 @@ void OGRSQLiteBaseDataSource::LoadRelationshipsFromForeignKeys() const
...
-            "ORDER BY m.name");
+            "'virts_geometry_columns') ";
+        if (!excludedTables.empty())
+        {
+            std::string oExcludedTablesList;
+            for (const auto &osExcludedTable : excludedTables)
+            {
+                oExcludedTablesList += !oExcludedTablesList.empty() ? "," : "";
+                oExcludedTablesList +=
+                    sqlite3_mprintf("'%q'", osExcludedTable.c_str());
```
rouault, line 743:
```suggestion
                char* pszEscapedName = sqlite3_mprintf("'%q'", osExcludedTable.c_str());
                oExcludedTablesList += pszEscapedName;
                sqlite3_free(pszEscapedName);
```

ogr/ogrsf_frmts/sqlite/ogrsqlitedatasource.cpp
```diff
@@ -284,7 +284,10 @@ std::vector<std::string> OGRSQLiteDataSource::GetRelationshipNames(
 
 {
     if (!m_bHasPopulatedRelationships)
-        LoadRelationshipsFromForeignKeys();
+    {
+        m_osMapRelationships.clear();
+        LoadRelationshipsFromForeignKeys({});
+    }
```
rouault, line 290: A possibly cleaner solution would be to have a *virtual* LoadRelationships() method in OGRSQLiteBaseDataSource that would do:
```
void OGRSQLiteDataSource::LoadRelationships()
{
        m_osMapRelationships.clear();
        LoadRelationshipsFromForeignKeys({});
        m_bHasPopulatedRelationships = true;
}
```
and call it from there
and have the LoadRelationships() method of GPKG override it
and remove the ``m_bHasPopulatedRelationships = true;`` statement from ogrsqlitedatasource.cpp:831

ogr/ogrsf_frmts/sqlite/ogrsqlitedatasource.cpp
```diff
@@ -305,7 +308,10 @@ OGRSQLiteDataSource::GetRelationship(const std::string &name) const
...
```
rouault, line 314: same comment as above

nyalldawson:
@rouault thanks! Do you think this is appropriate for backporting?

rouault:
At this stage of 3.8.x, the benefit vs risk is a bit low given than 3.8.5 will probably be the last point release in the 3.8.x series

PR #9292, opened 2024-02-24
Update gdaltransform.rst to hint on running gdalinfo on srcfile

jidanni:
Else one might think GCPs can be given in srcfile one per line,
```
<pixel> <line> <easting> <northing> [<elevation>]
```

### DIFF
--- a/doc/source/programs/gdaltransform.rst
+++ b/doc/source/programs/gdaltransform.rst
@@ -120,8 +120,7 @@ projection,including GCP-based transformations.
     not given, source projection/GCPs are read from the command-line :option:`-s_srs`
     or :option:`-gcp` parameters.
 
-    Ome might first inspect via gdalinfo <srcfile> to ensure all
-    projections/GCPs will be interpreted correctly.
+    Note that only the SRS and/or GCPs of this input file is taken into account, and not its pixel content.
 
 .. option:: <dstfile>
 
--- a/doc/source/programs/gdaltransform.rst
+++ b/doc/source/programs/gdaltransform.rst
@@ -116,9 +116,12 @@ projection,including GCP-based transformations.
 
 .. option:: <srcfile>
 
-    File with source projection definition or GCP's. If
-    not given, source projection is read from the command-line :option:`-s_srs`
-    or :option:`-gcp` parameters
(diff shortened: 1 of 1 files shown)


doc/source/programs/gdaltransform.rst
```diff
@@ -116,9 +116,12 @@ projection,including GCP-based transformations.
...
 
-    File with source projection definition or GCP's. If
-    not given, source projection is read from the command-line :option:`-s_srs`
-    or :option:`-gcp` parameters
+    File with source projection definition or GCPs. If
+    not given, source projection/GCPs are read from the command-line :option:`-s_srs`
+    or :option:`-gcp` parameters.
+
+    Ome might first inspect via gdalinfo <srcfile> to ensure all
+    projections/GCPs will be interpreted correctly.
```
rouault, line 124: The following should be clearer, shouldn't it ?

```suggestion
    Note that only the SRS and/or GCPs of this input file is taken into account, and not its pixel content.
```

jidanni: OK, but please be sure to also mention gdalinfo: 

> Note that only the SRS and/or GCPs of this input file is taken into account, and not its pixel content. Confirm using gdalinfo.

(And hyperlink gdalinfo.)

Else some users like me thought they could use csv/xyz files as GCP files etc.

A true life story:
I gave it several lines of
```
<pixel> <line> <easting> <northing> [<elevation>]
<pixel> <line> <easting> <northing> [<elevation>]...
```
And wondered why the output didn't look like `<easting> <northing>`
but more like `<pixel> <line>`. But never mind that.

rouault: > please be sure to also mention gdalinfo:

I don't see the point. Let's keep the text to the minimum needed.

jidanni: The main problem is the format of `<srcfile>` is never specified.

rouault: > The main problem is the format of `<srcfile>` is never specified.

==> now done per https://github.com/OSGeo/gdal/commit/ef28df5aa1f461ed3a81ffc3f939082116f871fc

PR #9294, opened 2024-02-24
Update index.rst noting offline individual HTML pages .ZIP

### LINKED ISSUE #9228: Offer HTML offline documentation, in addition to current PDF
Just like on https://github.com/OSGeo/PROJ/issues/4051, could the GDAL
project offer, in addition to the current PDF document download,
an additional HTML file download, please?

I.e., a zip file, of documentation only, consisting of separate .html
files. Just like the https://gdal.org/ website itself. (Not just one
massive .html file too.)

This zip file should not be stuck inside a larger download, but
independent. No, there is no more offline doc package:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=960964#12

Ah, on https://docs.generic-mapping-tools.org/dev/
in the lower left corner, one can download
https://docs.generic-mapping-tools.org/dev/gmt-docs-dev.zip
which is indeed ideal: individual files, with even an search
box (in its side panel) that works offline!

jidanni:
Closes #9228.

### DIFF
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -4,7 +4,10 @@ GDAL
 
 .. include:: ./about_no_title.rst
 
-This documentation is also available as a `PDF file <gdal.pdf>`_.
+This documentation is also available as a `PDF file <gdal.pdf>`_,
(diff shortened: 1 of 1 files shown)


jidanni:
And what do you know, gdal.pdf is also included in the .ZIP, (taking up about 4 of the 22MB).)

PR #9295, opened 2024-02-24
XLSX: Add OLCStringsAsUTF8 capability

theroggy:
The OLCStringsAsUTF8 capability seems to be missing for the XLSX driver.

Noticed this when using this capability pyogrio to write XLSX files and getting an UTF-8 error when trying to read it again:
https://github.com/geopandas/pyogrio/pull/361

### DIFF
--- a/ogr/ogrsf_frmts/xlsx/ogrxlsxdatasource.cpp
+++ b/ogr/ogrsf_frmts/xlsx/ogrxlsxdatasource.cpp
@@ -260,6 +260,8 @@ int OGRXLSXDataSource::TestCapability(const char *pszCap)
         return true;
     else if (EQUAL(pszCap, ODsCCurveGeometries))
         return true;
+    else if (EQUAL(pszCap, OLCStringsAsUTF8))
+        return true;
     else
         return false;
 }


ogr/ogrsf_frmts/xlsx/ogrxlsxdatasource.cpp
```diff
@@ -260,6 +260,8 @@ int OGRXLSXDataSource::TestCapability(const char *pszCap)
         return true;
     else if (EQUAL(pszCap, ODsCCurveGeometries))
         return true;
+    else if (EQUAL(pszCap, OLCStringsAsUTF8))
+        return true;
```
rouault, line 264: This is incorrect. OLCStringsAsUTF8 is a *layer* capability, and here this is the TestCapability() method of a *dataset*
Furthermore declaring UTF-8 is already done with https://github.com/OSGeo/gdal/blob/master/ogr/ogrsf_frmts/xlsx/ogrxlsxdatasource.cpp#L825

rouault: Demo:
```
$ python -c "from osgeo import ogr; ds = ogr.Open('autotest/ogr/data/xlsx/test.xlsx'); print(ds.GetLayer(0).TestCapability(ogr.OLCStringsAsUTF8))"
True
``

theroggy: Oops, sorry, thanks. I didn't know about the different levels of TestCapability, nor that OLCStringsAsUTF8 could be set as it is being done in ogrxlsxdatasource.cpp.
I'm doing more testing to try to find what exactly is happening, but because I can only reproduce it on a CI machine it is a pain to debug...

theroggy: I added extra debugging info in the code passed by the unit test "misbehaving" and it is quite weird:
- <Running on my Windows machine, OGR_L_TestCapability for OLCStringsAsUTF8 for a new XLSX file being created returns True
- on the Windows 2019 CI machine on github for the same code called from the same unit test TestCapability returns False.

I don't how it could be related, but you never know: another difference between my PC and the github image is that `locale.getpreferredencoding()` returns "UTF-8" in the tests on my computer and "cp1252" on github.

Can you think of any reason/explanation?

rouault: > Can you think of any reason/explanation?

not really, SetAdvertizeUTF8() has been set on the XLSX driver since day 1 of the driver (https://github.com/OSGeo/gdal/commit/5cf32d6590e58eb3bf7b9d1579a4f88c748f3fa9#diff-9174ca7e9f027d9789bd91e7a74cf446b66694a95efd915ac992ea913644f221R536). This doesn't depend on anything external
One explanation could be that for some reason on the CI machine, another driver than the XLSX one would recognize the file...

theroggy: I'm not sure when SetAdvertizeUTF8() is called in the XLSX code, but I wonder if it is too late for the way it is used in pyogrio.
The case I'm talking about is the function to create a new xlsx file. For this case, the main relevant things are:
- GDALCreate is called to create the new file
- GDALDatasetCreateLayer is called to create a new layer
- GDALSetMetadata is called to set some metadata to the dataset and/or layer

Then OGR_L_TestCapability is called to check the encoding needed...

Would it be possible that this is OK for most drivers, but that for XLSX SetAdvertizeUTF8() hasn't been called yet at that point?

I already tried to move the OGR_L_TestCapability call to after the fields have been added to the layer using amongst others OGR_L_CreateField, but that didn't help. It isn't ideal either because I can imaging that (in some cases ?) the encoding is also relevant for the column names being created?

rouault: > Would it be possible that this is OK for most drivers, but that for XLSX SetAdvertizeUTF8() hasn't been called yet at that point?

ah that was it! Fix in https://github.com/OSGeo/gdal/pull/9301

PR #9297, opened 2024-02-24
Update gdal_grid.rst to mention -ot defaults to a guess that might no…

jidanni:
…t be appropriate for the relevant driver

Please be aware that I cannot see how my changes render on GitHub, even when looking at Preview, so, well, I hope they look good.

P.S., proof of validity of what I am trying to mention:
```
$ gdal_grid --debug on ... n.gpkg
...
GDAL: GDALDriver::Create(GPKG,n.gpkg,...,Float64,...)
ERROR 6: Only Byte, Int16, UInt16 or Float32 supported
```

### DIFF
--- a/doc/source/programs/gdal_grid.rst
+++ b/doc/source/programs/gdal_grid.rst
@@ -48,6 +48,9 @@ computer.
 
 .. include:: options/ot.rst
 
+   If not set then a default type is used, which might not be supported
+   by the relevant driver.
+
 .. include:: options/of.rst
 
 .. option:: -txe <xmin> <xmax>


doc/source/programs/gdal_grid.rst
```diff
@@ -48,6 +48,9 @@ computer.
 
 .. include:: options/ot.rst
 
+   If not set then a default type is used, which might not be supported
+   by the relevant driver.
+
```
rouault, line 53: You cannot put content after a ``.. include`` statement.
I don't see this comment as really necessary as the user will get an explicit error when the situation arises. If you're keen on keeping this comment, you should move it just before the options detail for example, at the topo of the "Description"  paragraph

jidanni: Fixed in #9333.

PR #9303, opened 2024-02-25
Update index.rst to say the ZIP includes the PDF

jidanni:
https://github.com/OSGeo/gdal/pull/9294#issuecomment-1962284238


rouault:
I don't see the point

jidanni:
Many people will download both, not realizing the latter also contains the former.

rouault:
ok...

PR #9311, opened 2024-02-25
Disable my_test_sqlite3_ext in static builds

dg0yt:
Cf. https://lists.osgeo.org/pipermail/gdal-dev/2024-February/058508.html
The change disables the module for all static builds (`BUILD_SHARED_LIBS` off). But the key goal is to disable it in builds with `-static` CRT linkage on linux.
Alternatively, it could be flagged as `EXCLUDE_FROM_ALL`, but then it might bit rot.
AFAICS the module is neither used nor installed.


rouault:
> AFAICS the module is neither used

actually just by one test (https://github.com/OSGeo/gdal/blob/fe693c6d13ccfd3c33420961891645b274095825/autotest/ogr/ogr_virtualogr.py#L397), but this is optional

PR #9313, opened 2024-02-26
Make sure our vendored flatbuffers copy has a unique namespace

rouault:
Cf https://lists.osgeo.org/pipermail/gdal-dev/2024-February/058544.html


rouault:
CC @bjornharrtell

simoneves:
That's even hackier than how I fixed it locally, but at least I picked the same name! :)

bjornharrtell:
Looks sensible to me

elpaso (approved):

PR #9333, opened 2024-02-27
Update gdal_grid.rst to warn of errors due to not setting type

jidanni:
https://github.com/orgs/sphinx-doc/discussions/12005#discussioncomment-8584525 mentions how to add the paragraph between two includes!


rouault:
The resulting formatting is not idea, but let's go for that

PR #9336, opened 2024-02-27
[warp] fast track for same projection

elpaso:
Fixes  a downstream issue: https://github.com/qgis/QGIS/issues/56288

### DIFF
--- a/alg/gdaltransformer.cpp
+++ b/alg/gdaltransformer.cpp
@@ -428,6 +428,62 @@ CPLErr CPL_STDCALL GDALSuggestedWarpOutput2(GDALDatasetH hSrcDS,
     const int nInXSize = GDALGetRasterXSize(hSrcDS);
     const int nInYSize = GDALGetRasterYSize(hSrcDS);
 
+    /* ------------------------------------------------------------- */
+    /* Special case for warping on the same (or null) CRS.           */
+    /* ------------------------------------------------------------- */
+    if (pTransformArg && pfnTransformer == GDALGenImgProjTransform)
+    {
+        double adfGeoTransform[6];
+
+        if (GDALGetGeoTransform(hSrcDS, adfGeoTransform) == CE_None &&
+            !(adfGeoTransform[0] == 1.0 && adfGeoTransform[1] == 1.0 &&
+              adfGeoTransform[2] == 0.0 && adfGeoTransform[3] == 1.0 &&
+              adfGeoTransform[4] == 0.0 && adfGeoTransform[5] == 1.0))
+        {
+            GDALGenImgProjTransformInfo *psInfo{
+                static_cast<GDALGenImgProjTransformInfo *>(pTransformArg)};
+
+            if (psInfo->pSrcTransformer != GDALRPCTransform)
+            {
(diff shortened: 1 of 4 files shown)


alg/gdaltransformer.cpp
```diff
@@ -428,6 +428,76 @@ CPLErr CPL_STDCALL GDALSuggestedWarpOutput2(GDALDatasetH hSrcDS,
...
+    /* Special case for warping on the same (or null) CRS.           */
+    /* ------------------------------------------------------------- */
+    if (pTransformArg && pfnTransformer == GDALGenImgProjTransform)
+    {
+        double adfGeoTransform[6];
+
+        if (GDALGetGeoTransform(hSrcDS, adfGeoTransform) == CE_None &&
+            !(adfGeoTransform[0] == 1.0 && adfGeoTransform[1] == 1.0 &&
+              adfGeoTransform[2] == 0.0 && adfGeoTransform[3] == 1.0 &&
+              adfGeoTransform[4] == 0.0 && adfGeoTransform[5] == 1.0))
```
rouault, line 441: I still think this test is a band-aid to help a test case pass, but is not correct. Either we lack additional protection to not run this new code, or the failing test should be adapted if that makes sense

rouault: but we should definitely test that adfGeoTransform[2] and adfGeoTransform[4] are 0, ie. no rotation

elpaso: Changed to only test no rotation, all test are passing (locally), waiting for CI.

autotest/alg/warp.py
```diff
@@ -1669,3 +1669,73 @@ def test_warp_average_oversampling():
             assert math.isnan(got_data[(y + 4) * 14 + (14 - 1 - x)])
         for x in range(6):
             assert got_data[(y + 4) * 14 + (x + 4)] == 3.0
+
+
+###############################################################################
+# Test bug fix for https://github.com/qgis/QGIS/issues/56288
+
+
+def test_non_square():
```
rouault, line 1678: to fix failure on the windows-minimal config (that removes all optional drivers like XYZ), cf https://github.com/OSGeo/gdal/actions/runs/8080871845/job/22078186296?pr=9336:

```suggestion
@pytest.mark.require_driver("XYZ")
def test_non_square():
```

alg/gdaltransformer.cpp
```diff
@@ -428,6 +432,116 @@ CPLErr CPL_STDCALL GDALSuggestedWarpOutput2(GDALDatasetH hSrcDS,
...
+                for (int i = 0; i < 6; i++)
+                {
+                    fprintf(stderr,
+                            "GDALSuggestedWarpOutput2 "
+                            "psInfo->adfDstGeoTransform[%d] = %.18g\n",
+                            i, psInfo->adfDstGeoTransform[i]);
+                }
+            }
+
+            if (psInfo && !psInfo->pSrcTransformer &&
```
jjimenezshaw, line 473: Which case is covering this code?

jjimenezshaw:
Thanks @elpaso !

PR #9350, opened 2024-02-28
Doc: Add Python docstrings for some functions in gdal module

dbaston:


swig/include/python/docs/gdal_docs.i
```diff
@@ -0,0 +1,354 @@
...
+%feature("docstring") CPLGetNumCPUs "
+
+Return the number of processors detected by GDAL.
+
+Returns
+-------
+int
+";
+
+%feature("docstring") wrapper_GDALGetCacheMax "
```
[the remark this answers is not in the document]
dbaston: > Or maybe we need to attach the docstring to the wrapper_xxxx ?

Yes, this unfortunately seems to be the case when we use a `%rename` directive. I will add comments to the file indicating the end user methods.

swig/include/python/docs/gdal_docs.i
```diff
@@ -0,0 +1,365 @@
...
+>>> with gdal.config_options({'A': '3', 'B': '4'}):
+...     gdal.SetConfigOption('C', '5')
+...     gdal.GetConfigOptions()
+...
+{'C': '5', 'A': '3', 'B': '4'}
+
+See Also
+--------
+:py:func:`GetConfigOption`
+:py:func:`GetGlobalConfigOptions`
```
agiudiceandrea, line 83: @rouault, does GetGlobalConfigOptions actually exists?

rouault: apparently not

PR #9351, opened 2024-02-28
gdal_polygonize.py: handle error when the creation of the destination datasource fails

agiudiceandrea:
## What does this PR do?

Makes gdal_polygonize.py print an error message when the creation of the destination datasource fails.

Otherwise, in such case, the script will fails at https://github.com/OSGeo/gdal/blob/master/swig/python/gdal-utils/osgeo_utils/gdal_polygonize.py#L162 and a Python traceback error message will be printed.

### DIFF
--- a/swig/python/gdal-utils/osgeo_utils/gdal_polygonize.py
+++ b/swig/python/gdal-utils/osgeo_utils/gdal_polygonize.py
@@ -146,6 +146,9 @@ def gdal_polygonize(
         if not quiet:
             print("Creating output %s of format %s." % (dst_filename, driver_name))
         dst_ds = drv.CreateDataSource(dst_filename)
+        if dst_ds is None:
+            print('Cannot create datasource "%s"' % dst_filename)
+            return 1
 
     # =============================================================================
     #       Find or create destination layer.


swig/python/gdal-utils/osgeo_utils/gdal_polygonize.py
```diff
@@ -146,6 +146,9 @@ def gdal_polygonize(
         if not quiet:
             print("Creating output %s of format %s." % (dst_filename, driver_name))
         dst_ds = drv.CreateDataSource(dst_filename)
+        if dst_ds is None:
```
rouault, line 149: I don't think that change will do anything, because the exception happens above:
```
$ gdal_polygonize.py byte.tif /non_existing/bar.gpkg
Creating output /non_existing/bar.gpkg of format GPKG.
  File "/home/even/gdal/gdal/build_cmake/swig/python/osgeo_utils/auxiliary/gdal_argparse.py", line 226, in main
    self.doit(**kwargs)
  File "/home/even/gdal/gdal/build_cmake/swig/python/osgeo_utils/gdal_polygonize.py", line 351, in doit
    return gdal_polygonize(**kwargs)
  File "/home/even/gdal/gdal/build_cmake/swig/python/osgeo_utils/gdal_polygonize.py", line 148, in gdal_polygonize
    dst_ds = drv.CreateDataSource(dst_filename)
  File "/home/even/gdal/gdal/build_cmake/swig/python/osgeo/ogr.py", line 717, in CreateDataSource
    return _ogr.Driver_CreateDataSource(self, *args, **kwargs)
RuntimeError: sqlite3_open(/non_existing/bar.gpkg) failed: unable to open database file
```
Are exceptions returned to the user a bad thing?

agiudiceandrea: I was referring to the case in https://github.com/qgis/QGIS/issues/56588 (and in the other one about xtf).
In such case the processing log is:
```
Creating output C:/OSGeo4W_V2_6/test.xtf of format Interlis 2.
File "C:\OS2019~1\apps\Python39\lib\site-packages\osgeo_utils\auxiliary\gdal_argparse.py", line 226, in main 
self.doit(**kwargs) 
File "C:\OS2019~1\apps\Python39\lib\site-packages\osgeo_utils\gdal_polygonize.py", line 351, in doit 
return gdal_polygonize(**kwargs) 
File "C:\OS2019~1\apps\Python39\lib\site-packages\osgeo_utils\gdal_polygonize.py", line 162, in gdal_polygonize 
dst_layer = dst_ds.CreateLayer( 
AttributeError: 'NoneType' object has no attribute 'CreateLayer'
```

While with this change, it is more understandable:
```
Creating output C:/OSGeo4W_V2_6/test.xtf of format Interlis 2.
Cannot create datasource "C:/OSGeo4W_V2_6/test.xtf"
```

Anyway you are right that when the path is not existent it fails at https://github.com/OSGeo/gdal/blob/8f72c602dc3774eea3f30bfa25ec529d70f69611/swig/python/gdal-utils/osgeo_utils/gdal_polygonize.py#L148, or e.g. also at https://github.com/OSGeo/gdal/blob/8f72c602dc3774eea3f30bfa25ec529d70f69611/swig/python/gdal-utils/osgeo_utils/gdal_polygonize.py#L177 for some formats (like DXF) with
```
Creating output C:/OSGeo4W_V2_6/test.dxf of format DXF.
File "C:\OS2019~1\apps\Python39\lib\site-packages\osgeo_utils\auxiliary\gdal_argparse.py", line 226, in main 
self.doit(**kwargs) 
File "C:\OS2019~1\apps\Python39\lib\site-packages\osgeo_utils\gdal_polygonize.py", line 354, in doit 
return gdal_polygonize(**kwargs) 
File "C:\OS2019~1\apps\Python39\lib\site-packages\osgeo_utils\gdal_polygonize.py", line 180, in gdal_polygonize 
dst_layer.CreateField(fd) 
File "C:\OS2019~1\apps\Python39\lib\site-packages\osgeo\ogr.py", line 1922, in CreateField 
return _ogr.Layer_CreateField(self, *args, **kwargs) 
RuntimeError: DXF layer does not support arbitrary field creation, field 'DN' not created.
```

Not sure if it is really useful to gracefully catch all those exceptions...

agiudiceandrea: Clearly, feel free to close this PR, if it is not really useful.

rouault: @agiudiceandrea I'm trying to understand the situations where CreateDataSource() would return a None object without throwing an exception. gdal.UseExceptions() has been enabled in utilities in 3.7.0, and in the report the user is using 3.7.1, so this should normally not happen. But actually I just found this is a oddity of the ILI2 driver that I've just fixed per https://github.com/OSGeo/gdal/pull/9354
Anyway merging your fix as it doesn't hurt and can help in such weird cases

PR #9390, opened 2024-03-06
Update gdal_grid.rst about layers not found

jidanni:
Actually it really should quit and return $?=1 to the shell!


rouault:
Closing that one as we don't to document bugs that can be reasonably fixed, so @jidanni please create an issue about that

PR #9394, opened 2024-03-06
Update ogr2ogr.rst to make sure users realize only the last -sql counts

jidanni:


jratike80:
I do not believe that this behavior will change. Could it be enough to edit https://gdal.org/programs/ogr2ogr.html#cmdoption-ogr2ogr-sql and change "SQL statement to execute" into "A single SQL statement to execute"?

jidanni:
@jratike80 All I know is currently if the user uses -sql ... -sql ... only perhaps years later would he discover that one of them was not working. Therefore it would even be better if (you could make) the program also please print an error message and bomb out, as well as document it in that place you mentioned. Thanks.

jratike80:
Could you consider how to make this existing document more visible for the users https://gdal.org/programs/argument_syntax.html#argument-syntax? As an explanation it could be added  that many `-oo` options can be used in this case because of the ellipsis in the syntax definition. An `-sql` example could be added as an example of the opposite case, or this `-a_srs` example that gives EPSG:2393 TIFF. 
`gdal_translate -a_srs epsg:4326 -a_srs epsg:2393 cropped.tif to_delete.tif

For me it feels unlikely that lots of people have been trusting on the repeated -sql options. If they know SQL they do not 
select * from table where foo='bar';

jidanni:
Yes, true, the user should have noticed that there are no dots.

And they should also know about UNION.

Therefore I conclude that the documents have officially already done
their duty.

So the right thing is for somebody to please modify the program to
simply die with error upon getting more than one -sql option.

Just like it currently does for `;` in -sql.

PR #9398, opened 2024-03-06
SWIG: Exposes additional Dataset methods for the csharp wrapper (fixes #918)

### LINKED ISSUE #918: C# bindings are missing GDALDataset functions from GDAL/OGR unification

I expected the current version of C# bindings to have all the post-OGR-unification functions on GDALDataset — functions such as GetLayerCount, GetLayerByName, etc. Currently they do not. This class in the bindings only appears to have pre-2.0 functions. Without this, I am unable to get layers/features from datasets opened with GDALOpenEx.

1. Create a C# project and reference GDAL.NET 2.3.1
2. Use `OSGeo.GDAL.Gdal.OpenEx(...)` to open any vector dataset.
3. Realize that the returned `OSGeo.GDAL.Dataset` object provides no way to get to the layers/features in that dataset.

@szekerest's [GDAL.NET 2.3.1](https://www.nuget.org/packages/GDAL.NET/2.3.1)

andreasbaa:
SWIG: Exposes additional Dataset methods for the csharp wrapper:

- GetNextFeature (exposing _ppoBelongingLayer_ as a ref IntPtr and _pdfProgressPct_ as ref double),
- GetLayerCount, 
- GetLayer, 
- GetLayerByName, 
- ResetReading

### DIFF
--- a/swig/include/Dataset.i
+++ b/swig/include/Dataset.i
@@ -889,6 +889,38 @@ CPLErr AdviseRead(  int xoff, int yoff, int xsize, int ysize,
 
 #endif /* defined(SWIGPYTHON) || defined(SWIGJAVA) */
 
+#ifdef SWIGCSHARP
+  
+  %newobject GetNextFeature;
+  OGRFeatureShadow *GetNextFeature( OGRLayerShadow** ppoBelongingLayer = NULL,
+                                    double* pdfProgressPct = NULL,
+                                    GDALProgressFunc callback = NULL,
+                                    void* callback_data=NULL )
+  {
+    return GDALDatasetGetNextFeature( self, ppoBelongingLayer, pdfProgressPct,
+                                      callback, callback_data );
+  }
+
+  int GetLayerCount() {
+    return GDALDatasetGetLayerCount(self);
+  }
+
+  OGRLayerShadow *GetLayer(int index ) {
+    OGRLayerShadow* layer = (OGRLayerShadow*) GDALDatasetGetLayer(self, index);
+    return layer;
+  }
+
+  OGRLayerShadow *GetLayerByName(const char* layerName) {
+    OGRLayerShadow* layer = (OGRLayerShadow*) GDALDatasetGetLayerByName(self, layerName);
+    return layer;
+  }
+
+  void ResetReading()
+  {
+    GDALDatasetResetReading(self);
+  }
+
+#endif
 
 OGRErr AbortSQL() {
     return GDALDatasetAbortSQL(self);
--- a/swig/include/csharp/typemaps_csharp.i
+++ b/swig/include/csharp/typemaps_csharp.i
@@ -599,3 +599,19 @@ OPTIONAL_POD(int, int);
 %typemap(imtype) (void* callback_data) "string"
 %typemap(cstype) (void* callback_data) "string"
 %typemap(csin) (void* callback_data) "$csinput"
+
+
+/******************************************************************************
+ * GDALGetNextFeature typemaps                                                *
+ *****************************************************************************/
+
+%apply (double *defaultval) {double* pdfProgressPct};
+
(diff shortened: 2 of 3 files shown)


swig/include/Dataset.i
```diff
@@ -889,6 +889,38 @@ CPLErr AdviseRead(  int xoff, int yoff, int xsize, int ysize,
...
+  OGRLayerShadow *GetLayerByName(const char* layerName) {
+    OGRLayerShadow* layer = (OGRLayerShadow*) GDALDatasetGetLayerByName(self, layerName);
+    return layer;
+  }
+
+  void ResetReading()
+  {
+    GDALDatasetResetReading(self);
+  }
+
```
rouault, line 922: those already exist in ``#if defined(SWIGPYTHON) || defined(SWIGJAVA)`` branch at line 772. It would be better to move the existing definitions out of that branch, to have it in common for all bindings.  That would make the GetNextFeature() definition the only SWIGCSHARP specific one

andreasbaa: @rouault I agree. Updated the PR. Something like that?

swig/include/Dataset.i
```diff
@@ -890,6 +870,46 @@ CPLErr AdviseRead(  int xoff, int yoff, int xsize, int ysize,
...
+}
+
+void ResetReading()
+{
+  GDALDatasetResetReading(self);
+}
+
+int GetLayerCount() {
+  return GDALDatasetGetLayerCount(self);
+}
```
rouault, line 896:
```suggestion
  OGRLayerShadow *GetLayerByName(const char* layer_name) {
    OGRLayerShadow* layer = (OGRLayerShadow*) GDALDatasetGetLayerByName(self, layer_name);
    return layer;
  }

  void ResetReading()
  {
    GDALDatasetResetReading(self);
  }

  int GetLayerCount() {
    return GDALDatasetGetLayerCount(self);
  }
```

swig/include/Dataset.i
rouault: yes, looks good. I've adjusted indentation

swig/include/csharp/typemaps_csharp.i
```diff
@@ -599,3 +599,19 @@ OPTIONAL_POD(int, int);
...
+
+/******************************************************************************
+ * GDALGetNextFeature typemaps                                                *
+ *****************************************************************************/
+
+%apply (double *defaultval) {double* pdfProgressPct};
+
+%typemap(imtype) (OGRLayerShadow **ppoBelongingLayer)  "ref IntPtr"
+%typemap(cstype) (OGRLayerShadow **ppoBelongingLayer) "ref IntPtr"
+%typemap(csin) (OGRLayerShadow **ppoBelongingLayer)  "ref $csinput"
```
rouault, line 612: @andreasbaa I don't know C# at all, but is that sufficient to get a Layer instance from the CSharp method?

andreasbaa: @rouault Yes, I have done it like this in C#:
`var intPtr = new IntPtr();
double progress = 0;
var feature = ds.GetNextFeature(ref intPtr, ref progress, null, null);
var layer = new Layer(intPtr, false, null);
`

rouault: hum, that doesn't look very type-safe to me (it looks like you could potentially try to instance another GDAL object that would take a intPtr in its constructor). I guess that ideally we'd want to return a tuple (Feature, Layer), but I've no idea if that's doable in C#, and even less with SWIG. So this is probably a good enough solution for now

andreasbaa: I totally agree. I tried making SWIG return the Layer object directly, but did not manage to make swig translate the double pointer to a ref Layer. Hopefully someone can contribute to this in the future. Exposing the intpointer makes me at least able to utilize getnextfeature with the gmlas driver :) Big performance increase!

PR #9407, opened 2024-03-07
Update gdaltransform.rst removing invalid brackets

jidanni:


rouault:
they were intended. alternate fix in fdbbdcd939

rouault:
the fix in https://github.com/OSGeo/gdal/commit/956350a2980204ae65a5b637ea436560ffc13fa1 adds angle brackets in the synopsis, which is the appropriate fix to designate a place holder string that must be replaced by the actual value.  Angle brackets are meant for optional arguments

jidanni:
> the fix in [956350a](https://github.com/OSGeo/gdal/commit/956350a2980204ae65a5b637ea436560ffc13fa1) adds angle brackets in the synopsis, which is the appropriate fix to designate a place holder string that must be replaced by the actual value. Angle brackets are meant for optional arguments

@rouault I am very sorry to inform you that that is not the case,
- https://en.wikipedia.org/wiki/Usage_message#Pattern
- https://stackoverflow.com/questions/23242493/linux-unix-man-page-syntax-conventions/23242584#23242584
It must be late at night, as 99% of the man pages are OK, and it would take a lot of work to make it all backwards. Good thing I alerted you to the fact before further damage occurs. Thanks!

rouault:
> Angle brackets are meant for optional arguments

Ah, I mistyped. I meant square brackets. Fixed in my above comment

jidanni:
https://stackoverflow.com/questions/32085652/whats-the-meaning-of-in-git-command-synopsis indeed confirms the `[<>]` style. OK that indeed is more advanced than `[FILE]`.

PR #9420, opened 2024-03-07
Add geotiff as image_format option

cfis:
Add geotiff as supported image_format option to avoid warning message. This is foollow up to https://github.com/OSGeo/gdal/pull/9231


jratike80:
But are the tiles really always GeoTIFFs, or can they be just ordinary TIFFs?

PR #9421, opened 2024-03-07
gdaltransform/gdallocationinfo: make it output extra content at end of input line, add echo mode and field separator

### LINKED ISSUE #9411: Allow access to user input in output
### Feature description

Proj is great when it comes to returning the user's data,

```
echo 45.25 111.30 The Nurdsrtom Estate | proj +proj=utm +zone=12 -r

-2788964.34	14037151.76 The Nurdsrtom Estate
```

Not so with e.g., gdaltransform,

```
echo 300 -370 0 Silverstein Winery Ltd. | gdaltransform \
-gcp 0   -500 -111.89114803 40.75846686 \
-gcp 0   0    -111.89114717 40.76932606 \
-gcp 500 0    -111.87685039 40.76940631

-111.8825697384 40.761338402 0
```

So the user needs workarounds like,
```
gdaltransform -gcp ... input.csv | paste - input.csv | \
perl -anwle 'print join ",", @F[0,1,3];'
```

You might say it's like if when submitting GitHub bug reports, we need
to remember the order in which we submitted them, as when they get
answered they lack the original subject and bug number.

Yes, true, often we don't want "grandma to quote our entire email when
responding", especially for single queries, when we are very aware of
what we typed.

Not so for a whole .csv file of queries. In this case we need to keep
track: the 35th question got the 35th answer. I.e., we need to paste
line 35 of the input file to line 35 of the output file. Not that
hard, as we saw above, but still a hassle.

The most frustrating is gdallocationinfo -valonly.

It returns `Band1\nBand2\nBand1\nBand2\n...` Not only is there no
--option to get back what one typed in, but also the band separator
and record separator are the same. So we need hacks like:

```
gdallocationinfo -valonly -geoloc input|\
perl -pwe 's/\n/ / if $. % 2;'|paste -d ' ' - ...

gdallocationinfo -valonly c.tif|\
perl -nwle '$a[$. % 2]=$_; print join " ", reverse @a unless $. % 2;'
```
True, there is `gdallocationinfo -xml`, but that also won't let the
user have any of his original input back.

So, each program should have some option added, that will in one form or the other, pass back what the user inputted, when giving results.

Going back to the Proj example. One can see the authors were on their toes. If the user didn't want any extra stuff tacked on the answer, then he simply could snip it off before passing it in for evaluation.

That's smarter then just sending it to /dev/null no matter what. I mean even in high security situations, companies still say "regarding your account ***5678", not just "regarding your account" (which one?)

Proj even has
> -E  Causes  the  input  coordinates  to be copied to the output line prior to printing the converted values.

rouault:
Fixes #9411

- gdaltransform: make it output extra content at end of input line (unless -ignore_extra_input), and add -E echo mode, and -field_sep option
- gdallocationinfo: make it output extra content at end of input line (unless -ignore_extra_input), and add -E echo mode, and -field_sep option

### DIFF
--- a/apps/gdallocationinfo.cpp
+++ b/apps/gdallocationinfo.cpp
@@ -35,6 +35,8 @@
 #include "ogr_spatialref.h"
 #include <vector>
 
+#include <cctype>
+
 #ifdef _WIN32
 #include <io.h>
 #else
@@ -52,6 +54,8 @@ static void Usage(bool bIsError)
         bIsError ? stderr : stdout,
         "Usage: gdallocationinfo [--help] [--help-general]\n"
         "                        [-xml] [-lifonly] [-valonly]\n"
+        "                        [-E] [-field_sep <sep>] "
+        "[-ignore_extra_input]\n"
         "                        [-b <band>]... [-overview <overview_level>]\n"
         "                        [-l_srs <srs_def>] [-geoloc] [-wgs84]\n"
         "                        [-oo <NAME>=<VALUE>]... <srcfile> [<x> <y>]\n"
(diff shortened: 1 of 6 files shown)


jidanni:
Actually a ton of the other programs also need the fix.

rouault:
> a ton of the other programs

which ones?

jidanni:
> which ones?

Err... OK, I'll be sure to report them if I find them :-)

PR #9422, opened 2024-03-07
gdaldriver: Don't Lose papszOpenOptions

cfis:
When using gdal_translate, the source datasource is opened twice. However, the second time does not use any open options specified when orginally opening the data soruce. This is a probelm for OGC Tile data sources if the user specified TILEMATRIXSET or TILEMATRIX because the driver will default to WorldCRS84Quad even when that is not what was originally specified or to a different zoom level.

For example,  if you request TILEMATRIXSET="foo" then gdal open will the correct tilematrixset definition on the first open, but on the second open it will instead load the definition for the WorldCRS84Quad tilematrixset (assuming it is a listed tilematrixset).

This PR fixes the issue by passing `poSrcDS->papszOpenOptions` to the newly opened datasource.

### DIFF
--- a/gcore/gdaldriver.cpp
+++ b/gcore/gdaldriver.cpp
@@ -1049,7 +1049,8 @@ CPLErr GDALDriver::QuietDeleteForCreateCopy(const char *pszFilename,
                                      : nullptr,
                 nullptr};
             auto poSrcDSTmp = std::unique_ptr<GDALDataset>(GDALDataset::Open(
-                poSrcDS->GetDescription(), GDAL_OF_RASTER, apszAllowedDrivers));
+                poSrcDS->GetDescription(), GDAL_OF_RASTER, apszAllowedDrivers,
+                poSrcDS->papszOpenOptions));
             if (poSrcDSTmp)
             {
                 char **papszFileList = poSrcDSTmp->GetFileList();


rouault:
@cfis I'm a bit confused by your PR description compared to its content. What you modify in the PR is the GDALDriver::QuietDeleteForCreateCopy() method, so this would only make sense if you tried to overwrite a OGCAPI dataset, which doesn't make sense as it is a non-file based read-only driver.

cfis:
Ah sorry. Here is an example. When I run:

```
 ./gdal_translate -oo "CACHE=NO" -oo "API=tiles" -oo "IMAGE_FORMAT=GEOTIFF" -oo "TILEMATRIXSET=C1232_180_180" -oo "TILEMATRIX=0" OGCAPI:https://tiles.qa3.orbitalinsight.io/ogcapi/collections/1050010019917900---s---WV02---wv02---NA---NA out.tiff
```
The goal is copy the remote tiles and merge them together into a local tiff file. Note that `C1232_180_180` is a custom tilematrixset. We also support `WorldCRS84Quad`.

If you look at the curl log:

```
GET /ogcapi/collections/1050010019917900---s---WV02---wv02---NA---NA
GET /ogcapi/collections/1050010019917900---s---WV02---wv02---NA---NA/map/tiles
GET /ogcapi/collections/1050010019917900---s---WV02---wv02---NA---NA/map/tiles/C1232_180_180
GET /ogcapi/collections/1050010019917900---s---WV02---wv02---NA---NA/orbitalTiles/C1232_180_180 
GET /ogcapi/collections/1050010019917900---s---WV02---wv02---NA---NA/map/tiles/C1232_180_180/0/0/0.tif

GET /ogcapi/collections/1050010019917900---s---WV02---wv02---NA---NA HTTP/1.1
GET /ogcapi/collections/1050010019917900---s---WV02---wv02---NA---NA/map/tiles HTTP/1.1
GET /ogcapi/collections/1050010019917900---s---WV02---wv02---NA---NA/map/tiles/WebMercatorQuad
GET /ogcapi/tileMatrixSets/WebMercatorQuad

GET /ogcapi/collections/1050010019917900---s---WV02---wv02---NA---NA/map/tiles/C1232_180_180/0/9462/10413.tif 
etc....
```

Notice the two incorrect calls:

```
GET /ogcapi/collections/1050010019917900---s---WV02---wv02---NA---NA/map/tiles/WebMercatorQuad
GET /ogcapi/tileMatrixSets/WebMercatorQuad
```
Those are being generated from the code I updated.

rouault:
ah, sorry I've missed that QuietDeleteForCreateCopy() was actually called with the source dataset. So your fix is correct

PR #9428, opened 2024-03-08
GDAL: fix infinite loop when resampling

### LINKED ISSUE #9427: GTiff: endless loop when resampling (bilinear) under some circumstances
### What is the bug?

Discovered downstream in https://github.com/qgis/QGIS/issues/56769

Preliminary test case to reproduce (test.tif file is from downstream QGIS issue mentioned above).

```bash
$ gdalinfo test.tif 
Driver: GTiff/GeoTIFF
Files: test.tif
Size is 980, 1012
Coordinate System is:
PROJCRS["ETRS89 / UTM zone 32N",
    BASEGEOGCRS["ETRS89",
        ENSEMBLE["European Terrestrial Reference System 1989 ensemble",
            MEMBER["European Terrestrial Reference Frame 1989"],
            MEMBER["European Terrestrial Reference Frame 1990"],
            MEMBER["European Terrestrial Reference Frame 1991"],
            MEMBER["European Terrestrial Reference Frame 1992"],
            MEMBER["European Terrestrial Reference Frame 1993"],
            MEMBER["European Terrestrial Reference Frame 1994"],
            MEMBER["European Terrestrial Reference Frame 1996"],
            MEMBER["European Terrestrial Reference Frame 1997"],
            MEMBER["European Terrestrial Reference Frame 2000"],
            MEMBER["European Terrestrial Reference Frame 2005"],
            MEMBER["European Terrestrial Reference Frame 2014"],
            ELLIPSOID["GRS 1980",6378137,298.257222101,
                LENGTHUNIT["metre",1]],
            ENSEMBLEACCURACY[0.1]],
        PRIMEM["Greenwich",0,
        ID["EPSG",4258]],
    CONVERSION["UTM zone 32N",
        METHOD["Transverse Mercator",
            ID["EPSG",9807]],
        PARAMETER["Latitude of natural origin",0,
            ANGLEUNIT["degree",0.0174532925199433],
            ID["EPSG",8801]],
        PARAMETER["Longitude of natural origin",9,
            ANGLEUNIT["degree",0.0174532925199433],
            ID["EPSG",8802]],
        PARAMETER["Scale factor at natural origin",0.9996,
            SCALEUNIT["unity",1],
            ID["EPSG",8805]],
        PARAMETER["False easting",500000,
            LENGTHUNIT["metre",1],
            ID["EPSG",8806]],
        PARAMETER["False northing",0,
            LENGTHUNIT["metre",1],
            ID["EPSG",8807]]],
    CS[Cartesian,2],
        AXIS["(E)",east,
            ORDER[1],
            LENGTHUNIT["metre",1]],
        AXIS["(N)",north,
            ORDER[2],
            LENGTHUNIT["metre",1]],
    USAGE[
        SCOPE["Engineering survey, topographic mapping."],
        AREA["Europe between 6°E and 12°E: Austria; Belgium; Denmark - onshore and offshore; Germany - onshore and offshore; Norway including - onshore and offshore; Spain - offshore."],
        BBOX[38.76,6,84.33,12]],
    ID["EPSG",25832]]
Origin = (305000.500000000000000,5681012.500000000000000)
Pixel Size = (1.000000000000000,-1.000000000000000)
Metadata:
  AREA_OR_POINT=Area
Image Structure Metadata:
  INTERLEAVE=BAND
Corner Coordinates:
Upper Left  (  305000.500, 5681012.500) (  6d12'21.32"E, 51d14'49.54"N)
Lower Left  (  305000.500, 5680000.500) (  6d12'23.30"E, 51d14'16.82"N)
Upper Right (  305980.500, 5681012.500) (  6d13'11.80"E, 51d14'50.74"N)
Lower Right (  305980.500, 5680000.500) (  6d13'13.78"E, 51d14'18.02"N)
Center      (  305490.500, 5680506.500) (  6d12'47.55"E, 51d14'33.78"N)
Band 1 Block=980x2 Type=UInt32, ColorInterp=Gray
  NoData Value=-3.4028234663852886e+38
  Metadata:
    RepresentationType=ATHEMATIC
```

```python
from osgeo import gdal, ogr, osr


def main():

    ds = gdal.Open('test.tif')

    # Get raster band
    band = ds.GetRasterBand(1)

    scanline = band.ReadRaster1(xoff=0, yoff=0,
                        xsize=980, ysize=1012,
                        buf_xsize=956, buf_ysize=988,
                        buf_pixel_space=4, buf_line_space=956*4,
                        buf_type=gdal.GDT_UInt32,
                        resample_alg=gdal.GRIORA_Bilinear)

    print(scanline)

if __name__ == '__main__':
    main()
```

### Versions and provenance

verified in 3.9, will test master soon

### Additional context

_No response_

elpaso:
Fix #9427


gcore/gdalrasterband.cpp
```diff
@@ -1929,6 +1929,128 @@ uint64_t GDALRasterBand::GetNoDataValueAsUInt64(int *pbSuccess)
...
+        }
+        else if (fNoDataValue == -std::numeric_limits<float>::infinity())
+        {
+            dfNoDataValue = std::numeric_limits<float>::lowest();
+        }
+
+        if (dfNoDataValue == std::numeric_limits<float>::max())
+        {
+            dfReplacementVal = std::nextafter(
+                dfNoDataValue, std::numeric_limits<float>::lowest());
```
rouault, line 2026: I woud rather take the nextbefore() of max(), as we want to minimize the gap between dfNoDataValue and dfReplacementVal

gcore/gdalrasterband.cpp
```diff
@@ -1929,6 +1929,128 @@ uint64_t GDALRasterBand::GetNoDataValueAsUInt64(int *pbSuccess)
...
+        }
+    }
+    else if (dt == GDT_Float64)
+    {
+        dfNoDataValue = GDALClampDoubleValue(
+            dfNoDataValue, std::numeric_limits<double>::lowest(),
+            std::numeric_limits<double>::max());
+        if (dfNoDataValue == std::numeric_limits<double>::max())
+        {
+            dfReplacementVal = std::nextafter(
```
rouault, line 2041: same remark as for Float32

rouault (commented):
-

gcore/gdalrasterband.cpp
```diff
@@ -1929,6 +1929,128 @@ uint64_t GDALRasterBand::GetNoDataValueAsUInt64(int *pbSuccess)
...
+
+        if (dfNoDataValue == std::numeric_limits<float>::max())
+        {
+            dfReplacementVal = std::nextafter(
+                dfNoDataValue, std::numeric_limits<float>::lowest());
+        }
+        else
+        {
+            dfReplacementVal = std::nextafter(
+                dfNoDataValue, std::numeric_limits<float>::max());
```
rouault, line 2031: we probably need an explicit cast to float, to be sure that the float version is used
```suggestion
            dfReplacementVal = std::nextafter(
                static_cast<float>(dfNoDataValue), std::numeric_limits<float>::max());
```

autotest/cpp/test_gdal.cpp
```diff
@@ -2462,6 +2462,61 @@ TEST_F(test_gdal, GDALBufferHasOnlyNoData)
...
+              std::numeric_limits<int32_t>::max() - 1);
+
+    // Test floating point types
+
+    EXPECT_EQ(GDALGetRasterNoDataReplacementValue(
+                  GDT_Float32, std::numeric_limits<double>::lowest()),
+              std::nextafter(std::numeric_limits<float>::lowest(), 0.0));
+    EXPECT_EQ(GDALGetRasterNoDataReplacementValue(
+                  GDT_Float32, std::numeric_limits<double>::max()),
+              std::nextafter(std::numeric_limits<float>::max(), 0.0));
```
rouault, line 2510: the use of double for those float tests is suspicous. We should have at least a nominal test with std::numeric_limits<float>::lowest() and max()

rouault: ah, I realize the test is written like that because of the use of GDALClampDoubleValue().

gcore/gdalrasterband.cpp
```diff
@@ -1929,6 +1929,128 @@ uint64_t GDALRasterBand::GetNoDataValueAsUInt64(int *pbSuccess)
...
+        if (fNoDataValue == std::numeric_limits<float>::infinity())
+        {
+            dfNoDataValue = std::numeric_limits<float>::max();
+        }
+        else if (fNoDataValue == -std::numeric_limits<float>::infinity())
+        {
+            dfNoDataValue = std::numeric_limits<float>::lowest();
+        }
+
+        if (dfNoDataValue == std::numeric_limits<float>::max())
```
rouault, line 2023: Perhaps?
```suggestion
        if (fNoDataValue == std::numeric_limits<float>::max())
```

gcore/gdalrasterband.cpp
```diff
@@ -1929,6 +1929,128 @@ uint64_t GDALRasterBand::GetNoDataValueAsUInt64(int *pbSuccess)
...
+ * @since GDAL 3.10
+ */
+double GDALGetRasterNoDataReplacementValue(GDALDataType dt,
+                                           double dfNoDataValue)
+{
+
+    double dfReplacementVal = dfNoDataValue;
+    if (dt == GDT_Byte)
+    {
+        dfNoDataValue = GDALClampDoubleValue(
```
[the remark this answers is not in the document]
elpaso: yes, but my thinking was that the computational cost of an additional if, compared to an assignment was more or less the same, what I can do here is to make GDALClampDoubleValue modify dfNoDataValue and return TRUE if it was changed, and return 0 straight away if not.

gcore/gdalrasterband.cpp
elpaso: > I woud rather take the nextbefore() of max(), as we want to minimize the gap between dfNoDataValue and dfReplacementVal

To my knowledge nextbefore does not exist, nextafter does the job: https://cpp.sh/?source=%2F%2F+Example+program%0A%23include+%3Ciostream%3E%0A%23include+%3Cstring%3E%0A%23include+%3Cmemory%3E%0A%0Astruct+A%0A%7B%0A++++int+foo+%3D+1%3B%0A%7D%3B%0A%0Astruct+B%0A%7B%0A++++B(std%3A%3Aunique_ptr%3CA%3E+%26%26_a)%3A%0A+++++a(std%3A%3Amove(_a))+%7B%0A+++++++++%0A++++std%3A%3Acout+%3C%3C+%22a%3A+%22+%3C%3C+a+%3C%3C+std%3A%3Aendl%3B%0A++++std%3A%3Acout+%3C%3C+%22_a%3A+%22+%3C%3C+_a+%3C%3C+std%3A%3Aendl%3B%0A%7D%3B%0A+++++%0A++++std%3A%3Aunique_ptr%3CA%3E+a%3B%0A%7D%3B%0A%0Aint+main()%0A%7B%0A%0A++++std%3A%3Aunique_ptr%3CA%3E+a+%7B+new+A(+)+%7D%3B%0A++++a-%3Efoo+%3D+2%3B%0A++++%0A++++std%3A%3Acout+%3C%3C+%22a%3A+%22+%3C%3C+a+%3C%3C+std%3A%3Aendl%3B%0A++++%0A++++B+b%7Bstd%3A%3Amove(a)%7D%3B%0A++++%0A++++std%3A%3Acout+%3C%3C+%22a%3A+%22+%3C%3C+a+%3C%3C+std%3A%3Aendl%3B%0A%7D

What I will do is replace max/lowest as the second argument with 0 when appropriate to make it more readable.

gcore/gdalrasterband.cpp
elpaso: good catch!

gcore/gdalrasterband.cpp
```diff
@@ -1929,6 +1929,128 @@ uint64_t GDALRasterBand::GetNoDataValueAsUInt64(int *pbSuccess)
...
+        const float fNoDataValue{static_cast<float>(dfNoDataValue)};
+
+        if (fNoDataValue == std::numeric_limits<float>::infinity())
+        {
+            dfNoDataValue = std::numeric_limits<float>::max();
+        }
+        else if (fNoDataValue == -std::numeric_limits<float>::infinity())
+        {
+            dfNoDataValue = std::numeric_limits<float>::lowest();
+        }
```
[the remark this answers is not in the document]
elpaso: Why? The infinity in case of float is the possible result of the down cast but in this case there is no cast that could generate and infinity.

gcore/gdalrasterband.cpp
elpaso: No, I would say the check is logically correct but you made me think that it might be safer to change the comparison to
```suggestion
        if (dfNoDataValue >= std::numeric_limits<float>::max())
```
even if to my understanding the automatic promotion of `std::numeric_limits<float>::max()` to `double` should make it irrelevant.

I will add some comments to explain what was my intention with the code flow.

elpaso:
@rouault I went down into the rabbit hole of int64/uint64 max() not representable as double but I'm not entirely sure how this is handled by client code.

gcore/gdalrasterband.cpp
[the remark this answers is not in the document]
rouault: regarding computation cost, normally GetRasterNoDataReplacementValue shouldn't not be used inside a performance critical loop. The typical use pattern is to call it outside of the loop iterating over pixels.
Maybe the clamping works. With my mind set, it seems to be simpler to have this logic: "test if the provided nodata is in the validity range of the indicated data type. If it is not, exit early and return a 'random'=0 replace nodata value that is never going to be used anyway since no computed pixel will be equal to the nodata value since it is out of range"

gcore/gdalrasterband.cpp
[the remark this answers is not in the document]
rouault: > The infinity in case of float is the possible result of the narrowing cast

not necessarily. A Float32 dataset could have a +inf nodata value (so not just because of a narrowing cast. And if we implement an early exit when the input dfNodataValue is out of range of float32, but maybe still allowing nan,inf,-inf, the narrowing cast shouldn't cause issues), and here your point was apparently to use a replacement value of <float>::max() if a computation would happen. I don't really think that in practice the result of a correct computation would lead to +inf, so my intuition would be that this particular case is not needed. But if it is needed for Float32, for the same reason, it should be needed for Float32.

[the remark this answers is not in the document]
rouault: (sorry, my above comment was written before seeing your latest change. code review is hard :-))  Ok, so now we exit early if dfNoDataValue is +/- inf. So effectively the tests at line 2089-2090 could only trigger because of a narrowing cast, but it would seem weird to me that when dfNoDataValue <= std::numeric_limits<float>::max(),  static_cast<float>(dfNoDataValue) could be equal to std::numeric_limits<float>::infinity()

rouault:
> I went down into the rabbit hole of int64/uint64 max() not representable as double but I'm not entirely sure how this is handled by client code.

It would be prudent to indicate in the documentation of the function that it cannot be reliably trusted with UInt64/Int64 because their nodata values might not always be representable exactly as a double. I believe the function can only be trusted for UInt64/Int64 when abs(dfNoDataValue) < 2^53 (cf https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER)

autotest/cpp/test_gdal.cpp
```diff
@@ -2462,6 +2462,155 @@ TEST_F(test_gdal, GDALBufferHasOnlyNoData)
...
+    EXPECT_EQ(GDALGetRasterNoDataReplacementValue(
+                  GDT_Float32, std::numeric_limits<double>::max()),
+              0.0);
+    EXPECT_EQ(GDALGetRasterNoDataReplacementValue(
+                  GDT_Float32, std::numeric_limits<double>::infinity()),
+              0.0);
+    EXPECT_EQ(GDALGetRasterNoDataReplacementValue(
+                  GDT_Float32, -std::numeric_limits<double>::infinity()),
+              0.0);
+    // in range for float 32
```
rouault, line 2584: we'd need a test for in-range ``std::numeric_limits<float>::max()`` and ``std::numeric_limits<double>::lowest()``

gcore/gdalrasterband.cpp
```diff
@@ -1929,6 +1929,208 @@ uint64_t GDALRasterBand::GetNoDataValueAsUInt64(int *pbSuccess)
...
+ *        is out of range for the specified data type (dt).
+ *        For UInt64 and Int64 data type this function cannot reliably trusted
+ *        because their nodata values might not always be representable exactly
+ *        as a double, in particular the maximum absolute value for those types
+ *        is 2^53.
+ *
+ * @param dt Data type
+ * @param dfNoDataValue The no data value
+
+ * @since GDAL 3.10
```
rouault, line 1943:
```suggestion
 * @since GDAL 3.9
```

gcore/gdalrasterband.cpp
```diff
@@ -1929,6 +1929,208 @@ uint64_t GDALRasterBand::GetNoDataValueAsUInt64(int *pbSuccess)
...
 }
 
+/**
+ * \brief Returns a replacement value for a nodata value or 0 if dfNoDataValue
+ *        is out of range for the specified data type (dt).
+ *        For UInt64 and Int64 data type this function cannot reliably trusted
+ *        because their nodata values might not always be representable exactly
+ *        as a double, in particular the maximum absolute value for those types
+ *        is 2^53.
+ *
```
rouault, line 1939:
```suggestion
 *
 * The replacement value is a value that can be used in a computation
 * whose result would match by accident the nodata value, whereas it is
 * meant to be valid. For example, for a dataset with a nodata value of 0,
 * when averaging -1 and 1, one would get normally a value of 0. The
 * replacement nodata value can then be substitued to that 0 value to still
 * get a valid value, as close as practical to the true value, while being
 * different from the nodata value.
 *
```

autotest/cpp/test_gdal.cpp
```diff
@@ -2581,13 +2581,22 @@ TEST_F(test_gdal, GetRasterNoDataReplacementValue)
...
               0.0);
+
     // in range for float 32
     EXPECT_EQ(static_cast<float>(
                   GDALGetRasterNoDataReplacementValue(GDT_Float32, -1.0)),
               std::nextafter(float(-1.0), 0.0f));
     EXPECT_EQ(static_cast<float>(
                   GDALGetRasterNoDataReplacementValue(GDT_Float32, 1.1)),
               std::nextafter(float(1.1), 2.0f));
+    EXPECT_EQ(GDALGetRasterNoDataReplacementValue(
```
rouault, line 2592: you perhaps need to static_cast<float> the return of  GDALGetRasterNoDataReplacementValue?

rouault: ah, ignore my above remark. Seems you've addressed it already

gcore/gdal.h
```diff
@@ -1521,6 +1521,8 @@ CPLErr CPL_DLL CPL_STDCALL GDALSetRasterNoDataValueAsInt64(GDALRasterBandH,
 CPLErr CPL_DLL CPL_STDCALL GDALSetRasterNoDataValueAsUInt64(GDALRasterBandH,
                                                             uint64_t);
 CPLErr CPL_DLL CPL_STDCALL GDALDeleteRasterNoDataValue(GDALRasterBandH);
+double CPL_DLL CPL_STDCALL GDALGetRasterNoDataReplacementValue(GDALDataType,
+                                                               double);
```
rouault, line 1525:
```suggestion
double CPL_DLL GDALGetRasterNoDataReplacementValue(GDALDataType, double);
```

rouault:
@elpaso i've pushed a few extra commits in your branch for final tuning.  I've removed the "Raster" in the function name, because the convention (at least as I decypher it... wasn't born at the time it was forged :-)) is that it should be used for functions that take a GDALRasterBandH as first argument (although there are a few ones actually taking a GDALDatasetH...)

elpaso:
> @elpaso i've pushed a few extra commits in your branch for final tuning. I've removed the "Raster" in the function name, because the convention (at least as I decypher it... wasn't born at the time it was forged :-)) is that it should be used for functions that take a GDALRasterBandH as first argument (although there are a few ones actually taking a GDALDatasetH...)

thanks, I see you have also fixed the windows warning.

PR #9440, opened 2024-03-10
gdal_viewshed: add support for south-up source datasets (fixes #9432)

### LINKED ISSUE #9432: gdal_viewshed: Attempt to create 5x-1 dataset
### What is the bug?

User simply follows example on man page, and is accused of trying to create negative sized datasets.

```
$ gdal_viewshed -md 15 -ox 236850 -oy 2675610 m.tif xdg.tif
$ gdal_viewshed -md 20 -ox 236850 -oy 2675610 m.tif xdg.tif
0ERROR 1: Attempt to create 3x-1 dataset is illegal,sizes must be larger than zero.
ERROR 1: Cannot create dataset for xdg.tif
$ gdal_viewshed -md 33 -ox 236850 -oy 2675610 m.tif xdg.tif
0ERROR 1: Attempt to create 5x-1 dataset is illegal,sizes must be larger than zero.
ERROR 1: Cannot create dataset for xdg.tif

```

### Versions and provenance

Fresh Debian sid gdal.
[m.tif.gz](https://github.com/OSGeo/gdal/files/14546646/m.tif.gz)

### Additional context

Hmm, a mysterious wish for a negative sized dataset?

The user is simply following the (-md 500) example in the gdal_viewshed man page.

Perhaps the user does not understand the -md units.

Looking at the man page, these all mention units,
```
       -ox <value>
              The X position of the observer (in SRS units).

       -oy <value>
              The Y position of the observer (in SRS units).

       -oz <value>
              The height of the observer above the DEM surface in  the  height
              unit of the DEM. Default: 2

       -tz <value>
              The  height  of  the  target above the DEM surface in the height
              unit of the DEM. Default: 0
```
but not this
```
       -md <value>
              Maximum distance from observer to compute visibiliy.  It is also
              used to clamp the extent of the output raster.
```
So it should mention the units...

I conclude that I did not in fact request a negative sized output, and it must be a bug in the program. But I'll append the input .tif here so people can check. Had to zip it.

rouault:


alg/viewshed.cpp
```diff
@@ -282,29 +282,39 @@ GDALDatasetH GDALViewshedGenerate(
     }
 
     /* calculate the area of interest */
+    constexpr double EPSILON = 1e-8;
     int nXStart =
         dfMaxDistance > 0
-            ? (std::max)(0, static_cast<int>(std::floor(
-                                nX - adfInvGeoTransform[1] * dfMaxDistance)))
+            ? (std::max)(
```
elpaso, line 288: just curios why the `(std::max)` instead of `std::max` ?

elpaso (approved):

alg/viewshed.cpp
rouault: I've added a commit to remove those parentheses

hobu: > just curios why the `(std::max)` instead of `std::max` ?

Given include file ordering scenarios, it is my experience that we don't always get the behavior we hope for with NOMINMAX. For PDAL, especially when using `std::max` and `std::min` in header files, we always enclose the functions so we are never beholden to some external definition configuration of the NOMINMAX macro.

rouault: > Given include file ordering scenarios, it is my experience that we don't always get the behavior we hope for with NOMINMAX

``add_definitions(-DNOMINMAX)`` does the job nicely, at least for .cpp compilation units. But your point may be valid for public headers where you can't control easily that. As far as I can see, we only have one public header "gdalcachedpixelaccessor.h" with std::min use in it (but it is rather esoteric, so I don't think anyone outside of GDAL use it)

PR #9456, opened 2024-03-14
JPEG: ReadXMPMetadata(): only stop looking for XMP marker at Start Of Scan

rouault:
Some images have XMP after the quantization/Huffman table markers


elpaso:
Is this worth a test?

rouault:
> Is this worth a test?

I had to use an hexadecimal editor to move segments around, but I managed it :-) (and it made me realize that I didn't initially test the correct marker, although that had no practical consequence). Commit amended with correct marker test and new test case

PR #9466, opened 2024-03-15
doc: update the ogr2ogr default value for -gt in SQLite performance hints

ctoney:
This pull request updates the stated default value of the `ogr2ogr -gt` argument in the "Performance hints" section for SQLite/Spatialite. It updates to the default value 100000 as given in the `ogr2ogr` doc at:
https://gdal.org/programs/ogr2ogr.html#cmdoption-ogr2ogr-gt.

It also removes the recommendation
>Increasing to -gt 65536 or more ensures optimal performance while populating some table containing many hundredth thousand or million rows.

Should there be a recommendation to increase potentially >100000 in some cases?


rouault:
> Should there be a recommendation to increase potentially >100000 in some cases?

from my memories, the potential gains in increasing the -gt value further aren't that dramatic, so this is probably fine like that. Thanks

PR #9486, opened 2024-03-17
Doc: Update Python API docs for ogr.Feature

dbaston:
- Alphabetize
- Remove C signatures and parameters
- Add missing docstrings
- Remove docstrings for methods not exposed in Python
- Add cross-references and some examples

### DIFF
--- a/swig/include/python/docs/ogr_feature_docs.i
+++ b/swig/include/python/docs/ogr_feature_docs.i
@@ -233,7 +233,7 @@ Examples
 
 Fetch :py:const:`OFTDateTime` field value as a ISO8601 representation.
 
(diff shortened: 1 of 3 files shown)


swig/include/python/docs/ogr_feature_docs.i
```diff
@@ -1,1712 +1,1039 @@
...
 Parameters
 -----------
-hFeat:
-    handle to the feature to set to.
-hOtherFeat:
+other : Feature
     handle to the feature from which geometry, and field
     values will be copied.
-panMap:
+forgiving : bool, default = True
```
rouault, line 873: No default value in SetFromWithMap() constrary to SetFrom():

```suggestion
forgiving : bool
```

PR #9487, opened 2024-03-17
Python bindings: Invalidate layer ref when Dataset closes

dbaston:
## What does this PR do?

Prevents a crash when using a `Layer` after its parent `Dataset` closes. We already handle this for the `DataSource` - `Layer` relationship and the `Dataset` - `Band` relationship, but `Dataset` - `Layer` was missed.

#7965

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/swig/include/python/gdal_python.i
+++ b/swig/include/python/gdal_python.i
@@ -744,12 +744,12 @@ void wrapper_VSIGetMemFileBuffer(const char *utf8_path, GByte **out, vsi_l_offse
 
 %feature("pythonappend") GetMaskBand %{
     if hasattr(self, '_parent_ds') and self._parent_ds():
-        self._parent_ds()._add_band_ref(val)
+        self._parent_ds()._add_child_ref(val)
 %}
 
 %feature("pythonappend") GetOverview %{
     if hasattr(self, '_parent_ds') and self._parent_ds():
-        self._parent_ds()._add_band_ref(val)
+        self._parent_ds()._add_child_ref(val)
 %}
 
 %feature("shadow") ComputeStatistics %{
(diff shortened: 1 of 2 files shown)


swig/include/python/gdal_python.i
```diff
@@ -1580,7 +1580,23 @@ def ReleaseResultSet(self, sql_lyr):
...
+
+%feature("pythonappend") GetLayerByName %{
+    self._add_child_ref(val)
+%}
+
+%feature("pythonappend") GetLayerByIndex %{
+    self._add_child_ref(val)
+%}
+
+%feature("pythonappend") CreateLayer %{
```
rouault, line 1594: In theory, we should likely also instrument DeleteLayer() to make sure it removes the reference to avoid the array of reference to grow without control if someone would do an infinite loop of CreateLayer() + DeleteLayer(). But that might require an auxiliary data structure to map the layer index to its object. That could be for a follow-up pull request as I don't anticipate this scenario to be a nominal use case.

dbaston: Wouldn't the `WeakSet` that we are using for the container take care of this for us?

dbaston: Indeed:

```
ds = gdal.GetDriverByName("GPKG").Create("/tmp/test.gpkg", 0, 0, gdal.GDT_Unknown)
for i in range(100):
    lyr_name = f"ds_{i}"
    lyr = ds.CreateLayer(lyr_name)
    ds.DeleteLayer(lyr_name)

assert len(ds._child_references) == 1
```

rouault: oh, nice!

PR #9490, opened 2024-03-17
CSLConstList/CPLStringList improvements

rouault:
- CSLConstList/CPLStringList: add iterating facilities and better std::vector<std::string> operability

  * Add CPLStringList(const std::vector<std::string> &aosList) constructor
  * Add CPLStringList(std::initializer_list<const char *> oInitList) constructor
  * Add static CPLStringList BoundToConstList(CSLConstList) constructor-like
    (for cheap/non-copying instanciation of a CPLStringList from a CSLConstList)
  * Add CPLStringList::operator std::vector<std::string>(void)
  * Add cpl::Iterate(CSLConstList) which wraps a CSLConstList in a structure that
    can be used with C++ iterators.
  * Add cpl::IterateNameValue(CSLConstList) which wraps a CSLConstList in a
    structure that can be used with C++ iterators to get (name, value) pairs.
    like in ``for (const auto& [name, value]: cpl::IterateNameValue(papszList)) {}``
  * Add std::vector<std::string> cpl::ToVector(CSLConstList)

- gcore: use new CSLConstList/CPLStringList facilities

- apps: use new CSLConstList/CPLStringList facilities

### DIFF
--- a/port/cpl_string.h
+++ b/port/cpl_string.h
@@ -288,6 +288,7 @@ extern "C++"
 {
 #ifndef DOXYGEN_SKIP
 #include <string>
+#include <vector>
 #endif
 
 // VC++ implicitly applies __declspec(dllexport) to template base classes
@@ -452,9 +453,13 @@ extern "C++"
         CPLStringList();
         explicit CPLStringList(char **papszList, int bTakeOwnership = TRUE);
         explicit CPLStringList(CSLConstList papszList);
+        explicit CPLStringList(const std::vector<std::string> &aosList);
+        explicit CPLStringList(std::initializer_list<const char *> oInitList);
         CPLStringList(const CPLStringList &oOther);
         CPLStringList(CPLStringList &&oOther);
         ~CPLStringList();
+
+        static const CPLStringList BoundToConstList(CSLConstList papszList);
 
         CPLStringList &Clear();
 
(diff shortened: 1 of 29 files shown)


port/cpl_string.h
```diff
@@ -614,6 +646,191 @@ extern "C++"
...
+        using reference = value_type &;
+
+        CSLConstList m_papszList = nullptr;
+        bool m_bAtEnd = false;
+
+        inline const char *operator*() const
+        {
+            return *m_papszList;
+        }
+        inline CSLIterator &operator++()
```
elpaso, line 669: Should we add a clang-tidy rule to add a line break between methods ?
maybe  (untested):
SeparateDefinitionBlocks: Always
EmptyLineBeforeAccessModifier: LogicalBlock

port/cplstringlist.cpp
```diff
@@ -107,6 +156,27 @@ CPLStringList::CPLStringList(CPLStringList &&oOther) : CPLStringList()
...
+/*                           BoundToConstList()                         */
+/************************************************************************/
+
+/**
+ * Return a CPLStringList that wraps the passed list.
+ *
+ * The input list is *NOT* copied and must be kept alive while the
+ * return CPLStringList is used.
+ *
+ * @param papszListIn a NULL terminated list of strings to wrap into the CPLStringList
```
elpaso, line 169:
```suggestion
 * @param papszListIn a list of NULL terminated strings to wrap into the CPLStringList
```

elpaso (approved):
I only looked at the `port` changes and they looks good to me.

port/cpl_string.h
rouault: > Should we add a clang-tidy rule to add a line break between methods ?

yes, sounds like a good idea

port/cplstringlist.cpp
rouault: the original phrasing is actually correct. The list is terminated by a NULL pointer. This is a phrasing we use in a lot of places. Each string is assumed to be terminated by the NUL character, but that's kind of obvious to any user of a C API.

PR #9499, opened 2024-03-18
BAG/S102: make sure to report a subdataset for the bathymetry coverage when we expose one for the georeferenced metadata

rouault:
Otherwise GIS software based on GDAL might only expose the georeferenced data


rouault:
FYI @GlenRice-NOAA

elpaso (approved):

PR #9504, opened 2024-03-19
[WIP] Add new OGR driver for OpenDRIVE (XODR)

michikommader:
## What does this PR do?

This adds a read-only vector driver for the road description format [OpenDRIVE](https://www.asam.net/standards/detail/opendrive/). OpenDRIVE is an open industry standard in the automotive domain of driving simulation, maintained by ASAM e.V. It is increasingly intersecting with the public and GIS domains which raises the need for better interoperability with open-source spatial tools.

- This XODR driver is developed by German Aerospace Center (DLR) and licenced under MIT.
- This XODR driver implementation uses [libOpenDRIVE](https://github.com/pageldev/libOpenDRIVE) which is licenced under Apache 2.0.
- This XODR driver is supposed to be used as plug-in.

I am am open to a fruitful discussion in order to make OpenDRIVE directly usable through GDAL.

## Tasklist

 - [x] Make sure *newly added files* are correctly formatted (cf [pre-commit configuration](https://gdal.org/development/dev_practices.html#commit-hooks))
 - [x] Add [test](https://github.com/DLR-TS/gdal/blob/libopendrive-pr/autotest/ogr/ogr_xodr.py) case(s)
 - [x] Add [documentation](https://github.com/DLR-TS/gdal/blob/libopendrive-pr/doc/source/drivers/vector/xodr.rst)
 - [ ] Clarify handling of custom [`docker/ubuntu-small/DockerfileXODR`](https://github.com/DLR-TS/gdal/blob/libopendrive-pr/docker/ubuntu-small/DockerfileXODR) with GDAL maintainers: Which file location is better suited for future maintenance – the driver's source code directory or a separate repository?
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS: tested with Ubuntu 22.04, Windows 10
* Compiler: GCC on Linux, GCC 13.1.0 x86_64 (MCF threads) MinGW-w64 MSVCRT on Windows


autotest/ogr/ogr_xodr.py
```diff
@@ -0,0 +1,285 @@
...
+#
+# Project:  GDAL/OGR Test Suite
+# Purpose:  XODR driver testing.
+# Author:   Michael Scholz, German Aerospace Center (DLR)
+#           Gülsen Bardak, German Aerospace Center (DLR)
+#
+###############################################################################
+# Copyright 2024 German Aerospace Center (DLR), Institute of Transportation Systems
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
```
rouault, line 13: Please use the MIT licence to be consistent with (most of) the code base

autotest/ogr/ogr_xodr.py
```diff
@@ -0,0 +1,285 @@
...
+        ), "wrong geometry created for RoadSignal"
+    else:
+        assert (
+            wkt == "POINT (618366.942790883 5809541.22374025 103.556888074495)"
+        ), "wrong geometry created for dissolved RoadSignal"
```
rouault, line 285: It would be worth adding a test that runs the "test_ogrsf" compliance checking utility on the sample file. Something like:
```
def test_ogr_xodr_test_ogrsf():

    import test_cli_utilities

    if test_cli_utilities.get_test_ogrsf_path() is None:
        pytest.skip()

    ret = gdaltest.runexternal(
        test_cli_utilities.get_test_ogrsf_path() + " -ro " + xodr_file
    )

    assert "INFO" in ret
    assert "ERROR" not in ret
```

doc/source/drivers/vector/xodr.rst
```diff
@@ -0,0 +1,198 @@
+.. _vector.xodr:
+
+XODR -- OpenDRIVE Road Description Format
+=========================================
+
```
rouault, line 5:
```suggestion

.. versionadded:: 3.9

```

doc/source/drivers/vector/xodr.rst
```diff
@@ -0,0 +1,198 @@
...
+Driver capabilities
+-------------------
+
+.. supports_georeferencing::
+
+Specification version
+---------------------
+
+The currently supported OpenDRIVE version is 1.4 and basically depends on what is provided by libOpenDRIVE_.
```
rouault, line 22:
```suggestion
The currently supported OpenDRIVE version is 1.4 and depends on what is provided by libOpenDRIVE_. 
```

doc/source/drivers/vector/xodr.rst
```diff
@@ -0,0 +1,198 @@
...
+
+By definition, OpenDRIVE geometries are always referenced in a Cartesian coordinate system which defaults to having its origin at ``(0, 0, 0)``. If real-world coordinates are used, the OpenDRIVE header provides a PROJ.4 string with the definition of a projected spatial reference system:
+
+::
+
+  <header ...>
+    <geoReference><![CDATA[+proj=tmerc +lat_0=0 +lon_0=9 +k=0.9996 +x_0=500000 +y_0=0 +datum=WGS84 +units=m +no_defs]]></geoReference>
+  </header>
+
+The XODR driver uses this PROJ definition as spatial reference for creation of all OGR geometry layers.
```
rouault, line 49: Note: I assume you can't easily change the format. But it would be much better to use a WKT CRS string instead (cf https://www.ogc.org/standard/wkt-crs/)

doc/source/drivers/vector/xodr.rst
```diff
@@ -0,0 +1,198 @@
...
+  <header ...>
+    <geoReference><![CDATA[+proj=tmerc +lat_0=0 +lon_0=9 +k=0.9996 +x_0=500000 +y_0=0 +datum=WGS84 +units=m +no_defs]]></geoReference>
+  </header>
+
+The XODR driver uses this PROJ definition as spatial reference for creation of all OGR geometry layers. 
+
+Limitations
+-----------
+
+By default, OpenDRIVE XML files are opened by GDAL always in UTF-8 encoding, see :cpp:func:`VSIFOpenExL`.
```
rouault, line 54: I'm not sure to understand this sentence. Are you refering to the encoding of the filename, or the content of the file? If it is the encoding of the filename, then it is a general GDAL policy, and not appropriate documenting here.
If it is the content, then isn't it OpenDrive job to convert to UTF8 ?

doc/source/drivers/vector/xodr.rst
```diff
@@ -0,0 +1,198 @@
...
+
+Open options
+------------
+
+The following open options can be specified
+(typically with the ``-oo name=value`` parameters of :program:`ogrinfo` or :program:`ogr2ogr`):
+
+-  .. oo:: EPS
```
rouault, line 62: A longer name would be better (to be changed elsewhere as well)
```suggestion
-  .. oo:: EPSILON
```

doc/source/drivers/vector/xodr.rst
```diff
@@ -0,0 +1,198 @@
...
+- Convert the whole OpenDRIVE dataset with custom parameters :oo:`EPS` and :oo:`DISSOLVE_TIN` into a :ref:`GeoPackage <vector.gpkg>`:
+
+  ::
+
+    ogr2ogr -f "GPKG" CulDeSac.gpkg CulDeSac.xodr -oo EPS=0.9 -oo DISSOLVE_TIN=YES
+
+Convenient usage through docker image 
+-------------------------------------
+
+To use the XODR driver inside a docker image, first build the image from the corresponding docker image directory
```
rouault, line 99: Please remove the docker/ubuntu-small/DockerfileXODR file, and move the building of OpenDrive into the docker/ubuntu-full (full) build recipee. It would also be fine to move it as well in the alpine-normal (assuming OpenDrive is not too big). But we definitely don't want a new Dockerfile to maintain. 4 is already more than enough :-)

doc/source/drivers/vector/xodr.rst
```diff
@@ -0,0 +1,198 @@
...
+    
+    ::
+
+      docker run --name <container_name> -it gdal/xodr /bin/bash
+
+
+General building notes
+----------------------
+
+Building of the driver as plugin is tested to work on
```
rouault, line 131: Regarding building documentation, a section for libopendrive (similar to the other 66 libraries...) should be added to doc/source/development/building_from_source.rst

ogr/ogrsf_frmts/xodr/ogr_xodr.h
```diff
@@ -0,0 +1,261 @@
+/******************************************************************************
+ * $Id$
+ *
```
rouault, line 3:
```suggestion
```

ogr/ogrsf_frmts/xodr/ogr_xodr.h
```diff
@@ -0,0 +1,261 @@
...
+    OGRTriangulatedSurface triangulateSurface(odr::Mesh3D mesh);
+
+  public:
+    /**
+     * \param dissolveTriangulatedSurface True if original triangulated surface meshes from 
+     * libOpenDRIVE are to be dissolved into simpler geometries.
+     * Only applicable for layer types derived from meshes.
+    */
+    // TODO For lower memory consumption maybe better pass xodrRoadElements by reference?
+    OGRXODRLayer(RoadElements xodrRoadElements, std::string proj4Defn);
```
rouault, line 135: The coding standard in the code base is to pass objects as much as possible as const&. At least for std::string. Not sure how doable it is for your OpenDrive objects

ogr/ogrsf_frmts/xodr/ogr_xodr.h
```diff
@@ -0,0 +1,261 @@
...
+
+  protected:
+    VSILFILE *file;
+    RoadElements roadElements;
+    OGRSpatialReference spatialRef;
+    bool dissolveTIN;
+    OGRFeatureDefn *featureDefn;
+
+    /* Unique feature ID which is automatically incremented for any new road feature creation. */
+    int nNextFID;
```
rouault, line 95: Please use a default inline initialization or member variables, and use a m_ prefix. Applies to other similar places

ogr/ogrsf_frmts/xodr/ogr_xodr.h
```diff
@@ -0,0 +1,261 @@
...
+  private:
+    OGRXODRLayer **layers;
+    int nLayers;
+
+    /**
+     * Retrieves all necessary road elements from the underlying OpenDRIVE structure.
+     * 
+     * \param roads Roads of the dataset.
+    */
+    RoadElements createRoadElements(const std::vector<odr::Road> roads);
```
rouault, line 234: perhaps ?
```suggestion
    RoadElements createRoadElements(const std::vector<odr::Road>& roads);
```

ogr/ogrsf_frmts/xodr/ogr_xodr.h
```diff
@@ -0,0 +1,261 @@
...
+     * OGC Simple Feature geometries.
+    */
+    double eps;
+
+    /**
+     * Whether to dissolve triangulated surfaces which are created from libOpenDRIVE's meshes.
+    */
+    bool dissolveTIN;
+
+    int Open(const char *fileName, char **openOptions, int bUpdate);
```
rouault, line 251: bUpdate likely not needed if you don't support updates.

```suggestion
    bool Open(const char *fileName, CSLConstList openOptions);
```

ogr/ogrsf_frmts/xodr/ogr_xodr.h
```diff
@@ -0,0 +1,261 @@
...
+
+/*--------------------------------------------------------------------*/
+/*--------------------  Data source declarations ----------------------*/
+/*--------------------------------------------------------------------*/
+
+class OGRXODRDataSource : public GDALDataset
+{
+  private:
+    OGRXODRLayer **layers;
+    int nLayers;
```
rouault, line 227: Most recent GDAL drivers uses C++11 and beyond capabilities to avoid manual memory management as much as possible:
```suggestion
    std::vector<std::unique_ptr<OGRXODRLayer>> m_apoLayers{};
```

ogr/ogrsf_frmts/xodr/ogrxodrdatasource.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+OGRXODRDataSource::~OGRXODRDataSource()
+{
+    for (int i = 0; i < nLayers; i++)
+    {
+        delete layers[i];
+        layers[i] = nullptr;
+    }
+    CPLFree(layers);
+}
```
rouault, line 51: all this could be removed with the proposed std::vector<std::unique_ptr<>> m_apoLayers

ogr/ogrsf_frmts/xodr/ogrxodrdatasource.cpp
```diff
@@ -0,0 +1,186 @@
...
+ * DEALINGS IN THE SOFTWARE.
+ ****************************************************************************/
+
+#include "ogr_xodr.h"
+using namespace odr;
+using namespace pugi;
+using namespace std;
+
+CPL_CVSID("$Id$")
+
```
rouault, line 36:
```suggestion
```

ogr/ogrsf_frmts/xodr/ogrxodrdatasource.cpp
```diff
@@ -0,0 +1,186 @@
...
+        CPLError(CE_Failure, CPLE_OpenFailed,
+                 "Write or update access not supported by XODR driver.");
+        return CE_Failure;
+    }
+    else
+    {
+        file = VSIFOpenExL(fileName, "r", true);
+    }
+
+    if (file == nullptr)
```
rouault, line 70: Not necessarily. The user might be a GDAL /vsi file.  But I suspect your OpenDrive code only supports regular files. So you should rather use the standard VSIFOpen() or fopen().
Ah I believe I now understand your doc comment about UTF-8 and files. What does  odr::OpenDriveMap xodr(fileName, false); expect?  Because GDAL will provide a UTF-8 filename.

ogr/ogrsf_frmts/xodr/ogrxodrdatasource.cpp
```diff
@@ -0,0 +1,186 @@
...
+        return CE_Failure;
+    }
+
+    const char *openOptionValue =
+        CSLFetchNameValueDef(openOptions, "EPS", "1.0");
+    eps = CPLAtof(openOptionValue);
+    openOptionValue = CSLFetchNameValueDef(openOptions, "DISSOLVE_TIN", "NO");
+    dissolveTIN = CPLTestBool(openOptionValue);
+
+    odr::OpenDriveMap xodr(fileName, false);
```
rouault, line 84: can't this fail?

ogr/ogrsf_frmts/xodr/ogrxodrdatasource.cpp
```diff
@@ -0,0 +1,186 @@
...
+    return layers[iLayer];
+}
+
+int OGRXODRDataSource::TestCapability(CPL_UNUSED const char *capability)
+{
+    if (EQUAL(capability, ODsCCreateLayer))
+        return FALSE;
+    else if (EQUAL(capability, ODsCDeleteLayer))
+        return FALSE;
+    else if (EQUAL(capability, ODsCZGeometries))
```
rouault, line 117:
```suggestion
    if (EQUAL(capability, ODsCZGeometries))
```

ogr/ogrsf_frmts/xodr/ogrxodrdatasource.cpp
```diff
@@ -0,0 +1,186 @@
...
+        return TRUE;
+    return FALSE;
+}
+
+RoadElements
+OGRXODRDataSource::createRoadElements(const std::vector<odr::Road> roads)
+{
+    RoadElements elements;
+
+    for (odr::Road road : roads)
```
rouault, line 127: const ref if possible:
```suggestion
    for (const odr::Road &road : roads)
```
applies to inner loops as well

ogr/ogrsf_frmts/xodr/ogrxodrdriver.cpp
```diff
@@ -0,0 +1,86 @@
...
+ * DEALINGS IN THE SOFTWARE.
+ ****************************************************************************/
+
+#include "ogr_xodr.h"
+#include "cpl_conv.h"
+#include "cpl_error.h"
+
+CPL_CVSID(
+    "$Id$")  //TODO do we need this? Also the $Id$ in above licence headers?
+
```
rouault, line 36:
```suggestion
```

ogr/ogrsf_frmts/xodr/ogrxodrdriver.cpp
```diff
@@ -0,0 +1,86 @@
...
+extern "C" void CPL_DLL RegisterOGRXODR();
+
+static GDALDataset *OGRXODRDriverOpen(GDALOpenInfo *openInfo)
+{
+    if (openInfo->eAccess == GA_Update || openInfo->fpL == nullptr)
+        return nullptr;
+
+    OGRXODRDataSource *dataSource = new OGRXODRDataSource();
```
rouault, line 44:
```suggestion
    auto dataSource = std::make_unique<OGRXODRDataSource>();
```

ogr/ogrsf_frmts/xodr/ogrxodrdriver.cpp
```diff
@@ -0,0 +1,86 @@
...
+}
+
+static int OGRXODRDriverIdentity(GDALOpenInfo *openInfo)
+{
+    return EQUAL(CPLGetExtension(openInfo->pszFilename), "xodr");
+}
+
+void RegisterOGRXODR()
+{
+    if (GDALGetDriverByName("XODR") != nullptr)
```
rouault, line 63: As your driver might be built as a plugin, could you use the new way of doing it per https://gdal.org/development/rfc/rfc96_deferred_plugin_loading.html?  Cf ogr/ogrsf_frmts/xls for a random example to take inspiration for (separation between ogrxlsdriver.cpp and ogrxlsdrivercore.cpp)

ogr/ogrsf_frmts/xodr/ogrxodrlayer.cpp
```diff
@@ -0,0 +1,119 @@
...
+    roadMarkMeshIter = roadElements.roadMarkMeshes.begin();
+
+    roadObjectIter = roadElements.roadObjects.begin();
+    roadObjectMeshesIter = roadElements.roadObjectMeshes.begin();
+
+    roadSignalIter = roadElements.roadSignals.begin();
+    roadSignalMeshesIter = roadElements.roadSignalMeshes.begin();
+}
+
+OGRTriangulatedSurface OGRXODRLayer::triangulateSurface(odr::Mesh3D mesh)
```
rouault, line 91: you could likely save memory copies with:
```suggestion
std::unique_ptr<OGRTriangulatedSurface> OGRXODRLayer::triangulateSurface(odr::Mesh3D mesh)
```

ogr/ogrsf_frmts/xodr/ogrxodrlayerlane.cpp
```diff
@@ -0,0 +1,134 @@
...
+    {
+        // Skip lane(s) with id 0
+        laneIter++;
+        laneMeshIter++;
+        laneRoadIDIter++;
+    }
+
+    if (laneIter != roadElements.lanes.end())
+    {
+        feature = std::unique_ptr<OGRFeature>(new OGRFeature(featureDefn));
```
rouault, line 61:
```suggestion
        feature = std::make_unique<OGRFeature>(featureDefn);
```

ogr/ogrsf_frmts/xodr/ogrxodrlayerlane.cpp
```diff
@@ -0,0 +1,134 @@
...
+        odr::Mesh3D laneMesh = *laneMeshIter;
+
+        std::string laneRoadID = *laneRoadIDIter;
+
+        OGRTriangulatedSurface tin = triangulateSurface(laneMesh);
+
+        if (dissolveTIN)
+        {
+            OGRGeometry *dissolvedPolygon = tin.UnaryUnion();
+            feature->SetGeometry(dissolvedPolygon);
```
rouault, line 73:
```suggestion
            feature->SetGeometryDirectly(dissolvedPolygon);
```
otherwise memory leak. There are other similar instances to fix

ogr/ogrsf_frmts/xodr/ogrxodrlayerlaneborder.cpp
```diff
@@ -0,0 +1,115 @@
...
+    defineFeatureClass();
+}
+
+OGRFeature *OGRXODRLayerLaneBorder::GetNextFeature()
+{
+    std::unique_ptr<OGRFeature> feature;
+
+    if (laneIter != roadElements.lanes.end())
+    {
+        feature = std::unique_ptr<OGRFeature>(new OGRFeature(featureDefn));
```
rouault, line 52:
```suggestion
        feature = std::make_unique<OGRFeature>(featureDefn);
```

ogr/ogrsf_frmts/xodr/ogrxodrlayerlaneborder.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+        odr::Lane lane = *laneIter;
+        odr::Line3D laneOuter = *laneLinesOuterIter;
+        std::string laneRoadID = *laneRoadIDIter;
+
+        OGRLineString lineString;
+        for (auto vertexIter = laneOuter.begin(); vertexIter != laneOuter.end();
+             ++vertexIter)
+        {
+            odr::Vec3D laneVertex = *vertexIter;
```
rouault, line 62: may be simplified as?
```suggestion
        for (const auto& laneVertex: laneOuter)
        {
```

ogr/ogrsf_frmts/xodr/ogrxodrlayerlaneborder.cpp
```diff
@@ -0,0 +1,115 @@
...
+            lineString.addPoint(laneVertex[0], laneVertex[1], laneVertex[2]);
+        }
+        OGRGeometry *geometry = lineString.MakeValid();
+
+        feature->SetGeometry(geometry);
```
rouault, line 67:
```suggestion
        feature->SetGeometryDirectly(geometry);
```

ogr/ogrsf_frmts/xodr/ogrxodrlayerreferenceline.cpp
```diff
@@ -0,0 +1,101 @@
...
+    defineFeatureClass();
+}
+
+OGRFeature *OGRXODRLayerReferenceLine::GetNextFeature()
+{
+    std::unique_ptr<OGRFeature> feature;
+
+    if (roadIter != roadElements.roads.end())
+    {
+        feature = std::unique_ptr<OGRFeature>(new OGRFeature(featureDefn));
```
rouault, line 52: make_unique

ogr/ogrsf_frmts/xodr/ogrxodrlayerreferenceline.cpp
```diff
@@ -0,0 +1,101 @@
...
+        for (auto vertexIter = refLine.begin(); vertexIter != refLine.end();
+             ++vertexIter)
+        {
+            odr::Vec3D refLineVertex = *vertexIter;
+            lineString.addPoint(refLineVertex[0], refLineVertex[1],
+                                refLineVertex[2]);
+        }
+        OGRGeometry *geometry = lineString.MakeValid();
+
+        feature->SetGeometry(geometry);
```
rouault, line 67:
```suggestion
        feature->SetGeometryDirectly(geometry);
```

ogr/ogrsf_frmts/xodr/ogrxodrlayerroadmark.cpp
```diff
@@ -0,0 +1,113 @@
...
+    defineFeatureClass();
+}
+
+OGRFeature *OGRXODRLayerRoadMark::GetNextFeature()
+{
+    std::unique_ptr<OGRFeature> feature;
+
+    if (roadMarkIter != roadElements.roadMarks.end())
+    {
+        feature = std::unique_ptr<OGRFeature>(new OGRFeature(featureDefn));
```
rouault, line 53: make_unique

ogr/ogrsf_frmts/xodr/ogrxodrlayerroadmark.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+        odr::RoadMark roadMark = *roadMarkIter;
+        odr::Mesh3D roadMarkMesh = *roadMarkMeshIter;
+
+        OGRTriangulatedSurface tin = triangulateSurface(roadMarkMesh);
+
+        if (dissolveTIN)
+        {
+            OGRGeometry *dissolvedPolygon = tin.UnaryUnion();
+            feature->SetGeometry(dissolvedPolygon);
```
rouault, line 63:
```suggestion
            feature->SetGeometryDirectly(dissolvedPolygon);
```

ogr/ogrsf_frmts/xodr/ogrxodrlayerroadobject.cpp
```diff
@@ -0,0 +1,101 @@
...
+    defineFeatureClass();
+}
+
+OGRFeature *OGRXODRLayerRoadObject::GetNextFeature()
+{
+    std::unique_ptr<OGRFeature> feature;
+
+    if (roadObjectIter != roadElements.roadObjects.end())
+    {
+        feature = std::unique_ptr<OGRFeature>(new OGRFeature(featureDefn));
```
rouault, line 52: make_unique

ogr/ogrsf_frmts/xodr/ogrxodrlayerroadsignal.cpp
```diff
@@ -0,0 +1,157 @@
...
+    defineFeatureClass();
+}
+
+OGRFeature *OGRXODRLayerRoadSignal::GetNextFeature()
+{
+    std::unique_ptr<OGRFeature> feature;
+
+    if (roadSignalIter != roadElements.roadSignals.end())
+    {
+        feature = std::unique_ptr<OGRFeature>(new OGRFeature(featureDefn));
```
rouault, line 53: make_unique

ogr/ogrsf_frmts/xodr/ogrxodrlayerroadsignal.cpp
```diff
@@ -0,0 +1,157 @@
...
+            odr::Road road = roadElements.roads.at(roadId);
+
+            double s = roadSignal.s0;
+            double t = roadSignal.t0;
+            double z = roadSignal.zOffset;
+            odr::Vec3D xyz = road.get_xyz(s, t, z);
+
+            OGRPoint point(xyz[0], xyz[1], xyz[2]);
+            OGRGeometry *geometry = point.MakeValid();
+            feature->SetGeometry(geometry);
```
rouault, line 70:
```suggestion
            feature->SetGeometryDirectly(geometry);
```

rouault (commented):
Nice work generally. See my comments.

I would also like this to be triggered more by our CI.
Can you add a libopendrive build in .github/workflows/ubuntu_20.04/Dockerfile.ci and .github/workflows/ubuntu_22.04/Dockerfile.ci

Do you know how your driver (and libopendrive) is robust to corrupted / hostile datasets?  At the very minimum, I'd like to see a test with a "random file" that has xodr extension, but isn't valid XODR file. Ideally we'd want to stress-test it using OSSFuzz. Cf https://github.com/OSGeo/gdal/blob/master/fuzzers/README.TXT , https://github.com/OSGeo/gdal/blob/master/fuzzers/build.sh

rouault:
Is the some packaging of libopendrive in Debian, conda-forge, etc ?

doc/source/drivers/vector/xodr.rst
michikommader: As you already assumed, this currently cannot be changed in OpenDRIVE. The `<geoReference>` tag has been introduced many years ago as part of the dataset's `<header>` with version 1.4 – this was already a novelty – and is still persisting in the recently released version 1.8. The OpenDRIVE standard defines it to hold a [PROJ projection string](https://proj.org/en/9.4/usage/projections.html). I will forward your hint regarding preferred usage of WKT CRS strings to the committee.

rouault:
@michikommader Do you plan any further work on this any time soon ? I'm asking as I'm trying to keep the list of opened PRs low, and if something is not going to progress for some time, it is better to close the PR, and re-open it later when work restart

michikommader:
> @michikommader Do you plan any further work on this any time soon ?

Apologies for my inactivity. Yes, we already went through some of your valuable remarks internally but did not review and push changes yet. I plan to supply all the missing information in the first week of May.

michikommader:
> Is the some packaging of libopendrive in Debian, conda-forge, etc ?

Not yet. I myself am not maintainer of libOpenDRIVE and would rather see packaging responsibility on the maintainer side. We can separately discuss if our institute could take over this task. Also, ASAM e. V. as standardisation organisation behind OpenDRIVE could have an interest in that.

ogr/ogrsf_frmts/xodr/ogr_xodr.h
michikommader: I can't follow why you changed the return value of `Open()` from `int` to `bool`. Should we not be able to return any of the possible different `CPLErr`, such as other drivers do? We will for sure adapt your suggestion but I'd like to better understand the reason and intended driver behaviour :-).

rouault: There are a ton of legacy use of "int" in GDAL code base where "bool" should have rather been used since the returned values where TRUE or FALSE, and we try to use "bool" whenever we can in new code. I haven't checked here, but I strongly suspect you just need bool. And if you want to return CPLErr then use CPLErr as the return type.

doc/source/drivers/vector/xodr.rst
michikommader: We will add a specific section for libOpenDRIVE to `doc/source/development/building_from_source.rst`.

Should we then completely avoid "General building notes" in the actual driver documentation and also add these more comprehensive hints to `building_from_source.rst`?

ogr/ogrsf_frmts/xodr/ogr_xodr.h
michikommader: Wonderful, that makes total sense. Thanks for the clarification.

doc/source/drivers/vector/xodr.rst
michikommader: My sentence is relating to the actual content of an OpenDRIVE XML file, which not necessarily comes encoded in UTF-8. When reading through the doc of `VSIFOpenExL`/`VSIFOpenL`, I probably overread that it only tackles *filename* encoding.

Regarding the actual content encoding, libOpenDRIVE internally used pugixml which in default configuration tries to guess the input encoding (see [`load_file()`](https://github.com/DLR-TS/libOpenDRIVE/blob/9a0437f8a18d445d5c43fe2a4c9401d8a4b770f0/thirdparty/pugixml/pugixml.hpp#L1047) and [4.6. Encodings](https://pugixml.org/docs/manual.html#loading.encoding)). I will change my hint in the GDAL driver doc accordingly.

rouault: yes, VSIFOpenExL() only cares about the encoding of the filename, which should be UTF-8.  The encoding of the content of the file is another topic, which must be dealt by upper layers.

michikommader:
> Do you know how your driver (and libopendrive) is robust to corrupted / hostile datasets? At the very minimum, I'd like to see a test with a "random file" that has xodr extension, but isn't valid XODR file.

libOpenDRIVE forwards XML data loading directly to pugixml. According to pugixml's [exception guarantees](https://pugixml.org/docs/manual.html#dom.exception) "most pugixml functions have a no-throw exception guarantee". For [handling parsing errors](https://pugixml.org/docs/manual.html#loading.errors) the `xml_parse_result` would have to be inspected for its `xml_parse_status`. But, from GDAL driver side we do not have access to that. Unfortunately, libOpenDRIVE itself only uses simple `printf()` statements to report possible problems.

At least, libOpenDRIVE gives access to the parsed `pugi::xml_document`. Checking for the availability of "expected OpenDRIVE XML nodes" would allow basic validation if the parsing was successful. But, I don't see this as a very practicable way. If the dataset was corrupted "somewhere in the middle", pugixml would still have successfully parsed all previous content and made this accessible to the caller. For our use case of ensuring robustness in GDAL, I see this as a major deficit. The best solution would be to extend libOpenDRIVE to allow at least access to pugixml's `xml_parse_status` with which we could implement a somewhat meaningful error handling. I will approach the maintainers.

ogr/ogrsf_frmts/xodr/ogrxodrdatasource.cpp
michikommader: See my [other answer](https://github.com/OSGeo/gdal/pull/9504#issuecomment-2113442286). For now, I will create a basic check if an "OpenDRIVE" node has successfully been parsed during creation of the `odr::OpenDriveMap`. This would fail for non-OpenDRIVE input files.

If the input just contains random structural XML errors, but generally is OpenDRIVE, all content up to the first error will be parsed and the creation of the `odr::OpenDriveMap` will *not* fail, unfortunately.

ogr/ogrsf_frmts/xodr/ogrxodrdriver.cpp
michikommader: When following RFC 96 and ogr/ogrsf_frmts/xls simultaneously, I get confused. Is it sufficient just to extend `gdalallregister.cpp` with the following?

```
#if defined(DEFERRED_XODR_DRIVER)
    DeclareDeferredOGRXODRPlugin();
#endif
```

Or should we also extend `ogrregisterall.cpp` together with `drivers.ini`  with the following?

```
#ifdef XODR_ENABLED
    RegisterOGRXODR();
#endif
```

rouault: both are needed. Cf commit 69dd1e53435401e47692d14da45c6f452133c475 where the XLS driver was converted to implement RFC 96

autotest/ogr/ogr_xodr.py
michikommader: Oh, that's a good one 😅! After adding this test, a whole lot of new errors are exposed.

michikommader: Haven't consciously worked with OGR spatial and attribute filters yet. Apache Arrow streams also not yet considered. I will dig into that.

```
INFO: Testing layer ReferenceLine.
...
E           ERROR: Attribute filter seems to be ignored in GetFeatureCount() when spatial filter is set.
E         ? +++++
E           ERROR: Spatial filter (0) failed to eliminatea feature unexpectedly!
...
E           ERROR: GetFeatureCount() may not be taking attribute filter into account (nInclusiveCount = 41, nExclusiveCount = 41, nExclusiveCountWhileIterating = 0, nTotalCount = 41).
...
E           ERROR: GetArrowStream() with 1=0 filter returned 41 features, whereas 0 expected
...
E           ERROR: GetArrowStream() with ID <> '2233498' filter returned 41 features, whereas less than 41 expected
...
E           ERROR: ExecuteSQL() should have returned a layer without features.
E           ERROR: ExecuteSQL() should have returned a layer without features.

```

michikommader: As we are not explicitly addressing spatial and attribute filters in any of our layers, I suspect that we are still not implementing our layers 100 % correctly. Otherwise these tests which fall back to the default implementation of `OGRLayer` should by default run correctly, I suppose.

rouault: > Apache Arrow streams also not yet considered.

don't worry about those ones. They are just a consequence of regular feature-based filtering not working yet, as your driver relies on the generic GetArrowArray implementation that fallbacks to using GetNextFeature() and is thus subject to its bugs

> As we are not explicitly addressing spatial and attribute filters in any of our layers,

Instead of directly implementin GetNextFeature(), you could could rename it to GetNextRawFeature() and make your OGRxxxxLayer classes extend from ``public OGRLayer, public OGRGetNextFeatureThroughRaw<OGRxxxxLayer>``. Cf for example at ogr/ogrsf_frmts/lvbag/ogr_lvbag.h . The OGRGetNextFeatureThroughRaw class implements a GetNextFeature() method that calls GetNextRawFeature() and then evaluates the spatial and attribute filters

michikommader: I appreciate your precise guidance. Following your suggestion, everything works as expected! The new test is passing and it saved me probably ~~half~~ a weekend of "getting my head into something yet unknown".

michikommader:
I'll focus on satisfying cppcheck_2004 next days.

doc/source/development/building_from_source.rst
```diff
@@ -1145,6 +1145,19 @@ by the :ref:`raster.marfa` driver.
     Control whether to use LibQB3. Defaults to ON when LibQB3 is found.
 
 
+libOpenDRIVE
```
rouault, line 1148: to be moved before libQB3 for lexicographic order

.github/workflows/ubuntu_20.04/Dockerfile.ci
```diff
@@ -258,6 +258,25 @@ RUN mkdir sqlite \
     && cd .. \
     && rm -rf sqlite
 
+# Build libOpenDRIVE
+ARG OPENDRIVE_VERSION=0.5.0-gdal
+RUN . /buildscripts/bh-set-envvars.sh \
+    && if test "${OPENDRIVE_VERSION}" != ""; then ( \
```
rouault, line 264: bh-set-envvars.sh is not used in CI Dockerfiles:

```suggestion
RUN if test "${OPENDRIVE_VERSION}" != ""; then ( \
```

.github/workflows/ubuntu_20.04/Dockerfile.ci
```diff
@@ -258,6 +258,25 @@ RUN mkdir sqlite \
...
+    && tar xzf ${OPENDRIVE_VERSION}.tar.gz \
+    && rm -f ${OPENDRIVE_VERSION}.tar.gz \
+    && cd libOpenDRIVE-${OPENDRIVE_VERSION} \
+    && cmake . -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=Release \
+    -DCMAKE_INSTALL_PREFIX=/usr/ \
+    && make -j$(nproc) \
+    && make install \
+    && mkdir -p /build_thirdparty/usr/lib \
+    && cp -P /usr/lib/libOpenDrive*.so* /build_thirdparty/usr/lib \
+    && for i in /build_thirdparty/usr/lib/*; do strip -s $i 2>/dev/null || /bin/true; done \
```
rouault, line 275:
```suggestion
```

.github/workflows/ubuntu_22.04/Dockerfile.ci
```diff
@@ -104,6 +104,25 @@ RUN mkdir mongocxx \
     && make install \
     && cd ../.. \
     && rm -rf mongocxx
+    
+# Build libOpenDRIVE
+ARG OPENDRIVE_VERSION=0.5.0-gdal
+RUN . /buildscripts/bh-set-envvars.sh \
+    && if test "${OPENDRIVE_VERSION}" != ""; then ( \
```
rouault, line 111:
```suggestion
RUN if test "${OPENDRIVE_VERSION}" != ""; then ( \
```

.github/workflows/ubuntu_22.04/Dockerfile.ci
```diff
@@ -104,6 +104,25 @@ RUN mkdir mongocxx \
...
+    && tar xzf ${OPENDRIVE_VERSION}.tar.gz \
+    && rm -f ${OPENDRIVE_VERSION}.tar.gz \
+    && cd libOpenDRIVE-${OPENDRIVE_VERSION} \
+    && cmake . -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=Release \
+    -DCMAKE_INSTALL_PREFIX=/usr/ \
+    && make -j$(nproc) \
+    && make install \
+    && mkdir -p /build_thirdparty/usr/lib \
+    && cp -P /usr/lib/libOpenDrive*.so* /build_thirdparty/usr/lib \
+    && for i in /build_thirdparty/usr/lib/*; do strip -s $i 2>/dev/null || /bin/true; done \
```
rouault, line 122:
```suggestion
```

michikommader (commented):
@rouault In the current master branch you use `ARROW_VERSION=16.0.0-1` but when testing, I have to set it to `16.1.0-1` in order to successfully build the Ubuntu 24.04 Docker image.

ogr/ogrsf_frmts/xodr/ogrxodrdrivercore.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+/************************************************************************/
+/*                    OGRXODRDriverIdentify()                            */
+/************************************************************************/
+
+int OGRXODRDriverIdentify(GDALOpenInfo *poOpenInfo)
+{
+    CPLString extension = CPLString(CPLGetExtension(poOpenInfo->pszFilename));
+    extension.tolower();
+    return EQUAL(extension, "xodr");
```
rouault, line 41: EQUAL() is case insensitive

```suggestion
    return EQUAL(CPLGetExtension(poOpenInfo->pszFilename), "xodr");
```

ogr/ogrsf_frmts/xodr/ogrxodrdatasource.cpp
```diff
@@ -0,0 +1,188 @@
...
+using namespace pugi;
+using namespace std;
+
+bool OGRXODRDataSource::Open(const char *pszFilename, CSLConstList openOptions)
+{
+    VSILFILE *file = nullptr;
+    file = VSIFOpenL(pszFilename, "r");
+    if (file == nullptr)
+        return FALSE;
+
```
rouault, line 41: This leaks the VSILFILE*, and is not necessary due OGRXODRDriverOpen() testing that poOpenInfo->fpL != nullptr

```suggestion
```

ogr/ogrsf_frmts/xodr/ogrxodrlayerlaneborder.cpp
```diff
@@ -0,0 +1,120 @@
...
+        odr::Lane lane = *m_laneIter;
+        odr::Line3D laneOuter = *m_laneLinesOuterIter;
+        std::string laneRoadID = *m_laneRoadIDIter;
+
+        OGRLineString lineString;
+        for (const auto &laneVertex : laneOuter)
+        {
+            lineString.addPoint(laneVertex[0], laneVertex[1], laneVertex[2]);
+        }
+        OGRGeometry *geometry = lineString.MakeValid();
```
rouault, line 90: MakeValid() might return nullptr in case of failure, so this should be tested to avoid a nullptr dereference. Furthermore MakeValid() is only available or GEOS-enabled builds. So the choice is either to make GEOS a build requirement for the driver in ogr/ogrsf_frmts/CMakeLists.txt with ``ogr_dependent_driver(xodr OpenDRIVE "GDAL_USE_OPENDRIVE;GDAL_USE_GEOS")``, or to only call MakeValid() if ``OGRGeometryFactory::haveGEOS()`` and use the original geometry then.
Is it common/expected that XODR datasets return non-valid geometries?

ogr/ogrsf_frmts/xodr/ogrxodrlayerreferenceline.cpp
```diff
@@ -0,0 +1,106 @@
...
+        odr::Road road = (*m_roadIter).second;
+        odr::Line3D refLine = *m_referenceLineIter;
+
+        OGRLineString lineString;
+        for (const auto &refLineVertex : refLine)
+        {
+            lineString.addPoint(refLineVertex[0], refLineVertex[1],
+                                refLineVertex[2]);
+        }
+        OGRGeometry *geometry = lineString.MakeValid();
```
rouault, line 84: cf above remark about MakeValid()

ogr/ogrsf_frmts/xodr/ogrxodrlayerroadmark.cpp
```diff
@@ -0,0 +1,121 @@
...
+        feature = std::make_unique<OGRFeature>(m_poFeatureDefn.get());
+
+        odr::RoadMark roadMark = *m_roadMarkIter;
+        odr::Mesh3D roadMarkMesh = *m_roadMarkMeshIter;
+
+        std::unique_ptr<OGRTriangulatedSurface> tin =
+            triangulateSurface(roadMarkMesh);
+        if (m_bDissolveTIN)
+        {
+            OGRGeometry *dissolvedPolygon = tin->UnaryUnion();
```
rouault, line 90: UnaryUnion() could potentially return nullptr, leading to nullptr dereference below

ogr/ogrsf_frmts/xodr/ogrxodrlayerroadsignal.cpp
```diff
@@ -0,0 +1,167 @@
...
+            std::string roadId = roadSignal.road_id;
+            odr::Road road = m_roadElements.roads.at(roadId);
+
+            double s = roadSignal.s0;
+            double t = roadSignal.t0;
+            double z = roadSignal.zOffset;
+            odr::Vec3D xyz = road.get_xyz(s, t, z);
+
+            OGRPoint point(xyz[0], xyz[1], xyz[2]);
+            OGRGeometry *geometry = point.MakeValid();
```
rouault, line 119: MakeValid() has no action on a point. The point could be passed directly

doc/source/drivers/vector/xodr.rst
```diff
@@ -0,0 +1,195 @@
...
+
+This should print a list of supported OGR formats, including ``XODR`` in the first row:
+
+  ::
+
+    Supported Formats:
+      XODR -vector- (rov): OpenDRIVE - Open Dynamic Road Information for Vehicle Environment
+      PCIDSK -raster,vector- (rw+v): PCIDSK Database File       
+      PDS4 -raster,vector- (rw+vs): NASA Planetary Data System 4
+      ...
```
rouault, line 187: More simple to just run "ogrinfo --format XODR". Furthermore XODR might not always be in the first row

ogr/ogrsf_frmts/xodr/ogrxodrlayerlaneborder.cpp
michikommader: It should be rather uncommon that XODR datasets return non-valid geometries. But as we rely on libOpenDRIVE's internal linear approximation of parametric geometries, we should at least validate our derived Simple Feature version. I will make this more robust with your suggestions in mind. I see it reasonable to make GEOS dependency optional.

michikommader: > or to only call MakeValid() if `OGRGeometryFactory::haveGEOS()`

Does there exist any similarly convenient function to test if SFCGAL is available?

michikommader: > I will make this more robust with your suggestions in mind

I omit using `MakeValid()` now because handling of a potentially modified return geometry is not trivial. Instead, just a testing for `IsValid()` is performed. If that fails, the feature's geometry will stay unset and a warning is issued. I am not sure if allowing geometry-less features in such boundary cases is a reasonable approach? Theoretically, such cases should not occur.

Further, GEOS is a required dependency now because we need it for `UnaryUnion()`.

rouault: > Does there exist any similarly convenient function to test if SFCGAL is available?

I don't think so. But you could had a bool OGRGeometryFactory::haveSFCGAL() if needed next to haveGEOS() in ogr/ogrgeometryfactory.cpp)

> Instead, just a testing for `IsValid()` is performed. If that fails, the feature's geometry will stay unset and a warning is issued

That seems a bit extreme. A lot (most) drivers will report invalid geometries, and let the responsibility to the user to fix them with MakeValid() (ogr2ogr has a -makevalid option for example), or discard them, or just process them as it. So if I were you, I would just report the geometry as obtained from libopendrive, ignoring if it is valid or not

rouault:
@michikommader Anything left on your side before merging this PR ? It looks good to me (the CI failures are unrelated)

michikommader:
> @michikommader Anything left on your side before merging this PR ? It looks good to me (the CI failures are unrelated)

I will today commit a small simplification regarding your last suggestions with `MakeValid()`.

michikommader:
> @michikommader Anything left on your side before merging this PR ? It looks good to me (the CI failures are unrelated)

From my side, there is nothing more to add. The driver is well usable in practical applications.

As next steps, I will look into providing ready-to-use binary releases of the libOpenDRIVE dependency. If development of libOpenDRIVE progresses with new features, this XODR driver might require adjustments. I would prepare theses as pull requests in a similar manner here.

Thank you for your patience and guidance for making this driver part of GDAL!

rouault:
I've refreshed the master Docker images, so "ghcr.io/osgeo/gdal:ubuntu-full-latest" and  "ghcr.io/osgeo/gdal:alpine-normal-latest" contain the driver

dbaston:
Should a `FindOpenDrive.cmake` file be added to avoid the following output in the cmake configure step?

```
CMake Warning at cmake/helpers/CheckDependentLibraries.cmake:149 (find_package):
  By not providing "FindOpenDrive.cmake" in CMAKE_MODULE_PATH this project
  has asked CMake to find a package configuration file provided by
  "OpenDrive", but CMake did not find one.

  Could not find a package configuration file provided by "OpenDrive" with
  any of the following names:

    OpenDriveConfig.cmake
    opendrive-config.cmake

  Add the installation prefix of "OpenDrive" to CMAKE_PREFIX_PATH or set
  "OpenDrive_DIR" to a directory containing one of the above files.  If
  "OpenDrive" provides a separate development package or SDK, be sure it has
  been installed.
  cmake/helpers/CheckDependentLibraries.cmake:807 (gdal_check_package)
  gdal.cmake:261 (include)
  CMakeLists.txt:240 (include)
```

https://github.com/OSGeo/gdal/actions/runs/9569105782/job/26380878179#step:13:361

rouault:
> Should a `FindOpenDrive.cmake` file be added to avoid the following output in the cmake configure step?

we want to avoid creating new Find cmake files when possible. Here it is not necessary as opendrive comes with CMake config files. Hence https://github.com/OSGeo/gdal/pull/10391 should fix it

Issue #9050, opened 2024-01-10
Feature Request: Add Line of Sight function


Ryanf55:
## Expected behavior and actual behavior.

I would like to be able to call a C++ function in GDAL to compute the Line of Sight visibility between two points over a DEM raster band. Currently, there only seems to be viewshed support, which is a lot of overhead if I just need to check line of sight between two points.

There was a request for an algorithm on the forums; it never made it into a supported feature: https://gis.stackexchange.com/questions/228726/algorithm-or-gdal-module-to-calculate-los

## Desired behavior

GDAL could add a new function like so:
```c++
struct Loc {
    double x;
    double y;
    double z;
}

static bool IsLineOfSightVisible(const GDALRasterBand& band, const Loc& locA, const Loc& locB); 
```

If LocA and LocB, and the geodetic line between them are all above terrain, it returns true. Otherwise false.

If either of the locations are outside the raster band, or the raster band reports no data at any of the locations along the line, then it would also return false.

From my perspective, it would be even better if there were an API that I could pass in the locations as LLH and have GDAL deal with the projections/conversions internally.

## Papers

https://www.researchgate.net/publication/2411280_Efficient_Line-of-Sight_Algorithms_for_Real_Terrain_Data 

This recommends use of Bresenham's line of sight algorithm in 3D.

## Operating system

Ubuntu 22.04

## GDAL version and provenance

master

## Other notes

Pending the support of the GDAL maintainers to add this feature, and they agree on the algorithm I found, I am happy to try to implement this and contribute it.

rouault:
I'd recommend a C compatible signature like
```
bool CPL_DLL GDALIsLineOfSightVisible(GDALRasterBandH, double xA, double yA, double zA, double xB, double yB, double zB, char** papszOptions);
```
implemented in a alg/los.cpp file

It would be good if we could be consistent of the results of GDALViewshedGenerate(), in particular the curvature coefficient (https://gdal.org/programs/gdal_viewshed.html#cmdoption-gdal_viewshed-cc), although that could be a later added option.

CC'ing @szekerest the author of gdal_viewshed

Ryanf55:
Awesome. Thanks for the tips. I appreciate any further recommendations.

[...]

PR #9506, opened 2024-03-19
Implement basic Line-of-sight algorithm

Ryanf55:
## What does this PR do?

This implements a basic LOS check using 2D bresenham. Height is interpolated with a linear interpolation between points.
I would like preliminary feedback (see below). I used a 2D bresenham because the height values may be floating point. 

I think if the height values are known to be integer-type at compile time, I think some templating could be used to optimize it with the 3D bresenham algorithm. For now, I used linear interpolation because it should work with all cases, and convert the dataset's height to a double.

## Open Questions

1. Is there a way to unit test for asserts and only have the tests run if compiled in debug mode?
   *  Don't test for nullptr, mark it in the API as not-null
1. Are the asserts I added on the `int` limits of pixel coordinates correct?
   *  Yea, but the lower level library already checks those things
1. Is there a better function call structure for the bresenham algorithm taking in a callback? I was trying to use functional principles and keep the function pure.
    * This is fine, and is actually good for exposing the violation location
1. Are there any easy optimizations that can be made for LoS queries that line up directly with a row or column of raster data
   * Optimizations can come later
1. Is it inefficient to be calling RasterIO for single-pixel values like I did, or is there a better way? The reason I did single-pixel values is to avoid a large runtime heap allocation of data to service the LOS query. 
   * Yea, I added a TODO for other approachs
1. Is there a better local dataset I can use for testing LOS queries on than the publicly hosted one I used in `GDALIsLineOfSightVisible_through_mountain`?
   * Yea, autotest/cpp/data/n43.dt0
1. How should the uncertainty in the height values be handled? Ideally, for my use case, I can supply with which certainty I want the LOS algorithm, under the assumption that the error values follow a normal distribution. For example, I want the line of sight check to evaluate true if the line between points has  90% or higher certainty it is free from intersection with terrain. Right now, by assuming the dataset is exact, it means we're only 50% certain, which is not great odds for an aerial vehicle.
   * Ignore uncertainty for this PR, handle it later

## Limitations

This isn't the fastest way. Optimizations can come later, including GPU acceleration.

#9050

## Test Data

The autotest folder has some nice DEM data that can be used directly instead of relying on the ArduPilot terrain server. 
```bash
gdaldem color-relief autotest/cpp/data/n43.dt0 n43_color.txt n43_color.tif
```
**n43_color.txt**
```
75 112 147 141
125 120 159 152
175 130 165 159
225 145 177 171
275 180 192 180
325 212 201 180
375 212 184 163
425 212 193 179
475 212 207 204
```

## Tasklist

 - [ ] Address open questions above
 - [x] Add test case(s)
 - [x] Migrate from zipped/curl terrain from ArduPilot to file in the repo
 - [x] Add documentation
 - [ ] Squash commits per subsystem
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS: Ubuntu 22.04
* Compiler: G++11

## Future extensions

Add a CLI tool to test LOS between two points. 

## Test instructions (mainly for myself)

```bash
 cmake -S . -B build -DBUILD_TESTING=ON -DCMAKE_BUILD_TYPE=Debug
cmake --build build --target gdal_unit_test -j `nproc`
./build/autotest/cpp/gdal_unit_test --gtest_filter="test_alg.GDALIsLineOfSightVisible*"
# Or, with GDB
 gdb ./build/autotest/cpp/gdal_unit_test
>>> r --gtest_filter="test_alg.GDALIsLineOfSightVisible*"
```

### DIFF
--- a/alg/los.cpp
+++ b/alg/los.cpp
@@ -155,8 +155,10 @@ static bool IsAboveTerrain(const GDALRasterBandH hBand, const int x,
  * Check Line of Sight between two points.
  * Both input coordinates must be within the raster coordinate bounds.
  *
- * This algorithm will check line of sight using a 3D Bresenham algorithm.
+ * This algorithm will check line of sight using a Bresenham algorithm.
  * https://www.researchgate.net/publication/2411280_Efficient_Line-of-Sight_Algorithms_for_Real_Terrain_Data
+ * Line of sight is computed in raster coordinate space, and thus may not be appropriate.
+ * For example, datasets referenced against geographic coordinate at high latitudes may have issues.
  *
  * @param hBand The band to read the DEM data from. This must NOT be null.
  * 
--- a/alg/los.cpp
+++ b/alg/los.cpp
@@ -126,12 +126,14 @@ bool Bresenham2D(int x1, int y1, int x2, int y2, std::function< auto(int, int) -
 
 
 // Get the elevation of a single point.
+static
 bool GetElevation(const GDALRasterBandH hBand, const int x, const int y, double& val)
 {
     return GDALRasterIO(hBand, GF_Read, x, y, 1, 1, &val, 1, 1, GDT_Float64, 0, 0) == CE_None;
 }
 
 // Check a single location is above terrain.
+static
 bool IsAboveTerrain(const GDALRasterBandH hBand, const int x, const int y, const double z)
 {
     double terrainHeight;
--- a/alg/los.cpp
+++ b/alg/los.cpp
@@ -49,8 +49,9 @@
 // The callback is run at every point along the line,
 // which should return True if the point is above terrain.
 // Bresenham2D will return true if all points have LOS between the start and end.
-static bool Bresenham2D(int x1, int y1, int x2, int y2,
-                        std::function<auto(int, int)->bool> OnBresenhamPoint)
+static bool
+Bresenham2D(const int x1, const int y1, const int x2, const int y2,
+            std::function<auto(const int, const int)->bool> OnBresenhamPoint)
 {
     bool isAboveTerrain = true;
     int dx, dy;
--- a/alg/los.cpp
+++ b/alg/los.cpp
@@ -152,7 +152,7 @@ bool IsAboveTerrain(const GDALRasterBandH hBand, const int x, const int y, const
  * Check Line of Sight between two points.
  * Both input coordinates must be within the raster coordinate bounds.
  *
- * This algorithm will check line of sight using a 3D bresenham algorithm.
+ * This algorithm will check line of sight using a 3D Bresenham algorithm.
  * https://www.researchgate.net/publication/2411280_Efficient_Line-of-Sight_Algorithms_for_Real_Terrain_Data
  *
  * @param hBand The band to read the DEM data from.
--- a/alg/los.cpp
+++ b/alg/los.cpp
@@ -210,7 +210,7 @@ bool GDALIsLineOfSightVisible(const GDALRasterBandH hBand, const int xA,
     // Use an interpolated Z height with 2D bresenham for the remaining cases.
 
     // Lambda for computing the square of a number
-    auto SQUARE = [](const int n) -> int { return std::pow(n, 2); };
+    auto SQUARE = [](const double d) -> double { return d * d; };
 
     // Lambda for Linear interpolate like C++20 std::lerp.
     auto lerp = [](const double a, const double b, const double t)
@@ -219,8 +219,10 @@ bool GDALIsLineOfSightVisible(const GDALRasterBandH hBand, const int xA,
     // Lambda for getting Z test height given x-y input along the bresenham line.
     auto GetZValue = [&](const int x, const int y) -> double
     {
-        const auto rNum = SQUARE(x - xA) + SQUARE(y - yA);
-        const auto rDenom = SQUARE(xB - xA) + SQUARE(yB - yA);
+        const auto rNum = SQUARE(static_cast<double>(x - xA)) +
+                          SQUARE(static_cast<double>(y - yA));
+        const auto rDenom = SQUARE(static_cast<double>(xB - xA)) +
+                            SQUARE(static_cast<double>(yB - yA));
         /// @todo In order to reduce CPU cost and avoid a sqrt operation, consider
         /// the approach to just the ratio along x or y depending on whether
         /// the line is steep or shallow.
--- a/alg/los.cpp
+++ b/alg/los.cpp
@@ -86,7 +86,7 @@ bool Bresenham2D(int x1, int y1, int x2, int y2, std::function< auto(int, int) -
 	{
 		dy <<= 1;
 		balance = dy - dx;
-		dx <<= 1;
+		dx *= 2;
 
 		while (x != x2 && isAboveTerrain)
 		{
@@ -104,9 +104,9 @@ bool Bresenham2D(int x1, int y1, int x2, int y2, std::function< auto(int, int) -
 	}
 	else
 	{
-		dx <<= 1;
+		dx *= 2;
 		balance = dx - dy;
-		dy <<= 1;
+		dy *= 2;
 
 		while (y != y2  && isAboveTerrain)
 		{
--- a/alg/los.cpp
+++ b/alg/los.cpp
@@ -165,11 +165,9 @@ bool IsAboveTerrain(const GDALRasterBandH hBand, const int x, const int y, const
  */
 
 bool GDALIsLineOfSightVisible(
-    const GDALRasterBandH hBand, const int xA, const int yA, const double zA, const int xB, const int yB, const double zB, const char** papszOptions)
+    const GDALRasterBandH hBand, const int xA, const int yA, const double zA, const int xB, const int yB, const double zB, CPL_UNUSED CSLConstList papszOptions)
 {
-    if(hBand == nullptr) {
-        return false;
-    }
+    VALIDATE_POINTER1(hBand, "GDALIsLineOfSightVisible", false);
 
     // Perform a preliminary check of the start and end points.
     if (!IsAboveTerrain(hBand, xA, yA, zA)) {
--- a/alg/los.cpp
+++ b/alg/los.cpp
@@ -1,0 +1,94 @@
+/******************************************************************************
+ *
+ * Project:  Line of Sight
+ * Purpose:  Core algorithm implementation for line of sight algorithms.
+ * Author:   Ryan Friedman, ryanfriedman5410+gdal@gmail.com
+ *
+ ******************************************************************************
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a
+ * copy of this software and associated documentation files (the "Software"),
+ * to deal in the Software without restriction, including without limitation
+ * the rights to use, copy, modify, merge, publish, distribute, sublicense,
+ * and/or sell copies of the Software, and to permit persons to whom the
+ * Software is furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included
+ * in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+ * OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+ * THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ ****************************************************************************/
+
+// Test instructions:
+// * cmake -S . -B build -DBUILD_TESTING=ON -DCMAKE_BUILD_TYPE=Debug
+// * cmake --build build --target gdal_unit_test -j `nproc`
+// * ./build/autotest/cpp/gdal_unit_test --gtest_filter="test_alg.GDALIsLineOfSightVisible*"
+// Or, with GDB
+// *  gdb ./build/autotest/cpp/gdal_unit_test
+// * >>> r --gtest_filter="test_alg.GDALIsLineOfSightVisible*"
+
+#include "cpl_port.h"
+#include "gdal_alg.h"
+
+
+/************************************************************************/
+/*                        GDALIsLineOfSightVisible()                    */
+/************************************************************************/
+
+/**
+ * Check Line of Sight between two points.
+ *
+ * This algorithm will check line of sight using a 3D bresenham algorithm.
+ * Sightlines" published at
+ * https://www.researchgate.net/publication/2411280_Efficient_Line-of-Sight_Algorithms_for_Real_Terrain_Data
+ *
+ *
+ * @param hBand The band to read the DEM data from.
+ *
+ * @return True if the two points are within Line of Sight.
+ *
+ * @since GDAL 3.X
+ */
+
+bool GDALIsLineOfSightVisible(
+    GDALRasterBandH hBand, double xA, double yA, double zA, double xB, double yB, double zB, char** papszOptions)
+{
+    if(hBand == nullptr) {
+        return false;
+    }
+
+
+    // Always process as a double regardless of the input type.
+    double val;
+    const int iXa = static_cast<int>(xA);
+    const int iYa = static_cast<int>(yA);
+
+    if(GDALRasterIO(hBand, GF_Read, iXa, iYa, 1, 1, &val, 1, 1, GDT_Float64, 0, 0) != CE_None) {
+        return false;
+    }
+    if (zA < val) {
+        return false;
+    }
+
+    const int iXb = static_cast<int>(xB);
+    const int iYb = static_cast<int>(yB);
+
+    if(GDALRasterIO(hBand, GF_Read, iXb, iYb, 1, 1, &val, 1, 1, GDT_Float64, 0, 0) != CE_None) {
+        return false;
+    }
+    if (zB < val) {
+        return false;
+    }
+
+
+    // Perform a preliminary check of the start and end points first to check they are in LOS.
+    // ASSERT_TRUE(hBand->RasterIO(GF_Write, 0, 0, 1, 1, &val, 1, 1, GDT_Byte, 1, nullptr, 0, 0, 0, nullptr) == CE_None);
+
+    return true;
+}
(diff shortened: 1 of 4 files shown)


autotest/cpp/test_alg.cpp
```diff
@@ -273,4 +275,125 @@ TEST_F(test_alg, GDALAutoCreateWarpedVRT_alpha_band)
...
+{
+    auto const sizeX = 1;
+    auto const sizeY = 1;
+    auto const numBands = 1;
+    GDALDatasetUniquePtr poDS(
+        GDALDriver::FromHandle(GDALGetDriverByName("MEM"))
+            ->Create("", sizeX, sizeY, numBands, GDT_Int8, nullptr));
+    ASSERT_TRUE(poDS != nullptr);
+
+    uint8_t val = 42;
```
rouault, line 296: to be consistent with GDT_Int8
```suggestion
    int8_t val = 42;
```

autotest/cpp/test_alg.cpp
```diff
@@ -273,4 +275,125 @@ TEST_F(test_alg, GDALAutoCreateWarpedVRT_alpha_band)
...
+
+// Test GDALIsLineOfSightVisible() through a mountain (not a unit test)
+TEST_F(test_alg, GDALIsLineOfSightVisible_through_mountain)
+{
+    GDALAllRegister();
+    const std::string path = "/vsizip/vsicurl/https://terrain.ardupilot.org/"
+                             "SRTM1/ap_srtm1.zip/ap_srtm1.vrt";
+    const auto poDS = GDALDatasetUniquePtr(
+        GDALDataset::FromHandle(GDALOpen(path.c_str(), GA_ReadOnly)));
+    ASSERT_TRUE(poDS != nullptr);
```
rouault, line 361: This will not work on builds without curl support, and when network is off. So (something like)

```suggestion
    if (!poDS )
    {
        GTEST_SKIP() << "Cannot open << path;
    }
```

rouault: Alternatively we have a autotest/gdrivers/data/dted/n43_wgs72.dt0 dataset. Maybe you could try to find a good example with it, so we don't depend on curl? (for minimal builds the DTED driver might be unavailable, so we should gracefully skip if the dataset opening fails)

alg/los.cpp
```diff
@@ -0,0 +1,224 @@
...
+
+
+// Get the elevation of a single point.
+bool GetElevation(const GDALRasterBandH hBand, const int x, const int y, double& val)
+{
+    CPLAssert(x >= 0);
+    CPLAssert(x <= GDALGetRasterBandXSize(hBand));
+    CPLAssert(y >= 0);
+    CPLAssert(y <= GDALGetRasterBandYSize(hBand));
+
```
rouault, line 142: GDALRasterIO() already validates that, so those checks can be removed:
```suggestion
```

alg/los.cpp
```diff
@@ -0,0 +1,224 @@
...
+    // Lambda for Linear interpolate like C++20 std::lerp.
+    auto lerp = [](const int a, const int b, const float t) {
+        return a + t * (b - a);
+    };
+
+    // Lambda for getting Z test height given x input along the bresenham line.
+    auto GetZValue = [&](const int x) -> double {
+        const auto xDiff = xB - xA;
+        const auto xPercent = x / xDiff;
```
rouault, line 211: xB and xA are int, so xDiff is int.  x is int as well, so x / xDiff is an integer division, which doesn't seem to be what you want. There should be a cast of one member to float, so presumably:
```suggestion
        const auto xPercent = float(x) / xDiff;
```

alg/los.cpp
```diff
@@ -0,0 +1,224 @@
...
+		incy = -1;
+	}
+
+	auto x = x1;
+	auto y = y1;
+    int	balance;
+
+	if (dx >= dy)
+	{
+		dy <<= 1;
```
rouault, line 94: clearer and as efficient (the optimizer will generate the left shift for us)
```suggestion
		dy *= 2;
```

alg/los.cpp
```diff
@@ -0,0 +1,224 @@
...
+		balance = dy - dx;
+		dx <<= 1;
```
rouault, line 96:
```suggestion
		dx *= 2;
```

alg/los.cpp
```diff
@@ -0,0 +1,224 @@
...
+			}
+			balance += dy;
+			x += incx;
+		} 
+        isAboveTerrain &= OnBresenhamPoint(x, y);
+
+	}
+	else
+	{
+		dx <<= 1;
```
rouault, line 114:
```suggestion
		dx *= 2;
```

alg/los.cpp
```diff
@@ -0,0 +1,224 @@
...
+		balance = dx - dy;
+		dy <<= 1;
```
rouault, line 116:
```suggestion
		dy *= 2;
```

alg/los.cpp
```diff
@@ -0,0 +1,224 @@
...
+
+// Get the elevation of a single point.
+bool GetElevation(const GDALRasterBandH hBand, const int x, const int y, double& val)
+{
+    CPLAssert(x >= 0);
+    CPLAssert(x <= GDALGetRasterBandXSize(hBand));
+    CPLAssert(y >= 0);
+    CPLAssert(y <= GDALGetRasterBandYSize(hBand));
+
+    return GDALRasterIO(hBand, GF_Read, x, y, 1, 1, &val, 1, 1, GDT_Float64, 0, 0) == CE_None;
```
rouault, line 143: I suspect using GDALCachedPixelAccessor (gcore/gdalcachedpixelaccessor.h ) would lead to better performance. Single pixel picking with GDALRasterIO() can have pretty big penalty. But could be done as a follow up pull request

alg/los.cpp
Ryanf55: nice. The left shift seems like a hold-over from much older compilers that couldn't optimize it.

alg/los.cpp
Ryanf55: Great, thanks.

alg/los.cpp
Ryanf55: Thanks for the tip. I'll try it out if it's easy.

alg/los.cpp
Ryanf55: yea, I'll fix this. When it was all double math earlier, it was fine.

autotest/cpp/test_alg.cpp
Ryanf55: Thanks! Yea, I just wanted to know a good dataset to use with sufficient enough height difference. I'll swap it over.

alg/los.cpp
```diff
@@ -0,0 +1,212 @@
...
+/*                        GDALIsLineOfSightVisible()                    */
+/************************************************************************/
+
+/**
+ * Check Line of Sight between two points.
+ * Both input coordinates must be within the raster coordinate bounds.
+ *
+ * This algorithm will check line of sight using a 3D bresenham algorithm.
+ * https://www.researchgate.net/publication/2411280_Efficient_Line-of-Sight_Algorithms_for_Real_Terrain_Data
+ *
```
rouault, line 157: A comment could be added to state that the line of sight is computed in raster coordinate space, and thus may not be appropriate for example on datasets referenced against geographic coordinate at high latitudes.

alg/los.cpp
```diff
@@ -0,0 +1,212 @@
...
+
+/************************************************************************/
+/*                        GDALIsLineOfSightVisible()                    */
+/************************************************************************/
+
+/**
+ * Check Line of Sight between two points.
+ * Both input coordinates must be within the raster coordinate bounds.
+ *
+ * This algorithm will check line of sight using a 3D bresenham algorithm.
```
rouault, line 155:
```suggestion
 * This algorithm will check line of sight using a 3D Bresenham algorithm.
```

rouault:
compilers complain about:
```
/home/runner/work/gdal/gdal/alg/los.cpp:200:25: error: implicit conversion turns floating-point number into integer: 'const double' to 'int' [-Werror,-Wfloat-conversion]
        return lerp(zA, zB, xPercent);
               ~~~~     ^~
/home/runner/work/gdal/gdal/alg/los.cpp:200:21: error: implicit conversion turns floating-point number into integer: 'const double' to 'int' [-Werror,-Wfloat-conversion]
        return lerp(zA, zB, xPercent);
               ~~~~ ^~
```

alg/los.cpp
Ryanf55: For now, I added a TODO. I'd like to get some more representative test scenarios written up that I can use to benchmark it, but don't think it's necessary for the initial merge.

alg/los.cpp
Ryanf55: Add in cb79c61e07d4e80fc096b6b8840a4f061a9e65c5. 

If there are other approaches to do LOS that would work in these scenarios and we can measure the inaccuracy, please share some references and I can do some research.

alg/los.cpp
```diff
@@ -0,0 +1,232 @@
...
+    // Lambda for Linear interpolate like C++20 std::lerp.
+    auto lerp = [](const int a, const int b, const float t)
+    { return a + t * (b - a); };
+
+    // Lambda for getting Z test height given x input along the bresenham line.
+    auto GetZValue = [&](const int x) -> double
+    {
+        const auto xDiff = xB - xA;
+        const float xPercent = static_cast<float>(x) / xDiff;
+        return lerp(static_cast<int>(zA), static_cast<int>(zB), xPercent);
```
rouault, line 221: why casting zA and zB to int ? Shouldn't the interpolation be done on double to be consistant. And maybe use a double for xPercent too. In general, float should be rarely used compared to using double, unless there are good reasons to do so (memory savings or processing speed on big arrays typically, which isn't hte case here)

autotest/cpp/test_alg.cpp
```diff
@@ -273,4 +275,121 @@ TEST_F(test_alg, GDALAutoCreateWarpedVRT_alpha_band)
...
+            ->Create("", sizeX, sizeY, numBands, GDT_Int8, nullptr));
+    ASSERT_TRUE(poDS != nullptr);
+
+    auto pBand = poDS->GetRasterBand(1);
+    ASSERT_TRUE(pBand != nullptr);
+
+    const double x1 = 1;
+    const double y1 = 1;
+    const double x2 = 2;
+    const double y2 = 2;
```
rouault, line 322:
```suggestion
    const int x1 = 1;
    const int y1 = 1;
    const int x2 = 2;
    const int y2 = 2;
```

alg/los.cpp
Ryanf55: Oh duh, `lerp` should just be changed to take in doubles.

Ryanf55: Fixed!

jratike80:
> Run bresenham terrain checking from (x1, y1) to (x2, y2).
> The callback is run at every point along the line,
> which should return True if the point is above terrain.
> Bresenham2D will return true if all points have LOS between the start and end.
> 

Just wondering that if the DEM pixels between x and y are tested in order, could it be possible to report the pixel that breaks the line-of-sight? I mean the location "o" in this illustration. Maybe that information could be useful sometimes.
`x1,y1------------o           x2,y2`

Ryanf55:
> > Run bresenham terrain checking from (x1, y1) to (x2, y2).
> > The callback is run at every point along the line,
> > which should return True if the point is above terrain.
> > Bresenham2D will return true if all points have LOS between the start and end.
> 
> Just wondering that if the DEM pixels between (x1,y1) and (x2,y2y) are tested in order, could it be possible to report the pixel that breaks the line-of-sight? I mean the location "o" in this illustration. Maybe that information could be useful sometimes. `(x1,y1)------------o (x2,y2)`

yea, seems reasonable. I can add a pass-by-reference argument for the violation pixel. As far as the order.

rouault:
> a pass-by-reference argument for the violation pixel.

should rather be a ``int *pnStopX, int *pnStopY`` to be compatible of a C API.

Ryanf55:
> > a pass-by-reference argument for the violation pixel.
> 
> should rather be a `int *pnStopX, int *pnStopY` to be compatible of a C API.

Shall I default those to nullptr to have it treated as optional?

alg/los.cpp
Ryanf55: Resolved.

rouault:
> Shall I default those to nullptr to have it treated as optional?

yes, it is common to allow the passed pointers to be null, if the user isn't interested in that output

autotest/cpp/test_alg.cpp
```diff
@@ -273,4 +275,125 @@ TEST_F(test_alg, GDALAutoCreateWarpedVRT_alpha_band)
     GDALClose(hWarpedVRT);
 }
 
-}  // namespace
+// Test GDALIsLineOfSightVisible() reject null dataset
+TEST_F(test_alg, GDALIsLineOfSightVisible_null_dataset)
+{
+    EXPECT_FALSE(
+        GDALIsLineOfSightVisible(nullptr, 0, 0, 0.0, 0, 0, 0.0, nullptr));
+}
```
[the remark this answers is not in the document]
Ryanf55: It's now documented.

Ryanf55:
It took a bit of time, but I found some somewhat obvious hills in the `n43.dt0` dataset.
`gdallocationinfo autotest/cpp/data/n43.dt0 -wgs84 -79.9754364 43.8029960`. 

That should be enough to make some tests. 

In order to test all 8 cases of the bresenham algorithm (shallow steep, 4 quadrants) of the algorithm, it would be much easier to download a dataset that has a volcano looking dataset, such as Mt St Helens.

Would it be ok to add one more DTED dataset to the `autotest/cpp/data/` folder? If you are concerned about the space, I could zip that tile.

Ryanf55:
> > > Run bresenham terrain checking from (x1, y1) to (x2, y2).
> > > The callback is run at every point along the line,
> > > which should return True if the point is above terrain.
> > > Bresenham2D will return true if all points have LOS between the start and end.
> > 
> > 
> > Just wondering that if the DEM pixels between (x1,y1) and (x2,y2y) are tested in order, could it be possible to report the pixel that breaks the line-of-sight? I mean the location "o" in this illustration. Maybe that information could be useful sometimes. `(x1,y1)------------o (x2,y2)`
> 
> yea, seems reasonable. I can add a pass-by-reference argument for the violation pixel. As far as the order, that's probably an algorithm detail. Bresenham does them in order.

Could I actually save this for a follow up PR? I think I've done everything intended for the first minimal version and would like to keep this first merge the bare minimum.

jratike80:
> Could I actually save this for a follow up PR? I think I've done everything intended for the first minimal version and would like to keep this first merge the bare minimum.

I think that starting from the bare minimum is the right approach.

rouault:
> Or perhaps crop the dataset and save as an alternative format like `tif`.

That would be preferred. We try to keep the size of data files in the main repository as small as possible (and also with licensing as close as possible to MIT). We've also started to use https://github.com/OSGeo/gdal-test-datasets but it should be as a least resort solution, since that still implies to download / use a remote resource from the test.

Ryanf55:
> > Or perhaps crop the dataset and save as an alternative format like `tif`.
> 
> That would be preferred. We try to keep the size of data files in the main repository as small as possible (and also with licensing as close as possible to MIT). We've also started to use https://github.com/OSGeo/gdal-test-datasets but it should be as a least resort solution, since that still implies to download / use a remote resource from the test.

Alrighty. I've taken the PR out of draft. Test coverage is already decent with the existing dataset. Let me know if you are happy with it all, and what you would like to see before final merge.

alg/los.cpp
```diff
@@ -0,0 +1,232 @@
...
+    // TODO if both X's or Y's are the same, it could be optimized for vertical/horizontal lines.
+
+    // Use an interpolated Z height with 2D bresenham for the remaining cases.
+
+    // Lambda for Linear interpolate like C++20 std::lerp.
+    auto lerp = [](const double a, const double b, const double t)
+    { return a + t * (b - a); };
+
+    // Lambda for getting Z test height given x input along the bresenham line.
+    auto GetZValue = [&](const int x) -> double
```
rouault, line 217: GetZValue with just the "x" argument will not work for LoS computations where x1 == x2. You'd likely need to pass (x,y), and compute the ratio ``sqrt((SQUARE(x - x1) + SQUARE(y - y1)) / (SQUARE(x2 - x1) + SQUARE(y2 - y1)))``

Ryanf55: Good good. I'll add a test for a vertical and horizontal line. 

Also, exact vertical or horizontal line is a special case that can be optimized (don't call bresenham, just iterate directly).

rouault: > Also, exact vertical or horizontal line is a special case

but my point would also be valid for an almost vertical line (that is when the ratio abs((x2-x1)/(y2-y1)) is low.
An alternative implementation could be to determine which of abs(x2-x1) and abs(y2-y1) is the greater, and compute the ration according to x if dx > dy, and y otherwise.

Ryanf55: Fixed in b9ebc37210b693c3f449dae659a8143af04510f5, and I added tests for vertical/horizontal cases in 248e0de42b2c9c872300465c5defdc0cf99c0b73.

Ryanf55: Using sqrt operators over and over in this manner sort of defeats the purpose of using bresenham, but I want to get benchmarking set up first before optimizing, and not in this PR if that's ok.

alg/los.cpp
```diff
@@ -0,0 +1,237 @@
...
+    {
+        return true;
+    }
+
+    // TODO if both X's or Y's are the same, it could be optimized for vertical/horizontal lines.
+
+    // Use an interpolated Z height with 2D bresenham for the remaining cases.
+
+    // Lambda for computing the square of a number
+    auto SQUARE = [](const int n) -> int { return std::pow(n, 2); };
```
rouault, line 213: I'm wondering how smart optimizers are regarding that.... Below is more obvious (and shorter :-)):
```suggestion
    auto SQUARE = [](const int n) -> int { return n * n ; };
```

rouault: hum, actually we should do all those distance computations on doubles. Because if someone tries that on a raster with 25,000 pixels wide, 25000**2 = 2,500,000,000 which exceeds INT_MAX

Ryanf55: It might depend what you enable and your compiler. https://stackoverflow.com/questions/2940367/what-is-more-efficient-using-pow-to-square-or-just-multiply-it-with-itself#:~:text=So%20it%20looks%20like%20the,case%20(60%20times%20quicker).

https://baptiste-wicht.com/posts/2017/09/cpp11-performance-tip-update-when-to-use-std-pow.html

I'll take the suggestion for now, but can re-evaluate when benchmarking is set up.

Ryanf55: Are you saying convert the indices to double before calling SQUARE, and make SQUARE operate on doubles?

rouault: > Are you saying convert the indices to double before calling SQUARE, and make SQUARE operate on doubles?

yes

Ryanf55: 0a447cb319e6a9f103a0f5a170d6d430119f57da now uses double math for all of GetZValue()

Ryanf55: I should add a test for large rasters to check nothing overflows...

Ryanf55: I tried doing a test with a dataset close to overflow size, and even if its a 1xint_max, it can't allocate the memory. The test just hangs forever. If you have a way to create a dataset that large in a test to check overflow, let me know.

rouault: Adding tests for large datasets is difficult, and we want the test suite to be as fast as possible. So I would put that aside.

Ryanf55: Alright. Ready for another pass then.

rouault (approved):

PR #9522, opened 2024-03-20
SWIG bindings: add gdal.IsLineOfSightVisible()

rouault:
CC @Ryanf55

It might be easier to continue doing Python tests  (possibly porting your existing C++ tests)


Ryanf55:
Thanks for the info. I'll keep this in mind for the future. Thanks for getting this work in.

PR #9523, opened 2024-03-21
LOS: Add special cases for LOS checks for vertical and horizontal lines

Ryanf55:
## What does this PR do?

* This adds optimizations for the Line of Sight algorithm if the query is vertical/horizontal.  Bresenham shouldn't be used, because it's slower than direct iteration over the raster.
* This avoids a bunch of sqrt options in GetZValueFromXY and should speed things up.
* Doing LOS checks for vertical/horizontal is common when algorithms use raster-aligned operations.
* If the Z interpolation didn't use a square root, it probably wouldn't be worth doing this change. 
* Adds more tests! 

Builds on https://github.com/OSGeo/gdal/pull/9506

## Tasklist

 - [x] Add test case(s)
 - [x] Add documentation (no need)
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS: Ubuntu 22
* Compiler: G++11

### DIFF
--- a/alg/los.cpp
+++ b/alg/los.cpp
@@ -205,19 +205,113 @@ bool GDALIsLineOfSightVisible(const GDALRasterBandH hBand, const int xA,
         return true;
     }
 
-    // TODO if both X's or Y's are the same, it could be optimized for vertical/horizontal lines.
+    // Lambda for Linear interpolate like C++20 std::lerp.
+    auto lerp = [](const double a, const double b, const double t)
+    { return a + t * (b - a); };
+
+    // Lambda for getting Z test height given y input along the LOS line.
+    // Only to be used for vertical line checks.
+    auto GetZValueFromY = [&](const int y) -> double
+    {
+        // A ratio of 0.0 corresponds to being at yA.
+        const auto ratio =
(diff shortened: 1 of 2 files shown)


alg/los.cpp
```diff
@@ -205,19 +205,113 @@ bool GDALIsLineOfSightVisible(const GDALRasterBandH hBand, const int xA,
...
+                if (!IsAboveTerrain(hBand, x, yA, zTest))
+                {
+                    return false;
+                }
+            }
+            return true;
+        }
+
+        // If we get here, it's a coding error.
+        return false;
```
Ryanf55, line 262: Could we also add an assertion error here somehow?

rouault: I feel it is better to change the if (xA > xB) to a else to avoid code that will never be reached. You may add a CPLAssert(xA != xB) at the beginning of the lambda to make the contract clear.

Ryanf55: Good good. I added asserts that express the intent of the lamba much better than a comment.

The unit tests need more coverage on these - I found a copy-paste bug in these lambdas that wasn't caught.
I'll add more comprehensive testing once I get the other feature with reporting the LOS violation location.

PR #9532, opened 2024-03-23
gdal2tiles: Fix case where --exclude still writes fully transparent tiles

kevinmehall:
The alpha channel is only checked for a fully transparent tile when the source region is of nonzero size. When there are zero pixels to copy, the tile will also be transparent and should be skipped with `--exclude`.

This occurs when the edge of the input is aligned with the tile grid. To reproduce the issue, create a GeoTIFF with:

```console
$ gdal_create -outsize 1536 1536 -a_srs EPSG:3857 -a_ullr -114.65554242953658 114.65554242953658 114.65554242953658 -114.65554242953658 -bands 4 -burn 0 -co COMPRESS=DEFLATE empty.tif
```

This will be fully transparent so should produce no tiles with `--exclude`, but currently produces tiles along one edge:

```console
$ python3 swig/python/gdal-utils/osgeo_utils/gdal2tiles.py --exclude --xyz -w none empty.tif -z 11-20 /tmp/tiled_out 
...
$ find /tmp/tiled_out/20
/tmp/tiled_out/20
/tmp/tiled_out/20/524288
/tmp/tiled_out/20/524288/524291.png
/tmp/tiled_out/20/524287
/tmp/tiled_out/20/524287/524291.png
/tmp/tiled_out/20/524285
/tmp/tiled_out/20/524285/524291.png
/tmp/tiled_out/20/524289
/tmp/tiled_out/20/524289/524291.png
/tmp/tiled_out/20/524286
/tmp/tiled_out/20/524286/524291.png
/tmp/tiled_out/20/524284
/tmp/tiled_out/20/524284/524291.png
/tmp/tiled_out/20/524290
/tmp/tiled_out/20/524290/524291.png
```

### DIFF
--- a/swig/python/gdal-utils/osgeo_utils/gdal2tiles.py
+++ b/swig/python/gdal-utils/osgeo_utils/gdal2tiles.py
@@ -1375,9 +1375,6 @@ def create_base_tile(tile_job_info: "TileJobInfo", tile_detail: "TileDetail") ->
             wysize,
             band_list=list(range(1, dataBandsCount + 1)),
         )
-    elif tile_job_info.exclude_transparent:
-        # If there are no pixels to copy, the tile will be fully transparent
-        return
 
     # The tile in memory is a transparent file by default. Write pixel values into it if
     # any
@@ -2803,6 +2800,13 @@ class GDAL2Tiles(object):
                         ry = ysize - (ty * tsize) - rysize
                         if wysize != self.tile_size:
                             wy = self.tile_size - wysize
+
+                if self.options.exclude_transparent and (
+                    rxsize == 0 or rysize == 0 or wxsize == 0 or wysize == 0
+                ):
+                    if self.options.verbose:
+                        logger.debug("\tExcluding tile with no pixel coverage")
+                    continue
 
                 # Read the source raster if anything is going inside the tile as per the computed
                 # geo_query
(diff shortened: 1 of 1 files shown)


rouault:
@kevinmehall Your fix is reasonable, but I'm wondering if an even better fix wouldn't be to add a check before https://github.com/OSGeo/gdal/blob/3dfece75d6a209ddcc24677104315d4d074bc256/swig/python/gdal-utils/osgeo_utils/gdal2tiles.py#L2807 to avoid adding such tiles at all into the tile_details[] list. The --exclude option was designed more to exclude fully transparent tiles "in the middle" of the covered area, rather than tiles at the edges that have no area at all.

kevinmehall:
Agreed. Added a commit (you can squash-merge them if you'd prefer).

I'm also not sure whether it should check `self.options.exclude_transparent` or be enabled unconditionally. I didn't dig into `geo_query` deep enough to tell whether this can result in a non-rectangular set of tiles, or if it will always cut off an entire row or column of tiles that is only there due to rounding differences.

swig/python/gdal-utils/osgeo_utils/gdal2tiles.py
```diff
@@ -2801,6 +2801,13 @@ def generate_base_tiles(self) -> Tuple[TileJobInfo, List[TileDetail]]:
                         if wysize != self.tile_size:
                             wy = self.tile_size - wysize
 
+                if self.options.exclude_transparent and (
+                    rxsize == 0 or rysize == 0 or wxsize == 0 or wysize == 0
+                ):
```
rouault, line 2806:
```suggestion
                if rxsize == 0 or rysize == 0 or wxsize == 0 or wysize == 0:
```

Issue #9467, opened 2024-03-15
Inconsistent Output Dimensions from gdal.warp() with Identical outputBounds and Projection Conversion


jaiindimple:
### What is the bug?

I am encountering an issue with gdal.warp() where the output dimensions are inconsistent when mosaicing two sets of images, despite using the same outputBounds and source projection (EPSG:32613) for all images, and converting them to EPSG:4326. The first set of images produces an output with one set of dimensions, while another set results in a different set of dimensions.

### Steps to reproduce the issue

1. Define the area of interest (AOI) as a polygon with specific coordinates.
2. Obtain two set of images from [here](https://drive.google.com/drive/folders/1yE34jV_uq_z5SZsOoaMVTIpo2QHoiP5G?usp=sharing) to create mosaic.
3. Use gdal.Warp with outputBounds set to the bounds of the AOI, converting dataset to EPSG:4326 on both set of datasets.

Here's the code snippet used for creating the mosaic:

```python
from osgeo import gdal
from shapely.geometry import shape

aoi_geometry = {"type": "Polygon", "coordinates": [ [
 [-106.40856573808874, 35.51543260935861 [-106.40856573808874, 35.10139620198477], [-105.92962613828232, 35.10139620198477], [-105.92962613828232, 35.51543260935861],  [-106.40856573808874, 35.51543260935861], ]],}

bbox = shape(aoi_geometry).bounds
sorted_mosaic_day_wise = ['/Users/Downloads/2024_03_03/T13SDU_20240303T174231_B02_10m.tif',
                          '/Users/Downloads/2024_03_03/T13SDV_20240303T174231_B02_10m.tif']

gdal.Warp(
    '/Users/Downloads/20240303_mosaiced.tif',
    sorted_mosaic_day_wise,
    options=gdal.WarpOptions(
        multithread=True,
        dstSRS="EPSG:4326",
        outputBounds=bbox,
        resampleAlg="lanczos",
        dstNodata=0,
        srcNodata=0,
        creationOptions=[
            "COMPRESS=LZW",
            "NUM_THREADS=ALL_CPUS" ],))
```

Expected Behavior
The output mosaic should have consistent dimensions across different sets of input images when using the same outputBounds and projection. 
Example:-  Width - 4793  Height	- 4143

Actual Behavior
[...]

PR #9573, opened 2024-03-27
WARP: fix resolution calc with -te

elpaso:
Fix issue #9467

### DIFF
--- a/apps/gdalwarp_lib.cpp
+++ b/apps/gdalwarp_lib.cpp
@@ -3896,8 +3896,6 @@ static GDALDatasetH GDALWarpCreateOutput(
                         // Transform target extent to source CRS
                         double dfMinX = psOptions->dfMinX;
                         double dfMinY = psOptions->dfMinY;
-                        double dfMaxX = psOptions->dfMaxX;
-                        double dfMaxY = psOptions->dfMaxY;
 
                         psRTI->poReverseTransform->Transform(
                             1, &dfMinX, &dfMinY, nullptr, nullptr);
--- a/apps/gdalwarp_lib.cpp
+++ b/apps/gdalwarp_lib.cpp
@@ -3830,7 +3830,7 @@ static GDALDatasetH GDALWarpCreateOutput(
 
         if (bKnownTargetExtentButNotResolution)
         {
-            // Sample points along a grid
+            // Sample points along a grid in target CRS
             constexpr int nPointsX = 10;
             constexpr int nPointsY = 10;
             constexpr int nPoints = 3 * nPointsX * nPointsY;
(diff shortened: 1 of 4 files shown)


apps/gdalwarp_lib.cpp
```diff
@@ -3873,22 +3873,72 @@ static GDALDatasetH GDALWarpCreateOutput(
...
+            {
+                const GDALReprojectionTransformInfo *psRTI =
+                    static_cast<const GDALReprojectionTransformInfo *>(
+                        psTransformInfo->pReprojectArg);
+                if (psRTI && psRTI->poReverseTransform)
+                {
+
+                    // Compute new geotransform from transformed target extent
+                    double adfGeoTransform[6];
+                    if (GDALGetGeoTransform(hSrcDS, adfGeoTransform) == CE_None)
```
rouault, line 3893: below code assumes that gt[2] and gt[5] are 0. We should check this assumption here

apps/gdalwarp_lib.cpp
```diff
@@ -3873,22 +3873,72 @@ static GDALDatasetH GDALWarpCreateOutput(
...
-            psInfo->pfnTransform(hTransformArg, TRUE, nPoints, &padfX[0],
-                                 &padfY[0], &padfZ[0], &pabSuccess[0]);
+            bool transformedToSrcCRS{false};
+
+            GDALGenImgProjTransformInfo *psTransformInfo{
+                static_cast<GDALGenImgProjTransformInfo *>(hTransformArg)};
+
+            // If a transformer is available, use an extent that covers the
+            // target extent instead of the real source image extent
+            if (psTransformInfo && psTransformInfo->pReprojectArg)
```
rouault, line 3883: We should also likely check that psTransformInfo->pSrcTransformer == nullptr to avoid the below logic to trigger when we use a image pixel -> source CRS coordinates transformer that is not geotransform-based (such as RPC, geolocation arrays, ...)

autotest/utilities/test_gdalwarp_lib.py
```diff
@@ -3171,7 +3171,7 @@ def test_gdalwarp_lib_automatic_grid_sampling():
         outputBounds=[-7655830, -6385994, 7152182, 8423302],
         dstSRS="+proj=laea +lat_0=48.514 +lon_0=-145.204 +x_0=0 +y_0=0 +datum=WGS84 +units=m +no_defs",
     )
-    assert ds.GetRasterBand(1).Checksum() == 46790
+    assert ds.GetRasterBand(1).Checksum() == 35573
```
rouault, line 3174: In addition to the checksum we should check the resolution returned by ds.GetGeoTransform()

autotest/utilities/test_gdalwarp_lib.py
```diff
@@ -4056,3 +4056,71 @@ def test_gdalwarp_lib_conflicting_source_metadata(tmp_vsimem):
...
+            dstSRS="EPSG:4326",
+            outputBounds=bbox,
+            resampleAlg="lanczos",
+            dstNodata=0,
+            srcNodata=0,
+            format="MEM",
+        ),
+    )
+
+    assert ds.RasterXSize == 4793
```
rouault, line 4099: In addition to the checksum we should check the resolution returned by ds.GetGeoTransform(). Same for second part of the test

apps/gdalwarp_lib.cpp
elpaso: did you mean gt[2] and gt[4] ?

autotest/utilities/test_gdalwarp_lib.py
elpaso: About this test, the resolution is very different with this patch but I checked the result visually and it looks ok to me.

GT without this patch (rounded to int):
(-7655830, 30850, 0.0, 8423302, 0.0, -30852)

GT with this patch (rounded to int):
(-7655830, 19433, 0.0, 8423302, 0.0, -19434)

do you think it is acceptable?

rouault: > do you think it is acceptable?

looks reasonable

apps/gdalwarp_lib.cpp
elpaso: done

apps/gdalwarp_lib.cpp
elpaso: done

PR #9574, opened 2024-03-27
windows_build.yml: disable KEA driver because of linking error

rouault:
Latest good build
(https://github.com/OSGeo/gdal/actions/runs/8442156700/job/23122964154) only found the C component of HDF5, and thus didn't enable KEA driver which requireds HDF5 CXX
```-- Found HDF5: D:/a/gdal/gdal/sdk/release-1930/lib/hdf5.lib (found version "1.12.0") found components: C```

whereas new failing builds
(https://github.com/OSGeo/gdal/actions/runs/8456021611/job/23164833883?pr=9572)
detect HDF5 CXX now, but libgdal fails to link with
```keadataset.obj : error LNK2019: unresolved external symbol "public: static class H5::FileAccPropList const & const
H5::FileAccPropList::DEFAULT" (?DEFAULT@FileAccPropList@H5@@2ABV12@B) referenced in function "public: static class GDALDataset * __cdecl KEADataset::Open(class GDALOpenInfo *)"


szekerest:
The latest update was a month ago and it was unrelated to this problem

PR #9579, opened 2024-03-28
Doc: Update Python API docs for FieldDefn, FieldDomain

dbaston:
## What does this PR do?

- Correct SWIG directives so that FieldDomain docs are included
- Split FieldDefn and FieldDomain docs into separate files
- Remove C signatures, types
- Add some examples

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments


swig/include/python/docs/ogr_fielddomain_docs.i
```diff
@@ -0,0 +1,236 @@
...
+--------
+>>> d = ogr.CreateCodedFieldDomain('my_code', None, ogr.OFTInteger, ogr.OFSTNone, { 1 : 'owned', 2 : 'leased' })
+>>> d.GetDomainType() == ogr.OFDT_CODED
+True
+
+";
+
+%feature("docstring")  GetEnumeration "
+
+Get the enumeration as (code, value) pairs.
```
rouault, line 65: while true for the C API, it is perhaps a bit confusing to speak about "pairs" in the context of the Python API, given that we use a Python dict? feel free to disregard as I feel I'm a bit nitpicking :-)

dbaston: "Get the enumeration as mapping of codes to values." ?

rouault: > Get the enumeration as mapping of codes to values." ?

if that sounds better for you, I'm happy with that too.

dbaston, line 65:
```suggestion
Get the enumeration as mapping of codes to values.
```

swig/include/python/docs/ogr_fielddomain_docs.i
```diff
@@ -0,0 +1,236 @@
...
+--------
+>>> d = ogr.CreateCodedFieldDomain('my_code', None, ogr.OFTInteger, ogr.OFSTNone, { 1 : 'owned', 2 : 'leased' })
+>>> d.GetDomainType() == ogr.OFDT_CODED
+True
+
+";
+
+%feature("docstring")  GetEnumeration "
+
+Get the enumeration as mapping of codes to values.
```
dbaston, line 65:
```suggestion
Get the enumeration as a mapping of codes to values.
```

PR #9582, opened 2024-03-29
Python: Add Band.ReadAsMaskedArray

dbaston:
Modeled after the method of the same name on `MDArray`.

I'm wondering if `NODATA` values should also be masked.


rouault:
> I'm wondering if `NODATA` values should also be masked.

they definitely should, and AFAICS they are already by your patch because self.GetMaskFlags() is equal to GMF_NODATA when GetNoDataValue() returns a valid value.

Issue #9562, opened 2024-03-25
LOS: Add support for reporting where the LOS query failed


Ryanf55:
# Purpose

Design the support for LOS query where the caller wants to know where LOS was violated in raster coordinate space.

# Original Request

              > Run bresenham terrain checking from (x1, y1) to (x2, y2).
> The callback is run at every point along the line,
> which should return True if the point is above terrain.
> Bresenham2D will return true if all points have LOS between the start and end.
> 

Just wondering that if the DEM pixels between (x1,y1) and (x2,y2y) are tested in order, could it be possible to report the pixel that breaks the line-of-sight? I mean the location "o" in this illustration. Maybe that information could be useful sometimes.
`(x1,y1)------------o           (x2,y2)`

_Originally posted by @jratike80 in https://github.com/OSGeo/gdal/issues/9506#issuecomment-2008166314_
 
 
 ## API Change
 
 I think API needs to change. Perhaps it should be broken before the release, otherwise a new function needs to be added.
 
 ### Current
 
 ```c
 bool CPL_DLL GDALIsLineOfSightVisible(const GDALRasterBandH, const int xA,
                                      const int yA, const double zA,
                                      const int xB, const int yB,
                                      const double zB,
                                      CSLConstList papszOptions);
```

 ### Proposed
 
 ```c
 bool CPL_DLL GDALIsLineOfSightVisible(const GDALRasterBandH, const int xA,
                                      const int yA, const double zA,
                                      const int xB, const int yB,
                                      const double zB,
                                      int* xTerrainIntersection, int* yTerrainIntersection,
                                      CSLConstList papszOptions);
```

1) Is there a better name than `Violation`? That makes it sound bad. Perhaps `yLosBreak`
2) Because bresenham may flip the direction of the line, the same LOS query on a raster rotated 180 degrees would result in a different location on the raster. If anyone knows of existing tools that report LOS violations, and what sort of data is reported, that would be helpful to design this feature. Note the current implementation is greedy - ie - it returns false as soon as the first point is found along the sight line that's below terrain. A UI graph might need to query all spots along the line anyways.

rouault:
Perhaps ``int* xTerrainIntersection, int *yTerrainIntersection`` (and in that x,y order for consistency). I believe it would be desirable if that was always the first point when going from A to B, for predictability of the result.

Ryanf55:
[...]

PR #9585, opened 2024-03-29
LOS: Add intersection location parameters

Ryanf55:
## What does this PR do?

* Lock in ABI before release to support x and y terrain intersection locations
* Implementation will come later
* Attempts to also add in the SWIG binding docs (Please verify, particularly the param types for the pointer args)

#9562 

## Tasklist

 - [ ] Verify there isn't anything special for pointer parameters in the python swig bindings (perhaps it should be a python callback...)
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS: Ubuntu 22
* Compiler: G++11

### DIFF
--- a/swig/include/python/docs/gdal_operations_docs.i
+++ b/swig/include/python/docs/gdal_operations_docs.i
@@ -24,6 +24,12 @@ yB : int
     The Y location (raster row) of the second point to check on the raster.
 zB : float
     The Z location (height) of the second point to check.
+xTerrainIntersection : int
+    The X location where the LOS line intersects with terrain, or nullptr if it does not
+    intersect terrain. Currently ignored.
+yTerrainIntersection : int
+    The Y location where the LOS line intersects with terrain, or nullptr if it does not
+    intersect terrain. Currently ignored.
 options : dict/list, optional
     A dict or list of name=value of options for the line of sight algorithm (currently ignored).
 
--- a/swig/include/python/docs/gdal_operations_docs.i
+++ b/swig/include/python/docs/gdal_operations_docs.i
@@ -24,17 +24,13 @@ yB : int
     The Y location (raster row) of the second point to check on the raster.
 zB : float
     The Z location (height) of the second point to check.
-xTerrainIntersection : int
-    The X location where the LOS line intersects with terrain, or nullptr if it does not
-    intersect terrain. Currently ignored.
-yTerrainIntersection : int
-    The Y location where the LOS line intersects with terrain, or nullptr if it does not
-    intersect terrain. Currently ignored.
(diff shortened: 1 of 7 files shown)


swig/include/python/docs/gdal_operations_docs.i
```diff
@@ -24,6 +24,12 @@ yB : int
     The Y location (raster row) of the second point to check on the raster.
 zB : float
     The Z location (height) of the second point to check.
+xTerrainIntersection : int
```
Ryanf55, line 27: Should this use a `typemap`? I'm new to SWIG.
Reference: https://www.swig.org/Doc1.3/Python.html#Python_nn18

See `31.9.3 Using typemaps to return arguments`

rouault: This is going a bit tricky to map to SWIG. Yes, we'd likely need a typemap per language. For Python, I guess we'd want to return a triple (is_visible, xInterection, yIntersection). 

So in the #ifdef SWIGPYTHON case, we'd probably want to have a wrapper like

void IsLineOfSightVisible(GDALRasterBandShadow *band,
                          int xA, int yA, double zA,
                          int xB, int yB, double zB,
                          bool* pbIntersects, int *xTerrainIntersection, int *yTerrainIntersection,
                          char** options = NULL)

and a (bool* pbIntersects, int *xTerrainIntersection, int *yTerrainIntersection) typemap. One of the closest example to take inspiration from would be the (int *pnxvalid, int *pnyvalid, int* pisvalid) typemap in swig/include/python/typemaps_python.i and used in swig/include/Band.i

For other languages, I would omit for now the int *xTerrainIntersection, int *yTerrainIntersection parameter, and just use the current master signature

rouault:
@Ryanf55 I've pushed a commit in your branch to fix the Python bindings

Ryanf55:
> @Ryanf55 I've pushed a commit in your branch to fix the Python bindings

Nice. I'm happy if you wanna merge, or I can make time tomorrow to test it before merge.

swig/include/python/docs/gdal_operations_docs.i
```diff
@@ -29,6 +29,8 @@ options : dict/list, optional
 
 Returns
 -------
-bool
-    True if the two points are within Line of Sight.
+(bool, int, int)
+    First value of tuple is True if the two points are within Line of Sight.
+    Second value is the X location where the LOS line intersects with terrain (will be set in the future, currently set to -1).
```
dbaston, line 34: I think it would be more clear to refer to "X location" as a "column index".

Returning a `NamedTuple` here might be more clear.

Issue #9541, opened 2024-03-24
gdalinfo: also offer DD not just DMS, else need -json...


jidanni:
### Feature description

```
Corner Coordinates:
Upper Left  (  440720.000, 3751320.000) (117d38'28.21"W, 33d54'8.47"N)
Lower Left  (  440720.000, 3720600.000) (117d38'20.79"W, 33d37'31.04"N)
Upper Right (  471440.000, 3751320.000) (117d18'32.07"W, 33d54'13.08"N)
Lower Right (  471440.000, 3720600.000) (117d18'28.50"W, 33d37'35.61"N)
Center      (  456080.000, 3735960.000) (117d28'27.39"W, 33d45'52.46"N)

```

### Additional context

There should be a way to tell it our eyes are more used to DD than  DMS, even if we can post-process it.

rouault:
You can get that by inspecting the wgs84Extent member of gdalinfo -json

PR #9594, opened 2024-04-02
Update gdalinfo.rst to say how to get DD format

jidanni:
For https://github.com/OSGeo/gdal/issues/9541 .

### DIFF
--- a/doc/source/programs/gdalinfo.rst
+++ b/doc/source/programs/gdalinfo.rst
@@ -196,6 +196,8 @@ Example
     Center      (  456080.000, 3735960.000) (117d28'27.39"W, 33d45'52.46"N)
     Band 1 Block=512x16 Type=Byte, ColorInterp=Gray
 
+For DD coordinates instead of DMS, inspect the wgs84Extent member of gdalinfo -json:
+
 Example of JSON output with ``gdalinfo -json byte.tif``
 
 .. code-block:: json
--- a/doc/source/programs/gdalinfo.rst
+++ b/doc/source/programs/gdalinfo.rst
@@ -196,7 +196,7 @@ Example
     Center      (  456080.000, 3735960.000) (117d28'27.39"W, 33d45'52.46"N)
     Band 1 Block=512x16 Type=Byte, ColorInterp=Gray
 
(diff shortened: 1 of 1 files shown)


doc/source/programs/gdalinfo.rst
```diff
@@ -196,6 +196,8 @@ Example
     Center      (  456080.000, 3735960.000) (117d28'27.39"W, 33d45'52.46"N)
     Band 1 Block=512x16 Type=Byte, ColorInterp=Gray
 
+For DD coordinates instead of DMS, inspect the wgs84Extent member of gdalinfo -json:
+
```
jratike80, line 200: That sounds like the json format exists mainly for showing extent in DD, while I think that the main advantage is that json is computer-friendly.
I wonder if there is a good reason for DMS, or if decimal degrees could be used always.

jidanni: Yeah I kind of "just happened to squeeze the sentence into the right spot", " perhaps giving that effect."

On the other hand we don't want to get rid of DMS completely. Maybe some people still like it. Gives sort of like that early mariners feeling. Good for historical publishing. 

Anyway I just wanted the man page to sort of say something for people who are baffled why didn't the authors ever think of using DD format.

But you're entirely welcome to change it anyway you like because I'm sure you'll come up with a better way than me. 

I just didn't want to get my hands dirty to do any actual programming.

doc/source/programs/gdalinfo.rst
```diff
@@ -196,6 +196,8 @@ Example
     Center      (  456080.000, 3735960.000) (117d28'27.39"W, 33d45'52.46"N)
     Band 1 Block=512x16 Type=Byte, ColorInterp=Gray
 
+For DD coordinates instead of DMS, inspect the wgs84Extent member of gdalinfo -json:
```
rouault, line 199: It might not be ob

rouault, line 199:
```suggestion
For corner coordinates formatted as decimal degree instead of the above degree, minute, second, inspect the ``wgs84Extent`` member of gdalinfo -json:
```

Issue #9580, opened 2024-03-28
GPKG: srs_id -1 for "Undefined cartesian SRS" has meter units, should this be "undefined" units?


theroggy:
### What is the bug?

According to the GeoPackage specs, data in an "Undefined Cartesian SRS" should be saved with srs_id = -1: https://www.geopackage.org/spec/#r11.

In GDAL this was implemented in these PR's:
- reading: https://github.com/OSGeo/gdal/pull/3286
- writing: https://github.com/OSGeo/gdal/pull/3312

However, for reading as well as writing srs_id=-1 the unit is set/expected to be "meter". So, only if the unit is meter, the srs_id in GPKG will be set to srs_id -1.

For information, in r-spatial (https://github.com/r-spatial/sf/issues/2049) the choice was made to explicitly to use a WKT that specifies the units to be undefined when saving data without crs specified to GPKG, leading to a "custom CRS" being saved in the GPKG, so an srs_id >= 100.000 instead of -1, which doesn't seem to be according to the GPKG specs?

So following questions rise:
- Was it an active choice to use unit "meter"?
- Would it be more logical that for an "Undefined Cartesian SRS" in the spirit of the GPKG spec that the unit would also be "undefined"?

Issue surfaced when implementing support for "Undefined Cartesian SRS" in pyogrio: https://github.com/geopandas/pyogrio/pull/368

### Steps to reproduce the issue

xref: https://github.com/geopandas/pyogrio/pull/368

### Versions and provenance

All versions of GDAL

CC: @edzer , @martinfleis, @jorisvandenbossche

rouault:
It is tricky to do better than what we do currently. GDAL uses PROJ bases CRS models, which are OGC Topic 2 based, and for Unknown cartesian CRS we have opted for the rather loose EngineeringCRS representation, but an EngineeringCRS is abased on a Cartesian coordinate system, and the axis of such system *must* have a unit. Opting for metre is the obvious default. The issue fundamentally staims at the specification, which defines a "Unknown cartesian CRS" concept but which has no clear equivalent in OGC Topic 2.

[...]

PR #9595, opened 2024-04-02
GPKG: map Null CRS to a new srs_id=99999, and add a SRID layer creation option

rouault:
New layer creation option:
```
-  .. lco:: SRID
      :choices: <integer>
      :since: 3.9

      Forced ``srs_id`` of the entry in the ``gpkg_spatial_ref_sys`` table to point to.
      This may be -1 ("Undefined Cartesian SRS"), 0 ("Undefined geographic SRS"),
      99999 ("Undefined SRS"), a valid EPSG CRS code or an existing entry of the
      ``gpkg_spatial_ref_sys`` table. If pointing to a non-existing entry, only a warning
      will be emitted.
```

```
Coordinate Reference Systems
----------------------------

Starting with GDAL 3.9, a layer without any explicit CRS is mapped from/to a
custom entry of srs_id=99999 with the following properties:

- ``srs_name``: ``Undefined SRS``
- ``organization``: ``GDAL``
- ``organization_coordsys_id``: 99999
- ``definition``: ``LOCAL_CS["Undefined SRS",LOCAL_DATUM["unknown",32767],UNIT["unknown",0],AXIS["Easting",EAST],AXIS["Northing",NORTH]]``
- ``definition_12_063`` (when the CRS WKT extension is used): ``ENGCRS["Undefined SRS",EDATUM["unknown"],CS[Cartesian,2],AXIS["easting",east,ORDER[1],LENGTHUNIT["unknown",0]],AXIS["northing",north,ORDER[2],LENGTHUNIT["unknown",0]]]``
- ``description``: ``Custom undefined coordinate reference system``

Note that the use of a LOCAL_CS / EngineeringCRS is mostly to provide a valid
CRS definition to comply with the requirements of the GeoPackage specification
and to be compatible of other applications (or GDAL 3.8 or earlier), but the
semantics of that entry is intented to be "undefined SRS of any kind".
```


jratike80:
Is it good to use different `srs_id=99999` and `organization_coordsys_id=0`? There may be a slight possibility for a mess because geopackage developers can think that those two are always equal, because usually for the EPSG codes they are. But if someone cares about the organization then hopefully they check correctly the combination organization + organization_coordsys_id and notice that NONE:0 is not the same than GDAL:0.
_(Well, I think I wouldn't care about that difference myself, and at least not between NONE:-1 and GDAL:0)_

Maybe the definition could somehow distinct the new undefined from the two existing undefined variants. Something like "Custom undefined coordinate reference system". Or "Undefined coordinate reference system using undefined units". This suggestion does not feel most clear.

```
0 (NONE 0) = undefined Cartesian coordinate reference system
-1 (NONE -1) = undefined geographic coordinate reference system
99999 (GDAL 0) = undefined coordinate reference system
```

Was there someone who wished to have undefined undefined Cartesian and undefined undefined non-Cartesian?

rouault:
> Is it good to use different `srs_id=99999` and `organization_coordsys_id=0`?

ok, I've change organization_coordsys_id to 99999 as well

> Something like "Custom undefined coordinate reference system"

I've change to that

rouault:
Fixes https://github.com/OSGeo/gdal/issues/9580

PR #9598, opened 2024-04-02
VRT: add a new mode to apply chained processing steps that apply to several bands at the same time

rouault:
The following built-in algorithms are introduced, and typically applied in the following order:
- Dehazing: remove haze effects by applying (subsampled) gain and offset auxiliary datasets.
- BandAffineCombination: to perform an affine transformation combination of bands.
- Trimming: local thresholding of saturation
- LUT: apply a look-up table (band per band)

### DIFF
--- a/autotest/gdrivers/vrtprocesseddataset.py
+++ b/autotest/gdrivers/vrtprocesseddataset.py
@@ -1,0 +1,1258 @@
+#!/usr/bin/env pytest
+###############################################################################
+# $Id$
+#
+# Project:  GDAL/OGR Test Suite
+# Purpose:  Test VRTProcessedDataset support.
+# Author:   Even Rouault <even.rouault at spatialys.com>
+#
+###############################################################################
+# Copyright (c) 2024, Even Rouault <even.rouault at spatialys.com>
+#
+# Permission is hereby granted, free of charge, to any person obtaining a
+# copy of this software and associated documentation files (the "Software"),
+# to deal in the Software without restriction, including without limitation
+# the rights to use, copy, modify, merge, publish, distribute, sublicense,
+# and/or sell copies of the Software, and to permit persons to whom the
+# Software is furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included
+# in all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+# OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+# THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+# FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+# DEALINGS IN THE SOFTWARE.
+###############################################################################
+
+import struct
+
+import gdaltest
+import pytest
+
+from osgeo import gdal
+
+###############################################################################
+# Test error cases in general VRTProcessedDataset XML structure
+
+
+def test_vrtprocesseddataset_errors(tmp_vsimem):
+
+    with pytest.raises(Exception, match="Input element missing"):
+        gdal.Open(
+            """<VRTDataset subclass='VRTProcessedDataset'>
+                    </VRTDataset>
+                    """
+        )
+
+    with pytest.raises(
+        Exception,
+        match="Input element should have a SourceFilename or VRTDataset element",
+    ):
+        gdal.Open(
+            """<VRTDataset subclass='VRTProcessedDataset'>
+                    <Input/>
+                    </VRTDataset>
+                    """
+        )
+
+    with pytest.raises(Exception):  # "No such file or directory'", but O/S dependent
+        gdal.Open(
+            """<VRTDataset subclass='VRTProcessedDataset'>
+                    <Input><SourceFilename/></Input>
+                    </VRTDataset>
+                    """
+        )
+
+    with pytest.raises(
+        Exception,
+        match="Missing one of rasterXSize, rasterYSize or bands on VRTDataset",
+    ):
+        gdal.Open(
+            """<VRTDataset subclass='VRTProcessedDataset'>
+                    <Input><VRTDataset/></Input>
+                    </VRTDataset>
+                    """
+        )
+
+    src_filename = str(tmp_vsimem / "src.tif")
+    src_ds = gdal.GetDriverByName("GTiff").Create(src_filename, 10, 5, 3)
+    src_ds.GetRasterBand(1).Fill(1)
+    src_ds.GetRasterBand(2).Fill(2)
+    src_ds.GetRasterBand(3).Fill(3)
+    src_ds.Close()
+
+    with pytest.raises(Exception, match="ProcessingSteps element missing"):
+        gdal.Open(
+            f"""<VRTDataset subclass='VRTProcessedDataset'>
+        <Input>
+            <SourceFilename>{src_filename}</SourceFilename>
+        </Input>
+        </VRTDataset>
+            """
+        )
+
+    with pytest.raises(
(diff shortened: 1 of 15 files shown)


doc/source/drivers/raster/vrt_processed_dataset.rst
```diff
@@ -0,0 +1,261 @@
...
+LUT
+---
+
+Apply a look-up table (band per band), typically to get from UInt16 to Byte data types.
+
+The following required argument must be specified:
+
+- ``lut_{band}``: List of the form ``[src value 1]:[dest value 1],[src value 2]:[dest value 2],....``. {band} must be replaced by 1 to the number of bands.
+
+
```
dbaston, line 256: Note that interpolation will be used if pixel values are not found in the LUT. (maybe make this opt-out?)

doc/source/drivers/raster/vrt_processed_dataset.rst
```diff
@@ -0,0 +1,261 @@
...
+- ``top_rgb``: Maximum saturating RGB output value.
+
+- ``tone_ceil``: Maximum threshold beyond which we give up saturation.
+
+- ``top_margin``: Margin to allow for dynamics in brighest areas (between 0 and 1, should be close to 0)
+
+
+The following optional arguments may be specified:
+
+- ``relative_filenames``: Whether the trimming dataset filename is relative to the VRT. Allowed values are ``true`` and ``false``. Defaults to ``false``
```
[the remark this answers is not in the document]
rouault: ok, changed to that

doc/source/drivers/raster/vrt_processed_dataset.rst
```diff
@@ -0,0 +1,261 @@
...
+- ``green_band``: Index (one-based) of the green band. Defaults to 1.
+
+- ``blue_band``: Index (one-based) of the blue band. Defaults to 1.
+
+- ``nodata``: Override the input nodata value coming from the previous step (or the input dataset for the first step).
+
+- ``trimming_nodata``: Override the nodata value coming from the trimming dataset.
+
+
+LUT
```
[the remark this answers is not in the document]
rouault: > To me this raises the question of whether pixel functions can/should also be valid processing steps.

There's indeed some similarity. I didn't want to go into that complication for now, but that could probably be a further enhancement

doc/source/drivers/raster/vrt_processed_dataset.rst
rouault: I'm not sure to have understood your comment. What happens here is the same as the LUT option of a VRTComplexSource, that is if the input value is lower than the lowest src value 1, the output value is dest value 1. And if the input value is greater than the maximum src value N, the output value is dest value N. So this clamps to those bounds

dbaston: I mean that that with LUT `1:10,3:30`, I would have expected a pixel value of `2` to either remain `2` or end up as nodata (these are the two choices in https://docs.qgis.org/3.34/en/docs/user_manual/processing_algs/qgis/rasteranalysis.html#reclassify-by-table)

So it would be good to clarify that the algorithm performs linear interpolation if a pixel value is not found in the LUT.

rouault: > So it would be good to clarify that the algorithm performs linear interpolation if a pixel value is not found in the LUT.

ok, got it. Done per 07b6ab5ce6b15ed4fe94072762266a3acc553898

sgillies:
@rouault @dbaston Is this a feature being bought by a specific customer? I would guess so from the desire to do haze processing instead of a more generalized process. Can the functionality not be covered by the existing pixel functions or Python plugin? I'm ever uncomfortable with GDAL having lots of partially overlapping features.

rouault:
> Is this a feature being bought by a specific customer?

yes

>  I would guess so from the desire to do haze processing instead of a more generalized process. 

The initial processing functions are admittedly a bit specific to the specific use case, but the VRTProcessedDataset mechanism intends to be a generic framework where more processing steps could be added.

> Can the functionality not be covered by the existing pixel functions 

Not really, or at least not efficiently. The pixel functions are designed to operate band per band. Here some of the processing steps (namely dehazing and bandAffineCombination) involve mixing together pixel values from multiple bands to generate output values for multiple bands. That would be tedious to do with standard VRT pixel functions and would involve more I/O.

> or Python plugin?

There's a need for efficiency, and a Python plugin would likely run into issues with the GIL.

frmts/vrt/vrtprocesseddataset.cpp
```diff
@@ -0,0 +1,1342 @@
...
+    {
+        CPLError(CE_Failure, CPLE_OutOfMemory,
+                 "Out of memory allocating working buffer");
+        return false;
+    }
+
+    if (m_poSrcDS->RasterIO(GF_Read, nXOff, nYOff, nBufXSize, nBufYSize,
+                            abyInput.data(), nBufXSize, nBufYSize, eFirstDT,
+                            nFirstBandCount, nullptr,
+                            nFirstDTSize * nFirstBandCount,
```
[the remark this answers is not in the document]
rouault: suggestion applied

frmts/vrt/vrtprocesseddataset.cpp
```diff
@@ -0,0 +1,1342 @@
...
+                            nFirstDTSize * nFirstBandCount * nBufXSize,
```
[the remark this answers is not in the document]
rouault: suggestion applied

rouault:
@sgillies I'm not sure how to best address your concerns. I can see several potential ways forward:
- back out that PR in its whole, and handle that as a out-of-tree patch by the user who ordered this work. Probably not the preferred way from his point of view
- variant of the above that wouldn't require patching GDAL: try to make the new VRTProcessedDataset framework itself something that can be registered at run-time / a plugin. That is: make it possible to dynamically register a subclass of VRTDataset. However I'm not totally sure this (the changes in vrtdataset.cpp itself, mostly related to routing XML deserialization/serialization to the VRTProcessedDataset class) can be done at all or in a not-too-ugly way, without actually trying to do it, but I'm willing to give that a try if that'd be the preferred way
- just provide the framework, without the 4 initial built-in algorithms, since they can be registered by a plugin. There's no technical difficulty, since the built-in algorithms use the registration mechanism. However I don't really see this as a desirable solution, because it would make the testing of the framework itself in core GDAL impossible.
- better cross-documentation in the VRT driver doc between the existing VRTDerivedBand + pixel functions and the new VRTProcessedDataset explaining how they differ from each other

Any suggestion?

One item I didn't list above is "try reusing the existing VRTDerivedBand framework to cover the use cases of this PR" as this isn't a workable technical solution because of the points I raised in https://github.com/OSGeo/gdal/pull/9598#issuecomment-2032780187:
- need to compute efficiently several output bands from several input bands at once. Using standard pixel functions would require multiple processing of the same intermediate buffers, with computation costs that grow up exponentially with the number of steps
- easy of defining multiple steps in the same VRT file, whereas VRTDerivedBand would require one VRT per step
- something I didn't mention: need to provide to the algorithm the georeferenced window of interest so it can use auxiliary datasets (as done by the LocalScaleOffset and Trimming algorithms). The signature of existing VRTDerivedBand pixel functions doesn't allow this

sgillies:
@rouault I think this is the kind of feature that users should build for themselves. But you have much more experience maintaining a big open source project than I do, so I'll get over my concerns.

autotest/gdrivers/vrtprocesseddataset.py
```diff
@@ -0,0 +1,1258 @@
...
+            <Argument name="gain_dataset_filename_1">{gain_filename}</Argument>
+            <Argument name="gain_dataset_band_1">1</Argument>
+            <Argument name="offset_dataset_filename_1">{offset_filename}</Argument>
+            <Argument name="offset_dataset_band_1">1</Argument>
+        </Step>
+    </ProcessingSteps>
+    </VRTDataset>
+        """
+    )
+    assert struct.unpack("B" * 12, ds.GetRasterBand(1).ReadRaster()) == (
```
dbaston, line 718: I think it would improve readability to use numpy for new tests. Is there a packager/user who needs to run the test suite and has sufficiently recent versions of `pytest` etc. but doesn't have `numpy` available?

rouault: > I think it would improve readability to use numpy for new tests. Is there a packager/user who needs to run the test suite and has sufficiently recent versions of `pytest` etc. but doesn't have `numpy` available?

yes, in that instance, that's probably not necessary to be that conservative. I've added https://github.com/OSGeo/gdal/pull/9598/commits/174c886a340f7a95a08eb8d8dc6c13515d32dd7f that uses numpy functionality

PR #9600, opened 2024-04-03
gdal_translate: avoid dataset leak on invalid -exponent and -scale

dbaston:
## What does this PR do?

Also fix no-op assertions in vrt_read.py and comment out those that subsequently fail.

I came across this when trying to figure out why some tests in `tiff_read.py` fail if they run after `vrt.py`.

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/autotest/gcore/vrt_read.py
+++ b/autotest/gcore/vrt_read.py
@@ -1442,25 +1442,27 @@ def test_vrt_invalid_source_band():
 @gdaltest.enable_exceptions()
 def test_vrt_protocol():
 
+    files_opened_start = gdaltest.get_opened_files()
+
     with pytest.raises(Exception):
-        assert not gdal.Open("vrt://")
+        gdal.Open("vrt://")
     with pytest.raises(Exception):
-        assert not gdal.Open("vrt://i_do_not_exist")
+        gdal.Open("vrt://i_do_not_exist")
     with pytest.raises(Exception):
-        assert not gdal.Open("vrt://i_do_not_exist?")
+        gdal.Open("vrt://i_do_not_exist?")
 
     ds = gdal.Open("vrt://data/byte.tif")
     assert ds.RasterCount == 1
     assert ds.GetRasterBand(1).Checksum() == 4672
 
     with pytest.raises(Exception):
-        assert not gdal.Open("vrt://data/byte.tif?foo=bar")
+        gdal.Open("vrt://data/byte.tif?foo=bar")
     with pytest.raises(Exception):
-        assert not gdal.Open("vrt://data/byte.tif?bands=foo")
+        gdal.Open("vrt://data/byte.tif?bands=foo")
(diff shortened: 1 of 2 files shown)


autotest/gcore/vrt_read.py
```diff
@@ -1442,25 +1442,27 @@ def test_vrt_invalid_source_band():
 @gdaltest.enable_exceptions()
 def test_vrt_protocol():
 
+    files_opened_start = gdaltest.get_opened_files()
+
     with pytest.raises(Exception):
-        assert not gdal.Open("vrt://")
```
dbaston, line 1446: `AssertionError` subclasses `Exception`, so if `gdal.Open` returns a dataset then our assertion will fail and `pytest.raises` will find what it was looking for.

autotest/gcore/vrt_read.py
```diff
@@ -1666,9 +1669,17 @@ def test_vrt_protocol():
     ds = gdal.Open("vrt://data/tiff_with_subifds.tif?sd_name=1")
     assert ds.GetRasterBand(1).Checksum() == 35731
     with pytest.raises(Exception):
-        assert not gdal.Open("vrt://data/tiff_with_subifds.tif?sd=2&sd_name=1")
-        assert not gdal.Open("vrt://data/tiff_with_subifds.tif?sd=3")
-        assert not gdal.Open("vrt://data/tiff_with_subifds.tif?sd_name=sds")
```
dbaston, line 1671: With multiple operations under the same `pytest.raises`, only one of them needs to fail.

autotest/gcore/vrt_read.py
```diff
@@ -1620,10 +1622,10 @@ def test_vrt_protocol():
     assert ds.GetRasterBand(1).YSize == 10
 
     ## separated oo instances not allowed
-    with pytest.raises(Exception):
-        assert not gdal.Open(
-            "vrt://data/byte_with_ovr.tif?oo=GEOREF_SOURCES=TABFILE&oo=OVERVIEW_LEVEL=0"
-        )
+    # with pytest.raises(Exception, match="option should be specified once"):
+    #     gdal.Open(
+    #         "vrt://data/byte_with_ovr.tif?oo=GEOREF_SOURCES=TABFILE&oo=OVERVIEW_LEVEL=0"
```
dbaston, line 1627: `VRTDataset::OpenVRTProtocol` is using `CSLTokenizeString2` to parse the query string. It looks like this function discards the second `oo` that is expected to produce an error here.

rouault: No, CSLTokenizeString2() does the job properly, but there were missing return nullptr. Fixed per e7cf790a2d2c1d6de37d77c9d474018bfcb376c8. Rebasing on top of latest master should fix it

dbaston: Done

PR #9607, opened 2024-04-03
WMS: Use GDAL_HTTP_TIMEOUT

dbaston:


rouault:
Could be appropriate to document in https://gdal.org/drivers/raster/wms.html that <Timeout> defaults to GDAL_HTTP_TIMEOUT

PR #9611, opened 2024-04-03
autotest: update pyscripts tests to run in parallel

dbaston:
## What does this PR do?

Uses tmp_path fixture to prevent clobbering, enabling these tests to run in parallel.

Apparently I did this last August but somehow forgot to make a PR.

### DIFF
--- a/autotest/pyscripts/test_gdal2tiles.py
+++ b/autotest/pyscripts/test_gdal2tiles.py
@@ -74,6 +74,18 @@ def test_gdal2tiles_version(script_path):
     )
 
 
+def pil_available():
+    try:
+        import numpy  # noqa
+        from PIL import Image  # noqa
+
+        import osgeo.gdal_array as gdalarray  # noqa
+
+        return True
+    except ImportError:
+        return False
+
+
 def _verify_raster_band_checksums(filename, expected_cs=[]):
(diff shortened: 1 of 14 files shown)


autotest/pyscripts/test_gdal2tiles.py
```diff
@@ -89,23 +100,22 @@ def _verify_raster_band_checksums(filename, expected_cs=[]):
...
     shutil.copy(
         test_py_scripts.get_data_path("gdrivers") + "small_world.tif",
-        "tmp/out_gdal2tiles_smallworld.tif",
+        input_tif,
     )
 
-    os.chdir("tmp")
     test_py_scripts.run_py_script(
-        script_path, "gdal2tiles", "-q out_gdal2tiles_smallworld.tif"
+        script_path, "gdal2tiles", f"-q {input_tif} {output_dir}"
```
rouault, line 114: this no longer tests a non specified output dir. do we have another test where output_dir is not specified?

dbaston: I must have thought (incorrectly, I guess) that I couldn't call `os.chdir` when running tests in parallel. I reverted this to use chdir and added a comment explaining the purpose.

PR #9617, opened 2024-04-04
Update libkml.rst adding vital "config option" setting example

jidanni:
First, the user casually mentioning "a configuration option" will cause the user to look up and try each of
- -dsco NAME=VALUE: Dataset creation option (format specific)
- -lco  NAME=VALUE: Layer creation option (format specific)
- -oo   NAME=VALUE: Input dataset open option (format specific)
- -doo  NAME=VALUE: Destination dataset open option (format specific)

yes, even if each makes decreasing sense.

Finally, about to give up, he will try --help-general
stumbling upon
- --config key value: set system configuration option.

But that doesn't work either.

Then he thinks, "I know, perhaps I got the quoting wrong," and tries all five again...

My example brilliantly sweeps away all misunderstandings, pulling the user back before it's too late.


rouault:
closing for same reason as https://github.com/OSGeo/gdal/pull/9619

PR #9619, opened 2024-04-04
Update libkml.rst adding vital 'config option' setting example the mo…

jidanni:
…ment they are mentioned

Nonchalantly mentioning "configuration options" without saying how to set them will cause a nightmare. 

Yes in #9617 I added further examples below this today, but there also needs to be an example right where the user first sees and tries it. Sorry if I made the page ugly.

By the way, here if the user says NO the data just goes somewhere else. How about if the user does not like either the YES or NO choices and just wants the data to disappear, without bloating the KML or KMZ. Therefore perhaps mention what the user might do in that case. Thanks.

### DIFF
--- a/doc/source/drivers/vector/libkml.rst
+++ b/doc/source/drivers/vector/libkml.rst
@@ -572,7 +572,11 @@ except for some special fields as noted below.
   .. config:: LIBKML_USE_SIMPLEFIELD
      :choices: YES, NO
 
-     If ``NO``, export fields as <Data>.
+     If ``NO``, export fields as <Data>. Here's how to set it:
+
+.. code-block:: bash
+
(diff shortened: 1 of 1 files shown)


rouault:
There's a link just below the place to edit pointing to https://gdal.org/user/configoptions.html . Sorry but we can't repeat in every place the general GDAL concepts. This is going to bloat docs. There was a general discussion about that on the mailing list in thread https://lists.osgeo.org/pipermail/gdal-dev/2024-March/058784.html . So a general solution should be found, if any

dbaston:
> mentioning "configuration options" without saying how to set them

The text "configuration options" links to instructions on setting them. Or is this not referring to the HTML documentation?

PR #9628, opened 2024-04-05
autotest: remove duplicate definitions of ogr2ogr tests

dbaston:


rouault:
I'm not totally sure this is a good idea. I'd prefer ogr2ogr (cpp) tests not to be tied with ogr2ogr.py. The later should probably be just retired, or considered as frozen or an example of how to use the OGR Python API. This is more a youth error of mine where I enthusiastically ported ogr2ogr to Python but ogr2ogr.py hasn't evolved in recent years. And those days we have gdal.VectorTranslate() that does the job of having ogr2ogr in Python.

dbaston:
> I'd prefer ogr2ogr (cpp) tests not to be tied with ogr2ogr.py.

I'd agree. On the other hand, updating 1700 lines of duplicated tests to use `tmp_path` is also unappealing. `tmp_path` is needed to allow these to run in parallel, and `test_ogr2ogr_py.py` is the only test in `pyscripts` that cannot be run in parallel after #9611.

If `ogr2ogr.py` is no longer needed, it could be removed along with its tests.

rouault:
> If `ogr2ogr.py` is no longer needed, it could be removed along with its tests.

yes, please do so, and remove the link to it in https://gdal.org/api/python_samples.html. I was wondering if we could perhaps keep it, but point to let's say the version of 3.8.5 tag of it, but that's probably just unneeded nostalgia from me :-)

PR #9629, opened 2024-04-05
Update csv.rst: not just points!

jidanni:


rouault:
this change is not appropriate as the sentence mentions "from a CSV file which has columns for the X and Y coordinates,", so this implies points.  WKT or WKB encodings which allow non-points are mentionned above

jidanni:
Well I was thinking when the user sees just "points", he will give up, not examining the page further, when all along he could have still used e.g., `ogr2ogr -nlt` to express that what one is looking at are not just points!

PR #9637, opened 2024-04-05
gdal2xyz: Change -srcwin parameter type to integer.

eliepa:
## PR fixes -srcwin parameter type

Window rectangle corners shall be specified as integers.


jratike80:
Is it a problem if it is float? If I read right https://github.com/OSGeo/gdal/blob/master/apps/gdal_translate_lib.cpp#L291 gdal_translate is using doubles. What if the windows is computed by code that can return something else than integers due to re-projection or something?

eliepa:
Yes, there is a problem with float:
`x_off, y_off, x_size, y_size = srcwin`
and later:
`for x_i in range(0, x_size, x_skip):`
which causes an error.

PR #9645, opened 2024-04-05
ExecuteSQL(): error out if SetSpatialFilter() fails

### LINKED ISSUE #9623: ogr2ogr: catch invalid -spat arguments better
### Feature description

```
ogr2ogr -spat BLA 30 -90 50 orr.csv -dialect SQLITE -sql \
'SELECT XCOORD,YCOORD,POINTID FROM PLSSPoint_GCDB WHERE
POINTID LIKE "UT260060S0020E0%"' file.gdb.zip; echo $?
ERROR 1: Invalid geometry field index : 0
0
```
Maybe say
> ERROR 1: Invalid geometry field index : 0 (xmin)

or
> ERROR 1: Invalid geometry field index : 0 ("BLA") (xmin)

### Additional context

And more importantly, set $? to non-zero!

rouault:
fixes #9623

### DIFF
--- a/autotest/ogr/ogr_sql_test.py
+++ b/autotest/ogr/ogr_sql_test.py
@@ -1863,3 +1863,19 @@ def test_ogr_sql_ilike_utf8():
 
     lyr.SetAttributeFilter("'éven' ILIKE '%xen'")
     assert lyr.GetFeatureCount() == 0
+
+
+###############################################################################
+# Test error on setting a spatial filter during ExecuteSQL
+
+
+@gdaltest.enable_exceptions()
+def test_ogr_sql_test_execute_sql_error_on_spatial_filter_mem_layer():
+
+    ds = ogr.GetDriverByName("Memory").CreateDataSource("")
+    ds.CreateLayer("test", geom_type=ogr.wkbNone)
+    geom = ogr.CreateGeometryFromWkt("POLYGON((0 0,0 1,1 1,1 0,0 0))")
+    with pytest.raises(
+        Exception, match="Cannot set spatial filter: no geometry field present in layer"
+    ):
+        ds.ExecuteSQL("SELECT 1 FROM test", spatialFilter=geom)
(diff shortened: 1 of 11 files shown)


rouault:
Ultimately we'd rather want SetSpatialFilter() to return an error code rather than relying on CPLGetLastErrorType() / CPLGetErrorCount()... but that involves revisiting all drivers

ogr/ogrsf_frmts/generic/ogrlayer.cpp
```diff
@@ -1439,13 +1439,55 @@ OGRGeometryH OGR_L_GetSpatialFilter(OGRLayerH hLayer)
...
+/************************************************************************/
+
+//! @cond Doxygen_Suppress
+bool OGRLayer::ValidateGeometryFieldIndexForSetSpatialFilter(
+    int iGeomField, const OGRGeometry *poGeomIn, bool bIsSelectLayer)
+{
+    if (iGeomField == 0 && poGeomIn == nullptr &&
+        GetLayerDefn()->GetGeomFieldCount() == 0)
+    {
+        /* do nothing */
```
elpaso, line 1453: shouldn't this return false?

autotest/ogr/ogr_sql_test.py
```diff
@@ -1863,3 +1863,19 @@ def test_ogr_sql_ilike_utf8():
 
     lyr.SetAttributeFilter("'éven' ILIKE '%xen'")
     assert lyr.GetFeatureCount() == 0
+
+
+###############################################################################
+# Test error on setting a spatial filter during ExecuteSQL
+
+
+@gdaltest.enable_exceptions()
```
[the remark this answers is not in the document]
rouault: no, here we are testing with the default OGRSQL dialect, whereas https://github.com/OSGeo/gdal/pull/9645/files#diff-a9a9b765e9b436a2d6ca92b8250f870cae3b82b44159fd6b9b959ea272b6b69bR2486 is for the SQLITE dialect

ogr/ogrsf_frmts/generic/ogrlayer.cpp
rouault: no, this is a tolerance. Setting a null spatial filter on geometry field idx 0 when there are no geometry field can't harm, and is accepted silently for backward compatibility with existing practice. I'm adding a code comment to make that clearer

Issue #9618, opened 2024-04-04
Add "Perhaps try again with ..." upon raster / vector guess failure


jidanni:
### Feature description
<hr>

```
$ gdalinfo ...
ERROR ... not recognized as a supported file format.
```
Add "Perhaps try again with ogrinfo."

<hr>

```
$ ogrinfo ...
ERROR ... not recognized as a supported file format.
```
Add "Perhaps try again with gdainfo."

<hr>

### Additional context

Reasoning:

- It might not be entirely obvious, simply looking at the file suffix, what exactly the file contains.
- If users are expected to be experts about the contents of files, then they wouldn't need "info" programs in the first place.
- Just like when we go to the hospital and it turns out our problem A turns out to be problem B. At least the doctor will try and kindly refer us elsewhere instead of just being "dumbstruck", "I'm sorry Ms. Johnson, there is no cure." At least give the patient a glimmer of hope.
- Perhaps in the first place, or in the future, there should have been / should be a "jack of all trades" xxxinfo program, (like Unix file(1)), that is ready for both "genders", and even newer / future hybrid formats.

PR #9654, opened 2024-04-09
gdalinfo, ogrinfo: Improve error output when file cannotbe opened

dbaston:
## What does this PR do?

Updates `gdalinfo` to check if a file that cannot be opened has a name indicating a vector data type, and suggest the user call `ogrinfo` instead. Make the corresponding update to `ogrinfo`.

#9618

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/apps/gdalinfo_bin.cpp
+++ b/apps/gdalinfo_bin.cpp
@@ -148,8 +148,21 @@ MAIN_START(argc, argv)
 #ifdef __AFL_HAVE_MANUAL_CONTROL
             continue;
 #else
-        fprintf(stderr, "gdalinfo failed - unable to open '%s'.\n",
-                psOptionsForBinary->pszFilename);
+        VSIStatBuf sStat;
+        CPLString message;
+        message.Printf("gdalinfo failed - unable to open '%s'.",
+                       psOptionsForBinary->pszFilename);
+        if (VSIStat(psOptionsForBinary->pszFilename, &sStat) == 0)
+        {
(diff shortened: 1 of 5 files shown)


elpaso (approved):

apps/gdalinfo_bin.cpp
```diff
@@ -148,8 +148,32 @@ MAIN_START(argc, argv)
 #ifdef __AFL_HAVE_MANUAL_CONTROL
             continue;
 #else
-        fprintf(stderr, "gdalinfo failed - unable to open '%s'.\n",
-                psOptionsForBinary->pszFilename);
+        bool bMaybeVector = false;
+        VSIStatBuf sStat;
+        if (VSIStat(psOptionsForBinary->pszFilename, &sStat) == 0)
```
rouault, line 153: what about using GDALIdentifyDriver() instead and test if it is a vector driver? That would be more robust as not all drivers are extension specific, and that would work with non-file connection name

Same remark for ogrinfo

dbaston: Is there a potential cost to that? I worried about a bunch of filesystem or network activity just to generate a better error message, but that may be unfounded.

rouault: That should be acceptable. Most network requests should already have bien cached. Abd this is an error code path so performance is less of a concern

dbaston: I've made this change

Issue #1427, opened 2019-04-07
Set band name in netCDF files


giumas:
## Expected behavior and actual behavior.

I expected to be able to set the band name in a netCDF raster file, but the created band has the default `Band1` name.

## Steps to reproduce the problem.

Execute this Python script:

```
import logging
from osgeo import gdal
from osgeo import osr
import numpy as np

logging.basicConfig(level=logging.DEBUG, format="%(levelname)-9s %(name)s.%(funcName)s:%(lineno)d > %(message)s")
logger = logging.getLogger(__name__)

# fake data source
np.random.seed(0)
ny = 120
nx = 160
arr1 = np.random.rand(nx, ny)
gt = (360988.700, 0.200, 0.000, 4769988.900, 0.000, -0.200)
# output file
nc_path = "gdal.nc"

# prepare netCDF file
dst_drv = gdal.GetDriverByName('netCDF')
dst_ds = dst_drv.Create(nc_path, ny, nx, 1, gdal.GDT_Float32)
dst_ds.SetGeoTransform(gt)
srs = osr.SpatialReference() 
srs.ImportFromEPSG(32619)
dst_ds.SetProjection(srs.ExportToWkt())

# write band
dst_ds.GetRasterBand(1).SetDescription("Bathymetry")  // <<<<< Set the name to 'Bathymetry'
dst_ds.GetRasterBand(1).SetNoDataValue(np.nan)
dst_ds.GetRasterBand(1).WriteArray(arr1)

# finalize to disk and close
dst_ds.FlushCache()
dst_ds = None
```
Then, `gdalinfo gdal.nc`:
```

Warning 1: Recode from UTF-8 to CP_ACP failed with the error: "Invalid argument".
Driver: netCDF/Network Common Data Format
Files: C:\code\hyo2\backscatter\hyo2_openbst\examples\workground\gdal.nc
Size is 120, 160
Coordinate System is:
PROJCS["WGS 84 / UTM zone 19N",
    GEOGCS["WGS 84",
        DATUM["WGS_1984",
            SPHEROID["WGS 84",6378137,298.257223563,
                AUTHORITY["EPSG","7030"]],
            AUTHORITY["EPSG","6326"]],
        PRIMEM["Greenwich",0,
            AUTHORITY["EPSG","8901"]],
        UNIT["degree",0.0174532925199433,
            AUTHORITY["EPSG","9122"]],
        AUTHORITY["EPSG","4326"]],
[...]

PR #9657, opened 2024-04-10
netCDF: Add BAND_NAMES creation option

dbaston:
## What does this PR do?

Adds an option to control the names of netCDF variables created by GDAL. 

https://github.com/OSGeo/gdal/issues/1427
https://gis.stackexchange.com/questions/230642/setting-band-names-when-writing-multiple-layer-rasters-using-gdal-with-python
https://gis.stackexchange.com/questions/265204/gdal-creation-option-to-set-variable-name-to-netcdf

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/frmts/netcdf/netcdfdataset.cpp
+++ b/frmts/netcdf/netcdfdataset.cpp
@@ -9156,7 +9156,6 @@ netCDFDataset *netCDFDataset::CreateLL(const char *pszFilename, int nXSize,
 
         return poDS;
     }
-
     // Create the dataset.
     CPLString osFilenameForNCCreate(pszFilename);
 #if defined(_WIN32) && !defined(NETCDF_USES_UTF8)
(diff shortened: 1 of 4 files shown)


elpaso (commented):
code LGTM, the codecheck CI failure seems like a false positive. As a workaround you can possibly check for poDS != nullptr.

The other gcore test failure seems unrelated.

frmts/netcdf/netcdfdataset.cpp
```diff
@@ -9104,7 +9104,7 @@ netCDFDataset *netCDFDataset::CreateLL(const char *pszFilename, int nXSize,
 
     CPLReleaseMutex(hNCMutex);  // Release mutex otherwise we'll deadlock with
         // GDALDataset own mutex.
-    netCDFDataset *poDS = new netCDFDataset();
+    auto poDS = std::make_unique<netCDFDataset>();
```
rouault, line 9107: I'm not totally sure this change is safe. See the comments around the explicit "delete poDS" about mutex deadlocks

dbaston:
```
// Release mutex otherwise we'll deadlock with
        // GDALDataset own mutex.
```
Is this referring to use of `hNCMutex` by `netCDFDataset` ? Or is it referring to the `m_poPrivate->hMutex` in `GDALDataset` ?

rouault: I believe this was a deadlock between hNCMutex and https://github.com/OSGeo/gdal/blob/732d82cf7486be5ee12e1d67c60f09f051072e0e/gcore/gdaldataset.cpp#L324

dbaston: I've removed this change.

PR #9659, opened 2024-04-12
Update pds4.rst : Raster option VAR_Target is wrong, use VAR_TARGET_TYPE

kidpixo:
## What does this PR do?

This is only a small add to the documentation, this PR correct an example in it.

The suggested option `-co VAR_Target=Moon` in pds4.rst results in an error producing a PDS4 raster examples.
 
```
Warning 1: ${TARGET} could not be substituted
```
Inspecting the resulting xml files I found :

```xml
<Target_Identification>
    <name>${TARGET}</name>
    <type>Satellite</type>
    [...]
</Target_Identification>
```
where the option `-co VAR_TARGET_TYPE="Satellite"` were correctly interpolated.

I just changed the offending option to `-co VAR_TARGET_NAME=Moon` and it works.

By chance, I discovered that the variables names are case insensitive.

NONE

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments
 - [x] All CI builds and checks have passed : Github says "Able to merge. These branches can be automatically merged."
 
## Environment

**Not Appilcable**

* OS:
* Compiler:

## Final Words

Thanks for this great tool!

### DIFF
--- a/doc/source/drivers/raster/pds4.rst
+++ b/doc/source/drivers/raster/pds4.rst
@@ -421,7 +421,7 @@ its parameterized variables:
 
    $ gdal_translate input.tif output.xml -of PDS4 \
                -co VAR_TARGET_TYPE=Satellite \
-               -co VAR_TARGET_TYPE=Moon \
+               -co VAR_TARGET_NAME=Moon \
                -co VAR_OBSERVING_SYSTEM_NAME=LOLA \
                -co VAR_LOGICAL_IDENTIFIER=Lunar_LRO_LOLA_DEM_Global_64ppd.tif \
                -co VAR_TITLE="LRO LOLA Digital Elevation Model (DEM) 64ppd" \
@@ -439,7 +439,7 @@ Create a text file "myOptions.txt" with the below content
    #This is a comment
    #Conversion parameters for the LRO LOLA dataset
    -co VAR_TARGET_TYPE=Satellite
-   -co VAR_TARGET_TYPE=Moon
+   -co VAR_TARGET_NAME=Moon
    -co VAR_OBSERVING_SYSTEM_NAME=LOLA
    -co VAR_LOGICAL_IDENTIFIER=Lunar_LRO_LOLA_DEM_Global_64ppd.tif
    -co VAR_TITLE="LRO LOLA Digital Elevation Model (DEM) 64ppd"
--- a/doc/source/drivers/raster/pds4.rst
+++ b/doc/source/drivers/raster/pds4.rst
@@ -421,7 +421,7 @@ its parameterized variables:
 
    $ gdal_translate input.tif output.xml -of PDS4 \
                -co VAR_TARGET_TYPE=Satellite \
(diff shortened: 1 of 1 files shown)


doc/source/drivers/raster/pds4.rst
```diff
@@ -421,7 +421,7 @@ its parameterized variables:
 
    $ gdal_translate input.tif output.xml -of PDS4 \
                -co VAR_TARGET_TYPE=Satellite \
-               -co VAR_Target=Moon \
+               -co VAR_TARGET_TYPE=Moon \
```
jratike80, line 424: Did you mean "VAR_TARGET_NAME"?

kidpixo:
Argh, I was too quick , you are right.

doc/source/drivers/raster/pds4.rst
kidpixo: I was to quick , you're absolutely right.

kidpixo:
It should be fixed now,when the CI Is over .

doc/source/drivers/raster/pds4.rst
```diff
@@ -421,7 +421,7 @@ its parameterized variables:
 
    $ gdal_translate input.tif output.xml -of PDS4 \
                -co VAR_TARGET_TYPE=Satellite \
-               -co VAR_Target=Moon \
+               -co VAR_TARGET_NAME=Moon \
```
rouault, line 424:
```suggestion
               -co VAR_TARGET=Moon \
```

doc/source/drivers/raster/pds4.rst
```diff
@@ -439,7 +439,7 @@ Create a text file "myOptions.txt" with the below content
    #This is a comment
    #Conversion parameters for the LRO LOLA dataset
    -co VAR_TARGET_TYPE=Satellite
-   -co VAR_Target=Moon
+   -co VAR_TARGET_NAME=Moon
```
rouault, line 442:
```suggestion
   -co VAR_TARGET=Moon
```

rouault:
The correct fix was actually to use VAR_TARGET upper case.

kidpixo:
> The correct fix was actually to use VAR_TARGET upper case.

I see, the variable in the template was indeed `${TARGET}` and I guess the variable from user are interpreted as 
`VAR_TARGET` > variable `TARGET` in template.

Checking again, setting `VAR_TARGET_NAME` or `VAR_Target_NAME` produces  

```xml
<Target_Identification>
      <name>Mercury_2015</name>
```
that is the name of spheroid used in the projection `<cart:spheroid_name>Mercury_2015</cart:spheroid_name>` 

I got sidetracked from my own implementation I use to fill up  PDS4 template. 
I use a variable in the template as close as possible to the XPATH of the object, in this case something like `Target_Identification_name`.

PR #9669, opened 2024-04-15
VSICACHE: avoid EOF read

### LINKED ISSUE #9658: VSI_CACHE causing bad range requests in some circumstance (/vsizip//vsis3/)
### What is the bug?

I have a 4211788 byte (4MiB) zip file containing a ~10MiB tiff file in an S3 bucket.
I'm using gdal from python to open the last tiff file in the zip file.
The following works perfectly for me:
`gdal.Open("/vsizip//vsis3/cdn-misc.kx.gd/foss-tickets/file_example_TIFF_10MB.zip", gdal.GA_ReadOnly)`
However if I first turn on the VSI cache with `gdal.SetConfigOption("VSI_CACHE", "TRUE")`
Then it fails with:
`S3: Request at offset 4227072, after end of file`
Note that this *is* after the end of the file, and that it corresponds to the start of the 129th cache block - VSI cache blocks default to 32KiB in size.
So, it looks like the VSI cache is for some reason requesting a block outside the file.

I found out this works with `/vsizip//vsicurl/` too, not just `/vsizip/vsis3/`

I've only figured out how to show this from python so far.
```
from osgeo import gdal

gdal.UseExceptions()

opts = {
    'VSI_CACHE': 'TRUE',
    'CPL_DEBUG': 'ON',
}

for k, v in opts.items():
    gdal.SetConfigOption(k, v)

# Use either /vsizip//vsis3/
path = '/vsizip//vsis3/cdn-misc.kx.gd/foss-tickets/file_example_TIFF_10MB.zip'
# or /vsizip//vsicurl/
path = '/vsizip//vsicurl/https://cdn-misc.kx.gd/foss-tickets/file_example_TIFF_10MB.zip'

gdal_tile = gdal.Open(path, gdal.GA_ReadOnly)
gdal_tile = None
```

You should see one or other of the following:
`S3: Request at offset 4227072, after end of file`
`VSICURL: Request at offset 4227072, after end of file`

I think in this case the program succeeds anyway in spite of trying to read beyond the end of the file, but I'm not sure if that's always the case.

### Versions and provenance

macOS Sonoma 14.3.1
GDAL 3.8.5

### Additional context

Wasn't too hard to make a new test file that exhibits this behaviour, so I think this must be affecting the majority of zip files, at least when the vsi-cache is on and the user tries to access the last file in the zip.

elpaso:
Do not read past EOF even if the client (e.g. VSIZIP) code say so.

Fix #9658

### DIFF
--- a/port/cpl_vsil_cache.cpp
+++ b/port/cpl_vsil_cache.cpp
@@ -360,8 +360,15 @@ size_t VSICachedFile::Read(void *pBuffer, size_t nSize, size_t nCount)
     /*      Make sure the cache is loaded for the whole request region.     */
     /* ==================================================================== */
     const vsi_l_offset nStartBlock = m_nOffset / m_nChunkSize;
-    const vsi_l_offset nEndBlock =
-        (m_nOffset + nRequestedBytes - 1) / m_nChunkSize;
+    // Calculate last block
+    const vsi_l_offset nLastBlock = m_nFileSize / m_nChunkSize;
+    vsi_l_offset nEndBlock = (m_nOffset + nRequestedBytes - 1) / m_nChunkSize;
+
+    // if nLastBlock is not 0 consider the min value to avoid out-of-range reads
+    if (nLastBlock != 0 && nEndBlock > nLastBlock)
+    {
+        nEndBlock = nLastBlock;
+    }
 
     for (vsi_l_offset iBlock = nStartBlock; iBlock <= nEndBlock; iBlock++)
     {
--- a/autotest/gcore/vsifile.py
+++ b/autotest/gcore/vsifile.py
@@ -1306,24 +1306,13 @@ def test_vsifile_win32_network_path():
 def test_vsifile_eof_cache_read(tmp_path):
     """Test issue GH #9658"""
 
-    np = pytest.importorskip("numpy")
-    # Create an empty GTiff file
-    filename = f"/vsizip/{tmp_path}/test.zip/testraster.tif"
-    driver = gdal.GetDriverByName("GTiff")
-    size = 1000
-    ds = driver.Create(
-        filename, size, size, 1, gdal.GDT_Byte, options=["STREAMABLE_OUTPUT=YES"]
-    )
-    assert ds
-    data = np.random.randint(0, 255, size=(size, size))
-    band = ds.GetRasterBand(1)
-    assert band.WriteArray(data) == 0
-    ds = None
-
-    # Read past EOF
+    tmp_filename = str(tmp_path / "vsifile_eof_cache_read.bin")
+    f = gdal.VSIFOpenL(tmp_filename, "wb")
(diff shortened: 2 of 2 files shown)


port/cpl_vsil_cache.cpp
```diff
@@ -360,8 +360,15 @@ size_t VSICachedFile::Read(void *pBuffer, size_t nSize, size_t nCount)
...
-    const vsi_l_offset nEndBlock =
-        (m_nOffset + nRequestedBytes - 1) / m_nChunkSize;
+    // Calculate last block
+    const vsi_l_offset nLastBlock = m_nFileSize / m_nChunkSize;
+    vsi_l_offset nEndBlock = (m_nOffset + nRequestedBytes - 1) / m_nChunkSize;
+
+    // if nLastBlock is not 0 consider the min value to avoid out-of-range reads
+    if (nLastBlock != 0 && nEndBlock > nLastBlock)
+    {
+        nEndBlock = nLastBlock;
```
rouault, line 370: is this code path exercised by existing tests?

elpaso: I don't know. I was thinking at a test but I have no idea where to reproduce this scenario with vsicache and vsicurl in a test.

rouault: Perhaps add an assertion and run autotest/gcore/vsifile.py ? There are tests in it with VSI_CACHE=YES

> but I have no idea how to reproduce this scenario with vsicache and vsicurl in a test.

Testing on a regular file should be enough

elpaso: ok, done. But without using curl there is no way to catch the error message that was triggered by curl, so the test will always pass. 
The new code path is triggered by the test though.

rouault: I was thinking to a simpler test that would not involve zip or tiff, but just use the VSILFILE API and issue a gdal.VSIFReadL() with a size bigger than the the file size.

Something like:
```
f = gdal.VSIFOpenL(tmp_filename, "wb")
gdal.VSIFWriteL(b"x" * 100000, 100000, 1, f)
gdal.VSIFCloseL(f)
with gdal.config_option("VSI_CACHE", "YES"):
    f = gdal.VSIFReadL(tmp_filename, "rb")
    gdal.VSIFSeekL(f, 60000, 0)
    data = gdal.VSIFReadL(1, 75000, f)  # reads past end of file
    gdal.VSIFCloseL(f)
    assert data == b"x" * 40000
```

elpaso: ok, test updated and verified that it triggers the new code path.

olsen232:
The 4MB zip file I attached to #9658 was just an example to demonstrate the bug, but this was causing me real-world issues opening TIFFs from a 4GB zip that I haven't uploaded here - today I've verified that this fixes the real-world issue for me.

PR #9687, opened 2024-04-17
gdalinfo: use argparser

elpaso:


apps/gdalinfo_lib.cpp
```diff
@@ -2113,135 +2282,42 @@ GDALInfoOptions *
...
+        argParser->parse_args_without_binary_name(aosArgv.List());
+    }
+    catch (const std::exception &error)
+    {
+        CPLError(CE_Failure, CPLE_AppDefined, "%s", error.what());
+        return nullptr;
     }
 
-    return psOptions;
+    if (psOptionsForBinary != nullptr && psOptionsForBinary->osFilename.empty())
```
rouault, line 2315: I believe this can be removed as argparser should validate that already

apps/gdalinfo_lib.cpp
```diff
@@ -215,6 +216,174 @@ gdal_json_object_new_double_significant_digits(double dfVal,
...
+        .choices("WKT1", "WKT2", "WKT2_2015", "WKT2_2018")
+        .nargs(1)
+        .store_into(psOptions->osWKTFormat)
+        .help(_("WKT format used for SRS."));
+
+    argParser->add_argument("-sd")
+        .metavar("<sd>")
+        .nargs(1)
+        .store_into(psOptionsForBinary->nSubdataset)
+        .help(
```
rouault, line 318: In https://github.com/OSGeo/gdal/pull/9683/files, I use this shorter help message:
```
    argParser->add_argument("-sd")
        .metavar("<n>")
        .store_into(psOptions->nSubdataset)
        .help(_("Use subdataset of specified index (starting at 1), instead of "
                "the source dataset itself."));
```

apps/gdalinfo_lib.cpp
```diff
@@ -215,6 +216,174 @@ gdal_json_object_new_double_significant_digits(double dfVal,
...
+    argParser->add_argument("-proj4")
+        .flag()
+        .store_into(psOptions->bReportProj4)
+        .help(_("Report a PROJ.4 string corresponding to the file's coordinate "
+                "system."));
+
+    argParser->add_argument("-wkt_format")
+        .metavar("<WKT1|WKT2|WKT2_2015|WKT2_2018>")
+        .choices("WKT1", "WKT2", "WKT2_2015", "WKT2_2018")
+        .nargs(1)
```
rouault, line 310: nargs = 1 is the default:

```suggestion
```

Similar instances below

apps/gdalinfo_lib.cpp
```diff
@@ -215,6 +216,174 @@ gdal_json_object_new_double_significant_digits(double dfVal,
...
+
+    argParser->add_argument("-proj4")
+        .flag()
+        .store_into(psOptions->bReportProj4)
+        .help(_("Report a PROJ.4 string corresponding to the file's coordinate "
+                "system."));
+
+    argParser->add_argument("-wkt_format")
+        .metavar("<WKT1|WKT2|WKT2_2015|WKT2_2018>")
+        .choices("WKT1", "WKT2", "WKT2_2015", "WKT2_2018")
```
rouault, line 309: I would remove the choices() here and let the validation be done afterwards. Actually WKT2_2018 is now deprecated, and we should encourage using WKT2_2019 instead (there are aliases).  The .rst should be updated to display WKT2_2019 too

apps/gdalinfo_lib.cpp
```diff
@@ -2045,7 +2214,7 @@ static void GDALInfoReportMetadata(const GDALInfoOptions *psOptions,
         else
         {
             aosExtraMDDomainsExpanded =
-                CSLDuplicate(psOptions->papszExtraMDDomains);
+                CSLDuplicate(psOptions->aosExtraMDDomains);
```
rouault, line 2217:
```suggestion
                psOptions->aosExtraMDDomains;
```

apps/gdalinfo_lib.cpp
elpaso: What do you mean with "afterwards" ? In an `action()` ?

apps/gdalinfo_lib.cpp
```diff
@@ -215,6 +216,174 @@ gdal_json_object_new_double_significant_digits(double dfVal,
...
+                _("Read and display image statistics computing approximated "
+                  "values on overviews or a subset of all tiles if required."));
+    }
+
+    argParser->add_argument("-hist")
+        .store_into(psOptions->bReportHistograms)
+        .help(_("Report histogram information for all bands."));
+
+    // Helper to set an inverted logic flag
+    auto setInvFlag = [&argParser](const std::string &name, bool &storage,
```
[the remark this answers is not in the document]
elpaso: done

apps/gdalinfo_lib.cpp
elpaso: done

apps/gdalinfo_lib.cpp
elpaso: done

PR #9689, opened 2024-04-17
Doc: Link RFC 8 Development guidelines from Development practices chapter

rouault:


dbaston:
I'd consider copying the relevant information directly into `dev_practices.rst` (a living document) vs directing attention to RFC-8, which documents practices at a current point in time but no longer applies in its entirety (for example, about making Python code compatible with versions 2 and 3).

rouault:
> I'd consider copying the relevant information directly into `dev_practices.rst` (a living document) vs directing attention to RFC-8, which documents practices at a current point in time but no longer applies in its entirety (for example, about making Python code compatible with versions 2 and 3).

RFC-8 mentions it is a living document, so we could update it. Having both RFC-8 and replicating in dev_practices could be source of confusion. Or we should empty the content of RFC-8 and move it to dev_practices

dbaston:
> move it to dev_practices

I'd be inclined to go this route. It seems like a lower barrier to maintain/update this than an RFC, which implies (to me, maybe wrongly) some amount of voting/process/ceremony.

rouault:
> I'd be inclined to go this route.

ok, I've just done that in https://github.com/OSGeo/gdal/pull/9692 . Closing this PR as superseded

PR #9697, opened 2024-04-18
Python bindings: Avoid crash when accessing closed file

dbaston:


swig/include/python/gdal_python.i
```diff
@@ -352,6 +352,44 @@ void wrapper_VSIGetMemFileBuffer(const char *utf8_path, GByte **out, vsi_l_offse
 %clear (GByte **out, vsi_l_offset *length);
 
 
+%pythonappend VSIFCloseL %{
+    args[0].this = None
+%}
+
+%pythonprepend VSIFCloseL %{
```
dbaston, line 359: These `%pythonprepend` blocks are not necessary to avoid the crash, they just make it raise the standard python exception instead of e.g. `TypeError: in method 'VSIFEofL', argument 1 of type 'VSILFILE *'
`

rouault:
doesn't the Read() method need a similar protection?

dbaston:
> doesn't the Read() method need a similar protection?

It's "protected" here, in that it no longer causes a crash. For reasons unknown to me SWIG doesn't pick up either `%pythonprepend wrapper_VSIFReadL` or `%pythonprepend VSIFReadL`. I guess I could remove `%apply Pointer NONNULL {VSILFILE* fp}` from that function and throw an appropriate exception if `fp` is null.

PR #9698, opened 2024-04-18
autotest: Prevent unrelated tests from interfering with each other

dbaston:
## What does this PR do?

Makes various fixes to prevent unrelated tests from interfering with each other.

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/autotest/gcore/vsiadls.py
+++ b/autotest/gcore/vsiadls.py
@@ -53,7 +53,7 @@ def open_for_read(uri):
 
 
 ###############################################################################
-@pytest.fixture(autouse=True, scope="module")
+@pytest.fixture(autouse=True, scope="function")
 def startup_and_cleanup():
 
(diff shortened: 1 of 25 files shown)


autotest/gcore/vsiadls.py
```diff
@@ -53,23 +53,9 @@ def open_for_read(uri):
 
 
 ###############################################################################
-@pytest.fixture(autouse=True, scope="module")
+@pytest.fixture(autouse=True, scope="function")
```
rouault, line 56: any particular reason to change the scope? I assume this causes the webserver to be started and stopped for each test, which might slow down things (but perhaps not significantly?)  For other network file systems, we use  scope="module" for this startup_and_cleanup method

dbaston: No reason that I can see, so I assume it is an accidental change I made when troubleshooting something. I reverted it.

And yes, starting and stopping the webserver for each test does slow things down quite a bit.

PR #9714, opened 2024-04-20
Add recommendation for finding older GDAL using CMake

Ryanf55:
## What does this PR do?

* This adds a recommendation on how to find GDAL on Ubuntu 22 which is distributed with GDALL 3.4.

n/A

## Tasklist

 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS: Ubuntu 22 with GDAL 3.4 installed as binaries.
* Compiler: G++11.4

### DIFF
--- a/doc/source/development/cmake.rst
+++ b/doc/source/development/cmake.rst
@@ -29,6 +29,7 @@ particular, CMake will consult (and set) the cache variable
 Before GDAL 3.5, you can use the following to create the imported library target ``GDAL::GDAL``:
 
 .. code::
+
     find_package(GDAL CONFIG QUIET)
     if(NOT TARGET GDAL::GDAL)
         find_package(GDAL REQUIRED)
--- a/doc/source/development/cmake.rst
+++ b/doc/source/development/cmake.rst
@@ -6,7 +6,7 @@ Using GDAL in CMake projects
 
 .. versionadded:: 3.5
 
-The recommended way to use the GDAL library in a CMake project is to
+The recommended way to use the GDAL library 3.5 or higher in a CMake project is to
 link to the imported library target ``GDAL::GDAL`` provided by
 the CMake configuration which comes with the library. Typical usage is:
 
@@ -25,3 +25,22 @@ number of places. The lookup can be adjusted for all packages by setting
 the cache variable or environment variable ``CMAKE_PREFIX_PATH``. In
 particular, CMake will consult (and set) the cache variable
 ``GDAL_DIR``.
+
+Before GDAL 3.5, you can use the following to create the imported library target ``GDAL::GDAL``:
+
+.. code::
+    find_package(GDAL CONFIG QUIET)
+    if(NOT TARGET GDAL::GDAL)
+        find_package(GDAL REQUIRED)
+        if(NOT TARGET GDAL::GDAL)
+            add_library(GDAL IMPORTED)
+            if(DEFINED GDAL_LIBRARIES)
+                target_link_libraries(GDAL INTERFACE "${GDAL_LIBRARIES}")
+                add_library(GDAL::GDAL ALIAS GDAL)
+            else()
+                message(FATAL_ERROR "Missing GDAL_LIBRARIES")
+            endif()
+        endif()
+    endif()
+
+    target_link_libraries(MyApp PRIVATE GDAL::GDAL)


doc/source/development/cmake.rst
```diff
@@ -25,3 +25,22 @@ number of places. The lookup can be adjusted for all packages by setting
 the cache variable or environment variable ``CMAKE_PREFIX_PATH``. In
 particular, CMake will consult (and set) the cache variable
 ``GDAL_DIR``.
+
+Before GDAL 3.5, you can use the following to create the imported library target ``GDAL::GDAL``:
+
+.. code::
```
rouault, line 31:
```suggestion
.. code::

```

doc/source/development/cmake.rst
```diff
@@ -25,3 +25,23 @@ number of places. The lookup can be adjusted for all packages by setting
...
+    if(NOT TARGET GDAL::GDAL)
+        find_package(GDAL REQUIRED)
+        if(NOT TARGET GDAL::GDAL)
+            add_library(GDAL IMPORTED)
+            if(DEFINED GDAL_LIBRARIES)
+                target_link_libraries(GDAL INTERFACE "${GDAL_LIBRARIES}")
+                add_library(GDAL::GDAL ALIAS GDAL)
+            else()
+                message(FATAL_ERROR "Missing GDAL_LIBRARIES")
+            endif()
```
dg0yt, line 43: Which CMake version is this targeting?

 - CMake 3.14 added `GDAL::GDAL` to the find module.
 - `add_library(GDAL IMPORTED)` lacks a type (`INTERFACE`?).
 - The `FATAL_ERROR` shouldn't exist. Just rely on the required package supplying `GDAL_LIBRARIES`.
 - I'm also not sure if the quotes in `target_link_libraries` do the right thing.

Baseline: Test the suggested pattern with `find_package(GDAL MODULE)` and an old version CMake.

Ryanf55: I was intending to go down to 3.12, but realized by using a local version higher than that, this was untested and completely wrong. My apologies.

The intent was to cover was the situation a user has a separate `FindGDAL.cmake` script in their `CMAKE_MODULE_PATH`, which is [preferred over CMake's supplied version](https://cmake.org/cmake/help/latest/command/find_package.html#search-modes).

[Github search indicated](https://github.com/search?q=FindGDAL&type=code
) a lot of repos still have a `FindGDAL.cmake` script, which in no way guarantees that the GDAL::GDAL target will exist. 

To start, what about we target a minimum of 3.14 to keep things simple, because [Ubuntu 20 comes with 3.16](https://packages.ubuntu.com/focal/cmake), and leave people writing their own find script on their own.

If so, here's a tested version that I ran on ubuntu focal:
```cmake
cmake_minimum_required(VERSION 3.14)
project(MyAppExample)

find_package(GDAL CONFIG QUIET)
if(NOT TARGET GDAL::GDAL)

    find_package(GDAL REQUIRED)
endif()

add_executable(MyApp main.cpp)

target_link_libraries(MyApp PRIVATE GDAL::GDAL)
```

Create main.cpp:
```c++
#include <gdal/gdal_priv.h>

int main() {
    GDALAllRegister();

    return 0;
}
```

Put these in a folder and run the below commands, with changes to the path as needed.

```bash
docker run --net host -v /home/ryan/Dev/ros2_ws/src/gdal/examples/gdal_old_cmake:/ws -w /ws -it ubuntu:focal bash
apt-get update
DEBIAN_FRONTEND=noninteractive apt install -y cmake build-essential libgdal-dev
cmake  -B build
cmake --build build
```

dg0yt: > The intent was to cover was the situation a user has a separate `FindGDAL.cmake` script in their `CMAKE_MODULE_PATH`, which is [preferred over CMake's supplied version](https://cmake.org/cmake/help/latest/command/find_package.html#search-modes).

These proliferation of divergent find modules is part of the reasin why find modules are bad... You cannot give generic instructions for unknown find modules. If somebody has a custom find module, he must already know what he is doing.

Ryanf55: Proposed fix: https://github.com/OSGeo/gdal/pull/9743

PR #9726, opened 2024-04-22
TileDB: use GEOM_WKB type when creating geometry columns with TileDB >= 2.21

rouault:
Builds again TileDB < 2.21 will still use the BLOB type on creation, and builds again TileDB >= 2.21 will still support reading BLOB geometry columns.

CC @perrygeo


rouault:
CC @normanb @sgillies

PR #9729, opened 2024-04-23
Fix generated find_dependency

dg0yt:
The version number must be the second argument.


dg0yt:
The alternative is to handle the version number separately from other arguments. This would allow the explicit NAMES and CONFIGS arguments to be in front of the remaining arguments.

rouault:
@dg0yt This looks good to me. Any reason to let it as a draft?

dg0yt:
vcpkg CI is happy now.

PR #9731, opened 2024-04-23
Fixin two errors and adding tests for MiraMonVector driver

AbelPau:
## What does this PR do?
MiraMonVector: fix wrong use of number of records in string lists for field values
MiraMonVector: fix wrong word Lower instead of Lowest in metadata OpenOptionList
MiraMonVector: add new writing tests and improves the existing ones

## Tasklist

 - [x] Add test case(s)


rouault:
@abelpau Can you redo the pull request against master ? (there are a couple wrong changes by issuing a pull request from a branch forked off master against 3.9 directly). I'll tag it then as "backport release/3.9" so it is backported automatically

Regarding the new tests there's a lot of copy&paste. I guess the checks for the attributes don't need to be repeated for each geometry type. So they could be done in a dedicated test, and then the tests for the geometries would only check the geometry aspect.

PR #9734, opened 2024-04-23
Miramon: fixes and additional tests

rouault:
MiraMonVector: fix wrong use of number of records in string lists for field values
MiraMonVector: fix wrong word Lower instead of Lowest in metadata OpenOptionList
MiraMonVector: add new writing tests and improves the existing ones


rouault:
@abelpau See the updated coverage: https://coveralls.io/builds/67084419/source?filename=ogr%2Fogrsf_frmts%2Fmiramon%2Fogrmiramonlayer.cpp

AbelPau:
> @AbelPau See the updated coverage: https://coveralls.io/builds/67084419/source?filename=ogr%2Fogrsf_frmts%2Fmiramon%2Fogrmiramonlayer.cpp

This is a very low increase after all the work I did. I was convinced this would improve a lot. But I'll review and create more tests. No problem.

rouault:
> This is a very low increase after all the work I did. I was convinced this would improve a lot

Well, I see there has been significant progress, but writing tests is a bit of "diminishing return" process, and there's a point where you need to add lots of tests to cover the remaining code paths. That said, covering nominal code paths is a reasonable target to aim for and should generally require a "reasonable" effort.
On the contrary, covering error code paths requires a *LOT* of effort (for example the SQLite software has a test suite whose line count is 100 times the line count of the library itself, to have 100% line and branch coverage ! This is obviously an extreme case).
In some heavily used and tested parts of GDAL where we test nominal code paths and a few error code paths, the number of line of tests can be comparable to the number of line of actual code in the library.

rouault:
@abelPau Now this has been merged, if you do follow up enhancements/fixes, do them on a branch forked off from latest master
```
git checkout master
git pull origin master
git checkout -b name_of_new_branch
```

AbelPau:
Today I created more tests (and two minor fixes) and did a PR. I hope this time is all right!

PR #9739, opened 2024-04-24
[gdal_contour] argparser

elpaso:
Uses the new argparser for the gdalcontour options.

### DIFF
--- a/apps/gdal_contour.cpp
+++ b/apps/gdal_contour.cpp
@@ -159,8 +159,8 @@ GDALContourAppOptionsGetParser(GDALContourOptions *psOptions)
 
     group.add_argument("-fl")
         .metavar("<level>")
+        .nargs(argparse::nargs_pattern::at_least_one)
         .scan<'g', double>()
-        .append()
         .action([psOptions](const std::string &s)
                 { psOptions->adfFixedLevels.push_back(CPLAtof(s.c_str())); })
         .help(_("Name one or more \"fixed levels\" to extract."));
@@ -328,7 +328,7 @@ MAIN_START(argc, argv)
     /*      Create the output file.                                         */
     /* -------------------------------------------------------------------- */
     CPLString osFormat;
-    if (!sOptions.osFormat.empty())
+    if (sOptions.osFormat.empty())
     {
         const auto aoDrivers = GetOutputDriversFor(
             sOptions.aosDestFilename.c_str(), GDAL_OF_VECTOR);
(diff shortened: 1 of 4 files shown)


apps/gdal_utils_priv.h
```diff
@@ -243,6 +243,8 @@ std::string CPL_DLL GDALGridGetParserUsage();
 
 std::string CPL_DLL GDALBuildVRTGetParserUsage();
 
+std::string CPL_DLL GDALContourGetParserUsage();
```
rouault, line 246: as gdal_contour is a bin only utility, this function should not be exposed in gdal_utils_priv.h as it can't be used by the GDAL library

apps/gdal_contour.cpp
```diff
@@ -33,45 +33,193 @@
...
+        .scan<'g', double>()
+        .action(
+            [psOptions](const auto &d)
+            {
+                psOptions->bNoDataSet = true;
+                psOptions->dfNoData = CPLAtofM(d.c_str());
+            })
+        .help(_("Input pixel value to treat as \"nodata\"."));
+
+    argParser->add_argument("-f")
```
rouault, line 132: you can use add_output_format_argument()

apps/gdal_contour.cpp
```diff
@@ -89,225 +237,87 @@ static void CreateElevAttrib(const char *pszElevAttrib, OGRLayerH hLayer)
...
 
     argc = GDALGeneralCmdLineProcessor(argc, &argv, 0);
+    if (argc < 1)
+        exit(-argc);
 
     /* -------------------------------------------------------------------- */
     /*      Parse arguments.                                                */
     /* -------------------------------------------------------------------- */
+
+    CPLStringList aosArgv;
```
rouault, line 264: This list doesn't seem useful. You can directly use argc / argv

rouault:
When the PR is ready, can you separate the changes in gdalargumentparser.h/cpp in a separate initial commit?

elpaso:
> When the PR is ready, can you separate the changes in gdalargumentparser.h/cpp in a separate initial commit?

I can remove the changes completely and put them in a separate PR, I'm not using them here, I just though they were right in case one wanted to further modify the arg.

apps/gdal_utils_priv.h
elpaso: done

apps/gdal_contour.cpp
elpaso: done

Issue #9751, opened 2024-04-24
NumPy 2.0 support


jakirkham:
### Feature description

NumPy 2.0 is coming out soon ( https://github.com/numpy/numpy/issues/24300 ). NumPy 2.0.0rc1 packages for conda & wheels came out 3 weeks back ( https://github.com/numpy/numpy/issues/24300#issuecomment-2030603395 )

To prepare for NumPy 2.0, it might be worthwhile to start testing against NumPy 2 in CI

NumPy has put out [a migration guide]( https://numpy.org/devdocs/numpy_2_0_migration_guide.html#numpy-2-migration-guide ). More details are in [the release notes]( https://numpy.org/devdocs/release/2.0.0-notes.html )

If GDAL make use of NumPy's C API (and produces wheels that use it), having a release of GDAL with wheels built against NumPy 2.0.0rc1 would be helpful to ensure NumPy 1 & 2 compatible wheels (as wheels built against NumPy 1 won't be compatible with NumPy 2). More details in [this NumPy 2 ABI doc]( https://numpy.org/devdocs/dev/depending_on_numpy.html#numpy-2-abi-handling )

Also as NumPy is tracking ecosystem support for NumPy 2.0, it would be helpful to share GDAL current support status in issue (with any plans): https://github.com/numpy/numpy/issues/26191

rouault:
> I gave a try a few weeks ago locally, and everything went fine without requiring any change

Actually I should have written any change in the library itself. A few adjustments were made in the test suite per #9044

rouault:
Any Python packaging champion here (perhaps @pl-kevinwurster @rcoup) ?
[...]

PR #9752, opened 2024-04-24
CI: tweak alpine configuration to use numpy 2.0.0rc1

rouault:
Refs #9751

This commit should be reverted once alpine:edge switches to numpy 2 as it disables pyarrow related tests (however the alpine 32 bit config still has it)

### DIFF
--- a/.github/workflows/alpine/Dockerfile.ci
+++ b/.github/workflows/alpine/Dockerfile.ci
@@ -51,10 +51,6 @@ RUN apk add \
     poppler-dev \
     proj-dev \
     proj-util \
-    py3-pyarrow \
-    py3-pyarrow-pyc \
-    py3-numpy \
-    py3-numpy-dev \
     py3-pip \
     py3-setuptools \
     python3-dev \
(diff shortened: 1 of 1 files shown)


jakirkham:
Thanks Even! 🙏

FWIW Arrow 16.0.0 should have NumPy 2 compatibility

Refs:
* [Release notes]( https://arrow.apache.org/release/16.0.0.html )
* https://github.com/apache/arrow/issues/39532

rouault:
@dbaston Maybe you have some ideas of what might happen . In https://github.com/OSGeo/gdal/actions/runs/8824768156/job/24227829652?pr=9752 , I got a weird failure during test execution where pyarrow import worked (but was broken, probably because it relies on numpy 1 whereas this PR installs numpy2), despite I removed it from being installed in the Dockerfile.ci by this pull request. It seems like there was some remain of it from a previous Docker image where it was installed.

rouault:
@dbaston Please ignore my above comment. Nothing caching related. The issue is actually solved with:
```
# apache-arrow-dev actually comes with an embedded pyarrow version, which is not py3-pyarrow, and is non functional !
RUN mv /usr/lib/python3.12/site-packages/pyarrow /usr/lib/python3.12/site-packages/pyarrow.disabled
```

rouault:
closing that one as superseded per #9753

PR #9753, opened 2024-04-25
CI: add a configuration testing numpy 2.0.0rc1

rouault:
Refs #9751

This commit should be reverted once alpine:edge switches to numpy 2


jakirkham:
Thanks Even! 🙏

PR #9754, opened 2024-04-25
docs: pmtiles.rst - casing

missinglink:
update casing of `JSon` -> `JSON`


missinglink:
@rouault there are more of these in the docs, I saw some in the geojson driver doc's today too.

Would it be helpful to PR to change the others too? I wasn't sure if that is someone's prefered abbreviation method or meant to mean JavaScript/JSON or something like that?

rouault:
> there are more of these in the docs,

should be dealt by https://github.com/OSGeo/gdal/pull/9764

PR #9756, opened 2024-04-25
Update ogr2ogr.rst: some people thought it was possible to add GCPs to a vector file!

jidanni:


rouault:
I've opted for a variant of your proposal in 0efd3659cf that removes the confusing "Add the indicated..."

doc/source/programs/ogr2ogr.rst
```diff
@@ -510,7 +510,9 @@ output coordinate system or even reprojecting the features during translation.
 .. option:: -gcp <ungeoref_x> <ungeoref_y> <georef_x> <georef_y> [<elevation>]
 
     Add the indicated ground control point. This option may be provided
-    multiple times to provide a set of GCPs.
+    multiple times to provide a set of GCPs. (Note that unlike gdal_edit.py,
+    the GCPs here are used for calculating reprojection during translation,
```
jratike80, line 514: I think that instead of gdal_edit.py the opposite example could be gdal_translate which is kind of pair with ogr2ogr https://gdal.org/programs/gdal_translate.html#cmdoption-gdal_translate-gcp

jidanni: Holy moly. Gdaltranslate man page has the same problem. Thanks for alerting me.

OK, folks. Please revise all man pages with a -gcp option on them, to say clearly if their -gcp option affects calculations, or simply appends the GCPs to the output stream.

Currently I already guessed wrong several times and it is very sad! Thanks!

PR #9761, opened 2024-04-25
pyproject.toml: use numpy>=2.0.0rc1 for python >=3.9

rouault:
Refs #9751

### DIFF
--- a/swig/python/pyproject.toml
+++ b/swig/python/pyproject.toml
@@ -1,5 +1,8 @@
 [build-system]
-requires = ["setuptools>=67.0.0", "oldest-supported-numpy", "wheel"]
+requires = ["setuptools>=67.0.0",
+            "wheel",
+            "oldest-supported-numpy; python_version=='3.8'",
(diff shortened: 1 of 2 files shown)


pl-kevinwurster:
@rouault You may want to consider adding something to the PyPI readme about how users who:

* Are using Python 3.9+
* Are using GDAL 3.9+
* Cannot yet install Numpy 2.0 for some reason

Can use:

```
$ pip install "numpy==< their required version >"
$ pip install "gdal" --no-build-isolation
```

Otherwise I think these users will get an error when attempting to install the `gdal` Python package, right? I think the local build phase would attempt to install Numpy 2.0, and potentially error out.

rouault:
> You may want to consider adding something to the PyPI readme about how users who:

good suggestion. Done

hobu:
I recently refactored the PDAL Python bindings build system to use `scikit-build-core`, which enables ditching setup.py and using a "proper" pyproject.toml-based build configuration. 

https://github.com/PDAL/python/

PR #9771, opened 2024-04-26
Update csv.rst copying -oo line from above for -lco too

jidanni:


doc/source/drivers/vector/csv.rst
```diff
@@ -421,6 +421,9 @@ be stored in RAM temporarily before flushing to disk.
 Layer Creation options
 ----------------------
 
+The following layer creation options can be specified
+(typically with the -lco name=value parameters of ogrinfo or ogr2ogr):
+
```
jratike80, line 426: Most drivers have layer creation options, and quite many also dataset creation options. If that needs clarifications then instead of repeating the same note in every driver document then maybe do it here https://gdal.org/programs/ogr2ogr.html#cmdoption-ogr2ogr-lco (ogrinfo does not have layer creation options). Or write a new GDAL vocubulary document that collects the common keywords and how and where they are used.

rouault: yes, this was discussed recently in https://lists.osgeo.org/pipermail/gdal-dev/2024-March/058784.html. We should probably use a substitution  mechanism to have the same introduction text in all drivers, instead of doing manual modification in each of them. I'm ticketing that as https://github.com/OSGeo/gdal/issues/9775 and closing that PR

rouault:
Cf discussion at

PR #9773, opened 2024-04-26
MiraMonVector driver. Adding tests + one fix

AbelPau:
## What does this PR do?
New tests to improve coverage:

- Relation between polygons and its linestrings
- Test that a field cannot be added after feature creation.

Fix: Change the limit from 1000 to -1 in CPLReadLine2L() in order to not limit the read lines.

## Tasklist

 - [x] Add test case(s)


rouault:
> Fix: Change the limit from 1000 to -1 in CPLReadLine2L() in order to not limit the read lines.

This was actually changed from 1024 to 10000. Which looks good to me as we prefer having a limit

AbelPau:
> > Fix: Change the limit from 1000 to -1 in CPLReadLine2L() in order to not limit the read lines.
> 
> This was actually changed from 1024 to 10000. Which looks good to me as we prefer having a limit

Ok, no problem. I think 1024 is on the limit but 10000 it's unreachable! So, good

PR #9784, opened 2024-04-26
Doc: updates to support building with upcoming update proj-docs Docker image updated to Ubuntu 24.04

rouault:
I've tested locally those changes with a local build of the updated proj-docs Docker image per https://github.com/OSGeo/PROJ/pull/4128


rouault:
Doc build passes with the updated proj-docs Docker image

PR #9785, opened 2024-04-26
LOS: Implement simple violation location reporting

Ryanf55:
## What does this PR do?

* Implements basic support for reporting a location where LOS  is violated. It reports the location of the first detected violation.
* It's not symmetric due to bresenham's optimizations, but it will at least return one location where terrain is violated (if it is)

#9562 

## Tasklist

 - [x] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS: Ubuntu 22
* Compiler: G++11

## Instructions
```bash
cmake -S . -B build -DBUILD_TESTING=ON -DCMAKE_BUILD_TYPE=Debug
cmake --build build --target gdal_unit_test -j `nproc`
./build/autotest/cpp/gdal_unit_test --gtest_filter="test_alg.GDALIsLineOfSightVisible*"
# Or, with GDB
 gdb ./build/autotest/cpp/gdal_unit_test
>>> r --gtest_filter="test_alg.GDALIsLineOfSightVisible*"
```

### DIFF
--- a/alg/los.cpp
+++ b/alg/los.cpp
@@ -177,13 +177,11 @@ static bool IsAboveTerrain(const GDALRasterBandH hBand, const int x,
  *
  * @param[out] pnxTerrainIntersection The X location where the LOS line
  *             intersects with terrain, or nullptr if it does not intersect
- *             terrain. Not implemented currently (*pnxTerrainIntersection
- *             is set to -1)
+ *             terrain.
  *
  * @param[out] pnyTerrainIntersection The Y location where the LOS line
  *             intersects with terrain, or nullptr if it does not intersect
- *             terrain. Not implemented currently (*pnyTerrainIntersection
- *             is set to -1)
+ *             terrain.
  *
  * @param papszOptions Options for the line of sight algorithm (currently ignored).
  *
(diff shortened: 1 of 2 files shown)


rouault:
Can you update autotest/alg/los.py or autotest/cpp/test_alg.cpp?

Ryanf55:
> Can you update autotest/alg/los.py or autotest/cpp/test_alg.cpp?

Yup. Are you on with heuristic based testing that doesn't check for exact locations reported because it's currently an implementation detail that doesn't matter?
More importantly, the values get set to something between points A and B.

rouault:
> Are you on with heuristic based testing that doesn't check for exact locations

I would just test the current returned values. If one day the algorithm is changed, we can always update the expected results

rouault:
code formatting issue: cf https://github.com/OSGeo/gdal/actions/runs/8880363204/job/24380232858?pr=9785

Ryanf55:
> code formatting issue: cf https://github.com/OSGeo/gdal/actions/runs/8880363204/job/24380232858?pr=9785

Linter is fixed, only failure left is unrelated on windows: https://github.com/OSGeo/gdal/actions/runs/8887950310/job/24404034582?pr=9785#step:13:2102

rouault:
@Ryanf55 Can you adjust the documentation of the function that currently says that the intersection reporting is not currently implemented?

Ryanf55:
Done. Note - the implementation sets the values to -1 if there is no intersection. Do you think I should re-order that to set it to `nullptr`?

rouault:
> Do you think I should re-order that to set it to `nullptr`?

I don't understand what you mean. You can't set to nullptr the value of a ``int*``

PR #9789, opened 2024-04-28
Update gdaltransform.rst to hint that it is perfectly vector ready

jidanni:
Yes, `<pixel>` and `<line>` should just be called `<x>` and `<y>`, but I'm not that bold to rip things up here.

By the way I wonder if there is any _vector_ format that can contain GCPs so one needn't use `-gcp` but instead just have them in `srcfile`.

### DIFF
--- a/doc/source/programs/gdaltransform.rst
+++ b/doc/source/programs/gdaltransform.rst
@@ -109,7 +109,7 @@ projection,including GCP-based transformations.
 
 .. option:: -gcp <pixel> <line> <easting> <northing> [<elevation>]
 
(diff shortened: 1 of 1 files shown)


doc/source/programs/gdaltransform.rst
```diff
@@ -109,7 +109,7 @@ projection,including GCP-based transformations.
 
 .. option:: -gcp <pixel> <line> <easting> <northing> [<elevation>]
 
-    Provide a GCP to be used for transformation (generally three or more are required)
+    Provide a GCP to be used for transformation (generally three or more are required. Also unlike gdallocationinfo output, pixel and line need not be integers.)
```
rouault, line 112:
```suggestion
    Provide a GCP to be used for transformation (generally three or more are required). Pixel and line need not be integers.
```

PR #9790, opened 2024-04-28
Update ogr2ogr.rst to reveal all "unlisted" columns

jidanni:


rouault:
I find the addition regarding the FID column even more confusing, given that it is not used in the example, as the topic here is to just extract geometries. So I'm not applying this. I've just removed in fbdcb69066 the LIMIT 2

PR #9791, opened 2024-04-28
Update gdaltransform.rst replacing wimpy example with super high powered one

jidanni:


rouault:
I'm not feeling like this is appropriate for the gdaltransform manual. It goes far beyond demonstrating how to use the tool, and is specific to a very particular use case, and it involves other tools like sed or perl. This would rather belong to a blog post or a tutorial with advanced examples of GDAL

PR #9809, opened 2024-04-29
gdal2tiles: support an input dataset name not being a file

### LINKED ISSUE #9808: wish: gdal2tiles.py support for vrt:// syntax
### Feature description

gdal2tiles.py tests if a path is a regular file using info from `VSIStatL`, which means it cannot support input syntax like

```
"vrt://gdrivers/data/small_world.tif?outsize=200%,200%"
```

It would be handy to accept such input, because we can inline conversion to Byte with scaling without materializing full VRT, for example: 

```
"vrt://gdrivers/data/float64.tif?scale=true&ot=Byte"
```

 and this could work with remote sources at vsi endpoints as far as I can tell.

rouault:
(fixes #9808)


mdsumner:
very nice thank you!

PR #9815, opened 2024-04-30
MiraMon: Lost changes + update DBF date

AbelPau:
## What does this PR do?
Recovers lost code from april 14th-15th-16th. A Test has been modified because of decimal modifications:

- Writing the appropriate decimals for each case for length, perimeter, area, etc.: if a layer is in longitude/latitude, it has more decimals than if it is in UTM. If it is unknown, then it is deduced according to the location of the coordinates of the first point (uncommon case). 
- Simplification of the variables used for writing to the DBF. 
- Changes in the names of some variables to facilitate code readability. 
- Exhaustive explanation of the function MM_SprintfDoubleSignifFigures(). 
- Correction in an internet address of specifications.
- Assignment of the units 'm' in special cases where it is better to include the units than not.
- Correction of error in the case of Triangle entities

And something new: Fix little error about date in DBF. No test needed. 

## Tasklist

 - [x] Add test case(s)

### DIFF
--- a/ogr/ogrsf_frmts/miramon/mm_wrlayr.c
+++ b/ogr/ogrsf_frmts/miramon/mm_wrlayr.c
@@ -97,7 +97,8 @@ int MMCheckVersionForFID(struct MiraMonVectLayerInfo *hMiraMonLayer,
                          MM_INTERNAL_FID FID);
 
 // Extended DBF functions
-int MMCreateMMDB(struct MiraMonVectLayerInfo *hMiraMonLayer);
+int MMCreateMMDB(struct MiraMonVectLayerInfo *hMiraMonLayer,
+                 struct MM_POINT_2D *pFirstCoord);
 int MMAddDBFRecordToMMDB(struct MiraMonVectLayerInfo *hMiraMonLayer,
                          struct MiraMonFeature *hMMFeature);
 int MMAddPointRecordToMMDB(struct MiraMonVectLayerInfo *hMiraMonLayer,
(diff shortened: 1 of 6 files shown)


ogr/ogrsf_frmts/miramon/mm_wrlayr.c
```diff
@@ -6467,6 +6507,34 @@ MMWriteValueToszStringToOperate(struct MiraMonVectLayerInfo *hMiraMonLayer,
...
+    const struct MM_FIELD *camp, const char *valor)
+{
+    if (!hMiraMonLayer)
+        return 1;
+
+    if (!camp)
+        return 0;
+
+    if (MMResizeStringToOperateIfNeeded(hMiraMonLayer,
+                                        camp->BytesPerField + 10))
```
rouault, line 6521: why +10 ?

rouault:
Can you also propose a more descriptive commit message about what those changes do ?

ogr/ogrsf_frmts/miramon/mm_wrlayr.c
AbelPau: To avoid resizing everytime 1 bytes is needed. But it's not important. It could be +1 (for the \0 byte)

rouault: ok, thanks

AbelPau:
> Can you also propose a more descriptive commit message about what those changes do ?

Sure, I had to go, but I was going to do it at sometime. Thanks.

PR #9831, opened 2024-05-02
MiraMonVector: Little fixes in metadata files

AbelPau:
## What does this PR do?
Little fixes in MiraMon metadata files. More specifically:
 

- Making perimeter, area, and longitude symbolically representable. 
- Writing the correct HRS in polygon metadata.
- Writing CreationDate in metadata without a wrong timekeeper +00:00 
- Making units visible in all private fields in metadata (even if then don't have units, it's a defect behavior). 
- Making perimeter, area, and longitude symbolically representable 
- Making nodes visibles
 
No addicional tests are needed because are tiny modifications in some aspects used when visualizing layers in MiraMon 

## Tasklist


AbelPau:
@rouault this PR is essential. But if it can be included in the next release version, it would be ok.
I leave it checking and i'll make ready for review when it finishes.

PR #9832, opened 2024-05-02
[ogrtindex] Use GDALArgumentParser

elpaso:
```
Usage: ogrtindex [--help] [--long-usage] [--help-general]
                 [-lnum <n>]... [-lname <name>]... [-of <output_format>] [-tileindex <tileindex>] [-write_absolute_path]
                 [-skip_different_projection] [-t_srs <target_srs>] [-src_srs_name <field_name>]
                 [-src_srs_format {AUTO|WKT|EPSG|PROJ}] [-accept_different_schemas]
                 <output_dataset> <src_dataset>...

Program to generate a UMN MapServer compatible tile index for a set of OGR data sources.

Positional arguments:
  <output_dataset>                      Name of the output dataset. 
  <src_dataset>                         Name of the source dataset(s). [nargs: 1 or more] 

Optional arguments:
  -h, --help                            Shows short help message and exits. 
  --long-usage                          Shows long help message and exits. 
  --help-general                        Report detailed help on general options. 
  -lnum <n>                             Add layer number <n> from each source file in the tile index. [may be repeated]
  -lname <name>                         Add layer named <name> from each source file in the tile index. [may be repeated]
  -of <output_format>                   Output format. 
  -tileindex <tileindex>                Name to use for the dataset name. [default: "LOCATION"]
  -write_absolute_path                  Write absolute path of the source file in the tile index. 
  -skip_different_projection            Skip layers that are not in the same projection as the first layer. 
  -t_srs <target_srs>                   Extent of input files will be transformed to the desired target coordinate reference system. 
  -src_srs_name <field_name>            Name of the field to store the SRS of each tile. 
  -src_srs_format {AUTO|WKT|EPSG|PROJ}  Format of the source SRS to store in the tile index file. 
  -accept_different_schemas             Disable check for identical schemas for layers in input files. 

For more details, see the full documentation for ogrtindex at
https://gdal.org/programs/ogrtindex.html
```

### DIFF
--- a/apps/ogrtindex.cpp
+++ b/apps/ogrtindex.cpp
@@ -98,26 +98,28 @@ MAIN_START(nArgc, papszArgv)
 
     GDALArgumentParser argParser{"ogrtindex", true};
 
-    argParser.add_description("Program to generate a UMN MapServer compatible "
-                              "tile index for a set of OGR data sources.");
+    argParser.add_description(
+        _("Program to generate a UMN MapServer compatible "
+          "tile index for a set of OGR data sources."));
 
     argParser.add_epilog(
-        "For more details, see the full documentation for ogrtindex "
-        "at\nhttps://gdal.org/programs/ogrtindex.html");
+        _("For more details, see the full documentation for ogrtindex "
+          "at\nhttps://gdal.org/programs/ogrtindex.html"));
 
     argParser.add_argument("-lnum")
         .metavar("<n>")
         .append()
         .scan<'i', int>()
         .store_into(anLayerNumbers)
-        .help("Add layer number <n> from each source file in the tile index.");
+        .help(
+            _("Add layer number <n> from each source file in the tile index."));
 
     argParser.add_argument("-lname")
         .metavar("<name>")
         .append()
         .store_into(aosLayerNames)
-        .help(
-            "Add layer named <name> from each source file in the tile index.");
+        .help(_(
+            "Add layer named <name> from each source file in the tile index."));
 
     argParser.add_output_format_argument(osOutputFormat);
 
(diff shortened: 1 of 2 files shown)


apps/ogrtindex.cpp
```diff
@@ -120,132 +67,168 @@ static void Usage(bool bIsError)
...
-    bool bSrcSRSFormatSpecified = false;
     SrcSRSFormat eSrcSRSFormat = FORMAT_AUTO;
     size_t nMaxFieldSize = 254;
+    std::vector<std::string> aosSrcDatasets;
+    std::vector<std::string> aosLayerNames;
+    std::vector<int> anLayerNumbers;
+
+    GDALArgumentParser argParser{"ogrtindex", true};
+
+    argParser.add_description("Program to generate a UMN MapServer compatible "
```
rouault, line 101: Dummy``_()`` gettext macro missing here and below

apps/ogrtindex.cpp
```diff
@@ -120,132 +67,168 @@ static void Usage(bool bIsError)
...
+        .help("Write absolute path of the source file in the tile index.");
+
+    argParser.add_argument("-skip_different_projection")
+        .flag()
+        .store_into(bSkipDifferentProjection)
+        .help("Skip layers that are not in the same projection as the first "
+              "layer.");
+
+    argParser.add_argument("-t_srs")
+        .metavar("<target_srs>")
```
rouault, line 143:
```suggestion
        .metavar("<srs_def>")
```

apps/ogrtindex.cpp
elpaso: done

PR #9834, opened 2024-05-02
Deprecate GDALDestroyDriverManager() and OGRCleanupAll() for GDALCleanup()

rouault:
and another cleanup

CC @elpaso


elpaso (approved):
Thanks! Much cleaner now.

PR #9835, opened 2024-05-02
MiraMonVector fix: better condition & variable type

AbelPau:
## What does this PR do?
Fixes subtype condition: This is not an error, but it could become one if the subtype OFSTBoolean weren't set to "1". So, I'm doing that only to clarify the code.

Fixes the variable type: Throughout the code, the type is MM_EXT_DBF_N_FIELDS instead of "int". So, I'm leaving it as it should be.

No tests needed. No Backport release/3.9 needed.

## Tasklist


ogr/ogrsf_frmts/miramon/ogrmiramonlayer.cpp.bak
```diff
@@ -0,0 +1,2983 @@
+/******************************************************************************
```
rouault, line 1: this file should be removed

AbelPau: Ouch! I didn't see that. Sorry!

PR #9836, opened 2024-05-02
CMake: add html, man, latexpdf, doxygen, doxygen_check_warnings, clean_doc targets

### LINKED ISSUE #5484: doc/CMakeLists.txt is broken
The 2, optional, targets it defines are broken: "manpages" and "documents"

rouault:
Fixes #5484

Updated dev_documentation.rst:
```rst
Building documentation
######################

Documentation can be generated with Makefile targets, from the ``doc`` subdirectory
of the GDAL source repository (only on Unix systems).

The following targets are available:

* ``html``: build HTML documentation into the ``doc/build/html`` directory, where
  they can be viewed using a web browser.

* ``man``: build MAN pages into the ``doc/build/man`` directory.

* ``latexpdf``: build PDF documentation into the ``doc/build/pdf`` directory

* ``doxygen``: regenerate API Doxygen XML and HTML output, that is used by the
  ``html`` target. Doxygen content is not automatically rebuilt when source files
  are modified, hence this target must be explicitly run to refresh it.

* ``doxygen_check_warnings``: same as ``doxygen``, but errors out when Doxygen
  emits a warning (the ``doxygen`` target is tolerant to Doxygen warnings).
  This can be useful to reproduce one of the continuous integration checks that
  verifies that there are no Doxygen warnings.
  Requires Doxygen >= 1.9.3 to be warning free.

* ``clean``: clean the ``doc/build`` directory.

It is also possible to run those targets as CMake targets. In that case, the
output directory will be the ``doc/build`` subdirectory of the CMake
build directory. To only clean the documentation, the ``clean_doc`` target can
be invoked.
Note: those CMake targets are only available if the CMake BUILD_DOCS=ON variable
is set (it is set by default if build preconditions are met, that is if Doxygen,
Sphinx and make are available)
```

CC @dbaston @abellgithub

### DIFF
--- a/doc/CMakeLists.txt
+++ b/doc/CMakeLists.txt
@@ -1,170 +1,71 @@
 # CMake4GDAL project is distributed under MIT license. See accompanying file LICENSE.txt.
 
-# Documentation rules
 find_package(Doxygen)
-if (DOXYGEN_FOUND)
-  set(TARGET_HTML_DIR ${CMAKE_BINARY_DIR}/html)
-  file(MAKE_DIRECTORY ${TARGET_HTML_DIR})
-  # Generate translated docs. Should go first, because index.html page should be overwritten with the main one later
-  add_custom_command(
-    OUTPUT ${TARGET_HTML_DIR}/header.html ${TARGET_HTML_DIR}/footer.html ${TARGET_HTML_DIR}/stylesheet.css
-    COMMAND ${DOXYGEN_EXECUTABLE} -w html header.html footer.html stylesheet.css
-    WORKING_DIRECTORY ${TARGET_HTML_DIR})
-  # Generate Russian document
-  set(GDAL_DOXYGEN_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}")
-  set(GDAL_DOXYGEN_PROJECT_NAME "RU")
-  set(GDAL_DOXYGEN_LANGUAGE "Russian")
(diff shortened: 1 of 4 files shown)


dbaston:
Would it make sense to remove the Makefile and use CMake directly to build the docs? One benefit is that API doc targets could depend on non-doc targets where appropriate, so we could remove workarounds like

> update the timestamp of the rst files associated with the page where the documentation appears (e.g., touch doc/source/api/python/osgeo.ogr.rst)

from https://gdal.org/development/dev_documentation.html#python-api-documentation

rouault:
> Would it make sense to remove the Makefile and use CMake directly to build the docs?

yes, such further enhancements are totally reasonable. I just wanted to do something relatively quick to get documentation building easily accessible from CMake. Would you like following up on this in subsequent PRs ?

(I'm wondering if we could have things not requiring a Unix environment so doc building is available from non-Cygwin Windows (provided folks have installed Doxygen, sphinx, etc.), but that would be more a stretch goal...)

dbaston:
> Would you like following up on this in subsequent PRs ?

Yes, I can work on this.

rouault:
> Would it make sense to remove the Makefile and use CMake directly to build the docs?

Doing the 3.9.0 release, I see that, besides potential CI changes (doc_build.yml), the "doc/build_doc_snapshot.sh" script might also be impacted

PR #9845, opened 2024-05-03
[gdal_footprint] Use GDALArgumentParser

elpaso:


apps/gdal_footprint_lib.cpp
```diff
@@ -116,12 +117,177 @@ struct GDALFootprintOptions
...
+        .default_value("location")
+        .store_into(psOptions->osLocationFieldName)
+        .help(_("Specifies the name of the field in the resulting vector "
+                "dataset where the path of the input dataset will be stored."));
+
+    argParser->add_argument("-no_location")
+        .flag()
+        .store_into(psOptions->bClearLocation)
+        .help(_("Turns off the writing of the path of the input dataset as a "
+                "field in the output vector dataset."));
```
rouault, line 249: those 2 should actually be mutually exclusive

apps/gdal_footprint_lib.cpp
```diff
@@ -116,12 +117,177 @@ struct GDALFootprintOptions
...
+        .help(_("Enables writing the absolute path of the input dataset."));
+
+    argParser->add_layer_creation_options_argument(psOptions->aosLCO);
+
+    argParser->add_argument("-dsco")
+        .metavar("<NAME>=<VALUE>")
+        .append()
+        .action([psOptions](const std::string &s)
+                { psOptions->aosDSCO.AddString(s.c_str()); })
+        .help(_("Dataset creation option (format specific)."));
```
rouault, line 263: that one could be a candidate for a helper in GDALArgumentParser . It is at least used by gdal_contour and ogr2ogr

apps/gdal_footprint_lib.cpp
```diff
@@ -116,12 +117,177 @@ struct GDALFootprintOptions
...
+        group.add_argument("-ovr")
+            .metavar("<index>")
+            .scan<'i', int>()
+            .store_into(psOptions->nOvrIndex)
+            .help(_("Defines which overview level of source file must be used, "
+                    "when overviews are available on the source raster."));
+
+        group.add_argument("-srcnodata")
+            .metavar("<value>")
+            .nargs(nargs_pattern::at_least_one)
```
rouault, line 168: I'm not too sure about that one. The current parser accepts things like ``-srcnodata "1 2"`` (the double quotes are actually removed by the shell, so in terms of argv[], this is 2 values: ``-srcnodata`` and ``1 2``). I understand that ``nargs(nargs_pattern::at_least_one)`` would allow to specify ``-srcnodata 1 2`` (which would be more natural), but is that stored correctly into osSrcNoData, which expects a space separator?

In gdalbuildvrt, I've just used the following to only support the syntax with values packed together in the same argument
```
    argParser->add_argument("-srcnodata")
        .metavar("\"<value>[ <value>]...\"")
        .store_into(psOptions->osSrcNoData);
```

apps/gdal_footprint_lib.cpp
elpaso: I had the same thought, I will do it in a follow-up PR.

apps/gdal_footprint_lib.cpp
elpaso: You are right, nargs_pattern::at_least_one or ::any leads to problems when the argument is the last one and there are positional arguments. I will use the same code as you did in gdalbuildvrt.

elpaso: done

apps/gdal_footprint_lib.cpp
elpaso: done

PR #9848, opened 2024-05-04
Update gdaltransform.rst adding -i to GCP example

jidanni:


rouault:
curious what @jratike80 thinks about that? Personally I'm not sure if that's a good move as I find it adds a bit of complication/confusion: the user has to understand that "c" is an expression, that will be evaluated; the forward direction is done twice with the initial example and the one you added at the end.

jidanni:
I would love to put the GCPs in a file, and then use that file twice or thrice, but I don't know how.

It seems the only file formats that can store GCPs are some raster image formats? I would love to make an example where I used gdal_edit to add the GCPs to a file, and then reuse that file twice a (the third one, 0,0 can indeed be dropped.)

jratike80:
> I would love to put the GCPs in a file, and then use that file twice or thrice, but I don't know how.

Maybe optfile works? 
https://gdal.org/programs/raster_common_options.html#cmdoption-optfile

jratike80:
> curious what @jratike80 thinks about that?

The example is a bit too challenging for me. For understanding what it is about I would rather see the whole thing in an illustrated tutorial.

jidanni:
That --optfile is a brilliant solution. I'll start working on a new PR!

PR #9854, opened 2024-05-05
gdal_retile.py: resolution handling

DirkThalheim:
## What does this PR do?

This adds an option to gdal_retile for calculating image pyramids for specific resolutions/pixel sizes. This can be used instead of specifying the number of levels. For example: 
  gdal_retile.py -pyramidOnly -ps 4096 4096 -res 1.0,5.0,25.0 -targetDir out/pyramid input/*.tif 
Will create an image pyramid with 
* level 1: 1 m/pixel
* level 2: 5 m/pixel
* level 3: 25 m/pixel
(assuming the input data uses as units meter)

## Tasklist

 - [ ] Add test case(s)
 - [x] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:


rouault:
Existings tests are broken by the changes

PR #9867, opened 2024-05-06
Doc: Add Python docstring for gdal.FillNodata

dbaston:
Side note, I found the first result in the example a little surprising.


rouault:
> I found the first result in the example a little surprising.

what exactly was surprising ? at first sight, that seems plausible to me

dbaston:
I would have expected the results in the two columns to be the same.

dbaston:
Disregard, I posted a different example than the one I was troubleshooting.

PR #9883, opened 2024-05-09
MiraMonVector: adding which polygon cycles the linestring in the linestring metadata information

AbelPau:
## What does this PR do?
Fix: adding which polygon cycles the stringline in the stringline metadata information

It's not an error but it's useful metadata information.

## Tasklist

 - [x] Almost all CI builds and checks have passed
 - [x] Confirm that Linux Builds / Alpine gcc failure is not this pull request fault.

### DIFF
--- a/ogr/ogrsf_frmts/miramon/mm_wrlayr.c
+++ b/ogr/ogrsf_frmts/miramon/mm_wrlayr.c
@@ -5584,14 +5584,22 @@ static int MMWriteMetadataFile(struct MiraMonVectorMetaData *hMMMD)
         }
     }
 
-    // Writing OVERVIEW:ASPECTES_TECNICS in polygon metadata file.
-    // ArcSource=fitx_pol.arc
-    if (hMMMD->ePlainLT == MM_LayerType_Pol)
+    if (hMMMD->ePlainLT == MM_LayerType_Pol && hMMMD->aArcFile)
     {
+        // Writing OVERVIEW:ASPECTES_TECNICS in polygon metadata file.
+        // ArcSource=fitx_pol.arc
         fprintf_function(pF, LineReturn "[%s]" LineReturn,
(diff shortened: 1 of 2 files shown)


AbelPau:
It seems that Linux Builds / Alpine, gcc fails for a problem out of this PR scope.

ogr/ogrsf_frmts/miramon/mm_wrlayr.c
```diff
@@ -5868,7 +5878,11 @@ static int MMWriteVectorMetadataFile(struct MiraMonVectLayerInfo *hMiraMonLayer,
                    sizeof(hMMMD.hBB));
             hMMMD.pLayerDB = nullptr;
         }
-        return MMWriteMetadataFile(&hMMMD);
+        hMMMD.aArcFile = strdup_function(
+            get_filename_function(hMiraMonLayer->MMPolygon.pszLayerName));
+        nResult = MMWriteMetadataFile(&hMMMD);
+        free_function(hMMMD.aArcFile);
```
rouault, line 5884: shouldn't hMMMD.aArcFile be set to NULL after free_function, to avoid a potential reuse of freed memory ?

AbelPau: I think it's not necessary because it's in a structure created at the same function and set to 0.
memset(&hMMMD, 0, sizeof(hMMMD));

And after the use the function finalizes with return 0 and this structure is not used outside.

AbelPau:
> @rouault It seems that Linux Builds / Alpine, gcc fails for a problem out of this PR scope. Is this true?

It seems thant a second try has make it pass.

ogr/ogrsf_frmts/miramon/mm_wrlayr.c
rouault: ok, I missed this was a local structure

PR #9884, opened 2024-05-09
[gdal_create] Use GDALArgumentParser

elpaso:


apps/gdal_create.cpp
```diff
@@ -31,45 +31,169 @@
...
+        .help(_("Set the size of the output file."));
+
+    argParser->add_argument("-bands")
+        .metavar("<count>")
+        .store_into(psOptions->nBandCount)
+        .help(_("Set the number of bands in the output file."));
+
+    argParser->add_argument("-burn")
+        .metavar("<value>")
+        .nargs(nargs_pattern::at_least_one)
```
rouault, line 128: .append() should probably be added too, since the current synopsis indicates "[-burn <value>]..."

Testing a bit, I see we currently support all of 3 below variants:
```
gdal_create -outsize 1 1 -bands 3 -burn 1 2 3 out.tif
gdal_create -outsize 1 1 -bands 3 -burn "1 2 3" out.tif
gdal_create -outsize 1 1 -bands 3 -burn 1 -burn 2 -burn 3 out.tif
```
the later one is the only one actually documented (or at least this is my understanding of the synopsis and documentation), but the one tested per test_gdal_create_pdf_tif is `` -burn 1 2``
If that's not too inconvenient, we could try to still support the 3 variants?

elpaso: done

apps/gdal_create.cpp
```diff
@@ -31,45 +31,186 @@
...
+            {
+                if (s.find(' ') != std::string::npos)
+                {
+                    char **papszTokens = CSLTokenizeString(s.c_str());
+                    for (int i = 0; papszTokens[i] != nullptr; i++)
+                    {
+                        psOptions->adfBurnValues.push_back(
+                            CPLAtof(papszTokens[i]));
+                    }
+                    CSLDestroy(papszTokens);
```
rouault, line 142: Slightly more C++ way avoiding explicit CSLDestroy():
```suggestion
                    const CPLStringList aosTokens(CSLTokenizeString(s.c_str()));
                    for (int i = 0; i < aosTokens.size(); i++)
                    {
                        psOptions->adfBurnValues.push_back(
                            CPLAtof(aosTokens[i]));
                    }
```

rouault:
Failure on [build-windows-msys2-mingw](https://github.com/OSGeo/gdal/actions/runs/9031429017/job/24817643775?pr=9884#logs) unrelated, and likely related to msys2 having upgraded to gcc 14

PR #9895, opened 2024-05-10
MiraMonVector: Adding more matchs in the look-up table MM_m_idofic.csv

AbelPau:
## What does this PR do?
Adds more EPSG Spatial Reference Systems in the MiraMon look-up table MM_m_idofic.csv
With this update, MiraMon is expected to recognize more EPSG spatial reference systems.

No afectation in checks is expected. No changes in code.


rouault:
unrelated msys2-mingw failure will be fixed per https://github.com/OSGeo/gdal/pull/9894

PR #9896, opened 2024-05-10
[gdalmdiminfo] Use GDALArgumentParser

elpaso:
```
gdalmdiminfo --long-usage
Export manually or use this script as a wrapper for commands!!!!
Usage: gdalmdiminfo [--help] [--long-usage] [--help-general]
                        [-detailed] [-nopretty] [-array <array_name>] [-limit <number>] [-oo <NAME>=<VALUE>]...
                        [-if <format>]... [-arrayoption <NAME=VALUE>]... [-stats] [-stdout]
                        <dataset_name>

Lists various information about a GDAL multidimensional dataset.

Positional arguments:
  <dataset_name>             Input dataset. 

Optional arguments:
  -h, --help                 Shows short help message and exits. 
  --long-usage               Shows long help message and exits. 
  --help-general             Report detailed help on general options. 
  -detailed                  Most verbose output. Report attribute data types and array values. 
  -nopretty                  Outputs on a single line without any indentation. 
  -array <array_name>        Name of the array, used to restrict the output to the specified array. 
  -limit <number>            Number of values in each dimension that is used to limit the display of array values. 
  -oo <NAME>=<VALUE>         Open option(s) for input dataset. [may be repeated]
  -if <format>               Format/driver name(s) to try when opening the input file. [may be repeated]
  -arrayoption <NAME=VALUE>  Option passed to GDALGroup::GetMDArrayNames() to filter reported arrays. [may be repeated]
  -stats                     Read and display image statistics. 
  -stdout                    Write output to stdout instead of a JSON file. 

For more details, consult https://gdal.org/programs/gdalmdiminfo.html
```

### DIFF
--- a/apps/gdalmdiminfo_lib.cpp
+++ b/apps/gdalmdiminfo_lib.cpp
@@ -33,6 +33,7 @@
 #include "cpl_json.h"
 #include "cpl_json_streaming_writer.h"
 #include "gdal_priv.h"
+#include "gdalargumentparser.h"
 #include <limits>
 #include <set>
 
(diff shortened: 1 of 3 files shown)


apps/gdalmdiminfo_lib.cpp
```diff
@@ -1060,6 +1061,113 @@ static void WriteToStdout(const char *pszText, void *)
...
+        .store_into(psOptions->nLimitValuesByDim)
+        .help(_("Number of values in each dimension that is used to limit the "
+                "display of array values."));
+
+    if (psOptionsForBinary)
+    {
+        argParser->add_open_options_argument(
+            psOptionsForBinary->aosOpenOptions);
+
+        argParser->add_argument("-if")
```
rouault, line 1105: we have GDALArgumentParser::add_input_format_argument() for that

apps/gdalmdiminfo_lib.cpp
```diff
@@ -1060,6 +1061,113 @@ static void WriteToStdout(const char *pszText, void *)
...
+                _("Format/driver name(s) to try when opening the input file."));
+
+        argParser->add_argument("dataset_name")
+            .metavar("<dataset_name>")
+            .store_into(psOptionsForBinary->osFilename)
+            .help("Input dataset.");
+    }
+
+    argParser->add_argument("-arrayoption")
+        .metavar("<NAME=VALUE>")
```
rouault, line 1129:
```suggestion
        .metavar("<NAME>=<VALUE>")
```

apps/gdalmdiminfo_lib.cpp
```diff
@@ -1060,6 +1061,113 @@ static void WriteToStdout(const char *pszText, void *)
...
+
+    argParser->add_argument("-stats")
+        .flag()
+        .store_into(psOptions->bStats)
+        .help(_("Read and display image statistics."));
+
+    argParser->add_argument("-stdout")
+        .flag()
+        .store_into(psOptions->bStdoutOutput)
+        .help(_("Write output to stdout instead of a JSON file."));
```
rouault, line 1144:
```suggestion
    // Only used by gdalmdiminfo binary to write output to stdout instead of in a string, in JSON mode
    argParser->add_argument("-stdout")
        .flag()
        .hidden()
        .store_into(psOptions->bStdoutOutput);
```

apps/gdalmdiminfo_lib.cpp
elpaso: done

apps/gdalmdiminfo_lib.cpp
elpaso: done

PR #9907, opened 2024-05-11
Update osmconf.ini clarifying direction

jidanni:
In https://gdal.org/drivers/vector/osm.html#configuration we read
> The customization is essentially which OSM attributes and keys should be translated into OGR layer fields.

I therefore conclude the direction of the import is actually into OGR, and thus make it clear in the top of the file.


ogr/ogrsf_frmts/osm/data/osmconf.ini
```diff
@@ -1,5 +1,5 @@
 #
-# Configuration file for OSM import
+# Configuration file for exporting OSM data, importing it into OGR
```
rouault, line 2:
```suggestion
# Configuration file for importing OSM data into OGR
```

PR #9909, opened 2024-05-12
OSR importFromESRI() Arc/Info 7 .prj: fix importing LAMBERT_AZIMUTHAL…

rouault:
… with a 'radius of the sphere of reference'

Fixes https://lists.osgeo.org/pipermail/gdal-dev/2024-May/058990.html

### DIFF
--- a/ogr/ogr_srs_esri.cpp
+++ b/ogr/ogr_srs_esri.cpp
@@ -262,7 +262,8 @@ static CPLString OSR_GDS(char **papszNV, const char *pszField,
  * importFromESRI() by an automatic call to morphFromESRI().
  *
  * Currently only GEOGRAPHIC, UTM, STATEPLANE, GREATBRITIAN_GRID, ALBERS,
- * EQUIDISTANT_CONIC, TRANSVERSE (mercator), POLAR, MERCATOR and POLYCONIC
+ * EQUIDISTANT_CONIC, TRANSVERSE (mercator), POLAR, LAMBERT (Conic Conformal),
+ * LAMBERT_AZIMUTHAL, MERCATOR and POLYCONIC
  * projections are supported from old style files.
  *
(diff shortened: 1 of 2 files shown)


ogr/ogr_srs_esri.cpp
```diff
@@ -589,8 +627,18 @@ OGRErr OGRSpatialReference::importFromESRI(char **papszPrj)
                 }
                 if (!bFoundParameters)
                 {
-                    // If unknown, default to WGS84 so there is something there.
-                    SetWellKnownGeogCS("WGS84");
+                    if (bHasRadiusOfSphereOfReference)
+                    {
+                        OGRSpatialReference oGCS;
+                        oGCS.SetGeogCS("unknown", "unknown", "unknown",
+                                       dfRadiusOfSphereOfReference, 0);
```
jjimenezshaw, line 634: What if the units in the file are not meters, is this radius of the earth in meters? I don't know.

For instance, I have seen files from Trimble where the radius of the ellipsoid is in the linear units of the file, that can be foot

rouault: > What if the units in the file are not meters, is this radius of the earth in meters? I don't know.

importFromESRI() (seems to) assumes that Units is only for the projected linear units, not for the ellipsoid dimension. I'm not aware of a specification for ArcInfo 7.x .prj file, so we can't know for sure until someone exhibits a counter example

> For instance, I have seen files from Trimble where the radius of the ellipsoid is in the linear units of the file

ArcInfo 7.x .prj files ?

jjimenezshaw: > > For instance, I have seen files from Trimble where the radius of the ellipsoid is in the linear units of the file
> 
> ArcInfo 7.x .prj files ?

No. Trimble Site Calibration files (.CAL or .DC)

jjimenezshaw (approved):

Issue #8132, opened 2023-07-21
Feature Request: Improve Buffer


Meibes:
Hello,

according to the [GDAL docs](https://gdal.org/doxygen/classOGRGeometry.html) buffer accepts two inputs: dfDist and nQuadSegs. It also says: 

> This method is built on the GEOS library

According to the [GEOS docs](https://libgeos.org/doxygen/classgeos_1_1operation_1_1buffer_1_1BufferOp.html), GEOS offers additional parameters: nEndCapStyle and isSingleSided

Would it be much work to integrate these additional parameters into GDAL / the python GDAL/OGR bindings?

Thanks

rouault:
Those parameters are available through GEOSBufferWithParams_r(). It should be relatively straightforward to implement

PR #9924, opened 2024-05-14
OGRGeometry: Add BufferEx method

dbaston:
## What does this PR do?

Adds an `OGRGeometry::BufferEx` method to allow passing key/value options.

#8132

## Tasklist

 - [x] Add test case(s)
 - [x] Add documentation
 - [ ] Review
 - [ ] Adjust for comments


dbaston:
I'm not sure of the cleanest way to handle this. I guess I could expose `BufferEx` under a separate name. The tests pass despite the warning.

```
/home/dan/dev/gdal/swig/include/ogr.i:3713: Warning 511: Can't use keyword arguments with overloaded functions (OGRGeometryShadow::Buffer(double,int)).
/home/dan/dev/gdal/swig/include/ogr.i:3717: Warning 511: Can't use keyword arguments with overloaded functions (OGRGeometryShadow::Buffer(double,char **)).
```

ogr/ogrgeometry.cpp
```diff
@@ -4534,6 +4535,261 @@ OGRGeometryH OGR_G_Buffer(OGRGeometryH hTarget, double dfDist, int nQuadSegs)
...
+                CPLError(CE_Failure, CPLE_IllegalArg,
+                         "Invalid value for MITRE_LIMIT: %s", pszValue);
+            }
+        }
+        else if (EQUAL(pszParam, "QUADRANT_SEGMENTS"))
+        {
+            try
+            {
+                std::size_t end;
+                int nQuadSegs = std::stoi(pszValue, &end, 10);
```
dbaston, line 4681: This parameter parsing/validation is really verbose and makes me wonder if I'm missing something like `std::optional<int> CPLReadInt(std::string_view)` or `bool CPLReadInt(const char*, int*)`.

`std::from_chars` is the best solution I've found but I recently ran into problems with it being unsupported on recent versions of Apple clang.

rouault:
> I'm not sure of the cleanest way to handle this. I guess I could expose `BufferEx` under a separate name. The tests pass despite the warning.

yeah, we have a similar issue with Feature::SetField(). One solution might be, for Python, to expose different SWIG method names for the 2 forms of Buffer, and have a Python method in swig/include/python/ogr_python.i that redirects to the appropriate one.

ogr/ogrgeometry.cpp
```diff
@@ -4534,6 +4535,261 @@ OGRGeometryH OGR_G_Buffer(OGRGeometryH hTarget, double dfDist, int nQuadSegs)
...
+ *
+ * @param dfDist the buffer distance to be applied. Should be expressed into
+ *               the same unit as the coordinates of the geometry.
+ * @param papszOptions NULL terminated list of options (may be NULL)
+ *
+ * @return the newly created geometry, or NULL if an error occurs.
+ */
+
+OGRGeometry *OGRGeometry::BufferEx(UNUSED_IF_NO_GEOS double dfDist,
+                                   CSLConstList papszOptions) const
```
rouault, line 4571:
```suggestion
                                   UNUSED_IF_NO_GEOS CSLConstList papszOptions) const
```

ogr/ogrgeometry.cpp
rouault: > `std::from_chars` is the best solution I've found but I recently ran into problems with it being unsupported on recent versions of Apple clang.

well, it is now a requirement since it is used in apps/argparse/argparse.hpp

Otherwise, I wouldn't care much about validating that the provided string is in range and would just blindly use atoi() . We do that almost everywhere....

ogr/ogrgeometry.cpp
```diff
@@ -4534,6 +4535,261 @@ OGRGeometryH OGR_G_Buffer(OGRGeometryH hTarget, double dfDist, int nQuadSegs)
...
+ * </ul>
+ *
+ * This function is the same as the C function OGR_G_BufferEx().
+ *
+ * @param dfDist the buffer distance to be applied. Should be expressed into
+ *               the same unit as the coordinates of the geometry.
+ * @param papszOptions NULL terminated list of options (may be NULL)
+ *
+ * @return the newly created geometry, or NULL if an error occurs.
+ */
```
rouault, line 4568:
```suggestion
 * @since GDAL 3.10
 */
```

ogr/ogrgeometry.cpp
```diff
@@ -4534,6 +4535,261 @@ OGRGeometryH OGR_G_Buffer(OGRGeometryH hTarget, double dfDist, int nQuadSegs)
...
+ *
+ * This function is the same as the C++ method OGRGeometry::BufferEx().
+ *
+ * @param hTarget the geometry.
+ * @param dfDist the buffer distance to be applied. Should be expressed into
+ *               the same unit as the coordinates of the geometry.
+ * @param papszOptions NULL terminated list of options (may be NULL)
+ *
+ * @return the newly created geometry, or NULL if an error occurs.
+ */
```
rouault, line 4781:
```suggestion
 * @since GDAL 3.10
 */
```
```

**Comment by dbaston** [comment]
> One solution might be, for Python, to expose different SWIG method names for the 2 forms of Buffer, and have a Python method in swig/include/python/ogr_python.i that redirects to the appropriate one.

I guess the alternative is to disable kwargs for this function, which breaks the usage
```
ogr.CreateGeometryFromWkt("POINT (0 0)").Buffer(1, quadsecs=1)
```
That doesn't bother me, especially with "quadsecs" being a typo, but I'm fine to go either way.

rouault:
> is to disable kwargs [... ] That doesn't bother me, especially with "quadsecs" being a typo,

That would be fine too

PR #9938, opened 2024-05-16
MiraMonVector: Fixing the way that Z are written in MiraMon layers

AbelPau:
## What does this PR do?
During the translation from a 30 Gigabyte GeoPackage, it was detected that the MiraMonVector driver did not properly detect when the offset for heights exceeded the maximum possible value in a 4-byte variable (in version 1.1). This has been fixed, and an extra check has been added at the end of the translation to ensure that no offset overflows occur.

Once this was corrected, it was also noticed that a key feature for saving space with heights was not being properly used: the MiraMon format allows writing a height only once if it is the same for all vertices of a linestring. According to the specifications, if the number of heights for the linestring is negative, then the height applies to each vertex of the linestring. While this was not a bug, it will represent significant space savings in some 3D files with linestrings that have many vertices. This affected only the writing of linestrings, as the reading was done correctly.

For points, it was detected that the number of Z-values for each point (1) was being written as a positive number when the specifications dictate that it should be negative. This affected only the writing of linestrings, as the reading was done correctly.

Additionally, two minor improvements have been made:

- Informs the user that the layer obtained when the version 1.1 limit is reached is valid and contains N elements (not the total of them). This, at least, allows the user to take a look at the obtained layer.
- Increases some defines to reduce the number of reallocs needed for large layers. The increase does not affect small layers as the increases are only a few megabytes.

A backport to 3.9 would be wonderful.

## Tasklist

 - [x]  Add a test for this fix
 - [x] All CI builds and checks have passed except the specified below
 - [ ] Confirm that Failed Travis is not this PR fault.
 - [ ] Confirm that Failed autotest_gdrivers is not this PR fault.

### DIFF
--- a/autotest/ogr/ogr_miramon_vector.py
+++ b/autotest/ogr/ogr_miramon_vector.py
@@ -1173,7 +1173,16 @@ def test_ogr_miramon_write_basic_linestring(tmp_path):
     ds = None
 
 
-def test_ogr_miramon_write_basic_linestringZ(tmp_path):
+# There are to ways of writing/reading repeated Z's in a linestring file.
+# So let's test both ways (different Z's in each vertice or the same Z for all of them)
+@pytest.mark.parametrize(
+    "LinestringZ",
+    [
+        "LINESTRING Z (0 0 4,0 1 3,1 1 2)",
+        "LINESTRING Z (0 0 4,0 1 4,1 1 4)",
+    ],
+)
+def test_ogr_miramon_write_basic_linestringZ(tmp_path, LinestringZ):
 
     filename = str(tmp_path / "DataSetLINESTRING")
     ds = ogr.GetDriverByName("MiramonVector").CreateDataSource(filename)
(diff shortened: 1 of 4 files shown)


AbelPau:
I open it again because there was a little mistake and it's been solved

rouault:
> the MiraMon format allows writing a height only once if it is the same for all vertices of a linestring. According to the specifications, if the number of heights for the linestring is negative, then the height applies to each vertex of the linestring

it feels like a new test should be added to test such linestring and read it back

AbelPau:
> > the MiraMon format allows writing a height only once if it is the same for all vertices of a linestring. According to the specifications, if the number of heights for the linestring is negative, then the height applies to each vertex of the linestring
> 
> it feels like a new test should be added to test such linestring and read it back

I thought about that but it's been detected in a big file. So, I have to think how to make it happen in a little one. But Yes, you are right. I'll think about it.

rouault:
> I thought about that but it's been detected in a big file

but the write optimization is applied even on small files, right ? Wouldn't just writing "LINESTRING Z (1 2 10, 3 4 10)" would be sufficient ?

AbelPau:
> > I thought about that but it's been detected in a big file
> 
> but the write optimization is applied even on small files, right ? Wouldn't just writing "LINESTRING Z (1 2 10, 3 4 10)" would be sufficient ?

Yes, at least it should work. At this time there is "LINESTRING Z (0 0 4,0 1 3,1 1 2)" and I can add one that 4, 3 and 2 are the same value. I'm going to update that example. Thanks.

AbelPau:
> > I thought about that but it's been detected in a big file
> 
> but the write optimization is applied even on small files, right ? Wouldn't just writing "LINESTRING Z (1 2 10, 3 4 10)" would be sufficient ?

Added and waiting for checks. There are two failed but I think it's something that has nothing to do with this PR. If you can verify that, then it'll be ready if the remaining two checks are ok. Backport to 3.9 would be wonderful :)

PR #9943, opened 2024-05-16
Update ogrinfo.rst removing obsolete "# " not present today

jidanni:
Ogrinfo doesn't add the "# " anymore. Today we sync the man page with the current output.


jratike80:
I wonder if "#" has been used as a workaround for some text rendering issue. Anyway, they can go.

PR #9956, opened 2024-05-16
CPLHTTPFetch(): add a RETRY_CODES option, and C++'ify HTTP retry logic

### LINKED ISSUE #9941: GDAL_HTTP_RETRY_ALL Configuration option
### Feature description

It would be nice to be able to set the HTTP retry behaviour of GDAL to happen on every HTTP error.

For example, a (probably wrongly configured) WFS server is sometimes returning an error 400 on a random valid request.
Then GDAL doesn't retry the request (which would make sense if the error code was correct) causing failure of the download.

This would allow more flexibility I think, and help to deal with wrongly configured WFS servers.

### Additional context

_No response_

rouault:
*  CPLHTTPFetch(): add a RETRY_CODES option (GDAL_HTTP_RETRY_CODES config option)
    
    Fixes #9941
    
    ```
    -  .. config:: GDAL_HTTP_RETRY_CODES
          :choices: ALL or comma-separated list of codes
          :since: 3.10
    
          Specify which HTTP error codes should trigger a retry attempt.
          Valid values are ``ALL`` or a comma-separated list of HTTP codes.
          By default, 429, 500, 502, 503 or 504 HTTP errors are considered, as
          well as other situations with a particular HTTP or Curl error message.
    ```

*  Add C++ facilities for HTTP retry logic

### DIFF
--- a/port/cpl_http.cpp
+++ b/port/cpl_http.cpp
@@ -529,6 +529,7 @@ constexpr TupleEnvVarOptionName asAssocEnvVarOptionName[] = {
     {"GDAL_HTTP_NETRC_FILE", "NETRC_FILE"},
     {"GDAL_HTTP_MAX_RETRY", "MAX_RETRY"},
     {"GDAL_HTTP_RETRY_DELAY", "RETRY_DELAY"},
+    {"GDAL_HTTP_RETRY_CODES", "RETRY_CODES"},
     {"GDAL_CURL_CA_BUNDLE", "CAINFO"},
     {"CURL_CA_BUNDLE", "CAINFO"},
     {"SSL_CERT_FILE", "CAINFO"},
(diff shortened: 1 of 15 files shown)


port/cpl_http.h
```diff
@@ -179,8 +179,47 @@ CPL_C_END
...
+class CPLHTTPRetryContext
+{
+  public:
+    explicit CPLHTTPRetryContext(const CPLHTTPRetryParameters &oParams);
+
+    bool CanRetry(int response_code, const char *pszErrBuf,
+                  const char *pszCurlError);
+    bool CanRetry();
+
+    /** Returns the delay to apply. Only valid after a successful call to CanRetry() */
```
elpaso, line 204: Reading this note makes me wonder if it is worth adding a debug message to prevent misuse (calling this without calling CanRetry).

port/cpl_http.cpp
```diff
@@ -553,41 +554,65 @@ constexpr TupleEnvVarOptionName asAssocEnvVarOptionName[] = {
...
-                          strstr(pszCurlError, "Operation timed out") ||
-                          strstr(pszCurlError, "Connection reset by peer") ||
-                          strstr(pszCurlError, "Connection was reset"))))
+    bool bRetry = false;
+    if (pszRetriableCodes && pszRetriableCodes[0])
+    {
+        bRetry = EQUAL(pszRetriableCodes, "ALL") ||
+                 strstr(pszRetriableCodes, CPLSPrintf("%d", response_code));
+    }
+    else if (response_code == 429 || response_code == 500 ||
```
elpaso, line 602: Probably out of scope for this PR but a nice readability improvement would be to replace these literal error codes with constants from a header (I suppose CURL has one, or something like https://github.com/j-ulrich/http-status-codes-cpp/tree/main).

elpaso (approved):

port/cpl_http.h
rouault: done (testing m_nRetryCount == 0)

PR #9958, opened 2024-05-17
Update sql_sqlite_dialect.rst to include single quote example

jidanni:
Also folded monster long line.

(Single quotes have actually been part of the shell for 50 years.)

### DIFF
--- a/doc/source/user/sql_sqlite_dialect.rst
+++ b/doc/source/user/sql_sqlite_dialect.rst
@@ -68,7 +68,8 @@ be used in joins are searched from the master database.
 
 .. code-block:: shell
 
-    ogrinfo jointest.gpkg -dialect INDIRECT_SQLITE -sql "SELECT a.ID,b.ID FROM jointest a JOIN \"jointest2.shp\".\"jointest2\" b ON a.ID=b.ID"
+    ogrinfo jointest.gpkg -dialect INDIRECT_SQLITE -sql \
+    "SELECT a.ID,b.ID FROM jointest a JOIN \"jointest2.shp\".\"jointest2\" b ON a.ID=b.ID"
 
 The column names that can be used in the result column list, in WHERE, JOIN, ... clauses
 are the field names of the layers. Expressions, SQLite functions, spatial functions, etc...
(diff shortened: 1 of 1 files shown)


doc/source/user/sql_sqlite_dialect.rst
```diff
@@ -115,6 +116,12 @@ between double quotes, the internal double quotes must be escaped with \\
 
     ogrinfo p.shp -sql "SELECT geometry \"FROM\", \"3D\" FROM p"
 
+If single quotes are available in the shell, one can also use just
+
```
jratike80, line 120: Single quotes are available in the Windows shell but they can be used only on Linux (and Mac?). I am not sure if it is good to recommend using a syntax that does not work on all platforms.

There needs to be a comma after "geometry".

jidanni:
Well please mention that there, else, for us who have never used the other systems you mention, it looks like the author is just making things 50 times harder for themselves, because a vital page was creased in their K&R Book.

jratike80:
Rather write it into a supporting document "Good to know about Linux vs Windows" because the use of single vs double quotes is something that does not only affect the usage of SQLite SQL dialect.

jidanni:
I'm sorry I can't make sense of the diffs here on GitHub. All I know is the current version on the website seems like grandparents paying for their new house with a stack of one dollar bills. So needs about three words added acknowledging that simpler solutions are available on some systems. Thanks. No need to even mention what they are though.

jratike80:
Sorry, I don't understand your last comment. 

 Because you aim at clarity, then how about changing "If single quotes are available in the shell" into  "Linux and Mac users can use single outer quotes but that syntax does not work on Windows"

doc/source/user/sql_sqlite_dialect.rst
```diff
@@ -115,6 +116,12 @@ between double quotes, the internal double quotes must be escaped with \\
...
+.. code-block:: shell
+
+    ogrinfo p.shp -sql 'SELECT geometry "FROM", "3D" FROM p'
+
```
rouault, line 124: Dropping that change
```suggestion
```

rouault:
> Rather write it into a supporting document "Good to know about Linux vs Windows" because the use of single vs double quotes is something that does not only affect the usage of SQLite SQL dialect.

yes, that would be more appropriate in a more generic place, but not sure where it is belong to. Where do we stop in our explanations of what is not GDAL specific... ?

PR #9961, opened 2024-05-17
GDALDataset::RasterIO() / GDALDatasetRasterIO[Ex](): accept a const int* panBandList, instead of a int*, and related changes

rouault:
Currently GDALDataset::RasterIO() / GDALDatasetRasterIO[Ex]() take a "int* panBandList", whereas its semantics is really "const int*".
To do the changes fully one should also change the virtual method GDALDataset::IRasterIO() in a similar way, but that would affect out-of-tree drivers, so I've used a compromise where I changed in-tree code to use a BANDMAP_TYPE define that evaluates to "int *" for backward compatibility, but with an easy way to check it is const ready, with the below

 ```
// This macro can be defined to check that GDALDataset::IRasterIO()
// implementations do not alter the passed panBandList. It is not defined
// by default (and should not!), hence int* is used.
\#if defined(GDAL_BANDMAP_TYPE_CONST_SAFE)
\#define BANDMAP_TYPE const int *
\#else
\#define BANDMAP_TYPE int *
\#endif
```

I'm not totally sure this a good idea/approach

### DIFF
--- a/.github/workflows/ubuntu_24.04/build.sh
+++ b/.github/workflows/ubuntu_24.04/build.sh
@@ -5,10 +5,14 @@ set -e
 LD_LIBRARY_PATH="/opt/instantclient_19_9:/opt/instantclient_19_9/lib:${LD_LIBRARY_PATH}"
 export LD_LIBRARY_PATH
 
+# Use by swig/python/setup.py.in
+GDAL_PYTHON_CXXFLAGS="-DGDAL_BANDMAP_TYPE_CONST_SAFE"
+export GDAL_PYTHON_CXXFLAGS
+
 cmake ${GDAL_SOURCE_DIR:=..} \
     -DCMAKE_BUILD_TYPE=Release \
     -DCMAKE_C_FLAGS="-Werror" \
-    -DCMAKE_CXX_FLAGS="-Werror" \
+    -DCMAKE_CXX_FLAGS="-Werror -DGDAL_BANDMAP_TYPE_CONST_SAFE" \
     -DUSE_CCACHE=ON \
     -DOracle_ROOT=/opt/instantclient_19_9 \
     -DGDAL_USE_GEOTIFF_INTERNAL:BOOL=ON \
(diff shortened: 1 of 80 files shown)

### CODE REVIEW

**Comment by elpaso** [comment]
> but that would affect out-of-tree drivers

do you mean this would be an API break for third party drivers?

> I'm not totally sure this a good idea/approach

I can't think of a better one.

File: .github/workflows/ubuntu_24.04/build.sh [hunk 1]
```diff
@@ -5,10 +5,13 @@ set -e
 LD_LIBRARY_PATH="/opt/instantclient_19_9:/opt/instantclient_19_9/lib:${LD_LIBRARY_PATH}"
 export LD_LIBRARY_PATH
 
+GDAL_PYTHON_CXXFLAGS="-DBANDMAP_TYPE_CONST_SAFE"
```
elpaso, line 8: Perhaps just rename it to `GDAL_BANDMAP_CONST_SAFE` to make it immediately clear it's a GDAL flag/macro

alg/gdalpansharpen.cpp
```diff
@@ -1642,10 +1642,17 @@ void GDALPansharpenOperation::PansharpenResampleJobThreadFunc(void *pUserData)
...
+    for (int i = 0; i < psJob->nBandCount; ++i)
+        anBands.push_back(i + 1);
+    // This call to RasterIO() in a thread to poMEMDS shared between several
+    // threads is really risky, but works given the implementation details...
+    // Do not do this at home!
     CPL_IGNORE_RET_VAL(psJob->poMEMDS->RasterIO(
         GF_Read, psJob->nXOff, psJob->nYOff, psJob->nXSize, psJob->nYSize,
         psJob->pBuffer, psJob->nBufXSize, psJob->nBufYSize, psJob->eDT,
-        psJob->nBandCount, nullptr, 0, 0, psJob->nBandSpace, &sExtraArg));
+        psJob->nBandCount, anBands.data(), 0, 0, psJob->nBandSpace,
```
elpaso, line 1654: how it this (and previous) change related to the const correctness?

rouault: The issue is that this code abuses mutlithreading. It does multithreaded read on the same dataset. Before when passing panBandMap==nullptr a temporary array was allocated by GDALDataset::RasterIO(). I changed the later to rather use a member variable to hold the temporary array. And thus there was suddently a concurrent modification of that temporary array causing chaos.

.github/workflows/ubuntu_24.04/build.sh
rouault: done

PR #9962, opened 2024-05-17
Update sql_sqlite_dialect.rst to clarify Windows example

jidanni:
For #9958.

Linux users have a better way to deal with this, so we need to gently hint that this example is from Windows, lest Linux users think we are uneducated.

### DIFF
--- a/doc/source/user/sql_sqlite_dialect.rst
+++ b/doc/source/user/sql_sqlite_dialect.rst
@@ -110,7 +110,8 @@ they are not strictly needed.
     SELECT "p"."geometry", "p"."FROM", "p"."3D" FROM "poly" p
 
 When SQL statements are used in the command shell and the statement itself is put
-between double quotes, the internal double quotes must be escaped with \\
+between double quotes, the internal double quotes must be escaped with \\.
+Windows example:
 
 .. code-block:: shell
 


doc/source/user/sql_sqlite_dialect.rst
```diff
@@ -110,7 +110,8 @@ they are not strictly needed.
     SELECT "p"."geometry", "p"."FROM", "p"."3D" FROM "poly" p
 
 When SQL statements are used in the command shell and the statement itself is put
-between double quotes, the internal double quotes must be escaped with \\
+between double quotes, the internal double quotes must be escaped with \\.
+Windows example:
```
rouault, line 114: That' not true. The also example also work on Unix shells

jidanni: Fine. If nobody understands what I'm trying to say, then I'll just give up.

rouault: > If nobody understands what I'm trying to say

See my comment at https://github.com/OSGeo/gdal/pull/9958#issuecomment-2117472539 . We do understand but it is not obvious where to explain that detail. The documentation also needs to remain compact and straight-to-the-point it demonstrates. We can't recall in every place where we have SQL in the doc that depending on the shell used, different quoting rules may apply. By sticking to using double-quotes, we're at least compatible of (hopefully) all Unix and Windows shells.

PR #9970, opened 2024-05-21
Update ogrinfo.rst to mention vital prefix method

jidanni:
Without mentioning this, we see that sometimes
its just hit and miss what driver will be selected.
```
$ ogrinfo --formats|
 perl -awnle 'next unless /json/i; print $F[0];'|
 xargs -n 1 ogrinfo --format|egrep Extension\|Long\ N
  Long Name: GeoJSON
  Extensions: json geojson
  Long Name: GeoJSON Sequence
  Extensions: geojsonl geojsons
  Long Name: ESRIJSON
  Extension: json
  Long Name: TopoJSON
  Extensions: json topojson
 ```
Yes, individual driver pages do mention what I am mentioning here on ogrinfo, but only here do I tie them all together.

### DIFF
--- a/doc/source/programs/ogrinfo.rst
+++ b/doc/source/programs/ogrinfo.rst
@@ -226,6 +226,7 @@ edit data.
 
     The data source to open. May be a filename, directory or other virtual
     name. See the OGR Vector Formats list for supported datasources.
+    A prefix may be used to force choosing a particular driver. See example below.
 
 .. option:: <layer>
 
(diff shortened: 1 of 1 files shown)


jratike80:
Did you see https://gdal.org/programs/ogrinfo.html#cmdoption-ogrinfo-if

> Format/driver name to be attempted to open the input file(s). It is generally not necessary to specify it, but it can be used to skip automatic driver detection, when it fails to select the appropriate driver. This option can be repeated several times to specify several candidate drivers. Note that it does not force those drivers to open the dataset. In particular, some drivers have requirements on file extensions.

jidanni:
Regarding `-if`. Yes I saw it. But I recall that is is a "slightly different animal". However I dared not add to my wording exactly how it differs. But please fix my PR to elucidate the relation of the prefix stuff with the `-if` stuff. I'm sure the authors wouldn't have bothered to create two methods if in fact they do exactly the same thing! Thanks.

jratike80:
Authors have created exactly one method for ogrinfo and it is `-if`. The prefix thing is older and it exists in some drivers, for example https://gdal.org/drivers/vector/csv.html

> For files structured as CSV, but not ending with the ".csv" extension, the 'CSV:' prefix can be added before the filename to force loading by the CSV driver.

Because `-if` exists now I do not see why `PREFIX:` should be documented for ogrinfo.

rouault:
Closing this pull request as resolving https://github.com/OSGeo/gdal/issues/9978 will be a better resolution

PR #9950, opened 2024-05-16
GeoJSON/JSONFG: do not recognize GeoJSON files with a featureType feature property as JSONFG     

rouault:
Fixes https://github.com/OSGeo/gdal/issues/9946

PR #9982, opened 2024-05-22
[Backport 3.9] GeoJSON/JSONFG: do not recognize GeoJSON files with a featureType feature property as JSONFG

rouault:
Backport of #9950


rouault:
redone correctly in https://github.com/OSGeo/gdal/pull/9983

PR #9991, opened 2024-05-23
gdal_viewshed: use functions in implementation

abellgithub:
This changes the implementation to use functions and uses threads for the separable bits. It also merges the two result vectors into one.

### DIFF
--- a/alg/viewshed.cpp
+++ b/alg/viewshed.cpp
@@ -28,7 +28,7 @@
 #include <algorithm>
 #include <array>
 #include <atomic>
-#include <thread>
+#include <future>
 
 #include "gdal_alg.h"
 #include "gdal_priv_templates.hpp"
@@ -676,21 +676,19 @@ bool Viewshed::processFirstLine(int nX, int nY, int nLine,
     const auto [iLeft, iRight] =
         adjustHeight(nYOffset, nX, vThisLineVal.data() + nX);
 
-    std::thread t1(
-        [&, left = iLeft]()
+    auto t1 = std::async(std::launch::async, [&, left = iLeft]()
         {
             processLineLeft(nX, nYOffset, nX - 2, left - 1, vResult,
                             vThisLineVal, vLastLineVal);
         });
 
-    std::thread t2(
-        [&, right = iRight]()
+    auto t2 = std::async(std::launch::async, [&, right = iRight]()
         {
             processLineRight(nX, nYOffset, nX + 2, right, vResult, vThisLineVal,
                              vLastLineVal);
         });
-    t1.join();
-    t2.join();
+    t1.wait();
+    t2.wait();
 
     // Make the current line the last line.
     vLastLineVal = std::move(vThisLineVal);
@@ -741,21 +739,19 @@ bool Viewshed::processLine(int nX, int nY, int nLine,
         vResult[nX] = oOpts.outOfRangeVal;
 
     // process left half then right half of line
-    std::thread t1(
-        [&, left = iLeft]()
+    auto t1 = std::async(std::launch::async, [&, left = iLeft]()
         {
             processLineLeft(nX, nYOffset, nX - 1, left - 1, vResult,
                             vThisLineVal, vLastLineVal);
         });
 
-    std::thread t2(
-        [&, right = iRight]()
+    auto t2 = std::async(std::launch::async, [&, right = iRight]()
         {
             processLineRight(nX, nYOffset, nX + 1, right, vResult, vThisLineVal,
                              vLastLineVal);
         });
-    t1.join();
-    t2.join();
+    t1.wait();
+    t2.wait();
 
     // Make the current line the last line.
     vLastLineVal = std::move(vThisLineVal);
@@ -836,8 +832,7 @@ bool Viewshed::run(GDALRasterBandH band, GDALProgressFunc pfnProgress,
 
     // scan upwards
     std::atomic<bool> err(false);
-    std::thread tUp(
-        [&]()
+    auto tUp = std::async(std::launch::async, [&]()
         {
             std::vector<double> vLastLineVal = vFirstLineVal;
 
@@ -848,8 +843,7 @@ bool Viewshed::run(GDALRasterBandH band, GDALProgressFunc pfnProgress,
         });
 
     // scan downwards
-    std::thread tDown(
-        [&]()
+    auto tDown = std::async(std::launch::async, [&]()
         {
             std::vector<double> vLastLineVal = vFirstLineVal;
 
@@ -858,8 +852,8 @@ bool Viewshed::run(GDALRasterBandH band, GDALProgressFunc pfnProgress,
                     err = true;
         });
 
-    tUp.join();
-    tDown.join();
+    tUp.wait();
+    tDown.wait();
 
     if (!emitProgress(1))
         return false;
--- a/alg/viewshed.cpp
+++ b/alg/viewshed.cpp
@@ -28,6 +28,9 @@
 #include <algorithm>
 #include <array>
 #include <limits>
+//ABELL
+#include <iostream>
+#pragma GCC diagnostic ignored "-Wold-style-cast"
 
 #include "gdal_alg.h"
 #include "gdal_priv.h"
@@ -237,25 +240,27 @@ double CalcHeightAdjFactor(const GDALDataset *poDataset, double dfCurveCoeff)
     return 0;
 }
 
-double CalcHeightLine(int i, double Za, double Zo)
+double CalcHeightLine(int i, double Za, [[maybe_unused]] double Zo)
 {
     if (i == 1)
         return Za;
     else
-        return (Za - Zo) / (i - 1) + Za;
+        return Za * i / (i - 1);
 }
 
-double CalcHeightDiagonal(int i, int j, double Za, double Zb, double Zo)
+double CalcHeightDiagonal(int i, int j, double Za, double Zb,
+                          [[maybe_unused]] double Zo)
 {
-    return ((Za - Zo) * i + (Zb - Zo) * j) / (i + j - 1) + Zo;
+    return (Za * i + Zb * j) / (i + j - 1);
 }
 
-double CalcHeightEdge(int i, int j, double Za, double Zb, double Zo)
+double CalcHeightEdge(int i, int j, double Za, double Zb,
+                      [[maybe_unused]] double Zo)
 {
     if (i == j)
         return CalcHeightLine(i, Za, Zo);
     else
-        return ((Za - Zo) * i + (Zb - Zo) * (j - i)) / (j - 1) + Zo;
+        return (Za * i + Zb * (j - i)) / (j - 1);
 }
 
 }  // unnamed namespace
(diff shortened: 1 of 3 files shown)


abellgithub:
I ran clang-format locally, so I'm not sure what the issue is with the CI failure. Maybe different clang-format versions?

rouault:
> Maybe different clang-format versions?

Much likely. I'd strongly suggest using pre-commit to get the right version and automating reformatting. Cf https://gdal.org/development/dev_practices.html#commit-hooks

As this is about optimization, I would also be interested in seeing timings before and after the change (on a CMAKE_BUILD_TYPE=Release or RelWithDebInfo build)

alg/viewshed.cpp
```diff
@@ -426,499 +480,389 @@ bool Viewshed::run(GDALRasterBandH hBand, GDALProgressFunc pfnProgress,
...
+        dfResult += (dfZ - dfCellVal);
+        dfResult = std::max(0.0, dfResult);
     }
+    else
+        dfResult = (dfCellVal + oOpts.targetHeight < dfZ) ? oOpts.invisibleVal
+                                                          : oOpts.visibleVal;
+    dfCellVal = std::max(dfCellVal, dfZ);
+}
+
+/// Process the first line (the one with the X coordinate the same as the observer).
```
rouault, line 641:
```suggestion
/// Process the first line (the one with the Y coordinate the same as the observer).
```

alg/viewshed.cpp
```diff
@@ -426,499 +480,389 @@ bool Viewshed::run(GDALRasterBandH hBand, GDALProgressFunc pfnProgress,
...
-            if (oOpts.outputMode != OutputMode::Normal)
-                dfHeightResult[iPixel] = std::max(
-                    0.0, (dfZ - padfFirstLineVal[iPixel] + dfGroundLevel));
+    const auto [iLeft, iRight] =
+        adjustHeight(nYOffset, nX, vThisLineVal.data() + nX);
 
-            setVisibility(iPixel, dfZ, padfFirstLineVal, vResult);
-        }
-        else
+    std::thread t1(
```
rouault, line 679: Is it worth the trouble threading the processing of a single line ... ?

alg/viewshed.cpp
```diff
@@ -426,499 +480,389 @@ bool Viewshed::run(GDALRasterBandH hBand, GDALProgressFunc pfnProgress,
...
+/// @param nLine  Line number being processed.
+/// @param vLastLineVal  Vector in which to store the read line. Becomes the last line
+///    in further processing.
+/// @return True on success, false otherwise.
+bool Viewshed::processLine(int nX, int nY, int nLine,
+                           std::vector<double> &vLastLineVal)
+{
+    int nYOffset = nLine - nY;
+    std::vector<double> vResult(oOutExtent.xSize());
+    std::vector<double> vThisLineVal(oOutExtent.xSize());
```
rouault, line 720: isn't there a runtime cost in allocating those temporary buffers every time this method is invoked ?

alg/viewshed.cpp
```diff
@@ -426,499 +480,389 @@ bool Viewshed::run(GDALRasterBandH hBand, GDALProgressFunc pfnProgress,
...
-    }
+        vResult[nX] = oOpts.outOfRangeVal;
 
-    /* scan upwards */
-    std::copy(vFirstLineVal.begin(), vFirstLineVal.end(), vLastLineVal.begin());
-    for (int iLine = nY - 1; iLine >= nYStart; iLine--)
-    {
-        if (GDALRasterIO(hBand, GF_Read, nXStart, iLine, nXSize, 1,
-                         padfThisLineVal, nXSize, 1, GDT_Float64, 0, 0))
+    // process left half then right half of line
```
rouault, line 743: I would be worried that spawning 2 threads, presumably very short lived unless you have gigantic rasters (I'm not even sure that raster width == INT_MAX would be big enough), repeatedly for each line might be more costly than the computation benefit they bring.

abellgithub:
Before I threaded things, it was a wee bit slower than the original. Most of that was the merging of the output vector, which required the RasterIO code to potentially migrate the data vector to a different type. This could be changed, but I'm not sure it's worth it in code complexity.  I ran stuff locally on a large raster and the old stuff came out to 20+ seconds. The new stuff took about 15 seconds.

Why would running clang-format as a pre-commit hook change its behaivor?

alg/viewshed.cpp
abellgithub: If they're small rasters, why would anyone much care?

dbaston:
> Why would running clang-format as a pre-commit hook change its behaivor?

The hook specifies the version of clang-format:

https://github.com/OSGeo/gdal/blob/12dab86ca2d8b1a65c4c085e137c62294682ac1d/.pre-commit-config.yaml#L30-L31

abellgithub:
@dbaston : But how is that working locally if I have some other version?

rouault:
> But how is that working locally if I have some other version?

just fine. pre-commit utilities are cached in a dedicated directory. You may have different git repositories using pre-commit with different clang or whatever versions.

alg/viewshed.cpp
abellgithub: I'm sure there is. But it's small. Probably very small. malloc is very good and you're re-allocating vectors of the same size that you just recently destroyed. Also, from a code standpoint, it means that you don't have to worry about making sure that threads maintain local copies. I ran things for time after each change, and this made no noticeable difference in my test, but there are lots of cases and I won't swear that it won't make a difference for someone else's test case.

abellgithub:
> pre-commit utilities are cached in a dedicated directory. 

I don't understand but this is perhaps not the place for me to figure it out. I wasn't interested in having clang-format run on my code every time I did a commit, so I just ran it by hand when I was done.

rouault:
> I don't understand but this is perhaps not the place for me to figure it out.

cf https://stackoverflow.com/questions/62539288/where-does-pre-commit-install-environments

Investing a few seconds setting up pre-commit as documented in https://gdal.org/development/dev_practices.html#commit-hooks will end up paying off. This is literally 2 lines (+ 1 to fix the current issues), and you won't have to worry about that anymore.

alg/viewshed.cpp
abellgithub: It was when I tested it. It's trivial to remove the threading and try it.

rouault:
I've given the PR a try with a rather dummy all-zero dataset generated with:
```
gdal_create -outsize 40000 40000 -ot uint16 -a_srs EPSG:32631 -a_ullr 400000 5000000 440000 4960000 -co tiled=yes -co compress=zstd in.tif
```

On Linux x86_64 Ubuntu 20.04 Kernel 5.15, Intel(R) Core(TM) i7-10750H CPU @ 2.60GHz, 6-cores hyper-threaded (12 vCPUs), with 3.9, RelWithDebInfo,  ``gdal_viewshed -ox 420000 -oy 4980000 -oz 100 -co tiled=yes -co compress=zstd in.tif out.tif`` runs in 14.2 seconds. 
With this PR, its runs in 10.2 seconds wall-clock time / 21.4 seconds aggregated CPU time. Fine

Adding "taskset -c 11" to limit running on a single CPU, I get the same 14.2s timing with 3.9 (as expected), but I get 20.1 seconds wall-clock time with this PR, which isn't surprising given that the operating system has more work dealing with threads that can't be executed simultaneously

But more worrying, on my Windows VM (Windows 10 under Virtualbox), while the 3.9 version in Release mode runs in 19.5 seconds, with this PR is runs in 58.5 seconds ! (the VM has access to all CPUs). There may be partly an artifact on not running on bare metal, but I guess this isn't the only reason and it shows that Windows struggles with the thousands of short lived threads.

I believe this PR must be enhanced in 2 directions:
- if CPLGetNumCPUs() returns <= 1, then execute the lambdas run by std::thread() in a serial way and not spawn threads. People using GDAL on limited VMs, or running many parallel GDAL jobs on  the same VM, is a use case we should try not to degrade.
- and certainly more work, but we may need to use a thread pooling mechanism to have decent performance on Windows. Or, just fallback to the above strategy on Windows to not degrade current performance.

hobu:
>  but we may need to use a thread pooling mechanism to have decent performance on Windows

Is there a standard GDAL thread pool facility to use?

rouault:
> Is there a standard GDAL thread pool facility to use?

yes, we already discussed it with Andrew a bit: https://gdal.org/doxygen/classCPLWorkerThreadPool.html . It has probably not all the bells and whistles of more advanced multithreading frameworks, but that's what we use in a number of places already in the code base (warp, geotiff driver, etc.). We also have a singleton instance of a global thread pool with https://github.com/OSGeo/gdal/blob/master/gcore/gdal_thread_pool.h

abellgithub:
> But more worrying, on my Windows VM (Windows 10 under Virtualbox), while the 3.9 version in Release mode runs in 19.5 seconds, with this PR is runs in 58.5 seconds ! (the VM has access to all CPUs). There may be partly an artifact on not running on bare metal, but I guess this isn't the only reason and it shows that Windows struggles with the thousands of short lived threads.

I built and tested on my Windows laptop. It is 6 core i9 running Windows 10. I changed the threads to std::async because I read that it may use the system thread pool on Windows. It didn't really change much. With std::async/std::thread it runs about one second slower (15 vs. 14 secs) than without those threads for my test. On the other hand, on my OSX box things run faster with threads for left and right (about a second and a half faster with the threads).

So I'm not seeing the large performance differences you're reporting above. If you would want to try moving the calls outside of the threading and try again on your machine, maybe we'd learn more. I can push a branch if you'd like.

rouault:
> I can push a branch if you'd like.

yes I can give a try

rouault:
> Here is a branch that removes the threads for each line: https://github.com/abellgithub/gdal/tree/lessthread

much better! runs in ~ 15.8 s in the Windows VM . Actually the key change is the use of std::async.

rouault:
> unless you changed something in the branch above

I didn't.
Basically in my Windows VM:
- https://github.com/abellgithub/gdal/commit/6e1a4ef06a68a2e8a0b87c310355399dd40b66e3 (before async) : runs in ~ 58 s
- https://github.com/abellgithub/gdal/commit/05f1967d108c2bb64a3d0331c957b3a065516241 (async) : runs in ~ 15.1 s
- https://github.com/abellgithub/gdal/commit/dd9bfe8c03dbd4afd823bd8c8c2afffc101062ca (remove threading left/right) : runs in ~ 15.8 s  (I'm not totally sure about the small timing difference between that one and the previous one)

> but I'm happy with whatever you recommend at this point.

it seems that std::async() is perhaps a nicer alternative to using CPLWorkerThreadPool. The only potential downside is that the C++ standard doesn't seem to guarantee the thread pool behavior but it looks like at least MSVC implements it, and I assume you've tested with gcc / clang ? But given your last commit indeed removing threading left/right, std::async vs std::thread shouldn't make any difference. 

One small downside of threading is that writes in the destination file may occur in any order and thus the output file (when it is TIFF) is not binary identical between runs (the pixel content should be identical, but tiles/strips are written in a random order).
Related to that, it would still be nice to test GDAL_NUM_THREADS and CPLGetNumCPUs() as I suggested in https://github.com/OSGeo/gdal/pull/9991#issuecomment-2128154686 to really implement sequential execution when only 1 thread is available. That way we'd get a repeatable output file when that would be needed.

abellgithub:
Perhaps they changed something in the Windows OS to make start of threads much faster and I'm seeing the result of that, as all of these changes made little difference in my testing, regardless of system. :shrug:

I'm not sure why it would matter if the output is binary equivalent as long as the data is the same. I would think some formats would have differences if for no other reason than they contain some sort of timestamp. My personal bias is to keep the code as clean as possible unless there's some strong reason to do otherwise, but I'll defer to you. Let me know.

rouault:
> but I'll defer to you

ok, let's drop that complication (special behaviour for one thread) for now and see if someone complains. I believe however we need the std::async changes you did to minimize the risk of bad performance on (some) Windows configs

abellgithub:
I don't understand the CI failure. Perhaps just a system timeout?

Also, I added the pre-commit hooks but I seem to have to run them by hand. Is there something other than what's shown [here](https://gdal.org/development/dev_practices.html) that I need to do? I used conda instead of pip, but I wouldn't think that would matter. Thx.

PR #10013, opened 2024-05-27
Update gdal_contour.rst to add my pure guess as to what is going on

jidanni:
Just my wild guess as to what is going on!


jratike80:
By the documentation of the algorithm https://gdal.org/api/gdal_alg.html#_CPPv419GDALContourGenerate15GDALRasterBandHddiPdidPvii16GDALProgressFuncPv I think that your guess is wrong.

PR #10016, opened 2024-05-27
Update gdal_grid.rst to document "-zfield field_n"

jidanni:


jratike80:
I believe that the field_n thing is only true with CSV data https://gdal.org/drivers/vector/csv.html

> By default, the driver attempts to treat the first line of the file as a list of field names for all the fields. However, if one or more of the names is all numeric it is assumed that the first line is actually data values and dummy field names are generated internally (field_1 through field_n) and the first record is treated as a feature. 

Please check with some other point data source. If field_n works only with CSV I would withdraw this edit.

jidanni:
Well somebody please change my change to mention that "If you have CSV files, then..."

All I have are headerless CSV files.

All I know is I'm saving the next person a whole day of asking on StackExchange because they didn't think of checking the CSV docs.

I suppose CSV is the only format that can lack field names.

jratike80:
> I suppose CSV is the only format that can lack field names.

I hope so.

rouault:
> If field_n works only with CSV I would withdraw this edit.

yes, this is a CSV specific.

Closing this PR. 

People need to thing logically and we can't possibly give all situations that can arise with the tons of drivers that exist. When a utility asks for a field name, and you don't know it, you use ogrinfo to retrieve it, and if it's a header less CSV you'll see "field_1", "field_2", etc.

PR #10017, opened 2024-05-27
Update ogrmerge.rst making ready for primetime

jidanni:
I tested it and it works great.
```
$ ogrmerge --version
GDAL 3.9.0, released 2024/05/07
$ man ogrmerge | wc
    153     765    7272
$ man ogrmerge.py | wc
No manual entry for ogrmerge.py
      0       0       0
```

### DIFF
--- a/doc/source/programs/ogrmerge.rst
+++ b/doc/source/programs/ogrmerge.rst
@@ -1,7 +1,7 @@
 .. _ogrmerge:
 
 ================================================================================
-ogrmerge.py
+ogrmerge
 ================================================================================
 
 .. only:: html
@@ -15,7 +15,7 @@ Synopsis
 
 .. code-block::
 
-    ogrmerge.py [--help] [--help-general]
+    ogrmerge [--help] [--help-general]
                 -o <out_dsname> <src_dsname> [<src_dsname>]...
                 [-f format] [-single] [-nln <layer_name_template>]
                 [-update | -overwrite_ds] [-append | -overwrite_layer]
@@ -37,8 +37,8 @@ Description
 
 .. versionadded:: 2.2
 
-:program:`ogrmerge.py` script takes as input several vector datasets,
-each of them having one or several vector layers, and copy them in
+:program:`ogrmerge` takes as input several vector datasets,
+each of them having one or several vector layers, and copies them into
 a target dataset.
 
 There are essentially two modes:
@@ -56,7 +56,7 @@ output format is not VRT, final translation is done with :program:`ogr2ogr`
 or :py:func:`gdal.VectorTranslate`. So, for advanced uses, output to VRT,
 potential manual editing of it and :program:`ogr2ogr` can be done.
 
-.. program:: ogrmerge.py
+.. program:: ogrmerge
 
 .. include:: options/help_and_help_general.rst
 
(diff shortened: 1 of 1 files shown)


jratike80:
What certainly works for running Python scripts is syntax `python path_to_script`.

```
python  C:\OSGeo4W\apps\Python312\Scripts\ogrmerge.py
Usage: ogrmerge.py [--help] [--help-general]
...
```

There may be helper systems which make is possible to launch the script with easier syntax  `ogrmerge` but that is not guaranteed to work everywhere.

jidanni:
All I know is one doesn't want to hardwire "fortran", etc, into filenames that might get rewritten in a different language in the future.

jratike80:
But they are Python scripts and not compiled programs like gdalinfo.  Users must have a compatible Python version on their own computer for using those utilities. Internally Python must be started first before the script can be run. It is important to know that distinction.

If the script is rewritten in C++ some day the .py will naturally disappear from the filename.

jidanni:
Certainly such users not playing with a full deck will learn very quickly via error messages telling them to mend their ways.

Else with ".sh", ".this", ".that" extensions hanging out all over the place, GDAL/OGR will look like an unfinished college project. 

And as a matter of fact, of the thousands of programs in my /usr/bin , guess which are the only ones that even have any 'language extension' at all. It makes me wonder if they somehow slipped in against some Debian policy or rules of style or something.

jidanni:
A N D    T H E R E   you have it:

https://www.debian.org/doc/debian-policy/ch-files.html

> When scripts are installed into a directory in the system PATH, the script name should not include an extension such as .sh or .pl that denotes the scripting language currently used to implement it.

jratike80:
After this https://github.com/OSGeo/gdal/pull/5296 a part of the Python scripts can be launched without .py extension. See also https://github.com/OSGeo/gdal/issues/5281

The change does not affect older GDAL versions but I guess that the simple syntax that you suggest is OK but I would like to have some Python expert to confirm. 

> Certainly such users not playing with a full deck will learn very quickly via error messages telling them to mend their ways.

Glad that you also have trust in our users. I would still keep the .py for example on this list so the users can quickly see which tools require Python and GDAL Python bindings, and which not.

image

jidanni:
> I would still keep the .py for example on this list so the users can quickly see which tools require Python and GDAL Python bindings, and which not.

OK, but change
> bla_bla.py

to
> bla_bla (py)

or
> bla_bla (python)

please.

jratike80:
Maybe there should be then a clear mention about tool being a Python script.

Now
> gdal_calc.py
> Command line raster calculator with numpy syntax.

Then 
> gdal_calc
> Command line raster calculator with numpy syntax. The tool is a Python script that requires GDAL Python bindings https://gdal.org/api/python_bindings.html

doc/source/programs/ogrmerge.rst
```diff
@@ -183,18 +183,24 @@ Create a VRT with a layer for each input shapefiles
...
 
 .. code-block::
 
-    ogrmerge.py -single -o merged.shp france.shp germany.shp -src_layer_field_name country
+    ogrmerge -single -o merged.shp france.shp germany.shp -src_layer_field_name country
+
+Notes
+-----
+
+Older versions of ogrmerge were called ogrmerge.py.
```
rouault, line 205:
```suggestion
ogrmerge is a Python utility, and is only available if GDAL Python bindings are available.
```

PR #10018, opened 2024-05-27
Update gdal_grid.rst finally documenting vital -zfield field_3

jidanni:
Here we finally document that -zfield can use "field_3" ! How did we ever figure it out??? Trial and error!!

### DIFF
--- a/doc/source/programs/gdal_grid.rst
+++ b/doc/source/programs/gdal_grid.rst
@@ -473,6 +473,21 @@ in the VRT file in the following way:
 
     <GeometryField encoding="PointFromColumns" x="field_1" y="field_2" z="field_3"/>
 
+And, if both not using a VRT file, nor column headers:
+
+.. code-block:: bash
+
+   gdal_grid -l dem -oo X_POSSIBLE_NAMES=field_1 \
+      -oo Y_POSSIBLE_NAMES=field_2 -zfield field_3 dem.csv dem.tif
+
+or
+
+.. code-block:: bash
+
+    gdal_grid -l dem -oo X_POSSIBLE_NAMES=field_1 \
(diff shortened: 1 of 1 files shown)


jratike80:
Another option would have been to read the CSV format man page. Or check the source data with ogrinfo

ogrinfo foo.csv -al -so
INFO: Open of `foo.csv'
      using driver `CSV' successful.

```
Layer name: foo
Geometry: None
Feature Count: 2
Layer SRS WKT:
(unknown)
field_1: String (0.0)
field_2: String (0.0)
field_3: String (0.0)
field_4: String (0.0)
```

jidanni:
> Another option would have been to read the CSV format man page. Or check the source data with ogrinfo

In fact these are great things for me to mention in this same spot too, once this first PR gets processed.

jratike80:
Before adding the same information into every possible spot where CSV data can be utilized, try to consider if it could be possible enhance the documentation so that it is not needed.
Maybe the home page could have pointers:

Read the driver manual
Read the utility manual
Read about the common options
Learn about the configuration options

See the advanced, user provided, usage examples

jidanni:
Sure, there are two extremes, put on page 1 all the proper things one should do, or instead have an AI chatbot floating on the right margin of the man pages.

All I know is we underprivileged occasional headerless .csv users need all the tips we can get if we are going to be on the same footing as the other kids. Anyway the other kids can just chuckle as they skim over the extra words added that would be a lifesaver for us .csv users who, face it, just have a tougher time with OGR/GDAL, yes, even if we do read the CSV driver page several times.

rouault:
Closing for the same reason as #10016 . And I would argue that CSV files without headers are already quite an advanced use case

PR #10022, opened 2024-05-27
Update gdal_contour.rst to say must specify either -i or -fl or -e

jidanni:
Tell users on the man page. Don't wait for the traffic cop (error message) to tell them the exact same thing!

### DIFF
--- a/doc/source/programs/gdal_contour.rst
+++ b/doc/source/programs/gdal_contour.rst
@@ -94,6 +94,7 @@ be on the right, i.e. a line string goes clockwise around a top.
 .. option:: -i <interval>
 
     Elevation interval between contours. Ignored if -fl is used.
+    Must specify either -i or -fl or -e.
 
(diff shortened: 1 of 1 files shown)


doc/source/programs/gdal_contour.rst
```diff
@@ -94,6 +94,7 @@ be on the right, i.e. a line string goes clockwise around a top.
 .. option:: -i <interval>
 
     Elevation interval between contours. Ignored if -fl is used.
+    Must specify either -i or -fl or -e.
```
rouault, line 97: shouldn't that mention be moved in the above "description" part to avoid repeated it 3 times?

jidanni: One could say I am putting a warning label on each individual cigarette, instead of just the pack.

But that's the only way one can guarantee users won't come howling when they hit one of the booby trapped options...

PR #10052, opened 2024-05-28
HDF5: add support for libhdf5 >= 1.14.4.2 when built with Float16

rouault:
Fixes https://github.com/HDFGroup/hdf5/issues/4527

### DIFF
--- a/frmts/hdf5/hdf5dataset.cpp
+++ b/frmts/hdf5/hdf5dataset.cpp
@@ -44,6 +44,7 @@
 
 #include "cpl_conv.h"
 #include "cpl_error.h"
+#include "cpl_float.h"
 #include "cpl_string.h"
 #include "gdal.h"
 #include "gdal_frmts.h"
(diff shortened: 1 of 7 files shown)


imincik:
@rouault , great thanks for this patch. I confirm that GDAL 3.9.0 is building fine (including tests) with HDF5 1.14.4.3 .

frmts/hdf5/hdf5dataset.cpp
```diff
@@ -1132,6 +1178,28 @@ static herr_t HDF5AttrIterate(hid_t hH5ObjID, const char *pszAttrName,
                 psContext->m_osValue += szData;
             }
         }
+#ifdef HDF5_HAVE_FLOAT16
+        else if (H5Tequal(H5T_NATIVE_FLOAT16, hAttrNativeType) > 0)
+        {
+            for (hsize_t i = 0; i < nAttrElmts; i++)
+            {
+                const uint16_t nVal16 = static_cast<uint16_t *>(buf)[i];
+                const uint32_t nVal32 = CPLHalfToFloat(nVal16);
```
elpaso, line 1187: I'm confused by CPLHalfToFloat (which doesn't return a float) and the memcpy below (which assumes sizeof float), I know that they are both 32 bits but I am wondering if that assumption is enforced by standard or if is it "assert"ed somewhere.

rouault: We have a ``static_assert(sizeof(float) == 4);`` in port/cpl_conv.cpp

PR #10060, opened 2024-05-29
MiraMonVector: fix a case sensitive comparison

AbelPau:
## What does this PR do?
Solves the case of giving a false error when the translating a polígon with a capital extension reference to its arc.

Some other strstr() have been corrected to a version of windows stristr() called MM_stristr() cause it's not important to be case sensitive

## Tasklist

### DIFF
--- a/ogr/ogrsf_frmts/miramon/mm_gdal_functions.c
+++ b/ogr/ogrsf_frmts/miramon/mm_gdal_functions.c
@@ -1262,13 +1262,13 @@ reintenta_lectura_per_si_error_CreaCampBD_XP:
             if (11 > (read_bytes = fread_function(charset_cpg, 1, 10, f_cpg)))
             {
                 charset_cpg[read_bytes] = '\0';
-                p = strstr(charset_cpg, "UTF-8");
+                p = MM_stristr(charset_cpg, "UTF-8");
                 if (p)
                     pMMBDXP->CharSet = MM_JOC_CARAC_UTF8_DBF;
-                p = strstr(charset_cpg, "UTF8");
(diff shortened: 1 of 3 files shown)


rouault:
stristr() is not portable. You may need to implement a MMstristr() function that for example uses the EQUAL() (case-insensitive) macro iterating over the first string of stristr()

AbelPau:
> stristr() is not portable. You may need to implement a MMstristr() function that for example uses the EQUAL() (case-insensitive) macro iterating over the first string of stristr()

Thanks Even, I was trying to figure out that. Thanks!

rouault:
(note: if those files had been .cpp files, you could have used ``CPLString(pszHaystack).ifind(pszNeedle) != std::string::npos``)

ogr/ogrsf_frmts/miramon/mm_gdal_functions.c
```diff
@@ -1947,6 +1947,50 @@ char *MM_oemansi_n(char *szszChain, size_t n_bytes)
...
+
+    char *p1 = (char *)haystack;
+    char *p2 = (char *)needle;
+    char *p1Adv = (char *)haystack;
+
+    while (*++p2)
+    {
+        p1Adv++;
+    }
+
```
rouault, line 1970: I can't make sense of that. Why would you skip the first strlen(needle) character of haystack ? That's wrong, and that can cause an out-of-bounds access if strlen(haystack) < strlen(needle). And please use EQUALN() to simplify the code

AbelPau: Sorry, I've rewritten it again using the EQUALN() function. 
Now it's very more clear.

PR #10068, opened 2024-05-29
RFC96 Deferred driver loading adjustment to fix functional regression

rouault:
RFC96 adjusted with:

Adjustments done post GDAL 3.9.0, for GDAL 3.9.1
------------------------------------------------

After GDAL 3.9.0 release, it has been noticed that the following setup which
used to work in prior releases no longer worked:

- Step 1: building libgdal without support for driver X
- Step 2: building driver X as a plugin, discarding the libgdal share library,
          built at that stage
- Step 3: using driver X built as a plugin against libgdal built at step 1. In that
          scenario, driver X is expected to be loaded as if it was an out-of-tree drivers.

Such scenario is used when delivering a fully open-source libgdal without any
prior knowledge of which drivers could be later built as plugins, or for which
pre-configuring libgdal to support such drivers is not practical because they
rely on a proprietary SDK and the identification method and/or driver metadata
depends on the availability of the SDK include files (e.g. MrSID).

Starting with GDAL 3.9.1, the ``add_gdal_driver()`` function in the CMakeLists.txt
of drivers which use the ``CORE_SOURCES`` keyword must also declare the
``NO_SHARED_SYMBOL_WITH_CORE`` keyword, so that the files pointed by CORE_SOURCES
are built twice: once in libgdal with a ``GDAL_core_`` prefix, and another time
in the plugin itself with a ``GDAL_driver_`` prefix, by using the
PLUGIN_SYMBOL_NAME() macro of :file:`gdal_priv.h`.

Example in ogr/ogrsf_frmsts/oci/CMakeLists.txt:

.. code-block::

    add_gdal_driver(TARGET ogr_OCI
                    SOURCES ${SOURCE}
                    CORE_SOURCES ogrocidrivercore.cpp
                    PLUGIN_CAPABLE
                    NO_SHARED_SYMBOL_WITH_CORE)

Example in ogrocidrivercore.h:

.. code-block:: cpp

    #define OGROCIDriverIdentify \
       PLUGIN_SYMBOL_NAME(OGROCIDriverIdentify)
    #define OGROCIDriverSetCommonMetadata \
       PLUGIN_SYMBOL_NAME(OGROCIDriverSetCommonMetadata)

    int OGROCIDriverIdentify(GDALOpenInfo *poOpenInfo);

    void OGROCIDriverSetCommonMetadata(GDALDriver *poDriver);

A consequence of that change is that drivers built as a plugin against GDAL 3.9.0
will not be loadable by GDAL 3.9.1 (or later patch in the 3.9 series), because
they relied on driver-specific functions that are no longer exported by libgdal >= 3.9.1.

After that, things should work as they used to, and drivers built against libgdal 3.9.1
should work against libgdal 3.9.2 for example.

Also note that the above only affects *in-tree* plugin drivers. Out-of-tree plugin drivers
are not affected.


hobu:
> Step 1: building libgdal without support for driver X
    Step 2: building driver X as a plugin, discarding the libgdal share library,
    built at that stage
    Step 3: using driver X built as a plugin against libgdal built at step 1. In that
    scenario, driver X is expected to be loaded as if it was an out-of-tree drivers.

Confirmed this behavior now works as expected using this PR and compiling the MrSID driver against @conda-forge's GDAL 3.9.

rouault:
> After that, things should work as they used to, and drivers built against libgdal 3.9.1
> should work against libgdal 3.9.2 for example.

With the caveat that the Identification method (and metadata) embedded in libgdal 3.9.1 would be used, and not the one of the driver of 3.9.2. So such mix of versions is not generally desirable (for in-tree deferred loading compatible drivers. for out-of-tree drivers, no problem, as libgdal doesn't know anything about them)

Issue #10073, opened 2024-05-30
ogr2ogr -gcp doesn't return in error when not enough GCPs are specified (or probably some other error during GCP evaluation)


jidanni:
### Feature description

Apparently there are ERRORs, fatal ERRORs, and WARNINGs

I would make all ERRORs fatal.

E.g., when not given enough GCPs (#10074), ogr2ogr, as of version 3.9 says
> ERROR 1: Failed to compute GCP transform: Not enough points available

but still doesn't return an error ($?=1) to the shell. (Our Makefile trundles on. ~~Nobody has told it our back axle is gone...~~)

Here is ERROR 1 again, but at least this returns $?=1 to the shell:
```
ogr2ogr -f LIBKML all.kmz all.kml \
-sql 'SELECT name FROM "all"'
ERROR 1: SELECT from table all failed, no such table/featureclass.
```
Then there is the case of when returning errors, should the output file be created anyway?

PR #10074, opened 2024-05-30
Update ogr2ogr.rst: not two, but three GCPs

jidanni:
See #10073.


rouault:
2 GCPs will be supported by https://github.com/OSGeo/gdal/pull/10078

PR #10087, opened 2024-05-31
Clarify units and directional distortion of -simplify and -segmentize

jidanni:
Sure hope I got the details right!

### DIFF
--- a/doc/source/programs/ogr2ogr.rst
+++ b/doc/source/programs/ogr2ogr.rst
@@ -450,9 +450,23 @@ output coordinate system or even reprojecting the features during translation.
     topology per feature, in particular for polygon geometries, but not for a
     whole layer.
 
+    The specified value of this option is the tolerance used to merge
+    consecutive points of the output geometry using the
+    :cpp:func:`OGRGeometry::Simplify` method
+    The unit of the distance is in 
+    georeferenced units of the target vector dataset.
+    This option is applied after the reprojection implied by :option:`-t_srs`
+    Note for the same given tolerance value, actual on the ground
+    east-west tolerance decreases towards the poles.
+
 .. option:: -segmentize <max_dist>
 
-    Maximum distance between 2 nodes. Used to create intermediate points.
+    The specified value of this option is the maximum distance between two
+    consecutive points of the output geometry before intermediate points are added.
+    The unit of the distance is georeferenced units of the source raster.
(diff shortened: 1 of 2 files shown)


doc/source/programs/ogr2ogr.rst
```diff
@@ -450,9 +450,23 @@ output coordinate system or even reprojecting the features during translation.
...
 
+    The specified value of this option is the tolerance used to merge
+    consecutive points of the output geometry using the
+    :cpp:func:`OGRGeometry::Simplify` method
+    The unit of the distance is in 
+    georeferenced units of the target vector dataset.
+    This option is applied after the reprojection implied by :option:`-t_srs`
+    Note for the same given tolerance value, actual on the ground
+    east-west tolerance decreases towards the poles.
+
```
jratike80, line 461: The east-west toleranse applies only to some CRS (EPSG:4326, EPSG:3857). In most real world use cases of gdal_footprint this thing does not happen and this texts adds mostly confusion.

jidanni: Please mention it there.

And mention alternatives if such maps produced are not acceptable.

Thanks.

jratike80:
I do not think that we should start documenting for every utility that can be used with functions like buffer or simplify or area or lenght etc. that the on the ground length of a degree varies according to latitude.

In case of gdal_footprint I think that the user can safely simplify and segmentize in EPSG:4326 without even knowing that the Earth is not flat. The simplified or segmentized footprint works well as a mask or as an index of the location of the image.

doc/source/programs/ogr2ogr.rst
```diff
@@ -450,9 +450,23 @@ output coordinate system or even reprojecting the features during translation.
...
     whole layer.
 
+    The specified value of this option is the tolerance used to merge
+    consecutive points of the output geometry using the
+    :cpp:func:`OGRGeometry::Simplify` method
+    The unit of the distance is in 
+    georeferenced units of the target vector dataset.
+    This option is applied after the reprojection implied by :option:`-t_srs`
+    Note for the same given tolerance value, actual on the ground
+    east-west tolerance decreases towards the poles.
```
rouault, line 460:
```suggestion
    The unit of the distance is in 
    georeferenced units of the source vector dataset.
```

PR #10111, opened 2024-06-03
[argparse] update library

elpaso:
Update to PR https://github.com/p-ranav/argparse/pull/356

Also fixes the incompatibility between store_into and scan and action: when the three methods above were called, being all based on the (unique) action, the last one would overwrite the previous ones.

This issue was making the parser strictly dependent on the order of the scan/store_into/action calls making them mutually exclusive.

### DIFF
--- a/apps/argparse/README.TXT
+++ b/apps/argparse/README.TXT
@@ -1,3 +1,4 @@
 Provenance: https://github.com/p-ranav/argparse
 
-Retrieved from https://github.com/p-ranav/argparse/blob/b85a0a414d415fccdf6d77072421b1d3887c8f79/include/argparse/argparse.hpp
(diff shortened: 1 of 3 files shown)


rouault:
The test failure might require to apply https://github.com/OSGeo/gdal/pull/10093 first

elpaso:
> The test failure might require to apply #10093 first

I think that the issue comes from `.scan<'g', double>()`, before the changes https://github.com/p-ranav/argparse/pull/356 because `.scan` was internally implemented as an `.action` and only the last action was actually stored, `.scan` was completely ignored.

Now, after https://github.com/p-ranav/argparse/pull/356 you can call multiple `.actions` and they will be executed in order.

`.scan` is not compatible with the `-burn` variants because it expects a single double. I think the solution here is to remove the scan.

PR #10112, opened 2024-06-03
Update OAPIF examples 

tjay:
The OAPIF examples contain outdated URLs

## What does this PR do?

Update sample URLs to more current service URLs


tjay:
The requests from the current documentation cause a high load on our servers. This adjustment changes the URLs to the current OGC API implementation (xtraserver-webapi/ldproxy) and to services from the state of North Rhine-Westphalia with a smaller amount of data.

Issue #9978, opened 2024-05-21
Update Identify() method of drivers using a "DRIVER_NAME:" strategy to take into account poOpenInfo->papszAllowedDrivers


rouault:
A number of drivers may (or require) a "DRIVER_NAME:" prefix to help them identify a dataset. They could also take into account when poOpenInfo->papszAllowedDrivers has a single value which is the driver being tested.
Cf https://github.com/OSGeo/gdal/pull/9970#issuecomment-2122571722

PR #10122, opened 2024-06-03
GeoJSON/GeoJSONSeq/TopoJSON/ESRIJSON/JSONFG: relax identification checks 

rouault:
... when papszAllowedDrivers[] contains only the driver name

Refs https://github.com/OSGeo/gdal/issues/9978 (this PR will be followed by another one with other drivers)


rouault:
superseded per #10185

Issue #10123, opened 2024-06-04
Docs: mention user-agent default value


benjimin:
### Feature description

The [docs](https://gdal.org/user/configoptions.html) for `GDAL_HTTP_USERAGENT` _should_ say what the **default value** is.

#6376 added a default value for the HTTP user agent header used by GDAL. According to comments in the merged code, the default HTTP user agent is `GDAL/x.y.z` (denoting the version). The configuration docs should be updated to reflect this.

This is important because knowing the user agent header string is necessary to correctly configure firewalls, to allow legitimate access to data collections while blocking spurious scraping by generic web crawlers and AI trainers. (Note some major hosting providers such as AWS charge for egress, not ingress, so significant costs can be inadvertently incurred by permitting general machine access which isn't expected to be making productive use of the structured content of the geospatial datasets.)

jratike80:
Doxygen documentation about the thing https://gdal.org/doxygen/cpl__http_8h.html#ae83c06d51089bb340ed0b98a6ed8d01d.

benjimin:
Here is an extract from the current docs. Note that defaults are expressed for other configuration options but not for this one:

https://gdal.org/user/configoptions.html

- _**GDAL_HTTP_USERAGENT=value:** When set this string will be used to set the User-Agent header in the http request sent to the remote server._

- _**GDAL_HTTP_UNSAFESSL=[YES/NO]:** Defaults to `NO`. Set to "`YES`" to get the curl library to skip SSL host / certificate verification._

jratike80:
Would you like to try the "Edit on GitHub" button in the top-right corner of the page and create a pull request?

jratike80:
I made a quick code search and found some potentially different ways to define the user agent:

**WMTS**: GDAL WMS Driver, can be set in the XML file
**ogr_geocoding**: what is this? Defaults to GDAL/OGR version string
**ECW**: what is this? NCSecwSetConfig(NCSCFG_ECWP_CLIENT_HTTP_USER_AGENT, 
     "ECW GDAL Driver/" NCS_ECWJP2_FULL_VERSION_STRING_DOT_DEL);
**ogramigocloud**: what is this?  userAgent << "USERAGENT=gdal/AmigoCloud build:"

PR #10124, opened 2024-06-04
User configuration docs: mention default User-Agent

benjimin:
## What does this PR do?

Document the default value for this configuration variable.

#10123

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/doc/source/user/configoptions.rst
+++ b/doc/source/user/configoptions.rst
@@ -928,8 +928,9 @@ Networking options
 
 -  .. config:: GDAL_HTTP_USERAGENT
 
-      When set this string will be used to set the ``User-Agent`` header in the http
+      This string will be used to set the ``User-Agent`` header in the HTTP
       request sent to the remote server.
+      Defaults to "GDAL/x.y.z" where x.y.z is the current version.
 
 -  .. config:: GDAL_HTTP_UNSAFESSL
       :choices: YES, NO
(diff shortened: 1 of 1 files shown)


benjimin:
Would it be preferable to use the `:default: GDAL/version` annotation (rather than using prose to describe the default value)?

doc/source/user/configoptions.rst
```diff
@@ -928,8 +928,9 @@ Networking options
 
 -  .. config:: GDAL_HTTP_USERAGENT
 
-      When set this string will be used to set the ``User-Agent`` header in the http
+      This string will be used to set the ``User-Agent`` header in the HTTP
       request sent to the remote server.
+      Defaults to "GDAL/x.y.z" where x.y.z is the current version.
```
jratike80, line 934: I wonder the wording "current version". Should it rather be "where x.y.z is the GDAL version" or "GDAL build version"? I think that the current GDAL version is always the most recent one https://gdal.org/download.html#current-release

doc/source/user/configoptions.rst
```diff
@@ -928,8 +928,9 @@ Networking options
...
```
rouault, line 933:
```suggestion
      Defaults to "GDAL/x.y.z" where x.y.z is the GDAL build version.
```

PR #10126, opened 2024-06-04
VSIStatL(): allow trailing slash on mingw64 builds

rouault:


rouault:
cherry picked in 3.9 per f62e5a5295317dd86f5eb0c4f34078bd1674ea92

PR #10136, opened 2024-06-05
Block cache: Use unordered_set for hash block cache.

abellgithub:
Hashing should perform better than an ordered container.

A comment in the code said that tests depended on the comparator, but things still pass and there is nothing in the hash block cache code itself that suggests that order matters.

### DIFF
--- a/gcore/gdalhashsetbandblockcache.cpp
+++ b/gcore/gdalhashsetbandblockcache.cpp
@@ -48,7 +48,7 @@ class GDALHashSetBandBlockCache final : public GDALAbstractBandBlockCache
 {
     struct BlockHash
     {
-        std::size_t operator()(GDALRasterBlock *const& b) const noexcept
+        std::size_t operator()(GDALRasterBlock *const &b) const noexcept
         {
             std::hash<int> h;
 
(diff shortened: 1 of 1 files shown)


gcore/gdalhashsetbandblockcache.cpp
```diff
@@ -47,23 +46,32 @@
...
 {
-    struct BlockComparator
+    struct BlockHash
     {
-        // Do not change this comparator, because this order is assumed by
-        // tests like tiff_write_133 for flushing from top to bottom, left
-        // to right.
-        bool operator()(const GDALRasterBlock *const &lhs,
-                        const GDALRasterBlock *const &rhs) const
+        std::size_t operator()(GDALRasterBlock *const &b) const noexcept
```
rouault, line 51: The left shift could potentially trigger a sanitizer warning (what's the rationale for it by the way? a code comment could be useful)
```suggestion
        CPL_NOSANITIZE_UNSIGNED_INT_OVERFLOW
        std::size_t operator()(GDALRasterBlock *const &b) const noexcept
```

rouault (commented):
Are there timings measurement showing an improvement ? There's a discussion in https://stackoverflow.com/questions/1349734/why-would-anyone-use-set-instead-of-unordered-set showing that unordered_set is not always better than set

rouault:
(the Docs CI failure is unrelated. I've just fixed it in master)

dbaston:
The story I've heard (complete hearsay) is that `std::unordered_xyz` implementations are slower than alternatives outside of the STL but can't be substantially improved because it would require unacceptable API breakages. If we are going to change the implementation, it might also be worth testing third-party libraries that could be easily vendored.

I don't know much (anything) about the GDAL block cache's requirements in terms of set size, need for stable references, etc., so I don't know if something like https://github.com/martinus/unordered_dense might be appropriate.

gcore/gdalhashsetbandblockcache.cpp
abellgithub: Hash functions are magic and I have never spent much time on them. I stole this from elsewhere expecting that the original author spent more time than me. If we're pretty sure that we're working with 64 bits as size_t (which is probably most platforms now), you could probably just take the two numbers and force them into a uint64_t and run std::hash<uint64_t> on that. But if someone else wants to spend time finding a great hash function that would be fine. I just like simple.

abellgithub:
People spend lots of time working on algorithms. I generally expect that the std:: ones are good enough and will improve over time without my doing anything. This happened recently with std::sort, for example.

My understanding is that if you have unordered data, use a hash. This is why the std::unordered_<> containers were introduced. I haven't benchmarked this but I'm sure others have and if it weren't usually faster, there's no reason for them to exist.

If the size is sure to be small (not sure what counts as small), it's almost always better these days just to iterate contiguous memory when the equality test is cheap. This is why flat_set and flat_map have been introduced (but not until C++23).

Caching data is a big topic and people spend tons of time working on just the right thing for their application. I'm not suggesting this is perfect. It's just simple and hopefully a small improvement.

rouault:
> It's just simple and hopefully a small improvement.

Well, that's not clear to me what this PR improves :-) Performance: not clear, and it slightly increases the number of lines of code.

And despite tiff_write_133 not actually failing (probably that the array based block cache implementation is used now for it), the current logic has the advantage of flushing blocks in a predictable way: top-to-bottom, left-to-right. So in some respect, blocks are ordered.

abellgithub:
Why would the order of block flushing matter? Aren't you just dropping the data? If you want an LRU cache, that's something different from what's here.

Anyway, close if you don't want it. I've stated my thinking.

rouault:
> Why would the order of block flushing matter?

what we discussed in the gdal_viewshed thread: so that blocks are written in the file in a predictable way. The tiff_write_133 test checks a very particular feature of TIFF reading and writing, which is the possibility to generate TIFF files in a purely streaming way (/dev/stdout/) in some conditions (no compression, top-to-bottom left-to-right block ordering), and reading them similarly.

That's different from LRU. The global list of blocks in gdalrasterblock is LRU, but when flushing the blocks referenced in GDALArrayBandBlockCache or GDALHashSetBandBlockCache at dataset closing, we try to do that in a "raster friendly" order

PR #10156, opened 2024-06-07
OGR_PG_SKIP_CONFLICTS: optionally insert with ON CONFLICT DO NOTHING

jef-n:
speaking of small contributions…

refs norBIT/alkisimport#82

### DIFF
--- a/doc/source/drivers/vector/pg.rst
+++ b/doc/source/drivers/vector/pg.rst
@@ -467,6 +467,7 @@ The following configuration options are available:
 -  .. config:: OGR_PG_SKIP_CONFLICTS
       :choices: YES, NO
       :default: NO
+      :since: 3.10
 
       If set to "YES" (not the default),
       conflicts when inserting features will be skipped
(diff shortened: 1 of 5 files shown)


rouault:
@jef-n Could you add a small test for that into autotest/ogr/ogr_pg.py ?

rouault:
thanks for the added test. There's a code linting issue. You need to setup pre-commit. See https://gdal.org/development/dev_practices.html#commit-hooks (you'll need to run ``pre-commit run --all-files`` to fix the existing issues)

doc/source/drivers/vector/pg.rst
```diff
@@ -464,6 +464,16 @@ The following configuration options are available:
       create) the ``ogr_system_tables.metadata`` table to retrieve and store
       layer metadata.
 
+-  .. config:: OGR_PG_SKIP_CONFLICTS
+      :choices: YES, NO
+      :default: NO
```
rouault, line 469:
```suggestion
      :default: NO
      :since: 3.10
```

PR #10170, opened 2024-06-09
Contour: include minimum raster value as contour line when it exactly matches the first level

### LINKED ISSUE #10167: gdal_contour skips first contour when on edge
Here the `0.5` is missing. Only `1` and `1.5` were output. Gdal 3.9.

```
gdal_create -q -outsize 2 2 a.xyz
perl -awnle '$F[-1]=$F[0]; print "@F"' a.xyz > d.xyz
cat d.xyz
0.5 -0.5 0.5
1.5 -0.5 1.5
0.5 -1.5 0.5
1.5 -1.5 1.5
gdal_contour -q -i 0.5 -a Z d.xyz d.csv
cat d.csv
ID,Z
0,1
1,1.5
```

And it makes just as little sense as if `1.5` was missing, and only `0.5` and `1` were output.

Nor does the man page document that the first contour is skipped if there are no data points below it.

Which is good, as documenting that would be just as bad as saying that the last contour is skipped if there are no data points above it.

If contours on edges are to be skipped, then at least skip consistently.

But that would leave us with only one contour, at `1`.

Solution: fix the bug that arbitrarily skips the first contour, even though there are two or more data points right on it, that should become part of a new LINESTRING contour.

Else it is like buses that run every half hour from 8:30 AM to 5 PM (but oops, not the first bus.)

rouault:
Fixes #10167

There's currently an asymmetry since the maximum raster value is included. This is due how the marching square thresholding compares pixel values to the candidate contour levels. Add a special case to modify the comparison for the level that matches the minimum raster value.

### DIFF
--- a/alg/marching_squares/square.h
+++ b/alg/marching_squares/square.h
@@ -203,9 +203,9 @@ struct Square
     //
     //     ^
     //  -  |  +
-    Segments segments(double level) const
+    Segments segments(double level, double minLevel) const
     {
-        switch (marchingCase(level))
+        switch (marchingCase(level, minLevel))
         {
             case (ALL_LOW):
                 // debug("ALL_LOW");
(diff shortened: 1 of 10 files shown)


alg/marching_squares/square.h
```diff
@@ -269,16 +281,17 @@ struct Square
                 // The classical marching square algorithm says the ambiguity
                 // should be resolved between the two possible configurations by
                 // looking at the value of the center point. But in certain
-                // cases, this may lead to line contours from different levels
+                // cases, this may lead to line contours from different level, minLevels
```
elpaso, line 284: is this from a search replace or was it intended?
```suggestion
                // cases, this may lead to line contours from different level
```

alg/marching_squares/utility.h
```diff
@@ -36,14 +36,14 @@ namespace marching_squares
 
 // This is used to determine the maximum level value for polygons,
 // the one that spans all the remaining plane
-const double Inf = std::numeric_limits<double>::max();
+const double Inf = std::numeric_limits<double>::infinity();
```
elpaso, line 39: maybe constexpr

alg/marching_squares/utility.h
```diff
@@ -36,14 +36,14 @@ namespace marching_squares
...
 
 const double NaN = std::numeric_limits<double>::quiet_NaN();
 
 #define debug(format, ...) CPLDebug("MarchingSquare", format, ##__VA_ARGS__)
 
 // Perturb a value if it is too close to a level value
-inline double fudge(double level, double value)
+inline double fudge(double level, double value, double minLevel)
```
elpaso, line 46: if this API can be changed maybe better if it it was
```suggestion
inline double fudge(double value, double minLevel, double level)
```

elpaso (approved):

alg/marching_squares/square.h
rouault: done

alg/marching_squares/utility.h
rouault: done

alg/marching_squares/utility.h
rouault: done

PR #10184, opened 2024-06-10
Viewshed: create current extent

abellgithub:
Shift output extent to current extent to avoid incorrect assumptions.


rouault:
> Shift output extent to current extent to avoid incorrect assumptions.

is that a bug fix ? (specific to master?) doesn't this deserve a new regression test?

abellgithub:
No bug. Just an incremental change to what I'm doing. I can hold it if you want.

rouault:
> I can hold it if you want.

no, this PR is fine. Thanks for clarifying

PR #10188, opened 2024-06-10
MRF: Faster LERC V1 encoding

lucianpls:
## What does this PR do?

Make LERC V1 encoding in MRF about twice as fast, by avoiding temporary vector use

## Tasklist

 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:


hobu:
Could this be adapted to TIFF LERC encoding too, or is it not relevant there?

lucianpls:
> Could this be adapted to TIFF LERC encoding too, or is it not relevant there?

No, this is only for Lerc v1, which is not used in TIFF at all. I keep it maintained because it is the Esri elevation tile format.

PR #10193, opened 2024-06-11
[gdalwarp] Fix -ovr

### LINKED ISSUE #10174: gdalwarp -ovr {absolute_value} doesn't influence choice of output resolution
Cf https://lists.osgeo.org/pipermail/gdal-dev/2024-June/059079.html

``gdalwarp -ovr 0 -overwrite input.tif output.tif`` should take into account the resolution of the 1st overview level of the input dataset to compute the output dataset

elpaso:
Fix #10174 -ovr {absolute_value} doesn't influence choice of output resolution


apps/gdalwarp_lib.cpp
```diff
@@ -2502,10 +2502,21 @@ static GDALDatasetH GDALWarpDirect(const char *pszDest, GDALDatasetH hDstDS,
             CPLSPrintf("%g", psOptions->dfErrorThreshold));
     }
 
+    // Examine desired overview level and retrieve the corresponding dataset
+    // if it exists.
+    GDALDatasetH hDstDSOverview = nullptr;
+    if (psOptions->nOvLevel >= 0)
+    {
+        hDstDSOverview = GDALCreateOverviewDataset(
```
rouault, line 2510: This will leak hDstDSOverview, and more annoyingly this will crash if nSrcCount > 1.
I believe the call to GDALCreateOverviewDataset() should be moved to CreateOutput() and applied to all input datasets

elpaso: I see, it's much more complex that I imagined. 

I think I  have now the proper implementation but there is still a failing test that I don't know how to handle (what is the expected behavior):

```python
    # Crates a full res output file
    gdaltest.runexternal(f"{gdalwarp_path} {src_tif} {dst_tif}")
    # The output file has now a "wrong" resolution and no overwrite was specified
    gdaltest.runexternal(f"{gdalwarp_path} {src_tif} {dst_tif} -ovr 0")
    ds = gdal.Open(dst_tif)
    # Fail! Full res is used instead of overview
    assert ds.GetRasterBand(1).Checksum() == cs_ov0
```

What are we supposed to do here?
Error out or silently overwrite the output file (that has full res instead of the ovr res)?

The code path in this case skips createOutput completely.

elpaso: this is also unexpected (even if the output res is correct) https://github.com/OSGeo/gdal/pull/10193/files#diff-143c752f11c4a0dd7cde1c002e04f4df6f174faebddab77e602b5afff6dfe8f8R1139

autotest/utilities/test_gdalwarp.py
```diff
@@ -1113,6 +1118,37 @@ def test_gdalwarp_40(gdalwarp_path, tmp_path):
...
+    gdaltest.runexternal(f"{gdalwarp_path} {src_tif} {dst_tif} -ovr 0")
+
+    # Repeat with wrong resolution output file and no overwrite (takes a different code path)
+    os.unlink(dst_tif)
+    # Crates a full res output file
+    gdaltest.runexternal(f"{gdalwarp_path} {src_tif} {dst_tif}")
+    gdaltest.runexternal(f"{gdalwarp_path} {src_tif} {dst_tif} -ovr 0")
+    ds = gdal.Open(dst_tif)
+    # FIXME: This fails: is that expected?
+    # assert ds.GetRasterBand(1).Checksum() == cs_ov0
```
rouault, line 1142: yes, that's expected. You are updating an existing dataset, created at full resolution and size, with a overview of a source dataset. Hence the checksum of the output cannot match the source overview, because there is not the same number of pixels. I believe lines 1128 to 1143 don't bring much in terms o testing coverage and could probably be removed

apps/gdalwarp_lib.cpp
rouault: > What are we supposed to do here?
> Error out or silently overwrite the output file (that has full res instead of the ovr res)?

There should already be a warning emitted at line 2860 (in master): "cannot get overview level ...". This is probably enough for now

autotest/utilities/test_gdalwarp.py
elpaso: do you mind if I keep (most of) them ?  They have been helpful while developing, because they trigger different code paths (depending on whether there is or isn't an existing output file).

rouault: fine for me

PR #10194, opened 2024-06-11
csv.rst: Show users how to convert haphazard CSVs into WKT CSVs

jidanni:
Here we finally show users how to make CSVs that no longer require massive command line baggage to access. I.e., now CSV is no longer at a competitive disadvantage to other formats.

### DIFF
--- a/doc/source/drivers/vector/csv.rst
+++ b/doc/source/drivers/vector/csv.rst
@@ -177,6 +177,21 @@ say it must be surrounded by double-quotes, and double-quotes inside the
 string must be repeated for proper escaping), the name of such column(s)
 can be derived from the :oo:`GEOM_POSSIBLE_NAMES` open option.
 
+Let's see how it is possible to make our above CSV file not require any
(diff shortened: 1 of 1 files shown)


rouault:
I'm quite of worried about the CSV driver doc going out of control with lots of examples, partially overlapping. The new ones suggested combine other similar examples in this page demonstrating -lco GEOMETRY=AS_WKT or the use of Spatialite MakeLine(), but by using it with ogr2ogr instead of ogrinfo. Where do we stop in providing examples?

@jratike80 opinion?

jidanni:
Maybe the page should just show users how to convert haphazard CSVs into WKT format.

After doing that their CSV is just like any other OGR source and needs no further explanation.

jratike80:
In my opinion the peculiar use cases of GDAL utilities and CSV should be collected into an own tutorial. My motivation is not to hide the information from the CSV page but also to serve better those CSV enthusiasts.

PR #10213, opened 2024-06-14
[gpkg] Implement ALTER TABLE RENAME for rasters

### LINKED ISSUE #10201: GPKG: support ALTER TABLE ... RENAME TO ... for raster layers
### Feature description

The ExecuteSQL() of GPKG already catches this for vector layers, to do more than just table renaming, and also do necessary changes in auxiliary tables. It would be nice to support raster layers to. Cf https://github.com/qgis/QGIS/pull/57754

### Additional context

_No response_

elpaso:
Fix #10201

### DIFF
--- a/autotest/gdrivers/gpkg.py
+++ b/autotest/gdrivers/gpkg.py
@@ -4298,3 +4298,53 @@ def test_gpkg_gti_gpkg_ext(tmp_vsimem):
     ds = gdal.Open(filename)
     assert ds.GetDriver().ShortName == "GPKG"
     assert ds.GetRasterBand(1).Checksum() == 4672
+
+
+###############################################################################
+# Test rename a raster table with SQL
+
+
+@pytest.mark.parametrize("data_type", [gdal.GDT_Byte, gdal.GDT_UInt16])
+def test_gpkg_rename_raster_table(data_type, tmp_vsimem):
+
+    test_layer_path = str(tmp_vsimem / "test_gpkg_rename_raster_table.gpkg")
+
+    if data_type == gdal.GDT_UInt16:
+        src_ds = gdal.Open("data/int16.tif")
+    else:
+        src_ds = gdal.Open("data/small_world.tif")
+
+    ds = gdaltest.gpkg_dr.CreateCopy(
+        test_layer_path,
+        src_ds,
+        options=[
+            "TILE_FORMAT=PNG",
+            "RASTER_TABLE=weird'layer\"name",
+        ],
+    )
+    ds = None
+    src_ds = None
+
+    ds = gdal.OpenEx(test_layer_path, gdal.OF_RASTER | gdal.OF_UPDATE)
+    # Get layer name
+    layer_name = ds.GetMetadataItem("IDENTIFIER")
+    assert layer_name == "weird'layer\"name"
+
+    checksum = ds.GetRasterBand(1).Checksum()
+
+    ds.ExecuteSQL('ALTER TABLE "weird\'layer""name" RENAME TO bar')
+    ds.ExecuteSQL("VACUUM")
+    ds = None
+
+    ds = gdal.Open(test_layer_path, gdal.OF_RASTER)
+    layer_name = ds.GetMetadataItem("IDENTIFIER")
+    assert layer_name == "bar"
+    assert ds.GetRasterBand(1).Checksum() == checksum
+    ds = None
+
+    # Check that there is no more any reference to the layer
+    f = gdal.VSIFOpenL(test_layer_path, "rb")
+    content = gdal.VSIFReadL(1, 1000000, f).decode("latin1")
+    gdal.VSIFCloseL(f)
+
+    assert "weird" not in content
--- a/ogr/ogrsf_frmts/gpkg/ogrgeopackagedatasource.cpp
+++ b/ogr/ogrsf_frmts/gpkg/ogrgeopackagedatasource.cpp
@@ -5613,97 +5613,8 @@ int GDALGeoPackageDataset::Create(const char *pszFilename, int nXSize,
 
         if (bCreateTriggers)
         {
-            /* From D.5. sample_tile_pyramid Table 43. tiles table Trigger
-             * Definition SQL  */
-            pszSQL = sqlite3_mprintf(
-                "CREATE TRIGGER \"%w_zoom_insert\" "
-                "BEFORE INSERT ON \"%w\" "
-                "FOR EACH ROW BEGIN "
-                "SELECT RAISE(ABORT, 'insert on table ''%q'' violates "
-                "constraint: zoom_level not specified for table in "
-                "gpkg_tile_matrix') "
-                "WHERE NOT (NEW.zoom_level IN (SELECT zoom_level FROM "
-                "gpkg_tile_matrix WHERE lower(table_name) = lower('%q'))) ; "
-                "END; "
-                "CREATE TRIGGER \"%w_zoom_update\" "
-                "BEFORE UPDATE OF zoom_level ON \"%w\" "
-                "FOR EACH ROW BEGIN "
-                "SELECT RAISE(ABORT, 'update on table ''%q'' violates "
-                "constraint: zoom_level not specified for table in "
-                "gpkg_tile_matrix') "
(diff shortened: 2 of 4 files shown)


ogr/ogrsf_frmts/gpkg/ogrgeopackagedatasource.cpp
```diff
@@ -7324,6 +7324,158 @@ bool GDALGeoPackageDataset::DeleteVectorOrRasterLayer(const char *pszLayerName)
...
+    }
+
+    if (HasDataColumnsTable())
+    {
+        osSQL += sqlite3_mprintf(
+            "UPDATE gpkg_data_columns SET table_name = '%q' WHERE "
+            "lower(table_name) = lower('%q');",
+            pszNewLayerName, pszLayerName);
+    }
+
```
rouault, line 7420: if HasGriddedCoverageAncillaryTable(), you should also update the tile_matrix_set_name column of gpkg_2d_gridded_coverage_ancillary, and  tpudt_name of gpkg_2d_gridded_tile_ancillary

autotest/gdrivers/gpkg.py
```diff
@@ -4298,3 +4298,35 @@ def test_gpkg_gti_gpkg_ext(tmp_vsimem):
     ds = gdal.Open(filename)
     assert ds.GetDriver().ShortName == "GPKG"
     assert ds.GetRasterBand(1).Checksum() == 4672
+
+
+###############################################################################
+# Test rename a raster table with SQL
+
+
+def test_gpkg_rename_raster_table(tmp_vsimem):
```
rouault, line 4307: you could parametrize with input data types gdal.GDT_Byte to test for the base 'tiles' case and gdal.GDT_UInt16 to test the '2d-gridded-coverage' case

autotest/gdrivers/gpkg.py
```diff
@@ -4298,3 +4298,35 @@ def test_gpkg_gti_gpkg_ext(tmp_vsimem):
...
+
+    ds = gdal.OpenEx(test_layer_path, gdal.OF_RASTER | gdal.OF_UPDATE)
+    # Get layer name
+    layer_name = ds.GetMetadataItem("IDENTIFIER")
+    assert layer_name == "foo"
+    ds.ExecuteSQL("ALTER TABLE foo RENAME TO bar")
+    ds = None
+    ds = gdal.Open(test_layer_path, gdal.OF_RASTER)
+    layer_name = ds.GetMetadataItem("IDENTIFIER")
+    assert layer_name == "bar"
```
rouault, line 4331: I would also check the checksum is preserved after renaming, to check that everything went fine.

and it could be relevant to test for the absence of the "foo" string in the raw file (after vacuuming), as done in ogr_gpkg.py:3417-3425

ogr/ogrsf_frmts/gpkg/ogrgeopackagedatasource.cpp
```diff
@@ -7516,15 +7668,13 @@ OGRLayer *GDALGeoPackageDataset::ExecuteSQL(const char *pszSQLCommand,
         {
             const char *pszSrcTableName = papszTokens[2];
             const char *pszDstTableName = papszTokens[5];
-            OGRGeoPackageTableLayer *poSrcLayer =
-                dynamic_cast<OGRGeoPackageTableLayer *>(
-                    GetLayerByName(SQLUnescape(pszSrcTableName)));
-            if (poSrcLayer)
+            if (!RenameVectorOrRasterLayer(pszSrcTableName, pszDstTableName))
```
rouault, line 7671: There might be non-typical use cases where a table would not be recoginized as a vector or raster layers. In those cases, we would want to still forward the ALTER TABLE ... to be executed.

Example:
```
ogr2ogr poly.gpkg autotest/ogr/data/poly.shp
ogrinfo poly.gpkg -sql "create table my_hidden_table(id integer primary key)"
ogrinfo poly.gpkg  -oo LIST_ALL_TABLES=NO -sql "alter table my_hidden_table rename to my_hidden_table_renamed"
```
We need a new test case for that

ogr/ogrsf_frmts/gpkg/ogrgeopackagedatasource.cpp
elpaso: done

autotest/gdrivers/gpkg.py
elpaso: done

ogr/ogrsf_frmts/gpkg/ogrgeopackagedatasource.cpp
elpaso: done

autotest/gdrivers/gpkg.py
elpaso: done and test added

ogr/ogrsf_frmts/gpkg/ogrgeopackagedatasource.cpp
```diff
@@ -7324,6 +7324,273 @@ bool GDALGeoPackageDataset::DeleteVectorOrRasterLayer(const char *pszLayerName)
...
+    // Temporary remove foreign key checks
+    const GPKGTemporaryForeignKeyCheckDisabler
+        oGPKGTemporaryForeignKeyCheckDisabler(this);
+
+    if (SoftStartTransaction() != OGRERR_NONE)
+    {
+        return false;
+    }
+
+    osSQL = sqlite3_mprintf("UPDATE gpkg_contents SET table_name = '%q' WHERE "
```
rouault, line 7380: all below uses of sqlite3_mprintf() will leak. The return value needs to be freed with sqlite3_free

elpaso: done

ogr/ogrsf_frmts/gpkg/ogrgeopackagedatasource.cpp
```diff
@@ -7324,6 +7324,309 @@ bool GDALGeoPackageDataset::DeleteVectorOrRasterLayer(const char *pszLayerName)
...
+
+        pszSQL = sqlite3_mprintf(
+            "UPDATE gpkgext_relations SET mapping_table_name = '%q' WHERE "
+            "lower(mapping_table_name) = lower('%q');",
+            pszNewLayerName, pszLayerName);
+        osSQL += pszSQL;
+        sqlite3_free(pszSQL);
+    }
+
+    // Drop all triggers for the layer
```
rouault, line 7500: Did you find dropping and recreating triggers needed ? Normally given the SQLite versions we target now, SQLite should normally update existing triggers:

Cf
```
$ sqlite3
SQLite version 3.31.1 2020-01-27 19:55:54
Enter ".help" for usage hints.
Connected to a transient in-memory database.
Use ".open FILENAME" to reopen on a persistent database.
sqlite> create table foo(id);
sqlite> CREATE TRIGGER my_trigger BEFORE INSERT ON foo FOR EACH ROW BEGIN SELECT RAISE(ABORT, 'error'); end;
sqlite> select * from sqlite_master;
table|foo|foo|2|CREATE TABLE foo(id)
trigger|my_trigger|foo|0|CREATE TRIGGER my_trigger BEFORE INSERT ON foo FOR EACH ROW BEGIN SELECT RAISE(ABORT, 'error'); end
sqlite> alter table foo rename to bar;
sqlite> select * from sqlite_master;
table|bar|bar|2|CREATE TABLE "bar"(id)
trigger|my_trigger|bar|0|CREATE TRIGGER my_trigger BEFORE INSERT ON "bar" FOR EACH ROW BEGIN SELECT RAISE(ABORT, 'error'); end
```

elpaso: I've found it necessary only because the name of the trigger itself is not changed, so the test   `assert "weird" not in content` fails at https://github.com/OSGeo/gdal/pull/10213/files#diff-cf1e29b5cee8a8c395aacdaba2a8b883b9f09e8f03beb9b24527ddea1eb7bb46R4350

elpaso: Sure, done.

PR #10215, opened 2024-06-14
GRIB: make .idx reading compatible of /vsisubfile/

### LINKED ISSUE #10214: /vsisubfile reads GRIB2 files inconsistently
### What is the bug?

I have been working with some GRIB2 files recently and discovered that I could use `/vsisubfile` to read individual bands using the published byte ranges for each band. This is super useful because these files have hundreds of bands. It works **great** except for the first band. When I try to read the first band, GDAL interprets it as a multi-band dataset instead of a single-band dataset like all of the other bands.

You can run `gdalinfo` on an entire GRIB2 file like this:
```sh
gdalinfo /vsicurl/https://noaahrrr.blob.core.windows.net/hrrr/hrrr.20240510/conus/hrrr.t12z.wrfsfcf02.grib2
```
This properly describes all of the bands!

You can run `gdalinfo` on a slice in the middle of the file like this:
```sh
gdalinfo /vsisubfile/345203_103748,/vsicurl/https://noaahrrr.blob.core.windows.net/hrrr/hrrr.20240510/conus/hrrr.t12z.wrfsfcf02.grib2
```

When you run `gdalinfo` on the first slice (start byte = 0), GDAL reads it as a dataset with the full number of bands but only loads data for the first (correct) band:
```sh
gdalinfo /vsisubfile/0_345203,/vsicurl/https://noaahrrr.blob.core.windows.net/hrrr/hrrr.20240510/conus/hrrr.t12z.wrfsfcf02.grib2
```
The result can be loaded and used, but the structure is not what you expect from the operation.

I don't know a lot about the internal structure of the GRIB files but I assume that this is happening because the full dataset metadata is getting loaded somewhere around byte 0.

### Versions and provenance

Linux Ubuntu 24.04
GDAL 3.8.4, released 2024/02/08

### Additional context

_No response_

rouault:
Fixes #10214

### DIFF
--- a/frmts/grib/gribdataset.cpp
+++ b/frmts/grib/gribdataset.cpp
@@ -922,11 +922,16 @@ char **GRIBRasterBand::GetMetadata(const char *pszDomain)
 const char *GRIBRasterBand::GetMetadataItem(const char *pszName,
                                             const char *pszDomain)
 {
-    FindMetaData();
-    if (m_nGribVersion == 2 &&
-        CPLTestBool(CPLGetConfigOption("GRIB_PDS_ALL_BANDS", "ON")))
+    if (!((!pszDomain || pszDomain[0] == 0) &&
+          (EQUAL(pszName, "STATISTICS_MINIMUM") ||
+           EQUAL(pszName, "STATISTICS_MAXIMUM"))))
     {
-        FindPDSTemplateGRIB2();
+        FindMetaData();
+        if (m_nGribVersion == 2 &&
+            CPLTestBool(CPLGetConfigOption("GRIB_PDS_ALL_BANDS", "ON")))
+        {
+            FindPDSTemplateGRIB2();
+        }
     }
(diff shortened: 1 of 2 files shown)


frmts/grib/gribdataset.cpp
```diff
@@ -1309,22 +1321,42 @@ GRIBDataset::Inventory(VSILFILE *fp, GDALOpenInfo *poOpenInfo)
     std::unique_ptr<gdal::grib::InventoryWrapper> pInventories;
 
     VSIFSeekL(fp, 0, SEEK_SET);
-    CPLString sSideCarFilename = CPLString(poOpenInfo->pszFilename) + ".idx";
+    std::string osSideCarFileame(poOpenInfo->pszFilename);
```
elpaso, line 1324:
```suggestion
    std::string osSideCarFilename(poOpenInfo->pszFilename);
```

rouault: fixed

PR #10224, opened 2024-06-16
CI Windows: _DISABLE_CONSTEXPR_MUTEX_CONSTRUCTOR again

rouault:
Apply 5a9ef441292b5a92ab58fe54a9fcdab5b1e58bdd again as we still get failures...


rouault:
still doesn't work... sigh...

PR #10228, opened 2024-06-17
Viewshed: fix issue 10205

abellgithub:
Fix visibility of some cells. Add tests to verify base computations and issue.

### DIFF
--- a/autotest/cpp/test_viewshed.cpp
+++ b/autotest/cpp/test_viewshed.cpp
@@ -1,0 +1,188 @@
+///////////////////////////////////////////////////////////////////////////////
+//
+// Project:  C++ Test Suite for GDAL/OGR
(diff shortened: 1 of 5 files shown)


autotest/cpp/test_viewshed.cpp
```diff
@@ -0,0 +1,190 @@
+///////////////////////////////////////////////////////////////////////////////
+//
+// Project:  C++ Test Suite for GDAL/OGR
+// Purpose:  Test general OGR features.
+// Author:   Mateusz Loskot <mateusz@loskot.net>
+//
+///////////////////////////////////////////////////////////////////////////////
+// Copyright (c) 2006, Mateusz Loskot <mateusz@loskot.net>
```
rouault, line 8: you should change that to your name

PR #10230, opened 2024-06-18
Python bindings: Avoid crash when using orphaned subgeometry

dbaston:
Fixes https://github.com/OSGeo/gdal/issues/9920

Issue #9920, opened 2024-05-14
Python bindings: Segfault when accessing sub-geometry after parent has been destroyed


dbaston:
### What is the bug?

Accessing a sub-geometry after its parent has been destroyed causes a segfault. The sub-geometry should either keep a strong reference on its parent (may lead to unexpected behavior somewhere?) or the parent should invalidate references to its children on destruction.

### Steps to reproduce the issue

```
g = ogr.CreateGeometryFromWkt("Point (0 0)")
q = g.Buffer(1).GetGeometryRef(0)
q.GetPointCount() # segfault
```

### Versions and provenance

master

rouault:
I don't see a particular issue in trying to keep alive the parent in GetGeometryRef().

Not sure if you're aware of it, but ther are complications regarding life-time of geometries with RemoveGeometry(). But that's a more advanced/rare use case than the one of ``g.Buffer(1).GetGeometryRef(0)``, so it could be left aside

```python
>>> from osgeo import ogr
>>> g = ogr.CreateGeometryFromWkt("MultiPoint ((0 0))")
>>> point = g.GetGeometryRef(0)
>>> g.RemoveGeometry(0)
>>> point.ExportToWkt()
Segmentation fault (core dumped)
```

PR #10231, opened 2024-06-18
Python bindings: Make Geometry.RemoveGeometry invalidate references

dbaston:
## What does this PR do?

Continuation of #10230 to address https://github.com/OSGeo/gdal/issues/9920#issuecomment-2111025554

This fix is not without cost and I'm not sure it should be applied. It increases the runtime of 

```
python3 -m timeit -r 5 -s 'from osgeo import ogr; g=ogr.CreateGeometryFromWkt("MULTIPOINT ((0 0))")' 'q=g.GetGeometryRef(0)'
```
from 627 ns to 1490 ns.

#9920

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/autotest/ogr/ogr_geom.py
+++ b/autotest/ogr/ogr_geom.py
@@ -4519,3 +4519,13 @@ def test_ogr_geom_buffer_with_args():
 
     with pytest.raises(Exception, match="Unsupported buffer option"):
         geom.Buffer(1, {"QUALITY": "HIGH"})
+
+
+def test_ogr_subgeom_use_after_parent_free():
+
+    g = ogr.CreateGeometryFromWkt("POLYGON ((0 0, 1 0, 1 1, 0 0))")
+
+    exterior_ring = g.GetGeometryRef(0)
+    del g
+
+    assert exterior_ring.GetPointCount() > 0  # does not crash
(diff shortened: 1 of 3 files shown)


dbaston:
This check can still be circumvented, e.g.:

```
    defn = ogr.FeatureDefn()
    f = ogr.Feature(defn)
    f.SetGeometryDirectly(ogr.CreateGeometryFromWkt("MultiPoint ((0 0))"))

    g1 = f.GetGeometryRef()
    g2 = f.GetGeometryRef()
    point = g2.GetGeometryRef(0)
    g1.RemoveGeometry(0)

    with pytest.raises(Exception):
        point.ExportToWkt()  # crashes
```

rouault:
> This check can still be circumvented

I'd say your call if you think this PR is worthwhile or not

I'm thinking that an alternate would be to propose a GetGeometry(int) that would do GetGeometyRef(...).Clone(), and advertize using it if the parent geometry is going to be modified

dbaston:
> I'd say your call if you think this PR is worthwhile or not

As much as I feel like Python should never crash, this seems like an incomplete solution to an uncommon problem. I'd say the juice isn't worth the squeeze.

PR #10237, opened 2024-06-18
Viewshed: address potential issues with line-based interpolation

### LINKED ISSUE #10205: Viewshed: Some cells not properly calculated
### What is the bug?

The viewshed code doesn't calculate the height or visibility correctly along lines at angles n * (pi / 4) from the observer except those on the same raster line as the observer (0 and pi).

Found during code inspection and writing new tests. The logic of the viewshed algorithm should force all eight cells surrounding the observer to be visible, but this isn't always the case. This error may also propagate into the calculation for other cells.

### Versions and provenance

All

### Additional context

_No response_

abellgithub:
Add tests that have been verified by hand-computation.
Correct existing tests to match fixed visibility/height computations along lines (multiples of pi / 4)
Close #10205


autotest/cpp/test_viewshed.cpp
```diff
@@ -0,0 +1,259 @@
+///////////////////////////////////////////////////////////////////////////////
+//
+// Project:  C++ Test Suite for GDAL/OGR
+// Purpose:  Test general OGR features.
```
rouault, line 4:
```suggestion
// Purpose:  Test viewshed algorithm
```

Issue #10242, opened 2024-06-19
make install puts python packages in the wrong directory


strk:
### What is the bug?

after `make install` this fails:
> python -c 'import osgeo'

with this error:
> Traceback (most recent call last):
>  File "<string>", line 1, in <module>
> ModuleNotFoundError: No module named 'osgeo'

### Steps to reproduce the issue

in a `Debian GNU/Linux 12 (bookworm)` system build and install gdal , then run `python -c 'import osgeo'`

### Versions and provenance

Problem was found in release/3.6 branch

### Additional context

The default python system path does not include `/usr/local/lib/python3/dist-packages` but GDAL installs in there.

lnicola:
This seems semi-intentional, since https://github.com/OSGeo/gdal/pull/4517 added the Python version to that path, but only for non-Debian distros.

rouault:
@strk can you give https://github.com/OSGeo/gdal/pull/10243 a try ?

(Debian is particularly annoying with its patching of default Python paths)

strk:
I've updated the bug description because the default path does not indeed include /usr/local at all

rouault:
> I've updated the bug description because the default path does not indeed include /usr/local at all

the first time you run cmake, if you don't specify CMAKE_INSTALL_PREFIX, it will be set as a cached variable to ``/usr/local`` in CMakeLists.txt

strk:
(not) funny enough, `PYTHONPATH= python -c 'import sys; print(sys.path)'` does include `/usr/local/lib/python3.11/dist-packages` now that gdal (with the patch in PR #10243) installs something in there.

PR #10243, opened 2024-06-19
Python bindings: fix/workaround installation into /usr/local/lib/python{version}/dist-packages when prefix is /usr/local

rouault:
Fixes #10242

PR #10244, opened 2024-06-19
WIP: Fix python install on Debian bookworm

strk:
Trying to fix  python install path on Debian.

See #10242


strk:
Leaving this to @rouault in #10243

PR #10258, opened 2024-06-20
OGR layer algebra: honour PROMOTE_TO_MULTI=YES for Points

rouault:
Fixes https://lists.osgeo.org/pipermail/gdal-dev/2024-June/059103.html

### DIFF
--- a/autotest/ogr/ogr_layer_algebra.py
+++ b/autotest/ogr/ogr_layer_algebra.py
@@ -261,6 +261,31 @@ def test_algebra_intersection_3(D1, D2, C):
     assert is_same(D1, C), "D1 != C"
 
 
+def test_algebra_intersection_multipoint():
+
+    driver = ogr.GetDriverByName("MEMORY")
+    ds = driver.CreateDataSource("ds")
+    layer1 = ds.CreateLayer("layer1")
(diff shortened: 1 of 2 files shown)


jratike80:
Does promote_to_multi now change all points into multipoints? A shapefile has different shapetype for points and multipoints and I am wondering if that can make 
confusion. Multipoint is also a rather uncommon datatype in real world data and it may  be that some software and algorithms will have troubles with multipoints, even if they have only one member. And with constrained datastores (shape, PostGIS, GeoPackage) the digitizing tools should be prepared to create multipoints.

Based on my experience and memory users have not had much problems because points have not been turned into multipoints. I am not sure if this change is for good or bad.

rouault:
> Does promote_to_multi now change all points into multipoints?

yes

jratike80:
Well, maybe the new behaviour can give unintentional results only if
 a) user has a habit to write PROMOTE_TO_MULTI always, even when working with point data when the argument has not been doing anything, or 
 b) user works with source data that contains mixed geometries, including points, or
 c) if SQL query or layer algebra creates points and user did not think that it can happen.

Maybe there is no need to be worried.

PR #10261, opened 2024-06-21
DXF: add a DXF_CLOSED_LINE_AS_POLYGON=YES/NO configuration option

### LINKED ISSUE #10153: Polygon in dxf, after ogr2ogr, got lingstring in geojson
### What is the bug?

[dxf.zip](https://github.com/user-attachments/files/15730961/dxf.zip)
There are two polygon shapes in the dxf attachment. The shape with more points calls shapeA, the other one calls shapeB.
Got linestrings after converting.

Test a. With command "ogr2ogr -f GeoJSON test.json test.dxf -nlt POLYGON", shapeB output as polygon, shapeA output as linestring. The last point of the linestring has a different y from the first point.
Test b. using nuget gdal in c#, after getting geojson driver, creating data source, opening dxf data source, copying layer from the dxf, i got two linestrings. creating multilinestring geometries to containing each linestring, using function polygonize, shapeA got null and shapeB got an empty geometry collection

### Versions and provenance

nuget GDAL-3.8.5
binary 3.6.2

### Additional context

_No response_

rouault:
to control whether closed POLYLINE/LWPOLYLINE should be exposed as a polygon. Defaults to NO to match current behavior.

Fixes #10153

Also available in DWF driver as DWG_CLOSED_LINE_AS_POLYGON=YES/NO

Also, for DXF, expose existing reader configuration options as open options: INLINE_BLOCKS, MERGE_BLOCK_GEOMETRIES, TRANSLATE_ESCAPE_SEQUENCES, INCLUDE_RAW_CODE_VALUES, 3D_EXTENSIBLE_MODE, CLOSED_LINE_AS_POLYGON, HATCH_TOLERANCE ENCODING

CC @atlight  It was not obvious to me what the default of the CLOSED_LINE_AS_POLYGON should be. I see some software like KLayout have even more complicated heuristics/settigs for that policy. See "Polygon formation and LINE/POLYLINE interpretation"  at https://www.klayout.de/dxf_format.html


hanfangyu:
@rouault How to get the 3.10 jar package? We currently need to use the new features, and look forward to your reply. Thanks.

rouault:
> I encountered the same problem and needed a 3.10 java jar package to solve it.

build master from source. Cf https://gdal.org/en/latest/development/building_from_source.html#

PR #10264, opened 2024-06-21
Viewshed: support observers to left and right of input raster

abellgithub:
This adds support for observers to the left and right of the raster. Observers above and below the raster are not supported and providing such a location for an observer generates an error. It assumes that the space outside of the raster is empty (at a height below all cells in the raster).

Tests are included.


alg/viewshed.h
```diff
@@ -87,18 +87,52 @@ class Viewshed
...
+        {
+            return containsX(nX) && containsY(nY);
+        }
+
+        /// \brief  Clamp the argument to be in the window in the X dimension.
+        /// \param  nX  Value to clamp.
+        /// \return  Clamped value.
+        int clampX(int nX) const
+        {
+            return xSize() ? std::clamp(nX, xStart, xStop - 1) : xStart;
```
rouault, line 131: ``#include <algorithm>`` needed for std::clamp

alg/viewshed.cpp
```diff
@@ -280,12 +282,19 @@ bool Viewshed::calcOutputExtent(int nX, int nY)
...
     {
         CPLError(CE_Failure, CPLE_AppDefined,
-                 "The observer location falls outside of the DEM area");
+                 "Observer position above or below the raster "
+                 "not currently supported");
         return false;
     }
+    if (!oOutExtent.contains(nX, nY))
+    {
+        CPLError(CE_Warning, CPLE_AppDefined,
```
rouault, line 294: do we need to issue a warning if this is nominally supported ?

alg/viewshed.cpp
```diff
@@ -280,12 +282,19 @@ bool Viewshed::calcOutputExtent(int nX, int nY)
...
+                 "NOTE: The observer location falls outside of the DEM area");
+        //ABELL - Make sure observer Z is specified.
```
rouault, line 296: some check to add here ?

alg/viewshed.cpp
```diff
@@ -547,20 +575,95 @@ double doMax(int nXOffset, int nYOffset, double dfThisPrev, double dfLast,
...
 
-double doLine(int nXOffset, [[maybe_unused]] int nYOffset, double dfThisPrev,
-              [[maybe_unused]] double dfLast,
-              [[maybe_unused]] double dfLastPrev)
+}  // unnamed namespace
+
+/// Process the part of the first line to the left of the observer.
+///
+/// @param nX  X coordinate of the observer.
+/// @param iStart  X coordinate of the first cell to the left of the observer to be procssed.
```
rouault, line 583:
```suggestion
/// @param iStart  X coordinate of the first cell to the left of the observer to be processed.
```

alg/viewshed.cpp
```diff
@@ -547,20 +575,95 @@ double doMax(int nXOffset, int nYOffset, double dfThisPrev, double dfLast,
...
+    }
+    // For cells outside of the [start, end) range, set the outOfRange value.
+    std::fill(vResult.begin(), vResult.begin() + iEnd + 1, oOpts.outOfRangeVal);
 }
 
-}  // unnamed namespace
+/// Process the part of the first line to the right of the observer.
+///
+/// @param nX  X coordinate of the observer.
+/// @param iStart  X coordinate of the first cell to the right of the observer to be procssed.
```
rouault, line 624:
```suggestion
/// @param iStart  X coordinate of the first cell to the right of the observer to be processed.
```

autotest/pymod/test_cli_utilities.py
```diff
@@ -29,59 +29,14 @@
 ###############################################################################
 
 import os
-import sys
```
rouault, line 32: it would be good if those changes to test_cli_utilities.py were put in a dedicated PR

abellgithub: I didn't realize those were in there until after I pressed the button. Are you OK with them as they are?

alg/viewshed.cpp
abellgithub: I'm not sure. Normally the Z is zero by default and perhaps this is still fine.

alg/viewshed.cpp
abellgithub: I think so since it's a change in behavior from what exists. I don't think it hurts anything. It could be removed later.

autotest/pymod/test_cli_utilities.py
rouault: Actually, I'm afraid that's going to break Windows testing, because with MSVC builds, binaries go to $build_dir/apps/$config_name where config_name=Release, Debug, RelWithDebInfo, etc.
cmake/helpers/GdalSetRuntimeEnv.cmake tests care of correctly appending that directory in the PATH.

rouault:
For CI to trigger, you also need to rebase on top of latest master to fix a conflict (likely related to 082d0e78c54dd544a14947ac216fdcca8080595c)

autotest/pymod/test_cli_utilities.py
abellgithub: Is there some reason not to drive all this from cmake and avoid all of this? cmake knows where it stuck the binaries.

rouault: it is convenient to be able to run a single test file locally with pytest, without going through a whole test suite

abellgithub:
@rouault : I think issues have been addressed here.

autotest/cpp/test_viewshed.cpp
```diff
@@ -62,21 +62,19 @@ Viewshed::Options stdOptions(const Coord &observer)
     return stdOptions(observer.first, observer.second);
 }
 
-DatasetPtr runViewshed(int8_t *in, int edgeLength,
-                       const Viewshed::Options &opts)
+DatasetPtr runViewshed(const int8_t *in, int xlen, int ylen,
+                       Viewshed::Options opts)
```
rouault, line 66: any reason to have back out the change I did in master? (Coverity Scan complains when too large structures are passed by values)
```suggestion
                       const Viewshed::Options &opts)
```

abellgithub: No. I didn't realize you had made changes.

PR #10284, opened 2024-06-23
[Lint] OSM: more usage of std::vector, std::unique_ptr, and C++ cast operators

rouault:


ogr/ogrsf_frmts/osm/ogrosmdatasource.cpp
```diff
@@ -205,13 +205,14 @@ static CPLString GetInterestLayersForDSName(const CPLString &osDSName)
 {
     CPLMutexHolder oMutexHolder(&hMutex);
     GIntBig nPID = CPLGetPID();
-    for (int i = 0; i < (int)oListDSToBeOpened.size(); i++)
+    for (auto oIter = oListDSToBeOpened.begin();
```
elpaso, line 208: can you use a Range-based for loop here?  (since you are modernizing the code)

elpaso (approved):

ogr/ogrsf_frmts/osm/ogrosmdatasource.cpp
rouault: actually no, because in the loop, we user the iterator as the argument of the std::vector::erase() method

PR #10292, opened 2024-06-24
ogr2ogr: error out if WKT argument of -clipsrc/-clipdst is an invalid geometry

### LINKED ISSUE #10289: ogr2ogr -clipsrc/-clipdst: TopologyException: side location conflict: perhaps should be fatal
### Feature description

It seems this should be a fatal error. The output file made was a mess, at least in my case.
```
$ ogr2ogr m.kmz l.kmz -clipsrc 'POLYGON ((120.79382 24.20545 ...
ERROR 1: TopologyException: side location conflict at
  120.7946631515733 24.205405693211443.
  This can occur if the input geometry is invalid.
$ echo $?
0
```
To reproduce use some POLYGON that is really a figure 8, etc.
Version 3.9.

rouault:
If specifying -makevalid, MakeValid() will be run on it to make it valid.

For consistency, change behavior of -clipsrc/-clipdst {datasetname} to not automatically make valid invalid clip geometries, but do it only if -makevalid is specified.

Fixes #10289

### DIFF
--- a/apps/ogr2ogr_lib.cpp
+++ b/apps/ogr2ogr_lib.cpp
@@ -616,7 +616,8 @@ static OGRLayer *GetLayerAndOverwriteIfNecessary(GDALDataset *poDstDS,
 static std::unique_ptr<OGRGeometry> LoadGeometry(const std::string &osDS,
                                                  const std::string &osSQL,
                                                  const std::string &osLyr,
-                                                 const std::string &osWhere)
+                                                 const std::string &osWhere,
+                                                 bool bMakeValid)
 {
     auto poDS = std::unique_ptr<GDALDataset>(
         GDALDataset::Open(osDS.c_str(), GDAL_OF_VECTOR));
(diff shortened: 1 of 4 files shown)


apps/ogr2ogr_lib.cpp
```diff
@@ -2328,19 +2340,58 @@ GDALDatasetH GDALVectorTranslate(const char *pszDest, GDALDatasetH hDstDS,
             *pbUsageError = TRUE;
         return nullptr;
     }
+    if (psOptions->poClipSrc && !psOptions->poClipSrc->IsValid())
+    {
+        if (!psOptions->bMakeValid)
+        {
+            CPLError(CE_Failure, CPLE_IllegalArg,
+                     "-clipsrc geometry is invalid. You can try to make it "
+                     "valid with -makevalid");
```
dbaston, line 2349: The results of MakeValid can be surprising.

Do you think it makes sense to either extend this message along the lines of

> -clipsrc geometry is invalid. You can try to make it valid with -makevalid. The results should be manually inspected.

or instead have a separate warning when MakeValid is actually used?

> The -clipsrc geometry is invalid and was corrected with MakeValid. The results of this operation should be inspected.

rouault: > The results should be manually inspected.

Not sure if that would be very helpful, as the user has no easy way to know how the geometry was corrected by MakeValid()

dbaston: Maybe more clear to say "the output of ogr2ogr should be manually inspected."

rouault: ok, warning/error messages tuned

dbaston (approved):

PR #8133, opened 2023-07-21
JP2Grok: add new driver for JPEG 2000 format using Grok library

boxerab:
## What does this PR do?

This PR adds a new driver for the JPEG 2000 format using [Grok](https://github.com/GrokImageCompression/grok) library.

### Design Strategy

The existing `JP2OpenJPEG` driver was copied and refactored to isolate the OpenJPEG-specific bits
in a header file `opj_context` while keeping most of the generic logic intact. Then, a similar header file
`grk_context` was created with Grok-specific bits. So, this refactor could be used for both open source JPEG 2000
drivers.

### Testing

Basic performance testing of `gdal_translate` with default settings on a large HiRISE image shows this driver outperforming the existing open source driver by 6 times.

## What are related issues/pull requests?

https://github.com/OSGeo/gdal/pull/3449
https://github.com/OSGeo/gdal/pull/5117

## Tasklist

 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Updated Python API documentation (swig/include/python/docs/)
 - [ ] Review
 - [ ] Adjust for comments
 - [ ] All CI builds and checks have passed

PR #10294, opened 2024-06-25
JP2Grok: new Grok JPEG 2000 driver based on Grok JPEG 2000 Library

boxerab:
Let's try this again !

## What does this PR do?

This PR adds a new driver for the JPEG 2000 format using [Grok](https://github.com/GrokImageCompression/grok) library.

### Design Strategy

In a previous merged PR, common features supporting both OpenJPEG and Grok libraries were refactored into a template driver named `JP2DatasetBase`. This PR creates a new Grok driver inheriting from `JP2DatasetBase`.

### Performance

Basic performance testing of `gdal_translate` with default settings on a large HiRISE image shows this driver outperforming the existing open source driver by 6 times.

https://github.com/OSGeo/gdal/pull/3449
https://github.com/OSGeo/gdal/pull/5117
https://github.com/OSGeo/gdal/pull/8133

## Tasks

 - [ ] Support encode (currently disabled)
 - [ ] Address issues raised by @rouault in previous Grok PRs
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/doc/source/drivers/raster/jp2grok.rst
+++ b/doc/source/drivers/raster/jp2grok.rst
@@ -1,0 +1,552 @@
+
+.. _raster.JP2Grok:
+
+================================================================================
+JP2Grok -- JPEG2000 driver based on Grok library
+================================================================================
+
+.. shortname:: JP2Grok
+
+.. build_dependencies:: grok >= 9.6
+
+
+This driver is an implementation of a JPEG2000 reader/writer based on
+Grok library **v2**.
+
+The driver uses the VSI Virtual File API, so it can read JPEG2000
+compressed NITF files.
+
+XMP metadata can be extracted from JPEG2000 files, and will be stored as
+XML raw content in the xml:XMP metadata domain.
+
+The driver supports writing georeferencing information as GeoJP2 and
+GMLJP2 boxes.
+
+The driver supports creating files with
+transparency, arbitrary band count, and adding/reading metadata. Update
+of georeferencing or metadata of existing file is also supported.
+Optional intellectual property metadata can be read/written in the
+xml:IPR box.
+
+Driver capabilities
+-------------------
+
+.. supports_createcopy::
+
+.. supports_georeferencing::
+
+.. supports_virtualio::
+
+Georeferencing
+--------------
+
+Georeferencing information can come from different sources : internal
+(GeoJP2 or GMLJP2 boxes), worldfile .j2w/.wld sidecar files, or PAM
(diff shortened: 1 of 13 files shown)


doc/source/drivers/raster/jp2grok.rst
```diff
@@ -0,0 +1,552 @@
+
+.. _raster.JP2Grok:
```
rouault, line 2:
```suggestion
.. _raster.jp2grok:
```

doc/source/drivers/raster/jp2grok.rst
```diff
@@ -0,0 +1,552 @@
...
+JP2Grok -- JPEG2000 driver based on Grok library
+================================================================================
+
+.. shortname:: JP2Grok
+
+.. build_dependencies:: grok >= 9.6
+
+
+This driver is an implementation of a JPEG2000 reader/writer based on
+Grok library **v2**.
```
rouault, line 14:
```suggestion
Grok library.
```

rouault:
@boxerab Just looking at https://github.com/GrokImageCompression/grok/commits/master/ , I'm puzzled to see that Grok git history has been completely reset, and that the released packages are only the latest one. This isn't super re-assuring for audit purposes, especially after the XZ story.

rouault:
> Anyone is free to audit the code

good luck when you have zero history... Was there something in the history that wasn't appropriate and had to be removed? But in that case there are less drastic ways of removing a commit than just erasing the whole history, which is an important asset of open source projects. I'm not the only one to find that there's something fishy here: https://mastodon.social/@EvenRouault/112687807835976568 . I can't name a project that has reset their history like that. This is an important asset to understand the rationale of parts of the code and having an audit trail. My 2 cents if there's nothing to hide and you've kept a copy of your initial git repo before the big reset is that you re-push it, and re apply the later modifications on top of if

jratike80:
> As many continue to benefit from the project, it continues on its current path.

That applies to GDAL as well. We need to consider if and how out paths will meet.

rouault:
> I will note that GDAL has drivers for completely opaque black box software such as ERDAS ECW/JP2 SDK

indeed, that's what we mentioned during yesterday's monthly GDAL maintainer meeting. Given that we won't ship Grok itself as part of deliverables of the GDAL project, this isn't a blocking point for us to consider including the source code of the GDAL driver part. This is the responsibility of the end user to decide whether it fits their criteria

Issue #10296, opened 2024-06-25
ogrinfo: input file should be optional


jidanni:
### Feature description

Here the user is forced to include an irrelevant input file, when all the user is trying to do is get the result of a fully self-contained SELECT statement.
```
$ ogrinfo a.csv -dialect SQLITE -sql \
"SELECT ST_Buffer(ST_GeomFromText('POLYGON ((120.81313 24.18337...
```
So perhaps `ogrinfo` should simply have input files be optional.

(Or, on the man page have a "how to fake the input file" section.
("Just create a file.csv with the following lines.") Yes, simple, but ugly.)

rouault:
You can use the ":memory:" input file which is a in-memory empty SQLite file

ogrinfo :memory: -sql "SELECT ST_Buffer(ST_GeomFromText('POINT(0 0)'), 1)"

jratike80:
Why it does not work in this order:

```
ogrinfo -sql "SELECT ST_Buffer(ST_GeomFromText('POINT(0 0)'), 1)" :memory
ERROR 4: :memory: No such file or directory
```
I have been using on-the-fly generated data a lot when answering users questions. I had not thought about using the SQLite memory db. I was now considering to have some reserved datasource name like "virtualSQL". Somehow I was remembering the dummy "DUAL" table in Oracle https://www.w3resource.com/sql/sql-dual-table.php. I remember that DUAL was quite handy when I used Oracle a long time ago.

rouault:
> Why it does not work in this order:

it does for me. Perhaps you need to surround :memory: with double quotes

jratike80:
The error message, that itself contains colons, just fooled me and I did not notice that the ending colon was missing from `:memory:`. I used `:memory` and the error was
`ERROR 4: :memory: No such file or directory`

jidanni:
Got an idea: If ogrinfo doesn't find an input file, then it should just plug :memory: into itself internally.

PR #10300, opened 2024-06-25
Update ogrinfo.rst with tip how to workaround need for input file

jidanni:
#10296

### DIFF
--- a/doc/source/programs/ogrinfo.rst
+++ b/doc/source/programs/ogrinfo.rst
@@ -529,4 +529,10 @@ Adding a column to an input file:
 
    ogrinfo input.shp -sql "ALTER TABLE input ADD fieldX float"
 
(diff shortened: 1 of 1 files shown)


doc/source/programs/ogrinfo.rst
```diff
@@ -529,4 +529,10 @@ Adding a column to an input file:
 
    ogrinfo input.shp -sql "ALTER TABLE input ADD fieldX float"
 
+Sometimes there is no input file involved in a calculation. In such cases one may
+use the ``:memory:`` input file which is a in-memory empty SQLite file.
```
rouault, line 533:
```suggestion
use the ``:memory:`` input file which is a in-memory empty SQLite file (and the SQLite SQL dialect will be implicitly used).
```

PR #10303, opened 2024-06-25
update links in `generate_leaflet()`

### LINKED ISSUE #10304: OSM Toner doesn't work in gdal2tiles
### What is the bug?

https://github.com/OSGeo/gdal/blob/f160dfd68e518d14149aed751662065515bb065b/swig/python/gdal-utils/osgeo_utils/gdal2tiles.py#L3987-L3988

It seems that the service address has changed and no longer works.

Now the link looks like this: `https://tiles.stadiamaps.com/tiles/stamen_toner/16/10516/25320.png`, but requires registration (https://docs.stadiamaps.com/authentication).

I guess the best solution would be to remove this service?

For example, the following link doesn't work:

```
https://b.tile.stamen.com/toner/7/70/42.png
#> NS_ERROR_UNKNOWN_HOST
```

### Versions and provenance

Windows 11

### Additional context

_No response_

kadyb:
This simple PR updates links in the `generate_leaflet()` from `gdal2tiles`. This also fixes #10304 by removing OSM Toner.


jratike80:
Could there be a need the check the Stamen tiles? See https://gis.stackexchange.com/questions/482777/stamen-basemaps-not-loading-in-open-layers-plugin-anymore and https://docs.stadiamaps.com/guides/migrating-from-stamen-map-tiles/

kadyb:
@jratike80, sorry I didn't notice your comment earlier and created [separate issue](https://github.com/OSGeo/gdal/issues/10304), but yes, they are currently not working.

PR #10320, opened 2024-06-27
ogrinfo: auto-select layer in datasets with only one layer

dbaston:
## What does this PR do?

Avoids the need to specify the layer name in `ogrinfo` calls to single-layer datasets, e.g. 

```
ogrinfo -so tl_2022_us_county.shp
```
instead of
```
ogrinfo -so tl_2022_us_county.shp tl_2022_us_county
```

This matches the behavior when using `ogrinfo -json` or when calling `gdal.VectorInfo` from Python. It is also [described as](https://trac.osgeo.org/gdal/wiki/UserDocs/OgrInfo) "likely what is really wanted" by the author 17 years ago :)

Somewhat mixed feelings about this because forgetting `-al` is my reminder to also set `-so`. But I don't think that's a good reason to keep the current behavior (though there may be one).

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/apps/ogrinfo_lib.cpp
+++ b/apps/ogrinfo_lib.cpp
@@ -2106,7 +2106,7 @@ char *GDALVectorInfo(GDALDatasetH hDataset,
 
                 CPLJSONObject oLayer;
                 oLayerArray.Add(oLayer);
-                if (!psOptions->bAllLayers)
+                if (nLayerCount > 1 && !psOptions->bAllLayers)
                 {
                     if (!bJson)
                         Concat(osRet, psOptions->bStdoutOutput,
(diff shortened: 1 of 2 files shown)


jratike80:
Does the change mean that number of layers is checked first and if the dataset has only one layer, then ogrinfo comes automatically for that layer, and otherwise user gets a list? And all the open options etc are still checked, for example "list_all_tables" https://gdal.org/drivers/vector/gpkg.html#dataset-open-options.

The change may make some issues if users learn to run plain `ogrinfo my.gpkg` or `ogrinfo my.gml` in some odd scripts using GDAL utilities and parsing and piping the response. The script stops working when another layer is added to the dataset.

rouault:
I'm also ambivalent about the appropriateness of the change, which has some niceness, but might make things harder to reason about.
Currently, you have 3 graduation of details about dataset exploration:
- ogrinfo my.gpkg: just list layer names and geometry type
- ogrinfo -so -al my.gpkg: metadata of all layers
- ogrinfo -al my.gpkg: metadata + features
Perhaps the documentation is not obvious / clear enough about that

PR #10322, opened 2024-06-27
Update ogr2ogr.rst about when there is no input file

jidanni:
For https://github.com/OSGeo/gdal/issues/10296 .


jratike80:
I would document this in a wiki page "Interesting use cases".  There coud be even more interesting use cases, like

SELECT ST_HexagonalGrid(ST_GeomFromText('POLYGON ((0 0,0 100, 100 100, 0 0))'),10);

which gives

image

PR #10331, opened 2024-06-27
MRF: Allow open of MRF-in-TAR as MRF

lucianpls:
## What does this PR do?

Allow the use of a tar file containing a single MRF directly as an MRF.

It has always been possible to open an MRF (as any other gdal format) inside a tar file by using a vsitar path.
This change allows a tar file that starts with the MRF metadata file to be directly recognized and opened for reading without having to use the vsitar path. This makes an MRF inside a TAR file act as a single file raster format, making it easier to share. All the MRF files have to be inside the tar, with the main metadata (.mrf) being the first.

The identification of such an MRF-in-TAR candidate is done by directly inspecting the tar header and the beginning of the first file inside the tar, without invoking using /vsitar.

This change does not interfere with the use of any other file inside the tar using the normal /vsitar path. The only difference is that gdalinfo on such a mrf-in-tar file will return the MRF information, as opposed to listing the tar content as it does now.

The last file extension has to be .tar to be correctly opened by vsitar internally, the recommended file name for an MRF-in-TAR will be **filename.mrf.tar**

### DIFF
--- a/autotest/gdrivers/mrf.py
+++ b/autotest/gdrivers/mrf.py
@@ -236,10 +236,7 @@ def test_mrf_zen_test():
 def test_mrf_in_tar():
     import tarfile
     out_ds = gdal.Translate(
-        "plain.mrf",
-        "data/byte.tif",
-        format="MRF",
-        creationOptions=["COMPRESS=DEFLATE"]
+        "plain.mrf", "data/byte.tif", format="MRF", creationOptions=["COMPRESS=DEFLATE"]
     )
     out_ds = None
     with tarfile.TarFile("plain.mrf.tar", "w", format=tarfile.GNU_FORMAT) as tar:
--- a/autotest/gdrivers/mrf.py
+++ b/autotest/gdrivers/mrf.py
@@ -235,10 +235,10 @@ def test_mrf_zen_test():
 
 def test_mrf_in_tar():
     import tarfile
-    out_ds = gdal.Translate(
+
+    gdal.Translate(
         "plain.mrf", "data/byte.tif", format="MRF", creationOptions=["COMPRESS=DEFLATE"]
     )
-    out_ds = None
     with tarfile.TarFile("plain.mrf.tar", "w", format=tarfile.GNU_FORMAT) as tar:
         for ext in ("mrf", "idx", "pzp"):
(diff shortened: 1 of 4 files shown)


rouault (commented):
test ?

autotest/gdrivers/mrf.py
```diff
@@ -233,6 +233,24 @@ def test_mrf_zen_test():
...
+    with tarfile.TarFile("plain.mrf.tar", "w", format=tarfile.GNU_FORMAT) as tar:
+        for ext in ("mrf", "idx", "pzp"):
+            tar.add("plain." + ext)
+    for ext in ("mrf", "idx", "pzp", "mrf.aux.xml"):
+        gdal.Unlink("plain." + ext)
+    ds = gdal.Open("plain.mrf.tar")
+    cs = ds.GetRasterBand(1).Checksum()
+    ds = None
+    assert cs == 4672
+    gdal.Unlink("plain.mrf.tar")
```
rouault, line 251: would be better to write in tmp_path rather than in current directory. Something like:

```suggestion
def test_mrf_in_tar(tmp_path):
    import tarfile

    mrf_filename = str(tmp_path / "plain.mrf")   
    gdal.Translate(
        mrf_filename, "data/byte.tif", format="MRF", creationOptions=["COMPRESS=DEFLATE"]
    )
    
    mrf_tar_filename = str(tmp_path / "plain.mrf.tar")
    with tarfile.TarFile(mrf_tar_filename, "w", format=tarfile.GNU_FORMAT) as tar:
        for ext in ("mrf", "idx", "pzp"):
            tar.add(str(tmp_path / "plain." + ext))
    for ext in ("mrf", "idx", "pzp", "mrf.aux.xml"):
        gdal.Unlink(str(tmp_path / "plain." + ext))
    ds = gdal.Open(mrf_tar_filename)
    cs = ds.GetRasterBand(1).Checksum()
    ds = None
    assert cs == 4672
```

lucianpls: The inside file can't have a path, to make it even less likely to trigger by mistake.
I did change the test to use the tmp_path, it is a better example of how to build such a tar in python

PR #10347, opened 2024-07-01
GDALMDArray::GetResampled(): take into account good_wavelengths array…

rouault:
… for EMIT dataset orthorectification

When the /sensor_band_parameters/good_wavelengths variable is present, use it to decide of the current array along the band dimension are valid.


gcore/gdalmultidim_gltorthorectification.cpp
```diff
@@ -124,6 +127,32 @@ class GLTOrthoRectifiedArray final : public GDALPamMDArray
         newAr->m_poSRS.reset(oSRS.Clone());
         newAr->m_poSRS->SetDataAxisToSRSAxisMapping(
             {/*latIdx = */ 1, /* lonIdx = */ 2});
+        if (CPLTestBool(CSLFetchNameValueDef(papszOptions,
+                                             "USE_BAD_WAVELENGTHS", "YES")) &&
```
elpaso, line 131: Is this spelled correctly? Shouldn't it be 
```suggestion
                                             "USE_GOOD_WAVELENGTHS", "YES")) &&
```
according to https://github.com/OSGeo/gdal/pull/10347/files#diff-47bb821c4b705cacab4b4ad82536126da726d7e5c7dcd9ea04e64c7207131e81R8451

gcore/gdalmultidim_gltorthorectification.cpp
```diff
@@ -124,6 +127,32 @@ class GLTOrthoRectifiedArray final : public GDALPamMDArray
...
+            poParent->GetDimensionCount() == 3)
+        {
+            auto poGoodWaveLengths = poRootGroup->OpenMDArrayFromFullname(
```
elpaso, line 134:
```suggestion
            const auto poGoodWaveLengths = poRootGroup->OpenMDArrayFromFullname(
```

gcore/gdalmultidim_gltorthorectification.cpp
rouault: fixed

gcore/gdalmultidim_gltorthorectification.cpp
rouault: fixed

PR #10350, opened 2024-07-01
docs ogrinfo: add a note to -so option with -json

agiudiceandrea:
## What does this PR do?

Hopefully clarifies that the `-so` "Summary Only" option in `ogrinfo` is "implicit" / "irrilevant" when the `-json` option is provided and that the `-features` option should be used instead.
Not sure about the wording...

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:


elpaso (approved):
LGTM

PR #10370, opened 2024-07-06
WKT geometry importer: accept (non conformant) PointZ/PointZM without space as generated by current QGIS versions

rouault:
Cf thread at https://lists.osgeo.org/pipermail/qgis-developer/2024-July/066906.html

### DIFF
--- a/autotest/ogr/ogr_wkbwkt_geom.py
+++ b/autotest/ogr/ogr_wkbwkt_geom.py
@@ -373,76 +373,79 @@ def test_ogr_wkbwkt_test_broken_geom():
 # Test importing WKT SF1.2
 
 
-def test_ogr_wkbwkt_test_import_wkt_sf12():
-
-    list_wkt_tuples = [
+@pytest.mark.parametrize(
+    "input_wkt,expected_output_wkt",
+    [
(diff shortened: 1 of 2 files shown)


jgrocha:
@rouault Thanks for the PR.

Just a quick question: why GDAL does not use the 'Z' on the WKT output?

Example:
```python
from osgeo import ogr

a = ogr.CreateGeometryFromWkt("POINT Z (0 0 0)")
print(a.ExportToWkt())
print(a.GetZ())

b = ogr.Geometry(ogr.wkbPoint)
b.AddPoint(7, 7, 7)
print(b.ExportToWkt())
print(b.GetZ())
```
Output:
```
POINT (0 0 0)
0.0
POINT (7 7 7)
7.0
```
The same happens with Postgis:
```sql
select st_asewkt(st_pointfromtext('POINT Z (0 0 0)'));
```
Output:
```
POINT(0 0 0)
```

rouault:
> why GDAL does not use the 'Z' on the WKT output?

you need to use ExportToIsoWkt().  ExporttoWkt() uses an older variant

elpaso (approved):

PR #10376, opened 2024-07-07
OGR SQL lexer: fix memory leak on invalid expression (master only)

rouault:
Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=69985


elpaso:
I cannot topen the link (permission denied).

elpaso (approved):

rouault:
> I cannot open the link (permission denied).

the memleak could be triggered with something like: ``ogrinfo autotest/ogr/data/poly.shp -where "p." -al`` . One of the added test triggers this situation

PR #10404, opened 2024-07-10
Add scripts/update_CMakeLists.py to update cmake_minimum_required() of CMakeLists.txt in subdirectories

rouault:


dbaston:
Is it possible to do something along the lines of
```
# somewhere that is included by both the top-level CMakeLists.txt and cmake/helpers/SetupStandalonePlugin.cmake
set(CMAKE_VERSION_MIN 3.16)
set(CMAKE_VERSION_MAX 3.28)

# various CMakeLists.txt files
cmake_minimum_required(VERSION ${CMAKE_VERSION_MIN}...${CMAKE_VERSION_MAX})
```

rouault:
@dbaston you're right. I assumed that cmake_minimum_required() should be the first statement in the top CMakeLists.txt file, but calling a include() before actually works. Dropping that PR in favor of commit https://github.com/OSGeo/gdal/pull/10403/commits/3f50558a016e89074fe4b0abcc290c07f5b38d31 added to https://github.com/OSGeo/gdal/pull/10403

Issue #10410, opened 2024-07-12
Issue with construction of server path on the OAPIF driver


doublebyte1:
### What is the bug?

`ogrinfo OAPIF:https://maps.gnosis.earth/ogcapi/collections/Daraa:HydrographySrf/` triggers this error: 

```
ERROR 1: HTTP error code : 400
ERROR 1: <h1>GNOSIS Map Server (OGCAPI) - 400 Bad Request</h1><h3>Unsupported or malformed request</h3>
```

The same request works for urls that do not contain relative links, like `https://ogc-api.nrw.de/inspire-us-feuerwehr` or `https://emotional.byteroad.net/collections/hex350_grid_cardio_1920`

As pointed out by @jerstlouis, according to RFC 8288 / RFC 3986 Relative links should be resolved properly :
 
https://docs.ogc.org/is/19-072/19-072.html#_11b9b4f7-42fc-413a-b63a-e7fb060b5e4b
 
https://datatracker.ietf.org/doc/html/rfc8288#section-3.1

https://datatracker.ietf.org/doc/html/rfc3986#section-5

### Steps to reproduce the issue

`ogrinfo OAPIF:https://maps.gnosis.earth/ogcapi/collections/Daraa:HydrographySrf/`

### Versions and provenance

Ubuntu 22.04.4 LTS
GDAL - built from master branch on 11/07/2024

rouault:
@doublebyte1 @jerstlouis  The driver actually tries to resolve relative links, but it is confused, thinking the base URI is https://maps.gnosis.earth/ogcapi, and then concatenating to it values like "/ogcapi/api?f=json" returned from "https://maps.gnosis.earth/ogcapi"
Reading https://datatracker.ietf.org/doc/html/rfc3986#section-5 is far from being crystal clear to me what is supposed to be the base URI.
I assume that it should be the {root } described at "Figure 2 — Example URI and Components" of https://docs.ogc.org/is/19-072/19-072.html#_6805d238-a16a-4b2c-9cfa-d0551bf25587 ?
Some clarification could be welcome

doublebyte1:
@rouault we created a PR that partially addresses this: https://github.com/OSGeo/gdal/pull/10412

jerstlouis:
@rouault Essentially I believe it should behave exactly like an HTML page with a `<a href='...'>` link.

My understanding is that a URL staring with:
- `/` goes back to the root of the domain
- `./` keep the link's context and adds the rest
[...]

PR #10412, opened 2024-07-12
Fix partially the problem with server paths on OAPIF driver

doublebyte1:
There is an issue supporting server paths that contain relative links. This is described here: 
https://github.com/OSGeo/gdal/issues/10410

This PR  addresses relative paths beginning with /, but OGR should also handle paths starting with ./ and ../../ , so more work needs to be done.


rouault:
more extensive fix in https://github.com/OSGeo/gdal/pull/10413

PR #10455, opened 2024-07-20
[Lint] OSM: const'ify and smartptr'ify

rouault:


port/cpl_vsi_mem.cpp
```diff
@@ -266,13 +265,21 @@ bool VSIMemFile::SetLength(vsi_l_offset nNewLength)
             return false;
         }
 
-        const vsi_l_offset nNewAlloc = (nNewLength + nNewLength / 10) + 5000;
+        // If the first allocation is 1 MB or above, just take that value
+        // as the one to allocate
+        // Otherwise slightly reserve more to avoid too frequent reallocations.
+        const vsi_l_offset nNewAlloc =
```
elpaso, line 271: How are these changes related with OSM changes?

elpaso (approved):

port/cpl_vsi_mem.cpp
rouault: yes slight "optimization" I discovered when realizing that the OSM code didn't actually pre-allocate the /vsimem/ file as it was supposed to do

PR #10456, opened 2024-07-20
MRF/LERC: check that values are within target type range before casting

rouault:
CC @lucianpls

### DIFF
--- a/frmts/mrf/LERCV1/Lerc1Image.cpp
+++ b/frmts/mrf/LERCV1/Lerc1Image.cpp
@@ -25,6 +25,7 @@ Contributors:  Thomas Maurer
 #include <cstdint>
 #include <cmath>
 #include <cfloat>
+#include <climits>
 #include <string>
 #include <algorithm>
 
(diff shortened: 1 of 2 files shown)


lucianpls:
@rouault 
Lerc1Image is used in other projects, it can't depend on gdal headers.
How about using the explicit limits directly:
```
static int numBytesFlt(float z)
{
    if (z > 32767 || z < -32768 || z != int16_t(z))
        return 4;
    if (z > 127 || z < -128)
        return 2;
    return 1;
}
```

rouault:
> How about using the explicit limits directly:

done

lucianpls:
@rouault
GDALIsValueInRange<float>(NAN) returns false. It should be true, since NAN is a valid float value.

lucianpls:
> @rouault GDALIsValueInRange<float>(NAN) returns false. It should be true, since NAN is a valid float value for use as NDV in this case.

I'm not sure that NAN works as explicit NDV, not the infinities. But it does seem strange that infinities are considered in range, while NAN is not.

rouault:
> But it does seem strange that infinities are considered in range, while NAN is not.

yeah, I don't recall the reason for that. This was changed long time ago in 6e7704a9a47d62b67cca2fd1bd9cfe45df8266c4. Probably because NaN are generally annoying to work with, so GDALIsValueInRange() excludes them.

> I'm not sure that NAN works as explicit NDV, nor the infinities

Do you want me to add a std::isfinite() test in addition ?

lucianpls:
> Do you want me to add a std::isfinite() test in addition ?

No, it should be fine the way it looks now. Thanks!

PR #10457, opened 2024-07-20
GPKG raster: honor sparse dataset in reading and writing

rouault:


ogr/ogrsf_frmts/gpkg/gdalgeopackagerasterband.cpp
```diff
@@ -1229,6 +1231,144 @@ CPLErr GDALGPKGMBTilesLikeRasterBand::IReadBlock(int nBlockXOff, int nBlockYOff,
...
+#endif
+
+    sqlite3_stmt *hStmt = nullptr;
+    int rc = sqlite3_prepare_v2(m_poTPD->IGetDB(), pszSQL, -1, &hStmt, nullptr);
+    if (rc != SQLITE_OK)
+    {
+        CPLError(CE_Failure, CPLE_AppDefined, "failed to prepare SQL %s: %s",
+                 pszSQL, sqlite3_errmsg(m_poTPD->IGetDB()));
+        sqlite3_free(pszSQL);
+        return GDAL_DATA_COVERAGE_STATUS_UNIMPLEMENTED |
```
elpaso, line 1284: Is this the correct return value in case of errors? The doc say: `Flag returned by GDALGetDataCoverageStatus() when the driver does not implement GetDataCoverageStatus().` but here it is implemented.

rouault: > Is this the correct return value in case of errors? The doc say: `Flag returned by GDALGetDataCoverageStatus() when the driver does not implement GetDataCoverageStatus().` but here it is implemented.

yeah, we don't really have a status code for error. So I thought defaulting to that was the best. With GDAL_DATA_COVERAGE_STATUS_UNIMPLEMENTED we document that the result isn't totally reliable. And with GDAL_DATA_COVERAGE_STATUS_DATA we are "pessimistic", so that user code doesn't skip that region

PR #10459, opened 2024-07-20
GTI: implement GetDataCoverageStatus()

rouault:


autotest/gdrivers/gti.py
```diff
@@ -1359,6 +1380,12 @@ def test_gti_no_source(tmp_vsimem):
         is None
     )
 
+    if ogrtest.have_geos():
+        (flags, pct) = vrt_ds.GetRasterBand(1).GetDataCoverageStatus(
+            0, 0, vrt_ds.RasterXSize, vrt_ds.RasterYSize
+        )
+        assert flags == gdal.GDAL_DATA_COVERAGE_STATUS_EMPTY and pct == 0.0
+
```
elpaso, line 1388: do we have (and need) a test for pct > 0 and < 100 ?

elpaso (approved):

autotest/gdrivers/gti.py
rouault: test_gti_gap_between_sources() test case added for that

PR #10461, opened 2024-07-21
Create function GDALInterpolateAtPoint

jjimenezshaw:
## What does this PR do?
Creates the function `GDALInterpolateAtPoint` to interpolate a value in a raster band.
Interpolation methods are (so far) `nearest`, `bilinear` and `cubic`.
It can be used via

- `C` API `GDALRasterInterpolateAtPoint`
- `C++` API `Band::InterpolateAtPoint`  
- `python` via `InterpolateAtPoint` in a raster band
- CLI app `gdallocationinfo` with the new option `-r`

The `C` API is
``` C
CPLErr GDALRasterInterpolateAtPoint(GDALRasterBandH hBand, double dfPixel,
                                    double dfLine,
                                    GDALRIOResampleAlg eInterpolation,
                                    double *pdfRealValue, double *pdfImagValue)
```
For complex layers the implementation is not done yet (to be done in a later PR), but the API is ready. In `C++` `pdfImagValue` is by default a `null_ptr` to simplify the usage.

As sugested by @rouault , the implementation is extracted from the RPC code, that currently has a cache on the raster data. That makes the code more performant if several values are requested in the same area of the raster. The code is removed from `alg/gdal_rpc.cpp` and pasted with some modifications in `alg/gdal_interpolateatpoint.cpp`.

## Interpolation methods
Only `nearest`, `bilinear` and `cubic` methods are implemented so far. For the borders a minor adjust is done to be consistent between the top and left borders (that were extrapolating) with the bottom and right (that are downgrading to a simpler algorithm). Now both are downgrading to a simpler algorithm.

However this produces discontinuities in the returned values at the borders. In a later PR I would like to change the behaviour in the borders doing something similar as show in the images at https://en.wikipedia.org/wiki/Bicubic_interpolation

## Implementation problems
There are some questions/problems described in the comments.

https://lists.osgeo.org/pipermail/gdal-dev/2024-April/058899.html

## Tasklist

 - [x] Add test case(s)
 - [x] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/alg/gdal_interpolateatpoint.cpp
+++ b/alg/gdal_interpolateatpoint.cpp
@@ -1,0 +1,288 @@
+/******************************************************************************
+ * $Id$
+ *
+ * Project:  GDAL DEM Interpolation
+ * Purpose:  Interpolation algorithms with cache
+ * Author:   Frank Warmerdam, warmerdam@pobox.com
+ *
+ ******************************************************************************
+ * Copyright (c) 2001, Frank Warmerdam
+ * Copyright (c) 2008-2012, Even Rouault <even dot rouault at spatialys.com>
+ * Copyright (c) 2024, Javier Jimenez Shaw
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a
+ * copy of this software and associated documentation files (the "Software"),
+ * to deal in the Software without restriction, including without limitation
+ * the rights to use, copy, modify, merge, publish, distribute, sublicense,
+ * and/or sell copies of the Software, and to permit persons to whom the
+ * Software is furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included
+ * in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+ * OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+ * THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ ****************************************************************************/
+
+#include "gdal_interpolateatpoint.h"
+
+#include <algorithm>
+
+static bool
+GDALInterpExtractDEMWindow(GDALRasterBand *pBand,
+                           std::unique_ptr<DoublePointsCache> &cacheDEM, int nX,
+                           int nY, int nWidth, int nHeight, double *padfOut)
+{
+    constexpr int BLOCK_SIZE = 64;
+
+    // Request the DEM by blocks of BLOCK_SIZE * BLOCK_SIZE and put them
+    // in cacheDEM
+    if (!cacheDEM)
+        cacheDEM.reset(new DoublePointsCache{});
+
+    const int nXIters = (nX + nWidth - 1) / BLOCK_SIZE - nX / BLOCK_SIZE + 1;
+    const int nYIters = (nY + nHeight - 1) / BLOCK_SIZE - nY / BLOCK_SIZE + 1;
+    const int nRasterXSize = pBand->GetXSize();
+    const int nRasterYSize = pBand->GetYSize();
+    for (int iY = 0; iY < nYIters; iY++)
+    {
+        const int nBlockY = nY / BLOCK_SIZE + iY;
+        const int nReqYSize =
+            std::min(nRasterYSize - nBlockY * BLOCK_SIZE, BLOCK_SIZE);
+        const int nFirstLineInCachedBlock = (iY == 0) ? nY % BLOCK_SIZE : 0;
+        const int nFirstLineInOutput =
+            (iY == 0) ? 0
+                      : BLOCK_SIZE - (nY % BLOCK_SIZE) + (iY - 1) * BLOCK_SIZE;
+        const int nLinesToCopy = (nYIters == 1) ? nHeight
+                                 : (iY == 0)    ? BLOCK_SIZE - (nY % BLOCK_SIZE)
+                                 : (iY == nYIters - 1)
+                                     ? 1 + (nY + nHeight - 1) % BLOCK_SIZE
+                                     : BLOCK_SIZE;
+        for (int iX = 0; iX < nXIters; iX++)
+        {
+            const int nBlockX = nX / BLOCK_SIZE + iX;
+            const int nReqXSize =
+                std::min(nRasterXSize - nBlockX * BLOCK_SIZE, BLOCK_SIZE);
+            const uint64_t nKey =
+                (static_cast<uint64_t>(nBlockY) << 32) | nBlockX;
+            const int nFirstColInCachedBlock = (iX == 0) ? nX % BLOCK_SIZE : 0;
+            const int nFirstColInOutput =
+                (iX == 0)
+                    ? 0
+                    : BLOCK_SIZE - (nX % BLOCK_SIZE) + (iX - 1) * BLOCK_SIZE;
+            const int nColsToCopy = (nXIters == 1) ? nWidth
+                                    : (iX == 0) ? BLOCK_SIZE - (nX % BLOCK_SIZE)
+                                    : (iX == nXIters - 1)
+                                        ? 1 + (nX + nWidth - 1) % BLOCK_SIZE
+                                        : BLOCK_SIZE;
+
+#if 0
+            CPLDebug("RPC", "nY=%d nX=%d nBlockY=%d nBlockX=%d "
+                     "nFirstLineInCachedBlock=%d nFirstLineInOutput=%d nLinesToCopy=%d "
+                     "nFirstColInCachedBlock=%d nFirstColInOutput=%d nColsToCopy=%d",
+                     nY, nX, nBlockY, nBlockX, nFirstLineInCachedBlock, nFirstLineInOutput, nLinesToCopy,
+                     nFirstColInCachedBlock, nFirstColInOutput, nColsToCopy);
+#endif
+
+            std::shared_ptr<std::vector<double>> poValue;
+            if (!cacheDEM->tryGet(nKey, poValue))
+            {
+                poValue = std::make_shared<std::vector<double>>(nReqXSize *
+                                                                nReqYSize);
+                CPLErr eErr = pBand->RasterIO(
+                    GF_Read, nBlockX * BLOCK_SIZE, nBlockY * BLOCK_SIZE,
+                    nReqXSize, nReqYSize, poValue->data(), nReqXSize, nReqYSize,
+                    GDT_Float64, 0, 0, nullptr);
+                if (eErr != CE_None)
+                {
+                    return false;
+                }
+                cacheDEM->insert(nKey, poValue);
+            }
+
+            // Compose the cached block to the final buffer
+            for (int j = 0; j < nLinesToCopy; j++)
+            {
+                memcpy(padfOut + (nFirstLineInOutput + j) * nWidth +
+                           nFirstColInOutput,
+                       poValue->data() +
+                           (nFirstLineInCachedBlock + j) * nReqXSize +
+                           nFirstColInCachedBlock,
+                       nColsToCopy * sizeof(double));
+            }
+        }
+    }
+
+#if 0
+    CPLDebug("RPC_DEM", "DEM for %d,%d,%d,%d", nX, nY, nWidth, nHeight);
+    for(int j = 0; j < nHeight; j++)
+    {
+        std::string osLine;
+        for(int i = 0; i < nWidth; ++i )
+        {
+            if( !osLine.empty() )
+                osLine += ", ";
+            osLine += std::to_string(padfOut[j * nWidth + i]);
+        }
+        CPLDebug("RPC_DEM", "%s", osLine.c_str());
+    }
+#endif
+
+    return true;
+}
+
+/************************************************************************/
+/*                        GDALInterpolateAtPoint()                      */
+/************************************************************************/
+
+bool GDALInterpolateAtPoint(GDALRasterBand *pBand,
+                            GDALRIOResampleAlg eResampleAlg,
+                            std::unique_ptr<DoublePointsCache> &cache,
+                            const double dfXIn, const double dfYIn,
+                            double *pdfOutputValue)
+{
+    const int nRasterXSize = pBand->GetXSize();
+    const int nRasterYSize = pBand->GetYSize();
+    int bGotNoDataValue = FALSE;
+    const double dfNoDataValue = pBand->GetNoDataValue(&bGotNoDataValue);
+
+    if (eResampleAlg == GDALRIOResampleAlg::GRIORA_CubicSpline)
+    {
+        // Convert from upper left corner of pixel coordinates to center of
+        // pixel coordinates:
+        const double dfX = dfXIn - 0.5;
+        const double dfY = dfYIn - 0.5;
+        const int dX = static_cast<int>(std::floor(dfX));
+        const int dY = static_cast<int>(std::floor(dfY));
+        const double dfDeltaX = dfX - dX;
+        const double dfDeltaY = dfY - dY;
+
+        const int dXNew = dX - 1;
+        const int dYNew = dY - 1;
+        if (!(dXNew >= 0 && dYNew >= 0 && dXNew + 4 <= nRasterXSize &&
+              dYNew + 4 <= nRasterYSize))
+        {
+            goto bilinear_fallback;
+        }
+        // CubicSpline interpolation.
+        double adfElevData[16] = {0.0};
+        if (!GDALInterpExtractDEMWindow(pBand, cache, dXNew, dYNew, 4, 4,
+                                        adfElevData))
+        {
+            return FALSE;
+        }
+
+        double dfSumH = 0.0;
+        double dfSumWeight = 0.0;
+        for (int k_i = 0; k_i < 4; k_i++)
+        {
+            // Loop across the X axis.
+            for (int k_j = 0; k_j < 4; k_j++)
+            {
+                // Calculate the weight for the specified pixel according
+                // to the bicubic b-spline kernel we're using for
+                // interpolation.
+                const int dKernIndX = k_j - 1;
+                const int dKernIndY = k_i - 1;
+                const double dfPixelWeight =
+                    BiCubicSplineKernel(dKernIndX - dfDeltaX) *
+                    BiCubicSplineKernel(dKernIndY - dfDeltaY);
+
+                // Create a sum of all values
+                // adjusted for the pixel's calculated weight.
+                const double dfElev = adfElevData[k_j + k_i * 4];
+                if (bGotNoDataValue && ARE_REAL_EQUAL(dfNoDataValue, dfElev))
+                    continue;
+
+                dfSumH += dfElev * dfPixelWeight;
+                dfSumWeight += dfPixelWeight;
+            }
+        }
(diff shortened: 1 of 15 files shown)


jjimenezshaw:
**Problem with cubic interpolation**:
As mentioned in the python test `test_interpolateatpoint_cubic_several_points` 
```python
    raster_array_1 = np.array(
        (
            [1.0, 2.0, 1.5, -0.3],
            [1.0, 2.0, 1.5, -0.3],
            [1.0, 2.0, 1.5, -0.3],
            [1.0, 2.0, 1.5, -0.3],
        )
    )
    mem_ds.GetRasterBand(1).WriteArray(raster_array_1)

    # using https://tools.timodenk.com/cubic-spline-interpolation to get some samples
    got_cubic = mem_ds.GetRasterBand(1).InterpolateAtPoint(1.5, 1.5, gdal.GRIORA_Cubic)
    assert got_cubic == pytest.approx(
        1.75, 1e-6
    )  # I would expect 2.0, as it is a input point. To be studied
```
I would expect a value of 2.0, that is the pixel value at x=1.5. However it returns 1.75.
Debugging the interpolation code (copied from the RPC algorithm) I could see that all points have weights not null. Is that expected?

jjimenezshaw:
**Problem in python return code**
As shown in the test
```python
def test_interpolateatpoint_errors():

    ds = gdal.Open("data/byte.tif")

    res = ds.GetRasterBand(1).InterpolateAtPoint(1000, 120, gdal.GRIORA_Bilinear)
    assert res == 0.0  # I would expect None. To be studied
```
those coordinates are out of bound. I would expect it returning a  `None`, but it return `0.0`. I do not know what is happening in the swig code.

jjimenezshaw:
Please, feel free to suggest new names or locations for functions and files if needed.

gcore/gdal.h
```diff
@@ -2552,6 +2552,12 @@ int CPL_DLL GDALDimensionSetIndexingVariable(GDALDimensionH hDim,
                                              GDALMDArrayH hArray);
 bool CPL_DLL GDALDimensionRename(GDALDimensionH hDim, const char *pszNewName);
 
+CPLErr CPL_DLL GDALRasterInterpolateAtPoint(GDALRasterBandH hBand,
```
rouault, line 2555: this should be moved upper in the file, with the other functions that take GDALRasterBandH as their first arguement

gcore/gdal_priv.h
```diff
@@ -1804,6 +1807,14 @@ class CPL_DLL GDALRasterBand : public GDALMajorObject
 
     std::shared_ptr<GDALMDArray> AsMDArray() const;
 
+    /** Interpolates value given a resalmple algorithm
+     * @since GDAL 3.10
+     */
```
rouault, line 1812: redundant with the doc in the .cpp file:
```suggestion
```

gcore/gdalrasterband.cpp
```diff
@@ -8869,3 +8870,83 @@ std::shared_ptr<GDALMDArray> GDALRasterBand::AsMDArray() const
     return GDALMDArrayFromRasterBand::Create(
         poDS, const_cast<GDALRasterBand *>(this));
 }
+
+/************************************************************************/
+/*                             InterpolateAtPoint()                     */
+/************************************************************************/
+
+/**
+ * \brief interpolates the value between pixels using
```
rouault, line 8879:
```suggestion
 * \brief Interpolates the value between pixels using
```

gcore/gdalrasterband.cpp
```diff
@@ -8869,3 +8870,83 @@ std::shared_ptr<GDALMDArray> GDALRasterBand::AsMDArray() const
...
+ * a resampling algorithm
+ *
+ * @param dfPixel pixel coordinate as a double, where interpolation should be done.
+ * @param dfLine line coordinate as a double, where interpolation should be done.
+ * @param eInterpolation interpolation type. Only near, bilinear and cubic are allowed.
+ * @param pdfRealValue pointer to interpolated value (real part)
+ * @param pdfImagValue pointer to interpolated value (imaginary part)
```
rouault, line 8886:
```suggestion
 * @param pdfImagValue pointer to imaginary part of interpolated value (may be null if not needed)
```

gcore/gdalrasterband.cpp
```diff
@@ -8869,3 +8870,83 @@ std::shared_ptr<GDALMDArray> GDALRasterBand::AsMDArray() const
...
+/************************************************************************/
+
+/**
+ * \brief interpolates the value between pixels using
+ * a resampling algorithm
+ *
+ * @param dfPixel pixel coordinate as a double, where interpolation should be done.
+ * @param dfLine line coordinate as a double, where interpolation should be done.
+ * @param eInterpolation interpolation type. Only near, bilinear and cubic are allowed.
+ * @param pdfRealValue pointer to interpolated value (real part)
```
rouault, line 8885:
```suggestion
 * @param pdfRealValue pointer to real part of interpolated value
```

gcore/gdalrasterband.cpp
```diff
@@ -8869,3 +8870,83 @@ std::shared_ptr<GDALMDArray> GDALRasterBand::AsMDArray() const
...
+
+    return res ? CE_None : CE_Failure;
+}
+
+/************************************************************************/
+/*                        GDALRasterInterpolateAtPoint()                */
+/************************************************************************/
+
+/**
+ * \brief interpolates the value between pixels using
```
rouault, line 8935:
```suggestion
 * \brief Interpolates the value between pixels using
```

swig/include/python/docs/gdal_band_docs.i
```diff
@@ -528,6 +528,23 @@ Returns
...
+Parameters
+----------
+pixel : float
+line : float
+interpolation : GRIOResampleAlg (nearest, bilinear, cubic)
+
+Returns
+-------
+float:
+    Interpolated value, , or ``None`` if it has any error.
```
rouault, line 545:
```suggestion
    Interpolated value, or ``None`` if it has any error.
```

swig/include/python/gdal_python.i
```diff
@@ -4939,3 +4939,26 @@ def quiet_errors():
         tuple.max_part_count = max_part_count
         val = tuple
 %}
+
+%feature("shadow") InterpolateAtPoint %{
+def InterpolateAtPoint(self, *args, **kwargs):
+    """InterpolateAtPoint(Band self, pixel, line, interpolation) -> value
+
+    ... doc here TODO ...
```
rouault, line 4947: I'm not sure which of this inline function or gdal_band_docs.i wins over the other one, but if this is this one, it should be updated. Otherwise removed

autotest/gcore/interpolateatpoint.py
```diff
@@ -0,0 +1,143 @@
...
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Library General Public License for more details.
+#
+# You should have received a copy of the GNU Library General Public
+# License along with this library; if not, write to the
+# Free Software Foundation, Inc., 59 Temple Place - Suite 330,
+# Boston, MA 02111-1307, USA.
```
rouault, line 26: MIT licence is prefered:
```suggestion
# SPDX-License-Identifier: MIT
```

alg/gdal_interpolateatpoint.h
```diff
@@ -0,0 +1,85 @@
...
+
+#include <memory>
+
+typedef enum
+{
+    /*! Nearest neighbour (select on one input pixel) */ DRA_NearestNeighbour =
+        0,
+    /*! Bilinear (2x2 kernel) */ DRA_Bilinear = 1,
+    /*! Cubic Convolution Approximation (4x4 kernel) */ DRA_Cubic = 2
+} DEMResampleAlg;
```
rouault, line 50: DEMResampleAlg should be renamed to maybe GIAPResampleAlg (GIAP = GDALInterpolateAtPoint)

alg/gdal_interpolateatpoint.h
```diff
@@ -0,0 +1,85 @@
...
+class GDALDoublePointsCache
+{
+  public:
+    std::unique_ptr<DoublePointsCache> cache{};
+};
+
+int GDALInterpolateAtPoint(GDALRasterBand *pBand, DEMResampleAlg eResampleAlg,
+                           std::unique_ptr<DoublePointsCache> &cacheDEM,
+                           const double dfXIn, const double dfYIn,
+                           double *pdfDEMH);
```
rouault, line 64:
```suggestion
bool GDALInterpolateAtPoint(GDALRasterBand *pBand, DEMResampleAlg eResampleAlg,
                           std::unique_ptr<DoublePointsCache> &cache,
                           const double dfXIn, const double dfYIn,
                           double *pdfOutputValue);
```

apps/gdallocationinfo.cpp
```diff
@@ -137,6 +137,12 @@ MAIN_START(argc, argv)
         .help(_("Query the (overview_level)th overview (overview_level=1 is "
                 "the 1st overview)."));
 
+    std::string osResampling;
+    argParser.add_argument("-r")
```
rouault, line 141: doc/source/programs/gdallocationinfo.rst to update with the new option

alg/gdal_interpolateatpoint.cpp
```diff
@@ -0,0 +1,288 @@
...
+#endif
+
+    return true;
+}
+
+/************************************************************************/
+/*                        GDALInterpolateAtPoint()                      */
+/************************************************************************/
+
+int GDALInterpolateAtPoint(GDALRasterBand *pBand, DEMResampleAlg eResampleAlg,
```
rouault, line 143:
```suggestion
bool GDALInterpolateAtPoint(GDALRasterBand *pBand, DEMResampleAlg eResampleAlg,
```

gcore/gdalrasterband.cpp
```diff
@@ -8869,3 +8870,83 @@ std::shared_ptr<GDALMDArray> GDALRasterBand::AsMDArray() const
...
+                "Only near, bilinear and cubic interpolation methods allowed");
+
+            return CE_Failure;
+            break;
+    }
+    GDALRasterBand *pBand = const_cast<GDALRasterBand *>(this);
+    if (!m_oPointsCache)
+        m_oPointsCache = std::make_unique<GDALDoublePointsCache>();
+
+    auto res =
```
rouault, line 8921:
```suggestion
    bool res =
```

autotest/utilities/test_gdallocationinfo.py
```diff
@@ -299,3 +299,81 @@ def test_gdallocationinfo_out_of_raster_coordinates_valonly_multiband(
...
+    assert float(ret) == pytest.approx(125.795025, rel=1e-4)
+
+
+def test_gdallocationinfo_report_geoloc_interpolate_bilinear(gdallocationinfo_path):
+
+    ret = gdaltest.runexternal(
+        gdallocationinfo_path
+        + " -r bilinear -geoloc ../gcore/data/byte.tif 441319.09 3750601.80"
+    )
+    print(ret)
```
rouault, line 334:
```suggestion
```

autotest/utilities/test_gdallocationinfo.py
```diff
@@ -299,3 +299,81 @@ def test_gdallocationinfo_out_of_raster_coordinates_valonly_multiband(
...
+    assert "P,11.97" in ret
+    assert "Value: 137.2524" in ret
+
+
+def test_gdallocationinfo_report_interpolate_bilinear(gdallocationinfo_path):
+
+    ret = gdaltest.runexternal(
+        gdallocationinfo_path + " -r bilinear ../gcore/data/byte.tif 9.98 11.97"
+    )
+    print(ret)
```
rouault, line 347:
```suggestion
```

autotest/gcore/interpolateatpoint.py
```diff
@@ -0,0 +1,143 @@
...
+        (
+            [1.0, 2.0, 1.5, -0.3],
+            [1.0, 2.0, 1.5, -0.3],
+            [1.0, 2.0, 1.5, -0.3],
+            [1.0, 2.0, 1.5, -0.3],
+        )
+    )
+    mem_ds.GetRasterBand(1).WriteArray(raster_array_1)
+
+    # using https://tools.timodenk.com/cubic-spline-interpolation to get some samples
```
rouault, line 134: 2 potential reasons:
- perhaps because the cubic spline in this page is not the same method as the cubic of GDAL. The Cubic in GDAL is a Catmull-Rohm cubic (https://github.com/OSGeo/gdal/blob/b6b5546456fae7fff72560128d658c5e5ce597e5/alg/gdalwarpkernel.cpp#L3391)
- or the 0.5 pixel shift applied (cf comments "Convert from upper left corner of pixel coordinates to center of" in GDALRPCGetDEMHeight()). The reasoning is that the pixel value is exactly reached at the center of the pixel.

rouault:
> **Problem in python return code**

I've pushed a commit in your branch to fix/workaround that.

You didn't do anything wrong. Back in 2006, the IF_ERROR_RETURN_NONE typemap was half-restored in 86d4f1d47ae after having been removed in 1c6e19f94f64888eadb02ed05d1465c673bb27b6 . So the end result is that the IF_ERROR_RETURN_NONE typemap currently only eats the CPLErr return code but doesn't do anything with it. I had always that feeling that something didn't work as intended when using GetStatistics() which uses this typemap too, but never investigated in depth. I'm a bit hesitant if the typemap should be fixed now. Probably yes. Anyway this is out of scope for this PR.

autotest/gcore/interpolateatpoint.py
jjimenezshaw: I plotted the values with the three interpolations and got this:
(5 pixels in the x axis, the last two equal)
in any case, I will try to fix it in a later PR, because that interpolation affects RPC, and I do not want to add noise there.

Figure_1


```python
def test_plot():
    import matplotlib.pyplot as plt

    mem_ds = gdal.GetDriverByName("MEM").Create(
        "", xsize=5, ysize=4, bands=1, eType=gdal.GDT_Float32
    )
    raster_array_1 = np.array(
        (
            [1.0, 2.0, 1.5, 0.3, 0.3],
            [1.0, 2.0, 1.5, 0.3, 0.3],
            [1.0, 2.0, 1.5, 0.3, 0.3],
            [1.0, 2.0, 1.5, 0.3, 0.3],
        )
    )
    band = mem_ds.GetRasterBand(1)
    band.WriteArray(raster_array_1)

    t = np.arange(0.0, 5.0, 0.01)
    y = 1.5
    y_cubic = [band.InterpolateAtPoint(x, y, gdal.GRIORA_Cubic) for x in t]
    y_linear = [band.InterpolateAtPoint(x, y, gdal.GRIORA_Bilinear) for x in t]
    y_near = [band.InterpolateAtPoint(x, y, gdal.GRIORA_NearestNeighbour) for x in t]

    plt.plot(t, y_linear, "r", t, y_cubic, "b", t, y_near, "g")
    plt.grid(True)
    plt.show()
```

jjimenezshaw: > * The Cubic in GDAL is a Catmull-Rohm cubic

The code in `BiCubicKernel` seems to be similar to `GWKBSpline`
https://github.com/OSGeo/gdal/blob/b6b5546456fae7fff72560128d658c5e5ce597e5/alg/gdalwarpkernel.cpp#L3464

rouault: > The code in `BiCubicKernel` seems to be similar to `GWKBSpline`

oh! you're right. I didn't realize that. This was introduced in 8a7c8dcd701 . Ok, so we should rather map GRIORA_CubicSpline to it, and rename it internally to CubicBSpline
It would be nice to add proper mapping for Catmull-Rohm cubic / GRIORA_Cubic, but that can be done later

jjimenezshaw: > > The code in `BiCubicKernel` seems to be similar to `GWKBSpline`
> 
> oh! you're right. I didn't realize that. This was introduced in [8a7c8dc](https://github.com/OSGeo/gdal/commit/8a7c8dcd7016cd1798b86001e427360bb5338363) . Ok, so we should rather map GRIORA_CubicSpline to it, and rename it internally to CubicBSpline It would be nice to add proper mapping for Catmull-Rohm cubic / GRIORA_Cubic, but that can be done later

That was my impression yesterday night. But it opens questions respect to RPC code.
I see three enums about interpolation methods (are there more?):

- `DEMResampleAlg`
- `GDALRIOResampleAlg`
- `GDALResampleAlg`

The last two have to be equivalent if I understand correctly the comment before their definitions.
The first one is used only in RPC, right?
But based on what we have seen here, `DEMResampleAlg::DRA_Cubic` is equivalent to `GDALResampleAlg::GRA_CubicSpline`

At `gdal_rpc.cpp` is the only place I found that something is set to `DRA_Cubic`:
```C++
    const char *pszDEMInterpolation =
        CSLFetchNameValueDef(papszOptions, "RPC_DEMINTERPOLATION", "bilinear");
    if (EQUAL(pszDEMInterpolation, "near"))
    {
        psTransform->eResampleAlg = DRA_NearestNeighbour;
    }
    else if (EQUAL(pszDEMInterpolation, "bilinear"))
    {
        psTransform->eResampleAlg = DRA_Bilinear;
    }
    else if (EQUAL(pszDEMInterpolation, "cubic"))
    {
        psTransform->eResampleAlg = DRA_Cubic;
    }
```
that is "conflicting" with other places (where we use the `-r` argument) about the meaning of `cubic`.

So, me guess about what to do is:

- Implement an interpolation in `InterpolateAtPoint` for `GDALResampleAlg::GRA_Cubic` for `cubic` and rename the current one for `cubicspline`.
- When RPC is using `DRA_Cubic`, use `GRA_CubicSpline`. That will keep the current (3.9) behavior. Would it be a good idea to change that in GDAL 4.0, so RPC would have `cubic` and `cubicspline`, consistent with other places in the code?
- Change the API of the internal function `GDALInterpolateAtPoint` to use `GDALRIOResampleAlg` or  `GDALResampleAlg` instead of `DEMResampleAlg`...
- Question: What is the difference between `GDALRIOResampleAlg` and `GDALResampleAlg`? Which one should I use in the internal function `GDALInterpolateAtPoint`?

rouault: > The last two have to be equivalent if I understand correctly the comment before their definitions.

The values of the constants are designed to match when they are available in both, but some methods are present in one and not in the other one

> The first one is used only in RPC, right?

yes was purely an implementation detail of the RPC code

> But based on what we have seen here, DEMResampleAlg::DRA_Cubic is equivalent to GDALResampleAlg::GRA_CubicSpline

yes apparently.  so DRA_Cubic should be probably renamed to DRA_CubicSpline for consistence

> that is "conflicting" with other places (where we use the -r argument) about the meaning of cubic.

indeed. For existing RPC_DEMINTERPOLATION=cubic, we can't change the name. But for the new -r flag of gdallocationinfo, we should accept "cubicspline" for now, and reserve "cubic" for Catmull-Rohm cubic if we implement it at some later time

> Change the API of the internal function GDALInterpolateAtPoint to use GDALRIOResampleAlg

yes we could do that to avoid adding new enumerations.

> What is the difference between GDALRIOResampleAlg and GDALResampleAlg?

GDALRIOResampleAlg is for the methods supported by RasterIO() and the overview building code
GDALResampleAlg is for the warping code

> Which one should I use in the internal function GDALInterpolateAtPoint?

I would go for GDALRIOResampleAlg

apps/gdallocationinfo.cpp
jjimenezshaw: done

gcore/gdal.h
jjimenezshaw: done

swig/include/python/gdal_python.i
jjimenezshaw: done. This one wins over.

alg/gdal_interpolateatpoint.h
jjimenezshaw: returned to its original place.

alg/gdal_interpolateatpoint.h
jjimenezshaw: done

gcore/gdalrasterband.cpp
jjimenezshaw: done

jjimenezshaw:
I hope I addressed all the comments. Including make obvious that the interpolation is splinecubic. Next PR will be to include cubic.

PR #10463, opened 2024-07-21
geometry classes: avoid potential unlikely int overflows

rouault:


elpaso (approved):
LGTM but is this tested?

rouault:
> LGTM but is this tested?

for the OGRLineString change, yes.
for curvecollection and geometrycollection, it would be very difficult to test as it would really require to add 2 billions geometries. Would take an awful lot of time and CI workers would probably have not enough RAM

PR #10464, opened 2024-07-21
Add a cpl::contains(container, value) helper

rouault:
and use it in gcore

@dbaston Thoughts?   I'm quite tired of writing code like ``container.find(key) != containers.end()``. As of C++17, we could use ``container.count(key)``, as the closest alternative to C++20 ``container.contains(key)``, but using count() might complicate future modernization when we bump to C++20.  This ``cpl::contains(container, value)`` should be easily greppable to automate a change to ``container.contains(value)``

### DIFF
--- a/gcore/gdaldriver.cpp
+++ b/gcore/gdaldriver.cpp
@@ -1063,8 +1063,7 @@ CPLErr GDALDriver::QuietDeleteForCreateCopy(const char *pszFilename,
                 {
                     CPLString osFilename(pszFileInList);
                     osFilename.replaceAll('\\', '/');
-                    if (oSetExistingDestFiles.find(osFilename) !=
-                        oSetExistingDestFiles.end())
+                    if (cpl::contains(oSetExistingDestFiles, osFilename))
                     {
                         oSetExistingDestFilesFoundInSource.insert(osFilename);
                     }
@@ -1078,8 +1077,8 @@ CPLErr GDALDriver::QuietDeleteForCreateCopy(const char *pszFilename,
         {
             for (const std::string &osFilename : oSetExistingDestFiles)
             {
-                if (oSetExistingDestFilesFoundInSource.find(osFilename) ==
-                    oSetExistingDestFilesFoundInSource.end())
+                if (!cpl::contains(oSetExistingDestFilesFoundInSource,
+                                   osFilename))
                 {
                     VSIUnlink(osFilename.c_str());
                 }
--- a/port/cpl_port.h
+++ b/port/cpl_port.h
@@ -1161,6 +1161,14 @@ extern "C++"
     {
         return t;
     }
+
+    /** Emulates the C++20 .contains() method */
+    template <typename C, typename V>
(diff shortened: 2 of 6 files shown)


elpaso (approved):

gcore/gdaldriver.cpp
```diff
@@ -1063,8 +1063,7 @@ CPLErr GDALDriver::QuietDeleteForCreateCopy(const char *pszFilename,
                 {
                     CPLString osFilename(pszFileInList);
                     osFilename.replaceAll('\\', '/');
-                    if (oSetExistingDestFiles.find(osFilename) !=
-                        oSetExistingDestFiles.end())
+                    if (cpl::contains(oSetExistingDestFiles, osFilename))
                     {
                         oSetExistingDestFilesFoundInSource.insert(osFilename);
                     }
```
dbaston, line 1069:
```suggestion
                    oSetExistingDestFilesFoundInSource.insert(osFilename);
```

dbaston (approved):

gcore/gdaldriver.cpp
rouault: no, I actually confused myself initially doing that "optimization" :-), but oSetExistingDestFiles and oSetExistingDestFilesFoundInSource are distinct sets ...

port/cpl_port.h
```diff
@@ -1161,6 +1161,14 @@ extern "C++"
     {
         return t;
     }
+
+    /** Emulates the C++20 .contains() method */
+    template <typename C, typename V>
+    inline bool contains(const C &container, const V &value)
+    {
+        return container.count(value) != 0;
```
dg0yt, line 1169: Hm, doesn't `count` always need to do too much? It cannot stop at the first match.

rouault: not for std::set and std::map, which are up to now the only users of that

dg0yt: > not for std::set and std::map

... unless some call triggers the second overload of count? (Taking an "equivalent" key instead of an exact one.)

> up to now the only users of that

This is a very generic function interface, but the implementation is specific. It gets attention (read: review) when added or modified. It is unlikely to get attention when a different use is added elsewhere. Including in software *using* GDAL.

rouault: Thanks for pointing me to this new C++14 second overload / "equivalent" key concept. I have a bit of a hard time fully understanding the impact of that , but my understanding of https://stackoverflow.com/questions/20317413/what-are-transparent-comparators, is that it is only available if you define something like a std::key or std::map specially constructed with a std::less<>, so that shouldn't occur with the current state of the code base. That said in https://github.com/OSGeo/gdal/pull/10518 I'm proposing to use a std::find() based implementation to avoid any performance issue.

>  It is unlikely to get attention when a different use is added elsewhere. Including in software using GDAL.

This can't as this is protected by a ifdef GDAL_COMPILATION. So for GDAL internal use only

PR #10471, opened 2024-07-23
[gdal_contour] Allow to use -fl with -i and -e

### LINKED ISSUE #10172: Remove artificial gdal_contour -fl restriction
https://gdal.org/programs/gdal_contour.html says

>`-i <interval>`
>Elevation interval between contours. Ignored if -fl is used. Must specify either -i or -fl or -e.
>
>`-off <offset>`
>Offset from zero relative to which to interpret intervals. Ignored if -fl is used.

Actually if you think about it, `-fl` doesn't have any conflict at all with the other options.

Say the range of a certain bird is from 230 to 440 meters. Then the user should be allowed to use `-fl 230 440` to add these two supplementary special contours to their map, without ~~suddenly~~ becoming responsible for also listing all the other contours because now `-i` can no longer can be used (~~for "no good reason"~~), or needing to make two maps and use `ogrmerge` as a workaround.

elpaso:
Fixed levels are added to those automatically calculated from -i or -e.

Fixes #10172

### DIFF
--- a/alg/contour.cpp
+++ b/alg/contour.cpp
@@ -34,6 +34,7 @@
 #include "utility.h"
 #include "contour_generator.h"
 #include "segment_merger.h"
+#include <algorithm>
 
 #include "gdal.h"
 #include "gdal_alg.h"
@@ -686,15 +687,13 @@ CPLErr GDALContourGenerateEx(GDALRasterBandH hBand, void *hLayer,
         }
     }
 
-    // Global ok
     bool ok = false;
-    // Fixed levels ok
-    bool flOk = true;
 
     try
     {
         if (polygonize)
         {
+
             if (!fixedLevels.empty())
             {
                 // If the minimum raster value is larger than the first requested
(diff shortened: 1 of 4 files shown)


alg/contour.cpp
```diff
@@ -726,40 +730,46 @@ CPLErr GDALContourGenerateEx(GDALRasterBandH hBand, void *hLayer,
...
-                SegmentMerger<RingAppender, ExponentialLevelRangeIterator>
-                    writer(appender, levels, /* polygonize */ true);
-                ContourGeneratorFromRaster<decltype(writer),
-                                           ExponentialLevelRangeIterator>
-                    cg(hBand, useNoData, noDataValue, writer, levels);
-                ok = cg.process(pfnProgress, pProgressArg);
-            }
-            else
+
+            if (flOk)
```
rouault, line 736: hum, so you are doing a 2 pass approach: first user provided fixed levels, and then contour intervals.
I'm not completely sure this will lead to correct results. One obvious (non nominal case) is that the same level is in fixed level and in contour intervals, which will lead to duplicated lines. But even in the nominal case, when using with -p (polygonization), the fact that the levels are not sorted might lead to wrong results.
This should be checked.
My initial idea for the implementation would have rather to modify the fixed level array to merge into it the levels generated from the contour interval, and sort it.

apps/gdal_contour.cpp
```diff
@@ -82,7 +82,7 @@ GDALContourAppOptionsGetParser(GDALContourOptions *psOptions)
         "utility: http://gdal.org/gdal_contour.html"));
 
     argParser->add_extra_usage_hint(
-        _("One and only one of -i, -fl or -e must be specified."));
+        _("One one of -i, -fl or -e must be specified."));
```
rouault, line 85:
```suggestion
        _("One of -i, -fl or -e must be specified."));
```

autotest/utilities/test_gdal_contour.py
```diff
@@ -399,4 +399,232 @@ def test_gdal_contour_missing_fl_i_or_e(gdal_contour_path, testdata_tif, tmp_pat
     _, err = gdaltest.runexternal_out_and_err(
         gdal_contour_path + f" {testdata_tif} {contour_shp}"
     )
-    assert "One and only one of -i, -fl or -e must be specified." in err
+    assert "One one of -i, -fl or -e must be specified." in err
```
rouault, line 402:
```suggestion
    assert "One of -i, -fl or -e must be specified." in err
```

PR #10476, opened 2024-07-23
GPKG/SQLITE SQL: extend gdal_get_pixel_value()/gdal_get_layer_pixel_value() to support an interpolation method

rouault:
CC @jjimenezshaw


jjimenezshaw:
#10506 adds another interpolation method: `cubic`

rouault:
rebased on top of latest master to add support for cubic

PR #10499, opened 2024-07-25
HANA: Add support for REAL_VECTOR type

mrylov:
This PR adds support for the new data type called REAL_VECTOR that was introduced with HANA Cloud QRC1/2024.


ogr/ogrsf_frmts/hana/ogrhanadatasource.cpp
```diff
@@ -1048,11 +1051,14 @@ OGRErr OGRHanaDataSource::GetQueryColumns(
             continue;
 
         bool isArray = false;
-        bool isGeometry = false;
         CPLString tableName = rsmd->getTableName(clmIndex);
         CPLString columnName = rsmd->getColumnName(clmIndex);
         CPLString defaultValue;
         short dataType = rsmd->getColumnType(clmIndex);
+        // TODO: remove when HANA Client is fixed.
```
mrylov, line 1058: Oh, this needs to be fixed first.

ogr/ogrsf_frmts/hana/ogrhanalayer.cpp
```diff
@@ -114,65 +114,75 @@ CreateFieldDefn(const AttributeColumnDescription &columnDesc)
...
-        case odbc::SQLDataTypes::VarBinary:
-        case odbc::SQLDataTypes::LongVarBinary:
+        case QGRHanaDataTypes::Binary:
+        case QGRHanaDataTypes::VarBinary:
+        case QGRHanaDataTypes::LongVarBinary:
             ogrFieldType = OFTBinary;
             setFieldSize = true;
             break;
+        case QGRHanaDataTypes::RealVector:
+            ogrFieldType = OFTBinary;
```
rouault, line 178: Why not mapping this into as OGRRealList ? This would be much more user friendly

mrylov: Good point. I completely forgot about OGRRealList type, will consider your suggestion in the next commit.

rouault: > By converting a double value back to a float may produce a value different from an original one.

no, as double has more precision than float, this will round to the original float value when converted back to float as the extra bits of the mantissa are dropped.   So float -> double -> float is lossless (but not double -> float -> double obviously)

> Hence, we would prefer to stick to the binary representation for the REAL_VECTOR type.

I'm fine with that too if that's your final choice

mrylov: You are right @rouault, I've confused myself.
So, the binary representation is our final decision.

At the moment, we don't see any urgent need in treating REAL_VECTOR type as OFTRealList. I guess that not all data source providers properly support OFTRealList.
Once there is a need in supporting OGRRealList or OFTString for REAL_VECTOR type, we can always implement it later and give a control via data source open options.
For example:
```
VECTOR_DATA_TYPE: Controls how REAL_VECTOR data type is represented. Possible values are: OFTBinary, OFTRealList or OFTString.
```
or something like that.

Thank you and kind regards

PR #10505, opened 2024-07-25
All std::function as pool job

abellgithub:
Use queue for job queue.


port/cpl_worker_thread_pool.cpp
```diff
@@ -210,21 +191,23 @@ bool CPLWorkerThreadPool::SubmitJob(CPLThreadFunc pfnFunc, void *pData)
         wt->pInitData = nullptr;
         wt->poTP = this;
         wt->bMarkedAsWaiting = false;
+        //ABELL - Why should this fail? And this is a *pool* thread, not necessarily
```
rouault, line 194: this may fail because the underlying operating system operation, e.g. pthread_create, may fail (lack of OS resources, memory, etc) ...  I'd prefer you revert that part, since at least currently we return in error and the caller might know that something went wrong and return in error itself, whereas with your proposed change the failure is silently ignored.

abellgithub: I don't have any problem printing some sort of an error, but at that point your entire program is probably screwed, the pool should be marked invalid and so on.  I don't think the right answer is to do what was being done before (eliminating the queued jobs). The job queue is essentially independent of the worker thread list and if you're going to keep the pool, the jobs should remain queued.

abellgithub: I think an easier and cleaner design that eliminates this problem is just to create all the worker threads when the pool is Setup(). Then if you have an error, you can fail immediately and you don't have to deal with these error conditions.

abellgithub: I'm happy to change the code as I suggested above, but I'd rather do it in two steps -- first this and then change the worker thread queue.

rouault: > but at that point your entire program is probably screwed

yes probably. If that occurs currently, normally the call chain should error out all along. But in any case, I'd prefer we keep that part as it, in the scope of this PR, which isn't directly affected by that.

> I'm happy to change the code as I suggested above, but I'd rather do it in two steps -- first this and then change the worker thread queue.

yes, better in another PR, although I'm not entirely sure what you've in mind for this next PR (I'll likely have little time in the next 2 weeks to review)

abellgithub: You seem to be clinging hard to a problematic design. I'm trying to improve the code so that it's more useful and easier to maintain. I don't understand the resistance.

If you want I can throw an exception if the thread creation fails, which is what std::thread does, but having to clear the submitted jobs and carry on doesn't seem a good solution.

abellgithub:
@rouault : Thanks for merging this. It will simplify things for me. If you have thoughts on more changes in the spirit of the suggestions above, let me know and I'll try to work on it.

PR #10506, opened 2024-07-25
InterpolateAtPoint near borders

### LINKED ISSUE #10501: InterpolateAtPoint is not continuous at the borders
### What is the bug?

`GDALRasterInterpolateAtPoint` is using the interpolation algorithm from RPC (that suffers the same behavior).
In the borders the interpolations are downgraded from CubicSpline to Bilinear, and from Bilinear to Nearest.
You can see an example here, using the image from https://en.wikipedia.org/wiki/Bicubic_interpolation (see that we use CubicSpline, that is a bit different)
old
In the image there are cross-sections where you can see the discontinuities.

I am working in a PR to "extend" the image in the borders, assuming a constant pixel value, and let the interpolation compute the values.
It should look something like this:
new

As a bonus it will contain **Cubic** interpolation (in addition to CubicSpline) ;)

Request values in the borders of the image (like at [0 - 0.5) for bilinear or [0 - 1.5) for cubicspline)

### Versions and provenance

Tested in Ubuntu 20.04, but happens on every platform. For the RPC interpolation, it is happening the same.

### Additional context

_No response_

jjimenezshaw:
## What does this PR do?
Makes `GDALRasterInterpolateAtPoint` continuous near the borders of the image.
It also adds `cubic` interpolation.
cubic
(why adding cubic here? Unlike cubicspline, cubic matches the values of the grid in the nodes. That makes testing during development easier. See that in the images at x=0.5 and y=0.5, that are cutting at the nodes of the grid).
See that cubic may get values bigger than the maximum of the grid values.

### Implementation
How is it getting data near the borders, if there are no point to interpolate with? It "extends" the image with the pixel value at the border.

For instance, this is how the top left corner is "extended" to have a 4x4 kernel:
```
.___      a a a b
|a b  =>  a a a b
|c i      a a a b
          c c c i
```

Fixes #10501

## Tasklist

 - [x] Add test case(s)
 - [x] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/alg/gdal_interpolateatpoint.cpp
+++ b/alg/gdal_interpolateatpoint.cpp
@@ -31,6 +31,8 @@
 
 #include "gdal_interpolateatpoint.h"
 
+#include "gdalresamplingkernels.h"
+
 #include <algorithm>
 
 static bool
@@ -134,28 +136,6 @@ GDALInterpExtractDEMWindow(GDALRasterBand *pBand,
 #endif
 
     return true;
-}
-
-static double CubicKernel(double dfX)
-{
-    // http://en.wikipedia.org/wiki/Bicubic_interpolation#Bicubic_convolution_algorithm
-    // W(x) formula with a = -0.5 (cubic hermite spline )
-    // or
-    // https://www.cs.utexas.edu/~fussell/courses/cs384g-fall2013/lectures/mitchell/Mitchell.pdf
-    // k(x) (formula 8) with (B,C)=(0,0.5) the Catmull-Rom spline
-    double dfAbsX = fabs(dfX);
-    if (dfAbsX <= 1.0)
-    {
-        double dfX2 = dfX * dfX;
-        return dfX2 * (1.5 * dfAbsX - 2.5) + 1;
-    }
-    else if (dfAbsX <= 2.0)
-    {
-        double dfX2 = dfX * dfX;
-        return dfX2 * (-0.5 * dfAbsX + 2.5) - 4 * dfAbsX + 2;
-    }
-    else
-        return 0.0;
 }
 
 /************************************************************************/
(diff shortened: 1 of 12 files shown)


jjimenezshaw:
fyi @mdsumner @rouault

alg/gdal_interpolateatpoint.cpp
```diff
@@ -136,6 +136,28 @@ GDALInterpExtractDEMWindow(GDALRasterBand *pBand,
     return true;
 }
 
+static double CubicKernel(double dfX)
```
jjimenezshaw, line 139: Should this function be in the header file? Currently is not used anywhere else. It is copied from `alg/gdalwarpkernel.cpp`

rouault: > Should this function be in the header file

we should try to limit copy&paste. Perhaps as a inline function in a alg/gdalresamplingkernels.h header used by gdal_interpolateatpoint.cpp and gdalwarpkernel.cpp (I also considered in the existing alg/gdal_alg_priv.h). Or any equivalent suggestion you may have

jjimenezshaw: done

jjimenezshaw:
There is a persistent failing test `test_vsiaz_write_blockblob_chunk_size_1` https://github.com/OSGeo/gdal/actions/runs/10130627537/job/28012234353
that seems to be networking related. Not with my changes I guess.
Is it fixed on master? should I rebase in any case?

rouault:
> Is it fixed on master?

I've just pushed a commit to skip it on CI. It is indeed quite flaky especially on this macos target.

>  should I rebase in any case?

no, I'll merge it as it

PR #10518, opened 2024-07-28
cpl::contains(): use find() method for implementation to address …

rouault:
…https://github.com/OSGeo/gdal/pull/10464#discussion_r1694287217


dg0yt (approved):
Thanks.
IMO STL wouldn't call it `count` if it couldn't return more than `1`.
And the special GDAL function is the opportunity to choose the right implementation instead of a shortcut.

PR #10543, opened 2024-08-02
HANA: Add support for fast extent estimation

jatsqi:
## What does this PR do?

With QRC/1 2024 a new and faster way to estimate extents of spatial column has been added to HANA.
More precisely, users can now use special columns in [this table](https://help.sap.com/docs/hana-cloud-database/sap-hana-cloud-sap-hana-database-spatial-reference/spatial-extent-with-sys-m-st-geometry-columns?locale=en-US) to get an estimation of the extent of a spatial column.
Even though the stored extent is only an estimation, it is often good enough.
Especially for very large tables costly aggregation functions are avoided every time the extent is recomputed.
The extent information in the mentioned HANA table is only available after a delta-merge, so a fallback to the standard behaviour, i.e. via aggregation functions, was added in case the data is unavailable or has not yet been computed.

None.

## Tasklist

 - [x] Add test case(s)
 - [x] Add documentation (no documentation required, this change is not user-facing)
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/autotest/ogr/ogr_hana.py
+++ b/autotest/ogr/ogr_hana.py
@@ -1221,6 +1221,94 @@ def test_ogr_hana_38():
 
 
 ###############################################################################
+# Test reading a layer extent and verify a working fast extent implementation
+
+
+def test_ogr_hana_39():
+    conn = create_connection()
+
+    def delta_merge():
+        execute_sql(
+            conn,
+            f"MERGE DELTA OF {table_name}",
+        )
+
+    # Create test table
+    layer_name = get_test_name()
+    table_name = f'"{gdaltest.hana_schema_name}"."{layer_name}"'
+    execute_sql(
+        conn,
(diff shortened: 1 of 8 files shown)


rouault:
Python linting issue. Cf https://github.com/OSGeo/gdal/actions/runs/10214824460/job/28264531781?pr=10543

jatsqi:
Ah thanks @rouault, seems like I forgot to run the linter for this.

rouault:
@jatsqi Now that it has been merged, it seems it causes random failures in test_ogr_hana_39(). See https://github.com/OSGeo/gdal/actions/runs/10253377428/job/28365852783?pr=10549 . Can you have a look at that?

jatsqi:
@rouault I'll have a look right away 👍

jatsqi:
@rouault I did some digging and talked with a colleague about this, and it seems that it is not that straightforward to test this, but I'm convinced the implementation itself is correct.
I've opened a PR #10553 for this, so at least the pipeline is no longer failing for new PRs.

PR #10570, opened 2024-08-10
InterpolateAtPoint  complex number implementation

jjimenezshaw:
## What does this PR do?
Implementation of `GDALInterpolateAtPoint` for complex numbers, that was missing.

If templates are not welcome, the last commit can be deleted.

#10461 and #10506

## Tasklist

 - [x] Add test case(s)
 - [ ] Review
 - [ ] Adjust for comments


alg/gdal_interpolateatpoint.cpp
```diff
@@ -91,31 +108,37 @@ GDALInterpExtractDEMWindow(GDALRasterBand *pBand,
...
+                    eDataType, 0, 0, nullptr);
                 if (eErr != CE_None)
                 {
                     return false;
                 }
-                cacheDEM->insert(nKey, poValue);
+                cache->insert(nKey, poValue);
             }
 
+            double *padfAsDouble = reinterpret_cast<double *>(padfOut);
```
jjimenezshaw, line 130: See https://en.cppreference.com/w/cpp/numeric/complex , section "Array-oriented access"

alg/gdal_interpolateatpoint.cpp
```diff
@@ -34,23 +34,40 @@
...
+{
+    return ARE_REAL_EQUAL(dfNoDataValue, dfOut);
+}
+
+template <> bool areEqualReal(double dfNoDataValue, std::complex<double> dfOut)
+{
+    return ARE_REAL_EQUAL(dfNoDataValue, dfOut.real());
+}
+
+template <typename T>
```
rouault, line 51: given the "double *padfAsDouble = reinterpret_cast<double *>(padfOut)", it would be prudent to at least add a comment that T should be double or ``std::complex<double>`` (perhaps some std::enable_if_t magic or static_assert could enforce that, but a simple code comment would do)

jjimenezshaw: Added a comment. In addition, `areEqualReal` should fail because it is not defined for other types.

rouault:
Looks good to me and draft status could be removed

PR #10607, opened 2024-08-15
Add a 'make spelling' target to doc/

rouault:
- Fix various typos found by 'make spelling'
- Add a 'make spelling' target to doc/
- CI: enable 'make spelling'

@dbaston This may interfere with PR #10434 , especially since PR #10434 removes doc_build.yml which is modified by this PR to run 'make spelling' . How in the new setup 'make spelling' should be run?


dbaston:
> How in the new setup 'make spelling' should be run?

We could add it as a step to `readthedocs.yaml` but it seems like it takes a while to run, so maybe it's better kept running independently through GitHub Actions?

rouault:
> so maybe it's better kept running independently through GitHub Actions?

sounds good to me

PR #10662, opened 2024-08-27
OpenFileGDB writer: fix writing a Int32 field with value -21121

rouault:


nyalldawson (approved):

geographika:
For future reference - discussions about this at https://lists.osgeo.org/pipermail/gdal-dev/2024-August/059357.html

PR #10678, opened 2024-08-29
OpenFileGDB/filegdbtable.cpp: Use comments for Python code snippets in C/C++ instead of conditional compilation statements to prevent choking pre-processor

gorloffslava:
## What does this PR do?

Currently, [filegdbtable.cpp#L2734](https://github.com/OSGeo/gdal/blob/master/ogr/ogrsf_frmts/openfilegdb/filegdbtable.cpp#L2734) uses `#if 0 ... #endif` construct to comment out the Python code snippet. 

In many cases it works fine, but with some sets of pre-processor defines/macros, it chokes on this snippet as trying to expand some parts of that (we weren't able to narrow down that to the minimal sets of flags as fixing that is simply easier).

This tiny PR replaces `#if 0 ... #endif` construct with usual C/C++ comments. It shouldn't/doesn't affect the code compilation where it worked previously but should prevent pre-processor from erroring in all cases there. 

Please, find the errors log snippet attached.
[logs.txt](https://github.com/user-attachments/files/16800661/logs.txt)

### DIFF
--- a/ogr/ogrsf_frmts/openfilegdb/filegdbtable.cpp
+++ b/ogr/ogrsf_frmts/openfilegdb/filegdbtable.cpp
@@ -2730,55 +2730,8 @@ void FileGDBTable::GetMinMaxProjYForSpatialIndex(double &dfYMin,
         return;
     double dfMinLat;
     double dfMaxLat;
-    // Determined through experimentation, e.g with the following script
-    // ```python
-    // from osgeo import gdal, ogr, osr
-    // import struct
-    // gdal.RmdirRecursive('test.gdb')
-    // ds = ogr.GetDriverByName('FileGDB').CreateDataSource('test.gdb')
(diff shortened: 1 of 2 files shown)


rouault:
@gorloffslava I would prefer that you would move that Python code in a ``ogr/ogrsf_frmts/openfilegdb/find_srs_latitude_limits.py`` file, and mentions it as a C++ comment to replace this #if 0 / #endi section

gorloffslava:
@rouault done!

rouault:
@gorloffslava 

Can you please:
```
pip install pre-commit
pre-commit install
pre-commit run --files ogr/ogrsf_frmts/openfilegdb/find_srs_latitude_limits.py
```
and commit the updated file

BwL1289:
@rouault done

PR #10691, opened 2024-08-30
frmts/netcdf: Fix build failure with upstream netcdf caused by _FillValue macro renaming

gorloffslava:
## What does this PR do?
Starting [[c26f7ea](https://github.com/Unidata/netcdf-c/commit/c26f7eabf4a1cd25353f22734f439505fe636a45)](https://github.com/Unidata/netcdf-c/commit/c26f7eabf4a1cd25353f22734f439505fe636a45), netcdf-c exposes the `NC_FillValue` macro instead of `_FillValue` to avoid collisions with [[C++ standard library](https://github.com/Unidata/netcdf-c/issues/2858)](https://github.com/Unidata/netcdf-c/issues/2858). However, the previous macro, `_FillValue`, was fully removed causing netcdf-c consumers, including GDAL, fail to build.

It's unlikely that this naming change will be backported to the previous netcdf-c releases, so we have to account for both macros variants. We do so by introducing our own macro, `NCDF_FillValue`, and using that in places where `_FillValue` was previously used. If `NC_FillValue` is defined by `netcdf.h`, `NCDF_FillValue` expands to it and, if it's not, to `_FillValue`.

I was able to successfully build the GDAL against netcdf-c both with:
1. `_FillValue`: v4.9.2 (from GitHub sources, Amazon Linux 2023's dnf, and macOS homebrew)
2. `NC_FillValue`: v4.9.3dev (the latest commit from the `main`, `eda184e6af63d6eee1b3a59c61d1695eef44fcb4`).

### DIFF
--- a/frmts/netcdf/netcdfdataset.cpp
+++ b/frmts/netcdf/netcdfdataset.cpp
@@ -452,10 +452,10 @@ netCDFRasterBand::netCDFRasterBand(const netCDFRasterBand::CONSTRUCTOR_OPEN &,
     const char *pszNoValueName = nullptr;
 
     // Find attribute name, either _FillValue or missing_value.
-    int status = nc_inq_att(cdfid, nZId, _FillValue, &atttype, &attlen);
+    int status = nc_inq_att(cdfid, nZId, NCDF_FillValue, &atttype, &attlen);
     if (status == NC_NOERR)
     {
-        pszNoValueName = _FillValue;
+        pszNoValueName = NCDF_FillValue;
     }
     else
     {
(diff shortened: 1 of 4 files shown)


rouault:
I've cherry-picked this in 93d545a9173fd17c414bb11b2194c12d1966760e by just applying pre-commit code reformatting

rouault:
cherry-picked to 3.9 per e0342f516855c7bea378f356937c110dcf333614

BwL1289:
Thank you @rouault.

gorloffslava:
@rouault thank you!

Issue #8440, opened 2023-09-20
[meta-ticket] GDAL 4.0 potential changes related to API or behaviour breaks


rouault:
List of tickets tagged with 4.0 milestone: https://github.com/OSGeo/gdal/milestone/33

Below are just proposed ideas. Nothing decided

API breakage:

* C and C++ API:
- [ ]    #8288,  #8396, #8399

* C++ API (mostly):
- [ ] OGRLayer::GetNextFeature() and GetFeature() to return a ``std::unique_ptr<OGRFeature>``: involves lots of internal changes, affect C++ users
- [ ] or change OGRLayer::GetNextFeature() and GetFeature() to accept a ``OGRFeature&`` that they update. May help a bit performance in bulk reading operations, but probably not spectacular
- [ ] Remove reference count mechanism from OGRSpatialReference class itself and use ``std::shared_ptr<OGRSpatialReference>`` (or ``std::shared_ptr<const OGRSpatialReference>`` ?) instead : involves lots of internal changes, affect C++ users of the geometry and layer classes, although we might have to propagate it everywhere since OGRSpatialReferenceH would be an opaque type for ``std::shared_ptr<OGRSpatialReference>``
- [ ] Remove reference count mechanism from OGRFeatureDefn class itself and use ``std::shared_ptr<OGRFeatureDefn>`` instead: involves lots of internal changes, affect C++ users, may impact slightly the C API (OGR_FD_GetReferenceCount)

API behaviour changes:

- [ ]  #4681
- [ ]  #7452

Driver changes:

- [ ]  #4712

Driver removals?

Others :

- [ ] Changes in error handling mechanism (to be defined what changes would be desirable) ? Cf https://lists.osgeo.org/pipermail/gdal-dev/2022-July/056066.html

tbonfort:
- [ ] GTIFF: used TILED=YES by default
- [ ] GTIFF: use COMPRESS=LZW by default

elpaso:
- [ ]  Design and implement a unified way to describe a datasource with open options and subdatasets with a string, see the discussion here: https://lists.osgeo.org/pipermail/gdal-dev/2023-September/057641.html. I'm not sure it would be an API break though.

jjimenezshaw:
>  GTIFF: used TILED=YES by default

About this default without overviews, it can be really frustrating.
[...]

PR #10700, opened 2024-08-31
[Optim] OGRSQL: do not query geometry column(s) when not needed

rouault:
With this optimization, ``ogrinfo nz-building-outlines.parquet -sql "select stddev_pop(building_id) from \"nz-building-outlines\""`` goes from 2.8 s to 0.8 s

### DIFF
--- a/ogr/ogrsf_frmts/generic/ogr_gensql.cpp
+++ b/ogr/ogrsf_frmts/generic/ogr_gensql.cpp
@@ -2543,14 +2543,47 @@ int OGRGenSQLResultsLayer::Compare(const OGRField *pasFirstTuple,
 void OGRGenSQLResultsLayer::AddFieldDefnToSet(int iTable, int iColumn,
                                               CPLHashSet *hSet)
 {
-    if (iTable != -1 && iColumn != -1)
+    if (iTable != -1)
     {
         OGRLayer *poLayer = m_apoTableLayers[iTable];
-        if (iColumn < poLayer->GetLayerDefn()->GetFieldCount())
+        const auto poLayerDefn = poLayer->GetLayerDefn();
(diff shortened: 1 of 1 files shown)


jratike80:
From users point of view, is it better/faster to use this enhanced OGR SQL dialect instead of SQLite dialect that does not select the geometry automatically? What happens if user is using the OGR SQL with ogr2ogr instead of ogrinfo?

`ogr2ogr -sql "select stddev_pop(building_id) from \"nz-building-outlines\"" out.gpkg nz-building-outlines.parquet `

How does GDAL know when the geometry column is not needed and when it is just not used in the query because traditionally OGR SQL selects the geometry automatically?

Could we change the behavior of the OGR SQL dialect in GDAL version 4 and require explicit selection of the geometry field?

rouault:
> From users point of view, is it better/faster to use this enhanced OGR SQL dialect instead of SQLite dialect that does not select the geometry automatically?

yes, this is significantly slower through the SQLite dialect (we cannot automatically detect which columns are needed in advance, to ignore them, and even if we did, that would still be slower)

> What happens if user is using the OGR SQL with ogr2ogr instead of ogrinfo?

like ogrinfo, it will output one single feature with the stddev value and no geometry

> How does GDAL know when the geometry column is not needed and when it is just not used in the query because traditionally OGR SQL selects the geometry automatically?

Well, the code I edited know all fields, regular and geometry, that are needed in the select clause or in the where clause, so things should work just fine

> Could we change the behavior of the OGR SQL dialect in GDAL version 4 and require explicit selection of the geometry field?

Something to consider if GDAL 4 happens... (I've just added it to https://github.com/OSGeo/gdal/issues/8440)

rouault:
> yes, this is significantly slower through the SQLite dialect

5.2 s

PR #10713, opened 2024-09-01
Add OGR_G_GeodesicLength() and OGRCurve/OGRSurface/OGRGeometryCollection::get_GeodesicLength()

rouault:
and make OGR_G_GetLength() work on surfaces, suming the length of their exterior and interior rings.

and SQLite/GPKG: Add ST_Length(geom, use_ellipsoid)

### DIFF
--- a/ogr/ogr_api.cpp
+++ b/ogr/ogr_api.cpp
@@ -1604,6 +1604,8 @@ OGRErr OGR_G_RemoveGeometry(OGRGeometryH hGeom, int iGeom, int bDelete)
  * \brief Compute length of a geometry.
  *
  * Computes the length for OGRCurve or MultiCurve objects.
+ * For surfaces, compute the sum of the lengths of their exterior
+ * and interior rings (since 3.10).
  * Undefined for all other geometry types (returns zero).
  *
  * This function utilizes the C++ get_Length() method.
@@ -1612,6 +1614,9 @@ OGRErr OGR_G_RemoveGeometry(OGRGeometryH hGeom, int iGeom, int bDelete)
  * @return the length or 0.0 for unsupported geometry types.
  *
  * @since OGR 1.8.0
+ *
+ * @see OGR_G_GeodesicLength() for an alternative method returning lengths
+ * computed on the ellipsoid, and in meters.
  */
 
 double OGR_G_Length(OGRGeometryH hGeom)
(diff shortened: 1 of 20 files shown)


jratike80:
Maybe it could be mentioned somewhere that the algorithm takes a shortest route, and it can follow the other side of the Earth than what user may believe? I am not sure if this can make confusion, but sometimes users test with simple unrealistic geometries, like global polygon bbox(-180,-90,180,90)?

ogr/ogr_api.cpp
```diff
@@ -1643,6 +1651,61 @@ double OGR_G_Length(OGRGeometryH hGeom)
...
+double OGR_G_GeodesicLength(OGRGeometryH hGeom)
+
+{
+    VALIDATE_POINTER1(hGeom, "OGR_G_GeodesicLength", -1);
+
+    double dfLength = 0.0;
+
+    const auto poGeom = ToPointer(hGeom);
+    const OGRwkbGeometryType eType =
+        wkbFlatten(ToPointer(hGeom)->getGeometryType());
```
elpaso, line 1685:
```suggestion
        wkbFlatten(poGeom)->getGeometryType());
```

ogr/ogrgeometrycollection.cpp
```diff
@@ -1288,6 +1291,78 @@ double OGRGeometryCollection::get_GeodesicArea(
...
+ * @see get_Length() for an alternative method returning areas computed in
+ * 2D Cartesian space.
+ *
+ * @since GDAL 3.10
+ */
+double OGRGeometryCollection::get_GeodesicLength(
+    const OGRSpatialReference *poSRSOverride) const
+{
+    if (!poSRSOverride)
+        poSRSOverride = getSpatialReference();
```
elpaso, line 1327: For clarity, worth checking for  poSRSOverride not nullptr? I see it's not strictly necessary because it will be checked on the individual items.

elpaso (approved):

rouault:
> Maybe it could be mentioned somewhere that the algorithm takes a shortest route, and it can follow the other side of the Earth than what user may believe? I am not sure if this can make confusion, but sometimes users test with simple unrealistic geometries, like global polygon bbox(-180,-90,180,90)?

I've added:
```
 *
 * <a href="https://geographiclib.sourceforge.io/html/python/geodesics.html">Geodesics</a>
 * follow the shortest route on the surface of the ellipsoid.
```

ogr/ogr_api.cpp
rouault: done

ogr/ogrgeometrycollection.cpp
rouault: removed

PR #10730, opened 2024-09-04
/vsicurl/: honor 'Cache-Control: no-cache' header

rouault:


port/cpl_vsil_curl.cpp
```diff
@@ -1374,7 +1373,17 @@ vsi_l_offset VSICurlHandle::GetFileSizeOrHeaders(bool bSetError,
...
+                        if (bGetHeaders)
+                        {
+                            m_aosHeaders.SetNameValue(pszKey, pszValue);
+                        }
+                        if (EQUAL(pszKey, "Cache-Control") &&
+                            EQUAL(pszValue, "no-cache") &&
+                            CPLTestBool(CPLGetConfigOption(
+                                "CPL_VSIL_CURL_HONOR_CACHE_CONTROL", "YES")))
+                        {
+                            m_bCached = false;
```
sgillies, line 1385: @rouault does this mean that the results are not cached at all? Or that cached results aren't used without first validating with the server?

sgillies (approved):

port/cpl_vsil_curl.cpp
rouault: It checks if the server returns a 'Cache-Control: no-cache' response header. If it does, then, *when the file is closed*, any cached data about that file is invalidated. Which is the behaviour you could already get with the [CPL_VSIL_CURL_NON_CACHED](https://gdal.org/en/latest/user/configoptions.html#config-CPL_VSIL_CURL_NON_CACHED) config option.

PR #10731, opened 2024-09-04
gdal_contour: add a -gt option to define the transaction flush interval (default it to 100,000)

### LINKED ISSUE #10729: gdal_contour slow performance with GPKG output
### What is the bug?

Using GeoPackage as the output format when producing contour lines with gdal_contour results in far superior processing times (over 10 times) compared to using ESRI Shapefile as the output format.

My tests were made using this large [DEM](https://diffusion.mffp.gouv.qc.ca/Diffusion/DonneeGratuite/Foret/IMAGERIE/Produits_derives_LiDAR/21M/21M16NO/MNT_21M16NO.tif)

1. Run `gdal_contour -i 10 -a ELEV MNT_21M16NO.tif contour.shp` and check time (about 20 seconds for me)
2. Run `gdal_contour -i 10 -a ELEV MNT_21M16NO.tif contour.gpkg` and check time (about 410 seconds for me)

Both results are identical and each process consumes the same amount of RAM. However, I see far less CPU usage when the output is GPKG than with SHP.

### Versions and provenance

GDAL 3.9.2 (from OSGeo4W for installing QGIS 3.38.2)
Windows 10 Enterprise 22H2

### Additional context

_No response_

rouault:
Fixes #10729

### DIFF
--- a/apps/gdal_contour.cpp
+++ b/apps/gdal_contour.cpp
@@ -190,7 +190,7 @@ GDALContourAppOptionsGetParser(GDALContourOptions *psOptions)
                 else
                     psOptions->nGroupTransactions = atoi(s.c_str());
             })
-        .help(_("Group <n> features per transaction "));
+        .help(_("Group <n> features per transaction."));
 
     argParser->add_quiet_argument(&psOptions->bQuiet);
 
--- a/apps/gdal_contour.cpp
+++ b/apps/gdal_contour.cpp
@@ -64,6 +64,7 @@ struct GDALContourOptions
     bool bQuiet = false;
     std::string aosDestFilename;
     std::string aosSrcFilename;
+    GIntBig nGroupTransactions = 100 * 1000;
 };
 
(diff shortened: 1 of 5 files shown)


apps/gdal_contour.cpp
```diff
@@ -179,6 +180,18 @@ GDALContourAppOptionsGetParser(GDALContourOptions *psOptions)
...
+        .metavar("<n>|unlimited")
+        .action(
+            [psOptions](const std::string &s)
+            {
+                if (EQUAL(s.c_str(), "unlimited"))
+                    psOptions->nGroupTransactions = -1;
+                else
+                    psOptions->nGroupTransactions = atoi(s.c_str());
+            })
+        .help(_("Group <n> features per transaction "));
```
elpaso, line 193:
```suggestion
        .help(_("Group <n> features per transaction."));
```

elpaso (approved):

PR #10738, opened 2024-09-05
autotest: avoid deprecation warnings related to DataSource.Release()

rouault:


autotest/ogr/ogr_refcount.py
```diff
@@ -86,12 +86,10 @@ def test_ogr_refcount_2():
 def test_ogr_refcount_3():
```
dbaston, line 86: Marking the test as follows should squelch the warning while retaining coverage on `Release`:

```
@pytest.mark.filterwarnings("ignore::DeprecationWarning")
```

rouault: good suggestion, done

PR #10739, opened 2024-09-05
OGRGeometry classes cleanups and optimizations

rouault:
- OGRTriangulatedSurface::operator=(): simplify by making base operator to call empty()
- Tag OGRGeometry derived classes as final or non-final
- OGR geometry classes: mark default constructors directly in .h and removed useless overridden destructor for better code generation

The latter is interesting

For example before the default OGRMultiPolygon constructor would do essentially nothing but calling the OGRMultiSurface default constructor which would do essentially nothing but calling the default OGRGeometryCollection constructor:

```
00000000005ca440 <OGRMultiPolygon::OGRMultiPolygon()>:
  5ca440:       f3 0f 1e fa             endbr64 
  5ca444:       53                      push   %rbx
  5ca445:       48 89 fb                mov    %rdi,%rbx
  5ca448:       e8 33 1e d3 ff          callq  2fc280 <OGRMultiSurface::OGRMultiSurface()@plt>
  5ca44d:       48 8b 05 e4 be 0f 01    mov    0x10fbee4(%rip),%rax        # 16c6338 <vtable for OGRMultiPolygon@@Base-0x1a020>
  5ca454:       48 83 c0 10             add    $0x10,%rax
  5ca458:       48 89 03                mov    %rax,(%rbx)
  5ca45b:       5b                      pop    %rbx
  5ca45c:       c3                      retq   
  5ca45d:       90                      nop
  5ca45e:       66 90                   xchg   %ax,%ax

00000000005d5280 <OGRMultiSurface::OGRMultiSurface()>:
  5d5280:       f3 0f 1e fa             endbr64 
  5d5284:       53                      push   %rbx
  5d5285:       48 89 fb                mov    %rdi,%rbx
  5d5288:       e8 f3 66 d2 ff          callq  2fb980 <OGRGeometryCollection::OGRGeometryCollection()@plt>
  5d528d:       48 8b 05 74 11 0f 01    mov    0x10f1174(%rip),%rax        # 16c6408 <vtable for OGRMultiSurface@@Base-0x1b7a0>
  5d5294:       48 83 c0 10             add    $0x10,%rax
  5d5298:       48 89 03                mov    %rax,(%rbx)
  5d529b:       5b                      pop    %rbx
  5d529c:       c3                      retq   
  5d529d:       90                      nop
  5d529e:       66 90                   xchg   %ax,%ax
```

With the change, this method is no longer even defined. Calls to ``OGRGeometryCollection::OGRGeometryCollection()@plt`` are directly inlined

Similarly for the destructor (which is still defined), but called directly OGRGeometryCollection destructor, instead of going through OGRMultiSurface.


rouault:
@dbaston if you have a chance to review this. Not totally sure all items in this PR are good ideas (the 'final' bit might be risky if external code would derive OGR geometry classes, which I'm not aware of, which doesn't mean anything...)

dbaston:
Added some suggestions in https://github.com/rouault/gdal/pull/60. I would share your hesitation on marking classes final unless there's a benchmark showing a benefit (presumably from devirtualization?)

Feel free to email or send a message on Slack if I don't respond to mentions, since I watch the repository I get all messages which means the ones directed to me are sometimes missed...

rouault:
> Added some suggestions in https://github.com/rouault/gdal/pull/60.

Merged. Thanks

> I  would share your hesitation on marking classes final 

just reverted that commit https://github.com/OSGeo/gdal/pull/10739/commits/2c9f23a3ff72f19ef25f3bc8e70f57c278c07b13

PR #10742, opened 2024-09-06
netCDF driver: simplify identification logic by just checking runtime  availability of HDF4/HDF5 drivers

rouault:
... ignoring their compile-time availability, which makes it more robust in the case of plugin-only builds.

Also remove the GDAL_HAS_HDF4 and GDAL_HAS_HDF5 metadata items from the metadata of the netCDF driver. They are redundant with checking for GDALGetDriverByName("HDF4" / "HDF5") != nullptr.

CC @schwehr


frmts/netcdf/netcdfdrivercore.cpp
```diff
@@ -493,12 +489,7 @@ void netCDFDriverSetCommonMetadata(GDALDriver *poDriver)
 #ifdef NETCDF_HAS_HDF4
     poDriver->SetMetadataItem("NETCDF_HAS_HDF4", "YES");
 #endif
-#ifdef HAVE_HDF4
-    poDriver->SetMetadataItem("GDAL_HAS_HDF4", "YES");
-#endif
-#ifdef HAVE_HDF5
-    poDriver->SetMetadataItem("GDAL_HAS_HDF5", "YES");
-#endif
```
schwehr, line 501: Wondering if there is anything counting on these being there or absent.

rouault: not in GDAL at least: https://github.com/search?q=repo%3AOSGeo%2Fgdal%20GDAL_HAS_HDF5&type=code
and not in any Debian source code : https://codesearch.debian.net/search?q=GDAL_HAS_HDF5

rouault: For completness, adding this in the commit message
"Also remove the GDAL_HAS_HDF4 and GDAL_HAS_HDF5 metadata items
from the metadata of the netCDF driver. They are redundant with checking
for GDALGetDriverByName("HDF4" / "HDF5") != nullptr."

PR #10744, opened 2024-09-07
gdal::VectorX Prototype

jjimenezshaw:
## What does this PR do?
During my work on the file `alg/gdal_interpolateatpoint.cpp`, I saw that there was a repetitive pattern: doing everything on X and Y. Both as integer to manage the pixel, and as double to use the coordinates in the image.

I missed a functionality where x,y (or even x,y,z) could be used as an object, and do some basic operations on that, like subtract another point, multiply by a factor, etc. Something I am used to with Eigen library.

Once that development was finished I tried something, to see if it makes sense. This is it.

The main feature is the template class `VectorX`. Both the size and type are templated. You can make it double, int, even complex<double>. Size can be any, but I think the most common would be 2, maybe 3. Four predefined types are included: `Vector2i`, `Vector3i`, `Vector2d`, `Vector3d`

This PR is using it only in `alg/gdal_interpolateatpoint.cpp`.

**The idea is to use it internally. Not in any public API.**

It should make the code more readable, and I hope it will help to not forget the Y or the X during developments.

What do you think? Is it useful? In case it is included, I don't expect a massive migration, but just start using it when it is convenient.

#10461, #10506, #10570

## Tasklist

 - [x] Add test case(s)
 - [x] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/alg/gdal_interpolateatpoint.cpp
+++ b/alg/gdal_interpolateatpoint.cpp
@@ -17,6 +17,8 @@
 
 #include "gdalresamplingkernels.h"
 
+#include "gdal_vectorx.h"
+
 #include <algorithm>
 #include <complex>
 
@@ -36,10 +38,15 @@ template <> bool areEqualReal(double dfNoDataValue, std::complex<double> dfOut)
 template <typename T>
 bool GDALInterpExtractValuesWindow(GDALRasterBand *pBand,
                                    std::unique_ptr<DoublePointsCache> &cache,
-                                   int nX, int nY, int nWidth, int nHeight,
-                                   T *padfOut)
+                                   gdal::Vector2i point,
+                                   gdal::Vector2i dimensions, T *padfOut)
 {
     constexpr int BLOCK_SIZE = 64;
+
+    const int nX = point.x();
+    const int nY = point.y();
+    const int nWidth = dimensions.x();
+    const int nHeight = dimensions.y();
 
     // Request the DEM by blocks of BLOCK_SIZE * BLOCK_SIZE and put them
     // in cache
@@ -153,29 +160,32 @@ bool GDALInterpolateAtPointImpl(GDALRasterBand *pBand,
                                 std::unique_ptr<DoublePointsCache> &cache,
                                 const double dfXIn, const double dfYIn, T &out)
 {
-    const int nRasterXSize = pBand->GetXSize();
-    const int nRasterYSize = pBand->GetYSize();
+    const gdal::Vector2i rasterSize{pBand->GetXSize(), pBand->GetYSize()};
+    const gdal::Vector2d inLoc{dfXIn, dfYIn};
+
     int bGotNoDataValue = FALSE;
     const double dfNoDataValue = pBand->GetNoDataValue(&bGotNoDataValue);
 
-    if (dfXIn < 0 || dfXIn > nRasterXSize || dfYIn < 0 || dfYIn > nRasterYSize)
+    if (inLoc.x() < 0 || inLoc.x() > rasterSize.x() || inLoc.y() < 0 ||
+        inLoc.y() > rasterSize.y())
     {
         return FALSE;
     }
 
     // Downgrade the interpolation algorithm if the image is too small
-    if ((nRasterXSize < 4 || nRasterYSize < 4) &&
+    if ((rasterSize.x() < 4 || rasterSize.y() < 4) &&
         (eResampleAlg == GRIORA_CubicSpline || eResampleAlg == GRIORA_Cubic))
     {
         eResampleAlg = GRIORA_Bilinear;
     }
-    if ((nRasterXSize < 2 || nRasterYSize < 2) &&
+    if ((rasterSize.x() < 2 || rasterSize.y() < 2) &&
         eResampleAlg == GRIORA_Bilinear)
     {
         eResampleAlg = GRIORA_NearestNeighbour;
     }
 
-    auto outOfBorderCorrection = [](int dNew, int nRasterSize, int nKernelsize)
+    auto outOfBorderCorrectionSimple =
+        [](int dNew, int nRasterSize, int nKernelsize)
     {
         int dOutOfBorder = 0;
         if (dNew < 0)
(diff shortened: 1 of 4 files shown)


alg/gdal_point_templ.h
```diff
@@ -0,0 +1,263 @@
...
+    constexpr size_type size() const noexcept
+    {
+        return N;
+    }
+
+    RawPoint()
+    {
+    }
+
+    RawPoint(T x, T y)
```
rouault, line 57: I'm not the best expert on advanced C++ features, but I guess you'd need something like a constructor with a variadic number of arguments: https://stackoverflow.com/questions/28866559/writing-variadic-template-constructor
And you probably want to ensure that the number of items passed is exactly N

alg/gdal_point_templ.h
```diff
@@ -0,0 +1,263 @@
...
+        static_assert(N >= 2, "Invalid constructor params");
+        _values[0] = x;
+        _values[1] = y;
+    }
+
+    RawPoint(const std::array<T, N> &l) : _values{l}
+    {
+    }
+
+    std::array<T, N> array() const
```
rouault, line 68:
```suggestion
    const std::array<T, N>& array() const
```

alg/gdal_point_templ.h
```diff
@@ -0,0 +1,263 @@
+/******************************************************************************
+ * $Id$
+ *
```
rouault, line 3:
```suggestion
```

rouault (commented):
For testing, I assume you should be able to just include this .h file in a autotest/cpp/test_gdal_point_templ.cpp file added to gdal_unit_test

jjimenezshaw:
> What is "raw" about this point? 

TBH I do not remember. Maybe the point in the image coordinates, not in the CRS. But it should not be restricted to it. I am not happy with the name of the class, as I said in the PR description.  I didn't call it directly `Point` because it could be too generic; or `Vector` that is even more generic. Maybe just `Point` is a better idea. Suggestions are welcome.

> Are there use cases other than the one here?

I hope so. Any algorithm that uses the coordinates of a point in the raster; or even in a geometry. This is a prototype to see if it makes sense (I didn't want to the class without any usecase). Once @rouault said it could be useful, I will improve it, with better documentation and tests.
My intention is not to use it everywhere now. Just when the developer finds it useful. I find more manageable a point as an object, and not having x and y separated in two variables.

@dbaston this is the PR I mentioned in the meeting.

alg/gdal_vectorx_templ.h
```diff
@@ -0,0 +1,256 @@
...
+ * The above copyright notice and this permission notice shall be included
+ * in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+ * OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+ * THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
```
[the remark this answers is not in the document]
jjimenezshaw: done

alg/gdal_interpolateatpoint.cpp
```diff
@@ -17,6 +17,8 @@
 
 #include "gdalresamplingkernels.h"
 
+#include "gdal_vectorx_templ.h"
```
rouault, line 20: what about shortening that to "gdal_vectorx.h" ? The template thing is just an implementation detail that a user using gdal::Vector2i wouldn't notice

abellgithub:
If you're wanting this, why not just use Eigen or similar?

alg/gdal_interpolateatpoint.cpp
jjimenezshaw: Done.

BTW, if there is a better location for this file than the folder `alg`, I can move it.

jjimenezshaw:
> If you're wanting this, why not just use Eigen or similar?

Adding a dependency like Eigen for such a small thing I think is not worth it.

alg/gdal_interpolateatpoint.cpp
rouault: Would probably be better in gcore/

rouault:
> Adding a dependency like Eigen for such a small thing I think is not worth it.

yes, the aim is not to do advanced linear algebra, but just to transport simple coordinate tuples.  So perhaps that should be gdal::Tuple ?

rouault:
> It is more than just a tuple.

ok, gdal::VectorX is fine then. (ignore the Alpine failures. They are linked to package upgrades on their side)

alg/gdal_vectorx.h
```diff
@@ -0,0 +1,290 @@
...
+ ******************************************************************************
+ * Copyright (c) 2024, Javier Jimenez Shaw
+ *
+ * SPDX-License-Identifier: MIT
+ ****************************************************************************/
+
+#ifndef GDAL_VECTORX_H_INCLUDED
+#define GDAL_VECTORX_H_INCLUDED
+
+/*! @cond Doxygen_Suppress */
```
jjimenezshaw, line 15: I think this is a leftover of a copy-paste. Should this class be in Doxygen? I think so.

rouault: As we don't install the file, this is debatable. There should be at least one warning in the class introduction that the file is for GDAL internal use for now

jjimenezshaw:
> @jjimenezshaw anything left on your side before merging?

I will add tomorrow the comment about the usage

jjimenezshaw:
> @jjimenezshaw anything left on your side before merging?

I think I am fine.

PR #10761, opened 2024-09-10
Overview: fix nearest and mode resamplings to be exact with all data types

### LINKED ISSUE #10758: Overviews for UInt32 have wrong pixel values with Nearest Neighbour resampling
### What is the bug?

Hello.

We are experiencing a problem with overviews generated for UInt32 rasters. It seems like it is a bug in GDAL's resampling of the raster. An example of some sampled points from the full ortho and overview images from a test script:

```
Full: 306783378, Overview 0: 306783392, Overview 1: 306783392, Overview 2: 306783392, Overview 3: 306783392, Overview 4: 0, Overview 5: 0, Overview 6: 0
Full: 306783378, Overview 0: 306783392, Overview 1: 306783392, Overview 2: 306783392, Overview 3: 306783392, Overview 4: 0, Overview 5: 0, Overview 6: 0
Full: 306783378, Overview 0: 306783392, Overview 1: 306783392, Overview 2: 306783392, Overview 3: 306783392, Overview 4: 0, Overview 5: 0, Overview 6: 0
```

Notice the value going from 306783378 to 306783392 in the overviews. The problem arises with both Nearest Neighbour and Mode resampling, and also happens with no compression for both full raster and overviews.

The GTiff is generated with the following Python code:

```python
vrt_output_path = os.path.join(self.temp_dir, "ortho.vrt")
vrt = gdal.BuildVRT(vrt_output_path, dataset_file_paths)
vrt = None

gdal.UseExceptions()
creation_options = ["COMPRESS=LZW", "PREDICTOR=YES", "BIGTIFF=YES", "NUM_THREADS=ALL_CPUS", "RESAMPLING=NEAREST",
                    "OVERVIEW_RESAMPLING=NEAREST", "WARP_RESAMPLING=NEAREST", "OVERVIEWS=IGNORE_EXISTING"]
warp_options = gdal.WarpOptions(dstSRS="EPSG:3857", format="COG", multithread=True, creationOptions=creation_options, outputType=gdal.GDT_UInt32)
output_file_name = os.path.join(self.temp_dir, "ortho_cog.tif")
gdal.Warp(output_file_name, vrt_output_path, options=warp_options)
```

I have tried generating overviews with `gdaladdo` also, which results in the same problem.

It only seems to behave this way for pixels that have a value larger than what is supported by UInt16 (just a gut feeling). Is there some conversion step in the resampling that might cause this error?

I'm intending on digging down into the GDAL source code and debugging it, to try and fix it. Any pointers to where i should start?

1. Create a GeoTiff raster with UInt32 datatype and have pixel values above 2^16.
2. Create overviews (e.g. 2, 4, 8, 16)
3. Compare full sized raster pixel values with overview pixel values.

Sample script (using rasterio) to test - same problem when checking pixel values in QGIS:

```python
import rasterio
import numpy as np

tiff_path = "<path to a Uint32 tiff>"

with rasterio.open(tiff_path, mode="r") as ds:
    sample_point_x = np.linspace(0, ds.width - 1, num=1000, dtype=np.uint32)
    sample_point_y = np.linspace(0, ds.height - 1, num=1000, dtype=np.uint32)

    overviews = ds.overviews(1)
    full_band = ds.read(1)
    overview_bands = [(ds.read(1, out_shape=(ds.height // overview_level, ds.width // overview_level)), overview_level) for overview_level in overviews]

    for x in sample_point_x:
        for y in sample_point_y:
            full_value = full_band[y, x]
            overview_values = [band[min(int(round(y / level)), band.shape[0] - 1), min(int(round(x / level)), band.shape[1] - 1)] for band, level in overview_bands]
            all_equal = all([value == full_value for value in overview_values])
            if all_equal:
                continue
            print_string = f"Full: {full_value}"
            for i, value in enumerate(overview_values):
                print_string += f", Overview {i}: {value}"
            print(print_string)
```

Minimal reproducible example:
```python
import numpy as np
import rasterio
import numpy
import osgeo.gdal as gdal
import os
import tempfile

temp_dir = tempfile.TemporaryDirectory()
output_tiff_path = os.path.join(temp_dir.name, "temp.tif")

with rasterio.open(output_tiff_path, mode="w", driver="GTiff", width=64, height=64, count=1, dtype=numpy.uint32) as ds:
    data = np.zeros((64, 64), dtype=np.uint32)
    data[32, 32] = 0xffffff # values below 0xffffff works fine. Problems seems to happen above 0xfffffe
    ds.write(data, 1)

cog_driver = gdal.GetDriverByName("COG")
gdal_ds = gdal.Open(output_tiff_path)
gdal_ds.BuildOverviews("NEAREST", [2, 4, 8, 16])
gdal_ds = None

with rasterio.open(output_tiff_path, mode="r") as ds:
    overviews = ds.overviews(1)
    full_band = ds.read(1)
    overview_bands = [(ds.read(1, out_shape=(ds.height // overview_level, ds.width // overview_level)), overview_level) for overview_level in overviews]

    full_value = full_band[32, 32]
    overview_values = [
        band[min(int(round(32 / level)), band.shape[0] - 1), min(int(round(32 / level)), band.shape[1] - 1)] for
        band, level in overview_bands]
    print_string = f"Full: {full_value}"
    for i, value in enumerate(overview_values):
        print_string += f", Overview {i}: {value}"
    print(print_string)
```

Outputs:
`Full: 16777215, Overview 0: 16777216, Overview 1: 16777216, Overview 2: 16777216, Overview 3: 16777216`

### Versions and provenance

Tested on both Windows 11 GDAL 3.7.3, released 2023/10/30 and Windows 11 GDAL 3.9.2, released 2024/08/13

### Additional context

_No response_

rouault:
Fixes #10758

Also make sure that for other resampling methods, the working data type is large enough (e.g using Float64 for Int32/UInt32/Int64/UInt64). While doing so, fix an underlying bug in the convolution based code, where a hard-coded pixel stride value of 4 was used instead of sizeof(TWork), which I believe wasn't visible before.

### DIFF
--- a/gcore/overview.cpp
+++ b/gcore/overview.cpp
@@ -2099,9 +2099,39 @@ static CPLErr GDALResampleChunk_Gauss(const GDALOverviewResampleArgs &args,
 /*                      GDALResampleChunk_Mode()                        */
 /************************************************************************/
 
+template <class T> static inline bool IsSame(T a, T b)
+{
+    return a == b;
+}
+
+template <> bool IsSame<float>(float a, float b)
+{
+    return a == b || (std::isnan(a) && std::isnan(b));
+}
+
+template <> bool IsSame<double>(double a, double b)
+{
+    return a == b || (std::isnan(a) && std::isnan(b));
+}
+
+template <>
+bool IsSame<std::complex<float>>(std::complex<float> a, std::complex<float> b)
+{
+    return a == b || (std::isnan(a.real()) && std::isnan(a.imag()) &&
+                      std::isnan(b.real()) && std::isnan(b.imag()));
+}
+
+template <>
+bool IsSame<std::complex<double>>(std::complex<double> a,
+                                  std::complex<double> b)
+{
+    return a == b || (std::isnan(a.real()) && std::isnan(a.imag()) &&
+                      std::isnan(b.real()) && std::isnan(b.imag()));
+}
+
 template <class T>
-static CPLErr GDALResampleChunk_Mode_T(const GDALOverviewResampleArgs &args,
-                                       const T *pChunk, T *const pDstBuffer)
+static CPLErr GDALResampleChunk_ModeT(const GDALOverviewResampleArgs &args,
+                                      const T *pChunk, T *const pDstBuffer)
 
 {
     const double dfXRatioDstToSrc = args.dfXRatioDstToSrc;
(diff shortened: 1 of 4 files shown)


gcore/overview.cpp
```diff
@@ -4361,7 +4404,8 @@ CPLErr GDALRegenerateOverviewsEx(GDALRasterBandH hSrcBand, int nOverviewCount,
 
     const GDALDataType eSrcDataType = poSrcBand->GetRasterDataType();
     const GDALDataType eWrkDataType =
-        GDALDataTypeIsComplex(eSrcDataType)
+        (GDALDataTypeIsComplex(eSrcDataType) &&
+         !STARTS_WITH_CI(pszResampling, "NEAR"))
```
EmilRybergAkson, line 4408: Should this also include `&& !STARTS_WITH_CI(pszResampling, "MODE")` ?

rouault: Yes, but not only... Extra commit to make mode resampling exact too

PR #10762, opened 2024-09-10
Fix build issue of external code including gdal_priv.h with MSVC (master only)

rouault:
MSVC complains about "use of undefined type 'GDALDoublePointsCache'" since presumably it tries to locate the destructor of GDALDoublePointsCache, to which it doesn't have access


elpaso:
Is there another option to fix this? To go back for managed to manual memory handling seems a step back to me.

PR #10763, opened 2024-09-11
OpenFileGDB: Fixed the problem of not excluding straight line segments when parsin…

dingsl-giser:
Fixed the problem of not excluding straight line segments when parsing arcs.

## What does this PR do?
Fixed the problem of not excluding straight line segments when parsing circular arcs as `CURVE_ARC_INTERIOR_POINT` type.

When using openfilegdb driver to parse circular arc data in filegdb, circular arc is mistakenly parsed as `CURVE_ARC_INTERIOR_POINT` when it is a line.

ArcGIS:
image

QGIS:
image

This is the result of parsing the arc control point. It is actually a line, but it is classified as `CURVE_ARC_INTERIOR_POINT`:
image

### DIFF
--- a/ogr/ogrpgeogeometry.cpp
+++ b/ogr/ogrpgeogeometry.cpp
@@ -2451,7 +2451,8 @@ OGRErr OGRCreateFromShapeBin(GByte *pabyShape, OGRGeometry **ppoGeom,
                     if (nBits & EXT_SHAPE_ARC_IP)
                         CPLDebug("OGR", "  DefinedIP");
 #endif
-                    if ((nBits & EXT_SHAPE_ARC_IP) != 0)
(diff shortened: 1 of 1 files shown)


rouault:
@dingsl-giser Could you please provide a minimum reproducer for that (ideally one single geometry), so a regression test can be added

dingsl-giser:
> @dingsl-giser Could you please provide a minimum reproducer for that (ideally one single geometry), so a regression test can be added

@rouault This is a circular arc polyline in filegdb with an IsLine attribute I created, you can test it.
[test.gdb.zip](https://github.com/user-attachments/files/16987686/test.gdb.zip)

rouault:
Thanks, I've picked your commit and added an automate test based on your test dataset. Closing this PR as superseded per https://github.com/OSGeo/gdal/pull/10791

PR #10772, opened 2024-09-11
gdal_rasterize: fix (workaround) regressions related to argparse for -b / -burn / -init arguments (master only)

### LINKED ISSUE #10766: gdal_rasterize: issue with -burn (master)
@elpaso There are regressions compared to 3.9

The following command lines no longer work:

1) Source filename directly after -burn
```
$ rm -f out.tif; gdal_rasterize -ts 10 10 -burn 1 autotest/ogr/data/poly.shp out.tif
ERROR 1: Failed to parse 'autotest/ogr/data/poly.shp' as number: pattern 'autotest/ogr/data/poly.shp' does not match to the end
```

but adding ``-l poly`` after -burn works:

```
$ rm -f out.tif; gdal_rasterize -ts 10 10 -burn 1 -l poly autotest/ogr/data/poly.shp out.tif
```

2) Or (probably same underlying issue):
```
$ rm -f out.tif; gdal_rasterize -ts 10 10 autotest/ogr/data/poly.shp out.tif -burn 1
ERROR 1: Failed to parse 'autotest/ogr/data/poly.shp' as number: pattern 'autotest/ogr/data/poly.shp' does not match to the end
```

3) The non documented -burn "val1 val2 ..." no longer works:

$ rm -f out.tif; gdal_rasterize -ts 10 10 -burn "1 2" -l poly autotest/ogr/data/poly.shp out.tif
ERROR 1: Failed to parse '1 2' as number: pattern '1 2' does not match to the end

That point is minor compared to 1) and 2). I believe we have addressed that for some arguments of other utilities, but maybe I'm wrong / can't remember which ones

rouault:
Fixes #10766

argparser is confused by arguments that have at_least_one cardinality, if they immediately precede positional arguments, which is quite undersandable, because if we suppose that we have a -foo argument that takes strings. How would the parser be supposed to distinguish values of -foo from positional filename argument in: ``-foo value1 value2 filename`` ... ? For numeric values, it could probably be tweaked to correctly separate. The problem is that the parse stage doesn't know the type of values of an argument (as they are actions like others...). So my take is that is would be hard to fix argparse to deal with that without significant changes in its code. Basically, currently at_least_one cardinality only works for the last positional argument, and nothing else.
Hum actually my above theory is not completely right ... I see an interesting test case in https://github.com/p-ranav/argparse/blob/master/test/test_optional_arguments.cpp#L147 where there are both nargs_pattern::any positional and non-positional arguments. But in that case, the positional arguments must be specified first ...
For now, let's live with our workaround...

### DIFF
--- a/apps/gdal_rasterize_lib.cpp
+++ b/apps/gdal_rasterize_lib.cpp
@@ -104,12 +104,12 @@ GDALRasterizeOptionsGetParser(GDALRasterizeOptions *psOptions,
         _("This program burns vector geometries (points, lines, and polygons) "
           "into the raster band(s) of a raster image."));
 
+    // Dealt manually as argparse::nargs_pattern::at_least_one is problematic
     argParser->add_argument("-b")
         .metavar("<band>")
         .append()
         .scan<'i', int>()
         .nargs(argparse::nargs_pattern::at_least_one)
-        .store_into(psOptions->anBandList)
         .help(_("The band(s) to burn values into."));
 
     argParser->add_argument("-i")
@@ -132,22 +132,12 @@ GDALRasterizeOptionsGetParser(GDALRasterizeOptions *psOptions,
         auto &group = argParser->add_mutually_exclusive_group(
             psOptionsForBinary != nullptr);
 
+        // Dealt manually as argparse::nargs_pattern::at_least_one is problematic
         group.add_argument("-burn")
             .metavar("<value>")
             .scan<'g', double>()
             .append()
-            .nargs(argparse::nargs_pattern::at_least_one)
-            .action(
-                [psOptions](const std::string &s)
-                {
-                    const CPLStringList aosTokens(
-                        CSLTokenizeString2(s.c_str(), " ", 0));
-                    for (int i = 0; i < aosTokens.size(); i++)
-                    {
-                        psOptions->adfBurnValues.push_back(
-                            CPLAtof(aosTokens[i]));
-                    }
-                })
+            //.nargs(argparse::nargs_pattern::at_least_one)
             .help(_("A fixed value to burn into the raster band(s)."));
 
         group.add_argument("-a")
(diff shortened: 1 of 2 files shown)


elpaso (approved):

elpaso:
@rouault I believe we could make an attempt to exploit `.scan<'g', double>` for such cases but in the general case I agree we would need a pre-processing of argv.

Is this necessary: https://github.com/OSGeo/gdal/pull/10772/files#diff-fb66b6f479760850edc8943817cb9631b8c4451c3614e5ea7ee37bbf33edf90aR140 because it will probably alter the help string to accept a single argument, did you check the new help text?

jratike80:
Do we need to support `-foo value1 value2` or could we require that multiple arguments must be enclosed  `-foo "value1 value2"`
That seems to be used in gdalwarp
https://gdal.org/en/latest/programs/gdalwarp.html#cmdoption-gdalwarp-srcnodata

rouault:
> Is this necessary: https://github.com/OSGeo/gdal/pull/10772/files#diff-fb66b6f479760850edc8943817cb9631b8c4451c3614e5ea7ee37bbf33edf90aR140

yes, I've just tested that un-commenting breaks things like:
``rm -f out.tif; gdal_rasterize -ts 10 10 -burn 10 autotest/ogr/data/poly.shp out.tif``
(and the autotest check I've modified)

> because it will probably alter the help string to accept a single argument, did you check the new help text?

yes, this does affect a bit the help text.
```
With current master: [[-burn <value>...]...|[-a <attribute_name>]|[-3d]]
With this PR:        [[-burn <value>]...|[-a <attribute_name>]|[-3d]]
```

That said the doc page only advertizes (https://gdal.org/en/latest/programs/gdal_rasterize.html) repeating the argument, not being able to multi-value it. So I'm going to actually comment  nargs(argparse::nargs_pattern::at_least_one) for -b and -init too, so that we are in line with the doc.
We clearly lack consistence about that among utilities. But this would be a battle for another day.

PR #10721, opened 2024-09-03
Add new GCI_ constants in particular for InfraRed and SAR, and 'standardize' …

rouault:
…a band-level IMAGERY metadata domain

Implements discussion thread started at https://lists.osgeo.org/pipermail/gdal-dev/2024-August/thread.html#59403

Added values:
    * GCI_PanBand: Panchromatic band [0.40 - 1.00 um]
    * GCI_CoastalBand: Coastal band [0.40 - 0.45 um]
    * GCI_RedEdgeBand: Red-edge band [0.69 - 0.79 um]
    * GCI_NIRBand: Near-InfraRed (NIR) band [0.75 - 1.40 um]
    * GCI_SWIRBand: Short-Wavelength InfraRed (SWIR) band [1.40 - 3.00 um]
    * GCI_MWIRBand: Mid-Wavelength InfraRed (MWIR) band [3.00 - 8.00 um]
    * GCI_LWIRBand: Long-Wavelength InfraRed (LWIR) band [8.00 - 15 um]
    * GCI_TIRBand: Thermal InfraRed (TIR) band (MWIR or LWIR) [3 - 15 um]
    * GCI_OtherIRBand: Other infrared band [0.75 - 1000 um]

They can be in particular used for the GeoTIFF driver, being serialized in the GDAL_METADATA tag.

For spectral bands, the wavelength ranges are indicative only, and may vary depending on sensors.

STACIT driver enhanced to map STAC common_name to GCI_ constants

gdalinfo -json output enhanced to output STAC common_name from GCI_ constants

Items in the IMAGERY band-level metadata domain:
- CENTRAL_WAVELENGTH_UM: Central Wavelength in micrometers.
- FWHM_UM: Full-width half-maximum (FWHM) in micrometers.

Filled by the SENTINEL2 and ENVI drivers (if corresponding metadata items are found in the ENVI header)
May also be found in other drivers handling arbitrary GDAL metadata, such as the ones using the GDAL Persistent Auxiliary Mechanism (PAM / .aux.xml side car files) or VRT drivers. The GTiff driver also supports serializing and deserializing the band IMAGERY metadata domain in the ``GDAL_METADATA`` TIFF tag.

PR #10773, opened 2024-09-11
STACIT: STAC 1.1 support

m-mohr:
## What does this PR do?

Implement support for STAC 1.1 in STACIT driver.

Is waiting for #10721 

Disclaimer: Last time I wrote C/C++ was 10 years ago, this may take me a while...

None

## Tasklist

 - [ ] Add support for STAC 1.1 in STACIT
 - [ ] ~Add support for STAC 1.1 in STACTA~
 - [ ] Add support for STAC 1.1 in CLI (gdalinfo)?
 - [ ] Check common band names for validity
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/frmts/stacit/stacitdataset.cpp
+++ b/frmts/stacit/stacitdataset.cpp
@@ -60,7 +60,7 @@ struct AssetSetByProjection
 struct Asset
 {
     std::string osName{};
-    CPLJSONArray eoBands{};
+    CPLJSONArray bands{};
     std::map<std::string, AssetSetByProjection> assets{};
 };
 
@@ -139,8 +139,7 @@ int STACITDataset::Identify(GDALOpenInfo *poOpenInfo)
             return pszHeader[0] == '{';
         }
 
-        if (strstr(pszHeader, "\"stac_version\"") != nullptr &&
-            strstr(pszHeader, "\"proj:transform\"") != nullptr)
+        if (strstr(pszHeader, "\"stac_version\"") != nullptr)
         {
             return true;
         }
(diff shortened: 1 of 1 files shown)


frmts/stacit/stacitdataset.cpp
```diff
@@ -139,8 +139,7 @@ int STACITDataset::Identify(GDALOpenInfo *poOpenInfo)
             return pszHeader[0] == '{';
         }
 
-        if (strstr(pszHeader, "\"stac_version\"") != nullptr &&
-            strstr(pszHeader, "\"proj:transform\"") != nullptr)
```
m-mohr, line 143: Why is proj:transform checked for? It could appear both in properties or assets so it seems like this check may lead to false results.

frmts/stacit/stacitdataset.cpp
```diff
@@ -594,34 +607,37 @@ bool STACITDataset::SetupDataset(
...
+                else
+                    osCommonName = band["common_name"].ToString();
                 if (osCommonName == "red")
                     poVRTBand->SetColorInterpretation(GCI_RedBand);
                 else if (osCommonName == "green")
                     poVRTBand->SetColorInterpretation(GCI_GreenBand);
                 else if (osCommonName == "blue")
                     poVRTBand->SetColorInterpretation(GCI_BlueBand);
-                else if (osCommonName == "alpha")
-                    poVRTBand->SetColorInterpretation(GCI_AlphaBand);
```
m-mohr, line 615: There is no common name called 'alpha" in STAC.

frmts/stacit/stacitdataset.cpp
rouault: The driver requires the projection extension to be there, hence checking for it. And as Identify() needs to be quick (it only has a subset of the JSON and can't use a JSON parser), it doesn't get into the details to figure out where the element is present

frmts/stacit/stacitdataset.cpp
rouault: Really ? but please don't remove that there, otherwise this is going to conflict with PR https://github.com/OSGeo/gdal/pull/10721 that touches that code too

frmts/stacit/stacitdataset.cpp
m-mohr: Oh, now I remember what strstr does. I'll revert, thanks.

frmts/stacit/stacitdataset.cpp
m-mohr: Yes, no alpha. The full list of allowed values is here: https://github.com/stac-extensions/eo?tab=readme-ov-file#common-band-names

frmts/stacit/stacitdataset.cpp
```diff
@@ -594,34 +603,35 @@ bool STACITDataset::SetupDataset(
...
+            const auto &band = asset.bands[i];
+            const auto osBandName = band["name"].ToString();
             if (!osBandName.empty())
                 poVRTBand->SetDescription(osBandName.c_str());
 
             if (eInterp != GCI_Undefined)
             {
-                const auto osCommonName = eoBand["common_name"].ToString();
+                auto osCommonName = band["eo:common_name"].ToString();
+                if (osBandName.empty())
```
rouault, line 617: shouldn't that be?
```suggestion
                if (osCommonName.empty())
```

and at line 406 we check first "bands", and fallback to "eo:bands", whereas here this is first "eo:common_name", and fallback to "common_name"

m-mohr: Yes, good catch.

frmts/stacit/stacitdataset.cpp
m-mohr: The list was also extended a bit recently and the property in STAC 1.1 is a named eo:common_name and can be placed in asset or bands.

rouault: > The list was also extended a bit recently

yes, #10721 takes inspiration from it, but feedback of the GDAL community was that some of the details like "rededge071", "rededge075", etc. were much too specific, so we didn't go to that detail for the GCI_ColorInterpretation enumeration.

I'm wondering if #10721 shouldn't be merge first (should be soon. I'm just letting a couple extra days for potential new comments on it), so that you can work on a stable base

m-mohr: Yeah, go ahead. This will still take me a little longer anyway.

frmts/stacit/stacitdataset.cpp
m-mohr: It needs two out of the three proj: properties and transform is one them. So just checking proj:transform seems to exclude cases where proj:bbox and proj:shape are set. Should we update this to check whether any of them is present?

m-mohr: | fields | currently | new | 
| transform + shape | accepted | accepted | 
| transform + bbox | accepted | accepted |
| bbox + shape | rejected | accepted |

rouault: > Maybe worth a separate small PR though, it's unrelated to the STAC 1.1 updates.

yes

rouault:
https://github.com/OSGeo/gdal/pull/10721 has been merged. Please rebase on top of latest master

m-mohr:
Won't have time to work on this before December, but happy to reopen a new PR if this gets closed...

m-mohr:
Still not December ;)

m-mohr:
Closing for now, was a bit too optimistic about my time. If no one else works on it I may pick it up again at some point, others feel free to continue with the given code or start from scratch.

rouault:
continued in https://github.com/OSGeo/gdal/pull/11760

frmts/stacit/stacitdataset.cpp
rouault: taken into account per https://github.com/OSGeo/gdal/pull/11760/commits/8dd8c815d61e42b789e9d8f4ade372e7bb811b64

PR #10779, opened 2024-09-12
S102: add support for IHO S102 v3.0 specification

rouault:


frmts/hdf5/s102dataset.cpp
```diff
@@ -412,34 +424,45 @@ GDALDataset *S102Dataset::Open(GDALOpenInfo *poOpenInfo)
...
+    if (!poGroupQuality)
     {
-        CPLError(CE_Failure, CPLE_AppDefined,
-                 "Cannot find group /QualityOfSurvey");
-        return false;
+        pszNameOfQualityGroup = "QualityOfBathymetryCoverage";
+        poGroupQuality = poRootGroup->OpenGroup(pszNameOfQualityGroup);
+        if (!poGroupQuality)
+        {
+            CPLError(CE_Failure, CPLE_AppDefined,
```
elpaso, line 444: Can we specify in the error message which group exactly was expected?

frmts/hdf5/s102dataset.cpp
```diff
@@ -469,14 +491,42 @@ bool S102Dataset::OpenQualityOfSurvey(
...
+        }
+        else if (oType.GetClass() == GEDTC_COMPOUND &&
+                 oType.GetComponents().size() == 1 &&
+                 oType.GetComponents()[0]->GetType().GetClass() ==
+                     GEDTC_NUMERIC &&
+                 oType.GetComponents()[0]->GetType().GetNumericDataType() ==
+                     GDT_UInt32)
+        {
+            // seen in a S102 v3 product (102DE00CA22_UNC_MD.H5), although
+            // I believe this is non-conforment.
```
elpaso, line 514:
```suggestion
            // I believe this is non-conformant.
```

frmts/hdf5/s102dataset.cpp
```diff
@@ -469,14 +491,42 @@ bool S102Dataset::OpenQualityOfSurvey(
...
+
+            // Escape potentials single-quote and double-quote with anti-slash
```
elpaso, line 516: do you mean 'back' slash ?

frmts/hdf5/s102dataset.cpp
```diff
@@ -527,6 +578,8 @@ bool S102Dataset::OpenQualityOfSurvey(
 
     auto poDS =
         std::unique_ptr<GDALDataset>(poValuesArray->AsClassicDataset(1, 0));
+    if (!poDS)
+        return false;
```
elpaso, line 582: Maybe log an error here or is this nominal?

frmts/hdf5/s102dataset.cpp
rouault: We could based on the IHO spec version, but that's more code to add to fetch it, and I don't think it is necessary. The current message should be an already good enough hint, and it would only happen if users try to open a subdataset whose name hasn't been reported before on the main ile

frmts/hdf5/s102dataset.cpp
rouault, line 516:
```suggestion
            // Escape potentials single-quote and double-quote with back-slash
```

rouault: indeed, Frenchism..

frmts/hdf5/s102dataset.cpp
rouault: this shouldn't happen normally. And if it did, AsClassicDataset() is supposed to have emitted an error message

frmts/hdf5/s102dataset.cpp
rouault: fixed

rouault:
FYI @GlenRice-NOAA

GlenRice-NOAA:
Adding @giumas

giumas:
> WARNING: it seems there are some complexities (in particular multiple grids with different vertical datums) in the v3 spec that are not handled by this update

@rouault, should it be mentioned in the documentation that the driver supports only a partial implementation of S-102 PS v3.0? 
More in general, how much work would be required to support multiple grids with different vertical datums?

giumas:
> WARNING: it seems there are some complexities (in particular multiple grids with different vertical datums) in the v3 spec that are not handled by this update

@rouault can you clarify what you mean with this warning?

rouault:
> can you clarify what you mean with this warning?

well, hardly. That would imply that I actually fully read the spec and understand it, and I'm not in a mood of doing that... ;-) Mostly @GlenRice-NOAA warned me that there might be multiple grids in a S102 v3 and the driver definitely doesn't handle that and will just use the Group_001 one.

rouault:
warning in doc about incomplete S-102 v3 support adde in https://github.com/OSGeo/gdal/pull/10891

PR #10781, opened 2024-09-12
Doc: insert space character in :choices: output, to be nice with browsers line-break insertion logic

### LINKED ISSUE #10778: Docs: incorrect display of long lines (using Chrome on Windows 10)
### What is the bug?

Using Google Chrome 128.0.6613.138 on Windows 10:

image

The horizontal scrolling bar is not available.

Open https://gdal.org/en/latest/drivers/vector/filegdb.html using Google Chrome on Windows 10.

### Versions and provenance

Windows 10 Home 1903 64-bit 10.0.18362 Build 18362.778

### Additional context

_No response_

rouault:
Fixes #10778

### DIFF
--- a/doc/source/_extensions/configoptions.py
+++ b/doc/source/_extensions/configoptions.py
@@ -104,7 +104,9 @@ class Option:
         elif len(self.choices) == 1:
             text += f"={self.choices[0]}: "
         else:
-            text += f'=[{"/".join(self.choices)}]: '
+            # Insert space around /, so that all browsers produce line breaks
+            # Cf https://github.com/OSGeo/gdal/issues/10778
+            text += f'=[{" / ".join(self.choices)}]: '
 
         para += nodes.strong(text, text)
 


jratike80:
By https://stackoverflow.com/questions/24488950/css-to-break-a-line-only-on-a-certain-character there is also a character called ZERO WIDTH SPACE (U+2000B). But maybe such invisible space could make more confusion than traditional space.

rouault:
> By https://stackoverflow.com/questions/24488950/css-to-break-a-line-only-on-a-certain-character there is also a character called ZERO WIDTH SPACE (U+2000B). But maybe such invisible space could make more confusion than traditional space.

Thanks for the hint. Let's test this hypothesis in https://github.com/OSGeo/gdal/pull/10782

rouault:
ok, so we have the RTD preview of this pull request in https://gdal--10781.org.readthedocs.build/en/10781/drivers/vector/filegdb.html

rouault:
superseded per https://github.com/OSGeo/gdal/pull/10782

PR #10782, opened 2024-09-12
Doc: insert Unicode zero-width space character in :choices: output, to be nice with browsers line-break insertion logic

### LINKED ISSUE #10778: Docs: incorrect display of long lines (using Chrome on Windows 10)

rouault:
Fixes #10778

### DIFF
--- a/doc/source/_extensions/configoptions.py
+++ b/doc/source/_extensions/configoptions.py
@@ -104,7 +104,14 @@ class Option:
         elif len(self.choices) == 1:
             text += f"={self.choices[0]}: "
         else:
-            text += f'=[{"/".join(self.choices)}]: '
+            # Insert Unicode zero-width space around separating forward slash,
+            # so that all browsers produce line breaks.
+            # Cf https://github.com/OSGeo/gdal/issues/10778
+            zero_width_space = "\u200b"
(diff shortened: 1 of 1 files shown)


jratike80:
Wouldn't it be enough to add zero width space just on one side of `/`?

rouault:
> Wouldn't it be enough to add zero width space just on one side of `/`?

presumably, but this would lack symmetry (and I'm not going to enter political debates whether left or right should be preferred) :-) :-)

rouault:
@jratike80 @agiudiceandrea Could you check if https://gdal--10782.org.readthedocs.build/en/10782/drivers/vector/filegdb.html renders correctly for you?

jratike80:
Renders fine for me on Windows with Chrome 128.0.6613.120 and with Edge 128.0.2739.67

agiudiceandrea:
Renders fine also for me 
image

PR #10785, opened 2024-09-12
Regression test for fixed FlatGeobuf reader external singlepart MultiLineString

bjornharrtell:
@rouault this is for https://github.com/OSGeo/gdal/pull/10775#issuecomment-2347034735, I opted to do it as new PR to verify the expectation that it fails pre patch. Can merge into other PR when confirmed.

I have problems getting tests to run locally, unfortunately. They run, but doesn't seem to run against the compiled gdal or something, even though dev env is set etc.


autotest/ogr/ogr_flatgeobuf.py
```diff
@@ -874,6 +874,23 @@ def test_ogr_flatgeobuf_read_invalid_geometries(filename):
...
             pass
 
+###############################################################################
+# Check that we don't crash
+
+
+@pytest.mark.parametrize(
+    "filename",
+    [
+        "test_ogr_flatgeobuf_singlepart_mls_new.fgb",
```
rouault, line 884:
```suggestion
        "data/flatgeobuf/test_ogr_flatgeobuf_singlepart_mls_new.fgb",
```

rouault:
yes please this pull request into #10775, once it fails here for good reasons ;-)

bjornharrtell:
Failed to make a file as intended.. and didn't think about that I can verify the problem with the ogrinfo I have in my system. New file commited which gives this error from ogrinfo 3.8.4:

ERROR 1: Unexpected nullptr: MultiLineString ends data
ERROR 1: Failed to read geometry
ERROR 1: Fatal error parsing feature

bjornharrtell:
Now I don't understand why the test passes. @rouault any ideas?

autotest/ogr/ogr_flatgeobuf.py
```diff
@@ -874,6 +874,23 @@ def test_ogr_flatgeobuf_read_invalid_geometries(filename):
...
+# Check that we don't crash
+
+
+@pytest.mark.parametrize(
+    "filename",
+    [
+        "data/flatgeobuf/test_ogr_flatgeobuf_singlepart_mls_new.fgb",
+    ],
+)
+def test_ogr_flatgeobuf_read_singlepart_mls_new(filename):
```
rouault, line 887: I've just tested locally that you would get an exception if you enable exceptions like this (they are disabled by default in this module). But it would be better (in the branch where it is fixed) to actually check that we get a valid geometry by inspecting a feature. I'll try to improve your other PR to do that if you can't manage to make the test suite run (normally "source ./scripts/setdevenv.sh" should do it)

```suggestion
@gdaltest.enable_exceptions()
def test_ogr_flatgeobuf_read_singlepart_mls_new(filename):
```

bjornharrtell: I see, thanks @rouault

bjornharrtell:
Merged into https://github.com/OSGeo/gdal/pull/10775, closing.

PR #10799, opened 2024-09-14
CPLRecode(): make ISO-8859-2 and -15 and CP437/CP1250/CP1251/CP1252 to UTF-8 always available

rouault:
(but not other direction yet)

This will help for common use cases when using a minimal GDAL build lacking iconv support.


elpaso (approved):
LGTM, I'm curious about the use case though, what/who is using these "minimal" builds without recode?

PR #10801, opened 2024-09-14
Doc: Clarify behavior of gdalwarp -et option

dbaston:
## What does this PR do?

Attempts to add some clarity to documentation of gdalwarp's `-et` parameter, gleaned from review of `GDALApproxTransform`.

### DIFF
--- a/alg/gdaltransformer.cpp
+++ b/alg/gdaltransformer.cpp
@@ -3828,7 +3828,8 @@ static CPLXMLNode *GDALSerializeApproxTransformer(void *pTransformArg)
  * @param pBaseTransformArg the callback argument for the high precision
  * transformer.
  * @param dfMaxError the maximum cartesian error in the "output" space that
- * is to be accepted in the linear approximation.
+ * is to be accepted in the linear approximation, evaluated as a Manhattan
+ * distance.
  *
  * @return callback pointer suitable for use with GDALApproxTransform().  It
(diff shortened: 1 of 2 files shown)


rouault:
> gleaned from review of `GDALApproxTransform`.

Can we compare our reasoning ? Mine is that GDALCreateApproxTransformer() doc mentions that ``dfMaxError the maximum cartesian error in the "output" space`` (and implementation matches that), but as far as warping is concerned, given that GDALApproxTransform() checks that all points have the same Y to apply the piece-wise linear interpolatin, and that GWKWarpKernel passes target y lines, and invokes the inverse transformation, and the output space of the inverse transformer is source pixels, then the threshold is a fraction of source pixels.

dbaston:
Yes, that matches my reasoning. That said, to me

> the maximum cartesian error in the "output" space

should probably state that the error is evaluated as a Manhattan distance:

https://github.com/OSGeo/gdal/blob/ae91738520706b869701b52ff32e2645c60bba56/alg/gdaltransformer.cpp#L3976-L3995

which I described in this PR as "sum of the horizontal and vertical errors"

rouault:
> should probably state that the error is evaluated as a Manhattan distance:

yes, if we want to be fully accurate. Do you want to do that addition or should I just merge this PR as it ?

dbaston:
I'll add it.

PR #10802, opened 2024-09-14
CI: Fix ReadTheDocs test for [ci skip] with multiline commit messages

dbaston:
git log will emit a blank line which passes the grep -v test and prevents exiting.


dbaston:
[502b1b8](https://github.com/OSGeo/gdal/pull/10802/commits/502b1b854d54fe74cb6dbc1b671e4866b8f968fc) fixes this but still fails for long-ish commit messages.

Issue #10813, opened 2024-09-16
`OPTIMIZE_SIZE` Warper option from `gdalwarp` gives different re-projection output


aafaque33:
### What is the bug?

Recently we tried to upgrade the GDAL version in the docker container.

After upgrade of the GDAL from 3.5.3 to GDAL 3.9.1 we have found that one of the re-projection output was having differences (not in all pixels but kind of scan lines with slight pixel shift)
Verified that same shift happens in GDAL version 3.8

First we thought that it may be because of the `rasterio` as the issue was while using the `rio warp` but not with `gdalwarp` so logged the issue / discussion in `raserio`
- Look at detailed discussion here: https://github.com/rasterio/rasterio/discussions/3150

After discussion it was found that same issue haapens in `gdalwarp` if the `-co TILED=YES -co BLOCKXSIZE=512 -co BLOCKYSIZE=512` which `rio warp` is using by default.

It turns out after so much trial and error that it was because of the one of the `warper option` `OPTIMIZE_SIZE` that seems like before GDAL 3.8 was always `FALSE` for everything, But since [GDAL 3.8](https://gdal.org/en/latest/api/gdalwarp_cpp.html#_CPPv4N15GDALWarpOptions16papszWarpOptionsE:~:text=within%20the%20polygon.-,OPTIMIZE_SIZE,-%3A%20This%20defaults%20to) is automatically enabled (may be for TILED / BLOCKSIZE / Compressed options) when it is safe to do so, and that produces a slight pixel shift / difference in reprojected output

Testing the `gdalwarp` post GDAL 3.8 with that warper option disabled gives the same reprojected output as it was in older versions of `GDAL`

Not sure if it is expected behavior but shouldn't the reprojected output be same with or without that `warper option`? 

### Steps to reproduce the issue

**Note: Reprojection is from UTM Zone 22N to Geographic (EPSG:4326)**

### Data Files

Tif file with no reprojection done (Currently in UTM Zone 22N). Providing the google drive link as can't upload more than 25 MB file
- https://drive.google.com/file/d/1X3eP7O4nihHM7_mONIlekP6o5Dv7S1ut/view?usp=drive_link
 
## Run a docker container

```
> docker run --rm -it -v ~/docker:/data ghcr.io/osgeo/gdal:ubuntu-full-3.9.2 /bin/bash
```

## Run GDAL WARP with different scenarios**

```
> root@120bf665500e:/# gdalinfo --version
[...]

PR #10814, opened 2024-09-16
gdalwarper.cpp: more details about OPTIMIZE_SIZE

rouault:
and document XSCALE and YSCALE warping options

Ref #10813


rouault:
re-opened as a pull request again my fork in https://github.com/OSGeo/gdal/pull/10816

PR #10817, opened 2024-09-16
gdalwarper.cpp: more details about OPTIMIZE_SIZE

rouault:
and document XSCALE and YSCALE warping options

Ref #10813

### DIFF
--- a/alg/gdalwarper.cpp
+++ b/alg/gdalwarper.cpp
@@ -1167,6 +1167,20 @@ CPLErr GDALWarpDstAlphaMasker(void *pMaskFuncArg, int nBandCount,
  * will be selected, not just those whose center point falls within the
  * polygon.</li>
  *
+ * <li>XSCALE: Ratio expressing the resampling factor (number of destination
+ * pixels per source pixel) along the target horizontal axis.
+ * The scale is used to determine the number of source pixels along the x-axis
+ * that are considered by the resampling algorithm.
+ * Equals to one for no resampling, below one for downsampling
+ * and above one for upsampling. This is automatically computed, for each
+ * processing chunk, and may thus vary among them, depending on the
+ * shape of output regions vs input regions. Such variations can be undesired
+ * in some situations. If the resampling factor can be considered as constant
+ * over the warped area, setting a constant value can lead to more reproducible
+ * pixel output.</li>
+ *
+ * <li>YSCALE: Same as XSCALE, but along the horizontal axis.</li>
+ *
  * <li>OPTIMIZE_SIZE: This defaults to FALSE, but may be set to TRUE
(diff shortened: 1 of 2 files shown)


alg/gdalwarper.cpp
```diff
@@ -1167,6 +1167,17 @@ CPLErr GDALWarpDstAlphaMasker(void *pMaskFuncArg, int nBandCount,
  * will be selected, not just those whose center point falls within the
  * polygon.</li>
  *
+ * <li>XSCALE: Ratio expressing the resampling factor along the target
```
dbaston, line 1170:
```suggestion
 * <li>XSCALE: Ratio expressing the resampling factor (number of destination pixels per source pixel) along the target
```

alg/gdalwarper.cpp
```diff
@@ -1167,6 +1167,17 @@ CPLErr GDALWarpDstAlphaMasker(void *pMaskFuncArg, int nBandCount,
...
+ * horizontal axis. Equals to one for no resampling, below one for downsampling
+ * and above one for upsampling. This is automatically computed, for each
+ * processing chunk, and may thus vary among them, depending on the
+ * shape of output regions vs input regions. Such variations can be undesired
+ * in some situations. If the resampling factor can be considered as constant
+ * over the warped area, setting a constant value can lead to more reproducible
+ * pixel output.</li>
```
dbaston, line 1177: Describe `XSCALE=FROM_GRID_SAMPLING` ?

alg/gdalwarper.cpp
```diff
@@ -1167,6 +1167,17 @@ CPLErr GDALWarpDstAlphaMasker(void *pMaskFuncArg, int nBandCount,
  * will be selected, not just those whose center point falls within the
  * polygon.</li>
  *
+ * <li>XSCALE: Ratio expressing the resampling factor along the target
+ * horizontal axis. Equals to one for no resampling, below one for downsampling
+ * and above one for upsampling. This is automatically computed, for each
```
dbaston, line 1172: Add something like "The scale is used to determine the number of source pixels along the x-axis that are considered by the resampling algorithm." ?

(Does it have another purpose?)

alg/gdalwarper.cpp
rouault: applied, thanks

alg/gdalwarper.cpp
rouault: applied, thanks

alg/gdalwarper.cpp
rouault: Let's keep it as undocumented for now / implementation detail... (partly because I'm never satisfied with those settings, and also because this is tricky to describe accurately and in a understandable way)

PR #10822, opened 2024-09-17
gdalwarp: fix crash/infinite loop when using -tr one a 1x1 blank raster (3.8.0 regression)

### LINKED ISSUE #10827: abort with gdalwarp and bad (too coarse) resolution
### What is the bug?

I see an abort with a target resolution that is entirely inappropriate for the target. I found it by going from entirely automated targets and tweaking the resolution from suitable (0.0006)  to coarse (1.0). 

```bash
gdal_create /tmp/test_mds11.tif -ot Float32 -outsize 256 256 -burn 0 -a_srs EPSG:3857 -a_ullr 20017940 -1976356 20037508 -1995924

gdalwarp /tmp/test_mds11.tif -t_srs "EPSG:4326" /tmp/test_mds12.tif -tr 1 1
# corrupted size vs. prev_size
#Aborted
```

### Versions and provenance

Ubuntu 22.04.4 LTS 

GDAL 3.10.0dev-262ffc2002, released 2024/09/17

### Additional context

Originally in a different context in R I saw this and I will check again after feedback here. 

```
# malloc_consolidate(): invalid chunk size
# Aborted
```

rouault:
Fixes regression introduced by https://github.com/OSGeo/gdal/pull/7911
Test case: https://lists.osgeo.org/pipermail/gdal-dev/2024-September/059512.html

Fixes #10827

### DIFF
--- a/apps/gdalwarp_lib.cpp
+++ b/apps/gdalwarp_lib.cpp
@@ -4428,10 +4428,14 @@ static GDALDatasetH GDALWarpCreateOutput(
             nPixels = static_cast<int>((psOptions->dfMaxX - psOptions->dfMinX +
                                         (psOptions->dfXRes / 2.0)) /
                                        psOptions->dfXRes);
+            if (nPixels == 0)
+                nPixels = 1;
             nLines = static_cast<int>(
                 (std::fabs(psOptions->dfMaxY - psOptions->dfMinY) +
                  (psOptions->dfYRes / 2.0)) /
                 psOptions->dfYRes);
+            if (nLines == 0)
+                nLines = 1;
             adfDstGeoTransform[0] = psOptions->dfMinX;
             adfDstGeoTransform[3] = psOptions->dfMaxY;
             adfDstGeoTransform[1] = psOptions->dfXRes;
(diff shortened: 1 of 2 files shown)


misev:
I applied this commit to the release/3.8 branch and compiled it on Ubuntu 24.04 with cmake config
```
cmake ../gdal -DGDAL_BUILD_OPTIONAL_DRIVERS=OFF -DOGR_BUILD_OPTIONAL_DRIVERS=OFF
```
but it doesn't seem to work well:
```
$ export LD_LIBRARY_PATH="/usr/local/lib:$LD_LIBRARY_PATH"
$ ldd /usr/local/bin/gdalwarp 2>&1 | grep gdal -i
    libgdal.so.34 => /usr/local/lib/libgdal.so.34 (0x00007fc412c00000)
    
$ /usr/local/bin/gdalwarp test.vrt -t_srs EPSG:32631 -tr 1000 1000 -of VRT output.vrt -overwrite
Creating output file that is 0P x 0L.
ERROR 1: Attempt to create 0x0 dataset is illegal,sizes must be larger than zero.
```

rouault:
> but it doesn't seem to work well:

I've updated the PR to also handle the case where -tap is not specified

elpaso (approved):

PR #10867, opened 2024-09-24
Add read-only raster driver, SNAP_TIFF, for Sentinel Application Processing GeoTIFF files

rouault:
Such files are formulated in a way that makes it difficult to read them with the generic GTiff driver:
- single strip of uncompressed Float32 data type (1.7 GB on the sample file I worked on)
- more than 65,535 GCPs in the GeoTIFF TiePoints tag (> 1 million), which make it   impossible to read with libgeotiff. Furthermore those tiepoints form   actually a geolocation array, and are thus best represented using the GDAL GEOLOCATION formalism
- a unknown TIFF tag of code 65000, which happens to contain XML DIMAP metadata of several megabytes. Part of it contains information on the nodata value, offset and scale, and thus must be parsed to report appropriate band properties

All that combined calls for a specialized driver, instead of cluttering even more the generic GTiff driver.
To help for that, I've developed a simple TIFF reader in https://github.com/libertiff/libertiff and included it here.


jratike80:
It seems a bit like new QGIS nightly builds are now offering this SNAP_TIFF format as a default **output** format for some processes like clip raster. It may be only a QGIS bug but I make a note here. I can investigate the QGIS side better later but now I have a real work that I must do first.

rouault:
> It may be only a QGIS bug

much likely, because on the GDAL side, we don't expose any creation capabilities

rouault:
> It seems a bit like new QGIS nightly builds are now offering this SNAP_TIFF format as a default **output** format for some processes like clip raster.

I don't reproduce this with my QGIS dev build against a GDAL version with SNAP_TIFF. The GTiff driver is correctly selected. That said there was definitely a known issue with format selection in QGIS processing, but it has been fixed in https://github.com/qgis/QGIS/pull/57702

jratike80:
Raster Extraction - Clip raster by extent for example creates a GDAL command

`gdal_translate -projwin 360853.6639 6714413.7697 360961.8182 6714257.3265 -of SNAP_TIFF C:\data\dem2m.tif C:/Users/JRAHKONEN/AppData/Local/Temp/processing_giGcnu/ad37cba320a145848a03c12bed16db8d/OUTPUT.tif`

My QGIS version is from OSGeo4W:

```
3.39.0-Master
QGIS code revision
40246d3509
```
Similar error has happened before because Interlis format got selected https://gis.stackexchange.com/questions/477614/polygonize-error/477617, maybe that does not happen any more.

PR #10885, opened 2024-09-27
[gdal_contour] Fix regression when fixed level == raster max

### LINKED ISSUE #10854: gdal_contour behavior change between 3.8.4 and 3.9.0
### What is the bug?

It appears `gdal_contour` behavior has changed somewhere between 3.8.4 and 3.9.0.  I am generating hillshade polygons from hillshade raster images generated by `gdaldem`. In my application, the world is cut into tiles and it is important that the hillshade polygons match across tile boundaries. To accomplish this, the hillshade raster image for a tile overlaps with its neighbor tile images by 8 pixels to ensure continuity (that's a total of 16 overlapping pixels between any two neighbors).

This all worked fine until I upgraded to GDAL 3.9.0. The images below illustrate the differences:

Here is the output of `gdal_contour` with version 3.8.4 near the boundary between two tiles where tile A is north of tile B:
### Tile A
90488694-highlight-gdal-3 7 2
### Tile B
90488692-highlight-gdal-3 7 2
You can see that the polgons generated in the overlapping border areas between the tiles are identical

Here is the output of `gdal_contour` with version 3.9.0 in the same region:
### Tile A
90488694-highlight-gdal-3 9 0
### Tile B
90488692-highlight-gdal-3 9 0

Here Tile A polygons are different. Note that the small polygon (gold in Tile B) is generated, but is classified to a different level than in Tile B.

In the images above, the two white dots represent points along the edge between the tiles where they will eventually be cut.

Both sets of data were generated off the same two hillshade raster images:
90488692-90488694-hillshade-raster-overlap

You can see in the image above, that there is overlap between the neighboring hillshade raster images. In the picture one of the images was turned red and made slightly transparent. I have verified that the images are pixel-for-pixel equal in the overlap region.

Attached is a zip file containing a bash script, the two source hillshade raster images and the generated hillshade polygons from each GDAL version. You should be able to modify the bash script to point to where the different GDAL binary versions are located on your computer to run the script. The script also illustrates how `gdal_contour` is being called.

[gdal_contour_issue.zip](https://github.com/user-attachments/files/17081339/gdal_contour_issue.zip)

### Versions and provenance

Ubuntu 24.04
GDAL version 3.9.0

### Additional context

_No response_

elpaso:
Fix #10854

### DIFF
--- a/alg/contour.cpp
+++ b/alg/contour.cpp
@@ -773,7 +773,8 @@ CPLErr GDALContourGenerateEx(GDALRasterBandH hBand, void *hLayer,
                 // last level polygons
                 if (maxLevel == dfMaximum)
                 {
-                    dfMaximum = std::nextafter(dfMaximum, std::numeric_limits<double>::infinity());
+                    dfMaximum = std::nextafter(
+                        dfMaximum, std::numeric_limits<double>::infinity());
                 }
             }
 
--- a/alg/contour.cpp
+++ b/alg/contour.cpp
@@ -773,7 +773,7 @@ CPLErr GDALContourGenerateEx(GDALRasterBandH hBand, void *hLayer,
                 // last level polygons
                 if (maxLevel == dfMaximum)
                 {
-                    dfMaximum += 1e-10;
+                    dfMaximum = std::nextafter(dfMaximum, std::numeric_limits<double>::infinity());
                 }
             }
 
--- a/alg/contour.cpp
+++ b/alg/contour.cpp
@@ -196,7 +196,7 @@ struct PolygonContourWriter
     std::unique_ptr<OGRMultiPolygon> currentGeometry_ = {};
     OGRPolygon *currentPart_ = nullptr;
     OGRContourWriterInfo *poInfo_ = nullptr;
-    double currentLevel_;
+    double currentLevel_ = 0;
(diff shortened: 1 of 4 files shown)


alg/contour.cpp
```diff
@@ -749,6 +749,32 @@ CPLErr GDALContourGenerateEx(GDALRasterBandH hBand, void *hLayer,
...
+                        }
+                    }
+                }
+
+                // If the maximum raster value is equal to the last requested
+                // level, add a small value to the maximum to avoid skipping the
+                // last level polygons
+                if (maxLevel == dfMaximum)
+                {
+                    dfMaximum += 1e-10;
```
rouault, line 776: wouldn't the following be more robust, especially when dealing with values with large magnitude ?
```suggestion
                    dfMaximum = std::nextafter(dfMaximum, std::numeric_limits<double>::infinity());
```

rouault:
There's also some testing of it through autotest/cpp/gdal_unit_test (autotest/cpp/test_marching_squares_*.cpp files) which must be updated

elpaso:
> There's also some testing of it through autotest/cpp/gdal_unit_test (autotest/cpp/test_marching_squares_*.cpp files) which must be updated

The test was actually ok: I reverted part of my first commit.

PR #10888, opened 2024-09-27
gdalwarp: allow unlimited warp memory with -wm unlimited

dbaston:
There may be good reasons not to allow this. I was just looking for an easy way to rule out chunk boundaries when investigating artifacts observed in gdalwarp output.

### DIFF
--- a/apps/gdalwarp_lib.cpp
+++ b/apps/gdalwarp_lib.cpp
@@ -5754,11 +5754,20 @@ GDALWarpAppOptionsGetParser(GDALWarpAppOptions *psOptions,
         .action(
             [psOptions](const std::string &s)
             {
-                if (CPLAtofM(s.c_str()) < 10000)
+                if (EQUAL("unlimited", s.c_str()))
+                {
+                    psOptions->dfWarpMemoryLimit =
+                        std::numeric_limits<double>::infinity();
+                }
+                else if (CPLAtofM(s.c_str()) < 10000)
+                {
                     psOptions->dfWarpMemoryLimit =
                         CPLAtofM(s.c_str()) * 1024 * 1024;
+                }
                 else
+                {
                     psOptions->dfWarpMemoryLimit = CPLAtofM(s.c_str());
+                }
             })
         .help(_("Set max warp memory."));
 
--- a/doc/source/programs/gdalwarp.rst
+++ b/doc/source/programs/gdalwarp.rst
@@ -394,6 +394,9 @@ with control information.
     the "warp memory" allowed it will subdivide the chunk into smaller chunks
     and try again.
 
+    A -wm value of "unlimited" will prevent this chunking. If sufficient
+    memory is not available, the program will fail.
+
     If the -wm value is very small there is some extra overhead in doing many
     small chunks so setting it larger is better but it is a matter of
     diminishing returns.


apps/gdalwarp_lib.cpp
```diff
@@ -5754,11 +5754,20 @@ GDALWarpAppOptionsGetParser(GDALWarpAppOptions *psOptions,
         .action(
             [psOptions](const std::string &s)
             {
-                if (CPLAtofM(s.c_str()) < 10000)
+                if (EQUAL("unlimited", s.c_str()))
+                {
+                    psOptions->dfWarpMemoryLimit =
```
rouault, line 5759: FYI, we do have a CPLGetUsablePhysicalRAM() . Taking 80% of it might be a way to get a similar effect to unlimited without crashing. Although maybe 80% is too much because if you have a warp working buffer of that size, you'll need an input and output buffers of comparable size. So perhaps 25% would be the actual practical limit.

dbaston: I get a graceful exit if I don't have enough memory available: `ERROR 2: /home/dan/dev/gdal/alg/gdalwarpoperation.cpp, 694: cannot allocate 777600000000 bytes`

Would `-wm x%` be better? So I can still write `-wm 100%` if I want to live on the edge?

rouault: > I get a graceful exit if I don't have enough memory available: `ERROR 2: /home/dan/dev/gdal/alg/gdalwarpoperation.cpp, 694: cannot allocate 777600000000 bytes`

yes, because you are requesting way too much :-) There might be situation where you require almost too much, and might get caught by the OOM killer.

Anyway, I don't call for changes. "Unlimited" is pretty clear. This was a FYI that we have some infrastructure to get an idea of what is available (the name might be a bit misleading "CPLGetUsablePhysicalRAM". This is what the process can  use in total,  not necessarily what is remaining, and I wouldn't exclude that the OS might adveritze more memory than the OOM killer will really allow)

doc/source/programs/gdalwarp.rst
```diff
@@ -394,6 +394,9 @@ with control information.
     the "warp memory" allowed it will subdivide the chunk into smaller chunks
     and try again.
 
+    A -wm value of "unlimited" will prevent this chunking. If sufficient
```
rouault, line 397: we should though add a word of caution that "bigger is not necessarily better" and that people should only use that if they know what they are doing / that the default is very reasonable for most use cases (well maybe you have a different experience with your tests?). Even if people have enough RAM, there are definitely situations where a too big warping memory has undesirable consequences of processing speed, particularly for scenarios that might involve reprojection between polar projections and geographic where a very small target source window can translate to a huge source window

apps/gdalwarp_lib.cpp
jratike80: People tend to believe that for full speed 100% of anything is the best. If I understand right, in this case it may not be true. Should there a comment about that if there isn't already?

dbaston: > Should there a comment about that if there isn't already?

There's some discussion at https://gdal.org/en/latest/programs/gdalwarp.html#memory-usage

rouault: > https://gdal.org/en/latest/programs/gdalwarp.html#memory-usage

ah indeed, but might be good to link to it / add some short warning very clause to the mention of unlimited that will make people extremely excited ;-)
Or maybe ... if it is just an experimental flag ... just not document it .... ?

dbaston:
Going to close this and propose a different solution in a separate PR.

PR #10894, opened 2024-09-28
Warper: fix too lax heuristics about antimeridian warping for Avg/Sum/Q1/Q3/Mode algorithms

### LINKED ISSUE #10892: warp: resample 'average' (sometimes) results in 0 for shape (1,1)
### What is the bug?

I'm motivated to use the warper to get a singleton aggregate value, such as the mean elevation from this GEBCO overview. But with width and height set to 1, I get "0". Translate shows the correct value. 

```python
from osgeo import gdal
gdal.UseExceptions()
dsn = "vrt:///vsicurl/https://gebco2023.s3.valeria.science/gebco_2023_land_cog.tif?ovr=6"

resampleAlg = "average"

w = gdal.Warp("/vsimem/ovr6.tif", dsn, width = 1, height = 1,  resampleAlg = resampleAlg)
w.GetRasterBand(1).ReadAsArray()
# array([[0]], dtype=int16)

t = gdal.Translate("/vsimem/ovr6.tif", dsn, width = 1, height = 1,  resampleAlg = resampleAlg)
t.GetRasterBand(1).ReadAsArray()
# array([[-1899]], dtype=int16)

```

I've explored this by taking rasters filled with '1' from sizes 1,1 and up. I think the minimum size at which the problem occurs is ~9 but I haven't been able to pin that exactly due to other problems. 

Create a `x: 10, y: 9` raster filled with zeros, then compare warp to translate with resample 'average'. 

```python3
from osgeo import gdal
gdal.UseExceptions()

import numpy as np
width = 10
height = 9
driver = gdal.GetDriverByName("MEM")
ds = driver.Create("", width, height, 1, gdal.GDT_Int16)
ds.SetGeoTransform([0, 1, 0, 0, 0, -1])
ds.SetProjection("EPSG:4326")

band = ds.GetRasterBand(1)
data = np.full((height, width), 1, np.int16)
band.WriteArray(data)
resampleAlg = "average"

w = gdal.Warp("", ds, width = 1, height = 1, resampleAlg = resampleAlg, format = "MEM")
w.GetRasterBand(1).ReadAsArray()
#array([[0]], dtype=int16)

t = gdal.Translate("", ds, width = 1, height = 1, resampleAlg = resampleAlg, format = "MEM")
t.GetRasterBand(1).ReadAsArray()
# array([[1]], dtype=int16)

```

(I talk of minimum size because at `x: 9 y:8` warp gets the correct value. )

### Versions and provenance

```
gdalinfo --version
GDAL 3.10.0dev-082d0e78c54dd544a14947ac216fdcca8080595c, released 2024/06/20
```

### Additional context

I appreciate that Translate is the better choice in this situation, but I thought this might uncover potential other stuff, in the past I had used a target of 2x2 which I then took the average of those values to make this work. 

It's taken me a while to pin this down this far because of other issues, the above was run using the ubuntu small image

rouault:
Amends https://github.com/OSGeo/gdal/pull/6482

Fixes #10892

### DIFF
--- a/alg/gdalwarpkernel.cpp
+++ b/alg/gdalwarpkernel.cpp
@@ -6685,9 +6685,26 @@ static void GWKAverageOrModeThread(void *pData)
                 (nSrcXSize - padfX2[iDstX]) * poWK->dfXScale <
                     nThresholdWrapOverX)
             {
-                bWrapOverX = true;
-                std::swap(padfX[iDstX], padfX2[iDstX]);
-                padfX2[iDstX] += nSrcXSize;
+                // Check there is a discontinuity by checking at mid-pixel.
+                // NOTE: all this remains fragile. To confidently
+                // detect antimeridian warping we should probably try to access
+                // georeferenced coordinates, and not rely only on tests on
+                // image space coordinates. But accessing georeferenced
+                // coordinates from here is not trivial, and we would for example
+                // have to handle both geographic, Mercator, etc.
+                // Let's hope this heuristics is good enough for now.
+                double x = iDstX + 0.5 + poWK->nDstXOff;
+                double y = iDstY + poWK->nDstYOff;
+                double z = 0;
+                int bSuccess = FALSE;
+                poWK->pfnTransformer(psJob->pTransformerArg, TRUE, 1, &x, &y,
+                                     &z, &bSuccess);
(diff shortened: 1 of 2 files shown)


mdsumner:
Looks good Even I'm getting expected results and no regressions for the #6482 that I can see.

PR #10896, opened 2024-09-30
Add public gdal_typetraits.h header with gdal::CXXTypeTraits<T> and gdal::GDALDataTypeTraits<T>

rouault:


gcore/gdal_typetraits.h
```diff
@@ -0,0 +1,510 @@
...
+
+    static inline GDALExtendedDataType GetExtendedDataType()
+    {
+        return GDALExtendedDataType::CreateString();
+    }
+};
+
+//! @endcond
+
+/** Trait accepting a GDALDaaType and mapping it to correspond C++ type and
```
pjonsson, line 250: Missing a `t` in `GDALDataType`, and "correspond" should be "corresponding".

rouault: fixed

dbaston (approved):

PR #10902, opened 2024-09-30
CPLGetValueType(): do not recognize '01' as integer, but as string

rouault:
Fixes https://github.com/Toblerity/Fiona/issues/1454

### DIFF
--- a/apps/gdaltransform.cpp
+++ b/apps/gdaltransform.cpp
@@ -402,6 +402,11 @@ MAIN_START(argc, argv)
             if (fgets(szLine, sizeof(szLine) - 1, stdin) == nullptr)
                 break;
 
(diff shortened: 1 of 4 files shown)


theroggy:
@rouault small question (no underlying request for change)... I wondered, for this type of automatic detection of types, is it typically the first row of data that is inspected, or all rows, or a few rows?

rouault:
> I wondered, for this type of automatic detection of types, is it typically the first row of data that is inspected, or all rows, or a few rows?

It depends on drivers. The GML driver will do a full scan. The CSV driver doesn't do type detection by default, but has [options](https://gdal.org/en/latest/drivers/vector/csv.html#open-options) to enable it and limit the number of scanned bytes.

PR #10924, opened 2024-10-03
WMS: try to be robust to servers using CRS instead of SRS in 1.1.1 mode (and warn about that)

### LINKED ISSUE #10922: GDAL does not get subdatasets for a WMS service
### What is the bug?

GDAL does not return any dataset for the following WMS service:
`https://mapy.geoportal.gov.pl/wss/service/PZGIK/BDOO/WMS/aktualne`
The service contains 45 layers.

`gdalinfo "WMS:https://mapy.geoportal.gov.pl/wss/service/PZGIK/BDOO/WMS/aktualne?service=wms&request=getcapabilities"`

### Versions and provenance

Windows 11 / GDAL 3.9.2, released 2024/08/13 from Conda-Forge

### Additional context

Perhaps this issue is also a good moment for refactor the process of getting subdatasets from WMS services, as some of us have been discussing on the [gdal mailing list](https://www.mail-archive.com/gdal-dev@lists.osgeo.org/msg41499.html).

Current GetSubdataset implementation for WMS returns the path only to one combination of layer-SRS, where WMS service provides individual list of CRS for each layer and arbitrarily selected SRS is not usually the best (default) SRS for the presented area.

rouault:
Fixes #10922

### DIFF
--- a/frmts/wms/wmsmetadataset.cpp
+++ b/frmts/wms/wmsmetadataset.cpp
@@ -380,10 +380,34 @@ void GDALWMSMetaDataset::ExploreLayer(CPLXMLNode *psXML,
     const char *pszMaxXLocal = nullptr;
     const char *pszMaxYLocal = nullptr;
 
-    const char *pszSRSTagName =
-        VersionStringToInt(osVersion.c_str()) >= VersionStringToInt("1.3.0")
-            ? "CRS"
-            : "SRS";
+    const bool bIsWMS130 =
+        VersionStringToInt(osVersion.c_str()) >= VersionStringToInt("1.3.0");
+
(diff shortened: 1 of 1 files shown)


jratike80:
Are you sure that we should try to support every WMS server that does not follow the standard in some peculiar and uncommon way? I think we should not.

rouault:
> I think we should not.

yeah, hard to tell where to stop... Part of the mission of GDAL is to support non-conformant stuff when it is reasonable to do so. Maybe that will save us a support ticket next time someone hits that issue... ? It appears to be a ESRI WMS server

jratike80:
Oh no, it may really be much more easy to make a workaround than to make ESRI admit that their software has a bug.

PR #10932, opened 2024-10-04
docker: use ninja as build system

pjonsson:
## What does this PR do?

There are some nice features in CMake
that are limited to Ninja only, for
example JOB_POOL_LINK and Multi-config:

https://cmake.org/cmake/help/latest/prop_tgt/JOB_POOL_LINK.html
https://cmake.org/cmake/help/latest/generator/Ninja%20Multi-Config.html
https://www.scivision.dev/cmake-ninja-multi-config/

The first limits the number of linking
processes running concurrently, and the second
avoids having to re-generate the build files
to make a debug/release build.

Ninja is also rumored to be slightly faster
than make, and that seems to hold for GDAL
as well, a simple benchmark (time -p) with
a warm cache (second time image is built)
for ubuntu-small on a 48 core machine
(3 GHz dual Xeon Gold 6136):

PROJ:

make:

real 65.39
user 400.61
sys 34.91

ninja:

real 64.84
user 413.17
sys 34.97

GDAL:

make:

real 86.40
user 2466.07
sys 180.64

ninja:

real 82.43
user 2471.69
sys 169.74

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:

### DIFF
--- a/docker/alpine-normal/Dockerfile
+++ b/docker/alpine-normal/Dockerfile
@@ -6,7 +6,7 @@
 # or licensed under MIT (LICENSE.TXT) Copyright 2019 Even Rouault <even.rouault@spatialys.com>
 
 ARG ALPINE_VERSION=3.20
-FROM alpine:${ALPINE_VERSION} as builder
+FROM alpine:${ALPINE_VERSION} AS builder
 
 # Derived from osgeo/proj by Howard Butler <howard@hobu.co>
 LABEL maintainer="Even Rouault <even.rouault@spatialys.com>"
@@ -296,7 +296,7 @@ RUN if test "${GDAL_VERSION}" = "master"; then \
        ;do rm $i; done)
 
 # Build final image
-FROM alpine:${ALPINE_VERSION} as runner
+FROM alpine:${ALPINE_VERSION} AS runner
 
 RUN date
 
--- a/docker/alpine-normal/Dockerfile
+++ b/docker/alpine-normal/Dockerfile
@@ -62,6 +62,7 @@ RUN apk add --no-cache \
     make \
     mariadb-connector-c-dev \
     netcdf-dev \
+    ninja-build \
     odbc-cpp-wrapper-dev \
     ogdi-dev \
     openexr-dev \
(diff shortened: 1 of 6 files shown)


pjonsson:
I think this is a slight improvement, but I'm not the one who spends every day in the GDAL code so my opinion shouldn't count for much. Would you like me to complete this PR so all docker images use ninja for building the cmake-based projects?

rouault:
> Would you like me to complete this PR so all docker images use ninja for building the cmake-based projects?

yes, that's fine

pjonsson:
Ok, I think this is ready, all images build locally for me.

I don't know anything about Alpine, but I get a warning when installing ninja that I probably want samurai instead of ninja. I have used ninja extensively for almost a decade and I have never heard of samurai, so it might be a fine piece of software, but I don't think GDAL's docker images is the place for me to explore that.

The CI takes a long time here so I added a commit that has nothing to do with the title of the PR. The commit fixes the docker warning about different casing of `FROM` and `as`.

rouault:
> but I get a warning when installing ninja that I probably want samurai instead of ninja

ok, so "samurai" is https://pkgs.alpinelinux.org/package/edge/main/x86_64/samurai which is a "ninja-compatible build tool written in C", and the alpine samurai package actually alias "ninja" to "samurai". Must be a packager taste to prefer a C port of C++ ninj-a-build...
I believe it would be better to install ninja-build for the Alpine images, right at the beginning, and set the PATH, instead of doing it multiple times, and removing it (and the same for cmake)

pjonsson:
Updated Alpine to install cmake and ninja-build first and never remove the package.

I rebuilt alpine-small before pushing, but I hit the docker cache when rebuilding alpine-normal and I don't want to clear my local docker cache, so that image is not built before I pushed. I guess the CI for this PR will let me know if I missed something.

rouault:
> I guess the CI for this PR will let me know if I missed something.

unfortunately no. Building the Docker images is not covered by CI... (would be too long)

rouault:
> Building the Docker images is not covered by CI... (would be too long)

I've actually added a job (triggered only by changes in docker/ and doing x86_64 builds only) in https://github.com/OSGeo/gdal/pull/10947 . Can you rebase your pull request on top of latest master so it runs on your pull request ? Thanks

pjonsson:
This morning I realized I could dirty the cache by adding some non-existent path to PATH, so I've done that, and both alpine images build with the changes.

I rebased on top of latest master, the CI does not make mistakes.

PR #10945, opened 2024-10-05
OGRFeature: SetXXX() methods: more informative warning messages reporting field name and value for out-of-range values

### LINKED ISSUE #10915: Improve accuracy warning when types does not match
### What is the bug?

Hi, actually when we tranform/import/export some data with GDAL, it handles the conversion types, in cases where the selected output type (by the user or default gdal) does not match, we can get messages like:

```Warning 1: Integer overflow occurred when trying to set 32bit field.```

If the output is PG with the Varchar type, it could cause a similar issue for long strings.

The warning is great, is just that with it we can't know which column is triggering this, in case we want to change the column type to keep precision on the data.

Would be nice have something like this:

```Warning 1: Integer overflow occurred when trying to set 32bit field on column 'col_1'.```

This is not only for PG/Sqlite, is in general, the complex part is for some files, SQLite has Tables->Columns, PG has Schema->Table->Columns, the idea is get the needed info to reach the column that triggers this.

Other formats could also need other aspects to reach that column.

Thx!

Any SQLite file with a row that has a value of the max possible integer, which will overflow PG Int32.

ogr2ogr PG:"dbname='postgres' host='127.0.0.1' port='5432' user='postgres' password='postgres'" some.sqlite
```Warning 1: Integer overflow occurred when trying to set 32bit field.```

### Versions and provenance

GDAL 3.9.1, released 2024/06/22

### Additional context

_No response_

rouault:
Fixes #10915

### DIFF
--- a/ogr/ogrfeature.cpp
+++ b/ogr/ogrfeature.cpp
@@ -4440,8 +4440,8 @@ void OGRFeature::SetField(int iField, const char *pszValue)
             }
             else
             {
-                const char *papszValues[2] = {pszValue, nullptr};
-                SetField(iField, const_cast<char **>(papszValues));
+                const char *const papszValues[2] = {pszValue, nullptr};
+                SetField(iField, papszValues);
             }
         }
     }
@@ -5062,8 +5062,7 @@ void OGR_F_SetFieldStringList(OGRFeatureH hFeat, int iField,
 {
     VALIDATE_POINTER0(hFeat, "OGR_F_SetFieldStringList");
 
-    OGRFeature::FromHandle(hFeat)->SetField(
-        iField, const_cast<const char *const *>(papszValues));
+    OGRFeature::FromHandle(hFeat)->SetField(iField, papszValues);
 }
 
 /************************************************************************/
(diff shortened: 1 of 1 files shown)


latot:
Hi! great PR!

I think would be nice, for the integer cases (or similar ones), add if they are signed/unsigned ones in the message.

Thx!

rouault:
> I think would be nice, for the integer cases (or similar ones)

the actual value that fails to be set is put in the error message

latot:
Hi! yes, I means not the value's symbol (if is signed or not), the objective field type, in C++ a Integer 64 is Signed by default, while in other langs is not always the case, so I think would be good say explicitly if the field that is being used if is a signed or unsigned one.

rouault:
> This is also because this warning is for the final user, who does not needs to know in which lang this was done, helps for a clear description.

I let it to others to further tweak that if that's needed. I believe this PR as it is sufficient.

Signedness of OFTInteger/OFTInteger64 explicited in 46ad0f1fd14915b3a59740697a1065f6051e2572

PR #10948, opened 2024-10-05
VRTWarpedDataset::IRasterIO(): avoid a memory allocation and pixel copy when possible

rouault:


elpaso (commented):
Same changes have been reviewed in https://github.com/OSGeo/gdal/pull/10952/

PR #10950, opened 2024-10-06
docker: fix ccache

pjonsson:
The rsync commands use the wrong
directory, so the commands fail
and print an error message, but
the build continues.

Fix the commands so they work,
but leave the missing error
detection as is.

### DIFF
--- a/docker/alpine-normal/Dockerfile
+++ b/docker/alpine-normal/Dockerfile
@@ -143,12 +143,12 @@ RUN mkdir proj \
     && cd proj \
     && if test "${RSYNC_REMOTE}" != ""; then \
         echo "Downloading cache..."; \
-        rsync -ra ${RSYNC_REMOTE}/proj/$(uname -m)/ $HOME/; \
+        mkdir -p "$HOME/.cache"; \
+        rsync -ra ${RSYNC_REMOTE}/proj/$(uname -m)/ $HOME/.cache/; \
         echo "Finished"; \
         export CC="ccache gcc"; \
         export CXX="ccache g++"; \
-        mkdir -p "$HOME/.ccache"; \
-        export PROJ_DB_CACHE_DIR="$HOME/.ccache"; \
+        export PROJ_DB_CACHE_DIR="$HOME/.cache"; \
         ccache -M 100M; \
     fi \
     # IPO disabled since it crashes with gcc 13.2.1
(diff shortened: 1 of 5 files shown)


pjonsson:
The reason I discovered this was because I wanted to take a look at whether the builtin cache mechanism of buildx could be used instead of the rsync daemon that is currently used. I've browsed too many issues in the docker repositories, but the short summary of the discussions in Docker is that the build network mode that #7571 works around by disabling buildx is unlikely to come back.

Right now, multiple gdal repositories can share a single ccache for docker builds in the home directory, but that cache is split up on a per-image basis. Is the shared cache for multiple gdal repositories essential and must be preserved? If it's fine with a per-repository ccache I think I can get the current functionality from just using bind mounts, and completely remove the separate rsync container.

A second question since I just looked at the dockerfile for building: is there a reason to not limit the installation of the ccache package to non-release builds, and always enable `-DUSE_CCACHE=ON` when calling cmake in the dockerfiles? I assume that would eliminate some of the setting of environment variables and things would just work out of the box with the default values.

rouault:
> Is the shared cache for multiple gdal repositories essential and must be preserved?

this is nice to have for people doing frequent builds to save some time

>  If it's fine with a per-repository ccache I think I can get the current functionality from just using bind mounts, and completely remove the separate rsync container.

I believe the rationale for the current design was to make sure that each image doesn't compete with another one w.r.t cache size

> is there a reason to not limit the installation of the ccache package to non-release builds, and always enable -DUSE_CCACHE=ON when calling cmake in the dockerfiles?

I'm a bit lost with all the negations in your question ;-) So your question is: "why not enable ccache on release builds?" I believe the rationale was to make sure to have 'perfect' / 'clean' builds for release

pjonsson:
> I'm a bit lost with all the negations in your question ;-)

Sorry, I'll try to rephrase: why not remove the special CC/CXX exports/wrappers and always pass `-DUSE_CCACHE=ON` to cmake, but remove the ccache package from release builds?

And that question wasn't meant for additional things to this PR, it was more of a "since I'm looking at this line right now, that question crossed my mind".

rouault:
> why not remove the special CC/CXX exports/wrappers and always pass `-DUSE_CCACHE=ON` to cmake

hum, I believe I found USE_CCACHE didn't work well with the cross-compilers hence those workarounds

> but remove the ccache package from release builds?

could potentially be done. A bit of a micro-optimization

PR #10955, opened 2024-10-07
docker: use buildkit cache mounts

pjonsson:
The "docker build --network" parameter
no longer accepts general networks
when using BuildKit, and BuildKit has
been the default for all platforms
except Windows since Docker 23.0.

One value that is still accepted
by "docker build --network" is "host",
so use the host network when building
images, and run the rsync cache container
on the host network as well.

Being able to use BuildKit allows
us to use cache mounts for ccache,
so the second consecutive invocation
of docker build on a machine will
already have a warm cache and rsync
will not copy anything.


pjonsson:
This is a minimal proof of concept PR for how we can keep the rsync cache daemon working without having to disable BuildKit in Docker. I have tried to minimize the commit to highlight the changes, a completed PR would perform similar changes on the other images, and remove some unused code in utils.sh.

This PR uses the cache mounts which is a small performance improvement for local builds since rsync doesn't have to perform any work to keep the cache in sync. BuildKit has other features that could be interesting too, for example being able to lazily build the target image without having to build all other images (even unused ones) defined in a multipart Dockerfile.

For completeness: I mentioned it in some other PR already but  I've read numerous tickets on Docker and the network parameter to `docker build` is unlikely to come back, especially anytime soon.

Does this look like an acceptable solution to you? If so, would you like to preserve the separation of the PROJ cache so ubuntu-small and ubuntu-full has their own cache?

rouault:
> Does this look like an acceptable solution to you?

Probably... I'm just ignorant of --mount=type=cache : where is this cache actually stored? how do you wipe it ?

pjonsson:
I'm fairly sure that the mount cache-type is just like any other thing that can be cached in docker (ENV command, RUN command, etc.), so on a conceptual level I think the answer is "somewhere inside buildkit". If you're talking about physical location, it's my `/var/lib/docker/overlay2` that grows over time, but I doubt that path is mandated in any specification, it's just where docker chose to store things.

You clear the cache with `docker system prune`, and despite the scary name of the command, I think it's fairly harmless as long as you run it without parameters (you will be prompted for confirmation).

rouault:
(read the docs build failure unrelated, and fixed in latest master)

rouault:
@pjonsson Would you mind trying doing --release builds that use docker buildx to confirm things work fine with them too ? (as this isn't teste by CI)

pjonsson:
Tested building all images with `--with-multi-arch --gdal HEAD --release`. That found the `PROJ_DB_CACHE_PARAM` variable being undefined, so trying uncovered an error, but probably not the one you were looking for.

On a technical note, from Docker 23.0, I don't think there is much difference under the hood between `docker build` and `docker buildx`.

rouault:
> but probably not the one you were looking for.

the most interesting errors are the ones you don't know in advance :-)

Thanks for all your efforts on improving this part of the infrastructure!

PR #10954, opened 2024-10-07
Doc: use 'version_selector' instead of 'display_version' which is deprecated by sphinx-rtd-theme 3.0.0

rouault:

PR #10956, opened 2024-10-07
Doc: stick to sphinx_rtd_theme 2.x for now

rouault:


rouault:
hum, despite sticking to rtd_them 2.x, the version selector still disappears as with 3.0 (https://github.com/OSGeo/gdal/pull/10954#issuecomment-2397158054), so there have been other changes on RTD infrastructure.

Dropping this PR...

PR #10962, opened 2024-10-07
Arrow/Parquet: use recommended item names for GeoArrow [multi]line, [multi]polygon, multipoint

rouault:


mdsumner:
excellent looks good , getting expected schema from nanoarrow now

elpaso (approved):
Can you please add a test for this change?

rouault:
> Can you please add a test for this change?

done

PR #10952, opened 2024-10-06
VRTWarpedDataset::IRasterIO(): optimize I/O when requesting whole image at a resolution that doesn't match an overview

rouault:
Fixes https://github.com/rasterio/rasterio/issues/3203

With that optimization:

```
$ gdalwarp /vsicurl/https://sentinel-cogs.s3.us-west-2.amazonaws.com/sentinel-s2-l2a-cogs/15/T/VK/2023/10/S2B_15TVK_20231008_0_L2A/TCI.tif in.vrt

$ CPL_DEBUG=VSICURL GDAL_INGESTED_BYTES_AT_OPEN=32768 GDAL_HTTP_MULTIRANGE=YES GDAL_HTTP_MERGE_CONSECUTIVE_RANGES=YES GDAL_DISABLE_READDIR_ON_OPEN=EMPTY_DIR python -c "from osgeo import gdal; ds = gdal.Open('in.vrt'); ds.ReadRaster(0, 0, ds.RasterXSize, ds.RasterYSize, ds.RasterXSize // 5, ds.RasterYSize // 5)"
/home/even/gdal/gdal/build_cmake/swig/python/osgeo/gdal.py:311: FutureWarning: Neither gdal.UseExceptions() nor gdal.DontUseExceptions() has been explicitly called. In GDAL 4.0, exceptions will be enabled by default.
  warnings.warn(
VSICURL: GetFileSize(https://sentinel-cogs.s3.us-west-2.amazonaws.com/sentinel-s2-l2a-cogs/15/T/VK/2023/10/S2B_15TVK_20231008_0_L2A/TCI.tif)=200959614  response_code=200
VSICURL: Downloading 0-32767 (https://sentinel-cogs.s3.us-west-2.amazonaws.com/sentinel-s2-l2a-cogs/15/T/VK/2023/10/S2B_15TVK_20231008_0_L2A/TCI.tif)...
VSICURL: Got response_code=206
VSICURL: GetFileList(/vsicurl/https://sentinel-cogs.s3.us-west-2.amazonaws.com/sentinel-s2-l2a-cogs/15/T/VK/2023/10/S2B_15TVK_20231008_0_L2A)
VSICURL: GetFileSize(https://sentinel-cogs.s3.us-west-2.amazonaws.com/sentinel-s2-l2a-cogs/15/T/VK/2023/10/S2B_15TVK_20231008_0_L2A/TCI.xml)=0  response_code=404
[...]

PR #10964, opened 2024-10-08
Use stricter typing for void* pTransformerArg

rouault:
@elpaso Implements what you discussed in https://github.com/OSGeo/gdal/pull/10952#discussion_r1791522527

The last commit ("Make GDALTransformerArg a typedef of an anonymous struct GDALTransformerArgOpaque*") is slightly backwards incompatible


rouault:
> The last commit ("Make GDALTransformerArg a typedef of an anonymous struct GDALTransformerArgOpaque*") is slightly backwards incompatible

We might also postpone this to 3.11, as there's no emergency in merging this, especially if @dbaston would want a cleaner way of doing proper C++ typing.

PR #10965, opened 2024-10-08
Doc: Add description of gdalwarp source pixel weighting

dbaston:


doc/source/programs/gdalwarp.rst
```diff
@@ -508,9 +508,26 @@ Once the dimensions of the output image have been determined,
...
+performs the resampling algorithm selected with :option:`-r`.  Depending on the
+resampling algorithm and relative scales of the source and destination rasters,
+source pixels may be weighted either according to the approximate fraction of
+the source pixel that is covered by the destination pixel (e.g., "mean" and
+"sum" resampling), or by horizontal and vertical Cartesian distances between
+the center of the source pixel and the center of the target pixel (e.g.,
+bilinear or cubic spline resampling). In the latter case, the relative weight
+of an individual source pixel is determined by the product of the weights
+determined for its row and column; the diagonal Cartesian distance is not
+calculated.
```
rouault, line 530: I'm wondering if the last part of this sentence ("the diagonal Cartesian distance is not calculated.") is strictly needed, but I don't mind either to let it. As far as I understand, we use fairly standard bilinar/bicubic resampling. I'm not sure what a method taking into diagonal distances would look like

dbaston: I might have expected that each of the source pixels was weighted according to the distance between the source pixel center and the destination pixel center (reprojected to the source image space.) The way GDAL is doing it is more efficient since it calls the weighting function fewer times. That may be the standard approach, I haven't looked into what any other software is doing.

I agree that it seems like excessive detail but I didn't see it described elsewhere, so ...

dbaston:
If an image would be helpful I could clean up and SVGify this example I was looking at that shows source pixel values and weights using cubic spline resampling. (The red point is the destination pixel transformed back into source image space)

image

rouault (approved):

rouault:
> If an image would be helpful I could clean up and SVGify this example I was looking at that shows source pixel values (bottom) and weights (top) using cubic spline resampling.

That could be useful, but perhaps not in the gdalwarp documentation page itself, since I suspect it would distract most end users. Perhaps in alg/  and referenced from the Doxygen doc of the warping code?

rouault:
The RtD build failure has been solved in master

PR #10967, opened 2024-10-08
Remove "trademarked" terms

nyalldawson:
🤬


rouault:
Thanks
(out of curiosity, did someone make some fuss about that... ? no need to name them)

nyalldawson:
> did someone make some fuss about that... ?

Yes, they have.

rouault:
> Yes, they have.

ok, so for good measure, let's get rid of undue advertizing: d306d4c161

PR #10972, opened 2024-10-09
RFC102 implementation: embedded resources

rouault:
Implements https://github.com/OSGeo/gdal/pull/10913


doc/source/development/building_from_source.rst
```diff
@@ -224,6 +224,36 @@ All cached entries can be viewed using ``cmake -LAH`` from a build directory.
...
+
+.. option:: USE_ONLY_EMBEDDED_RESOURCE_FILES=ON/OFF
+
+    .. versionadded:: 3.11
+
+    Even if EMBED_RESOURCE_FILES=ON, by default GDAL will still try to locate
+    resource files on the file system, and fallback to the embedded version if
+    not found. By setting USE_ONLY_EMBEDDED_RESOURCE_FILES=OFF, no attempt
+    at localing resource files on the file system is made.
+
```
jratike80, line 256: This feels odd. For me USE_ONLY_EMBEDDED_RESOURCE_FILES=ON would make more sense.

rouault: good catch. Was what I meant indeed. Fixed

rouault:
PROJ specificities are now covered by a dedicated PROJ RFC-8: https://github.com/OSGeo/PROJ/pull/4274

ogr/ogrsf_frmts/gml/embedded_resources_gen2.c
```diff
@@ -0,0 +1,36 @@
+// File generated by generate_embedded_resources.py. DO NOT EDIT
```
pjonsson, line 1: My experience with keeping generated sources under version control hasn't been great, so consider the alternative of having CMake generate these files somewhere inside `${CMAKE_CURRENT_BINARY_DIR}`. (Not suggesting to replace the Python script with CMake-code, just that CMake sets up a task that calls the Python script to generate the files.)

doc/source/development/building_from_source.rst
```diff
@@ -224,6 +224,36 @@ All cached entries can be viewed using ``cmake -LAH`` from a build directory.
...
+    Default is OFF for shared library builds (BUILD_SHARED_LIBS=ON), and ON
+    for static library builds (BUILD_SHARED_LIBS=OFF).
+    When ON, resource files needed by GDAL will be embedded into the GDAL library
+    and/or relevant plugins.
+
+.. option:: USE_ONLY_EMBEDDED_RESOURCE_FILES=ON/OFF
+
+    .. versionadded:: 3.11
+
+    Even if EMBED_RESOURCE_FILES=ON, by default GDAL will still try to locate
```
pjonsson, line 252: There are currently two options, either fallback to the use the embedded resource, or refuse to use the filesystem. Is there a "user-extensible" use-case with resources where "use the embedded resource if present, but fallback to look in the filesystem" would make sense?

And language wise, my gut feeling is that "by default" should be moved to after "on the file system" in this sentence.

ogr/ogrsf_frmts/gml/embedded_resources_gen2.c
rouault: I'll let that as an exercise for later. We do not add new .gfs files frequently. Last time was years ago

doc/source/development/building_from_source.rst
rouault: > where "use the embedded resource if present, but fallback to look in the filesystem" would make sense?

I fail to see the use case. "Use the embedded resource if present": the presence of the embedded ressource is fully determined by EMBED_RESOURCE_FILES=ON/OFF.

> And language wise, my gut feeling is that "by default" should be moved to after "on the file system" in this sentence.

thanks, fixed

ogr/ogrsf_frmts/gml/embedded_resources_gen2.c
rouault: Actually, we don't have a Python build requirement currently

PR #10974, opened 2024-10-09
docker: enable more caching

pjonsson:
This changes a lot of things all over the place, but the big picture is:
- Release builds pass `--no-cache` to Docker so they don't accidentally pick up something from the Docker cache. For local experimentation, that can be overridden by passing `--docker-cache` to build.sh.
- The deb packages+lists are cached in Docker for non-release builds. This avoids fetching the same thing from the network repeatedly. The cache for apt can only be used by one build at the time so that is mounted with `sharing=locked`. To avoid blocking concurrent docker image builds, apt-get calls were put in a separate `RUN` statement from a long-running compilation.
- Separation of rsync and ccache behavior, which lets the user disable the gdal rsync daemon but still benefit from ccache and the Docker cache.

It's always tempting to change everything at once, but I have tried to restrain myself so this PR would still be reviewable. Each commit does one thing, and viewing the commit without showing whitespace changes should give a reasonably sized change.

I have built all images locally with both `--release` and without it.

### DIFF
--- a/docker/util.sh
+++ b/docker/util.sh
@@ -17,7 +17,7 @@ fi
 
 usage()
 {
-    echo "Usage: build.sh [--push] [--docker-repository repo] [--tag name] [--gdal tag|sha1|master] [--gdal-repository repo] [--proj tag|sha1|master] [--release] [--docker-cache|--no-docker-cache]"
+    echo "Usage: build.sh [--push] [--docker-repository repo] [--tag name] [--gdal tag|sha1|master] [--gdal-repository repo] [--proj tag|sha1|master] [--release] [--docker-cache|--no-docker-cache] [--no-rsync-daemon]"
     # Non-documented: --test-python
     echo ""
     echo "--push: push image to docker repository (defaults to ghcr.io)"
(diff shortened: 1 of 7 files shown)


docker/util.sh
```diff
@@ -194,15 +211,17 @@ echo "Using GDAL_REPOSITORY=${GDAL_REPOSITORY}"
 IMAGE_NAME="${TARGET_IMAGE}-${TAG_NAME}"
 REPO_IMAGE_NAME="${DOCKER_REPO}/${IMAGE_NAME}"
 
+BUILD_ARGS=(
+    $DOCKER_CACHE_PARAM \
```
rouault, line 215: it seems to me that DOCKER_CACHE_PARAM might not be initialized in the non-release case if neither --docker-cache nor --no-docker-cache have been specified

rouault:
shellcheck is complaining:
```
shellcheck -e SC2086,SC2046,SC2164,SC2054,SC2129 $(find . -name '*.sh' -a -not -name ltmain.sh -a -not -wholename "./autotest/*" -a -not -wholename "./.github/*")

In ./docker/util.sh line 216:
    $DOCKER_CACHE_PARAM \
    ^-----------------^ SC2206 (warning): Quote to prevent word splitting/globbing, or split robustly with mapfile or read -a.

For more information:
  https://www.shellcheck.net/wiki/SC2206 -- Quote to prevent word splitting/g...
```

PR #10980, opened 2024-10-09
PG: avoid error when the original search_path is empty

rouault:
Fixes https://lists.osgeo.org/pipermail/gdal-dev/2024-October/059608.html


dgboss:
Thanks for the quick fix!

PR #10982, opened 2024-10-10
heif: add tile reading

bradh:
## What does this PR do?

libheif recently added tiling support (https://github.com/strukturag/libheif/wiki/Read-and-Writing-Tiled-Images) as part
of the OGC TB-20 activities.

This provides detection of the new API (not yet released) and use of it where available.

N/A

## Tasklist

 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments


cmake/helpers/CheckDependentLibraries.cmake
```diff
@@ -449,6 +449,23 @@ gdal_check_package(MONGOCXX "Enable MongoDBV3 driver" CAN_DISABLE)
...
+    {
+        struct heif_image_tiling tiling;
+        return 0;
+    }
+    "
+    LIBHEIF_SUPPORTS_TILES
+)
+if (LIBHEIF_SUPPORTS_TILES)
+    set(LIBHEIF_SUPPORTS_TILES 1)
+    add_definitions(-DLIBHEIF_SUPPORTS_TILES)
```
rouault, line 466: it would be if we could avoid a general directive like ``add_definitions()`` that will apply to the whole GDAL code base and use instead something like:
```
    set_property(TARGET HEIF::HEIF APPEND PROPERTY
                 INTERFACE_COMPILE_DEFINITIONS "LIBHEIF_SUPPORTS_TILES")
```

rouault (commented):
Would you mind editing one of the CI configurations to build libheif master from source so this gets a bit of testing? Lets say .github/workflows/alpine/Dockerfile.ci (installation of libheif-dev package to be removed)

And corresponding testing in autotest/gdrivers/heif.py would be great too

frmts/heif/heifdataset.cpp
```diff
@@ -637,14 +655,88 @@ GDALHEIFRasterBand::GDALHEIFRasterBand(GDALHEIFDataset *poDSIn, int nBandIn)
...
+    int nStride = 0;
+    const uint8_t *pSrcData = heif_image_get_plane_readonly(
+        hImage, heif_channel_interleaved, &nStride);
+    if (eDataType == GDT_Byte)
+    {
+        for (int y = 0; y < nBlockYSize; y++)
+        {
+            for (int x = 0; x < nBlockXSize; x++)
+            {
+                (static_cast<GByte *>(pImage))[y * nBlockXSize + x] =
```
rouault, line 719: is the a risk that nBlockXSize * nBlockYSize might overflow INT_MAX ? If so, then y * nBlockXSize + x here and in the else clause below should be rewritten as ``static_cast<size_t>(y) * nBlockXSize + x``

frmts/heif/heifdataset.cpp
```diff
@@ -637,14 +655,88 @@ GDALHEIFRasterBand::GDALHEIFRasterBand(GDALHEIFDataset *poDSIn, int nBandIn)
...
+#endif
+                                       : heif_chroma_interleaved_RGBA),
+        decode_options, nBlockXOff, nBlockYOff);
 
+    if (err.code != heif_error_Ok)
+    {
+        CPLError(CE_Failure, CPLE_AppDefined, "%s",
+                 err.message ? err.message : "Cannot decode image");
+        poGDS->m_bFailureDecoding = true;
+        return CE_Failure;
```
rouault, line 707:
```suggestion
        heif_decoding_options_free(decode_options);
        return CE_Failure;
```

rouault:
ah, I now realize that my suggestion of updating the alpine CI wasn't perhaps the best. It is currently broken in master due to another reason (an libxml2 update that breaks libspatialite)
Perhaps the ubuntu 24.04 then

frmts/heif/heifdataset.cpp
bradh: Should be fixed.

Out of interest, would defining `x` and `y` to be `size_t` instead of `int` be an acceptable alternative?

rouault: you could do that as well. up to you. You'll just get a compiler warning when comparing x/y to nBlockX/YSize about mismatched signedness, requiring a cast of nBlockX/YSize to size_t

cmake/helpers/CheckDependentLibraries.cmake
bradh: Should be fixed.

cmake/helpers/CheckDependentLibraries.cmake
```diff
@@ -449,6 +449,23 @@ gdal_check_package(MONGOCXX "Enable MongoDBV3 driver" CAN_DISABLE)
...
+    int main()
+    {
+        struct heif_image_tiling tiling;
+        return 0;
+    }
+    "
+    LIBHEIF_SUPPORTS_TILES
+)
+if (LIBHEIF_SUPPORTS_TILES)
+    set(LIBHEIF_SUPPORTS_TILES 1)
```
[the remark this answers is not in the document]
bradh: Should be fixed.

bradh:
Test failures look legitimate (although gdal_translate shows it correctly). Investigating.

frmts/heif/heifdataset.cpp
```diff
@@ -637,14 +655,92 @@ GDALHEIFRasterBand::GDALHEIFRasterBand(GDALHEIFDataset *poDSIn, int nBandIn)
                                         "IMAGE_STRUCTURE");
     }
 #endif
+
+#ifdef LIBHEIF_SUPPORTS_TILES
+    nBlockXSize = poDSIn->m_tiling.tile_width;
```
rouault, line 660: the 0x0 block size error must come from m_tiling.tile_width/tile_heigh being 0. Perhaps that's the way libheif signals a non-tiled dataset? In which case you might want to go to to the scanline oriented approach below and potentially use the "old" IReadBlock() way too ? Just guessing, having not checked how libheif deals with tiling...

bradh: Its is supposed to be uniform handling (i.e. non-tiled is a single tile).

The problem was in the overview data set - I hadn't added tiling for that. Should now be fixed.

bradh:
Is there a pattern for enabling a pytest based on the cmake configuration results?

rouault:
> Is there a pattern for enabling a pytest based on the cmake configuration results?

Kind of. For example the ADBC driver defines a driver metadata item for a compile-time capability:
https://github.com/OSGeo/gdal/pull/11003/files#diff-a74e58abe10ad4241b0610648027d423ce96a8d93ac32786e5fc7d149d7e4a1fR95

and then the tests check for it with https://github.com/OSGeo/gdal/pull/11003/files#diff-12f03fcbd903ace6895703f38de777aad70b78a5b89fe10f9f7ddeb672237f77R21 and pytest.skip() if needed

bradh:
> > Is there a pattern for enabling a pytest based on the cmake configuration results?
> 
> Kind of. For example the ADBC driver defines a driver metadata item for a compile-time capability: https://github.com/OSGeo/gdal/pull/11003/files#diff-a74e58abe10ad4241b0610648027d423ce96a8d93ac32786e5fc7d149d7e4a1fR95
> 
> and then the tests check for it with https://github.com/OSGeo/gdal/pull/11003/files#diff-12f03fcbd903ace6895703f38de777aad70b78a5b89fe10f9f7ddeb672237f77R21 and pytest.skip() if needed

Thanks. Added that.

autotest/gdrivers/heif.py
```diff
@@ -117,6 +117,26 @@ def test_heif_rgba_16bit():
...
+
+    ds = gdal.Open("data/heif/uncompressed_comp_RGB_tiled.heif")
+    assert ds
+    assert ds.RasterXSize == 30
+    assert ds.RasterYSize == 20
+    assert ds.RasterCount == 3
+    assert ds.GetRasterBand(1).DataType == gdal.GDT_Byte
+    assert ds.GetRasterBand(1).GetBlockSize() == [15, 5]
+    assert ds.GetRasterBand(2).GetBlockSize() == [15, 5]
+    assert ds.GetRasterBand(3).GetBlockSize() == [15, 5]
```
rouault, line 137: it could be worth to actually read pixel values, to make sure IReadBlock() is correctly implemented. I understand that libheif might not always return exact pixel values but we could use statistics-based testing like done above in test_heif_exif_endian()

bradh: This should be occurring through the existing tests (since they all go through the same API, if its available).

I guess there is a corner case where it works for the single-tile case but is off for the M by N case though.

rouault: > I guess there is a corner case where it works for the single-tile case but is off for the M by N case though.

yes, that was my thinking that you could possibly have got wrong taking into account non-zero nBlockXOff/nBlockYOff. Perhaps not now, but the next guy in 5 years that will touch that code ;-)

autotest/gdrivers/heif.py
```diff
@@ -117,6 +117,248 @@ def test_heif_rgba_16bit():
...
+    assert (
```
rouault, line 138: GDAL builds don't strictly require numpy / gdal_array support so:
```suggestion
    pytest.importorskip("osgeo.gdal_array")
    assert (
```

rouault:
other_checks CI failure unrelated and fixed in master per https://github.com/OSGeo/gdal/pull/11050

Squashing & merging

autotest/gdrivers/heif.py
bradh: Thanks for this.

frmts/heif/heifdrivercore.cpp
```diff
@@ -86,6 +86,9 @@ void HEIFDriverSetCommonMetadata(GDALDriver *poDriver)
 
     poDriver->pfnIdentify = HEIFDriverIdentifySimplified;
     poDriver->SetMetadataItem(GDAL_DCAP_OPEN, "YES");
+#ifdef LIBHEIF_SUPPORTS_TILES
+    poDriver->SetMetadataItem("HEIF_SUPPORTS_TILES", "YES");
```
bradh, line 90: @rouault Should this have been in the "HEIF" metadata domain?

I'm doing a follow-up PR to make sure libheif supports what we are testing (and to provide info for debugging), and noted that you used this in the AVIF vs HEIF check. I can do the same here if preferred.

rouault: We don't have a very strict policy on this. I'd say that you report metadata items under the non-default metadata domain for things that are so technical that they are mostly for internal GDAL use (autotest typically).

PR #10985, opened 2024-10-10
Python bindings: add a osgeo.gdal_fsspec module that on import will register GDAL VSI file system handlers as fsspec AbstractFileSystem

rouault:
This enables using GDAL virtual file systems with other libraries of the Python ecosystem that accept [fsspec](https://filesystem-spec.readthedocs.io/en/latest/) paths

"""Module exposing GDAL Virtual File Systems (VSI) as a "gdalvsi" fsspec implementation.

   Importing "osgeo.gdal_fsspec" requires the Python "fsspec"
   (https://filesystem-spec.readthedocs.io/en/latest/) module to be available.

   A generic "gdalvsi" fsspec protocol is available. All GDAL VSI file names must be
   simply prefixed with "gdalvsi://". For example:

   - "gdalvsi://data/byte.tif" to access relative file "data/byte.tif"
   - "gdalvsi:///home/user/byte.tif" to access absolute file "/home/user/byte.tif"
   - "gdalvsi:///vsimem/byte.tif" (note the 3 slashes) to access VSIMem file "/vsimem/byte.tif"
   - "gdalvsi:///vsicurl/https://example.com/byte.tif (note the 3 slashes) to access "https://example.com/byte.tif" through /vsicurl/
"""

### DIFF
--- a/.github/workflows/cmake_builds.yml
+++ b/.github/workflows/cmake_builds.yml
@@ -432,7 +432,7 @@ jobs:
             cfitsio freexl geotiff libjpeg-turbo libpq libspatialite libwebp-base pcre pcre2 postgresql \
             sqlite tiledb zstd cryptopp cgal doxygen librttopo libkml openssl xz \
             openjdk ant qhull armadillo blas blas-devel libblas libcblas liblapack liblapacke blosc libarchive \
-            arrow-cpp pyarrow libaec libheif libavif cmake
+            arrow-cpp pyarrow libaec libheif libavif cmake fsspec
     - name: Check CMake version
       shell: bash -l {0}
       run: |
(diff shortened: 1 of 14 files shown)


sgillies:
@rouault can we do without so many new forms of identifiers? As a community, I think we want fewer ways to reference the same thing, not more ways to reference the same thing.

Like, I imagine the fsspec usage would be: 

```
fs = fsspec.filesystem("gdalvsi")
fs.open("/vsicurl/https://example.com/foo.tif")
```

The existing `/vsi*/` files could remain the one way to reference datasets for GDAL, if we want.

Is `"vsizip://my.zip/foo.tif"` only for fsspec or will it work with GDALOpen too?

rouault:
> can we do without so many new forms of identifiers?

I also had the same hesitation, and I would be OK with just having a single "gdalvsi" fsspec protocol taking GDAL paths as you suggest

> Is "vsizip://my.zip/foo.tif" only for fsspec or will it work with GDALOpen too?

No that would be only for fsspec, if we keep the "sub-classed" vsiXXX fsspec protocol

rouault:
Pull request simplified to register a single "gdalvsi" fsspec protocol as suggested by @sgillies

PR #10995, opened 2024-10-12
docker: add ccache for upstream builds

pjonsson:
This PR removes the build support for openjpeg since that hasn't been active since the upgrade to Ubuntu 20.04, and adds ccache for the non-OSGeo projects that are built as part of producing the ubuntu-full image.


rouault:
https://github.com/OSGeo/gdal/actions/runs/11305617898/job/31445313025?pr=10995:
```
Found unknown binary file docker/ubuntu-full/tiledb-cmake-ccache.patch of kind "Git commit 16c3aa25b9e9401b91cd70f2118c7a86f12140b4". If it is legit, add the following line in scripts/binaries_allow_list.csv:
"docker/ubuntu-full/tiledb-cmake-ccache.patch",c1ad2776c1599060613fad9ca784ce140ea7ef795b612d55b7472f0e4e053ad2
```

PR #11003, opened 2024-10-13
Add read-only OGR ADBC (Arrow Database Connectivity) driver

rouault:
Cf https://arrow.apache.org/adbc/current/index.html for what ADBC is.

Depends on the adbc-driver-manager library.

The driver is read-only, and there is no support for spatial data currently.

Beyond official ADBC drivers (adbc-driver-sqlite, adbc-driver-postgresql, adbc-driver-snowflake, adbc-driver-bigquery,
etc.), it can also be used to read Parquet or DuckDB datasets using libduckdb, if libduckdb is installed and can be loaded through dynamic shared library opening.


ogr/ogrsf_frmts/adbc/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
...
+  NO_SHARED_SYMBOL_WITH_CORE
+  STRONG_CXX_WFLAGS)
+
+if(NOT TARGET ogr_ADBC)
+    return()
+endif()
+
+gdal_standard_includes(ogr_ADBC)
+target_include_directories(ogr_ADBC PRIVATE $<TARGET_PROPERTY:ogrsf_generic,SOURCE_DIR>)
+gdal_target_link_libraries(ogr_ADBC PRIVATE AdbcDriverManager::adbc_driver_manager_shared)
```
paleolimbot, line 19: This is OK, but I think you can also vendor `c/driver_manager/adbc_driver_manager.cc/h` if that is easier for packaging/makes it more likely to get this into a default build (this is what the R package does since we can't use CMake).

ogr/ogrsf_frmts/adbc/ogradbcdataset.cpp
```diff
@@ -0,0 +1,381 @@
...
+    if (STARTS_WITH(pszFilename, "ADBC:"))
+    {
+        pszFilename += strlen("ADBC:");
+        poTmpOpenInfo =
+            std::make_unique<GDALOpenInfo>(pszFilename, GA_ReadOnly);
+        poTmpOpenInfo->papszOpenOptions = poOpenInfo->papszOpenOptions;
+        poOpenInfo = poTmpOpenInfo.get();
+    }
+    const char *pszADBCDriverName =
+        CSLFetchNameValue(poOpenInfo->papszOpenOptions, "ADBC_DRIVER");
```
paleolimbot, line 175: If at all possible, in R it would help to be able to initialize via the init function address:

https://github.com/apache/arrow-adbc/blob/cd24bc096612feeef31bff2a9271616cc94e1634/c/include/arrow-adbc/adbc_driver_manager.h#L69-L79

...because in R this is how we pass around driver information (including DuckDB and Snowflake). In Python I believe the driver packages we ship always have a shared library + entrypoint (although I would have to look up if this is easily accessible to help something like GDAL load the driver). This can, of course, be handled some other time, but if there's any way to squeeze in supporting a driver installed using `pip install adbc_driver_snowflake` or `install.packages("adbcsnowflake")`, this is how our documentation encourages users of those environments to get the driver.

ogr/ogrsf_frmts/adbc/ogradbcdataset.cpp
```diff
@@ -0,0 +1,381 @@
...
+        if (pszSQL[0])
+        {
+            auto poLayer = CreateLayer(pszSQL, osLayerName.c_str());
+            if (!poLayer)
+                return false;
+            poLayer->m_bIsParquetLayer = bIsParquetLayer;
+            m_apoLayers.emplace_back(std::move(poLayer));
+        }
+    }
+    else if (bIsDuckDB || bIsSQLite3)
```
paleolimbot, line 318: Would `AdbcConnectionGetObjects()` help here?

https://github.com/apache/arrow-adbc/blob/cd24bc096612feeef31bff2a9271616cc94e1634/c/include/arrow-adbc/adbc.h#L1504-L1614

paleolimbot (commented):
This is amazing!

With respect to geometry support, we're in the process of thinking through how to mark Geometry while pushing the minimum geometry awareness into the general-purpose drivers (that's GDAL's job, after all!): https://github.com/apache/arrow-adbc/issues/2098 . I think we will be able to mark Snowflake geometry as some extension type denoting GeoJSON, PostGIS as something denoting EWKB, and BigQuery as `geoarrow.wkt` (with spherical edges). Getting DuckDB to mark their output has been a feature request for a while but I think it's generally thought of as a good idea: https://github.com/duckdb/duckdb_spatial/issues/153 .

I am not sure at what point I'll be able to get there, but I have wanted to do the reverse as well (expose OGR as an ADBC driver). I am in the process of finishing up the documentation for a framework to make it a bit easier: https://github.com/apache/arrow-adbc/pull/2186

docker/ubuntu-full/Dockerfile
```diff
@@ -483,6 +501,31 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
...
+    && dpkg -i libadbc-driver-sqlite102_14-1_amd64.deb \
+    && dpkg -i libadbc-driver-sqlite-dev_14-1_amd64.deb \
+    && dpkg -i libadbc-driver-snowflake102_14-1_amd64.deb \
+    && dpkg -i libadbc-driver-snowflake-dev_14-1_amd64.deb \
+    && rm -f libadbc*.deb; \
+    fi
+
+# Install libduckdb
+RUN if test "$(uname -p)" = "x86_64"; then \
+    wget -q https://github.com/duckdb/duckdb/releases/download/v1.1.1/libduckdb-linux-amd64.zip \
```
pjonsson, line 523: Could the `wget -q` commands be changed to `curl -LO -fsS` so the build fails with an error message on HTTP errors?

ogr/ogrsf_frmts/adbc/ogradbcdataset.cpp
rouault: > If at all possible, in R it would help to be able to initialize via the init function address:

Hum, would you suggest passing a function pointer as an hexadecimal address as a string like "0xDEADBEEF" ? I believe that's technically undefined behaviour in C/C++ for function pointers, also that will work in most cases. However I remember some time ago to have experimenting with the [CHERI architecture](https://www.cl.cam.ac.uk/research/security/ctsrd/cheri/) and pointers are kind of "fat pointers" and can't be "instanciated" from their address. And there's a security issue in allowing to pass this. I could do that, but protected with a configuration option to enable it explicitly.

docker/ubuntu-full/Dockerfile
rouault: done, thanks

ogr/ogrsf_frmts/adbc/CMakeLists.txt
rouault: I'm a bit hesitant on that:
- that would mean extra code to maintain in GDAL
- When we vendor things, we make an effort to rename their symbols, which is an extra layer of maintainance on GDAL side. If adbc_driver_manager code itself has some provision for a #define ADBC_DRIVER_MANAGER_NAMESPACE that could allow to prefix its symbols with a user specified prefix, that would save that effort on GDAL side.
- what about the Windows side of things? Initially, I tried to use Conda to test on Windows on our CI, but it failed, and then I realized that
https://github.com/conda-forge/arrow-adbc-split-feedstock/blob/6f2af5a9b0de9cbdf895ec0ab1e516c306abac67/recipe/meta.yaml#L31 disabled Windows packaging. Is it a Conda thing, or is there's something more fundamental in adbc_driver_manager itself that prevents it from working on Windows?
- if we vendor, what about the stability of the interface between adbc_driver_manager and ADBC drivers themselves? I mean if we forget to update the vendored  adbc_driver_manager, will that affect its working with ADBC drivers that would have been built against a more recent external ADBC interface?
- with respect to my above symbol renaming, do ADBC drivers link with adbc_driver_manager itself? If so, then symbol renaming is not possible, and vendoring would be a bit risky.

ogr/ogrsf_frmts/adbc/ogradbcdataset.cpp
rouault: Interesting. I've used in https://github.com/OSGeo/gdal/pull/11003/commits/637405badd26cf4d7f4312466c9de95a8a168ead 
However I've noticed what I believe is a discrepency between the API documentation and driver implementations regarding the depth. I would expect ADBC_OBJECT_DEPTH_DB_SCHEMAS to be enough to get the table list since its doc mentions "Return metadata on catalogs, schemas, and table", but I actually need to go one level deeper (ADBC_OBJECT_DEPTH_TABLES) to get the table list. At least with SQLite, DuckDB and PostgreSQL. With ADBC_OBJECT_DEPTH_DB_SCHEMAS, db_schema_tables is null or an empty list.

ogr/ogrsf_frmts/adbc/CMakeLists.txt
paleolimbot: The TL;DR is that this is something we need to make possible, test, and document from our side: we want people to federate to data sources using ADBC exactly like you're doing here and there won't be widely available linux packages anytime soon.

Perhaps a safer alternative, if slightly more verbose, would be to use just the driver loader part of the driver manager:

https://github.com/apache/arrow-adbc/blob/cd24bc096612feeef31bff2a9271616cc94e1634/c/include/arrow-adbc/adbc_driver_manager.h#L34-L67

...and call the driver callbacks directly:

https://github.com/apache/arrow-adbc/blob/cd24bc096612feeef31bff2a9271616cc94e1634/c/driver/framework/base_driver_test.cc#L78-L145

One of the things we may do soon is make a header-only C++ wrapper that supports that process.

> that would mean extra code to maintain in GDAL

If it worked using `FetchContent` to an official ADBC release would that be sufficient? I don't think there will be a widely available "system" ADBC for quite some time that will let R users in particular benefit from this (since most of them choose not to use conda!).

> with respect to my above symbol renaming, do ADBC drivers link with adbc_driver_manager itself? If so, then symbol renaming is not possible, and vendoring would be a bit risky.

ADBC drivers definitely do not use the `AdbcXXX` symbols, but some (or all?) of them export them (I think this is confusing and I wish they didn't). Loading a driver using the init function and calling the driver callbacks definitely doesn't depend on the `AdbcXXX` symbols at all.

> what about the Windows side of things?

The driver manager definitely works on Windows (it's bundled into the Python package that builds on MSVC and vendored into the R package that builds using mingw). I'm not sure of the linking details of Python + conda or why it is not packaged for conda windows.

ogr/ogrsf_frmts/adbc/ogradbcdataset.cpp
paleolimbot: That's a great point! I was thinking a base 10 integer address string but an opt-in configuration option won't help R or Rust users here. Would a C or C++ API hook to inject `AdbcLoadDriver` or similar be secure? What I'm hoping for here is the ability for GDAL to use a driver installed via `install.packages()`, `cargo`, or `pip` since the drivers are seeing quite a bit of action at the moment and getting the latest version via system is difficult or impossible to do in some environments. With the ability to inject `AdbcLoadDriver`, something like sf could provide that by hooking up to the `adbcdrivermanager` package (or pyogrio could do it by hooking up to `adbc_driver_manager`).

ogr/ogrsf_frmts/adbc/CMakeLists.txt
rouault: > there won't be widely available linux packages anytime soon.

IMHO, that's something to consider in the top of the priority list for success of ADBC rather than investing in all other workarounds.

> If it worked using `FetchContent` to an official ADBC release would that be sufficient?

not really, at least for GDAL as packaged by Linux distributions. I believe some build systems disable network after they have downloaded the tarball and installed all dependencies to make sure it is fully reproducible

ogr/ogrsf_frmts/adbc/ogradbcdataset.cpp
rouault: > Would a C or C++ API hook to inject `AdbcLoadDriver` or similar be secure?

definitely. The slightly annoying part is that setting the hook should be done in GDAL core (as the ADBC driver may be a plugin, that cannot be directly in it), but that's not much a big deal.

ogr/ogrsf_frmts/adbc/CMakeLists.txt
paleolimbot: Got it! It's not a workaround in the sense that it's documented behaviour:

https://github.com/apache/arrow-adbc/blob/cd24bc096612feeef31bff2a9271616cc94e1634/c/include/arrow-adbc/adbc.h#L930-L935

...it's just documented behaviour that we haven't made as easy to use as the driver manager's `AdbcXX` symbols yet. With the ability to inject `AdbcLoadDriver`,  this would mean you don't necessarily need the driver manager at all (i.e., you would only need `adbc.h`, which is intended to be vendored in this way and defines an ABI with very intentional forward/backward compatibility guarantees).

ogr/ogrsf_frmts/adbc/ogradbcdataset.cpp
```diff
@@ -0,0 +1,512 @@
...
+    {
+        auto poLayerList = CreateLayer(
+            "SELECT name FROM sqlite_master WHERE type IN ('table', 'view')",
```
elpaso, line 323: this branch is also executed when `bIsDuckDB` is it correct to query `sqlite_master` in that case?

ogr/ogrsf_frmts/adbc/ogradbcdataset.cpp
```diff
@@ -0,0 +1,512 @@
...
+            return false;
+        }
+
+        for (const auto &poFeature : poLayerList.get())
+        {
+            const char *pszNamespace = poFeature->GetFieldAsString(0);
+            const char *pszTableName = poFeature->GetFieldAsString(1);
+            const std::string osStatement = CPLSPrintf(
+                "SELECT * FROM \"%s\".\"%s\"",
+                CPLString(pszNamespace).replaceAll("\"", "\"\"").c_str(),
```
elpaso, line 367: If we have an SQL quoting function available somewhere this is when it could be of use.

ogr/ogrsf_frmts/generic/ogrlayerarrow.h
```diff
@@ -34,4 +36,73 @@ bool CPL_DLL OGRCloneArrowArray(const struct ArrowSchema *schema,
...
+    }
+
+    /** Move assignment operator */
+    inline OGRArrowArrayStream &operator=(OGRArrowArrayStream &&other)
+    {
+        if (this != &other)
+        {
+            clear();
+            memcpy(&m_stream, &(other.m_stream), sizeof(m_stream));
+            memset(&(other.m_stream), 0, sizeof(m_stream));
```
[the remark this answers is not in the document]
rouault: That would result in a double-free. The way to invalidate a ArrowArray/ArrowSchema/ArrowArrayStream is to invalidate its release member, but zero'ing all stuff is even clearer. Cf https://arrow.apache.org/docs/format/CDataInterface.html#moving-an-array

ogr/ogrsf_frmts/adbc/ogr_adbc.h
```diff
@@ -0,0 +1,210 @@
...
+    std::vector<std::unique_ptr<OGRFeature>> m_apoFeatures{};
+
+    CPL_DISALLOW_COPY_ASSIGN(OGRArrowArrayToOGRFeatureAdapterLayer)
+
+  public:
+    explicit OGRArrowArrayToOGRFeatureAdapterLayer(const char *pszName)
+    {
+        m_poLayerDefn = new OGRFeatureDefn(pszName);
+        m_poLayerDefn->SetGeomType(wkbNone);
+        m_poLayerDefn->Reference();
```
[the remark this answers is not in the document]
rouault: Something that we discussed with @dbaston for the scope of his future works, but besides the enormous effort to change that within OSGeo/GDAL, there's an obvious backward compatiiblity issue for external C++ users. Could very well be a 4.0 topic

ogr/ogrsf_frmts/adbc/ogradbcdataset.cpp
```diff
@@ -0,0 +1,512 @@
...
+        }
+    }
+
+    if (pszSQL)
+    {
+        if (pszSQL[0])
+        {
+            auto poLayer = CreateLayer(pszSQL, osLayerName.c_str());
+            if (!poLayer)
+                return false;
```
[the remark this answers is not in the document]
rouault: CreateLayer() should have emitted one

ogr/ogrsf_frmts/adbc/ogradbcdataset.cpp
rouault: yes, this is fine. DuckDB immitates SQLite: https://duckdb.org/docs/guides/meta/list_tables.html#see-also

ogr/ogrsf_frmts/generic/ogrlayerarrow.h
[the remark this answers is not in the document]
rouault: code comment added

autotest/ogr/ogr_adbc.py
```diff
@@ -0,0 +1,337 @@
...
+
+    assert "INFO" in ret
+    assert "ERROR" not in ret
+
+
+###############################################################################
+# Run test_ogrsf on a partitioned Parquet dataset
+
+
+def test_ogr_adbc_test_ogrsf_parquet_star():
```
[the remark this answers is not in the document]
rouault: test function renamed

doc/source/drivers/vector/adbc.rst
```diff
@@ -0,0 +1,111 @@
...
+
+Connection string
+-----------------
+
+Several connection strings are supported:
+
+- ``ADBC:{some_uri}``, together with the ``ADBC_DRIVER`` open option.
+- a SQLite3 database filename, if the ``adbc_driver_sqlite`` is available.
+- a DuckDB database filename, if the :file:`libduckdb.so`, :file:`libduckdb.dylib`
+  or :file:`duckdb.dll` is available (and is in a system location, or can be
```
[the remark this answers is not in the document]
rouault: fixed

doc/source/drivers/vector/adbc.rst
```diff
@@ -0,0 +1,111 @@
...
+  located through LD_LIBRARY_PATH on Linux, DYLD_LIBRARY_PATH on MacOSX or PATH on Windows).
+- a Parquet database filename, if the :file:`libduckdb.so`, :file:`libduckdb.dylib`
+  or :file:`duckdb.dll` is available (and is in a system location, or can be
```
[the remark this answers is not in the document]
rouault: fixed

doc/source/drivers/vector/adbc.rst
```diff
@@ -0,0 +1,111 @@
...
+:program:`ogrinfo`).
+It returns for each table a OGR feature with the following fields (some
+potentially unset or with an empty string): ``catalog_name``, ``schema_name``,
+``table_name``, ``table_type``.
+
+Examples
+--------
+
+- Assuming :file:`libduckdb.so`, :file:`libduckdb.dylib` or :file:`duckdb.dll`
+  is available (and is in a system location, or can be located through
```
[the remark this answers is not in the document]
rouault: fixed

doc/source/drivers/vector/adbc.rst
```diff
@@ -0,0 +1,111 @@
...
+
+  Convert a Parquet file to GeoPackage:
+
+  ::
+
+      ogr2ogr out.gpkg in.parquet
+
+
+- Assuming :file:`libduckdb.so`, :file:`libduckdb.dylib` or :file:`duckdb.dll`
+  is available (and is in a system location, or can be located through
```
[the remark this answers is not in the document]
rouault: fixed

ogr/ogrsf_frmts/adbc/ogradbcdataset.cpp
rouault: added :
```
/** Replace all occurences of ch by it repeated twice.
 * Typically used for SQL string literal or identifier escaping.
 */
std::string CPL_DLL OGRDuplicateCharacter(const std::string &osStr, char ch);
```

ogr/ogrsf_frmts/adbc/ogradbcdataset.cpp
rouault: I've added a new public header gdal_adbc.h per 8d86043696960632c662345c3dff7066367eb651 that can be used to call a ``void GDALSetAdbcDriverInitFunc(AdbcDriverInitFunc init_func)`` to define the entry point.
This is only compile tested on my side. Actual functional testing would be appreciated

paleolimbot: I will take a look today/this evening and report back!

paleolimbot: No pressure on any of this (all ADBC is great!), but I put together an example of what I had in mind. This turned out to be more like "inject `AdbcLoadDriver`" than just one init function. Basically, if the embedding application can have some flexibility about how to load the drivers it can use ones already statically linked in, or updated versions of them installed via `pip` or some other mechanism. Using the "load" mechanism frees GDAL from the driver manager completely since then it can just use the callbacks.

Experiment in GDAL: https://github.com/rouault/gdal/compare/adbc...paleolimbot:gdal:adbc-suggestions

Changes to sf: https://github.com/r-spatial/sf/compare/main...paleolimbot:adbc-experiments

Changes to the R driver manager: https://github.com/apache/arrow-adbc/compare/main...paleolimbot:r-adbcdrivermanager-init

rouault: @paleolimbot Cool! I've merged your changes with minor adaptations

doc/source/drivers/vector/adbc.rst
```diff
@@ -0,0 +1,129 @@
...
+
+Custom driver entry point
+-------------------------
+
+A custom driver entry point can be specified by applications by calling
+:cpp:func:`GDALSetAdbcLoadDriverOverride` (defined in header :file:`gdal_adbc.h`)
+before using the driver. The specified init function will be used by the
+GDAL ADBC driver as a way of specifying the ADBC driver for all calls in the
+same thread. Setting both the init function and the ``ADBC_DRIVER`` open
+option is not supported.
```
paleolimbot, line 99:
```suggestion
GDAL ADBC driver as a way of locating and loading the ADBC driver if GDAL was
not built with ADBC Driver Manager support or if an embedding application has
an updated or augmented collection of drivers available.
```

ogr/ogrsf_frmts/adbc/ogradbcdataset.cpp
paleolimbot: Thank you!

doc/source/drivers/vector/adbc.rst
rouault: done

rouault:
Merging

PR #11005, opened 2024-10-14
netCDF CreateCopy(): fix taking into account NETCDF_DIM_EXTRA when source dataset is not georeferenced

rouault:
Fixes https://lists.osgeo.org/pipermail/gdal-dev/2024-October/059618.html


mdsumner:
confirmed fix

PR #11013, opened 2024-10-15
[3.9] docker/ubuntu-full/Dockerfile: revert to using wget for https://github.com/ stuff

rouault:
I've observed while building 3.9.3 image systematic failures when using curl on https://github.com/ content, when run under the arm64 architecture under docker buildx (with qemu)

Cf:
```
* Host github.com:443 was resolved.
* IPv6: (none)
* IPv4: 140.82.121.3
*   Trying 140.82.121.3:443...
* Connected to github.com (140.82.121.3) port 443
* ALPN: curl offers h2,http/1.1
} [5 bytes data]
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
} [512 bytes data]
*  CAfile: /etc/ssl/certs/ca-certificates.crt
*  CApath: /etc/ssl/certs
{ [5 bytes data]
* TLSv1.3 (IN), TLS handshake, Server hello (2):
{ [122 bytes data]
* TLSv1.3 (IN), TLS handshake, Encrypted Extensions (8):
{ [19 bytes data]
* TLSv1.3 (IN), TLS handshake, Certificate (11):
{ [3137 bytes data]
* TLSv1.3 (IN), TLS handshake, CERT verify (15):
{ [79 bytes data]
* TLSv1.3 (IN), TLS handshake, Finished (20):
{ [36 bytes data]
* TLSv1.3 (OUT), TLS change cipher, Change cipher spec (1):
} [1 bytes data]
* TLSv1.3 (OUT), TLS handshake, Finished (20):
} [36 bytes data]
* SSL connection using TLSv1.3 / TLS_AES_128_GCM_SHA256 / X25519 / id-ecPublicKey
* ALPN: server accepted h2
* Server certificate:
*  subject: CN=github.com
*  start date: Mar  7 00:00:00 2024 GMT
*  expire date: Mar  7 23:59:59 2025 GMT
*  subjectAltName: host "github.com" matched cert's "github.com"
*  issuer: C=GB; ST=Greater Manchester; L=Salford; O=Sectigo Limited; CN=Sectigo ECC Domain Validation Secure Server CA
*  SSL certificate verify ok.
*   Certificate level 0: Public key type EC/prime256v1 (256/128 Bits/secBits), signed using ecdsa-with-SHA256
*   Certificate level 1: Public key type EC/prime256v1 (256/128 Bits/secBits), signed using ecdsa-with-SHA384
*   Certificate level 2: Public key type EC/secp384r1 (384/192 Bits/secBits), signed using ecdsa-with-SHA384
} [5 bytes data]
* Send failure: Broken pipe
* OpenSSL SSL_write: Broken pipe, errno 32
* Closing connection
curl: (55) Send failure: Broken pipe
```

### DIFF
--- a/docker/ubuntu-full/Dockerfile
+++ b/docker/ubuntu-full/Dockerfile
@@ -75,7 +75,7 @@ RUN . /buildscripts/bh-set-envvars.sh \
 # Build likbkea
 ARG KEA_VERSION=1.5.2
 RUN . /buildscripts/bh-set-envvars.sh \
-    && curl -LO -fsS https://github.com/ubarsc/kealib/archive/kealib-${KEA_VERSION}.zip \
+    && wget -q https://github.com/ubarsc/kealib/archive/kealib-${KEA_VERSION}.zip \
     && unzip -q kealib-${KEA_VERSION}.zip \
     && rm -f kealib-${KEA_VERSION}.zip \
     && cd kealib-kealib-${KEA_VERSION} \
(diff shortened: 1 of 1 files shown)


pjonsson:
Is this an intermittent error, or something reproducible?

I did a quick search, and it seems qemu has failed on curl for others as well:

https://gitlab.com/qemu-project/qemu/-/issues/1471
https://unix.stackexchange.com/questions/400224/curl-fails-on-tls-connection-sslv3-alert-bad-record-mac

so your revert is probably the easiest way forward.

pjonsson:
It's not the same error message, but here's another more recent curl issue with Ubuntu 24.04 on ARM64 running on qemu connecting to Microsoft's servers:

https://bugs.launchpad.net/ubuntu/+source/curl/+bug/2073448

If this is a different manifestation of that bug, I guess there isn't much we can do.

rouault:
> Is this an intermittent error, or something reproducible?

100% reproducible

PR #11019, opened 2024-10-15
Zarr V2 creation: fix bug when creating dataset with partial blocks …

### LINKED ISSUE #11016: ZARR:  ZARR_V2 created from PNG compression problem with (relatively) small blocks
### What is the bug?

When creating a ZARR from PNG 20x20x4 (RGBA) and I use compression for ZARR_V2 with small block sizes, I get an error. 

It seems to be about the relative size of the blocks compared to the source, but I didn't detect the pattern. 

```
gdal_translate gdrivers/data/png/rgba16.png lz4.zarr -of ZARR -co FORMAT=ZARR_V2 -co BLOCKSIZE=4,5,6 -co COMPRESS=LZ4  -co INTERLEAVE=BAND

Input file size is 20, 20
0...10...20...30...40...50ERROR 1: Decompressed tile lz4.zarr/lz4/0.1.2 has not expected size after filters
```

It's fine with COMPRESS=NONE, but other compression-on options error also. 

With larger block sizes it's seemingly ok

```
gdal_translate gdrivers/data/png/rgba16.png lz4.zarr -of ZARR -co FORMAT=ZARR_V2 -co BLOCKSIZE=4,25,26 -co COMPRESS=LZ4  -co INTERLEAVE=BAND

Input file size is 20, 20
```

### Versions and provenance

GDAL 3.10.0dev-e3aaf5d09b-dirty, released 2024/10/15

Linux, ubuntu 22.04

### Additional context

Originally reported here with example below: https://lists.osgeo.org/pipermail/gdal-dev/2024-October/059625.html

```
gdal_translate "/vsicurl/https://github.com/cran/png/blob/master/inst/img/Rlogo.png?raw=true" lz4.zarr -of ZARR -co FORMAT=ZARR_V2 -co BLOCKSIZE=4,25,26 -co COMPRESS=LZ4 -co INTERLEAVE=BAND

Input file size is 100, 76
0ERROR 1: Decompressed tile lz4.zarr/lz4/0.0.0 has not expected size after
filters
```

rouault:
…and need to re-read them in the writing process when compression is involved

Embarassingly, we were trying to decompress chunks with the ... compressor ... instead of the decompressor.

Fixes #11016

### DIFF
--- a/autotest/gdrivers/zarr_driver.py
+++ b/autotest/gdrivers/zarr_driver.py
@@ -5523,3 +5523,21 @@ def test_zarr_read_cf1_zarrv3():
         ds.GetSpatialRef().ExportToProj4()
         == "+proj=utm +zone=11 +ellps=clrk66 +units=m +no_defs"
     )
+
+
+###############################################################################
+# Test bug fix for https://github.com/OSGeo/gdal/issues/11016
+
+
+@gdaltest.enable_exceptions()
+def test_zarr_write_partial_blocks_compressed(tmp_vsimem):
+
+    out_filename = "/vsimem/test.zarr"
+    src_ds = gdal.Open("data/small_world.tif")
+    gdal.Translate(
+        out_filename,
+        src_ds,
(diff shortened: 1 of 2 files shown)


mdsumner:
Looks good, thanks Even!

PR #11033, opened 2024-10-16
CSharp bindings: Add VSIGetMemFileBuffer bindings

### LINKED ISSUE #11032: Add VSIGetMemFileBuffer to C# / .NET bindings
### Feature description

Currently there are no way to read out the in-memory bytes of an in memory dataset with the .NET bindings, as far as i know. It would be nice if the VSIGetMemFileBuffer was added to the .NET bindings (and i guess the other ones as well). Currently we are working around this by using P/Invoke directly on the gdal library file, but this of course means we are missing out on exception handling etc.

### Additional context

_No response_

EmilRybergAkson:
Fixes #11032

This is the first time I have dealt with configuration SWIG, and I'm not a C++ expert, so there might be a better way of generating the bindings. I tried to take inspiration of how the other SWIG interface code was written.

It is intentional that the `vsi_l_offset` is mapped to "int" in C#, since the use case will in most cases probably be:

```csharp
IntPtr ptr = Gdal.GetMemFileBuffer(vsiPath, out int length, false);
byte[] buffer = new byte[length];
Marshal.Copy(ptr, buffer, 0, length);
```

And Marshal.Copy only supports int for buffer length. It seems like unsigned long and unsigned long long pointers are handled similarly in other parts of the binding code (casted to signed int), so hopefully this is OK.

### DIFF
--- a/swig/include/csharp/gdal_csharp.i
+++ b/swig/include/csharp/gdal_csharp.i
@@ -381,17 +381,17 @@ public CPLErr SetGCPs(GCP[] pGCPs, string pszGCPProjection) {
 
 %rename (GetMemFileBuffer) wrapper_VSIGetMemFileBuffer;
 
-%typemap(cstype) (GIntBig *pnDataLength) "out int";
-%typemap(imtype) (GIntBig *pnDataLength) "out int";
-%apply (int *OUTPUT) {(GIntBig *pnDataLength)}
+%typemap(cstype) (vsi_l_offset *pnDataLength) "out ulong";
+%typemap(imtype) (vsi_l_offset *pnDataLength) "out ulong";
+%apply (unsigned long long *OUTPUT) {(vsi_l_offset *pnDataLength)}
 %typemap(cstype) (int bUnlinkAndSeize) "bool";
 %typemap(csin) (int bUnlinkAndSeize) "$csinput ? 1 : 0";
 
 %inline {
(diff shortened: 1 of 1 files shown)


rouault:
> It is intentional that the `vsi_l_offset` is mapped to "int" in C#

probably not... ? Seeing https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/integral-numeric-types, it looks like it should be "long" for a 64 bit integer, actually  "ulong" since it is  a unsigned one

EmilRybergAkson:
> > It is intentional that the `vsi_l_offset` is mapped to "int" in C#
> 
> probably not... ? Seeing https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/integral-numeric-types, it looks like it should be "long" for a 64 bit integer, actually "ulong" since it is a unsigned one

Yes. I just meant that it was intentional that it was cast to int on the C# side since the Marshal.Copy does not have ulong overload for buffer length, so it was to keep it "user friendly" on the C# side. I noticed that in `cpl.i` `vsi_l_offset` is changed to `GIntBig`, so I thought it might have been OK to cast to signed int. But I guess to keep it as close as possible to the C/C++ API that it should be ulong!

rouault:
@szekerest This looks plausible to me, but would appreciate your opinion

PR #11048, opened 2024-10-19
PDF: remove write support for GEO_ENCODING=OGC_BP

rouault:
for 3.11

Several reasons:
- support for SRS has always been partial (contrary to the default ISO32000 encoding)
- and the actual trigger for this pull request: lately the company behind the OGC Best Practice encoding has behaved in a unfriendly way with respect to our community. No need to promote their techs further.

CC @nyalldawson  Will that be an issue for QGIS?


nyalldawson:
> Will that be an issue for QGIS?

Well, we'll need to remove that export options at least. I'm unsure how widely used it is. 

@andreasneumann do you know if this option is required for compliance? Perhaps you could check with the Swiss user group?

rouault:
> I'm unsure how widely used it is.

I had some memories that PDF created with the OGC_BP couldn't be properly read with the T....Go toolbar due to it requiring them to be signed (it could open them, but wouldn't take into account the georeferenced info). But I could be wrong.

rouault:
gentle ping @andreasneumann

andreasneumann:
@rouault - sorry for not being responsive. I don't have a clue what exactly this would mean? Do you have more information on it?

I assume that Geospatial PDF (layers, feature info, coordinate display) would still work, but that compatibility with the TxxxGx toolbar wouldn't work anymore? Is this assumption correct?

I asked people in the QGIS-CH steering committee - but they did not know either.

rouault:
> I don't have a clue what exactly this would mean? Do you have more information on it?

in QGIS, you can choose between the ISO32000 encoding (the default choice) and the OGC BP one. This pull request is about removing the later one. I assume most users let the default choice

>  but that compatibility with the TxxxGx toolbar wouldn't work anymore?

I've no idea what TxxxGx does those days. I assume they support ISO32000 geospatial PDFs, in addition to their own encoding (my memories were that many years ago writing with their encoding no longer worked properly in their toolbar since they required those PDFs to be signed with their signing key, or at least that was my impression)

andreasneumann:
> in QGIS, you can choose between the ISO32000 encoding (the default choice) and the OGC BP one. This pull request is about removing the later one. I assume most users let the default choice

I agree with you. Probably users just leave the default ISO32000 encoding.

Please go ahead and remove that OGC_BP support.

Issue #11042, opened 2024-10-17
gdalwarp witg lanczos creates aliased result


zdila:
### What is the bug?

Gdalwarp with `-r lanczos` can cause aliased edges.

lanczos: 
![image](https://github.com/user-attachments/assets/2026a9ba-dd53-4ab6-937d-6bd23df9791f)

cubic:
![image](https://github.com/user-attachments/assets/b402d00f-8ab1-4520-9270-664b1b8f65d2)

I have three input vrt files but I can reproduce it only on single one.

### Steps to reproduce the issue

```
gdalwarp -s_srs 'EPSG:8353' -t_srs 'EPSG:3857' -ct '+proj=pipeline +step +inv +proj=krovak +lat_0=49.5 +lon_0=24.8333333333333 +alpha=30.2881397527778 +k=0.9999 +x_0=0 +y_0=0 +ellps=bessel +step +inv +proj=hgridshift +grids=Slovakia_JTSK03_to_JTSK.gsb +step +proj=krovak +lat_0=49.5 +lon_0=24.8333333333333 +alpha=30.2881397527778 +k=0.9999 +x_0=0 +y_0=0 +ellps=bessel +step +inv +proj=krovak +lat_0=49.5 +lon_0=24.8333333333333 +alpha=30.2881397527778 +k=0.9999 +x_0=0 +y_0=0 +ellps=bessel +step +proj=push +v_3 +step +proj=cart +ellps=bessel +step +proj=helmert +x=485.021 +y=169.465 +z=483.839 +rx=-7.786342 +ry=-4.397554 +rz=-4.102655 +s=0 +convention=coordinate_frame +step +inv +proj=cart +ellps=WGS84 +step +proj=pop +v_3 +step +proj=webmerc +lat_0=0 +lon_0=0 +x_0=0 +y_0=0 +ellps=WGS84' -tr 0.29858214173896972949 0.29858214173896972949 -tap -r lanczos -of GTiff -co TILED=YES -co BIGTIFF=YES -co COMPRESS=WEBP -co WEBP_LEVEL=85 -co NUM_THREADS=ALL_CPUS -wo NUM_THREADS=ALL_CPUS -multi -te 2118692.6 6294091.5 2120157.9 6295276.7 Ortofoto_2021_stred_jtsk_rgb/all.vrt stred-extract-warped.tif
```

### Versions and provenance

Tried with 2 versions:

1. Debian linux testing: GDAL 3.9.2, released 2024/08/13
[...]

PR #11051, opened 2024-10-19
Warper: improve performance of multi-band bicubic warping

rouault:
On extracted2.tif dataset from
https://github.com/OSGeo/gdal/issues/11042#issuecomment-2420556574 (12500 x 10000 pixels, 3 bands, Byte) ``gdalwarp -s_srs EPSG:8353 -t_srs EPSG:3857 -r cubic extracted2.tif tmp.tif -overwrite`` goes from 30.4 seconds to 26.0.

and  improve performance of Byte/UInt16 multi-band bicubic warping with XSCALE=1 and YSCALE=1 on SSE2 / x86_64

``gdalwarp -s_srs EPSG:8353 -t_srs EPSG:3857 -r cubic extracted2.tif tmp.tif -overwrite -wo XSCALE=1 -wo YSCALE=1`` goes from 19.0 seconds to 15.7.

### DIFF
--- a/alg/gdalwarpkernel.cpp
+++ b/alg/gdalwarpkernel.cpp
@@ -2849,11 +2849,14 @@ static bool GWKBilinearResampleNoMasks4SampleT(const GDALWarpKernel *poWK,
 // or http://en.wikipedia.org/wiki/Cubic_Hermite_spline : CINTx(p_1,p0,p1,p2)
 // http://en.wikipedia.org/wiki/Bicubic_interpolation: matrix notation
 
-// TODO(schwehr): Use an inline function.
-#define CubicConvolution(distance1, distance2, distance3, f0, f1, f2, f3)      \
-    (f1 + 0.5 * (distance1 * (f2 - f0) +                                       \
-                 distance2 * (2.0 * f0 - 5.0 * f1 + 4.0 * f2 - f3) +           \
-                 distance3 * (3.0 * (f1 - f2) + f3 - f0)))
+template <typename T>
+static inline T CubicConvolution(T distance1, T distance2, T distance3, T f0,
+                                 T f1, T f2, T f3)
+{
+    return (f1 + T(0.5) * (distance1 * (f2 - f0) +
+                           distance2 * (2 * f0 - 5 * f1 + 4 * f2 - f3) +
+                           distance3 * (3 * (f1 - f2) + f3 - f0)));
+}
 
 /************************************************************************/
 /*                       GWKCubicComputeWeights()                       */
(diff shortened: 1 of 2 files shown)


alg/gdalwarpkernel.cpp
```diff
@@ -5827,6 +5854,142 @@ static CPLErr GWKRealCase(GDALWarpKernel *poWK)
...
+#if defined(__x86_64) || defined(_M_X64)
+
+static inline float Convolute4x4(const __m128 row0, const __m128 row1,
+                                 const __m128 row2, const __m128 row3,
+                                 const __m128 weightsXY0,
+                                 const __m128 weightsXY1,
+                                 const __m128 weightsXY2,
+                                 const __m128 weightsXY3)
+{
+    return XMMHorizontalAdd(
```
elpaso, line 5872: just wondering if you could gain even more speed exploiting pipelining by splitting the addition in two separate (and indipendent) instructions followed by a final addition. Not sure if that will actually be faster though.

rouault: > by splitting the addition in two separate (and indipendent) instructions followed by a final addition

not sure what you mean exactly. I'm afraid I'd need to see the exact code you're thinking to. Pipelining is probably achieved by the fact that when it is used to process 2 rows at a time, there are 2 invokations of that inline function in the same for() loop and probably the CPU will pipeline if possible.

I spent a significant time coming to the current code and while its perhaps not the optimum (the optimum could be very well micro-architecture dependent because the latency and throughput of SSE instructions evolve over them), it is a significant improvement over current situation

elpaso: something like (untested):
```c++
const auto mul0 = _mm_mul_ps(row0, weightsXY0);
const auto mul1 = _mm_mul_ps(row1, weightsXY1);
const auto mul2 = _mm_mul_ps(row2, weightsXY2);
const auto mul3 = _mm_mul_ps(row3, weightsXY3);
const auto add1 = _mm_add_ps(mul0, mul1);
const auto add2 = _mm_add_ps(mul2, mul3);
return _mm_add_ps(add1, add2);
```

rouault: Well, that's what the method does currently (except the order of additions, but that doesn't really matter), but there's a lacking key component which is the final XMMHorizontalAdd() that adds together the 4 float of the SSE registry to return a single float scalar value. Basically this functions starts with 16 input pixel values, spread over 4 SSE registers, multiply them by 16 weighting values spread over the 4 weights SSE registers, and computes a single output value

rouault: ok, I now see your point. Indeed that could be slightly better. I've ust tried it, and the difference in timing falls into the measurement noise, but I've adopted your suggestion (with final horizontal add) as it could indeed result in better pipeling. Thanks for raising that.

PR #11065, opened 2024-10-21
gdalbuildvrt: add a -co switch

### LINKED ISSUE #11062: Missing `-co` option on `gdalbuildvrt`
### What is the bug?

There's no `-co`, so no way to set the block size options added in GDAL 3.7.

`gdalbuildvrt --help`

### Versions and provenance

3.10.0

### Additional context

_No response_

rouault:
Fixes #11062

### DIFF
--- a/swig/include/python/gdal_python.i
+++ b/swig/include/python/gdal_python.i
@@ -2485,6 +2485,17 @@ mapGRIORAMethodToString = {
     gdalconst.GRIORA_Gauss: 'gauss',
 }
 
+def _addCreationOptions(new_options, creationOptions):
+    """Update new_options with creationOptions formatted as expected by utilities"""
+    if isinstance(creationOptions, str):
+        new_options += ['-co', creationOptions]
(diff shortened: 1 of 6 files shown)


swig/include/python/gdal_python.i
```diff
@@ -2485,6 +2485,17 @@ mapGRIORAMethodToString = {
     gdalconst.GRIORA_Gauss: 'gauss',
 }
 
+def _addCreationOptions(new_options, creationOptions):
```
dbaston, line 2488: This is both modifying `new_options` and returning it, which could be unexpected.

rouault: good point, I was following a bit too much "papszFoo = CSLAddString(papszFoo, pszBar)"
Updated to no longer return a value and just update in place new_options:
```
def _addCreationOptions(new_options, creationOptions):
    """Update new_options with creationOptions formatted as expected by utilities"""
```

lnicola (approved):

Issue #10889, opened 2024-09-27
argument parsing: throw error when non-numeric values provided to numeric arguments


dbaston:
### Feature description

In a handful of cases (e.g., `gdalwarp -wm`) the argument parsing code uses `CPLAtofM` to parse numeric arguments. This function returns a value of zero when given a non-numeric argument, causing invalid arguments to be silently replaced with zero.

rouault:
> This function returns a value of zero when given a non-numeric argument, causing invalid arguments to be silently replaced with zero.

like atof()... ?

PR #11066, opened 2024-10-21
Remove some unchecked float parsing in command line utils

dbaston:
# What does this PR do?

Addresses some cases where invalid values to command line arguments would be silently accepted and converted to zero or otherwise misinterpreted.

 #10889

## Tasklist

 - [x] Add test case(s)
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/apps/gdal_translate_lib.cpp
+++ b/apps/gdal_translate_lib.cpp
@@ -3276,6 +3276,12 @@ GDALTranslateOptionsNew(char **papszArgv,
             psOptions->asScaleParams[nIndex].bHaveScaleSrc = false;
             if (i < argc - 2 && ArgIsNumeric(papszArgv[i + 1]))
             {
+                if (!ArgIsNumeric(papszArgv[i + 2]))
+                {
+                    CPLError(CE_Failure, CPLE_IllegalArg,
+                             "Value of -scale must be numeric");
+                    return nullptr;
+                }
                 psOptions->asScaleParams[nIndex].bHaveScaleSrc = true;
                 psOptions->asScaleParams[nIndex].dfScaleSrcMin =
                     CPLAtofM(papszArgv[i + 1]);
@@ -3287,6 +3293,12 @@ GDALTranslateOptionsNew(char **papszArgv,
                 psOptions->asScaleParams[nIndex].bHaveScaleSrc &&
                 ArgIsNumeric(papszArgv[i + 1]))
             {
+                if (!ArgIsNumeric(papszArgv[i + 2]))
+                {
+                    CPLError(CE_Failure, CPLE_IllegalArg,
+                             "Value of -scale must be numeric");
+                    return nullptr;
+                }
                 psOptions->asScaleParams[nIndex].dfScaleDstMin =
                     CPLAtofM(papszArgv[i + 1]);
                 psOptions->asScaleParams[nIndex].dfScaleDstMax =
--- a/apps/ogr2ogr_lib.cpp
+++ b/apps/ogr2ogr_lib.cpp
@@ -7522,22 +7522,16 @@ static std::unique_ptr<GDALArgumentParser> GDALVectorTranslateOptionsGetParser(
 
     argParser->add_argument("-segmentize")
         .metavar("<max_dist>")
-        .action(
-            [psOptions](const std::string &s)
-            {
-                psOptions->eGeomOp = GEOMOP_SEGMENTIZE;
-                psOptions->dfGeomOpParam = CPLAtofM(s.c_str());
-            })
+        .store_into(psOptions->dfGeomOpParam)
+        .action([psOptions](const std::string &)
+                { psOptions->eGeomOp = GEOMOP_SEGMENTIZE; })
         .help(_("Maximum distance between 2 nodes."));
 
     argParser->add_argument("-simplify")
         .metavar("<tolerance>")
-        .action(
-            [psOptions](const std::string &s)
-            {
-                psOptions->eGeomOp = GEOMOP_SIMPLIFY_PRESERVE_TOPOLOGY;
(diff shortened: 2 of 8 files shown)


apps/gdal_translate_lib.cpp
```diff
@@ -3415,15 +3427,55 @@ GDALTranslateOptionsNew(char **papszArgv,
...
+                    throw std::invalid_argument(
+                        "Failed to parse first value of -outsize: " +
+                        sOutSizeX);
+                }
+            }
             else
-                psOptions->nOXSizePixel = atoi((*aosOutSize)[0].c_str());
+            {
+                size_t pos;
+                psOptions->nOXSizePixel = std::stoi(sOutSizeX, &pos);
```
rouault, line 3447: std::stoi itself can throw an exception, so there are maybe situations where the below informative exception couldn't be reached and a raw one thrown
Same remark for nOYSizePixel

dbaston: How would you feel about the following?

```
            if (sOutSizeX.back() == '%')
            {
                psOptions->dfOXSizePct = parse_or_throw<double>(
                    sOutSizeX, "Failed to parse first value of -outsize");
            }
            else
            {
                psOptions->nOXSizePixel = parse_or_throw<int>(
                    sOutSizeX, "Failed to parse first value of -outsize");
            }

            if (sOutSizeY.back() == '%')
            {
                psOptions->dfOYSizePct = parse_or_throw<double>(
                    sOutSizeY, "Failed to parse second value of -outsize");
            }
            else
            {
                psOptions->nOYSizePixel = parse_or_throw<int>(
                    sOutSizeY, "Failed to parse second value of -outsize");
            }
```

with

```
/**
* Read a single value of a given type from a string_view.
*
* An invalid_value exception will be thrown if the value:
* - cannot be read
* - does not occupy the entire string_view (i.e, if there are trailing non-numeric characters)
* - overflows the desired type
*
* The exception text will be constructed from the provided message base and the unparseable value.
*
* @param s text from which a value should be read
* @param message_base Text used to construct an exception message in case of failure
* @return a value of the desired type
*/
template <typename T>
T parse_or_throw(std::string_view s, std::string_view message_base)
{
    T value;

    if constexpr (std::is_integral_v<T>)
    {
        auto [ptr, ec] = std::from_chars(s.begin(), s.end(), value);

        if (ec == std::errc() && ptr == s.end())
        {
            // no error
            return value;
        }
    }
    else
    {
        // Unfortunately we need to make a copy to to interface with CPLStrtodM
        // because we cannot guarantee that the string_view is null-terminated.
        std::string s_str(s);

        char *end;
        double parsed = CPLStrtodM(s_str.c_str(), &end);
        if (*end == '\0')
        {
            // no error
            if constexpr (std::is_same_v<T, double>)
            {
                return parsed;
            }

            if (parsed >= std::numeric_limits<T>::lowest() &&
                parsed <= std::numeric_limits<T>::max())
            {
                return static_cast<T>(parsed);
            }
        }
    }

    std::ostringstream ss;
    ss << message_base << ": " << s;
    throw std::invalid_argument(ss.str());
}
```

rouault: LGTM

autotest/utilities/test_ogr2ogr.py
```diff
@@ -290,18 +290,55 @@ def test_ogr2ogr_13(ogr2ogr_path, tmp_path):
...
     assert ds is not None and ds.GetLayer(0).GetFeatureCount() == 10
     feat = ds.GetLayer(0).GetNextFeature()
     assert feat.GetGeometryRef().GetGeometryRef(0).GetPointCount() == 36
 
 
+###############################################################################
+# Test -simplify
+
+
+@pytest.mark.require_geos
```
rouault, line 318: For performance reasons, I would suggest to avoid adding more tests in test_foo.py when it is possible to have them in test_foo_lib.py  (forking a process is way slower than in-process execution of the utlity), which should be possible here

dbaston: Makes sense, I should have looked deeper when I didn't find a test for `-simplify` in this file already.

apps/ogr2ogr_lib.cpp
```diff
@@ -7526,7 +7526,8 @@ static std::unique_ptr<GDALArgumentParser> GDALVectorTranslateOptionsGetParser(
             [psOptions](const std::string &s)
             {
                 psOptions->eGeomOp = GEOMOP_SEGMENTIZE;
-                psOptions->dfGeomOpParam = CPLAtofM(s.c_str());
```
dbaston, line 7529: It looks like this particular case could have been fixed (more idiomatically?) by adding `.scan<'g', double>()` to the call chain.

What's odd is that (as far as I can see), `.action` still requires you to repeat the string-to-number conversion yourself.

rouault: > What's odd is that (as far as I can see), `.action` still requires you to repeat the string-to-number conversion yourself.

you can use ``.store_into(reference_to_double_variable)`` as well

dbaston: I missed that. So the original code can simply be replaced by the following and will error if `-simplify` cannot be parsed to a double.

```
    argParser->add_argument("-simplify")
        .metavar("<tolerance>")
        .store_into(psOptions->dfGeomOpParam)
        .action(
            [psOptions](const std::string &)
            {
                psOptions->eGeomOp = GEOMOP_SIMPLIFY_PRESERVE_TOPOLOGY;
            })
        .help(_("Distance tolerance for simplification."));
```

dbaston:
I've backed out the API additions for now. These remaining commits can be applied independently.

PR #11070, opened 2024-10-22
heif: fix memory leak of image data

bradh:
## What does this PR do?

In reworking the heif driver to support tiles (in #10982), I inadvertently introduced a memory leak. This fixes that.

N/A

## Tasklist

 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:

### DIFF
--- a/frmts/heif/heifdataset.cpp
+++ b/frmts/heif/heifdataset.cpp
@@ -748,7 +748,7 @@ CPLErr GDALHEIFRasterBand::IReadBlock(int nBlockXOff, int nBlockYOff,
             }
         }
     }
-
+    heif_image_release(hImage);
     return CE_None;
 }
 #else


frmts/heif/heifdataset.cpp
```diff
@@ -748,7 +748,7 @@ CPLErr GDALHEIFRasterBand::IReadBlock(int nBlockXOff, int nBlockYOff,
             }
         }
     }
-
+    heif_image_release(hImage);
```
rouault, line 751: shouldn't that be done too in the error code path at line 719 ?

bradh: I don't _think_ so. That is the error path for where we are filling in `hImage` and it will be untouched on failure.

(Reason for emphasis - I'm on low sleep, so maybe missing something here).

rouault: you're right. Just checked that https://github.com/strukturag/libheif/blob/bee0699e0e4eca18325d4d886d0590990ad51443/libheif/api/libheif/heif.cc#L1481 only sets a not-null hImage in the success code path, so no need to do heif_image_release()

PR #11071, opened 2024-10-22
[rfc] Add RFC 103 - OGR_SCHEMA open option

elpaso:
Formatted preview: https://gdal--11071.org.readthedocs.build/en/11071/development/rfc/rfc103_schema_open_option.html

### DIFF
--- a/doc/source/development/rfc/rfc103_schema_open_option.rst
+++ b/doc/source/development/rfc/rfc103_schema_open_option.rst
@@ -44,6 +44,7 @@ The structure of the JSON document will be similar to the one produced by the `o
 with the notable exception that (for the scope of this RFC) only the type of the fields will be considered:
 
 .. code-block:: json
+
     {
     "fields": [
         {
@@ -61,6 +62,7 @@ with the notable exception that (for the scope of this RFC) only the type of the
 In case of multi-layered datasets, the schema will be specified as a dictionary with the layer name as the key:
 
 .. code-block:: json
+    
     {
     "layer1": {
         "fields": [
--- a/doc/source/development/rfc/rfc103_schema_open_option.rst
+++ b/doc/source/development/rfc/rfc103_schema_open_option.rst
@@ -1,0 +1,105 @@
+.. _rfc-103:
+
+===================================================================
+RFC 103 add a SCHEMA open option to selected OGR drivers
+===================================================================
+
+=============== =============================================
+Author:         Alessandro Pasotti
+Contact:        elpaso at itopen.it
+Status:         Draft
+Created:        2024-10-22
+=============== =============================================
+
+Summary
+-------
+
+This RFC enables users to specify a SCHEMA open option in the OGR
+drivers that support it.
+
+The new option will be used to override the auto-detected fields types.
+
+Motivation
+----------
+
+Several OGR drivers must guess the attribute data type: CSV, GeoJSON, SQLite,
+the auto-detected types are not always correct and the user has no way to
+override them at opening tim.
+
+For the details please see the discussion attached to the issue: https://github.com/OSGeo/gdal/issues/10943
+
+Implementation
+--------------
+
+A new open option named SCHEMA will be added to the following drivers:
+
+- CSV
+- GeoJSON
+- SQLite
+- GML
+
+The option will be used to specify the schema in the form of a JSON document (or a path to a JSON file).
+
+The structure of the JSON document will be similar to the one produced by the `ogrinfo -json` command
+with the notable exception that (for the scope of this RFC) only the type of the fields will be considered:
+
+.. code-block:: json
+    {
+    "fields": [
+        {
+        "name": "field1",
+        "type": "string"
+        },
+        {
+        "name": "field2",
+        "type": "integer"
+        }
+    ]
+    }
+
+
+In case of multi-layered datasets, the schema will be specified as a dictionary with the layer name as the key:
+
+.. code-block:: json
+    {
+    "layer1": {
+        "fields": [
+        {
+            "name": "field1",
+            "type": "string"
+        },
+        {
+            "name": "field2",
+            "type": "integer"
+        }
+        ]
+    },
+    "layer2": {
+        "fields": [
+        {
+            "name": "field1",
+            "type": "string"
+        },
+        {
+            "name": "field2",
+            "type": "integer"
+        }
+        ]
+    }
+    }
+
+A preliminary draft of the implementation can be found at:
+https://github.com/elpaso/gdal/commits/enhancement-gh10943-fields-schema-override/
+
+
+Errors and warnings
+-------------------
+
+- If the schema is not a valid JSON document, a critical error will be raised.
+
+- If the schema is a valid JSON document but does not contain the expected fields or it is a no-op
+  (does not contain any actionable instruction), a warning will be raised and the schema will be ignored.
+
+- Additional JSON properties will be ignored while parsing the schema.
+
+- If the schema contains a field that is not present in the dataset, a warning will be raised and the field will be ignored.
--- a/doc/source/development/rfc/rfc103_schema_open_option.rst
+++ b/doc/source/development/rfc/rfc103_schema_open_option.rst
@@ -1,7 +1,7 @@
 .. _rfc-103:
 
 ===================================================================
-RFC 103 add a SCHEMA open option to selected OGR drivers
+RFC 103 add a OGR_SCHEMA open option to selected OGR drivers
 ===================================================================
 
 =============== =============================================
@@ -14,7 +14,7 @@ Created:        2024-10-22
 Summary
 -------
 
-This RFC enables users to specify a SCHEMA open option in the OGR
+This RFC enables users to specify a OGR_SCHEMA open option in the OGR
 drivers that support it.
 
 The new option will be used to override the auto-detected fields types.
@@ -31,7 +31,7 @@ For the details please see the discussion attached to the issue: https://github.
 Implementation
 --------------
 
-A new open option named SCHEMA will be added to the following drivers:
+A new reserved open option named OGR_SCHEMA will be added to the following drivers:
 
 - CSV
 - GeoJSON
@@ -55,7 +55,9 @@ with the notable exception that (for the scope of this RFC) only the type of the
         },
         {
         "name": "field2",
-        "type": "integer"
+        "type": "integer",
+        "width":11,
+        "precision":5
         }
     ]
     }
@@ -76,7 +78,9 @@ In case of multi-layered datasets, the schema will be specified as a list of lay
             },
             {
             "name": "field2",
-            "type": "integer"
+            "type": "integer",
+            "width":11,
+            "precision":5
             }
         ]
         },
@@ -89,12 +93,17 @@ In case of multi-layered datasets, the schema will be specified as a list of lay
             },
             {
             "name": "field2",
-            "type": "integer"
+            "type": "integer",
+            "width":11,
+            "precision":5
             }
         ]
         }
     ]
     }
+
+
+The new option will be used by applications such as `ogr2ogr` to override the auto-detected fields types.
 
 A preliminary draft of the implementation can be found at:
 https://github.com/elpaso/gdal/commits/enhancement-gh10943-fields-schema-override/
--- a/doc/source/development/rfc/rfc103_schema_open_option.rst
+++ b/doc/source/development/rfc/rfc103_schema_open_option.rst
@@ -40,10 +40,115 @@ A new reserved open option named OGR_SCHEMA will be added to the following drive
 
 The option will be used to specify the schema in the form of a JSON document (or a path to a JSON file).
 
-The schema document will not need to to contain a definition for every field and it will allow partial overrides of the auto-detected fields types.
+The schema document will allow both "Full" and "Patch" modes.
+"Patch" mode will be the default and will allow partial overrides of the auto-detected fields types
+while "Full" mode will replace all the auto-detected types with the ones specified in the schema.
 
 The structure of the JSON document will be similar to the one produced by the `ogrinfo -json` command
-with the notable exception that (for the scope of this RFC) only the information related to the type of the fields will be considered:
+with the notable exception that (for the scope of this RFC) only the information related to the type of
+the fields will be considered.
+
+The JSON schema for the OGR_SCHEMA open option will be as follows:
+
+.. code-block:: json
+
+    {
+        "$schema": "http://json-schema.org/draft-07/schema#",
+        "description": "Schema for OGR_SCHEMA open option",
+
+        "oneOf": [
+            {
+                "$ref": "#/definitions/dataset"
+            }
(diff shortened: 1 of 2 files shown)


doc/source/development/rfc/rfc103_schema_open_option.rst
```diff
@@ -0,0 +1,105 @@
...
+-------------------
+
+- If the schema is not a valid JSON document, a critical error will be raised.
+
+- If the schema is a valid JSON document but does not contain the expected fields or it is a no-op
+  (does not contain any actionable instruction), a warning will be raised and the schema will be ignored.
+
+- Additional JSON properties will be ignored while parsing the schema.
+
+- If the schema contains a field that is not present in the dataset, a warning will be raised and the field will be ignored.
```
jratike80, line 111: What happens if some fields are missing?

elpaso: Nothing: auto-detected types will be used. The idea is that you can override only a part of the schema.

elpaso: I have added a sentence to clarify that partial overrides are possible (in fact, they probably are the main use case).

doc/source/development/rfc/rfc103_schema_open_option.rst
```diff
@@ -0,0 +1,105 @@
...
+Implementation
+--------------
+
+A new open option named SCHEMA will be added to the following drivers:
+
+- CSV
+- GeoJSON
+- SQLite
+- GML
+
```
jratike80, line 40: GML can already use ,xsd and .gfs, and csv can have a .csvt sidecar, but maybe the more options the better. Especially manual writing of .csvt file is not convenient at all.

doc/source/development/rfc/rfc103_schema_open_option.rst
jratike80: I believe that it does not take long before someone wants to have an option to rename the fields by the same.
```
{
            "name": "field1",
            "new_name": "description"
            "type": "string"
 },
```

jratike80: And then someone would like to have an alternative for the OGR SQL EXCLUDE for removing some fields

> The syntax * EXCLUDE ([fields]) can be used to select all fields except those listed in parentheses.

They will not have that option if the fields which are missing from the schema are autodetected. But it is not possible to make both you and them happy at the same time, and you are the author...

doc/source/development/rfc/rfc103_schema_open_option.rst
```diff
@@ -0,0 +1,113 @@
...
+drivers that support it.
+
+The new option will be used to override the auto-detected fields types.
+
+Motivation
+----------
+
+Several OGR drivers must guess the attribute data type: CSV, GeoJSON, SQLite,
+the auto-detected types are not always correct and the user has no way to
+override them at opening tim.
```
rouault, line 27:
```suggestion
Several OGR drivers must guess the attribute data type: CSV, GeoJSON, SQLite, etc.
The auto-detected types are not always correct and the user has no way to
override them at opening them.
```

doc/source/development/rfc/rfc103_schema_open_option.rst
```diff
@@ -0,0 +1,113 @@
...
+
+Errors and warnings
+-------------------
+
+- If the schema is not a valid JSON document, a critical error will be raised.
+
+- If the schema is a valid JSON document but does not contain the expected fields or it is a no-op
+  (does not contain any actionable instruction), a warning will be raised and the schema will be ignored.
+
+- Additional JSON properties will be ignored while parsing the schema.
```
rouault, line 111: I would suggest that we ignore the schema if an expected JSON property is found with the error message mentionning that property. That way this will enable use to add more capabilites in the future, while giving the user a certainty that what he specified is fully taken into account.

doc/source/development/rfc/rfc103_schema_open_option.rst
rouault: I believe we should add a required top-level JSON property "schema_type" whose only supported value currently will be "patch", to mean that we alter the autodetected schema for the parts we specify with the JSON document. If in the future we would want to allow full schema replacement, that would be done with "schema_type" = "full" or something like that

doc/source/development/rfc/rfc103_schema_open_option.rst
```diff
@@ -0,0 +1,113 @@
...
+The structure of the JSON document will be similar to the one produced by the `ogrinfo -json` command
+with the notable exception that (for the scope of this RFC) only the type of the fields will be considered:
+
+.. code-block:: json
+
+    {
+    "fields": [
+        {
+        "name": "field1",
+        "type": "string"
```
rouault, line 54: We would need also to support "subtype" (for String JSON, or Integer Boolean).  And perhaps "width" and "precision" too

doc/source/development/rfc/rfc103_schema_open_option.rst
```diff
@@ -0,0 +1,113 @@
...
+Several OGR drivers must guess the attribute data type: CSV, GeoJSON, SQLite,
+the auto-detected types are not always correct and the user has no way to
+override them at opening tim.
+
+For the details please see the discussion attached to the issue: https://github.com/OSGeo/gdal/issues/10943
+
+Implementation
+--------------
+
+A new open option named SCHEMA will be added to the following drivers:
```
rouault, line 34: We should mention this open option will be a reserved one. If a driver uses it, it must be for that purpose

And I believe we should modify ogr2ogr so that -mapFieldType uses the capability offered by the SCHEMA open option when available

jratike80:
Oh, names. PostGIS driver has already `ACTIVE_SCHEMA` and `SCHEMAS`. OAPIF has `IGNORE_SCHEMA`.
Perhaps this new open option should have some more original name, like OGR_SCHEMA.

rouault:
> Perhaps this new open option should have some more original name, like OGR_SCHEMA.

+1

doc/source/development/rfc/rfc103_schema_open_option.rst
elpaso: The rationale for ignoring was to allow the _easy_ workflow `ogrinfo -json` -> edit the field types -> use that JSON document as an input for `ogr2ogr`. If we error out on unsupported properties the user will need to remove all unsupported properties. 

It is fine for me, just wanted to point this out.

Same observation goes for the "schema_type" recommended below.

doc/source/development/rfc/rfc103_schema_open_option.rst
elpaso: It makes sense.
Is it `subtype` also exposed by `ogrinfo -json` ?

doc/source/development/rfc/rfc103_schema_open_option.rst
sgillies: @jratike80 I agree that field renaming will be very useful. Instead of an implicit operation, where `new_name` means replace `name` or `new_type` means replace `type`, I'd prefer to use an explicit operation (sort of like https://jsonpatch.com/#replace) which says: replace field `name` with `{"name": "NEW_NAME", "type": "NEW_TYPE"}`.

sgillies: Now that I think about it, field renaming can be done with SQL, right? While overriding the original field type cannot.

jratike80: SQL is flexible and type casting is supported. I seem to have used this kind of SQL with ogr2ogr
```
SELECT 
'FIN' AS "country",
CAST("mtk_id" as text) as "id", 
'CC BY 4.0' AS "license",
ST_Simplify("geometry",5) AS "geometry"
FROM "source"
```
I am not sure how well GDAL detects those CASTed attributes generally, but they have worked for my use case with  GeoPackage as the outputformat.

However, some users feel that a need to know some SQL is weird, and having several SQL dialects, from which GDAL makes a selection by hidden rules, makes more confusion for beginners. But it is hard to set the limits when to use SQL, when to add a new specific open option or ogr2ogr switch, or when to write an OGR VRT file.

For me the OGR_SCHEMA feels more user friendly than SQL CAST especially because I fear that different dialects may not do exactly similar casts. By supporting also renaming users would not need to to one part of mapping with OGR_SCHEMA and another one with SQL.

rouault: > For me the OGR_SCHEMA feels more user friendly than SQL CAST

not only that, but type casting at the SQL level is too late. If a driver has wrongly guessed that a field was Integer32 wheras later records hold 64-bit values, the driver will present truncated/wrong values to the SQL engine. There's no way to "fix" that at that level. That must be something done by the driver itself

doc/source/development/rfc/rfc103_schema_open_option.rst
```diff
@@ -0,0 +1,125 @@
...
+
+    {
+    "layers":[
+        {
+        "name": "layer1",
+        "fields": [
+            {
+            "name": "field1",
+            "type": "string",
+            "subtype": "JSON"
```
rouault, line 79: Actually I see we have spelled it "subType" in ogrinfo -json output:
```
      "fields":[
        {
          "name":"a",
          "type":"String",
          "subType":"JSON",
          "nullable":true,
          "uniqueConstraint":false
        }
```

Schema in https://github.com/OSGeo/gdal/blob/master/apps/data/ogrinfo_output.schema.json#L117

doc/source/development/rfc/rfc103_schema_open_option.rst
rouault: > If we error out on unsupported properties the user will need to remove all unsupported properties.

If the error message is sufficiently explicit, I don't think that's much a concern. But there are clearly pros/cons

doc/source/development/rfc/rfc103_schema_open_option.rst
elpaso: For `schema_type` I was thinking to make it a layer-level property, this way a combination of patch and full can be used for individual layers.

I'm working on a json schema document to clarify this.

doc/source/development/rfc/rfc103_schema_open_option.rst
elpaso: I have added a JSON schema to the RFC: https://github.com/OSGeo/gdal/pull/11071/files#diff-9d0c9dc12d0b4a70d63be37cb390f1e9970d226296c30e8883c73a911c82806bR55

doc/source/development/rfc/rfc103_schema_open_option.rst
```diff
@@ -17,153 +17,198 @@ Summary
...
+The new option will be used to override the auto-detected fields types and to rename detected fields.
 
 Motivation
 ----------
 
 Several OGR drivers must guess the attribute data type: CSV, GeoJSON, SQLite,
 the auto-detected types are not always correct and the user has no way to
-override them at opening tim.
+override them at opening them.
+
```
jratike80, line 28: override them at opening time.

doc/source/development/rfc/rfc103_schema_open_option.rst
```diff
@@ -0,0 +1,275 @@
...
+            "name": "field2",
+            "newName": "new_field2"
+          }
+        ]
+      }
+    ]
+    }
+
+
+The new option will be used by applications such as `ogr2ogr` to override the auto-detected fields types and to override the auto-detected (and possibly laundered) field names.
```
rouault, line 257:
```suggestion
The new option will be used by applications such as `ogr2ogr` (for the ``-mapFieldType`` switch) to override the auto-detected fields types and to override the auto-detected (and possibly laundered) field names.
```

doc/source/development/rfc/rfc103_schema_open_option.rst
```diff
@@ -0,0 +1,275 @@
...
+
+In case of multi-layered datasets, the schema will be specified as a list of layers, each with its own fields definition and Patch/Full mode:
+
+.. code-block:: json
+
+    {
+    "layers":[
+        {
+        "name": "layer1",
+        "schema_type": "Full",
```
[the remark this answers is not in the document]
elpaso: definitely, ogrinfo already uses `subType` so I'd go with camel case. I will amend the schema.

doc/source/development/rfc/rfc103_schema_open_option.rst
```diff
@@ -0,0 +1,275 @@
...
+        "fields": [
+            {
+            "name": "field1",
+            "type": "String",
+            "subType": "JSON"
+            },
+            {
+            "name": "field2",
+            "newName": "new_field2"
```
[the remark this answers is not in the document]
elpaso: That makes sense to me. As an alternative we could treat the fields as an object such as:
```json
"fields" : { "field1" { "name" : "field1_renamed" ...}}
```
that would mean that we abandon the idea to use the output of `ogrinfo -json` as a template for the schema, but perhaps we have already lost that train.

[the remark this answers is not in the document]
jratike80: I hope not to lose everything from the `ogrinfo -json`. It would be rather demanding to write a working schema from a scratch, and having a separate option or utility for printing the default schema feels like duplication. But rather duplication than leaving users alone.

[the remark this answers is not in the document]
sgillies: @elpaso yeah, the original sin (so to speak) is that the OGR schema is represented as an array of fields (this is OGR's internal model) instead of as a dictionary. Thus we have to reference fields by "the item in the `fields` array where `name` is `val`" instead of just "`fields[val]`". I think it makes more sense to accept that than to change it.

rouault:
this PR can be rebased on top of latest master, and its status and voting history updated

elpaso:
> this PR can be rebased on top of latest master, and its status and voting history updated

done

doc/source/development/rfc/rfc103_schema_open_option.rst
```diff
@@ -0,0 +1,255 @@
...
+
+- PRs for the other drivers:
+    - htpps://github.com/OSGeo/gdal/pull/11479: CSV
+    - htpps://github.com/OSGeo/gdal/pull/11464: GeoJSON
+    - htpps://github.com/OSGeo/gdal/pull/11499: SQLite
+
+Voting history
+--------------
+
++1 from PSC members JukkaR, JavierJS, HowardB and EvenR
```
rouault, line 255:
```suggestion
+1 from PSC members JukkaR, JavierJS, HowardB and EvenR

.. below is an allow-list for spelling checker.

.. spelling:word-list::
    CSV
    GeoJSON
    SQLite
```

jratike80: What was changed?

rouault: addition of the spelling allow list

Issue #11067, opened 2024-10-21
Option explodecollections doesn't "explode" MULTIPOLYGON EMPTY geometries


theroggy:
### What is the bug?

Not a life-threatening problem, but when an input file contains e.g. MULTIPOLYGON EMPTY geometries and gdal.VectorTranslate() is ran on it with `explodecollections=True`, these MULTIPOLYGON EMPTY geometries aren't converted to POLYGON EMPTY.

E.g. for GeoPackage output, this leads to a non-conforming file being created and the following warning being emitted:
```
Warning 1: A geometry of type MULTIPOLYGON is inserted into layer parcels of geometry type POLYGON, which is not normally allowed by the GeoPackage specification, but the driver will however do it. To create a conformant GeoPackage, if using ogr2ogr, the -nlt option can be used to override the layer geometry type. This warning will no longer be emitted for this combination of layer and feature geometry type.
```

### Steps to reproduce the issue

``` python
from osgeo import gdal

gdal.UseExceptions()
options = gdal.VectorTranslateOptions(explodeCollections=True)
ds_output = gdal.VectorTranslate(
    srcDS="/vsicurl/https://github.com/geofileops/geofileops/raw/main/tests/data/polygon-parcel.gpkg",
    destNameOrDestDS="C:/temp/empty_exploded.gpkg",
    options=options,
)
```

### Versions and provenance

Tested on windows 11 with gdal 3.9.2 (installed from conda-forge)

rouault:
Ah ah I guess I deserved to be punished after posting https://erouault.blogspot.com/2024/09/those-concepts-in-geospatial-field-that.html :-)

jratike80:
Call it as workaround, but this seems to work:

`ogr2ogr -f gpkg mp.gpkg mp.jml -explodecollections -nlt polygon`

But I found something that does feel like a real bug.

Command
`ogr2ogr -f gpkg mp.gpkg mp.jml  -nlt polygon`

writes an invalid geometry into gpkg

` POLYGON ((-0.179209774695773 -0.867144071108579,0.687934296412806 -0.936515596797265,0.358419549391546 -1.78631678648367,-0.439352996028347 -1.67647853747659,-0.179209774695773 -0.867144071108579),(-1.39899243472184 -0.76308678257555,-0.422010114606175 -0.809334466368007,-0.653248533568463 -1.60132605131384,-0.809334466368007 -1.56085932799544,-1.39899243472184 -0.76308678257555))`

Test file 
[...]

PR #11076, opened 2024-10-22
ogr2ogr: fix -explodecollections on empty geometries

rouault:
Fix #11067

### DIFF
--- a/apps/ogr2ogr_lib.cpp
+++ b/apps/ogr2ogr_lib.cpp
@@ -6177,9 +6177,9 @@ bool LayerTranslator::Translate(
         int nIters = 1;
         std::unique_ptr<OGRGeometryCollection> poCollToExplode;
         int iGeomCollToExplode = -1;
+        OGRGeometry *poSrcGeometry = nullptr;
         if (bExplodeCollections)
         {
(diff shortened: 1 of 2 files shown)


apps/ogr2ogr_lib.cpp
```diff
@@ -6416,9 +6418,36 @@ bool LayerTranslator::Translate(
...
-                    poCollToExplode->removeGeometry(0, FALSE);
-                    poDstGeometry.reset(poPart);
+                    if (poSrcGeometry && poCollToExplode->IsEmpty())
+                    {
+                        const OGRwkbGeometryType eSrcType =
+                            poSrcGeometry->getGeometryType();
+                        const OGRwkbGeometryType eSrcFlattenType =
+                            wkbFlatten(eSrcType);
+                        OGRwkbGeometryType eDstType = eSrcType;
+                        if (eSrcFlattenType == wkbMultiPoint)
```
elpaso, line 6428: Can we make it  switch?

elpaso (approved):

apps/ogr2ogr_lib.cpp
rouault: done

PR #11088, opened 2024-10-22
Doc: Clarify behavior of gdal_translate -scale and -exponent

dbaston:


doc/source/programs/gdal_translate.rst
```diff
@@ -156,23 +156,33 @@ resampling, and rescaling pixels in the process.
     formats that support serializing statistics computations (GeoTIFF, VRT...)
     Note that the values specified after :option:`-scale` are only used to compute a scale and
     offset to apply to the input raster values. In particular, ``src_min`` and
-    ``src_max`` are not used to clip input values.
+    ``src_max`` are not used to clip input values unless :option:`-exponent`
+    is also specified. If clipping without exponential scaling is desired,
```
dbaston, line 160:
```suggestion
    is also specified.
    Source values that are outside the range of ``src_min`` and ``src_max`` will be scaled to values outside the range of ``dst_min`` and ``dst_max``. 
    If clipping without exponential scaling is desired,
```

dbaston: Added "Instead of being clipped, " to reinforce that we are talking about the case where "exponent" is not applied.

(I could also have imagined that out-of-range values are left alone, hence the addition)

jratike80:
The `-exponent 1` trick is nice!

Issue #11095, opened 2024-10-23
Add some create options to JPEGXL Driver to set distances for all extra channels


lidq92:
### Feature description

Is your feature request related to a problem? Please describe.
A bad behavior is found when applying lossy compression of multispectral images using JPEG XL.
This may be caused by the default distance for extra channels (e.g., 0 for lossless compression?).

Describe the solution you'd like
Add some options to set distances for all extra channels.

Describe alternatives you've considered
Set the default distance for extra channels the same as the main channels (first three or RGB).

### Additional context

https://gis.stackexchange.com/questions/487069/gdals-jpegxl-driver-shows-bad-behaviours-for-lossy-compression-of-multispectral

jratike80:
Getting test data from https://github.com/boazarad/ARAD_1K requires registration. Can you suggest anything else?

jratike80:
Ok, I have now a local 96 band image "coffee" from https://github.com/hacarus/hsi-open-dataset/tree/master. That has a "Creative Commons Attribution-NonCommercial 4.0 International License" so OK for testing but not suitable for the GDAL test set, if such need will arise later.
Please test if you can reproduce your issue with the coffee dataset.

You can combine a 96 band tiff with these commands
```
gdalbuildvrt -separate coffee.vrt coffee*.png
gdal_translate coffee.vrt coffee.tif
```

jratike80:
I think that I am using GDAL in a right way, and for me the lossy compression with the highest recommended distance=3 leads to  a  **bigger** file size than the lossless version:

```
gdal_translate -of JPEGXL -co EFFORT=7 -co lossless=yes coffee96.tif coffee96_lossless.jxl
file size: 986076
gdal_translate -of JPEGXL -co EFFORT=7 -co distance=3 coffee96.tif coffee96_lossy3.jxl
file size:  1016334
```
I do not know if the "coffee" image is too small for reasonable tests:
`Size is 192, 256`

lidq92:
> I found that by default the QUALITY/DISTANCE setting only applied to the R,G,B channels, and extra ones were encoded as lossless. Fixe per PR #11099

Thank you! I will have a try based on your branch.

lidq92:
@rouault After updating libheif to a version >= 1.18.0-rc1 (actually I used master 1.19.0), I successfully ran the code based on [rouault/gdal/fix_11095](https://github.com/rouault/gdal/tree/fix_11095).

I have two further questions:
[...]

PR #11098, opened 2024-10-23
JP2Lura: fix Identify() method

rouault:
I discovered accidentally the method was mis-identifying a EXR dataset as recognized by the JP2Lura driver while investigating https://github.com/OSGeo/gdal/issues/11095. This dates back to 2e650d8cace54238f863ca705cd20030409cf295 where for some reason (I suspect wrong copy&paste) the Identify() logic was completely changed


elpaso (approved):

elpaso:
A test would be nice to have so that we don't break it in the future.

PR #11099, opened 2024-10-23
JPEGXL: writer: propagate DISTANCE/QUALITY setting to extra channel

### LINKED ISSUE #11095: Add some create options to JPEGXL Driver to set distances for all extra channels

rouault:
Fixes #11095

### DIFF
--- a/frmts/jpegxl/jpegxl.cpp
+++ b/frmts/jpegxl/jpegxl.cpp
@@ -1798,16 +1798,35 @@ GDALDataset *JPEGXLDataset::CreateCopy(const char *pszFilename,
         }
     }
 
+    constexpr int DEFAULT_EFFORT = 5;
+    const int nEffort = atoi(CSLFetchNameValueDef(
+        papszOptions, "EFFORT", CPLSPrintf("%d", DEFAULT_EFFORT)));
+    const char *pszDistance = CSLFetchNameValue(papszOptions, "DISTANCE");
+    const char *pszQuality = CSLFetchNameValue(papszOptions, "QUALITY");
+    const char *pszAlphaDistance =
+        CSLFetchNameValue(papszOptions, "ALPHA_DISTANCE");
     const char *pszLossLessCopy =
         CSLFetchNameValueDef(papszOptions, "LOSSLESS_COPY", "AUTO");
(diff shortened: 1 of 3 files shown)


frmts/jpegxl/jpegxl.cpp
```diff
@@ -1798,16 +1798,35 @@ GDALDataset *JPEGXLDataset::CreateCopy(const char *pszFilename,
...
+    const int nEffort = atoi(CSLFetchNameValueDef(
+        papszOptions, "EFFORT", CPLSPrintf("%d", DEFAULT_EFFORT)));
+    const char *pszDistance = CSLFetchNameValue(papszOptions, "DISTANCE");
+    const char *pszQuality = CSLFetchNameValue(papszOptions, "QUALITY");
+    const char *pszAlphaDistance =
+        CSLFetchNameValue(papszOptions, "ALPHA_DISTANCE");
     const char *pszLossLessCopy =
         CSLFetchNameValueDef(papszOptions, "LOSSLESS_COPY", "AUTO");
-    if (EQUAL(pszLossLessCopy, "AUTO") || CPLTestBool(pszLossLessCopy))
+    if ((EQUAL(pszLossLessCopy, "AUTO") && !pszDistance && !pszQuality &&
```
elpaso, line 1810: Not sure if I read this correctly but it seems to me it can be reduced to 
if (EQUAL(pszLossLessCopy, "AUTO") && ((first group of checks) || (second group of checks))

elpaso (approved):

frmts/jpegxl/jpegxl.cpp
rouault: actually not because the second group test for *not* equal to AUTO

PR #11107, opened 2024-10-24
JPEGXL: allow a minimum DISTANCE of 0.01, max of 25 and revise QUALITY to DISTANCE formula

### LINKED ISSUE #11095: Add some create options to JPEGXL Driver to set distances for all extra channels

rouault:
to be in sync with latest libjxl

Fixes #11095

### DIFF
--- a/autotest/gdrivers/jpegxl.py
+++ b/autotest/gdrivers/jpegxl.py
@@ -127,7 +127,7 @@ def test_jpegxl_rgba_distance():
 
 
 @pytest.mark.parametrize(
-    "quality,equivalent_distance", [(100, 0), (90, 1), (10, 12.65)]
+    "quality,equivalent_distance", [(100, 0), (10, 15.266666666666667)]
 )
 def test_jpegxl_rgba_quality(quality, equivalent_distance):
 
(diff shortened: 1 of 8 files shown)


lidq92:
Thank you very much for the quick update! I have one more question related to the distance and quality.

The third part of the mapping function from quality to distance is different between [libjxl](https://github.com/libjxl/libjxl/blob/c6355600c5932e3a9f1b95a4cadfa9b718fbec8f/lib/jxl/encode.cc#L1551) and [GDAL's JPEG XL Driver](https://github.com/OSGeo/gdal/blob/master/frmts/jpegxl/jpegxl.cpp#L2140). And the max(d) value allowed is different, i.e., 25 for libjxl, 15 for JPEG XL Driver of GDAL. I'm very interested in the reasons behind it. 

You're really great, thank you for your answers and updates!

rouault:
> I'm very interested in the reasons behind it.

just that libjxl has apparenly changed its validity range for distance and its quality->distance formula since last time we sync with it. Just amened my pull request to resync

PR #11110, opened 2024-10-24
Doc: document that JXL-in-GeoTIFF requires tiled=yes

jratike80:
I suppose that recommending 256x256, 512x512, or 1024x1024 sized blocks makes sense, because the "group" in JPEG XL is 256x256 pixels.

### DIFF
--- a/doc/source/drivers/raster/gtiff.rst
+++ b/doc/source/drivers/raster/gtiff.rst
@@ -485,9 +485,10 @@ This driver supports the following creation options:
       * ``LERC_ZSTD`` is available when ``LERC`` and ``ZSTD`` are available.
 
       * ``JXL`` is for JPEG-XL, and is only available when using internal libtiff and building GDAL against
-        https://github.com/libjxl/libjxl . Supported data types are ``Byte``, ``UInt16`` and ``Float32`` only.
-        For GDAL < 3.6.0, JXL compression may only be used alongside ``INTERLEAVE=PIXEL`` (the default) on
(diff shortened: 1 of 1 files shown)


jratike80:
Or could it be possible to switch into `tiled=yes` automatically when the compression is JPEG XL, because stripes do not make sense?

rouault:
> Or could it be possible to switch into `tiled=yes` automatically when the compression is JPEG XL, because stripes do not make sense?

someone using BLOCKYSIZE=256 could potentially achieve reasonable stripe performance

doc/source/drivers/raster/gtiff.rst
```diff
@@ -485,9 +485,10 @@ This driver supports the following creation options:
       * ``LERC_ZSTD`` is available when ``LERC`` and ``ZSTD`` are available.
 
       * ``JXL`` is for JPEG-XL, and is only available when using internal libtiff and building GDAL against
-        https://github.com/libjxl/libjxl . Supported data types are ``Byte``, ``UInt16`` and ``Float32`` only.
-        For GDAL < 3.6.0, JXL compression may only be used alongside ``INTERLEAVE=PIXEL`` (the default) on
-        datasets with 4 bands or less.
+        https://github.com/libjxl/libjxl . JXL compression should be used with the ``TILED=YES`` creation
```
rouault, line 488:
```suggestion
        https://github.com/libjxl/libjxl . It is recommended to use JXL compression with the ``TILED=YES`` creation
```

PR #11111, opened 2024-10-24
Remove gdalwarpsimple and ogrdissolve

dbaston:
## What does this PR do?

Removes the `gdalwarpsimple` and `ogrdissolve` utilities.

`gdalwarpsimple` is what was referred to as "the old gdalwarp" in the GDAL 1.2 release notes. It does not build without warnings, and I cannot see any sign of it being used.

`ogrdissolve` is in the source tree but not included in the build system.


dbaston:
Other non-testing applications that are not built by default: `dumpoverviews`, `gdalflattenmask`, `gdal2ogr`.

rouault (approved):

PR #11138, opened 2024-10-27
SQLite/GPKG: turn on SQLite 'PRAGMA secure_delete=1' by default

rouault:
unless the user specifies a value of this pragma through OGR_SQLITE_PRAGMA.

For example, Debian and Conda-Forge SQLite3 builds already turn on secure_delete. Make sure that this is done for all SQLite3 builds.

### DIFF
--- a/autotest/ogr/ogr_gpkg.py
+++ b/autotest/ogr/ogr_gpkg.py
@@ -10540,3 +10540,49 @@ def test_gpkg_create_more_than_2000_fields(tmp_vsimem):
         with pytest.raises(Exception, match="Limit of 2000 columns reached"):
             lyr.CreateField(ogr.FieldDefn("foo"))
         assert lyr.GetLayerDefn().GetFieldCount() == 2000 - 2
+
+
+###############################################################################
+# Test that secure_delete is turned on
+
+
+@gdaltest.enable_exceptions()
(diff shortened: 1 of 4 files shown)


jratike80:
I make a few tests. SECURE_DELETE=ON makes deleting 5.6 million rows from a regular, non-spatial table 6 times slower, 2.5 sec vs 15 sec.
However, deleting massive amounts of data from SQLite is always very slow when tables have triggers, and that is the case with spatial geopackage tables.

Deleting a 5.6 million table with the same data as in the previous tests but now with r-tree, triggers and everything in geopackage requires this much time:
secure_delete=no 326 seconds
secure_delete=yes 370 seconds

Considerations about similar, but more compicated use case https://gis.stackexchange.com/questions/486885/better-control-cascading-triggers-in-sqlite

rouault:
> Deleting a 5.6 million table with the same data as in the previous tests but now with r-tree, triggers and everything in geopackage requires this much time:

Will be vastly improved with https://github.com/OSGeo/gdal/pull/11139

rouault:
> Conda-Forge SQLite3 builds already turn on secure_delete

On deeper review, was actually only true of the Linux and Mac builds. Submitted https://github.com/conda-forge/sqlite-feedstock/pull/116 to do that on Windows too

PR #11139, opened 2024-10-27
GPKG: optimize speed of 'DELETE FROM table_name', especially on ones with RTree

rouault:
On a .gpkg of 1.6 GB with a table of 3.2 million features, this decreases the time from 1 minute to 12 seconds (with secure delete enabled, or 1 second with secure delete disabled)

### DIFF
--- a/ogr/ogrsf_frmts/gpkg/ogrgeopackagetablelayer.cpp
+++ b/ogr/ogrsf_frmts/gpkg/ogrgeopackagetablelayer.cpp
@@ -9140,3 +9140,179 @@ OGRErr OGRGeoPackageTableLayer::GetExtent3D(int iGeomField,
 
     return OGRERR_NONE;
 }
+
+/************************************************************************/
+/*                           Truncate()                                 */
+/************************************************************************/
+
+/** Implements "DELETE FROM {table_name}" in an optimzed way.
+ *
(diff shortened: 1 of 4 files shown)


ogr/ogrsf_frmts/gpkg/ogrgeopackagetablelayer.cpp
```diff
@@ -9140,3 +9140,179 @@ OGRErr OGRGeoPackageTableLayer::GetExtent3D(int iGeomField,
...
+        explicit ReenableTriggers(sqlite3 *hDB) : m_hDB(hDB)
+        {
+        }
+
+        ~ReenableTriggers()
+        {
+            sqlite3_db_config(m_hDB, SQLITE_DBCONFIG_ENABLE_TRIGGER, 1,
+                              nullptr);
+        }
+        CPL_DISALLOW_COPY_ASSIGN(ReenableTriggers)
```
elpaso, line 9183: Is this necessary?

elpaso (approved):

ogr/ogrsf_frmts/gpkg/ogrgeopackagetablelayer.cpp
rouault: yes, to please -Weffc++

PR #11170, opened 2024-10-30
CI: remove alpine_numpy2 config, now that alpine:edge has switched to numpy 2.1.2

rouault:


rouault:
cherry-picked to 3.10 per 2f350c58061d9daaa74ab17355b4a1fa07a8e291

Issue #11181, opened 2024-10-31
Test failures with data/hdf5/u8be.h5 and MSYS2 mingw-w64-x86_64-hdf5-1.14.5-1


rouault:
The build-windows-msys2-mingw CI target is now failing on tests that involve the autotest/gdrivers/data/hdf5/u8be.h5 file, since MSYS2 has updated their mingw-w64-x86_64-hdf5 package from 1.14.3-2 to 1.14.5-1

Last good build: https://github.com/OSGeo/gdal/actions/runs/11599287020/job/32297101088  (OSGeo/GDAL master)
First failed build: https://github.com/rouault/gdal/actions/runs/11604312201/job/32312870698  (a branch of mine on ~ same master + unrelated changes)

msys2 related update is at https://github.com/msys2/MINGW-packages/pull/22342

GDAL autotest test failures seem to suggest that the file can be opened, but that the array of type H5T_STD_U8BE (unsigned integer 8 bit big endian  (note of myself: why does HDF5 need to distinguish little and big endian for a single-byte data type...)):
```
2024-10-31T01:42:31.2420877Z 26: FAILED hdf5.py::test_hdf5_3 - RuntimeError: `HDF5:"data/hdf5/u8be.h5"://TestArray' does not exist in the file system, and is not recognized as a supported dataset name.
2024-10-31T01:42:31.2423598Z 26: FAILED hdf5.py::test_hdf5_4 - RuntimeError: `HDF5:"data/hdf5/u8be.h5"://TestArray' does not exist in the file system, and is not recognized as a supported dataset name.
2024-10-31T01:42:31.2426093Z 26: FAILED hdf5multidim.py::test_hdf5_multidim_basic - AssertionError: assert [] == ['TestArray']
```

A local Linux build of hdf5 1.14.5 doesn't show any difference in the output of the h5dump utility compared to earlier versions, so it might be either an issue with the API, or a Windows specific issue, or a msys2 packaging one...

PR #11182, opened 2024-10-31
CI: build-windows-msys2-mingw: disable hdf5 because of test failures with hdf5 1.14.5-1

rouault:
Refs #11181


rouault:
cherry-picked to 3.10 per ae97323c47ce5329a927f95cb7775ef07aa65493

PR #11199, opened 2024-11-02
Add gdal_minmax_element.hpp public header, that can also be vendored, to find the min/max elements in a buffer

rouault:
Refs https://github.com/qgis/QGIS/pull/59285

NaN values are taken into account for float/double
Contains a SSE2 optimized version for int8/uint8/int16/uint16/int32/uint32/float/double in the no-nodata case (also taking into account NaN), which is about 10x times than std::min_element / std::max_element
Also add SSE2 optimization for uint8, int8, uint16, int16, uint32, int32, float and double nodata cases

<details>
<summary>
With gcc 9.4 (not so different with gcc 14.2.1, which is just slightly bit better for the "(nodata case, using std::min_element with nodata aware comparison)" cases for uint8/int8/uint16/int16, and significantly better for uint32/int32, but still behind our manual optimization ):
</summary>

```bash
$ (cd perftests && g++ -DGDAL_COMPILATION -std=c++17 -I../../port -I../port -I../../gcore -I../gcore -I../../ogr -O2 ../../perftests/testperf_gdal_minmax_element.cpp  -o testperf_gdal_minmax_element  -Wl,-rpath,/home/even/gdal/gdal/build_cmake ../libgdal.so.36.3.11.0  && ./testperf_gdal_minmax_element)
uint8:
min at idx 210901 (optimized)
-> elapsed=697957
min at idx 210901 (using std::min_element)
-> elapsed=19272114
min at idx 349447 (nodata case, optimized)
-> elapsed=963895
min at idx 349447 (nodata case, using std::min_element with nodata aware comparison)
-> elapsed=22531919
--------------------
int8:
min at idx 0 (optimized)
-> elapsed=639479
min at idx 0 (using std::min_element)
-> elapsed=16144175
min at idx 0 (nodata case, optimized)
-> elapsed=742246
min at idx 0 (nodata case, using std::min_element with nodata aware comparison)
-> elapsed=16512409
--------------------
uint16:
min at idx 17343 (optimized)
-> elapsed=1318264
min at idx 17343 (using std::min_element)
-> elapsed=16481840
min at idx 119863 (nodata case, optimized)
-> elapsed=1514535
min at idx 119863 (nodata case, using std::min_element with nodata aware comparison)
-> elapsed=18770894
--------------------
int16:
min at idx 5640996 (optimized)
-> elapsed=1276417
min at idx 5640996 (using std::min_element)
-> elapsed=16151810
min at idx 5640996 (nodata case, optimized)
-> elapsed=1418040
[...]

PR #11200, opened 2024-11-03
Add GDALRasterComputeMinMaxLocation / GDALRasterBand::ComputeMinMaxLocation, and map it to SWIG

rouault:
and add a gdal_minmax_location.py script:
    
```
$ python ../swig/python/gdal-utils/osgeo_utils/samples/gdal_minmax_location.py byte.tif
Minimum=74.0 at (col,line)=(9,17), (X,Y)_georef=(441290.0,3750270.0), (long,lat)_WGS84=(-117.6358076,33.8929309)
Maximum=255.0 at (col,line)=(2,18), (X,Y)_georef=(440870.0,3750210.0), (long,lat)_WGS84=(-117.6403456,33.8923662)
```

(on top of PR #11199)

### DIFF
--- a/gcore/gdal_minmax_element.hpp
+++ b/gcore/gdal_minmax_element.hpp
@@ -72,6 +72,7 @@ namespace GDAL_MINMAXELT_NS
 namespace detail
 {
 
+#ifdef GDAL_MINMAX_ELEMENT_USE_SSE2
 /************************************************************************/
 /*                            compScalar()                              */
 /************************************************************************/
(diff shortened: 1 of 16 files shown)


gcore/gdal_minmax_element.hpp
```diff
@@ -0,0 +1,971 @@
...
+                                           dfNoDataValue);
+}
+
+namespace detail
+{
+
+#ifdef NOT_EFFICIENT
+
+/************************************************************************/
+/*                         minmax_element()                             */
```
bdezonia, line 698: So there are algorithms for finding both the min and max element of a list of numbers at the same time that beat the speed of iterating the list and checking each number versus min and each numeber versus mac and updating the min and max as you go. You can do less comparisons by pulling out two numbers at a time, comparing them to each other, and checking the min one to overall min and the max one versus overall max. I think this ends up doing 3/4's the number of overall comparisons than the naive approach. I have lost the reference for this but I coded it in my java library. See here; https://github.com/bdezonia/zorbage/blob/master/src/main/java/nom/bdezonia/zorbage/algorithm/MinMaxElement.java

rouault: Thanks for the suggestion but saving 1/4 of operations isn't going to improve performance in a significant way.

Current timings are for a 10 million Byte vector:
- using std::minmax_element(): 15,062,444
- using NOT_EFFICIENT code path: 9,625,260
- using SSE2 optimized gdal::min_element() + gdal::max_element() (enabled currently): 1,341,901

So your suggestion would perhaps reduce NOT_EFFICIENT code path timing, but it would be still far away from the  twice repeated SSE2 optimized code paths. Actually I've just tested it and it results in... 50% slower code than the naive version... Presumably to degraded pipelining

jratike80:
The gdal-dev version from OSGeo4W installed today shows version `GDAL 3.11.0dev-d9b65d7609, released 2024/11/05`. The date is too old for this minmax feature, but if the version string refers to the number of last included commit, the d9b65d7609 removed the Lura driver and it was a few days later. So maybe the GDALRasterComputeMinMaxLocation should be available. However, running gdal_minmax_location.py prints an error:
`AttributeError: 'Band' object has no attribute 'ComputeMinMaxLocation'`

 Should I wait for new nightly builds, or could there be something wrong in the OSGeo4W?

rouault:
> GDAL 3.11.0dev-d9b65d7609, released 2024/11/05

@jef-n Is the date of the system where OSGeo4W is build up-to-date ?

> so maybe the GDALRasterComputeMinMaxLocation should be available

yes, it should be available. Commit d9b65d7609e62d89744b6b1cbdf7e35d4163e212 is the merge request from yesterday, so it must include GDALRasterComputeMinMaxLocation. Are you sure your PYTHONPATH is correctly set to point to the updated GDAL python bindings?

Issue #10067, opened 2024-05-29
Allow writing VRT from gdal_calc


dbaston:
### Feature description

`gdal_calc` performs its calculations eagerly. It would be useful to be able to describe a transformation of remote datasets (e.g., NDVI) and store the result as a VRT.

Possible implementations:
- Translate the expression to a Python pixel function (simpler, would require user of the VRT to set `GDAL_VRT_ENABLE_PYTHON=YES`
- Translate simple expressions to C++ pixel functions (more complex, would likely result in lots of nesting)

dbaston:
@hobu I've done a quick test with the [exprtk library](https://www.partow.net/programming/exprtk/). It seems a bit overkill for this purpose, but it ships as a single MIT-licensed header and integrates very easily as a GDAL pixel function. For example, the following evaluates successfully:

```
<VRTDataset rasterXSize="1" rasterYSize="1">
<VRTRasterBand dataType="Float64" band="1" subClass="VRTDerivedRasterBand">
  <PixelFunctionType>expression</PixelFunctionType>
  <PixelFunctionArguments expression="if (A > B) 1.5*C ; else A" />
  <SimpleSource>
    <SourceFilename relativeToVRT="0">/tmp/pytest-of-dan/pytest-79/test_vrt_pixelfn_expression_co0/source_0.tif</SourceFilename>
    <SourceBand>1</SourceBand>
  </SimpleSource>
  <SimpleSource>
    <SourceFilename relativeToVRT="0">/tmp/pytest-of-dan/pytest-79/test_vrt_pixelfn_expression_co0/source_1.tif</SourceFilename>
    <SourceBand>1</SourceBand>
  </SimpleSource>
  <SimpleSource>
    <SourceFilename relativeToVRT="0">/tmp/pytest-of-dan/pytest-79/test_vrt_pixelfn_expression_co0/source_2.tif</SourceFilename>
    <SourceBand>1</SourceBand>
  </SimpleSource>
</VRTRasterBand>
</VRTDataset>
```

It would nice to figure out a way to specify a name for each `<SimpleSource>` and make that source name visible to the expression parser.

dbaston:
[...]

PR #11209, opened 2024-11-05
VRT Pixel Functions: Add function to evaluate arbitrary expression

dbaston:
## What does this PR do?

Adds a C++ pixel function called "expression" that can evaluate an arbitrary expression (or indeed, a mini-program) using the [exprtk library](https://www.partow.net/programming/exprtk/index.html). This is a single MIT-licensed header that is easily integrated into GDAL. It appears to be quite widely used. However, given that a major purpose of this is to extend the utility of VRT to users that don't want to allow functions defined in Python (possibly because of security concerns), I'm a little uncomfortable with the scope of expressions allowed by this library. Some of these, such as file I/O, can be disabled (and I've done so.)

I've also looked at:
- [tinyexpr](https://github.com/codeplea/tinyexpr), which seems easily integrated but lacks conditionals
- [muparser](https://github.com/beltoforion/muparser), a little harder to integrate (includes multiple header/object files) and supports a reasonable variety of functions including a ternary conditional operator

I need to do more work to investigate details such as nodata handling, but wanted to share the concept sooner rather than later.

An alternative implementation would expose `exprtk` as an additional pixel function language alongside Python, instead of hooking into the existing C++ pixel function machinery.

An example VRT that is allowed is:

```
<VRTDataset rasterXSize="1" rasterYSize="1">
  <VRTRasterBand dataType="Float64" band="1" subClass="VRTDerivedRasterBand">
    <PixelFunctionType>expression</PixelFunctionType>
    <PixelFunctionArguments expression="(NIR-R)/(NIR+R)" />
    <SimpleSource name="NIR">
      <SourceFilename relativeToVRT="0">source_0.tif</SourceFilename>
      <SourceBand>1</SourceBand>
    </SimpleSource>
    <SimpleSource name="R">
      <SourceFilename relativeToVRT="0">source_1.tif</SourceFilename>
      <SourceBand>1</SourceBand>
    </SimpleSource>
  </VRTRasterBand>
</VRTDataset>
```


frmts/vrt/pixelfunctions.cpp
```diff
@@ -15,6 +15,12 @@
 #include "gdal.h"
 #include "vrtdataset.h"
 
+#define exprtk_disable_caseinsensitivity
+#define exprtk_disable_rtl_io
+#define exprtk_disable_rtl_io_file
+#define exprtk_disable_rtl_vecops
+#define exprtk_disable_string_capabilities
+#include <exprtk.hpp>
```
rouault, line 23: Maybe the following to avoid namespace collisions in case GDAL is integrated with another software also using exprtk?
```suggestion
#define exprtk gdal_exprtk
#include <exprtk.hpp>
```

rouault:
That's super cool, and so compact as an integration!

Any hints on the performance ? In particular it would be cool if exprtk could use SIMD to vectorize computations, but I'm probably asking too much.  Regarding performance, I'd suspect the biggest issue to be the GDALCopyWords() done for each pixel. Creating a temporary array with all the double values and copying in one-go to the output data type would certainly be beneficial (this remark applies to existing pixel functions).

Can exptrtk can be used with non-double variables ? (in case that would make things faster for integer only computations)

Thinking out loud: for ultimate performance, we should probabl use some LLVM-based solution where the expression would be just-in-time compiled to the target architecture, a bit like [numba](https://numba.pydata.org/) does.

dbaston:
> Any hints on the performance ? 

Haven't looked into this yet. In particular, comparing to Python would be interesting.

> Creating a temporary array with all the double values and copying in one-go to the output data type would certainly be beneficial (this remark applies to existing pixel functions).

Maybe they could be rewritten as templates...

> Can exptrtk can be used with non-double variables ? (in case that would make things faster for integer only computations)

The docs note "Support for various numeric types (float, double, long double, [MPFR/GMP](https://www.mpfr.org/))"

> Thinking out loud: for ultimate performance, we should probabl use some LLVM-based solution where the expression would be just-in-time compiled to the target architecture, a bit like [numba](https://numba.pydata.org/) does.

I also came across [mathpresso](https://github.com/kobalicek/mathpresso) that does something along these lines.

dbaston:
@rouault A limitation of the pixel function integration is that we can only output a single band. exprtk would happily accept an expression like `"var q[2] := {B2, B3}; B1 * q;"` but we have no way to return the second band to the user. Do you think looking into a `VRTProcessedDataset` integration might make sense instead?

On performance, I've done basic tests like summing up a 24-band netCDF. Performance is roughly equivalent to Python with `perf` showing the following for exprtk:

```
  23.81%  gdal_translate  libgdal.so.36.3.11.0        [.] ExprPixelFunc
  15.50%  gdal_translate  libgdal.so.36.3.11.0        [.] exprtk::details::vec_add_op<double>::process
  12.32%  gdal_translate  libgdal.so.36.3.11.0        [.] (anonymous namespace)::GDALCopyWordsFromT<short>
  11.21%  gdal_translate  libgdal.so.36.3.11.0        [.] GDALCopyWords64
   6.31%  gdal_translate  libnetcdf.so.19             [.] ncx_getn_short_short
   4.06%  gdal_translate  [unknown]                   [k] 0xffffffff98a0109c
   3.31%  gdal_translate  libc.so.6                   [.] __memset_avx2_unaligned_erms
   2.47%  gdal_translate  [unknown]                   [k] 0xffffffff98a00158
``` 

and numpy:

```
  16.49%  gdal_translate  libc.so.6                                          [.] __memmove_avx_unaligned_erms
  13.34%  gdal_translate  _multiarray_umath.cpython-310-x86_64-linux-gnu.so  [.] DOUBLE_add_FMA3__AVX2
  12.76%  gdal_translate  libgdal.so.36.3.11.0                               [.] (anonymous namespace)::GDALCopyWordsFromT<short>
   6.90%  gdal_translate  [unknown]                                          [k] 0xffffffff98a00158
   6.82%  gdal_translate  libnetcdf.so.19                                    [.] ncx_getn_short_short
   4.28%  gdal_translate  [unknown]                                          [k] 0xffffffff98a0109c
   4.23%  gdal_translate  ld-linux-x86-64.so.2                               [.] do_lookup_x
   3.13%  gdal_translate  libc.so.6                                          [.] __memset_avx2_unaligned_erms
   1.98%  gdal_translate  libpython3.10.so.1.0                               [.] _PyEval_EvalFrameDefault
```

rouault:
> Do you think looking into a `VRTProcessedDataset` integration might make sense instead?

It accepts only a single (potentially multiband) input dataset. That said the Input can be a VRTDataset, so you can assemble several single-band datasets as a virtual multiple one. Hard to tell which way is the preferred one. Would having it available in both VRTDerivedRasterBand and VRTProcessedDataset be overkill ? Having it available in VRTDerivedRasterBand makes it probably easier/intutive to write formulas that are different per output-band, even if exprtk allows to return a tuple.

dbaston:
> Would having it available in both VRTDerivedRasterBand and VRTProcessedDataset be overkill ?

Not necessarily. I'll look into `VRTProcessedDataset` some more.

> Having it available in VRTDerivedRasterBand makes it probably easier/intutive to write formulas that are different per output-band

On the other hand, this requires repeating all of the `SimpleSource` definitions (and re-reading the input pixels) for every output band...unless I'm missing something.

rouault:
> On the other hand, this requires repeating all of the `SimpleSource` definitions (and re-reading the input pixels) for every output band.

that's true

frmts/vrt/vrtprocesseddatasetfunctions.cpp
```diff
@@ -1468,6 +1472,218 @@ static CPLErr TrimmingProcess(
     return CE_None;
 }
 
+/************************************************************************/
+/*                    ExpressionInit()                                  */
+/************************************************************************/
+
+class ExpressionData
```
rouault, line 1479: wrap it into a anonymous namespace to avoid potential ODR issues

autotest/gdrivers/vrtprocesseddataset.py
```diff
@@ -1198,6 +1198,123 @@ def test_vrtprocesseddataset_trimming_errors(tmp_vsimem):
...
+        pytest.param(
+            "return [ALL_BANDS]",
+            np.array([[[1, 2]], [[3, 4]], [[5, 6]]]),
+            np.array([[[1, 2]], [[3, 4]], [[5, 6]]]),
+            None,
+            id="multiple bands in, multiple bands out (2)",
+        ),
+        pytest.param(
+            """
+             var chunk[10];
```
rouault, line 1224: That's interesting... from a security point of view.
- can that trigger arbitrary memory allocation ?
- does exprtk does bound checking when accessing the arrays ?
I suspect you could also cause infinite /very long looping ?

dbaston: > can that trigger arbitrary memory allocation

Array sizes can be up to 2 billion, so you can cause a large allocation with:

```             
var out[2000000000];
return [out];
```

> does exprtk do bound checking when accessing the arrays ?

Not by default, but enabled in fd94c9e0ed1a10edf90031312128f5826d8e1cdf

frmts/vrt/vrtexpression.cpp
```diff
@@ -0,0 +1,199 @@
...
+                         context.type_size;
+        auto nIndexAccessed = (static_cast<std::uint8_t *>(context.access_ptr) -
+                               static_cast<std::uint8_t *>(context.base_ptr)) /
+                              context.type_size;
+
+        std::ostringstream oss;
+        oss << "Attempted to access index " << nIndexAccessed
+            << " in a vector of " << nElements
+            << " elements when evaluating VRT expression.",
+            throw std::runtime_error(oss.str());
```
rouault, line 28:
```suggestion
            << " elements when evaluating VRT expression.";
        throw std::runtime_error(oss.str());
```

autotest/gdrivers/vrtprocesseddataset.py
rouault: > Array sizes can be up to 2 billion, so you can cause a large allocation with:

could we improve exprtk to offer a way to limit the allocations?

ArashPartow: > could we improve exprtk to offer a way to limit the allocations?

@rouault: Yes the maximum vector size that can be created in the expression can be set explicitly:

https://www.partow.net/programming/exprtk/readme.html#1473


```
parser_t parser; 
parser.settings().set_max_local_vector_size(12345); 

```

third_party/exprtk/exprtk.hpp
```diff
@@ -3188,7 +3188,10 @@ namespace exprtk
                }
             }
 
-            return (g.token_list_.size() - stride_ + 1);
+            // This return value appears to be unused and may cause undefined
+            // behavior when stride_ > (g.token_list_.size() + 1)
+            // return (g.token_list_.size() - stride_ + 1);
+            return 0;
```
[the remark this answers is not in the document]
dbaston: @ArashPartow . Here's a minimal example: 

test.cpp

```
#include "../gdal/third_party/exprtk/exprtk.hpp"

int main() {
    exprtk::expression<double> expression;
    exprtk::parser<double> parser;
    exprtk::symbol_table<double> table;

    double x = 8;
    table.add_variable("X", x);
    expression.register_symbol_table(table);
    parser.compile("X", expression);

    double value = expression.value();

    return value != 8;
}
```
running

```
clang++ -fsanitize=undefined -fsanitize=unsigned-integer-overflow test.cpp && ./a.out
```

produces
```
../gdal/third_party/exprtk/exprtk.hpp:3193:42: runtime error: unsigned integer overflow: 1 - 2 cannot be represented in type 'unsigned long'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior ../gdal/third_party/exprtk/exprtk.hpp:3193:42 in 
../gdal/third_party/exprtk/exprtk.hpp:3193:52: runtime error: unsigned integer overflow: 18446744073709551615 + 1 cannot be represented in type 'unsigned long'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior ../gdal/third_party/exprtk/exprtk.hpp:3193:52 in
```

Originally showed up in our CI here (https://github.com/OSGeo/gdal/actions/runs/11784348075/job/32823345209#step:17:29078)

rouault:
The ASAN failures are unrelated to this PR and are avoided in master per fbaa99cb08d1b4d6f832bfa711879ea2bb3a0b35. You can rebase on top of that.

Perhaps we could limit the amount of memory to some default value, let's say 10 MB, and if above that, fail with an error like "you need to set the GDAL_VRT_MAX_MEMORY_EXPR configuration option".

frmts/vrt/vrtexpression.cpp
```diff
@@ -0,0 +1,224 @@
...
+    bool m_bIsCompiled{false};
+
+    Impl()
+    {
+        using settings_t = std::decay_t<decltype(m_oParser.settings())>;
+
+        m_oParser.register_vector_access_runtime_check(m_oVectorAccessCheck);
+
+        int nMaxVectorSize = std::atoi(
+            CPLGetConfigOption("GDAL_EXPRTK_MAX_VECTOR_SIZE", "100000"));
```
rouault, line 57: maybe use your new CPLParseMemorySize ?

frmts/vrt/vrtexpression.cpp
```diff
@@ -0,0 +1,224 @@
...
+
+        if (nMaxVectorSize > 0)
+        {
+            m_oParser.settings().set_max_local_vector_size(nMaxVectorSize);
+        }
+
+        bool bEnableLoops =
+            CPLTestBool(CPLGetConfigOption("GDAL_EXPRTK_ENABLE_LOOPS", "YES"));
```
rouault, line 65: I see in exptrk doc that there is a "Loop Iteration Checks" paragraph with ways to limit the number of iterations or runtime. For a single target pixel computation, it would be crazy that we spend more than  let's say 1 second (and if that threshold is reached, we would error out for following pixels).

ArashPartow: Here is an example that demonstrates the **_expression evaluation_** timeout functionality:  https://www.partow.net/programming/exprtk/code/exprtk_loop_timeout_rtc.cpp

This is an example that demonstrates both _**expression evaluation and compilation**_ timeout functionality:
https://github.com/google/oss-fuzz/blob/master/projects/exprtk/exprtk_fuzzer.cpp

frmts/vrt/vrtexpression.cpp
dbaston: We're limiting the number of elements rather than the memory size. I've renamed the config option to make that less confusing.

third_party/exprtk/exprtk.hpp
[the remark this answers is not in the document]
ArashPartow: @dbaston  I have been able to replicate that sanitizer diagnostic.

Initially, in C++, arithmetic operations between unsigned integer types cannot either overflow or underflow, hence there is no undefined behaviour. In short this particular check is broken, and that is probably why it is not on/enabled by default when enabling UBSAN.

The for the sequence of operations should be:

```
 1. uint - uint     -> uint
 2. uint + int
    2.1 uint + uint -> uint
```


In any case people more familiar than myself seem to think as much: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=106469

I suspect `"sanitize=unsigned-integer-overflow"` is similar to many other checks that can be enabled, where these checks are better off being used periodically on a developer's box than to have them be run on CI continuously.


That being said, the return values in that expression stopped being used a long time ago, the operator calls now register the encountered errors internally in the `token_scanner`.  I've updated the library, where the change is to set all the return values to be zero, a further refactoring may be done down the road to remove the return values altogether. This change remove the diagnostic that is raised. Let me know if this fixes the issue you're seeing.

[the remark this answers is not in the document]
rouault: > Initially, in C++, arithmetic operations between unsigned integer types cannot either overflow or underflow, hence there is no undefined behaviour.

They actually overflow or underflow (ie UINT_MAX + 1 == 0), but that is indeed defined behavior. That said, my experience with the GDAL library is that 75% of the situations where unsigned integer overflow/underflow happened were actual bugs, hence we enable it in our UBSAN continuous integration.  We have a special function decorator ``#define CPL_NOSANITIZE_UNSIGNED_INT_OVERFLOW     __attribute__((no_sanitize("unsigned-integer-overflow")))`` to tag the few functions where unsigned integer overflow/underflow is the desired behavior  (like when doing a horizontal difference filter for integer data)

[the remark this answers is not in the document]
ArashPartow: @rouault  Just to clarify my statement about UBSAN: is that it should be on (used) and the default set of checks are fine. 

The UBSAN checks that are not on by **default**, are probably so because they are problematic in certain contexts giving false positives and require more attention be provided when analysing their results. 

That being said, if defined behaviour (in this case of wrap around post unsigned operations) provides incorrect results, that is most likely a bug in the code. _As an example the wheels on a car rotating (aka behaving correctly) are not the cause of the crash the car might be in._

frmts/vrt/vrtexpression.cpp
dbaston: @rouault I added a `GDAL_EXPRTK_TIMEOUT_SECONDS` configuration option in 8e06984befdba8bd8a3c2354fac39e73a26b6a65 . You could exceed the timeout without being caught if you don't use a loop, but hopefully that possibility is excluded by the limits on expression length (`"GDAL_EXPRTK_MAX_EXPRESSION_LENGTH)` and complexity (not currently exposed as a config option).

@ArashPartow exprtk was reporting an iteration _count_ violation when a timeout occurred; see 1d8819d9ea8830b2826df688c2ac9a19dea3e87d

rouault:
you'll need to rebase your PR on top of latest master and run `scripts/collect_config_options.py`

rouault:
For the msys2-mingw64 "'file too big" error, see https://github.com/qgis/QGIS/blob/ac7a7fc78eadae3682c57387d2863128f6e65901/src/app/CMakeLists.txt#L493

frmts/vrt/vrtexpression.h
```diff
@@ -0,0 +1,45 @@
...
+
+#include "cpl_error.h"
+
+#include <string_view>
+#include <vector>
+
+/**
+ * Class to support evaluation of an expression using the exprtk library.
+ */
+class GDALExpressionEvaluator
```
dbaston, line 23: Maybe instead `gdal::ExprtkEvaluator` or `gdal::vrt::ExprtkExpression` ?

rouault: > `gdal::ExprtkEvaluator`

would be fine to me as there's nothing vrt specific in it as far as I can see and we could potentially moved it upwards in the hierarchy if needed

rouault:
The ASAN failure on test_ogrmerge.py is fixed/worked around in latest master per 671c0ed6aa43843061bbc8fb9b3a9b5a554a51d0

ASAN also reports a unsigned integer overflow (actually an underflow) at
https://github.com/OSGeo/gdal/blob/0f47d71f30c37484d45cdcd55a98219321c367f9/third_party/exprtk/exprtk.hpp#L5131
If a negative value for upper_bound makes sense (@ArashPartow  ?), that line should probably be rewritten as
```
upper_bound(static_cast<int>(vsize) - static_cast<int>(remainder ? loop_batch_size : 0))
```
because even if subtracting a - b where a and b are unsigned and a < b is legal, casting that value which is a value > INT_MAX to int is perhaps not *undefined* behavior, but it is probably implementation defined

rouault:
Could you add in autotest/gdrivers/data/vrt a sample VRT file using an expressionsand not referencing any external raster, so it can be easily used as a seed for fuzzing? (fuzzers/build_seed_corpus.sh has logic to grap all files below autotest/gcore/data and autotest/gdrivers/data and use them as input for the generic gdal_fuzzer)
You may create it by modifying the arraysource_array_constant.vrt sample file added per https://github.com/OSGeo/gdal/pull/11435/files

rouault:
> Like this?

Yes

> Is there some way to tell the fuzzer to make random expressions?

It will figure that itself, by grabbing C strings from libgdal as candidates and randomly mutating until it explores new code paths. Optionally one can help it by providing a dictionary of strings it should try (https://github.com/google/oss-fuzz/blob/master/docs/getting-started/new_project_guide.md#dictionaries), but that's generally not necessary.

ArashPartow:
> The ASAN failure on test_ogrmerge.py is fixed/worked around in latest master per [671c0ed](https://github.com/OSGeo/gdal/commit/671c0ed6aa43843061bbc8fb9b3a9b5a554a51d0)
> 
> ASAN also reports a unsigned integer overflow (actually an underflow) at https://github.com/OSGeo/gdal/blob/0f47d71f30c37484d45cdcd55a98219321c367f9/third_party/exprtk/exprtk.hpp#L5131 If a negative value for upper_bound makes sense (@ArashPartow ?), that line should probably be rewritten as
> 

@rouault  would you be able to show me the specific ASAN error generated?  As I'd like to see if I can reproduce it locally.

rouault:
> @rouault would you be able to show me the specific ASAN error generated? As I'd like to see if I can reproduce it locally.

@ArashPartow  Cf https://github.com/OSGeo/gdal/actions/runs/12142044186/job/33855583979:

```
2024-12-03T15:45:36.5894302Z gdrivers/vrtprocesseddataset.py::test_vrtprocesseddataset_expression[procedural] /home/runner/work/gdal/gdal/third_party/exprtk/exprtk.hpp:5131:50: runtime error: unsigned integer overflow: 10 - 16 cannot be represented in type 'unsigned long'
```
the "procedural" test is at https://github.com/OSGeo/gdal/pull/11209/files#diff-4b00ec67b2717f7cdd201ce1583aeca261f3dc7b8074fc52d9bcc274c322e68eR1225

dbaston:
Unfortunately I've just noticed that this PR increases the size of `libgdal.so` in a release build from 16M to 23M 😳

rouault:
> Unfortunately I've just noticed that this PR increases the size of `libgdal.so` in a release build from 16M to 23M 😳

aouch... I see a mention of exprtk_disable_enhanced_features in https://github.com/ArashPartow/exprtk/blob/7b993904a21639304edd4db261f6e2cdcf6d936b/readme.txt#L5858 that could help but it warns about degraded performance.
exprtk_disable_string_capabilities could perhaps help as we won't need that

dbaston:
I think it's also worth plugging in something like https://github.com/Blake-Madden/tinyexpr-plusplus for comparison. It may be that we want to just handle simpler cases and leave vectors and looping to users willing to enable Python. (I was thinking of a Python follow-on for the VRT processed datasets)

ArashPartow:
> unsigned integer overflow

@rouault  I think that's not an ASAN diagnostic, that is the compiler itself and I think we discussed it before - it's an invalid diagnostic being provided by clang, as unsigned int can't have under or overflow.   https://github.com/OSGeo/gdal/pull/11209#discussion_r1843795186

ArashPartow:
wrt: 
> I think it's also worth plugging in something like https://github.com/Blake-Madden/tinyexpr-plusplus for comparison. It may be ..

Here's a list of such libraries, that may be easier to integrate: 

https://github.com/ArashPartow/math-parser-benchmark-project

rouault:
> I think that's not an ASAN diagnostic,

a UBSAN one more exactly with -fsanitize=undefined-integer-overflow.

> it's an invalid diagnostic being provided by clang, as unsigned int can't have under or overflow

maybe, but the result of 10 - 16 is a large unsigned value, bigger than INT_MAX, and thus casting it to int is not defined by the C++ standard but by the compiler implementations. Most implementations will convert it to the expected negative value due to how complement-to-two representation works but this is not guaranteed by the standard a weird implementation could do something else.

ArashPartow:
> Unfortunately I've just noticed that this PR increases the size of `libgdal.so` in a release build from 16M to 23M 😳

> but it warns about degraded performance. exprtk_disable_string_capabilities could perhaps help as we won't need that

@dbaston /  @rouault   There is a similar discussion here: https://github.com/LMMS/lmms/pull/7571#issuecomment-2486932328

rouault:
Idea: if we don't easily manage to reduce exprtk binary size, we could possibly make it a build-time option, and possibly offer the possibility to build it as a plugin, abusing GDAL driver plugin infrastructure. Not saying this should be done in the scope of this PR. This PR could just make it a build-time option defaulting to OFF.

dbaston:
I've added support for the [muparser](https://github.com/beltoforion/muparser) math library to both derived raster bands and processed datasets. I've also added a "batch_size" argument to the processed datasets, which, combined with some preprocessing of the expression, works around the lack of vector support in muparser for cases like aggregating every N bands:

```
<VRTDataset rasterXSize="1440" rasterYSize="721" subClass="VRTProcessedDataset">  <Input>
<SourceFilename relativeToVRT="0">NETCDF:/home/dan/data/era5/era5_hour_202101.nc:tp</SourceFilename>
  </Input>
  <ProcessingSteps>
    <Step name="expression">
        <Algorithm>Expression</Algorithm>
        <Argument name="dialect">muparser</Argument>
        <Argument name="batch_size">24</Argument>
        <Argument name="expression">sum(BANDS)</Argument>
    </Step>
  </ProcessingSteps>
  <OutputBands count="FROM_LAST_STEP" dataType="Float32" />
</VRTDataset>


gdal_translate sum.vrt out_exprtk.tif
```

Profiling this example, in which 1488 bands are reduced to 62, shows more overhead for `muparser`. But the runtime is essentially the same, 29-30 seconds. Maybe there are workflows where the expression evaluation would be a more significant part of runtime?

exprtk:
```
  52.88%  gdal_translate  libgdal.so.36.3.11.0       [.] (anonymous namespace)::GDALCopyWordsFromT<short>
   5.01%  gdal_translate  libgdal.so.36.3.11.0       [.] exprtk::details::vec_add_op<double>::process
   3.80%  gdal_translate  libz.so.1.2.11             [.] adler32_z
   3.35%  gdal_translate  libc.so.6                  [.] __memmove_avx_unaligned_erms
   3.02%  gdal_translate  libgdal.so.36.3.11.0       [.] ExpressionProcess
   2.88%  gdal_translate  [unknown]                  [k] 0xffffffff9240109c
   2.13%  gdal_translate  [vdso]                     [.] 0x00000000000006e5
   1.44%  gdal_translate  libgdal.so.36.3.11.0       [.] (anonymous namespace)::GDALCopyWordsFromT<double>
   1.40%  gdal_translate  libz.so.1.2.11             [.] 0x0000000000002d3b
   0.96%  gdal_translate  libz.so.1.2.11             [.] 0x00000000000033d8
   0.83%  gdal_translate  libgdal.so.36.3.11.0       [.] gdal::ExprtkExpression::Evaluate
   0.70%  gdal_translate  libz.so.1.2.11             [.] 0x00000000000033ed
   0.64%  gdal_translate  libz.so.1.2.11             [.] 0x00000000000033ff
   0.61%  gdal_translate  libz.so.1.2.11             [.] 0x0000000000002d2f
   0.60%  gdal_translate  libgdal.so.36.3.11.0       [.] std::vector<double, std::allocator<double> >::_M_default_append
   0.59%  gdal_translate  libz.so.1.2.11             [.] 0x0000000000002dbf
   0.57%  gdal_translate  libz.so.1.2.11             [.] 0x00000000000033df
   0.56%  gdal_translate  libz.so.1.2.11             [.] 0x0000000000002d59
   0.53%  gdal_translate  libhdf5_serial.so.103.3.0  [.] 0x00000000002a8a64
   0.53%  gdal_translate  libz.so.1.2.11             [.] 0x00000000000033f6
```

muparser:
```
  51.00%  gdal_translate  libgdal.so.36.3.11.0       [.] (anonymous namespace)::GDALCopyWordsFromT<short>
   9.13%  gdal_translate  libgdal.so.36.3.11.0       [.] mu::ParserBase::ParseCmdCodeBulk
   6.71%  gdal_translate  libgdal.so.36.3.11.0       [.] mu::MathImpl<double>::Sum
   3.65%  gdal_translate  libz.so.1.2.11             [.] adler32_z
   2.75%  gdal_translate  libc.so.6                  [.] __memmove_avx_unaligned_erms
   2.72%  gdal_translate  [unknown]                  [k] 0xffffffff9240109c
   1.59%  gdal_translate  libgdal.so.36.3.11.0       [.] ExpressionProcess
   1.53%  gdal_translate  libgdal.so.36.3.11.0       [.] (anonymous namespace)::GDALCopyWordsFromT<double>
   1.32%  gdal_translate  libz.so.1.2.11             [.] 0x0000000000002d3b
   0.89%  gdal_translate  libz.so.1.2.11             [.] 0x00000000000033d8
   0.76%  gdal_translate  libz.so.1.2.11             [.] 0x00000000000033ed
   0.64%  gdal_translate  libz.so.1.2.11             [.] 0x00000000000033ff
   0.55%  gdal_translate  libz.so.1.2.11             [.] 0x0000000000002d2f
   0.53%  gdal_translate  libz.so.1.2.11             [.] 0x0000000000002dbf
   0.52%  gdal_translate  libz.so.1.2.11             [.] 0x00000000000033df
   0.51%  gdal_translate  libhdf5_serial.so.103.3.0  [.] 0x00000000002a8a64
   0.50%  gdal_translate  libz.so.1.2.11             [.] 0x0000000000002d59
   0.49%  gdal_translate  libgdal.so.36.3.11.0       [.] gdal::MuParserExpression::Evaluate
   0.46%  gdal_translate  libz.so.1.2.11             [.] 0x0000000000002d48
```

frmts/vrt/vrtprocesseddatasetfunctions.cpp
```diff
@@ -1468,6 +1471,259 @@ static CPLErr TrimmingProcess(
...
+        else
+        {
+            return m_adfResults;
+        }
+    }
+
+  private:
+    int BatchCount() const
+    {
+        return (m_nInBands + m_nNominalBatchSize - 1) / m_nNominalBatchSize;
```
rouault, line 1606: - could this be stored in a const variable at construction time to avoid potential recomputation?
- if someone would use an insane large value for m_nNominalBatchSize, the addition could overflow int. You can for example use the DIV_ROUND_UP(m_nInBands, m_nNominalBatchSize) from gcore/gdal_priv.h

frmts/vrt/vrtprocesseddatasetfunctions.cpp
```diff
@@ -1468,6 +1471,259 @@ static CPLErr TrimmingProcess(
...
+    }
+
+    int m_nInBands;
+    int m_nNominalBatchSize;
+    int m_nCurrentBatchSize;
```
rouault, line 1611:
```suggestion
    const int m_nInBands;
    const int m_nNominalBatchSize;
    const int m_nCurrentBatchSize;
```

frmts/vrt/vrtexpression_muparser.cpp
```diff
@@ -0,0 +1,156 @@
...
+    return m_pImpl->Compile();
+}
+
+void MuParserExpression::RegisterVariable(std::string_view osVariable,
+                                          double *pdfValue)
+{
+    m_pImpl->Register(osVariable, pdfValue);
+}
+
+void MuParserExpression::RegisterVector(std::string_view osVariable,
```
rouault, line 121: not that it likely matters much performance wise, but given that you have to std::string(osVariable) at least twice in this method (including one in a loop), just taking a const std::string& as a parameter could be slightly better here

rouault:
> Maybe there are workflows where the expression evaluation would be a more significant part of runtime?

the time spent in GDALCopyWords seems quite large. I'm wondering if it is expected or if there's some inefficiency we could solve. Do you have a link to the dataset you use here so I can have a look?

dbaston:
https://drive.google.com/file/d/1XRc7NDY8dNM4giYMxezSGD1YmEC1kcyY/view?usp=sharing

rouault:
> the time spent in GDALCopyWords seems quite large. I'm wondering if it is expected or if there's some inefficiency we could solve

After investigation, we are probably close to optimal, or at least, I can't spot any low-hanging fruit.
Most of the time is spent acquiring the source data, converting from Int16 to Float64 and re-arranging it in a pixel-interleaved way.
``python3 -c "from osgeo import gdal; ds = gdal.Open('era5_hour_202101.nc'); ds.ReadAsArray(buf_type = gdal.GDT_Float64, interleave = 'pixel')"`` just takes slightly less time than reading the VRT.
If acquiring in a interleave = 'band' way, this cuts the time by 2, but that would require reworking the VRTProcessedDataset logic, and I'm not sure it would help at all since we would need to re-order data anyway in step processing functions.
What I noticed though with this particular dataset is that by default the processing chunk is the whole raster, given the netCDF block size is the raster dimensions, which thus consumes 1440 * 721 * 1488 * sizeof(double) = 11.5 GB to acquire input data, and 1440 * 721 * 62 * sizeof(double) = 0.5 GB for the output data. In https://github.com/OSGeo/gdal/pull/11569, I've implemented a logic to dynamically reduce the block size if it is too large compared to the amount of RAM. Of course when doing so, this increases the processing time, but that's better than crashing the machine. With that particular dataset and the 40% threshold I put, with 32 GB RAM, you can still process the dataset as a single block.

rouault:
> After investigation, we are probably close to optimal, or at least, I can't spot any low-hanging fruit.

actually I now realize I had already implemented a cache-friendly array transposition function for the multidim code. I've generalized it and used it in VRTProcessedDataset and it gives a nice perf boost: https://github.com/OSGeo/gdal/pull/11572

frmts/vrt/vrtprocesseddatasetfunctions.cpp
dbaston: done, thanks

frmts/vrt/vrtexpression_muparser.cpp
dbaston: Excessive copies avoided in 8d977e8b00dfac7cec952d094fa4c7bf27bf3c01 without changing interface

dbaston:
I think this is getting close. A few issues:

- If `exprtk` is optional and non-default, do we want to continue to vendor it (1.6 MB), have CMake download it if the feature is enabled, or rely on the user to provide it? Or do we want to remove this support for the time being if we can't show a clear benefit? I'm inclined not to vendor it but I can be irrational about trying to keep the repo small.

- I'm struggling to resolve the [linking issue](https://github.com/OSGeo/gdal/actions/runs/12659450440/job/35278591984?pr=11209#step:10:1493) in the msys2 build. I have this reproduced locally and will continue flailing at it.

- I'm also not sure how to resolve the [Doxygen warnings](https://github.com/OSGeo/gdal/actions/runs/12659450442/job/35278583755?pr=11209#step:4:4491). I would have expected them to be resolved by adding` GDAL_VRT_ENABLE_MUPARSER` and `GDAL_VRT_ENABLE_EXPRTK` to the `PREDEFINED` block in the `Doxyfile`, but this has no effect locally. The issue may also resolve itself depending on how we handle `exprtk`.

- Still need doc updates to reflect the use of `muparser` as a default but I'll do this after we make a decision on how to handle `exprtk`.

rouault:
> * If `exprtk` is optional and non-default, do we want to continue to vendor it (1.6 MB), have CMake download it if the feature is enabled, or rely on the user to provide it? Or do we want to remove this support for the time being if we can't show a clear benefit? I'm inclined not to vendor it but I can be irrational about trying to keep the repo small.

Also +1 not to vendor it, and rely on CMake FindXXXX() logic (cf cmake/helpers/CheckDependentLibraries.cmake) to detect an external exprtk library

> * I'm struggling to resolve the [linking issue](https://github.com/OSGeo/gdal/actions/runs/12659450440/job/35278591984?pr=11209#step:10:1493) in the msys2 build

Isn't it some issue related to declspec(dllexport) / declspec(dllimport) being inappropriately used  in the context of a vendored library

> * I'm also not sure how to resolve the [Doxygen warnings](https://github.com/OSGeo/gdal/actions/runs/12659450442/job/35278583755?pr=11209#step:4:4491).

as it is about non-public API, you can just wrap the offending files between ``*! @cond Doxygen_Suppress */``  and   ``*! @endcond */``

hobu:
+1 to requiring user-provided exprtk. It is available on conda-forge for our CI, and it looks like it is starting to be provided by some of the other packaging systems. If we vendor it, we will get complaints about it. We lose either way 😄

dbaston:
> @dbaston I've pushed to your branch a commit (cherry-picked from master) that should hopefully fix the mysterious unrelated crash under the ASAN CI config (no idea why it just popped up on your pull request!)

Sorry, I didn't notice this message and apparently clobbered your commit with a force-push. I'll rebase from master again to pick it up.

Is the commit history in this PR important to you? I think the simplest way to remove the vendored exprtk from the history is to squash to a single commit. (Or maybe two commits, with the patch to the vendored muparser kept separate)

rouault:
> Is the commit history in this PR important to you?

Squashing in a single commit is probably going to be the simplest solution . I tend to prefer atomic history in general like "Add function foo()", "use function foo in driver A", "use function foo in driver B", but only if each individual commit builds properly and doesn't break a ``git bisect`` section, and when doing week-long developments it can be a headache to reconstruct a clean history (I'd wish github had a function where we could ask it to run CI on each commit to ensure the history is clean). We're not yet at a position to be as demanding as the Linux kernel project is to its contributors :-)

rouault:
We'll need to document the variables related to EXPRTK and MUPARSER dependencies in https://gdal.org/en/latest/development/building_from_source.html#cmake-package-dependent-options

frmts/vrt/vrtexpression.h
```diff
@@ -0,0 +1,194 @@
...
+#endif
+#if GDAL_VRT_ENABLE_MUPARSER
+    if (EQUAL(pszDialect, "muparser"))
+    {
+        return std::make_unique<gdal::MuParserExpression>(pszExpression);
+    }
+#endif
+    CPLError(CE_Failure, CPLE_IllegalArg, "Unknown expression dialect: %s",
+             pszDialect);
+    return nullptr;
```
rouault, line 189:
```suggestion
MathExpression::Create([[maybe_unused]] const char *pszExpression,
                        const char *pszDialect)
{
    if (EQUAL(pszDialect, "exprtk"))
    {
#if GDAL_VRT_ENABLE_EXPRTK
        return std::make_unique<gdal::ExprtkExpression>(pszExpression);
#else
        CPLError(CE_Failure, CPLE_IllegalArg,
                 "Dialect '%s' is not supported by this GDAL build. A GDAL "
                 "build against exprtk would be needed.", pszDialect);
#endif
    }
    else if (EQUAL(pszDialect, "muparser"))
    {
#if GDAL_VRT_ENABLE_MUPARSER
        return std::make_unique<gdal::MuParserExpression>(pszExpression);
#else
        CPLError(CE_Failure, CPLE_IllegalArg,
                 "Dialect '%s' is not supported by this GDAL build. A GDAL "
                 "build against muparser would be needed.", pszDialect);
#endif
    }
    else
    {
        CPLError(CE_Failure, CPLE_IllegalArg, "Unknown expression dialect: %s",
                 pszDialect);
    }
    return nullptr;
```

rouault:
https://github.com/dbaston/gdal/pull/2 : CI: add muparser to conda windows build

rouault:
https://github.com/dbaston/gdal/pull/3: ExprPixelFunc(): fix memory leak in error code path

rouault:
The PR LGTM. Does @hobu want to give it some additional review ?

hobu:
Is one of EXPRTK or MUPARSER a hard dependency of GDAL going forward?

dbaston:
> Is one of EXPRTK or MUPARSER a hard dependency of GDAL going forward?

They are both optional. CMake is setup to build with muparser if it's available, and to build with ExprTk only if requested (because of the effect on binary size). I would recommend that standard packages (conda, debian, etc.) build with muparser.

hobu:
Is the "commonly expected usage" going to require an expression engine? IMO if that's the case, we should just vendor ExprTk and be done with it.

dbaston:
I think the default would be to have an expression engine, yes. I don't recommend building w/ExprTk by default because it adds ~7MB to a ~16MB library. We could vendor muparser but it seems to be widely available.

cmake/helpers/CheckDependentLibraries.cmake
```diff
@@ -443,6 +443,9 @@ gdal_check_package(IDB "enable ogr_IDB driver" CAN_DISABLE)
 gdal_check_package(rdb "enable RIEGL RDB library" CONFIG CAN_DISABLE)
 include(CheckDependentLibrariesTileDB)
 
+gdal_check_package(ExprTk "Enable C++ Mathematical Expression Tooklit Library (ExprTk) for VRT expressions" DISABLED_BY_DEFAULT)
+gdal_check_package(muparser "Enable muparser library for VRT expressions" CAN_DISABLE)
```
rouault, line 447: to hint packagers (although I suspect they won't pay much attention to that, so we should make some noise in the release notes)
```suggestion
gdal_check_package(muparser "Enable muparser library for VRT expressions" RECOMMENDED CAN_DISABLE)
```

rouault:
Merged!

rouault:
We now need a ``gdal raster calc --input=A:in.tif --input=B:in2.tif --output out.tif --calc "A+B"``  or shorter ``gdal raster calc A:in.tif B:in2.tif out.tif --calc "A+B"``

dbaston:
That's the plan! (https://github.com/OSGeo/gdal/issues/10067)

PR #11213, opened 2024-11-05
OGRLayer::GetArrowStream(): add a DATETIME_AS_STRING=YES/NO option

### LINKED ISSUE #11212: ogr2ogr: for e.g. for GPKG files, naive datetimes in src are represented as being UTC in dst file
### What is the bug?

Datetime columns are not always treated correctly/consistently:
- For GeoPackage, when a file is "translated" to a geopackage file:
    - naive datetimes (no timezone information) are interpreted/written as being UTC to the destination file, which changes the the datetime information significantly. This only happens when the data is internally treated via arrow ([more info](https://gdal.org/en/latest/programs/ogr2ogr.html#known-issues)). With the classic internal treatment (e.g. when explodecollection,... is specified) this does not occur.
    - If the source geopackage has a column with localized timezone (e.g. +4), the datetime in that column is converted to UTC. This is not ideal as the localization information is lost, but the time stays ~correct, so this problem isn't as significant.
- For FlatGeoBuffer the timezone information is always ignored when this is the destination file format.

This issue is already being discussed in the context of the integration in pyogrio via the C API [here](https://github.com/geopandas/pyogrio/issues/487). Just posting it here as well for completeness sake and to document that the same behaviour occurs in ogr2ogr as well when arrow is used internally.

**UPDATE**: added FlatgeoBuffer to the script + with versus without using arrow

``` python
import warnings
from pathlib import Path
import geopandas as gpd
import pandas as pd
from osgeo import gdal
from shapely import Point

warnings.filterwarnings("ignore")
gdal.UseExceptions()

input_gdf = gpd.GeoDataFrame(
    data={
        "datetime_naive": pd.to_datetime(["2021-01-01 00:00:00", "2021-01-01 00:00:00", "2021-01-01 00:00:00"]),
        "datetime_utc": pd.to_datetime(["2021-01-01 00:00:00+00:00", "2021-01-01 00:00:00+00:00", "2021-01-01 00:00:00+00:00"]),
        "datetime_tz_local": pd.to_datetime(["2021-01-01 00:00:00+04:00", "2021-01-01 00:00:00+04:00", "2021-01-01 00:00:00+04:00"]),
    },
    geometry=[Point(0, 0), Point(0, 0), Point(0, 0)],
    crs=31370,
)

for suffix in [".gpkg", ".fgb"]:
    for arrow in ["YES", "NO"]:
        gdal.SetConfigOption("OGR2OGR_USE_ARROW_API", arrow)
        src = Path(f"C:/temp/src_arrow-{arrow}{suffix}")
        src.unlink(missing_ok=True)
        input_gdf.to_file(src)

        dst = Path(f"C:/temp/dst_arrow-{arrow}{suffix}")
        dst.unlink(missing_ok=True)
        ds_output = gdal.VectorTranslate(srcDS=src, destNameOrDestDS=dst)
        ds_output = None

        src_gdf = gpd.read_file(src)
        dst_gdf = gpd.read_file(dst)

        print(f"=== result for {suffix}, {arrow=} ===")
        print(src_gdf.drop(columns=["geometry"]))
        print(dst_gdf.drop(columns=["geometry"]))
```

Output:
```
=== result for .gpkg, arrow='YES' ===
  datetime_naive              datetime_utc         datetime_tz_local
0     2021-01-01 2021-01-01 00:00:00+00:00 2021-01-01 00:00:00+04:00
1     2021-01-01 2021-01-01 00:00:00+00:00 2021-01-01 00:00:00+04:00
2     2021-01-01 2021-01-01 00:00:00+00:00 2021-01-01 00:00:00+04:00
             datetime_naive              datetime_utc         datetime_tz_local
0 2021-01-01 00:00:00+00:00 2021-01-01 00:00:00+00:00 2020-12-31 20:00:00+00:00
1 2021-01-01 00:00:00+00:00 2021-01-01 00:00:00+00:00 2020-12-31 20:00:00+00:00
2 2021-01-01 00:00:00+00:00 2021-01-01 00:00:00+00:00 2020-12-31 20:00:00+00:00
=== result for .gpkg, arrow='NO' ===
  datetime_naive              datetime_utc         datetime_tz_local
0     2021-01-01 2021-01-01 00:00:00+00:00 2021-01-01 00:00:00+04:00
1     2021-01-01 2021-01-01 00:00:00+00:00 2021-01-01 00:00:00+04:00
2     2021-01-01 2021-01-01 00:00:00+00:00 2021-01-01 00:00:00+04:00
  datetime_naive              datetime_utc         datetime_tz_local
0     2021-01-01 2021-01-01 00:00:00+00:00 2021-01-01 00:00:00+04:00
1     2021-01-01 2021-01-01 00:00:00+00:00 2021-01-01 00:00:00+04:00
2     2021-01-01 2021-01-01 00:00:00+00:00 2021-01-01 00:00:00+04:00
=== result for .fgb, arrow='YES' ===
  datetime_naive              datetime_utc         datetime_tz_local
0     2021-01-01 2021-01-01 00:00:00+00:00 2021-01-01 00:00:00+04:00
1     2021-01-01 2021-01-01 00:00:00+00:00 2021-01-01 00:00:00+04:00
2     2021-01-01 2021-01-01 00:00:00+00:00 2021-01-01 00:00:00+04:00
  datetime_naive datetime_utc datetime_tz_local
0     2021-01-01   2021-01-01        2021-01-01
1     2021-01-01   2021-01-01        2021-01-01
2     2021-01-01   2021-01-01        2021-01-01
=== result for .fgb, arrow='NO' ===
  datetime_naive              datetime_utc         datetime_tz_local
0     2021-01-01 2021-01-01 00:00:00+00:00 2021-01-01 00:00:00+04:00
1     2021-01-01 2021-01-01 00:00:00+00:00 2021-01-01 00:00:00+04:00
2     2021-01-01 2021-01-01 00:00:00+00:00 2021-01-01 00:00:00+04:00
  datetime_naive              datetime_utc         datetime_tz_local
0     2021-01-01 2021-01-01 00:00:00+00:00 2021-01-01 00:00:00+04:00
1     2021-01-01 2021-01-01 00:00:00+00:00 2021-01-01 00:00:00+04:00
2     2021-01-01 2021-01-01 00:00:00+00:00 2021-01-01 00:00:00+04:00
```

### Versions and provenance

- OS: Windows 11
- gdal version: 3.9.2, installed from conda-forge

### Additional context

_No response_

rouault:
* GetArrowStream(): add DATETIME_AS_STRING=YES/NO. Defaults to NO. Added in GDAL 3.11. Whether DateTime fields should be returned as a (normally ISO-8601 formatted) string by drivers. The aim is to be able to handle mixed timezones (or timezone naive values) in the same column. All drivers must honour that option, and potentially fallback to the OGRLayer generic implementation if they cannot (which is the case for the Arrow, Parquet and ADBC drivers).
When DATETIME_AS_STRING=YES, the TIMEZONE option is ignored.

Fixes https://github.com/geopandas/pyogrio/issues/487

* OGRLayer::GetArrowStream(): when DATETIME_AS_STRING=YES, expose "GDAL:OGR:type":"DateTime" metadata in the ArrowSchema of DateTime fields

* CreateFieldFromArrowSchema(): take into account GDAL:OGR:Type=DataTime when ArrowSchema.format='u' (string)

* ogr2ogr: GPKG/FlatGeoBuf -> other format: in Arrow code path, use DATETIME_AS_STRING to preserve origin timezone

Fixes #11212

### DIFF
--- a/apps/ogr2ogr_lib.cpp
+++ b/apps/ogr2ogr_lib.cpp
@@ -3997,7 +3997,8 @@ static int GetArrowGeomFieldIndex(const struct ArrowSchema *psLayerSchema,
 /************************************************************************/
 
 static CPLStringList
-BuildGetArrowStreamOptions(const GDALVectorTranslateOptions *psOptions,
+BuildGetArrowStreamOptions(OGRLayer *poSrcLayer, OGRLayer *poDstLayer,
+                           const GDALVectorTranslateOptions *psOptions,
                            bool bPreserveFID)
 {
     CPLStringList aosOptionsGetArrowStream;
@@ -4021,6 +4022,31 @@ BuildGetArrowStreamOptions(const GDALVectorTranslateOptions *psOptions,
             "MAX_FEATURES_IN_BATCH",
             CPLSPrintf("%d", psOptions->nGroupTransactions));
     }
+
+    auto poSrcDS = poSrcLayer->GetDataset();
+    auto poDstDS = poDstLayer->GetDataset();
+    if (poSrcDS && poDstDS)
+    {
+        auto poSrcDriver = poSrcDS->GetDriver();
+        auto poDstDriver = poDstDS->GetDriver();
+
+        const auto IsArrowNativeDriver = [](GDALDriver *poDriver)
+        {
+            return EQUAL(poDriver->GetDescription(), "ARROW") ||
+                   EQUAL(poDriver->GetDescription(), "PARQUET") ||
+                   EQUAL(poDriver->GetDescription(), "ADBC");
+        };
+
+        if (poSrcDriver && poDstDriver && !IsArrowNativeDriver(poSrcDriver) &&
+            !IsArrowNativeDriver(poDstDriver))
+        {
+            // For non-Arrow-native drivers, request DateTime as string, to
+            // allow mix of timezones
+            aosOptionsGetArrowStream.SetNameValue(GAS_OPT_DATETIME_AS_STRING,
+                                                  "YES");
+        }
+    }
+
     return aosOptionsGetArrowStream;
 }
 
@@ -4085,8 +4111,8 @@ bool SetupTargetLayer::CanUseWriteArrowBatch(
             }
         }
 
-        const CPLStringList aosGetArrowStreamOptions(
-            BuildGetArrowStreamOptions(psOptions, bPreserveFID));
+        const CPLStringList aosGetArrowStreamOptions(BuildGetArrowStreamOptions(
+            poSrcLayer, poDstLayer, psOptions, bPreserveFID));
         if (poSrcLayer->GetArrowStream(streamSrc.get(),
                                        aosGetArrowStreamOptions.List()))
         {
--- a/autotest/cpp/CMakeLists.txt
+++ b/autotest/cpp/CMakeLists.txt
@@ -110,7 +110,7 @@ gdal_standard_includes(gdal_unit_test)
 target_compile_options(gdal_unit_test PRIVATE ${GDAL_CXX_WARNING_FLAGS})
 target_compile_definitions(gdal_unit_test PRIVATE -DGDAL_TEST_ROOT_DIR="${GDAL_ROOT_TEST_DIR}")
 target_include_directories(
-  gdal_unit_test PRIVATE $<TARGET_PROPERTY:appslib,SOURCE_DIR> $<TARGET_PROPERTY:gdal_vrt,SOURCE_DIR>)
+  gdal_unit_test PRIVATE $<TARGET_PROPERTY:appslib,SOURCE_DIR> $<TARGET_PROPERTY:gdal_vrt,SOURCE_DIR> $<TARGET_PROPERTY:ogrsf_generic,SOURCE_DIR>)
 if (GDAL_USE_SQLITE3)
   target_compile_definitions(gdal_unit_test PRIVATE -DHAVE_SQLITE3)
   target_include_directories(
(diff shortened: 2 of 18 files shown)


rouault:
@theroggy @jorisvandenbossche  I'm thinking that in this DATETIME_AS_STRING=YES mode, in the ArrowSchema of datetime fields exposed as string (format='u'), we should probably also set the metadata field with a hint for the DateTime semantics. Any suggestion of an appropriate value for it?

jorisvandenbossche:
Thanks a lot for looking into this! 


> we should probably also set the metadata field with a hint for the DateTime semantics. Any suggestion of an appropriate value for it?

Would you just want to indicate that the original GDAL/OGR type was a DateTime? Or is there more information about the column that GDAL can know at that point?  
For the type, maybe something like `"gdal:type": "DateTime"` ? (there is not yet any precedence where you store some information like this is any file format?)

rouault:
> Would you just want to indicate that the original GDAL/OGR type was a DateTime?

actually, I'm just remembering that we have already something. https://gdal.org/en/latest/doxygen/classOGRLayer.html#a3ffa8511632cbb7cff06a908e6668f55 mentions:
```
Starting with GDAL 3.8, the ArrowSchema::metadata field filled by the get_schema() callback may be set with the potential following items:
    "GDAL:OGR:alternative_name": value of OGRFieldDefn::GetAlternativeNameRef()
    "GDAL:OGR:comment": value of OGRFieldDefn::GetComment()
    "GDAL:OGR:default": value of OGRFieldDefn::GetDefault()
    "GDAL:OGR:subtype": value of OGRFieldDefn::GetSubType()
    "GDAL:OGR:width": value of OGRFieldDefn::GetWidth() (serialized as a string)
    "GDAL:OGR:unique": value of OGRFieldDefn::IsUnique() (serialized as "true" or "false")
    "GDAL:OGR:domain_name": value of OGRFieldDefn::GetDomainName()
 ```
 Those are only filled when they cannot be expressed with an Arrow concept.
 So logically that should be extended with "GDAL:OGR:type": "DateTime" in that situation

PR #11230, opened 2024-11-09
Emit a CE_Warning when CPLSet[Thread]ConfigOption() is called with a unknown config option, in debug mode

rouault:
```shell
$ gdalinfo --config BAD_OPTION=TEST --debug on --version
Warning 1: Unknown configuration option 'BAD_OPTION'.
```


doc/source/user/configoptions.rst
```diff
@@ -68,6 +68,20 @@ they can be limited to only the current thread with
 For boolean options, the values YES, TRUE or ON can be used to turn the option on;
 NO, FALSE or OFF to turn it off.
 
+How to detect if the passed configuration option is known of GDAL
```
pjonsson, line 71: Here and a couple of other places in this PR mention "of GDAL", and those should be "to GDAL".

doc/source/user/configoptions.rst
```diff
@@ -68,6 +68,20 @@ they can be limited to only the current thread with
...
+-----------------------------------------------------------------
+
+By default GDAL will not warn if the name of the configuration option is unknown
+of it. Starting with GDAL 3.11, if you set the :config:`CPL_DEBUG` configuration
```
pjonsson, line 75: Can remove the words "of it".

doc/source/user/configoptions.rst
```diff
@@ -68,6 +68,20 @@ they can be limited to only the current thread with
...
+option to ``ON`` (or any value that is not ``OFF``, ``FALSE``, ``NO``), a GDAL
+warning will be emitted for unknown configuration options.
+
+.. code-block:: shell
+
+    $ gdalinfo --config BAD_OPTION=TEST --debug on --version
+    Warning 1: CPLSetConfigOption() called with key=BAD_OPTION, which is unknown of GDAL
```
pjonsson, line 82: I think a "user facing" warning for a command line program would normally say something like `Unknown option 'BAD_OPTION'`. I guess this might be hard to change since there is also giving warnings when using the GDAL API to worry about.

doc/source/user/configoptions.rst
rouault: thanks, fixed

doc/source/user/configoptions.rst
rouault: thanks, fixed

doc/source/user/configoptions.rst
rouault: changed to more neutral "Unknown configuration option '%s'."

port/cpl_conv.cpp
```diff
@@ -1921,13 +2018,18 @@ static void NotifyOtherComponentsConfigOptionChanged(const char *pszKey,
  * value provided during the last call will be used.
  *
  * Options can also be passed on the command line of most GDAL utilities
- * with the with '--config KEY VALUE'. For example,
- * ogrinfo --config CPL_DEBUG ON ~/data/test/point.shp
+ * with the with '--config KEY VALUE' (or '--config KEY=VALUE' since GDAL 3.10).
+ * For example, ogrinfo --config CPL_DEBUG ON ~/data/test/point.shp
```
jratike80, line 2022: with the with?

rouault: fixed, thanks

dbaston:
I don't know if this is a good idea, just tossing something out there. What if the CLI populates a global set of config options that have been specified. `CPLGetConfigOption` removes items from that set. On program exit, we check to see if the set was empty, and issue a warning with the names of config options that were specified but never used. This would avoid the need to maintain a header of known options, and also would produce a warning if a user specified an option that is known to GDAL but not applicable to their particular usage.

rouault:
> On program exit, we check to see if the set was empty, and issue a warning with the names of config options that were specified but never used.

There's actually code that essentially does that, currently protected by if #DEBUG_CONFIG_OPTIONS, but at the time I wrote it, I was nervous about the performance implications of enabling that in production builds at each CPLGetConfigOption() access. Perhaps that's a not founded worry, since calling CPLGetConfigOption() has already performance implications, and shouldn't be done in performance sensitive loops.
But more fundamentally users specifying config options that are not going to be read by the code can be OK in some use cases, for example when defining them in the GDAL configuration file (where they can apply to different type of workflows).

scripts/collect_config_options.py
```diff
@@ -0,0 +1,159 @@
...
+            explore(filename)
+        if filename.endswith(".c") or filename.endswith(".cpp"):
+            collect_config_options(filename)
+
+
+for dirname in ("alg", "port", "gcore", "frmts", "ogr", "apps", "swig"):
+    explore(os.path.join(os.path.dirname(__file__), "..", dirname))
+
+c_file = "// clang-format off\n"
+c_file += "constexpr static const char* const apszKnownConfigOptions[] =\n"
```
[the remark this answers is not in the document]
rouault: thanks, integrated

dbaston:
If the behavior is opt-in (`CPL_DEBUG=ON`) then it would be useful to know that a config option you specified was not valid in the context you tried to use it in. If that spits out warnings for various tokens etc. that are stored in a configuration file, I think that's OK. Probably out-of-scope here, though.

jratike80:
> That said there's a blind spot with this PR regarding config options specified as environment variables.

But the options written into the  `[configoptions]` section of the configuration file are checked, right?

rouault:
> But the options written into the `[configoptions]` section of the configuration file are checked, right?

with this PR, yes they will be checked, as they are set through CPLSetConfigOption(): https://github.com/OSGeo/gdal/blob/d9b65d7609e62d89744b6b1cbdf7e35d4163e212/port/cpl_conv.cpp#L2302

agiudiceandrea:
Hi @rouault, it looks like various config options are missing in the list, e.g. `GDAL_PDF_DPI`: trying to set it using `CPLSetConfigOption` (when DEBUG is ON) generates the following errors:
```
CPLError: Unknown configuration option 'GDAL_PDF_DPI'.
Warning 1: Unknown configuration option 'GDAL_PDF_DPI'.
```
Is this intended?

rouault:
> Is this intended?

nope, thanks for raising this. fixed per https://github.com/OSGeo/gdal/commit/a5f476eb97442b5a19ffa5d01e08331ee8e5ecb1

PR #11234, opened 2024-11-10
LVBAG: only run IsValid() if bFixInvalidData

rouault:
This will speed up processing if not needing to fix invalid data. May be related to https://lists.osgeo.org/pipermail/gdal-dev/2024-November/059794.html

CC @yorickdewid


yorickdewid:
Agreed

Issue #10809, opened 2024-09-16
perf: gdalwarp 3.4.3 takes seconds, gdalwarp 3.5.0 and later takes 30+ minutes


pjonsson:
### What is the bug?

The command
```shell
 gdalwarp -q -geoloc -r cubic -of COG -co COMPRESS=DEFLATE -co NUM_THREADS=1 --config GDAL_NUM_THREADS 1 /tmp/tmp_xnrpustp/oa02.vrt /tmp/output_a2odmztn/oa02.tif
 ```
 took seconds with GDAL 3.4.3, but takes 38+ CPU minutes with GDAL 3.9.2 when using the GDAL ubuntu-small Docker image. I don't have exact numbers for GDAL 3.5.x, but that was also measured in minutes rather than seconds. Around 15 months ago I bisected GDAL, and landed on #5520 where the performance changed. The PR mentions 
"goes from 1.26 s to 3.52 s" when testing on a Sentinel 5P product, so I understand some slowdown was expected.

Performance numbers are on a Ryzen 5900x boosting to 4.55-4.65 GHz. System is not overheating, the fans are not audibly louder than when the system is idle. IO is to a fast SSD, but there is no IO on the system when the CPU is going at full speed.

The commands have the environment variable `GDAL_CACHEMAX=256`, and that value is tuned for the workload using GDAL 3.4.3.

The example is using files from the product `S3B_OL_2_WFR____20230309T115352_20230309T115652_20230310T233800_0180_077_066_1800_MAR_O_NT_003.SEN3`, but from what I have observed there is nothing unique with the mentioned product, the GDAL performance is similar for other OL_2_WFR___ products.

### Steps to reproduce the issue

The VRT is created with `gdal_translate -q -ot Float32 -unscale NETCDF:Oa02_reflectance.nc:Oa02_reflectance -co NUM_THREADS=1 --config GDAL_NUM_THREADS 1 /tmp/tmp_xnrpustp/oa02.vrt`, and then modified based on https://gis.stackexchange.com/questions/103116/map-project-a-raster-having-separate-latitude-and-longitude-raster-bands.

The people who initially wrote the code that constructed the GDAL commands are gone, so I don't have any more information than that. Attaching the modified VRT, and the latitude/longitude VRTs that it uses.

[vrts.zip](https://github.com/user-attachments/files/17009974/vrts.zip)
### Versions and provenance

[...]

PR #11235, opened 2024-11-10
docker: enable IPO

pjonsson:
This reduces the size of the small Ubuntu
image by 4 MB. There also seems to be
a slight speedup, the user time for the example
from #10809 goes from 34.x seconds to
33.x seconds on my computer.


rouault:
My memories of IPO (at least on PROJ I believe) was that it leads to very long compilation times, in particular for the --release builds doing emulated compilation for arm64. Could you conditionalize that to the x86_64 arch ?

pjonsson:
Ok, updated to only enable it for x86-64.

PR #11236, opened 2024-11-10
autotest: update internal googletest to 1.15.2 to avoid CMake warnings

rouault:


rouault:
backported to 3.10 per a676daeea55cc792e89ee55d5451dd9a6763a777

PR #11252, opened 2024-11-12
Doc: Python: move files around and add executable examples using myst-nb

rouault:
[myst-nb](https://myst-nb.readthedocs.io/en/latest/index.html) is one of the components used by [Jupyter Books](https://jupyterbook.org/en/stable/explain/sphinx.html). 
Inspiration from [PDAL](https://github.com/PDAL/PDAL/blob/master/doc/development/documentation-quickstart.md), but I didn't want to go all-in by converting all our doc to Jupyter Book / MyST markdown. Restrict that for now to pages that need notebooks.

==> [Rendered output](https://gdal--11252.org.readthedocs.build/en/11252/api/python/python_examples.html)

To speed-up development, a notebook page can be rendered to HTML with ``mystnb-docutils-html5 --nb-read-as-md=true source/api/python/python_examples.myst > out.html`` . Much faster than doing a full "make html" build, but the output is not fully conformant to the neat output of "make html". That said, code is executed, which is the most important thing when doing trials&errors


doc/source/api/python/python_examples.myst
```diff
@@ -0,0 +1,99 @@
...
+a GDAL raster dataset and its bands.
+
+```{code-cell}
+
+from osgeo import gdal
+import json
+
+gdal.UseExceptions()
+
+with gdal.Open("data/byte.tif") as ds:
```
dbaston, line 27: I would try to limit the cells to 6-10 lines (excluding imports etc.) so that outputs are more easily related to the code that produced them.

dbaston (approved):

doc/source/api/python/python_examples.myst
rouault: I couldn't find a way to split a cell in several parts, with continuation, especially with the use of the ``with`` syntax

dbaston: The `with` block can't be split between cells, but otherwise each cell starts picks up with the state of the previous cell.

image

rouault: split as suggested, and added a paragraph in dev_documentation.rst about it

Issue #11253, opened 2024-11-12
Crash in GDALCopyWords64 when switching to CPL_CPU_REQUIRES_ALIGNED_ACCESS


schwehr:
### What is the bug?

I hit trouble going to 8f2f8efd89db360811f4aab8c7e6a67b307cc12c, so I'm backing up and setting CPL_CPU_REQUIRES_ALIGNED_ACCESS. I think the crash happens in here: https://github.com/OSGeo/gdal/blob/566204a0969479efccbe7671ba7a40e0375a5ead/gcore/rasterio.cpp#L3326-L3332

```c++
            for (decltype(nWordCount) i = 0; i < nWordCount; i++)
            {
                memcpy(static_cast<GByte *>(pDstData) + nDstPixelStride * i,
                       static_cast<const GByte *>(pSrcData) +
                           nSrcPixelStride * i,
                       nDstDataTypeSize);
            }
```

I get a sandbox crash in GDALCopyWords64 when I enable CPL_CPU_REQUIRES_ALIGNED_ACCESS in a test that tries to read this grib. And it works fine from gdalinfo (with both asan and msan builds).

```
gdalinfo gfs_193.20230720.i0000.f168_nodata.grb2 -stats -mm | grep -v GRIB
Files: gfs_193.20230720.i0000.f168_nodata.grb2
Size is 1440, 721
Coordinate System is:
    DATUM["unnamed",
        ELLIPSOID["Sphere",6371229,0,
            LENGTHUNIT["metre",1,
                ID["EPSG",9001]]]],
    PRIMEM["Greenwich",0,
        ANGLEUNIT["degree",0.0174532925199433,
            ID["EPSG",9122]]],
    CS[ellipsoidal,2],
        AXIS["latitude",north,
            ORDER[1],
            ANGLEUNIT["degree",0.0174532925199433,
                ID["EPSG",9122]]],
        AXIS["longitude",east,
            ORDER[2],
            ANGLEUNIT["degree",0.0174532925199433,
                ID["EPSG",9122]]]]
Data axis to CRS axis mapping: 2,1
Origin = (-180.125000000000000,90.125000000000000)
Pixel Size = (0.250000000000000,-0.250000000000000)
Corner Coordinates:
Upper Left  (-180.1250000,  90.1250000) (180d 7'30.00"W, 90d 7'30.00"N)
Lower Left  (-180.1250000, -90.1250000) (180d 7'30.00"W, 90d 7'30.00"S)
[...]

PR #11254, opened 2024-11-12
GDALCopyWords64(): tidy code

rouault:
Technically it is safer to cast a pointer to uintptr_t than our custom type. At least, it better documents the intent. In practice, I doubt this changes much.

CC @schwehr  w.r.t #11253 . Very low chance that it fixes your issue


schwehr:
Looks like nice cleanup.

PR #11270, opened 2024-11-13
OpenJPEG: add a TPSOT_COMPLIANT open option that defaults to YES

rouault:
Fixes the issue reported in https://lists.osgeo.org/pipermail/gdal-dev/2024-November/059805.html Requires openjpeg >= https://github.com/uclouvain/openjpeg/pull/1560 (likely 2.5.3 when it is released)


rouault:
@kidanger I've just ammended this PR to actually fold that new TPSOT_COMPLIANT open into the existing STRICT one. So the default STRICT=YES open option will disable the TPSOT fixing heuristics

rouault:
Dropping this PR as it is no longer needed, since I've actually tuned the openjpeg one to also disable the TPSOT fixing logic, when opj_decoder_set_strict_mode(, true) is called.

PR #11271, opened 2024-11-13
Doc: Make CLI examples referenceable

dbaston:
## What does this PR do?

Tests out a pattern for enabling cross-references to specific code snippets. This makes it easy to jump from the description of a CLI argument to an example that uses it.

Sphinx allows a code block to have a name attribute, but to enable cross-referencing it must also have a caption. So I took the description of each snippet (previously written in plain text) and moved it into the caption. The default style for a caption looks a bit awkward:

image

so this PR overrides the CSS so that the "caption" is more of a title:

image

If the pattern looks OK, it can be applied to documentation for the other CLI tools.


rouault:
LGTM. Maybe describe that in https://gdal.org/en/latest/development/dev_documentation.html#sphinx-rst-style-guide ?

dbaston:
Going to spend a bit more time on this. It doesn't work well for multi-command examples. Maybe a custom directive is the way to go, if it can be done without too much code.

dbaston:
Ended up making a little extension. I think the cross-referencing works fairly well.

image

doc/source/development/dev_documentation.rst
```diff
@@ -341,6 +341,46 @@ To reference a method or function::
...
+
+To document a command-line tool, use the ``example`` directive.
+An example may have a title as well as an id that can be used to generate
+a cross-reference from the same document.
+
+.. code-block:: rst
+
+    .. example::
+       :title: Listing files
+       :id: `basic-ls`
```
rouault, line 356: From the actual documentation pages, it doesn't look those backwards single quote are needed:
```suggestion
       :id: basic-ls
```

rouault (approved):
great addition!

PR #11283, opened 2024-11-15
GPKG: make CreateCopy() work on vector datasets

### LINKED ISSUE #11282: Change in GPKG driver CopyDataSource behavior in 3.10.0
### What is the bug?

The following usage of the geopackage driver `CopyDataSource` worked in 3.9.3 but does not work in 3.10.0. I'm trying to copy a geopackage vector, but `CreateCopy` seems to always expect a raster dataset.

I looked at the 3.10.0 changelog but didn't see anything that would necessarily explain this change, so I thought it might be unexpected. Thanks for taking a look!

```python
from osgeo import ogr

ogr.UseExceptions()

driver = ogr.GetDriverByName('GPKG')
vector = driver.CreateDataSource('foo.gpkg')
layer = vector.CreateLayer('foo', geom_type=ogr.wkbPoint)
feature = ogr.Feature(layer.GetLayerDefn())
geometry = ogr.CreateGeometryFromWkt('POINT (0 0)')
feature.SetGeometry(geometry)
layer.CreateFeature(feature)

source_vector = ogr.Open('foo.gpkg')
driver.CopyDataSource(source_vector, 'copy.gpkg')
```

```python
<class 'osgeo.gdal.Dataset'>
  File "C:\Users\dmf\projects\repro\gdal_gpkg\gpgk_vector_copy.py", line 21, in <module>
    driver.CopyDataSource(source_vector, 'copy.gpkg')
  File "C:\Users\dmf\projects\invest\env-gdal310\Lib\site-packages\osgeo\gdal.py", line 3992, in CopyDataSource
    return self.CreateCopy(utf8_path, ds, options = options or [])
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\dmf\projects\invest\env-gdal310\Lib\site-packages\osgeo\gdal.py", line 3892, in CreateCopy
    return _gdal.Driver_CreateCopy(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: Only 1 (Grey/ColorTable), 2 (Grey+Alpha), 3 (RGB) or 4 (RGBA) band dataset supported
```

### Versions and provenance

Windows 10
`GDAL 3.10.0, released 2024/11/01` raises the error
`GDAL 3.9.3, released 2024/10/07` does not

### Additional context

_No response_

rouault:
Fixes #11282, which is due to a9745e5f8df that mapped CopyDataSource() onto CreateCopy(), but the GPKG driver implementation of CreateCopy() wasn't ready to deal with vector dataset.

### DIFF
--- a/ogr/ogrsf_frmts/gpkg/ogrgeopackagedatasource.cpp
+++ b/ogr/ogrsf_frmts/gpkg/ogrgeopackagedatasource.cpp
@@ -6005,6 +6005,21 @@ GDALDataset *GDALGeoPackageDataset::CreateCopy(const char *pszFilename,
                                                GDALProgressFunc pfnProgress,
                                                void *pProgressData)
 {
+    const int nBands = poSrcDS->GetRasterCount();
+    if (nBands == 0)
+    {
+        GDALDataset *poDS = nullptr;
+        GDALDriver *poThisDriver =
+            GDALDriver::FromHandle(GDALGetDriverByName("GPKG"));
+        if (poThisDriver != nullptr)
+        {
+            poDS = poThisDriver->DefaultCreateCopy(pszFilename, poSrcDS,
+                                                   bStrict, papszOptions,
+                                                   pfnProgress, pProgressData);
+        }
+        return poDS;
+    }
+
     const char *pszTilingScheme =
         CSLFetchNameValueDef(papszOptions, "TILING_SCHEME", "CUSTOM");
 
@@ -6018,7 +6033,6 @@ GDALDataset *GDALGeoPackageDataset::CreateCopy(const char *pszFilename,
         apszUpdatedOptions.SetNameValue("RASTER_TABLE", osBasename);
     }
 
-    const int nBands = poSrcDS->GetRasterCount();
     if (nBands != 1 && nBands != 2 && nBands != 3 && nBands != 4)
     {
         CPLError(CE_Failure, CPLE_NotSupported,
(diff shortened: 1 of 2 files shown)


ogr/ogrsf_frmts/gpkg/ogrgeopackagedatasource.cpp
```diff
@@ -6005,6 +6005,21 @@ GDALDataset *GDALGeoPackageDataset::CreateCopy(const char *pszFilename,
                                                GDALProgressFunc pfnProgress,
                                                void *pProgressData)
 {
+    const int nBands = poSrcDS->GetRasterCount();
+    if (nBands == 0)
+    {
+        GDALDataset *poDS = nullptr;
+        GDALDriver *poThisDriver =
+            GDALDriver::FromHandle(GDALGetDriverByName("GPKG"));
+        if (poThisDriver != nullptr)
```
elpaso, line 6014: can this actually happen?

elpaso (approved):

ogr/ogrsf_frmts/gpkg/ogrgeopackagedatasource.cpp
rouault: Really unlikely. That would be an equivalent of the [TOCTOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) pattern for GDAL, where one thread would CreateCopy() whereas another one would deregister the GPKG driver...

PR #11287, opened 2024-11-16
CMake: use add_compile_options() instead of setting CMAKE_CXX_FLAGS for -fno-finite-math-only

### LINKED ISSUE #11286: clang++-20 does  warns about -fno-finite-math-only when linking
### What is the bug?

When linking, clang++-20 warns if option -fno-finite-math-only is used.

> clang++-20: warning: argument unused during compilation: '-fno-finite-math-only' [-Wunused-command-line-argument]

This message appears nearly 70 times when building gdal with this compiler, each instance appears to be linking.
Since this is only a warning and the build completes this is probably not a high priority.

-fno-finite-math-only is set in
  cmake/helpers/GdalCompilationFlags.cmake
which appears to have replaced parts of gdal.cmake in v3.10. The flag was originally set for clang in dfdd28cc7e64baf4e94028221247dc9e4b47bc34

```
echo "int main() { return 0 ; } " > finite.cpp ; 
echo compiling ; clang++-20 -c -o finite.o finite.cpp ; echo linking ; clang++-20 -o finite finite.o -fno-finite-math-only
```
> compiling
> linking
> clang++-20: warning: argument unused during compilation: '-fno-finite-math-only' [-Wunused-command-line-argument]

### Versions and provenance

Linux Ubuntu 24-10 Oracular with llvm-20, clang-20 and related packages from https://apt.llvm.org/oracular llvm-toolchain-oracular/main 

### Additional context

Splitting compile and build flags may be a non-trivial  task;
since this is only a warning and the build completes this is probably not a high priority.

rouault:
Fixes #11286

@andrew-aitchison Can you confirm it fixes your issue?

### DIFF
--- a/cmake/helpers/GdalCompilationFlags.cmake
+++ b/cmake/helpers/GdalCompilationFlags.cmake
@@ -159,7 +159,7 @@ if (CMAKE_CXX_COMPILER_ID STREQUAL "IntelLLVM" OR CMAKE_CXX_COMPILER_ID STREQUAL
   check_cxx_compiler_flag(-fno-finite-math-only HAVE_FLAG_NO_FINITE_MATH_ONLY)
   if (HAVE_FLAG_NO_FINITE_MATH_ONLY)
     # Intel CXX compiler based on clang defaults to -ffinite-math-only, which breaks std::isinf(), std::isnan(), etc.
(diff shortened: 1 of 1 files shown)


andrew-aitchison:
Confirm issue fixed

PR #11304, opened 2024-11-20
DXF writer: do not set 0 as the value for DXF code 5 (HANDLE)

### LINKED ISSUE #11299: Failure to open converted DXF file in AutoCAD after converting from GeoJSON using GDAL.Native 3.9.1 while ogr2ogr converted file opens fine
### What is the bug?

I'm encountering an issue when trying to convert a GeoJSON file to DXF format. I'm using GDAL.Native version 3.9.1 for the conversion. The conversion process itself appears to be successful as it generates a DXF file without any obvious errors during the conversion stage.

However, when I attempt to open the resulting DXF file in AutoCAD, it fails to load. Oddly enough, when I use ogr2ogr (also in version 3.9.1) to perform the same conversion from GeoJSON to DXF, the generated DXF file can be opened in AutoCAD without any issues.

Here is the example geojson content:
{"type":"FeatureCollection","features":[{"type":"Feature","id":0,"properties":{"SubClasses":"AcDbEntity : AcDbLwPolyline","Layer":"mylayer","EntityHandle":"12E0"},"geometry":{"type":"Polygon","coordinates":[[[1,1],[1,2],[2,2],[2,1],[1,1]]]}}]}

C# codes:
```
GdalConfiguration.ConfigureOgr();
using var iSrc = Ogr.Open("data.json", (int)Access.GA_ReadOnly);
if (iSrc == null) return;

using var driver = Ogr.GetDriverByName("DXF");
using var oSrc = driver.CreateDataSource("data-dotnet.dxf", null);
for (var idx = 0; idx < iSrc.GetLayerCount(); idx++)
{
    using var iLyr = iSrc.GetLayerByIndex(idx);
    if (iLyr == null) continue;

    var layerName = iLyr.GetName();
    using var oLyr = oSrc.CopyLayer(iLyr, layerName, null);
}
```

ogr2ogr command:
`ogr2ogr -f "DXF" data-ogr2ogr.dxf data.json`

### Versions and provenance

Windows 11
Gdal.Native 3.9.1
Gdal binary 3.9.1

### Additional context

[geojson-dxf.zip](https://github.com/user-attachments/files/17811750/geojson-dxf.zip)

data.json
data-dotnet.dxf
data-ogr2ogr.dxf

rouault:
as it is apparently a reserved value.

Fixes #11299

### DIFF
--- a/autotest/ogr/ogr_dxf.py
+++ b/autotest/ogr/ogr_dxf.py
@@ -436,20 +436,28 @@ def test_ogr_dxf_12(tmp_path):
 
     dst_feat = ogr.Feature(feature_def=lyr.GetLayerDefn())
     dst_feat.SetGeometryDirectly(ogr.CreateGeometryFromWkt("LINESTRING(10 12, 60 65)"))
+    dst_feat.SetFID(0)
     lyr.CreateFeature(dst_feat)
+    # 80 is the minimum handle value we set in case of inapproriate or unsed
+    # initial FID value
+    assert dst_feat.GetFID() >= 80
     dst_feat = None
 
     dst_feat = ogr.Feature(feature_def=lyr.GetLayerDefn())
     dst_feat.SetGeometryDirectly(
         ogr.CreateGeometryFromWkt("POLYGON((0 0,100 0,100 100,0 0))")
     )
+    dst_feat.SetFID(79)
     lyr.CreateFeature(dst_feat)
+    assert dst_feat.GetFID() == 79
     dst_feat = None
 
     # Test 25D linestring with constant Z (#5210)
     dst_feat = ogr.Feature(feature_def=lyr.GetLayerDefn())
     dst_feat.SetGeometryDirectly(ogr.CreateGeometryFromWkt("LINESTRING(1 2 10,3 4 10)"))
+    dst_feat.SetFID(79)
     lyr.CreateFeature(dst_feat)
+    assert dst_feat.GetFID() > 79
     dst_feat = None
 
     # Test 25D linestring with different Z (#5210)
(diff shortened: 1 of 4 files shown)


atlight:
> Fixes #11303

That would be #11299 .

Honestly the code at https://github.com/OSGeo/gdal/blob/1b32bc87538a116bb7d4d6ccef1b9831fb04acb8/ogr/ogrsf_frmts/dxf/ogrdxfwriterlayer.cpp#L173 just looks wrong. There is no such thing as an "entity ID"; DXF files have an "entity handle". The value of the `EntityHandle` field, if set, should be used as the entity handle, otherwise we should just be making one up. Using the FID for the entity handle is nonsense and should never have been done.

rouault:
> That would be #11299 .

corrected, thanks

> The value of the EntityHandle field, if set, should be used as the entity handle, otherwise we should just be making one up. Using the FID for the entity handle is nonsense and should never have been done.

yes, I agree that looks dubious. But perhaps as a temporary measure (especially for the sake of a 3.10 backport), we can live with that in the mean time ? I can create an issue to remember we should do something along the line you mention.

atlight (approved):

atlight:
> But perhaps as a temporary measure (especially for the sake of a 3.10 backport), we can live with that in the mean time ? I can create an issue to remember we should do something along the line you mention.

Yes, if this fixes the user's immediate issue I am not opposed to it going in, but it is certainly something that needs addressing properly down the track. 

Incidentally, it would help me if you could briefly explain why ogr2ogr doesn't appear to hit this code path? It seems to always write an entity handle of `20000` no matter what `id` value I set in the user's GeoJSON minimal example.

rouault:
> if you could briefly explain why ogr2ogr doesn't appear to hit this code path?

ogr2ogr by default doesn't set the FID of the new feature from the FID of the source feature, when the target driver has no layer FID creation option. It lets its to OGRUnsetFID == -1.
Whereas GDALDataset::CopyLayer() always set the FID of the new feature from the source one.

PR #11314, opened 2024-11-20
CMake: export -DGDAL_DEBUG as PUBLIC for debug builds

### LINKED ISSUE #11311: Missing DEBUG macro
### What is the bug?

Currently we have issues switching from an old GDAL version (3.3.3 - nmake build systems) to a new one (3.5+ - cmake build system) as the linker fails in debug mode. Specifically, we have the linker problem with `OGRSpatialReference::FindMatches`.

With vcpk, on the other hand, we don't have the linker issue. That's why I used `dumpbin` to inspect the `gdald.lib`:
- Result of `dumpbin` with GDAL compiled via vcpkg:
`?FindMatches@OGRSpatialReference@@QEBAPEAPEAUOGRSpatialReferenceHS@@PEAPEADPEAHPEAPEAH@Z`
(`public: struct OGRSpatialReferenceHS * __ptr64 * __ptr64 __cdecl OGRSpatialReference::FindMatches(char * __ptr64 * __ptr64,int * __ptr64,int * __ptr64 * __ptr64)const __ptr64`)
- Result of `dumpbin` with self compiled GDAL:
`?FindMatches@OGRSpatialReference@@QEBAPEAPEAXPEAPEADPEAHPEAPEAH@Z`
(`public: void * __ptr64 * __ptr64 __cdecl OGRSpatialReference::FindMatches(char * __ptr64 * __ptr64,int * __ptr64,int * __ptr64 * __ptr64)const __ptr64`)

As you can see, the return type differs. The reason for this is this:
```cpp
#ifdef DEBUG
typedef struct OGRSpatialReferenceHS *OGRSpatialReferenceH;
typedef struct OGRCoordinateTransformationHS *OGRCoordinateTransformationH;
#else
/** Opaque type for a Spatial Reference object */
typedef void *OGRSpatialReferenceH;
/** Opaque type for a coordinate transformation object */
typedef void *OGRCoordinateTransformationH;
#endif
```

For some reason `DEBUG` is defined during building self compiled GDAL, but during compiling via vcpkg. Unfortunately, I have not yet found where this macro is defined, as it is not actually a predefined one (`_DEBUG` is predefined, but not `DEBUG`, see [here](https://learn.microsoft.com/en-us/cpp/preprocessor/predefined-macros)). If you create a simple CMake project that has only one executable, set the debug mode in MSVC, and copy in the above `#ifdef` from GDAL, you will also see that `DEBUG` is not defined this case.

So the question is: Where is `DEBUG` defined? And shouldn't it be exported so that it is active for the GDAL headers when it is included?

1) Build GDAL with default settings
2) Create a simple CMake project with just a executable

### Versions and provenance

Windows 10/11

### Additional context

_No response_

rouault:
Fixes #11311

@SunBlack can you give this a try?

### DIFF
--- a/gdal.cmake
+++ b/gdal.cmake
@@ -196,6 +196,10 @@ if (MINGW AND BUILD_SHARED_LIBS)
     set_target_properties(${GDAL_LIB_TARGET_NAME} PROPERTIES SUFFIX "-${GDAL_SOVERSION}${CMAKE_SHARED_LIBRARY_SUFFIX}")
 endif ()
 
+# Some of the types in our public headers are dependent on whether GDAL_DEBUG
+# is defined or not
+target_compile_definitions(${GDAL_LIB_TARGET_NAME} PUBLIC $<$<CONFIG:DEBUG>:GDAL_DEBUG>)
+
 # Install properties
 if (GDAL_ENABLE_MACOSX_FRAMEWORK)
   set(FRAMEWORK_VERSION ${GDAL_VERSION_MAJOR}.${GDAL_VERSION_MINOR})
--- a/autotest/postinstall/test_cpp/test_cpp.cpp
+++ b/autotest/postinstall/test_cpp/test_cpp.cpp
@@ -9,6 +9,7 @@
 #endif
 #include <cpl_vsi_virtual.h>
 #include <ogr_geometry.h>
+#include <ogr_spatialref.h>
 
 int main(int argc, char **argv)
 {
@@ -16,5 +17,13 @@ int main(int argc, char **argv)
     OGRGeometryFactory::createFromWkt("POINT(1 2)", nullptr, &poGeom);
     OGRGeometryFactory::destroyGeometry(poGeom);
     std::cout << GDALVersionInfo("RELEASE_NAME") << std::endl;
+
+    // Check fix for https://github.com/OSGeo/gdal/issues/11311
+    OGRSpatialReference oSRS;
+    int nEntries = 0;
+    int *panMatchCondidence = nullptr;
+    oSRS.FindMatches(nullptr, &nEntries, &panMatchCondidence);
+    CPLFree(panMatchCondidence);
+
     return (0);
 }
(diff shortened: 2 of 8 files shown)


gdal.cmake
```diff
@@ -196,6 +196,10 @@ if (MINGW AND BUILD_SHARED_LIBS)
     set_target_properties(${GDAL_LIB_TARGET_NAME} PROPERTIES SUFFIX "-${GDAL_SOVERSION}${CMAKE_SHARED_LIBRARY_SUFFIX}")
 endif ()
 
+# To also export for external users what we did with add_compile_definitions($<$<CONFIG:DEBUG>:DEBUG>)
+# in cmake/helpers/GdalCompilationFlags.cmake
+target_compile_definitions(${GDAL_LIB_TARGET_NAME} PUBLIC $<$<CONFIG:DEBUG>:DEBUG>)
```
dg0yt, line 201: I'm not sure if this a good idea. At least as long as the macro is not prefixed with `GDAL_` or similar.

rouault: well, DEBUG is used in public GDAL include files, so ...

dg0yt: This really needs a prefix...

SunBlack: I think @dg0yt meant to say: `DEBUG` is such a generic name that you can't be sure that another library isn't using it or being influenced by it in some way. So setting it now could have an effect on other libraries. But if the `DEBUG` flag is set for the debug build in every case: Why not switch to `_DEBUG`, which is what the compilers set?

rouault: ok, I've revised the PR to export GDAL_DEBUG for BUILD_TYPE=DEBUG, and I've changed the tests in the public headers to test for "#if defined(DEBUG) || defined(GDAL_DEBUG)``. That should be backwards compatible.

rouault: > On Linux, I can mix a debug build of my app with a release lib of GDAL...

with the C API yes, but not if you use some parts of the C++ API

SunBlack: What do you think about also renaming the macro within the cpp files and `GdalCompilationFlags.cmake` on this way `DEBUG` could be completely removed with GDAL 4.0.

rouault: > What do you think about also renaming the macro within the cpp files and `GdalCompilationFlags.cmake` on this way `DEBUG` could be completely removed with GDAL 4.0.

That could probably be done in GDAL 3.11. I didn't want to do the full rename in this PR, to keep it in a state where it is backportable to the 3.10 maintenance branch

PR #11318, opened 2024-11-20
CMake: MSVC 64-bit cleanup

dg0yt:
## What does this PR do?

This PR revises CMake code for MSVC.
- Variable `CMAKE_CL_64` is discouraged:
  https://cmake.org/cmake/help/latest/variable/CMAKE_CL_64.html
- Macro `_WIN64` is defined by MSVC preprocessor:
  https://learn.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-170
- `/machine:x64` breaks arm64.
  (Ported from https://github.com/microsoft/vcpkg/commit/dcf8a700dee158c3690e8a824a7ff8e24687a6ec#diff-d9dc7eb8a09a06ac0d7d8954d0422637dd33206bd73bee02952f1222f9438630R9.)

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS: Windows
* Compiler: MSVC

---

No local testing due to no MSVC.


cmake/modules/packages/FindDB2.cmake
```diff
@@ -46,7 +46,7 @@ if(UNIX)
         endforeach(db2_libdir)
     endforeach(db2_install_path)
 elseif(WIN32)
-    if (CMAKE_CL_64) # 64-bit build, DB2 64-bit installed
+    if (CMAKE_SIZEOF_VOID_P EQUAL "8") # 64-bit build, DB2 64-bit installed
```
rouault, line 49: can you delete this file entirely? DB2 has been dropped long time ago, so it is totally unused

dg0yt:
Removed FindDB2.cmake.  
There are interesting hits for `git grep DB2` (frmts/drivers.ini, and HACK_FOR_IBM_DB2_V72), but this out of scope for this PR.

rouault:
> frmts/drivers.ini,

just removed that one

> HACK_FOR_IBM_DB2_V72

hum, that one dates back to times where I was hardly born. It doesn't related with the DB2 driver that was added (and eventually removed) after this HACK_FOR_IBM_DB2_V72

Issue #11330, opened 2024-11-21
Doc: build warnings (->failures) on ReadTheDocs infra due to new example directive


rouault:
CC @dbaston

For some reasons, master builds fail on RTD infra due to warnings like:
```
/home/docs/checkouts/readthedocs.org/user_builds/gdal/checkouts/latest/doc/source/programs/gdal_calc.rst:39: WARNING: Can't find example gdal-calc-multiple-expr
/home/docs/checkouts/readthedocs.org/user_builds/gdal/checkouts/latest/doc/source/programs/gdal_calc.rst:47: WARNING: Can't find example gdal-calc-3d-sum
/home/docs/checkouts/readthedocs.org/user_builds/gdal/checkouts/latest/doc/source/programs/gdal_calc.rst:147: WARNING: Can't find example gdal-calc-allbands-1
/home/docs/checkouts/readthedocs.org/user_builds/gdal/checkouts/latest/doc/source/programs/gdal_calc.rst:147: WARNING: Can't find example gdal-calc-allbands-2
/home/docs/checkouts/readthedocs.org/user_builds/gdal/checkouts/latest/doc/source/programs/gdaladdo.rst:100: WARNING: Can't find example refresh-vrt
/home/docs/checkouts/readthedocs.org/user_builds/gdal/checkouts/latest/doc/source/programs/gdaladdo.rst:124: WARNING: Can't find example refresh-tiff
/home/docs/checkouts/readthedocs.org/user_builds/gdal/checkouts/latest/doc/source/programs/gdaladdo.rst:170: WARNING: Can't find example use-rrd
/home/docs/checkouts/readthedocs.org/user_builds/gdal/checkouts/latest/doc/source/programs/gdalbuildvrt.rst:174: WARNING: Can't find example vrtnodata
/home/docs/checkouts/readthedocs.org/user_builds/gdal/checkouts/latest/doc/source/programs/gdalbuildvrt.rst:183: WARNING: Can't find example separate
/home/docs/checkouts/readthedocs.org/user_builds/gdal/checkouts/latest/doc/source/programs/gdalbuildvrt.rst:221: WARNING: Can't find example filelist
/home/docs/checkouts/readthedocs.org/user_builds/gdal/checkouts/latest/doc/source/programs/gdallocationinfo.rst:40: WARNING: Can't find example xml
[...]

PR #11331, opened 2024-11-21
Doc: Fix example xrefs in singlehtml build

dbaston:
Should fix https://github.com/OSGeo/gdal/issues/11330


rouault:
confirming locally that this fixes the issue. Thanks for the quick fix

PR #11346, opened 2024-11-24
libcsf: address (false positive) Coverity warnings

rouault:
Coverity Scan has false positives about calls to CsfGetAttrIndex() that might return NR_ATTR_IN_BLOCK whereas they can not. Adding a CsfGetAttrBlockAndIdx() function that returns both the block and index should make it happy.

@kordejong can you double check that?


kordejong:
Hi Even. Thanks for your patch. My colleague Oliver Schmitz (@OliverSchmitz) will have a look at it. He is the principle maintainer of PCRaster. CSF is PCRaster's native raster format.

OliverSchmitz:
Hello Even, the modifications look reasonable. I also checked them with our PCRaster builds, all tests pass. Should be fine to merge

PR #11359, opened 2024-11-26
Add a hint in the documentation, that the configuration file was created with SameMinorVersion.

### LINKED ISSUE #11315: [CMake] Version check fails
### What is the bug?

In our CMake file we have this line:

```cmake
find_package(GDAL 3.3 CONFIG REQUIRED)
```
But this fails in case we building a dev build of CMake

```
CMake Error at CMakeLists.txt:38 (find_package):
  Could not find a configuration file for package "GDAL" that is compatible
  with requested version "3.3".

  The following configuration files were considered but not accepted:

    C:/dev/dependencies/gdal/lib/cmake/gdal/GDALConfig.cmake, version: 3.11.0dev
```

Note: Had this issue already on our CI when checking our via
```sh
git clone  --depth 1 --branch v3.10.0 https://github.com/OSGeo/gdal.git .
```
Shouldn't the `dev` removed on the commits with a release tag? Didn't checked the CMake source in which cases the `dev` bill be added and when not.

See bug decription

### Versions and provenance

Windows & Linux

### Additional context

_No response_

SunBlack:
Closes #11315 as all other improvements should be implemented on the CMake side.

Is there any Job within the workflow, where I can see the preview of the doc, as I locally don't have Sphinx installed?

### DIFF
--- a/doc/source/development/cmake.rst
+++ b/doc/source/development/cmake.rst
@@ -10,7 +10,7 @@ The recommended way to use the GDAL library 3.5 or higher in a CMake project is 
 link to the imported library target ``GDAL::GDAL`` provided by
 the CMake configuration which comes with the library. Typical usage is:
 
-.. code::
+.. code:: cmake
 
     find_package(GDAL CONFIG REQUIRED)
 
@@ -26,10 +26,26 @@ the cache variable or environment variable ``CMAKE_PREFIX_PATH``. In
 particular, CMake will consult (and set) the cache variable
 ``GDAL_DIR``.
 
+If a specific minor version is required, you can search for this via:
+
+.. code:: cmake
+
+    find_package(GDAL 3.10 CONFIG REQUIRED)
+
+If more than one minor version is to be supported at the same time,
+``${GDAL_VERSION}`` itself must be evaluated.
+
+.. code:: cmake
+
+    find_package(GDAL CONFIG REQUIRED)
+    if(GDAL_VERSION VERSION_LESS "3.7" OR GDAL_VERSION VERSION_GREATER "3.9")
+      message(FATAL_ERROR "Required at least GDAL version 3.7 - 3.9, but found ${GDAL_VERSION}.")
+    endif()
+
 Before GDAL 3.5, it is recommended to use `find module supplied with CMake <https://cmake.org/cmake/help/latest/module/FindGDAL.html>`__.
 This also creates the ``GDAL::GDAL`` target. It requires CMake version 3.14.
 
-.. code::
+.. code:: cmake
 
     cmake_minimum_required(VERSION 3.14)
 


doc/source/development/cmake.rst
```diff
@@ -26,10 +26,27 @@ the cache variable or environment variable ``CMAKE_PREFIX_PATH``. In
 particular, CMake will consult (and set) the cache variable
 ``GDAL_DIR``.
 
+To ensure API compatibility, it is recommended to search for a specific
```
rouault, line 29: I'm not sure the 'recommended' phrasing applies to most use cases. This is rather for people that would want to stick with a particular version. However, in the general case the C API within the 3.x series is backwards compatible, and the C++ one is also close to be backwards compatible for users of the API (a bit less for out-of-tree plugins), so checks of the type ``GDAL_VERSION VERSION_GREATER_EQUAL "x.y"`` are what most people will need

rouault: > Wouldn't that speak in favor of using `SameMajorVersion` instead of `SameMinorVersion`?

yes, if we didn't have those occasional breaks in the C++ plugin API. An out of tree driver might use ``find_package(GDAL 3.7)`` to really mean they exactly want 3.7, because 3.8 might break them.
That said they could also write ``find_package(GDAL 3.7.0...3.7.99)`` if we do your suggested change

@dg0yt thoughts?

dg0yt: I would choose SameMajorVersion. TBH in vcpkg we effectively apply SameMajorVersion, although via a wrapper. This probably saves a lot of patching in other packages. But vcpkg doesn't build plugins.

SunBlack: > But we'd need some words of caution regarding breakage in C++ plugin API that might require out-of-tree plugin driver to add more precision version checking

Can you briefly explain what you mean by that? It doesn't change anything about how GDAL is built. And if, for example, GDAL is built with the internal dependencies, but then the same dependency is used again in a different version in your project, there may already be issues if the ABI is different. So I don't see where this would change anything.

rouault: > Can you briefly explain what you mean by that?

hum not sure I remember ;-)  What I had in mind is what I mentionned about external out-ot-tree drivers that could need to do ``find_package(GDAL 3.7)`` if they are just compatible of that version. But I suppose that would also work with a SameMajorVersion strategy?

SunBlack: That's exactly the point you made above. If someone is only compatible with 3.7.x, they would have to write `find_package(GDAL 3.7)` instead of `find_package(GDAL 3.7.0...3.7.99)`, because with `SameMajorVersion` `find_package(GDAL 3.7)` means `>=3.7`.

What is the plan, since there is already a GDAL 4.0 milestone: Should 4.0 already come after 3.11 or is it still a collection for the distant future? If 4.0 is already on the horizon, we could also postpone this change and possibly already create a PR with 4.0 as the target label so that it is not forgotten.

rouault: > What is the plan, since there is already a GDAL 4.0 milestone:

GDAL 4.0 is purely hypothetical for now. Nobody seems excited about it (I'm not personnally)

I believe it's fine to switch to SameMajorVersion in 3.11. If that's a breakage, that will be a relatively minor one. We can document it in MIGRATION_GUIDE.TXT

SunBlack: Ok, should I then use this PR to customize the readme for 3.10.1 as the behavior is currently and then the customization with `SameMajorVersion` in a separate PR? Because I think the documentation should also contain the note that there was a change.

rouault: > Ok, should I then use this PR to customize the readme for 3.10.1 as the behavior is currently and then the customization with `SameMajorVersion` in a separate PR? Because I think the documentation should also contain the note that there was a change.

if you wish. Not sure which "readme" you refer to exactly?

SunBlack: Sorry, not readme, I meant the documentaiton I'm touching in this PR.

rouault: To be sure of your intentions, do you expect this PR to be backported to the stable branch as well so it appears in the "stable" doc  ?
and then you'll do a follow-up PR chaning to SameMajorVersion for master only?

SunBlack: Since in most cases there are several patch releases per minor release before the next minor release comes out, my idea was to make this PR for the next 3.10 release and then make the CMake changes for 3.11 including the documentation adjustments in a separate PR.

PR #11368, opened 2024-11-26
Docker: add options to add Oracle, ECW, MrSID drivers to ubuntu-full amd64

rouault:


msmitherdc:
Just an fyi that there is an updated MrSiD sdk for a more recent gcc version here: https://bin.extensis.com/download/developer/MrSID_DSDK-9.5.5.5244-rhel9.x86-64.gcc1131.zip

rouault:
> Just an fyi that there is an updated MrSiD sdk for a more recent gcc version here: https://bin.extensis.com/download/developer/MrSID_DSDK-9.5.5.5244-rhel9.x86-64.gcc1131.zip

thanks for the hint. PR updated with that.

PR #11372, opened 2024-11-27
OpenFileGDB: be robust to unusual .gdbindexes files with weird/corrupted/not-understood entries

### LINKED ISSUE #11295: SetAttributeFilter / ApplicationException
### What is the bug?

Dear all,

i'm opening a GeoDB file ad on a layer would like to set a AttributeFilter but i obtain this applicationException

error

What i'm doing wrong ?

Thanks in advance

Try to set AttributeFilter

### Versions and provenance

Windows

### Additional context

_No response_

rouault:
Fixes #11295


ogr/ogrsf_frmts/openfilegdb/filegdbtable.cpp
```diff
@@ -2547,7 +2548,7 @@ int FileGDBTable::GetIndexCount()
         returnErrorAndCleanupIf(static_cast<GUInt32>(pabyEnd - pabyCur) <
                                     sizeof(GUInt32),
                                 VSIFree(pabyIdx));
-        GUInt32 nIdxNameCarCount = GetUInt32(pabyCur, 0);
+        const GUInt32 nIdxNameCarCount = GetUInt32(pabyCur, 0);
```
elpaso, line 2551: Was `Car` actually meant to be `Char` ?

ogr/ogrsf_frmts/openfilegdb/filegdbtable.cpp
```diff
@@ -2557,13 +2558,56 @@ int FileGDBTable::GetIndexCount()
             ReadUTF16String(pabyCur, nIdxNameCarCount));
         pabyCur += 2 * nIdxNameCarCount;
 
+        // 4 "magic fields"
+        returnErrorAndCleanupIf(static_cast<GUInt32>(pabyEnd - pabyCur) <
```
elpaso, line 2562: I do not follow: if the comparison is between a sum of `size_t` (that I assume to be `size_t` itself) why cast to GUInt32 ?

ogr/ogrsf_frmts/openfilegdb/filegdbtable.cpp
rouault: addressed in https://github.com/OSGeo/gdal/pull/11410

ogr/ogrsf_frmts/openfilegdb/filegdbtable.cpp
rouault: addressed in https://github.com/OSGeo/gdal/pull/11410

PR #11375, opened 2024-11-27
CI: add minimal testing for s390x / big endian

rouault:
Use cross-compiler for fast compilation, and only run a subset of tests through Docker qemu emulation.


rouault:
First "cold" run: 33 minutes
Second "hot" run using compilation cache: 16 minutes

PR #11380, opened 2024-11-28
Zarr: Check SRS axis order when setting data axis to SRS axis mapping

skogler:
## What does this PR do?

We recently encountered issues when resampling EPSG:4326 Zarr files using gdalwarp, where the resulting output files would have correct bounds, but incorrectly sampled data (swapped latitude and longitude). This happens for example when warping a Zarr to a GTif file as  follows:

```bash
# Create a Zarr in 4326 spanning the whole world
gdal_create -of Zarr -a_srs EPSG:4326 -a_ullr -180 90 180 -90 -ot Byte -outsize 512 256 -burn 127 /tmp/empty.zarr

# Convert it to a Tiff using gdalwarp 
gdalwarp /tmp/empty.zarr /tmp/empty.tif
```
If you open `/tmp/empty.tif` in QGIS, you will see that only the +-90 longitude range is filled with nonzero values.

When debugging this, I found that the `data_axis_to_srs_axis_mapping` in the `gdalmdiminfo` output was not consistent with the axis order declared in the CRS and in the Zarr file. 
For example, if the Zarr file contains an array that is ordered latitude first, GDAL reports the axis mapping as [2, 1]. I think it should be [1, 2] in this case, because EPSG:4326 also has latitude ordered first.

After this patch, the initialization code explicitly checks:
- Is the SRS is oriented north-up?
- Are we using authority compliant axis order?

Only if both are true, the axis order is swapped. We believe that this is necessary since the axes in Zarr can be ordered in any way, but are annotated as X(longitude) or Y(latitude) using either name-based or attribute-based matching.

We tested this using UTM (XY order) and 4326 (YX order) files, and it resolves the problem for us.

This might not be the best way to do this, but it solves a critical issue for us. We are happy to receive feedback and change the logic if needed though!

## Tasklist

 - [x] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS: Linux 6.12
* Compiler: gcc 14.2.1
* GDAL: 3.10

### DIFF
--- a/autotest/gdrivers/zarr_driver.py
+++ b/autotest/gdrivers/zarr_driver.py
@@ -1137,6 +1137,33 @@ def test_zarr_read_classic_2d():
         gdal.RmdirRecursive("/vsimem/test.zarr")
 
 
+def test_zarr_read_classic_2d_authority_compliant_crs_with_latlon_order():
+
+    try:
+        src_ds = gdal.Open("data/small_world_400pct_1band.vrt")
+        src_srs = src_ds.GetSpatialRef()
+        assert src_srs is not None
+        assert src_srs.GetAuthorityCode(None) == "4326"
+
+        filename = "/vsimem/test.zarr"
+        gdal.GetDriverByName("ZARR").CreateCopy(filename, src_ds, strict=False)
+        ds = gdal.Open(filename)
+
+        assert ds is not None
+        srs = ds.GetSpatialRef()
+        assert srs is not None
+        assert srs.GetAuthorityCode(None) == "4326"
+        # 1. Axis order of CRS is lat/lon because axis mapping strategy is authority compliant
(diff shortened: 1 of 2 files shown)


rouault:
@skogler Thanks for the PR. I believe it is correct. However I've proposed something more in line with the initial intended logic in https://github.com/OSGeo/gdal/pull/11381 .  That issue with axis order is tricky since the conventions in the multidimensional part of GDAL and in the classic 2D one are switched ...

skogler:
@rouault  Thanks for the quick response and fix! Yeah, the axis order thing was quite difficult to debug since it is swapped in the 2D API and gdalinfo and gdalmdiminfo output different values. 

I tested your proposed fix with my use case and it works perfectly, thanks!

PR #11381, opened 2024-11-28
Zarr: fix incorrect DataAxisToSRSAxisMapping for EPSG geographic CRS

rouault:
Resolves issue of https://github.com/OSGeo/gdal/pull/11380


bstadlbauer:
Thank you so much for the quick turn-around 🙇

PR #11390, opened 2024-11-29
Assorted set of fixes for WMTS issue

### LINKED ISSUE #11387: Missing EPSG/ESRI codes since updating to GDAL 3
### What is the bug?

Hello

I've been using GDAL for a while, in part to use WMTS datasets from NASA's Trek API such as those listed here: https://trek.nasa.gov/tiles/apidoc/trekAPI.html?body=mars
I've stayed on GDAL 2.4 for way too long, but recently updated. However, now NASA Trek's WMTS no longer load in GDAL:

> ERROR 1: PROJ: proj_create: crs not found: EPSG:104905
> ERROR 1: Cannot parse CRS 'urn:ogc:def:crs:EPSG::104905'

I'm using GDAL 3.8.3 and PROJ 7.2.1 currently (those packaged in my Linux distribution), but I also tried with the latest and get the same error. It worked with GDAL 2.4.2.

The code is this one: https://epsg.io/104905 + https://spatialreference.org/ref/esri/104905/. There is also 104903 for the Moon, etc.

Looking deeper, I can see that previously that was handled by WKT files in GDAL_DATA (specifically esri_extra.wkt for EPSG:104905), but now that's handled by proj.db in PROJ_LIB, and the code seems to be missing from that database.

I don't know if there is a good reason for this regression or if that is a bug, and I don't know if this is considered a GDAL or a PROJ issue. Sorry about that.
I guess I will try to add the codes to proj.db manually on the short term, but I wanted to inform you about this issue.

gdalinfo https://trek.nasa.gov/tiles/Mars/EQ/THEMIS_NightIR_ControlledMosaics_100m_v2_oct2018/1.0.0/WMTSCapabilities.xml

### Versions and provenance

GDAL 3.8.3 (Windows 10 manual build + openSUSE Leap 15.6 package from distrib)
GDAL 3.9.3 (Windows 10 using latest OSGeo4W release)

### Additional context

_No response_

rouault:
Fixes #11387

- OGRSpatialReference::importFromEPSG(): tries with ESRI when it looks like an ESRI code, but with a warning when that succeeds
- WMTS: for geographic CRS with official lat,lon order, be robust to bounding box and TopLeftCorner being in the wrong axis order and emits a warning
- HTTP driver: re-emit warnings/errors raised by underlying driver

### DIFF
--- a/autotest/gdrivers/data/wmts/WMTSCapabilities_THEMIS_NightIR_ControlledMosaics_100m_v2_oct2018.xml
+++ b/autotest/gdrivers/data/wmts/WMTSCapabilities_THEMIS_NightIR_ControlledMosaics_100m_v2_oct2018.xml
@@ -1,0 +1,102 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Capabilities xmlns="http://www.opengis.net/wmts/1.0" xmlns:ows="http://www.opengis.net/ows/1.1" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:gml="http://www.opengis.net/gml" xsi:schemaLocation="http://www.opengis.net/wmts/1.0 http://schemas.opengis.net/wmts/1.0/wmtsGetCapabilities_response.xsd" version="1.0.0">
+    <!-- Service Identification --> 
+    <ows:ServiceIdentification>
+        <ows:Title>THEMIS_NightIR_ControlledMosaics_100m_v2_oct2018</ows:Title>
+        <ows:ServiceType>OGC WMTS</ows:ServiceType>
+        <ows:ServiceTypeVersion>1.0.0</ows:ServiceTypeVersion>
+    </ows:ServiceIdentification> <!-- Operations Metadata --> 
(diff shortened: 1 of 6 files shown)


elpaso (approved):
LGTM. What puzzles me is if we really want/need to support a misconfigured server confusing ESRI and EPSG ?

jratike80:
>  What puzzles me is if we really want/need to support a misconfigured server confusing ESRI and EPSG 

The WMTS and WMS standards do not define too well other CRS namespaces. WMS 1.1 defines only "EPSG" and "AUTO". WMS 1.3.0 introduced "AUTO2" and "CRS" with codes 1, 84, 83, 27, and 88. I think I have never seen any other of those despite CRS:84. It  is vaguelly said that other CRS's may exist, but standards do not have any examples. Therefore it has been common to invent own fake "EPSG" codes like 900913, or to use EPSG: instead of ESRI: because some clients think that if it is not EPSG then it is an error. Even this ArcGIS server seems to be configured to announce non-existing <CRS>EPSG:102139</CRS> https://gtkdata.gtk.fi/arcgis/services/Rajapinnat/GTK_Kalliopera_WMS/MapServer/WMSServer?service=WMS&version=1.3.0&request=GetCapabilities

PR #11394, opened 2024-11-29
Various hardening checks when trying to insert huuuge geometries to PostGIS

rouault:


port/cpl_string.cpp
```diff
@@ -2785,12 +2785,19 @@ char *CPLUnescapeString(const char *pszInput, int *pnLength, int nScheme)
 char *CPLBinaryToHex(int nBytes, const GByte *pabyData)
 
 {
-    char *pszHex = static_cast<char *>(CPLMalloc(nBytes * 2 + 1));
+    CPLAssert(nBytes >= 0);
+    char *pszHex = static_cast<char *>(
+        VSI_MALLOC_VERBOSE(static_cast<size_t>(nBytes) * 2 + 1));
+    if (!pszHex)
+    {
+        pszHex = CPLStrdup("");
```
elpaso, line 2793: should we emit a warning here?

elpaso (approved):

port/cpl_string.cpp
rouault: > should we emit a warning here?

VSI_MALLOC_VERBOSE() emits a CPLError(CE_Failure, CPLE_OutOfMemory,...) including the file and line from which it has been invoked when it fails

PR #11429, opened 2024-12-03
netCDF: add support to identify a geolocation array for a variable that lacks a 'coordinates' attribute

rouault:
but that has side "lon" and "lat" 2D variables whose dimensions are the same as the last 2 ones of the variable of interest

Fixes https://lists.osgeo.org/pipermail/gdal-dev/2024-December/059870.html


aborruso:
Thank you very much

PR #11431, opened 2024-12-03
VRTProcessedDataset: allow to define a BandCountFromLastStep element

### LINKED ISSUE #11430: VRT processed dataset: automatically create output bands
### Feature description

A VRT processed dataset lets you either (a) manually declare each output band using a `<VRTRasterBand>` tag, or (b) automatically create the same number of bands as the input. It would be nice if there was also some way the final `Step` could automatically create the output bands during its initialization phase.

For example, I have a netCDF with 1488 bands representing hourly data, and I'd like to use a VRT expression to reduce this to 124 bands of daily data. Currently, I would need to know the correct number of output bands and create them manually. Then, if the input file changes, the VRT would no longer have the correct number of output bands.

### Additional context

_No response_

rouault:
to specify that the number of bands of the VRT dataset should be the number of output bands returned by the initialization function of the last step.

Fixes #11430

CC @dbaston


dbaston:
What do you think about a slightly more general syntax like

```
<OutputBands count="FROM_SOURCE|FROM_LAST_STEP|1,2,3..." type="FROM_SOURCE|FROM_LAST_STEP|Int16,Float32,..." />
```

If both `<OutputBands>` and `<VRTRasterBand>` are present, an error is thrown.

If neither `<OutputBands>` nor `<VRTRasterBand>` is present, the behavior is the same as `<OutputBands count="FROM_SOURCE" type="FROM_SOURCE" />`

rouault:
> What do you think about a slightly more general syntax like

yes, I although thought we could have a more general syntax. I like your proposal

> If both `<OutputBands>` and `<VRTRasterBand>` are present, an error is thrown.

I would still allow that, at least when their number and type matches what OutputBands asks. A typical use case is someone doing "gdalinfo -stats the.vrt" (or just opening in QGIS) which will add explicit VRTRasterBand elements with the statistics. The generic XMLInit() code currently starts by instanciating those bands, then VRTProcessedDataset::XMLInit() temporarily inhibts them, and checks that the number of output bands of the last steps matches the number of VRTRasterBand . If so, it uses the later.

dbaston:
I merged this with #11209 locally and after trivial modification to my init function I was able to open the VRT below. Very nice!
```
<VRTDataset subClass="VRTProcessedDataset">
  <Input>
    <SourceFilename relativeToVRT="1">../byte.tif</SourceFilename>
  </Input>
  <OutputBands count="FROM_LAST_STEP" type="FROM_LAST_STEP"/>
  <ProcessingSteps>
    <Step>
      <Algorithm>Expression</Algorithm>
      <Argument name="expression">return [B1, 2*B1, 3*B1]</Argument>
    </Step>
  </ProcessingSteps>
</VRTDataset>
```

doc/source/drivers/raster/vrt_processed_dataset.rst
```diff
@@ -132,6 +130,48 @@ The value of ``Algorithm`` must be a registered VRTProcessedDataset function. At
...
+
+* if there are explicit ``VRTRasterBand`` elements, they must have a
+  ``subClass="VRTProcessedRasterBand"`` attribute
+
+* it there are no explicit ``VRTRasterBand`` elements, the number and data types
+  of input bands are used implicitly.
+
+Both ``OutputBands`` and  ``VRTRasterBand`` elements may be defined. The information
+specified by ``OutputBands`` will be used in priority, and ``VRTRasterBand`` elements
+will be used only if there are compatible with the band count and data type specified
```
dbaston, line 170:
```suggestion
will be used only if they are compatible with the band count and data type specified
```

doc/source/drivers/raster/vrt_processed_dataset.rst
```diff
@@ -132,6 +130,48 @@ The value of ``Algorithm`` must be a registered VRTProcessedDataset function. At
...
 
+Starting with GDAL 3.11, a ``OutputBands`` element can be
+defined as a child element of ``VRTDataset``, with the following 2 attributes:
+
+* ``count`` whose value can be ``FROM_SOURCE`` to indicate that the output band
+  count must be the same as the number of bands of the input dataset,
+  ``FROM_LAST_STEP`` to indicate that it must be the number of output bands
+  returned by the initialization function of the last step, or an integer value.
+
+* ``type`` whose value can be ``FROM_SOURCE`` to indicate that the output band
```
dbaston, line 141: I know I suggested `type`, but do you think this should be `dataType` for consistency with other tags?

dbaston (approved):

doc/source/drivers/raster/vrt_processed_dataset.rst
rouault: fixed

doc/source/drivers/raster/vrt_processed_dataset.rst
rouault: yes, I had also the same thought. Changed to dataType

PR #11439, opened 2024-12-05
ogr2ogr: fix crash with -ct and using Arrow code path (e.g source is GeoPackage) (3.10.0 regression)

### LINKED ISSUE #11438: ogr2ogr with a PROJ pipeline fails when the source is a GeoPackage on GDAL 3.10.0
### What is the bug?

When applying a PROJ affine transformation pipeline with ogr2ogr to a GeoPackage file, the operation fails with the error message "Segmentation fault (core dumped)". The output file is either unreadable (for GeoPackage output) or has no geometry (for GeoJSON output).

The following example translates and rotates points given in UTM-based coordinates. This used to work fine in previous versions of GDAL/OGR, including 3.9.3. The error occurred with various setups with conda on my local machine and with an Ubuntu-based docker image, but not when using an Alpine-based docker image. The operation also succeeds when the source data is given in a different file format (e.g. GeoJSON or Shapefile).

Create a source file with points named `testpoints.geojson` of the following content inside a directory named `data` :

```
{
"type": "FeatureCollection",
"name": "testpoints",
"crs": { "type": "name", "properties": { "name": "urn:ogc:def:crs:EPSG::32632" } },
"features": [
{ "type": "Feature", "properties": { }, "geometry": { "type": "Point", "coordinates": [ 768212.618588526034728, 6102308.445965819060802 ] } },
{ "type": "Feature", "properties": { }, "geometry": { "type": "Point", "coordinates": [ 768231.531171231181361, 6102310.891558409668505 ] } }
]
}

```
From the parent folder, start a docker container and execute the following commands.

```
docker run -it -v "$(pwd)"/data:/data --rm ghcr.io/osgeo/gdal:ubuntu-small-3.10.0
cd data
ogr2ogr testpoints.gpkg testpoints.geojson
ogr2ogr -ct "+proj=pipeline +step +proj=affine +xoff=-768107.4470771276 +yoff=-6102340.838122625 +step +proj=affine +xoff=768107.4470771276 +yoff=6102340.838122625 +s11=0.7071067811865476 +s12=-0.7071067811865476 +s21=0.7071067811865476 +s22=0.7071067811865476" transformed_points.gpkg testpoints.gpkg
```

### Versions and provenance

GDAL 3.10.0, released 2024/11/01

Failed on Windows 11 and on Ubuntu 24.04.1 LTS (WSL) with conda-forge, both
Also failed with the docker image ghcr.io/osgeo/gdal:ubuntu-small-3.10.0
Succeeded with the docker image ghcr.io/osgeo/gdal:alpine-small-3.10.0

### Additional context

_No response_

rouault:
Fixes #11438

Passing "--config OGR2OGR_USE_ARROW_API=NO" or "-t_srs {srs_def}" can be used as a workaround

### DIFF
--- a/apps/ogr2ogr_lib.cpp
+++ b/apps/ogr2ogr_lib.cpp
@@ -4099,12 +4099,11 @@ bool SetupTargetLayer::CanUseWriteArrowBatch(
             {
                 return false;
             }
-            auto poSrcSRS = m_poUserSourceSRS ? m_poUserSourceSRS
-                                              : poSrcLayer->GetLayerDefn()
-                                                    ->GetGeomFieldDefn(0)
-                                                    ->GetSpatialRef();
-            if (!poSrcSRS ||
-                !OGRGeometryFactory::isTransformWithOptionsRegularTransform(
+            const auto poSrcSRS = m_poUserSourceSRS ? m_poUserSourceSRS
+                                                    : poSrcLayer->GetLayerDefn()
+                                                          ->GetGeomFieldDefn(0)
+                                                          ->GetSpatialRef();
+            if (!OGRGeometryFactory::isTransformWithOptionsRegularTransform(
                     poSrcSRS, m_poOutputSRS, nullptr))
             {
                 return false;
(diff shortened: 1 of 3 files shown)


jratike80:
OGR2OGR_USE_ARROW_API is documented in https://gdal.org/en/latest/programs/ogr2ogr.html but not in 
https://gdal.org/en/latest/user/configoptions.html. Is that intentional? If it affects only ogr2ogr then it may be ok this way.

PR #11443, opened 2024-12-06
KEA: use native chunksize for copying RAT

gillins:
## What does this PR do?

Changes the chunksize used for copying the RAT to be the default native HDF5 chunksize (10000) rather and 1000. This should improve performance while copying to a KEA file.

https://github.com/ubarsc/kealib/pull/68

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:

### DIFF
--- a/frmts/kea/keacopy.cpp
+++ b/frmts/kea/keacopy.cpp
@@ -111,7 +111,6 @@ static bool KEACopyRasterData(GDALRasterBand *pBand,
     return true;
 }
 
-constexpr int RAT_CHUNKSIZE = 1000;
 
 // copies the raster attribute table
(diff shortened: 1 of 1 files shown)


frmts/kea/keacopy.cpp
```diff
@@ -111,7 +111,6 @@ static bool KEACopyRasterData(GDALRasterBand *pBand,
     return true;
 }
 
-constexpr int RAT_CHUNKSIZE = 1000;
```
rouault, line 114:
```suggestion
```

PR #11454, opened 2024-12-06
Update T-Kartor sponsorship name

hobu:


doc/source/sponsors/index.rst
```diff
@@ -160,7 +160,7 @@ the health of the project:
 
     .. container:: horizontal-logo
 
-        `T-Kartor USA <https://www.t-kartorusa.com/>`__
+        `T-Kartor <https://www.t-kartorusa.com/>`__
```
rouault, line 163: The old URL redirects to that one, so better use it?

```suggestion
        `T-Kartor <https://www.t-kartor.com//>`__
```

PR #11459, opened 2024-12-08
ADBC: add spatial support for DuckDB databases and GeoParquet

rouault:
- Automate loading duckdb_spatial extension when installed, and when the dataset is DuckDB or Parquet
- Retrieve geometries (GEOMETRY type) as OGR geometries
- Read GeoParquet metadata to figure out spatial extent, CRS and geometry type per geometry column
- Use duckdb_spatial ST_Intersects() for faster spatial filtering (when done with OGRLayer::SetSpatialFilter()), potentially leveraging DuckDB RTree when it is available.
- Use GeoParquet bounding box column in complement to above
- Passthrough forward of WHERE claused expresse through OGRLayer::SetAttributeFilter()


ogr/ogrsf_frmts/adbc/ogradbclayer.cpp
```diff
@@ -127,16 +789,46 @@ int OGRADBCLayer::TestCapability(const char *pszCap)
...
+                OGRDuplicateCharacter(GetDescription(), '\'').c_str(),
+                pszGeomColName,
+                OGRDuplicateCharacter(pszGeomColName, '"').c_str()));
+            return poTmpLayer &&
+                   std::unique_ptr<OGRFeature>(poTmpLayer->GetNextFeature());
+        }
+        else if (!m_geomColBBOX[m_iGeomFieldFilter].osXMin.empty())
+        {
+            // Let's assume that the presence of a geometry bounding box
+            // column is sufficient enough to pretent to have fast spatial
```
paleolimbot, line 825: pretent -> pretend?

rouault: fixed, thanks

PR #11462, opened 2024-12-09
Ogrinfo: command line help text fixes

jratike80:
Fix wrong command line help text for the "summary only" option and add points at the end of some help texts for making them consistent.

### DIFF
--- a/apps/ogrinfo_lib.cpp
+++ b/apps/ogrinfo_lib.cpp
@@ -2305,18 +2305,18 @@ static std::unique_ptr<GDALArgumentParser> GDALVectorInfoOptionsGetParser(
     argParser->add_argument("-al")
         .store_into(psOptions->bAllLayers)
         .help(_("List all layers (used instead of having to give layer names "
-                "as arguments)"));
+                "as arguments)."));
 
     {
         auto &group = argParser->add_mutually_exclusive_group();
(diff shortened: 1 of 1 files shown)


apps/ogrinfo_lib.cpp
```diff
@@ -2305,18 +2305,18 @@ static std::unique_ptr<GDALArgumentParser> GDALVectorInfoOptionsGetParser(
...
-                "as arguments)"));
+                "as arguments)."));
 
     {
         auto &group = argParser->add_mutually_exclusive_group();
         group.add_argument("-so", "-summary")
             .store_into(psOptions->bSummaryParser)
-            .help(_("Summary only: list all layers (used instead of having to "
-                    "give layer names as arguments)"));
+            .help(_("Summary only: Show only summary information like "
```
elpaso, line 2314:
```suggestion
            .help(_("Summary only: show only summary information like "
```

elpaso (approved):

apps/ogrinfo_lib.cpp
jratike80: @elpaso , did I accept your suggestion correctly? I am not so familiar with editing in GitHub.

elpaso: You did :+1:

PR #11479, opened 2024-12-11
[CSV] OGR_SCHEMA support - RFC 103

elpaso:
Implementation of RFC 103 - OGR_SCHEMA open option for CSV driver

https://gdal--11071.org.readthedocs.build/en/11071/development/rfc/rfc103_schema_open_option.html

Also, promote Jane to CTO.

### DIFF
--- a/autotest/ogr/ogr_csv.py
+++ b/autotest/ogr/ogr_csv.py
@@ -26,6 +26,7 @@
 # Boston, MA 02111-1307, USA.
 ###############################################################################
 
+import json
 import math
 import pathlib
 import sys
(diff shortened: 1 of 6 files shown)


autotest/ogr/ogr_csv.py
```diff
@@ -3182,6 +3183,225 @@ def test_ogr_csv_invalid_wkt(tmp_vsimem):
...
+
+        if schema:
+            open_options.append("OGR_SCHEMA=" + schema)
+        else:
+            open_options = []
+
+        # Validate the JSON schema
+        if not expected_warning and schema:
+            schema = json.loads(schema)
+            #################gdaltest.validate_json(schema, "ogr_fields_override.schema.json")
```
elpaso, line 3319:
```suggestion
            gdaltest.validate_json(schema, "ogr_fields_override.schema.json")
```

rouault:
It could be worth checking what happens in a case of a CSV like the following with the special WKT column, if you exclude it in fullSchema mode, or if you alter its type to a non string value:

```
id,WKT,foo
1,"POINT (1 2)",bar
```

doc/source/drivers/vector/csv.rst
```diff
@@ -413,6 +413,15 @@ The following open options are supported:
 
       Maximum number of bytes for a line (-1=unlimited).
 
+-  .. oo:: OGR_SCHEMA
+      :choices: <filename>|<json string>
+      :since: 3.11.0
+
+      Partially or totally overrides the auto-detected schema to use for creating the layer.
```
rouault, line 420: We could mention this option has precedence over the AUTODETECT_TYPE open option and the .csvt side car file.

elpaso: done

elpaso:
> It could be worth checking what happens in a case of a CSV like the following with the special WKT column, if you exclude it in fullSchema mode, or if you alter its type to a non string value:
> 
> ```
> id,WKT,foo
> 1,"POINT (1 2)",bar
> ```

I have added a test case for WKT, can you please check if the tested expected behavior makes sense to you?

see: https://github.com/OSGeo/gdal/pull/11479/commits/c4920807ae9257090150446886248b23f24f4bd9

PR #11487, opened 2024-12-12
Add a 'gdal raster edit' command to override SRS, extent and metadata

rouault:


jratike80:
Late comment, but how about having `-a_bbox` instead of `-a_ullr`? At least for me it is easier to remember the minx,miny,maxx,maxy order.

rouault:
> Late comment, but how about having `-a_bbox` instead of `-a_ullr`

Are you refering to https://github.com/OSGeo/gdal/pull/11487/files#diff-9b7ac11ef53a8c359d463cb4baac1fdd36fb16cc84af402e505974e63d6451fcR198  ? If so, it is just that gdal edit is underneath redirecting to gdal_translate, and thus using its existing -a_ullr. An implementation detail.

the user has only access to gdal raster edit --extent=xmin,ymin,xmax,ymax   , which will be rewritten to generate a north-up image. If we want to allow south-up image, then we'd probably need some explicit --south-up-geotransform or something like that?

jratike80:
> Are you refering to...

Yes, and I did not understand what I was reading.  `gdal raster edit --extent=xmin,ymin,xmax,ymax` is fine. But well, we use bbox in some other places, is extent any better for this use case?

rouault:
> But well, we use bbox in some other places, is extent any better for this use case?

yes good observation. I was hesitating about that too. For some reason, I associated "bbox" with filtering use cases, hence an alternate name, but I'm not strongly minded about that, and I'm open to changes. Deciding about good new names is going to be a huge challenge... Perhaps you can create a ticket about that? I've added a "gdal_cli" issue label to categorize everything related to the new "gdal" frontend

jratike80:
I just started to review the OGC standards but I fear I can find there contradicting examples. Usually that happens when comparing two or more standards.

rouault:
> I just started to review the OGC standards

oh noooooooooooooooo!!!!!!!!!!!!!!!! lol

PR #11493, opened 2024-12-14
Doc: clarify VSICurlPartialClearCache()

rouault:
Cf https://github.com/rasterio/rasterio/pull/3276#issuecomment-2543169811

### DIFF
--- a/port/cpl_vsil_curl.cpp
+++ b/port/cpl_vsil_curl.cpp
@@ -6295,6 +6295,13 @@ void VSICurlClearCache(void)
  * mechanisms can prevent opening new files, or give an outdated version of
  * them.
  *
+ * The filename prefix must start with the name of a known virtual file system
+ * (such as "/vsicurl/", "/vsis3/")
+ *
+ * VSICurlPartialClearCache("/vsis3/b") will clear all cached state for any file
+ * or directory starting with that prefix, so potentially "/vsis3/bucket",
+ * "/vsis3/basket/" or "/vsis3/basket/object".
+ *
  * @param pszFilenamePrefix Filename prefix
  * @since GDAL 2.4.0
  */


port/cpl_vsil_curl.cpp
```diff
@@ -6295,6 +6295,12 @@ void VSICurlClearCache(void)
  * mechanisms can prevent opening new files, or give an outdated version of
  * them.
  *
+ * The filename prefix must start with the prefix of a known virtual file system.
+ *
+ * VSICurlPartialClearCache("/vsis3/f") will clear all cached state for any file
+ * or directory starting with that prefix, so potentially "/vsis3/fu",
+ * "/vsis3/foo/" or "/vsis3/foo/bar".
```
sgillies, line 6302: Me, I'm trying not to use `foo` and `bar` anymore.

```suggestion
 * VSICurlPartialClearCache("/vsis3/b") will clear all cached state for any file
 * or directory starting with that prefix, so potentially "/vsis3/bucket",
 * "/vsis3/basket/" or "/vsis3/basket/object".
```

Do I understanding correctly that `/vsis3/` can't be truncated? Because that's the name of a handler and not a URL path?

Using "prefix" in the documentation is ambiguous, I think, because S3 also uses the term "prefix", but in a different sense.

rouault: > do I understanding correctly that `/vsis3/` can't be truncated?

yes

> Using "prefix" in the documentation is ambiguous

yes, any alternative suggestion? That said prefix here matches VSIGetFileSystemsPrefixes()

rouault: > yes, any alternative suggestion? That said prefix here matches VSIGetFileSystemsPrefixes()

changing to:
 * The filename prefix must start with the name of a known virtual file system
 * (such as "/vsicurl/", "/vsis3/", etc.)

sgillies (approved):
Thank you @rouault . "filename prefix" works for me.

Issue #11453, opened 2024-12-06
CMAKE FindSFCGAL.cmake for libSFCGAL-osg.2.0.0


chris-english:
### Feature description

In the event that one has compiled libSFCGAL-osg.2.0.0, CMAKE FindSFCGAL only looks for SFCGAL and SFCGALd. Would:

find_library(SFCGAL_LIBRARY_RELEASE NAMES SFCGAL-osg
             HINTS ${SC_SFCGAL_LIBDIR}
             PATH ${SFCGAL_ROOT}
             PATH_SUFFIXES lib Lib
             )

be sufficient edit to FindSFCGAL.cmake that it would be found?

rouault:
CC @lbartoletti 

@chris-english For my education, what is "libSFCGAL-osg" ? Which distribution channel provides it ?

chris-english:
I managed to build it after following this OpenSceneGraph [PR](https://github.com/openscenegraph/OpenSceneGraph/pull/1281/commits/26f6c9b5a63715dde3c9f85c8faf035d3ec0f657). OpenSceneGraph is currently starved for resources as seemingly all devs are now working on Vulkan GPU. Yet, there remains a large group still working with CPU, including me.

chris-english:
But given how Find is configured, even pointing to lib and include and GDAL_USE_SFCGAL=ON bring no joy in cmake <point-to-things> ..

dg0yt:
A general change might be multiple NAMES + NAMES_PER_DIR.
And while writing this comment, I see that PATHS is misspelled in that find modules.
~~~diff
-find_library(SFCGAL_LIBRARY_RELEASE NAMES SFCGAL
+find_library(SFCGAL_LIBRARY_RELEASE
+             NAMES SFCGAL SFCGAL-osg
+             NAMES_PER_DIR
              HINTS ${SC_SFCGAL_LIBDIR}
-             PATH ${SFCGAL_ROOT}
+             PATHS ${SFCGAL_ROOT}
              PATH_SUFFIXES lib Lib
             )
~~~~

chris-english:
The @dg0yt suggested change to cmake find library does find SFCGAL-osg, but then undefined references:
[ 86%] Linking CXX shared library libgdal.so
/usr/bin/ld: ogr/CMakeFiles/ogr.dir/ogrgeometry.cpp.o: in function `OGRGeometry::OGRexportToSFCGAL(OGRGeometry const*)':
[...]

PR #11495, opened 2024-12-15
Fix CMake find modules

dg0yt:
## What does this PR do?

This PR fixes oversights in the find modules which may lead to unexpected search results.
In particular, it adds NAMES_PER_DIR for all cases of multi-name library and program search. This is needed to give precedence to the order of locations as indicated by HINTS, PATHS etc. over the order of NAMES.

This was noticed in the discussion of https://github.com/OSGeo/gdal/issues/11453.

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments


cmake/modules/3.20/FindPostgreSQL.cmake
```diff
@@ -177,6 +177,7 @@ endif()
 function(__postgresql_find_library _name)
   find_library(${_name}
    NAMES ${ARGN}
+   NAMES_PER_DIR
```
rouault, line 180: would you mind contributing that upstream to https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/FindPostgreSQL.cmake?ref_type=heads#L183 ?

rouault (approved):

cmake/modules/3.20/FindPostgreSQL.cmake
dg0yt: On a second look, this spot is not relevant (but doesn't do harm) because it is not multi-name. Upstream anyways, or revert?

rouault: better not change if not relevant

PR #11496, opened 2024-12-15
gdalbuildvrt: Emit warning on invalid value for -vrtnodata

dbaston:


apps/gdalbuildvrt_lib.cpp
```diff
@@ -436,6 +436,21 @@ static CPLString GetProjectionName(const char *pszProjection)
 /*                           AnalyseRaster()                            */
 /************************************************************************/
 
+static void checkNoDataValues(const std::vector<BandProperty> &asProperties)
+{
+    for (const auto &oProps : asProperties)
+    {
+        if (oProps.bHasNoData && GDALDataTypeIsInteger(oProps.dataType) &&
```
rouault, line 443: probably better to use ``bool GDALIsValueExactAs(double dfValue, GDALDataType eDT)``

dbaston: I thought there must be something like this but didn't look hard enough. Thanks!

PR #11498, opened 2024-12-16
Fix cppcheck nullPointerOutOfMemory warnings

rouault:


lpechacek:
As a side note on the patch titled "Fix cppcheck nullPointerOutOfMemory warnings". The omitted check on `fopen()` return status causes build failure of Rasterio in openSUSE Build Service. The thing is that Rasterio tests try to open a TIFF file which triggers `CPLGetPhysicalRAM()` in GDAL. `CPLGetPhysicalRAM()` attempts to open `/proc/meminfo` but `/proc` is not mounted in the build environment. `fopen()` returns `NULL` and the process subsequently dies in `fgets()`. Build time tests fail and packages are not generated.

The minimal reproducer for the SEGV is as follows:
```
fmn:/tmp # cat gdal-segv.c 
#include <gdal/cpl_conv.h>

int main(void)
{
	CPLGetPhysicalRAM();
	return 0;
}
fmn:/tmp # gcc -o gdal-segv gdal-segv.c -lgdal
fmn:/tmp # ./gdal-segv 
Segmentation fault (core dumped)
fmn:/tmp # rpm -q gdal
gdal-3.10.1-217.2.i586
```

So, the fix has wider application than a rare out-of-memory situation.

rouault:
> So, the fix has wider application than a rare out-of-memory situation.

ok, I've cherry-picked it to the release/3.10 branch as well (for 3.10.2) in 8c7e61ece464b95feb7d67390f32b745328819b5 . You can also cherry-pick in your GDAL build recipe on top of 3.10.1

PR #11499, opened 2024-12-16
[sqlite] OGR_SCHEMA implementation - RFC 103

elpaso:
Implementation of RFC 103 - OGR_SCHEMA open option for SQLite driver

https://gdal--11071.org.readthedocs.build/en/11071/development/rfc/rfc103_schema_open_option.html

### DIFF
--- a/ogr/ogrsf_frmts/sqlite/ogrsqlitelayer.cpp
+++ b/ogr/ogrsf_frmts/sqlite/ogrsqlitelayer.cpp
@@ -900,11 +900,18 @@ OGRFeature *OGRSQLiteLayer::GetNextRawFeature()
                 {
                     const GIntBig nVal =
                         sqlite3_column_int64(m_hStmt, iRawField);
-                    if (nVal >= INT_MIN && nVal <= INT_MAX)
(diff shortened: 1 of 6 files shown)


ogr/ogrsf_frmts/sqlite/ogrsqlitelayer.cpp
```diff
@@ -900,11 +900,18 @@ OGRFeature *OGRSQLiteLayer::GetNextRawFeature()
                 {
                     const GIntBig nVal =
                         sqlite3_column_int64(m_hStmt, iRawField);
-                    if (nVal >= INT_MIN && nVal <= INT_MAX)
-                        poFeature->SetFieldSameTypeUnsafe(
-                            iField, static_cast<int>(nVal));
+                    if (poFieldDefn->GetSubType() == OFSTBoolean)
```
elpaso, line 903: @rouault this change is only for consistency with other drivers but I'm not entirely sure , what do you think?

rouault: Seems reasonable

PR #11504, opened 2024-12-17
Internal libtiff: fix writing a Predictor=3 file with non-native endianness

rouault:


rouault:
cherry picked to 3.10 per 42e8bb712e

PR #11508, opened 2024-12-17
/vsicurl?: accept header.<key>=<value> to specify HTTP request headers

### LINKED ISSUE #11503: /vsicurl/: Add query parameter equivalent for `GDAL_HTTP_HEADERS`
### Feature description

The `/vsicurl/` VFS permits an alternate syntax that allows specifying options like `/vsicurl?option_i=val_i&url=(etc.)`.
While the `GDAL_HTTP_HEADER_FILE` global configuration option has an equivalent query parameter `header_file`, `GDAL_HTTP_HEADERS` seems to have no such equivalent.
I want to be able to dynamically set certain headers per dataset, meaning I cannot use the regular configuration option (which I understand to be global).
I'm currently able to use the `header_file` option instead, but this requires managing and disposing of a temporary file or `/vsimem/` location, which can be error-prone.

So a dataset like 
`/vsicurl?header_file=%2Ftmp%2Fheaders&url=(etc.)`
with `/tmp/headers` containing
```
my-header: Foo
other-header: Bar
```
could instead be expressed as e.g.
`/vsicurl?headers=my-header%3A%20Foo%2Cother-header%3A%20Bar&url=(etc.)`
using the same syntax as `GDAL_HTTP_HEADERS` but URL-encoded.

### Additional context

_No response_

rouault:
Fixes #11503

### DIFF
--- a/autotest/gcore/vsicurl.py
+++ b/autotest/gcore/vsicurl.py
@@ -1609,3 +1609,27 @@ def test_vsicurl_VSICURL_QUERY_STRING(server, filename, query_string):
             assert statres.size == 3
         finally:
             gdal.SetPathSpecificOption(full_filename, "VSICURL_QUERY_STRING", None)
+
+
+###############################################################################
+# Test /vsicurl?header.foo=bar&
+
+
+@gdaltest.enable_exceptions()
+def test_vsicurl_header_option(server):
+
+    gdal.VSICurlClearCache()
+
+    handler = webserver.SequentialHandler()
+    handler.add(
+        "HEAD",
+        "/test_vsicurl_header_option.bin",
+        200,
+        {"Content-Length": "3"},
+        expected_headers={"foo": "bar", "Accept": "application/json"},
+    )
+
+    with webserver.install_http_handler(handler):
+        full_filename = f"/vsicurl?header.foo=bar&header.Accept=application%2Fjson&url=http%3A%2F%2Flocalhost%3A{server.port}%2Ftest_vsicurl_header_option.bin"
+        statres = gdal.VSIStatL(full_filename)
+        assert statres.size == 3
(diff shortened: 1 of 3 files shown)


rouault:
@sgillies in case you have some suggestion, although I anticipate your lack of enthusiasm for this new GDAL idiosynchracy ;-)

I did consider the "/vsicurl?header:foo=bar&url=...." syntax, with ':' as the separator between "header" and "foo", but then I realize that strictly speaking ':' is a reserved character that should in principle be URL encoded, which would hurt for readability, hence the choice of dot ...

sgillies:
@rouault I'm not a fan, it's true 😄 

What about specifying headers in curl style? Like

`"-H Accept:application/json /vsicurl?url=http://localhost/test_vsicurl_header_option.bin"`

rouault:
> What about specifying headers in curl style? Like
> 
> `"-H Accept:application/json /vsicurl?url=http://localhost/test_vsicurl_header_option.bin"`

Hum, not super convinced... And my main objection I have is that it would be yet another way of specifying /vsicurl/ arguments, whereas we have a framework in place (I'm perhaps wrong, but I had the impression /vsicurl? and URL arguments was derived from a suggestion of yours ... ?)

sgillies:
Yeah, it was my suggestion. I thought it would be nice to be able to use standard URL parsers to deserialize the options out of a GDAL filename. I'm not opposed to this PR, just not excited about supporting GDAL filenames that are loaded with headers. Separating *what* from *how* is a good thing, no?

rouault:
> Separating _what_ from _how_ is a good thing, no?

maybe... but if you look at the /vsicurl? existing options, this is mostly (only actually) about the how. I'd stick with that framework for consistency

PR #11517, opened 2024-12-19
[gdal][api] GDALDriverHasOpenOption

elpaso:
Add a function to check whether an open option is supported by the driver.

### DIFF
--- a/gcore/gdaldriver.cpp
+++ b/gcore/gdaldriver.cpp
@@ -1483,12 +1483,12 @@ bool GDALDriver::HasOpenOption(const char *pszOpenOptionName) const
     if (pszOpenOptionName == nullptr)
         return false;
 
+    // Const cast is safe here since we are only reading the metadata
     auto pszOOMd{const_cast<GDALDriver *>(this)->GetMetadataItem(
         GDAL_DMD_OPENOPTIONLIST)};
     if (pszOOMd == nullptr)
         return false;
 
-    // Const cast is safe here since we are only reading the metadata
     const CPLXMLTreeCloser oXml{CPLParseXMLString(pszOOMd)};
     for (CPLXMLNode *option = oXml->psChild; option != nullptr;
          option = option->psNext)
--- a/gcore/gdaldriver.cpp
+++ b/gcore/gdaldriver.cpp
@@ -1478,6 +1478,23 @@ bool GDALDriver::CanVectorTranslateFrom(
     return bRet;
 }
 
+bool GDALDriver::HasOpenOption(const char *pszOpenOptionName) const
+{
+    if (pszOpenOptionName == nullptr)
+        return false;
+
+    auto pszOOMd{const_cast<GDALDriver *>(this)->GetMetadataItem(
+        GDAL_DMD_OPENOPTIONLIST)};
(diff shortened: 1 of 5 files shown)


gcore/gdaldriver.cpp
```diff
@@ -1478,6 +1478,23 @@ bool GDALDriver::CanVectorTranslateFrom(
...
+
+    auto pszOOMd{const_cast<GDALDriver *>(this)->GetMetadataItem(
+        GDAL_DMD_OPENOPTIONLIST)};
+    if (pszOOMd == nullptr)
+        return false;
+
+    // Const cast is safe here since we are only reading the metadata
+    const std::string osOpenOptions{pszOOMd};
+    const std::string osKey =
+        "<Option name='" + std::string(pszOpenOptionName) + "'";
```
rouault, line 1494: I guess that will work in practice, but that's a bit fragile. We could potentially have a driver using escaped double quote ``"<Option name=\"foo\" ....>"``. Or reordering attributes with the name not being first.
The clean solution would be to use CPLXMLParseString() and browse through the structure. I'm OK though with that simplified implementation, but perhaps add a code comment reflecting my observation.

elpaso: I thought that being us in control of the source XML it could be acceptable but better be robust if we can.

gcore/gdaldriver.cpp
```diff
@@ -1478,6 +1478,28 @@ bool GDALDriver::CanVectorTranslateFrom(
...
+    if (pszOpenOptionName == nullptr)
+        return false;
+
+    auto pszOOMd{const_cast<GDALDriver *>(this)->GetMetadataItem(
+        GDAL_DMD_OPENOPTIONLIST)};
+    if (pszOOMd == nullptr)
+        return false;
+
+    // Const cast is safe here since we are only reading the metadata
+    const auto oXml{CPLParseXMLString(pszOOMd)};
```
rouault, line 1492: CPLParseXMLString() doesn't return a smart pointer, instead use:

```suggestion
    const CPLXMLTreeCloser oXml{CPLParseXMLString(pszOOMd)};
```

gcore/gdaldriver.cpp
```diff
@@ -1478,6 +1478,28 @@ bool GDALDriver::CanVectorTranslateFrom(
...
+    if (pszOpenOptionName == nullptr)
+        return false;
+
+    auto pszOOMd{const_cast<GDALDriver *>(this)->GetMetadataItem(
+        GDAL_DMD_OPENOPTIONLIST)};
+    if (pszOOMd == nullptr)
+        return false;
+
+    // Const cast is safe here since we are only reading the metadata
+    const CPLXMLTreeCloser oXml{CPLParseXMLString(pszOOMd)};
```
rouault, line 1492:
```suggestion
    // Const cast is safe here since we are only reading the metadata
    auto pszOOMd{const_cast<GDALDriver *>(this)->GetMetadataItem(
        GDAL_DMD_OPENOPTIONLIST)};
    if (pszOOMd == nullptr)
        return false;

    const CPLXMLTreeCloser oXml{CPLParseXMLString(pszOOMd)};
```

PR #11537, opened 2024-12-22
OGR VRT: accept SrcRegion value to be any geometry type as well as SetSpatialFilter()

### LINKED ISSUE #11518: OGRVRTLayer throws CE_Failure with description ignoring it
### What is the bug?

I'm using Mapserver (v8.2.2) and a VRT file with a SrcRegion defined. And when performing a WMS getfeatureinfo request the request fails. To my opinion this is due to the type of error thrown in `ogrvrtlayer.cpp`.

Case:
I'm doing a WMS request that results in an error thrown by the OGR VRT layer: `CPLError(CE_Failure, CPLE_AppDefined,"Spatial filter should be polygon when a SrcRegion is defined. Ignoring it");`. However the `poSrcRegion` is correctly set and execution continues.

This exception is caught by mapserver with `if (CPLGetLastErrorType() == CE_Failure)` that results in failure.

Expected result:
OGR VRT layer to throw a CE_Warning like this: `CPLError(CE_Warning, CPLE_AppDefined,"Spatial filter should be polygon when a SrcRegion is defined. Ignoring it");`

**Dataset:**
BAG geopackage: https://service.pdok.nl/lv/bag/atom/bag.xml

**VRT file:**
```
<OGRVRTDataSource>
    <OGRVRTLayer name="pand">
        <SrcDataSource>bag-light.gpkg</SrcDataSource>
        <SrcLayer>pand</SrcLayer>
        <LayerSRS>EPSG:28992</LayerSRS>
        <GeometryField>
            <!-- Clip is only implemented at GeometryField level and not at OGRVRTLayer level -->
            <SrcRegion clip="false">POLYGON((120000 435000, 155000 435000, 155000 470000, 120000 470000, 120000 435000))</SrcRegion>
        </GeometryField>
    </OGRVRTLayer>
</OGRVRTDataSource>
```

No idea how to reproduce this with ogr2ogr since I cannot find out how the `m_poFilterGeom` is set that seems to be something different then a polygon.

### Versions and provenance

GDAL 3.6, 3.9.3, 3.10.0 self-built from binaries.

### Additional context

_No response_

rouault:
Fixes #11518

### DIFF
--- a/doc/source/drivers/vector/vrt.rst
+++ b/doc/source/drivers/vector/vrt.rst
@@ -247,7 +247,7 @@ layer name, and may have the following subelements:
 
 
 - **SrcRegion** (optional) : This element is used to
-  define an initial spatial filter for the source features, such that only features intersecting ``SrcRegion`` will be returned be included in the VRT layer. This spatial
+  define an initial spatial filter for the source features, such that only features intersecting ``SrcRegion`` will be returned in the VRT layer. This spatial
   filter will be combined with any spatial filter explicitly set on the
   VRT layer with the SetSpatialFilter() method. The value of the element
   must be a valid WKT string defining a geometry in the spatial reference system of the source layer. An optional **clip**
--- a/doc/source/drivers/vector/vrt.rst
+++ b/doc/source/drivers/vector/vrt.rst
@@ -250,7 +250,7 @@ layer name, and may have the following subelements:
   define an initial spatial filter for the source features. This spatial
   filter will be combined with any spatial filter explicitly set on the
   VRT layer with the SetSpatialFilter() method. The value of the element
-  must be a valid WKT string defining a polygon. An optional **clip**
+  must be a valid WKT string defining a geometry. An optional **clip**
   attribute can be set to "TRUE" to clip the geometries to the source
   region, otherwise the source geometries are not modified.
 
--- a/doc/source/drivers/vector/vrt.rst
+++ b/doc/source/drivers/vector/vrt.rst
@@ -247,10 +247,10 @@ layer name, and may have the following subelements:
 
 
 - **SrcRegion** (optional) : This element is used to
-  define an initial spatial filter for the source features. This spatial
+  define an initial spatial filter for the source features, such that only features intersecting ``SrcRegion`` will be returned be included in the VRT layer. This spatial
   filter will be combined with any spatial filter explicitly set on the
   VRT layer with the SetSpatialFilter() method. The value of the element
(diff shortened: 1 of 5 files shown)


doc/source/drivers/vector/vrt.rst
```diff
@@ -250,7 +250,7 @@ layer name, and may have the following subelements:
   define an initial spatial filter for the source features. This spatial
```
dbaston, line 250:
```suggestion
  define an initial spatial filter for the source features, such that only features intersecting ``SrcRegion`` will be returned be included in the VRT layer. This spatial
```

doc/source/drivers/vector/vrt.rst
```diff
@@ -250,7 +250,7 @@ layer name, and may have the following subelements:
...
   filter will be combined with any spatial filter explicitly set on the
   VRT layer with the SetSpatialFilter() method. The value of the element
-  must be a valid WKT string defining a polygon. An optional **clip**
+  must be a valid WKT string defining a geometry. An optional **clip**
```
dbaston, line 253:
```suggestion
  must be a valid WKT string defining a geometry in the spatial reference system of the source layer. An optional **clip**
```

doc/source/drivers/vector/vrt.rst
```diff
@@ -247,10 +247,10 @@ layer name, and may have the following subelements:
 
 
 - **SrcRegion** (optional) : This element is used to
-  define an initial spatial filter for the source features. This spatial
+  define an initial spatial filter for the source features, such that only features intersecting ``SrcRegion`` will be returned be included in the VRT layer. This spatial
   filter will be combined with any spatial filter explicitly set on the
```
jratike80, line 251: I think either "will be returned" or "will be included", not both

doc/source/drivers/vector/vrt.rst
```diff
@@ -247,10 +247,10 @@ layer name, and may have the following subelements:
 
 
 - **SrcRegion** (optional) : This element is used to
-  define an initial spatial filter for the source features. This spatial
+  define an initial spatial filter for the source features, such that only features intersecting ``SrcRegion`` will be returned be included in the VRT layer. This spatial
```
rouault, line 250:
```suggestion
  define an initial spatial filter for the source features, such that only features intersecting ``SrcRegion`` will be returned in the VRT layer. This spatial
```

doc/source/drivers/vector/vrt.rst
rouault: fixed, thanks

PR #11546, opened 2024-12-25
docs: Instruct how to uninstall custom GDAL

DelazJ:
because it took me some readings on the Internet to find how I could rollback to the default install after the new install didn't work properly. It may help other beginners like me.
 

## Tasklist

 - [ ] ADD YOUR TASKS HERE
 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

## Environment

* OS:
* Compiler:

### DIFF
--- a/doc/source/development/building_from_source.rst
+++ b/doc/source/development/building_from_source.rst
@@ -108,6 +108,12 @@ for the shared lib, *e.g.* ``set (GDAL_LIB_OUTPUT_NAME gdal_x64 CACHE STRING "" 
 
 Refer to :ref:`using_gdal_in_cmake` for how to use GDAL in a CMake project.
 
+To uninstall GDAL, from the build directory:
+
+.. code-block:: bash
+
+    cmake --build . --target uninstall
+
 Building on Windows
 +++++++++++++++++++
 


doc/source/development/building_from_source.rst
```diff
@@ -108,6 +108,12 @@ for the shared lib, *e.g.* ``set (GDAL_LIB_OUTPUT_NAME gdal_x64 CACHE STRING ""
 
 Refer to :ref:`using_gdal_in_cmake` for how to use GDAL in a CMake project.
 
+To uninstall GDAL, from the build directory:
```
rouault, line 111: maybe add this word of caution?
```suggestion
To uninstall GDAL, from the build directory (be careful: this does not restore the files that pre-existed to the last installation, but just removes the last installed files!)
```

DelazJ: @rouault Does that mean that the previously installed GDAL may get broken and people should "ideally" reinstall it from packages?

rouault: > Does that mean that the previously installed GDAL may get broken and people should "ideally" reinstall it from packages?

if people installed their custom GDAL in the same directory that the system one (they should not !), yes it will be broken

DelazJ: OK! That probably explains why mine is still working; I blindy follow the instructions in gdal docs and it landed in /usr/local.
Thanks.

PR #11558, opened 2024-12-30
Add 'gdal raster buildvrt'

rouault:
Equivalent of gdalbuildvrt (not all options ported)

@hobu With the addition of VSIGlob(), "gdal raster buildvrt /vsis3/my_bucket/*.tif out.vrt" works

### DIFF
--- a/doc/source/programs/gdal_raster_buildvrt.rst
+++ b/doc/source/programs/gdal_raster_buildvrt.rst
@@ -1,0 +1,181 @@
+.. _gdal_raster_buildvrt_subcommand:
+
+================================================================================
+"gdal raster buildvrt" sub-command
+================================================================================
+
+.. versionadded:: 3.11
+
+.. only:: html
+
+    Build a virtual dataset (VRT).
+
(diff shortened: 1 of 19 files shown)


dbaston:
What do you think about `gdal raster mosaic` as a name for this?

rouault:
> What do you think about `gdal raster mosaic` as a name for this?

Good suggestion. We should probably also allow non-VRT output as users might rightfully expect to be able to generate a COG. Can you file an issue about that?

doc/source/programs/gdal_raster_buildvrt.rst
```diff
@@ -0,0 +1,181 @@
...
+--------
+
+.. example::
+   :title: Make a RGB virtual mosaic from 3 single-band input files
+   :id: separate
+
+   .. code-block:: bash
+
+       gdal raster biuldvrt --separate red.tif green.tif blue.tif rgb.vrt
+
```
jratike80, line 167: biuldvrt -> buildvrt

PR #11569, opened 2025-01-03
VRTProcessedDataset: reduce block size when amount of RAM for temp buffers would be too large

rouault:
and also: gcore: use GDALCopyWords64() instead of GDALCopyWords() for microscopic perf improvement

### DIFF
--- a/frmts/vrt/vrtprocesseddataset.cpp
+++ b/frmts/vrt/vrtprocesseddataset.cpp
@@ -301,12 +301,29 @@ CPLErr VRTProcessedDataset::Init(const CPLXMLNode *psTree,
 
     const auto poSrcFirstBand = m_poSrcDS->GetRasterBand(1);
     poSrcFirstBand->GetBlockSize(&m_nBlockXSize, &m_nBlockYSize);
+    bool bUserBlockSize = false;
     if (const char *pszBlockXSize =
             CPLGetXMLValue(psTree, "BlockXSize", nullptr))
+    {
+        bUserBlockSize = true;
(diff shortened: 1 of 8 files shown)


frmts/vrt/vrtprocesseddataset.cpp
```diff
@@ -521,6 +538,66 @@ CPLErr VRTProcessedDataset::Init(const CPLXMLNode *psTree,
         return CE_Failure;
     }
 
+    // Use only up to 40% of RAM to acquire source bands and generate the output
+    // buffer.
+    const auto nMaxMemAllowed = CPLGetUsablePhysicalRAM() / 10 * 4;
```
dbaston, line 543: This behavior seems similar to the `-wm` argument of `gdalwarp`. Would it makes sense to have a configuration option like `GDAL_WORKING_MEMORY` that can be used in multiple places? Or would that be too confusing if, for example, our source for the processed VRT is a warped VRT?

rouault: > Or would that be too confusing if, for example, our source for the processed VRT is a warped VRT?

yes there's indeed a risk of confusion. One could imagine there could be some memory allocator in GDAL that would know which components of GDAL need memory allocation and distribute what is available amongst them. Like component A would claim SIZE_A , B would claim SIZE_B, etc.  and then if sume(SIZE_i)> max_allowed, the allocator would tell X it just can use SIZE_X * max_allowed / sum(SIZE_i). But I don't see how that could be implemented in our architecture.

PR #11572, opened 2025-01-03
Add GDALTranspose2D() for fast 2D matrix transposition

rouault:
and use it in VRTProcessedDataset

On the use case of https://github.com/OSGeo/gdal/pull/11209#issuecomment-2568585769 , "gdal_translate src.vrt dst.tif" runtime is reduced from 38 seconds to 22 seconds on my machine.

CC @dbaston


gcore/rasterio.cpp
```diff
@@ -5673,3 +5673,193 @@ void GDALDeinterleave(const void *pSourceBuffer, GDALDataType eSourceDT,
...
+    }
+
+#define GDALTranspose2D_internal(DST_TYPE_CST, DST_TYPE, DST_IS_COMPLEX)       \
+    case DST_TYPE_CST:                                                         \
+        GDALTranspose2D<DST_TYPE, DST_IS_COMPLEX>(                             \
+            pSrc, eSrcType, static_cast<DST_TYPE *>(pDst), nSrcWidth,          \
+            nSrcHeight);                                                       \
+        break
+
+    switch (eDstType)
```
dbaston, line 5843: I'm finding a couple of things about this tricky to follow - the switch statement with `case` blocks behind a macro and the different signatures of `GDALTranspose2D`, each with its own internally defined macros. It feels like it should be possible to simplify this dispatch with a variadic template, but maybe it isn't feasible. I struggled with the same thing at https://github.com/OSGeo/gdal/pull/11573/files#diff-558d98f8123abd945a2e0d3cda118167d0853a589a1c307945470e2432b92482R733, which is a much simpler case. (And the best templated solution I came up with is ugly, requiring an unused tag parameter, https://github.com/dbaston/gdal/commit/75015bb9d3b7b33c70d834eaaa17300d7b288d92)

GEOS uses a template to dispatch based two coordinate sequence types, which again is much simpler because there are only 16 possible combinations. https://github.com/libgeos/geos/blob/main/include/geos/geom/CoordinateSequences.h#L43

I don't have a change to suggest at this point.

rouault: I've tried to improve a bit the switch() so that the case's look apparent instead of being hidden in the macro

dbaston:
> 38 seconds to 22 seconds

Can confirm with 30s -> 20s here. Nice!

PR #11573, opened 2025-01-03
VRT Processed Dataset: Add BandScaleOffset algorithm

dbaston:
Adds an algorithm to unscale rasters before they are input to other algorithms such as expression evaluation.

### DIFF
--- a/doc/source/drivers/raster/vrt_processed_dataset.rst
+++ b/doc/source/drivers/raster/vrt_processed_dataset.rst
@@ -12,6 +12,9 @@ to apply chained processing steps that may apply to several bands at the same ti
 The following built-in algorithms are introduced, and may typically be applied
 in the following order:
 
+- BandScaleOffset (since GDAL 3.11): apply per-band scale and offset values from
+  :cpp:func:`GDALRasterBand::GetScale` and :cpp:func:`GDALRasterBand::GetOffset`.
+
 - LocalScaleOffset: apply per-pixel gain and offset coming (typically subsampled)
   from auxiliary datasets. Can be used for dehazing processing.
 
--- a/frmts/vrt/vrtprocesseddataset.cpp
+++ b/frmts/vrt/vrtprocesseddataset.cpp
@@ -574,6 +574,9 @@ CPLErr VRTProcessedDataset::Init(const CPLXMLNode *psTree,
             auto poBand =
                 new VRTProcessedRasterBand(this, i + 1, eOutputBandType);
             poBand->CopyCommonInfoFrom(poSrcBand);
+            poBand->SetNoDataValue(adfNoData[i]);
+            poBand->SetScale(1);
+            poBand->SetOffset(0);
             SetBand(i + 1, poBand);
         }
     }
@@ -713,7 +716,8 @@ bool VRTProcessedDataset::ParseStep(const CPLXMLNode *psStep, bool bIsFinalStep,
         for (int i = 1; i <= m_poSrcDS->GetRasterCount(); ++i)
         {
             int bHasVal = false;
-            const double dfVal = GetRasterBand(i)->GetOffset(&bHasVal);
+            const double dfVal =
+                m_poSrcDS->GetRasterBand(i)->GetOffset(&bHasVal);
             oStep.aosArguments.AddNameValue(
                 CPLSPrintf("offset_%d", i),
                 CPLSPrintf("%.17g", bHasVal ? dfVal : 0.0));
@@ -726,7 +730,8 @@ bool VRTProcessedDataset::ParseStep(const CPLXMLNode *psStep, bool bIsFinalStep,
         for (int i = 1; i <= m_poSrcDS->GetRasterCount(); ++i)
         {
             int bHasVal = false;
-            const double dfVal = GetRasterBand(i)->GetScale(&bHasVal);
+            const double dfVal =
+                m_poSrcDS->GetRasterBand(i)->GetScale(&bHasVal);
             oStep.aosArguments.AddNameValue(
                 CPLSPrintf("scale_%d", i),
                 CPLSPrintf("%.17g", bHasVal ? dfVal : 1.0));
--- a/frmts/vrt/vrtprocesseddatasetfunctions.cpp
+++ b/frmts/vrt/vrtprocesseddatasetfunctions.cpp
@@ -12,6 +12,7 @@
 
 #include "cpl_minixml.h"
 #include "cpl_string.h"
+#include "gdal_typetraits.h"
 #include "vrtdataset.h"
 
 #include <algorithm>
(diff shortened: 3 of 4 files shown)


frmts/vrt/vrtprocesseddataset.cpp
```diff
@@ -574,6 +574,9 @@ CPLErr VRTProcessedDataset::Init(const CPLXMLNode *psTree,
             auto poBand =
                 new VRTProcessedRasterBand(this, i + 1, eOutputBandType);
             poBand->CopyCommonInfoFrom(poSrcBand);
+            poBand->SetNoDataValue(adfNoData[i]);
+            poBand->SetScale(1);
```
dbaston, line 578: I guess we would want to copy the scale and offset from `poSrcBand` if we are not using the `BandScaleOffset` algorithm, but making this behavior conditional seems messy and I'm not sure there is a use case for applying algorithms to scaled values.

frmts/vrt/vrtprocesseddataset.cpp
```diff
@@ -713,7 +716,8 @@ bool VRTProcessedDataset::ParseStep(const CPLXMLNode *psStep, bool bIsFinalStep,
         for (int i = 1; i <= m_poSrcDS->GetRasterCount(); ++i)
         {
             int bHasVal = false;
-            const double dfVal = GetRasterBand(i)->GetOffset(&bHasVal);
+            const double dfVal =
```
dbaston, line 719: Pretty sure this was a typo in the original; `GetRasterBand(1)` returns `nullptr` at this point.

frmts/vrt/vrtprocesseddatasetfunctions.cpp
```diff
@@ -604,6 +605,196 @@ LUTProcess(const char * /*pszFuncName*/, void * /*pUserData*/,
...
+
+    if (!bIsFinalStep)
+    {
+        *pnOutBands = nInBands;
+    }
+
+    auto data = std::make_unique<BandScaleOffsetData>(nInBands, padfInNoData,
+                                                      papszFunctionArgs);
+
+    // Input NoData value is meaningless after scaling, so we set the output NoData value to NaN.
```
dbaston, line 690: Not sure using `NaN` is appropriate here or if I need to look elsewhere for a declared NoData value.

frmts/vrt/vrtprocesseddataset.cpp
rouault: indeed. Would you mind issuing a PR against release/3.10 with this fix and the line at line 733 ?

frmts/vrt/vrtprocesseddatasetfunctions.cpp
```diff
@@ -1514,6 +1702,16 @@ void GDALVRTRegisterDefaultProcessedDatasetFuncs()
         GDT_Float64, nullptr, 0, nullptr, 0, LUTInit, LUTFree, LUTProcess,
         nullptr);
 
+    GDALVRTRegisterProcessedDatasetFunc(
+        "BandScaleOffset", nullptr,
+        "<ProcessedDatasetFunctionArgumentsList>"
+        "   <Argument name='scale_{band}' type='builtin' />"
+        "   <Argument name='offset_{band}' type='builtin' />"
+        "</ProcessedDatasetFunctionArgumentsList>",
+        GDT_Float64, BandScaleOffsetSupportedTypes.data(),
```
rouault, line 1711:
```suggestion
        GDT_Unknown, BandScaleOffsetSupportedTypes.data(),
```

doc/source/drivers/raster/vrt_processed_dataset.rst
```diff
@@ -12,6 +12,9 @@ to apply chained processing steps that may apply to several bands at the same ti
 The following built-in algorithms are introduced, and may typically be applied
 in the following order:
 
+- BandScaleOffset (since GDAL 3.11): apply per-band scale and offset values from
+  :cpp:func:`GDALRasterBand::GetScale` and :cpp:func:`GDALRasterBand::GetOffset`.
```
rouault, line 16: an example would be welcome

frmts/vrt/vrtprocesseddatasetfunctions.cpp
rouault: BandAffineCombination has an optional
"   <Argument name='dst_nodata' type='double' "
        "description='Override output nodata value'/>"

frmts/vrt/vrtprocesseddatasetfunctions.cpp
rouault: code coverage confirms this should be fixed: https://coveralls.io/builds/71575735/source?filename=frmts%2Fvrt%2Fvrtprocesseddatasetfunctions.cpp#L735

frmts/vrt/vrtprocesseddataset.cpp
dbaston: done, https://github.com/OSGeo/gdal/pull/11580

frmts/vrt/vrtprocesseddatasetfunctions.cpp
dbaston: Fixed, thanks

frmts/vrt/vrtprocesseddatasetfunctions.cpp
dbaston: Should I add that here?

frmts/vrt/vrtprocesseddataset.cpp
dbaston: Thinking about this more, if we say (as these two lines imply) that algorithms must be run on the unscaled data, is the `BandScaleOffset` algorithm desirable at all? Or should the band scale/offset just be applied automatically when reading from the source raster?

frmts/vrt/vrtprocesseddatasetfunctions.cpp
rouault: > Should I add that here?

I don't have a strong opinion if we need to provide a override dstnodata. Just documenting that the algorithm standardize on NaN should be fine

frmts/vrt/vrtprocesseddataset.cpp
rouault: > Or should the band scale/offset just be applied automatically when reading from the source raster?

yes I guess applying it automatically should satisfy 99% of use cases. Do we want to provide some option in the XML to disable the automatic unscaling ?

dbaston: "unscale" attribute of "Input" with default value of "true" ? So `<Input unscale="false" />` if you want to preserve the current behavior?

rouault: yes, that sounds good.

doc/source/drivers/raster/vrt_processed_dataset.rst
jratike80: The example could describe how to set back the scale and offset values of the input data, which were "unscaled" earlier.
User may have data like the recommended DGIWG DEM variant that is scaled into Int16, and it may be surprise that data gets automatically unscaled, but it must be explicitly scaled back if Int16 with scale/offset is still desired.
I apologize if I have misunderstood something.

dbaston: > and it may be surprise that data gets automatically unscaled, but it must be explicitly scaled back if Int16 with scale/offset is still desired

I think a challenge with automatically re-scaling is knowing what scale factor to use. If we've applied an arbitrary processing algorithm to the inputs, I don't think we can assume the input scale/offset are still appropriate.

rouault:
@dbaston Do you plan to modify this PR to automatically apply unscaling ?

dbaston:
> @dbaston Do you plan to modify this PR to automatically apply unscaling ?

I was waiting to hear feedback on the issue @jratike80 raised about re-scaling. I think re-scaling should be done through e.g. `gdal_translate processed.vrt -ot x -scale y -offset z`.

rouault:
superseded per https://github.com/OSGeo/gdal/pull/11623

PR #11602, opened 2025-01-07
[Backport 3.10] netCDF: fix 3.10.1RC1 regression regarding reporting of values of extra  dimensions

rouault:
Fixes https://lists.osgeo.org/pipermail/gdal-dev/2025-January/060021.html


edzer:
Great thanks - fix confirmed!

PR #11606, opened 2025-01-08
Add CPLErrorOnce() and CPLDebugOnce()

rouault:
to emit an error/debug message only once during the life-time of a process.

### DIFF
--- a/alg/gdalwarpkernel.cpp
+++ b/alg/gdalwarpkernel.cpp
@@ -1240,13 +1240,8 @@ CPLErr GDALWarpKernel::PerformWarp()
         if (pafUnifiedSrcDensity != nullptr)
         {
             // Typically if there's a cutline or an alpha band
-            static bool bHasWarned = false;
-            if (!bHasWarned)
-            {
-                bHasWarned = true;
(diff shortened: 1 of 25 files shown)


rouault:
> like Python warnings, then? Use it in code as many times as we want, but a unique error message/code combination appears no more than once?

yes, this just institutionalizes a practice that has been done at hand up to now. Some GDAL errors/warnings are sometimes in tight loops or would be too frequently emitted.

dbaston:
This is a nice addition. I had a couple of ideas, please feel free to disregard:
- should we allow the warning to be raised a small N number of times? Since the implementation isn't testing the uniqueness of the message I can imagine cases where seeing a message like `"Value %d for field %s does not fit into a short and will be clamped."` with a few different values of `%d` and `%s` could make it easier to understand the source of the error.
- should we automatically append something like "Further messages of this type will be suppressed" to the text?

rouault:
> * should we allow the warning to be raised a small N number of times?

I can see the point, but dont' feel like going into that complication currently.

> should we automatically append something like "Further messages of this type will be suppressed" to the text?

good idea, done

PR #11609, opened 2025-01-08
[SQLite] Transaction field handling: intial implementation

elpaso:
Fix out of sync (not restored) fields after a ROLLBACK.

See: https://github.com/qgis/QGIS/pull/59797#issuecomment-2544133498

This is an SQLite only fix, if the approach sounds reasonable we should consider supporting more drivers. 

Funded by: QGIS.org


ogr/ogrgeomfielddefn.cpp
```diff
@@ -123,6 +123,49 @@ OGRGeomFieldDefn::~OGRGeomFieldDefn()
...
+ * @return a reference to the current object.
+ * @since GDAL 3.11
+ */
+OGRGeomFieldDefn &OGRGeomFieldDefn::operator=(const OGRGeomFieldDefn &oOther)
+{
+    if (&oOther != this)
+    {
+        SetName(oOther.pszName);
+        SetType(oOther.eGeomType);
+        SetSpatialRef(oOther.poSRS);
```
rouault, line 162: Looking at SetSpatialRef() implementation I can see it could have an issue if this.poSRS == oOther.poSRS and the reference count is 1

The method should be modified to:
```
void OGRGeomFieldDefn::SetSpatialRef(const OGRSpatialReference *poSRSIn)
{
    if (m_poSRS == poSRSIn)
    {
        return;
    }
    if (m_bSealed)
    {
        CPLError(
            CE_Failure, CPLE_AppDefined,
            "OGRGeomFieldDefn::SetSpatialRef() not allowed on a sealed object");
        return;
    }
    if (poSRS != nullptr)
        const_cast<OGRSpatialReference *>(poSRS)->Release();
    poSRS = poSRSIn;
    if (poSRS != nullptr)
        const_cast<OGRSpatialReference *>(poSRS)->Reference();
}
```

ogr/ogrsf_frmts/sqlite/ogrsqlitelayer.cpp
```diff
@@ -3555,11 +3553,165 @@ OGRErr OGRSQLiteLayer::CommitTransaction()
...
+            {
+                case FieldChangeType::DELETE:
+                {
+                    GetLayerDefn()->AddFieldDefn(
+                        oFieldChange.poFieldDefn.get());
+
+                    // Now move the field to the right place
+                    // from the last position to its original position
+                    const int iFieldCount = GetLayerDefn()->GetFieldCount();
+                    int *panOrder =
```
rouault, line 3598: I would use a std::vector<int> . Use of CPLMalloc/CPLCalloc in this driver are old-school habits

ogr/ogrsf_frmts/sqlite/ogrsqlitetablelayer.cpp
```diff
@@ -1710,6 +1717,13 @@ OGRSQLiteTableLayer::CreateGeomField(const OGRGeomFieldDefn *poGeomFieldIn,
         }
     }
 
+    if (m_poDS->IsInTransaction())
```
rouault, line 1720: There's a bit of dissymetry with the call to AddFieldDefn() in CreateField() having been done *before* the "if IsInTransation())" logic and here where AddGeomFieldDefn() is done *after* the IsInTransaction() test  (note that if you symmetrize things you need to add/remove a -1 index to GetFieldCount/GetGeomFieldCount)

ogr/ogrsf_frmts/sqlite/ogrsqlitetablelayer.cpp
```diff
@@ -2150,7 +2164,27 @@ OGRErr OGRSQLiteTableLayer::DeleteField(int iFieldToDelete)
         eErr = m_poDS->SoftCommitTransaction();
         if (eErr == OGRERR_NONE)
         {
-            eErr = m_poFeatureDefn->DeleteFieldDefn(iFieldToDelete);
+
+            if (m_poDS->IsInTransaction())
```
rouault, line 2168: I'm not sure the IsInTransaction() logic is appropriate here. The field was unconditionally added at line 1727

ogr/ogrsf_frmts/sqlite/ogrsqlitetablelayer.cpp
elpaso: the reason for the dissimetry is because poGeomField is (already was) a unique ptr whic would be null after the move. I will add a comment.

ogr/ogrsf_frmts/sqlite/ogrsqlitetablelayer.cpp
elpaso: It seems correct to me: if we are in a transaction we don't want the object to be destroyed immediately (because it would crash if someone is still accessing to it) but instead we steal it and keep it alive until either the layer is destroyed or another transaction is started.  

This is was I understood (or misunderstood) to be the expected behavior.

rouault: Disregard my comment. I misread the diff... (line 1727 is another method)

ogr/ogrgeomfielddefn.cpp
elpaso: done

ogr/ogrsf_frmts/sqlite/ogrsqlitelayer.cpp
elpaso: done

ogr/ogr_feature.h
```diff
@@ -643,8 +649,31 @@ class CPL_DLL OGRFeatureDefn
...
     virtual void AddFieldDefn(const OGRFieldDefn *);
     virtual OGRErr DeleteFieldDefn(int iField);
+
+    /**
+     * @brief StealFieldDefn takes ownership of the field definition at index detaching
+     *        it from the feature definition.
+     * This is a mechanism for the application to take over ownership of the field definition from the feature definition without copying.
+     * @param iField index of the field definition to detach.
+     * @return a pointer to the detached field definition or nullptr if the index is out of range.
+     * @note The caller is responsible for deleting the returned field definition.
```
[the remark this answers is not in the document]
elpaso: done

ogr/ogrsf_frmts/sqlite/ogrsqlitetablelayer.cpp
elpaso: comment added

rouault:
In the documentation of RollbackTransaction() (in ogr/ogrsf_frmts/ogrsf_frmts.dox), it should probably be mentioned that OGRFeature* instances acquired or created between the StartTransaction() and RollbackTransaction() should be destroyed before RollbackTransaction() if the field structure has been modified during the transaction.
In particular, the following is invalid:

```
lyr->StartTransaction();
lyr->DeleteField(...);
f = new OGRFeature(lyr->GetLayerDefn());
lyr->RollbackTransaction();
// f is in a inconsistent state at that point, given its array of fields doesn't match the updated layer definition, and thus it cannot even be safely deleted !
```
this must be corrected as:
```
lyr->StartTransaction();
lyr->DeleteField(...);
f = new OGRFeature(lyr->GetLayerDefn());
delete f;
lyr->RollbackTransaction();
```

ogr/ogrfielddefn.cpp
```diff
@@ -108,6 +108,63 @@ OGRFieldDefn::~OGRFieldDefn()
...
+        CPLFree(pszName);
+        pszName = CPLStrdup(oOther.pszName);
+        CPLFree(pszAlternativeName);
+        pszAlternativeName = CPLStrdup(oOther.pszAlternativeName);
+        eType = oOther.eType;
+        eJustify = oOther.eJustify;
+        nWidth = oOther.nWidth;
+        nPrecision = oOther.nPrecision;
+        CPLFree(pszDefault);
+        pszDefault = CPLStrdup(oOther.pszDefault);
```
rouault, line 155: Given that CPLStrdup(nullptr) return an empty string and not a nullptr:

```suggestion
        pszDefault = oOther.pszDefault ? CPLStrdup(oOther.pszDefault) : nullptr;
```

ogr/ogrfielddefn.cpp
```diff
@@ -108,6 +108,63 @@ OGRFieldDefn::~OGRFieldDefn()
...
+/**
+ * @brief OGRFieldDefn::OGRFieldDefn copy constructor.
+ * @param oOther the object to copy.
+ * @since GDAL 3.11
+ */
+OGRFieldDefn::OGRFieldDefn(const OGRFieldDefn &oOther)
+    : pszName(CPLStrdup(oOther.pszName)),
+      pszAlternativeName(CPLStrdup(oOther.pszAlternativeName)),
+      eType(oOther.eType), eJustify(oOther.eJustify), nWidth(oOther.nWidth),
+      nPrecision(oOther.nPrecision), pszDefault(CPLStrdup(oOther.pszDefault)),
```
rouault, line 124:
```suggestion
      nPrecision(oOther.nPrecision),
      pszDefault(oOther.pszDefault ? CPLStrdup(oOther.pszDefault) : nullptr),
```

ogr/ogrfielddefn.cpp
elpaso: done

ogr/ogrsf_frmts/ogrsf_frmts.h
```diff
@@ -395,6 +401,33 @@ class CPL_DLL OGRLayer : public GDALMajorObject
 
   protected:
     //! @cond Doxygen_Suppress
+
+    enum class FieldChangeType : char
+    {
+        ADD,
+        ALTER,
+        DELETE
```
rouault, line 409: It appears there must be a #define DELETE somewhere in a Windows system header...  So maybe switch to Add/Alter/Delete or add a _FIELD suffix ?

ogr/ogrsf_frmts/ogrsf_frmts.dox
```diff
@@ -3049,6 +3049,31 @@ form depending on the limitations of the format driver.
...
+
+ Instead, the feature should be destroyed before the rollback:
+
+ \code
+ lyr->StartTransaction();
+ lyr->DeleteField(...);
+ f = new OGRFeature(lyr->GetLayerDefn());
+ ...
+ delete f;
+ \encode
```
rouault, line 3075:
```suggestion
 \endcode
```

rouault:
macos_build CI failure unrelated and will be addressed per https://github.com/OSGeo/gdal/pull/11621

PR #11612, opened 2025-01-08
Add GDALExpandPackedBitsToByteAt0Or255() and GDALExpandPackedBitsToByteAt0Or1()

rouault:
and use them in GTiff and LIBERTIFF drivers


hobu:
Does this really need to be in the public API? I get the convenience angle, but it is hyper specific, and named in such a way that makes it difficult to know why I might use or not use it as a public API consumer.

rouault:
> Does this really need to be in the public API?

good point. Prototypes moved from gdal.h to gdal_priv.h

PR #11614, opened 2025-01-08
Add 'gdal raster overview add' and 'gdal raster overview delete'

rouault:
Docs previews: https://gdal--11614.org.readthedocs.build/en/11614/programs/gdal_raster_overview_add.html and https://gdal--11614.org.readthedocs.build/en/11614/programs/gdal_raster_overview_delete.html

### DIFF
--- a/apps/CMakeLists.txt
+++ b/apps/CMakeLists.txt
@@ -16,6 +16,8 @@ add_library(
   gdalalg_raster_convert.cpp
   gdalalg_raster_edit.cpp
   gdalalg_raster_pipeline.cpp
+  gdalalg_raster_overview_add.cpp
+  gdalalg_raster_overview_delete.cpp
   gdalalg_raster_read.cpp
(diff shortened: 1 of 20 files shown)


jratike80:
> Thoughts ?
- I have always been thinking that `-ro` in gdaladdo is a programmer's view, but for a user it means external overviews. Would it be dangerous to have `gdal raster overview add --external` instead, even it is not explicit - overviews can be external for non-tiff formats anyway.
- What is the meaning of `--read-only` with `gdal raster overview delete`? To delete the overviews if they are external, but not if they are internal, or?
- "Integral overview levels" feel somehow odd but my English is not so good that I could tell why.

rouault:
@jratike80 Thanks for the feedbak. "--read-only" renamed to "--external" to both "add" and "remove".   "Integral..." rephrased to "A list of overview levels to build. Each overview level must be an integer value greater or equal to 2."

PR #11615, opened 2025-01-08
Explicitly set CMAKE_CXX_SCAN_FOR_MODULES=0

hobu:
Something about my configuration was causing the following complaint

> CMAKE_CXX_COMPILER_CLANG_SCAN_DEPS-NOTFOUND 

Some searching showed https://discourse.cmake.org/t/cmake-3-28-cmake-cxx-compiler-clang-scan-deps-notfound-not-found/9244/3 

My configuration didn't have `clang-tools` installed, so I didn't have `clang-scan-deps` available. I shouldn't have *needed* it, but I don't understand why our CMake configuration thinks it does.

This PR tells CMake that it doesn't need it. When we upgrade to C++20 modules 😄, we can switch it back on.

### DIFF
--- a/cmake/helpers/GdalCAndCXXStandards.cmake
+++ b/cmake/helpers/GdalCAndCXXStandards.cmake
@@ -8,3 +8,8 @@ if (NOT CMAKE_C_STANDARD)
     set(CMAKE_C_STANDARD 99)
     set(CMAKE_C_STANDARD_REQUIRED ON)
 endif()
+
+# explicitly tell CMake not to do module
+# dependency scanning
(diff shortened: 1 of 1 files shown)


rouault:
Looks reasonable. By chance do we known if this has been fixed in the latest CMake versions?  If so, we could possibly add a test on the CMake version (although that might bring little benefits)

hobu:
```
cmake version 3.31.2
```

I'm using a pretty current version. I think there's something about our dependency chain causing this check to be triggered. That chain is so large there's probably no way to prevent it.

PR #11616, opened 2025-01-09
Add a data/gdal_algorithm.schema.json to validate 'gdal [command] [subcommand] --json-usage' output

rouault:


hobu:
Sorry I'm not tracking so closely. What do all of the CLI schema documents do? Are they for users to validate things? Is there documentation of how to use them?

rouault:
> Are they for users to validate things? Is there documentation of how to use them?

yes, mostly as a documentation to users of our output, and a way for ourselves to crosscheck we generate expected output.

PR #11623, opened 2025-01-09
VRT processed dataset: unscale source raster values

dbaston:
Alternative to #11573 . Introduces an attribute `scale` to `<Input>` with default value `auto`. This will cause input datasets with a defined scale or offset to be unscaled to Float64 (also causing output bands to be Float64). `<Input unscale="false">` prevents this and keeps the current behavior. `<Input scale="true">` invokes the unscaling machinery regardless of the defined scale/offset values in the input dataset, so that the output type is always consistent.

~~I don't love the implementation which ends up touching `VRTProcessedDataset` in several places. I think a cleaner implementation could involve creating a method like `GDALRasterBand* GDALRasterBand::GetUnscaled()`. If the band was not scaled it could return a reference to itself; for a band with scaling it could return a a wrapper `GDALScaledRasterBand` with appropriate overloads of `IRasterIO`, `GetNoDataValue`, etc. This would also provide a consolidated location for an optimized (SSE?) implementation of the unscaling.~~

cc @jratike80

### DIFF
--- a/autotest/gdrivers/data/vrt/processed_OutputBands_FROM_LAST_STEP.vrt
+++ b/autotest/gdrivers/data/vrt/processed_OutputBands_FROM_LAST_STEP.vrt
@@ -1,5 +1,5 @@
 <VRTDataset subClass="VRTProcessedDataset">
-  <Input>
+  <Input unscale="AUTO">
     <SourceFilename relativeToVRT="1">../byte.tif</SourceFilename>
   </Input>
   <OutputBands count="FROM_LAST_STEP" dataType="FROM_LAST_STEP"/>
--- a/autotest/gdrivers/vrtprocesseddataset.py
+++ b/autotest/gdrivers/vrtprocesseddataset.py
@@ -10,10 +10,14 @@
 # SPDX-License-Identifier: MIT
 ###############################################################################
 
+import os
+
 import gdaltest
 import pytest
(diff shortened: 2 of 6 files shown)


rouault:
> which ends up touching `VRTProcessedDataset` in several places

indeed. I thought you were going to silently add the unscale processing functions as a first step. Wouldn't that simplify things ?

> I think a cleaner implementation could involve creating a method like GDALRasterBand* GDALRasterBand::GetUnscaled()

or just use the dataset returned by GDALTranslate("", hSrcDS, options = "-f VRT -unscale") which will handle all details, including reading from overviews.

dbaston:
> Wouldn't that simplify things ?

In some ways. But there's still a problem where the `BandScaleOffset` algorithm can't set the scale/offset properties on the output, so you would get different behavior if you include it as an explicit processing step vs implicitly via `<Input unscale="true">`. This seems like it would be confusing.

> or just use the dataset returned by GDALTranslate("", hSrcDS, options = "-f VRT -unscale") 

I'll give this a try.

rouault:
macos_build failure unrelated, and should be solved by rebasing on top of latest master

rouault:
@dbaston Please also edit frmts/vrt/data/gdalvrt.xsd to add the new attribute

PR #11625, opened 2025-01-10
cpl_error.h: Move the include of stdbool for CPLErrorOnce into the header.

schwehr:
Callers of things in cpl_error.h should not have to be aware of internally required includes.

Minor cleanup from https://github.com/OSGeo/gdal/commit/ad15a8e141c4841ef1696ad88d0ea55272c82d01

### DIFF
--- a/port/cpl_error.h
+++ b/port/cpl_error.h
@@ -17,9 +17,7 @@
 #include "cpl_port.h"
 
 #include <stdarg.h>
-CPL_C_START
 #include <stdbool.h>
-CPL_C_END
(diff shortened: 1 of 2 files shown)


port/cpl_error.h
```diff
@@ -17,6 +17,9 @@
 #include "cpl_port.h"
 
 #include <stdarg.h>
+CPL_C_START
+#include <stdbool.h>
+CPL_C_END
```
rouault, line 22:
```suggestion
#include <stdbool.h>
```

PR #11626, opened 2025-01-10
Remove uses of #ifdef CPL_HAS_GINT64

rouault:
The code base has assumed for many many years that we have access to 64-bit integer types.


schwehr:
Thanks for the great cleanup!

PR #11640, opened 2025-01-12
Add RFC 105 text: Add and use safe path manipulation functions

rouault:
[Rendered view](https://gdal--11640.org.readthedocs.build/en/11640/development/rfc/rfc105_safe_path_manipulation_functions.html)

### DIFF
--- a/doc/source/development/rfc/rfc105_safe_path_manipulation_functions.rst
+++ b/doc/source/development/rfc/rfc105_safe_path_manipulation_functions.rst
@@ -26,7 +26,7 @@ Motivation
 ----------
 
 Above mentioned functions return a ``const char *`` string, which is owned by
-a thread-locale rotating buffer of 10 strings of up to 2048 bytes each, managed
+a thread-local rotating buffer of 10 strings of up to 2048 bytes each, managed
 by the `CPLGetStaticResult() <https://github.com/OSGeo/gdal/blob/ea26bd087b3e34b91ef0315ca3889f39445f2e1f/port/cpl_path.cpp#L53>`__
 function of port/cpl_path.cpp
 
@@ -79,7 +79,7 @@ oss-fuzz, and probably latent ones still uncovered.
 The later in date is https://issues.oss-fuzz.com/issues/388868487 , fixed
 by https://github.com/OSGeo/gdal/pull/11638 , and shows how subtle those bugs
 can be. In that instance, the issue is due to complex chains of calls where we
-end up using more than 10 instances of the thread-locale storage.
+end up using more than 10 instances of the thread-local storage.
 
 Implementation
 --------------
@@ -89,12 +89,12 @@ systematic approach to suppress their root case:
 
 - Start from the implementation of the CPLXXXX() functions, copy them to
   C++ CPLXXXXSafe() functions that return a ``std::string`` instead of a ``const char *``,
-  and modify them to no longer use thread-locale rotating buffer of strings,
+  and modify them to no longer use thread-local rotating buffer of strings,
   but rely on std::string natural concatenation routines instead
   of the convoluted C concatenation logic, resulting in clearer code.
 
 - make the now superseded CPLXXXX() functions call the Safe versions, and store
-  the result in the thread-locale rotating buffer of strings. So those functions
+  the result in the thread-local rotating buffer of strings. So those functions
   will mostly behave as their current implementation
 
   That is a simple as:
@@ -186,7 +186,7 @@ Out-of-scope (for the candidate implementation related to this RFC)
 -------------------------------------------------------------------
 
 We have exactly the same issue with ``const char* CPLSPrintf(const char* pszFmt, ....)``
-which uses its own thread-locale rotating buffer of 10 strings of size up to
+which uses its own thread-local rotating buffer of 10 strings of size up to
 8000 bytes each.
 
 The safer alternative exists as ``CPLString::Printf()`` and it would be worth
--- a/doc/source/development/rfc/rfc105_safe_path_manipulation_functions.rst
+++ b/doc/source/development/rfc/rfc105_safe_path_manipulation_functions.rst
@@ -8,7 +8,7 @@ RFC 105: Add and use safe path manipulation functions
 Author:        Even Rouault
 Contact:       even.rouault @ spatialys.com
 Started:       2025-01-11
-Status:        Adopte, implemented in 3.11
+Status:        Adopted, implemented in 3.11
 Target:        GDAL 3.11
 ============== =============================================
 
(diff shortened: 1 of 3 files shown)


doc/source/development/rfc/rfc105_safe_path_manipulation_functions.rst
```diff
@@ -0,0 +1,244 @@
...
+CPLGetFormFilename(), CPLGetFormCIFilename(), etc.), that returns a result stored
+in more or less ephemeral storage, to avoid potential security issues related
+to their mis-use. It also covers converting most of the code base to the safer
+alternatives.
+
+Motivation
+----------
+
+Above mentioned functions return a ``const char *`` string, which is owned by
+a thread-locale rotating buffer of 10 strings of up to 2048 bytes each, managed
```
lnicola, line 29: Nit:

```suggestion
a thread-local rotating buffer of 10 strings of up to 2048 bytes each, managed
```

doc/source/development/rfc/rfc105_safe_path_manipulation_functions.rst
```diff
@@ -0,0 +1,244 @@
...
+        do_something_else(osExt.c_str());
+
+
+However, experience has shown that it is extremely easy to forget that, and over
+the years, this has caused several heap buffer overflows bugs detected by
+oss-fuzz, and probably latent ones still uncovered.
+The later in date is https://issues.oss-fuzz.com/issues/388868487 , fixed
+by https://github.com/OSGeo/gdal/pull/11638 , and shows how subtle those bugs
+can be. In that instance, the issue is due to complex chains of calls where we
+end up using more than 10 instances of the thread-locale storage.
```
lnicola, line 82:
```suggestion
end up using more than 10 instances of the thread-local storage.
```

doc/source/development/rfc/rfc105_safe_path_manipulation_functions.rst
```diff
@@ -0,0 +1,244 @@
...
+
+Implementation
+--------------
+
+Rather to react on a case by case when fuzzers found issues, we want a
+systematic approach to suppress their root case:
+
+- Start from the implementation of the CPLXXXX() functions, copy them to
+  C++ CPLXXXXSafe() functions that return a ``std::string`` instead of a ``const char *``,
+  and modify them to no longer use thread-locale rotating buffer of strings,
```
lnicola, line 92:
```suggestion
  and modify them to no longer use thread-local rotating buffer of strings,
```

doc/source/development/rfc/rfc105_safe_path_manipulation_functions.rst
```diff
@@ -0,0 +1,244 @@
...
+  but rely on std::string natural concatenation routines instead
+  of the convoluted C concatenation logic, resulting in clearer code.
+
+- make the now superseded CPLXXXX() functions call the Safe versions, and store
+  the result in the thread-locale rotating buffer of strings. So those functions
```
lnicola, line 97:
```suggestion
  the result in the thread-local rotating buffer of strings. So those functions
```

doc/source/development/rfc/rfc105_safe_path_manipulation_functions.rst
```diff
@@ -0,0 +1,244 @@
...
+of drivers from the need of calling CPLGetExtensionSafe(), as well as a
+convenience `GDALOpenInfo::IsExtensionEqualToCI(const char* pszExtension)` method.
+So, all in all, the performance impact of those changes is thought/hoped to be
+in the hardly measurable category.
+
+Out-of-scope (for the candidate implementation related to this RFC)
+-------------------------------------------------------------------
+
+We have exactly the same issue with ``const char* CPLSPrintf(const char* pszFmt, ....)``
+which uses its own thread-locale rotating buffer of 10 strings of size up to
```
lnicola, line 189:
```suggestion
which uses its own thread-local rotating buffer of 10 strings of size up to
```

jjimenezshaw:
Thank you @rouault . Great analysis.

lnicola:
Are there plans to require C++17 "soon"? It has `std::filesystem`, which might be an alternative. There's also the question of non-ASCII file paths, but it's better left unasked (and I don't think this RFC makes support for them worse).

LGTM otherwise.

jjimenezshaw:
> The safe moniker implies something scary about the older usage.

That's the idea of this RFC, right?

rouault:
> I also would like to get to a point where these CPL functions are simple wrappers of std::filesystem, but there are likely to be platform peculiarities that CPL is pasting over in ways that are not compatible.

Indeed... How std::filesystem on MSVC would work with /vsi paths... ?

> We should at least attempt to wrap std::filesystem implementations for all the CPL path handling stuff and run the test suite and see how much actually falls out

Even doing the "dumb" replacement I suggest in the RFC is going to be very painful, because there are several hundreds of locations where the sed magic isn't powerful enough to automate it and that will require hand editing. I'll let std::filesystem exploration to someone braver than me :-)

> before we build another big chunk of code that we have to maintain going forward...

The change of the CPLxxxx() functions to return std::string and no longer use C concatenation function was the easy part and it is not that much (300 hundred lines of code maybe?)

> The safe moniker implies something scary about the older usage.

Well, they are somewhat scary to use, so better have people ran away from them :-) Joke aside, I'm fine with the "2" is there's a consensus it is better than "Safe". I also considerd a "Cpp" suffix to indicate they are for use by C++ code, but wasn't convinced. We have some precedent of using "2" suffix.

doc/source/development/rfc/rfc105_safe_path_manipulation_functions.rst
```diff
@@ -0,0 +1,244 @@
...
+Status:        Draft
+Target:        GDAL 3.11
+============== =============================================
+
+Summary
+-------
+
+This RFC adds safe versions, for use by C++ code, of all functions of
+cpl_path.cpp (such as CPLGetPath(), CPLGetDirname(), CPLGetBasename(), CPLGetExtension(),
+CPLGetFormFilename(), CPLGetFormCIFilename(), etc.), that returns a result stored
```
mdsumner, line 20: ? should be CPLFormFilename() and CPLFormCIFilename() (no "Get")

rouault, line 20:
```suggestion
CPLFormFilename(), CPLFormCIFilename(), etc.), that returns a result stored
```

rouault: indeed, thanks! Fixed

doc/source/development/rfc/rfc105_safe_path_manipulation_functions.rst
```diff
@@ -0,0 +1,217 @@
...
+
+===================================================================
+RFC 105: Add and use safe path manipulation functions
+===================================================================
+
+============== =============================================
+Author:        Even Rouault
+Contact:       even.rouault @ spatialys.com
+Started:       2025-01-11
+Status:        Adopte, implemented in 3.11
```
rouault, line 11:
```suggestion
Status:        Adopted, implemented in 3.11
```

PR #11646, opened 2025-01-13
Avoid clang warning in -fsanitize=undefined -O mode (master only)

### LINKED ISSUE #11645: clang loop optimizer pragma warns on rasterio.cpp.
### What is the bug?

[rasterio.cpp](https://github.com/OSGeo/gdal/blob/493ed18ef8ad85813501573043bb9e1053d0f47f/gcore/rasterio.cpp#L5380C2-L5380C14) triggers a complaint with UBSAN mode with recent clang (like in the last couple weeks), but as this is a custom environment, I don't know if this happens with older clangs. I've seen `pass-failed` happen a few years ago with non-gdal code.

```
#if defined(__clang__)
#pragma clang loop vectorize(disable)
#endif
```

``` 
gcore/rasterio.cpp:5954:5: error: loop not vectorized: the optimizer was unable to perform the requested transformation; the transformation might be disabled or specified as part of an unsupported transformation ordering [-Werror,-Wpass-failed=transform-warning]
 5954 |     for (size_t i = 0; i < nIters; ++i)
      |     ^
gcore/rasterio.cpp:6049:5: error: loop not vectorized: the optimizer was unable to perform the requested transformation; the transformation might be disabled or specified as part of an unsupported transformation ordering [-Werror,-Wpass-failed=transform-warning]
 6049 |     for (size_t i = 0; i < nIters; ++i)
      |     ^
2 errors generated.
```

Not sure why the loop doesn't unroll.

I don't have great details on how to reproduce. Sorry.

### Versions and provenance

Hit this GDAL from head on 2025-01-10.

### Additional context

Not sure the best way to disable this in the GDAL code base and is that reasonable to do? Locally, I set `-Wno-pass-failed` for now and then the UBSAN run goes just fine.

I looked at CPL_NOSANITIZE_UNSIGNED_INT_OVERFLOW and maybe this is correct `__attribute__((no_sanitize_undefined))`?

I will investigate more.

rouault:
Fixes #11645

### DIFF
--- a/gcore/rasterio.cpp
+++ b/gcore/rasterio.cpp
@@ -5944,6 +5944,11 @@ __attribute__((optimize("tree-vectorize")))
 #if defined(__GNUC__)
 __attribute__((noinline))
 #endif
+#if defined(__clang__) && !defined(__INTEL_CLANG_COMPILER)
+// clang++ -O2 -fsanitize=undefined fails to vectorize, ignore that warning
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wpass-failed"
+#endif
 static void
 GDALInterleave2Byte(const uint8_t *CPL_RESTRICT pSrc,
                     uint8_t *CPL_RESTRICT pDst, size_t nIters)
@@ -5957,6 +5962,9 @@ GDALInterleave2Byte(const uint8_t *CPL_RESTRICT pSrc,
         pDst[2 * i + 1] = pSrc[i + 1 * nIters];
     }
 }
+#if defined(__clang__) && !defined(__INTEL_CLANG_COMPILER)
+#pragma clang diagnostic pop
+#endif
 
 #endif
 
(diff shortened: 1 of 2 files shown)


schwehr:
I copied it over and it works for me. Thanks!

schwehr (approved):
LGTM

rouault:
I had to revert the commit enable the ``-O`` level since it caused runtime issues like:
```
/usr/bin/../lib/gcc/x86_64-linux-gnu/13/../../../../include/c++/13/bits/basic_string.h:490:51: runtime error: unsigned integer overflow: 19 - 24 cannot be represented in type 'size_type' (aka 'unsigned long')
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /usr/bin/../lib/gcc/x86_64-linux-gnu/13/../../../../include/c++/13/bits/basic_string.h:490:51 
```

schwehr:
Should be fine without the `-O`. Thanks!

PR #10341, opened 2024-06-29
ogr2ogr: speed-up -clipsrc/-clipdst, and add -skipinvalid

rouault:
-   ogr2ogr: speed-up -clipsrc/-clipdst by avoiding GEOS when possible
    
    If the envelope of the feature's geometry is contained in the envelope
    of the -clipsrc/-clipdst geometry, we can just avoid any GEOS operation.

- ogr2ogr: add -skipinvalid to skip features whose geometry is not valid w.r.t Simple Features

Issue #11652, opened 2025-01-14
ogr2ogr clipsrc uses bounding box instead of shape


Thorwyn101:
### What is the bug?

I want to clip the content of a shapefile to a given polygon. But instead of the real shape, only the polygon's bounding box is used (the polygon is larger than the given features in northeast direction):

![image](https://github.com/user-attachments/assets/1dbd8e5d-459e-4f88-80c4-369fba11e78f)

GDAL 3.9.3 (released 2024/10/07) yields the desired output:

![image](https://github.com/user-attachments/assets/4cc4c260-f606-4e42-bdbe-a6c9e3239a39)

### Steps to reproduce the issue

I tried the polygon as WKT and Geopackage/ESRI Shapefile. EPSG 4326 is set for input shapefile and WKT. Following code is used:

`ogr2ogr.exe -lco ENCODING=UTF-8 -clipsrc "POLYGON((6.33 49.95, ...))" output.shp input.shp`

`ogr2ogr.exe -lco ENCODING=UTF-8 -clipsrc shape.gpkg output.shp input.shp`

In both cases the features are clipped to the bounding box of the polygon.

### Versions and provenance

Windows 11, GDAL 3.10.0, released 2024/11/01

jratike80:
Please provide copy-pastable commands by using for example Natural Earth vectors as test data.

jratike80:
I had a try with GDAL 3.11.0dev-c6bedc3cd4, released 2024/12/20 and ne_50m_admin_0_countries.shp with commands:

```
ogr2ogr -clipsrc "POLYGON (( -0.236 15.059, 24.98 23.792, 27.719 4.778, 12.867 2.247, -0.236 15.059 ))" test.shp ne_50m_admin_0_countries.shp

ogr2ogr -clipdst "POLYGON (( -0.236 15.059, 24.98 23.792, 27.719 4.778, 12.867 2.247, -0.236 15.059 ))" test2.shp ne_50m_admin_0_countries.shp
```

I cannot reproduce your observation, but I wonder why some countries get selected, not clipped.

<img width="222" alt="image" src="https://github.com/user-attachments/assets/188166fd-75d4-4263-85f4-2b6d47ba4ca1" />

Thorwyn101:
Data: ne_10m_minor_islands.shp (red areas)
Area: south west of Finland

[...]

PR #11655, opened 2025-01-15
ogr2ogr: fix -clipsrc/-clipdst when a input geometry is within the envelope of a non-rectangular clip geometry, but doesn't intersect it (3.10.0 regression)

rouault:
Fixes #10341

This fixes commit 388d3ae7e47cba64e1050023bed43c4b04fe4277 (PR #10341) "ogr2ogr: speed-up -clipsrc/-clipdst by avoiding GEOS when possible", which only worked if the clipping geometry was a rectangle.


agiudiceandrea:
Fixes #11652.

PR #11667, opened 2025-01-16
autotest: update tests for gdal_calc

dbaston:
- simplify generation of temporary file names, since they'll be cleaned up automatically
- separate independent tests into separate functions
- add tests for untested arguments
- avoid copying input dataset for every test function
- use vsimem where possible

### DIFF
--- a/autotest/pyscripts/test_gdal_calc.py
+++ b/autotest/pyscripts/test_gdal_calc.py
@@ -129,8 +129,8 @@ def test_gdal_calc_py_1(script_path, tmp_path, stefan_full_rgba):
         f"-Z {infile} --Z_band=2 --calc=Z --overwrite --format GTiff --outfile {out[2]}",
     )
 
-    for i, checksum in zip(
-        range(test_count), (input_checksum[0], input_checksum[1], input_checksum[1])
+    for i, checksum in enumerate(
+        (input_checksum[0], input_checksum[1], input_checksum[1])
     ):
         check_file(out[i], checksum, i + 1)
 
@@ -182,7 +182,7 @@ def test_gdal_calc_py_2(script_path, tmp_path, stefan_full_rgba):
         f"--overwrite --outfile {out[2]}",
     )
 
-    for i, checksum in zip(range(test_count), (12368, 62785, 47132)):
+    for i, checksum in enumerate((12368, 62785, 47132)):
         check_file(out[i], checksum, i + 1)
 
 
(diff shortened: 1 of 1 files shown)


dbaston:
Should the `@optfile` syntax work in Windows CMD? The way the test was written previously, it passed even though the call to `gdal_calc` failed. I'm trying to figure out if it should succeed. When I call gdal_calc from CMD (outside of the test framework), it appears to strip all of the `\` characters from the optfile, causing gdal_calc to fail because it can't find the input dataset.

rouault:
> Should the `@optfile` syntax work in Windows CMD?

I've no idea. If that's an annoyance, just skip the test on Windows

jratike80:
> Should the @optfile syntax work in Windows CMD?

```
gdal_calc @optfile.txt
```
Maybe the optfile itself does not work on Windows? Like having \ as newlines, while it maybe should be ^.

dbaston:
Here are the optfile contents:

```
-A C:\Users\dbaston\AppData\Local\Temp\pytest-of-dbaston\pytest-5\gdal_calc0\stefan_full_rgba.tif --calc=A --overwrite --outfile C:\Users\dbaston\AppData\Local\Temp\pytest-of-dbaston\pytest-5\test_gdal_calc_py_7a___0\out.tif
```

Calling `gdal_calc` with `--optfile` is successful:

```
C:\dev\gdal\build>python swig\python\osgeo_utils\gdal_calc.py --optfile C:\Users\dbaston\AppData\Local\Temp\pytest-of-dbaston\pytest-5\test_gdal_calc_py_7a___0\opt
```

Calling `gdal_calc` with `@opt` fails -- the `\` characters appear to be stripped on reading:
```
C:\dev\gdal\build>python swig\python\osgeo_utils\gdal_calc.py @C:\Users\dbaston\AppData\Local\Temp\pytest-of-dbaston\pytest-5\test_gdal_calc_py_7a___0\opt
Could not open file "C:UsersdbastonAppDataLocalTemppytest-of-dbastonpytest-5gdal_calc0stefan_full_rgba.tif": No such file or directory
```

After editing the optfile to replace `\` with `/`, it can be used with `@`:
```
C:\dev\gdal\build>python swig\python\osgeo_utils\gdal_calc.py @C:\Users\dbaston\AppData\Local\Temp\pytest-of-dbaston\pytest-5\test_gdal_calc_py_7a___0\opt2.txt
```

I'm going to skip the test on Windows for now.

PR #11673, opened 2025-01-16
[gdal_contour] Do not include min/max when polygonize

### LINKED ISSUE #11564: gdal_contour generates polygon for value not specified in fixed levels
### What is the bug?

When using gdal_contour and specifying fixed levels to contour, it seems to output the lowest value in the raster data regardless of the fixed levels specified in the command.

Using the supplied grib2 file (unzip it first), run the following command

`gdal_contour -p -amin value -fl 20 22 22 24 26 28 30 32 34 36 38 40 44 48 52 56 60 64 68 72 76 80 ./base-ref.grib2 ./base-ref.json`

The output JSON will have all the fixed levels, but will contour the lowest value in the grib file (-10) which drastically bloats the output.

[base-ref.grib2.zip](https://github.com/user-attachments/files/18286248/base-ref.grib2.zip)

### Versions and provenance

Mac OS: 15.0
GDAL 3.10.0, released 2024/11/01

### Additional context

_No response_

elpaso:
Fixes #11564 by not including raster min/max when polygonize and fixed levels are specified.

MIN and MAX literals can be used to specify raster min/max values, e.g.:

-fl MIN 10 20 30 MAX

### DIFF
--- a/alg/contour.cpp
+++ b/alg/contour.cpp
@@ -517,7 +517,9 @@ an averaged value from the two nearby points (in this case (12+3+5)/3).
  *   FIXED_LEVELS=f[,f]*
  *
  * The list of fixed contour levels at which contours should be generated.
- * This option has precedence on LEVEL_INTERVAL
+ * This option has precedence on LEVEL_INTERVAL. MIN and MAX can be used
+ * as special values to represent the minimum and maximum values of the
+ * raster.
  *
  *   NODATA=f
  *
--- a/alg/contour.cpp
+++ b/alg/contour.cpp
@@ -141,8 +141,11 @@ struct PolygonContourWriter
             currentGeometry_->addGeometryDirectly(currentPart_);
         }
 
-        OGRPolygonContourWriter(previousLevel_, currentLevel_,
-                                *currentGeometry_, poInfo_);
+        if (currentGeometry_->getNumGeometries() > 0)
+        {
+            OGRPolygonContourWriter(previousLevel_, currentLevel_,
+                                    *currentGeometry_, poInfo_);
+        }
 
         currentGeometry_.reset(nullptr);
         currentPart_ = nullptr;
@@ -607,7 +610,19 @@ CPLErr GDALContourGenerateEx(GDALRasterBandH hBand, void *hLayer,
         fixedLevels.resize(aosLevels.size());
         for (size_t i = 0; i < fixedLevels.size(); i++)
         {
-            fixedLevels[i] = CPLAtof(aosLevels[i]);
+            // Handle min/max values
+            if (EQUAL(aosLevels[i], "MIN"))
+            {
+                fixedLevels[i] = std::numeric_limits<double>::lowest();
+            }
+            else if (EQUAL(aosLevels[i], "MAX"))
+            {
+                fixedLevels[i] = std::numeric_limits<double>::max();
+            }
+            else
+            {
+                fixedLevels[i] = CPLAtof(aosLevels[i]);
+            }
             if (i > 0 && !(fixedLevels[i] >= fixedLevels[i - 1]))
             {
                 CPLError(CE_Failure, CPLE_AppDefined,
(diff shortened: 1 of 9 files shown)


doc/source/programs/gdal_contour.rst
```diff
@@ -146,12 +155,111 @@ Examples
...
+        yllcorner    0
+        cellsize     1
+        4 15
+        25 36
+
+        $ gdal_contour test.asc -f GeoJSON /vsistdout/ -i 10 -fl 15 -p -amin min -amax max
+
+    This would create 10-meter polygonal contours from the DEM data in :file:`test.asc` and produce a GeoJSON output
+    with the contour min and max elevations in the ``min`` and ``max`` attributes, the values of these fields will
+    be: (4.0, 10.0), (10, 15.0), (15, 20.0)  and (20, 36.0).
```
jratike80, line 264: Shoud it be (10, 15.0), (15, 20.0)  and (20, 30.0)?

I think I understand what happens, but not how to describe it so that a newcomer understands. Perhaps something like
"Creates contours at regular 10 meter intervals and adds extra contour for a fixed 15 m level. Finally turns areas between the contours into polygons."

jratike80: For those values, shouldn't user define `-i 10 -fl MIN 15 MAX -p`? Otherwise it would not match with the title...

jratike80: Maybe a bit different approach would have been easier to explain. First improve documentation of -p

> "Generate contour polygons rather than contour lines. The min and max values of the raster will be used as the lower and upper limit for the data range of the first and last polygon, respectively.

Then introduce `exlude_min` and `exclude_max` for preventing the default behavior so that it would have similar effect on all `-i` and `-fl` combinations, when used together with `-p`, and no effect without `-p`.

elpaso: > For those values, shouldn't user define `-i 10 -fl MIN 15 MAX -p`? Otherwise it would not match with the title. Or do regular intervals start/stop at min/max automatically? It does make sense, most users want to get the polygon between the sea and 10 m level automatically, but it is not clear from the documentation.

yes, regular intervals get min/max automatically. I didn't change that.

doc/source/programs/gdal_contour.rst
```diff
@@ -122,6 +122,15 @@ be on the right, i.e. a line string goes clockwise around a top.
...
     Generate contour polygons rather than contour lines.
 
+    When this mode is selected the polygons are created for values between
+    each level specified by :option:`-i` or :option:`-fl` or :option:`-e`,
+    in case :option:`-fl` is used alone at least two fixed levels must be specified.
+
+    The minimum and maximum values from the raster are not automatically added to
+    the fixed levels list but the special values `MIN`` and `MAX`` (case insensitive)
+    can be used to include them.
+
```
jratike80, line 132: ... but when option `-i` is used together with `-p`, the minimum and maximum values from the raster ARE automatically added, and there is no way to get rid of them.

I know this is not a common use case, but there can be lot of writing to the `-fl` if the DEM has a large height range.

doc/source/programs/gdal_contour.rst
```diff
@@ -146,12 +155,112 @@ Examples
...
+        cellsize     1
+        4 15
+        25 36
+
+        $ gdal_contour test.asc -f GeoJSON /vsistdout/ -i 10 -p -amin min -amax max
+
+    This would create 10-meter polygonal contours from the DEM data in :file:`test.asc`
+    and produce a GeoJSON output with the contour min and max elevations in the ``min``
+    and ``max`` attributes, not including the minimum and maximum values from the raster.
+
```
jratike80, line 189: I do not understand. If there is only `-i`, doesn't it mean, that min/max are used just like before the change?
Or maybe I just need another cup of coffee.

elpaso: You a re right:  I removed the `not`.

alg/contour.cpp
```diff
@@ -607,7 +610,19 @@ CPLErr GDALContourGenerateEx(GDALRasterBandH hBand, void *hLayer,
         fixedLevels.resize(aosLevels.size());
         for (size_t i = 0; i < fixedLevels.size(); i++)
         {
-            fixedLevels[i] = CPLAtof(aosLevels[i]);
+            // Handle min/max values
+            if (EQUAL(aosLevels[i], "MIN"))
```
rouault, line 614: The Doxygen for that option (lines 517-519) should also mention "MIN" and "MAX" as valid values.

elpaso: done

PR #11681, opened 2025-01-17
Add 'gdal vector clip' and 'gdal vector pipeline read ... ! clip ... ! write ...'

rouault:


rouault:
@sgillies @jratike80  Thoughts ?  Cf https://gdal--11681.org.readthedocs.build/en/11681/programs/gdal_vector_clip.html for the options. I've tried to be as much consistent as possible with the raster clip (https://gdal--11681.org.readthedocs.build/en/11681/programs/gdal_raster_clip.html)  although the later only currently accepts raster 'like' datasets, whereas for this vector clip I accept both raster and vector. For raster, accepting a vector like could imply doing more than just rectangular clipping, and potentially something similar to gdalwarp -cutline

jratike80:
> Thoughts?

I would add a couple of words somewhere and explain that clip often yields mixed geometry types. And add `-nlt GEOMETRY` into the GeoPackage example. Do I remember right, that `-nlt` then affects all output layers? Alternatively there could be an option to filter the output by geometry type, but it might become a black box that user cannot control and understand.

> If the specified dataset is a vector dataset, its polygonal geometries are unioned together to form the clipping geometry.

Idea is OK, but polygon layers tend to be dirty with gaps and overlaps between individual polygon features.  The clipped result may be surprising. Users should be warned.

> For raster, accepting a vector like could imply doing more than just rectangular clipping, and potentially something similar to gdalwarp -cutline

Maybe also similar to gdal_rasterize -i for clipping holes. They would be nice options to have in the future. For the raster --like, maybe mention that bounds which are taken from the template image are rectangular, and bounds include nodata.

rouault:
Thanks @jratike80 for the feedback

> I would add a couple of words somewhere and explain that clip often yields mixed geometry types. And add -nlt GEOMETRY into the GeoPackage example

Some words added (as well as improving the code). Currently we don't have any equivalent of -nlt with "gdal" utility. We probably need some "geom_op" step to "gdal vector pipeline" to do such thing, and other operations like buffering, densifying, etc.  Or a distinct step per geometry operation... (although I'm a bit afraid of the doc and code overhead of having one step per type of geometry operation)

> Idea is OK, but polygon layers tend to be dirty with gaps and overlaps between individual polygon features. The clipped result may be surprising

That's what we do already with ogr2ogr. Warning added

dbaston:
What do you think about accepting `--xmin` `--xmax` etc. as an alternative to `--bbox` ? I'd be willing to implement it.

rouault:
> What do you think about accepting `--xmin` `--xmax` etc. as an alternative to `--bbox` ?

Is there a strong reason for offering that alternative ?  (we should try to avoid alternate syntaxes as much as possible)

dbaston:
Unlike `--bbox`, I can look at it and immediately understand what the numbers mean.

rouault:
> Unlike `--bbox`, I can look at it and immediately understand what the numbers mean.

I'm not super convinced we need that, unless we have several concordant opinions that it is needed.
Even X and Y terminology is rather ambiguous. If you are a CRS standard obedient, X might mean easting or no(r)thing depending on the coordinate system definition of the CRS. Here we made the choice of the godless CRS practitionners by sticking to "GIS friendly" convention for X and Y.
(One could argue that our bbox definition is also not super faithful to GeoJSON as we don't support anti-meridian crossing situations where [170,-10,-170,10] would mean square of 20x20 degree centered around (180,0) )

dbaston:
> I'm not super convinced we need that, unless we have several concordant opinions that it is needed.

That's fine, hence my question. I'm just scarred from recently writing some code for GEOS, where bounding boxes are constructed as xmin,xmax,ymin,ymax.

PR #11684, opened 2025-01-17
Remove OZI OZF2/OZFX3 raster driver

rouault:
With the blessing of @jratike80 (any reason for that particular one by the way? perhaps they've switched to a new format ?)

(by the way I see that Geoserver has "bug stomp" parties, GDAL should organize "driver stomp" from time to time :-) :-))


jratike80:
> With the blessing of @jratike80 (any reason for that particular one by the way? perhaps they've switched to a new format ?)

OZF2/OZF3 driver has always been read-only. OZF files can be written with the img2ozf utility, that writes nowadays ozf4 format. The encryption of ozf3 is reverse engineered, but ozf4 not.  So the only use case for the driver is if user happens to have ozf2 or ozf3 maps, that the user has not converted themself from jpeg, tiff, gif etc. with img2ozf. Thus, probably maps which have been bought many years ago. That is certainly not a big group of users.

agiudiceandrea:
Please see https://github.com/qgis/QGIS/issues/62810.

PR #11688, opened 2025-01-17
VFK: FIX invalid parcel "banana" geometries (fixes #3376)

### LINKED ISSUE #3376: VFK: Generated invalid parcel "banana" geometries
OGR VFK driver ver. 2.4.4 built on Debian from source generates invalid geometries at least for PAR layer(table) in situations called "banana" polygon. Instead of creating polygon with hole that touches exterior ring at one point it creates polygon in "banana" shape or even self intersection polygon type. The invalid polygon can be detected with ST_IsValid() function after importing into Postgis database.

seidlmic:
Fixes invalid polygon geometry building by replacing old algoritm with GEOS BuildArea() function. (Added OGRGeometry::BuildArea()/OGR_G_BuildArea() and GEOSBuildArea() to load polygons geometries in VFK.

Introduces on GEOS drive with about 10% performance penalty.

Thanks to Petr Zima as the main author.


ogr/ogrgeometry.cpp
```diff
@@ -6885,6 +6885,100 @@ OGRGeometryH OGR_G_Polygonize(OGRGeometryH hTarget)
...
+ *
+ * This method is built on the GEOSBuildArea_r() function of the GEOS
+ * library, check it for the definition of the geometry operation.
+ * If OGR is built without the GEOS library, this method will always fail,
+ * issuing a CPLE_NotSupported error.
+ *
+ * @return a newly allocated geometry now owned by the caller,
+ *         or NULL on failure.
+ *
+ * @since OGR 3.8.4
```
rouault, line 6913:
```suggestion
 * @since OGR 3.11
```

ogr/ogrgeometry.cpp
```diff
@@ -6885,6 +6885,100 @@ OGRGeometryH OGR_G_Polygonize(OGRGeometryH hTarget)
...
+ * library, check it for the definition of the geometry operation.
+ * If OGR is built without the GEOS library, this function will always fail,
+ * issuing a CPLE_NotSupported error.
+ *
+ * @param hGeom handle on the geometry to polygonize.
+ *
+ * @return a handle on newly allocated geometry now owned by the caller,
+ *         or NULL on failure.
+ *
+ * @since OGR 3.8.4
```
rouault, line 6971:
```suggestion
 * @since OGR 3.11
```

ogr/ogrsf_frmts/vfk/CMakeLists.txt
```diff
@@ -29,3 +29,7 @@ if (GDAL_USE_SQLITE3)
   gdal_target_link_libraries(ogr_VFK PRIVATE SQLite::SQLite3)
   target_compile_definitions(ogr_VFK PRIVATE -DHAVE_SQLITE)
 endif ()
+if (GDAL_USE_GEOS)
```
rouault, line 32: Is the driver still useful from a practical point of view without the GEOS dependency ? If not, then making it conditioned to GEOS presence should be done in ogr/ogrsf_frmts/CMakeLists.txt y changing

ogr_dependent_driver(vfk "Czech Cadastral Exchange Data Format" "GDAL_USE_SQLITE3")

to

ogr_dependent_driver(vfk "Czech Cadastral Exchange Data Format" "GDAL_USE_SQLITE3;GDAL_USE_GEOS")

by the way seeing SQLite3 is already a mandatory dependency, one could remove the if() test at line 27, and perhaps other conditional code in the driver, but that's admintedly not directly in the scope of this PR

rouault:
> Thanks to Petr Zima as the main author.

you can add at the end of the commit message (preferably using the email associated to his github account if he has one) so that git properly attributes authorship
```
Co-authored-by: name <email@domain.com>
```

rouault:
ok, so we have a test failure on configuration https://github.com/OSGeo/gdal/actions/runs/12838383582/job/35807308362?pr=11688 where we have SQLite support but not GEOS. So depending on the answer to my question of https://github.com/OSGeo/gdal/pull/11688#discussion_r1920869304, either make the driver GEOS dependent, or add a "@pytest.mark.require_geos" decoration to the failed test

ogr/ogrsf_frmts/vfk/CMakeLists.txt
seidlmic: The drive can be built without GEOS. In that case drive will warn about this situation during conversion and building polygon geometries (layers BUD, PAR} will be skipped. All other data will be imported.

```
Warning 6: GEOS support not enabled. Unable to build geometry for PAR.
Warning 6: GEOS support not enabled. Unable to build geometry for BUD.
```

seidlmic:
I have decided for  "@pytest.mark.require_geos" and not fully require GEOS.

PR #11695, opened 2025-01-20
[SQLite] Support SAVEPOINT

elpaso:
This actually fixes transactional editing in QGIS for GPKG:
https://github.com/qgis/QGIS/pull/59797#issuecomment-2544133498

Followup https://github.com/OSGeo/gdal/pull/11609

I have also moved the IsGenerated flag to the field definition class as recommended by Even.

### DIFF
--- a/ogr/ogr_feature.h
+++ b/ogr/ogr_feature.h
@@ -1270,8 +1270,6 @@ class CPL_DLL OGRFeature
 
     int IsFieldSet(int iField) const;
 
-    bool IsFieldGenerated(int iField) const;
-
     void UnsetField(int iField);
 
     bool IsFieldNull(int iField) const;
--- a/ogr/ogr_feature.h
+++ b/ogr/ogr_feature.h
@@ -82,6 +82,9 @@ class CPL_DLL OGRFieldDefn
     int bNullable;
     int bUnique;
 
+    // Used by drivers (GPKG) to track generated fields
+    bool m_bGenerated = false;
+
     std::string m_osDomainName{};  // field domain name. Might be empty
 
     std::string m_osComment{};  // field comment. Might be empty
@@ -190,6 +193,16 @@ class CPL_DLL OGRFieldDefn
     int IsUnique() const
     {
         return bUnique;
+    }
+
+    bool IsGenerated() const
+    {
+        return m_bGenerated;
+    }
+
+    void SetGenerated(bool bGeneratedIn)
+    {
+        m_bGenerated = bGeneratedIn;
     }
 
     void SetUnique(int bUniqueIn);
@@ -1243,6 +1256,8 @@ class CPL_DLL OGRFeature
     }
 
     int IsFieldSet(int iField) const;
+
+    bool IsFieldGenerated(int iField) const;
 
     void UnsetField(int iField);
 
--- a/ogr/ogr_feature.h
+++ b/ogr/ogr_feature.h
@@ -196,7 +196,10 @@ class CPL_DLL OGRFieldDefn
     }
 
     /**
-     * @brief IsGenerated
+     * @brief Return whether the field is a generated field.
+     *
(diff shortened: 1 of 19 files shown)


ogr/ogrfeature.cpp
```diff
@@ -1590,6 +1590,61 @@ int OGRFeature::IsFieldSet(int iField) const
     }
 }
 
+/************************************************************************/
+/*                             IsFieldGenerated()                       */
+/************************************************************************/
+
+/**
+ * \brief Test if a field is a generated field.
```
rouault, line 1598:
```suggestion
 * \brief Test if a field is a generated field.
 *
 * At time of writing, only the GeoPackage and PG drivers fill that information. Consequently,
 * only a returned value equal to TRUE can be fully trusted.
```

ogr/ogrfeature.cpp
```diff
@@ -1590,6 +1590,61 @@ int OGRFeature::IsFieldSet(int iField) const
...
+    return oFieldDefn ? oFieldDefn->IsGenerated() : false;
+}
+
+/************************************************************************/
+/*                          OGR_F_IsFieldGenerated()                    */
+/************************************************************************/
+
+/**
+ * \brief Test if a field is a generated field.
+ *
```
rouault, line 1622:
```suggestion
 *
 * At time of writing, only the GeoPackage driver fills that information. Consequently,
 * only a returned value equal to TRUE can be fully trusted.
```

ogr/ogr_feature.h
```diff
@@ -192,6 +195,26 @@ class CPL_DLL OGRFieldDefn
         return bUnique;
     }
 
+    /**
+     * @brief IsGenerated
```
rouault, line 199:
```suggestion
     * @brief Return whether the field is a generated field.
     *
     * At time of writing, only the GeoPackage and PG drivers fill that information. Consequently,
     * only a returned value equal to TRUE can be fully trusted.
```

ogr/ogrfeature.cpp
```diff
@@ -1590,6 +1590,61 @@ int OGRFeature::IsFieldSet(int iField) const
     }
 }
 
+/************************************************************************/
```
rouault, line 1593: Is it really useful to have a method at OGRFeature level that is a small convenience layer on top of the OGRFieldDefn API ? and in the C API in particular. We don't do this for other properties of OGRFieldDefn

ogr/ogrsf_frmts/gpkg/ogrgeopackagedatasource.cpp
```diff
@@ -7723,7 +7723,56 @@ OGRLayer *GDALGeoPackageDataset::ExecuteSQL(const char *pszSQLCommand,
         SoftRollbackTransaction();
         return nullptr;
     }
-
+    else if (STARTS_WITH_CI(osSQLCommand, "SAVEPOINT"))
```
rouault, line 7726: As there's a bit of copy & paste related to that handling of transaction related SQL (including also the above lines for BEGIN/COMMIT/ROLLBACK) between the SQLite and GPKG driver, I'm wondering if this could be factored ?

Perhaps a  "bool OGRSQLiteBaseDataSource::ProcessTransactionSQL(const std::string& osSQLCommand)" method that would return true if it has fully processed the command ?

rouault (commented):
Good job!

ogr/ogrfeature.cpp
elpaso: I don't know, I've got the inspiration from OGR_F_IsFieldNull and OGR_F_IsFieldSetAndNotNull. I can remove them if you prefer.

rouault: IsFieldNull is used extensively, hence a rather justified need for a shortcut helper. My slight preference would be to not increase the API surface for IsGenerated/SetGenerated at the OGRFeature level as it is likely going to be of marginal use

Issue #11693, opened 2025-01-20
MapInfo Tab file  GDAL "OGR_Dr_CopyDataSource" fucntion fails.


vdrasutis:
### What is the bug?

Hello.

We had GDAL build with some customizations. We recently used GDAL v3.5.1.
Now I build with **GDAL v3.5.3**.

We use this method to copy source data:
```
[DllImport(GdalWrapper.GdalDLL, CallingConvention = CallingConvention.Cdecl)]
private static extern IntPtr OGR_Dr_CopyDataSource(IntPtr hDriver, IntPtr hSrcDs, string newName, string[] parameters);
```
**Issue:** 
And now "OGR_Dr_CopyDataSource" copying data source fails with error: **"CreateFeature() failed: invalid feature id 1"**.

Attached [GDAL_tab.zip](https://github.com/user-attachments/files/18475870/GDAL_tab.zip), which contains simple LinqPad (I am opening with LinqPad v7 x64) C# Program it uses GDAL nuget. Also added simple TAB file format results generated by our app which I want to clone/ export as TAB file format results.

**Note:** 

1. We do not use in app GDAL nuget and have own object mapping.
2. I managed to repeat similar error when used CreateDataSource separately and populated each object reparately so when feature was created and I tried to SetFID it was throwind same/similar error (note tried to pass same option "FID_ASSIGNMENT=NO_AUTO" but it also did not worked):

```
// dstLayer.SetFIDAssignment(OGR.FID_ASSIGNMENT.FID_ASSIGN_AUTO);  // Enable automatic FID assignment
long dstFidBeforeSet = dstFeature.GetFID();
// Console.WriteLine($"FID before: {dstFidBeforeSet}."); 	// -1
dstFeature.SetFID(fid);		// But if I set FID from source I get error "Error cloning feature with ID 1: CreateFeature() failed: invalid feature id 1"
// long dstFidAfterSet = dstFeature.GetFID();          		// 1... and etc
// Console.WriteLine($"FID after: {dstFidAfterSet}.");

// Create feature in destination layer
int createResult = dstLayer.CreateFeature(dstFeature);
``` 

So would you suggest, how to fix  "OGR_Dr_CopyDataSource" usage or need to manually create destination Data Source and copy objects/create from source?

### Steps to reproduce the issue

[GDAL_tab.zip](https://github.com/user-attachments/files/18475870/GDAL_tab.zip)

### Versions and provenance

OS: Windows
App language: C#
[...]

PR #11698, opened 2025-01-20
Issue template: ask users to declare their potential use of AI [ci skip]

rouault:
Rationale: https://github.com/OSGeo/gdal/issues/11693#issuecomment-2602847660


hobu:
> We are not very keen of "AI" (Artificial stupIdity) tools, so at the very least if you have used such tools in your experiments, tell us where exactly. And better not use GDAL at all, and consider other projects that have time to loose with such insanity.

How about something like this:

> - [ ] I'm using LLM/AI-suggested procedures
> - [ ] AI (Copilot or something similar) supported my development of this PR

With language that says

> Using GDAL still requires thinking. While LLMs can sometimes provide useful guidance about how to use GDAL, the most precise documentation is the source code itself, with the test suite being the next most useful place for discovering how to use a feature. Please tell us in the ticket if you are using LLM-suggested procedures. understand what the LLM is suggesting

rouault:
> How about something like this:

thanks. suggestion adopted

jratike80 (approved):

PR #11704, opened 2025-01-22
Doc: Java JNI in ubuntu-full Docker image support

ravhed:
## What does this PR do?

The Docker image ubuntu-full builds the Java GDAL JNI bindings. The existing documentation of the image didn't mention this. This commit adds a note that the GDAL Java JNI is also available in the ubuntu-full image.


rouault:
thanks!

PR #11710, opened 2025-01-24
Update gdal_contour.rst

jidanni:
See https://en.wikipedia.org/wiki/Digital_elevation_model#Types

Also the user worries that their XYZ CSV file might not be an official DEM, though it will work just fine.

### DIFF
--- a/doc/source/programs/gdal_contour.rst
+++ b/doc/source/programs/gdal_contour.rst
@@ -27,7 +27,7 @@ Description
 -----------
 
 The :program:`gdal_contour` generates a vector contour file from the input
-raster elevation model (DEM).
+raster elevation model (e.g., a raster DEM).
 
 The contour line-strings are oriented consistently and the high side will
 be on the right, i.e. a line string goes clockwise around a top.


jratike80:
What if another user, maybe an OGC member, worries that they cannot create contours from their temperature data? Only "Regularly gridded coverage data file" could make them satisfied (maybe, needs to be discussed in the SWG). They might also argue that XYZ CSV is not raster, it is point data that can expressed as a raster, but only in a special case when it makes up a regular grid.

I do not believe that the meaning of DEM has ever misunderstood, not even by any OGC member. The word "raster" appears there already `raster elevation model (DEM).` But if you think that `raster elevation model (e.g., a raster DEM)` is needed to make it clear, it is OK for me.

rouault:
I believe the current description is enough clear / unambiguous, so in the interest of keeping text shorter when possible, I'll close without merging

jidanni:
I mean it
> The gdal_contour generates a vector contour file from the  input  raster elevation model (DEM).

looks like it should say
> input digital elevation model (DEM).

or
> input  raster elevation model (REM).

It looks rather asleep at the wheel.

jratike80:
Or just "The gdal_contour generates a vector contour file from the input raster."
The utility does not know if the input is DEM or DSM, or some temperature, pressure etc. data. That we deal with digital data is obvious for most users, I believe.
In https://gdal.org/en/stable/programs/gdaldem.html "elevation raster" and "DEM raster" are used. Non-uniform, but still correctly understood.
REM? It makes me do rapid eye movements, never heard in this context.

PR #11713, opened 2025-01-24
Warping: fix inconsistent replacement of valid value that collides with the dstnodata value

### LINKED ISSUE #11711: gdal warp changes 0's into 1's when called with a specific combination of options
### What is the bug?

Not too long ago we upgraded our systems from gdal 3.9.1 to 3.10.1. After the upgrade, we found that all our binary masks were filled with 1's.

This appeared to be caused by a very specific combination of warp options. A minimal reproducer is given below. It starts with an uint8 array full of zeros, and after warping, all the data has become 1's. Note that all the given warp options seem necessary to reproduce the issue, where the resampling can be anything except nearest, the dstNodata must be 0, and the outputType can be any unsigned integer type, to reproduce the issue.

```python
import numpy as np
from osgeo import gdal, gdal_array

gdal.UseExceptions()

data = np.zeros((3, 4), np.uint8)
print('before:', data)
ds = gdal_array.SaveArray(data, '/vsimem/data.tif')
ds.SetGeoTransform((0, 1, 0, 0, 0, 1))
warpoptions = {
    'format': 'GTiff',
    'resampleAlg': gdal.GRA_Bilinear,  # any non-nearest resampling
    "dstNodata": 0,
    "outputType": gdal.GDT_Byte,  # any unsigned int type
}
ds2 = gdal.Warp('/vsimem/data2.tif', ds, **warpoptions)
data2 = ds2.GetRasterBand(1).ReadAsArray()
print('after:', data2)
```

For me, this outputs:
```
before: [[0 0 0 0]
 [0 0 0 0]
 [0 0 0 0]]
after: [[1 1 1 1]
 [1 1 1 1]
 [1 1 1 1]]
```

### Versions and provenance

OS: ubuntu (gdal's own ubuntu-small docker)
version where this worked: 3.9.1
version with this bug: 3.10.1

### Additional context

_No response_

rouault:
Fixes #11711

While technically a bug fix, I'm somewhat 50% 50% to backport that to 3.10, in particular as it affects probably quite common use cases like Byte nearest resampling. Thoughts @dbaston ?

### DIFF
--- a/alg/gdalwarper.h
+++ b/alg/gdalwarper.h
@@ -463,6 +463,8 @@ class CPL_DLL GDALWarpKernel
 
     GWKTieStrategy eTieStrategy;
 
+    bool bWarnedAboutDstNoDataReplacement = false;
+
     /*! @endcond */
 
     GDALWarpKernel();
--- a/alg/gdalwarpkernel.cpp
+++ b/alg/gdalwarpkernel.cpp
@@ -190,6 +190,7 @@ static CPLErr GWKBilinearNoMasksOrDstDensityOnlyDouble(GDALWarpKernel *poWK);
 static CPLErr GWKCubicNoMasksOrDstDensityOnlyShort(GDALWarpKernel *poWK);
 static CPLErr GWKCubicSplineNoMasksOrDstDensityOnlyShort(GDALWarpKernel *poWK);
 static CPLErr GWKNearestShort(GDALWarpKernel *poWK);
+static CPLErr GWKNearestUnsignedShort(GDALWarpKernel *poWK);
 static CPLErr GWKNearestNoMasksOrDstDensityOnlyFloat(GDALWarpKernel *poWK);
 static CPLErr GWKNearestFloat(GDALWarpKernel *poWK);
 static CPLErr GWKAverageOrMode(GDALWarpKernel *);
@@ -1307,9 +1308,11 @@ CPLErr GDALWarpKernel::PerformWarp()
         bNoMasksOrDstDensityOnly)
         return GWKBilinearNoMasksOrDstDensityOnlyUShort(this);
 
-    if ((eWorkingDataType == GDT_Int16 || eWorkingDataType == GDT_UInt16) &&
-        eResample == GRA_NearestNeighbour)
+    if (eWorkingDataType == GDT_Int16 && eResample == GRA_NearestNeighbour)
         return GWKNearestShort(this);
+
+    if (eWorkingDataType == GDT_Int16 && eResample == GRA_NearestNeighbour)
+        return GWKNearestUnsignedShort(this);
 
     if (eWorkingDataType == GDT_Float32 && eResample == GRA_NearestNeighbour &&
         bNoMasksOrDstDensityOnly)
@@ -1521,6 +1524,47 @@ template <> double GWKClampValueT<double>(double dfValue)
 #endif
 
 /************************************************************************/
+/*                             AvoidNoData()                            */
+/************************************************************************/
+
+template <class T>
+inline void AvoidNoData(const GDALWarpKernel *poWK, int iBand,
+                        GPtrDiff_t iDstOffset)
+{
+    GByte *pabyDst = poWK->papabyDstImage[iBand];
+    T *pDst = reinterpret_cast<T *>(pabyDst);
+
+    if (poWK->padfDstNoDataReal != nullptr &&
+        poWK->padfDstNoDataReal[iBand] == static_cast<double>(pDst[iDstOffset]))
+    {
+        if constexpr (std::numeric_limits<T>::is_integer)
+        {
+            if (pDst[iDstOffset] ==
+                static_cast<T>(std::numeric_limits<T>::lowest()))
+            {
+                pDst[iDstOffset] =
+                    static_cast<T>(std::numeric_limits<T>::lowest() + 1);
+            }
+            else
+                pDst[iDstOffset]--;
+        }
+        else
+        {
+            if (pDst[iDstOffset] == std::numeric_limits<T>::max())
+            {
+                pDst[iDstOffset] =
+                    std::nextafter(pDst[iDstOffset], static_cast<T>(0));
+            }
+            else
+            {
+                pDst[iDstOffset] = std::nextafter(
+                    pDst[iDstOffset], std::numeric_limits<T>::max());
+            }
+        }
+    }
+}
+
+/************************************************************************/
 /*                         GWKSetPixelValueRealT()                      */
 /************************************************************************/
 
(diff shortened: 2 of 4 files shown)


vincentschut:
as the original bug issuer of #11711 : we don't need a backport, we have fixed our code internally to avoid hitting this corner case.

dbaston:
I wouldn't be inclined to backport it because it changes behavior in a way that might not be expected. Since there is no unambiguously correct behavior in this situation I think it would be good to emit a one-time warning like ("value %d in the source dataset has been changed to %d in the destination dataset to avoid being treated as NoData. To avoid this, select a different NoData value for the destination dataset.")

rouault:
> I think it would be good to emit a one-time warning like ("value %d in the source dataset has been changed to %d in the destination dataset to avoid being treated as NoData. To avoid this, select a different NoData value for the destination dataset.")

done

schwehr:
@sgillies @rouault  Another rasterio failure from a warp change with 6074ce4975c1e0ed9e7592beb5c21a9dd2a184a9

```
assert 1314520 == 1309625
 +  where 1314520 = <function count_nonzero at 0x3217fdd4b2b0>(array([14, 17, 11, ..., 42, 43, 45], shape=(1315638,), dtype=uint8))
 +    where <function count_nonzero at 0x3217fdd4b2b0> = np.count_nonzero
test3d = True, count_nonzero = 1309625
path_rgb_byte_tif = 'data/RGB.byte.tif'

    @pytest.mark.parametrize("test3d,count_nonzero", [(True, 1309625), (False, 437686)])
    def test_reproject_array_interface(test3d, count_nonzero, path_rgb_byte_tif):
        class DataArray:
            def __init__(self, data):
                self.data = data
    
            def __array__(self, dtype=None):
                return self.data
    
            @property
            def dtype(self):
                return self.data.dtype
    
        with rasterio.open(path_rgb_byte_tif) as src:
            if test3d:
                source = DataArray(src.read())
            else:
                source = DataArray(src.read(1))
        out = DataArray(np.empty(source.data.shape, dtype=np.uint8))
        reproject(
            source,
            out,
            src_transform=src.transform,
            src_crs=src.crs,
            src_nodata=src.nodata,
            dst_transform=DST_TRANSFORM,
            dst_crs="EPSG:3857",
            dst_nodata=99,
        )
        assert isinstance(out, DataArray)
>       assert np.count_nonzero(out.data[out.data != 99]) == count_nonzero
E       assert 1314520 == 1309625
E        +  where 1314520 = <function count_nonzero at 0x3217fdd4b2b0>(array([14, 17, 11, ..., 42, 43, 45], shape=(1315638,), dtype=uint8))
E        +    where <function count_nonzero at 0x3217fdd4b2b0> = np.count_nonzero
```

rouault:
@schwehr That's likely expected and should be filed on rasterio side.

schwehr:
This commit triggered a test failure with the Earth Engine Data Catalog SMAP processing. Here is the fix:

- "should not specify both srcNodata and dstNodata when the value is the same"
- https://github.com/google/earthengine-catalog/commit/0c9f8062d9998b06ea6939c4100fe046a55235ca
- https://github.com/google/earthengine-catalog/pull/1272

david-eidmann:
I think an error with a failure would be way better than a warning. We had a workflow where we received images with no nodatavalue set in the metadata, but we knew nodata = 0 by convention. So we warped them with '-dst_nodata 0'. The warnings from gdal got mangled in the outputs of various tool calls and remained unnoticed. Hence we got images where all nodata pixels were transformed to valid ones due to this "fix", causing huge issues in the pipelines by contaminating various products....

jratike80:
> but we knew nodata = 0 by convention. So we warped them with '-dst_nodata 0'.

Why didn't you warp them with `-srs_nodata 0`?
https://gdal.org/en/stable/programs/gdalwarp.html#cmdoption-gdalwarp-srcnodata

david-eidmann:
One could argue that we did not want to overwrite srs_nodata if they will be set at some point. But i cannot reconstruct our thought process at this time. In hindsight you are always smarter and we definitely had a bad workflow design in this place. But a working system without warnings (before the patch) remains mostly questioned.

Nevertheless, the changes introduced here can alter data quite heavily with only a warning (for example scene classifications with integer classes). I could imagine we are not the only ones affected in the future and would like to save others the hassle. Although i see instances where this feature is desired, I'd suggest protecting it behind a specifically desired option (for example setting -srs_protect_valid_pixels_offset or so and fail if not set).

jratike80:
An example about a use case when the current behavior is desired. The original image does not have nodata but the water areas are mostly pure black. When warped, the process creates nodata triangles at the edges and they needs to be transparent. But before this fix the water areas turned into nodata and transparent as well.

image

For my mind, if some pixel value in the source data means nodata, it is not too much to ask the user to tell it. Either by adding nodata into the image metadata with gdal edit, or by telling it with an warp program argument. Or a better alternative might be to add alpha or mask band into the output and not rely on nodata that tends to make problems for example with lossy compression.

Now, is it easier to use ` -srs_protect_valid_pixels_offset ` if user does not have nodata in the data, or to use `-srs_nodata` if user does have nodata in the data, and for some reason does not want to tell it in the image metadata, that feels like a proper solution.

david-eidmann:
> For my mind, if some pixel value in the source data means nodata, it is not too much to ask the user to tell it.

--> I completely see your use case and aggree it is not too much to ask the user to tell it. However, as this has not been necessary in the past and there are different "best" behaviours  depending on the use case, i would still prefere a protection mechanism. Purely hypothetical, a user's intention could be to "expand" the nodata area, by mapping the existing nodata to a new val which are both supposed to be nodata. Even when setting -srs_nodata i can think of cases where i would not expect the data modification with +=1.
I just find it not very user friendly, to have defaults with stong affects on data like this one. Warnings are a good thing, but many wrappers can swallow them.

> Now, is it easier to use `-srs_protect_valid_pixels_offset` if user does not have nodata in the data, or to use `-srs_nodata` if user does have nodata in the data, and for some reason does not want to tell it in the image metadata, that feels like a proper solution.

--> I am not completely sure if you agree with my proposed solution or not.

Heres a wrapup of my suggestion:

1. dst_nodata value NOT PRESENT in srs values:
 --> all good, continue

2. dst_nodata value PRESENT in srs values:
 --> Now we have users who would prefere the value-bumping with +=1 and users who would be catched completely off guard by this behaviour if unnoticed. Hence I would prefere an ERROR. If the new behaviour is intended, it could be triggered with -srs_protect_valid_pixels_offset (or a better named param) as suggested
If we need an error-free solution, i would stick with the old approach and prefere a warning that srs_data contained dst_nodata and is now nodata. This aligns with historic behaviour and also catches the cases where users set -srs_nodata but actually intend to expand the nodata mask.

PR #11716, opened 2025-01-25
CI: Point ExprTk to release branch

ArashPartow:
Use the release branch instead of the master branch.

### DIFF
--- a/docker/alpine-normal/Dockerfile
+++ b/docker/alpine-normal/Dockerfile
@@ -232,7 +232,7 @@ RUN --mount=type=cache,id=alpine-normal-spatialite,target=$HOME/.cache \
     fi
 
 # Download exprtk
-RUN curl -LO -fsS https://github.com/ArashPartow/exprtk/raw/refs/heads/master/exprtk.hpp \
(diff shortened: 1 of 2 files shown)


rouault:
@ArashPartow  is there some documentation of the difference between the ExprTk branches: master ,release ? and the logic for tagging. @dbaston was a bit concerned of force pushes that break git history. With stories like https://en.wikipedia.org/wiki/XZ_Utils_backdoor in mind, it is a bit concerning to have funky git practices that make auditing complicated.

ArashPartow:
@rouault  release is the stable branch, The master branch may change rapidly and be reverted and have force pushes etc. The tags are only on release and cmake branches. 

The tags are semver based making it easier for users that deal in semver style versioning, as ExprTk has its own non-standard versioning which can be found at the bottom of the file.

--------------------

wrt XZ the changes all happened on master.

rouault:
> wrt XZ the changes all happened on master.

my point was that force pushes, even on master, are generally frowned upon, as that make reviewing activity quite hard. Anyway that's your project. Thanks for the explanations

PR #11718, opened 2025-01-26
RFC106 implementation: Metadata items to reflect driver update capabilities

rouault:


autotest/gcore/misc.py
```diff
@@ -314,6 +318,66 @@ def misc_6_internal(datatype, nBands, setDriversDone):
                 dst_ds.GetMetadataItem("", None)
             dst_ds = None
 
+            if (
```
dbaston, line 321: I would pull this into a separate test...there is so much going on in "misc_6" already. I would be happy to try a PR if you want.

dbaston (approved):

autotest/gcore/misc.py
rouault: > I would be happy to try a PR if you want.

fine. Yes that's messy

PR #11721, opened 2025-01-27
Doc: CMake: mention that disabling the use of a package does not disable its detection

### LINKED ISSUE #11707: HDF5 dependency cannot be excluded
### What is the bug?

When configuring the source I can use the cmake define `-DGDAL_USE_HDF5=OFF` to disable HDF5 support. What happens is that HDF5 is still detected due to the following code in **cmake/helpers/CheckDependentLibraries.cmake**:
```
if(HAVE_KEA)
    # CXX is only needed for KEA driver
    gdal_check_package(HDF5 "Enable HDF5" COMPONENTS "C" "CXX" CAN_DISABLE VERSION 1.10)
else()
    gdal_check_package(HDF5 "Enable HDF5" COMPONENTS "C" CAN_DISABLE VERSION 1.10)
endif()
```

KEA driver depends on HDF5 but if both KEA and HDF5 are disabled, the configure system still looks for HDF5 libraries. I think that something like the following will fix the issue:

```
if(HAVE_KEA)
    # CXX is only needed for KEA driver
    gdal_check_package(HDF5 "Enable HDF5" COMPONENTS "C" "CXX" CAN_DISABLE VERSION 1.10)
else()
    if(GDAL_USE_HDF5)
        gdal_check_package(HDF5 "Enable HDF5" COMPONENTS "C" CAN_DISABLE VERSION 1.10)
    endif()
endif()
```

just run cmake with `-DGDAL_USE_HDF5=no` and `-DGDAL_USE_KEA=OFF` on a system with hdf5 installed

I get 

```
...

-- Found HDF5: /usr/lib64/libhdf5.so;/usr/lib64/libsz.so;/usr/lib64/libz.so;/usr/lib64/libdl.a;/usr/lib64/libm.so (found version "1.14.4-3") found components: C

...

 * HDF5
   Enable HDF5

...

```

in the output log.

### Versions and provenance

Gentoo Linux

```
$ gdalinfo --version
GDAL 3.9.1, released 2024/06/22

```

### Additional context

_No response_

rouault:
Fixes #11707

### DIFF
--- a/doc/source/development/building_from_source.rst
+++ b/doc/source/development/building_from_source.rst
@@ -293,6 +293,8 @@ following option:
 
     Control whether a found dependency can be used for the GDAL build.
 
+    Note that disabling the use of a package does not disable its detection.
+
 It is also possible to ask GDAL to disable the use of any external dependency
 (besides the required one, PROJ) by default by setting the following option to
 OFF. Individual libraries shall then be enabled explicitly with
--- a/doc/source/development/building_from_source.rst
+++ b/doc/source/development/building_from_source.rst
@@ -293,7 +293,7 @@ following option:
 
     Control whether a found dependency can be used for the GDAL build.
 
-    Note that disabling the use of a package does not disable its detection.
+    Note that CMake will still attempt to detect a package even if it has been disabled.
 
 It is also possible to ask GDAL to disable the use of any external dependency
 (besides the required one, PROJ) by default by setting the following option to


doc/source/development/building_from_source.rst
```diff
@@ -293,6 +293,8 @@ following option:
 
     Control whether a found dependency can be used for the GDAL build.
 
+    Note that disabling the use of a package does not disable its detection.
```
dbaston, line 296:
```suggestion
    Note that CMake will still attempt to detect a package even if it has been disabled.
```

dbaston (approved):

PR #11724, opened 2025-01-27
gdal raster mosaic/stack: turn gdalbuildvrt -strict mode, and add a --resolution=same mode mode

### LINKED ISSUE #11717: gdal raster mosaic/stack: fail if number of bands, projection, color interpretation are not the same within soures
cf comment https://github.com/OSGeo/gdal/pull/11715#discussion_r1929138106

rouault:
- gdal raster mosaic/stack: add a --resolution=same mode, and default to it
- gdalbuildvrt: add '-resolution same' mode to check all source rasters have the same resolution
- gdal raster mosaic/stack: turn gdalbuildvrt -strict mode

Fixes #11717 

CC @dbaston

### DIFF
--- a/apps/gdalalg_raster_mosaic.cpp
+++ b/apps/gdalalg_raster_mosaic.cpp
@@ -46,15 +46,18 @@ GDALRasterMosaicAlgorithm::GDALRasterMosaicAlgorithm()
     AddArg("band", 'b', _("Specify input band(s) number."), &m_bands);
     AddOverwriteArg(&m_overwrite);
     {
-        auto &arg = AddArg("resolution", 0,
-                           _("Target resolution (in destination CRS units)"),
-                           &m_resolution)
-                        .SetMetaVar("<xres>,<yres>|average|highest|lowest");
+        auto &arg =
+            AddArg("resolution", 0,
+                   _("Target resolution (in destination CRS units)"),
+                   &m_resolution)
+                .SetDefault("same")
+                .SetMetaVar("<xres>,<yres>|same|average|highest|lowest");
         arg.AddValidationAction(
             [this, &arg]()
             {
                 const std::string val = arg.Get<std::string>();
-                if (val != "average" && val != "highest" && val != "lowest")
+                if (val != "average" && val != "highest" && val != "lowest" &&
+                    val != "same")
                 {
                     const auto aosTokens =
                         CPLStringList(CSLTokenizeString2(val.c_str(), ",", 0));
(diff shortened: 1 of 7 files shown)


doc/source/programs/gdalbuildvrt.rst
```diff
@@ -92,6 +92,8 @@ changed in later versions.
 
     `user` must be used in combination with the :option:`-tr` option to specify the target resolution.
 
+    `same` (added in GDAL 3.11) checks that all source rasters have the same resolution and errors out when this is not the case.
```
dbaston, line 95:
```suggestion
    `same` (default, added in GDAL 3.11) checks that all source rasters have the same resolution and errors out when this is not the case.
```

dbaston: Also, remove "is the default" from `average` above.

apps/gdalalg_raster_mosaic.cpp
```diff
@@ -185,7 +188,17 @@ bool GDALRasterMosaicAlgorithm::RunImpl(GDALProgressFunc pfnProgress,
               "VRT");
 
     CPLStringList aosOptions;
-    if (!m_resolution.empty())
+    aosOptions.push_back("-strict");
+
+    aosOptions.push_back("-program_name");
+    aosOptions.push_back("gdal raster mosaic");
+
+    if (m_resolution.empty())
```
dbaston, line 196: When would this be true?

apps/gdalbuildvrt_lib.cpp
```diff
@@ -800,7 +811,8 @@ std::string VRTBuilder::AnalyseRaster(GDALDatasetH hDS,
             {
                 CPLString osExpected = GetProjectionName(pszProjectionRef);
                 CPLString osGot = GetProjectionName(proj);
-                return CPLSPrintf("gdalbuildvrt does not support heterogeneous "
+                return m_osProgramName +
```
dbaston, line 814: Is there a reason to do both string concatenation and printf?

dbaston (approved):

apps/gdalalg_raster_mosaic.cpp
rouault: good point. that can't. removed

apps/gdalbuildvrt_lib.cpp
rouault: basically because it is more convenient to write it that way than adding %s and .c_str()

PR #11725, opened 2025-01-27
Doc: robots.txt: lame attempt at preventing AI robots to scrap us

rouault:
Rationale: mostly because I'm fed up about the hype around that, but perhaps although because they're violating the MIT license of our content by not attributing what is derived from it.


rouault:
I know this is probably super unpopular given that most our users probably love those tools and probably want them to hallucinate on the most recent content.... I let other GDAL committers merge or just close, or maybe our stale bot will do the job lol

hobu:
This doesn't keep them out of our codebase, just our docs. Does it provide much value in that regard?

rouault:
Probably not. That's mostly a political statement.

PR #11726, opened 2025-01-28
Completely remove XPM driver

rouault:
This driver makes no sense in a geospatial context. Users can use ImageMagick to convert from/to XPM if they need to.


schwehr (approved):
Thanks for the cleanup!

PR #11727, opened 2025-01-28
ADRG: remove write support

rouault:


autotest/gdrivers/adrg.py
```diff
@@ -196,3 +108,23 @@ def test_adrg_zna_18():
 
 
 ###############################################################################
+# Test reading dataset with 2 subdataset image
+
+
+def test_adrg_read_2subdatasets():
+
+    ds = gdal.Open("data/adrg/subdataset/TRANSH01.THF")
```
schwehr, line 116: Might want to use `with gdal.Open("data/adrg/subdataset/TRANSH01.THF") as ds` and same with the 2nd open in this test.

schwehr (approved):

PR #11731, opened 2025-01-28
Remove R Object Data Store (*.rda) driver

rouault:


dbaston:
Talked this over with some colleagues that work with R in more different environments than I do. The only use case we can imagine is converting raster data into something that can be read with base R in an environment where GDAL is not available. Even in this situation, a user would probably prefer to use with R + terra/gdalraster/etc. to write an `RDS`. So, it seems unlikely that someone would need this driver. On the other hand, the `rda` format is still used and this appears to be a relatively simple driver.

I'd be fine with removing it, it can always be resurrected if needed.

dbaston (approved):

rouault:
> Do we want to go through deprecation before removal, as in #7943 ?

if our users would pay license fees, that would probably be fair.  Given they don't, not worth the burden IHMO: a git revert will take less energy. And probably the git revert should be issued by the people who want to resurrect drivers, to test their true motivation.

kadyb:
CC @rhijmans

mdsumner:
It's easy enough to craft something in base R, any raw binary format can be read and shaped in a line of code, and {ff} can be interchanged with RRASTER. There's plenty of packages out there with handcrafted code for FLT and EHdr and so forth. I thought there might be some existing workflows with this RDA format, but I think it's worth removing this and finding out, and I will help with any efforts to workaround or restore this.  😀

PR #11735, opened 2025-01-28
Remove RDB driver

rouault:
@tnixeu ok for that, or is the driver still useful here ?


tnixeu:
@rouault We are fine with the pull request. At the moment the driver is not useful. If there is demand from users, we will re-evaluate whether it makes sense have the driver in gdal or add support via a dynamically loaded driver.

schwehr (approved):

Issue #9327, opened 2024-02-26
Unable to Open SVG


evan203:
## Expected behavior and actual behavior.

I expected to be able to open a SVG file. The following SVG is generated from a PDF using `pdftocairo -svg`. I have also tried to use a svg generated from adobe illustrator (from the same pdf). 

## Steps to reproduce the problem.

```
ogrinfo "Asset 1.svg"
ERROR 4: `Asset 1.svg' not recognized as a supported file format.
ogrinfo failed - unable to open 'Asset 1.svg'.
```
`ogrinfo --formats` lists I have SVG as a supported format: `SVG -vector- (rov): Scalable Vector Graphics`

## Operating system

Windows 11 23H2

## GDAL version and provenance

GDAL 3.8.4 from osgeo4w. 

## Relevant files  

[Asset 1.pdf](https://github.com/OSGeo/gdal/files/14410233/Asset.1.pdf)

[Asset 1.svg](https://github.com/OSGeo/gdal/assets/63885438/2eddf67b-235a-4afd-8628-f6a20ec8a7c7)

[AI exported asset 1.svg](https://github.com/OSGeo/gdal/assets/63885438/7b31a404-9272-4aaf-a5ba-932ded5295c7)

rouault:
As documented in https://gdal.org/drivers/vector/svg.html : "Currently, it will only read SVG files that are the output from Cloudmade Vector Stream Server"

rouault:
> Does Cloudmade Vector Stream Server even exist any more?

apparently not : https://qgis-user.osgeo.narkive.com/p6nWarcj/add-svg-as-vector-layer "It looks like this product is "Cloudmade
Vector Stream Server", which appears to be a dead web service?

I guess the driver could be extended to be more generic. The Cloudmade stuff allowed to produced georeferenced datasets. A more generic SVG reader would produce arbitrary vector coordinates.
Reopening for a potential contributor willing to have a take on that.

nyalldawson:
@rouault is that a new requirement? I distinctly remember opening a Inkscape exported svg using ogr sometime around ~5 years ago

rouault:
> is that a new requirement?

no, it's not new. As far as I can see, it exists since day 1 (71ab469144a3666489283d3362e05a21e9470991)

nyalldawson:
@rouault 

Guess my memory is a bit leaky then! :laughing: (Maybe I went Inkscape SVG -> svg to dxf -> OGR...)

[...]

PR #11739, opened 2025-01-28
Remove SVG driver

rouault:
The driver could only read files from Cloudmade Vector Stream Server, which no longer exists.


nyalldawson (approved):

hobu:
This driver wasn't support for actual SVG? https://en.wikipedia.org/wiki/SVG

rouault:
> This driver wasn't support for actual SVG? https://en.wikipedia.org/wiki/SVG

no, it could I guess, but it is restricted to the formulation of the defunct  Cloudmade Vector Stream Server. Nobody has cared to do that improvement, so I assume generic SVG doesn't matter

hobu:
> so I assume generic SVG doesn't matter

👍 

#9327 showed someone attempting to use it but failing for the reason that it was tied to the CloudMade server implementation. If demand for this were significant there would be a lot more ticket traffic about it.

Routhwick:
Suffice to say that cloudmade.com itself went offline with practically no fanfare last northern fall or thereabouts, [according to IA](https://web.archive.org/web/*/https://cloudmade.com/).

PR #11741, opened 2025-01-29
Remove raster and vector SDTS drivers

rouault:
only last commit is relevant. Based on top of a previous PR


hobu:
After some discussion with @msmitherdc, we don't think there is any significant holdings or usage of the SDTS drivers at this time.

PR #11743, opened 2025-01-29
Remove OGR Tiger driver

rouault:


dbaston:
As far as I can see, older versions of US Census boundaries are only distributed in the TIGER format. 
https://www.census.gov/geographies/mapping-files/time-series/geo/tiger-line-file.2006.html#list-tab-790442341

dbaston:
> They can use GDAL Docker images of past versions to convert to GeoPackage

I'm not sure how feasible that is for the community that would use these files.
https://gis.stackexchange.com/q/323161/18189
https://gis.stackexchange.com/questions/172855/converting-historical-1992-2000-2006-tiger-line-files-to-modern-shapefile-form

jratike80:
> I'm not sure how feasible that is for the community that would use these files.

There are very few questions really about using data in the old TIGER/Line format from 2006 or earlier https://gis.stackexchange.com/search?tab=newest&q=TIGER&searchOn=3. No problem for removing the driver for my mind.

dbaston:
New question added today, FWIW https://gis.stackexchange.com/q/498346/18189

PR #11744, opened 2025-01-29
Remove OGDI support (VMAP) + all past cumulative removals

rouault:
only last commit is relevant. Based on top of a previous PR

Mostly used by the defense industry as far as I know. They are smart enough to maintain that. I particularly don't have much appetite to be involved anymore with libogdi which is an obsolete C code base with an awful build system.

Today's illustration that libogdi is obsolete: https://github.com/libogdi/ogdi/issues/31 and https://github.com/libogdi/ogdi/issues/30


elpaso (approved):

sebastic:
`frmts/drivers.ini` still includes `OGR_OGDI`, shouldn't that also be removed?

Other removed drivers are also still included there.

rouault:
> `frmts/drivers.ini` still includes `OGR_OGDI`, shouldn't that also be removed?

yes, thanks. Gone with 754168f98b

PR #11755, opened 2025-01-29
gml:CircleByCenterPoint(): return a 5-point CIRCULARSTRING for compliance with ISO/IEC 13249-3:2011

### LINKED ISSUE #11750: GML: gml:CircleByCenterPoint is converted to WKT which is no longer compliant
### What is the bug?

When reading a GML file containing a `gml:CircleByCenterPoint` geometry, the geometry is reported as a `CURVEPOLYGON` with a `CIRCULARSTRING`, consisting of three points, the first point being equal to the third point.

This is compliant with earlier editions of the SQL/MM standard (ISO/IEC 13249-3), as e.g. documented in the PostGIS documentation, https://postgis.net/docs/manual-3.5/using_postgis_dbmanagement.html#CircularString.

In [ISO/IEC 13249-3:2016](https://www.iso.org/obp/ui#iso:std:iso-iec:13249:-3:ed-5:v1:en), the current version of the standard, an `ST_CircularCurve` with equal segment start and end point is no longer permitted, Instead, type `ST_Circle` is introduced to deal with circles. This is described in ISO/IEC 13249-3:2016 Annex D Incompatibilities with ISO/IEC 13249-3:2011.

The WKT reported when converting a `gml:CircleByCenterPoint` geometry, will fail in e.g.
- Oracle, with [ORA-13346: the coordinates defining an arc are collinear](https://docs.oracle.com/en/error-help/db/ora-13346/index.html?r=23ai)
- SQL server see https://learn.microsoft.com/en-us/sql/relational-databases/spatial/circularstring?view=sql-server-ver16

It would be better if a  `gml:CircleByCenterPoint` would be converted to a `CURVEPOLYGON` with an `CIRCULARSTRING` defined by 5 points. That encoding will be recognized by more databases and other tools, hence increasing interoperability.

I noticed this issue with the WKT encoding of circles when reading GML. The underlying problem as described above possibly occurs in other drivers, I did not investigate that.

The following geometry:

```xml
<gml:Polygon srsName="http://www.opengis.net/def/crs/EPSG/0/25832"
             srsDimension="2">
    <gml:exterior>
        <gml:Ring>
            <gml:curveMember>
                <gml:Curve>
                    <gml:segments>
                        <gml:CircleByCenterPoint numArc="1">
                            <gml:pos>1 2</gml:pos>
                            <gml:radius uom="m">2</gml:radius>
                        </gml:CircleByCenterPoint>
                    </gml:segments>
                </gml:Curve>
            </gml:curveMember>
        </gml:Ring>
    </gml:exterior>
</gml:Polygon>
```

is reported as `CURVEPOLYGON (CIRCULARSTRING (-1 2,3 2,-1 2))`.

`SELECT sdo_geom.VALIDATE_GEOMETRY_WITH_CONTEXT(sdo_util.FROM_WKTGEOMETRY('CURVEPOLYGON (CIRCULARSTRING (-1 2,3 2,-1 2))') , 0,005) FROM dual;` fails with ORA-13346.

A better WKT representation would e.g. be `CURVEPOLYGON (CIRCULARSTRING (-1 2,1 0,3 2,1 4,-1 2))`.

Image

Relevant code (I think):

https://github.com/OSGeo/gdal/blob/abaccaa755f5fd4abd481992b843fdcb46e8433b/ogr/gml2ogrgeometry.cpp#L2042-L2057

### Versions and provenance

GDAL 3.10.1, released 2025/01/08, installed from OSGeo4W.

### Additional context

_No response_

rouault:
Fixes #11750

### DIFF
--- a/autotest/ogr/ogr_gml_geom.py
+++ b/autotest/ogr/ogr_gml_geom.py
@@ -2398,7 +2398,7 @@ def test_gml_CircleByCenterPoint():
     geom = ogr.CreateGeometryFromGML(gml)
 
     ogrtest.check_feature_geometry(
-        geom, ogr.CreateGeometryFromWkt("CIRCULARSTRING (-1 2,3 2,-1 2)")
+        geom, ogr.CreateGeometryFromWkt("CIRCULARSTRING (-1 2,1 4,3 2,1 0,-1 2)")
     )
 
 
--- a/ogr/gml2ogrgeometry.cpp
+++ b/ogr/gml2ogrgeometry.cpp
@@ -2043,8 +2043,14 @@ GML2OGRGeometry_XMLNode_Internal(const CPLXMLNode *psNode, const char *pszId,
         p.setX(dfCenterX - dfRadius);
         p.setY(dfCenterY);
         poCC->addPoint(&p);
+        p.setX(dfCenterX);
+        p.setY(dfCenterY + dfRadius);
+        poCC->addPoint(&p);
         p.setX(dfCenterX + dfRadius);
         p.setY(dfCenterY);
+        poCC->addPoint(&p);
+        p.setX(dfCenterX);
+        p.setY(dfCenterY - dfRadius);
         poCC->addPoint(&p);
         p.setX(dfCenterX - dfRadius);
         p.setY(dfCenterY);


autotest/ogr/ogr_gml_geom.py
```diff
@@ -2396,9 +2396,10 @@ def test_gml_CircleByCenterPoint():
 
     gml = "<gml:CircleByCenterPoint><gml:pos>1 2</gml:pos><gml:radius>2</gml:radius></gml:CircleByCenterPoint>"
     geom = ogr.CreateGeometryFromGML(gml)
+    print(geom)
```
schwehr, line 2399: Remove the debugging print.

rouault: thanks, removed

schwehr (approved):

heidivanparys:
@rouault Thanks for creating this pull request. One comment, the title should be updated from "… for compliance with ISO/IEC 13249-3:201**1** to "… for compliance with ISO/IEC 13249-3:201**6**".

… and while writing this, the pull request for merged. Well, my comment is here for reference then 🙂.

rouault:
> One comment, the title should be updated from

ah thanks, too many numbers and too many standards :-)

PR #11764, opened 2025-01-30
Reduce file opening

rouault:
- CPLGetPhysicalRAM(): cache result to avoid multiple file openings on repeated calls
- /vsis3/: cache content of ~/.aws/config and ~/.aws/credentials to avoid multiple file openings on repeated calls

Helps reducing the number of openat() syscalls in MapServer intensive scenarios

### DIFF
--- a/port/cpl_aws.cpp
+++ b/port/cpl_aws.cpp
@@ -608,7 +608,10 @@ static EC2InstanceCertainty IsMachinePotentiallyEC2Instance()
         }
     }
 
-    return IsMachinePotentiallyEC2InstanceFromLinuxHost();
+    static const EC2InstanceCertainty
+        eIsMachinePotentiallyEC2InstanceFromLinuxHost =
+            IsMachinePotentiallyEC2InstanceFromLinuxHost();
+    return eIsMachinePotentiallyEC2InstanceFromLinuxHost;
 
 #elif defined(_WIN32)
     if (!CPLTestBool(CPLGetConfigOption("CPL_AWS_AUTODETECT_EC2", "YES")))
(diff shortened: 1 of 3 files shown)


port/cpl_aws.cpp
```diff
@@ -1093,38 +1094,62 @@ static bool ReadAWSCredentials(const std::string &osProfile,
     osAccessKeyId.clear();
     osSessionToken.clear();
 
-    VSILFILE *fp = VSIFOpenL(osCredentials.c_str(), "rb");
-    if (fp != nullptr)
+    static std::mutex goMutex;
+    std::lock_guard oGuard(goMutex);
+
+    // Cache last request and result to avoid repeated file accesses.
```
pedros007, line 1100: Is the credentials cache from`osCredentialsFilename`cache populated once at process start and never again? Users could (should) update the value periodically, so a long-running GDAL processes might get access denied from AWS but it's possible new credentials were updated in the file.

rouault: > Is the credentials cache from osCredentialsFilename cache populated once at process start and never again?

yes

> but it's possible new credentials were updated in the file.

is it really realistic that ~/.aws/config or ~/.aws/credentials content change during the lifetime of a process ? Aren't those used for 'static' credentials ?

pedros007: > is it really realistic that ~/.aws/config or ~/.aws/credentials content change during the lifetime of a process ? Aren't those used for 'static' credentials ?

The typical GDAL user would probably not rotate credentials in the lifetime of a process (such as in a single gdalwarp call).  However, some companies/AWS IAM policies require users to rotate static credentials every few hours.  My particular use case (long-running Mapserver) is probably an edge case where it's easy enough for me to restart a Mapserver FastCGI process.

rouault: ok, so this basically means that the whole idea is busted. I'll revert that commit and just keep the CPLGetPhysicalRAM() change

rouault:
closing that one, and just keeping the CPLGetPhysicalRAM() change in https://github.com/OSGeo/gdal/pull/11764

PR #11770, opened 2025-01-31
Update ogrinfo.rst to note -dialect is ignored with -where.

jidanni:
I tried to combine this with my patch 62 but failed.


jratike80:
What do you think, could it be reported rather in -dialect?
`-dialect SQL dialect that is used for performing the statement that is set with the -sql option. Has no effect on -where.`

But maybe the whole `-dialect` part should be re-written to clarify what dialect gets selected and when.

jidanni:
All I know is the man page should warn in all the places it can!

PR #11776, opened 2025-01-31
CPLGetPhysicalRAM(): cache result to avoid multiple file openings on repeated calls

rouault:


dbaston (approved):
Simplified diff: https://github.com/OSGeo/gdal/pull/11776/files?w=1

pedros007:
> Simplified diff: https://github.com/OSGeo/gdal/pull/11776/files?w=1

Woah 🤯  I had no idea you could append `?w=1` to remove whitespace from a PR.  Thanks for this tip!

PR #11783, opened 2025-02-03
MiraMonVector: Fix issue 393742177 oss-fuzz

AbelPau:
## What does this PR do?

Fixes a memory leak that happens in the case where polygon zero creates the database but some error occurs in the creation of the polygon zero. The database must be destroyed and created again next time the polygon zero is created. It's been reproduced in the case of skipFailures==true.

## Tasklist


dbaston:
@rouault would you want a test for something like this?

PR #11787, opened 2025-02-04
Zarr V3: update with latest Zarr V3 specification

rouault:
- Rename 'endian' codec to 'bytes'
- Always generate a "codecs" array with the "bytes" codec in it. On reading, be tolerant with its absence but raise a warning as it being deprecated
- On reading, raise a warning on implicit groups, as being deprecated
- Support the 'zstd' codec, which isn't yet standardized, but is the default one used by Python zarr v3.0.2

Not handled: 'crc32c' and 'sharding' codecs


mdsumner:
awesome

(I don't have much to offer yet but fwiw I'm watching and will be trying this out)

frmts/zarr/zarr_v3_codec.cpp
```diff
@@ -29,18 +29,76 @@ ZarrV3Codec::ZarrV3Codec(const std::string &osName) : m_osName(osName)
...
+        abySrc.data(), abySrc.size(), &pOutputData, &nOutputSize,
+        m_aosCompressorOptions.List(), m_pCompressor->user_data);
+    if (bRet)
+    {
+        abyDst.resize(nOutputSize);
+    }
+    else if (nOutputSize > abyDst.size())
+    {
+        CPLError(CE_Failure, CPLE_AppDefined,
+                 "%s codec:Encode(): output buffer too small",
```
elpaso, line 62: I know we are not translating error messages but I really think we should eventually. 
Can we introduce _() at least in new code?

elpaso (approved):

frmts/zarr/zarr_v3_codec.cpp
rouault: cf my comment in another ticket. I believe this should be part of a more general move, if we eventually go into that direction.  A more practical reason is that the _() macro isn't in any of our headers currently, and we should be cautious how/where to introduce it to avoid toe-stepping over gettext...

PR #11788, opened 2025-02-04
SQLite/GPKG: run PRELUDE_STATEMENTS after end of initialization, in particular after Spatialite loading

### LINKED ISSUE #11782: -oo PRELUDE_STATEMENTS="select setdecimalprecision(1)" not working
### Feature description

Perhaps allow a --config SQL_PREAMBLE=... etc.
allowing users to adjust SQL global settings without
needing surprising workarounds, e.g.,
https://gis.stackexchange.com/questions/490023/how-to-set-spatialite-global-settings-in-gdal-ogr/490034#490034

rouault:
Fixes #11782

### DIFF
--- a/autotest/ogr/ogr_gpkg.py
+++ b/autotest/ogr/ogr_gpkg.py
@@ -5870,6 +5870,27 @@ def test_ogr_gpkg_prelude_statements(tmp_vsimem):
 
 
 ###############################################################################
+# Test PRELUDE_STATEMENTS open option
+
+
+def test_ogr_gpkg_prelude_statements_after_spatialite_loading(tmp_vsimem):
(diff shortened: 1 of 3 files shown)


jratike80:
Is it so that the Spatialite extension and geopackage extensions are loaded automatically and therefore using their functions in prelude_statements is easy?

A theoretical question, but maybe some day a user wants to use some other extension in prelude_statements. Must they load the extension with configuration option OGR_SQLITE_LOAD_EXTENSIONS=, or is is possible to do all with prelude_statements and allow loading extensions, really load an extension, and use a function from this extension?

dbaston (approved):

PR #11793, opened 2025-02-04
muparser: avoid uncaught exception on invalid variable name

dbaston:


frmts/vrt/vrtexpression_muparser.cpp
```diff
@@ -25,17 +25,31 @@ class MuParserExpression::Impl
...
     void Register(std::string_view osVariable, double *pdfValue)
     {
-        m_oParser.DefineVar(std::string(osVariable), pdfValue);
+        try
+        {
+            m_oParser.DefineVar(std::string(osVariable), pdfValue);
+        }
+        catch (const mu::Parser::exception_type &)
+        {
+            m_bCompileFailed = true;
```
rouault, line 40: no interesting error message coming from mu::Parser we could CPLError() ?

rouault (approved):

frmts/vrt/vrtexpression_muparser.cpp
dbaston: The message in the exception says that the variable name is invalid but doesn't say what the name was. Still, I've moved the site of the `CPLError` call to print an improved message.

rouault: Not to be addressed by this PR necessarily, but I'm wondering about the actual benefits of passing std::string_view as argument if we end up creating a std::string. This is slightly more verbose in term of source code, an probably also a bit in term of binary size ?

dbaston: Actual benefit is to `std::string_view` is negative, as you suggest. I think the longer-term improvement would be for `CPLDebug` and friends to accept a `std::string_view` argument instead of requiring everything be a C string. I haven't looked into it at all though, maybe there's some reason that won't work. Or maybe we should hold out for a couple of years and switch to `std::format`.

rouault (approved):

Issue #10066, opened 2024-05-29
Allow writing vector VRT from gdal.VectorTranslate


dbaston:
### Feature description

The vector VRT format offers a tremendously useful capability for lazily performing transformations on vector layers. However, the files currently must be authored by hand. The proposed enhancement would allow ogr2ogr to write VRT files, using arguments such as `-sql` and `-t_srs` to create and serialize a `OGRVRTLayer`, `OGRVRTWarpedLayer`, etc.

Implementing `pfnVectorTranslateFrom` in `OGRVRTDriver` may be a possible means to implement the functionality.

dbaston:
This is made unnecessary by the introduction of the `.gdalg.json` format.

PR #11794, opened 2025-02-04
"gdal vector filter": add --where and --fields; add "gdal vector select"

rouault:


doc/source/programs/gdal_vector_filter.rst
```diff
@@ -0,0 +1,150 @@
...
+
+Common options
+++++++++++++++
+
+.. option:: --strict
+
+    Strict mode (default). Error out when a field specified with :option:`--fields`
+    does not exist in the source layer(s).
+
+.. option:: --lax
```
dbaston, line 75: I might consider a name that more explicitly describes the behavior, like `--ignore-missing-fields`

dbaston:
I would consider putting `--fields` in a separate command (`select` ?) that could be chained with `filter`, or not. (A bit like the verbs in [dplyr](https://dplyr.tidyverse.org), I guess)

doc/source/programs/gdal_vector_filter.rst
```diff
@@ -0,0 +1,150 @@
...
+      -o, --output <OUTPUT>                                Output vector dataset [required]
+
+    Common Options:
+      -h, --help                                           Display help message and exit
+      --version                                            Display GDAL version and exit
+      --json-usage                                         Display usage as JSON document and exit
+      --drivers                                            Display driver list as JSON document and exit
+      --config <KEY>=<VALUE>                               Configuration option [may be repeated]
+      --progress                                           Display progress bar
+      --strict                                             Strict mode (default: true)
```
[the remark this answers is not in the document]
rouault: strict and lax removed

doc/source/programs/gdal_vector_filter.rst
```diff
@@ -0,0 +1,150 @@
...
+
+    Bounds to which to filter the dataset. They are assumed to be in the CRS of
+    the input dataset.
+    The X and Y axis are the "GIS friendly ones", that is X is longitude or easting,
+    and Y is latitude or northing.
+    Note that filtering does not clip geometries to the bounding box.
+
+.. option:: --where <WHERE>|@<filename>
+
+    Attribute query (like SQL WHERE)
```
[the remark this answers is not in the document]
rouault: fixed

apps/gdalalg_vector_filter.cpp
```diff
@@ -30,8 +33,218 @@ GDALVectorFilterAlgorithm::GDALVectorFilterAlgorithm(bool standaloneStep)
...
+/************************************************************************/
+
+class GDALVectorFilterAlgorithmLayer final
+    : public OGRLayer,
+      public OGRGetNextFeatureThroughRaw<GDALVectorFilterAlgorithmLayer>
+{
+  private:
+    bool m_bIsOK = true;
+    OGRLayer *const m_poSrcLayer;
+    OGRFeatureDefn *const m_poFeatureDefn = nullptr;
```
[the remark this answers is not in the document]
rouault: - m_poSrcLayer transformed into a reference
- for m_poFeatureDefn, we can't do much about that since OGRFeatureDefn includes its own ref counting mechanism. Changing that (or studying if we can...) is I believe in the todo list of @dbaston.  I've some doubts we can do that in the GDAL 3.x series

doc/source/programs/gdal_vector_filter.rst
rouault: done

rouault:
> I would consider putting `--fields` in a separate command (`select` ?)

done: "gdal vector select" added

jratike80:
Does `--fields` support `EXCLUDE`?

rouault:
> Does `--fields` support `EXCLUDE`?

I've adde an --exclude switch

doc/source/programs/gdal_vector_select.rst
```diff
@@ -0,0 +1,132 @@
...
+    no explicit name for the geometry field, ``_ogr_geometry_`` must be used to
+    select the unique geometry field.
+
+    Specifying a non-existing source field name results in an error.
+
+.. option:: --ignore-missing-fields
+
+    By default, if a field specified by --fields does not exist in the input
+    layer(s), an error is emitted and the processing is stopped.
+    When Specifying :option:`--ignore-missing-fields`, only a warning is
```
dbaston, line 100:
```suggestion
    When specifying :option:`--ignore-missing-fields`, only a warning is
```

doc/source/programs/migration_guide_to_gdal_cli.rst
```diff
@@ -182,3 +182,14 @@ Vector commands
...
+
+* Selecting features from a shapefile based on an attribute query, and restricting to a few fields
+
+.. code-block::
+
+    ogr2ogr -where "country='Greenland'" -select population,_ogr_geometry_ out.gpkg in.shp
+
+    ==>
+
+    gdal vector filter --where "country='Greenland'" --fields population,_ogr_geometry_ in.shp out.gpkg
```
dbaston, line 195:
```suggestion
    gdal vector pipeline ! read in.shp ! filter --where "country='Greenland'" ! select --fields population,_ogr_geometry_ ! write out.gpkg
```

doc/source/programs/gdal_vector_select.rst
```diff
@@ -0,0 +1,132 @@
...
+    ``a field with " double quote`` with a Unix shell:
+
+    .. code-block:: bash
+
+        --fields "regular_field,\"a_field_with space, and comma\",\"a field with \\\" double quote\""
+
+    A field is only selected once, even if mentioned several times in the list.
+
+    Geometry fields can also be specified in the list. If the source layer has
+    no explicit name for the geometry field, ``_ogr_geometry_`` must be used to
```
dbaston, line 91: I think needed to include `_ogr_geometry_` explicitly might be surprising for users. I can imagine including it by default unless an `--exclude-geom` argument is passed. Would be curious what others think.

apps/gdalalg_vector_select.cpp
```diff
@@ -0,0 +1,335 @@
...
+        m_poFeatureDefn->Reference();
+    }
+
+    ~GDALVectorSelectAlgorithmLayer() override
+    {
+        if (m_poFeatureDefn)
+            m_poFeatureDefn->Dereference();
+    }
+
+    bool IncludeFields(const std::vector<std::string> &selectedFields,
```
dbaston, line 102: The only thought I have on the implementation is whether the processing could be handled by creating an in-memory VRT that could either be materialized or not, depending on the output format. I don't think anything about this interface precludes that, so I could look into it later on as part of https://github.com/OSGeo/gdal/issues/10066.

autotest/utilities/test_gdalalg_vector_select.py
```diff
@@ -0,0 +1,214 @@
...
+    lyr.SetAttributeFilter("EAS_ID = 170")
+    lyr.ResetReading()
+    f = lyr.GetNextFeature()
+    assert f["EAS_ID"] == 170
+    assert lyr.GetNextFeature() is None
+    assert lyr.TestCapability(ogr.OLCFastFeatureCount) == 0
+    assert lyr.GetFeatureCount() == 1
+    lyr.SetAttributeFilter(None)
+
+    # Test spatial select on reslt layer
```
dbaston, line 67:
```suggestion
    # Test spatial select on result layer
```

doc/source/programs/gdal_vector_select.rst
dbaston: FWIW, the R [sf](https://github.com/r-spatial/sf) package behaves this way (some discussion at https://github.com/r-spatial/sf/issues/121).

apps/gdalalg_vector_select.cpp
dbaston: Same comment could apply to #11827

doc/source/programs/gdal_vector_select.rst
rouault: fixed

doc/source/programs/gdal_vector_select.rst
```diff
@@ -0,0 +1,132 @@
...
+    select the unique geometry field.
+
+    Specifying a non-existing source field name results in an error.
+
+.. option:: --ignore-missing-fields
+
+    By default, if a field specified by --fields does not exist in the input
```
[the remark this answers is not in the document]
rouault: fixed

doc/source/programs/gdal_vector_select.rst
rouault: > I think needing to include `_ogr_geometry_` explicitly might be surprising for users. I can imagine including it by default unless an `--exclude-geom` argument is passed

that's a good question. I'm not sure either. One thing that is certain is that the current difference in behavior between OGRSQL where "SELECT" automatically selects the geometry field (unless excluded thanks to your recent work on EXCLUDE) compared to all other SQL dialects where the geometry field must be explicitly selected surprises users a lot.

apps/gdalalg_vector_select.cpp
rouault: Good question. Regarding VRT'ization of all of this, I believe both GDAL and OGR VRT will have to be extended to support accepting a datasource and a pipeline string, rather than trying to extend the capabilities of the VRT drivers themselves to support all algorithms, which would involve duplicating all the code.

autotest/utilities/test_gdalalg_vector_select.py
rouault: fixed

apps/gdalalg_vector_select.cpp
rouault: ticketed at https://github.com/OSGeo/gdal/issues/11837

doc/source/programs/gdal_vector_select.rst
rouault: ticketed at https://github.com/OSGeo/gdal/issues/11839

Issue #11795, opened 2025-02-04
SWIG: Text passed to GDALAlgorithm::ReportError does not surface in Python exception message


dbaston:
### What is the bug?

When an error is issued during instantiation of `GDALAlgorithm`, the message passed to `GDALAlgorithm::ReportError` does not make it into the Python exception message. Instead, a Python`RuntimeError` is issued with an empty message. 

### Steps to reproduce the issue

Modify line 36 of `gdalalg_raster_edit.cpp` such that the short name is invalid:

```
        AddArg("dataset", '@', _("Dataset (in-place updated)"), &m_dataset,
```

Run `pytest autotest/utilities/test_gdalalg_raster_edit.py`

All tests fail with the following:

```
    def InstantiateSubAlgorithm(self, *args):
        r"""InstantiateSubAlgorithm(Algorithm self, char const * algName) -> Algorithm"""
>       return _gdal.Algorithm_InstantiateSubAlgorithm(self, *args)
E       RuntimeError
```

### Versions and provenance

master

dbaston:
Is the issue here that `GDALAlgorithm::AddArg` calls `GDALAlgorithm::ReportError` with `CE_Failure`, but the operation does not actually fail (the argument is still added, ignoring the invalid components of it). Would it be appropriate to change these to `CE_Warning`, which both allows the error message to surface in Python and seems more consistent with the actual behavior of `AddArg` ?

For context, I'm trying to add an algorithm for `gdal raster calc`, and it's more challenging to debug if the error messages don't surface.

rouault:
> is the issue here that `GDALAlgorithm::AddArg` calls `GDALAlgorithm::ReportError` with `CE_Failure`

I don't know. I didn't investigate

> Would it be appropriate to change these to CE_Warning

no, I don't think so. This is a real error that should be fixed in the code.

(The "low priority" tag is more than it is a one-time programmer issue that shouldn't affect end users, because the error is systematically emitted when we run the algorithm, and so we can't miss it, at least when using the command line)

rouault:
> Sorry, I'm not clear on what "this" refers to here. 

I meant that the situation that causes the CE_Failure message to be emitted is quite severe

> Do you mean that the signature of `GDALAlgorithm::AddArg` should be changed to actually fail if the argument is invalid?

I'm not convinced that this would help for the SWIG case. In nominal situations, there's no reason for the function to fail.

[...]

PR #11799, opened 2025-02-05
SWIG: Avoid losing error message

dbaston:
Fixes https://github.com/OSGeo/gdal/issues/11795

Current code checks `CPLGetLastErrorType` to see if there is an error. If so, `Py_XDECREF(resultobj)` is called, which causes the error to be reset. The subsequent call to `CPLGetLastErrorMsg()` then returns an empty string.

### DIFF
--- a/swig/python/modify_cpp_files.cmake
+++ b/swig/python/modify_cpp_files.cmake
@@ -42,7 +42,7 @@ string(REPLACE "#define SWIGPYTHON"
        _CONTENTS "${_CONTENTS}")
 
 string(REPLACE "return resultobj;"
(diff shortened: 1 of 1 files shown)


swig/python/modify_cpp_files.cmake
```diff
@@ -42,7 +42,7 @@ string(REPLACE "#define SWIGPYTHON"
        _CONTENTS "${_CONTENTS}")
 
 string(REPLACE "return resultobj;"
```
dbaston, line 44:
```suggestion
# patch to avoid memory leaks on exception (see 594fe48)
string(REPLACE "return resultobj;"
```

PR #11819, opened 2025-02-07
Make all Transform methods of OGRCoordinateTransformation and GDALTransformerFunc return FALSE as soon as one point fails to transform

### LINKED ISSUE #11817: Inconsistent return values from Transform methods in OGRCoordinateTransformation
### What is the bug?

`OGRCoordinateTransformation::Transform()` seems to give different return values depending on which call signature is used.

The first example below uses
`(size_t nCount, double *x, double *y, double *z = nullptr, int *pabSuccess = nullptr)`

which I would expect to return `TRUE` (and `success` would contain a vector of `{TRUE, FALSE}`). Instead it returns `FALSE`.

The second example uses
`(size_t nCount, double *x, double *y, double *z, double *t, int *pabSuccess)`

which returns `TRUE`. 

`TransformWithErrorCodes()` also returns `TRUE`.

Ex. 1
```cpp
// transform_test_ex1.cpp

#include <iostream>
#include <vector>

#include "ogr_spatialref.h"

int main() {
    OGRSpatialReference oSourceSRS{}, oDestSRS{};
    OGRCoordinateTransformation *poCT = nullptr;

    oSourceSRS.importFromEPSG(4326);
    oSourceSRS.SetAxisMappingStrategy(OAMS_TRADITIONAL_GIS_ORDER);
    oDestSRS.importFromEPSG(5070);
    oDestSRS.SetAxisMappingStrategy(OAMS_TRADITIONAL_GIS_ORDER);

    poCT = OGRCreateCoordinateTransformation(&oSourceSRS, &oDestSRS);
    
    std::vector<double> x = {-114, -1140};
    std::vector<double> y = {47, 47};
    std::vector<int> success(2);
    
    if (poCT->Transform(2, x.data(), y.data(), nullptr, success.data()))
        std::cout << "Transform returned TRUE" << std::endl;
    else
        std::cout << "Transform returned FALSE" << std::endl;

    OGRCoordinateTransformation::DestroyCT(poCT);
    return 0;
}
```

```
$ g++ -I /usr/include/gdal transform_test_ex1.cpp -lgdal -lproj -o transform_test

$ ./transform_test 
ERROR 1: PROJ: aea: Invalid longitude
Transform returned FALSE
```

Ex. 2
```cpp
// transform_test_ex2.cpp

#include <iostream>
#include <vector>

#include "ogr_spatialref.h"

#include <iostream>
#include <vector>

#include "ogr_spatialref.h"

int main() {
    OGRSpatialReference oSourceSRS{}, oDestSRS{};
    OGRCoordinateTransformation *poCT = nullptr;

    oSourceSRS.importFromEPSG(4326);
    oSourceSRS.SetAxisMappingStrategy(OAMS_TRADITIONAL_GIS_ORDER);
    oDestSRS.importFromEPSG(5070);
    oDestSRS.SetAxisMappingStrategy(OAMS_TRADITIONAL_GIS_ORDER);

    poCT = OGRCreateCoordinateTransformation(&oSourceSRS, &oDestSRS);
    
    std::vector<double> x = {-114, -1140};
    std::vector<double> y = {47, 47};
    std::vector<int> success(2);
    
    if (poCT->Transform(2, x.data(), y.data(), nullptr, nullptr, success.data()))
        std::cout << "Transform returned TRUE" << std::endl;
    else
        std::cout << "Transform returned FALSE" << std::endl;

    OGRCoordinateTransformation::DestroyCT(poCT);
    return 0;
}
```

```
$ g++ -I /usr/include/gdal transform_test_ex2.cpp -lgdal -lproj -o transform_test

$ ./transform_test 
ERROR 1: PROJ: aea: Invalid longitude
Transform returned TRUE
```

### Versions and provenance

ghcr.io/osgeo/gdal:ubuntu-full-latest

```
$ gdalinfo --version
GDAL 3.11.0dev-f1386937cde9e540784909294fdd45cda3ee65d2, released 2025/01/15

$ g++ --version
g++ (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
```

### Additional context

_No response_

rouault:
Fixes #11817

Kind of a breaking change, but the current behavior was highly inconsistent and hard to reason about.

New paragraph in MIGRATION_GUIDE.TXT:

- The following methods OGRCoordinateTransformation::Transform(size_t nCount, double *x, double *y, double *z, double *t, int *pabSuccess) and OGRCoordinateTransformation::TransformWithErrorCodes(size_t nCount, double *x, double *y, double *z, double *t, int *panErrorCodes) are modified to return FALSE as soon as at least one point fails to transform (to be consistent with the other form of Transform() that doesn't take a "t" argument), whereas previously they would return FALSE only if no transformation was found. When FALSE is returned the pabSuccess[] or panErrorCodes[] arrays indicate which point succeeded or failed to transform.

  The GDALTransformerFunc callback and its implementations (GenImgProjTransformer, RPCTransformer, etc.) are also modified to return FALSE as soon as at least one point fails to transform.

### DIFF
--- a/alg/gdaltransformer.cpp
+++ b/alg/gdaltransformer.cpp
@@ -185,10 +185,9 @@ CPLErr CPL_STDCALL GDALSuggestedWarpOutput(GDALDatasetH hSrcDS,
                                     adfExtent, 0);
 }
 
-static int GDALSuggestedWarpOutput2_MustAdjustForRightBorder(
+static bool GDALSuggestedWarpOutput2_MustAdjustForRightBorder(
     GDALTransformerFunc pfnTransformer, void *pTransformArg, double *padfExtent,
-    CPL_UNUSED int nPixels, int nLines, double dfPixelSizeX,
-    double dfPixelSizeY)
+    int /* nPixels*/, int nLines, double dfPixelSizeX, double dfPixelSizeY)
 {
     double adfX[21] = {};
     double adfY[21] = {};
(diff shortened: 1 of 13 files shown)


ctoney:
This seems a little awkward. If FALSE is returned, does it mean modification in-place still may have happened but we need to check `pabSuccess` to find out?

rouault:
> If FALSE is returned, does it mean modification in-place still may have happened but we need to check `pabSuccess` to find out?

yes. I can't really think of a better alternative

rouault:
> yes. I can't really think of a better alternative

well, there's an alternative, which would to return the number of points that have succeeded in being transformed, which would allow quickly to discriminate all-failed, all-succeeded, mixed success, but that wouldn't be any longer a boolean return value.

ctoney:
That sounds good, but trade-offs for breaking change I would leave up to you.

It's not too bad the way it is. I ended with something like
```cpp
    int res = poCT->Transform(pts_in.nrow(), xbuf.data(), ybuf.data(),
                              nullptr, nullptr, success.data());

    if (!res &&
        std::find(success.begin(), success.end(), TRUE) == success.end()) {

        // all failed
    } else if (!res) {
        // some failed
    } else {
        // all succeeded
    }

File: alg/gdaltransformer.cpp [hunk 1]
```diff
@@ -601,13 +588,15 @@ CPLErr CPL_STDCALL GDALSuggestedWarpOutput2(GDALDatasetH hSrcDS,
                     double ayTemp[2] = {padfY[i], padfY[i]};
                     double azTemp[2] = {padfZ[i], padfZ[i]};
                     int abSuccess[2] = {FALSE, FALSE};
+                    CPLTurnFailureIntoWarning(true);
```
elpaso, line 591: side note: should we have a context manager for this settings temporary changes?

elpaso (approved):

alg/gdaltransformer.cpp
rouault: good idea. I had the same thought too when writing that code. ticketed as https://github.com/OSGeo/gdal/issues/11831

PR #11822, opened 2025-02-07
Python bindings: add syntaxic sugar to get/set algorithm arguments + doc

rouault:
- Python bindings: add syntaxic sugar to get/set algorithm arguments
- Doc: add a few examples of how to use GDAL algorithms from Python

### DIFF
--- a/doc/source/programs/gdal_cli_from_python.rst
+++ b/doc/source/programs/gdal_cli_from_python.rst
@@ -1,0 +1,128 @@
+.. _gdal_cli_from_python:
+
+================================================================================
+How to use "gdal" CLI algorithms from Python
+================================================================================
+
+Raster commands
+---------------
+
+* Getting information on a raster dataset as JSON
+
+.. code-block:: python
+
+    from osgeo import gdal
+    import json
+
+    gdal.UseExceptions()
+    alg = gdal.GetGlobalAlgorithmRegistry()["raster"]["info"]
+    alg["input"] = "byte.tif"
+    alg.Run()
+    info = json.loads(alg["output-string"])
+    print(info)
+
+
+* Converting a georeferenced netCDF file to cloud-optimized GeoTIFF
+
+.. code-block:: python
+
+    from osgeo import gdal
+    gdal.UseExceptions()
+    alg = gdal.GetGlobalAlgorithmRegistry()["raster"]["convert"]
+    alg["input"] = "in.nc"
+    alg["output"] = "out.tif"
+    alg["output-format"] = "COG"
+    alg["overwrite"] = True  # if the output file may exist
+    alg.Run()
+    alg.Finalize() # ensure output dataset is closed
+
+or
+
+.. code-block:: python
+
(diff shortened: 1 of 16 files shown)


doc/source/programs/gdal_cli_from_python.rst
```diff
@@ -0,0 +1,128 @@
...
+
+* Getting information on a raster dataset as JSON
+
+.. code-block:: python
+
+    from osgeo import gdal
+    import json
+
+    gdal.UseExceptions()
+    alg = gdal.GetGlobalAlgorithmRegistry()["raster"]["info"]
```
dbaston, line 18: Accessing an "algorithm registry" still feels a bit low-level to me. What do you think about something like:

```
with gdal.run("raster info", {"input" :"byte.tif"}) as result:
    info = json.loads(result.output)
```

doc/source/programs/gdal_cli_from_python.rst
```diff
@@ -0,0 +1,128 @@
...
+    print(info)
+
+
+* Converting a georeferenced netCDF file to cloud-optimized GeoTIFF
+
+.. code-block:: python
+
+    from osgeo import gdal
+    gdal.UseExceptions()
+    alg = gdal.GetGlobalAlgorithmRegistry()["raster"]["convert"]
```
dbaston, line 31: or

```
with gdal.run("raster convert", {"input":"in.nc", "output":"out.tif", "output-format":"COG", "overwrite": True}) as result:
    ds = result.output

# ds is closed on __exit__

ds = gdal.run("raster convert", {"input":"in.nc", "output":"out.tif", "output-format":"COG", "overwrite": True}).result
# ds must be manually closed
```

doc/source/programs/gdal_cli_from_python.rst
rouault: An algorithm could have several output arguments (like generating a main raster and a mask). So ``result.output`` would only be allowed if the algorithm declare a single output argument ?

doc/source/programs/gdal_cli_from_python.rst
rouault: Life-time of object may be tricky... e.g "gdal raster edit" takes a "dataset" argument that is both an input and output (edit in place)

doc/source/programs/gdal_cli_from_python.rst
dbaston: > Is this needed at all? A Python user can already do subprocess.run(["raster", "info"])

Maybe they can, though if they're like me and they've only been using Python since 2.1, they still have to look up the documentation for `subprocess` every single time. How do I capture stdout again?

> No need for a context manager.

Not in this example - the idea was to avoid having to call a `Finalize()` method in subsequent examples.

doc/source/programs/gdal_cli_from_python.rst
dbaston: Could it be considered to have no output?

PR #11827, opened 2025-02-09
Add 'gdal vector sql', as standalone or part of 'gdal vector pipeline'

rouault:


doc/source/programs/gdal_vector_sql.rst
```diff
@@ -0,0 +1,105 @@
...
+
+    Usage: gdal vector sql [OPTIONS] <INPUT> <OUTPUT> <statement>|@<filename>
+
+    Apply SQL statement(s) to a dataset.
+
+    Positional arguments:
+      -i, --input <INPUT>                                  Input vector dataset [required]
+      -o, --output <OUTPUT>                                Output vector dataset [required]
+      --sql <statement>|@<filename>                        SQL statement(s) [may be repeated] [required]
+
```
jratike80, line 28: The number of `--sql`s must match the number of the layers defined later with `--output-layer`, right?

doc/source/programs/gdal_vector_sql.rst
```diff
@@ -0,0 +1,105 @@
...
+      -h, --help                                           Display help message and exit
+      --version                                            Display GDAL version and exit
+      --json-usage                                         Display usage as JSON document and exit
+      --drivers                                            Display driver list as JSON document and exit
+      --config <KEY>=<VALUE>                               Configuration option [may be repeated]
+      --progress                                           Display progress bar
+
+    Options:
+      -l, --layer, --input-layer <INPUT-LAYER>             Input layer name(s) [may be repeated]
+      -f, --of, --format, --output-format <OUTPUT-FORMAT>  Output format
```
jratike80, line 39: I believe that `-l` gets skipped when `-sql` is used.

doc/source/programs/gdal_vector_sql.rst
```diff
@@ -0,0 +1,105 @@
...
+      --co, --creation-option <KEY>=<VALUE>                Creation option [may be repeated]
+      --lco, --layer-creation-option <KEY>=<VALUE>         Layer creation option [may be repeated]
+      --overwrite                                          Whether overwriting existing output is allowed
+      --update                                             Whether to open existing dataset in update mode
+      --overwrite-layer                                    Whether overwriting existing layer is allowed
+      --append                                             Whether appending to existing layer is allowed
+      --output-layer <OUTPUT-LAYER>                        Output layer name(s) [may be repeated]
+      --dialect <DIALECT>                                  SQL dialect (e.g. OGRSQL, SQLITE)
+
```
jratike80, line 48: It may get complicated for the users to define these options correctly for multi-layer outputs if they want to override this but append to that etc.

doc/source/programs/gdal_vector_sql.rst
```diff
@@ -0,0 +1,105 @@
...
+    Can be repeated to generated multiple output layers.
+
+.. option:: --dialect <DIALECT>
+
+    SQL dialect. In some cases can be used to use the (unoptimized) :ref:`ogr_sql_dialect` instead
+    of the native SQL of an RDBMS by passing the ``OGRSQL`` dialect value.
+    The :ref:`sql_sqlite_dialect` dialect can be chosen with the ``SQLITE``
+    and ``INDIRECT_SQLITE`` dialect values, and this can be used with any datasource.
+
```
jratike80, line 82: My attempt on this:

SQL dialect. By default the native SQL of an RDBMS is used. If a datasource does not support SQL natively, the default is to use the OGRSQL dialect. OGRSQL can by used with any datasource. The SQLITE dialect can be chosen with the ``SQLITE`` and ``INDIRECT_SQLITE`` dialect values, and they can also be used with any datasource. Overriding the default dialect may be beneficial because the capabilities of the SQL dialects vary. What SQL dialects a driver supports can be checked with "gdal vector info".
```
gdal vector info --format "PostgreSQL"
Supported SQL dialects: NATIVE OGRSQL SQLITE
gdal vector info --format "ESRI Shapefile"
Supported SQL dialects: OGRSQL SQLITE

```

doc/source/programs/gdal_vector_sql.rst
```diff
@@ -0,0 +1,105 @@
...
+Examples
+--------
+
+.. example::
+   :title: Generate a GeoPackage file with a layer sorted by descending population
+
+   .. code-block:: bash
+
+        $ gdal vector sql in.gpkg out.gpkg --sql="SELECT * FROM country ORDER BY pop DESC"
+
```
jratike80, line 99: The example works but I guess that the name of the GPKG layer would be "SELECT", so maybe it would be better to include also `--output-layer`.

apps/gdalalg_vector_sql.cpp
```diff
@@ -0,0 +1,306 @@
...
+
+/************************************************************************/
+/*                   GDALVectorSQLAlgorithmDataset                      */
+/************************************************************************/
+
+namespace
+{
+class GDALVectorSQLAlgorithmDataset final : public GDALDataset
+{
+    GDALDataset *m_poSrcDS = nullptr;
```
elpaso, line 55: Is there a way to replace raw pointers with smart ones in this (new) class?

apps/gdalalg_vector_sql.cpp
```diff
@@ -0,0 +1,306 @@
...
+    if (m_sql.size() == 1)
+    {
+        auto outDS = std::make_unique<GDALVectorSQLAlgorithmDataset>(poSrcDS);
+        outDS->SetDescription(poSrcDS->GetDescription());
+
+        OGRLayer *poLayer = poSrcDS->ExecuteSQL(
+            m_sql[0].c_str(), nullptr,
+            m_dialect.empty() ? nullptr : m_dialect.c_str());
+        if (!poLayer)
+            return false;
```
elpaso, line 237: how is the actual error reported?

apps/gdalalg_vector_sql.cpp
```diff
@@ -0,0 +1,306 @@
...
+        // names
+        std::set<std::string> setOutputLayerNames;
+        std::vector<std::string> aosLayerNames;
+        for (const std::string &sql : m_sql)
+        {
+            auto poLayer = poSrcDS->ExecuteSQL(
+                sql.c_str(), nullptr,
+                m_dialect.empty() ? nullptr : m_dialect.c_str());
+            if (!poLayer)
+                return false;
```
elpaso, line 260: error reporting?

apps/gdalalg_vector_sql.h
```diff
@@ -0,0 +1,63 @@
...
+/************************************************************************/
+/*                      GDALVectorSQLAlgorithm                          */
+/************************************************************************/
+
+class GDALVectorSQLAlgorithm /* non final */
+    : public GDALVectorPipelineStepAlgorithm
+{
+  public:
+    static constexpr const char *NAME = "sql";
+    static constexpr const char *DESCRIPTION =
```
elpaso, line 29: isn't this potentially localized ?

doc/source/programs/gdal_vector_sql.rst
```diff
@@ -0,0 +1,105 @@
...
+================================================================================
+"gdal vector sql" sub-command
+================================================================================
+
+.. versionadded:: 3.11
+
+.. only:: html
+
+    Apply SQL statement(s) to a dataset
```
elpaso, line 11:
```suggestion
    Apply SQL statement(s) to a dataset.
```

doc/source/programs/gdal_vector_sql.rst
```diff
@@ -0,0 +1,105 @@
...
+.. include:: gdal_options/of_vector.rst
+
+.. include:: gdal_options/co_vector.rst
+
+.. include:: gdal_options/overwrite.rst
+
+.. option:: --sql <sql_statement>|@<filename>
+
+    SQL statement to execute that returns a table/layer (typically a SELECT
+    statement)
```
elpaso, line 72:
```suggestion
    statement).
```

gcore/gdalalgorithm.cpp
```diff
@@ -267,9 +257,24 @@ bool GDALAlgorithmArg::Set(const std::string &value)
...
+
+    return true;
+}
+
+bool GDALAlgorithmArg::Set(const std::string &value)
+{
+    if (m_decl.GetType() != GAAT_STRING)
+    {
+        CPLError(CE_Failure, CPLE_AppDefined,
+                 "Calling Set(std::string) on argument '%s' of type %s is not "
```
elpaso, line 270: prepare for localization? _()

elpaso (approved):
Nice job! 
I just left some minor comments.

doc/source/programs/gdal_vector_sql.rst
rouault: yes, I've added doc for --output-layer mentionning this

doc/source/programs/gdal_vector_sql.rst
rouault: I agree -l / --input-layer is confusing in that context. Removed for that utility

doc/source/programs/gdal_vector_sql.rst
rouault: --append should work if they correctly specify layer names with --output-layer

doc/source/programs/gdal_vector_sql.rst
rouault: done. thanks

doc/source/programs/gdal_vector_sql.rst
rouault: done. thanks

apps/gdalalg_vector_sql.cpp
rouault: I've changed to using a reference instead. But ownership of the source dataset is a bit complicated. It is a broader subject, and I'm not sure we can easily fix that in the GDAL 3.x series (nor we'd have the will to break for GDAL 4)

apps/gdalalg_vector_sql.cpp
rouault: in most cases a CPLError() should have been emitted by ExecuteSQL(), but thinking deeper, there might be cases where the statement is valid but is not a SELECT-like one and thus return a null layer without error. I've added handling for that case. Thanks.

apps/gdalalg_vector_sql.h
rouault: not sure we could do that with a constexpr. Anyway our localization story is inexistent and would be a separated topic

doc/source/programs/gdal_vector_sql.rst
rouault: fixed

gcore/gdalalgorithm.cpp
rouault: same as above. This should be a more general move to autoconvert all CPLError() calls if we decide to go for localization

PR #11843, opened 2025-02-12
[mysql] Remove deprecated reconnect option for mysql >= 80034

### LINKED ISSUE #11842: ogr2ogr: MySQL driver uses deprecated MYSQL_OPT_RECONNECT option
### What is the bug?

Hi,

Thanks for your great work on GDAL, ogr2ogr and the MySQL driver. 

I've recently found the `ogr2ogr` MySQL driver is using a deprecated `MYSQL_OPT_RECONNECT` option in MySQL 8. This emits a warning to standard out for now, but may fail in future.

The line in question is here: 
https://github.com/OSGeo/gdal/blob/51ee2933c08896648613488926bff36939a66be7/ogr/ogrsf_frmts/mysql/ogrmysqldatasource.cpp#L230C30-L230C49

The option is marked as deprecated in MySQL's documentation here:
https://dev.mysql.com/doc/c-api/8.4/en/c-api-auto-reconnect.html

It looks like there is no substitute, so perhaps the easiest course of action is to remove that line from the driver.

Thanks again... Scott

Command to run:

```
ogr2ogr -overwrite -skipfailures -progress -f MySQL MySQL:my_db,host=mysql,user='username',password='password',port=3306 -nln my_table -lco SPATIAL_INDEX=NO -t_srs EPSG:4326 -s_srs EPSG:4326 my_shapefile.shp
```

Output:

```
WARNING: MYSQL_OPT_RECONNECT is deprecated and will be removed in a future version.
```

### Versions and provenance

I checked the source in GitHub and the MySQL driver is still using this option as of 3.10. Both `stable` and `master` branches also have this code. 

I'm currently forced to use an older version of GDAL 3.2.3 to match my environment due to limitations with `cmake`. I don't expect anything to happen with older versions like this of course.

### Additional context

This is just a deprecation warning at present, but future versions of MySQL may change this behaviour. 

Thanks again for your work! Let me know if I can be of assistance.

elpaso:
MYSQL_OPT_RECONNECT is deprecated since mysql 80034

Fixes #11842

### DIFF
--- a/ogr/ogrsf_frmts/mysql/ogrmysqldatasource.cpp
+++ b/ogr/ogrsf_frmts/mysql/ogrmysqldatasource.cpp
@@ -217,6 +217,7 @@ int OGRMySQLDataSource::Open(const char *pszNewName, char **papszOpenOptionsIn,
         CSLDestroy(papszTableNames);
         return FALSE;
     }
+#if defined(LIBMYSQL_VERSION_ID) && (LIBMYSQL_VERSION_ID < 80034)
     else
     {
         // Enable automatic reconnection
@@ -229,6 +230,7 @@ int OGRMySQLDataSource::Open(const char *pszNewName, char **papszOpenOptionsIn,
         // and at any point on more recent versions.
         mysql_options(hConn, MYSQL_OPT_RECONNECT, &reconnect);
     }
+#endif
 
     bDSUpdate = bUpdate;
 


rouault:
> MYSQL_OPT_RECONNECT is deprecated since mysql 80034

for my information, do we know why it is deprecated: does mysql automatically reconnect without being explictly told ?

elpaso:
> > MYSQL_OPT_RECONNECT is deprecated since mysql 80034
> 
> for my information, do we know why it is deprecated: does mysql automatically reconnect without being explictly told ?

I did some research but I only find some comments regarding the fact that the automatic reconnect didn't work reliably in all situations and didn't leave the connection in a clean state, it was off by default and AFAIK it has no replacements.

I also tried to write some reconnection logic base when mysql_ping() returns false but I couldn't make it work. I could dig deeper if you think this is important.

rouault:
> I could dig deeper if you think this is important.

no, definitely not worth it. Thanks for the explanation !

rouault (approved):

PR #11844, opened 2025-02-12
Undo executive order 14172

rouault:


nyalldawson (approved):
Needs some unit tests 👍

rouault:
> Needs some unit tests 👍

done. There were indeed needed. Uncensoring tend to be more complicated than I anticipated, thanks to the "big data" / Arrow code path. I knew I was going to regret adding this stuff

NathanW2:
+1

ogr/ogrfeature.cpp
```diff
@@ -4199,6 +4199,41 @@ void OGR_F_SetFieldDouble(OGRFeatureH hFeat, int iField, double dfValue)
...
+/************************************************************************/
+/*                       SetFieldSameTypeUnsafe()                       */
+/************************************************************************/
+
+//! @cond Doxygen_Suppress
+void OGRFeature::SetFieldSameTypeUnsafe(int i, char *pszValueTransferred)
+{
+    if (pszValueTransferred &&
+        (pszValueTransferred[0] == 'g' || pszValueTransferred[0] == 'G') &&
+        EQUAL(pszValueTransferred, "Gulf of America"))
```
strk, line 4211: This would need some localization too, if it's going to really be used, some external mapping table...

elpaso (approved):
Lovely :)

ogr/ogrfeature.cpp
jratike80: Certainly not my dogs.

robe2:
I approve this message

alpha-beta-soup:
If a user hypothetically has data that has fields `name` and `alt_name` with values `Dinali` and `Mount McKinley`, and runs it through an OGR translate, the user will be left with values of `Dinali` and `Dinali`, and also get a warning message?

rouault:
> the user will be left with values of `Dinali` and `Dinali`, and also get a warning message?

yes

rouault:
> but I'm happy to go on record as humbly suggesting that this sort of interfering with user data sets a poor precedent.

yes, I totally agree this change goes beyond what is normally acceptable for a software that should be neutral. That said, the project shouldn't be hostage of my somewhat vain attempt at countering foolishness... If someone wants to revert that (or implement something more subtle to be compatible of your scenario) and find someone to merge it, then that's fine...

jratike80:
I would suggest to keep the change for 4 years and consider it again then. Everyone can see that it is a small, a bit childish,  targeted expression of opionion from a frustrated individual. It is not meant to be a precedent for a large scale name war.

rouault:
I've reverted the change (b9526c1ac59c7f44ec935fb75d429f41e512d092). At least git history will keep track of this vain attempt at protesting.

jbednar (approved):

PR #11850, opened 2025-02-13
Add 'gdal raster calc'

dbaston:
## What does this PR do?

Adds a `gdal raster calc` tool that provides much of the functionality of `gdal_calc.py`. Some differences with `gdal_calc.py`:

- VRT output is supported, using the `muparser` expression engine
- input names are not restricted to A-Z; the syntax is `--input X=file.tif` rather than `-X file.tif`
- rather than have an `--allBands` flag, the expression can either refer to all bands of a raster (`--calc "X + 3"`) or a single band of a raster (`--calc "X[1] + 3"`)
- the `--extent union` and `--extent intersection` arguments are not supported, at least for now. I think these may best be handled by separate calls to `gdal raster mosaic/stack` and `gdal raster calc` but could be persuaded otherwise without much difficulty
- the various nodata handling flags of `gdal_calc.py` are not supported (yet?)
- there isn't (yet) a good way to use a subset of bands as an input. I'd suggest `--input name=dsn[band(s)]`
- Python isn't (yet) available as expression engine

Feedback wanted!

#10067 
#11209 
#11656

### DIFF
--- a/apps/gdalalg_raster_calc.cpp
+++ b/apps/gdalalg_raster_calc.cpp
@@ -101,7 +101,8 @@ struct SourceProperties
     int nX{0};
     int nY{0};
     std::array<double, 6> gt{};
-    OGRSpatialReference *srs{nullptr};
+    std::unique_ptr<OGRSpatialReference, OGRSpatialReferenceReleaser> srs{
+        nullptr};
 };
 
 static std::optional<SourceProperties>
@@ -136,7 +137,7 @@ UpdateSourceProperties(SourceProperties &out, const std::string &dsn,
         if (options.checkSRS && out.srs)
         {
             const OGRSpatialReference *srs = ds->GetSpatialRef();
-            srsMismatch = srs && !srs->IsSame(out.srs);
+            srsMismatch = srs && !srs->IsSame(out.srs.get());
         }
     }
 
@@ -461,7 +462,8 @@ GDALCalcCreateVRTDerived(const std::vector<std::string> &inputs,
         out.nX = ds->GetRasterXSize();
         out.nY = ds->GetRasterYSize();
         out.nBands = 1;
-        out.srs = ds->GetSpatialRef() ? ds->GetSpatialRef()->Clone() : nullptr;
+        out.srs.reset(ds->GetSpatialRef() ? ds->GetSpatialRef()->Clone()
+                                          : nullptr);
         ds->GetGeoTransform(out.gt.data());
     }
 
@@ -476,7 +478,7 @@ GDALCalcCreateVRTDerived(const std::vector<std::string> &inputs,
         auto props = UpdateSourceProperties(out, dsn, options);
         if (props.has_value())
         {
-            sourceProps[source_name] = props.value();
+            sourceProps[source_name] = std::move(props.value());
         }
         else
         {
@@ -503,7 +505,7 @@ GDALCalcCreateVRTDerived(const std::vector<std::string> &inputs,
     ds->SetGeoTransform(out.gt.data());
     if (out.srs)
     {
-        ds->SetSpatialRef(OGRSpatialReference::FromHandle(out.srs));
+        ds->SetSpatialRef(OGRSpatialReference::FromHandle(out.srs.get()));
     }
 
     return ds;
--- a/apps/gdalalg_raster_calc.cpp
+++ b/apps/gdalalg_raster_calc.cpp
@@ -34,10 +34,15 @@ struct GDALCalcOptions
     bool checkExtent{true};
 };
 
-static bool ShouldReplace(const std::string &str, size_t from, size_t to)
+static bool MatchIsCompleteVariableNameWithNoIndex(const std::string &str,
+                                                   size_t from, size_t to)
 {
     if (to < str.size())
     {
+        // If the character after the end of the match is:
+        // * alphanumeric or _ : we've matched only part of a variable name
+        // * [ : we've matched a variable that already has an index
+        // * ( : we've matched a function name
         if (std::isalnum(str[to]) || str[to] == '_' || str[to] == '[' ||
             str[to] == '(')
         {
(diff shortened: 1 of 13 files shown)


rouault (commented):
Great work! Just a few comments below

doc/source/programs/gdal_raster_calc.rst
```diff
@@ -0,0 +1,104 @@
...
+.. option:: --calc
+
+    An expression to be evaluated using the `muparser <https://beltoforion.de/en/muparser>`__ math parser library.
+    The expression may refer to individual bands of each input (e.g., ``X[1] + 3``) or it may be applied to all bands
+    of an input (``X + 3``). If the expression contains a reference to all bands of multiple inputs, those inputs
+    must either have the same the number of bands, or a single band.
+
+    For example, if inputs ``A`` and ``B`` each have three bands, and input ``C`` has a single band, then the argument
+    ``--calc "A + B + C"`` is would be equivalent to ``--calc "A[1] + B[1] + C[1]" --calc "A[2] + B[2] + C[2]" --calc "A[3] + B[3] + C[3]"``.
+
```
jratike80, line 62: I don't undestand the example. If C has one band, is is really possible to use `C[2]` and `C[3]`? Why?
"is would be equivalent to" should probably be "would be equivalent".

apps/gdalalg_raster_calc.cpp
```diff
@@ -0,0 +1,604 @@
...
+    int nY{0};
+    std::array<double, 6> gt{};
+    OGRSpatialReferenceH srs{nullptr};
+};
+
+static std::optional<SourceProperties>
+UpdateSourceProperties(SourceProperties &out, const std::string &dsn,
+                       const GDALCalcOptions &options)
+{
+    GDALDatasetH ds = GDALOpen(dsn.c_str(), GA_ReadOnly);
```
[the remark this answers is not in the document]
dbaston: done

autotest/utilities/test_gdalalg_raster_calc.py
```diff
@@ -0,0 +1,449 @@
...
+
+@pytest.mark.parametrize("output_format", ("tif", "vrt"))
+def test_gdalalg_raster_calc_basic_1(calc, tmp_vsimem, output_format):
+
+    np = pytest.importorskip("numpy")
+
+    infile = "../gcore/data/rgbsmall.tif"
+    outfile = tmp_vsimem / "out.tif"
+
+    calc.GetArg("input").Set([infile])
```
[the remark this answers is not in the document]
dbaston: done

doc/source/programs/gdal_raster_calc.rst
dbaston: A sloppy mistake, fixed in 648dfd983511ac6b4e4b27b4cd9e0722bc3df15e

apps/gdalalg_raster_calc.cpp
```diff
@@ -0,0 +1,604 @@
...
+
+    return true;
+}
+
+/**
+ *  Add a band subscript to all instances of a specified variable that
+ *  do not already have such a subscript. For example, "X" would be
+ *  replaced with "X[3]" but "X[1]" would be left untouched.
+ */
+static std::string SetBandIndices(const std::string &origExpression,
```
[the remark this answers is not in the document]
dbaston: Line 49 should prevent this. But I'll look at whether I can make `test_gdalalg_raster_calc_expression_rewriting` less convoluted.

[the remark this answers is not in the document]
dbaston: Hopefully improved by ed4b07a225751def3a53c03e88af3c6ad848eba7

apps/gdalalg_raster_calc.cpp
```diff
@@ -0,0 +1,617 @@
...
+    return expression;
+}
+
+struct SourceProperties
+{
+    int nBands{0};
+    int nX{0};
+    int nY{0};
+    std::array<double, 6> gt{};
+    OGRSpatialReference *srs{nullptr};
```
rouault, line 104: I assume srs is leaked.  The following (or something along that line) should rather be used.  (we should probably have a ``typdef std::unique_ptr<OGRSpatialReference, OGRSpatialReferenceReleaser> OGRSpatialReferencePtr`` in ogr_spatialref.h)
```suggestion
    std::unique_ptr<OGRSpatialReference, OGRSpatialReferenceReleaser> srs{nullptr};
```

rouault: or maybe just store it as a non-pointer object ?

apps/gdalalg_raster_calc.cpp
```diff
@@ -0,0 +1,619 @@
...
+                return false;
+            }
+            name = "X";
+        }
+        else
+        {
+            name = input.substr(0, pos);
+        }
+
+        std::string dsn = input.substr(pos + 1, input.size());
```
rouault, line 371: This causes an issue with unsigned-integer-overflow sanitizer when pos == std::string::npos:
```
 /home/runner/work/gdal/gdal/apps/gdalalg_raster_calc.cpp:371:44: runtime error: unsigned integer overflow: 18446744073709551615 + 1 cannot be represented in type 'size_type' (aka 'unsigned long')
```
and the use of input.size() is probably OK, but not needed, so I guess this should be:

```suggestion
        std::string dsn = (pos == std::string::npos) ? input : input.substr(pos + 1);
```

apps/gdalalg_raster_calc.cpp
```diff
@@ -0,0 +1,619 @@
...
+        datasets[name] = dsn;
```
rouault, line 372: To please Coverity Scan...
```suggestion
        datasets[name] = std::move(dsn);
```

apps/gdalalg_raster_calc.cpp
```diff
@@ -0,0 +1,619 @@
...
+                return false;
+            }
+            name = "X";
+        }
+        else
+        {
+            name = input.substr(0, pos);
+        }
+
+        std::string dsn = (pos == std::string::npos) ? input : input.substr(pos + 1);
```
rouault, line 371:
```suggestion
        std::string dsn =
            (pos == std::string::npos) ? input : input.substr(pos + 1);
```

PR #11854, opened 2025-02-14
jpegxl.py: Minor cleanup.

### LINKED ISSUE #11833: gdrivers/jpegxl.py: test_jpegxl_rasterio has an unused variable i
### What is the bug?

Minor issue in the test. It doesn't currently hurt anything, but may be that the test doesn't cover as much as it should or the test could be slightly simplified if there aren't actually two things to test.

The `i` is not used in the for loop, so each iteration is the same. What was supposed to be changing?

https://github.com/OSGeo/gdal/blob/ebf06798dfe7e0e2a7d008d41fbbfc32ed15c0ec/autotest/gdrivers/jpegxl.py#L236-L243

Introduced here: https://github.com/OSGeo/gdal/commit/27203405bc33723163b3c9a66a7f8dbb82be3e9f#diff-e9e8ff60a80f6c886611c34cd7f719ccf067cd9f4f645fb1bb3cd669de194a89R205

By inspection.

### Versions and provenance

In current master head.

### Additional context

_No response_

schwehr:
- Remove extra empty lines
- Rename `outfilename` and `tmpfilename` -> `filename`
- Rename `/vsimem/out.jxl` to be specific to each test function
- Use `with open()` to load data with `.read()`
- Fix spelling: `withing` -> `within`
- Add a comment explaining the double run in `test_jpegxl_rasterio()`

Fixes #11833

### DIFF
--- a/autotest/gdrivers/jpegxl.py
+++ b/autotest/gdrivers/jpegxl.py
@@ -70,9 +70,9 @@ def test_jpegxl_rgba():
 
 
 @pytest.mark.parametrize("lossless", ["YES", "NO", None])
-def test_jpegxl_rgba_lossless_param(lossless):
+def test_jpegxl_rgba_lossless_param(tmp_vsimem, lossless):
     src_ds = gdal.Open("../gcore/data/stefan_full_rgba.tif")
-    filename = "/vsimem/jpegxl_rgba_lossless_param.jxl"
+    filename = tmp_vsimem / "jpegxl_rgba_lossless_param.jxl"
     options = []
     if lossless:
         options += ["LOSSLESS=" + lossless]
@@ -90,9 +90,6 @@ def test_jpegxl_rgba_lossless_param(lossless):
     else:
         assert cs == src_ds.GetRasterBand(1).Checksum()
 
-    ds = None
-    gdal.GetDriverByName("JPEGXL").Delete(filename)
-
 
 def test_jpegxl_rgba_lossless_no_but_lossless_copy_yes():
     src_ds = gdal.Open("../gcore/data/stefan_full_rgba.tif")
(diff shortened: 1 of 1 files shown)


schwehr:
Just an FYI: I put the port of jpegxl.py to google3 here...

https://github.com/schwehr/gdal-autotest2/blob/master/python/gdrivers/jpegxl_test.py

dbaston:
> What if there was a subclass of or thing exactly like pathlib.Path that worked with the vsi system and it understood what is possible with vsi paths?

The path manipulation functions should work as-is, but I can see writing a subclass to get methods like `.exists()` to work with in-memory files etc.

autotest/gdrivers/jpegxl.py
```diff
@@ -70,9 +70,9 @@ def test_jpegxl_rgba():
 
 
 @pytest.mark.parametrize("lossless", ["YES", "NO", None])
-def test_jpegxl_rgba_lossless_param(lossless):
+def test_jpegxl_rgba_lossless_param(tmp_vsimem, lossless):
     src_ds = gdal.Open("../gcore/data/stefan_full_rgba.tif")
-    filename = "/vsimem/jpegxl_rgba_lossless_param.jxl"
+    filename = str(tmp_vsimem / "jpegxl_rgba_lossless_param.jxl")
```
dbaston, line 75: Harmless, and I wouldn't bother changing it, but there should be no need for `str` here? Or is something still lacking support for `os.PathLike` ?

rouault: you're right. I've removed str() usage in ``str(tmp_vsimem / "xxx")``.  The only places where I needed to use str() is when appending a .aux.xml suffix

schwehr:
Thanks Even for adding that commit and merging it!

PR #11856, opened 2025-02-16
DXF: Don't hide block entities on layer 0 when that layer is frozen

atlight:
## What does this PR do?

Layer 0 is a magical layer in DWG/DXF files when it comes to block insertion. Freezing layer 0 in AutoCAD does not affect block entities on layer 0. This change brings GDAL into line with that behaviour.

None - own motion

## Tasklist

 - [x] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments
 - [ ] ADD YOUR TASKS HERE

## Environment

* OS:
* Compiler:

### DIFF
--- a/autotest/ogr/ogr_dxf.py
+++ b/autotest/ogr/ogr_dxf.py
@@ -3881,7 +3881,7 @@ def test_ogr_dxf_53():
 # Test frozen and off layers
 
 
-def test_ogr_dxf_54():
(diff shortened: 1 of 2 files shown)


jratike80:
Perhaps you could also create an issue that this PR resolves. If I understand right the current behaviour is a bug.

rouault (approved):

PR #11857, opened 2025-02-17
GTiff: CreateCopy(): fix incorrect writing of mask band when COPY_SRC_OVERVIEWS=YES and INTERLEAVE=BAND (master only)

### LINKED ISSUE #11851: GTiff driver: CreateCopy doesn't preserve mask on band interleaved COGs
### What is the bug?

After create a COG via CreateCopy() with the GTIFF driver and an explicit mask, the mask read back doesn't preserve the original bytes. When we read back the mask, the mask is expanded from the 1-bit mask to the 8-bit mask through the pathway of gtiffoddbitsband.h which is the correct path. However, the mask is not a 1-bit mask when I add logging in this file; instead, it's the original 8-bit mask (with 255s and 0s). So when it gets expanded to 8-bits, it's no longer the original mask:
E.g., 
The mask 255 0 255 255 is stored verbatim and becomes 255 255 ... 255 255 0 0 ... 0 0 255 255
Instead of being stored as 128 (10000000) 192 (11000000) which becomes 255 0 255 255.

So my guess is that somewhere along the line we either copy the mask without turning it back into a bitmask, or we lose metadata tracking that the mask is encoded a certain way. 

The issue *does not* occur when either COPY_SRC_OVERVIEWS="YES" is removed or INTERLEAVE is set to PIXEL (it only fails on band interleaving). This is using the 'GTiff' driver.

Example gist: https://gist.github.com/orglofch/40cf4d9136978b0384141cc74d1039b6.

The final mask is 0 0 255 255 instead of 0 255 255 255.

### Versions and provenance

In current master head. This failure only started occurring after https://github.com/OSGeo/gdal/commit/3bdc81a205bad8342688873e4ed5716f1208e8f5.

### Additional context

_No response_

rouault:
Fixes #11851

### DIFF
--- a/autotest/gcore/cog.py
+++ b/autotest/gcore/cog.py
@@ -1996,18 +1996,19 @@ def test_cog_preserve_ALPHA_PREMULTIPLIED_on_copy(tmp_vsimem):
 
 
 @gdaltest.enable_exceptions()
-def test_cog_write_interleave_tile(tmp_vsimem):
+@pytest.mark.parametrize("INTERLEAVE", ["TILE", "BAND"])
+def test_cog_write_interleave_tile_or_band(tmp_vsimem, INTERLEAVE):
     out_filename = str(tmp_vsimem / "out.tif")
 
     with gdal.quiet_errors():
         gdal.GetDriverByName("COG").CreateCopy(
             out_filename,
             gdal.Open("data/rgbsmall.tif"),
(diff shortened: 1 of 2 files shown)


schwehr (approved):
Thank you! Confirmed this fixes the original issue represented by the test that @orglofch shared.

PR #11862, opened 2025-02-18
Add RFC 108 text: driver removals for GDAL 3.11

rouault:


doc/source/development/rfc/rfc108_driver_removal_3_11.rst
```diff
@@ -0,0 +1,160 @@
...
+- raster:
+
+  * ADRG: defense-related format. Write support was mostly to test
+    the read side of the driver
+  * ELAS: not sure what it is about and that this is actively used
+  * PAux: not sure what it is about and that this is actively used
+  * MFF: not sure what it is about and that this is actively used
+  * MFF2/HKV: not sure what it is about and that this is actively used
+  * LAN: not sure what it is about and that this is actively used
+  * BYN: Canadian grid-related format. Write support doesn't seem useful
```
jratike80, line 92: LAN is an old ERDAS format before HFA (.img).  It is a simple format, probably "raw", and it does not support compression. Write support doesn't seem useful.

doc/source/development/rfc/rfc108_driver_removal_3_11.rst
```diff
@@ -0,0 +1,160 @@
...
+
+- raster:
+
+  * ADRG: defense-related format. Write support was mostly to test
+    the read side of the driver
+  * ELAS: not sure what it is about and that this is actively used
+  * PAux: not sure what it is about and that this is actively used
+  * MFF: not sure what it is about and that this is actively used
+  * MFF2/HKV: not sure what it is about and that this is actively used
+  * LAN: not sure what it is about and that this is actively used
```
rouault, line 91:
```suggestion
  * LAN: old ERDAS format before HFA (.img). "raw format", does not support compression. Write support doesn't seem useful.
```

doc/source/development/rfc/rfc108_driver_removal_3_11.rst
rouault: I've integrated your comment. Thanks

doc/source/development/rfc/rfc108_driver_removal_3_11.rst
```diff
@@ -0,0 +1,160 @@
...
+
+GDAL is an ever growing code base, with probably more than 50% of the code
+base being used by less of 1% of users. This is isn't very sustainable, and quite
+demotivating for current maintainers who know nothing/little about most drivers,
+have no evidence that they are used, and regularly stumble upon them when doing
+whole codebase enhancements.
+The decision of which driver to removal comes from a mix of "intuition" and
+community feedback from
+https://docs.google.com/spreadsheets/d/12yk0p8rRK4rAYO_VwXboatWkiILAjE2Cepu32lsHhM8/edit?gid=0#gid=0
+
```
dbaston, line 32: Why not consider available data such as [traffic to documentation pages](https://rpubs.com/dbaston/1267034)? This would show e.g. ~90 hits/month to the documentation for the TIGER driver, which is more than retained drivers such as JSONFG.

doc/source/development/rfc/rfc108_driver_removal_3_11.rst
```diff
@@ -0,0 +1,160 @@
...
+----------
+
+GDAL is an ever growing code base, with probably more than 50% of the code
+base being used by less of 1% of users. This is isn't very sustainable, and quite
+demotivating for current maintainers who know nothing/little about most drivers,
+have no evidence that they are used, and regularly stumble upon them when doing
+whole codebase enhancements.
+The decision of which driver to removal comes from a mix of "intuition" and
+community feedback from
+https://docs.google.com/spreadsheets/d/12yk0p8rRK4rAYO_VwXboatWkiILAjE2Cepu32lsHhM8/edit?gid=0#gid=0
```
dbaston, line 31: The community was [solicited for feedback](https://lists.osgeo.org/pipermail/gdal-dev/2025-January/060111.html) about removing _write_ support for certain drivers, with a link to this spreadsheet. The spreadsheet is more broad in its scope and indicates complete removal of certain drivers. The spreadsheet shows Interlis as the only vector driver being considered for removal, but this RFC proposes to remove 5 drivers (of which Interlis is not one?). It's not easy to understand the connection between the community feedback and the list in this RFC.

doc/source/development/rfc/rfc108_driver_removal_3_11.rst
jratike80: TIGER may be a special case and hit statistics contain mishits because most TIGER data are not in that format.

JSONFG is a format of the future, not yet accepted as an OGC standard, but it will be soon. TIGER is a format from the past. The choise which one to keep is clear.

dbaston: I'm not suggesting that we remove JSONFG because it has less web traffic than TIGER. But I don't think we should disregard that traffic either. Given relatively recent [discussion](https://forum.ipums.org/t/basic-questions-about-digital-block-geometry-for-1980-to-2000-for-us-territories/3994) or [usage](https://www.mdpi.com/2571-6255/6/8/300) of data in this format, I would not write off the site traffic as accidental.

doc/source/development/rfc/rfc108_driver_removal_3_11.rst
rouault: Interlis write support removal has already been removed prior to this RFC, given blessing from his maintainer.

> It's not easy to understand the connection between the community feedback and the list in this RFC.

Don't try too hard to find logic ;-)

doc/source/development/rfc/rfc108_driver_removal_3_11.rst
jratike80: In the "usage" link they have used `Data: County subdivision lines from US Census Bureau 2018 TIGER/line files`, thus shapefiles, but it does not mean that some other researcher wants and needs to use historical TIGER data for a good reason, like in the "discussion" link.

doc/source/development/rfc/rfc108_driver_removal_3_11.rst
```diff
@@ -0,0 +1,160 @@
...
+
+Funding
+-------
+
+Funded by GDAL Sponsorship Program (GSP) (this RFC text)
+
+Voting history
+--------------
+
+TBD
```
rouault, line 151:
```suggestion
+1 from PSC members JukkaR, JavierJS, KurtS, HowardB, DanielM, and -0 from DanB.
```

doc/source/development/rfc/rfc108_driver_removal_3_11.rst
```diff
@@ -0,0 +1,160 @@
...
+
+=====================================================================
+RFC 108: Driver removals for GDAL 3.11
+=====================================================================
+
+============== =============================================
+Author:        Even Rouault
+Contact:       even.rouault @ spatialys.com
+Started:       2025-02-18
+Status:        Draft
```
rouault, line 11:
```suggestion
Status:        Adopted, implemented
```

PR #11863, opened 2025-02-18
CSV: emit error when unbalanced double quotes are detected

### LINKED ISSUE #11845: Silent data loss for certain kinds of corrupt CSV input
### What is the bug?

I originally filed this as [geopandas issue #3511](https://github.com/geopandas/geopandas/issues/3511), but it appears that the behavior I'm seeing is specific to GDAL, and so they asked me to duplicate the issue over here.

I have observed that for certain kinds of corrupt CSV input, there is silent data loss from the GeoPandas `read_file()` method. The behavior persists using both fiona and pyogrio.  The reason for that is apparently due to the common GDAL implementation underneath both of those Python libraries.

I've developed a minimal Python test case that demonstrates the problem, documented in [geopandas issue #3511](https://github.com/geopandas/geopandas/issues/3511). Since I'm not using GDAL directly, I'm a little unsure what sort of code example would be best here.  I get the same results from both fiona and pyogrio, even though their implementations are somewhat different, so I'm hoping that a clear example might be enough for you to reproduce.  

This 1-row CSV file is valid:

```
Column
POINT (4.444444 -44.444444)
```

This 1-row CSV file can be loaded successfully, even though it is corrupt &mdash; it has a trailing quote (`"`):

```
Column
"POINT (6.666666 -77.777777)""
```

However, if you combine the rows together, you sometimes get silent data loss.  With the rows in this order, only the second row is returned:

```
Column
"POINT (6.666666 -77.777777)""
POINT (4.444444 -44.444444)
```

With the rows in this order, both are returned:

```
Column
POINT (4.444444 -44.444444)
"POINT (6.666666 -77.777777)""
```

There may be other scenarios where there is silent data loss, but this is the one that I was able to reproduce reliably on my side.

### Versions and provenance

The minimal test case in [geopandas issue #3511](https://github.com/geopandas/geopandas/issues/3511) was developed on my M3 Macbook running Sequoia, but the original problem was discovered on Ubuntu 24.04, so it does not appear to be platform-specific.  

Here's version info for GDAL on my Macbook, installed via Homebrew:
 
```
$ gdalinfo --version
GDAL 3.10.1, released 2025/01/08
```

On Ubuntu, we're using the libgdal34t64 package, which is at version 3.8.4+dfsg-3ubuntu3.

### Additional context

In the original data (which I can't share here), we lost > 2000 rows out of around 8500 rows in one particular CSV file. I was able to work around the behavior by normalizing the CSV using Python's `csv.reader` and `csv.writer` before passing it to GeoPandas. If I normalize using `dialect="excel", quoting=csv.QUOTE_ALL`, then all of the rows are returned as expected.  I'm uncertain why this works.

rouault:
Fixes #11845

### DIFF
--- a/autotest/ogr/data/csv/unbalanced_double_quotes.csv
+++ b/autotest/ogr/data/csv/unbalanced_double_quotes.csv
@@ -1,0 +1,3 @@
+id,txt
+1,"foo""
+2,bar
--- a/autotest/ogr/ogr_csv.py
+++ b/autotest/ogr/ogr_csv.py
@@ -3483,6 +3483,22 @@ def test_ogr_schema_override_wkt(tmp_vsimem):
 
 
 ###############################################################################
+# Test reading CSV with unbalanced double-quotes
+
+
+@gdaltest.enable_exceptions()
+def test_ogr_csv_unbalanced_double_quotes():
+
+    with ogr.Open("data/csv/unbalanced_double_quotes.csv") as ds:
+        lyr = ds.GetLayer(0)
+        with pytest.raises(
+            Exception,
+            match="CSV file has unbalanced number of double-quotes. Corrupted data will likely be returned",
+        ):
+            lyr.GetNextFeature()
+
+
+###############################################################################
 
 
 if __name__ == "__main__":
--- a/port/cpl_csv.cpp
+++ b/port/cpl_csv.cpp
@@ -706,6 +706,14 @@ CSVReadParseLineGeneric(void *fp, const char *(*pfnReadLine)(void *, size_t),
     {
         CPLError(CE_Failure, CPLE_OutOfMemory, "%s", e.what());
     }
+
+    if (bInString)
+    {
+        CPLError(CE_Failure, CPLE_AppDefined,
+                 "CSV file has unbalanced number of double-quotes. Corrupted "
+                 "data will likely be returned");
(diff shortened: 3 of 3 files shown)


autotest/ogr/data/csv/unbalanced_double_quotes.csv
```diff
@@ -0,0 +1,3 @@
+id,txt
+1,"foo""
+2,bar
```
jratike80, line 3: How do you define unbalanced?
`1,"foo""""` with 5 `"` is balenced and means string `foo"`?
Is it allowed that strings contain newlines?
```
1,"foo
2,bar"
```

rouault: > How do you define unbalanced?

Rule: all starting double quote should have a matching ending double quote, and repeated double-quote inside a string is an escaped double uote.
Unbalanced: when the above is not true in a file

rouault: > Is it allowed that strings contain newlines?
> 
> ```
> 1,"foo
> 2,bar"
> ```

yes, at least the OGR CSV driver allows it (which explains quite a bunch of complication in the code)

elpaso (approved):

PR #11865, opened 2025-02-18
Add a GDAL_RELEASE_NICKNAME define

rouault:


louwers (approved):

rouault:
Will be essentially backed out per https://github.com/OSGeo/gdal/pull/11899

PR #11880, opened 2025-02-22
Fix build when iterators are raw pointers

georgthegreat:
Similar to #11879, `typename Iterator` might be just a raw pointer. The most corrent way would be using `std::iterator_traits` to get the type properly, but it seems that current static_cast can be simply removed.

### DIFF
--- a/apps/argparse/argparse.hpp
+++ b/apps/argparse/argparse.hpp
@@ -1007,8 +1007,7 @@ public:
     if ((dist = static_cast<std::size_t>(std::distance(start, end))) >=
         num_args_min) {
       if (num_args_max < dist) {
-        end = std::next(start, static_cast<typename Iterator::difference_type>(
-                                   num_args_max));
+        end = std::next(start, num_args_max));
       }
       if (!m_accepts_optional_like_value) {
         end = std::find_if(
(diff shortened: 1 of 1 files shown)


rouault:
There's a typo in this commit (extra parenthesis), but I would be more confortable if this change was submitted upstream to https://github.com/p-ranav/argparse and accepted there prior to here

georgthegreat:
Thanks!
I have noticed the typo too, but was waiting for two other PRs to be merged.
I will create the PR to upstream, let's keep this one open for now.

georgthegreat:
I filed an upstream PR here:
https://github.com/p-ranav/argparse/pull/399

georgthegreat:
@rouault it seems that upstream is not quite active.
Maybe we can accept this change without upstream accepting it?
The code in gdal is different from upstream anyway.

rouault:
can you precise which environment requires this change ? GDAL tests on quite a large number of platforms and we haven't hit such issue

georgthegreat:
We use patched libc++ with iterators replaced with raw pointers.
You can observe our sources e. g. here:
https://github.com/yandex/perforator/blob/c680583aea53e44ebee0189ac5d8b146f09bb5d2/contrib/libs/cxxsupp/libcxx/include/__config_site#L21

gdal is used internally in our monorepo, but libc++ is the same.

jratike80:
> We use patched libc++ with iterators replaced with raw pointers.

Would the change help anybody else than you in the GDAL community?

rouault (approved):

georgthegreat:
> > We use patched libc++ with iterators replaced with raw pointers.
> 
> Would the change help anybody else than you in the GDAL community?

Yes, it will make the code simpler by removing unnecessaries.

PR #11896, opened 2025-03-01
zarr_driver.py: Minor cleanup.

schwehr:
- Break up massive wkt string in `test_zarr_read_crs`
- Wrap a few places with `gdal.quiet_errors` where ZARR_V3 gives "Warning 1: fill_value = null is invalid" on open
- Remove a commented out debugging print

Issue #9949, opened 2024-05-16
autotest: set up Python docstring testing


dbaston:
### Feature description

pytest can be used to invoke the examples in our docstrings: https://docs.pytest.org/en/7.1.x/how-to/doctest.html

This could be a useful way to make sure the examples run as expected.

dbaston:
I've done some exploration on this; see https://github.com/dbaston/gdal/commit/b8168d56c3afd9d8845845de29105f909b387c9e

It seems to work fairly well, we can invoke tests from the build root using e.g.  `pytest -s --doctest-modules swig/python/osgeo/gdal.py -k Algorithm`

No extra modules are needed.

rouault:
> It seems to work fairly well, we can invoke tests from the build root using e.g. `pytest -s --doctest-modules swig/python/osgeo/gdal.py -k Algorithm`

nice!

PR #11919, opened 2025-03-04
Add CPLQuietWarningsErrorHandler

dbaston:
As discussed in https://github.com/OSGeo/gdal/pull/11896#discussion_r1976500777

### DIFF
--- a/swig/include/python/gdal_python.i
+++ b/swig/include/python/gdal_python.i
@@ -4957,7 +4957,7 @@ def config_option(key, value, thread_local=True):
 
 @contextlib.contextmanager
 def quiet_errors():
-    """Temporarily install an error handler that silents all warnings and errors.
+    """Temporarily install an error handler that silences all warnings and errors.
 
        Returns
        -------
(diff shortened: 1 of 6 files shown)


swig/include/python/gdal_python.i
```diff
@@ -4975,6 +4975,28 @@ def quiet_errors():
...
+       .. versionadded: 3.11
+
+       Returns
+       -------
+            A context manager
+
+       Example
+       -------
+
+       >>> with gdal.ExceptionMgr(useExceptions=False), gdal.quiet_errors():
```
rouault, line 4991: I assume:
```suggestion
       >>> with gdal.ExceptionMgr(useExceptions=False), gdal.quiet_warnings():
```

maybe we should have a test for that?

dbaston: I tested it manually by modifying https://github.com/OSGeo/gdal/issues/9949 . I can write an automated test (would need to involve `subprocess`, I guess?)

dbaston: (test added in 1c404c681301476ff25843e3b7a080a693812d22)

rouault (approved):

PR #11923, opened 2025-03-05
[cli][gdal_contour] Port gdal_contour to cli

elpaso:
```bash
$ gdal raster contour --help

Usage: gdal raster contour [OPTIONS] <INPUT> <OUTPUT>

Creates a vector contour from a raster elevation model (DEM).

Positional arguments:

Common Options:

Options:
  -b, --band <BAND>                                    Specify input band number. (default: 1)
  -l, --nln, --layer <LAYER>                           Layer name
  --elevation-name <ELEVATION-NAME>                    Name of the elevation field.
  --min-name <MIN-NAME>                                Name of the minimum elevation field.
  --max-name <MAX-NAME>                                Name of the maximum elevation field.
  --3d                                                 Force production of 3D vectors instead of 2D.
  --srcnodata <SRCNODATA>                              Input pixel value to treat as 'nodata'.
  --interval <INTERVAL>                                Elevation interval between contours.
                                                       Mutually exclusive with --levels, --exp-base
  --levels <LEVELS>                                    List of contour levels to generate. [may be repeated]
                                                       Mutually exclusive with --interval, --exp-base
  -e, --exp-base <EXP-BASE>                            Base for exponential contour level generation.
                                                       Mutually exclusive with --interval, --levels
  --off, --offset <OFFSET>                             Offset to apply to contour levels.
  -p, --polygonize                                     Create polygons instead of lines.
  --group-transactions <GROUP-TRANSACTIONS>            Group n features per transaction (default 100 000).

Advanced Options:
  --oo, --open-option <KEY=VALUE>                      Open options [may be repeated]

For more details, consult https://gdal.org/programs/gdal_raster_contour.html

```


apps/gdalalg_raster_contour.cpp
```diff
@@ -0,0 +1,221 @@
...
+                                  &hBand, &hDstDS, &hLayer);
+
+    if (eErr == CE_None)
+    {
+        eErr = GDALContourGenerateEx(hBand, hLayer, papszStringOptions,
+                                     pfnProgress, pProgressData);
+    }
+
+    // Print debug papszStringOptions
+    for (int i = 0; papszStringOptions[i] != nullptr; i++)
```
elpaso, line 207: to be removed...

elpaso:
@dbaston @hobu @jratike80 (from a discussion with @rouault) the contour command consumes a raster and produces a vector, I thought it should go into the raster `gdal raster contour ...` category but I'm open to suggestions, for example should it be a top level command `gdal contour ...` ?

apps/gdalalg_raster_contour.cpp
```diff
@@ -0,0 +1,221 @@
...
+    // gdal_contour specific options
+    AddArg("band", 'b', _("Specify input band number."), &m_band).SetDefault(1);
+    AddLayerNameArg(&m_outputLayerName).AddAlias("nln");
+    AddArg("aelev", 'a', _("Name of the elevation field."),
+           &m_elevAttributeName);
+    AddArg("amin", 0, _("Name of the minimum elevation field."), &m_amin);
+    AddArg("amax", 0, _("Name of the maximum elevation field."), &m_amax);
+    AddArg("3d", 0, _("Force production of 3D vectors instead of 2D."), &m_3d);
+    AddArg("inodata", 0, _("Ignore nodata values in the input raster."),
+           &m_iNodata);
```
rouault, line 60: I would not bind that one for now, unless 50% + of users beg for it. It seems quite esoteric, and I don't know why contour would need that one whereas other utilities have no similar option . At worse users can create a VRT with no nodata
```suggestion
```

apps/gdalalg_raster_contour.h
```diff
@@ -0,0 +1,84 @@
...
+/*                     GDALRasterContourAlgorithm                       */
+/************************************************************************/
+
+class GDALRasterContourAlgorithm final : public GDALAlgorithm
+{
+  public:
+    static constexpr const char *NAME = "contour";
+    static constexpr const char *DESCRIPTION =
+        "Creates a vector contour from a raster elevation model (DEM).";
+    static constexpr const char *HELP_URL =
```
jratike80, line 30: Height/elevation may be the most common use case, but the algorithm does not know if the data contains heights, temperatures, pressures or whatever. "Elevation" appears throughout in the files and I think it can be so, but a user documentation should mention other sort of interpretations of the source data.

jratike80: "gdal raster contour" feels natural to me, but will the other "dem" tools be similarly "gdal raster hillshade" and so on, even they produce raster output?

I did not find where this --usage comes, but the output is vector even here.

> Positional arguments:
>   -i, --input <INPUT>                                  Input raster dataset [required]
>   -o, --output <OUTPUT>                                Output raster dataset (created by algorithm) [required]

apps/gdalalg_raster_contour.cpp
```diff
@@ -0,0 +1,221 @@
...
+    AddArg("snodata", 0, _("Set the nodata value for the output vector."),
+           &m_sNodata);
```
rouault, line 62: The description is incorrect. gdal_contour doc says "Input pixel value to treat as "nodata"
I guess we should call it "srcnodata" for clarity
But I'm not sure either we need it at all for now, as that could be seen as a pre-processing step to apply.

apps/gdalalg_raster_contour.cpp
```diff
@@ -0,0 +1,221 @@
...
+        .AddMetadataItem(GAAMDI_REQUIRED_CAPABILITIES,
+                         {GDAL_DCAP_RASTER, GDAL_DCAP_CREATECOPY});
+    AddOpenOptionsArg(&m_openOptions);
+    AddInputFormatsArg(&m_inputFormats)
+        .AddMetadataItem(GAAMDI_REQUIRED_CAPABILITIES, {GDAL_DCAP_RASTER});
+    AddInputDatasetArg(&m_inputDataset, openForMixedRasterVector
+                                            ? GDAL_OF_RASTER | GDAL_OF_VECTOR |
+                                                  GDAL_OF_MULTIDIM_RASTER
+                                            : GDAL_OF_RASTER);
+    AddOutputDatasetArg(&m_outputDataset, GDAL_OF_RASTER);
```
rouault, line 48: to answer @jratike80 question:
```suggestion
    AddOutputDatasetArg(&m_outputDataset, GDAL_OF_VECTOR);
```

apps/gdalalg_raster_contour.cpp
```diff
@@ -0,0 +1,221 @@
...
+#ifndef _
+#define _(x) (x)
+#endif
+
+/************************************************************************/
+/*          GDALRasterContourAlgorithm::GDALRasterContourAlgorithm()    */
+/************************************************************************/
+
+GDALRasterContourAlgorithm::GDALRasterContourAlgorithm(
+    bool openForMixedRasterVector)
```
rouault, line 32: you don't need that flag for contour (this is only useful for things like "gdal info" that tries to forward either to "gdal raster info" or "gdal vector info")
```suggestion
)
```

apps/gdalalg_raster_contour.cpp
```diff
@@ -0,0 +1,221 @@
...
+    : GDALAlgorithm(NAME, DESCRIPTION, HELP_URL), m_outputLayerName("contour"),
+      m_elevAttributeName(""), m_amin(""), m_amax(""), m_levels{}
+{
+
+    AddProgressArg();
+    AddOutputFormatArg(&m_outputFormat)
+        .AddMetadataItem(GAAMDI_REQUIRED_CAPABILITIES,
+                         {GDAL_DCAP_RASTER, GDAL_DCAP_CREATECOPY});
```
rouault, line 40:
```suggestion
                         {GDAL_DCAP_VECTOR});
```

apps/gdalalg_raster_contour.cpp
```diff
@@ -0,0 +1,221 @@
...
+    AddOpenOptionsArg(&m_openOptions);
+    AddInputFormatsArg(&m_inputFormats)
+        .AddMetadataItem(GAAMDI_REQUIRED_CAPABILITIES, {GDAL_DCAP_RASTER});
+    AddInputDatasetArg(&m_inputDataset, openForMixedRasterVector
+                                            ? GDAL_OF_RASTER | GDAL_OF_VECTOR |
+                                                  GDAL_OF_MULTIDIM_RASTER
+                                            : GDAL_OF_RASTER);
```
rouault, line 47:
```suggestion
    AddInputDatasetArg(&m_inputDataset, openForMixedRasterVector, GDAL_OF_RASTER);
```

apps/gdalalg_raster_contour.cpp
```diff
@@ -0,0 +1,221 @@
...
+    AddArg("amax", 0, _("Name of the maximum elevation field."), &m_amax);
+    AddArg("3d", 0, _("Force production of 3D vectors instead of 2D."), &m_3d);
+    AddArg("inodata", 0, _("Ignore nodata values in the input raster."),
+           &m_iNodata);
+    AddArg("snodata", 0, _("Set the nodata value for the output vector."),
+           &m_sNodata);
+    AddArg("interval", 0, _("Elevation interval between contours."),
+           &m_interval)
+        .SetMutualExclusionGroup("levels");
+    AddArg("fl", 0, _("List of contour levels to generate."), &m_levels)
```
rouault, line 66:
```suggestion
    AddArg("levels", 0, _("List of contour levels to generate."), &m_levels)
```

apps/gdalalg_raster_contour.h
elpaso: It's consistent with https://gdal.org/en/stable/programs/gdal_contour.html though. 
We should change that too.

apps/gdalalg_raster_contour.cpp
```diff
@@ -0,0 +1,221 @@
...
+                                                  GDAL_OF_MULTIDIM_RASTER
+                                            : GDAL_OF_RASTER);
+    AddOutputDatasetArg(&m_outputDataset, GDAL_OF_RASTER);
+    AddCreationOptionsArg(&m_creationOptions);
+
+    // gdal_contour specific options
+    AddArg("band", 'b', _("Specify input band number."), &m_band).SetDefault(1);
+    AddLayerNameArg(&m_outputLayerName).AddAlias("nln");
+    AddArg("aelev", 'a', _("Name of the elevation field."),
+           &m_elevAttributeName);
```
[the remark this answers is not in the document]
elpaso: We need to decide whether we want to keep the highest possible option compatibility with the legacy options in  https://gdal.org/en/stable/programs/gdal_contour.html or not. 
I assumed yes but I'm happy to change my assumption.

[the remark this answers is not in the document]
rouault: > We need to decide whether we want to keep the highest possible option compatibility with the legacy options

no, we don't need/want that. This is a sort of reboot and a unique chance to get things "right"
I'm not saying that my proposal is necessary right...

apps/gdalalg_raster_contour.cpp
elpaso: done

apps/gdalalg_raster_contour.cpp
elpaso: done

apps/gdalalg_raster_contour.cpp
elpaso: done

apps/gdalalg_raster_contour.cpp
elpaso: removed

apps/gdalalg_raster_contour.cpp
elpaso: renamed but kept

rouault:
We lack a documentation page (and an entry for the man page in doc/source/conf.py)

rouault (approved):

PR #11930, opened 2025-03-06
Add gdal raster resize

### LINKED ISSUE #11908: gdal reproject: add "size" to match gdalwarp '-ts', and allow (one) value to be 0
### Feature description

I want to be able to set the target size, '-te' and '-ts' is my normal preference, and being able to have one of the sizes set to 0 is a real benefit.   

Certainly interested to help implement this, though I guess if the trad apps and library versions remain there's no afaic. I wasn't much able to be active when the RFC was discussed. 

### Additional context

 (I wonder if I'm missing something, because there doesn't seem to be the equivalent of '-outsize' for 'translate'  conversion either?)

rouault:
Fixes #11908

### DIFF
--- a/apps/gdalalg_raster_resize.cpp
+++ b/apps/gdalalg_raster_resize.cpp
@@ -1,0 +1,103 @@
+/******************************************************************************
+ *
+ * Project:  GDAL
+ * Purpose:  "resize" step of "raster pipeline"
+ * Author:   Even Rouault <even dot rouault at spatialys.com>
+ *
+ ******************************************************************************
+ * Copyright (c) 2025, Even Rouault <even dot rouault at spatialys.com>
+ *
+ * SPDX-License-Identifier: MIT
+ ****************************************************************************/
+
+#include "gdalalg_raster_resize.h"
+
+#include "gdal_priv.h"
+#include "gdal_utils.h"
+
+//! @cond Doxygen_Suppress
+
+#ifndef _
+#define _(x) (x)
+#endif
+
+/************************************************************************/
+/*        GDALRasterResizeAlgorithm::GDALRasterResizeAlgorithm()        */
+/************************************************************************/
+
+GDALRasterResizeAlgorithm::GDALRasterResizeAlgorithm(bool standaloneStep)
+    : GDALRasterPipelineStepAlgorithm(NAME, DESCRIPTION, HELP_URL,
+                                      standaloneStep)
+{
+    auto &sizeArg = AddArg("size", 0, _("Target size in pixels"), &m_size)
+                        .SetMinCount(2)
+                        .SetMaxCount(2)
+                        .SetRequired()
+                        .SetRepeatedArgAllowed(false)
+                        .SetDisplayHintAboutRepetition(false)
+                        .SetMetaVar("<width>,<height>")
+                        .SetMutualExclusionGroup("resolution-size");
+    sizeArg.AddValidationAction(
+        [&sizeArg]()
+        {
+            const auto &val = sizeArg.Get<std::vector<int>>();
+            CPLAssert(val.size() == 2);
+            if (!(val[0] >= 0 && val[1] >= 0))
+            {
+                CPLError(CE_Failure, CPLE_AppDefined,
+                         "Target size should be positive or 0.");
+                return false;
+            }
+            return true;
+        });
+
+    AddArg("resampling", 'r', _("Resampling method"), &m_resampling)
+        .SetChoices("nearest", "bilinear", "cubic", "cubicspline", "lanczos",
+                    "average", "mode")
+        .SetDefault("nearest")
+        .SetHiddenChoices("near");
+}
+
+/************************************************************************/
+/*              GDALRasterResizeAlgorithm::RunStep()                    */
+/************************************************************************/
+
+bool GDALRasterResizeAlgorithm::RunStep(GDALProgressFunc, void *)
+{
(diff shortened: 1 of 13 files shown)


dbaston:
I wasn't sure what "resize" meant in this context, but my initial guess was that it changed the extent of the image. I usually think of changing resolution as "resampling" ([ArcGIS](https://pro.arcgis.com/en/pro-app/latest/tool-reference/data-management/resample.htm) and [grass](https://grass.osgeo.org/grass-stable/manuals/r.resample.html) appear to use the term for changes in resolution without changes in extent.) But other tools such as [terra](https://rdrr.io/github/rspatial/terra/man/resample.html) use the term "resampling" more broadly, to mean "anything gdalwarp does". Something like "set-pixel-size" may be more clear. Would be curious what others think about this.

rouault:
> Would be curious what others think about this.

CC @mdsumner @hobu as this was discussed in https://github.com/OSGeo/gdal/issues/11908

mdsumner:
"resize" is clear to me in the sense of changing the number of pixels in each direction, maintaining the extent (but totally appreciate other perspectives)

I don't like thinking in terms of "resolution", when I'm at the very first stages of exploring I want "just enough pixels to make a picture", so I'm constantly using the high level GDALWarp() function as the first step, because I don't have to think about the structure of the source data where you can make very poor choices for "resolution" depending on the units and the actual extent of the data. I think that is a real missed opportunity for many users and it doesn't get used or understood enough. 

changing the extent to me is "crop" (but a problematic term because you could crop in the other direction and extend the domain)

(I didn't know terra used the warper in resample ... I PR'd in  for `project(by_util = TRUE)` because I didn't know another way to get the high level warp via terra)

mdsumner:
"resize" as the concept makes sense to me in terms of the arguments '-outsize' and '-ts' - in R we use "dimension" because there's a generic `dim()` that returns the size-in-pixels of an array, but of course Python heavily uses the term "shape" and I've started getting used to that. I don't think "reshape" would make sense, though.  dimension/shape/size all kind of conflate with "rank" too, I don't think there's any easy way out of these terminology struggles

apps/gdalalg_raster_resize.cpp
```diff
@@ -0,0 +1,102 @@
...
+    sizeArg.AddValidationAction(
+        [&sizeArg]()
+        {
+            const auto &val = sizeArg.Get<std::vector<int>>();
+            CPLAssert(val.size() == 2);
+            if (!(val[0] >= 0 && val[1] >= 0))
+            {
+                CPLError(CE_Failure, CPLE_AppDefined,
+                         "Target size should be positive or 0.");
+                return false;
```
jratike80, line 49: Is is checked that the target size is in integers, or would the value be rounded or truncated automatically?

doc/source/programs/gdal_raster_pipeline.rst
```diff
@@ -94,6 +94,18 @@ Details for options can be found in :ref:`gdal_raster_edit_subcommand`.
...
+* resize [OPTIONS]
+
+.. code-block::
+
+    Resize a raster dataset.
+
+    Options:
+      --size <width>,<height>                              Target size in pixels [required]
+      -r, --resampling <RESAMPLING>                        Resampling method. RESAMPLING=nearest|bilinear|cubic|cubicspline|lanczos|average|mode
+
```
jratike80, line 106: Should we mention what is the default? I guess that nearest is the default.

doc/source/programs/gdal_raster_resize.rst
```diff
@@ -0,0 +1,90 @@
...
+================================================================================
+"gdal raster resize" sub-command
+================================================================================
+
+.. versionadded:: 3.11
+
+.. only:: html
+
+    Resize a raster dataset.
+
```
jratike80, line 12: Would it be good to add "without changing the georeferenced extents"? It may feel obvious, but I know that for some users it is not.

jratike80:
> Something like "set-pixel-size" may be more clear. Would be curious what others think about this.

This ESRI user has an issue due to being able to set only the pixel size, not the exact number of pixels:
https://gis.stackexchange.com/questions/490766/how-to-increase-the-number-of-cells-in-my-raster-without-resampling

apps/gdalalg_raster_resize.cpp
rouault:
```
$ gdal raster resize --size=10.5,10 byte.tif out.tif  --overwrite
ERROR 5: resize: Expected list of integer value for argument '--size', but got '10.5,10'.
Usage: gdal raster resize [OPTIONS] <INPUT> <OUTPUT>
Try 'gdal raster resize --help' for help.
```

doc/source/programs/gdal_raster_resize.rst
rouault: done

doc/source/programs/gdal_raster_pipeline.rst
rouault: good point. default added

rouault:
> changing the extent to me is "crop" (but a problematic term because you could crop in the other direction and extend the domain)

We have called it clip: https://gdal.org/en/latest/programs/gdal_raster_clip.html

jratike80:
> We have called it clip:
For a rectangular bbox crop and clip feels the same to me, but a possible future option to use cutline like in https://gdal.org/en/stable/programs/gdalwarp.html#cmdoption-gdalwarp-cutline fits better to clip.

PR #11931, opened 2025-03-07
GTiff: add a few missing default attributes in the creation option list

ctoney:
Adds the `default` attribute in the creation option list for a few options where it was missing in GTiff: PREDICTOR, TILED, BLOCKXSIZE, BIGTIFF. The values are set to match what is currently shown at https://gdal.org/en/latest/drivers/raster/gtiff.html. The motivation is so that default values can be read when the XML from `GDALGetDriverCreationOptionList()` is parsed.


jratike80:
I browsed the file throught and noticed variation in the boolean values. I know that both YES/NO and TRUE/FALSE do work, but do we prefer one or another?  Linenumbers and values:

```
1263  "default='NO'
1310 "default='FALSE'
1319 default='YES'
1367 default='FALSE'
1401 default='NO'
1426 default='NO'
1481 default='FALSE'
1484 default='FALSE'
```

There are also some more options missing the default value in the description
```
NBITS 
TFW
RBP
RPCTXT
BLOCKYSIZE
PHOTOMETRIC
ALPHA
```

A bunch of options beginning from the line 1404 probably do not have defaults

ctoney:
>There are also some more options missing the default value in the description

Not sure we can set some of these, e.g., `BLOCKYSIZE` default depends on whether it's tiled or striped. `NBITS` would depend on the apparent pixel type and don't think there would be defaults anyway. `PHOTOMETRIC`: "Default is `MINISBLACK`, but if the input image has 3 or 4 bands of Byte type, then RGB will be selected" (should we set `MINISBLACK` in the default attribute?). `RBP` and `RPCTXT`, it depends, can't tell from https://gdal.org/en/latest/drivers/raster/gtiff.html what the default should be assigned. Similar for `ALPHA` and `TFW`. It probably makes sense that some options will not have defaults in that list, but I'm not sure in some particular cases.

rouault (approved):

rouault:
@ctoney Can you amend your commit to use a valid email ? github refuse merging because of "Invalid email for web commit."

ctoney:
@rouault, sorry about that, does it work now? The commit shows to be associated with jctoney@gmail.com which is primary for my account. But I had recently set my work email, which is a secondary, as the display email in my public profile. Maybe that was the problem? Primary and public profile are both now set the same as jctoney@gmail.com, as it always was before.

rouault:
yes, worked that time

PR #11933, opened 2025-03-08
Doc: Fix axis order of points in osr.CoordinateTransformation examples

dbaston:
cc @ctoney

### DIFF
--- a/swig/include/python/docs/osr_coordinatetransformation_docs.i
+++ b/swig/include/python/docs/osr_coordinatetransformation_docs.i
@@ -32,8 +32,15 @@ tuple
 
 Examples
 --------
->>> ct.TransformBounds(-72.5, 44.2, -72.4, 44.3, 21)
-(7415356.140468472, -51238192.683464445, 7454323.154814391, -51210287.42581475)
+>>> wgs84 = osr.SpatialReference()
+>>> wgs84.ImportFromEPSG(4326)
+0
+>>> vt_sp = osr.SpatialReference()
+>>> vt_sp.ImportFromEPSG(5646)
+0
+>>> ct = osr.CoordinateTransformation(wgs84, vt_sp)
+>>> ct.TransformBounds(44.2,-72.5, 44.3, -72.4, 21)
+(1640416.6667, 619626.4287465283, 1666641.4901271078, 656096.7597360199)
 
 ";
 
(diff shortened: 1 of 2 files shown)


ctoney:
Thanks! Those are basically the edits I made, although repeating setup of the `ct` object for `TransformBounds()` and `TransformPoints()`  as you've done here is better. I didn't succeed all the way through build/review so was not confident enough that I had it right. I also wondered whether any additional files might be generated in the steps I was missing, that would need to be in the PR. Seeing this helps a lot.

swig/include/python/docs/osr_coordinatetransformation_docs.i
```diff
@@ -77,13 +84,19 @@ list
...
-[(7390620.052019633, -51202148.77747277, 0.0), (7387261.070131293, -51200373.68798984, 0.0)]
+>>> wgs84 = osr.SpatialReference()
+>>> wgs84.ImportFromEPSG(4326)
+0
+>>> vt_sp = osr.SpatialReference()
+>>> vt_sp.ImportFromEPSG(5646)
+0
+>>> ct = osr.CoordinateTransformation(wgs84, vt_sp)
+>>> ct.TransformPoints([(44.26, -72.58), (44.26, -72.59)])
+[(1619458.1108559777, 641509.1883246159, 0.0), (1616838.2913193079, 641511.9008312856, 0.0)]
```
jratike80, line 96: I get slightly different result (Proj  9.5.1). Maybe tell in the exaple that coordinates are lat-lon, here I use lon-lat because gdaltransform awaits that.
```
gdaltransform -s_srs epsg:4326 -t_srs epsg:5646
-72.58, 44.26
1619457.99756675 641509.73316664 0
```

ctoney:
My previous results match what is in the PR but were with PROJ 9.3.1

```py
>>> ct.TransformPoints([(44.26, -72.58), (44.26, -72.59)])
[(1619458.1108559777, 641509.1883246159, 0.0), (1616838.2913193079, 641511.9008312856, 0.0)]
```

swig/include/python/docs/osr_coordinatetransformation_docs.i
dbaston: Added a comment to note that inputs are lat/lon

rouault (approved):

PR #11935, opened 2025-03-09
Fix warnings when building against Poppler 25.03.00

rouault:


frmts/pdf/pdfio.h
```diff
@@ -54,8 +54,18 @@ class VSIPDFFileStream final : public BaseStream
     virtual void reset() override;
 #endif
 
+    static void resetNoCheckReturnValue(Stream *str)
+    {
+#if POPPLER_MAJOR_VERSION > 25 ||                                              \
+    (POPPLER_MAJOR_VERSION == 25 && POPPLER_MINOR_VERSION >= 2)
+        CPL_IGNORE_RET_VAL(str->reset());
```
dbaston, line 61: I was wondering if we should consider a macro definition for `CPL_IGNORE_RET_VAL` such that it can tolerate a `void` return....

```
#define CPL_IGNORE_RET_VAL(x) static_cast<void>(x)
```

But I'm guessing this situation is unusual enough that it's not worth it to change anything.

dbaston (approved):
I generally like to avoid macros, but if would the #ifelse

frmts/pdf/pdfio.h
rouault: I have some vague recollection that one of the analyzers didn't like a cast to void (but I may be wrong)

Issue #9923, opened 2024-05-14
Docs: Use CMake to build docs directly


dbaston:
> Would it make sense to remove the Makefile and use CMake directly to build the docs?

yes, such further enhancements are totally reasonable. I just wanted to do something relatively quick to get documentation building easily accessible from CMake. Would you like following up on this in subsequent PRs ?

(I'm wondering if we could have things not requiring a Unix environment so doc building is available from non-Cygwin Windows (provided folks have installed Doxygen, sphinx, etc.), but that would be more a stretch goal...)

_Originally posted by @rouault in https://github.com/OSGeo/gdal/issues/9836#issuecomment-2091115198_

dbaston:
This is a bit tricky because the docs include some `.rst` files (driver summaries) that are generated at build time by a Python script invoked by `make`. Since Sphinx requires all `rst` files to be under a single root, this would require us to copy/link the static `rst` files into the build directory (as is done for `autotest`). This could be avoided by building the driver summaries with a Sphinx extension instead of the standalone Python script, but it's not clear if this would reduce overall complexity.

rouault:
> Since Sphinx requires all `rst` files to be under a single root, this would require us to copy/link the static `rst` files into the build directory

although not super elegant, why not just forcing the working directory to be PROJECT_SOURCE_DIR/doc when running the script that generates the .rst file ?

dbaston:
Are there other cases where the build process writes to the tree for an out-of-tree build? I didn't want to be the first.

rouault:
> Are there other cases where the build process writes to the tree for an out-of-tree build?

can't think of any. Doc building not being a default build target, I wouldn't mind an exception

dbaston:
> This is a bit tricky because the docs include some .rst files (driver summaries) that are generated at build time by a Python script invoked by make

We could also avoid this issue by generating the driver summaries using Sphinx instead of an external script.

PR #11951, opened 2025-03-12
Doc: Build docs using CMake

dbaston:
This PR:

- Uses CMake to build documentation, removing the Makefile and associated use of Unix tools (sed, etc.). When using CMake, documentation outputs are written to the build tree; the source tree is left alone. CMake also sets the documentation build environment so that GDAL does not need to be installed. Documentation can still be built in-tree by manually invoking `sphinx-build`, as is done in ReadTheDocs.
- Sets Doxygen targets to fail on warnings, and removes the separate "check doxygen warnings" CI job.
- Links Doxygen targets (XML and HTML) to source files, so that they are rebuilt when source files change.
- Links Doxygen, Python bindings, and .rst files to Sphinx targets. Combined with the above, this should somewhat simplify the workflow for adding/editing Python API docstrings. However, it will still be necessary to touch the associated `rst` file - `sphinx-build` has no way to know that this is out of date.

Resolves: #9923

Have not tested the ReadTheDocs build yet; I assume I'll need to cycle on this for a bit.


dbaston:
I don't know if `doc/build_doc_snapshot.sh` is used for anything (and must be updated), or if it can be removed.

rouault:
> I don't know if `doc/build_doc_snapshot.sh` is used for anything

Running it is one of the steps of HOWTO-RELEASE when generating a feature version, to create the gdalXYZdoc.zip artifact (cf https://download.osgeo.org/gdal/3.10.0/)

doc/CMakeLists.txt
```diff
@@ -28,49 +24,145 @@ if (BUILD_DOCS)
...
+            COMMAND ${CMAKE_COMMAND} -E env BUILDDIR=${SPHINX_BUILD_DIR}
+                    ${SPHINX_BUILD} -M latexpdf
+                                    ${CMAKE_CURRENT_SOURCE_DIR}/source
+                                    ${CMAKE_CURRENT_BINARY_DIR}/build
+                                    ${SPHINX_BUILD_OPTS}
+    )
+
+    add_custom_target(latexpdf DEPENDS ${CMAKE_CURRENT_BINARY_DIR}/build/gdal.pdf)
+
+    # Sphinx manpage documentation
```
rouault, line 140: mkgdaldist.sh will need to be updated because it does a (cd doc; make man)

dbaston: I could replace `make man` with a direct `sphinx-build` invocation, or something like 

```
mkdir build-man
cd build-man
cmake ..
cmake --build . --target man
```

The `man` target does not currently depend on GDAL. But building docs for some utilities (like `gdalinfo`) requires executing the utility and collecting the output. So we either need to build GDAL to generate the man pages, or we need to be satisfied with the output of whatever GDAL version is found in the current path. Any thoughts on this?

rouault: > The `man` target does not currently depend on GDAL. But building docs for some utilities (like `gdalinfo`) requires executing the utility and collecting the output. So we either need to build GDAL to generate the man pages, or we need to be satisfied with the output of whatever GDAL version is found in the current path. Any thoughts on this?

we probably need to do a GDAL build of the version we release :-)   A minimal build with -DGDAL_BUILD_OPTIONAL_DRIVERS=OFF -DOGR_BUILD_OPTIONAL_DRIVERS=OFF should do it

dbaston: Added in [1806ba8](https://github.com/OSGeo/gdal/pull/11951/commits/1806ba8e1f17669d585f250efe11282f14a48743), and then changed in [8456bb6](https://github.com/OSGeo/gdal/pull/11951/commits/8456bb6fda84e5b16372bcc874cd285bdd643738) to avoid building GDAL once for the man pages and again for the doc snapshot.

rouault: @dbaston Did you try running the update ./mkgdaldist.sh script ?

dbaston: Yes, via `./mkgdaldist.sh 3110 -branch cmake-build-docs -url https://github.com/dbaston/gdal`

rouault (approved):

doc/source/development/dev_documentation.rst
```diff
@@ -29,35 +43,21 @@ The following targets are available:
...
-* ``doxygen_check_warnings``: same as ``doxygen``, but errors out when Doxygen
-  emits a warning (the ``doxygen`` target is tolerant to Doxygen warnings).
-  This can be useful to reproduce one of the continuous integration checks that
-  verifies that there are no Doxygen warnings.
-  Requires Doxygen >= 1.9.3 to be warning free.
+* ``doxygen_html``: generate Doxygen HTML documentation (this is distinct from the
+  API documentation included in the ``html`` target above).
 
-* ``spelling``: runs spell checking on the documentation, covering as well as
+* ``spelling``: runs spell checking on the documentation, including
```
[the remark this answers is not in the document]
dbaston: Fixed, thanks for the catch

PR #11969, opened 2025-03-17
PDF: Fix default value for DPI open option description

agiudiceandrea:
## What does this PR do?

If I'm not mistaken, the default value for the DPI open option ( = GDAL_PDF_DPI config option) is 150, not 72, according to https://gdal.org/en/stable/drivers/raster/pdf.html.

This PR just fixes the DPI open option description.


rouault (approved):

rouault:
backported to 3.10 per e172a8298d

PR #11971, opened 2025-03-17
gdal2tiles.py: apply srcnodata values in non-reprojected datasets

rouault:
Fixes --srcnodata from being ignored when the dataset does not need to be reprojected.

(Continuation of PR #11965)


charlieforward9:
Thanks for looking at that @rouault

PR #11976, opened 2025-03-18
Python bindings: gdal.UseExceptions(): do not emit ModuleNotFound message when numpy not available

### LINKED ISSUE #11954: Spurious `ModuleNotFound` message when using `ReadRaster` without NumPy
### What is the bug?

When using `ReadRaster` in Python REPL under [the Ubuntu-based Docker images](https://github.com/OSGeo/gdal/tree/master/docker#small-ghcrioosgeogdalubuntu-small-latest) with a virtualenv, GDAL unexpectedly emits the message `ModuleNotFoundError: No module named 'numpy'` to stderr while reading pixel values.

Create an image with this Dockerfile:

```
FROM ghcr.io/osgeo/gdal:ubuntu-small-latest
RUN apt-get update && apt-get install -y python3-pip python3-venv
RUN /usr/bin/python3 -m venv /opt/venv
RUN /opt/venv/bin/python3 -m pip install --upgrade pip
RUN /opt/venv/bin/python3 -m pip install 'GDAL>=3'
COPY example.tif ./
```

Read 100x100 pixels via `ReadRaster` with the virtualenv Python:

```
docker run --rm gdal:latest /opt/venv/bin/python3 -c 'import osgeo.gdal; osgeo.gdal.UseExceptions(); ds = osgeo.gdal.Open("example.tif"); print(len(ds.ReadRaster(0, 0, 100, 100)))'
```

As expected the length of the byte string is printed, but the stderr also includes an unexpected error string:

```
ModuleNotFoundError: No module named 'numpy'
20000
```

### Versions and provenance

Docker 27.4.1 running on MacOS 15.1. GDAL from [documented Docker images](https://github.com/OSGeo/gdal/tree/master/docker#small-ghcrioosgeogdalubuntu-small-latest).

### Additional context

After trying to find the location within `gdal.py` where the message originates, I reached a call to `_gdal.Dataset_ReadRaster` and couldn’t look deeper.

rouault:
Fixes #11954

### DIFF
--- a/swig/include/python/python_exceptions.i
+++ b/swig/include/python/python_exceptions.i
@@ -358,6 +358,11 @@ static void popErrorHandler()
 
 %pythoncode %{
 
+import importlib.util
+
+def _has_gdal_array():
+    return importlib.util.find_spec("osgeo.gdal_array") and importlib.util.find_spec("numpy")
+
 def UseExceptions():
     """ Enable exceptions in all GDAL related modules (osgeo.gdal, osgeo.ogr, osgeo.osr, osgeo.gnm).
         Note: prior to GDAL 3.7, this only affected the calling module"""
@@ -367,21 +372,25 @@ def UseExceptions():
         gdal._UseExceptions()
     except ImportError:
         pass
-    try:
-        from . import gdal_array
-        gdal_array._UseExceptions()
-    except ImportError:
-        pass
+
+    if _has_gdal_array():
+        try:
+            from . import gdal_array
+            gdal_array._UseExceptions()
+        except ImportError:
+            pass
     try:
         from . import ogr
         ogr._UseExceptions()
     except ImportError:
         pass
+
     try:
         from . import osr
         osr._UseExceptions()
     except ImportError:
         pass
+
     try:
         from . import gnm
         gnm._UseExceptions()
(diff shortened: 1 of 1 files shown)


swig/include/python/python_exceptions.i
```diff
@@ -362,16 +362,19 @@ def UseExceptions():
     """ Enable exceptions in all GDAL related modules (osgeo.gdal, osgeo.ogr, osgeo.osr, osgeo.gnm).
         Note: prior to GDAL 3.7, this only affected the calling module"""
 
+    import importlib.util
```
elpaso, line 365: Wouldn't be better to import this at the top level instead of from within the functions?

swig/include/python/python_exceptions.i
```diff
@@ -392,16 +395,19 @@ def DontUseExceptions():
...
         from . import gdal
         gdal._DontUseExceptions()
     except ImportError:
         pass
-    try:
-        from . import gdal_array
-        gdal_array._DontUseExceptions()
-    except ImportError:
-        pass
+    if importlib.util.find_spec("osgeo.gdal_array") and importlib.util.find_spec("numpy"):
```
elpaso, line 405: is it possible to wrap this code into a "private" function  to avoid duplication?

swig/include/python/python_exceptions.i
rouault: fixed

swig/include/python/python_exceptions.i
rouault: fixed

PR #11983, opened 2025-03-18
gcore: add more explicit error message when attempting to write an implicit mask band

### LINKED ISSUE #11973: Improve error message when writing to non-existent mask band
### Feature description

The following code produces confusing error messages:

```
>>> from osgeo import gdal
>>> ds = gdal.GetDriverByName("MEM").Create("", 10, 10)
>>> ds.GetRasterBand(1).GetMaskBand().Fill(255)
ERROR 8: Attempt to write to read only dataset in GDALRasterBand::Fill().
3
>>> import numpy as np
>>> ds.GetRasterBand(1).GetMaskBand().WriteArray(np.full((10, 10), 255))
ERROR 1: Write operation not permitted on dataset opened in read-only mode
```

The error here is that the mask band has not been created, not that the dataset was opened in read-only mode.

### Additional context

_No response_

rouault:
Fixes #11973

### DIFF
--- a/gcore/gdal_priv.h
+++ b/gcore/gdal_priv.h
@@ -1670,6 +1670,10 @@ class CPL_DLL GDALRasterBand : public GDALMajorObject
     virtual int IGetDataCoverageStatus(int nXOff, int nYOff, int nXSize,
                                        int nYSize, int nMaskFlagStop,
                                        double *pdfDataPct);
+
+    virtual bool
+    EmitErrorMessageIfWriteNotSupported(const char *pszCaller) const;
+
     //! @cond Doxygen_Suppress
     CPLErr
     OverviewRasterIO(GDALRWFlag eRWFlag, int nXOff, int nYOff, int nXSize,
@@ -2006,6 +2010,9 @@ class CPL_DLL GDALAllValidMaskBand : public GDALRasterBand
                      GSpacing nLineSpace,
                      GDALRasterIOExtraArg *psExtraArg) override;
 
+    bool
+    EmitErrorMessageIfWriteNotSupported(const char *pszCaller) const override;
+
     CPL_DISALLOW_COPY_ASSIGN(GDALAllValidMaskBand)
 
   public:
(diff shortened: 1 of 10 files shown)


gcore/gdalallvalidmaskband.cpp
```diff
@@ -82,6 +82,20 @@ CPLErr GDALAllValidMaskBand::IRasterIO(GDALRWFlag eRWFlag, int, int, int, int,
...
 
+/************************************************************************/
+/*                   EmitErrorMessageIfWriteNotSupported()              */
+/************************************************************************/
+
+bool GDALAllValidMaskBand::EmitErrorMessageIfWriteNotSupported(
+    const char *pszCaller)
+{
+    ReportError(CE_Failure, CPLE_NoWriteAccess,
+                "%s: attempt to write to a all-valid implicit mask band.",
```
elpaso, line 93: I'm non-native English speaker so I might be wrong but this sounds better to me:
```suggestion
                "%s: attempt to write to an all-valid implicit mask band.",
```

gcore/gdal_priv.h
```diff
@@ -1670,6 +1670,9 @@ class CPL_DLL GDALRasterBand : public GDALMajorObject
     virtual int IGetDataCoverageStatus(int nXOff, int nYOff, int nXSize,
                                        int nYSize, int nMaskFlagStop,
                                        double *pdfDataPct);
+
+    virtual bool EmitErrorMessageIfWriteNotSupported(const char *pszCaller);
```
elpaso, line 1674: maybe const?
```suggestion
    virtual bool EmitErrorMessageIfWriteNotSupported(const char *pszCaller) const;
```

elpaso (approved):
LGTM, just left a few minor comments.

gcore/gdalallvalidmaskband.cpp
rouault: fixed

gcore/gdal_priv.h
rouault: fixed

dbaston (approved):

PR #12000, opened 2025-03-21
Add 'gdal vector edit', and also add a --active-layer to most vector steps to restrict the layer to modify

rouault:


jratike80 (commented):
The parts which I understand look good to me.

PR #12004, opened 2025-03-21
gdal raster calc: use tolerance for extent check

dbaston:


apps/gdalalg_raster_calc.cpp
```diff
@@ -162,8 +162,10 @@ UpdateSourceProperties(SourceProperties &out, const std::string &dsn,
         double ymin =
             source.gt[3] + source.nX * source.gt[4] + source.nY * source.gt[5];
 
-        // TODO use a tolerance here
-        if (xmax != xmaxOut || ymin != yminOut)
+        constexpr double EXTENT_RTOL = 1e-6;
+
+        if (std::abs((xmax - xmaxOut) / xmaxOut) > EXTENT_RTOL ||
```
rouault, line 167: there's a risk of division by zero. I would rather do a comparison like std::fabs(xmax - xmaxOut) > RTOL * std::fabs(gt[1]) where RTOL = 1e-3 maybe? and similarly for y

dbaston: Much better idea, thanks

rouault (approved):

PR #12011, opened 2025-03-22
gdal raster reproject: add --srcnodata, --dstnodata, --addalpha, --warp-option, --transform-option, --error-threshold

rouault:


doc/source/programs/gdal_raster_reproject.rst
```diff
@@ -129,6 +162,106 @@ Standard options
...
+.. option:: --dstnodata <DSTNODATA>
+
+    Set nodata values
+    for output bands (different values can be supplied for each band).  If more
+    than one value is supplied all values should be quoted to keep them together
+    as a single operating system argument.  New files will be initialized to this
+    value and if possible the nodata value will be recorded in the output
+    file. Use a value of ``None`` to ensure that nodata is not defined.
+    If this argument is not used then nodata values will be copied from the source dataset.
+
```
jratike80, line 194: Should there be a note that only some formats support bandwise nodata values, and for example GeoTIFF does not?

doc/source/programs/gdal_raster_reproject.rst
```diff
@@ -129,6 +162,106 @@ Standard options
...
+.. option:: --addalpha
+
+    Create an output alpha band to identify nodata (unset/transparent) pixels.
+
```
jratike80, line 198: Maybe mention value 0 will be written to alpha band to mean nodata, and 255 to mean data when the output datatype is "byte". I do not know how it behaves with other datatypes.

doc/source/programs/gdal_raster_reproject.rst
```diff
@@ -129,6 +162,106 @@ Standard options
...
+    option is specified, in which case an exact transformer, i.e.
+    ``err_threshold=0``, will be used.
+
+Nodata / source validity mask handling
+--------------------------------------
+
+Invalid values in source pixels, either identified through a nodata value
+metadata set on the source band, a mask band, an alpha band or the use of
+:option:`--srcnodata` will not be used in interpolation.
+The details of how it is taken into account depends on the resampling kernel:
```
jratike80, line 222: Maybe mention that in this context alpha is a binary mask and 0 means nodata and all other values mean valid pixels.
I am not sure what is the right thing to do when alpha is used to define partial transparency, should user then use `srcnodata=None`?

doc/source/programs/gdal_raster_reproject.rst
rouault: note added

doc/source/programs/gdal_raster_reproject.rst
rouault: note added

doc/source/programs/gdal_raster_reproject.rst
rouault: > Maybe mention that in this context alpha is a binary mask and 0 means nodata and all other values mean valid pixels.

note added

> I am not sure what is the right thing to do when alpha is used to define partial transparency, should user then use srcnodata=None?

Generally you shouldn't mix both nodata and alpha. I don't remember the exact behavior. I prefer not going into that complication in the doc...

jratike80: > Generally you shouldn't mix both nodata and alpha. 

I don't do that, but I have seen some others, maybe experienced PhotoShop users sometimes do. The behavior varies, if I remember right. Agree, don't try to document now.

doc/source/programs/gdal_raster_reproject.rst
```diff
@@ -129,6 +162,111 @@ Standard options
...
+    warping option (see :cpp:member:`GDALWarpOptions::papszWarpOptions`) to be
+    set to ``YES``, if it is not explicitly set.
+
+    If ``--srcnodata`` is not explicitly set, but the source dataset has nodata values,
+    they will be taken into account, with ``UNIFIED_SRC_NODATA`` at ``PARTIAL``
+    by default.
+
+.. option:: --dstnodata <DSTNODATA>
+
+    Set nodata values
```
elpaso, line 187: odd line break?

doc/source/programs/gdal_raster_reproject.rst
```diff
@@ -129,6 +162,111 @@ Standard options
...
+    value and if possible the nodata value will be recorded in the output
+    file. Use a value of ``None`` to ensure that nodata is not defined.
+    If this argument is not used then nodata values will be copied from the source dataset.
+    Note that a number of output formats, including GeoTIFF, do not support
+    different per-band nodata values, but a single one for all bands.
+
+.. option:: --addalpha
+
+    Create an output alpha band to identify nodata (unset/transparent) pixels.
+    Value 0 is used for fully transparent pixels, and for Byte datasets, 255
```
dbaston, line 200: Not clear from this what happens for non-Byte datasets.

doc/source/programs/gdal_raster_reproject.rst
```diff
@@ -129,6 +162,111 @@ Standard options
...
+
+    Set a transformer option suitable to pass to :cpp:func:`GDALCreateGenImgProjTransformer2`.
+    See :cpp:func:`GDALCreateRPCTransformerV2()` for RPC specific options.
+
+.. option:: --et, --error-threshold <ERROR-THRESHOLD>
+
+    Error threshold for transformation approximation, expressed as a number of
+    source pixels. Defaults to 0.125 pixels unless the ``RPC_DEM`` transformer
+    option is specified, in which case an exact transformer, i.e.
+    ``err_threshold=0``, will be used.
```
dbaston, line 218:
```suggestion
    ``--error-threshold=0``, will be used.
```

doc/source/programs/gdal_raster_reproject.rst
```diff
@@ -129,6 +162,111 @@ Standard options
...
+
+- for nearest resampling, for each target pixel, the coordinate of its center
+  is projected back to source coordinates and the source pixel containing that
+  coordinate is identified. If this source pixel is invalid, the target pixel
+  is considered as nodata.
+
+- for bilinear, cubic, cubicspline and lanczos, for each target pixel, the
+  coordinate of its center is projected back to source coordinates and a
+  corresponding source pixel is identified. If this source pixel is invalid, the
+  target pixel is considered as nodata.
```
dbaston, line 237: In this case, valid pixels within the kernel radius would not be considered? Maybe clarify this.

doc/source/programs/gdal_raster_reproject.rst
```diff
@@ -129,6 +162,111 @@ Standard options
...
+threshold then the remaining source points are approximated using linear
+interpolation between the start and middle point, and between the middle and
+end point.
+If the error exceeds the threshold, the scanline is split into two sections and
+the approximator is recursively applied to each section until the error is less
+than the threshold or all points have been exactly computed.
+
+The error threshold (in source dataset pixels) can be controlled with the
+:option:`--error-threshold` switch. If you want to compare a true pixel-by-pixel reprojection
+use :option:`--error-threshold` which disables this approximator entirely.
```
dbaston, line 268:
```suggestion
use ``--error-threshold=0`` which disables this approximator entirely.
```

doc/source/programs/gdal_raster_reproject.rst
rouault: fixed

doc/source/programs/gdal_raster_reproject.rst
rouault: clarfied:  The maximum value for the alpha
    band, for fully opaque pixels, depends on the data type and the presence of
    the NBITS band metadata item. If it is present, the maximum value is 2^NBITS-1.
    Otherwise, if NBITS is not set and the alpha band is of type UInt16
    (resp. Int16), 65535 (resp. 32767) is used. Otherwise, 255 is used. The
    maximum value can also be overriden with ``--wo DST_ALPHA_MAX=<value>``.

doc/source/programs/gdal_raster_reproject.rst
rouault: fixed

doc/source/programs/gdal_raster_reproject.rst
rouault: clarification added

doc/source/programs/gdal_raster_reproject.rst
rouault: fixed

PR #12022, opened 2025-03-25
Patch 1

elapocha:
## Tasklist

 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments
 - [ ] ADD YOUR TASKS HERE

## Environment

* OS:
* Compiler:

### DIFF
--- a/docker/ubuntu-full/Dockerfile
+++ b/docker/ubuntu-full/Dockerfile
@@ -563,7 +563,8 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
     && apt-get update \
     && apt-get install -y -V libarrow${ARROW_SOVERSION} \
     && apt-get install -y -V libparquet${ARROW_SOVERSION} \
-    && apt-get install -y -V libarrow-dataset${ARROW_SOVERSION}
+    && apt-get install -y -V libarrow-dataset${ARROW_SOVERSION} \
+    && apt-get clean
 
 # Manually install ADBC packages from Ubuntu 22.04 as there are no 24.04 packages at time of writing.
 RUN if test "$(uname -p)" = "x86_64"; then \
@@ -596,6 +597,7 @@ RUN if test "$(uname -p)" = "x86_64"; then \
       apt-get update \
       && apt-get install -y -V libaio1t64 \
       && ln -s libaio.so.1t64 /usr/lib/x86_64-linux-gnu/libaio.so.1 \
+      && apt-get clean
   ) ; fi \
   ; fi
 
--- a/docker/ubuntu-full/Dockerfile
+++ b/docker/ubuntu-full/Dockerfile
@@ -564,7 +564,8 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
     && apt-get install -y -V libarrow${ARROW_SOVERSION} \
     && apt-get install -y -V libparquet${ARROW_SOVERSION} \
     && apt-get install -y -V libarrow-dataset${ARROW_SOVERSION} \
-    && apt-get clean
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists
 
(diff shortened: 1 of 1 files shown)


docker/ubuntu-full/Dockerfile
```diff
@@ -596,6 +598,8 @@ RUN if test "$(uname -p)" = "x86_64"; then \
       apt-get update \
       && apt-get install -y -V libaio1t64 \
       && ln -s libaio.so.1t64 /usr/lib/x86_64-linux-gnu/libaio.so.1 \
+      && apt-get clean \   
+      && rm -rf /var/lib/apt/lists
```
rouault, line 602:
```suggestion
      && rm -rf /var/lib/apt/lists \
```

rouault:
From the missing backslash, it looks this PR hasn't been tested locallly. Furthermore in 7f11471e62872a53b989c350a9cde4e7b1d1f821 the cleaning of the apt cache was removed. @pjonsson what are your thoughts about that PR ?

docker/ubuntu-full/Dockerfile
```diff
@@ -563,7 +563,9 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
     && apt-get update \
     && apt-get install -y -V libarrow${ARROW_SOVERSION} \
     && apt-get install -y -V libparquet${ARROW_SOVERSION} \
-    && apt-get install -y -V libarrow-dataset${ARROW_SOVERSION}
+    && apt-get install -y -V libarrow-dataset${ARROW_SOVERSION} \
+    && apt-get clean \
```
pjonsson, line 567: This clears the packages/lists from Docker cache volume mounted in lines 524-525.

pjonsson:
> [..] in 7f11471 the cleaning of the apt cache was removed.

It seems my commit missed updating the RUN statement that the second part of this PR updates, sorry about that. There should either be the cache mounted for the RUN statement in the same way as lines 524-525, or no cache and files/lists cleared like in this PR.

rouault:
@elpocha Can you make the changes suggested in https://github.com/OSGeo/gdal/pull/12022#issuecomment-2751395652 ?

elapocha:
> @elapocha Can you make the changes suggested in [#12022 (comment)](https://github.com/OSGeo/gdal/pull/12022#issuecomment-2751395652) ?

Sure. In the comment https://github.com/OSGeo/gdal/pull/12022#issuecomment-2751395652 there are two solutions described - first one is a cache mounted for RUN statement and the second one is a solution from this PR. @pjonsson was there any reason for using a cache mounted for RUN statement earlier?

pjonsson:
The cache volume is used to improve build performance, so I suggest going with that solution unless there is an observed problem with it.

(And the apt cache should be left on the cache volume so other builds can avoid downloading things again. Docker will garbage collect the cache volume when the total cache size exceeds the configured cache size limits.)

docker/ubuntu-full/Dockerfile
```diff
@@ -563,7 +563,7 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
     && apt-get update \
     && apt-get install -y -V libarrow${ARROW_SOVERSION} \
     && apt-get install -y -V libparquet${ARROW_SOVERSION} \
-    && apt-get install -y -V libarrow-dataset${ARROW_SOVERSION}
+    && apt-get install -y -V libarrow-dataset${ARROW_SOVERSION}
```
pjonsson, line 566: There's a trailing whitespace left on this line.

pjonsson:
I had a quick look at the cache mounts and they appear similar to the other cache mounts. If this passes the CI I'm pretty sure it works as intended.

rouault:
> If this passes the CI I'm pretty sure it works as intended.

CI doesn't test building Docker images...

pjonsson:
> CI doesn't test building Docker images...

You added that in commit 14c74159b35ca?

PR #12023, opened 2025-03-25
Python bindings: Accept NumPy types in Driver.Create

dbaston:


swig/include/python/typemaps_python.i
```diff
@@ -2468,6 +2468,39 @@ DecomposeSequenceOf4DCoordinates( PyObject *seq, int nCount, double *x, double *
...
 
+%typemap(in) (GDALDataType eType)
+{
+    if (PyInt_Check($input))
+    {
+        $1 = static_cast<GDALDataType>(PyInt_AsLong($input));
+    }
+    else
+    {
+        PyObject* gdal_array = PyImport_ImportModule("osgeo.gdal_array");
```
rouault, line 2479: Py_DECREF() should be called on gdal_array after use

dbaston: Fixed, thanks

rouault (approved):

PR #12039, opened 2025-03-27
MiraMonVector: Fixing metadata about reference system

AbelPau:
## What does this PR do?
When a layer has documented both the horizontal and vertical reference systems, the horizontal reference system was not correctly written in the appropiate MiraMon REL file. This PR fixes it.

## Tasklist


ogr/ogrsf_frmts/miramon/ogrmiramonlayer.cpp
```diff
@@ -183,6 +185,22 @@ OGRMiraMonLayer::OGRMiraMonLayer(GDALDataset *poDS, const char *pszFilename,
...
+                pszTargetKey = "GEOCCS";
+            else if (poSRS->IsLocal())
+                pszTargetKey = "LOCAL_CS";
+
+            if (!poSRS->IsLocal())
+            {
+                pszAuthorityName = poSRS->GetAuthorityName(pszTargetKey);
+                pszAuthorityCode = poSRS->GetAuthorityCode(pszTargetKey);
+            }
+
```
rouault, line 203: This shouldn't be necessary. This is the default behavior of GetAuthorityName() / GetAuthorityCode()

AbelPau: Do you propose using nullptr argument?
pszAuthorityName = poSRS->GetAuthorityName(nullptr);
pszAuthorityCode = poSRS->GetAuthorityCode(nullptr);

In that case this layer returns nullptr in both pszAuthorityName and pszAuthorityCode. This it's not correct, is it?

https://creaf-my.sharepoint.com/:u:/g/personal/a_pau_creaf_uab_cat/EURGVUIt5XhFg6oaA4s3VYUBaZhF3dOF9hDoO7p4-ZqnoA?e=s9fbJ2

rouault: > Do you propose using nullptr argument?

oh, I missed you wanted to extract the horizontal part only. So this is fine

PR #12040, opened 2025-03-27
autotest: avoid deprecation warning about pytest.importorskip("osgeo.gdal_array")

rouault:
Fixes warnings like:
```
gcore/hfa_rfc40.py:27
   /Users/runner/work/gdal/gdal/build/autotest/gcore/hfa_rfc40.py:27: PytestDeprecationWarning:
   Module 'osgeo.gdal_array' was found, but when imported by pytest it raised:
       ImportError("cannot import name '_gdal_array' from 'osgeo' (/Users/runner/work/gdal/gdal/build/swig/python/osgeo/__init__.py)")
   In pytest 9.1 this warning will become an error by default.
   You can fix the underlying problem, or alternatively overwrite this behavior and silence this warning by passing exc_type=ImportError explicitly.
   See https://docs.pytest.org/en/stable/deprecations.html#pytest-importorskip-default-behavior-regarding-importerror
   pytest.importorskip("osgeo.gdal_array")
```

### DIFF
--- a/autotest/alg/sieve.py
+++ b/autotest/alg/sieve.py
@@ -13,6 +13,7 @@
 ###############################################################################
 
 
+import gdaltest
 import pytest
 
 from osgeo import gdal
@@ -176,7 +177,7 @@ def test_sieve_5():
 
 def test_sieve_6():
 
-    pytest.importorskip("osgeo.gdal_array")
+    gdaltest.importorskip_gdal_array()
     numpy = pytest.importorskip("numpy")
 
     # Try 3002. Should run in less than 10 seconds
--- a/autotest/alg/transformgeoloc.py
+++ b/autotest/alg/transformgeoloc.py
@@ -14,6 +14,7 @@
 ###############################################################################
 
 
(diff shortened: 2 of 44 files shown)


dbaston:
Is there a situation with a correctly built/installed GDAL where we would need this? Can we avoid the issue by updating the MacOS CI so that gdal_array is properly imported?

rouault:
> Is there a situation with a correctly built/installed GDAL where we would need this? 

Python bindings built without numpy support, and then numpy installed

dbaston:
> Python bindings built without numpy support, and then numpy installed

Do we gain something from having a CI that is configured that way? There's definitely some friction in requiring a bit of boilerplate in every test that uses numpy.

rouault:
> Do we gain something from having a CI that is configured that way?

it is more than it is a potential testing configuration by users, although unusual.
A potential way forward would be to have a ``np = gdaltest.importnumpyorskip()`` that would do the gdal_array import.

PR #12043, opened 2025-03-28
Add gdal geom vector make-valid, segmentize, simplify, buffer, swap-xy

rouault:


autotest/utilities/test_gdalalg_vector_geom_set_type.py
```diff
@@ -320,3 +320,25 @@ def test_gdalalg_vector_geom_set_type_type_autocomplete():
         f"{gdal_path} completion gdal vector geom set-type --geometry-type GEOMETRYC"
```
jratike80, line 320: What does "GEOMETRYC" stand for? GeometryCollection shortened or what?

doc/source/programs/gdal_vector_geom_buffer.rst
```diff
@@ -0,0 +1,132 @@
...
+    This determines the maximum error in the approximation to the true buffer curve.
+    The default value of 8 gives less than 2% max error in the
+    buffer distance.
+    For a max error of < 1%, use QS = 12.
+    For a max error of < 0.1%, use QS = 18.
+    The error is always less than the buffer distance
+    (in other words, the computed buffer curve is always inside
+    the true curve).
+
+.. option:: --quadrant-segments <QUADRANT-SEGMENTS>
```
jratike80, line 99: This line appears already (line 84). Perhaps copy-paste without edit.

autotest/utilities/test_gdalalg_vector_geom_set_type.py
rouault: yes that's for testing completion

doc/source/programs/gdal_vector_geom_segmentize.rst
```diff
@@ -0,0 +1,65 @@
...
+Synopsis
+--------
+
+.. program-output:: gdal vector geom segmentize --help-doc
+
+Description
+-----------
+
+:program:`gdal vector geom segmentize` ensures segments of LineStrings or rings
+of Polygons are no longer than the specified max length, by inserting
```
jratike80, line 24: Is it rather "not longer"?

doc/source/programs/gdal_vector_geom_buffer.rst
rouault: thanks, fixed

apps/gdalalg_vector_geom_buffer.cpp
```diff
@@ -0,0 +1,168 @@
...
+                    poSrcFeature->StealGeometry(i)))
+            {
+                poGeom.reset(poGeom->BufferEx(m_opts.m_distance,
+                                              m_aosBufferOptions.List()));
+                if (poGeom)
+                {
+                    poGeom->assignSpatialReference(m_srcLayer.GetLayerDefn()
+                                                       ->GetGeomFieldDefn(i)
+                                                       ->GetSpatialRef());
+                    poSrcFeature->SetGeomFieldDirectly(i, poGeom.release());
```
dbaston, line 120:
```suggestion
                    poSrcFeature->SetGeomField(i, std::move(poGeom));
```

doc/source/programs/gdal_vector_geom_simplify.rst
```diff
@@ -0,0 +1,71 @@
...
+Synopsis
+--------
+
+.. program-output:: gdal vector geom simplify --help-doc
+
+Description
+-----------
+
+:program:`gdal vector geom simplify` simplify geometries ensuring that no
+segment is longer than the specified tolerance.
```
jratike80, line 24: New description for simplification :) but how to express it right?
Perhaps mention Douglas-Peucker, I guess that it is what is used, and the desciption could be something like 
"Simplifies a geometry by removing vertices. The tolerance value defines the maximum distance that is allowed between the original vertex and the simplified curve."

I think that GEOS supports also Visvalingam–Whyatt, but that can be reserved for the future.

apps/gdalalg_vector_geom_make_valid.cpp
```diff
@@ -0,0 +1,144 @@
...
+                    if (!bIsGeomCollection && !m_opts.m_keepLowerDim)
+                    {
+                        poGeom.reset(
+                            OGRGeometryFactory::removeLowerDimensionSubGeoms(
+                                poGeom.get()));
+                    }
+                    poGeom->assignSpatialReference(m_srcLayer.GetLayerDefn()
+                                                       ->GetGeomFieldDefn(i)
+                                                       ->GetSpatialRef());
+                    poSrcFeature->SetGeomFieldDirectly(i, poGeom.release());
```
dbaston, line 99:
```suggestion
                    poSrcFeature->SetGeomField(i, std::move(poGeom));
```

doc/source/programs/gdal_vector_geom_simplify.rst
```diff
@@ -0,0 +1,71 @@
...
+
+.. include:: gdal_options/oo.rst
+
+.. include:: gdal_options/if.rst
+
+Examples
+--------
+
+.. example::
+   :title: Make sure that segments of geometries are not smaller than one km (assuming the CRS is in meters)
```
jratike80, line 67: Rather large tolerance for practical use, even OK in theory, and the description is wrong.

apps/gdalalg_vector_geom_make_valid.cpp
```diff
@@ -0,0 +1,144 @@
...
+    std::unique_ptr<OGRFeature> poSrcFeature) const
+{
+    const int nGeomFieldCount = poSrcFeature->GetGeomFieldCount();
+    for (int i = 0; i < nGeomFieldCount; ++i)
+    {
+        if (IsSelectedGeomField(i))
+        {
+            auto poGeom =
+                std::unique_ptr<OGRGeometry>(poSrcFeature->StealGeometry(i));
+            if (poGeom)
```
dbaston, line 82: Pass through features that are (a) non-polygonal or (b) already valid ?

apps/gdalalg_vector_geom_make_valid.cpp
```diff
@@ -0,0 +1,144 @@
...
+/************************************************************************/
+/*                 GDALVectorGeomMakeValidAlgorithm()                   */
+/************************************************************************/
+
+GDALVectorGeomMakeValidAlgorithm::GDALVectorGeomMakeValidAlgorithm(
+    bool standaloneStep)
+    : GDALVectorGeomAbstractAlgorithm(NAME, DESCRIPTION, HELP_URL,
+                                      standaloneStep, m_opts)
+{
+    AddArg("keep-lower-dim", 0,
```
dbaston, line 33: Other options are available (could be added later)

https://github.com/libgeos/geos/blob/1e4d0b01fe307b347fb629793baf8488146d041b/capi/geos_c.h.in#L3221

apps/gdalalg_vector_geom_simplify.cpp
```diff
@@ -0,0 +1,150 @@
...
+                    poSrcFeature->StealGeometry(i)))
+            {
+                poGeom.reset(
+                    poGeom->SimplifyPreserveTopology(m_opts.m_tolerance));
+                if (poGeom)
+                {
+                    poGeom->assignSpatialReference(m_srcLayer.GetLayerDefn()
+                                                       ->GetGeomFieldDefn(i)
+                                                       ->GetSpatialRef());
+                    poSrcFeature->SetGeomFieldDirectly(i, poGeom.release());
```
dbaston, line 105:
```suggestion
                    poSrcFeature->SetGeomField(i, std::move(poGeom));
```

doc/source/programs/gdal_vector_geom_simplify.rst
rouault: what is wrong exactly ?

rouault: the phrasing was indeed not totally accurate. I've reworded in a more neutral way: "Simplify geometries using a tolerance of one km (assuming the CRS is in meters)"

jratike80: Simplify does not do that. The simplified line can be however long, if no part of the simplified line is not moved more than by the tolerance from the original line.  Simplifying a straight line with lots of vertices will keep only the start and end points.

apps/gdalalg_vector_geom_buffer.cpp
rouault: fixed

apps/gdalalg_vector_geom_make_valid.cpp
rouault: fixed

apps/gdalalg_vector_geom_make_valid.cpp
rouault: fixed

apps/gdalalg_vector_geom_simplify.cpp
rouault: fixed

apps/gdalalg_vector_geom_make_valid.cpp
rouault: I've added a commit that adds a --method=linework|structure argument

PR #12045, opened 2025-03-28
Doc: Doxygen fixes

dbaston:
1) `FAIL_ON_WARNINGS` is not a real Doxygen option, not sure how that ended up there. It is ignored by my Doxygen install but causes a failure on Windows with Doxygen from conda. The intended behavior is already captured in `WARN_AS_ERROR=FAIL_ON_WARNINGS_PRINT`.

2) Incredibly, Doxygen substitutes environment variables referenced with parentheses like `$(VARNAME)`. If `$(VARNAME)` expands to a Windows-style path with `\` , it then tried to interpret each component of a path as a Doxygen directive and unsurprisingly fails.

### DIFF
--- a/doc/CMakeLists.txt
+++ b/doc/CMakeLists.txt
@@ -71,7 +71,6 @@ if (BUILD_DOCS)
     if (DOXYGEN_FAIL_ON_WARNINGS) 
             string(JOIN "\n" DOXYFILE_CONTENTS
                    ${DOXYFILE_CONTENTS}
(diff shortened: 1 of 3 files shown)


rouault (approved):

rouault:
@dbaston  github refuses to merge because of "Invalid email for web commit".

dbaston:
I can't make any sense of that message, but it made me notice that my committer name was incomplete. Fixed that and merged

Issue #11993, opened 2025-03-20
netCDF: GeoTransform metadata ignored, resulting in precision loss


dbaston:
### What is the bug?

When writing a netCDF, GDAL writes a `GeoTransform` attribute to the `crs` attribute, e.g.:

```
$ ncinfo -v crs test.nc

<class 'netCDF4._netCDF4.Variable'>
|S1 crs()
    grid_mapping_name: latitude_longitude
    long_name: CRS definition
    longitude_of_prime_meridian: 0.0
    semi_major_axis: 6378137.0
    inverse_flattening: 298.257223563
    crs_wkt: GEOGCS["WGS 84",DATUM["WGS_1984",SPHEROID["WGS 84",6378137,298.257223563,AUTHORITY["EPSG","7030"]],AUTHORITY["EPSG","6326"]],PRIMEM["Greenwich",0,AUTHORITY["EPSG","8901"]],UNIT["degree",0.0174532925199433,AUTHORITY["EPSG","9122"]],AXIS["Latitude",NORTH],AXIS["Longitude",EAST],AUTHORITY["EPSG","4326"]]
    spatial_ref: GEOGCS["WGS 84",DATUM["WGS_1984",SPHEROID["WGS 84",6378137,298.257223563,AUTHORITY["EPSG","7030"]],AUTHORITY["EPSG","6326"]],PRIMEM["Greenwich",0,AUTHORITY["EPSG","8901"]],UNIT["degree",0.0174532925199433,AUTHORITY["EPSG","9122"]],AXIS["Latitude",NORTH],AXIS["Longitude",EAST],AUTHORITY["EPSG","4326"]]
    GeoTransform: 24.999861111111112422 0.00027777777777777777754 0 80.000138888888884026 0 -0.00027777777777777777754 
unlimited dimensions: 
current shape = ()
filling on, default _FillValue of  used
```

This is useful, because recomputing the geotransform from coordinate variables is lossy. However, when reading the netCDF, GDAL prefers the lossy recalculation over the `GeoTransform` attribute.

There is some recognition that this is not ideal:

https://github.com/OSGeo/gdal/blob/08f6aed2490dbdc5eb105d27b51566f4be851411/frmts/netcdf/netcdfdataset.cpp#L4155-L4159

but this was last touched more than 10 years ago, in c42e4e50d9acee9cddcfdfffd36c88b701ca9e29

It's tempting to just update the code to always use the `GeoTransform` value, but are there files for which this would be inappropriate?

[...]

PR #12047, opened 2025-03-29
netCDF: Use GeoTransform attribute to prevent precision loss

dbaston:
Increase precision of netCDF `GeoTransform` attribute and preferentially use it instead of calculated geotransform. Resolves https://github.com/OSGeo/gdal/issues/11993

### DIFF
--- a/frmts/netcdf/netcdfdataset.cpp
+++ b/frmts/netcdf/netcdfdataset.cpp
@@ -5561,7 +5561,7 @@ CPLErr netCDFDataset::AddProjectionVars(bool bDefsOnly,
                 for (int i = 0; i < 6; i++)
                 {
                     osGeoTransform +=
-                        CPLSPrintf("%.16g ", m_adfGeoTransform[i]);
+                        CPLSPrintf("%.17g ", m_adfGeoTransform[i]);
                 }
                 CPLDebug("GDAL_netCDF", "szGeoTransform = %s",
                          osGeoTransform.c_str());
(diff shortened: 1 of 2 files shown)


frmts/netcdf/netcdfdataset.cpp
```diff
@@ -4153,22 +4153,48 @@ void netCDFDataset::SetProjectionFromVar(
...
+                    double dfMaxDiff = 0.0;
+                    for (int i = 0; i < 6; i++)
+                    {
+                        dfMaxDiff =
+                            std::max(dfMaxDiff,
+                                     std::abs(adfTempGeoTransform[i] -
+                                              adfGeoTransformFromAttribute[i]));
+                    }
+
+                    if (dfMaxDiff > 0)
```
rouault, line 4182: we should probably allow for some tolerance value, especially if the x/y arrays are stored as float as that sometimes happen

dbaston: updated

frmts/netcdf/netcdfdataset.cpp
```diff
@@ -4153,22 +4153,54 @@ void netCDFDataset::SetProjectionFromVar(
...
+                if (bGotCfGT)
+                {
+                    constexpr double GT_RELERROR_WARN_THRESHOLD = 1e-6;
+                    double dfMaxAbsoluteError = 0.0;
+                    for (int i = 0; i < 6; i++)
+                    {
+                        double dfAbsoluteError =
+                            std::abs(adfTempGeoTransform[i] -
+                                     adfGeoTransformFromAttribute[i]);
+                        if (dfAbsoluteError > adfGeoTransformFromAttribute[i] *
```
rouault, line 4180: we need to take the absolute value (I suspect pre-commit will have to be run):
```suggestion
                        if (dfAbsoluteError > std::abs(adfGeoTransformFromAttribute[i]) *
```

rouault (approved):

Issue #12051, opened 2025-03-30
[task] Minor cleanup of zarr_driver.py


schwehr:
### What is the bug?

Not a bug really. Just a cleanup process. I want to bring some cleanup back to gdal/autotest from creating https://github.com/schwehr/gdal-autotest2/blob/master/python/gdrivers/zarr_test.py for internal use in google3.

Items:

- [ ] Remove redundant second Mkdir in test_zarr_create_group_errors.
- [ ] Rename `format` → `zarr_format` to avoid the [Built-in Function](https://docs.python.org/3/library/functions.html#format)
- [ ] Remove the `gdal.Of_UPDATE` flag from the post `update()` code of test_zarr_update_with_filters [here](https://github.com/OSGeo/gdal/blob/6b94067434956d4463f49ce6745c53d752e765a2/autotest/gdrivers/zarr_driver.py#L3534)
- [x] ~~`gdal.Open('ZARR:"data/zarr/not_existing.zarr"')` is tested 2x. Once [here](https://github.com/OSGeo/gdal/blob/6b94067434956d4463f49ce6745c53d752e765a2/autotest/gdrivers/zarr_driver.py#L1288C26-L1288C62) in test_zarr_read_classic and also [here](https://github.com/OSGeo/gdal/blob/6b94067434956d4463f49ce6745c53d752e765a2/autotest/gdrivers/zarr_driver.py#L1233).~~ Because only one uses `gdal.OF_MULTIDIM_RASTER`, it's okay

Some minor other notes:

- zarr_group.cpp and zarr_array.cpp call push_back in a loop
- zarr_attribute.cpp uses `anDims.size() == 0` rather than `anDims.empty()`

### Steps to reproduce the issue

N/A

### Versions and provenance

head

### Additional context

None of this is very important. Just minor cleanup

PR #12052, opened 2025-03-30
zarr_driver.py: Remove redundant second Mkdir in test_zarr_create_group_errors.

schwehr:
Please double check this. I think it's correct, but I'm not 100%.

This is one of the items in #12051


rouault:
@schwehr  You need to update to latest master. @dbaston  has done cleanups just today

PR #12057, opened 2025-03-31
gdaldem: hillshade and slope: automatically set scale for geographic …

rouault:
…and projected CRS; add -xscale and -yscale

This avoids in particular the need for defining the -s 111120 switch when the raster uses a geographic CRS

### DIFF
--- a/doc/source/programs/gdaldem.rst
+++ b/doc/source/programs/gdaldem.rst
@@ -24,7 +24,7 @@ Generate a shaded relief map:
 .. code-block::
 
      gdaldem hillshade <input_dem> <output_hillshade>
-                 [-z <zfactor>] [-s <scale>]
+                 [-z <zfactor>] [[-s <scale>] | [-xscale <xscale> -yscale <yscale>]]
                  [-az <azimuth>] [-alt <altitude>]
                  [-alg ZevenbergenThorne] [-combined | -multidirectional | -igor]
                  [-compute_edges] [-b <Band>] [-of <format>] [-co <NAME>=<VALUE>]... [-q]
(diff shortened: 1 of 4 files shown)


jratike80:
Should users be warned somehow, in the documentation or by warning messages, that scale 111120 is only correct at the equator and results will be badly off near the poles?

rouault:
> Should users be warned somehow, in the documentation or by warning messages, that scale 111120 is only correct at the equator and results will be badly off near the poles?

The new behavior  when not setting -scale, -xscale or -yscale is to compute xscale = yscale * cos(mean_latitude), which should yield to reasonable results. But you're right that too clause to the pole, this may still be insufficient. Doc and runtime warning added

doc/source/programs/gdaldem.rst
```diff
@@ -24,7 +24,7 @@ Generate a shaded relief map:
 .. code-block::
 
      gdaldem hillshade <input_dem> <output_hillshade>
-                 [-z <zfactor>] [-s <scale>]
+                 [-z <zfactor>] [[-s <scale>] | [-xscale <xscale> -xscale <yscale>]]
```
jratike80, line 27: `[-xscale <xscale> -xscale <yscale>]]` should  be `[-xscale <xscale> -yscale <yscale>]]`.

rouault: thanks, fixed

doc/source/programs/gdaldem.rst
```diff
@@ -34,7 +34,7 @@ Generate a slope map:
 .. code-block::
 
      gdaldem slope <input_dem> <output_slope_map>
-                 [-p] [-s <scale>]
+                 [-p] [[-s <scale>] | [-xscale <xscale> -xscale <yscale>]]
```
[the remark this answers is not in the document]
rouault: fixed

PR #12069, opened 2025-04-02
VRT Derived Bands: Allow mul and sum to apply constant factor to a single band

dbaston:


rouault (approved):

autotest/gdrivers/vrtderived.py
```diff
@@ -1265,11 +1260,33 @@ def handle(ecls, ecode, emsg):
...
-        os.remove("tmp/derived.vrt")
-    except OSError:
-        pass
+# Test multiplication / summation by a constant factor
+
+
+@pytest.mark.parametrize("fn", ["sum", "mul"])
+def test_vrt_pixelfn_constant_factor(tmp_vsimem, fn):
+
+    np = pytest.importorskip("numpy")
```
rouault, line 1269:
```suggestion
    gdaltest.importorskip_gdal_array()
    np = pytest.importorskip("numpy")
```

rouault:
@dbaston There's a test failure on Windows: https://github.com/dbaston/gdal/actions/runs/14440899536/job/40490611506  . I suspect that must be related to ``Trace = f'open("{tmp_path}/num_sources_skip_true.txt", "wt").write(str(len(in_ar)))'``.  On Windows '{tmp_path}' will expand to a string with backslashes. But when reinjected in ret, ``{Trace}`` will be evaluated in a Python context at runtime, and so those backslashes should be escaped.

PR #12071, opened 2025-04-03
docs: Merge "about" information into a single file

DelazJ:
The idea is to avoid (and limit) "kind of duplicate files"


doc/source/index.rst
```diff
@@ -2,7 +2,8 @@
 GDAL
 ================================================================================
 
-.. include:: ./about_no_title.rst
+.. include:: ./about.rst
+   :start-line: 6
```
DelazJ, line 6: @rouault what is your feeling about using [:start-after:](https://docutils.sourceforge.io/docs/ref/rst/directives.html#include) with a developer comment we would add in the about file, instead of line number?

```suggestion
   :start-after: .. descriptionstartshere
```

doc/source/about.rst
```diff
@@ -4,4 +4,23 @@
 What is GDAL?
 ================================================================================
```
DelazJ, line 6:
```suggestion

.. descriptionstartshere

```

doc/source/index.rst
rouault: sounds great!

rouault (approved):

PR #12073, opened 2025-04-03
Port gdal_grid as "gdal vector grid"

rouault:
The -a invdist|average|... switch of gdal_grid is modeled as a subprogram of "gdal vector grid", so that options specified to each algorithm can really be modeled as options for self-discovery/auto-completion/better doc.


dbaston:
I have always thought that "gdal_grid" is not very descriptive. Pretty much every GDAL program is operating on a grid. What do you think about "gdal vector interpolate-grid" or "gdal raster interpolate" or "gdal raster interpolate-points" ? Sorry to be leaving comments after this is merged but it's hard for me to keep up.

rouault:
What do you think about "gridding" ? It is the idiomatic term for the operation. "interpolate" doesn't capture the fact that the output is a raster (and thus people could think it is a kind of segmentize) and "interpolate-grid" is a bit verbose.
(sorry for the high pace. I'm trying to have a substantial subset of utilities ported in time for 3.11)

dbaston:
> What do you think about "gridding" 

It's the participle instead of the infinitive :slightly_smiling_face: ... I do see the term used by Surfer, but it seems to be less common.

Surfer - https://support.goldensoftware.com/hc/en-us/articles/231348728-A-Basic-Understanding-of-Surfer-Gridding-Methods-Part-1
QGIS - https://docs.qgis.org/3.40/en/docs/gentle_gis_introduction/spatial_analysis_interpolation.html
ArcGIS - https://pro.arcgis.com/en/pro-app/latest/tool-reference/spatial-analyst/an-overview-of-the-interpolation-tools.htm
SAGA - https://saga-gis.sourceforge.io/saga_tool_doc/2.1.3/grid_spline.html
GRASS - https://grasswiki.osgeo.org/wiki/Interpolation#Reinterpolation_of_%22sparse%22_(scattered_points_or_lines)_maps
scipy - https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.griddata.html

jratike80:
Generic Mapping Tools call it gridding https://docs.generic-mapping-tools.org/dev/gmt.html, and it is not uncommon term with point clouds.

PR #12077, opened 2025-04-03
GDALArgument: add Set[Min|Max][Included|Excluded]() to set allowed range

rouault:
and use it in existing algorithms

### DIFF
--- a/apps/data/gdal_algorithm.schema.json
+++ b/apps/data/gdal_algorithm.schema.json
@@ -124,6 +124,22 @@
         "default": {
           "$comment": "Default value (optional)"
         },
+        "min_value": {
+          "type": "number",
+          "$comment": "Minimum value allowed"
+        },
+        "min_value_is_included": {
+          "type": "boolean",
+          "$comment": "Whether min_value is included"
+        },
+        "max_value": {
+          "type": "number",
+          "$comment": "Maximum value allowed"
+        },
+        "max_value_is_included": {
+          "type": "boolean",
+          "$comment": "Whether max_value is included"
+        },
         "required": {
           "type": "boolean",
           "$comment": "Whether this argument is required"
--- a/apps/gdalalg_vector_geom_simplify.cpp
+++ b/apps/gdalalg_vector_geom_simplify.cpp
@@ -30,24 +30,11 @@ GDALVectorGeomSimplifyAlgorithm::GDALVectorGeomSimplifyAlgorithm(
     : GDALVectorGeomAbstractAlgorithm(NAME, DESCRIPTION, HELP_URL,
                                       standaloneStep, m_opts)
 {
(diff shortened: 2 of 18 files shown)


apps/gdalalg_vector_geom_simplify.cpp
```diff
@@ -30,24 +30,11 @@ GDALVectorGeomSimplifyAlgorithm::GDALVectorGeomSimplifyAlgorithm(
...
-                    "Value of 'tolerance' should be strictly positive.");
-                return false;
-            }
-            return true;
-        });
+    AddArg("tolerance", 0, _("Distance tolerance for simplification."),
+           &m_opts.m_tolerance)
+        .SetPositional()
+        .SetRequired()
+        .SetMinValueExcluded(0);
```
jratike80, line 37: ST_Simplify(0) in PostGIS and corresponding things is other software is a common trick for fixing some sort of geometry problems like repeated points. There may be better options nowadays but is there any reason not to allow `.SetMinValueIncluded(0)`? On the other hand, the value must have been strictly positive even before.

rouault: good point. Changed to SetMinValueIncluded(0)

apps/data/gdal_algorithm.schema.json
```diff
@@ -124,6 +124,22 @@
...
+          "$comment": "Whether min_value is included"
+        },
+        "max_value": {
+          "type": "number",
+          "$comment": "Maximum value allowed"
+        },
+        "max_value_is_excluded": {
+          "type": "boolean",
+          "$comment": "Whether max_value is excluded"
+        },
```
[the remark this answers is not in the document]
rouault: oops, no good catch ! too many search&replace of "min" by "max" and "in" by "ex" ...  Fixed

gcore/gdalalgorithm.cpp
```diff
@@ -581,27 +582,133 @@ bool GDALAlgorithmArg::ValidateChoice(const std::string &value) const
...
         const auto &values = Get<std::vector<std::string>>();
         for (const std::string &value : values)
         {
             if (!ValidateChoice(value))
-                return false;
+                ret = false;
         }
     }
 
+    const auto ValidateIntRange = [this, &ret](int val)
```
dbaston, line 602: Consider making this and `ValidateRealRange` member functions? Can `ValidateRealRange` be used for ints, since 64 bit ints are not handled?

rouault: > Consider making this and ValidateRealRange member functions? 

done

> Can `ValidateRealRange` be used for ints, since 64 bit ints are not handled?

for the error messages, for the int case, I explicitly cast to int to get better output

PR #12080, opened 2025-04-03
Fix read out of bounds in GDALInterpolateAtPoint

### LINKED ISSUE #12079: InterpolateAtPoint returns uninitialized value at edge
### What is the bug?

When reading from the right or bottom edge of a raster with `InterpolateAtPoint` and nearest-neighbor interpolation, an invalid read is performed and a garbage value returned.

running `iaptest.py`  under valgrind: 

```
from osgeo import gdal

gdal.UseExceptions()

nx, ny = 3, 3

ds = gdal.GetDriverByName("GTiff").Create("/tmp/test.tif", nx, ny)
ds.SetGeoTransform((0, 1, 0, ny, 0, -1))
ds.GetRasterBand(1).Fill(7)
print(ds.GetRasterBand(1).InterpolateAtPoint(nx, ny, gdal.GRIORA_NearestNeighbour))
```

outputs

```
==714812== Invalid read of size 8
==714812==    at 0x4852E5E: memmove (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==714812==    by 0x667C25B: bool GDALInterpExtractValuesWindow<double>(GDALRasterBand*, std::unique_ptr<lru11::Cache<unsigned long, std::shared_ptr<std::vector<double, std::allocator<double> > >, lru11::NullLock, std::unordered_map<unsigned long, std::_List_iterator<lru11::KeyValuePair<unsigned long, std::shared_ptr<std::vector<double, std::allocator<double> > > > >, std::hash<unsigned long>, std::equal_to<unsigned long>, std::allocator<std::pair<unsigned long const, std::_List_iterator<lru11::KeyValuePair<unsigned long, std::shared_ptr<std::vector<double, std::allocator<double> > > > > > > > >, std::default_delete<lru11::Cache<unsigned long, std::shared_ptr<std::vector<double, std::allocator<double> > >, lru11::NullLock, std::unordered_map<unsigned long, std::_List_iterator<lru11::KeyValuePair<unsigned long, std::shared_ptr<std::vector<double, std::allocator<double> > > > >, std::hash<unsigned long>, std::equal_to<unsigned long>, std::allocator<std::pair<unsigned long const, std::_List_iterator<lru11::KeyValuePair<unsigned long, std::shared_ptr<std::vector<double, std::allocator<double> > > > > > > > > > >&, gdal::VectorX<int, 2ul>, gdal::VectorX<int, 2ul>, double*) (gdal_interpolateatpoint.cpp:127)
==714812==    by 0x667AB1C: bool GDALInterpolateAtPointImpl<double>(GDALRasterBand*, GDALRIOResampleAlg, std::unique_ptr<lru11::Cache<unsigned long, std::shared_ptr<std::vector<double, std::allocator<double> > >, lru11::NullLock, std::unordered_map<unsigned long, std::_List_iterator<lru11::KeyValuePair<unsigned long, std::shared_ptr<std::vector<double, std::allocator<double> > > > >, std::hash<unsigned long>, std::equal_to<unsigned long>, std::allocator<std::pair<unsigned long const, std::_List_iterator<lru11::KeyValuePair<unsigned long, std::shared_ptr<std::vector<double, std::allocator<double> > > > > > > > >, std::default_delete<lru11::Cache<unsigned long, std::shared_ptr<std::vector<double, std::allocator<double> > >, lru11::NullLock, std::unordered_map<unsigned long, std::_List_iterator<lru11::KeyValuePair<unsigned long, std::shared_ptr<std::vector<double, std::allocator<double> > > > >, std::hash<unsigned long>, std::equal_to<unsigned long>, std::allocator<std::pair<unsigned long const, std::_List_iterator<lru11::KeyValuePair<unsigned long, std::shared_ptr<std::vector<double, std::allocator<double> > > > > > > > > > >&, double, double, double&) (gdal_interpolateatpoint.cpp:381)
==714812==    by 0x6678624: GDALInterpolateAtPoint(GDALRasterBand*, GDALRIOResampleAlg, std::unique_ptr<lru11::Cache<unsigned long, std::shared_ptr<std::vector<double, std::allocator<double> > >, lru11::NullLock, std::unordered_map<unsigned long, std::_List_iterator<lru11::KeyValuePair<unsigned long, std::shared_ptr<std::vector<double, std::allocator<double> > > > >, std::hash<unsigned long>, std::equal_to<unsigned long>, std::allocator<std::pair<unsigned long const, std::_List_iterator<lru11::KeyValuePair<unsigned long, std::shared_ptr<std::vector<double, std::allocator<double> > > > > > > > >, std::default_delete<lru11::Cache<unsigned long, std::shared_ptr<std::vector<double, std::allocator<double> > >, lru11::NullLock, std::unordered_map<unsigned long, std::_List_iterator<lru11::KeyValuePair<unsigned long, std::shared_ptr<std::vector<double, std::allocator<double> > > > >, std::hash<unsigned long>, std::equal_to<unsigned long>, std::allocator<std::pair<unsigned long const, std::_List_iterator<lru11::KeyValuePair<unsigned long, std::shared_ptr<std::vector<double, std::allocator<double> > > > > > > > > > >&, double, double, double*, double*) (gdal_interpolateatpoint.cpp:418)
==714812==    by 0x7A4688D: GDALRasterBand::InterpolateAtPoint(double, double, GDALRIOResampleAlg, double*, double*) const (gdalrasterband.cpp:9845)
==714812==    by 0x7A46959: GDALRasterInterpolateAtPoint (gdalrasterband.cpp:9871)
==714812==    by 0x57DEF71: GDALRasterBandShadow_InterpolateAtPoint (gdal_wrap.cpp:8070)
==714812==    by 0x57DEF71: _wrap_Band_InterpolateAtPoint (gdal_wrap.cpp:44047)
==714812==    by 0x5820AF: ??? (in /usr/bin/python3.12)
==714812==    by 0x54B30B: PyObject_Call (in /usr/bin/python3.12)
==714812==    by 0x5DB55A: _PyEval_EvalFrameDefault (in /usr/bin/python3.12)
==714812==    by 0x5D58EA: PyEval_EvalCode (in /usr/bin/python3.12)
==714812==    by 0x608B41: ??? (in /usr/bin/python3.12)
==714812==  Address 0x4eaae10 is 16 bytes after a block of size 80 in arena "client"
==714812== 

```

### Versions and provenance

master

### Additional context

_No response_

jjimenezshaw:
## What does this PR do?
fixes a read out of bounds when reading with `InterpolateAtPoint` using `GRIORA_NearestNeighbour` one pixel just at the right or bottom border out of the image.

fixes #12079

## Tasklist

 - [x] Add test case(s)
 - [ ] ~Add documentation~
 - [ ] ~Updated Python API documentation (swig/include/python/docs/)~
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/alg/gdal_interpolateatpoint.cpp
+++ b/alg/gdal_interpolateatpoint.cpp
@@ -174,9 +174,9 @@ bool GDALInterpolateAtPointImpl(GDALRasterBand *pBand,
         // Allow input coordinates right at the bottom or right edge
         // with GRIORA_NearestNeighbour.
         // "introduce" them in the pixel of the image.
-        if (dfXIn == rasterSize.x())
+        if (dfXIn >= rasterSize.x() && dfXIn <= rasterSize.x() + 1e-5)
             dfXIn -= 0.25;
-        if (dfYIn == rasterSize.y())
+        if (dfYIn >= rasterSize.y() && dfYIn <= rasterSize.y() + 1e-5)
             dfYIn -= 0.25;
     }
     const gdal::Vector2d inLoc{dfXIn, dfYIn};
--- a/alg/gdal_interpolateatpoint.cpp
+++ b/alg/gdal_interpolateatpoint.cpp
@@ -178,6 +178,11 @@ bool GDALInterpolateAtPointImpl(GDALRasterBand *pBand,
     {
         return FALSE;
     }
+    if (eResampleAlg == GRIORA_NearestNeighbour &&
+        (inLoc.x() == rasterSize.x() || inLoc.y() == rasterSize.y()))
+    {
+        return FALSE;
+    }
 
     // Downgrade the interpolation algorithm if the image is too small
     if ((rasterSize.x() < 4 || rasterSize.y() < 4) &&
--- a/alg/gdal_interpolateatpoint.cpp
+++ b/alg/gdal_interpolateatpoint.cpp
@@ -165,9 +165,20 @@ template <typename T>
 bool GDALInterpolateAtPointImpl(GDALRasterBand *pBand,
                                 GDALRIOResampleAlg eResampleAlg,
                                 std::unique_ptr<DoublePointsCache> &cache,
-                                const double dfXIn, const double dfYIn, T &out)
+                                double dfXIn, double dfYIn, T &out)
 {
     const gdal::Vector2i rasterSize{pBand->GetXSize(), pBand->GetYSize()};
+
+    if (eResampleAlg == GRIORA_NearestNeighbour)
+    {
+        // Allow input coordinates right at the bottom or right edge
+        // with GRIORA_NearestNeighbour.
+        // "introduce" them in the pixel of the image.
+        if (dfXIn == rasterSize.x())
+            dfXIn -= 0.25;
+        if (dfYIn == rasterSize.y())
+            dfYIn -= 0.25;
+    }
     const gdal::Vector2d inLoc{dfXIn, dfYIn};
 
     int bGotNoDataValue = FALSE;
@@ -175,11 +186,6 @@ bool GDALInterpolateAtPointImpl(GDALRasterBand *pBand,
 
     if (inLoc.x() < 0 || inLoc.x() > rasterSize.x() || inLoc.y() < 0 ||
         inLoc.y() > rasterSize.y())
-    {
-        return FALSE;
-    }
-    if (eResampleAlg == GRIORA_NearestNeighbour &&
-        (inLoc.x() == rasterSize.x() || inLoc.y() == rasterSize.y()))
     {
         return FALSE;
     }
--- a/alg/gdal_interpolateatpoint.cpp
+++ b/alg/gdal_interpolateatpoint.cpp
@@ -124,13 +124,20 @@ bool GDALInterpExtractValuesWindow(GDALRasterBand *pBand,
             // Compose the cached block to the final buffer
             for (int j = 0; j < nLinesToCopy; j++)
             {
-                memcpy(padfAsDouble + ((nFirstLineInOutput + j) * nWidth +
-                                       nFirstColInOutput) *
-                                          nTypeFactor,
-                       poValue->data() +
-                           ((nFirstLineInCachedBlock + j) * nReqXSize +
-                            nFirstColInCachedBlock) *
-                               nTypeFactor,
+                const size_t dstOffset =
+                    (static_cast<size_t>(nFirstLineInOutput + j) * nWidth +
+                     nFirstColInOutput) *
+                    nTypeFactor;
+                const size_t srcOffset =
+                    (static_cast<size_t>(nFirstLineInCachedBlock + j) *
+                         nReqXSize +
+                     nFirstColInCachedBlock) *
+                    nTypeFactor;
+                if (srcOffset + nColsToCopy * nTypeFactor > poValue->size())
+                {
+                    return false;
+                }
+                memcpy(padfAsDouble + dstOffset, poValue->data() + srcOffset,
                        nColsToCopy * sizeof(T));
             }
         }
--- a/autotest/gcore/interpolateatpoint.py
+++ b/autotest/gcore/interpolateatpoint.py
@@ -44,14 +44,23 @@ def test_interpolateatpoint_outofrange():
     assert res is None
     res = ds.GetRasterBand(1).InterpolateAtPoint(10, 1200, gdal.GRIORA_Bilinear)
     assert res is None
+    res = ds.GetRasterBand(1).InterpolateAtPoint(-1, 0, gdal.GRIORA_NearestNeighbour)
+    assert res is None
+    res = ds.GetRasterBand(1).InterpolateAtPoint(0, -0.5, gdal.GRIORA_NearestNeighbour)
+    assert res is None
+
+
+def test_interpolateatpoint_right_at_border():
+
+    ds = gdal.Open("data/byte.tif")
     res = ds.GetRasterBand(1).InterpolateAtPoint(20, 20, gdal.GRIORA_NearestNeighbour)
-    assert res is None
+    assert res == pytest.approx(107.0, 1e-5)
     res = ds.GetRasterBand(1).InterpolateAtPoint(18, 20, gdal.GRIORA_NearestNeighbour)
-    assert res is None
+    assert res == pytest.approx(99.0, 1e-5)
     res = ds.GetRasterBand(1).InterpolateAtPoint(20, 18, gdal.GRIORA_NearestNeighbour)
-    assert res is None
+    assert res == pytest.approx(123.0, 1e-5)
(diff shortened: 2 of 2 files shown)


jjimenezshaw:
The first commit avoids reading from a buffer out of range. This is safer, but that is not enough for the correct behaviour. The second commit limits the usage of NearestNeighbour in the pixels just out of range (at the border)

alg/gdal_interpolateatpoint.cpp
```diff
@@ -124,13 +124,18 @@ bool GDALInterpExtractValuesWindow(GDALRasterBand *pBand,
...
             {
-                memcpy(padfAsDouble + ((nFirstLineInOutput + j) * nWidth +
-                                       nFirstColInOutput) *
-                                          nTypeFactor,
-                       poValue->data() +
-                           ((nFirstLineInCachedBlock + j) * nReqXSize +
-                            nFirstColInCachedBlock) *
-                               nTypeFactor,
+                const size_t dstOffset =
+                    ((nFirstLineInOutput + j) * nWidth + nFirstColInOutput) *
```
rouault, line 128: In the unlikely case where someone would try to extract more than 2 billion pixels at at time (will need to run pre-commit to reformat):

```suggestion
                    (static_cast<size_t>(nFirstLineInOutput + j) * nWidth + nFirstColInOutput) *
```

alg/gdal_interpolateatpoint.cpp
```diff
@@ -124,13 +124,18 @@ bool GDALInterpExtractValuesWindow(GDALRasterBand *pBand,
...
+                    nTypeFactor;
+                const size_t srcOffset =
+                    ((nFirstLineInCachedBlock + j) * nReqXSize +
```
rouault, line 131:
```suggestion
                    (static_cast<size_t>(nFirstLineInCachedBlock + j) * nReqXSize +
```

alg/gdal_interpolateatpoint.cpp
jjimenezshaw: done

autotest/gcore/interpolateatpoint.py
```diff
@@ -44,6 +44,16 @@ def test_interpolateatpoint_outofrange():
     assert res is None
     res = ds.GetRasterBand(1).InterpolateAtPoint(10, 1200, gdal.GRIORA_Bilinear)
     assert res is None
+    res = ds.GetRasterBand(1).InterpolateAtPoint(20, 20, gdal.GRIORA_NearestNeighbour)
```
dbaston, line 47: Do we want to reeturn the same result as 19.999999999999, 19.9999999999 here? If we are using this function after transforming georeferenced coordinates to pixel/line, and we have an image that spans from (0, 0) to (1, 1), we should return a result for any 0 <= x <= 1, 0 <= y <= 1. 

```
gdal_create -burn 7 -outsize 10 10 -a_ullr 0 1 1 0 -a_srs "EPSG:4326" /tmp/square.tif
gdallocationinfo /tmp/square.tif -wgs84 1 1 # should succeed
```

Alternatively, we could have an `InterpolateAtCoordinate` or something that does the georeferenced -> pixel/line transformation and fudges computed pixel/line for this (literal) edge case.

jjimenezshaw: @dbaston this problem only happens for NearestNeighbour, that has discontinuities. What you are suggesting is to add a different behaviour for the coordinate at the border of the image, right? (in the example (19,19) will give the same pixel as (20,20) just because it is the border of the image)
If that is what we want, it can be done; but I would like an agreement before doing it.

About `InterpolateAtCoordinate`, I think @rouault did something, but I do not remember if it was in the API or in a CLI.

rouault: It's InterpolateAtGeolocation() ( https://gdal.org/en/latest/doxygen/classGDALRasterBand.html#aeacc1fae8ea4fae66d0e304b9e9ca3bb ) . It doesn't do any special tweaking of col, line coordinates

dbaston: Yes, if we think about this as a point in polygon test, I'm suggesting that all pixels contain points on their left and top edges, but the rightmost column of pixels also includes points falling on their right edge, and the bottom row of pixels also includes points on their bottom edge.

jjimenezshaw: @dbaston I added a commit that allows input coordinates at the bottom or right edge. If we like it we can continue. If not, we can just delete it.

dbaston: Thank you, @jjimenezshaw . After pulling your latest commit and changing the following in `gdallocationinfo.cpp`, I get the result I expect.

```
         bool bPixelReport = true;

-        if (iPixel < 0 || iLine < 0 || iPixel >= GDALGetRasterXSize(hSrcDS) ||
-            iLine >= GDALGetRasterYSize(hSrcDS))
+        if (iPixel < 0 || iLine < 0 || dfPixel > static_cast<double>(GDALGetRasterXSize(hSrcDS)) ||
+            dfLine > static_cast<double>(GDALGetRasterYSize(hSrcDS)))
         {
             if (bAsXML)
                 osXML += "<Alert>Location is off this file! No further details "
```

```
gdal_create -burn 7 -outsize 10 10 -a_ullr 0 1 1 0 -a_srs "EPSG:4326" /tmp/square.tif
gdallocationinfo /tmp/square.tif -wgs84 1 1
Report:
  Location: (10P,0L)
  Band 1:
    Value: 7
```

I guess the question is, does everyone else expect it? I think it _is_ reasonable to expect that a raster contains its corner points.

ctoney:
>I guess the question is, does everyone else expect it? I think it _is_ reasonable to expect that a raster contains its corner points.

It's otherwise confusing that a raster would contain one of its corner points but not the other three, or two of its edges and not the other two. Thanks for bringing this up. I'm glad to learn that adjusting the edge cases as in ae7ad34 is a reasonable thing to do.

ctoney:
@dbaston, with your change above https://github.com/OSGeo/gdal/pull/12080#discussion_r2029361537, would you expect a test for `-geoloc` to also work for the lower right corner (or any bottom/right edge)? E.g., the test that uses the upper left corner at https://github.com/OSGeo/gdal/blob/ddf842fdc83981de1032c176addae83726ae6fe2/autotest/utilities/test_gdallocationinfo.py#L91 but add a similar test for lower right? Would it work like this and report 20P,20L?

```py
# Test -geoloc at lower right corner

def test_gdallocationinfo_lr(gdallocationinfo_path):

    ret = gdaltest.runexternal(
        gdallocationinfo_path + " -geoloc ../gcore/data/byte.tif 441920.000 3750120.000"
    )
    ret = ret.replace("\r\n", "\n")
    expected_ret = """Report:
  Location: (20P,20L)
  Band 1:
    Value: 107"""
    assert ret.startswith(expected_ret)
```

rouault:
merged as part of https://github.com/OSGeo/gdal/pull/12087

PR #12088, opened 2025-04-05
Doc: Add caveat to description of gdalwarp -tap

dbaston:


rouault (approved):

doc/source/programs/gdalwarp.rst
```diff
@@ -232,7 +232,8 @@ with control information.
     includes the minimum extent (edges lines/columns that are detected as
     blank, before actual warping, will be removed starting with GDAL 3.8).
     Alignment means that xmin / resx, ymin / resy,
-    xmax / resx and ymax / resy are integer values.
+    xmax / resx and ymax / resy are integer values. It does not necessarily
+    mean that the output grid aligns with the input grid.
```
jratike80, line 236: How could we make it more clear that we mean "xmax divided by resx", and not "xmax and/or resx"?

dbaston: "Alignment means that xmin and ymin are an integer number of pixels from the origin (0, 0)" ?

PR #12091, opened 2025-04-05
Rename 'gdal vector grid' to 'gdal vector gridding'

rouault:


rouault:
@dbaston I know you're not a big fan of it, but is gridding slightly better than grid ?

dbaston:
Between the two I'd probably use "gdal vector grid" since it is the same verb tense as the other commands, and searching for "gdal interpolation" yields hits for "gdal_grid" which could maybe lead one to find "gdal vector grid".

rouault:
ok, dropping this PR then

PR #12092, opened 2025-04-05
GDALAlgorithm: add suggestion for bad algorithm/argument name based on Damerau-Levenshtein distance

rouault:
Demo:
```
$ gdal rastr
ERROR 1: Algorithm 'rastr' is unknown. Do you mean 'raster'?

$ gdal raster nifo
ERROR 1: Algorithm 'nifo' is unknown. Do you mean 'info'?

$ gdal raster info --frmt=json
ERROR 5: info: Option '--frmt' is unknown. Do you mean '--format'?
```

### DIFF
--- a/port/cpl_levenshtein.cpp
+++ b/port/cpl_levenshtein.cpp
@@ -1,0 +1,91 @@
+// SPDX-License-Identifier: MIT
+// Copyright (c) 2019, Guilherme Agostinelli
+// Ported from https://github.com/guilhermeagostinelli/levenshtein/blob/master/levenshtein.cpp
+
+#include "cpl_levenshtein.h"
+
+#include <algorithm>
+#include <cstring>
+#include <limits>
+#include <vector>
+
+/** Computes the Levenshtein distance between 2 words.
+ *
+ * If transpositionAllowed = true, then it is the Damerau-Levenshtein distance
+ *
+ * Return SIZE_T_MAX in case of error.
+ */
+size_t CPLLevenshteinDistance(const char *word1, const char *word2,
+                              bool transpositionAllowed)
+{
+    const size_t size1 = strlen(word1);
+    const size_t size2 = strlen(word2);
+
+    // If one of the words has zero length, the distance is equal to the size of the other word.
+    if (size1 == 0)
+        return size2;
+    if (size2 == 0)
+        return size1;
+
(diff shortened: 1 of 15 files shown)


port/cpl_levenshtein.cpp
```diff
@@ -0,0 +1,91 @@
...
+    const size_t size1 = strlen(word1);
+    const size_t size2 = strlen(word2);
+
+    // If one of the words has zero length, the distance is equal to the size of the other word.
+    if (size1 == 0)
+        return size2;
+    if (size2 == 0)
+        return size1;
+
+    // Would cause enormous amount of memory to be allocaded
```
lnicola, line 30:
```suggestion
    // Would cause enormous amount of memory to be allocated
```

port/cpl_levenshtein.cpp
```diff
@@ -0,0 +1,91 @@
...
+    if (size1 >= 32768 || size2 >= 32768)
+    {
+        return strcmp(word1, word2) == 0 ? 0
```
lnicola, line 33: `size1 + size2` would be more accurate. 

But 2GB is still arguably quite a bit of memory.

port/cpl_levenshtein.cpp
```diff
@@ -0,0 +1,91 @@
...
+    // Verification matrix i.e. 2D array which will store the calculated distance.
+    const size_t dimFastSize = size2 + 1;
+    std::vector<unsigned short> verif;
+    try
+    {
+        verif.resize((size1 + 1) * dimFastSize);
+    }
+    catch (const std::exception &)
+    {
+        return std::numeric_limits<size_t>::max();
```
lnicola, line 46: Ditto

lnicola:
Algorithm 'Dameau-Levenshtein' is unknown. Do you mean 'Damreau-Levenshtein'? :⁠-⁠)

One way to reduce the memory usage is to limit the maximum distance and only compute the elements in that radius from the "diagonal", but it's probably not worth the complexity just for argument parsing.

port/cpl_levenshtein.cpp
rouault: returning SIZE_T_MAX is on purpose (and documented in the function doc) to indicate that the exact value is unknown

Issue #11867, opened 2025-02-19
ZARR: reference stores online (within zip )


mdsumner:
### Feature description

A feature request, or question -  can we access this ZARR reference store (in Parquet) without downloading? 

```
https://archive.podaac.earthdata.nasa.gov/podaac-ops-cumulus-docs/ghrsst/open/docs/MUR-JPL-L4-GLOB-v4.1_combined-ref.parq.zip/
```

I thought this might work (it doesn't appear to need auth for the zip itself): 

```
"ZARR:/vsizip//vsicurl/https://archive.podaac.earthdata.nasa.gov/podaac-ops-cumulus-docs/ghrsst/open/docs/MUR-JPL-L4-GLOB-v4.1_combined-ref.parq.zip/MUR-JPL-L4-GLOB-v4.1_combined-ref.parq")
```

downloading it then works with /vsizip 

```
wget https://archive.podaac.earthdata.nasa.gov/podaac-ops-cumulus-docs/ghrsst/open/docs/MUR-JPL-L4-GLOB-v4.1_combined-ref.parq.zip

gdalmdiminfo "ZARR:/vsizip/MUR-JPL-L4-GLOB-v4.1_combined-ref.parq.zip/MUR-JPL-L4-GLOB-v4.1_combined-ref.parq"
```

but, with caveat, so vars are missing (maybe an install/config problem on my side)

```
ERROR 1: Filter shuffle not handled
```

These experimental parquet/Zarr stores are listed here: https://podaac.jpl.nasa.gov/dataset/MUR-JPL-L4-GLOB-v4.1

martindurant:
Happy to answer questions about the parquet format of kerchunk references. It is *not*  a valid zarr format except when interpreted in the context of references (kerchunk, fsspec.implementations.reference and virtualizarr in python-land). I was not aware of anyone using it outside of my bubble!

https://fsspec.github.io/kerchunk/spec.html#parquet-references

mdsumner:
Ok well one step at a time I guess. Still confused whether this is format for Zarr or just for python but it's not going away. You gave some guidance ala "vsikerchunk" a while back, this is long haul for me and very much appreciate your efforts, aligning to the spec and these recent fixes 👌

martindurant:
> this is format for Zarr or just for python

[...]

PR #12099, opened 2025-04-07
ZARR: add support for reading Kerchunk JSON and Parquet reference stores.

rouault:
Fixes https://github.com/OSGeo/gdal/issues/11867#issuecomment-2669515845

The driver supports reading "virtual" Zarr datasets where the Zarr logical hierarchy
does not directly translate into a hierarchy of files on the filesystem, but
through a file that contains a catalog of references to chunk locations (possibly remote).
Such files are generated by the Python [Kerchunk](https://fsspec.github.io/kerchunk/spec.html) library.

There are 2 types of Kerchunk reference stores:

- JSON reference stores, where the Zarr dataset is entirely described in a single JSON file.
  This comes in two versions: [Version 0](https://fsspec.github.io/kerchunk/spec.html#version-0) and [Version 1](https://fsspec.github.io/kerchunk/spec.html#version-1>).

  Such stores can be opened with:

    ::

        gdalmdiminfo "/vsikerchunk_json_ref//path/to/reference.json"

  or sometimes, when the header of the file is sufficient to be recognized
  as a Kerchunk reference store, with just:

    ::

        gdalmdiminfo "/path/to/reference.json"

  Note that the "templates" and "gen" features of Version 1 are not supported.

- Parquet reference stores, where the Zarr dataset is described in a :file:`.zmetadata` file that
  contains only the JSON definitions of the groups and arrays. The chunks themselves
  are cataloged in Parquet files in subdirectories:
  https://fsspec.github.io/kerchunk/spec.html#parquet-references

  Such stores can be opened with:

    ::

        gdalmdiminfo 'ZARR:"/path/to/directory/where/.zmetadata/is/located"'

  That is exactly like a regular Zarr dataset.

  To be able to read pixel values, the :ref:`vector.parquet` driver must be
  available.

As JSON reference stores can be very large and slow to parse, the :oo:`CACHE_KERCHUNK_JSON`
open option can be set to ``YES`` to ask to generate and use a local Parquet
reference store that is cached in :file:`${HOME}/.gdal/zarr_kerchunk_cache`.
This implies the :ref:`vector.parquet` driver is available.

The driver does not rotate cached stores in the local cache. It is the responsibility
of the user to manage its content and remove obsolete datasets.

It is also possible to convert JSON reference store into a Parquet one using
the :co:`CONVERT_TO_KERCHUNK_PARQUET_REFERENCE` creation option set to ``YES``.

For example:

::

    gdal_translate -of ZARR -co CONVERT_TO_KERCHUNK_PARQUET_REFERENCE=YES store.json store.parq

Also a few related enhancements:
- addition of FixedScaleOffset filter decompressor to read ZARR:"/vsicurl/https://projects.pawsey.org.au/vzarr/NSIDC_SEAICE_PS_N25km.parquet"
- addition of imagecodecs_tiff decompressor for https://sentinel-1-global-coherence-earthbigdata.s3.us-west-2.amazonaws.com/data/wrappers/zarr-all.json (https://guide.cloudnativegeo.org/kerchunk/kerchunk-in-practice.html)


martindurant:
Nice! Thanks for doing this.

The original issue mentioned "within ZIP files" - but this is not required, right?

rouault:
> The original issue mentioned "within ZIP files" - but this is not required, right?

That's orthogonal to this pull request. Instead of specifying a regular .json filename, that could be a /vsicurl/ one , or a /vsizip/ one, or a combination of both , like  "/vsizip//vsicurl/http//example.com/my.zip/path/to/kerchunk_store.json"

rouault:
> I guess `elementsize = 1` is an alias for "shuffle = no" ??

yes, that's logically a no-op. I've added a commit to support elementsize = 1

mdsumner:
nice! fixed, thanks

PR #12105, opened 2025-04-07
gdal_translate: expand output to include pixels partially within -projwin

dbaston:
Currently, `gdal_translate` rounds `-projwin` to the nearest pixels to avoid a grid shift. This may result in an output dataset that does not fully cover `-projwin`. This can be confusing, particularly when `-projwin` is specified in a different CRS from the source dataset. This PR modifies this behavior such that the output extent always includes `-projwin`.

This breaks several driver tests. I can fix those if the change is conceptually OK.


rouault:
> if the change is conceptually OK.

looks good to me!

dbaston:
In the end I needed to update expected results for tests in the VRT, WCS, ESRIC, and OGCAPI drivers. I'm not super-comfortable with the 50k-line diff in the OGCAPI tests, but :man_shrugging:

rouault (approved):

tbonfort:
I believe this change has the side-effect of triggering unwanted `Computed -srcwin 0 -1 4096 4098 falls partially outside source raster extent. Pixels outside the source raster extent will be set to the NoData value (if defined), or zero.` messages a bit further down, when the -srcwin extents exactly match some of the original borders (which is an operation that happens frequently when chopping up a dataset into tiles)
In godal we treat gdal warnings as errors, so this is rather impactful.

dbaston:
Do you have a reproducer handy?

tbonfort:
[empty.zip](https://github.com/user-attachments/files/21316196/empty.zip)

```gdal_translate -ot Byte -r nearest -projwin 73.297485 35.249927 73.385793 35.161618 empty.tif toto.tif```

**Comment by tbonfort** [comment]
I understand/suspect that the issue is due to the rounding of the projwin values. However, given that gdal_translate_lib takes in arguments as strings, small discrepancies between the formatted floating point values and the actual double representation stored in the dataset geotransform are inevitable.

**Comment by dbaston** [comment]
The implementation does take into account such discrepancies. It could arguably be made to use larger tolerances (the current tolerance is 0.1% of a pixel dimension) but the behavior in this case does not appear to be new:

```
$ gdal_translate --version
GDAL 3.8.4, released 2024/02/08

$ gdal_translate -ot Byte -r nearest -projwin 73.297485 35.249927 73.385793 35.161618 /tmp/empty.tif toto.tif
Input file size is 10545, 4202
Warning 1: Computed -srcwin 0 -1 4096 4096 falls partially outside raster extent. Going on however.
```

PR #12111, opened 2025-04-08
Add 'gdal raster viewshed'

rouault:


doc/source/programs/gdal_raster_viewshed.rst
```diff
@@ -0,0 +1,169 @@
...
+   Pixel value to set for visible areas. (Not supported in cumulative mode) Default: 255
+
+.. option:: --mode normal|DEM|ground|cumulative
+
+   Sets what information the output contains.
+
+   - ``normal`` (the default) returns a raster of type Byte containing visible locations.
+
+   - ``DEM`` and ``ground`` return a raster of type Float64 containing the minimum target
+     height for target to be visible from the DEM surface or ground level respectively.
```
jratike80, line 134: What is the difference between DEM surface and ground level?

doc/source/programs/gdal_raster_viewshed.rst
```diff
@@ -0,0 +1,169 @@
...
+
+
+   .. figure:: ../../images/gdal_viewshed.png
+
+
+   Create a viewshed raster with a radius of 500 for a person standing at location (-10147017, 5108065).
+
+   .. code-block:: bash
+
+       gdal raster viewshed --max-distance=500 --pos=-10147017, source.tif destination.tif
```
jratike80, line 169: gdal raster viewshed --max-distance=500 --pos=-10147017,5108065 source.tif destination.tif

rouault: fixed

doc/source/programs/gdal_raster_viewshed.rst
rouault: I've added the following:
"""
  That is to say, if the minimum target height for the target to be visible at a
  point is ``h`` and the value of the input raster at that point is ``E``,
  for ``DEM``, ``E + h`` will be the output value.
  For ``ground``, ``h`` will be output value.
"""

rouault:
CC @abellgithub I'm merging this in time for 3.11 beta1, but feel free to give it a review

abellgithub:
Why the changes for nodata?

rouault:
> Why the changes for nodata?

It is quite surprising that nodata is ignored, so better warn the user when the algorithm encounters it. Actually I got hit by that since my test dataset had -32768 has nodata value and I didn't realize it initially and was wondering why I got big negative values

Issue #11917, opened 2025-03-04
Doc: Condense tutorial presentation


dbaston:
### Feature description

The tutorials (e.g., [Raster API tutorial](https://gdal.org/en/latest/tutorials/raster_api_tut.html) provide example code in C, C++, and Python. The presentation of this code could be condensed using something like [sphinx-tabs](https://sphinx-tabs.readthedocs.io/en/latest/).

geographika:
Other candidates for use of sphinx-tabs are:

- https://gdal.org/en/stable/tutorials/index.html
- https://gdal.org/en/stable/tutorials/osr_api_tut.html#alternate-interfaces

Sometimes the driver examples (and command line programs) don't work under Windows due to different quote-escaping etc. so tabs could also be useful in these cases e.g. https://gdal.org/en/stable/drivers/raster/stacit.html#examples

geographika:
@dbaston @rouault - I'm currently moving the other tutorials to tabs.

Quick question - in the Vector tutorial https://gdal.org/en/stable/tutorials/vector_api_tut.html#reading-from-ogr there are examples for both "With GDAL 2.3 and C++11" and "With GDAL < 2.3 and C++" (as well as C). Should the GDAL < 2.3 examples still be kept?
If so, I can add them to another tab, but maybe they should be removed altogether from the latest docs?

dbaston:
> Should the GDAL < 2.3 examples still be kept?

I think they should be removed.

rouault:
> I think they should be removed.

mentions of anything older than 3.0 can be removed

geographika:
All tutorials with code example in several languages have been updated to use tabs. Closing.

Issue #11918, opened 2025-03-04
Docs: Update Python coding patterns in tutorials


dbaston:
### Feature description

Python tutorial code does not use exceptions and relies on older idioms such as `ds = None` to write a raster.

If exceptions are to become a future default in GDAL, they should be used in our example code.

Sionigdha:
Hi @dbaston , this makes a lot of sense.

I’d be interested in helping modernize a limited subset of the Python tutorial
examples to reflect current best practices (e.g. using exceptions, context
managers, and avoiding older idioms like ds = None).
Before starting, could you clarify:
- which tutorial(s) would be the best place to begin (raster vs vector)?
- whether examples should assume UseExceptions() is enabled?
- and whether consistency with existing docs or showcasing “recommended” usage
is the higher priority?
Happy to start with a small, focused PR and iterate based on feedback.

PR #12114, opened 2025-04-09
Docs: Add sphinx-tabs to the Raster Tutorial

geographika:
## What does this PR do?

Adds the [sphinx-tabs](https://sphinx-tabs.readthedocs.io/en/latest/) extension to the GDAL docs, allowing users to focus on the raster tutorial in their language of choice (C++/C/Python). 

image

#11917

If this PR is accepted, other tutorials can be updated similarly in separate PRs. 

## Tasklist

 - [X] Make sure code is correctly formatted (cf [pre-commit configuration](https://gdal.org/development/dev_practices.html#commit-hooks))
 - [X] Add documentation

### DIFF
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -6,6 +6,7 @@ numpy
 sphinx
 sphinx_bootstrap_theme
 sphinx-markdown-tables
+sphinx-tabs
 sphinxcontrib-bibtex
 sphinxcontrib-jquery
 sphinx-rtd-theme >= 3.0.0
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -83,6 +83,7 @@ extensions = [
     "sphinxcontrib_programoutput_gdal",
     "sphinxcontrib.spelling",
     "myst_nb",
+    "sphinx_tabs.tabs",
 ]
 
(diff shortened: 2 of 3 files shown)


dbaston (approved):
Looks great!

Nice how switching the language for one block switches it for all of them.

dbaston:
> the Python code is in the default mode of exceptions not being enabled. It could be appropriate to have an initial gdal.UseExceptions() and remove a few checks that are not needed then (like gdal.Open() will not return NULL on failure but emit an exception)

This is ticketed (in general) as https://github.com/OSGeo/gdal/issues/11918

The C++ code could see some updates as well :slightly_smiling_face:

geographika:
@rouault @dbaston - thanks for the feedback. I can run through the Python tutorial and make a follow-up PR. Would it make sense to have the Python code as a script that could be run as part of the test suite?
Sphinx could even read the code from the .py file, but that might be over-thinking it.

dbaston:
> Would it make sense to have the Python code as a script that could be run as part of the test suite?

We're not talking about a lot of code here, nor is it something that would be updated very often. So I'm inclined to think it's not worth it.

Running the Python code from the API documentation seems like a higher priority.

PR #12115, opened 2025-04-09
Add 'gdal vector concat' (port of Python ogrmerge script)

rouault:
- Can be used standalone or as the first step of "gdal vector pipeline"
- Strongly inspired from Python ogrmerge, adding a 3rd mode (which is the default) to fully control how source layers are grouped together.


nyalldawson:
> That "cat" is short for "concatenate" is probably obvious for Linux users

To me "cat" would be a dump of the contents, like how the Linux cat command is most often used to view file contents. I agree that "concat" would be less ambiguous.

PR #12120, opened 2025-04-10
Doc: Remove unused requirements.txt

geographika:
## What does this PR do?

Removes `sphinx_bootstrap_theme` and `sphinx-markdown-tables` which are no longer used to build the GDAL docs. 
Also removes `recommonmark`, which again does not appear to be used in the GDAL project, and is archived as of 2022 https://github.com/readthedocs/recommonmark

@dbaston - should https://github.com/OSGeo/gdal/blob/master/doc/environment.yml be kept in sync with requirements.txt? I presume this is for a Conda setup but I can find no mention of it in the docs.

### DIFF
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -4,8 +4,6 @@ breathe
 fsspec
 numpy
 sphinx
-sphinx_bootstrap_theme
-sphinx-markdown-tables
 sphinx-tabs
 sphinxcontrib-bibtex
 sphinxcontrib-jquery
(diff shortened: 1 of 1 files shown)


rouault:
> should https://github.com/OSGeo/gdal/blob/master/doc/environment.yml be kept in sync with requirements.txt?

I suspect this file is unused. You can try removing it and we'll see if CI will complain

dbaston:
> @dbaston - should https://github.com/OSGeo/gdal/blob/master/doc/environment.yml be kept in sync with requirements.txt? I presume this is for a Conda setup but I can find no mention of it in the docs.

IIRC the pre-RTD build environment used conda but also installed some packages from PyPI. Since our Python dependencies don't need to link to system libraries, it seemed simpler to just use `pip`. The `environment.yml` can be removed as far as I'm concerned.

dbaston (approved):

PR #12125, opened 2025-04-10
Doc: Simplify conda gdal-master instructions

dbaston:
Updating these with a command that worked with me. I couldn't run the instructions as written because I don't have write access to the base environment on the system I'm working on. But my understanding is that mamba is no longer necessary anyway.

Also updated the install command to explicitly include `-c conda-forge`; a vanilla `conda install -c gdal-master gdal` happily installs GDAL 3.6 (surprise!)

### DIFF
--- a/doc/source/download.rst
+++ b/doc/source/download.rst
@@ -198,6 +198,8 @@ build can be installed with the following command:
 ::
 
     conda install -c gdal-master -c conda-forge gdal-master::gdal
+        
+As with released versions of GDAL, additional drivers can be installed using `gdal-master::libgdal-{driver_name}`.
 
 
 .. _pixi:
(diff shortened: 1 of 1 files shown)


doc/source/download.rst
```diff
@@ -192,24 +192,13 @@ GDAL master Conda builds
...
 
-    conda update -n base -c conda-forge conda
-    conda install -n base --override-channels -c conda-forge mamba 'python_abi=*=*cp*'
-    conda create --name gdal_master_env
-    conda activate gdal_master_env
-
-Then install GDAL from the ``gdal-master`` channel:
-
-::
+    conda install -c gdal-master -c conda-forge gdal-master::gdal
```
dbaston, line 200:
```suggestion
    conda install -c gdal-master -c conda-forge gdal-master::gdal
        
As with released versions of GDAL, additional drivers can be installed using `gdal-master::libgdal-{driver_name}`.
```

rouault (approved):

PR #12128, opened 2025-04-11
Python bindings: expose gdal_translate -eco and -epo

dbaston:
1) Sets `GDALTranslateOptions::bQuiet` to `true` by default. Otherwise, it is only `true` when called constructed with a progress callback. Possibly this variable could be removed altogether, as there are other methods for squelching errors.

2) Exposes `gdal_translate` options `-epo` and `-eco` as `errorIfWindowPartiallyOutsideSource` and `errorIfWindowCompletelyOutsideSource`. I would love for these to be True by default, but...backwards compatibility. (Maybe they could be the default for `gdal raster clip` though? ) I also don't love the names, maybe there is a better way to do it with a single text argument.


jratike80:
So this case, where the `-projwin` is in EPSG:2391 and is the minimum envelope that covers the whole source raster that is in EPSG:3067, is an `-epo` case, right? No complaint, just want to know.

image

rouault (approved):

PR #12132, opened 2025-04-12
Add 'gdal vfs list/copy/delete'

rouault:
- Add 'gdal vfs list', alias 'gdal vfs ls' (replacement for gdal_ls.py sample script)
- Add 'gdal vfs copy', alias 'gdal vfs cp' (replacement for gdal_cp.py sample script)
- Add 'gdal vfs delete', alias 'gdal vfs rm' (replacement for gdal_rm.py sample script)


jratike80:
Would it be too long as `vfs list`, stands for "virtual file system - list"?

rouault:
I'm open for ``vfs`` instead of ``fs``. The advantage of ``ls`` is that this is a standard Unix name (and the ``-l`` and ``-R`` flags are exactly the ones of ``ls``). That said we have the concept of aliases for command names, so the long name could be ``list``, with ``ls`` as alias

rouault:
> Would it be too long as `vfs list`, stands for "virtual file system - list"?

done. As well as the addition of "copy" and "delete"

apps/gdalalg_vfs_list.cpp
```diff
@@ -0,0 +1,307 @@
...
+    AddArg("long-listing", 'l', _("Use a long listing format"), &m_longListing)
+        .AddAlias("long");
+    AddArg("recursive", 'R', _("List subdirectories recursively"),
+           &m_recursive);
+    AddArg("depth", 0, _("Maximum depth in recursive mode"), &m_depth)
+        .SetMinValueIncluded(1);
+    AddArg("absolute-path", 0, _("Display absolute path"), &m_absolutePath)
+        .AddAlias("abs");
+    AddArg("tree", 0, _("Use a hiearchical presentation for JSON output"),
+           &m_JSONAsTree);
```
jratike80, line 50: ...hierarchical..

doc/source/programs/gdal_vfs_list.rst
```diff
@@ -0,0 +1,76 @@
...
+    the immediate subdirectories, etc.
+
+.. option:: --abs, --absolute-path
+
+    Whether to report file names as absolute paths. By default, they are relative
+    to the input file name.
+
+.. option:: --tree
+
+    Use a hiearchical presentation for JSON output, instead of a flat list.
```
jratike80, line 65: ...hierarchical...

apps/gdalalg_vfs_list.cpp
rouault: fixed

doc/source/programs/gdal_vfs_list.rst
rouault: fixed

elpaso (approved):

hobu:
Sorry I didn't see this until just now. I wonder if it should maybe be named `gdal vsi {ls|delete|copy}` to align it with everything else and to denote that it can and should take VSI links.

rouault:
> that it can and should take VSI links.

it can also accept regular paths. Can people watching this PR vote whether they prefer "vfs" or "vsi" ?

jratike80:
They have 3 letters both but I do not understand what they really mean, so please select your favourite. +0 for both from me.

msmitherdc:
i guess vsi

rouault:
ok let's go for vsi ==> https://github.com/OSGeo/gdal/pull/12142

jratike80:
Finally found what VSI means - naturally "Virtual System Interface" https://lists.osgeo.org/pipermail/gdal-dev/2005-September/006426.html

PR #12134, opened 2025-04-13
Python bindings: make alg['datatype'] accept a GDT_ constant

rouault:


gcore/gdalalgorithm.cpp
```diff
@@ -4911,6 +4913,28 @@ char **GDALAlgorithmArgGetChoices(GDALAlgorithmArgH hArg)
     return CPLStringList(hArg->ptr->GetChoices()).StealList();
 }
 
+/************************************************************************/
+/*                  GDALAlgorithmArgGetMetadataItem()                   */
+/************************************************************************/
+
+/** Return the values of the metadate item of an argument.
```
elpaso, line 4920:
```suggestion
/** Return the values of the metadata item of an argument.
```

elpaso (approved):

PR #12135, opened 2025-04-13
GTiff: support creating Float16, and read it as Float16 (instead of Float32 with NBITS=16)

rouault:


frmts/gtiff/gtiffdataset_write.cpp
```diff
@@ -1895,6 +1909,13 @@ static void DiscardLsb(GByte *pabyBuffer, GPtrDiff_t nBytes, int iBand,
                                         nPlanarConfig, panMaskOffsetLsb,
                                         bHasNoData, nNoDataValue);
     }
+    else if (nBitsPerSample == 16 && nSampleFormat == SAMPLEFORMAT_IEEEFP)
+    {
+        GFloat16 fNoDataValue = static_cast<GFloat16>(dfNoDataValue);
```
elpaso, line 1914:
```suggestion
       const GFloat16 fNoDataValue = static_cast<GFloat16>(dfNoDataValue);
```

elpaso (approved):

PR #12142, opened 2025-04-14
Rename 'gdal vfs' as 'gdal vsi'

rouault:
Thanks to old good friend "sed" for its assistance


jratike80:
Finally found what VSI means - naturally "Virtual System Interface" https://lists.osgeo.org/pipermail/gdal-dev/2005-September/006426.html

jratike80:
Maybe @warmerdam was thinking that VSI could do also something else than virtual file systems in the future? Or maybe he did not think so much but was just coding? Would it be better to keep it secret for another 20 years?

The name `vfs` was not so original, I was just suggesting it instead of `fs` because I was thinking that they **are** virtual.

rouault:
added commit: "Doc: change 'GDAL Virtual file system (VSI)' to 'GDAL Virtual System Interface (VSI)'"

elpaso (approved):

PR #12146, opened 2025-04-15
Enable the generated tile layer by default

ismailsunni:
## What does this PR do?

After generating tiles with gdal2tiles, we got leaflet.html to see the result, but the layer is not enabled by default. I believe it makes sense to enable it by default.

## Tasklist

 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments
 - [ ] ADD YOUR TASKS HERE

## Environment

* OS:
* Compiler:


jratike80 (approved):
Other viewers do set the layer visible already.

rouault (approved):

PR #12157, opened 2025-04-17
Add PowerShell environment setup script

geographika:
Similar to https://github.com/OSGeo/gdal/commit/60bd384411bb783194b24582c1bbd424cb5f01df this adds a PowerShell script for setting up a development environment for the Python bindings:

```ps1
cd C:\docs\gdal\build
cmake --build . --target python_binding

C:\docs\gdal\scripts\setdevenv.ps1
python
>>> import osgeo
```


rouault:
For completness, mention it in https://gdal.org/en/latest/development/dev_environment.html#setting-development-environment-variables ?

PR #12160, opened 2025-04-17
gdal raster clip: error if bbox outside source raster extent

dbaston:


dbaston:
I'd be inclined to take this one step farther and error out if the bbox is partially outside the source raster extent AND a NoData value has not been defined. This is the case where we get filler values of zero written to the output. Thoughts?

rouault:
> I'd be inclined to take this one step farther and error out if the bbox is partially outside the source raster extent AND a NoData value has not been defined.

sounds good.

Shouldn't we have a "--skip-bbox-checks" flag or something like that that would disable -epo / -eco ? I can imagine there are use cases where you are happy with generating empty tiles.

dbaston:
> Shouldn't we have a "--skip-bbox-checks" flag or something like that that would disable -epo / -eco ? I can imagine there are use cases where you are happy with generating empty tiles.

I added a `--allow-bbox-outside-source` and then made `-epo` the default (regardless of input NoData values being defined or not). Had to change an existing test case to make this work.

PR #12162, opened 2025-04-17
Python bindings: Accept CRS definition in osr.SpatialReference constructor

### LINKED ISSUE #12006: Python: Improve osr.SpatialReference constructor
### Feature description

It would be nice to replace the awkward pattern of

```
wgs84 = osr.SpatialReference()
wgs84.ImportFromEPSG(4326)
```

with

```
wgs84 = osr.SpatialReference(epsg=4326)
```

Various keyword arguments could be made available to match the `ImportFromABCD` functions.

### Additional context

_No response_

dbaston:
Resolves #12006

### DIFF
--- a/swig/include/python/docs/osr_spatialreference_docs.i
+++ b/swig/include/python/docs/osr_spatialreference_docs.i
@@ -1,5 +1,26 @@
 %feature("docstring") OSRSpatialReferenceShadow "
 Python proxy of an :cpp:class:`OGRSpatialReference`.
+
+Create a new spatial reference object. An empty object will be created
+unless exactly one of the following parameters is provided.
+
+Parameters
+----------
+epsg : int, optional
+    EPSG CRS code, as understood by :py:meth:`ImportFromEPSG`
+esri : list, optional
+    ESRI ``.prj`` text, as understood by :py:meth:`ImportFromESRI`
+proj4 : str, optional
+    PROJ CRS string, as understood by :py:meth:`ImportFromProj4`
+wkt : str, optional
+    WKT CRS string, as understood by :py:meth:`ImportFromWkt`
+
+Examples
+--------
+>>> osr.SpatialReference(epsg=5646).GetName()
+'NAD83 / Vermont (ftUS)'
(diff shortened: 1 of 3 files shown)


swig/include/python/docs/osr_spatialreference_docs.i
```diff
@@ -1,5 +1,26 @@
...
+Parameters
+----------
+epsg : int, optional
+    EPSG CRS code, as understood by :py:meth:`ImportFromEPSG`
+esri : list, optional
+    ESRI ``.prj`` text, as understood by :py:meth:`ImportFromESRI`
+proj4 : str, optional
+    PROJ CRS string, as understood by :py:meth:`ImportFromProj4`
+wkt : str, optional
+    WKT CRS string, as understood by :py:meth:`ImportFromWkt`
```
rouault, line 16: can we add PROJJSON to it, both serialized or as a dict ?

rouault: > What is the import function for that?

SetFromUserInput(). Cf test_osr_import_projjson in osr_basic.py
I think the simplest would be to change OSRNewSpatialReference() / OGRSpatialReference(const char*) constructor to delegate to SetFromUserInput() which would include the current importFromWkt() behavior

rouault:
(apologies for the build failure. Fixed in latest master HEAD)

swig/include/python/osr_python.i
```diff
@@ -53,9 +53,34 @@ def _WarnIfUserHasNotSpecifiedIfUsingExceptions():
...
+            if type(args[0]) is dict:
+                import json
+                user_input = json.dumps(args[0])
+            else:
+                user_input = args[0]
+
         try:
             with ExceptionMgr(useExceptions=True):
-                this = _osr.new_SpatialReference(*args, **kwargs)
+                this = _osr.new_SpatialReference()
```
rouault, line 83: the try / finally / ExceptionMgr is likely unnecessary now that we no longer pass arguments to the constructor.  And probably all the below stuff until line 96

rouault: > And probably all the below stuff until line 96

actually I'm probably wrong about that. That must be standard SWIG initialization

rouault (approved):

PR #12165, opened 2025-04-17
Fix build issues with isnan() on GFloat16

rouault:
Cf https://lists.osgeo.org/pipermail/gdal-dev/2025-April/060462.html

### DIFF
--- a/gcore/gdalrasterband.cpp
+++ b/gcore/gdalrasterband.cpp
@@ -4078,7 +4078,7 @@ struct GDALNoDataValues
           bGotFloat16NoDataValue(false), hfNoDataValue(GFloat16(0.0f))
     {
         dfNoDataValue = poRasterBand->GetNoDataValue(&bGotNoDataValue);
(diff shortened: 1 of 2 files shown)


gcore/gdalrasterband.cpp
```diff
@@ -6306,9 +6308,10 @@ static inline double GetPixelValue(GDALDataType eDataType, bool bSignedByte,
             break;
         case GDT_Float16:
         {
+            using namespace std;
```
gdt, line 6311: I certainly don't claim this is wrong, but I wonder why there are two instances of `using namespace std;` rather than `std::`, as the latter seems narrower/safer and seems to be the prevailing idiom.

rouault: > I certainly don't claim this is wrong, but I wonder why there are two instances of `using namespace std;` rather than `std::`, as the latter seems narrower/safer and seems to be the prevailing idiom.

there's some subtelty here I don't fully understand myself, but without that using namespace things don't build for me. Maybe @eschnett has a bit more clues

Issue #12005, opened 2025-03-21
Python: Add Dataset.ReadAsMaskedArray


dbaston:
### Feature description

The `Band` version of this was added in https://github.com/OSGeo/gdal/pull/9582 but would be good to have for `Dataset` too.

PR #12172, opened 2025-04-18
Python bindings: Add Dataset.ReadAsMaskedArray

dbaston:
Resolves https://github.com/OSGeo/gdal/issues/12005

### DIFF
--- a/swig/include/python/gdal_python.i
+++ b/swig/include/python/gdal_python.i
@@ -599,15 +599,18 @@ void wrapper_VSIGetMemFileBuffer(const char *utf8_path, GByte **out, vsi_l_offse
   def ReadAsMaskedArray(self, xoff=0, yoff=0, win_xsize=None, win_ysize=None,
                   buf_xsize=None, buf_ysize=None, buf_type=None,
                   resample_alg=gdalconst.GRIORA_NearestNeighbour,
+                  mask_resample_alg=gdalconst.GRIORA_NearestNeighbour,
                   callback=None,
                   callback_data=None):
       """
       Read a window of this raster band into a NumPy masked array.
 
       Values of the mask will be ``True`` where pixels are invalid.
+        
+      Starting in GDAL 3.11, if resampling (``buf_xsize`` != ``xsize``, or ``buf_ysize`` != ``ysize``) the mask
+      band will be resampled using the algorithm specified by ``mask_resample_alg``.
 
-      See :py:meth:`ReadAsArray` for a description of arguments.
-
+      See :py:meth:`ReadAsArray` for a description of additional arguments.
       """
       import numpy
       array = self.ReadAsArray(xoff=xoff, yoff=yoff,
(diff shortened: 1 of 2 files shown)


swig/include/python/gdal_python.i
```diff
@@ -1110,6 +1110,42 @@ CPLErr ReadRaster1( double xoff, double yoff, double xsize, double ysize,
...
+        all_valid = all(self.GetRasterBand(band).GetMaskFlags() == GMF_ALL_VALID for band in band_list)
+
+        if all_valid:
+            return np.ma.masked_array(arr, False)
+       
+        masks = [self.GetRasterBand(band).GetMaskBand().ReadAsArray(
+                xoff=xoff, yoff=yoff,
+                win_xsize=xsize, win_ysize=ysize,
+                buf_xsize=buf_xsize, buf_ysize=buf_ysize,
+                resample_alg=gdalconst.GRIORA_Mode) != 255
```
rouault, line 1143: Perhaps indicate in the doc that when downsampling the mask resampling algorithm is Mode (ie majority) ?

dbaston: Documented this and also changed behavior of `Band.ReadAsMaskedArray` to match.

dbaston:
Hmm, I'm not sure how my tests were passing locally before. Right now I'm puzzled by

```
(Pdb) band.GetMaskBand().ReadAsArray()
array([[  0, 255],
       [  0,   0]], dtype=uint8)
(Pdb) band.GetMaskBand().ReadAsArray(buf_xsize=1, buf_ysize=1, resample_alg=gdal.GRIORA_Mode)
array([[255]], dtype=uint8)
```

rouault:
> Right now I'm puzzled by

I believe this is due to:  https://github.com/OSGeo/gdal/blob/master/gcore/overview.cpp#L4453 / https://github.com/OSGeo/gdal/blob/master/gcore/overview.cpp#L5798

dbaston:
I don't think GRA_Mode downscaling works for mask bands. If I'm following the code right, it performs a downscaled read from the parent band (using GRA_Mode) and then sets all valid pixels to 255. Which is different from performing a full resolution read on the parent, setting valid pixels to 255, and then downscaling.

rouault:
> I don't think GRA_Mode downscaling works for NoData mask bands.

yes the logic I pointed out where the mask band of the mask band is the mask band itself should probably be modified for Mode, and just get the GetMaskBand() of the mask band, which would be a AllValid mask band

dbaston:
@rouault any thoughts on me marking `test_numpy_rw_masked_array_mask_downsampling` as `xfail`, then proceeding along these lines in a separate PR?

>  I think we need to read from the parent band at full resolution, reclassify, and then downscale

dbaston:
> Only for mode ?

It's hard for me to picture what cubic resampling of a binary mask should yield, so it wouldn't bother me to keep the status quo there.

I'm approaching this from the expectation that resampling a mask band should yield the same results as resampling an ordinary band with the same values (0 or 255). Is that the wrong way to look at it?

rouault:
> Is that the wrong way to look at it?

I'm not sure what the "right" way is. I'm just underlying that the overview resampling code (also used by RasterIO() for non-nearest resampling) uses the mask band as its own mask band, ignoring 0 values. Both approaches (current GDAL one or the one you suggest) have their pros and cons depending use cases (if you want to be "optimistic" or "pessimistic" regarding pixel validity). Perhaps we should have a way for the user to specify which it prefers, but that would further complicates things...

dbaston:
I've exposed a `mask_resample_alg` argument in both ReadAsMaskedArray() methods, which I think is a resonable addition: I can imagine a user wanting to read pixels with `GRIORA_Average` and have the mask resampled with `GRIORA_Mode`, `GRIORA_Min`, or `GRIORA_Max` (I realize these last two don't currently exist).

This also sidesteps questions about the behavior of those algorithms for now.

PR #12189, opened 2025-04-21
Warn about INIT_DEST in warp operation instead of failing

sgillies:
#11978 changed a no-op condition to a hard failure. This breaks every recent version of Rasterio, which sets `INIT_DEST=NO_DATA` by default: https://github.com/rasterio/rasterio/blob/0fe62a7106034021e0375746b2581ad595f44f86/rasterio/_warp.pyx#L584-L585.

Instead, how about restoring the no-op, but with a warning?

### DIFF
--- a/alg/gdalwarpoperation.cpp
+++ b/alg/gdalwarpoperation.cpp
@@ -801,16 +801,17 @@ CPLErr GDALWarpOperation::InitializeDestinationBuffer(void *pDstBuffer,
         {
             if (psOptions->padfDstNoDataReal == nullptr)
             {
-                CPLError(CE_Failure, CPLE_AppDefined,
+                CPLError(CE_Warning, CPLE_AppDefined,
                          "BAND_INIT was set to NO_DATA, but a NoData value was "
                          "not defined.");
-                return CE_Failure;
             }
-
-            adfInitRealImag[0] = psOptions->padfDstNoDataReal[iBand];
-            if (psOptions->padfDstNoDataImag != nullptr)
+            else
             {
-                adfInitRealImag[1] = psOptions->padfDstNoDataImag[iBand];
+                adfInitRealImag[0] = psOptions->padfDstNoDataReal[iBand];
+                if (psOptions->padfDstNoDataImag != nullptr)
+                {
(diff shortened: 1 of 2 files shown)


jratike80:
If I understand right, the gdalwarp binary has always required some input for the destination nodata

```
gdalwarp  -dstnodata test.tif nodata.tif --debug on 

ERROR 1: Missing dst_dataset_name
```

Out of curiosity, what Rasterio is doing with such input? Does it copy the source nodata value, or set nodata into 0, or what?

sgillies:
@jratike80 Rasterio uses GDAL's C++ warper, not gdalwarp or its core C function.

rouault:
Sean isn't possible to modify instead Rasterio to not set INIT_DEST to NO_DATA when there is none but instead likely to 0 ?

dbaston:
> https://github.com/OSGeo/gdal/pull/11978 changed a no-op condition to a hard failure

It's not a no-op -- it causes an initialization to zero, though nothing in the code suggests to me that this behavior is intentional. Try running the below snippet both with and without `INIT_DEST`.

```
src = gdal.GetDriverByName("MEM").Create("", 1, 1)
src.SetGeoTransform((0, 1, 0, 1, 0, -1))
src.GetRasterBand(1).Fill(7)

dst = gdal.GetDriverByName("MEM").Create("", 2, 2)
dst.SetGeoTransform((0, 1, 0, 2, 0, -1))
dst.GetRasterBand(1).Fill(1)

gdal.Warp(dst, src, warpOptions={"INIT_DEST":"NO_DATA"})

print(dst.ReadAsArray())
```

If the desired behavior in rasterio is "initialize to the NoData unless there isn't one, in which case initialize to zero", why not code that explicitly?

sgillies:
A new Rasterio release to fix its usage isn't in the cards right now. That's why I'm asking if we can have a warning instead of a warp-stopping failure.

Now it could be that Rasterio user code is less susceptible that Rasterio's tests, and that I could just update Rasterio's tests to not blindly set INIT_DEST like they do. I'll look into that.

dbaston:
I don't have any problem making it a warning for the current release, I'd just prefer not to enshrine the current behavior. Would you object to making it a failure in 3.12?

Issue #12129, opened 2025-04-11
gdal_rasterize all_touched not working as expected


johanvdw:
### What is the bug?

gdal_rasterize with the option all-touched no longer correctly converts all pixels.

The image below shows:
* the shapefile
* dark blue - the resulting raster
* pink: the two pixels which are missing

The problem does not occur in version 3.5.1, but occurs in version 3.6.2 and current master.

![Image](https://github.com/user-attachments/assets/9ea987fe-7202-4d5f-ac27-aa17d1ca0d71)

### Steps to reproduce the issue

Command to reproduce
```bash
gdal_rasterize -q -a_nodata -9999 -at -burn 1 -l mask -of GTiff -te 163600 167680 164000 168000 -tr 20 20 -co COMPRESS=DEFLATE mask.shp mask.tif
```

[gdal-rasterize.zip](https://github.com/user-attachments/files/19702589/gdal-rasterize.zip)

### Versions and provenance

- versions 3.5.1 (correct) (compiled from source in debian linux)
- version 3.6.2 (incorrect) (debian bookworm package)
- version GDAL 3.11.0dev-c4a2e0b926: incorrect (compiled from source in debian linux)

### Additional context

originally reported by @sachagobeyn in https://github.com/watem-sedem/pywatemsedem/issues/55

jratike80:
I can confirm with GDAL 3.10.2 on Windows . Maybe easier to check the result with an 8-bit output
`gdal_rasterize -q -a_nodata 255 -at -burn 1 -of GTiff -te 163600 167680 164000 168000 -tr 20 20 -co COMPRESS=DEFLATE -ot byte mask.shp mask8.tif`

johanvdw:
bisect reveals an issue with:
[9f57e090aa23051ca69ef0bb3c9c992a69553a65] gdal_rasterize: fix ALL_TOUCHED on polygons whose boundaries coordinates are aligned on pixels (fixes #6414) (#6415)

johanvdw:
> And the geometry as WKT is `POLYGON (( 163600 167680, 163600 168000, 164000 168000, 164000 167680, 163600 167680 ))`

No that's a box containing all data, not the actual geometry.

That would be: 
`
[...]

PR #12195, opened 2025-04-22
gdal_rasterize: fix/simplify vertical/horizontal detection

johanvdw:
## What does this PR do?

This fixes https://github.com/OSGeo/gdal/issues/12129

I could not come up with cases where floor would be more appropriate than the absolute difference, but maybe I don't see the full context here.
In addition, one could argue that EPSILON_INTERSECT_ONLY should be used instead of the fixed value of 0.01

I've added a test based on the initial failing file.

### DIFF
--- a/autotest/alg/rasterize.py
+++ b/autotest/alg/rasterize.py
@@ -998,3 +998,44 @@ def test_rasterize_bugfix_gh8918(wkt):
     )
 
     assert target_ds.GetRasterBand(1).Checksum() == 36
+
+###############################################################################
+
+def test_rasterize_bugfix_gh12129():
+    # Setup working spatial reference
+    sr_wkt = 'LOCAL_CS["arbitrary"]'
+    sr = osr.SpatialReference(sr_wkt)
+
+    # Create a memory raster to rasterize into.
+    target_ds = gdal.GetDriverByName("MEM").Create("", 20, 16, 1, gdal.GDT_Byte)
+    target_ds.SetGeoTransform((163600, 20, 0, 168000, 0, -20))
+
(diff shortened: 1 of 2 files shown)


autotest/alg/rasterize.py
```diff
@@ -998,3 +998,51 @@ def test_rasterize_bugfix_gh8918(wkt):
...
+
+    print(target_ds.GetRasterBand(1))
+    checksum = target_ds.GetRasterBand(1).Checksum()
+    expected = 120
+    if checksum != expected:
+        print("checksum: " + str(checksum))
+        gdal.GetDriverByName("GTiff").CreateCopy(
+            "/tmp/rasterize_gh12129.tif", target_ds
+        )
+        pytest.fail("Did not get expected image checksum")
```
dbaston, line 1048:
```suggestion
    assert target_ds.GetRasterBand(1).Checksum() == 120
```

autotest/alg/rasterize.py
```diff
@@ -998,3 +998,43 @@ def test_rasterize_bugfix_gh8918(wkt):
...
+    # Run the algorithm.
+    gdal.RasterizeLayer(
+        target_ds,
+        [1],
+        rast_mem_lyr,
+        burn_values=[1],
+        options=["ALL_TOUCHED=YES"],
+    )
+
+    assert target_ds.GetRasterBand(1).Checksum() == 120
```
rouault, line 1040:
```suggestion
    # 121 on s390x
    assert target_ds.GetRasterBand(1).Checksum() in (120, 121)
```

rouault (approved):

rouault:
Fixes #12129

PR #12198, opened 2025-04-22
GDALWarp: Harden NoData value parsing

dbaston:
- Issue error if `-srcnodata` and `-dstnodata` cannot be parsed
- Issue error if numeric input to `INIT_DEST` cannot be parsed
- Temporarily allow `INIT_DEST=NO_DATA`  without a NoData value, to accommodate rasterio usage

### DIFF
--- a/alg/gdalwarpoperation.cpp
+++ b/alg/gdalwarpoperation.cpp
@@ -802,7 +802,7 @@ CPLErr GDALWarpOperation::InitializeDestinationBuffer(void *pDstBuffer,
             if (psOptions->padfDstNoDataReal == nullptr)
             {
                 CPLError(CE_Failure, CPLE_AppDefined,
-                         "BAND_INIT was set to NO_DATA, but a NoData value was "
+                         "INIT_DEST was set to NO_DATA, but a NoData value was "
                          "not defined.");
                 return CE_Failure;
             }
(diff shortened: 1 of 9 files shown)


alg/gdalwarpoperation.cpp
```diff
@@ -801,8 +801,10 @@ CPLErr GDALWarpOperation::InitializeDestinationBuffer(void *pDstBuffer,
         {
             if (psOptions->padfDstNoDataReal == nullptr)
             {
-                CPLError(CE_Failure, CPLE_AppDefined,
-                         "BAND_INIT was set to NO_DATA, but a NoData value was "
+                // TODO: Change to CE_Failure for GDAL 3.12
+                // See https://github.com/OSGeo/gdal/pull/12189
+                CPLError(CE_Warning, CPLE_AppDefined,
+                         "INIT_DEST was set to NO_DATA, but a NoData value was "
                          "not defined.");
```
rouault, line 808:
```suggestion
                         "INIT_DEST was set to NO_DATA, but a NoData value was "
                         "not defined. This warning will become a failure in a future GDAL release.");
```

rouault (approved):

alg/gdalwarpoperation.cpp
```diff
@@ -801,9 +801,12 @@ CPLErr GDALWarpOperation::InitializeDestinationBuffer(void *pDstBuffer,
...
-                CPLError(CE_Failure, CPLE_AppDefined,
-                         "BAND_INIT was set to NO_DATA, but a NoData value was "
-                         "not defined.");
+                // TODO: Change to CE_Failure for GDAL 3.12
+                // See https://github.com/OSGeo/gdal/pull/12189
+                CPLError(CE_Warning, CPLE_AppDefined,
+                         "INIT_DEST was set to NO_DATA, but a NoData value was "
+                         "not defined. This warning will become a failure in a "
+                         "future GDAL release.");
                 return CE_Failure;
```
QuLogic, line 810: Is this left-in `return` not still a failure instead of _just_ a warning? At the least, it means it won't go through all the bands in the `for` loop it's in, nor those `memset` below.

PR #12203, opened 2025-04-26
gdal vector pipeline: ensure sources are opened as vectors

dbaston:
Resolves https://lists.osgeo.org/pipermail/gdal-dev/2025-April/060482.html

I suspect there is a better way to do this, though.


apps/gdalalg_vector_pipeline.cpp
```diff
@@ -548,6 +548,18 @@ bool GDALVectorPipelineAlgorithm::ParseCommandLineArguments(
         inputArg->GetType() == GAAT_DATASET_LIST &&
         inputArg->Get<std::vector<GDALArgDatasetValue>>().size() == 1)
     {
+        if (inputArg->GetType() == GAAT_DATASET_LIST)
```
pjonsson, line 551: I don't know anything about the code/problem, but doesn't line 548 guarantee that this condition is always true?

PR #12206, opened 2025-04-27
Fix warnings with gcc 15.1 and arrow 20.0

rouault:


rouault:
CC @BishopGIS for the libopencad change

PR #12208, opened 2025-04-27
[cli] Port gdal_fillnodata to CLI

elpaso:
```bash
Usage: gdal raster fillnodata [OPTIONS] <INPUT> <OUTPUT>

Fill nodata raster regions by interpolation from edges.

Positional arguments:

Common Options:

Options:
  --overwrite                                          Whether overwriting existing output is allowed
  -b, --band <BAND>                                    Input band (1-based index) (default: 1)
  -d, --max-distance <MAX_DISTANCE>                    The maximum distance (in pixels) that the algorithm will search out for values to interpolate. (default: 100)
  -s, --smoothing-iterations <SMOOTHING_ITERATIONS>    The number of 3x3 average filter smoothing iterations to run after the interpolation to dampen artifacts. The default is zero smoothing iterations. (default: 0)
  --mask <MASK>                                        Use the first band of the specified file as a validity mask (zero is invalid, non-zero is valid).
  --strategy <STRATEGY>                                By default, pixels are interpolated using an inverse distance weighting (inv_dist). It is also possible to choose a nearest neighbour (nearest) strategy.. STRATEGY=inv_dist|nearest (default: inv_dist)

Advanced Options:

For more details, consult https://gdal.org/programs/gdal_raster_fillnodata.html
```

### DIFF
--- a/apps/CMakeLists.txt
+++ b/apps/CMakeLists.txt
@@ -24,8 +24,10 @@ add_library(
   gdalalg_raster_convert.cpp
   gdalalg_raster_create.cpp
   gdalalg_raster_edit.cpp
+
   gdalalg_raster_contour.cpp
   gdalalg_raster_footprint.cpp
+  gdalalg_raster_fillnodata.cpp
   gdalalg_raster_hillshade.cpp
   gdalalg_raster_index.cpp
   gdalalg_raster_mosaic.cpp
--- a/apps/CMakeLists.txt
+++ b/apps/CMakeLists.txt
@@ -24,7 +24,6 @@ add_library(
   gdalalg_raster_convert.cpp
   gdalalg_raster_create.cpp
   gdalalg_raster_edit.cpp
-
   gdalalg_raster_contour.cpp
   gdalalg_raster_footprint.cpp
   gdalalg_raster_fillnodata.cpp
--- a/apps/gdalalg_raster_fillnodata.cpp
+++ b/apps/gdalalg_raster_fillnodata.cpp
@@ -71,7 +71,7 @@ GDALRasterFillNodataAlgorithm::GDALRasterFillNodataAlgorithm() noexcept
     SetAutoCompleteFunctionForFilename(mask, GDAL_OF_RASTER);
 
     mask.AddValidationAction(
-        [this, &mask]()
+        [&mask]()
         {
             // Check if the mask dataset is a valid raster dataset descriptor
             // Try to open the dataset as a raster
--- a/apps/gdalalg_raster_fillnodata.cpp
+++ b/apps/gdalalg_raster_fillnodata.cpp
@@ -149,13 +149,6 @@ bool GDALRasterFillNodataAlgorithm::RunImpl(GDALProgressFunc pfnProgress,
     }
 
     GDALRasterBand *dstBand{poRetDS->GetRasterBand(1)};
-    if (!dstBand)
-    {
-        ReportError(CE_Failure, CPLE_AppDefined,
-                    "Cannot get destination band.");
-        return false;
-    }
-
     // Prepare options to pass to GDALFillNodata
     CPLStringList aosFillOptions;
     aosFillOptions.AddNameValue("INTERPOLATION", m_strategy.c_str());
--- a/apps/gdalalg_raster_fillnodata.cpp
+++ b/apps/gdalalg_raster_fillnodata.cpp
@@ -83,11 +83,11 @@ GDALRasterFillNodataAlgorithm::GDALRasterFillNodataAlgorithm() noexcept
 
     AddArg("strategy", 0,
            _("By default, pixels are interpolated using an inverse distance "
-             "weighting (inv_dist). It is also possible to choose a nearest "
+             "weighting (invdist). It is also possible to choose a nearest "
              "neighbour (nearest) strategy."),
            &m_strategy)
         .SetDefault(m_strategy)
-        .SetChoices("inv_dist", "nearest");
+        .SetChoices("invdist", "nearest");
 }
 
 /************************************************************************/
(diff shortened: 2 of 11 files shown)


apps/CMakeLists.txt
```diff
@@ -24,8 +24,10 @@ add_library(
   gdalalg_raster_convert.cpp
   gdalalg_raster_create.cpp
   gdalalg_raster_edit.cpp
+
```
rouault, line 27:
```suggestion
```

apps/gdalalg_raster_fillnodata.cpp
```diff
@@ -0,0 +1,180 @@
...
+            return static_cast<bool>(poDS);
+        });
+
+    AddArg("strategy", 0,
+           _("By default, pixels are interpolated using an inverse distance "
+             "weighting (inv_dist). It is also possible to choose a nearest "
+             "neighbour (nearest) strategy."),
+           &m_strategy)
+        .SetDefault(m_strategy)
+        .SetChoices("inv_dist", "nearest");
```
rouault, line 90: maybe normalize to "invdist" to be consistent with "gdal vector grid invdist" ? (and adapt in Run() to transform to inv_dist)

apps/gdalalg_raster_fillnodata.cpp
```diff
@@ -0,0 +1,180 @@
...
+        return false;
+    }
+
+    CPLStringList aosOptions;
+    aosOptions.AddString("-of");
+    aosOptions.AddString(m_format.c_str());
+    aosOptions.AddString("-b");
+    aosOptions.AddString(CPLSPrintf("%d", m_band));
+
+    for (const auto &co : m_creationOptions)
```
rouault, line 120: This is a suboptimal strategy in particular if outputing to a compressed dataset, because it will lead to the  band to being written twice leading to a larger file than needed.
I'd suggest to outputing to a temporary LZW  tiled compressed TIFF file (name generated with CPLGenerateTempFilename()), apply  GDALFillNodata() and finally translate to final file

doc/source/programs/gdal_raster_fillnodata.rst
```diff
@@ -0,0 +1,77 @@
...
+
+.. example::
+   :title: Fill nodata areas in a raster
+
+   The command specifies to use the second band of the input raster, 50 px max distance,
+   3 smoothing iterations and the `nearest` strategy for interpolation.
+   The output will be saved in `output.tif`.
+
+   .. code-block:: bash
+        gdal raster fillnodata -b 2 -max-distance 50 -smoothing-iterations 3 \
```
rouault, line 75:
```suggestion
        gdal raster fillnodata -b 2 --max-distance 50 --smoothing-iterations 3 \
```

apps/gdalalg_raster_fillnodata.cpp
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    GDALRasterBand *dstBand{poRetDS->GetRasterBand(1)};
+    if (!dstBand)
+    {
+        ReportError(CE_Failure, CPLE_AppDefined,
+                    "Cannot get destination band.");
+        return false;
+    }
+
```
rouault, line 158: shouldn't happen given all the above
```suggestion
```

doc/source/programs/gdal_raster_fillnodata.rst
```diff
@@ -0,0 +1,77 @@
...
+
+    Specifies the number of smoothing iterations to apply to the filled raster.
+    This can help to reduce artifacts in the filled areas.
+    Default is 0 iterations.
+
+.. option:: --strategy <STRATEGY>
+
+    Select the interpolation <STRATEGY> to use.
+    By default, pixels are interpolated using an inverse distance
+    weighting (`inv_dist`). It is also possible to choose a nearest
```
rouault, line 56: if following above suggestion

```suggestion
    weighting (`invdist`). It is also possible to choose a nearest
```

apps/gdalalg_raster_fillnodata.cpp
elpaso: I'm not sure I understand what you mean, the steps here are:
1. copy the original raster band (only the required band) to a new dataset (because we don't want to alter the original)
2. apply fillnodata on the newly copied band

Are you suggesting instead to:
1. copy (and potentially change the format of) the original raster band to LZW tiled compressed TIFF (regardless of the user-required destination format which might be another one)
2. apply fillnodata on the  newly copied band
3. translate to the user required format

?

It seems to me potentially more expensive, but I admit I didn't look at the internals of fillnodata.

jratike80:
Multiband GeoTIFFs are common and then it is not enough to fill one band. One band is used for analysing where nodata exist (fine in TIFF because the same nodata value applies all bands), but all the bands must be interpolated and filled. At least if I understand right the purpose of the tool.

elpaso:
> Multiband GeoTIFFs are common and then it is not enough to fill one band. One band is used for analysing where nodata exist (fine in TIFF because the same nodata value applies all bands), but all the bands must be interpolated and filled. At least if I understand right the purpose of the tool.

AFAIK the algorithm takes a single band as input.

elpaso:
> > Multiband GeoTIFFs are common and then it is not enough to fill one band. One band is used for analysing where nodata exist (fine in TIFF because the same nodata value applies all bands), but all the bands must be interpolated and filled. At least if I understand right the purpose of the tool.
> 
> AFAIK the algorithm takes a single band as input.

Which doesn't mean that it would not be a good idea to enhance it and implement multi-band support, though.

apps/gdalalg_raster_fillnodata.cpp
rouault: > Are you suggesting instead to:

yes, and it is indeed more expansive, but otherwise offering creation option could be a lie to the user. We could potentially only do that only if compression is selected in creation options (and restrict that detection to GTiff output format), but that would be a complication.

Anyway, we can disregard that point for this PR, but then please create an enhancement ticket about the issue, and make sure to add a warning in the documentation.

autotest/utilities/test_gdalalg_raster_fillnodata.py
```diff
@@ -0,0 +1,158 @@
...
+###############################################################################
+
+import gdaltest
+import pytest
+
+from osgeo import gdal
+
+pytest.importorskip("numpy")
+gdaltest.importorskip_gdal_array()
+
```
rouault, line 21:
```suggestion


```

jratike80:
> AFAIK the algorithm takes a single band as input.

OK, I believed that the algorithm could patch unintentional nodata holes (totally black pixels which tend to appear within water bodies etc.) also in multiband images.

PR #12217, opened 2025-04-28
CMake: Export GDAL library targets

dbaston:
## What does this PR do?

When building a separate project, this allows `find_package(GDAL CONFIG)` to import from a GDAL build tree using `cmake -DCMAKE_FIND_ROOT_PATH=/home/dan/dev/gdal/build-release ..`

See https://cmake.org/cmake/help/latest/guide/importing-exporting/index.html#exporting-targets-from-the-build-tree

### DIFF
--- a/gdal.cmake
+++ b/gdal.cmake
@@ -553,6 +553,11 @@ install(
   PUBLIC_HEADER DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}
   FRAMEWORK DESTINATION "${FRAMEWORK_DESTINATION}")
 
+# Generate targets file for importing directly from GDAL build tree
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -281,3 +281,7 @@ if (BUILD_TESTING AND EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/perftests")
   add_subdirectory(perftests)
 endif ()
 # vim: ts=4 sw=4 sts=4 et
+
+export(TARGETS ${GDAL_LIB_TARGET_NAME}
```
rouault, line 285: This should rather be moved to gdal.cmake close to lines that do similar things (line 546-563)

rouault (approved):

PR #12219, opened 2025-04-28
[cli] gdal raster sieve

elpaso:
```
Usage: gdal raster sieve [OPTIONS] <INPUT> <OUTPUT>

Remove small polygons from a raster dataset.

Positional arguments:

Common Options:

Options:
  --overwrite                                          Whether overwriting existing output is allowed
  --mask <MASK>                                        Use the first band of the specified file as a validity mask (all pixels with a value other than zero will be considered suitable for inclusion in polygons)
  -b, --band <BAND>                                    Input band (1-based index)
  -s, --size-threshold <SIZE-THRESHOLD>                Minimum size of polygons to keep (default: 2)
  -c, --connectedness                                  Consider diagonal pixels as connected

Advanced Options:
  --oo, --open-option <KEY=VALUE>                      Open options [may be repeated]

For more details, consult https://gdal.org/programs/gdal_raster_sieve.html
```

### DIFF
--- a/apps/gdalalg_raster_sieve.cpp
+++ b/apps/gdalalg_raster_sieve.cpp
@@ -49,10 +49,12 @@ GDALRasterSieveAlgorithm::GDALRasterSieveAlgorithm()
     AddCreationOptionsArg(&m_creationOptions);
     AddOverwriteArg(&m_overwrite);
 
-    auto &mask{AddArg("mask", 0,
-                      _("Use the first band of the specified file as a "
-                        "validity mask (zero is invalid, non-zero is valid)"),
-                      &m_maskDataset)};
+    auto &mask{
+        AddArg("mask", 0,
+               _("Use the first band of the specified file as a "
+                 "validity mask (all pixels with a value other than zero "
+                 "will be considered suitable for inclusion in polygons)"),
+               &m_maskDataset)};
 
     SetAutoCompleteFunctionForFilename(mask, GDAL_OF_RASTER);
 
@@ -119,53 +121,40 @@ bool GDALRasterSieveAlgorithm::RunImpl(GDALProgressFunc pfnProgress,
                                     : m_outputDataset.GetName()};
     std::unique_ptr<GDALDataset> workingDSHandler;
 
+    CPLStringList options;
+    options.AddString("-b");
+    options.AddString(CPLSPrintf("%d", m_band));
+
     if (useTemporaryFile)
     {
         // Translate the single required band to uncompressed TILED GTiff
-        CPLStringList options;
-        options.AddString("-b");
-        options.AddString(CPLSPrintf("%d", m_band));
         options.AddString("-of");
         options.AddString("GTiff");
         options.AddString("-co");
         options.AddString("TILED=YES");
-        GDALTranslateOptions *translateOptions =
-            GDALTranslateOptionsNew(options.List(), nullptr);
-
-        GDALDatasetH hSrcDS =
-            GDALDataset::ToHandle(m_inputDataset.GetDatasetRef());
-        workingDSHandler.reset(GDALDataset::FromHandle(
-            GDALTranslate(m_outputDataset.GetName().c_str(), hSrcDS,
-                          translateOptions, nullptr)));
-        GDALTranslateOptionsFree(translateOptions);
     }
     else
     {
         // Translate the single required band to the destination format
-        CPLStringList options;
-        options.AddString("-b");
-        options.AddString(CPLSPrintf("%d", m_band));
-        options.AddString("-of");
-        options.AddString(m_format.c_str());
+        if (!m_format.empty())
+        {
+            options.AddString("-of");
+            options.AddString(m_format.c_str());
+        }
         for (const auto &co : m_creationOptions)
         {
-            if (co.find('=') != std::string::npos)
-            {
-                options.AddString(co.c_str());
-            }
-            else
-            {
-                options.AddString(CPLSPrintf("-co %s", co.c_str()));
-            }
+            options.AddString("-co");
+            options.AddString(co.c_str());
         }
-        GDALTranslateOptions *translateOptions =
-            GDALTranslateOptionsNew(options.List(), nullptr);
-        GDALDatasetH hSrcDS =
-            GDALDataset::ToHandle(m_inputDataset.GetDatasetRef());
-        workingDSHandler.reset(GDALDataset::FromHandle(GDALTranslate(
-            workingFileName.c_str(), hSrcDS, translateOptions, nullptr)));
-        GDALTranslateOptionsFree(translateOptions);
     }
+
+    GDALTranslateOptions *translateOptions =
+        GDALTranslateOptionsNew(options.List(), nullptr);
+
+    GDALDatasetH hSrcDS = GDALDataset::ToHandle(m_inputDataset.GetDatasetRef());
+    workingDSHandler.reset(GDALDataset::FromHandle(GDALTranslate(
+        workingFileName.c_str(), hSrcDS, translateOptions, nullptr)));
+    GDALTranslateOptionsFree(translateOptions);
 
     if (!workingDSHandler)
     {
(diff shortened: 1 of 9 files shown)


elpaso:
@rouault I'm not so sure about how to handle the `mask` in general and `--nomask` (which is not supported now).

There are a few options here:
. automatically use the mask band (if any), disable with --nomask
. accept a mask specifier with dataset/band-number
. accept a dataset specifier and take the first band as a mask (current implementation, same as fillnodata).

Perhaps we can think about how to best handle the mask parameter at a later stage and keep it simple (and limited) for now.

apps/gdalalg_raster_sieve.cpp
```diff
@@ -0,0 +1,245 @@
...
+                                  GDAL_OF_RASTER, nullptr, nullptr, nullptr));
+            return static_cast<bool>(poDS);
+        });
+
+    AddBandArg(&m_band);
+    AddArg("size-threshold", 's', _("Minimum size of polygons to keep"),
+           &m_sizeThreshold)
+        .SetDefault(m_sizeThreshold);
+
+    AddArg("connectedness", 'c', _("Consider diagonal pixels as connected"),
```
rouault, line 75: can we change this to be a integer parameter, with values 4 or 8, as in gdalalg_raster_polygonize.cpp ?

apps/gdalalg_raster_sieve.cpp
```diff
@@ -0,0 +1,245 @@
...
+            }
+            else
+            {
+                options.AddString(CPLSPrintf("-co %s", co.c_str()));
+            }
+        }
+        GDALTranslateOptions *translateOptions =
+            GDALTranslateOptionsNew(options.List(), nullptr);
+        workingDSHandler.reset(GDALDataset::FromHandle(
+            GDALTranslate(workingFileName.c_str(),
```
rouault, line 234:
```suggestion
            GDALTranslate(m_outputDataset.GetName().c_str(),
```

apps/gdalalg_raster_sieve.cpp
```diff
@@ -0,0 +1,245 @@
...
+                          GDALDataset::ToHandle(workingDSHandler.get()),
+                          translateOptions, nullptr)));
+        GDALTranslateOptionsFree(translateOptions);
+    }
```
rouault, line 238:
```suggestion
        VSIUnlink(workingFileName.c_str());
    }
```

apps/gdalalg_raster_sieve.cpp
rouault: The .py test file hasn't been committed, and is likely lacking a test for that code path

apps/gdalalg_raster_sieve.cpp
```diff
@@ -0,0 +1,245 @@
...
+        for (const auto &co : m_creationOptions)
+        {
+            if (co.find('=') != std::string::npos)
+            {
+                options.AddString(co.c_str());
+            }
+            else
+            {
+                options.AddString(CPLSPrintf("-co %s", co.c_str()));
+            }
```
rouault, line 229: Test likely lacking for that code path:
```suggestion
           options.AddString("-co");
           options.AddString(co.c_str());
```

apps/gdalalg_raster_sieve.cpp
```diff
@@ -0,0 +1,245 @@
...
```
rouault, line 159:
```suggestion
           options.AddString("-co");
           options.AddString(co.c_str());
```

apps/gdalalg_raster_sieve.cpp
```diff
@@ -0,0 +1,245 @@
...
+        GDALTranslateOptionsFree(translateOptions);
+    }
+    else
+    {
+        // Translate the single required band to the destination format
+        CPLStringList options;
+        options.AddString("-b");
+        options.AddString(CPLSPrintf("%d", m_band));
+        options.AddString("-of");
+        options.AddString(m_format.c_str());
```
rouault, line 149: m_format could be empty

apps/gdalalg_raster_sieve.cpp
```diff
@@ -0,0 +1,245 @@
...
+
+    // If we were using a temporary file we need to translate to the final desired format
+    if (useTemporaryFile)
+    {
+        // Translate the temporary dataset to the final format
+        CPLStringList options;
+        options.AddString("-b");
+        options.AddString(CPLSPrintf("%d", m_band));
+        options.AddString("-of");
+        options.AddString(m_format.c_str());
```
rouault, line 219: m_format could be empty

doc/source/programs/gdal_raster_sieve.rst
```diff
@@ -0,0 +1,64 @@
...
+
+Description
+-----------
+
+:program:`gdal raster sieve` removes raster polygons smaller than a provided threshold size (in pixels)
+    and replaces them with the pixel value of the largest neighbour polygon.
+
+    The input dataset is read as integer data which means that floating point
+    values are rounded to integers. Re-scaling source data may be necessary in
+    some cases (e.g. 32-bit floating point data with min=0 and max=1).
```
rouault, line 28:
```suggestion
and replaces them with the pixel value of the largest neighbour polygon.

The input dataset is read as integer data which means that floating point
values are rounded to integers. Re-scaling source data may be necessary in
some cases (e.g. 32-bit floating point data with min=0 and max=1).
```

rouault:
> Perhaps we can think about how to best handle the mask parameter at a later stage and keep it simple (and limited) for now.

yes, I'm not sure why those algorithms have a specific parameter for the band mask and not just use the one returned by GetMaskBand()

jratike80:
`--mask <MASK>   Use the first band of the specified file as a validity mask (zero is invalid, non-zero is valid)`

I do not understand what this description means.

apps/gdalalg_raster_sieve.cpp
elpaso: We can but I guess I'm not the only one who didn't get why a two state toggle takes two integer numbers as arguments. Now I know that 4 and 8 are the adjacent pixels to consider but it took me a while to get there.

Can we find a better naming and turn it into a boolean toggle?

For instance:

--include-diagonal-pixels Include diagonal pixels in the neighbors.

apps/gdalalg_raster_sieve.cpp
elpaso: Sorry, added now.

apps/gdalalg_raster_sieve.cpp
rouault: @dbaston  Opinion whether --connectedness=4/8 is more or less intuitive than --include-diagonal-pixels ?

The concept of 4-connected/8-connected is at least referenced at https://en.wikipedia.org/wiki/Pixel_connectivity#4-connected / https://en.wikipedia.org/wiki/Pixel_connectivity#8-connected

elpaso: My point is that the argument is not an integer but a boolean because it only has two possible states. A native English speaker will probably find the best way to put the concept of diagonal pixel inclusion into understandable words.

jratike80: An image would make it understandable easily - are these connected or not?
image
I do not know how to make it understandable for beginning users with few words. I fear that `--connectedness=edge/corner` is not evident for everybody either. The best option so far is  --include-diagonal-pixels for my mind.

autotest/utilities/test_gdalalg_raster_sieve.py
```diff
@@ -0,0 +1,136 @@
...
+    tmp_filename = str(tmp_vsimem / "tmp.grd")
+
+    gdal.FileFromMemBuffer(tmp_filename, open("../alg/data/sieve_src.grd", "rb").read())
+
+    alg = get_alg()
+    alg["input"] = tmp_filename
+    alg["output"] = result_tif
+    alg["size-threshold"] = 2
+    alg.Run()
+
```
rouault, line 131:
```suggestion
    alg.Run()
    alg.Finalize()  # ensure output file is closed, so it can be later overwritten
```

apps/gdalalg_raster_sieve.cpp
dbaston: I agree that a flag is more clear. `--include-diagonal-pixels` or maybe `--connect-diagonal-pixels`

jratike80: `--connect-diagonal-pixels` is good and sounds with "connectedness".

elpaso: I have changed to `--connect-diagonal-pixels` and expanded the documentation, I hope it is clearer now.

rouault (approved):

PR #12222, opened 2025-04-28
Add 'gdal raster pixelinfo' (port of gdallocationinfo)

rouault:


doc/source/programs/gdal_raster_pixelinfo.rst
```diff
@@ -0,0 +1,236 @@
...
+
+    - ``bilinear`` applies a bilinear convolution kernel.
+
+    - ``cubic`` applies a cubic convolution kernel.
+
+    - ``cubicspline`` applies a B-Spline convolution kernel.
+
+.. option:: --overview <overview_level>
+
+    Query the (overview_level)th overview (overview_level=1 is the 1st overview),
```
jratike80, line 101: Is this true?

doc/source/programs/gdal_raster_pixelinfo.rst
```diff
@@ -0,0 +1,236 @@
...
+
+.. include:: gdal_options/if.rst
+
+.. option:: -b <band>
+
+    Selects a band to query. Multiple bands can be listed. By default all
+    bands are queried.
+
+.. option:: -p, --pos, --position <column,line> or <X,Y>
+
```
jratike80, line 54: X means always longitude or easting, Y means always latitude or northing.

doc/source/programs/gdal_raster_pixelinfo.rst
```diff
@@ -0,0 +1,236 @@
...
+    Several x,y tuples may be specified.
+
+.. option:: --position-type pixel|geoloc|user-crs
+
+    Type of position:
+
+    - ``pixel`` means that the position is set as column, line (default)
+
+    - ``geoloc`` means that th position is a georeferenced
```
jratike80, line 75: the position...

doc/source/programs/gdal_raster_pixelinfo.rst
```diff
@@ -0,0 +1,236 @@
...
+      coordinate expressed in the CRS of the dataset.
+
+    - ``user-crs`` means that th position is a georeferenced
```
jratike80, line 78: the position...

doc/source/programs/gdal_raster_pixelinfo.rst
rouault: fixed

doc/source/programs/gdal_raster_pixelinfo.rst
rouault: fixed

doc/source/programs/gdal_raster_pixelinfo.rst
rouault: fixed

doc/source/programs/gdal_raster_pixelinfo.rst
rouault: fixed

doc/source/programs/gdal_raster_pixelinfo.rst
```diff
@@ -0,0 +1,238 @@
...
+.. option:: --position-type pixel|geoloc|user-crs
+
+    Type of position:
+
+    - ``pixel`` means that the position is set as column, line (default)
+
+    - ``geoloc`` means that the position is a georeferenced
+      coordinate expressed in the CRS of the dataset.
+
+    - ``user-crs`` means that the position is a georeferenced
```
dbaston, line 80: Would `--position-type geoloc --position-crs "EPSG:4326"` be unclear? I'm not sure why `--position-type user-crs` is also needed.

doc/source/programs/gdal_raster_pixelinfo.rst
```diff
@@ -0,0 +1,238 @@
...
+
+    - ``bilinear`` applies a bilinear convolution kernel.
+
+    - ``cubic`` applies a cubic convolution kernel.
+
+    - ``cubicspline`` applies a B-Spline convolution kernel.
+
+.. option:: --ovr, --overview <index>
+
+    Query the (overview_level)th overview (overview_level=0 is the 1st overview),
```
dbaston, line 103: overview_level=1 seems more intuitive for the 1st overview, but I'm guessing this would create an inconsistency somewhere?

doc/source/programs/gdal_raster_pixelinfo.rst
```diff
@@ -0,0 +1,238 @@
...
+:program:`gdal raster pixelinfo` provide a mechanism to query information about
+a pixel given its location in one of a variety of coordinate systems.
+
+It supports outputting either as GeoJSON or CSV.
+
+The following items will be reported (when known):
+
+- Input coordinates
+- Input coordinates converted to column, line
+- Pixel value per selected band(s), with unscaled value when offset/scale is associated to the band
```
dbaston, line 32: On the one hand, this is convenient (don't have to specify if you want unscaling or not). On the other hand, it makes the output less predictable. (Imagine tossing a bunch of rasters at this and then suddenly the desired output is in `band_1_unscaled_value` instead of `band_1_value`). For example, I have a bunch of temperature data downloaded from Copernicus. The files I downloaded a couple of years ago scale the temperature into an int16, the newer ones just use a float. I'm not sure what the use case for reading scaled values is, so I would unscale by default into `band_1_value`.

doc/source/programs/gdal_raster_pixelinfo.rst
```diff
@@ -0,0 +1,238 @@
...
+
+:program:`gdal raster pixelinfo` provide a mechanism to query information about
+a pixel given its location in one of a variety of coordinate systems.
+
+It supports outputting either as GeoJSON or CSV.
+
+The following items will be reported (when known):
+
+- Input coordinates
+- Input coordinates converted to column, line
```
dbaston, line 31: I would consider making some of these opt-in to get a smaller output dataset.

doc/source/programs/gdal_raster_pixelinfo.rst
```diff
@@ -0,0 +1,238 @@
...
+
+.. program-output:: gdal raster pixelinfo --help-doc
+
+Description
+-----------
+
+:program:`gdal raster pixelinfo` provide a mechanism to query information about
+a pixel given its location in one of a variety of coordinate systems.
+
+It supports outputting either as GeoJSON or CSV.
```
dbaston, line 26: Why not create OGR features and write to any format? I can see wanting this output to Parquet or Postgres

doc/source/programs/gdal_raster_pixelinfo.rst
```diff
@@ -0,0 +1,238 @@
...
+    bands are queried.
+
+.. option:: -p, --pos, --position <column,line> or <X,Y>
+
+    Required.
+    This can be specified either as an option, a positional value after the
+    dataset name, or when called from :program:`gdal`, as (space separated)
+    values provided on the standard input.
+
+    By default, when :option:`--position-type` is not specified, or set to ``pixel``,
```
dbaston, line 60: Have you thought about having positions that are polygons or lines? This is something that people use exactextract for (output all the pixels intersecting a polygon). Long-term it would be good to have that functionality in GDAL itself.

doc/source/programs/gdal_raster_pixelinfo.rst
```diff
@@ -0,0 +1,238 @@
...
+    this is a column, line tuple (possibly with fractional part).
+    If :option:`--position-type` is set to ``geoloc``, this is a georeferenced
+    coordinate expressed in the CRS of the dataset.
+    If :option:`--position-crs` is specified, this is a georeferenced
+    coordinate expressed in the CRS defined by :option:`--position-crs`.
+
+    X means always longitude or easting, Y means always latitude or northing.
+
+    Several x,y tuples may be specified.
```
dbaston, line 69: Thoughts on the ability to read positions from a dataset?

doc/source/programs/gdal_raster_pixelinfo.rst
```diff
@@ -0,0 +1,238 @@
...
+          ]
+        }
+
+
+.. example::
+   :title: Reporting on point at WGS84 coordinates longitude=-117.6355 and latitude=33.8970 on the file :file:`byte.tif`, with CSV output format
+
+   .. code-block:: console
+
+       gdal raster pixelinfo --of=csv --position-crs=WGS84 byte.tif -117.6355 33.8970
```
jratike80, line 215: Shouldn't there be also `--position-type geoloc` or `--position-type user-crs` in this example? The default is said to be `pixel`.

doc/source/programs/gdal_raster_pixelinfo.rst
rouault: What about merging both options in a single one:   --position-crs=pixel|dataset|<crs-def> ?

doc/source/programs/gdal_raster_pixelinfo.rst
rouault: - gdallocationinfo -ovr use 1-based index
- gdal raster footprint --overview use 0-based index (like the GDAL API)
- gdal_translate -ovr and gdalwarp -ovr use 0-based index

doc/source/programs/gdal_raster_pixelinfo.rst
rouault: I would see that as a potential follow-up enhancement

doc/source/programs/gdal_raster_pixelinfo.rst
rouault: I would see that as a potential follow-up enhancement

doc/source/programs/gdal_raster_pixelinfo.rst
rouault: The tricky part is the interactive mode where you do "gdal raster pixelinfo my.tif"  and can input input coordinates interactively. We want a text output in that mode. I would say that binary mode only makes sense if we accept input coordinates from a vector dataset, so as a later enhancement

doc/source/programs/gdal_raster_pixelinfo.rst
dbaston: > What about merging both options in a single one:  --position-crs=pixel|dataset|<crs-def> ?

Seems good to me.

doc/source/programs/gdal_raster_pixelinfo.rst
dbaston: Sure, I just raise the topic for the purpose of planning the interface.

doc/source/programs/gdal_raster_pixelinfo.rst
rouault: fixed

doc/source/programs/gdal_raster_pixelinfo.rst
rouault: band_1_raw_value and band_1_unscaled_value now systematically emitted

doc/source/programs/gdal_raster_pixelinfo.rst
rouault: just outputting the floating point column, line coordinates instead of the _int and _double version

doc/source/programs/gdal_raster_pixelinfo.rst
rouault: done

PR #12223, opened 2025-04-28
Add STACIT driver examples

geographika:
## What does this PR do?

Adds command-line and Python examples to the STACIT driver page, using the tabs Sphinx plugin. 

image

This is a sample pull request to get feedback on if this is a good approach to use for other drivers. 

## Questions

cc @dbaston @rouault

- What is the best approach for handling the old and new command-line interfaces? Include both with a ` # from GDAL 3.11+` note? Only use the new approach for the latest docs?
- I have added `bash` (Linux) examples, and `PowerShell` (Windows) examples - as quoting can differ between the two operating systems (I ran into issues with the current examples). My personal preference is to not also include `cmd` Windows examples, but maybe these should also be added?
- I added the Python examples as separate scripts, and parts can then be included in the docs. This allows them to be linted (and possibly automate testing them as a CI task). Is this a good approach?

Questions specific to these examples:

- When should `Open` be used in Python, and when `OpenEx`? Here I needed to force the driver in Python as noted in https://lists.osgeo.org/pipermail/gdal-dev/2024-March/058667.html or I got `RuntimeError: `https://planetarycomputer.microsoft.com/api/stac/v1/search?collections=naip&bbox=-100,40,-99,41&datetime=2019-01-01T00:00:00Z%2F..' does not exist in the file system, and is not recognized as a supported dataset name.`

### DIFF
--- a/doc/source/drivers/raster/stacit.rst
+++ b/doc/source/drivers/raster/stacit.rst
@@ -141,6 +141,7 @@ Open a subdataset returned by the above request:
    .. code-tab:: bash
 
     gdalinfo "STACIT:\"https://planetarycomputer.microsoft.com/api/stac/v1/search?collections=naip&bbox=-100,40,-99,41&datetime=2019-01-01T00:00:00Z%2F..\":asset=image"
+    # from GDAL 3.11+
     gdal raster info "STACIT:\"https://planetarycomputer.microsoft.com/api/stac/v1/search?collections=naip&bbox=-100,40,-99,41&datetime=2019-01-01T00:00:00Z%2F..\":asset=image"
 
    .. code-tab:: ps1
--- a/doc/source/drivers/raster/stacit.rst
+++ b/doc/source/drivers/raster/stacit.rst
@@ -10,13 +10,17 @@ STACIT - Spatio-Temporal Asset Catalog Items
 
 .. built_in_by_default::
 
-This driver supports opening STAC API ItemCollections, with the input usually being a `STAC API search query <https://github.com/radiantearth/stac-api-spec/tree/main/item-search>`_ or the results saved as a JSON file. Items in the response must include projection information following the `Projection Extension Specification <https://github.com/stac-extensions/projection/>`_.
+This driver supports opening STAC API ItemCollections, with the input usually being a `STAC API search query <https://github.com/radiantearth/stac-api-spec/tree/main/item-search>`_
+or the results saved as a JSON file. Items in the response must include projection information following the
+`Projection Extension Specification <https://github.com/stac-extensions/projection/>`_.
 It builds a virtual mosaic from the items.
 
 A STACIT dataset which has no subdatasets is actually a :ref:`raster.vrt` dataset.
 Thus, translating it into VRT will result in a VRT file that directly references the items.
 
-Note that `STAC API ItemCollections <https://github.com/radiantearth/stac-api-spec/blob/main/fragments/itemcollection/README.md>`_ are not the same as  `STAC Collections <https://github.com/radiantearth/stac-spec/tree/master/collection-spec>`_. STAC API ItemCollections are GeoJSON FeatureCollections enhanced with STAC entities.
+Note that `STAC API ItemCollections <https://github.com/radiantearth/stac-api-spec/blob/main/fragments/itemcollection/README.md>`_
+are not the same as  `STAC Collections <https://github.com/radiantearth/stac-spec/tree/master/collection-spec>`_.
+STAC API ItemCollections are GeoJSON FeatureCollections enhanced with STAC entities.
 
 Open syntax
 -----------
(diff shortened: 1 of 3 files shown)


rouault:
> * When should `Open` be used in Python, and when `OpenEx`?

Open(name) is an alias of OpenEx(name, gdal.OF_RASTER | gdal.OF_VERBOSE_ERROR)

>  Here I needed to force the driver in Python 

yes, this is expected as there isn't enough context otherwise for the driver to guess if a URL can be recognized by it without downloading the file

doc/source/drivers/raster/stacit.rst
```diff
@@ -108,17 +112,47 @@ Examples
...
+
+    gdalinfo 'STACIT:\"https://planetarycomputer.microsoft.com/api/stac/v1/search?collections=naip&bbox=-100,40,-99,41&datetime=2019-01-01T00:00:00Z%2F..\"'
+    # from GDAL 3.11+
+    gdal raster info 'STACIT:\"https://planetarycomputer.microsoft.com/api/stac/v1/search?collections=naip&bbox=-100,40,-99,41&datetime=2019-01-01T00:00:00Z%2F..\"'
+
+   .. tab:: Python
 
+      .. literalinclude :: examples/stacit.py
+         :language: python
+         :end-before: # subdatasets
```
dbaston, line 134:
```suggestion
         :start-after: # Example 1
         :end-before: # Example 2
```

A convention like this would make the `stacit.py` file more clear, I think.

doc/source/drivers/raster/stacit.rst
```diff
@@ -108,17 +112,47 @@ Examples
...
+
+   .. code-tab:: ps1
+
+    gdalinfo 'STACIT:\"https://planetarycomputer.microsoft.com/api/stac/v1/search?collections=naip&bbox=-100,40,-99,41&datetime=2019-01-01T00:00:00Z%2F..\"'
+    # from GDAL 3.11+
+    gdal raster info 'STACIT:\"https://planetarycomputer.microsoft.com/api/stac/v1/search?collections=naip&bbox=-100,40,-99,41&datetime=2019-01-01T00:00:00Z%2F..\"'
+
+   .. tab:: Python
 
+      .. literalinclude :: examples/stacit.py
```
dbaston, line 132: I'd suggest mirroring the organization of the rst files, so an example included from `drivers/raster/stacit.rst` is found in `examples/drivers/raster/stacit.py`

dbaston (commented):
> What is the best approach for handling the old and new command-line interfaces? Include both with a  # from GDAL 3.11+ note? Only use the new approach for the latest docs?

Here, the purpose being to demonstrate the usage of a _driver_, I don't think both are needed. I might suggest the old one only, since the new one is unstable. Hopefully the commands we are using are simple enough that any differences are rather trivial?

> I have added bash (Linux) examples, and PowerShell (Windows) examples - as quoting can differ between the two operating systems (I ran into issues with the current examples). My personal preference is to not also include cmd Windows examples, but maybe these should also be added?

This makes sense. I don't think we need both bash and Powershell unless they are actually difference (and here I can see that they are)

> I added the Python examples as separate scripts, and parts can then be included in the docs. This allows them to be linted (and possibly automate testing them as a CI task). Is this a good approach?

Sounds good to me.

doc/source/drivers/raster/stacit.rst
```diff
@@ -108,17 +112,47 @@ Examples
...
+    gdal raster info "STACIT:\"https://planetarycomputer.microsoft.com/api/stac/v1/search?collections=naip&bbox=-100,40,-99,41&datetime=2019-01-01T00:00:00Z%2F..\""
+
+
+   .. code-tab:: ps1
+
+    gdalinfo 'STACIT:\"https://planetarycomputer.microsoft.com/api/stac/v1/search?collections=naip&bbox=-100,40,-99,41&datetime=2019-01-01T00:00:00Z%2F..\"'
+    # from GDAL 3.11+
+    gdal raster info 'STACIT:\"https://planetarycomputer.microsoft.com/api/stac/v1/search?collections=naip&bbox=-100,40,-99,41&datetime=2019-01-01T00:00:00Z%2F..\"'
+
+   .. tab:: Python
```
dbaston, line 130: Curious, why not `.. code-tab:: py` ?

geographika: I discovered it wasn't supported: https://sphinx-tabs.readthedocs.io/en/latest/
> Sphinx’s include directive does not work inside code-tab. Use tab or group-tab instead.

dbaston: Ah, thanks.

doc/source/drivers/raster/stacit.rst
geographika: Restructured as suggested.

doc/source/drivers/raster/stacit.rst
geographika: Done

geographika:
> Here, the purpose being to demonstrate the usage of a _driver_, I don't think both are needed. I might suggest the old one only, since the new one is unstable. Hopefully the commands we are using are simple enough that any differences are rather trivial?

I'd like to include both, as it makes it easy to see the two approaches side-by-side, which will help people migrate. 
The comment `# from GDAL 3.11+` however isn't great, maybe `# new client API (GDAL 3.11+)`. 

Alternatively, I can leave hidden as an RST comment until the new API is stabilised? If it does change it should be easy to find/replace.

doc/source/drivers/raster/stacit.rst
```diff
@@ -108,16 +112,48 @@ Examples
...
 
 Open a subdataset returned by the above request:
 
-::
+.. tabs::
+
+   .. code-tab:: bash
 
     gdalinfo "STACIT:\"https://planetarycomputer.microsoft.com/api/stac/v1/search?collections=naip&bbox=-100,40,-99,41&datetime=2019-01-01T00:00:00Z%2F..\":asset=image"
+    gdal raster info "STACIT:\"https://planetarycomputer.microsoft.com/api/stac/v1/search?collections=naip&bbox=-100,40,-99,41&datetime=2019-01-01T00:00:00Z%2F..\":asset=image"
```
rouault, line 144:
```suggestion
    # from GDAL 3.11+
    gdal raster info "STACIT:\"https://planetarycomputer.microsoft.com/api/stac/v1/search?collections=naip&bbox=-100,40,-99,41&datetime=2019-01-01T00:00:00Z%2F..\":asset=image"
```

rouault (approved):

PR #12259, opened 2025-05-03
gdal raster reclassify: allow it to accept a MEM input dataset

rouault:
FYI @dbaston : using ``VRTDerivedRasterBand::AddSimpleSource(GDALRasterBand*)`` allows it to work with a unnamed / MEM source dataset


dbaston (approved):
Nice simplification, thanks!

PR #12267, opened 2025-05-04
gdal raster calc: add support for GDALG output

rouault:
CC @dbaston


apps/gdalalg_raster_fill_nodata.cpp
```diff
@@ -103,11 +103,18 @@ bool GDALRasterFillNodataAlgorithm::RunImpl(GDALProgressFunc pfnProgress,
     if (!m_overwrite && !m_outputDataset.GetName().empty() &&
```
dbaston, line 103:
```suggestion
    if (!m_outputDataset.GetName().empty() &&
```

rouault: fixed

PR #12274, opened 2025-05-05
CMake: unset GDAL/OGR_ENABLE_DRIVER_xxxx when setting GDAL/OGR_BUILD_OPTIONAL_DRIVERS=ON the first time

rouault:


rouault:
CC @abellgithub

PR #12275, opened 2025-05-05
Add linking to -latomic for std::atomic<uint64_t> on architectures where it is needed

rouault:


gdt:
Is this going to add -latomic on other systems?  As I understand it, somewhat fuzzily, sometimes this is built in, and sometimes you need a library.  But maybe this only runs if the atomic ops are missing.

As I mentioned on list, pkgsrc has a makefile  fragment to ensure 64-bit atomics are present, either by linking in the library (which I think the program has to link), or changing the cpu flags:
```
if !defined(ATOMIC64_MK)

ATOMIC64_MK:=

.include "../../mk/bsd.fast.prefs.mk"

.if ${MACHINE_ARCH} == "i386"
.  if empty(CFLAGS:M*march=*)
CFLAGS+=        -march=i586
.  endif
.endif

# Non-NetBSD vendors tend to include libatomic as part of gcc.
.if ${OPSYS} == "NetBSD" && \
    (${MACHINE_ARCH} == "m68k" || \
    ${MACHINE_ARCH} == "mipsel" || \
    ${MACHINE_ARCH} == "mipseb" || \
    ${MACHINE_ARCH} == "powerpc" || \
    ${MACHINE_ARCH} == "sh3el" || \
    ${MACHINE_ARCH} == "sh3eb" || \
    ${MACHINE_ARCH} == "sparc" || \
    ${MACHINE_ARCH} == "vax" || \
    !empty(MACHINE_ARCH:Mearm*))
.  include "../../devel/libatomic/buildlink3.mk"
.endif

.endif
```

rouault:
> Is this going to add -latomic on other systems?

perhaps you could check the diff ;-)

cmake/helpers/configure.cmake
```diff
@@ -397,6 +398,39 @@ else ()
     HAVE_SHARED_MUTEX
   )
 
+  # std::atomic<int64_t> requires linking against -latomic on some esoteric platforms
```
gdt, line 401: Perhaps "many 32-bit CPUs" would give the reader a hint in the right direction, vs esoteric.

gdt:
> perhaps you could check the diff ;-)

I tried to read it but find cmake confusing.  Trying again, it seems like there's a test compile without -latomic, and nothing is added if that works.   So that means on i386 where -mi586 or higher is in force, that will succeed.    And then on most other 32-bit platforms, the first test will fail and it will try -latomic.

So this seems like a good fix.

cmake/helpers/configure.cmake
rouault: rephrased

PR #12279, opened 2025-05-06
gdal raster tile: add --kml option

rouault:


rouault:
cherry-picked to 3.11 per 6f8432b16a 99cd1d452a

PR #12283, opened 2025-05-06
GPKG: call sqlite3_errmsg() in error messages

### LINKED ISSUE #12247: Include internal SQLite3-Error-Messages into GDAL-Error-Messages
### Feature description

Some drivers are internally based on the **SQLite SDK**, e.g. the Geopackage driver. If there occures an error while using the SQLite-Api, it would be nice if the **last error-reason** for an SQLite error would be determined and reported as part of the GDAL error. Otherwise, it is very difficult to determine the real cause of the error in real-life-scenarios (without being able to change the source code).
For example in `ogrgeopackagetablelayer.cpp`: In some calls of sqlite3-functions, such as in `OGRGeoPackageTableLayer::CreateOrUpsertFeature`, this has already been implemented and the result of `sqlite3_errmsg` is included in the `CPLError`-message. In other places, such as `OGRGeoPackageTableLayer::ResetStatementInternal`, it is not.

### Additional context

_No response_

rouault:
Fix #12247

### DIFF
--- a/ogr/ogrsf_frmts/gpkg/ogrgeopackagedatasource.cpp
+++ b/ogr/ogrsf_frmts/gpkg/ogrgeopackagedatasource.cpp
@@ -2714,7 +2714,7 @@ bool GDALGeoPackageDataset::OpenRaster(
                 }
                 sqlite3_stmt *hSQLStmt = nullptr;
                 int rc =
-                    sqlite3_prepare_v2(hDB, pszSQL, -1, &hSQLStmt, nullptr);
+                    SQLPrepareWithError(hDB, pszSQL, -1, &hSQLStmt, nullptr);
 
                 if (rc == SQLITE_OK)
                 {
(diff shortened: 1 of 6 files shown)


rouault:
cherry-picked to 3.11 per 41df7321e1

ogr/ogrsf_frmts/gpkg/ogrgeopackagedatasource.cpp
```diff
@@ -10153,8 +10146,8 @@ bool GDALGeoPackageDataset::AddFieldDomain(
             if (sqlite_err != SQLITE_OK && sqlite_err != SQLITE_DONE)
             {
                 CPLError(CE_Failure, CPLE_AppDefined,
-                         "failed to execute insertion: %s",
-                         sqlite3_errmsg(hDB));
+                         "failed to execute insertion '%s': %s",
+                         sqlite3_errmsg(hDB), sqlite3_errmsg(hDB));
```
nyalldawson, line 10150: Is this double up intended?

rouault: good catch. fixed in d8530eb3b004271b77566e5c6fee30e37712a37d

PR #12290, opened 2025-05-06
Support long schema/table/column name

fechen123:
## What does this PR do?
Support long schema/table/column name. It used to only support length up to 64, not it is supported up to 128, which is the maximum length of the table/column name in Oracle.  

N/A

## Tasklist

 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments
 - [ ] ADD YOUR TASKS HERE

## Environment

* OS:
* Compiler:


jratike80:
Do you have any better reference than this one https://docs.oracle.com/en/database/oracle/oracle-database/23/odpnt/EFCoreIdentifier.html. Maybe add a comment line to the code that 128 means bytes, not characters.

fechen123:
> Do you have any better reference than this one https://docs.oracle.com/en/database/oracle/oracle-database/23/odpnt/EFCoreIdentifier.html. Maybe add a comment line to the code that 128 means bytes, not characters.

ok, I will add a comment line.

PR #12291, opened 2025-05-06
OGRGeometry::OGRexportToSFCGAL(): fix for 3D geometries for SFCGAL >= 1.5.2

rouault:
@lbartoletti  When testing https://gitlab.com/sfcgal/SFCGAL/-/merge_requests/383 and running autotest/ogr/ogr_geometry.py, I saw SFCGAL tests where skipped even when the GDAL build was against SFCGAL. I realize this was due to the SFCGAL presence detection test failing due to broken 3D geometries. By default OGR export to pre-ISO WKB ...


lbartoletti:
OK, Thanks! Indeed, on SFCGAL, we use only ISO-WKB

Issue #12292, opened 2025-05-07
ogrmerge: warnings when creating KMZ from Chinese filenames, and about DateTime fields


jidanni:
### What is the bug?

Just the other day I did
$ ogrmerge -f LIBKML -overwrite_ds -o street_name_groups.kmz 安順.kml 崇德.kml
>Warning 1: The output driver does not natively support DateTime type for field timestamp. Misconversion can happen. -mapFieldType can be used to control field type conversion...

>Warning 1: One or several characters couldn't be converted correctly from UTF-8 to CP437.  This warning will not be emitted anymore

Convert from UTF-8 to CP437? No thanks!

And there's no workaround. Also the problem is not mentioned in
https://gdal.org/en/latest/drivers/vector/kml.html#encoding-issues

Certainly the problem is OGR/GDAL was never tested on other than ASCII filenames?!

GDAL 3.10.3, released 2025/04/01

Here are the input KML files if you want to test: [files.cpio.gz](https://github.com/user-attachments/files/20076130/files.cpio.gz)

### Q&A
> Do you have any particular evidence that something got damaged when writing your .KMZ?

No. Maybe the only thing bad is the error message.

jratike80:
Please try if you could edit the title to describe the problem better and sharp.

Have you tested other gdal utilities? I do not get code page warning on Windows with `ogr2ogr -f LIBKML test安順.kml 安順.kml`.

ogrmerge is a Python script and Python may require extra tricks with codepages. Fortunately there will be a new native program
https://gdal.org/en/latest/programs/gdal_vector_concat.html in GDAL 3.11. It seems to work without issues for me.

```
gdal vector concat 安順.kml 崇德.kml output安順崇德.kml
Warning 1: Several drivers matching kml extension. Using LIBKML
Warning 1: The output driver does not natively support DateTime type for field timestamp. Misconversion can happen. -mapFieldType can be used to control field type conversion.
Warning 1: The output driver does not natively support DateTime type for field begin. Misconversion can happen. -mapFieldType can be used to control field type conversion.
Warning 1: The output driver does not natively support DateTime type for field end. Misconversion can happen. -mapFieldType can be used to control field type conversion.
```

jratike80:
About how to turn off the messages, by reading https://gdal.org/en/stable/drivers/vector/libkml.html#fields it seems that the warnings should not exist at all because datetime is supported. And ogr2ogr creates those 3 fields into the output KML as DateTime.

```
[...]

PR #12294, opened 2025-05-07
Update ogrmerge.rst to mention gdal vector concat

jidanni:
See https://github.com/OSGeo/gdal/issues/12292 .

P.S., I certainly got the link syntax wrong, especially the spaces. Please fix it for me. Thanks.

### DIFF
--- a/doc/source/programs/ogrmerge.rst
+++ b/doc/source/programs/ogrmerge.rst
@@ -62,7 +62,7 @@ potential manual editing of it and :program:`ogr2ogr` can be done.
 
 .. note::
 
-    Starting with GDAL 3.11 there is a newer :program:`gdal vector concat` program available.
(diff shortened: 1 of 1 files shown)


doc/source/programs/ogrmerge.rst
```diff
@@ -60,6 +60,10 @@ potential manual editing of it and :program:`ogr2ogr` can be done.
 
     ogrmerge is a Python utility, and is only available if GDAL Python bindings are available.
 
+.. note::
+
+    Starting with GDAL 3.11 there is a newer :program:`gdal vector concat` program available.
```
rouault, line 65:
```suggestion
    Starting with GDAL 3.11, a newer :ref:`gdal_vector_concat` program is available.
```

rouault:
> New alternative for ogrmerge.py might be an exception.

That's probably fine here as there's a 1:1 mapping between them. And we need to advertize them a bit so people can try them

PR #12306, opened 2025-05-08
Add a glossary and example link

geographika:
The GDAL docs contain a lot of abbreviations and technical terms. A glossary defining these terms and abbreviations could be of use to a reader. 

This PR adds a glossary page to the GDAL docs, with an example link, as a starting point for discussions to see if this would be a good idea to include and expand upon. 

The glossary uses the standard Sphinx glossary approach, similar to MapServer https://mapserver.org/glossary.html

image

Terms can link to the glossary using the following syntax. An example is included in this PR. 

```
:term:`EPSG`
```

image

There is an OSGeo Lexicon mailing list that was looking to "Coordinate definitions for terminology used within the OSGeo context" - https://lists.osgeo.org/mailman/listinfo/lexicon. I've emailed the list to see if there are any definition datasets that could be used as a starting point.

### DIFF
--- a/doc/source/glossary.rst
+++ b/doc/source/glossary.rst
@@ -108,6 +108,12 @@ The GDAL glossary contains terms and acronyms found throughout the GDAL document
 
         Well-Known Text. Text representation of geometries described in the Simple Features for SQL (SFSQL) specification.
 
+    WKT-CRS
+
+        Well-Known Text for Coordinate Reference Systems. A text format that defines how to describe coordinate reference
+        systems and transformations between them in a standardized way.
+        See the `OGC WKT-CRS standard <https://www.ogc.org/standards/wkt-crs/>__.
+
     WKB
 
         Well-Known Binary. Binary representation of geometries described in the Simple Features for SQL (SFSQL) specification.
--- a/doc/source/glossary.rst
+++ b/doc/source/glossary.rst
@@ -1,0 +1,128 @@
+.. _glossary:
+
+================================================================================
+Glossary
+================================================================================
+
+The GDAL glossary contains terms and acronyms found throughout the GDAL documentation
+
+.. glossary::
+
+    Affine Transformation
+
+        A geometric change that preserves points, straight lines, and ratios, including scaling, rotating, or translating.
+
+    API
+
+        Application Programming Interface. A set of rules that lets different software programs communicate and share data or functions.
+
+    DEM
+
+        Digital Elevation Model. A raster representation of the height of the earth's surface.
+
+    Euclidean Distance
+
+        A measure of the straight-line distance between two points in space.
+
+    EPSG
+
+        European Petroleum Survey Group. A group of geospatial experts for the petroleum industry (1986–2005),
+        now part of IOGP (International Association of Oil & Gas Producers). Known for creating the EPSG Geodetic Parameter Dataset,
+        a widely used database of coordinate systems and geodetic parameters.
+
+    Gridding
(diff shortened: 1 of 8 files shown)


jratike80 (approved):

geographika:
Note docs builds were failing due to https://github.com/sphinx-doc/sphinx/issues/13533 which seems to now be resolved: https://github.com/sphinx-doc/sphinx/issues/13533#issuecomment-2862775896

geographika:
@dbaston @rouault any opinions on if adding a glossary should be pursued? The entries in this PR were just to show the structure - obviously a base set of terms would be added before merging, and then built upon as I go through the drivers and tutorials. 
There were also a couple of replies on the lexicon list with links to previous datasets that could be used to help populate an inital glossary: https://lists.osgeo.org/pipermail/lexicon/2025-May/thread.html

rouault:
@geographika Yes definitely a good idea! There is so much jargon that can be explained

doc/source/glossary.rst
```diff
@@ -0,0 +1,18 @@
...
+
+    EPSG
+
+        European Petroleum Survey Group. A group of geospatial experts for the petroleum industry (1986–2005),
+        now part of IOGP (International Association of Oil & Gas Producers). Known for creating the EPSG Geodetic Parameter Dataset,
+        a widely used database of coordinate systems and geodetic parameters.
+
+    VSI
+
+        Virtual System Interface. An interface for accessing non-standard datasets, such as
```
dbaston, line 17:
```suggestion
        Virtual System Interface. An interface for accessing files and datasets in non-filesystem locations, such as
```

dbaston:
> any opinions on if adding a glossary should be pursued?

I like it!

jratike80 (approved):
Good start.

doc/source/glossary.rst
```diff
@@ -0,0 +1,128 @@
...
+
+        Virtual System Interface. An interface for accessing files and datasets in non-filesystem locations, such as
+        in-memory files, zip files, and over network protocols.
+
+        .. seealso::
+            :ref:`virtual_file_systems`.
+
+    WKT
+
+        Well-Known Text. Text representation of geometries described in the Simple Features for SQL (SFSQL) specification.
```
rouault, line 109: There should be a note that WKT is also used in the context of CRS, and WKT CRS != WKT geometry

geographika: Added a WKT-CRS entry just after WKT

geographika:
@dbaston @rouault - although only a few entries, it would be good to merge this so I can start linking to terms in other places in the docs in follow-up PR.

doc/source/glossary.rst
rouault: > Added a WKT-CRS entry just after WKT

you probably forgot to push that commit

doc/source/glossary.rst
```diff
@@ -0,0 +1,134 @@
...
+
+    WKT-CRS
+
+        Well-Known Text for Coordinate Reference Systems. A text format that defines how to describe coordinate reference
+        systems and transformations between them in a standardized way.
+        See the `OGC WKT-CRS standard <https://www.ogc.org/standards/wkt-crs/>__.
```
rouault:
```suggestion
        See the `OGC WKT-CRS standard <https://www.ogc.org/standards/wkt-crs/>`__.
```

rouault (approved):

PR #12307, opened 2025-05-08
GDALRegenerateOverviewsMultiBand(): raise threshold to go to on-disk …

### LINKED ISSUE #12303: GDALRegenerateOverviewsMultiBand has problems with large rasters.
### What is the bug?

GDALRegenerateOverviewsMultiBand has one code path that builds the overviews in a temporary, uncompressed TIF file and then copies the data to the actual destination. From what I recall, this function used to be able to deal with large, tiled and pixel interleaved rasters very well, it doesn't seem to be the case any longer.

The code in question is part of the commit 831df5a commit 
This presents two problems, especially for large, pixel interleaved rasters.

1) The temporary uncompressed TIF can be huge, much larger than the intended output size. Even if it isn't a problem by itself, the temporary TIF exists disk at the same time with the output, at least doubling the IO and storage required, for a while. The temporary file is unlinked but still using file system space until it gets closed.

2) The copy to the final destination is frequently done one band at a time, using GdalRasterBandCopyWholeRaster(). For pixel interleaved, tiled outputs, this usually ends up with tiles being written multiple times, with partial bands data, due to the memory block cache being under pressure. As far as I know, many tiled datasets don't deal very well with tiles being written multiple times with different data, in the best case resulting in bloated output.

The path described above is triggered when the memory required to generate an overview is larger than 10MB, which is dependent on the block size, number of bands and data type.
I used an MRF with BLOCKSIZE=1024, RGB byte, total size of 11k by 11k pixels. This triggers the described path when trying to generate MRF internal overviews.  It does not necessarily show the full extent of the problem, since both the disk space and the block cache might be able to take the 100MB or so temporary files, even if the content is all black, or something that will compress a lot (JPEG tiles for example).

### Versions and provenance

master branch

### Additional context

I'm not sure if it's affecting many raster formats, the function is used by gtiff, jpeg, mbtiles, mrf, rmf, tiledb and gpkg drivers.

rouault:
…temporary file

to be at least to accomodate of 1024 pixel block size and 3 bands. And also make GTiff temporary file tiled and compressed.

And flush overview bands when going through that temporary file code path.

Fixes #12303

I've verified that no temporary file is created with:
```
$ gdal_translate autotest/gdrivers/data/small_world.tif tmp.mrf -of mrf -co blocksize=1024 -outsize 11000 11000
$ gdaladdo -r average tmp.mrf 2
```

### DIFF
--- a/autotest/gcore/tiff_ovr.py
+++ b/autotest/gcore/tiff_ovr.py
@@ -2681,7 +2681,31 @@ def test_tiff_ovr_external_mask_update(tmp_vsimem):
 # of the default scanlines based one
 
 
-def test_tiff_ovr_fallback_to_multiband_overview_generate():
+@pytest.mark.parametrize(
+    "config_options",
+    [
+        {
+            "GDAL_OVR_CHUNK_MAX_SIZE": "1000",
+            "GDAL_OVR_CHUNK_MAX_SIZE_FOR_TEMP_FILE": "10000",
+        },
+        {
+            "GDAL_OVR_CHUNK_MAX_SIZE": "1000",
+            "GDAL_OVR_CHUNK_MAX_SIZE_FOR_TEMP_FILE": "10000",
+            "GDAL_OVR_TEMP_DRIVER": "MEM",
+        },
+        {
+            "GDAL_OVR_CHUNK_MAX_SIZE": "1000",
+            "GDAL_OVR_CHUNK_MAX_SIZE_FOR_TEMP_FILE": "10000",
+            "GDAL_OVR_TEMP_DRIVER": "GTIFF",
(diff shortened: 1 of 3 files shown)


lucianpls:
@rouault 
How about calling itself recursively with the natural chunk when possible, in the external file code path? It would solve the remaining problems, and likely avoid the flushing and compressing the temporary file. It would use more files, but they're small and short lived.
It looks to be fairly easy, using the existing XOFF/YOFF/XSIZE/YSIZE parameters. First pass chunks and call itself (all bands), only using the external file when it can't be chunked any further.

lucianpls:
OK, I'll try!

PR #12308, opened 2025-05-08
cmake_builds.yml:  numpy>=2

mdsumner:
Can we bump the numpy issue in the builds? 

## What does this PR do?

Unpins the numpy version from 1.26.4 in docker builds. 

https://lists.osgeo.org/pipermail/gdal-dev/2024-July/059198.html

A slightly cheeky PR since I'm not otherwise sure how to test this in context.   I really want to have numpy>2 when leveraging the docker image 🙏

### DIFF
--- a/.github/workflows/cmake_builds.yml
+++ b/.github/workflows/cmake_builds.yml
@@ -243,7 +243,7 @@ jobs:
         source myvenv/bin/activate
         # Set PATH so that gdal-config is found
         PYTHON_CMD=python3 && PATH=$GITHUB_WORKSPACE/install-gdal/bin:$PATH $PYTHON_CMD -m pip uninstall numpy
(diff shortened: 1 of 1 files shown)


rouault:
I'm not sure this PR is needed. We do have CI target that have numpy >= 2, for example the Alpine edge based ones:  https://github.com/OSGeo/gdal/actions/runs/14880474452/job/41787180023 : "Installing py3-numpy-dev (2.2.4-r0)"

> I really want to have numpy>2 when leveraging the docker image 🙏

That would be a different stories as we rely as much as possible on using packages of the distribution. So for alpine:3.21 based, they are already at numpy 2.1.3 (https://pkgs.alpinelinux.org/packages?name=py3-numpy&branch=v3.21&repo=&arch=x86_64&origin=&flagged=&maintainer=)  . The ubuntu 24.04 based ones will still ship with numpy 1.26.4 , until we upgrade to 26.04

mdsumner:
ah ok, thanks for the explanation

PR #12313, opened 2025-05-08
Add tabs to MDM raster docs

geographika:
PR also switches  `.. tab:: Python` to `.. group-tab:: Python`. This ensures all tabs in the group are switched to when one is selected, similar to `.. code-tab:`.

`.. code-tab:` cannot be used here, as it does not work with `literalinclude`.


rouault (approved):

geographika:
> @geographika Can you add "Doc: " as a prefix for the title of your commit messages ?

Sorry, only just saw this. Yes will include this in future.

PR #12317, opened 2025-05-08
Add 'gdal raster rgb-to-palette' (port of rgb2pct.py)

### LINKED ISSUE #12287: Task: port rgb2pct as 'gdal raster rgb-to-palette'
### Feature description

-

### Additional context

_No response_

rouault:
Fixes #12287

### DIFF
--- a/apps/gdalalg_raster_rgb_to_palette.cpp
+++ b/apps/gdalalg_raster_rgb_to_palette.cpp
@@ -53,6 +53,7 @@ GDALRasterRGBToPaletteAlgorithm::GDALRasterRGBToPaletteAlgorithm()
         .SetDefault(m_colorCount)
         .SetMinValueIncluded(2)
         .SetMaxValueIncluded(256);
+    AddArg("color-map", 0, _("Color map filename"), &m_colorMap);
 }
 
 /************************************************************************/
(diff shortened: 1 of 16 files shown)


apps/gdalalg_raster_rgb_to_palette.cpp
```diff
@@ -0,0 +1,323 @@
...
+    {
+        m_format = GetOutputDriverForRaster(m_outputDataset.GetName().c_str());
+        if (m_format.empty())
+        {
+            ReportError(CE_Failure, CPLE_AppDefined,
+                        "Cannot guess output driver from output filename");
+            return false;
+        }
+    }
+    if (EQUAL(m_format.c_str(), "JPEG") || EQUAL(m_format.c_str(), "WEBP"))
```
elpaso, line 79: Don't we have a capability to query for color table support? 
This approach seems a bit fragile to me.

apps/gdalalg_raster_rgb_to_palette.cpp
```diff
@@ -0,0 +1,323 @@
...
+    {
+        ReportError(CE_Failure, CPLE_NotSupported,
+                    "Input dataset must have at least 3 bands");
+        return false;
+    }
+    else if (nSrcBandCount >= 4)
+    {
+        ReportError(
+            CE_Warning, CPLE_AppDefined,
+            "Only R,G,B bands of input dataset will be taken into account");
```
elpaso, line 98: if alpha is not taken, shouldn't this check  be >= 3 ?

apps/gdalalg_raster_rgb_to_palette.cpp
```diff
@@ -0,0 +1,323 @@
...
+        }
+    }
+
+    if (nFound < 3)
+    {
+        if (nFound > 0)
+        {
+            ReportError(
+                CE_Warning, CPLE_AppDefined,
+                "Assuming first band is red, second green and thirst blue, "
```
elpaso, line 137:
```suggestion
                "Assuming first band is red, second green and third is blue, "
```

apps/gdalalg_raster_rgb_to_palette.cpp
```diff
@@ -0,0 +1,323 @@
...
+        CPLStringList aosTmpCO;
+        if (!bNeedTmpFile)
+            aosTmpCO = CPLStringList(m_creationOptions);
+        poTmpDS.reset(poTmpDriver->Create(
+            osTmpFilename.c_str(), poSrcDS->GetRasterXSize(),
+            poSrcDS->GetRasterYSize(), 1, GDT_Byte, aosTmpCO.List()));
+        bOK = poTmpDS != nullptr;
+        if (bOK && bNeedTmpFile)
+        {
+            VSIUnlink(osTmpFilename.c_str());
```
elpaso, line 260: Am I misreading this or you are deleting the temp file created at line 254?

elpaso: I guess I misinterpreted what bNeedTmpFile means, does it mean that it needs to be kept instead of unlinked?

apps/gdalalg_raster_rgb_to_palette.cpp
rouault: > Don't we have a capability to query for color table support?

no, but (most )drivers that do not support color table will error out. Here I anticipated for quite common use cases to emit a clean error message

apps/gdalalg_raster_rgb_to_palette.cpp
rouault: we want to warn if there is an alpha band, hence >= 4

PR #12336, opened 2025-05-10
Doc: fix (just a few) missing target references

rouault:
found by setting ``nitpicky = True`` in conf.py


godofcodeeyes:
The performance optimizations here are quite clever. Nice engineering work!

PR #12335, opened 2025-05-10
Doc: use :source_file: whereever possible

rouault:

PR #12341, opened 2025-05-11
dev_documentation.rst: make :source_file: custom directive more visible

rouault:
CC @geographika @DelazJ   borrows content from Seth's note at https://github.com/OSGeo/gdal/pull/12335#issuecomment-2869659885


DelazJ:
@rouault Is this different from using https://www.sphinx-doc.org/en/master/usage/extensions/extlinks.html?

Issue #12192, opened 2025-04-22
3.11.0beta fails to build from source for arm32 android


dg0yt:
### What is the bug?

Test build of 3.11.0beta in vcpkg fails for the arm-neon-android triplet with:
~~~
FAILED: gcore/CMakeFiles/gcore.dir/overview.cpp.o 
[...]

PR #12343, opened 2025-05-11
gdal_minmax_element.hpp: fix build issue with ARM32 Neon optimizations

rouault:
Fixes https://github.com/OSGeo/gdal/issues/12192#issuecomment-2869854141


dg0yt:
This works for vcpkg arm-neon-android. Thank you.

PR #12345, opened 2025-05-11
autotest/requirements.txt: remove mentions of python < 3.7.

rouault:
Our baseline since GDAL 3.9 is Python >= 3.8


rouault:
> Do we even need `pytest-sugar` in `requirements.txt` ?

That provides a neater output by default.

PR #12351, opened 2025-05-12
gdaladdo / gdal raster overview add/delete: avoid issues with COG

rouault:
Fixes https://lists.osgeo.org/pipermail/gdal-dev/2025-May/060558.html

### DIFF
--- a/gcore/gdalalgorithm.cpp
+++ b/gcore/gdalalgorithm.cpp
@@ -2193,7 +2193,7 @@ bool GDALAlgorithm::ProcessDatasetArg(GDALAlgorithmArg *arg,
 
         CPLStringList aosOpenOptions;
         CPLStringList aosAllowedDrivers;
-        if (arg->GetName() == GDAL_ARG_NAME_INPUT)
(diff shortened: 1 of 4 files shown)


gcore/gdalalgorithm.cpp
```diff
@@ -2215,6 +2215,20 @@ bool GDALAlgorithm::ProcessDatasetArg(GDALAlgorithmArg *arg,
         if (osDatasetName == "-" && (flags & GDAL_OF_UPDATE) == 0)
             osDatasetName = "/vsistdin/";
 
+        if ((flags & GDAL_OF_UPDATE) != 0 && m_callPath.size() == 4 &&
```
elpaso, line 2218:
```suggestion
        // Handle special case of overview delete in GTiff which would fail 
        // if it is COG without IGNORE_COG_LAYOUT_BREAK
        if ((flags & GDAL_OF_UPDATE) != 0 && m_callPath.size() == 4 &&
```

elpaso (approved):

PR #12366, opened 2025-05-13
VRT: 'transpose' for vrt:// connection

mdsumner:
Implements 'transpose' for the vrt connection options. This allows nomination of a variable by name and the `iXDim` and `iYDim` index positions for passing to `AsClassicDataset`. 

## Further examples 

```bash
"vrt://autotest/gdrivers/data/netcdf/foo_5dimensional.nc?transpose=/temperature:3,4"

"vrt://autotest/gdrivers/data/zarr/array_dimensions.zarr?transpose=/var:1,0"

```

The original motivation was this HDF5, which doesn't have georeferencing in 2D form, but by using multidim cast to classic we get the transform that derives, and can add further augmentation like 'a_srs'. 

 (needs earthdata auth). 

```bash
## export GDAL_HTTP_HEADER_FILE=...
gdalinfo "vrt:///vsicurl/https://gpm1.gesdisc.eosdis.nasa.gov/data/GPM_L3/GPM_3IMERGHHE.07/1998/001/3B-HHR-E.MS.MRG.3IMERG.19980101-S000000-E002959.0000.V07B.HDF5?transpose=//Grid/Intermediate/IRinfluence:1,2"

Driver: VRT/Virtual Raster
Files: none associated
Size is 3600, 1800
Origin = (-179.999996947394237,-89.999996946545821)
Pixel Size = (0.099999998304108,0.099999996607273)
Metadata:
  CodeMissingValue=-9999
  coordinates=time lon lat
  DimensionNames=time,lon,lat
  LongName=
Relative influence of infrared precipitation in the merged
                        microwave-infrared precipitation estimate

Corner Coordinates:
Upper Left  (-179.9999969, -89.9999969)
Lower Left  (-179.9999969,  89.9999969)
Upper Right ( 179.9999969, -89.9999969)
Lower Right ( 179.9999969,  89.9999969)
Center      (   0.0000000,   0.0000000)
Band 1 Block=291x1800 Type=Int16, ColorInterp=Undefined
  NoData Value=-9999
  Metadata:
    DIM_time_INDEX=0
    DIM_time_UNIT=seconds since 1980-01-06 00:00:00 UTC
    DIM_time_VALUE=567648000

```
## Tasklist

 - [x] Add test case(s)
 - [x] Add documentation
 - [ ] Review
 - [ ] Adjust for comments
 - [ ] ADD YOUR TASKS HERE

### DIFF
--- a/doc/source/drivers/raster/vrt.rst
+++ b/doc/source/drivers/raster/vrt.rst
@@ -2178,8 +2178,8 @@ an interface to the function :cpp:func:`GDALMDArray::AsClassicDataset` in the mu
 This can be valuable for reorienting an array that presents X and Y in YX or some other order. (There's a
 possible added advantage that a valid geotransform may be provided that the classic 2D model doesn't yet infer,
 because the multidimensional model can derive one from coordinates referenced in that form).
-The usage syntax is "vrt://somefile.extension?transpose=varname:iXDim,iYDim" with a no-op case
-"vrt://somefile.extension?transpose=varname:0,1" and "vrt://somefile.extension?transpose=varname:1,0" would be a
+The usage syntax is ``vrt://somefile.extension?transpose=varname:iXDim,iYDim`` with a no-op case
+``vrt://somefile.extension?transpose=varname:0,1`` and ``vrt://somefile.extension?transpose=varname:1,0`` would be a
 transpose on the first two axes. There must be two unique axis indexes with values between 0 and the maximum available.
 This option is mutually exclusive with ``sd_name`` and ``sd``.
 
--- a/doc/source/drivers/raster/vrt.rst
+++ b/doc/source/drivers/raster/vrt.rst
@@ -2162,7 +2162,7 @@ The effect of the ``eco`` option (added in GDAL 3.8) is that ``srcwin`` or ``pro
 The effect of the ``sd_name`` option (added in GDAL 3.9) is to choose an individual subdataset by
 name for sources that have multiple subdatasets. This means that rather than a fully-qualified description
 such as "NETCDF:myfile.nc:somearray" we may use "vrt://myfile.nc?sd_name=somearray". This option
-is mutually exclusive with ``sd``.
+is mutually exclusive with ``sd``, and with ``transpose``.
 
 The effect of the ``sd`` option (added in GDAL 3.9) is to choose an individual subdataset by
 number for sources that have multiple subdatasets. This means that rather than a fully-qualified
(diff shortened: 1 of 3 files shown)


frmts/vrt/vrtdataset.cpp
```diff
@@ -1481,6 +1481,58 @@ GDALDataset *VRTDataset::OpenVRTProtocol(const char *pszSpec)
...
+            {
+                CPLError(CE_Failure, CPLE_IllegalArg,
+                         "Invalid transpose option: %s", pszValue);
+                return nullptr;
+            }
+            int index_x = atoi(aosTransposeIndex[0]);
+            int index_y = atoi(aosTransposeIndex[1]);
+
+            auto poMDimDS = std::unique_ptr<GDALDataset>(
+                GDALDataset::Open(osFilename, GDAL_OF_MULTIDIM_RASTER));
```
rouault, line 1509: as we are reopening the source dataset, it means that "vrt:///some_dataset?some_option=some_value&transpose=...." would loose the effect of some_option.  So the evaluation of transpose must be done prior to this loop.

mdsumner: ah, whoops, thanks I will re-work

mdsumner: what option could be lost, though? They are all picked up and saved for Translate ( ? - I'm missing something about lost effects ??)

this works for example

```
gdalinfo "vrt://autotest/gdrivers/data/zarr/array_dimensions.zarr?a_srs=EPSG:3857&a_ullr=0,2,2,-2&transpose=/var:1,0"
```

I agree it's better that the transpose is isolated for good style and future-proofing, I just misunderstand the danger I think

rouault: > what option could be lost, though?

"bands" for example: https://github.com/OSGeo/gdal/blob/16b13502e820ad07f47a668c797c57660a19b440/frmts/vrt/vrtdataset.cpp#L1198

mdsumner: I force pushed a change to put the transpose loop on its own prior to sd_name/sd and added tests for exclusivity with those options. thanks!

doc/source/drivers/raster/vrt.rst
```diff
@@ -2162,16 +2162,26 @@ The effect of the ``eco`` option (added in GDAL 3.8) is that ``srcwin`` or ``pro
...
+This option is mutually exclusive with ``sd_name``, and with ``transpose``.
+
+The effect of the ``transpose`` option (added in GDAL 3.12) is to specify just one array by name from a
+multidimensional dataset and nominate the indexes for the two axes that define the 2D dataset. This is
+an interface to the function :cpp:func:`GDALMDArray::AsClassicDataset` in the multidimensional raster model.
+This can be valuable for reorienting an array that presents X and Y in YX or some other order. (There's a
+possible added advantage that a valid geotransform may be provided that the classic 2D model doesn't yet infer,
+because the multidimensional model can derive one from coordinates referenced in that form).
+The usage syntax is "vrt://somefile.extension?transpose=varname:iXDim,iYDim" with a no-op case
+"vrt://somefile.extension?transpose=varname:0,1" and "vrt://somefile.extension?transpose=varname:1,0" would be a
```
rouault, line 2182:
```suggestion
The usage syntax is ``vrt://somefile.extension?transpose=varname:iXDim,iYDim`` with a no-op case
``vrt://somefile.extension?transpose=varname:0,1`` and ``vrt://somefile.extension?transpose=varname:1,0`` would be a
```

rouault (approved):

doc/source/drivers/raster/vrt.rst
rouault, line 2182:
```suggestion
The usage syntax is ``vrt://somefile.extension?transpose=varname:iXDim,iYDim`` with a no-op case
``vrt://somefile.extension?transpose=varname:0,1`` and ``vrt://somefile.extension?transpose=varname:1,0`` would be a
```

Issue #12300, opened 2025-05-07
Expose GEOS coverage simplification


dbaston:
### Feature description

Add a `gdal vector geom simplify-coverage` (or `gdal vector simplify-coverage` ?) that simplifies shared boundaries of a set of polygons, as described in https://lin-ear-th-inking.blogspot.com/2023/03/simplifying-polygonal-coverages-with-jts.html

### Additional context

Unfortunately this algorithm requires reading all of the input features into memory before doing any processing, so the scaffolding of `GDALVectorGeomAbstractAlgorithm` cannot be readily used. Specifically, there isn't an obvious way to implement a `TranslateFeature` method.

jratike80:
Could it somehow fit as a new option into ogr_layer_algebra https://gdal.org/en/stable/programs/ogr_layer_algebra.html?

rouault:
> Could it somehow fit as a new option into ogr_layer_algebra https://gdal.org/en/stable/programs/ogr_layer_algebra.html?

I'd rather see it as Dan proposed (probably under 'geom') rather than a new mode ogr_layer_algebra, given the later combines 2 layers

PR #12370, opened 2025-05-13
Add gdal vector geom simplify-coverage

dbaston:
Implements coverage simplification as described in #12300.

Some things on my mind:

I implemented this as a pipeline algorithm, although it cannot be performed in a streaming way.

I didn't see a way to use `GDALVectorAbstractAlgorithm` since that ultimately requires one-to-one or one-to-many feature translation. I didn't see a good way to extend it for many-to-many. Since I didn't use `GDALVectorAbstractAlgorithm`, I ended up duplicating bits of its functionality. I'd like to follow this up with a `clean-coverage` algorithm and figured I could generalize some of the new code then.

In the custom `GDALDataset` overload, I did not implement a few methods such as `GetFeature` by ID and dataset-level feature iteration. It's not clear to me if/when these would be called in a pipeline scenario. 

In an earlier version of this implementation, instead of creating a custom `GDALDataset` overload, I created a dataset with the memory driver and wrote the results there. This is much simpler, except that I didn't see a good way to create output layers with the same fields as the inputs. Is there something like `GDALDataset::CopyLayer` that copies only the layer definition, but not the contents?

I interacted with GEOS directly in the algorithm instead of first creating an `OGRGeometryCollection::SimplifyCoverage`. This allows the implementation to be a bit more memory-efficient (for example, clearing inputs as soon as they are not needed) but I can see an argument for containing all GEOS interfacing in one place.

It seems like this algorithm should implicitly perform a `PROMOTE_TO_MULTI` sort of behavior, but I haven't figured out a clean way to do that yet.


apps/gdalalg_vector_geom_simplify_coverage.cpp
```diff
@@ -0,0 +1,352 @@
...
+    }
+
+    OGRFeature *GetNextFeature(OGRLayer **, double *, GDALProgressFunc,
+                               void *) override
+    {
+        // FIXME implement?
+        CPLError(CE_Failure, CPLE_AppDefined,
+                 "GDALDataset::GetNextFeature not implemented.");
+        return nullptr;
+    }
```
rouault, line 88: The base implementation of GDALDataset will do the right thing:

```suggestion
```

apps/gdalalg_vector_geom_simplify_coverage.cpp
```diff
@@ -0,0 +1,352 @@
...
+            finishGEOS_r(m_GEOSContext);
+        }
+    }
+
+    int TestCapability(const char *pszCap) override
+    {
+        if (EQUAL(pszCap, OLCRandomRead))
+        {
+            return true;
+        }
```
rouault, line 123:
```suggestion
```

apps/gdalalg_vector_geom_simplify_coverage.cpp
```diff
@@ -0,0 +1,352 @@
...
+        // FIXME implement?
+        CPLError(CE_Failure, CPLE_AppDefined, "Not implemented.");
+        return nullptr;
+    }
+
+    GIntBig GetFeatureCount(int bForce) override
+    {
+        return OGRLayer::GetFeatureCount(bForce);
+    }
+
```
rouault, line 144:
```suggestion
```

apps/gdalalg_vector_geom_simplify_coverage.h
```diff
@@ -0,0 +1,70 @@
...
+    static constexpr const char *HELP_URL =
+        "/programs/gdal_vector_geom_simplify_coverage.html";
+
+    explicit GDALVectorGeomSimplifyCoverageAlgorithm(
+        bool standaloneStep = false);
+
+    struct Options
+    {
+        double tolerance;
+        bool preserveBoundary;
```
rouault, line 43:
```suggestion
        double tolerance = 0;
        bool preserveBoundary = false;
```

doc/source/programs/gdal_vector_geom_simplify_coverage.rst
```diff
@@ -0,0 +1,91 @@
...
+--------
+
+.. program-output:: gdal vector geom simplify-coverage --help-doc
+
+Description
+-----------
+
+:program:`gdal vector geom simplify-coverage` simplifies shared boundaries of a
+polygonal dataset.
+
```
jratike80, line 25: Maybe the description could be improved. Actually also the outer boundary is simplified if not especially denied with `--preserve-external-boundary`. And what really is preserved is the topology of the coverage so that the shared boundaries remain shared, even they are not the same after simplification.

apps/gdalalg_vector_geom_simplify_coverage.cpp
dbaston: removed

apps/gdalalg_vector_geom_simplify_coverage.cpp
dbaston: removed

apps/gdalalg_vector_geom_simplify_coverage.h
dbaston: done

doc/source/programs/gdal_vector_geom_simplify_coverage.rst
dbaston: ... simplifies boundaries of a polygonal dataset, such that shared boundaries are preserved without introducing gaps or overlaps. Gaps or overlaps already present in the input dataset will not be corrected.

?

jratike80: Feels good.

apps/gdalalg_vector_geom_simplify_coverage.cpp
```diff
@@ -0,0 +1,326 @@
...
+    int TestCapability(const char *) override
+    {
+        // FIXME implement?
+        return false;
+    }
+
+  private:
+    GDALDataset &m_srcDS;
```
rouault, line 77: This member variable isn't used

rouault:
Memory leak reported by ASAN: https://github.com/dbaston/gdal/actions/runs/15022562126/job/42215139934
```
2025-05-14T14:55:42.3812332Z Direct leak of 176 byte(s) in 2 object(s) allocated from:
2025-05-14T14:55:42.3820167Z     #0 0x7feefd9260c1 in operator new(unsigned long) (/usr/lib/llvm-18/lib/clang/18/lib/linux/libclang_rt.asan-x86_64.so+0x1080c1) (BuildId: bf9e3747052c46cd49f7974555b1700dce30b8c0)
2025-05-14T14:55:42.3821320Z     #1 0x7feedb846d01 in GEOSCoverageSimplifyVW_r (/lib/x86_64-linux-gnu/libgeos_c.so.1+0x30d01) (BuildId: 1f19a68dc21881d713aaefcc10347bebb83f1fdf)
2025-05-14T14:55:42.3822432Z     #2 0x7feeefdb81b1 in GDALVectorGeomSimplifyCoverageAlgorithmLayer::Materialize() /home/runner/work/gdal/gdal/apps/gdalalg_vector_geom_simplify_coverage.cpp:220:37
2025-05-14T14:55:42.3823640Z     #3 0x7feeefdb37ce in GDALVectorGeomSimplifyCoverageAlgorithmLayer::GetNextFeature() /home/runner/work/gdal/gdal/apps/gdalalg_vector_geom_simplify_coverage.cpp:129:18
2025-05-14T14:55:42.3825025Z     #4 0x7feeeeb05565 in OGRLayer::FeatureIterator::FeatureIterator(OGRLayer*, bool) /home/runner/work/gdal/gdal/ogr/ogrsf_frmts/generic/ogrlayer.cpp:6053:41
2025-05-14T14:55:42.3825907Z     #5 0x7feeeeac63c0 in OGRLayer::begin() /home/runner/work/gdal/gdal/ogr/ogrsf_frmts/generic/ogrlayer.cpp:6106:12
2025-05-14T14:55:42.3826875Z     #6 0x7feeeeac5e62 in OGRLayer::GetFeatureCount(int) /home/runner/work/gdal/gdal/ogr/ogrsf_frmts/generic/ogrlayer.cpp:165:27
2025-05-14T14:55:42.3827667Z     #7 0x7feeeeac6e77 in OGR_L_GetFeatureCount /home/runner/work/gdal/gdal/ogr/ogrsf_frmts/generic/ogrlayer.cpp:189:42
2025-05-14T14:55:42.3828524Z     #8 0x7feed5eed41a in OGRLayerShadow_GetFeatureCount /home/runner/work/gdal/gdal/build-asan/swig/python/extensions/ogr_wrap.cpp:4674:33
2025-05-14T14:55:42.3829438Z     #9 0x7feed5eed41a in _wrap_Layer_GetFeatureCount /home/runner/work/gdal/gdal/build-asan/swig/python/extensions/ogr_wrap.cpp:10734:46
2025-05-14T14:55:42.3829923Z 
```

apps/gdalalg_vector_geom_simplify_coverage.cpp
rouault: If you keep OLCRandomRead, then you should implement GetFeature(), which you can actually implement by materializing and then returning m_features[idx]

apps/gdalalg_vector_geom_simplify_coverage.cpp
```diff
@@ -0,0 +1,316 @@
...
+
+        if (m_pos >= m_features.size())
+        {
+            return nullptr;
+        }
+
+        auto &ret = m_features[m_pos];
+        m_pos++;
+
+        return ret.release();
```
rouault, line 146: I suspect that given you relese m_features[m_pos], if you do ResetReading() and iterate again, you'll get a null feature. Perhaps you should return ret->Clone() instead ?

To check your implementation of OGRLayer is conformant, you could run test_ogrsf on a .gdalg file. See test_gdalalg_vector_clip_test_ogrsf() for example

apps/gdalalg_vector_geom_simplify_coverage.cpp
```diff
@@ -0,0 +1,316 @@
...
+    {
+        if (!m_materialized)
+        {
+            m_srcLayer.ResetReading();
+            m_features.clear();
+        }
+        m_pos = 0;
+    }
+
+    OGRFeature *GetNextFeature() override
```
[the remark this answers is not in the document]
dbaston: Maybe it's better to go back to just creating a dataset with the MEM driver? Do you see a downside to that?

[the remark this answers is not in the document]
rouault: > Maybe it's better to go back to just creating a dataset with the MEM driver?

that could be more convenient indeed

rouault:
> we don't have that, but you can do something like:
> 
> ```
> for( int i = 0; i < poSrcLayerDefn->GetFieldCount(); ++i )
>     m_poLayerDefn->AddFieldDefn(poSrcLayerDefn->GetFieldDefn(i);
> m_poLayerDefn->SetGeomType(wkbNone);
> for( int i = 0; i < poSrcLayerDefn->GetGeomFieldCount(); ++i )
>     m_poLayerDefn->AddGeomFieldDefn(poSrcLayerDefn->GetGeomFieldDefn(i);
> ```

I was totally forgetting we have OGRFeatureDefn::Clone(), which is actually used in gdalalg_vector_geom_set_type.cpp

dbaston:
> I was totally forgetting we have OGRFeatureDefn::Clone(), which is actually used in gdalalg_vector_geom_set_type.cpp

Unless I'm missing something, it only helps if I'm creating my own `OGRLayer` implementation...I can't use a FeatureDefn to create a Layer.

rouault:
> Unless I'm missing something, it only helps if I'm creating my own `OGRLayer` implementation...I can't use a FeatureDefn to create a Layer.

you could add a OGRMemLayer::OGRMemLayer(const OGRFeatureDefn& ) constructor in frmts/mem/ogrmemlayer.cpp

apps/gdalalg_vector_geom_simplify_coverage.cpp
```diff
@@ -0,0 +1,220 @@
...
+                return false;
+            }
+            poSimplified->assignSpatialReference(
+                dstLayer.GetLayerDefn()->GetGeomFieldDefn(0)->GetSpatialRef());
+            features[i]->SetGeometry(std::move(poSimplified));
+
+            if (dstLayer.CreateFeature(features[i].release()) != CE_None)
+            {
+                return false;
+            }
```
rouault, line 155: Using release() will leak as CreateFeature(), perhaps unexpectingly doesn't take ownership of the passed feature.

```suggestion
            if (dstLayer.CreateFeature(features[i].get()) != CE_None)
            {
                return false;
            }
            features[i].reset();
```

dbaston: Fixed, thanks

dbaston:
`test_gdalalg_vector_geom_simplify_coverage_test_ogrsf` Failing with 

```
E           ERROR: Driver GDALG does not declare GDAL_DCAP_UPDATE
```

rouault:
> `E ERROR: Driver GDALG does not declare GDAL_DCAP_UPDATE`

cf https://github.com/OSGeo/gdal/blob/master/apps/test_ogrsf.cpp#L4434  . You should call OGRMemLayer::SetUpdatable(false) (after you've filled it)

rouault:
you should rebase on top of latest master to make CI happy to take into account the change of signature of RunStep()

apps/gdalalg_vector_pipeline.h
```diff
@@ -219,6 +221,76 @@ class GDALVectorPipelinePassthroughLayer /* non final */
...
+            GetGDALDriverManager()->GetDriverByName("MEM");
+        m_ds.reset(cpl::down_cast<MEMDataset *>(
+            poMemDriver->Create("", 0, 0, 0, GDT_Unknown, nullptr)));
+    }
+
+    virtual bool Process(OGRLayer &srcLayer, OGRLayer &dstLayer) = 0;
+
+    bool AddProcessedLayer(OGRLayer &srcLayer)
+    {
+        OGRMemLayer *poDstLayer = cpl::down_cast<OGRMemLayer *>(
```
rouault, line 243: We could test srcLayer.TestCapability(OLCStringsAsUTF8), and if set pass the ADVERTIZE_UTF8=YES layer creation option

frmts/mem/memdataset.h
```diff
@@ -135,6 +135,9 @@ class CPL_DLL MEMDataset CPL_NON_FINAL : public GDALDataset
 
     OGRLayer *GetLayer(int) override;
 
+    OGRLayer *CreateLayer(const OGRFeatureDefn &oDefn,
```
rouault, line 138: if you return a OGRMemLayer* you could save the cpl::down_cast<> in gdalalg_vector_geom_simplify_coverage.cpp

apps/gdalalg_vector_pipeline.h
dbaston: Done in 20d580e62ed3d90a3655e305f13a8c8d53f92060

frmts/mem/memdataset.h
dbaston: Good idea, thanks

rouault (approved):

PR #12378, opened 2025-05-14
Add 'gdal raster update'

rouault:
This is the equivalent of "gdalwarp in.tif existing.tif".

I considered calling it 'patch' as an alternative. Thoughts ?

### DIFF
--- a/autotest/utilities/test_gdalalg_raster_update.py
+++ b/autotest/utilities/test_gdalalg_raster_update.py
@@ -1,0 +1,156 @@
+#!/usr/bin/env pytest
+# -*- coding: utf-8 -*-
+###############################################################################
+# Project:  GDAL/OGR Test Suite
+# Purpose:  'gdal raster update' testing
+# Author:   Even Rouault <even dot rouault @ spatialys.com>
+#
+###############################################################################
+# Copyright (c) 2025, Even Rouault <even dot rouault at spatialys.com>
+#
+# SPDX-License-Identifier: MIT
+###############################################################################
+
+import gdaltest
+import pytest
+
+from osgeo import gdal
+
+np = pytest.importorskip("numpy")
+gdaltest.importorskip_gdal_array()
+
+
+def test_gdalalg_raster_update_in_mem():
+
+    out_ds = gdal.GetDriverByName("MEM").Create("", 3, 3)
+    out_ds.SetGeoTransform([0, 1, 0, 0, 0, -1])
+
+    src_ds = gdal.GetDriverByName("MEM").Create("", 1, 1)
+    src_ds.SetGeoTransform([1, 1, 0, -1, 0, -1])
+    src_ds.GetRasterBand(1).Fill(1)
+
+    assert gdal.Run("gdal", "raster", "update", input=src_ds, output=out_ds)
+
+    np.testing.assert_array_equal(
+        out_ds.ReadAsArray(), np.array([[0, 0, 0], [0, 1, 0], [0, 0, 0]])
+    )
+
+
+def test_gdalalg_raster_update_gtiff(tmp_vsimem):
+
+    dst_filename = tmp_vsimem / "dst.tif"
+    out_ds = gdal.GetDriverByName("GTIFF").Create(dst_filename, 3, 3)
+    out_ds.SetGeoTransform([0, 1, 0, 0, 0, -1])
+    out_ds.Close()
+
+    src_filename = tmp_vsimem / "src.tif"
+    src_ds = gdal.GetDriverByName("GTIFF").Create(src_filename, 1, 1)
+    src_ds.SetGeoTransform([1, 1, 0, -1, 0, -1])
+    src_ds.GetRasterBand(1).Fill(1)
+    src_ds.Close()
+
+    tab_pct = [0]
+
+    def my_progress(pct, msg, user_data):
+        assert pct >= tab_pct[0]
+        tab_pct[0] = pct
+        return True
+
+    with gdal.Run(
+        "gdal",
+        "raster",
+        "update",
+        input=src_filename,
+        output=dst_filename,
+        progress=my_progress,
+        resampling="cubic",
+        warp_option={"SKIP_NOSOURCE": "YES"},
+        transform_option={"SRC_METHOD": "GEOTRANSFORM"},
+        error_threshold=0,
+    ) as alg:
+        ret_ds = alg.Output()
+
+        np.testing.assert_array_equal(
+            ret_ds.ReadAsArray(), np.array([[0, 0, 0], [0, 1, 0], [0, 0, 0]])
+        )
+
+    assert tab_pct[0] == 1.0
+
+
+def test_gdalalg_raster_update_src_is_same_as_dst_pointer():
+
+    ds = gdal.GetDriverByName("MEM").Create("", 3, 3)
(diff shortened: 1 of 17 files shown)


alonfaraj:
Maybe apply or paste?

jratike80:
Overlay?

rouault:
> Are the overviews of the target updated as well?

no, comment added in doc and link to gdal raster overview refresh

> Using cutline is not supported?

Good point, just added the --geometry / --geometry-crs options of gdal raster clip

doc/source/programs/gdal_raster_update.rst
```diff
@@ -0,0 +1,122 @@
...
+Description
+-----------
+
+:program:`gdal raster update` can be used to update an existing target raster dataset
+with the pixel values of the source raster, doing reprojection if the source
+and target datasets do not have the same coordinate reference systems.
+The extent, size, resolution or coordinate reference system of the target dataset
+are not modified by this operation.
+
+Overviews are *not* updated by this command, but may be if using :ref:`gdal_raster_overview_refresh`.
```
dbaston, line 29: Any thoughts on inverting this behavior? Update any existing overviews unless `--no-update-overviews` is specified?

doc/source/programs/gdal_raster_update.rst
```diff
@@ -0,0 +1,122 @@
...
+
+Standard options
+++++++++++++++++
+
+.. option:: -r, --resampling <RESAMPLING>
```
dbaston, line 34: Worth pulling this out into a separate file that can be included?

doc/source/programs/gdal_raster_update.rst
```diff
@@ -0,0 +1,122 @@
...
+    If the input geometry is GeoJSON, its CRS is assumed to be WGS84, unless there is
+    a CRS defined in the GeoJSON geometry or :option:`--geometry-crs` is specified.
+    If the input geometry is WKT, its CRS is assumed to be the one of the input dataset,
+    unless :option:`--geometry-crs` is specified.
+    The X and Y axis are the "GIS friendly ones", that is X is longitude or easting,
+    and Y is latitude or northing.
+
+.. option:: --geometry-crs <CRS>
+
+    CRS in which the coordinates values of :option:`--geometry`
```
dbaston, line 88:
```suggestion
    CRS in which the coordinate values of :option:`--geometry`
```

doc/source/programs/gdal_raster_update.rst
```diff
@@ -0,0 +1,122 @@
...
+    are expressed. If not specified, it is assumed to be the CRS of the input
+    dataset.
+    The bounds are reprojected from the geometry-crs to the CRS of the
+    input dataset.
```
dbaston, line 92:
```suggestion
    The geometry will be reprojected from the geometry-crs to the CRS of the
    dataset being updated.
```

?

doc/source/programs/gdal_raster_update.rst
```diff
@@ -0,0 +1,122 @@
...
+
+    Set a warp option.  The :cpp:member:`GDALWarpOptions::papszWarpOptions` docs show all options.
+    Multiple options may be listed.
+
+.. option:: -to <NAME>=<VALUE>
+
+    Set a transformer option suitable to pass to :cpp:func:`GDALCreateGenImgProjTransformer2`.
+    See :cpp:func:`GDALCreateRPCTransformerV2()` for RPC specific options.
+
+.. option:: --et, --error-threshold <ERROR-THRESHOLD>
```
dbaston, line 107: Beyond the scope of this PR, but I'm wondering it's worth reconsidering the default value for `--error-threshold`. We haven't found a huge performance boost from the default value, and have had some surprising errors introduced with non-zero values.

autotest/utilities/test_gdalalg_raster_update.py
```diff
@@ -0,0 +1,145 @@
...
+    out_ds = gdal.GetDriverByName("MEM").Create("", 3, 3)
+    out_ds.SetGeoTransform([0, 1, 0, 0, 0, -1])
+
+    src_ds = gdal.GetDriverByName("MEM").Create("", 1, 1)
+    src_ds.SetGeoTransform([1, 1, 0, -1, 0, -1])
+    src_ds.GetRasterBand(1).Fill(1)
+
+    assert gdal.Run("gdal", "raster", "update", input=src_ds, output=out_ds)
+
+    assert out_ds.ReadRaster() == b"\x00\x00\x00\x00\x01\x00\x00\x00\x00"
```
dbaston, line 30: Can we use numpy for readability?

dbaston:
> I considered calling it 'patch' as an alternative. Thoughts ?

I like "update"

doc/source/programs/gdal_raster_update.rst
rouault: done

doc/source/programs/gdal_raster_update.rst
rouault: fixd

doc/source/programs/gdal_raster_update.rst
rouault: fixed

doc/source/programs/gdal_raster_update.rst
rouault: > and have had some surprising errors introduced with non-zero values.

I'd be curious about the exact situation where that would happen

dbaston: I'll see what I can dig up. Maybe the issue is the assumption that the error is greatest at the midpoint between the points that are actually transformed. Would it make sense that that is not always true for certain transformations?

rouault: > Would it make sense that that is not always true for certain transformations?

I can imagine that could happen for polar projections <--> geographic

doc/source/programs/gdal_raster_update.rst
```diff
@@ -0,0 +1,122 @@
...
+
+Advanced options
+++++++++++++++++
+
+.. option:: --wo, --warp-option <NAME>=<VALUE>
+
+    Set a warp option.  The :cpp:member:`GDALWarpOptions::papszWarpOptions` docs show all options.
+    Multiple options may be listed.
+
+.. option:: -to <NAME>=<VALUE>
```
jratike80, line 102: Should it be `--to`?

doc/source/programs/gdal_raster_update.rst
rouault: > Should GDAL try to be smart and disable the approximator for such cases?

We have already some exceptions/tuning for such cases in various places (look for "pole" in gdaltransformer.cpp, gdalwarper.cpp, gdalwarpoperation.cpp). It is hard to do in the general case however
The approximate transformer helps substantially for UTM <--> geographic since the accurate Transverse Mercator we use is not cheap, especially if using nearest neighbour where the limiting factor becomes PROJ

doc/source/programs/gdal_raster_update.rst
rouault: > Update any existing overviews unless `--no-update-overviews` is specified?

done

mdsumner:
> > I considered calling it 'patch' as an alternative. Thoughts ?
> 
> I like "update"

me too (patch sounds like "raster patches")

PR #12383, opened 2025-05-15
gdal raster reproject: add a -j/--num-threads option and default to ALL_CPUS

rouault:

PR #12384, opened 2025-05-15
gdal raster reproject: fix performance issue when outputing to COG

rouault:
We recover performance by allowing a step to do the job of the following step, if it declares to be able to do so. This enables us to replace warp to VRT + translate  with warp to final format, which is very beneficial for COG output given that there is a co-operation between the COG driver and the warping code.

With this PR:
```
$ time gdal raster pipeline read byte.tif ! reproject --size 10000,10000 -r cubic --t_srs EPSG:4326 -j 0 ! write out.tif --overwrite --of=cog --co compress=zstd

real    0m5,748s
user    0m5,595s
sys     0m0,152s

$ time gdalwarp byte.tif out.tif -ts 10000 10000 -r cubic -wo NUM_THREADS=0 -co NUM_THREADS=0 -of COG -co COMPRESS=ZSTD -overwrite

real    0m5,280s
user    0m5,110s
sys     0m0,168s

$ time gdal raster pipeline read byte.tif ! reproject --size 10000,10000 -r cubic --t_srs EPSG:4326 -j 8 ! write out.tif --overwrite --of=cog --co compress=zstd

real    0m2,599s
user    0m11,262s
sys     0m0,527s

$ time gdal raster reproject --size 10000,10000 -r cubic --t_srs EPSG:4326 -j 8 --overwrite --of=cog --co compress=zstd byte.tif out.tif

real	0m2,594s
user	0m11,633s
sys	0m0,464s

$ time gdalwarp byte.tif out.tif -ts 10000 10000 -r cubic -wo NUM_THREADS=8 -co NUM_THREADS=8 -of COG -co COMPRESS=ZSTD -overwrite
Creating output file that is 10000P x 10000L.

real    0m2,258s
user    0m10,186s
sys     0m0,344s
```

Compared to master + https://github.com/OSGeo/gdal/pull/12383:
```
$ time gdal raster pipeline read byte.tif ! reproject --size 10000,10000 -r cubic --t_srs EPSG:4326 -j 0 ! write out.tif --overwrite --of=cog --co compress=zstd

real    0m8,631s
user    0m8,558s
sys     0m0,072s

$ time gdal raster pipeline read byte.tif ! reproject --size 10000,10000 -r cubic --t_srs EPSG:4326 -j 8 ! write out.tif --overwrite --of=cog --co compress=zstd --co num_threads=8

real    0m6,256s
user    0m17,601s
sys     0m0,388s
```

### DIFF
--- a/apps/gdalalg_raster_pipeline.h
+++ b/apps/gdalalg_raster_pipeline.h
@@ -29,9 +29,15 @@ class GDALRasterPipelineStepRunContext
   public:
     GDALRasterPipelineStepRunContext() = default;
 
+    // Progress callback to use during execution of the step
     GDALProgressFunc m_pfnProgress = nullptr;
     void *m_pProgressData = nullptr;
-    GDALRasterPipelineStepAlgorithm *m_poNextStep = nullptr;
+
+    // If there is a step in the pipeline immediately following step to which
+    // this instance of GDALRasterPipelineStepRunContext is passed, and that
+    // this next step is usable by the current step (as determined by
+    // CanHandleNextStep()), then this member will point to this next step.
+    GDALRasterPipelineStepAlgorithm *m_poNextUsableStep = nullptr;
 
   private:
     CPL_DISALLOW_COPY_ASSIGN(GDALRasterPipelineStepRunContext)
--- a/apps/gdalalg_raster_pipeline.h
+++ b/apps/gdalalg_raster_pipeline.h
@@ -22,6 +22,8 @@
 /*                  GDALRasterPipelineStepRunContext                    */
 /************************************************************************/
 
+class GDALRasterPipelineStepAlgorithm;
+
 class GDALRasterPipelineStepRunContext
 {
   public:
@@ -29,6 +31,7 @@ class GDALRasterPipelineStepRunContext
 
     GDALProgressFunc m_pfnProgress = nullptr;
     void *m_pProgressData = nullptr;
+    GDALRasterPipelineStepAlgorithm *m_poNextStep = nullptr;
 
   private:
     CPL_DISALLOW_COPY_ASSIGN(GDALRasterPipelineStepRunContext)
@@ -40,6 +43,22 @@ class GDALRasterPipelineStepRunContext
 
 class GDALRasterPipelineStepAlgorithm /* non final */ : public GDALAlgorithm
 {
+  public:
+    const GDALArgDatasetValue &GetOutputDataset() const
+    {
+        return m_outputDataset;
+    }
+
+    const std::string &GetOutputFormat() const
+    {
+        return m_format;
+    }
+
+    const std::vector<std::string> &GetCreationOptions() const
+    {
+        return m_creationOptions;
+    }
+
   protected:
     using StepRunContext = GDALRasterPipelineStepRunContext;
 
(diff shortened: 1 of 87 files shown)


apps/gdalalg_raster_pipeline.h
```diff
@@ -18,13 +18,50 @@
...
 
 class GDALRasterPipelineStepAlgorithm /* non final */ : public GDALAlgorithm
 {
+  public:
+    const GDALArgDatasetValue &GetOutputDataset() const
+    {
+        return m_outputDataset;
+    }
+
+    const std::string &GetFormat() const
```
elpaso, line 52:
```suggestion
    const std::string &GetOutputFormat() const
```
just to clearly distinguish it from the input format.

apps/gdalalg_vector_pipeline.cpp
```diff
@@ -168,9 +169,29 @@ bool GDALVectorPipelineStepAlgorithm::RunImpl(GDALProgressFunc pfnProgress,
...
+                pScaledData.reset(GDALCreateScaledProgress(
+                    0.0, bIsStreaming || bCanHandleNextStep ? 1.0 : 0.5,
+                    pfnProgress, pProgressData));
+            }
+
+            GDALVectorPipelineStepRunContext stepCtxt;
+            stepCtxt.m_pfnProgress = pScaledData ? GDALScaledProgress : nullptr;
+            stepCtxt.m_pProgressData = pScaledData.get();
+            if (bCanHandleNextStep)
+                stepCtxt.m_poNextStep = &writeAlg;
```
elpaso, line 191: I see the idea here is to set m_poNextStep only if it can be handled, wouldn't be clearer to always set it and explicitly mark as supported in the context, something like stepCtxt.m_bCanHandleNextStep = true?

apps/gdalalg_raster_reproject.cpp
```diff
@@ -132,21 +132,71 @@ GDALRasterReprojectAlgorithm::GDALRasterReprojectAlgorithm(bool standaloneStep)
...
 {
     CPLAssert(m_inputDataset.GetDatasetRef());
     CPLAssert(m_outputDataset.GetName().empty());
     CPLAssert(!m_outputDataset.GetDatasetRef());
 
     CPLStringList aosOptions;
-    aosOptions.AddString("-of");
-    aosOptions.AddString("VRT");
+    std::string outputFilename;
+    if (ctxt.m_poNextStep)
```
elpaso, line 163: see comment below about conveying the "can be handled" meaning within the pointer to the next step.

elpaso (approved):

gcore/gdalalgorithm.cpp
```diff
@@ -1453,7 +1453,15 @@ GDALAlgorithm::GDALAlgorithm(const std::string &name,
     AddArg("config", 0, _("Configuration option"), &m_dummyConfigOptions)
         .SetMetaVar("<KEY>=<VALUE>")
         .SetOnlyForCLI()
-        .SetCategory(GAAC_COMMON);
+        .SetCategory(GAAC_COMMON)
```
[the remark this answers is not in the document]
rouault: no. There was a subtle difference due to GDALRasterPipellineAlgorithm::RunStep() calling RunStep() on its steps rather than Run() that required this change

apps/gdalalg_vector_pipeline.cpp
rouault: That seems more complicated to me. It would be little useful for a step to know there is a next step that it cannot handle

elpaso: It's about clarity and readability: if someone reads `if (ctxt.m_poNextStep)` she may think: if it's null the current step must be the last one, right?  Wrong, it's not, it's just that m_poNextStep is not just carrying the information about the next step but what it really also means is that the current step can handle it.

Anyway, not a big issue.

rouault: maybe rename it as ``m_poNextUsableStep`` ?

elpaso: > maybe rename it as `m_poNextUsableStep` ?

Sounds perfect!

Issue #11952, opened 2025-03-12
Allow `gdalbuildvrt` to accept predefined pixel functions


ar-siddiqui:
### Feature description

One use case of gdalbuildvrt is to mosaic overlapping rasters with the pixel value for overlapped pixels derived from pixel functions. For example, this is useful for flood maps, where you want the highest flood depth to take priority whenever there are overlapping rasters.

There are some predefined pixel functions available https://gdal.org/en/stable/drivers/raster/vrt.html#default-pixel-functions (including the `max` which I need), but currently, the only way to use them is to manipulate the created VRT manually. 

It would be nice to have a CLI flag that can take predefined functions and add the pixel function in VRT.

### Additional context

From my research, I couldn't find this functionality in other gdal tools such as `gdal_merge` or `gdaltindex`. It would be nice to have this functionality in these tools as well.

dbaston:
@rouault A challenge I'm finding with this is a pixel function has no good way (that I can see) to determine which sources  actually overlap a given pixel. We have the `SkipNonContributingSources` tag that prevents a source from participating if it doesn't intersect the window we're trying to read, but we still get zero values passed to the pixel function if a source _partially_ intersects the window we're reading.

rouault:
> but we still get zero values passed to the pixel function if a source _partially_ intersects the window we're reading.

we should probably initialize to the nodata value, but that would mean that all pixel functions should be aware of it (typically 0 is a good value for add)

dbaston:
> we should probably initialize to the nodata value

What if we don't have a NoData value? I was thinking of a non-standard NaN (this is how R handles `NA`, its version of NoData), but that only works if `SourceTransferType` is floating point. To handle all data types, wouldn't the pixel function need to be provided with a mask?

dbaston:
> What if we don't have a NoData value?

I think we can just issue a warning. Providing the pixel functions with a mask seems too disruptive, and creating a VRT with multiple sources and no way to represent "no source" seems like it should be discouraged anyway.

@rouault I will make the built-in pixel functions NoData-aware, and add the `propagateNoData` argument (used for min and max pixel functions) to give some control over how NoData should be handled.

PR #12385, opened 2025-05-15
BuildVRT: Support pixel functions

dbaston:
## What does this PR do?

Adds a `-pixel-function` argument to `gdalbuildvrt` and `gdal raster mosaic` to specify how overlapping inputs should be handled.

I don't think the name "pixel function" is especially descriptive, but it's been around for a while so people seem to know what it means. (Like VSI?)

Keeping this as draft for now because I want to figure out a better solution for non-contributing sources. The failing "min" test is a good illustration of the problem.

#11952

## Tasklist

 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/apps/gdalbuildvrt_lib.cpp
+++ b/apps/gdalbuildvrt_lib.cpp
@@ -2507,6 +2507,13 @@ GDALBuildVRTOptionsNew(char **papszArgv,
             psOptions->ymax = (*oTE)[3];
         }
 
+        if (psOptions->osPixelFunction.empty() &&
+            !psOptions->aosPixelFunctionArgs.empty())
+        {
+            throw std::runtime_error(
+                "Pixel function arguments provided without a pixel function");
+        }
+
         return psOptions.release();
     }
     catch (const std::exception &err)
--- a/apps/gdalbuildvrt_lib.cpp
+++ b/apps/gdalbuildvrt_lib.cpp
@@ -248,6 +248,8 @@ class VRTBuilder
     bool bNoDataFromMask = false;
     double dfMaskValueThreshold = 0;
     const CPLStringList aosCreateOptions;
+    std::string osPixelFunction{};
+    const CPLStringList aosPixelFunctionArgs;
     const bool bWriteAbsolutePath;
 
     /* Internal variables */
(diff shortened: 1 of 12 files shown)


apps/gdalbuildvrt_lib.cpp
```diff
@@ -2151,10 +2177,25 @@ GDALBuildVRTOptionsGetParser(GDALBuildVRTOptions *psOptions,
...
+            .flag()
+            .store_into(psOptions->bSeparate)
+            .help(_("Place each input file into a separate band."));
+
+        group.add_argument("-pixel-function")
+            .metavar("<function>")
+            .action(
+                [psOptions](const std::string &s)
+                {
+                    // FIXME make sure it's a real function?
```
dbaston, line 2193: I think we should validate the name. The only reason not to - a pixel function that is available in the VRT runtime env but not the VRT creation env - seems too niche to worry about.

doc/source/programs/gdalbuildvrt.rst
```diff
@@ -191,12 +192,22 @@ changed in later versions.
...
     VRT bands, unless :option:`-b` is specified to select a subset of them.
     Before GDAL 3.8, only the first band of each input file was placed into a
     new VRT band, and :option:`-b` was ignored.
 
+.. option:: -pixel-function
+
+    Specify a function name to calculate a value from overlapping inputs.
+    For a list of available pixel functions, see :ref:`builtin_pixel_functions`.
+    If no function is specified, values will be taken from the last overlapping input.
+    This option is mutually exvlusive with with :option:`-separate`.
```
rouault, line 207:
```suggestion
    This option is mutually exclusive with with :option:`-separate`.
```

frmts/vrt/vrtderivedrasterband.cpp
```diff
@@ -400,6 +417,24 @@ void VRTDerivedRasterBand::SetPixelFunctionLanguage(const char *pszLanguage)
...
 
+/************************************************************************/
+/*                 SetSkipNonContributingSources()                      */
+/************************************************************************/
+
+/** Whether sources that do not intersect the VRTRasterBand RasterIO() requested
+ * region should be omitted. By default, data for all sources, including ones
+ * that do not intersect it, are passed to the pixel function. By setting this
+ * parameter to false, only sources that intersect the requested region will be
+ * passed.
```
jratike80, line 428: Isn't false the default, "please don't skip"?

rouault (approved):

dbaston:
Still need to add docs for `--pixel-function-arg` and make it throw an error when called without `--pixel-function`.

dbaston:
Oops, looks like `gdal raster mosaic` is missing the `--pixel-function-arg` implementation.

PR #12386, opened 2025-05-15
Document 2 more Multi-dimensional array API Python Functions

geographika:
I added in titles to allow an automatic TOC for the page as it was easy to become lost on the page, and this allows users to quickly jump to a class. 

One question on the SWIG Python bindings - nearly all functions in `gdal.py` have the arguments `(self, *args)`, this is unlike the MapServer bindings which have arguments that mirror their C functions, e.g. `(self, postData, url)`. 

With the example of `CreateDimension` the function is fully documented in `gdalmultidim.cpp`:

```c++
std::shared_ptr<GDALDimension> GDALGroup::CreateDimension(
    CPL_UNUSED const std::string &osName, CPL_UNUSED const std::string &osType,
    CPL_UNUSED const std::string &osDirection, CPL_UNUSED GUInt64 nSize,
    CPL_UNUSED CSLConstList papszOptions)
{
    CPLError(CE_Failure, CPLE_NotSupported,
             "CreateDimension() not implemented");
    return nullptr;
}
```

Then each of the drivers that supports MultiDimensions has its own implementation, e.g. in `memdataset.cpp`. 
Then in the SWIG bindings interface file `MultiDimensional.i` we create a wrapper function:

```c++
%newobject CreateDimension;
  GDALDimensionHS *CreateDimension( const char *name,
                                    const char* type,
                                    const char* direction,
                                    GUIntBig size,
                                    char **options = 0 ) {
    return GDALGroupCreateDimension(self, name, type, direction, size, options);
  }
```

I would have expected SWIG to generate a Python function similar to:

```python
def CreateDimension(name: str, type: str, direction: str, size: int, options=None) -> 'GDALDimension':
    # Python types would be a bonus, but even named parameters would be useful
```

However, in `gdal.py` it outputs:

```python
    def CreateDimension(self, *args):
        r"""
        CreateDimension(Group self, char const * name, char const * type, char const * direction, GUIntBig size, char ** options=None) -> Dimension
```

One final question. The `MultiDimensional.i` file uses the parameter name `const char* type`, however in Python `type` is a reserved word, so I used `dim_type`. Should these parameter names always match? Does it make any difference to the Python bindings what they are called (apart from for clarity in the help/docstrings)?

### DIFF
--- a/swig/include/python/docs/gdal_mdm_docs.i
+++ b/swig/include/python/docs/gdal_mdm_docs.i
@@ -1,3 +1,71 @@
+%feature("docstring")  CreateAttribute "
+
+Create an attribute within a :py:class:`osgeo.gdal.MDArray` or :py:class:`osgeo.gdal.Group`.
+
+See :cpp:func:`GDALIHasAttribute::CreateAttribute`.
+
+Parameters
+----------
+name : str
+    name
+dimensions : list
+    List of dimensions, ordered from the slowest varying
+    dimension first to the fastest varying dimension last.
+    Might be empty for a scalar array (if supported by driver)
+data_type: :py:class:`osgeo.gdal.ExtendedDataType`
+    Attribute data type
+options: dict/list
+    an optional dict or list of driver specific ``NAME=VALUE`` option strings.
+
+Returns
+-------
+
+Attribute:
+    the new :py:class:`osgeo.gdal.Attribute` or ``None`` on failure.
+
+Examples
+--------
+>>> numeric_attr = ar.CreateAttribute('numeric_attr', [], gdal.ExtendedDataType.Create(gdal.GDT_Float64))
+>>> string_attr = ar.CreateAttribute('string_attr', [], gdal.ExtendedDataType.CreateString())
+
+";
+
+%feature("docstring")  CreateDimension "
+
+Create a dimension within a :py:class:`osgeo.gdal.Group`.
+
+See :cpp:func:`GDALGroup::CreateDimension`.
+
+Parameters
+----------
+name : str
+    Dimension name
+dim_type : str
+    Dimension type (might be empty, and ignored by drivers)
+direction: str
+    Dimension direction (might be empty, and ignored by drivers)
+size : int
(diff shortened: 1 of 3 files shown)


rouault:
> One question on the SWIG Python bindings - nearly all functions in `gdal.py` have the arguments `(self, *args)`, this is unlike the MapServer bindings which have arguments that mirror their C functions, e.g. `(self, postData, url)`.

Seieng https://github.com/swig/swig/blob/c39c1401ccf2d87f163459847a96925608db0166/Source/Modules/python.cxx#L2291 , I bet this is due to https://github.com/OSGeo/gdal/blob/ff1d8eb143e276163dc957e1db400638484a48f7/swig/include/gdal.i#L29

swig/include/python/docs/gdal_mdm_docs.i
```diff
@@ -1,3 +1,71 @@
...
+
+Returns
+-------
+
+Attribute:
+    the new :py:class:`osgeo.gdal.Attribute` or ``None`` on failure.
+
+Examples
+--------
+>>> numeric_attr = ar.CreateAttribute('numeric_attr', [], gdal.ExtendedDataType.Create(gdal.GDT_Float64))
```
dbaston, line 28: I think it would be good for the examples to be runnable from scratch (except for module imports) and this would be a requirement in any case for #9949

dbaston:
Does putting this near the top cause the linked class names to be shortened?

```
.. py:currentmodule:: osgeo.gdal
```

Compare 
image

with 

image

dbaston:
The Python argument names _should_ matter, but for some functions they don't because`%feature("kwargs")` isn't set. Should we set `kwargs` and then rename parameters as needed to make them Python-friendly?

dbaston:
@geographika I was suggesting a more limited change so that the functions you're documenting here have signatures with `*args` and `**kwargs` like most of the other functions. (Committed that to your branch in bd562b189cf2d63e76a89bfcf70928a7a1ff47b1). Switching to have the actual argument names in the Python signatures seems like a good idea, but probably something that should be done project-wide.

PR #12392, opened 2025-05-16
GDALOverviews: Limit external file size in GDALRegenerateOverviewsMultiBand

lucianpls:
## What does this PR do?
In GDALRegenerateOverviewsMultiBand() (ROMB), when too much memory would be needed for generating an overview block, the overviews are generated in a temporary file and then copied to the actual overview dataset. The size of this temporary file is not limited, and it has extra problems with pixel interleaved output, since the bands are copied one at a time for the full overview.
This issue was reported in issue #12303 and was partially addressed in #12307, which raised the allowable memory limit which triggers the problem code path, but did not limit the size of the temporary file or the pixel interleaved problems.

This PR addresses the remaining problems by chunking the output overview when possible, and using the temporary file path for a single chunk at a time.

#12303, #12307

## Test case
To trigger the code path in question, set GDAL_OVER_CHUNK_MAX_SIZE_FOR_TEMP_FILE to 10485760 and build power of 2 overviews for a 11000 by 11000 8bit, pixel interleaved RGB MRF with a blocksize of 1024

```
gdal_translate -outsize 11000 11000  -co BLOCKSIZE=1024 any_small_rgb.jpg test_file.mrf
gdaladdo -r average test_file.mrf
```

## Changes

The "chunking" code triggers when a temporary file would be used, but an output chunk (tile) smaller than the whole overview does exist. The chunk would still exceed the memory usage limit. It then calls ROMB() for every chunk, with extra parameters that limit the processing extent.
On the next pass through ROMB, the external file will be created, this time limited to the chunk extent. This part was also modified, using a VRT to effect the scaling, with the correct resampling, before writing the chunk data in the external file.
Finally, the chunk data is transferred from the temporary output to the destination overview.

In addtion, an extra flag had to be added to the RasterIO ExtraArguments structure, to indicate that the source overviews should not be used for the respective operation, to avoid having the input data read from the same overview that is being generated.

## Tasklist

 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments
 - [ ] ADD YOUR TASKS HERE

## Environment

* OS:
* Compiler:

### DIFF
--- a/gcore/gdal.h
+++ b/gcore/gdal.h
@@ -213,7 +213,7 @@ typedef struct
         (s).pfnProgress = CPL_NULLPTR;                                         \
         (s).pProgressData = CPL_NULLPTR;                                       \
         (s).bFloatingPointWindowValidity = FALSE;                              \
-        (s).bUseOnlyThisScale = FALSE;                                            \
+        (s).bUseOnlyThisScale = FALSE;                                         \
     } while (0)
 
 /** Value indicating the start of the range for color interpretations belonging
--- a/gcore/gdal.h
+++ b/gcore/gdal.h
@@ -191,8 +191,11 @@ typedef struct
     /*! Height in pixels of the area of interest. Only valid if
      * bFloatingPointWindowValidity = TRUE */
     double dfYSize;
-    /*! Do not use overviews even if it would be possible */
-    int bDoNotUseOverviews;
+    /*! Indicate if overviews should be considered. Tested in
+        GDALBandGetBestOverviewLevel(), mostly reserved for use by
+        GDALRegenerateOverviewsMultiBand()
+    */
+    int bUseOnlyThisScale;
 } GDALRasterIOExtraArg;
 
 #ifndef DOXYGEN_SKIP
@@ -210,7 +213,7 @@ typedef struct
         (s).pfnProgress = CPL_NULLPTR;                                         \
         (s).pProgressData = CPL_NULLPTR;                                       \
         (s).bFloatingPointWindowValidity = FALSE;                              \
-        (s).bDoNotUseOverviews = FALSE;                                        \
+        (s).bUseOnlyThisScale = FALSE;                                            \
     } while (0)
 
 /** Value indicating the start of the range for color interpretations belonging
--- a/gcore/gdal.h
+++ b/gcore/gdal.h
@@ -191,6 +191,8 @@ typedef struct
     /*! Height in pixels of the area of interest. Only valid if
      * bFloatingPointWindowValidity = TRUE */
     double dfYSize;
+    /*! Do not use overviews even if it would be possible */
+    int bDoNotUseOverviews;
 } GDALRasterIOExtraArg;
 
 #ifndef DOXYGEN_SKIP
@@ -208,6 +210,7 @@ typedef struct
         (s).pfnProgress = CPL_NULLPTR;                                         \
         (s).pProgressData = CPL_NULLPTR;                                       \
         (s).bFloatingPointWindowValidity = FALSE;                              \
+        (s).bDoNotUseOverviews = FALSE;                                        \
     } while (0)
 
 /** Value indicating the start of the range for color interpretations belonging
--- a/autotest/gcore/tiff_ovr.py
+++ b/autotest/gcore/tiff_ovr.py
@@ -2719,7 +2719,7 @@ def test_tiff_ovr_fallback_to_multiband_overview_generate(config_options):
 
     ds = gdal.Open(filename)
     cs = ds.GetRasterBand(1).GetOverview(0).Checksum()
-    assert cs == 37308
+    assert cs == 36766
     ds = None
 
     gdal.GetDriverByName("GTiff").Delete(filename)
--- a/autotest/gcore/tiff_ovr.py
+++ b/autotest/gcore/tiff_ovr.py
@@ -3069,3 +3069,56 @@ def test_tiff_ovr_huge_raster_with_ovr_huge_block(tmp_vsimem):
     with pytest.raises(Exception):
         with gdal.Open(tmp_vsimem / "tmp.tif", gdal.GA_Update) as ds:
             ds.BuildOverviews("AVERAGE", [2])
+
+
+###############################################################################
+
+
+@gdaltest.enable_exceptions()
+def test_tiff_ovr_huge_reduction_factor_nodata(tmp_vsimem):
+
+    ds = gdal.GetDriverByName("GTIFF").Create(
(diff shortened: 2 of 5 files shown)


rouault:
@lucianpls Can you rebase on top of latest master ? There is a conflict preventing CI from running

lucianpls:
> @lucianpls Can you rebase on top of latest master ? There is a conflict preventing CI from running

Yes, I'll work on it tomorrow, I didn't expect it to change again this fast.

rouault:
The conflict is with 363a7ee4e18c217a247b705897c2f019fe686e6c

gcore/gdal.h
```diff
@@ -191,6 +191,8 @@ typedef struct
     /*! Height in pixels of the area of interest. Only valid if
      * bFloatingPointWindowValidity = TRUE */
     double dfYSize;
+    /*! Do not use overviews even if it would be possible */
+    int bDoNotUseOverviews;
```
rouault, line 195: The addition of this member is a bit problematic. There would be a lot of drivers that would need to be modified to honour it. That would require a review of all IRasterIO() implementations

lucianpls: The problem is that the scaling VRT ends up calling RasterIO on the source band, which in the MRF use case finds the overview and reads from it instead of the intended source band.
I tried to use a RasterBandProxy to stub the GetOverviewCount(), it still calls the proxied band RasterIO.
Do you have any suggestion?

I can avoid this issue for the MRF driver by not reporting the overviews while (re)building them. There aren't many drivers that call the ROMB function, it should be possible to do something similar for them. It would still be a potential pitfall for new drivers, but a lesser one than the one I am trying to fix.

rouault: perhaps add a VRTSimpleSource::SetUseOverviews() method to set a boolean that you would call when creating your VRT ?

lucianpls: The VRT calls IRasterIO to do the scaling on the source band, which does find the overviews. I don't see how a VRT flag could avoid that, short of doing it's own scaling like the ROMB does.

lucianpls: Re: bDoNotUseOverviews, would it be more acceptable if we rename the symbol somehow, to denote that it's only to be set and used by the ROMB function? The default behavior is still the same as before.

rouault: > The VRT calls IRasterIO to do the scaling on the source band, which does find the overviews. I don't see how a VRT flag could avoid that, short of doing it's own scaling like the ROMB does.

ah sorry I thought this was all in the VRT driver. I now see the flag is tested in GDALRasterBand::IRasterIO().
Another solution that would avoid exposing bDoNotUseOverviews to user API would be to set/unset a thread-local boolean.

If we need to keep bDoNotUseOverviews (perhaps to be renamed bUseOverviews to avoid a negation that is always confusing)  , then we need to:
- make it clear in its documentation where it is honored
- update RASTERIO_EXTRA_ARG_CURRENT_VERSION to 2, and only test bDoNotUseOverviews if GDALRasterIOExtraArg::nVersion >= 2 (in case users wouldn't use INIT_RASTERIO_EXTRA_ARG())

lucianpls: The negation is to make zero the default value, I find it safer that way. Possibly due to my electronics background.

rouault: > I find it safer that way

not really. If people don't initialize, they will get a random value from the stack...

lucianpls: I mean that one can use memset the size of struct to zero, leaving everything at default. That code does not require changes when the size of struct changes, and a cleaner INIT_RASTERIO_EXTRA_ARG for example.

lucianpls:
It is missing the NoData and Mask handling, not sure I fully understand that part yet.

lucianpls:
@rouault 
Thinking about this, if we get the VRT with a smaller block to do the scaling, the temporary dataset is not needed at all. 
This assumes that the VRT with a small block can complete a RasterIO to the output raster block without using an excessive amount of memory, which I believe to be true.

rouault:
``autotest/cpp/gdal_unit_test --gtest_filter=test_gdal.GDALRegenerateOverviewsMultiBand_very_large_block_size`` fails with `` C++ exception with description "std::bad_alloc" thrown in the test body.``

rouault:
> Thinking about this, if we get the VRT with a smaller block to do the scaling, the temporary dataset is not needed at all.
> This assumes that the VRT with a small block can complete a RasterIO to the output raster block without using an excessive amount of memory, which I believe to be true.

hum, I'm not able to anticipate. Best is to try e.g. with:
```
gdal_create huge.tif -outsize 50000 50000  -co TILED=YES -co BLOCKXSIZE=1024 -co BLOCKYSIZE=1024 -co SPARSE_OK=YES -co COMPRESS=LZW -bands 3
(ulimit -v 1000000; GDAL_OVR_TEMP_DRIVER=GTIFF gdaladdo -r average huge.tif 2048 --config GDAL_TIFF_OVR_BLOCKSIZE 1024)
```

lucianpls:
@rouault 
Now it seems to be failing due to a slight shift when using NEAR overviews with VRT.
The new code does produce the same result with AVERAGE (38196), CUBIC (44957) and BILINEAR (47433)

lucianpls:
It is not a misalignment, it is fine, NEAR sampling is not uniquely defined when the sampling point is exactly at the intersection of pixels, as it is when scaling by 2. Can we just change the expected checksum?

rouault:
> Can we just change the expected checksum?

which test(s) are affected ?

It seems CI hard crashes on autotest/gcore/tiff_ovr.py::test_tiff_ovr_huge_raster_with_ovr_huge_block

lucianpls:
> > Can we just change the expected checksum?
> 
> which test(s) are affected ?
> 
> It seems CI hard crashes on autotest/gcore/tiff_ovr.py::test_tiff_ovr_huge_raster_with_ovr_huge_block

It was failing, not crashing. That's the test, it is using NEAR resampling and fails with a different checksum. I modified the test, it should pass now.

rouault:
> It was failing, not crashing.

well, the 2 CI configs (can't remember which ones) I checked did hard crash (segfault). I'll let you check after the current run of CI checks

lucianpls:
@rouault 
Is the nodata mask correctly dealt with by the VRT?

lucianpls:
@rouault 
Thanks for the help and patience!
I think that the old code path, the one that uses the temporary, only generates exceptions (huge output blocks or huge scale difference). It could be replaced with just the right error message and failure.

PR #12398, opened 2025-05-16
t-shirt art directory cleanup

hobu:


rouault:
The new .svg appears to be an image put in a .svg which is a bit deceptive (and heavy 4 MB)

hobu:
> The new .svg appears to be an image put in a .svg 

ah, I thought this was an SVG. Don't merge this until I clean it up.

PR #12400, opened 2025-05-16
VRTDerivedBand: Make source_names work like other builtin arguments

dbaston:
When adding the `SOURCE_NAMES` argument that is used to pass source names to the VRT expression parser, I didn't notice that a mechanism for such internal arguments already exists. This updates `SOURCE_NAMES` to use that existing machinery.

### DIFF
--- a/frmts/vrt/pixelfunctions.cpp
+++ b/frmts/vrt/pixelfunctions.cpp
@@ -1637,10 +1637,11 @@ static const char pszExprPixelFuncMetadata[] =
     "   <Argument name='dialect' "
     "             description='Expression dialect' "
     "             type='string-select'"
-    "             default='exprtk'>"
+    "             default='muparser'>"
     "       <Value>exprtk</Value>"
     "       <Value>muparser</Value>"
-    "    </Argument>"
+    "   </Argument>"
+    "   <Argument type='builtin' value='source_names' />"
     "</PixelFunctionArgumentsList>";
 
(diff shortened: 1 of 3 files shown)


frmts/vrt/pixelfunctions.cpp
```diff
@@ -1637,10 +1637,11 @@ static const char pszExprPixelFuncMetadata[] =
     "   <Argument name='dialect' "
     "             description='Expression dialect' "
     "             type='string-select'"
-    "             default='exprtk'>"
+    "             default='muparser'>"
```
rouault, line 1640: not for this PR, but shouldn't this metadata  have #ifdef GDAL_VRT_ENABLE_MUPARSER / GDAL_VRT_ENABLE_EXPRTK to determine the allowed ``<Value>``s and default value ?

rouault (approved):

frmts/vrt/vrtderivedrasterband.cpp
```diff
@@ -791,19 +792,50 @@ CPLErr VRTDerivedRasterBand::GetPixelFunctionArguments(
                         std::pair<CPLString, CPLString>(osName, osValue));
                 if (osType == "builtin")
                 {
+                    const CPLString &osArgName = osValue;
+                    CPLString osVal;
                     double dfVal;
```
rouault, line 797:
```suggestion
                    CPLString osVal;
                    double dfVal = 0;
```

PR #12409, opened 2025-05-18
gdal raster calc: check that formula is correct in RunStep() even when writing to VRT, GDALG or stream

rouault:


apps/gdalalg_raster_calc.cpp
```diff
@@ -602,6 +615,35 @@ bool GDALRasterCalcAlgorithm::RunImpl(GDALProgressFunc pfnProgress,
         return false;
     }
 
+    if (!m_skipCheckExpression)
+    {
+        const bool bIsVRT =
+            m_format == "VRT" ||
+            (m_format.empty() &&
+             EQUAL(
+                 CPLGetExtensionSafe(m_outputDataset.GetName().c_str()).c_str(),
```
dbaston, line 624: Why two different methods for checking extensions?

apps/gdalalg_raster_calc.cpp
```diff
@@ -572,6 +572,19 @@ GDALRasterCalcAlgorithm::GDALRasterCalcAlgorithm() noexcept
         .SetRequired()
         .SetPackedValuesAllowed(false)
         .SetMinCount(1);
+
+    AddArg("skip-check-expression", 0,
```
dbaston, line 576: no-check-expression for consistency with no-check-extent and no-check-crs?

rouault: OK. Note this is a hidden option just for unit test purposes for now. I'll add a code comment

apps/gdalalg_raster_calc.cpp
rouault: Yeah, that's a bit annoying. We should probably have a method that takes both the format and filename arguments and return the format. I'll let that for another PR

apps/gdalalg_raster_calc.cpp
```diff
@@ -602,6 +617,35 @@ bool GDALRasterCalcAlgorithm::RunImpl(GDALProgressFunc pfnProgress,
...
+                 CPLGetExtensionSafe(m_outputDataset.GetName().c_str()).c_str(),
+                 "VRT"));
+        const bool bIsGDALG =
+            m_format == "GDALG" ||
+            (m_format.empty() &&
+             cpl::ends_with(m_outputDataset.GetName(), ".gdalg.json"));
+        if (m_format == "stream" || bIsVRT || bIsGDALG)
+        {
+            // Try reading a single pixel to check formulas are valid.
+            std::vector<GByte> dummyData(vrt->GetRasterCount());
```
dbaston, line 635: Only other thought I had was modifying the VRT XML to read from a dummy source instead (`/vsidevnull` ? :upside_down_face: ) to avoid a potentially expensive read (poorly chunked dataset, for example)

rouault: yeah, your right. I had a similar thought. Probably that creating a fake 1x1 GeoTIFF file in /vsimem/ should do the trick

rouault: > yeah, your right. I had a similar thought. Probably that creating a fake 1x1 GeoTIFF file in /vsimem/ should do the trick

done

dbaston (approved):

PR #12410, opened 2025-05-19
VRT pixel functions: Handle NoData values

dbaston:
Makes pixel functions NoData-aware for _non-complex_ inputs. For most functions with <= 2 inputs, a NoData input results in a NoData output. For functions with > 2 inputs, the `propagateNoData` parameter is used to determine whether to ignore `NoData` inputs or cause them to produce `NoData` outputs.

There is definitely room to simplify the pixel function implementations but I didn't pursue that here, just focused on getting correct behavior for now.

Will conflict with #12406

### DIFF
--- a/autotest/gdrivers/vrtderived.py
+++ b/autotest/gdrivers/vrtderived.py
@@ -1456,3 +1456,81 @@ def test_vrt_pixelfn_reclassify_nan(tmp_vsimem):
         dst,
         np.array([[1, 2]]),
     )
+
+
+@pytest.mark.parametrize(
+    "pixelfn,values,nodata_value,pixelfn_args,expected",
+    [
+        ("dB", [7], 7, {}, 7),
+        ("diff", [3, 7], 7, {}, 7),
+        ("diff", [7, 3], 7, {}, 7),
+        ("div", [3, 7], 7, {}, 7),
+        ("div", [7, 3], 7, {}, 7),
+        ("exp", [7], 7, {}, 7),
+        ("interpolate_linear", [7, 10, 20, 7], 7, {"t0": 0, "dt": 10, "t": 5}, 7),
+        ("interpolate_linear", [7, 10, 20, 7], 7, {"t0": 0, "dt": 10, "t": -1}, 7),
+        ("interpolate_linear", [7, 10, 20, 7], 7, {"t0": 0, "dt": 10, "t": 10}, 10),
+        ("interpolate_linear", [7, 10, 20, 7], 7, {"t0": 0, "dt": 10, "t": 11}, 11),
+        ("interpolate_linear", [7, 10, 20, 7], 7, {"t0": 0, "dt": 10, "t": 20}, 20),
+        ("interpolate_exp", [7, 10, 20, 7], 7, {"t0": 0, "dt": 10, "t": 5}, 7),
+        ("interpolate_exp", [7, 10, 20, 7], 7, {"t0": 0, "dt": 10, "t": -1}, 7),
+        ("interpolate_exp", [7, 10, 20, 7], 7, {"t0": 0, "dt": 10, "t": 10}, 10),
+        ("interpolate_exp", [7, 10, 20, 7], 7, {"t0": 0, "dt": 10, "t": 11}, 10.717734),
+        ("interpolate_exp", [7, 10, 20, 7], 7, {"t0": 0, "dt": 10, "t": 20}, 20),
+        ("inv", [7], 7, {}, 7),
+        ("inv", [float("nan")], 7, {}, float("nan")),
+        ("inv", [float("nan")], float("nan"), {}, float("nan")),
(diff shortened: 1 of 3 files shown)


rouault (approved):

jratike80:
Slightly related, what is supposed to happen with:

```
gdal_create -ot byte -burn 0 -outsize 100 100 zero.tif -a_srs epsg:3857 -a_ullr 0 100 100 0
gdal_create -ot byte -burn 1 -outsize 100 100 one.tif -a_srs epsg:3857 -a_ullr 0 100 100 0

zerotest.vrt
=======

<VRTDataset rasterXSize="100" rasterYSize="100">
  <SRS dataAxisToSRSAxisMapping="1,2">PROJCS["WGS 84 / Pseudo-Mercator",GEOGCS["WGS 84",DATUM["WGS_1984",SPHEROID["WGS 84",6378137,298.257223563,AUTHORITY["EPSG","7030"]],AUTHORITY["EPSG","6326"]],PRIMEM["Greenwich",0,AUTHORITY["EPSG","8901"]],UNIT["degree",0.0174532925199433,AUTHORITY["EPSG","9122"]],AUTHORITY["EPSG","4326"]],PROJECTION["Mercator_1SP"],PARAMETER["central_meridian",0],PARAMETER["scale_factor",1],PARAMETER["false_easting",0],PARAMETER["false_northing",0],UNIT["metre",1,AUTHORITY["EPSG","9001"]],AXIS["Easting",EAST],AXIS["Northing",NORTH],EXTENSION["PROJ4","+proj=merc +a=6378137 +b=6378137 +lat_ts=0 +lon_0=0 +x_0=0 +y_0=0 +k=1 +units=m +nadgrids=@null +wktext +no_defs"],AUTHORITY["EPSG","3857"]]</SRS>
  <GeoTransform>  0.0000000000000000e+00,  1.0000000000000000e+00,  0.0000000000000000e+00,  1.0000000000000000e+02,  0.0000000000000000e+00, -1.0000000000000000e+00</GeoTransform>
  <VRTRasterBand dataType="Byte" band="1">
  	<ColorInterp>Gray</ColorInterp>
	<PixelFunctionType>div</PixelFunctionType>
    <SimpleSource>
      <SourceFilename relativeToVRT="1">one.tif</SourceFilename>
      <SourceBand>1</SourceBand>
      <SourceProperties RasterXSize="100" RasterYSize="100" DataType="Byte" BlockXSize="100" BlockYSize="81" />
      <SrcRect xOff="0" yOff="0" xSize="100" ySize="100" />
      <DstRect xOff="0" yOff="0" xSize="100" ySize="100" />
    </SimpleSource>
    <SimpleSource>
      <SourceFilename relativeToVRT="1">zero.tif</SourceFilename>
      <SourceBand>1</SourceBand>
      <SourceProperties RasterXSize="100" RasterYSize="100" DataType="Byte" BlockXSize="100" BlockYSize="81" />
      <SrcRect xOff="0" yOff="0" xSize="100" ySize="100" />
      <DstRect xOff="0" yOff="0" xSize="100" ySize="100" />
    </SimpleSource>
  </VRTRasterBand>
</VRTDataset>

gdal_translate zerotest.vrt test.tif 
```

rouault:
> Slightly related, what is supposed to happen with:

- Internally, computations are done on floating point, so 1 / 0 will evaluate to NaN
- then the code that converts NaN to Byte will make it a zero

dbaston:
We special-case 1/0 to evaluate to Inf, which apparently becomes 255 as Byte.

rouault:
> We special-case 1/0 to evaluate to Inf, which apparently becomes 255 as Byte.

oops you're right. I made a confusion with 0 / 0 (that normally is evaluated as NaN, but here we handle it as finite / 0 = Inf).
@jratike80 and I now see why I was confused because your VRT is actually wrong. It lacks a ``subclass="VRTDerivedRasterBand"`` attribute on the ``<VRTRasterBand>`` element. Lacking it, this is seen as a standard VRT of two sources, overlapping each other, hence the zero.tif is the one rendered...

dbaston:
> It lacks a subclass="VRTDerivedRasterBand" attribute on the <VRTRasterBand> element. Lacking it, this is seen as a standard VRT of two sources, overlapping each other, hence the zero.tif is the one rendered...

I've been thinking about tightening up this parsing since I have been confused by similar mistakes myself. (I think it must be `subClass` actually, `subclass` would be ignored ?)

rouault:
> I've been thinking about tightening up this parsing 

What would you suggest ? Well, for that particular instance, in (base) VRTRasterBand::XMLInit() we could check that if PixelFunctionType is present we have subClass=VRTDerivedRasterBand and emit a warning if not.

> (I think it must be `subClass` actually, `subclass` would be ignored ?)

ah the code is confusing. It looks for "subclass" on parsing, but the XML parser does case insensitive comparisons, but the generator and XML schema advertize "subClass".

PR #12412, opened 2025-05-19
gdal CLI: fix handling of --format

### LINKED ISSUE #12411: gdal driver gti create --format option not working
### What is the bug?

Based on the docs at https://gdal.org/en/latest/programs/gdal_driver_gti_create.html `--format` and `--output-format` should be interchangeable. 

The following works fine for creating a GTI:

```
gdal driver gti create --layer dem --dst-crs EPSG:2154 --output-format FlatGeobuf *.asc tileindex.gti.fgb
```

However if `--format` is used the command gives the same output as if using `gdal --format FlatGeobuf`:

```
gdal driver gti create --layer dem --dst-crs EPSG:2154 --format FlatGeobuf *.asc tileindex.gti.fgb
Format Details:
  Short Name: FlatGeobuf
  Long Name: FlatGeobuf
  Supports: Vector
  Extension: fgb
```

```
gdal driver gti create --layer dem --dst-crs EPSG:2154 --format FlatGeobuf *.asc tileindex.gti.fgb
```

### Versions and provenance

Windows 11
GDAL 3.12.0dev-2cf7f89c26

### Additional context

_No response_

rouault:
Fixes #12411

### DIFF
--- a/apps/gdal.cpp
+++ b/apps/gdal.cpp
@@ -16,6 +16,7 @@
 #include "gdal.h"
 
 #include <cassert>
+#include <utility>
 
 // #define DEBUG_COMPLETION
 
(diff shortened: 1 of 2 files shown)


dbaston:
Right now we have all of

```
gdal --format MEM
gdal --formats
gdal --drivers
```

and also

```
gdal vector --drivers
gdal vector --formats
```
What if the new CLI uses `--format` to mean exclusively `--output-format` ? (so we _always_ hide it from `GDALGeneralCmdLineProcessor`), and then we get driver info only using `--drivers` ?

```
gdal --drivers
gdal --drivers MEM
gdal --drivers --format text
gdal --drivers --format json
gdal --drivers MEM --format json
```

Or, driver information really being a separate command, rather than a modifier to a comand:

```
gdal drivers
gdal drivers MEM
gdal drivers --format text
gdal drivers --format json
gdal drivers MEM --format json
gdal vector drivers --format json
gdal vector drivers "ESRI Shapefile" --format json
```

rouault:
> Or, driver information really being a separate command, rather than a modifier to a comand:

that would probably be better. I'll keep this PR as it though for now so it can be backported to 3.11 to fix the immediate issue

PR #12418, opened 2025-05-20
VRT pixel functions: Add mean, median, geometric_mean, harmonic_mean, mode

dbaston:


frmts/vrt/pixelfunctions.cpp
```diff
@@ -1997,6 +1997,308 @@ static CPLErr ReclassifyPixelFunc(void **papoSources, int nSources, void *pData,
...
+        return !values.empty();
+    }
+
+    double GetValue() const
+    {
+        std::sort(values.begin(), values.end());
+        if (values.size() == 1)
+        {
+            return values[0];
+        }
```
rouault, line 2156: The general case for odd counts will work just fine for count == 1:

```suggestion
```

rouault (approved):

PR #12424, opened 2025-05-20
Add 'gdal raster color-merge' (port of hsv_merge.py sample script)

rouault:
`gdal raster color-merge` allows the user to colorize a grayscale image with a RGB one.

It does the following steps:
- read the RGB (Red,Green,Blue) or RGBA (Red,Green,Blue,Alpha) dataset
- transforms its RGB components into the HSV (Hue,Saturation,Value) color space
- substitute the value component with the one from the grayscale raster
- transform back to RGB.

This can be for example to colorize a hillshade raster generated by `gdal_raster_hillshade` with an hypsometric rendering of a DEM generated by `gdal_raster_color_map`.


apps/gdalalg_raster_color_merge.h
```diff
@@ -0,0 +1,59 @@
...
+class GDALRasterColorMergeAlgorithm /* non final*/
+    : public GDALRasterPipelineStepAlgorithm
+{
+  public:
+    explicit GDALRasterColorMergeAlgorithm(bool standaloneStep = false);
+
+    static constexpr const char *NAME = "color-merge";
+    static constexpr const char *DESCRIPTION =
+        "Replace the color value of a RGB/RGBA dataset by the one of a "
+        "grayscale raster";
```
jratike80, line 33: This description does not make much sense to me Isn't the direction rather the opposite, from RGB to what-used-to-be grayscale?

rouault: would "Use a grayscale raster to replace the intensity of a RGB/RGBA dataset" make more sense ?  (knowing that "intensity" is a bit improper here as it is actually the V(alue) component in HSV color space, but intensity should be a reasonable quick summary)

jratike80: For me, yes. But some regular GIS user may think that it is the hillshade image that gets colored, not the RGB image that gets intensified.

rouault: > But some regular GIS user may think that it is the hillshade image that gets colored, not the RGB image that gets intensified.

isn't that the same thing: replace (hue, saturation) of the hillshade with (hue, saturation) of the RGB   vs replace "value" (intensity) of the RGB with value of the hillshade ?

I'm open to better wording if you can come with something :-)

jratike80: I think it is good enough. And you described one use case already as "This can be for example to colorize a hillshade raster generated by gdal_raster_hillshade with an hypsometric rendering of a DEM".

doc/source/programs/gdal_raster_color_merge.rst
```diff
@@ -0,0 +1,91 @@
...
+
+- read the RGB (Red,Green,Blue) or RGBA (Red,Green,Blue,Alpha) dataset
+
+- transform its RGB components into the HSV (Hue,Saturation,Value) color space
+
+- substitute the value component with the one from the grayscale raster
+
+- transform back to RGB.
+
+This can be for example to colorize a hillshade raster generated by
```
elpaso, line 35:
```suggestion
This can be used for example to colorize a hillshade raster generated by
```

autotest/utilities/test_gdalalg_raster_color_merge.py
```diff
@@ -0,0 +1,362 @@
...
+def test_gdalalg_raster_color_merge_out_of_memory(tmp_vsimem, xsize, ysize, req_object):
+
+    if xsize == 100 * 1000 * 1000 and gdaltest.is_travis_branch("sanitize"):
+        # otherwise it would throw AddressSanitizer: out of memory: allocator is trying to allocate 0x8e1bc9bf040000 bytes
+        pytest.skip("skip on ASAN")
+
+    gdal.GetDriverByName("GTiff").Create(
+        tmp_vsimem / "in.tif", 2, 2, 3, options=["COMPRESS=DEFLATE", "BLOCKYSIZE=2"]
+    )
+    with gdal.VSIFile(tmp_vsimem / "in.tif", "rb") as f:
```
elpaso, line 325: why you read data from a newly created file and rewrite them to the same file?

doc/source/programs/gdal_raster_color_merge.rst
rouault: fixed

autotest/utilities/test_gdalalg_raster_color_merge.py
rouault: I actually truncate the file by writing all bytes but the last 3 ones, to make it invalid and check correct I/O error propagation. I'll add a code comment

thareUSGS:
Excellent port from Python. Based on a simple request for this functionality, Frank did all the coding. I just added the ability to loop at the line level so huge merges could be completed. 

I will be testing this right away.

rouault:
> I just added the ability to loop at the line level so huge merges could be completed.

by the way, the new implementation is fully streamable, so you can genreate a .gdalg.json file (cf https://gdal.org/en/stable/drivers/raster/gdalg.html) from it ("gdal raster color-merge rgb.tif grayscale.tif out.gdalg.json"), and do piece-wise reading after.

PR #12436, opened 2025-05-21
VSICurlHandle::AdviseRead() (GTIFF multithreading /vsicurl/ reading): implement retry strategy

### LINKED ISSUE #12426: GDAL_HTTP_MAX_RETRY appears to be ineffective when threading is enabled
### What is the bug?

Working with @afrank150, we discovered some odd behavior that might be related to #11552, or at least the resulting output is the same. It seems like HTTP retries don't happen in a multithreaded environment.

In #11552, I reported that a server failure can lead to what looks like valid output. A translate command runs to completion even when the server is killed, and the output is corrupt but there's no indication of an error.

This seems to be a similar case but approached from another direction... if threading is enabled, an intermittent failure can lead to some corrupted data in an image where a retry should have occurred. With one thread, a retry does occur and the output looks fine. With multiple threads, no retry occurs and the output is corrupted.

Here's an example:
Image

This is a similar setup to what I did for #11552 where I use a local nginx server. In the old issue, I killed the server, but here I just temporarily move the file being served and then move it back. So the server temporarily responds with 404 errors during some of the range requests. With a single thread, GDAL waits the configured time and retries, but with multiple threads it doesn't appear that any retry happens.

First step is to generate some test data, so this snippet will create a 10k by 10k image with a vertical gradient. It will make the error pretty obvious.

```Python
from osgeo import gdal
import numpy as np

gdal.UseExceptions()
drv = gdal.GetDriverByName("GTiff")
ds = drv.Create("gradient.tif", 10000, 10000, 1, gdal.GDT_UInt16)
rb = ds.GetRasterBand(1)
rb.WriteArray(np.tile(np.arange(0, 10000, 1).reshape((10000, 1)), (1, 10000)))
ds = None
```

The image should look like this:
Image

Next, start a local nginx server that will serve images out of the current directory.

```bash
docker run --rm --name nginx -it -v $PWD:/usr/share/nginx/html:ro -p 80:80 nginx:latest
```

Now, run a `gdal_translate` command to pull the image through the server. There is a parallel shell command started first that will sleep for 0.5 seconds, move the gradient file, sleep another 0.2 seconds, then move the file back. Since the translate takes longer than that span of time, one or more 404 errors will be returned in the middle of the translate. (I set `GDAL_HTTP_RETRY_CODES=ALL` to ensure that a retry should happen.)

```bash
bash -c "sleep 0.5; mv gradient.tif gradient.tif.moved; sleep 0.2; mv gradient.tif.moved gradient.tif"&; GDAL_NUM_THREADS=2 GDAL_HTTP_MAX_RETRY=5 GDAL_HTTP_RETRY_DELAY=1 GDAL_HTTP_RETRY_CODES=ALL CPL_VSIL_CURL_ALLOWED_EXTENSIONS=tif gdal_translate /vsicurl/http://localhost/gradient.tif translate.tif
```

With two threads, it will report the error and finish, and the image will not be a smooth gradient anymore:
Image

The output looks like this. The shell is reporting that the other bash command finishes right in the middle of execution, but the translate reports a 404 error. It completes successfully.

```
[1] 22658
Input file size is 10000, 10000
0...10...20ERROR 1: Request for 40040146-50040145 failed with response_code=404
[1]  + done       bash -c
                          ...30...40...50...60...70...80...90...100 - done.
```

If I add `CPL_CURL_VERBOSE=YES` and capture the output, I can see where the 404 happens:
```
> GET /gradient.tif HTTP/1.1
Host: localhost
User-Agent: GDAL/3.11.0
Accept: */*
Range: bytes=40040146-50040145

< HTTP/1.1 404 Not Found
< Server: nginx/1.27.5
< Date: Tue, 20 May 2025 21:39:39 GMT
< Content-Type: text/html
< Content-Length: 153
< Connection: keep-alive
<
* Connection #0 to host localhost left intact
ERROR 1: Request for 40040146-50040145 failed with response_code=404
```

But if I search the entire log for that range of bytes, `40040146-50040145`, it only appears in these two places and nowhere else. There does not appear to be any retry.

Now if I remove the output (`rm translate.tif`) and try again, setting `GDAL_NUM_THREADS=1`, I get a different result. The output reports a retry:

```
[1] 22937
Input file size is 10000, 10000
0...10...20Warning 1: HTTP error code: 404 - http://localhost/gradient.tif. Retrying again in 1.0 secs
[1]  + done       bash -c
                          ...30...40...50...60...70...80...90...100 - done.
```

The output looks correct:
Image

If I capture the verbose curl output again, the same range of bytes is requested twice in a row with the retry log message in between:

```
> GET /gradient.tif HTTP/1.1
Host: localhost
User-Agent: GDAL/3.11.0
Accept: */*
Range: bytes=16384-32767

< HTTP/1.1 404 Not Found
< Server: nginx/1.27.5
< Date: Tue, 20 May 2025 21:47:07 GMT
< Content-Type: text/html
< Content-Length: 153
< Connection: keep-alive
<
* Connection #0 to host localhost left intact
Warning 1: HTTP error code: 404 - http://localhost/gradient.tif. Retrying again in 1.0 secs
* Couldn't find host localhost in the .netrc file; using defaults
* Found bundle for host: 0x55a892401070 [serially]
* Can not multiplex, even if we wanted to
* Re-using existing connection with host localhost
> GET /gradient.tif HTTP/1.1
Host: localhost
User-Agent: GDAL/3.11.0
Accept: */*
Range: bytes=16384-32767
```

### Versions and provenance

I have tested and observed this behavior in the official Docker image `ghcr.io/osgeo/gdal:ubuntu-small-3.11.0`. @afrank150 has also observed it on a Macbook.

### Additional context

_No response_

rouault:
Fixes #12426

### DIFF
--- a/autotest/gcore/tiff_read.py
+++ b/autotest/gcore/tiff_read.py
@@ -4993,7 +4993,9 @@ def test_tiff_read_multi_threaded(
 
 
 @pytest.mark.parametrize("use_dataset_readraster", [True, False])
-@pytest.mark.parametrize("advise_read", [True, False])
+@pytest.mark.parametrize(
+    "advise_read,test_retry", [(True, False), (True, True), (False, False)]
+)
 @pytest.mark.skipif(
     platform.system() == "Darwin" or gdaltest.is_travis_branch("mingw64"),
     reason="fails randomly",
@@ -5003,7 +5005,9 @@ def test_tiff_read_multi_threaded(
     not check_libtiff_internal_or_at_least(4, 0, 11),
     reason="libtiff >= 4.0.11 required",
 )
-def test_tiff_read_multi_threaded_vsicurl(use_dataset_readraster, advise_read):
+def test_tiff_read_multi_threaded_vsicurl(
+    use_dataset_readraster, advise_read, test_retry
+):
 
     webserver_process = None
     webserver_port = 0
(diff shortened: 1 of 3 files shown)


rouault:
backported to 3.11 per 3a46f249c99e2182e9be316e86ceae0d86baaeb5

PR #12448, opened 2025-05-22
Add 'gdal raster pansharpen' (port of gdal_pansharpen.py)

rouault:


jratike80:
Wondering if it could make sense to sharpen with common RGB images, by creating a grayscale version on-the-fly. I know I have been experimenting something as follows, I just don't remember why:

`gdal_calc.py -R input.tif --R_band=1 -G input.tif --G_band=2 -B input.tif --B_band=3 --outfile=result.tif --calc="R*0.2989+G*0.5870+B*0.1140"`

May be a useless idea.

jratike80:
Sorry, I must have clicked the mouse by accident.

rouault:
> Wondering if it could make sense to sharpen with common RGB images, by creating a grayscale version on-the-fly. I know I have been experimenting something as follows, I just don't remember why:

I don't think that makes sense, because the whole point of pansharpening is that your R,G,B (or whatever) color bands are less resoluted than the panchromatic band (all reflectance), and you use the later to improve the resolution of the spectral bands. Computing a greyscale version of the R,G,B will never enhance their resolution

jratike80:
I was thinking about a case when RGB and other data come from different sources or instruments, like a device recording RGB(N) with one instrument and thermal infrared with another one. Or if there happens to be aerial images and satellite data from the same place with about same aquisition times. But that is an uncommon use case and can be handled by pre-prossing the higher resolion data.

PR #12456, opened 2025-05-24
Simplify header file inclusion

sunpoet:
It is the same as [frmts/libertiff/libertiffdataset.cpp](https://github.com/OSGeo/gdal/blob/master/frmts/libertiff/libertiffdataset.cpp#L35).


rouault:
@sunpoet It would be great to at least locally test that your changes build before submitting a pull request. I've addressed that in 89fe104183b3937232008dfac86850a5c05b5dc6

PR #12477, opened 2025-05-28
Apply transform to distance calculation for pitch masking.

abellgithub:
Adjust distance calculation for transform.


abellgithub:
No. The test I had didn't use a transform.

I will add a test or two if this checks out OK.

PR #12483, opened 2025-05-29
gdal vector simplify-coverage: implement progress callback

rouault:
CC @dbaston

### DIFF
--- a/apps/gdalalg_vector_pipeline.cpp
+++ b/apps/gdalalg_vector_pipeline.cpp
@@ -994,7 +994,7 @@ GDALVectorNonStreamingAlgorithmDataset::
 /************************************************************************/
 
 bool GDALVectorNonStreamingAlgorithmDataset::AddProcessedLayer(
-    OGRLayer &srcLayer)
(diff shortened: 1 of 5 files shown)


dbaston:
I didn't add a progress callback because > 90% of the time was inside GEOS.

rouault:
@dbaston Do you believe the GEOS team would welcome enhancements to add progress report and cancellation for this algorithm? (to be checked if that's possible in that case). I don't see any precedent in the C API (the interrupt callback is the closest, but doesn't cover the progress part)

dbaston:
Some timings from simplifying TIGER state boundaries with a tolerance of 1e-3...

97% of runtime in GDALVectorSimplifyCoverageOutputDataset::Process(OGRLayer&, OGRLayer&)
93% of runtime in geos::coverage::CoverageSimplifier::simplify(double) 

which is split into
40% in geos::coverage::CoverageRingEdges::build
52% in geos::coverage::TPVWSimplifier::simplify

So, a roughly 50/50 split between figuring out the shared edges and progressively removing vertices. And < 5% GEOS conversion overhead.

PR #12485, opened 2025-05-29
HDF5/BAG: do not try opening datasets with one dimension size > INT_MAX

rouault:


elpaso (approved):
Changes looking good but I see that h5 paDims is
`typedef unsigned long long` , is it an internal limitation of GDAL using `int` for a size dimension that (AFAIK) should actually be an unsigned type?

PR #12488, opened 2025-05-30
PGDUMP: Use OGR_PG_SKIP_CONFLICTS

fitnr:
## What does this PR do?

Add support for the `OGR_PG_SKIP_CONFLICTS` config setting to the PGDump driver. The additional code and test follow the pattern in #10156, which added the option for the PostgreSQL driver.

The update adds `on conflict do nothing` to the end of each INSERT statement when the option is true. 

## Tasklist

 - [x] Add test case(s)
 - [x] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/doc/source/drivers/vector/pgdump.rst
+++ b/doc/source/drivers/vector/pgdump.rst
@@ -182,6 +182,13 @@ The following layer creation options are supported:
       DROP TABLE from being emitted. Set to IF_EXISTS
       in order DROP TABLE IF EXISTS to be emitted (needs PostgreSQL >= 8.2)
 
+-  .. config:: OGR_PG_SKIP_CONFLICTS
+      :choices: YES, NO
+      :default: NO
+
+      If set to "YES" (not the default), conflicts when inserting features
+      will be skipped (only applies when PG_USE_COPY is off).
+
 -  .. lco:: SRID
 
       Set the SRID of the geometry. Defaults to -1, unless a SRS
--- a/ogr/ogrsf_frmts/pgdump/ogr_pgdump.h
+++ b/ogr/ogrsf_frmts/pgdump/ogr_pgdump.h
@@ -121,6 +121,7 @@ class OGRPGDumpLayer final : public OGRLayer
     bool m_bCopyStatementWithFID = true;
     bool m_bNeedToUpdateSequence = false;
     bool m_bGeomColumnPositionImmediate = true;
+    int bSkipConflicts = false;
     std::vector<std::string> m_aosDeferredGeomFieldCreationCommands{};
     std::vector<std::string> m_aosDeferrentNonGeomFieldCreationCommands{};
     std::vector<std::string> m_aosSpatialIndexCreationCommands{};
--- a/ogr/ogrsf_frmts/pgdump/ogrpgdumplayer.cpp
+++ b/ogr/ogrsf_frmts/pgdump/ogrpgdumplayer.cpp
@@ -47,7 +47,9 @@ OGRPGDumpLayer::OGRPGDumpLayer(OGRPGDumpDataSource *poDSIn,
           OGRPGDumpEscapeColumnName(pszTableName).c_str()))),
       m_pszFIDColumn(pszFIDColumnIn ? CPLStrdup(pszFIDColumnIn) : nullptr),
       m_poFeatureDefn(new OGRFeatureDefn(pszTableName)), m_poDS(poDSIn),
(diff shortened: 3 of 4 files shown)


doc/source/drivers/vector/pgdump.rst
```diff
@@ -182,6 +182,13 @@ The following layer creation options are supported:
       DROP TABLE from being emitted. Set to IF_EXISTS
       in order DROP TABLE IF EXISTS to be emitted (needs PostgreSQL >= 8.2)
 
+-  .. config:: OGR_PG_SKIP_CONFLICTS
```
rouault, line 185: I understand the symmetry with the PG driver, but it would be best that this is dealt as a SKIP_CONFLICTS layer creation option to avoid proliferation of global configuration option

fitnr: That makes sense. I've felt the pain point in needing to use a mix of `--config` and `-lco` in a command.

It seems like such an option should be supported for the PG driver too, or would that be unnecessary?

rouault: > or would that be unnecessary?

would be nice, but not a requirement for the scope of this PR

port/cpl_known_config_options.h
```diff
@@ -780,7 +780,7 @@ constexpr static const char* const apszKnownConfigOptions[] =
    "OGR_PG_ENABLE_METADATA", // from ogrpgdatasource.cpp, ogrpgtablelayer.cpp
    "OGR_PG_JSON_TYPE", // from ogrpgdumplayer.cpp
    "OGR_PG_RETRIEVE_FID", // from ogrpgtablelayer.cpp
-   "OGR_PG_SKIP_CONFLICTS", // from ogrpgtablelayer.cpp
+   "OGR_PG_SKIP_CONFLICTS", // from ogrpgdumplayer.cpp, ogrpgtablelayer.cpp
```
rouault, line 783:
```suggestion
   "OGR_PG_SKIP_CONFLICTS", // from ogrpgtablelayer.cpp
```

ogr/ogrsf_frmts/pgdump/ogr_pgdump.h
```diff
@@ -107,6 +107,7 @@ class OGRPGDumpLayer final : public OGRLayer
     bool m_bCopyActive = false;
     bool m_bFIDColumnInCopyFields = false;
     int m_bCreateTable = false;
+    int m_bSkipConflicts = false;
```
rouault, line 110:
```suggestion
    bool m_bSkipConflicts = false;
```

ogr/ogrsf_frmts/pgdump/ogr_pgdump.h
```diff
@@ -139,7 +140,7 @@ class OGRPGDumpLayer final : public OGRLayer
   public:
     OGRPGDumpLayer(OGRPGDumpDataSource *poDS, const char *pszSchemaName,
                    const char *pszLayerName, const char *pszFIDColumn,
-                   int bWriteAsHexIn, int bCreateTable);
+                   int bWriteAsHexIn, int bCreateTable, int bSkipConflicts);
```
rouault, line 143:
```suggestion
                   int bWriteAsHexIn, int bCreateTable, bool bSkipConflicts);
```

doc/source/drivers/vector/pgdump.rst
```diff
@@ -182,6 +182,13 @@ The following layer creation options are supported:
       DROP TABLE from being emitted. Set to IF_EXISTS
       in order DROP TABLE IF EXISTS to be emitted (needs PostgreSQL >= 8.2)
 
+-  .. lco:: SKIP_CONFLICTS
+      :choices: ON, OFF
+      :default: OFF
```
rouault, line 187:
```suggestion
      :default: OFF
      :since: 3.12
```

ogr/ogrsf_frmts/pgdump/ogrpgdumplayer.cpp
```diff
@@ -40,14 +40,15 @@ OGRPGDumpLayer::OGRPGDumpLayer(OGRPGDumpDataSource *poDSIn,
                                const char *pszSchemaNameIn,
                                const char *pszTableName,
                                const char *pszFIDColumnIn, int bWriteAsHexIn,
-                               int bCreateTableIn)
+                               int bCreateTableIn, int bSkipConflictsIn)
```
rouault, line 43:
```suggestion
                               int bCreateTableIn, bool bSkipConflictsIn)
```

rouault (approved):

PR #12493, opened 2025-05-30
CI 20.04: update to GEOS main

rouault:


dbaston:
It looks like this is the only CI config that is building GEOS, should we use the occasion to build GEOS master?

dbaston (approved):

PR #12496, opened 2025-05-30
Doc: Add image for gdal raster footprint

dbaston:
Ends up adding about 60kb of binary files. Current autotest directory is about 150mb, so maybe I don't need to worry about it?


rouault:
> Ends up adding about 60kb of binary files. Current autotest directory is about 150mb, so maybe I don't need to worry about it?

yes that's fine

Issue #9772, opened 2024-04-26
CSV output HEADERS control


jidanni:
### Feature description

One can control HEADERS on CSV input, but not CSV output.

### Additional context

- https://gdal.org/drivers/vector/csv.html#open-options
  - SEPARATOR=[AUTO/COMMA/SEMICOLON/TAB/SPACE/PIPE]
  - HEADERS=[YES/NO/AUTO]
- https://gdal.org/drivers/vector/csv.html#layer-creation-options
  - SEPARATOR=[COMMA/SEMICOLON/TAB/SPACE]
  - *No HEADERS control provided. Must use "sed 1d" etc. to remove them.*

i.e.,
https://gdal.org/drivers/vector/csv.html#open-options has
> HEADERS=[YES/NO/AUTO]: Defaults to AUTO. Whether the first line of the file contains column names or not. When set to AUTO, GDAL will assume the first line is column names if none of the values are strictly numeric.

That's fine, but it only concerns input. Therefore
https://gdal.org/drivers/vector/csv.html#layer-creation-options
should give a method to remove headers upon output. No more needing to do
`ogr2ogr /vsistdout/ ... | sed 1d`
Perhaps allow `-lco HEADERS=NO`.

jidanni:
Alas, just workarounds,
https://gis.stackexchange.com/questions/460413/how-to-remove-the-first-header-line-when-creating-a-csv-with-ogr2ogr/463367#463367

How to get rid of headers on ogr2ogr -f CSV -sql OUTPUT?
Sure there is `ogr2ogr -oo HEADERS=NO` and `-oo X_POSSIBLE_NAMES=field_1`, `field_2`, etc., but that is for -f CSV **input**. How to say "no headers, please" for -f CSV **output**?

```
$ ogr2ogr -f CSV /vsistdout/ malate.geojson \
-sql 'SELECT "addr:housenumber" FROM malate'
addr:housenumber
1743
680
1632
1771
```
See, there it is, that wretched header, right there on line 1 of my output.

> Why is it bothering you so much?

That's beyond the scope of this question.

> I see. Well why don't you just pipe the output through `sed 1d`?
Voila, problem solved. Get on with life.

That would be cheating. One would think that if there is a way to deal with no HEADER in input, there would also be a way to get that kind of output too.
Upon finding none in
[...]

PR #12499, opened 2025-05-31
CSV: Allow to use the PIPE as separator for layer creation

agiudiceandrea:
## What does this PR do?

Allows to use the PIPE as separator for layer creation.

## Tasklist

 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments
 - [ ] ADD YOUR TASKS HERE

## Environment

* OS:
* Compiler:

### DIFF
--- a/doc/source/drivers/vector/csv.rst
+++ b/doc/source/drivers/vector/csv.rst
@@ -472,7 +472,7 @@ The following layer creation options are supported:
       This option also creates .prj file which stores coordinate system information.
 
 -  .. lco:: SEPARATOR
-      :choices: COMMA, SEMICOLON, TAB, SPACE
+      :choices: COMMA, SEMICOLON, TAB, SPACE, PIPE
       :default: COMMA
 
       Field separator character.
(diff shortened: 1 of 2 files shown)


rouault:
> Allows to use the PIPE as separator for layer creation.

For my information, in which use case it that useful ?

jratike80:
Pipe is already supported for reading "Starting with GDAL 3.7, pipe separated values files with a ".psv" extension are also recognized" so maybe some people like it. I wonder what is the next wish for a separator.

agiudiceandrea:
> For my information, in which use case it that useful ?

I've recently found very useful to use the pipe as separator, since my data (almost) always doesn't contain such char. It seems to me it's also easier to visually separate the columns when the pipe is used. Every now and then some colleagues ask me to provide the data separated by pipe.
I'm also a bit allergic to inconsistencies... ogr can read CSV with pipe, why it cannot write the same? :-)
(another improvement in this direction would be to address https://github.com/OSGeo/gdal/issues/9772 as well)

rouault (approved):

PR #12504, opened 2025-06-01
CSV: add a HEADER=YES/NO layer creation option

### LINKED ISSUE #9772: CSV output HEADERS control

rouault:
Fixes #9772


agiudiceandrea:
Great!
Is the name `HEADER` instead of `HEADERS` (like for the open option) intended?

rouault:
> Is the name `HEADER` instead of `HEADERS` (like for the open option) intended?

both are possible. I was wondering why the open option is called HEADERS plural and not just HEADER, given there's just a single header line, hence this alias for the creation option. We could potentially also offer HEADER as an alias for the open option as well

PR #12505, opened 2025-06-01
gdal raster mosaic: add completion/choices for --pixel-function and --pixel-function--arg

rouault:
CC @dbaston 

```
$ gdal raster mosaic --pixel-function=<TAB<>TAB>
cmul                dB2amp              exp                 imag                inv                 median              mul                 pow                 scale
complex             dB2pow              expression          intensity           log10               min                 norm_diff           real                sqrt
conj                diff                geometric_mean      interpolate_exp     max                 mod                 phase               reclassify          sum
dB                  div                 harmonic_mean       interpolate_linear  mean                mode                polar               replace_nodata

$ gdal raster mosaic --pixel-function=median --pixel-function-arg=<TAB>
==>
$ gdal raster mosaic --pixel-function=median --pixel-function-arg=propagateNoData=

$ gdal raster mosaic --pixel-function=real --pixel-function-arg=<TAB><TAB>
**                                                     No pixel function arguments for pixel function real
```

### DIFF
--- a/apps/gdalalg_raster_mosaic.cpp
+++ b/apps/gdalalg_raster_mosaic.cpp
@@ -17,6 +17,7 @@
 
 #include "gdal_priv.h"
 #include "gdal_utils.h"
+#include "vrtdataset.h"
 
 //! @cond Doxygen_Suppress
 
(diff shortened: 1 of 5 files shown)


dbaston:
Tested this locally and the the autocomplete works well to select a pixel function and an argument name. But accidentally hitting tab after a pixel function arg is selected comes up with a list of recently accessed files on my system.

```
 gdal raster mosaic --pixel-function max --pixel-function-arg propagateNoData=
```

tab tab gives choices like 

```
create_lat_raster.py                      get_dem_diff.py
```

For a bool argument we should be able to offer YES/NO. But this is pretty niche, maybe further autocomplete should be disabled once an argument is selected?

rouault:
> But accidentally hitting tab after a pixel function arg is selected comes up with a list of recently accessed files on my system.

I've added a commit fixing that

PR #12506, opened 2025-06-02
MiraMonVector: Fixing decimal figures error

AbelPau:
## What does this PR do?
Fixed an issue affecting the translation of numeric fields with defined decimal precision to MiraMon format (e.g., from a Shapefile).
A test has been added that converts a minimal Shapefile to MiraMon format and verifies the output for the affected fields.

## Tasklist

 - [x] Add test case(s)
 - [x] Review


ogr/ogrsf_frmts/miramon/ogrmiramonlayer.cpp
```diff
@@ -2725,10 +2726,25 @@ OGRErr OGRMiraMonLayer::TranslateFieldsValuesToMM(OGRFeature *poFeature)
...
-                    szChain, sizeof(szChain),
-                    phMiraMonLayer->pLayerDB->pFields[iField].nNumberOfDecimals,
-                    poFeature->GetFieldAsDouble(iField));
+                // According to NOTE_3812_20250527
+                if (phMiraMonLayer->pLayerDB->pFields[iField]
+                            .nNumberOfDecimals > 0 &&
+                    phMiraMonLayer->pLayerDB->pFields[iField]
+                            .nNumberOfDecimals < MAX_RELIABLE_SF_DOUBLE)
+                {
+                    snprintf(szChain, sizeof(szChain), "%.*f",
```
rouault: Better use CPLsnprintf() to make sure that a decimal point is used, instead of the decimal separator of the current locale (e.g. when running within QGIS in a locale where comma would be the decimal separator)

```suggestion
                    CPLsnprintf(szChain, sizeof(szChain), "%.*f",
```

AbelPau: Thanks. Updated!

PR #12507, opened 2025-06-02
GDALRasterBand: define +, -, *, /, AsType(), gdal::min()/max() to perform band lazy-evaluated arithmetics from C++, C and SWIG (Python)

rouault:
Documentation in https://gdal--12507.org.readthedocs.build/en/12507/user/band_algebra.html

Uses VRTDerivedRasterBand and pixel functions underneath.

Demo program converting a RGB image to grey level:

- Python:

 ```python
from osgeo import gdal
gdal.UseExceptions()
    
with gdal.Open("rgbsmall.tif") as ds:
    R = ds.GetRasterBand(1)
    G = ds.GetRasterBand(2)
    B = ds.GetRasterBand(3)
    greylevel = (0.299 * R + 0.587 * G + 0.114 * B).astype(gdal.GDT_Byte)
    gdal.GetDriverByName("GTiff").CreateCopy("greylevel.tif", greylevel)
```

- C++:

```cpp
#include <gdal_priv.h>

int main()
{
    GDALAllRegister();

    auto poDS = std::unique_ptr<GDALDataset>(GDALDataset::Open("rgbsmall.tif"));
    auto& R = *(poDS->GetRasterBand(1));
    auto& G = *(poDS->GetRasterBand(2));
    auto& B = *(poDS->GetRasterBand(3));
    auto greylevel = (0.299 * R + 0.587 * G + 0.114 * B).AsType(GDT_Byte);

    auto poGTiffDrv = GetGDALDriverManager()->GetDriverByName("GTiff");
    std::unique_ptr<GDALDataset>(
        poGTiffDrv->CreateCopy("greylevel.tif", greylevel.GetDataset(), false, nullptr, nullptr, nullptr)).reset();

    return 0;
}
```

CC @dbaston I've experimented with the above as a potentially alternative/complement to gdal raster calc for C++ & Python users. Thoughts?

### DIFF
--- a/frmts/vrt/pixelfunctions.cpp
+++ b/frmts/vrt/pixelfunctions.cpp
@@ -2347,6 +2347,8 @@ static CPLErr NormDiffPixelFunc(void **papoSources, int nSources, void *pData,
 
 static const char pszMinMaxFuncMetadataNodata[] =
     "<PixelFunctionArgumentsList>"
+    "   <Argument name='k' description='Optional constant term' type='double' "
+    "default='nan' />"
     "   <Argument type='builtin' value='NoData' optional='true' />"
     "   <Argument name='propagateNoData' description='Whether the output value "
     "should be NoData as as soon as one source is NoData' type='boolean' "
@@ -2354,10 +2356,11 @@ static const char pszMinMaxFuncMetadataNodata[] =
     "</PixelFunctionArgumentsList>";
 
 template <class Comparator>
-static CPLErr MinOrMaxPixelFunc(void **papoSources, int nSources, void *pData,
-                                int nXSize, int nYSize, GDALDataType eSrcType,
-                                GDALDataType eBufType, int nPixelSpace,
-                                int nLineSpace, CSLConstList papszArgs)
+static CPLErr MinOrMaxPixelFunc(double dfK, void **papoSources, int nSources,
+                                void *pData, int nXSize, int nYSize,
+                                GDALDataType eSrcType, GDALDataType eBufType,
+                                int nPixelSpace, int nLineSpace,
+                                CSLConstList papszArgs)
 {
     /* ---- Init ---- */
     if (GDALDataTypeIsComplex(eSrcType))
@@ -2399,9 +2402,14 @@ static CPLErr MinOrMaxPixelFunc(void **papoSources, int nSources, void *pData,
                 }
             }
 
-            if (!bPropagateNoData && std::isnan(dfRes))
+            if (std::isnan(dfRes))
             {
-                dfRes = dfNoData;
+                if (!bPropagateNoData)
+                    dfRes = dfNoData;
+            }
+            else if (!std::isnan(dfK) && Comparator::compare(dfK, dfRes))
+            {
+                dfRes = dfK;
             }
 
             GDALCopyWords(&dfRes, GDT_Float64, 0,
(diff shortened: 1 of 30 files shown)


rouault:
Python support now added !

rouault:
Ready for review

rouault:
Documentation in https://gdal--12507.org.readthedocs.build/en/12507/user/band_algebra.html

swig/include/python/gdal_python.i
```diff
@@ -246,6 +246,93 @@ static void readraster_releasebuffer(CPLErr eErr,
             yield entry
     finally:
         CloseDir(dir)
+
+  def where(cond_band, then_band, else_band):
```
dbaston, line 250: I would stick to the general GDAL Python convention of Pascal case method names.

gcore/gdalcomputedrasterband.cpp
```diff
@@ -0,0 +1,687 @@
...
+        {
+            bIsFirstBand = false;
+            bFirstBandHasNoData = bHasNoData;
+            dfFirstNoData = dfNoData;
+        }
+        else if (bHasNoData != bFirstBandHasNoData)
+        {
+            if (bWarnAboutNoData)
+            {
+                CPLError(CE_Warning, CPLE_AppDefined,
```
dbaston: Would it be better to use a `ComplexSource` to handle differing nodata values among sources? Ignoring nodata seems like it could cause surprises.

frmts/vrt/pixelfunctions.cpp
```diff
@@ -2347,17 +2347,20 @@ static CPLErr NormDiffPixelFunc(void **papoSources, int nSources, void *pData,
...
+    "   <Argument name='k' description='Optional constant term' type='double' "
+    "default='nan' />"
     "   <Argument type='builtin' value='NoData' optional='true' />"
     "   <Argument name='propagateNoData' description='Whether the output value "
     "should be NoData as as soon as one source is NoData' type='boolean' "
     "default='false' />"
     "</PixelFunctionArgumentsList>";
 
 template <class Comparator>
-static CPLErr MinOrMaxPixelFunc(void **papoSources, int nSources, void *pData,
```
dbaston, line 2357: Why doesn't `dfK` come through via `papszArgs` ?

gcore/gdalrasterband.cpp
```diff
@@ -10193,3 +10193,1993 @@ CPLErr GDALRasterBand::SplitRasterIO(GDALRWFlag eRWFlag, int nXOff, int nYOff,
...
+ *
+ * This method is the same as the C function GDALRasterBandAddBand()
+ *
+ * @since 3.12
+ * @throw std::runtime_error if both bands do not have the same dimensions.
+ */
+GDALComputedRasterBand
+GDALRasterBand::operator+(const GDALRasterBand &other) const
+{
+    ThrowIfNotSameDimensions(*this, other);
```
dbaston, line 10232: Is this the first time a public GDAL API throws exceptions?

swig/include/python/gdal_python.i
rouault: The intent here was to match numpy.where()

gcore/gdalrasterband.cpp
rouault: probably, but as we must return an object, I don't see the alternative

frmts/vrt/pixelfunctions.cpp
rouault: it ultimately does through the caller of this function which needs to know if "k" is set to decide if it can take the SSE2 optimized code path

swig/include/python/gdal_python.i
rouault: I've added an alias gdal.Where = gdal.where (and same for minimum/maximum/mean)

gcore/gdalcomputedrasterband.cpp
rouault: good suggestion. Implemented

PR #12523, opened 2025-06-04
Make sure GDALDestroyDriverManage() destroy the TLS cache of GDALThreadSafeDataset (for calling thread)

rouault:


rouault:
Something's wrong on windows

Issue #12524, opened 2025-06-05
Unable to create correct alpha band in GeoPackage using 'gdal raster mosaic'


matsamentet:
### What is the bug?

I have a GeoTIFF RGB raster which uses the color 255,255,252 around the edges (without defining it as nodata in the file). I want to translate this to a GeoPackage with this color transparent using an alpha band. I have been using gdalwarp with the combination of `-srcnodata "255 255 252" -dstalpha` to successfully accomplish this. 

With the new CLI, as far as I can see, the equivalent is` gdal raster mosaic`. I am not able to accomplish the same result however. Instead of creating an alpha band where the specified color is transparent, the color is defined as "Nodata" in the gpkg_metadata table (which does not make sense as RGB GeoPackages does not support the concept on Nodata according to the format documentation ) 

### Steps to reproduce the issue

Old gdalwarp command with correct result:
`gdalwarp -f gpkg -dstalpha -srcnodata "255 255 252" rgb_raster.tif gdalwarp_output.gpkg`

New CLI with incorrect output:
`gdal raster mosaic --of gpkg --add-alpha --src-nodata "255,255,252" --dst-nodata 0,0,0  rgb_raster.tif gdalmosaic_output.gpkg`

### Versions and provenance

Windows 11
GDAL 3.11.0 from conda-forge

rouault:
gdal raster mosaic isn't the way to go for your use case.  You could achieve what you want (ab)using (abuse is similar to using galwarp) with ``gdal raster reproject --add-alpha --src-nodata "255,255,252" --dst-nodata 0,0,0  rgb_raster.tif gdalmosaic_output.gpkg``

We perhaps need a dedicated "gdal raster nodata-to-alpha" command to avoid abusing other commands

matsamentet:
> gdal raster mosaic isn't the way to go for your use case. You could achieve what you want (ab)using (abuse is similar to using galwarp) with `gdal raster reproject --add-alpha --src-nodata "255,255,252" --dst-nodata 0,0,0 rgb_raster.tif gdalmosaic_output.gpkg`
> 
> We perhaps need a dedicated "gdal raster nodata-to-alpha" command to avoid abusing other commands

Understood. A similar usecase for me is having multiple input files, which was also handled by gdalwarp. As far as I can see from the docs, gdal raster reproject only supports single file input. It looks like multiple input files and nodata-to-alpha in the same operation is not currently possible with the new CLI? Perhaps something which could be achieved with piplines ?

PR #12525, opened 2025-06-05
gdal raster mosaic & stack: allow them to be the first step of a raster pipeline

rouault:
Related to #12524


apps/gdalalg_raster_pipeline.cpp
```diff
@@ -557,19 +573,26 @@ bool GDALRasterPipelineAlgorithm::ParseCommandLineArguments(
         return false;
     }
 
-    if (steps.front().alg->GetName() != GDALRasterReadAlgorithm::NAME)
+    if (steps.front().alg->GetName() != GDALRasterReadAlgorithm::NAME &&
```
elpaso: Can't we use a property/method/capability of the alg to determine if it's suitable as first step instead of hardcoding the classes ?

frmts/vrt/vrtsources.cpp
```diff
@@ -188,7 +188,8 @@ CPLErr VRTSimpleSource::FlushCache(bool bAtClosing)
 
 void VRTSimpleSource::UnsetPreservedRelativeFilenames()
 {
-    if (!STARTS_WITH(m_osSourceFileNameOri.c_str(), "http://") &&
+    if (m_bRelativeToVRTOri &&
```
elpaso, line 191: random question: is it necessary to check for other protocols (ftp://) ?

rouault: Seing this originates from c3fb4e938c4ca9352897ec0d504737c140a13ffa  , the test is specific to the HTTP driver that recognizes only ``http://`` or ``https://`, so that seems fine as it

apps/gdalalg_raster_pipeline.cpp
rouault: good idea. I've added a commit adding a CanBeFirstStep() virtual method

elpaso (approved):

PR #12528, opened 2025-06-05
Add np.bool to map of gdal codes

WeatherGod:
Don't promote a boolean array to float64 when trying to write the data

## What does this PR do?
Fix a performance issue when writing out a numpy array of booleans. With this change, saving out a 6000x6000 array to a geotiff is now 66% faster! Memory savings is also useful.

## Tasklist

 - [ ] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [x] Adjust for comments

### DIFF
--- a/autotest/gcore/numpy_rw.py
+++ b/autotest/gcore/numpy_rw.py
@@ -367,7 +367,7 @@ def test_numpy_rw_13():
     ds.GetRasterBand(1).WriteArray(ar)
 
     # Try reading into unsupported array type
-    ar = numpy.empty([1, 2], dtype=numpy.dtype('S1'))
+    ar = numpy.empty([1, 2], dtype=numpy.dtype("S1"))
     with pytest.raises(
         Exception, match="array does not have " "corresponding GDAL data type"
     ):
@@ -435,7 +435,7 @@ def test_numpy_rw_13():
     for i in range(3):
         ds.GetRasterBand(i + 1).WriteArray(ar[i])
 
-    ar = numpy.empty([3, 1, 2], dtype=numpy.dtype('S1'))
+    ar = numpy.empty([3, 1, 2], dtype=numpy.dtype("S1"))
     with pytest.raises(
         Exception, match="array does not have " "corresponding GDAL data type"
     ):
@@ -659,7 +659,7 @@ def test_numpy_rw_16():
     assert ds is None
 
     # Unsupported data type
-    array = numpy.empty([1, 1], numpy.dtype('S1'))
+    array = numpy.empty([1, 1], numpy.dtype("S1"))
     with gdal.quiet_errors():
         ds = gdal_array.OpenArray(array)
     assert ds is None
--- a/autotest/gcore/numpy_rw.py
+++ b/autotest/gcore/numpy_rw.py
@@ -260,6 +260,7 @@ def test_numpy_rw_10_bis(options):
 def test_numpy_rw_11():
 
     type_tuples = [
+        ("bool", gdal.GDT_Byte, numpy.bool_, True),
         ("uint8", gdal.GDT_Byte, numpy.uint8, 255),
         ("uint16", gdal.GDT_UInt16, numpy.uint16, 65535),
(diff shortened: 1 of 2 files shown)


WeatherGod:
The results should be identical, but I haven't figured out how to run the test suite yet. For my software, the output files are identical. Basically, BandWriteArray() would see what gdal datatype code corresponds to a numpy boolean and fail that. So, it thinks it is some exotic type and does an `astype(np.float64)`. In the end, so long as you originally specified GDT_Byte, it'll get written out correctly, but that detour was getting painful in one of my programs that was saving out many hires boolean geotiffs.

rouault:
Error with numpy 1.26:
```
`np.bool` was a deprecated alias for the builtin `bool`. To avoid this error in existing code, use `bool` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.bool_` here.
The aliases was originally deprecated in NumPy 1.20; for more details and guidance see the original release note at:
    https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations. Did you mean: 'bool_'?
    ```

**Comment by WeatherGod** [comment]
Ah, should have realized that you support pre-2.0 numpy. Easy enough to fix.

**Comment by rouault** [comment]
There's a test to adjust:
```
28:     def test_numpy_rw_13():
28:     
28:         drv = gdal.GetDriverByName("MEM")
28:         ds = drv.Create("", 2, 1, 1, gdal.GDT_Byte)
28:         ar = numpy.empty([1, 2], dtype=numpy.uint8)
28:         ar[0][0] = 100
28:         ar[0][1] = 200
28:         ds.GetRasterBand(1).WriteArray(ar)
28:     
28:         # Try reading into unsupported array type
28:         ar = numpy.empty([1, 2], dtype=numpy.bool_)
28: >       with pytest.raises(
28:             Exception, match="array does not have " "corresponding GDAL data type"
28:         ):
28: E       Failed: DID NOT RAISE <class 'Exception'>
28: 
28: D:/a/gdal/gdal/autotest/gcore/numpy_rw.py:370: Failed
```

**Comment by WeatherGod** [comment]
Before I go mucking about in the unit tests, can you think of any reason why `np.bool_` was not supported? Looking elsewhere in gdal_array.i, I see other indications that bool was supposed to be mapped to Byte, so this just seems like an oversight that just simply got codified by time?

**Comment by rouault** [comment]
I believe the intent of the test was to pick a type we didn't support at the time. There's no reason to *not* support bool. We just need to substitue bool in the test by another numpy type we don't support

**Comment by WeatherGod** [comment]
Ok, and I see a few more places where bool_ was used as a placeholder for an unsupported type. I'm going to try replacing them with `np.dtype('S1')`. Might also be worth duplicating some of these uint8 tests with equivalent bool_ tests.

**Comment by WeatherGod** [comment]
Tried as hard as I could, but I could not successfully do a development build of gdal to test, so I'm going to have to rely on the CI/CD here.

**Comment by rouault** [comment]
> Tried as hard as I could, but I could not successfully do a development build of gdal to test

https://gdal.org/en/stable/development/dev_environment.html might help hopefully

**Comment by WeatherGod** [comment]
Yeah, that was the page I found and to be clear, I was able to *build* gdal, but any attempt to import osgeo in python would result in some sort of error depending on my approach for building.

File: autotest/gcore/numpy_rw.py [hunk 1]
```diff
@@ -257,9 +257,9 @@ def test_numpy_rw_10_bis(options):
 # Test all datatypes
 
 
-def test_numpy_rw_11():
-
-    type_tuples = [
+@pytest.mark.parametrize(
+    "name,in_dt,np_dt,val",
+    [
         ("uint8", gdal.GDT_Byte, numpy.uint8, 255),
```
WeatherGod, line 263: Was there a reason why the `np.bool_` test was excluded?

rouault: > Was there a reason why the `np.bool_` test was excluded?

As it was written, it didn't pass, given creating a GDAL raster with GDT_Byte returns uint8, not bool.

